PR #7834, opened 2024-06-24
front: refacto stops table

anisometropie:
Pure refacto, no change in behavior

- moved table columns to a separate file, with the use of `cx` for classes
- keyColumn doesn’t need type params, types is inferred from the type of the array `Column<PathWaypointColumn>[]`
- some simplifications

### DIFF
--- a/front/src/modules/timesStops/TimesStops.tsx
+++ b/front/src/modules/timesStops/TimesStops.tsx
@@ -1,12 +1,6 @@
-import React, { useMemo, useState, useEffect } from 'react';
+import React, { useState, useEffect } from 'react';
 
-import {
-  keyColumn,
-  type Column,
-  checkboxColumn,
-  createTextColumn,
-  DynamicDataSheetGrid,
-} from 'react-datasheet-grid';
+import { DynamicDataSheetGrid } from 'react-datasheet-grid';
 import { useTranslation } from 'react-i18next';
 
 import type { ManageTrainSchedulePathProperties } from 'applications/operationalStudies/types';
@@ -16,9 +10,9 @@ import type { SuggestedOP } from 'modules/trainschedule/components/ManageTrainSc
 import type { PathStep } from 'reducers/osrdconf/types';
 import { useAppDispatch } from 'store';
 import { removeElementAtIndex } from 'utils/array';
+import { marginRegExValidation } from 'utils/physics';
 
-import { marginRegExValidation } from './consts';
-import timeColumn from './TimeColumnComponent';
+import { useInputColumns } from './TimeStopsColumns';
 import type { PathWaypointColumn } from './types';
 import { formatSuggestedViasToRowVias } from './utils';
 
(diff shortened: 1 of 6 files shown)


front/src/modules/timesStops/TimesStops.tsx
```diff
@@ -84,76 +80,12 @@ const TimesStops = ({ pathProperties, pathSteps = [], startTime }: TimesStopsPro
...
-        disabled: ({ rowIndex }) => rowIndex === pathProperties.allWaypoints.length - 1,
-      },
-    ],
+  const columns = useMemo<Column<PathWaypointColumn>[]>(
+    () => inputColumns(t, pathProperties),
     [t, pathProperties.allWaypoints.length]
   );
 
-  const removeVia = (rowData: PathWaypointColumn) => {
+  const removeVia = (rowData: PathWaypointColumn) => () => {
```
SharglutDev: I find it less clear, a function returning a function. Do you think we could leave it like it was ?

theocrsb:
LGTM and tested

front/src/modules/timesStops/TimesStops.tsx
anisometropie: done

front/src/modules/timesStops/TimeColumnComponent.tsx
```diff
@@ -62,4 +70,71 @@ const timeColumn: Partial<Column<string | null | undefined, string, string>> = {
   isCellEmpty: ({ rowData }) => !rowData,
 };
 
+export const inputColumns = (
```
SharglutDev: As the file is called `TimeColumnComponent`, we should either : 
- move `inputColumns` in another file
- rename `TimeColumnComponent` to `TimesStopsColumns` for example

front/src/modules/timesStops/utils.ts
SharglutDev, line 1: It doesn't come from your PR but we have a another bug in the component : 

https://github.com/OpenRailAssociation/osrd/assets/113104137/6349d2d2-f8cf-4b5b-8c5d-8c7b2c57146d

In `formatSuggestedViasToRowVias` function, we don't take into account the updated pathSteps with track/offset (instead of uic) corresponding to the ops added by map clicked. It can be fixed by putting this in the return. Do you want to fix it here or I open a new issue/PR for this ?

```
    if ('track' in op && 'offsetOnTrack' in op) {
      const correspondingOp = pathSteps.find((step) => step.id === op.opId);
      if (correspondingOp) {
        return {
          ...op,
          isMarginValid: op.theoreticalMargin
            ? marginRegExValidation.test(op.theoreticalMargin)
            : true,
          onStopSignal: correspondingOp.onStopSignal || false,
          arrival: i === 0 ? startTime?.substring(11, 19) : correspondingOp.arrival,
          stopFor: correspondingOp.stopFor,
          theoreticalMargin: correspondingOp.theoreticalMargin,
        };
      }
    }
    return {
      ...op,
      name: op.name || t('waypoint', { id: op.opId }),
      isMarginValid: op.theoreticalMargin ? marginRegExValidation.test(op.theoreticalMargin) : true,
      onStopSignal: op.onStopSignal || false,
      arrival: i === 0 ? startTime?.substring(11, 19) : op.arrival,
    };
  });
```

SharglutDev (changes requested):

front/src/modules/timesStops/utils.ts
anisometropie: probably better to fix in another PR

front/src/modules/timesStops/TimeColumnComponent.tsx
anisometropie: done

front/src/modules/timesStops/utils.ts
SharglutDev: Ok I'll do it

SharglutDev (approved):
Lgtm and tested

PR #7838, opened 2024-06-25
front: add suggestions using existing list of tags

### LINKED ISSUE #6876: Infra editor: Add speed tag suggestion in the text input
We should have suggestions using the existing list of tags.

Image

theocrsb:
In the speed limit part in infra, reusing 'SelectImprovedSNCF' present in the 'ElectrificationMetadataForm.tsx' for add suggestion of tags an possibily to create a new tag.

close #6876

### DIFF
--- a/front/public/locales/en/translation.json
+++ b/front/public/locales/en/translation.json
@@ -420,6 +420,7 @@
           "save-speed-section": "Save the speed limit"
         },
         "add-new-speed-limit": "Add new speed limit",
+        "add-new-tag": "Add a new composition code",
         "add-sign": "Add a sign from {{signType}} type",
         "add-track-range": "Click to link to:",
         "additional-speed-limit": "Additional speed limits",
--- a/front/public/locales/fr/translation.json
+++ b/front/public/locales/fr/translation.json
@@ -403,6 +403,7 @@
           "save-speed-section": "Sauvegarder la vitesse limite"
         },
         "add-new-speed-limit": "Ajouter une nouvelle limitation",
+        "add-new-tag": "Ajouter un nouveau code de composition",
         "add-sign": "Ajouter un panneau de type {{signType}}",
         "add-track-range": "Cliquer pour lier à :",
         "additional-speed-limit": "Limitation de vitesse additionnelles",
--- a/front/src/applications/editor/tools/rangeEdition/speedSection/SpeedSectionMetadataForm.tsx
+++ b/front/src/applications/editor/tools/rangeEdition/speedSection/SpeedSectionMetadataForm.tsx
@@ -1,4 +1,4 @@
-import React, { type FC, useContext, useEffect, useMemo, useState } from 'react';
+import React, { type FC, useContext, useEffect, useState } from 'react';
 
 import { cloneDeep, isEmpty, isEqual, map, size } from 'lodash';
 import { useTranslation } from 'react-i18next';
@@ -13,6 +13,7 @@ import type {
   SpeedSectionEntity,
 } from 'applications/editor/tools/rangeEdition/types';
 import type { ExtendedEditorContextType } from 'applications/editor/types';
+import SelectImprovedSNCF from 'common/BootstrapSNCF/SelectImprovedSNCF';
 
 import SpeedInput from './SpeedInput';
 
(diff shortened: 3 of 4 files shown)


front/public/locales/en/translation.json
```diff
@@ -302,7 +302,8 @@
         "empty-linked-track-section": "Click on a track section to link it to the electrification",
         "only-show-n": "Only show the first {{count}} track sections",
         "show-more-ranges_one": "See an additional track section",
-        "show-more-ranges_other": "See {{count}} additional track sections"
+        "show-more-ranges_other": "See {{count}} additional track sections",
+        "add-new-tag": "Add a new composition code"
```
SharglutDev: As this key concerns only speed limits, you should add it in the `speed-edition` section in the translation file.

front/public/locales/fr/translation.json
```diff
@@ -287,7 +287,8 @@
         "empty-linked-track-section": "Cliquez sur un tronçon d'itinéraire pour l'associer",
         "only-show-n": "Ne montrer que les {{count}} premières sections de lignes",
         "show-more-ranges_one": "Voir une section de ligne de plus",
-        "show-more-ranges_other": "Voir {{count}} sections de lignes de plus"
+        "show-more-ranges_other": "Voir {{count}} sections de lignes de plus",
+        "add-new-tag": "Ajouter un nouveau code de composition"
```
SharglutDev: Same

front/src/applications/editor/tools/rangeEdition/speedSection/SpeedSectionMetadataForm.tsx
```diff
@@ -103,24 +97,22 @@ const SpeedSectionMetadataForm: FC = () => {
...
+                    i === currentIndex ? { ...pair, tag: tagSelected } : pair
+                  )
+                );
+              }}
+              withSearch
+              withNewValueInput
+              addButtonTitle={t('Editor.tools.range-edition.add-new-tag')}
+              bgWhite
+            />
+            <div className="d-flex flex-row align-items-center my-1">
```
SharglutDev, line 100: By moving this div here, it now looks like that : 

image

But before it was like that, on one line : 

image

To keep this UI, you could : 
- move this div before the select
- add `...attrs` at the end of SelectImproved props (after l.45)
- change the SelectImprovedSNCF props (l.47) like that : `SelectProps<T> & Omit<React.HTMLAttributes<HTMLDivElement>, 'value' | 'onChange'>)` so that we can add more html attributes with props
- add `{...attrs}` to the first div props in the jsx (l.133)
- pass `className="flex-grow-1 mr-2"` in props to SelectImporvedSNCF in SpeedSectionMetadataForm

SharglutDev (changes requested):

front/src/applications/editor/tools/rangeEdition/speedSection/SpeedSectionMetadataForm.tsx
theocrsb: ok but if we try to pass 'inline' in props of SelectImprovedSNCF the classnames are overwritten.
I put a div around the component

front/public/locales/fr/translation.json
theocrsb: Done

front/public/locales/en/translation.json
theocrsb: Done

front/public/locales/en/translation.json
```diff
@@ -447,7 +447,8 @@
         "remove-track-range": "Click to detatch from:",
         "speed-limits": "Speed limits",
         "toggle-psl": "Permanent speed limit",
-        "toggle-psl-help": "Click on a track section first to be able to activate this option"
+        "toggle-psl-help": "Click on a track section first to be able to activate this option",
+        "add-new-tag": "Add a new composition code"
```
SharglutDev: Keys need to be sorted alphabetically (same for fr file)

SharglutDev (changes requested):

front/public/locales/en/translation.json
theocrsb: Done

SharglutDev (approved):
Lgtm and tested. Good job

PR #7830, opened 2024-06-24
core: use nodes in stdcm priority queue and replace weight by clear comparison

Erashin:
Fixes #7581 and #7582.

PR #7844, opened 2024-06-26
core: stdcm: improve heuristic by using lookahead data

### LINKED ISSUE #7642: core: stdcm: further improve heuristic
This isn't planned soon, it's just to keep track of the possible enhancements if we have some issues with stdcm performances

---

The current stdcm implementation navigates a graph block by block by running simulations and checking conflicts for the current block. The cost of a path is (mostly) the time it gets to reach the last simulated point, and the heuristic estimates the minimum time it takes to get from this point to the destination.

But the stdcm edges actually have some "lookahead", we know in advance what blocks we'll use next. This is a requirement for proper conflict detection. It's something we can use for the heuristic: if the current block is on a normal path but the lookahead goes in the wrong direction, we *could* dismiss it entirely, but we don't.

---

The implementation would be a little more complex. The estimation is currently split in two: the simulated section, and the cached remaining distance. We'd need 3 sections: the simulated section, an estimation of the time it takes to go through the lookahead, and then the cached remaining distance. This isn't entirely trivial to implement and requires some good testing (especially checking that the heuristic is never pessimistic).

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/7642

Work in progress: the tests fail, and it will conflict hard with https://github.com/OpenRailAssociation/osrd/pull/7830 once merged (this PR won't be merged first)

1. Small refactor of the heuristic to wrap in in a class
2. Cache the MRSPs
3. Use lookahead data to correctly identify which edges are going the wrong way
4. Improve how we evaluate the number of passed steps (prerequisite for this to work out)

Benchmark results:

image
Average -10% time.

But there's a few requests that used to cause an error and have been fixed, which increase the average response time (that's the +55s in the first max). Ignoring all errors, the average time goes from 3.79 to 2.58s (32% less time). 

image

### DIFF
--- a/core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/Envelope.java
+++ b/core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/Envelope.java
@@ -186,7 +186,6 @@ public final class Envelope implements Iterable<EnvelopePart>, SearchableEnvelop
     }
 
     private long interpolateUS(double position, boolean isArrivalAt) {
-        assert continuous : "interpolating times on a non continuous envelope is a risky business";
         var envelopePartIndex = findLeft(position);
         assert envelopePartIndex >= 0 : "Trying to interpolate time outside of the envelope";
         var envelopePart = get(envelopePartIndex);
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/STDCMHeuristic.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/STDCMHeuristic.kt
@@ -15,6 +15,7 @@ import fr.sncf.osrd.utils.units.Offset
 import fr.sncf.osrd.utils.units.meters
 import java.util.*
 import kotlin.math.max
+import kotlin.math.min
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 
@@ -33,9 +34,6 @@ typealias STDCMAStarHeuristic = (STDCMEdge, Offset<Block>?, Int) -> Double
  * been reached.
  *
  * Because it's optimistic, we know that we still find the best (fastest) solution.
- *
- * It could eventually be improved further by using lookahead or route data, but this adds a fair
- * amount of complexity to the implementation.
  */
 class STDCMHeuristicBuilder(
     private val blockInfra: BlockInfra,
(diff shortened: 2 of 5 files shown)


core/src/main/kotlin/fr/sncf/osrd/stdcm/STDCMHeuristic.kt
```diff
@@ -51,6 +49,8 @@ class STDCMHeuristicBuilder(
     fun build(): STDCMAStarHeuristic {
         logger.info("Start building STDCM heuristic...")
         // One map per number of reached pathfinding step
+        // maps[n][block] = min time it takes to go from the start of the block to the destination,
+        // if we're at the n-th step of the path
         val maps = mutableListOf<MutableMap<BlockId, Double>>()
```
Erashin: If we need such a big comment, it might be worth to rename this variable more explicitly. `remainingTimeEstimations` like in other parts of the code? Anything clearer really.

core/src/main/kotlin/fr/sncf/osrd/stdcm/STDCMHeuristic.kt
```diff
@@ -74,17 +80,38 @@ class STDCMHeuristicBuilder(
...
+            val allBlocks = mutableListOf(currentBlock)
+            allBlocks.addAll(lookahead)
 
-                return@res remainingTime
-            }
-            return@res Double.POSITIVE_INFINITY
+            // Account for the steps that will be passed in the lookahead
+            val expectedIndex =
+                getExpectedStepIndex(
+                    allBlocks.subList(0, allBlocks.size - 1),
```
Erashin, line 100: Might wanna add a comment to explain why we don't want the last block here as well.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/Envelope.java
```diff
@@ -186,7 +186,6 @@ public double maxSpeedInRange(double beginPos, double endPos) {
     }
 
     private long interpolateUS(double position, boolean isArrivalAt) {
-        assert continuous : "interpolating times on a non continuous envelope is a risky business";
```
Erashin, line 189: Why?

eckter: Interpolating the *speed* would have been risky, but we're interpolating the *time*, which is continuous. It's not differentiable but we don't care about that. 

That assert in this specific method never actually made sense, we just haven't had a need to interpolate times on a non continuous envelope yet. But we do now, we use the MRSP to estimate a fastest time.

Erashin (approved):
GG!

PR #7846, opened 2024-06-26
editoast: project study scenario last update fix

### LINKED ISSUE #7093: last updated time of Project/Study/Scenario is not updated
### What happened?

When I update a project/study/scenario, the `last_modification` field is not changed : it remains equal to the `creation_date `field

I expect the `last_modification` date to be updated to the date and hour when I edit my project/study/scenario

1. Create a project -> creation and modification date are set to the same date
2. Update the project (the name for example) -> modification date is not updated

- **Browser**: Chrome
- **OS**: W11
- **Env**: Staging internal

staging 550477c

Wadjetz:
fix #7093

- [x] Study is already fixed

### DIFF
--- a/editoast/src/views/projects.rs
+++ b/editoast/src/views/projects.rs
@@ -284,6 +284,7 @@ impl From<ProjectPatchForm> for Changeset<Project> {
             .flat_budget(Some(project.budget))
             .flat_image(Some(project.image))
             .flat_tags(project.tags)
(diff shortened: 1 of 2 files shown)


woshilapin (approved):
Nice, thank you 🍻

flomonster (approved):
LGTM

> [!NOTE]
> Not tested (only reviewed)

PR #7849, opened 2024-06-26
gateway: change type of field amr to an Option

younesschrifi:


younesschrifi:
Fix was done in sncf gitlab repo

PR #7851, opened 2024-06-26
front: add e2e tests deleting a rolling stock

### LINKED ISSUE #7774: front: add e2e tests deleting a rolling stock

Maymanaf:
Closes: #7774


Akctarus:
With @RomainValls, we found that the test does not pass 100% of the time, because the "delete" button is not visible if the rolling stock name is too long. The bug is known and is being resolved

Maymanaf:
> With @RomainValls, we found that the test does not pass 100% of the time, because the "delete" button is not visible if the rolling stock name is too long. The bug is known and is being resolved

I added a data-testidto directly locate the buttons, and limited the length of the generated name by UUID to 6 characters, avoiding this problem

Akctarus (approved):
All tests passed, LGTM

SharglutDev (commented):
Keep getting this when trying to : `npx playwright test tests/009-rollingstock-editor.spec.ts --ui --project=chromium`. Am I missing something to launch the test ?

Capture d’écran 2024-07-05 à 08 57 32

front/playwright.config.ts
```diff
@@ -12,7 +12,7 @@ export default defineConfig({
   testDir: './tests',
 
   /* Maximum time one test can run for. */
-  timeout: 50 * 1000,
+  timeout: 60 * 1000,
```
SharglutDev, line 15: Why do you need to increase the timeout timer ? Same for the retries

SharglutDev (commented):
Tested and it works nice, just left one question for the config.

front/playwright.config.ts
Maymanaf: We keep getting warnings that some old tests are slow in the CI.
image
 To avoid any risk of timeout problems and upcoming tests for search, I increased the timeout to 1 minute. For retries, I set it to 3 to anticipate potential failures related to changes in the code for V1/V2

RomainValls (approved):
LGTM

SharglutDev (approved):
Lgtm and tested, good job !

PR #7852, opened 2024-06-26
gateway: fix gateway amr

ghost:
Fixes issue with Gateway default value on amr


ghost:
Should not be required, was an issue locally with something else unrelated.

flomonster (approved):

PR #7857, opened 2024-06-27
front: fix shift train in space time chart

### LINKED ISSUE #7856: Moving trains in space time chart modify schedule points
### What happened?

When a train schedule is drag and drop in the space time chart the scheduled point are changed

Scheduled points should stay the same when a train schedule change its departure time.

1. Create a train schedule with a scheduled point
2. Drag and drop the train schedule
3. The speed of the train changes because the scheduled points are changed

- **Browser**: Firefox v126
- **OS**: Nixos v24.05
- **Env**: All

645314359c47866b68af33235e7104fb9dd386bf

flomonster:
Scheduled points must remain unchanged when shifting a train.

close #7856


SharglutDev (approved):
Lgtm

PR #7864, opened 2024-06-27
editoast: create editoast_models and move model v2 database tools

Wadjetz:


editoast/editoast_models/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod database;
```
flomonster:
```suggestion
pub mod database;

pub use database::DbConnection;
pub use database::DbConnectionPool;
```

flomonster (approved):
LGTM

editoast/editoast_models/src/database/connection_pool.rs
```diff
@@ -106,7 +106,7 @@ impl DbConnectionPoolV2 {
     /// 1. Once a connection is used, it should be dropped **AS SOON AS POSSIBLE**. Failing to do so
     ///    may lead to deadlocks. Example:
     ///
-    /// ```rust
+    /// ```ignore
```
woshilapin: Why do we need to ignore all the doc-tests in this file?

editoast/editoast_models/src/lib.rs
woshilapin: If you expose the `struct`  directly, maybe you don't need to expose the module anymore?

```suggestion
mod database;
```

editoast/editoast_models/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "editoast_models"
+version = "0.1.0"
+edition = "2021"
```
woshilapin, line 4:
```suggestion
edition = "2021"
license = "LGPL-3.0"
```

editoast/src/views/test_app.rs
```diff
@@ -23,7 +25,7 @@ use crate::{
     error::InternalError,
     infra_cache::InfraCache,
     map::MapLayers,
-    modelsv2::{database::connection_pool::create_connection_pool, DbConnectionPoolV2},
+    // modelsv2::{database::connection_pool::create_connection_pool, DbConnectionPoolV2},
```
woshilapin: Can be removed.

woshilapin (changes requested):
Yes, I did it, I reviewed it all 🎉 Thank you for this refactorization. Don't forget the license.

editoast/editoast_models/src/database/connection_pool.rs
Wadjetz: Rust doc tries to compile this code and it causes errors, ignore tell the rust doc to ignore the code inside

woshilapin: OK, I get that, but why did we need to change it then? Was it not launched before? Why was it fine before?

Wadjetz: i think it because we move it to a separate crate (library type crate)

leovalais: Yeah doc tests cannot be run for a bin.

editoast/editoast_models/src/lib.rs
leovalais: +1

leovalais (approved):
Thanks for this big refactoring 😄 
(I don't believe I've checked this many "viewed" since the switch to diesel_async 😅)

I'll summarize here what we discussed yesterday so that we're all on the same page.

### On errors

1. Although the crate `editoast_models` now exists, we still can't move models into it as they all rely on `InternalError`.
2. The first step would be to redesign the failures of `ModelV2`'s API.
3. Then we'd have to figure out a way to forward the information of the errors defined in sub-crates up to `editoast_views` without manually reading all fields of all variants.
4. Only then will we be able to start moving models around.


### On `diesel*` dependencies

The goal is to remove the `diesel` dependencies in all crates but `editoast_models`.

1. We need to finalize the pool V2 and tests migration (which depends on dropping TSV1 ofc 🙃)
2. The `DbConnection` type must be made opaque.
3. All db queries in the views must be moved into the most relevant model.

That's all I believe?

editoast/editoast_models/src/database/connection_pool.rs
woshilapin: So I added functioning doc-tests to the PR in [ca951fa](https://github.com/OpenRailAssociation/osrd/pull/7864/commits/ca951faa07d6faa82c727f0d90fded1b9921bc76). cc @leovalais

woshilapin (approved):

editoast/editoast_models/src/database/connection_pool.rs
leovalais: Even better! Thanks @woshilapin :D

PR #7869, opened 2024-06-27
core: stdcm: minor cleanup + bugfix

eckter:
* Fix formatting of edge / node comments. It got weird when going to java to kotlin, then  worse with the auto-formatter
* Remove unused fields in `STDCMEdge`
* Small bugfix on V1 with time since departure


Erashin (approved):
Ty for the catch, had not thought to run the v1 benchmark on my previous pr.

PR #7828, opened 2024-06-24
security: update rel in release information

bloussou:

PR #7874, opened 2024-06-27
front: use noreferrer for external license links

emersion:
When clicking a link to an external website, the Web browser might send the URL of the OSRD webapp in the external HTTP request. The URL might contain potentially sensitive information such as a private/internal domain name or URL parameters. Make the Web browser always omit this information by using rel="noreferrer".

No need for noopener here like #7828 did, because the link is opened in the same window (we aren't using target="_blank").


Uriel-Sautron (approved):
Thank you for these explanations

theocrsb (approved):

clarani (approved):

PR #7876, opened 2024-06-27
front: adapt updating train tsv2

### LINKED ISSUE #7488: front: adapt updating train v2
Refers to #7023 

### Description and goal

Update a train from the timetable with the new train schedule v2 endpoints

### Acceptance criteria

- [ ] we can still open the train update interface by clicking on the pen icon in the timetable train card
- [ ] when entering the interface, all of the train infos are already displayed (start time, path, margins...)
- [ ] updating a train update it also in the timetable

### Implementation plan
- add a new state value in ScenarioV2 (trainIdToEdit or EditingTrainId ?) to store the id of the train which is currently being edited by the user
- in TimetableTrainCardV2, in editTrainSchedule, remove the updateTrainScheduleIDsToModify dispatch and only setTrainIdToEdit instead
- in ScenarioV2, pass trainIdToEdit to ManageTrainScheduleV2 as a props
- in ManageTrainScheduleV2, pass the trainIdToEdit to the useSetupItineraryForTrainUpdate. This hook will dispatch in the store the correct data so the tabs are correctly pre-filled with the train data

- in the hook, you should:
  - fetch the whole trainSchedule from the backend with its id
  - fetch the other data you need to fill the store:
    - rollingStock (used to fetch the pathBlocks, and to get the data for the powerRestrictions)
    - pathBlocks (used to fetch the pathProperties)
    - pathProperties
  -  fill the following data in the store:
      - rollingStockId & rollingStockComfortV2
      - pathSteps: correspond to the steps of the path of the train. You should merge the data from path, schedule and margins.
      - initialSpeed
      - labels
      - usingElectricalProfiles
      - speedLimitByTag
      - powerRestrictionRanges
      - startTime

achrafmohye:
Closes #7488


emersion:
Seems like the TypeScript build is failing!

```
[front]        |  ERROR(TypeScript)  Property 'ch' does not exist on type '{ secondary_code?: string | null | undefined; uic: number; } & { deleted?: boolean | undefined; id: string; }'.
[front]        |  FILE  /app/src/applications/operationalStudies/hooks.ts:120:94
[front]        | 
[front]        |     118 |                 const correspondingOp = suggestedOperationalPoints.find(
[front]        |     119 |                   (suggestedOp) =>
[front]        |   > 120 |                     'uic' in step && suggestedOp.uic === step.uic && suggestedOp.ch === step.ch
[front]        |         |                                                                                              ^^
[front]        |     121 |                 );
[front]        |     122 |
[front]        |     123 |                 const correspondingSchedule = trainSchedule.schedule?.find(
[front]        | 
[front]        |  ERROR(TypeScript)  Type 'number | undefined' is not assignable to type 'number'.
[front]        |   Type 'undefined' is not assignable to type 'number'.
[front]        |  FILE  /app/src/applications/operationalStudies/views/v2/ScenarioV2.tsx:313:42
[front]        | 
[front]        |     311 |                 displayTrainScheduleManagement === MANAGE_TRAIN_SCHEDULE_TYPES.edit) && (
[front]        |     312 |                 <div className="scenario-managetrainschedule">
[front]        |   > 313 |                   <ManageTrainScheduleV2 trainIdToEdit={trainIdToEdit} />
[front]        |         |                                          ^^^^^^^^^^^^^
[front]        |     314 |                 </div>
[front]        |     315 |               )}
[front]        |     316 |               {displayTrainScheduleManagement === MANAGE_TRAIN_SCHEDULE_TYPES.import && (
[front]        | 
[front]        | [TypeScript] Found 2 errors. Watching for file changes.
```

front/src/modules/trainschedule/components/TimetableV2/TimetableTrainCardV2.tsx
```diff
@@ -56,23 +57,25 @@ const TimetableTrainCardV2 = ({
...
   const dispatch = useAppDispatch();
   const { updateTrainScheduleIDsToModify } = useOsrdConfActions();
-
   const [postTrainSchedule] =
     osrdEditoastApi.endpoints.postV2TimetableByIdTrainSchedule.useMutation();
   const [getTrainSchedule] = osrdEditoastApi.endpoints.postV2TrainSchedule.useLazyQuery();
   const [deleteTrainSchedule] = osrdEditoastApi.endpoints.deleteV2TrainSchedule.useMutation();
 
   const changeSelectedTrainId = (trainId: number) => {
+    setTrainIdToEdit(trainId);
```
emersion: I think this line isn't strictly necessary - `changeSelectedTrainId()` maybe doesn't need to affect any state related to train edition?

front/src/applications/operationalStudies/views/v2/ScenarioV2.tsx
```diff
@@ -69,7 +70,6 @@ const ScenarioV2 = () => {
   const infraId = useInfraID();
   const timetableId = useSelector(getTimetableID);
   const selectedTrainId = useSelector(getSelectedTrainId);
-  const trainIdUsedForProjection = useSelector(getTrainIdUsedForProjection);
```
emersion: Hm, is this unnecessary now? What is the relationship between the train used for projection and the edited train?

front/src/applications/operationalStudies/views/v2/ManageTrainScheduleV2.tsx
```diff
@@ -23,7 +24,11 @@ import SimulationSettings from 'modules/trainschedule/components/ManageTrainSche
 import TrainSettings from 'modules/trainschedule/components/ManageTrainSchedule/TrainSettings';
 import { formatKmValue } from 'utils/strings';
 
-const ManageTrainScheduleV2 = () => {
+type ManageTrainScheduleV2Props = {
+  trainIdToEdit: number;
```
emersion: This should probably be `number | undefined` in the case where `displayTrainScheduleManagement === MANAGE_TRAIN_SCHEDULE_TYPES.add`, I think?

front/src/applications/operationalStudies/hooks.ts
```diff
@@ -26,12 +27,11 @@ import type { ManageTrainSchedulePathProperties } from './types';
  * Hook to relaunch the pathfinding when editing a train
  */
 export const useSetupItineraryForTrainUpdate = (
-  setPathProperties: (pathProperties: ManageTrainSchedulePathProperties) => void
+  setPathProperties: (pathProperties: ManageTrainSchedulePathProperties) => void,
+  trainIdToEdit: number
 ) => {
-  const { getInfraID, getTrainScheduleIDsToModify, getUsingElectricalProfiles } =
```
emersion, line 31: Should we delete `getTrainScheduleIDsToModify` from the codebase now that we're using a separate state for this?

Seems like there is one spot which still uses this in `TimetableV2`, it should probably use the new state instead.

front/src/applications/operationalStudies/views/v2/ManageTrainScheduleV2.tsx
```diff
@@ -55,7 +60,7 @@ const ManageTrainScheduleV2 = () => {
   }, [pathSteps]);
 
   // TODO TS2 : test this hook in simulation results issue
-  // useSetupItineraryForTrainUpdate(setPathProperties);
+  useSetupItineraryForTrainUpdate(setPathProperties, trainIdToEdit);
```
emersion: We should probably guard this behind `if (trainIdToEdit)`?

front/src/applications/operationalStudies/hooks.ts
```diff
@@ -47,21 +47,41 @@ export const useSetupItineraryForTrainUpdate = (
...
+          }
+          if ('operational_point' in step) {
+            return {
+              operational_point: step.operational_point,
+            };
+          }
+          return {
+            trigram: step.trigram,
+          };
+        });
```
emersion: Hm, why is this necessary? It seems like this copies over fields from the input, but without this, the path items would already have the required fields?

front/src/applications/operationalStudies/hooks.ts
```diff
@@ -128,8 +194,8 @@ export const useSetupItineraryForTrainUpdate = (
       }
     };
 
-    if (trainScheduleIDsToModify.length > 0) setupItineraryForTrainUpdate();
-  }, [trainScheduleIDsToModify]);
+    if (trainIdToEdit) setupItineraryForTrainUpdate();
```
emersion: Nit: since `useSetupItineraryForTrainUpdate()` is only called when a train is being updated, we always have a train ID to work with, so I don't think this `if (trainIdToEdit)` is necessary here.

front/src/applications/operationalStudies/hooks.ts
```diff
@@ -88,20 +108,66 @@ export const useSetupItineraryForTrainUpdate = (
...
+                    }
+                  : {
+                      ...step,
+                      positionOnPath: pathfindingResult.path_items_positions[i],
+                      arrival: correspondingSchedule?.arrival,
+                      stopFor: correspondingSchedule?.stop_for,
+                      locked: correspondingSchedule?.locked,
+                      onStopSignal: correspondingSchedule?.on_stop_signal,
+                      coordinates: stepsCoordinates[i],
+                    };
```
emersion: This could probably be de-duplicated with something like this?

```js
const { kp, name, /* etc */ } = correspondingOp || correspondingSchedule || {};
return { kp, name, /* etc */ };
```

front/src/applications/operationalStudies/hooks.ts
```diff
@@ -88,20 +108,66 @@ export const useSetupItineraryForTrainUpdate = (
...
+              });
+
+              const findCorrespondingMargin = (
+                stepId: string,
+                margins: { boundaries: string[]; values: string[] }
+              ) => {
+                const marginIndex = margins.boundaries.findIndex(
+                  (boundaryId) => boundaryId === stepId
+                );
+                return marginIndex !== -1 ? margins.values[marginIndex + 1] : undefined;
```
emersion, line 143: Should this be `[marginIndex]`, or is `[marginIndex + 1]` correct here? Do we want to grab the next margin here?

front/src/applications/operationalStudies/hooks.ts
```diff
@@ -88,20 +108,66 @@ export const useSetupItineraryForTrainUpdate = (
...
+              };
+
+              if (trainSchedule.margins) {
+                updatedPathSteps.forEach((step) => {
+                  const correspondingMargin = findCorrespondingMargin(
```
emersion: Nit: the intermediate variable is unnecessary here, we can assign directly to `step.theoreticalMargin`.

emersion:
It seems like the names of the waypoints are missing when editing:

out

Same for the rolling stock:

out

emersion (commented):
Thanks for working on this, very nice to finally be able to edit train schedules!

Here's a first round of comments. It's the first time I'm reviewing a PR of this size, and I'm not familiar with this part of the code, so let me know if I got anything wrong! A few questions are also included because I didn't understand everything.

front/src/applications/operationalStudies/hooks.ts
SharglutDev: This looks weird but its the way margins work now. See https://osrd.fr/en/docs/reference/design-docs/timetable/#modifiable-fields. We could link this doc in a comment if you think it's necessary.

front/src/modules/trainschedule/components/TimetableV2/TimetableTrainCardV2.tsx
SharglutDev: Agreed, I think the state should change only when we edit the train (what you did on this file l.78). Is there any reason why you did it here as well ?

front/src/applications/operationalStudies/views/v2/ScenarioV2.tsx
SharglutDev: The trainIdUsedForProjection is vital for all the simulation logic, it should'nt be removed in this file and these functions. I think it doesn't have anything to do with the edition of a train. Interested by your answer to @emersion question :)

front/src/applications/operationalStudies/views/v2/ManageTrainScheduleV2.tsx
SharglutDev: And it throws a typescript error in `ScenarioV2.tsx`.

front/src/applications/operationalStudies/views/v2/ManageTrainScheduleV2.tsx
SharglutDev: Regarding this, maybe there should also be a way to `setTrainIdToEdit(undefined)` when leaving the ManageTrainSchedule "edit view" either when we click on the "Retourner aux résultats de simulation" button
Capture d’écran 2024-07-05 à 09 40 06

or clicking outside of the modal (which also close it).

What do you think @emersion ?

front/src/applications/operationalStudies/hooks.ts
SharglutDev: Fyi : `trainScheduleIDsToModify` was originally an array of ids because the endpoint to update trainschedules takes an array of trainschedule ids. Clients asked to be able to edit several trains at ones but the feature has never been implemented. But as it's not in the roadmap, I think it make sense that the new state takes only one id instead of an array. It can be changed later if needed.

front/src/applications/operationalStudies/hooks.ts
SharglutDev: This block of code is used pathfinding/utils.ts but because we need this to increment pathSteps in store which doesn't have exactly the same properties. But as you pointed, `trainSchedule.path` and `PathfindingInputV2['path_items']` have exactly the same format.

front/src/modules/trainschedule/components/TimetableV2/TimetableTrainCardV2.tsx
achrafmohye: Thanks for your feedback :)

front/src/applications/operationalStudies/hooks.ts
achrafmohye: Resolved, done some refacttoring of the code while accommodating the type restrictions

front/src/applications/operationalStudies/hooks.ts
achrafmohye: resolved :)

front/src/applications/operationalStudies/hooks.ts
```diff
@@ -88,20 +108,63 @@ export const useSetupItineraryForTrainUpdate = (
...
                 pathfindingResult.length
               );
 
-              const allWaypoints = upsertViasInOPs(suggestedOperationalPoints, formatedPathSteps);
+              const updatedPathSteps: PathStep[] = trainSchedule.path.map((step, i) => {
+                const correspondingOp = suggestedOperationalPoints.find(
+                  (suggestedOp) =>
+                    'uic' in step &&
+                    suggestedOp.uic === step.uic &&
+                    suggestedOp.ch === step.secondary_code
```
achrafmohye, line 102: is this condition to find the **correspondingOp** seems good ?

front/src/modules/trainschedule/components/TimetableV2/TimetableTrainCardV2.tsx
SharglutDev: Seems like it's still here ?

front/src/applications/operationalStudies/hooks.ts
achrafmohye: When adding new train, it seems to resolve the OD problem, to be confirmed

front/src/applications/operationalStudies/hooks.ts
```diff
@@ -25,13 +26,13 @@ import type { ManageTrainSchedulePathProperties } from './types';
 /**
  * Hook to relaunch the pathfinding when editing a train
  */
+
 export const useSetupItineraryForTrainUpdate = (
-  setPathProperties: (pathProperties: ManageTrainSchedulePathProperties) => void
+  setPathProperties: (pathProperties: ManageTrainSchedulePathProperties) => void,
+  trainIdToEdit: number | undefined
```
SharglutDev: There is a typescript error on l.51 because of this. You can remove the undefined because we call this hook only if trainIdToEdit is defined.

```suggestion
  trainIdToEdit: number
```

front/src/applications/operationalStudies/hooks.ts
```diff
@@ -25,13 +26,13 @@ import type { ManageTrainSchedulePathProperties } from './types';
 /**
  * Hook to relaunch the pathfinding when editing a train
  */
+
 export const useSetupItineraryForTrainUpdate = (
```
SharglutDev: We decided that we prefer each hooks in its one file to improve re-usability. 
Could you : 
- create a `hooks` folder in applications/operationalStudies
- create a `useSetupItineraryForTrainUpdate.ts` file and move the hook inside

front/src/applications/operationalStudies/hooks.ts
```diff
@@ -88,20 +91,63 @@ export const useSetupItineraryForTrainUpdate = (
...
+                  ...step,
+                  kp,
+                  name,
+                  positionOnPath: pathfindingResult.path_items_positions[i],
+                  arrival,
+                  stopFor,
+                  locked,
+                  onStopSignal,
+                  coordinates: stepsCoordinates[i],
+                };
```
SharglutDev: There is an issue here because when you `{...step}` in the return to update `pathSteps`, if the `step` has an `uic` or a `trigram` property, you pass `secondary_code` in `PathStep` instead of `ch`. I think typescript doesn't say anything because `step` doesn't always have a `secondary_code` property so he thinks it's fine. 

Because of that, when you edit a train, `checkCurrentConfig()` update the `secondary_code` of the `path_items` with the `pathSteps` `ch` property but as you passed in it `secondary_code` instead of `ch`, he never finds it and doesn't send any `secondary_code` to the `path_items` (which is optional). So when the edit train interface opens, when you try to associate each `trainSchedule.path` with an existing op in `pathProperties.operational_points`, he doesn't find it because there isn't any `secondary_code` (that's why you said it worked better by just checking the `uic` alone). 

And because of that, `pathSteps` are broken and we don't see the origin/vias/destination names in the ItineraryV2 component.

One way to fix : 

```suggestion
                const { kp, name, ch } = correspondingOp || {};

                const {
                  arrival,
                  stop_for: stopFor,
                  locked,
                  on_stop_signal: onStopSignal,
                } = correspondingSchedule || {};

               const stepWithoutSecondaryCode =
                  'secondary_code' in step ? omit(step, ['secondary_code']) : step;

                return {
                  ...stepWithoutSecondaryCode,
                  ch,
                  kp,
                  name,
                  positionOnPath: pathfindingResult.path_items_positions[i],
                  arrival,
                  stopFor,
                  locked,
                  onStopSignal,
                  coordinates: stepsCoordinates[i],
                } as PathStep;
```

front/src/modules/trainschedule/components/ManageTrainSchedule/TimetableManageTrainScheduleV2.tsx
```diff
@@ -2,48 +2,50 @@ import React, { useState } from 'react';
...
-import { useOsrdConfActions } from 'common/osrdContext';
 import TrainAddingSettings from 'modules/trainschedule/components/ManageTrainSchedule/TrainAddingSettings';
 
 import AddTrainScheduleV2Button from './AddTrainScheduleV2Button';
-import SubmitConfUpdateTrainSchedulesV2 from './SubmitConfUpdateTrainSchedulesV2';
+import UpdateTrainSchedulesV2 from './UpdateTrainSchedulesV2';
 
 type TimetableManageTrainScheduleProps = {
   displayTrainScheduleManagement: string;
+  trainIdToEdit: number | undefined;
```
SharglutDev:
```suggestion
  trainIdToEdit?: number;
```

front/src/modules/trainschedule/components/ManageTrainSchedule/TimetableManageTrainScheduleV2.tsx
```diff
@@ -2,48 +2,50 @@ import React, { useState } from 'react';
...
   setDisplayTrainScheduleManagement: (type: string) => void;
   setTrainResultsToFetch: (trainScheduleIds?: number[]) => void;
   infraState?: InfraState;
+  setTrainIdToEdit: (trainScheduleIds?: number) => void;
```
SharglutDev:
```suggestion
  setTrainIdToEdit: (trainIdToEdit?: number) => void;
```

front/src/modules/trainschedule/components/ManageTrainSchedule/TrainSettings.tsx
```diff
@@ -74,6 +74,7 @@ export default function TrainSettings() {
     setName(nameFromStore);
     setDepartureTime(departureTimeFromStore);
     setInitialSpeed(initialSpeedFromStore);
+    setStartTime(startTimeFromStore.substring(0, 19));
```
SharglutDev, line 77: The startTime stored in the trainschedule is in UTC so we need to convert it. 
Maybe you could create a function `convertUTCToLocalTime()` in `timeManipulation.ts`
`
export function convertIsoUtcToLocalTime(isoDate: string) {
   return dayjs.utc(isoDate).local().format()
}
`

```suggestion
    setStartTime(convertIsoUtcToLocalTime(startTimeFromStore).substring(0, 19));
```

front/src/modules/trainschedule/components/TimetableV2/TimetableV2.tsx
```diff
@@ -102,7 +102,7 @@ const TimetableV2 = ({
 
   // Avoid keeping this on refresh
   useEffect(() => {
-    dispatch(updateTrainScheduleIDsToModify([]));
+    setTrainIdToEdit(undefined);
   }, []);
```
SharglutDev: You can remove this. When we refresh the page, the state is reinitialized as well.

front/src/modules/trainschedule/components/TimetableV2/TimetableV2.tsx
```diff
@@ -125,7 +125,7 @@ const TimetableV2 = ({
           data-testid="scenarios-add-train-schedule-button"
           onClick={() => {
             setDisplayTrainScheduleManagement(MANAGE_TRAIN_SCHEDULE_TYPES.add);
-            dispatch(updateTrainScheduleIDsToModify([]));
+            setTrainIdToEdit(undefined);
```
SharglutDev: You don't need this. We already deal with that when leaving the "edit mode"

```suggestion
```

front/src/modules/trainschedule/components/TimetableV2/TimetableV2.tsx
```diff
@@ -41,13 +42,12 @@ const TimetableV2 = ({
...
   setSpaceTimeData,
+  setTrainIdToEdit,
+  trainIdToEdit,
 }: TimetableV2Props) => {
   const { t } = useTranslation(['operationalStudies/scenario', 'common/itemTypes']);
 
-  const { getTrainScheduleIDsToModify } = useOsrdConfSelectors();
-  const { updateTrainScheduleIDsToModify } = useOsrdConfActions();
   const trainIdUsedForProjection = useSelector(getTrainIdUsedForProjection);
-  const trainScheduleIDsToModify = useSelector(getTrainScheduleIDsToModify);
```
SharglutDev, line 50: Don't forget to remove all `trainScheduleIDsToModify` relations in the store (index, selector, actions, tests)

front/src/reducers/osrdconf/types.ts
```diff
@@ -66,7 +66,7 @@ export interface OsrdStdcmConfState extends OsrdConfState {
   standardStdcmAllowance?: StandardAllowance;
 }
 
-export type PathStep = (
+export type PathItem =
   | TrackOffset
   | {
       operational_point: string;
```
SharglutDev:
```suggestion
export type PathItem = ArrayElement<PathfindingInputV2['path_items']>
```

front/src/modules/trainschedule/components/TimetableV2/TimetableTrainCardV2.tsx
```diff
@@ -56,30 +56,30 @@ const TimetableTrainCardV2 = ({
...
   const dispatch = useAppDispatch();
-  const { updateTrainScheduleIDsToModify } = useOsrdConfActions();
-
   const [postTrainSchedule] =
     osrdEditoastApi.endpoints.postV2TimetableByIdTrainSchedule.useMutation();
   const [getTrainSchedule] = osrdEditoastApi.endpoints.postV2TrainSchedule.useLazyQuery();
   const [deleteTrainSchedule] = osrdEditoastApi.endpoints.deleteV2TrainSchedule.useMutation();
 
-  const changeSelectedTrainId = (trainId: number) => {
+  const changeSelectedTrainId = (trainId: number | undefined) => {
```
SharglutDev:
```suggestion
  const changeSelectedTrainId = (trainId: number) => {
```

front/src/modules/trainschedule/components/TimetableV2/TimetableTrainCardV2.tsx
```diff
@@ -56,30 +56,30 @@ const TimetableTrainCardV2 = ({
...
     dispatch(updateSelectedTrainId(trainId));
   };
 
-  const editTrainSchedule = () => {
-    dispatch(updateTrainScheduleIDsToModify([train.id]));
+  const editTrainSchedule = (trainId: number | undefined) => {
+    setTrainIdToEdit(trainId);
     setDisplayTrainScheduleManagement(MANAGE_TRAIN_SCHEDULE_TYPES.edit);
   };
```
SharglutDev:
```suggestion
  const editTrainSchedule = () => {
    setTrainIdToEdit(train.id);
    setDisplayTrainScheduleManagement(MANAGE_TRAIN_SCHEDULE_TYPES.edit);
  };
```

front/src/modules/trainschedule/components/TimetableV2/TimetableTrainCardV2.tsx
```diff
@@ -56,30 +56,30 @@ const TimetableTrainCardV2 = ({
...
 
   const deleteTrain = async () => {
     if (isSelected) {
       // we need to set selectedTrainId to undefined, otherwise just after the delete,
       // some unvalid rtk calls are dispatched (see rollingstock request in SimulationResults)
+      setTrainIdToEdit(undefined);
```
SharglutDev: Not needed. We deal with that when leaving the edit mode.

```suggestion
```

front/src/modules/trainschedule/components/ManageTrainSchedule/UpdateTrainSchedulesV2.tsx
```diff
@@ -0,0 +1,112 @@
...
+  setIsWorking,
+  setDisplayTrainScheduleManagement,
+  setTrainResultsToFetch,
+  trainIdToEdit,
+  setTrainIdToEdit,
+}: UpdateTrainSchedulesProps) => {
+  const { getConf, getName, getDepartureTime, getStartTime } = useOsrdConfSelectors();
+  const confName = useSelector(getName);
+  const simulationConf = useSelector(getConf);
+  const departureTime = useSelector(getDepartureTime);
```
SharglutDev:
```suggestion
```

front/src/modules/trainschedule/components/ManageTrainSchedule/UpdateTrainSchedulesV2.tsx
```diff
@@ -0,0 +1,112 @@
...
+            callSuccess = false;
+            dispatch(setFailure(castErrorToFailure(e)));
+          });
+
+        if (callSuccess) {
+          setTrainResultsToFetch([trainIdToEdit]);
+          dispatch(
+            setSuccess({
+              title: t('trainUpdated'),
+              text: `${confName}: ${departureTime}`,
```
SharglutDev: Use the new util created in the `TrainSettings.tsx` comment

```suggestion
              text: `${confName}: ${convertIsoUtcToLocalTime(startTimeFromStore).substring(0, 19)}`,
```

front/src/applications/operationalStudies/views/v2/ScenarioV2.tsx
```diff
@@ -46,6 +46,7 @@ const ScenarioV2 = () => {
   const [collapsedTimetable, setCollapsedTimetable] = useState(false);
   const [trainSpaceTimeData, setTrainSpaceTimeData] = useState<TrainSpaceTimeData[]>([]);
   const [trainResultsToFetch, setTrainResultsToFetch] = useState<number[]>();
+  const [trainIdToEdit, setTrainIdToEdit] = useState<number | undefined>();
```
SharglutDev:
```suggestion
  const [trainIdToEdit, setTrainIdToEdit] = useState<number>();
```

front/src/applications/operationalStudies/views/v2/ManageTrainScheduleV2.tsx
```diff
@@ -23,7 +24,11 @@ import SimulationSettings from 'modules/trainschedule/components/ManageTrainSche
 import TrainSettings from 'modules/trainschedule/components/ManageTrainSchedule/TrainSettings';
 import { formatKmValue } from 'utils/strings';
 
-const ManageTrainScheduleV2 = () => {
+type ManageTrainScheduleV2Props = {
+  trainIdToEdit: number | undefined;
```
SharglutDev:
```suggestion
  trainIdToEdit?: number;
```

SharglutDev (changes requested):

front/src/applications/operationalStudies/hooks.ts
SharglutDev: Actually getTrainScheduleIDsToModify is still used in TSV1, so we will remove it totally when dropping v1.

front/src/modules/trainschedule/components/TimetableV2/TimetableV2.tsx
SharglutDev: My bad for that, we still need this prop in TSV1. We will remove it totally when dropping v1.

front/src/applications/operationalStudies/hooks/useSimulationResults.ts
```diff
@@ -0,0 +1,57 @@
+import { useSelector } from 'react-redux';
+
+import { osrdEditoastApi } from 'common/api/osrdEditoastApi';
+import { useInfraID } from 'common/osrdContext';
+import useSpeedSpaceChart from 'modules/simulationResult/components/SpeedSpaceChart/useSpeedSpaceChart';
+import { getSelectedTrainId } from 'reducers/osrdsimulation/selectors';
+
+/**
+ * Hook to relaunch the pathfinding when editing a train
+ */
```
SharglutDev:
```suggestion
```

front/src/modules/trainschedule/components/ManageTrainSchedule/UpdateTrainSchedulesV2.tsx
SharglutDev, line 1: Could you rename this file to `UpdateTrainScheduleV2Button.tsx` (to match the naming of the add button file)

front/src/applications/operationalStudies/views/v2/ManageTrainScheduleV2.tsx
```diff
@@ -55,7 +60,9 @@ const ManageTrainScheduleV2 = () => {
   }, [pathSteps]);
 
   // TODO TS2 : test this hook in simulation results issue
```
SharglutDev, line 69:
```suggestion
```

front/src/modules/trainschedule/components/ManageTrainSchedule/UpdateTrainSchedulesV2.tsx
```diff
@@ -0,0 +1,112 @@
+import React from 'react';
```
kmer2016: I think it would be better to move the `submitConfUpdateTrainSchedules` function to the `ManageTrainSchedule/helper` folder. This will allow us to have a very small component that just returns a button. We could then move it directly into the parent component.

front/src/modules/trainschedule/components/ManageTrainSchedule/UpdateTrainSchedulesV2.tsx
```diff
@@ -0,0 +1,112 @@
...
+        if (callSuccess) {
+          setTrainResultsToFetch([trainIdToEdit]);
+          dispatch(
+            setSuccess({
+              title: t('trainUpdated'),
+              text: `${confName}: ${formatToIsoDate(startTime, true)}`,
+            })
+          );
+          dispatch(updateSelectedTrainId(trainIdToEdit));
+          setIsWorking(false);
```
kmer2016: Call `setIsWorking(false)` a single time in `finally` bloc instead of call in `try` and `catch` blocs.```suggestion

```suggestion
```

front/src/modules/trainschedule/components/ManageTrainSchedule/UpdateTrainSchedulesV2.tsx
```diff
@@ -0,0 +1,112 @@
...
+          setDisplayTrainScheduleManagement(MANAGE_TRAIN_SCHEDULE_TYPES.none);
+          setTrainIdToEdit(undefined);
+        }
+      } catch (e) {
+        setIsWorking(false);
```
kmer2016:
```suggestion
```

front/src/modules/trainschedule/components/ManageTrainSchedule/UpdateTrainSchedulesV2.tsx
```diff
@@ -0,0 +1,112 @@
...
+        dispatch(setFailure(castErrorToFailure(e)));
+      }
```
kmer2016:
```suggestion
      } finally {
        setIsWorking(false);
      }
```

front/src/modules/trainschedule/components/ManageTrainSchedule/TrainSettings.tsx
SharglutDev: Actually we can keep this line like that. But, as the back send us the startTime in UTC, we should convert it to local time when updating the store. So in adjustConfWithTrainToModify, we should do `dispatch(updateStartTime(convertIsoUtcToLocalTime(start_time)));` with the new utils.

kmer2016 (commented):
I think we can create a hook to manage train edition. it can provide : 
- the states variable `trainIdToEdit`/`setTrainIdToEdit`,
- a function to reset it because many componenet need `setTrainIdToEdit` only for set it to `undefined`,
- the function which send update to the backend (`submitConfUpdateTrainSchedules`).
what do you think @SharglutDev
But this can be done in another PR

emersion:
Seems like there are conflicts with the `dev` branch, this PR needs a rebase.

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -88,20 +85,67 @@ export const useSetupItineraryForTrainUpdate = (
...
+                  'secondary_code' in step ? omit(step, ['secondary_code']) : step;
+
+                return {
+                  ...stepWithoutSecondaryCode,
+                  ch,
+                  kp,
+                  name,
+                  positionOnPath: pathfindingResult.path_items_positions[i],
+                  arrival,
+                  stopFor,
```
SharglutDev: As the back send us the duration in ISO8601 (ex: "PTxxS"), we need to convert it back to seconds

```suggestion
                  stopFor: stopFor ? ISO8601Duration2sec(stopFor) : stopFor,
```

front/src/modules/trainschedule/components/TimetableV2/TimetableTrainCardV2.tsx
emersion: Sounds like it's fixed now!

front/src/applications/operationalStudies/views/v2/ScenarioV2.tsx
emersion: This is gone in the last version of this PR.

front/src/applications/operationalStudies/hooks.ts
emersion: Cool, let's keep this for later then!

front/src/applications/operationalStudies/views/v2/ManageTrainScheduleV2.tsx
emersion: Makes sense to me, and I now see the logic for this in `leaveManageTrainSchedule()`, so this is resolved.

front/src/reducers/osrdconf/types.ts
```diff
@@ -66,23 +67,9 @@ export interface OsrdStdcmConfState extends OsrdConfState {
...
-      secondary_code?: string | null;
-      trigram: string;
-    }
-  | {
-      /** An optional secondary code to identify a more specific location */
-      secondary_code?: string | null;
-      /** The [UIC](https://en.wikipedia.org/wiki/List_of_UIC_country_codes) code of an operational point */
-      uic: number;
-    }
-) & {
```
emersion, line 87: Nice cleanup!

front/src/applications/operationalStudies/hooks.ts
emersion: I see!

front/src/applications/operationalStudies/hooks.ts
emersion: Hm, it sounds a bit weird to me to compare a `ch` with a `secondary_code`, but I'm no train expert. Would someone else be able to chime in and tell if it makes sense?

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -88,20 +85,67 @@ export const useSetupItineraryForTrainUpdate = (
...
+
+                const {
+                  arrival,
+                  stop_for: stopFor,
+                  locked,
+                  on_stop_signal: onStopSignal,
+                } = correspondingSchedule || {};
+
+                const stepWithoutSecondaryCode =
+                  'secondary_code' in step ? omit(step, ['secondary_code']) : step;
```
emersion: Nit: I don't think we need to check whether `step` has a `secondary_code` or not here: `omit()` will do nothing if there is no such property.

front/src/modules/trainschedule/components/ManageTrainSchedule/UpdateTrainSchedulesV2.tsx
```diff
@@ -0,0 +1,112 @@
...
+      try {
+        await putV2TrainScheduleById({
+          id: trainIdToEdit,
+          trainScheduleForm: trainSchedule,
+        })
+          .unwrap()
+          .catch((e) => {
+            callSuccess = false;
+            dispatch(setFailure(castErrorToFailure(e)));
+          });
```
emersion: I think we can omit this `catch()` call, and the `try`/`catch` block will take care of handling errors?

front/src/modules/trainschedule/components/ManageTrainSchedule/UpdateTrainSchedulesV2.tsx
achrafmohye: Good idea, like discussed i'll move the file to a hooks folder at the same level as the `helper` folder

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -88,20 +85,67 @@ export const useSetupItineraryForTrainUpdate = (
...
+
+                return {
+                  ...stepWithoutSecondaryCode,
+                  ch,
+                  kp,
+                  name,
+                  positionOnPath: pathfindingResult.path_items_positions[i],
+                  arrival,
```
SharglutDev: For arrival, we get an ISO8601 format (ex: "PT60M" > arrival needs to be 60 minutes after the departure). So as we want to display the arrival, in the timestops table, as a time (ex: "11:35:00"), we need to be able to add the start time and the arrival and format it in timestring.

To do that, you can create a new util (that will be very handy for other computations) : 
`
/**
*Add up a ISO8601 date with another one in local time
*/
export const combineIsoDates(startTime: string, duration: string) => dayjs(startTime)
            .add(ISO8601Duration2sec(duration), 'second')
            .format()
`

```suggestion
                  arrival: arrival ? combineIsoDates(start_time, arrival) : arrival,
```

SharglutDev (changes requested):

emersion:
> It seems like the names of the waypoints are missing when editing

I've tested locally again this PR and this ^ is now fixed.

front/src/applications/operationalStudies/hooks.ts
emersion: This has been removed.

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
emersion: Suggestion for alternative naming: `addDurationToIsoDate`

front/src/applications/operationalStudies/hooks.ts
kmer2016: Yes, that makes sense, secondary_code is a bit of a synonym for ch code. But I can't remember why sometimes we use `secondary_code` and other times `ch` code. Perhaps @SharglutDev  could tell us a bit more?

emersion: I see! Perhaps `ch` is the SNCF-specific name, and `secondary_code` is the generic name?

front/src/utils/date.ts
```diff
@@ -138,3 +140,14 @@ export function convertUTCDateToLocalDate(date: number) {
...
+
+export function convertIsoUtcToLocalTime(isoUtcString: string): string {
+  const date = new Date(isoUtcString);
+  const offset = date.getTimezoneOffset() * 60000;
+  const localTime = new Date(date.getTime() - offset);
+  return localTime.toISOString();
+}
+
+export function addDurationToIsoDate(startTime: string, duration: string) {
+  dayjs(startTime).add(ISO8601Duration2sec(duration), 'second').format();
```
emersion: This might be missing a `return` here?

front/src/utils/date.ts
```diff
@@ -138,3 +140,14 @@ export function convertUTCDateToLocalDate(date: number) {
   const timeDifferenceMinutes = new Date().getTimezoneOffset();
   return Math.abs(timeDifferenceMinutes) * 60 + date;
 }
+
+export function convertIsoUtcToLocalTime(isoUtcString: string): string {
+  const date = new Date(isoUtcString);
+  const offset = date.getTimezoneOffset() * 60000;
+  const localTime = new Date(date.getTime() - offset);
+  return localTime.toISOString();
```
emersion: Hmm. This will return a `Date` with the `Z` suffix, but the time isn't actually in UTC timezone… Is this really what we want here?

Maybe we can use dayjs here instead?

```js
return dayjs(isoUtcString).local().format()
```

front/src/modules/trainschedule/components/ManageTrainSchedule/TrainSettings.tsx
```diff
@@ -32,7 +32,9 @@ export default function TrainSettings() {
   const [name, setName] = useState<string>(nameFromStore);
   // TODO TS2 : remove departureTime when drop v1
   const [departureTime, setDepartureTime] = useState<string>(departureTimeFromStore);
-  const [startTime, setStartTime] = useState(startTimeFromStore.substring(0, 19));
+  const [startTime, setStartTime] = useState(
+    convertIsoUtcToLocalTime(startTimeFromStore).substring(0, 19)
```
emersion: Ah, here is the logic that I assume strips the `Z` suffix? If we keep this approach, we should move the `substring()` call into `convertIsoUtcToLocalTime()` I think?

front/src/utils/date.ts
SharglutDev: agreed

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -88,20 +87,68 @@ export const useSetupItineraryForTrainUpdate = (
...
+                const stepWithoutSecondaryCode = omit(step, ['secondary_code']);
+
+                return {
+                  ...stepWithoutSecondaryCode,
+                  ch,
+                  kp,
+                  name,
+                  positionOnPath: pathfindingResult.path_items_positions[i],
+                  arrival: arrival
+                    ? addDurationToIsoDate(trainSchedule.start_time, arrival)
```
SharglutDev:
```suggestion
                    ? addDurationToIsoDate(trainSchedule.start_time, arrival).substring(11, 19)
```

front/src/applications/operationalStudies/views/v2/ManageTrainScheduleV2.tsx
```diff
@@ -61,6 +66,16 @@ const ManageTrainScheduleV2 = () => {
...
+  // TODO TS2 : test this hook in simulation results issue
+  if (trainIdToEdit) {
+    useSetupItineraryForTrainUpdate(setPathProperties, trainIdToEdit);
+  }
+
+  // const { data: pathWithElectrifications = { electrification_ranges: [] as RangedValue[] } } =
+  //   osrdEditoastApi.endpoints.getPathfindingByPathfindingIdElectrifications.useQuery(
+  //     { pathfindingId: pathFindingID as number },
+  //     { skip: !pathFindingID }
+  //   );
```
emersion: Remnant from a debugging session? :P 

Or is this for a future feature?

front/src/applications/operationalStudies/hooks.ts
kmer2016: it seems to me that's right

front/src/modules/trainschedule/components/ManageTrainSchedule/TrainSettings.tsx
emersion: Note to self: we still need the `substring()` even with dayjs, to strip the timezone suffix (`+02:00`).

emersion (approved):
LGTM, nice work!

front/src/applications/operationalStudies/hooks.ts
SharglutDev: I think the back called this prop `secondary_code` because we might not always put the ch code in it but another one sometimes ? (but we chosed to use only the ch for now)

SharglutDev (changes requested):

front/src/applications/operationalStudies/views/v2/ManageTrainScheduleV2.tsx
SharglutDev: Yep it should be removed

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/adjustConfWithTrainToModify.ts
```diff
@@ -144,27 +146,21 @@ export function adjustConfWithTrainToModifyV2(
...
-  dispatch(updateDepartureTime(start_time));
+  dispatch(updateStartTime(convertIsoUtcToLocalTime(start_time)));
   dispatch(updatePathSteps({ pathSteps, resetPowerRestrictions: true }));
   dispatch(updateInitialSpeed(initial_speed ? msToKmh(initial_speed) : 0));
 
-  if (options?.use_electrical_profiles === usingElectricalProfiles)
+  if (
+    options?.use_electrical_profiles !== undefined &&
+    options.use_electrical_profiles !== usingElectricalProfiles
+  )
```
SharglutDev:
```suggestion
  const trainUsingElectricalProfiles = options?.use_electrical_profiles ?? true
  if (
    trainUsingElectricalProfiles !== usingElectricalProfiles
  )
```

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/adjustConfWithTrainToModify.ts
```diff
@@ -144,27 +146,21 @@ export function adjustConfWithTrainToModifyV2(
...
     dispatch(toggleUsingElectricalProfiles());
 
-  if (speed_limit_tag) dispatch(updateSpeedLimitByTag(speed_limit_tag));
+  dispatch(updateSpeedLimitByTag(speed_limit_tag));
 
-  if (labels) dispatch(updateLabels(labels));
+  dispatch(updateLabels(labels || []));
 
   if (power_restrictions) {
```
SharglutDev:
```suggestion
```

SharglutDev (changes requested):

SharglutDev (approved):
Lgtm and tested, good job !

PR #7881, opened 2024-06-28
Fix large uri for batch retrieve train schedules

### LINKED ISSUE #7860: Opening a big tsv2 timetable crashes - URI Too large
### What happened?

Opening this scenario with 1900+ trains lead to a page with no train:

image

image

After a while the conflicts appeared:

image

I'm expecting the app to survive, even if the space time chart is known broken for the moment.

1. Go to a big tsv2 timetable

- **Browser**: Chrome
- **OS**: W11
- **Env**: rec

d9655c0

flomonster:
close #7860

Browser have a URI length limit:

- Microsoft Internet Explorer: 2,083 characters
- Microsoft Edge: 2,083 characters
- Google Chrome: 32,779 characters
- Mozilla Firefox: more than 64,000 characters
- Apple Safari: more than 64,000 characters
- Google Android: 8,192 characters

The idea is to move query parameters to the body of the request.

> [!NOTE]
> This implies moving `GET` endpoint to `POST`.

Which endpoint is involved:

- [x] Get batch train schedules
- [x] Simulation summary
- [x] Project path

### DIFF
--- a/editoast/src/views/v2/train_schedule.rs
+++ b/editoast/src/views/v2/train_schedule.rs
@@ -9,15 +9,12 @@ use std::ops::DerefMut;
 use std::sync::Arc;
 
 use actix_web::web::{Data, Json, Path, Query};
-use actix_web::HttpRequest;
-use actix_web::{delete, get, put, HttpResponse};
+use actix_web::{delete, get, post, put, HttpResponse};
 use editoast_derive::EditoastError;
 use editoast_schemas::train_schedule::TrainScheduleBase;
 use itertools::Itertools;
 use serde::Deserialize;
 use serde::Serialize;
-use serde_qs::actix::QsQuery;
-use serde_qs::Config;
 use thiserror::Error;
 use tracing::info;
 use utoipa::IntoParams;
@@ -75,7 +72,6 @@ editoast_common::schemas! {
     TrainScheduleBase,
     TrainScheduleForm,
     TrainScheduleResult,
-    BatchDeletionRequest,
     SimulationSummaryResult,
     InfraIdQueryParam,
     projection::schemas(),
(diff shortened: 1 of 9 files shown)


editoast/src/views/v2/train_schedule.rs
```diff
@@ -486,10 +471,10 @@ fn train_simulation_input_hash(
     format!("simulation_{osrd_version}.{infra_id}.{infra_version}.{hash_simulation_input}")
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, IntoParams)]
-struct SimulationBatchParams {
+#[derive(Debug, Clone, Deserialize, ToSchema)]
+struct SimulationBatchForm {
     infra: i64,
```
woshilapin: I believe we try to have always `infra_id` for naming an identifier (else, we don't know if it's an `Infra`).
```suggestion
    infra_id: i64,
```

woshilapin (commented):
Need to update all the tests, but looks good so far 🎉

SharglutDev (approved):
Lgtm and tested for front. Nice PR !

editoast/src/views/v2/train_schedule.rs
```diff
@@ -198,29 +186,26 @@ async fn get(
...
     responses(
         (status = 200, description = "Retrieve a list of train schedule", body = Vec<TrainScheduleResult>)
     )
 )]
-#[get("")]
+#[post("")]
 async fn get_batch(
     db_pool: Data<DbConnectionPoolV2>,
-    params: QsQuery<GetBatchParams>,
+    data: Json<BatchRequest>,
```
leovalais, line 205: Do we still use QsQuery somewhere? If not `serde_qs` could be removed from `Cargo.toml`.

leovalais (approved):
LGTM, not tested

editoast/src/views/v2/train_schedule.rs
flomonster: We still use it in the path properties endpoint. We have a (limited) list of properties.

hamz2a (approved):
LGTM.

editoast/src/views/v2/train_schedule.rs
leovalais: Since it's the only usage we could probably add the props to the payload to remove the paradigm and the dependency altogether wdyt?

PR #7896, opened 2024-07-01
front: fix "invalid Date" errors

### LINKED ISSUE #7836: "Invalid date" displayed in operational studies
### What happened?

When creating a project / study / scenario the front displays "updated on invalid date" / "modifiée le invalid date". 

See:
image

I would expect it to display a normal date, or no date at all

1. Create any project, study or scenario ?

- **Browser**: Firefox
- **OS**: Windows 11

24485e2

emersion:
_See individual commits._

Closes: https://github.com/OpenRailAssociation/osrd/issues/7836

### DIFF
--- a/front/src/utils/date.ts
+++ b/front/src/utils/date.ts
@@ -5,7 +5,6 @@ import utc from 'dayjs/plugin/utc';
 
 import i18n from 'i18n';
 
-dayjs.locale('fr');
 dayjs.extend(utc);
 dayjs.extend(timezone);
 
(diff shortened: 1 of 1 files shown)


theocrsb:
LGTM

front/src/utils/date.ts
```diff
@@ -19,16 +18,19 @@ export function formatIsoDate(date: Date) {
...
-      dayjs.locale('fr');
+      locale = 'fr';
       break;
     default:
-      dayjs.locale('en-gb');
+      locale = 'en';
   }
-  const dateToUTC = dayjs(`${date}Z`); // The 'Z' is to ensure we have an UTC date
+  // Force interpreting the date in UTC
+  const dateToUTC = dayjs(date).utc(true);
```
emersion, line 30: Maybe we should just drop the UTC conversion altogether here? (In practice browsers ignore the `Z` suffix, so dropping the UTC conversion would retain current behavior. I tried to keep the UTC conversion to retain the original intent of the old code, but maybe that's not a good idea.)

Math-R (approved):
tested and seems fine to me

front/src/utils/date.ts
Math-R: we can discuss that if you want but it's seems good to me

emersion: Alright, let's leave it as-is for now. Would be nice to audit all of the call-sites at some point to find out if any `Date` values really need this.

theocrsb (approved):

PR #7897, opened 2024-07-01
 front: reference map when no infra (or not existing) is selected

### LINKED ISSUE #6541: front: irrelevant infra retrieval error popup when empty database
### What happened?

When you open reference map with no infra in database, an error popup fires and says: "Unable to retrieve infra".
This is normal: there's no infra at all!

This is not an error, there should be no error messages.

1. use OSRD without any data
2. go to reference map
3. see the popup
4. be completely stunned

- **Browser**: Firefox v123.0b3
- **OS**: Archlinux latest
- **Env**: Local

e220992

sim51:
Fix #6541

 - map layers : don't display them if the infra is undefined (avoid 404)
 - InfraSelector : remove the code that check that the current infra exist (not it's role and if the infra doesn't exist display an error toast)
 - map reference: if the current infra doesn't exist, then we set it as undefined in the state. It avoid to have "no selected infra" in the header whereas in reality we have one in the store


SharglutDev (approved):
Lgtm and tested !

PR #7898, opened 2024-07-01
Add codegen filter to ModelV2

leovalais:
Adds an option to parametrize which implementation ModelV2 has to generate. This is required to support single PK field models (such as `Timetable`).

No impact on compilation times.

refs #8111

### DIFF
--- a/editoast/editoast_derive/src/modelv2/args.rs
+++ b/editoast/editoast_derive/src/modelv2/args.rs
@@ -1,6 +1,6 @@
 #![allow(clippy::manual_unwrap_or_default)]
 
-use super::RawIdentifier;
+use super::{crud::Crud, RawIdentifier};
 use darling::{
     ast,
     util::{self, PathList},
@@ -22,12 +22,24 @@ pub(super) struct ModelArgs {
     pub(super) changeset: GeneratedTypeArgs,
     #[darling(multiple, rename = "identifier")]
     pub(super) identifiers: Vec<RawIdentifier>,
+    #[darling(rename = "gen")]
+    pub(super) impl_plan: ImplPlan,
     #[darling(default)]
     pub(super) preferred: Option<RawIdentifier>,
     #[darling(default)]
     pub(super) batch_chunk_size_limit: Option<usize>,
 
     pub(super) data: ast::Data<util::Ignored, ModelFieldArgs>,
+}
+
+#[derive(Debug, PartialEq, Eq, FromMeta)]
+pub(super) struct ImplPlan {
+    #[darling(default)]
+    pub(super) ops: Crud,
+    #[darling(default)]
+    pub(super) batch_ops: Crud,
+    #[darling(default)]
+    pub(super) list: bool,
 }
 
 #[derive(FromMeta, Default, Debug, PartialEq)]
--- a/editoast/editoast_derive/src/modelv2/codegen.rs
+++ b/editoast/editoast_derive/src/modelv2/codegen.rs
@@ -22,6 +22,7 @@ mod row_decl;
 mod update_batch_impl;
 mod update_impl;
 
+use quote::ToTokens;
 use syn::parse_quote;
 
 use self::changeset_builder_impl_block::BuilderType;
@@ -52,6 +53,7 @@ use self::row_decl::RowFieldDecl;
 use self::update_batch_impl::UpdateBatchImpl;
 use self::update_impl::UpdateImpl;
 
+use super::args::ImplPlan;
 use super::identifier::Identifier;
 use super::utils::np;
 use super::ModelConfig;
@@ -111,6 +113,24 @@ impl Identifier {
     }
 }
 
+impl ImplPlan {
+    fn has_read(&self) -> bool {
+        self.ops.read || self.batch_ops.read
+    }
+
+    fn has_create(&self) -> bool {
+        self.ops.create || self.batch_ops.create
+    }
+
+    fn has_update(&self) -> bool {
+        self.ops.update || self.batch_ops.update
+    }
+
+    fn has_upsert(&self) -> bool {
+        self.has_update() || self.has_create()
+    }
+}
+
 impl ModelConfig {
     pub(crate) fn model_impl(&self) -> ModelImpl {
         ModelImpl {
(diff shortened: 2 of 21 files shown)


editoast/editoast_derive/src/modelv2/args.rs
```diff
@@ -30,6 +32,26 @@ pub(super) struct ModelArgs {
...
+
+impl Default for ImplPlan {
+    fn default() -> Self {
+        ImplPlan {
+            ops: Crud::ALL,
+            batch_ops: Crud::ALL,
+            list: true,
+        }
+    }
+}
```
flomonster: We could use `Derivative`.

flomonster (approved):
LGTM, the fact that we haven't seen any improvement in compilation time seems to indicate that ModelV2 may not be the source of the problem.

editoast/editoast_derive/src/modelv2/args.rs
```diff
@@ -30,6 +33,20 @@ pub(super) struct ModelArgs {
...
 
+#[derive(Debug, PartialEq, Eq, Derivative, FromMeta)]
+#[derivative(Default)]
+pub(super) struct ImplPlan {
+    #[darling(default)]
+    #[derivative(Default(value = "Crud::ALL"))]
+    pub(super) ops: Crud,
+    #[darling(default)]
+    #[derivative(Default(value = "Crud::ALL"))]
+    pub(super) batch_ops: Crud,
```
Castavo: Shouldn't the default be some kind of `Crud:NONE` ? 
From the usage on the models, it feels like what you intended.

Indeed, you specify `ops = crud` quite often, but not `batch_ops = crud`, instead `batch_ops` isn't mentioned

editoast/editoast_derive/src/modelv2/codegen.rs
```diff
@@ -111,6 +113,24 @@ impl Identifier {
     }
 }
 
+impl ImplPlan {
```
Castavo, line 116: Why is this not in `args.rs` ?

editoast/editoast_derive/src/modelv2/args.rs
Castavo: This comment is also true for `list`

leovalais: There are two different "default values" involved here.

1. The ones specified in `derivative` are used when there's no `ImplPlan` to parse, i.e. without `#[model(gen(...))]`. In this case, we implement all traits available (how it worked before this feature), hence the `Crud::ALL`.
2. When there's a `#[model(gen(...))]` annotation but some fields are omitted however, the `#[darling(default)]` applies. For example, `#[model(gen(ops = crud))]` will generate an `ImplPlan { ops: Crud { <crud>: true}, batch_ops: Crud { <crud>: false }, list: false }` as the `Default` of `Crud` sets everything to `false` (cf. `#[derive(Default)] struct Crud`).

editoast/editoast_derive/src/modelv2/codegen.rs
leovalais: Because these functions are only useful in this module. I could have written `fn has_read(plan: &ImplPlan) -> bool` but I think it makes sense for these to be methods instead. Wdyt?

If they were defined in `args.rs`, they would need to be `pub(super)` at least, giving them more visibility than necessary.

editoast/editoast_derive/src/modelv2/args.rs
Castavo: Wow okay, that was a skill issue on my part
That's quite convoluted

editoast/src/modelsv2/documents.rs
```diff
@@ -6,6 +6,7 @@ use editoast_derive::ModelV2;
 
 #[derive(Debug, Default, Clone, ModelV2)]
 #[model(table = editoast_models::tables::document)]
+#[model(gen(ops = crud))]
```
Castavo:
```suggestion
#[model(gen(ops = crd))]
```

editoast/editoast_derive/src/modelv2/args.rs
Castavo: It feels quite counterintuitive that the less you write in `gen` the less code is added, but not writing `gen` at all writes the most code

This might be nitpicking, resolve this thread if you think it is

Castavo (approved):

editoast/editoast_derive/src/modelv2/args.rs
leovalais: I see what you mean, but if not writing `gen` at all generates no code, you end up in a situation where `derive(ModelV2)` works without `gen` configuration but doesn't generate anything useful. It'd be like if `#[derive(Deserialize)]` doesn't make your type deserializable unless you add an extra `#[serde(please_do_gen_deserialize)]`.

Maybe we should make `gen` required to eliminate this behavior. Or maybe `gen` isn't the best naming for that. Not sure what to do.

leovalais: I've also added https://github.com/OpenRailAssociation/osrd/pull/7898#discussion_r1742236086 in a comment as the behavior isn't exactly straightforward.

flomonster: Actually, making `gen` mandatory isn't so bad. Besides the fact that it's more intiutive in terms of code generation (identified by @Castavo ). It forces the API user to ask what he really needs.
We could check if `ops` and `ops_batch` are both empty and throw a compilation error with a meaningful message.

woshilapin (approved):
Looks ok but as stated by @flomonster, if we don't see any compilation improvement, do we need the additional complexity?

leovalais:
@woshilapin I drafted this PR a while ago hoping it would help with compilation times. It didn't help so I abandoned the idea. However @Castavo noticed a problem with models having only a single primary key (such as `Timetable` 👀) for which the `ModelV2` traits are currently implemented by hand. Being able to select which operations we choose to generate will be necessary to support this case => https://github.com/OpenRailAssociation/osrd/issues/8111. Whether we want to provide a user-facing api for `ImplPlan` is another debatable question, but we need the internals.

editoast/editoast_derive/src/modelv2/args.rs
leovalais: Done. I'll let you resolve if you're ok with the change. Auto-merge is on.

PR #7899, opened 2024-07-01
front: avoid x scroll on map search

### LINKED ISSUE #7185: Unnecessary horizontal scroll in map search
### What happened?

image

The horizontal scroll shouldn't be there as long as there's some room to display the PR names.

1. Go to any instance of the map
2. Search for a PR

- **Browser**: Chrome
- **OS**: W11
- **Env**: rec

a42cad3

sim51:
Fix #7185

Using flex with grow/shrink instead of width 100% and specify x-overflow hidden (even if now the case should not be possible).
Now it works on both firefox & chrome

### DIFF
--- a/front/src/styles/scss/common/map/_mapSearch.scss
+++ b/front/src/styles/scss/common/map/_mapSearch.scss
@@ -4,25 +4,34 @@
   }
   .search-results {
     height: 30vh;
-    overflow: auto;
+    display: flex;
+    overflow-y: auto;
+    overflow-x: hidden;
     padding-right: 0.5rem;
+    width: 100%;
+    flex-direction: column;
+
     .search-result-item {
       all: unset;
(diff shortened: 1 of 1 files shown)


SharglutDev (commented):
Thank you for the fix ! Do we need width/flex grow/shrink ? It seems it's working without

sim51:
Have you tested on chrome ? 
The initial bug is only visible on chrome not firefox.

To avoid scroll and use the full size of a div, I generally use flex, grow & shrink pattern. 
It works fine even in responsive.

SharglutDev:
Yes even on chrome, it works without the grow/shrink/width:100%. But if you think it's gonna be more resilient with these properties, its fine by me.

SharglutDev (approved):
Lgtm and tested

PR #7903, opened 2024-07-01
core: fallback for speed limit tags

### LINKED ISSUE #7715: Speed-limit tag fallback in simulation
First part of https://github.com/osrd-project/osrd-confidential/issues/364

Use the correct speed-limit tag for fallback in train simulation speed limit (in case the actual speed-limit tag is not available on track section).
Just output simulation warnings if any (or logs).

bougue-pe:
Only handling the speed for now, no work on output, warnings and logs.
Extra-code to handle long-names currently used on speedLimitTag (both in infra or in requested simu).

also: little fix on `Speed.toString()`

Test:
- ✅ hand-tested locally that fallback works (for long-names) on small-infra: adding a speed section on `Messagerie - ME100` somewhere else than South-East(to be able to select it), then using that speedLimitTag on a train to trigger fallback to MA100 on South-East.

Fix #7715

### DIFF
--- a/core/kt-osrd-rjs-parser/src/main/kotlin/fr/sncf/osrd/RawInfraRJSParser.kt
+++ b/core/kt-osrd-rjs-parser/src/main/kotlin/fr/sncf/osrd/RawInfraRJSParser.kt
@@ -1,11 +1,16 @@
 package fr.sncf.osrd
 
+import com.charleskorn.kaml.Yaml
 import com.google.common.primitives.Doubles
 import fr.sncf.osrd.geom.LineString
 import fr.sncf.osrd.railjson.schema.common.graph.EdgeDirection
 import fr.sncf.osrd.railjson.schema.common.graph.EdgeEndpoint
 import fr.sncf.osrd.railjson.schema.geom.RJSLineString
-import fr.sncf.osrd.railjson.schema.infra.*
+import fr.sncf.osrd.railjson.schema.infra.RJSInfra
+import fr.sncf.osrd.railjson.schema.infra.RJSRoute
+import fr.sncf.osrd.railjson.schema.infra.RJSSwitch
+import fr.sncf.osrd.railjson.schema.infra.RJSSwitchType
+import fr.sncf.osrd.railjson.schema.infra.RJSTrackSection
 import fr.sncf.osrd.railjson.schema.infra.trackobjects.RJSRouteWaypoint
 import fr.sncf.osrd.railjson.schema.infra.trackobjects.RJSSignal
 import fr.sncf.osrd.railjson.schema.infra.trackranges.RJSElectrification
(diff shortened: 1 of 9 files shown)


core/kt-osrd-rjs-parser/src/main/resources/speed_limit_tags.yml
```diff
@@ -0,0 +1,224 @@
+V200:
```
Erashin, line 1: Checked all of the data, everything checks out (gosh that was long).
Always wondered where 8.33 default speed came from in a couple of tests, is it officially all of the rolling stocks' default speed (wasn't in the ticket linked to the pr)?
EDIT: ah yeah it's 30kjm/h in m/s, nvm :)

core/kt-osrd-rjs-parser/src/main/kotlin/fr/sncf/osrd/RawInfraRJSParser.kt
```diff
@@ -821,5 +900,7 @@ fun parseRJSInfra(rjsInfra: RJSInfra): RawInfra {
         parseSignal(builder, rjsSignal)
     }
 
+    parseSpeedLimitTags(builder, "speed_limit_tags.yml")
```
Erashin: Feels like the file's `resourcePath` shouldn't be a method parameter, or should at least be a static value, it should never change.

core/src/test/kotlin/fr/sncf/osrd/sim_infra_adapter/PathPropertiesTests.kt
```diff
@@ -477,6 +485,124 @@ class PathPropertiesTests {
...
+                blocks.last(),
+                routes = listOf("rt.DH2->buffer_stop.7")
+            )
+
+        /*
+        On the last block, speed limit sections from small infra are
+
+        0m       100m      600m     1000m     1500m     1600m
+        |=========|=========|=========|=========|=========|
+        | 39.444  |      31.111       | 83.333  |   3.0   |     no-tag (to be respected any time)
```
Erashin, line 523:
```suggestion
        | 39.444  |      31.111       | 8.333  |   3.0   |     no-tag (to be respected any time)
```

core/kt-osrd-rjs-parser/src/main/resources/speed_limit_tags.yml
bougue-pe: It was indeed missing from the issue linked to the PR (added it just now): it's 30 km/h that we are using as a default for now for every speedLimitTag.

core/src/test/kotlin/fr/sncf/osrd/sim_infra_adapter/PathPropertiesTests.kt
bougue-pe: This is correct I think: It's the default speed limit of the section, and it's 300 km/h as seen [here](https://github.com/OpenRailAssociation/osrd/blob/dd067f8d106b7006fbf5932a90676cc28ddbd718/tests/data/infras/small_infra/infra.json#L2928) (and not overridden by any other speed section).

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/RawInfraImpl.kt
```diff
@@ -511,9 +577,46 @@ class RawInfraImpl(
...
+                            ?: speedSection.speedByTrainTag[speedLimitTagPool[fallbackTagId]?.name]
+
+                    if (fallbackSpeed != null) {
+                        if (infraTagSpeed == null || fallbackSpeed > infraTagSpeed) {
+                            infraTagSpeed = fallbackSpeed
+                            infraTagId = fallbackTagId
+                        }
+                    }
+                }
+            }
```
Castavo: This feels large enough to become a separate method.
Maybe it could be a method of `SpeedSection` or of `SpeedLimitTagDescriptor`

core/kt-osrd-rjs-parser/src/main/kotlin/fr/sncf/osrd/RawInfraRJSParser.kt
Castavo: I agree that if you're not using this parameter elsewhere (like in a test 👀), it should be removed.
And if it is used elsewhere I agree it should be a const

Castavo (commented):
Cool
Don't we want to check whether tags inside a fallback list are present inside the config file ?

core/kt-osrd-rjs-parser/src/main/kotlin/fr/sncf/osrd/RawInfraRJSParser.kt
bougue-pe: Done https://github.com/OpenRailAssociation/osrd/pull/7903/commits/8c7694d388f024f624de3db69c97ce03a8527567#diff-9325fe5faad772881e72d6024d73bade0606de7420d369192dd9486480914c48

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/RawInfraImpl.kt
bougue-pe: Done: method separated in https://github.com/OpenRailAssociation/osrd/pull/7903/commits/8c7694d388f024f624de3db69c97ce03a8527567#diff-5d11fb62d051b79ae994acaff5b58f193131909011f76ad3547e4c777e4dfc85R637
Although IMO it should be a method of `SpeedLimitTagDescriptor`, but currently using `speedLimitTagPool` from `RawInfraImpl`, added a [TODO](https://github.com/OpenRailAssociation/osrd/pull/7903/commits/8c7694d388f024f624de3db69c97ce03a8527567#diff-5d11fb62d051b79ae994acaff5b58f193131909011f76ad3547e4c777e4dfc85R656-R657) on that.

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/RawInfraImpl.kt
```diff
@@ -276,6 +342,13 @@ class RawInfraImpl(
     }
 
     init {
+        // check that SpeedLimitTag fallbacks are present in the list of SpeedLimitTag
+        for (tag in speedLimitTagPool.values) {
+            for (fallbackTag in tag.fallbackList) {
+                assert(speedLimitTagPool.containsKey(fallbackTag))
+            }
+        }
```
bougue-pe, line 350: > @Castavo:
> Don't we want to check whether tags inside a fallback list are present inside the config file ?

👍 Good idea, added the check

core/kt-osrd-rjs-parser/src/main/kotlin/fr/sncf/osrd/RawInfraRJSParser.kt
```diff
@@ -83,6 +83,7 @@ import kotlinx.serialization.builtins.serializer
 import mu.KotlinLogging
 
 val logger = KotlinLogging.logger {}
+val SPEED_LIMIT_TAGS_RESOURCE_PATH = "speed_limit_tags.yml"
```
Erashin: I don't know if we want this to be available in the package (for now only one class in it, later on maybe more) and all the packages using `kt-osrd-rjs-parser`. Maybe change this to:
`private const val SPEE...`?

Erashin (approved):

core/kt-osrd-rjs-parser/src/main/kotlin/fr/sncf/osrd/RawInfraRJSParser.kt
bougue-pe: Done, thanks (amended fixup commit): https://github.com/OpenRailAssociation/osrd/compare/8c7694d388f024f624de3db69c97ce03a8527567..b9073205c73c4dbf5a8573c8fef01d08accb99a0

Castavo (approved):
LGTM
Did not test though

PR #7906, opened 2024-07-01
Isolate editoast search engine into a dedicated crate

leovalais:
Since the incremental computing prototype is postponed, we might as well do this now.

- Moves everything in `editoast_search` expect the endpoint.
- Adapt paths generated by the `Search` derive macro.
- Changes functions returning an editoast custom `Result` to return
    a standard one instead.
- Forwards search engine errors up to the API level, with updated i18n
  keys.
- Merges `views/search/mod.rs` and `views/search/objects.rs` into
  `views/search.rs`.


kmer2016 (approved):
lgtm

flomonster (approved):
LGTM (reviewed + tested).
Sorry for the delay.

PR #7910, opened 2024-07-02
front: fix via modal display

### LINKED ISSUE #7886: Weird alignment in the via modal for long names
### What happened?

Create a path with rolling stock 2TGVDUPL, using the rocket : PLY VSG DN SXA
The alignment "minus" sign of the VSG point is weird 
image
It is ok for shorter names like Dijon : 
image

I expected the same alignment

as above

- **Browser**: Chrome
- **OS**: W11
- **Env**: dev dgexsol

028ae7b

SarahBellaha:
Closes #7886

### DIFF
--- a/front/src/modules/trainschedule/styles/ManageTrainSchedule/_itinerary.scss
+++ b/front/src/modules/trainschedule/styles/ManageTrainSchedule/_itinerary.scss
@@ -58,10 +58,24 @@
       }
 
       &-name {
-        max-width: 47%;
+        width: 45%;
         white-space: nowrap;
         overflow: hidden;
         text-overflow: ellipsis;
+        
+        &.reduced-text {
+          width: 41%;
+        }
+      }
+
+      &-uic {
+        width: 15%;
+        text-align: end;
+        padding-right: 0.5rem;
+      }
+
+      &-ch {
+        width: 6%;
       }
     }
   }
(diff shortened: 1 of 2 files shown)


theocrsb:
thank you for your PR. behavior is fixed for some names but some still have weird behavior

image

front/src/modules/trainschedule/styles/ManageTrainSchedule/_itinerary.scss
```diff
@@ -58,10 +58,24 @@
       }
 
       &-name {
-        max-width: 47%;
+        width: 45%;
```
theocrsb, line 61: I feel like the behavior is better with 40%

Uriel-Sautron (approved):
LGTM, tested the behavior is ok

front/src/modules/trainschedule/styles/ManageTrainSchedule/_itinerary.scss
Yohh: that's a UI decision, we should be careful with this kind of suggestions based on personal preferences
also, the via selected is styled by `.reduce-text`, wich has `width: 41%;`

Yohh:
> thank you for your PR. behavior is fixed for some names but some still have weird behavior
> 
> image

nothing like that for me, did your switch work? are you really on the sba/front/fix-via-modal-7886 branch?

theocrsb:
> > thank you for your PR. behavior is fixed for some names but some still have weird behavior
> > image
> 
> nothing like that for me, did your switch work? are you really on the sba/front/fix-via-modal-7886 branch?

My bad. I tried again and it's ok. maybe switch doesn't work.

theocrsb (approved):
LGTM and tested

clarani (approved):

PR #7915, opened 2024-07-02
core: test ram limit

younesschrifi:


eckter:
Context : not to be merged, this is just to figure out what happens when reaching RAM limits in deployment envs (we've seen some weird behaviors). No need to review.

PR #7922, opened 2024-07-02
front: keep common value on switch when changing switch types

### LINKED ISSUE #7604: infrastructure editor switch form is cleared when switch type is changed
### What happened?

I start creating a switch and note a name and time : 

image

I realised I had the wrong type (2 ports instead of 3 : I change the type, and the information I had written before are erased :

image

I expected to keep the information previously written

as above

- **Browser**: Chrome
- **OS**: W11
- **Env**: staging dgexsol

625d185

sim51:
Fix #7604


SharglutDev (approved):
Lgtm and tested !

PR #7923, opened 2024-07-02
front: fix chart pointers and train details

### LINKED ISSUE #7801: TSV2 : pointer on the space time chart is not on the curve when margins are included
### What happened?

When you create a train with margins or by fixing some passing times, the pointer in the space time chart follows the basic curve (the fastest one) which is not visible

The pointer should follow the real train curve 

1. Create a projet/study/scenario
2. Create a train by fixing at least a margin or a passing time 
3. Go to simulation results
4. Move the mouse on the space time chart

- **Browser**: Edge
- **OS**: windows
- **Env**: recette

dev c57458e

emersion:
_See individual commits._

Closes: https://github.com/OpenRailAssociation/osrd/issues/7801

* * *

This has one unfortunate consequence: the space-speed chart now (correctly) displays pointers that are not aligned vertically for a given timestamp. That is, the pointer for the eco curve will lag behind the pointer for the base curve. See below:

out

I don't see a good way to fix it when hovering the space-time chart... Maybe only show the pointer for eco, not for base?

### DIFF
--- a/front/src/applications/operationalStudies/hooks.ts
+++ b/front/src/applications/operationalStudies/hooks.ts
@@ -234,77 +234,79 @@ export const useSimulationResults = () => {
           setFormattedPowerRestrictions(powerRestrictionsWithHandled);
         }
 
-        // Format chart synchronizer data
-        const {
-          baseHeadPositions,
-          baseTailPositions,
-          baseSpeeds,
-          marginSpeeds,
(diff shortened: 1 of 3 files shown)


Uriel-Sautron (approved):
Good work, tested !!

clarani (approved):

PR #7924, opened 2024-07-02
Set editoast tokio flavor multi-thread

flomonster:
- Bump to rustc 1.79.0
- Use tokio multi thread scheduler instead of current thread.

Tokio doc: https://docs.rs/tokio/latest/tokio/attr.main.html#multi-threaded-runtime

> [!NOTE]
> When printing the flavor in the main we have `MultiThread` instead of `CurrentThread`.
> However we still observe `CurrentThread` when printing inside an endpoint.

### DIFF
--- a/editoast/Cargo.lock
+++ b/editoast/Cargo.lock
@@ -8,7 +8,7 @@ version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5f7b0a21988c1bf877cf4759ef5ddaac04c1c9fe808c9142ecb78ba97d97a28a"
 dependencies = [
- "bitflags 2.5.0",
+ "bitflags 2.6.0",
  "bytes",
  "futures-core",
  "futures-sink",
@@ -44,7 +44,7 @@ dependencies = [
  "actix-service",
  "actix-utils",
  "actix-web",
- "bitflags 2.5.0",
+ "bitflags 2.6.0",
  "bytes",
  "derive_more",
  "futures-core",
@@ -69,7 +69,7 @@ dependencies = [
  "actix-utils",
  "ahash",
  "base64 0.22.1",
- "bitflags 2.5.0",
+ "bitflags 2.6.0",
  "brotli",
  "bytes",
  "bytestring",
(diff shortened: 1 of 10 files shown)


ghost (approved):
That sounds like an incredibly simple yet perfect solution

flomonster:
> That sounds like an incredibly simple yet perfect solution

This does not fix the usage of `join_all` futures.

woshilapin (approved):
Digging a bit into `actix-web`, it seems to basically launch as many actix worker as there is CPU available by default. An actix worker is basically an OS thread with a mono-thread `tokio` runtime in it. Here is how it works.

1. [We set up the number of actix worker in `editoast`](https://github.com/OpenRailAssociation/osrd/blob/dev/editoast/src/main.rs#L426-L429)
2. This number of worker is stored into [`threads` of the `ServerBuilder`](https://github.com/actix/actix-net/blob/server-v2.3.0/actix-server/src/builder.rs#L33) into our instance of [`HTTPServer`](https://github.com/actix/actix-web/blob/web-v4.8.0/actix-web/src/server.rs#L81)
3. This `threads` number is used to [instantiate as many `ServerWorker`](https://github.com/actix/actix-net/blob/910c181251251da38948b9297464aab84edcd719/actix-server/src/accept.rs#L52-L66)
4. In the `ServerWorker::start`, [the handle on the current `tokio::Runtime` is gathered](https://github.com/actix/actix-net/blob/910c181251251da38948b9297464aab84edcd719/actix-server/src/worker.rs#L299) 
5. Always in `ServerWorker::start`, [a new OS thread is spawned](https://github.com/actix/actix-net/blob/910c181251251da38948b9297464aab84edcd719/actix-server/src/worker.rs#L327-L329) (remember, `ServerWorker::start` is called as many times as there is workers)
6. Finally, inside this new OS thread, [a new mono-threaded `tokio` runtime is launched](https://github.com/actix/actix-net/blob/910c181251251da38948b9297464aab84edcd719/actix-server/src/worker.rs#L398-L407)

Note that a lot is happening in `ServerWorker::start` which I'm not entirely sure the impact. But basically, it probably doesn't matter if our `fn main` is multi-threaded or not since actix is basically running a mono-threaded `tokio` runtime in independent OS threads from what I understand.

That said, I believe that defaulting to a multi-threaded `tokio` runtime is good anyway, even if it might not help us in what we're looking for.

ghost:
> > That sounds like an incredibly simple yet perfect solution
> 
> This does not fix the usage of `join_all` futures.

My bad, I thought it would.

PR #7925, opened 2024-07-02
front: if current infra is the one we delete, change the store

### LINKED ISSUE #7798: After deleting an infrastructure I can still see it in the map but I get an error
### What happened?

I have deleted the infrastructure `small_infra Brieuc` but the map and title has not been refreshed:
image
So I can still see my infrastructure and I have a popup indicating it doesn't exists when I open the infrastructure selection popup.

After deleting an infrastructure I expect that the map is empty with a message "Aucune infrastructure sélectionnée" and a the infrastructure button is flashing on the map

1. Select an infrastructure
2. Go on the map
3. Delete this infrastructure
4. Go on the map (you can still see it)
5. Reopen the infrastructure selection popup

- **Browser**: Firefox 

c57458e

sim51:
Fix #7798


emersion (approved):

theocrsb (approved):
LGTM and tested

SharglutDev (approved):
Lgtm and tested, good job !

PR #7926, opened 2024-07-02
editoast: specify versions for all dependencies

woshilapin:
`*` versions might create some unwanted behavior in case of `cargo update`.
It also homogenizes all our dependencies' versions.


flomonster (approved):
LGTM, just need to rebase

PR #7930, opened 2024-07-03
front: fix via added by map click not updated

### LINKED ISSUE #7929: Can't edit a via that has been selected by map click
### What happened?

https://github.com/OpenRailAssociation/osrd/assets/113104137/6349d2d2-f8cf-4b5b-8c5d-8c7b2c57146d

We should be able to edit this kind of via aswell

See video

- **Browser**: Firefox 126
- **OS**: Ubuntu
- **Env**: Dev

dev bbab375

SharglutDev:
close #7929

### DIFF
--- a/front/src/modules/timesStops/utils.ts
+++ b/front/src/modules/timesStops/utils.ts
@@ -25,11 +25,27 @@ export const formatSuggestedViasToRowVias = (
       formattedOps.unshift(operationalPoints[originIndexInOps]);
     }
   }
-  return formattedOps.map((op, i) => ({
-    ...op,
-    name: op.name || t('waypoint', { id: op.opId }),
-    isMarginValid: op.theoreticalMargin ? marginRegExValidation.test(op.theoreticalMargin) : true,
-    onStopSignal: op.onStopSignal || false,
-    arrival: i === 0 ? startTime?.substring(11, 19) : op.arrival,
-  }));
+  return formattedOps.map((op, i) => {
+    const isMarginValid = op.theoreticalMargin
+      ? marginRegExValidation.test(op.theoreticalMargin)
+      : true;
+
+    const correspondingOp = !('uic' in op)
+      ? pathSteps.find((step) => step.id === op.opId)
+      : undefined;
+
+    const { arrival, onStopSignal, name } = correspondingOp || op;
+
+    return {
+      ...op,
+      isMarginValid,
+      arrival: i === 0 ? startTime?.substring(11, 19) : arrival,
+      onStopSignal: onStopSignal || false,
+      name: name || t('waypoint', { id: op.opId }),
+      ...(correspondingOp && {
+        stopFor: correspondingOp.stopFor,
+        theoreticalMargin: correspondingOp.theoreticalMargin,
+      }),
+    };
+  });
 };


front/src/modules/timesStops/utils.ts
```diff
@@ -25,11 +25,28 @@ export const formatSuggestedViasToRowVias = (
...
+          theoreticalMargin: correspondingOp.theoreticalMargin,
+        };
+      }
+    }
+    return {
+      ...op,
+      isMarginValid: op.theoreticalMargin ? marginRegExValidation.test(op.theoreticalMargin) : true,
+      onStopSignal: op.onStopSignal || false,
+      arrival: i === 0 ? startTime?.substring(11, 19) : op.arrival,
+    };
```
Yohh: instead of nesting condition and give 2 returns, can't you manage `isMarginValid` , `onStopSignal` and `arrival` then return the result?

SharglutDev: I could propose you that but I think I prefer the first syntax. Do you have another idea ? 

```
    let opToFormat = {
      ...op,
      isMarginValid: op.theoreticalMargin ? marginRegExValidation.test(op.theoreticalMargin) : true,
      onStopSignal: op.onStopSignal || false,
      arrival: i === 0 ? startTime?.substring(11, 19) : op.arrival,
    };
    if (!('uic' in op)) {
      const correspondingOp = pathSteps.find((step) => step.id === op.opId);
      if (correspondingOp) {
        opToFormat = {
          ...opToFormat,
          name: op.name || t('waypoint', { id: op.opId }),
          arrival: i === 0 ? startTime?.substring(11, 19) : correspondingOp.arrival,
          stopFor: correspondingOp.stopFor,
          theoreticalMargin: correspondingOp.theoreticalMargin,
        };
      }
    }

    return opToFormat;
  });
```

theocrsb (approved):
LGTM and tested

front/src/modules/timesStops/utils.ts
SarahBellaha: Maybe something like that ?

```suggestion
    const correspondingOp = pathSteps.find((step) => step.id === op.opId);
    const isMarginValid = op.theoreticalMargin
      ? marginRegExValidation.test(op.theoreticalMargin)
      : true;
    const { arrival } = correspondingOp || op;
    const { onStopSignal } = correspondingOp || op;
    const { name } = correspondingOp || op;

    return {
      ...op,
      isMarginValid,
      arrival: i === 0 ? startTime?.substring(11, 19) : arrival,
      onStopSignal: onStopSignal || false,
      name: name || t('waypoint', { id: op.opId }),
      ...(correspondingOp && {
        stopFor: correspondingOp.stopFor,
        theoreticalMargin: correspondingOp.theoreticalMargin,
      }),
    };
```

emersion: Alternatively, keep the current logic with 2 returns, and use a const for the `isMarginValid` field, since that's the only common one between the two cases?

```js
const isMarginValid = op.theoreticalMargin ? marginRegExValidation.test(op.theoreticalMargin) : true;
if (correspondingOp) {
  return {
    ...op,
    name: op.name || t('waypoint', { id: op.opId }),
    isMarginValid,
    // etc
  };
}

return {
  ...op,
  isMarginValid,
  // etc
};
```

Yohh: that's the way I was thinking about

SarahBellaha: I updated my suggestion ⏫

SharglutDev: I think for code readability/maintainability that it is still better with 2 returns. We have only one big condition between "has 'uic'" or not instead of a small condition for each properties which is harder to track which one comes from `op` or from `correspondingOp` . I can add @emersion suggestion to remove the repetition for isMarginValid.
Is that ok for all of you ?

Yohh: I disagree with both your readability/maintainability arguments, the solution proposed by @SarahBellaha seems to me cleaner and far more maintainable, anyway your solution and the one from @emersion are both functional, so feel free! it was just a suggestion

emersion (approved):
Sure, LGTM :)

Yohh (approved):
nice, LGTM :muscle:

SarahBellaha (approved):
LGTM :)

Akctarus (approved):
Tested, LGTM

clarani (approved):
LGTM :white_check_mark:

Issue #6980, opened 2024-03-26
Rework editoast's testing utils


leovalais:
# Description and goal

This is a required step in order to complete #6397.

Blocks the isolation of `models/`.

## Context

We expect the `models` crate to be the only one to depend on `diesel`, and the `views` crate to be the only one to depend on `actix`. This has several implications on the architecture we currently use for tests, where we mix `diesel` and `actix` thoroughly.

## The connection pool

We use `Data<DbPool>` everywhere. Its full type is `actix_web::Data<diesel_async::pooled_connection::deadpool::Pool<PgConnection>>`... The connection pool being a `editoast_models` object, we'll have to strip `actix_web` from this type. 

It wouldn't be very convenient to use an `Arc<Pool<...>>` because the `editoast_models` crate would then have to re-export `diesel_async` to allow its users to fully qualify the type.

Better create an opaque wrapper `struct ConnectionPool(Arc<Pool<PgConnection>>)`. It might also make the management of test connections easier (see below).
For the same reason, a `type Connection = diesel_async::AsyncPgConnection` might be a good idea.

## Avoiding flaky tests

Currently, `TestFixture<T>` weakly protect us from test data leaking, but it's fragile, especially for data created using endpoints. It also forces us to keep track of every object created by the test.

A better model would be to open a transaction at the beginning of the test and never commit it. `diesel` provides us with such a mechanism through https://docs.rs/diesel-async/0.4.1/diesel_async/trait.AsyncConnection.html#method.begin_test_transaction. But that only works for a single connection. Test data inserted using this connection is only available to its following requests (it's just a transaction after all). We need to make sure that the endpoints use that same connection, so there may be some logic to define into `ConnectionPool`. 

```rust
impl ConnectionPool {
    // Runtime
    fn new() -> Self
    // With a flag to ensure that get() always returns the same Connection
    fn for_tests() -> Self
    async fn get() -> Result<Connection, E>
}

#[rstest]
async fn test() {
    let pool = ConnectionPool::for_tests();
    let app = create_test_service(&pool).await; // make sure that the endpoints use the same pool, hence the same connection
    ...
}
```

## What about `TestFixture<T>`

If everything runs within a transaction that's never committed, we don't need their `Drop` behavior. So we don't need the fixtures anymore. And since the `#[fixture]` is kinda useless (stack overflows), we might as well take the opportunity to remove those altogether.

# Acceptance criteria

[...]

PR #7932, opened 2024-07-03
editoast: refactor infra tests part 7

hamz2a:
Part of https://github.com/OpenRailAssociation/osrd/issues/6980

It's better to review the PR commit by commit.
I created intermediate functions `{function}_v2` to facilitate the migration, but I deleted everything at the end.

I also refactored `persist_railjson` to use `DbConnection` directly instead of `DbConnectionPool`. Consequently, I removed the macro `persist!` that was in there and `futures::try_join!`, and I added a transaction.

### DIFF
--- a/editoast/src/fixtures.rs
+++ b/editoast/src/fixtures.rs
@@ -474,7 +474,7 @@ pub mod tests {
         Infra::changeset()
             .name("small_infra".to_owned())
             .last_railjson_version()
-            .persist(railjson, db_pool)
+            .persist(railjson, db_pool.get().await.unwrap().deref_mut())
             .await
             .unwrap()
     }
--- a/editoast/src/fixtures.rs
+++ b/editoast/src/fixtures.rs
@@ -4,6 +4,7 @@ pub mod tests {
     use std::io::Cursor;
     use std::ops::{Deref, DerefMut};
     use std::sync::Arc;
+    use uuid::Uuid;
 
     use editoast_models::create_connection_pool;
     use editoast_models::DbConnection;
@@ -261,8 +262,7 @@ pub mod tests {
         } = scenario_fixture_set().await;
 
         let pathfinding = pathfinding(db_pool()).await;
-        let mut rs_name = "fast_rolling_stock_".to_string();
-        rs_name.push_str(name);
+        let rs_name = format!("fast_rolling_stock_{}_{name}", Uuid::new_v4()).to_string();
         let rolling_stock = named_fast_rolling_stock(&rs_name, db_pool()).await;
         let ts_model = make_train_schedule(
             db_pool(),
--- a/editoast/src/generated_data/mod.rs
+++ b/editoast/src/generated_data/mod.rs
@@ -86,38 +86,39 @@ pub trait GeneratedData {
 }
 
 /// Refresh all the generated data of a given infra
+#[tracing::instrument(level = "debug", skip_all, fields(infra_id))]
 pub async fn refresh_all(
     db_pool: Arc<DbConnectionPoolV2>,
-    infra: i64,
+    infra_id: i64,
     infra_cache: &InfraCache,
 ) -> Result<()> {
     // The other layers depend on track section layer.
     // We must wait until its completion before running the other requests in parallel
-    TrackSectionLayer::refresh_pool(db_pool.clone(), infra, infra_cache).await?;
-    debug!("⚙️ Infra {infra}: track section layer is generated");
+    TrackSectionLayer::refresh_pool(db_pool.clone(), infra_id, infra_cache).await?;
+    debug!("⚙️ Infra {infra_id}: track section layer is generated");
     // The analyze step significantly improves the performance when importing and generating together
     // It doesn’t seem to make a different when the generation step is ran separately
     // It isn’t clear why without analyze the Postgres server seems to run at 100% without halting
     sql_query("analyze")
         .execute(db_pool.get().await?.deref_mut())
         .await?;
-    debug!("⚙️ Infra {infra}: database analyzed");
+    debug!("⚙️ Infra {infra_id}: database analyzed");
     futures::try_join!(
-        SpeedSectionLayer::refresh_pool(db_pool.clone(), infra, infra_cache),
-        SignalLayer::refresh_pool(db_pool.clone(), infra, infra_cache),
-        SwitchLayer::refresh_pool(db_pool.clone(), infra, infra_cache),
-        BufferStopLayer::refresh_pool(db_pool.clone(), infra, infra_cache),
-        ElectrificationLayer::refresh_pool(db_pool.clone(), infra, infra_cache),
-        DetectorLayer::refresh_pool(db_pool.clone(), infra, infra_cache),
-        OperationalPointLayer::refresh_pool(db_pool.clone(), infra, infra_cache),
-        PSLSignLayer::refresh_pool(db_pool.clone(), infra, infra_cache),
-        NeutralSectionLayer::refresh_pool(db_pool.clone(), infra, infra_cache),
-        NeutralSignLayer::refresh_pool(db_pool.clone(), infra, infra_cache),
+        SpeedSectionLayer::refresh_pool(db_pool.clone(), infra_id, infra_cache),
+        SignalLayer::refresh_pool(db_pool.clone(), infra_id, infra_cache),
+        SwitchLayer::refresh_pool(db_pool.clone(), infra_id, infra_cache),
+        BufferStopLayer::refresh_pool(db_pool.clone(), infra_id, infra_cache),
+        ElectrificationLayer::refresh_pool(db_pool.clone(), infra_id, infra_cache),
+        DetectorLayer::refresh_pool(db_pool.clone(), infra_id, infra_cache),
+        OperationalPointLayer::refresh_pool(db_pool.clone(), infra_id, infra_cache),
+        PSLSignLayer::refresh_pool(db_pool.clone(), infra_id, infra_cache),
+        NeutralSectionLayer::refresh_pool(db_pool.clone(), infra_id, infra_cache),
+        NeutralSignLayer::refresh_pool(db_pool.clone(), infra_id, infra_cache),
     )?;
-    debug!("⚙️ Infra {infra}: object layers is generated");
+    debug!("⚙️ Infra {infra_id}: object layers is generated");
     // The error layer depends on the other layers and must be executed at the end.
-    ErrorLayer::refresh_pool(db_pool.clone(), infra, infra_cache).await?;
-    debug!("⚙️ Infra {infra}: errors layer is generated");
+    ErrorLayer::refresh_pool(db_pool.clone(), infra_id, infra_cache).await?;
+    debug!("⚙️ Infra {infra_id}: errors layer is generated");
     Ok(())
 }
 
@@ -171,7 +172,10 @@ pub mod tests {
     use crate::modelsv2::fixtures::create_empty_infra;
     use editoast_models::DbConnectionPoolV2;
 
-    #[rstest] // Slow test
+    #[rstest]
+    // Slow test
+    // PostgreSQL deadlock can happen in this test, see section `Deadlock` of [DbConnectionPoolV2::get] for more information
+    #[serial_test::serial]
     async fn refresh_all_test() {
         let db_pool = DbConnectionPoolV2::for_tests();
         let infra = create_empty_infra(db_pool.get_ok().deref_mut()).await;
--- a/editoast/src/generated_data/mod.rs
+++ b/editoast/src/generated_data/mod.rs
@@ -209,20 +209,21 @@ pub mod tests {
     use rstest::rstest;
     use std::ops::DerefMut;
 
-    use crate::fixtures::tests::db_pool;
     use crate::generated_data::clear_all;
-    use crate::generated_data::refresh_all;
+    use crate::generated_data::refresh_all_v2;
     use crate::generated_data::update_all;
     use crate::modelsv2::fixtures::create_empty_infra;
     use editoast_models::DbConnectionPoolV2;
 
     #[rstest] // Slow test
     async fn refresh_all_test() {
-        let db_pool_v2 = DbConnectionPoolV2::for_tests();
-        let infra = create_empty_infra(db_pool_v2.get_ok().deref_mut()).await;
-        assert!(refresh_all(db_pool(), infra.id, &Default::default())
-            .await
-            .is_ok());
+        let db_pool = DbConnectionPoolV2::for_tests();
+        let infra = create_empty_infra(db_pool.get_ok().deref_mut()).await;
+        assert!(
+            refresh_all_v2(db_pool.into(), infra.id, &Default::default())
+                .await
+                .is_ok()
+        );
     }
 
     #[rstest]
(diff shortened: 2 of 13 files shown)


editoast/src/generated_data/mod.rs
```diff
@@ -120,6 +130,41 @@ pub async fn refresh_all(
...
 
+pub async fn refresh_all_v2(
+    db_pool: Arc<DbConnectionPoolV2>,
+    infra: i64,
+    infra_cache: &InfraCache,
+) -> Result<()> {
+    // The other layers depend on track section layer.
+    // We must wait until its completion before running the other requests in parallel
+    TrackSectionLayer::refresh_pool_v2(db_pool.clone(), infra, infra_cache).await?;
+    debug!("⚙️ Infra {infra}: track section layer is generated");
```
woshilapin: For this log and all the other below, I would advise to put the `infra` variable as a structure data in the logs. 2 solutions:

- add `#[tracing::instrument(level = "debug", skip_all, fields(infra_id = infra))]` at the top of the function
- or change the log into `debug!(infra_id = infra, "Infra: track section layer is generated");`

editoast/src/views/infra/railjson.rs
```diff
@@ -239,11 +243,13 @@ mod tests {
...
 
-        let conn = &mut db_pool.get().await.unwrap();
-        assert!(Infra::delete_static(conn, res.infra).await.unwrap());
+        let res: PostRailjsonResponse = app.fetch(req).assert_status(StatusCode::OK).json_into();
+
+        assert!(
+            Infra::delete_static(db_pool.get_ok().deref_mut(), res.infra)
+                .await
+                .unwrap()
+        );
```
woshilapin, line 256: I'm not entirely sure, do we want to test for deletion OR do we NEED to delete? What happens if, for some reason, the test panics just above, does the DB stay in a clumsy situation?

editoast/src/views/infra/auto_fixes/mod.rs
```diff
@@ -380,56 +368,77 @@ mod tests {
...
+            .expect("Failed to refresh infra");
 
         // Check the only initial issues are "overlapping_speed_sections" warnings
-        let (infra_errors_before_all, before_all_count) = query_errors(conn, &small_infra).await;
+        let (infra_errors_before_all, before_all_count) =
+            query_errors(db_pool.get_ok().deref_mut(), &small_infra).await;
         assert!(infra_errors_before_all
             .iter()
             .all(|e| matches!(e.sub_type, InfraErrorType::OverlappingSpeedSections { .. })));
+        dbg!(infra_errors_before_all, before_all_count);
```
woshilapin: This can probably be removed.

editoast/src/views/infra/auto_fixes/mod.rs
```diff
@@ -380,56 +368,77 @@ mod tests {
...
+        small_infra
+            .refresh_v2(db_pool.clone(), true, &infra_cache)
+            .await
+            .expect("Failed to refresh infra");
 
         // Check that some new issues appeared
-        let (infra_errors_before_fix, before_fix_count) = query_errors(conn, &small_infra).await;
+        let (infra_errors_before_fix, before_fix_count) =
+            query_errors(db_pool.get_ok().deref_mut(), &small_infra).await;
+        dbg!(before_fix_count, before_all_count);
```
woshilapin: This can probably be removed.

editoast/src/fixtures.rs
```diff
@@ -465,14 +465,15 @@ pub mod tests {
...
+        let conn = &mut db_pool.get().await.unwrap();
         let railjson: RailJson = serde_json::from_str(include_str!(
             "../../tests/data/infras/small_infra/infra.json"
         ))
         .unwrap();
         Infra::changeset()
             .name("small_infra".to_owned())
             .last_railjson_version()
-            .persist(railjson, db_pool)
+            .persist(railjson, conn)
```
woshilapin: `conn` should probably be inlined here, to drop the connection as soon as possible.

```suggestion
            .persist(railjson, db_pool.get().await.unwrap().deref_mut())
```

editoast/src/modelsv2/railjson.rs
```diff
@@ -57,27 +38,88 @@ pub async fn persist_railjson(
...
-        persist!(RouteModel, routes),
-        persist!(SignalModel, signals),
-        persist!(SwitchModel, switches),
-        persist!(SpeedSectionModel, speed_sections),
-        persist!(SwitchTypeModel, extended_switch_types),
-        persist!(NeutralSectionModel, neutral_sections),
-    )
-    .map(|_| ())
+
+    connection
```
woshilapin, line 50: Removing the `futures::try_join!()` in favor of this big block should have a performance impact if I'm not wrong. `try_join!()` does launch all the futures concurrently, but I do believe this new code will run sequentially.

woshilapin (changes requested):
I'm curious about why we do need a lot of `#[serial_test::serial]` in commit 3040c347c5a9c7fd0a23aeb2e24ea2ae2aa89e3a.

Also, I'd like to be convinced that removing `futures::try_join_all!()` is fine. But to me, it should affect the time of execution for `persist_railjson` unless I'm missing something.

Apart from that, great work in separating commits in this PR. Thank you, it really eased the review for this big PR.

editoast/src/generated_data/mod.rs
hamz2a: Done.

editoast/src/views/infra/railjson.rs
hamz2a: I think we're testing the delete after creation. If the test panics, we won't have any issues in the DB since we don't commit the changes.

editoast/src/views/infra/auto_fixes/mod.rs
hamz2a: Done.

editoast/src/views/infra/auto_fixes/mod.rs
hamz2a: Done.

editoast/src/fixtures.rs
hamz2a: Done.

hamz2a:
> I'm curious about why we do need a lot of `#[serial_test::serial]` in commit [3040c34](https://github.com/OpenRailAssociation/osrd/commit/3040c347c5a9c7fd0a23aeb2e24ea2ae2aa89e3a).
> 

I added `#[serial_test::serial]` because I was getting deadlock errors.

editoast/src/modelsv2/railjson.rs
woshilapin: OK, after discussion, we acknowledge that it might be a performance regression, but will avoid starving the DB pool which might kill the entire service, it's a balance ⚖️

woshilapin:
> > I'm curious about why we do need a lot of `#[serial_test::serial]` in commit [3040c34](https://github.com/OpenRailAssociation/osrd/commit/3040c347c5a9c7fd0a23aeb2e24ea2ae2aa89e3a).
> 
> I added `#[serial_test::serial]` because I was getting deadlock errors.

We dug a little bit to try to understand. It's a deadlock detected by the Postgresql instance itself. It might be due to concurrence on the generated id for infras since we do create very long transactions when working with infras in tests.

We did not find an obvious solution to work around the problem. Let's keep it like this.

woshilapin (approved):

editoast/src/fixtures.rs
```diff
@@ -261,8 +262,7 @@ pub mod tests {
         } = scenario_fixture_set().await;
 
         let pathfinding = pathfinding(db_pool()).await;
-        let mut rs_name = "fast_rolling_stock_".to_string();
-        rs_name.push_str(name);
+        let rs_name = format!("fast_rolling_stock_{}_{name}", Uuid::new_v4()).to_string();
```
leovalais, line 265: Which was failing without the use of UUIDS?

editoast/src/generated_data/mod.rs
```diff
@@ -164,18 +165,18 @@ pub mod tests {
...
 
-    use crate::fixtures::tests::db_pool;
     use crate::generated_data::clear_all;
     use crate::generated_data::refresh_all;
     use crate::generated_data::update_all;
     use crate::modelsv2::fixtures::create_empty_infra;
     use editoast_models::DbConnectionPoolV2;
 
     #[rstest] // Slow test
+    #[serial_test::serial]
```
leovalais, line 178: If I understand correctly is this causes a non trivial Postgres issue. We could add a comment explaining the error and why serial fixes it. (Especially since we're serializing a test marked as a `// Slow test` :D)

editoast/src/modelsv2/railjson.rs
leovalais: It could be interesting to measure how much the performance is degraded. As far as I know, this function is either used for:

1. tests, where no big infra is imported -- so it's irrelevant
2. importing an infra using the frontend, we're however limited by the size of request payloads, which kinda breaks the whole feature (which should be removed until we find a solution imo)
3. loading a recently imported infra while testing dags or creating a dump, in which case a small perf degradation is painless, but a big one can be pretty impactful on the workflow

If performance of `persist` becomes problematic, we could make it accept an `Either<&mut conn, db_pool>` and make the queries concurrent if the pool is provided. We could then provide the pool for the CLI import, but keep the single connection for the endpoint to avoid starvation.

editoast/src/views/infra/mod.rs
```diff
@@ -667,29 +657,22 @@ pub mod tests {
...
+
+        let request = TestRequest::post()
+            .uri(format!("/infra/{}/clone/?name=cloned_infra", empty_infra.id).as_str())
             .to_request();
-        let response = call_service(&app, req).await;
-        let cloned_infra_id: i64 = assert_status_and_read!(response, StatusCode::OK);
-        let cloned_infra = Infra::retrieve(conn, cloned_infra_id)
+
+        let cloned_infra_id: i64 = app.fetch(request).assert_status(StatusCode::OK).json_into();
+        let cloned_infra = Infra::retrieve(db_pool.get_ok().deref_mut(), cloned_infra_id)
```
leovalais, line 672: `Infra::exists` could be used instead.

editoast/src/generated_data/mod.rs
```diff
@@ -87,19 +87,20 @@ pub trait GeneratedData {
...
+    db_pool: Arc<DbConnectionPoolV2>,
     infra: i64,
     infra_cache: &InfraCache,
 ) -> Result<()> {
     // The other layers depend on track section layer.
     // We must wait until its completion before running the other requests in parallel
     TrackSectionLayer::refresh_pool(db_pool.clone(), infra, infra_cache).await?;
-    debug!("⚙️ Infra {infra}: track section layer is generated");
-    let mut conn = db_pool.get().await?;
+    debug!(infra_id = infra, "Infra: track section layer is generated");
```
leovalais: Maybe adding a `#[tracing::instrument(fields(infra_id)` instead to create a span would be more idiomatic? @woshilapin a confirmation ? 😅

leovalais (changes requested):
Thanks for this work! PoolV1 is almost gone! 😄

editoast/src/fixtures.rs
hamz2a: `delete_used_rolling_stock_should_fail` is a `tsv1` test

editoast/src/generated_data/mod.rs
hamz2a: I just added `#[tracing::instrument...`

editoast/src/views/infra/mod.rs
hamz2a: Done.

editoast/src/modelsv2/railjson.rs
woshilapin: Well, the problem with having a `pool` directly is that we might starve the pool... and therefore, `/health` which does check the DB is up and running won't have an available connection, then take more than a second to respond. If the deployment is configured to restart the service when `/health` is too slow (which is our case), then it starts to be a bigger problem.

editoast/src/generated_data/mod.rs
hamz2a: Done.

editoast/src/generated_data/mod.rs
```diff
@@ -68,12 +69,11 @@ pub trait GeneratedData {
...
 
     async fn refresh_pool(
-        pool: Arc<DbConnectionPool>,
+        pool: Arc<DbConnectionPoolV2>,
         infra: i64,
         infra_cache: &InfraCache,
     ) -> Result<()> {
-        let mut conn = pool.get().await?;
-        Self::refresh(&mut conn, infra, infra_cache).await
+        Self::refresh(pool.get().await?.deref_mut(), infra, infra_cache).await
```
leovalais, line 76: Better to get rid of as much `_pool` variants as possible. Ideally all functions that use a single connection would take a connection as argument.

editoast/src/generated_data/mod.rs
```diff
@@ -86,37 +86,39 @@ pub trait GeneratedData {
 }
 
 /// Refresh all the generated data of a given infra
+#[tracing::instrument(level = "debug", skip_all, fields(infra_id))]
 pub async fn refresh_all(
-    db_pool: Arc<DbConnectionPool>,
-    infra: i64,
+    db_pool: Arc<DbConnectionPoolV2>,
```
leovalais, line 91: Likewise about the pool

editoast/src/generated_data/mod.rs
```diff
@@ -164,18 +166,19 @@ pub mod tests {
...
 
-    use crate::fixtures::tests::db_pool;
     use crate::generated_data::clear_all;
     use crate::generated_data::refresh_all;
     use crate::generated_data::update_all;
     use crate::modelsv2::fixtures::create_empty_infra;
     use editoast_models::DbConnectionPoolV2;
 
     #[rstest] // Slow test
+    /// For more details, see `DbConnectionPoolV2::get`, Deadlocks section.
```
leovalais:
```suggestion
    // PostgreSQL deadlock can happen in this test, see section `Deadlock` of [DbConnectionPoolV2::get] for more information
```

`///` is for documentation, not comments

When referencing an object, better use `[ns::obj]` to allow the CI check the link isn't broken.

leovalais (changes requested):
Thanks for the changes! A few more and 🚀 
I'll test your PR on Monday.

There are commits like `merge branch dev` or `rebase` in your history. Those will need to be squashed.

editoast/src/generated_data/mod.rs
hamz2a: Done.

editoast/src/generated_data/mod.rs
hamz2a: As we discussed, we will leave it as it is.

editoast/src/generated_data/mod.rs
hamz2a: As we discussed, we will leave it as it is.

leovalais (approved):
Great job and thank you for the investigation of the deadlocks bug and for rearranging the history!

Once this is merged we'll need to check that all non-TSV1 endpoints are using `DbPoolV2` and that we didn't miss any test that was using `TestFixture`.

After that we can probably work on making the `DbConnection` opaque to allow us to remove any `diesel` from `views` :)

PR #7933, opened 2024-07-03
core: skip work schedules on track chunks with no routes

### LINKED ISSUE #7871: stdcm: null pointer exception in an empty timetable
### What happened?

The error:
```json
{"status":500,"type":"core:unknown_error","context":{"stack_trace":["RawInfraImpl.kt:609","BlockAvailability.kt:145","BlockAvailability.kt:106","STDCMEndpointV2.kt:79","FkRegex.java:153","FkRegex.java:217","FkChain.java:72","TkFork.java:98","TkDataDog.kt:27","TkDataDog.kt:16","TkDataDog.kt:15","TkWrap.java:58","TkFallback.java:84","TkFallback.java:66","TkWrap.java:58","TkSlf4j.java:110","BkBasic.java:123","BkBasic.java:99","BkSafe.java:46","BkWrap.java:51","BkParallel.java:81","ThreadPoolExecutor.java:1144","ThreadPoolExecutor.java:642","Thread.java:1583"],"url":"http://osrd-rec-core.osd-rec01.svc.cluster.local/v2/stdcm","exception_type":"class java.lang.NullPointerException"},"message":"An unknown exception was thrown"}
```

No error

1. create an empty scenario with TSv2
2. try a STDCM path like :
image

- **Browser**: Firefox 

d9655c0

Erashin:
Fixes #7871.

For more context: in the current infra, there are currently 1265 track sections which are fully not covered by routes (and 9 which are partially covered). As a result, the corresponding 2465 track chunks are not covered by routes, and hence have no zones, which is the root of this bug. There might be a bug in route generation, it is getting looked into. For now, we can just skip the corresponding work schedules (the method isn't used elsewhere).

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
@@ -17,6 +17,10 @@ import fr.sncf.osrd.utils.units.Offset
 import fr.sncf.osrd.utils.units.meters
 import kotlin.math.max
 import kotlin.math.min
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+val blockAvailabilityLogger: Logger = LoggerFactory.getLogger("BlockAvailability")
 
 data class BlockAvailability(
     val fullInfra: FullInfra,
@@ -143,6 +147,12 @@ private fun convertWorkSchedules(
                 val chunkEndOffset = chunkStartOffset + infra.getTrackChunkLength(chunk).distance
                 if (chunkStartOffset > range.end || chunkEndOffset < range.begin) continue
                 val zone = infra.getTrackChunkZone(chunk)
+                if (zone == null) {
+                    blockAvailabilityLogger.info(
+                        "Skipping part of work schedule [${entry.startTime}; ${entry.endTime}] because it is on a track not fully covered by routes: $track",
+                    )
+                    continue
+                }
                 res.add(
                     SpacingRequirement(
                         infra.getZoneName(zone),
--- a/core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/api/InterlockingInfra.kt
+++ b/core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/api/InterlockingInfra.kt
@@ -32,7 +32,7 @@ interface LocationInfra : TrackNetworkInfra, TrackInfra, TrackProperties {
 
     fun getDetectorName(det: DetectorId): String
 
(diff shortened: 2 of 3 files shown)


core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
```diff
@@ -143,6 +147,13 @@ private fun convertWorkSchedules(
                 val chunkEndOffset = chunkStartOffset + infra.getTrackChunkLength(chunk).distance
                 if (chunkStartOffset > range.end || chunkEndOffset < range.begin) continue
                 val zone = infra.getTrackChunkZone(chunk)
+                if (zone == null) {
+                    // TODO: raise a warning
+                    blockAvailabilityLogger.warn(
+                        "Skipping work schedule because it is on a track not fully covered by routes: $track",
```
bougue-pe: I wonder if this really is a problem at that point: skipping occupation on something that can't be occupied anyway (so no wrong planning done using that resource), if I'm correct.
So that warning may be removed (or, as mentioned, replaced by some check at load on infra, but this is more work).
It may also just be downgraded to info, because actually trying to use the zone may be a better sign of something off in data than just no-route track-chunks.

Then (if that log stay), I'd add detail to recover the affected work:
```suggestion
                    blockAvailabilityLogger.info(
                        "Skipping part of work schedule [${entry.startTime}; ${entry.endTime}] because it is on a track not fully covered by routes: $track",
```

bougue-pe (approved):
Thanks for this fix!
It can go as-is IMO, let you judge if suggestions are worthy.

If the cost is reasonable, adding a test on that would be nice (requires a testing infra with missing routes, maybe complicated to get?).

core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
Erashin: You are correct, it can't be occupied anyway, since there are no route for this chunk. However, I'd still prefer to raise a warning, because as a user, I'd rather know why my work schedule got plain out ignored when I tried inputting them without having to search through the code :)
I like your commit suggestion though, it adds more details.

Erashin (commented):
For the tests, ngl, I don't think it's worth it. First off, I'd only want to test `convertWorkSchedules`, but I wanna keep the method private, so I'd have to test `makeBlockAvailability`, which would require a very specific infra setup, since I'd have to check its output (and not just the generated work schedules). All that for a check that i hope will disappear once we look into route generation.

bougue-pe (approved):

bloussou:
will fix : https://github.com/OpenRailAssociation/osrd/pull/7933

Edit (by @Erashin): #7945

Khoyo (approved):

Issue #7135, opened 2024-04-08
Isolate `editoast::modelsv2` into a separate crate `editoast_models`


leovalais:
# D&G

> [!WARNING]
> Blocked by #7031

Move every object interacting with the database into `editoast_models`.

> [!NOTE]
> Don't move ModelV1 into this crate since this will be dropped soon
> after the TrainScheduleV2 backend is fully implemented.

# AC

- [ ] `editoast_models` crate is created within editoast's workspace
- [ ] All models must implement at least, `Debug`, `Serialize`, `Deserialize`, `ToSchema` (and `PartialEq` when it makes sense, likewise for `Eq`)
- [ ] All schemas defined in a file must be included in a `schema!` expression
- [ ] All schemas defined in the crate are available using `editoast_models::schemas()` (generated using `schema!`)
- [ ] All models are moved to `editoast_models` crate
- [ ] `editoast::modelsv2` module is removed
- [ ] No `actix` dependency is present in `editoast_models` (or any other dep that don't belong to the crate)
- [ ] The database queries present in the views sould be moved into `editoast_models` **as much as possible**
- [ ] The part of the editor CRUD that interacts with the database is moved to `editoast_models`, while maintaining a decent API for the views
- [ ] All pertinent unit tests are ported and pass
- [ ] No regression is observed
- [ ] The crate's public API is as minimal as possible
- [ ] The public API is well documented
- [ ] The CI has been adaptated

leovalais:
Done

PR #7939, opened 2024-07-04
editoast: refactor layers get records and app health

Wadjetz:
part of #7135

### DIFF
--- a/editoast/editoast_models/src/db_connection_pool.rs
+++ b/editoast/editoast_models/src/db_connection_pool.rs
@@ -1,18 +1,20 @@
+use std::sync::Arc;
+
+use diesel::sql_query;
 use diesel::ConnectionError;
 use diesel::ConnectionResult;
 use diesel_async::pooled_connection::deadpool::Object;
 use diesel_async::pooled_connection::deadpool::Pool;
-
 use diesel_async::pooled_connection::AsyncDieselConnectionManager;
 use diesel_async::pooled_connection::ManagerConfig;
 use diesel_async::AsyncPgConnection;
+use diesel_async::RunQueryDsl;
 use futures::future::BoxFuture;
 use futures::Future;
 use futures_util::FutureExt as _;
 use openssl::ssl::SslConnector;
 use openssl::ssl::SslMethod;
 use openssl::ssl::SslVerifyMode;
-use std::sync::Arc;
 use url::Url;
 
 #[cfg(feature = "testing")]
@@ -326,6 +328,11 @@ impl DbConnectionPoolV2 {
     pub fn for_tests_no_transaction() -> Self {
         Self::new_for_tests(false)
     }
+}
+
+pub async fn ping_database(conn: &mut DbConnection) -> Result<(), EditoastModelsError> {
+    sql_query("SELECT 1").execute(conn).await?;
+    Ok(())
 }
 
 pub fn create_connection_pool(
--- a/editoast/src/views/mod.rs
+++ b/editoast/src/views/mod.rs
@@ -25,16 +25,22 @@ pub mod work_schedules;
 mod test_app;
 
 use std::ops::DerefMut as _;
+use std::sync::Arc;
+use std::time::Duration;
 
+use futures::TryFutureExt;
 pub use openapi::OpenApiRoot;
 
 use actix_web::get;
 use actix_web::web::Data;
 use actix_web::web::Json;
-use diesel::sql_query;
-use redis::cmd;
+use editoast_derive::EditoastError;
+use editoast_models::ping_database;
+use editoast_models::DbConnectionPoolV2;
 use serde_derive::Deserialize;
 use serde_derive::Serialize;
+use thiserror::Error;
+use tokio::time::timeout;
 use utoipa::ToSchema;
 
 use crate::client::get_app_version;
@@ -49,7 +55,6 @@ use crate::infra_cache::operation;
 use crate::models;
 use crate::modelsv2;
 use crate::RedisClient;
-use editoast_models::DbConnectionPoolV2;
 
 crate::routes! {
     (health, version, core_version),
@@ -95,6 +100,17 @@ editoast_common::schemas! {
     work_schedules::schemas(),
 }
 
+#[derive(Debug, Error, EditoastError)]
+#[editoast_error(base_id = "app_health")]
+pub enum AppHealthError {
+    #[error("Timeout error")]
+    Timeout,
+    #[error(transparent)]
+    Database(#[from] editoast_models::EditoastModelsError),
+    #[error(transparent)]
+    Redis(#[from] redis::RedisError),
+}
+
 #[utoipa::path(
     responses(
         (status = 200, description = "Check if Editoast is running correctly", body = String)
@@ -105,14 +121,25 @@ async fn health(
     db_pool: Data<DbConnectionPoolV2>,
     redis_client: Data<RedisClient>,
 ) -> Result<&'static str> {
-    use diesel_async::RunQueryDsl;
-    sql_query("SELECT 1")
-        .execute(db_pool.get().await?.deref_mut())
-        .await?;
-
-    let mut conn = redis_client.get_connection().await?;
-    cmd("PING").query_async::<_, ()>(&mut conn).await.unwrap();
+    timeout(
+        Duration::from_millis(500),
+        check_health(db_pool.into_inner(), redis_client.into_inner()),
+    )
+    .await
+    .map_err(|_| AppHealthError::Timeout)??;
     Ok("ok")
+}
+
+async fn check_health(
+    db_pool: Arc<DbConnectionPoolV2>,
+    redis_client: Arc<RedisClient>,
+) -> Result<()> {
+    let mut db_connection = db_pool.clone().get().await?;
+    tokio::try_join!(
+        ping_database(db_connection.deref_mut()).map_err(AppHealthError::Database),
+        redis_client.ping_redis().map_err(|e| e.into())
+    )?;
+    Ok(())
 }
 
 #[derive(ToSchema, Serialize, Deserialize)]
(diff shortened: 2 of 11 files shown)


editoast/editoast_models/src/app_health.rs
```diff
@@ -0,0 +1,13 @@
...
+use crate::DbConnection;
+use crate::EditoastModelsError;
+
+pub struct AppHealth;
+
+impl AppHealth {
+    pub async fn run_simple_request(conn: &mut DbConnection) -> Result<usize, EditoastModelsError> {
+        Ok(sql_query("SELECT 1").execute(conn).await?)
+    }
+}
```
leovalais: This probably doesn't need to be encapsulated in a struct. A `ping_database` function would be enough I'd say.

editoast/editoast_models/src/app_health.rs
```diff
@@ -0,0 +1,13 @@
+use diesel::sql_query;
+use diesel_async::RunQueryDsl;
+
+use crate::DbConnection;
+use crate::EditoastModelsError;
+
+pub struct AppHealth;
+
+impl AppHealth {
```
woshilapin: I find it weird to have the concept of "health" in the models. It's not really something about our business logic, nothing related to trains or related objects, it's purely technical. Although, I understand that you're trying to make disappear all traces of DB in `views`. Not entirely sure how to handle that. @leovalais  any idea? I'm available to talk about it.

editoast/src/views/layers/mvt_utils.rs
leovalais, line 1: Not sure about the refactoring in this file. Why move the query somewhere else in `views/` instead of `models/`? Am I missing something?

leovalais (changes requested):

editoast/editoast_models/src/app_health.rs
leovalais: I agree that the health check shouldn't stay in models. I suggested below a `ping_database` function which would `SELECT 1`. This way, when looking at the `/health` endpoint it becomes clear that's everything it does. Wdyt?

leovalais: (We could also `PING` redis btw.)

woshilapin: I wonder if we also should ping Redis as we're on it?

editoast/editoast_models/src/app_health.rs
```diff
@@ -0,0 +1,13 @@
...
+    pub async fn run_simple_request(conn: &mut DbConnection) -> Result<usize, EditoastModelsError> {
+        Ok(sql_query("SELECT 1").execute(conn).await?)
```
flomonster: Ideally, we should set a timeout for this query. Let's say the database takes more than `x` ms to respond the health should respond with a 500. Same for redis.

editoast/editoast_models/src/db_connection_pool.rs
```diff
@@ -320,6 +322,10 @@ impl DbConnectionPoolV2 {
     }
 }
 
+pub async fn ping_database(conn: &mut DbConnection) -> Result<usize, EditoastModelsError> {
```
leovalais:
```suggestion
pub async fn ping_database(conn: &mut DbConnection) -> Result<(), EditoastModelsError> {
```

editoast/editoast_models/src/error.rs
```diff
@@ -12,4 +12,6 @@ pub enum EditoastModelsError {
     #[allow(dead_code)]
     #[error("Test connection not initialized")]
     TestConnection,
+    #[error("Timeout error")]
+    TimeoutError,
```
leovalais: This variant doesn't seem to be constructed within the crate but rather in the views.

editoast/src/views/mod.rs
```diff
@@ -105,14 +109,23 @@ async fn health(
...
+async fn check_health(
+    db_pool: Arc<DbConnectionPoolV2>,
+    redis_client: Arc<RedisClient>,
+) -> Result<()> {
+    ping_database(db_pool.get().await?.deref_mut()).await?;
     let mut conn = redis_client.get_connection().await?;
-    cmd("PING").query_async::<_, ()>(&mut conn).await.unwrap();
-    Ok("ok")
+    cmd("PING").query_async::<_, ()>(&mut conn).await?;
+    Ok(())
```
flomonster: We could join it.

```
join!(ping_database(db_pool.get().await?.deref_mut()), cmd("PING").query_async::<_, ()>(&mut conn));
```

editoast/src/models/layers/mod.rs
```diff
@@ -0,0 +1,5 @@
+mod geo_json_and_data;
+
+pub use geo_json_and_data::create_and_fill_mvt_tile;
+pub use geo_json_and_data::GeoJsonAndData;
+pub use geo_json_and_data::GeoPoint;
```
leovalais:
```suggestion
pub mod geo_json_and_data;
```

wouldn't that be enough?

editoast/src/models/layers/mod.rs
leovalais, line 1: We try to enforce the rust 2018 module naming convention (layer.rs instead of mod.rs) in the new crates

editoast/src/models/mod.rs
leovalais, line 1: likewise

editoast/src/views/mod.rs
```diff
@@ -105,14 +109,23 @@ async fn health(
...
 
+async fn check_health(
+    db_pool: Arc<DbConnectionPoolV2>,
+    redis_client: Arc<RedisClient>,
+) -> Result<()> {
+    ping_database(db_pool.get().await?.deref_mut()).await?;
     let mut conn = redis_client.get_connection().await?;
-    cmd("PING").query_async::<_, ()>(&mut conn).await.unwrap();
-    Ok("ok")
+    cmd("PING").query_async::<_, ()>(&mut conn).await?;
```
leovalais: this could probably have the same timeout and be run at the same time as the ping

leovalais (changes requested):

editoast/src/views/mod.rs
```diff
@@ -105,14 +117,31 @@ async fn health(
...
 
+async fn check_health(
+    db_pool: Arc<DbConnectionPoolV2>,
+    redis_client: Arc<RedisClient>,
+) -> Result<()> {
+    let mut db_connection = db_pool.clone().get().await?;
+    let (_, _) = tokio::join!(
+        ping_database(db_connection.deref_mut()),
+        ping_redis(redis_client)
+    );
```
leovalais: These potential errors should be caught

leovalais (changes requested):

editoast/src/views/mod.rs
```diff
@@ -105,14 +117,31 @@ async fn health(
...
+    Ok(())
+}
+
+async fn ping_redis(redis_client: Arc<RedisClient>) -> Result<()> {
```
woshilapin: Since you moved `ping_database` into `db_connection_pool.rs`, maybe this one can be moved to `redis_utils.rs`? I'd be fine to both have them here, or both in their respective module.

front/public/locales/fr/errors.json
```diff
@@ -195,6 +195,9 @@
     },
     "work_schedule": {
       "NameAlreadyUsed": "Un groupe de planches travaux avec le nom '{{name}}' existe déjà"
+    },
+    "app_health": {
+      "TimeoutError": "Le serveur n'a pas répondu à temps"
```
woshilapin: I believe you'll have to change it also in [front/public/locales/en/errors.json](https://github.com/OpenRailAssociation/osrd/blob/88e6cd365311201b7b7468ef7b2eb9d94b0d1ee5/front/public/locales/en/errors.json)

editoast/src/views/layers/mvt_utils.rs
woshilapin: Especially, this refactorization seems to not be related to the PR. Can we do that in a separate PR? If not, please explain in the description of the commit/PR why we do need that?

front/public/locales/en/errors.json
```diff
@@ -193,6 +193,11 @@
     },
     "work_schedule": {
       "NameAlreadyUsed": "A group of work schedules with '{{name}}' already exists"
+    },
+    "app_health": {
+      "Timeout": "Timeout error",
+      "Database": "Database error",
+      "Redis": "Redis error"
```
woshilapin: Maybe a better translation is expected?

```suggestion
      "Timeout": "Service has not responded in time",
      "Database": "Database is in error",
      "Redis": "Redis is in error"
```

leovalais (approved):
Tested, thanks! 🚀

editoast/src/models/layers/geo_json_and_data.rs
leovalais, line 1: Oops, in my previous comment I suggested to move this file to `models/`, but I meant `modelsv2` (as we'd like to just remove the `models/` folder once TSV1 is gone). Sorry about that! 🫤

editoast/src/views/layers/mvt_utils.rs
woshilapin: I don't understand why this comment has been resolved. I still don't understand why a PR seems to do 2 unrelated things: refactor layers and app health. It feels like it should have been 2 different PR and at the very least 2 different commits. If it's not possible, please explain why in the commit description (and possibly also in the PR description).

leovalais: Agreed. I missed this point in my first reviews, thanks for insisting on it.

Given that the layer refactoring has already been reviewed and doesn't seem need further modifications, I'd favor dividing the commit in two (but I'm fine with both, +1 about the commit description).

woshilapin: Thank you very much for the split in commits 🎉

woshilapin (approved):

clarani (approved):

PR #7952, opened 2024-07-04
editor: improves switch types cache handling

### LINKED ISSUE #7946: front: editor: enable to select switch for speed restriction by routes
### What happened?

See:

https://github.com/OpenRailAssociation/osrd/assets/32448554/a19a29f5-f51b-41de-b49a-0c93b1c98b46

_No response_

See the video

- **Browser**: Firefox 

staging dc11fc8

jacomyal:
This commit fixes #7946.

Details:
- Adds a `?` to handle case where switch types have not been loaded yet in SwitchList.tsx
- Adds a very simple cache to `useSwitchTypes` to avoid reloading switch types any time a new component mounts with that hook

### DIFF
--- a/front/src/applications/editor/tools/switchEdition/useSwitchTypes.ts
+++ b/front/src/applications/editor/tools/switchEdition/useSwitchTypes.ts
@@ -6,7 +6,7 @@ import { osrdEditoastApi } from 'common/api/osrdEditoastApi';
 
 import type { SwitchType } from './types';
 
-// Client prefered order
+// Client preferred order
 const trackNodeTypeOrder = [
   'link',
   'point_switch',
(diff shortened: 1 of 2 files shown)


front/src/applications/editor/tools/switchEdition/useSwitchTypes.ts
```diff
@@ -15,30 +15,47 @@ const trackNodeTypeOrder = [
   'double_slip_switch',
 ];
 
+let switchTypesCache: Record<number, SwitchType[]> = {};
+
 export default function useSwitchTypes(infraID: number | undefined) {
-  const [data, setData] = useState<SwitchType[]>([]);
+  const [data, setData] = useState<SwitchType[]>(
+    !isNil(infraID) ? switchTypesCache[infraID] || [] : []
```
Yohh, line 22: I m not sure to understand, if true why adding a new condition to possibly set an empty array?

jacomyal: Here is what this does:
1. `data` must always be an array
2. If `infraID` is nil, then `data` is initially set to `[]`
3. Else, we check if we have a value in the cache that has already been loaded. If we have some, we set `data` to this value (`switchTypesCache[infraID]`)
4. Else, we set `data` to `[]`

@Yohh Does that answer your question?

Yohh: yes thanks

Yohh (approved):
LGTM tested

clarani (approved):

Issue #6685, opened 2024-02-20
editoast: better way to infer/compute error context type


sim51:
### Description and goal

In the error derive, we compute the type of error's context ( ie list its properties and their type).

For now the code is pretty simple and do manually the cast between simple rust type and openapi.
You can check the function `fn error_context_to_openapi_object`  in `src/view/mod.rs`  for example. 

A better way can be found to do it, specially if the error derive if we can save in inventory something that be directly given to utoapi. 

### Acceptance criteria

A better way to compute context type which support more types with less code

sim51:
Some hints : 
- https://github.com/rust-lang-nursery/lazy-static.rs
- https://docs.rs/utoipa/latest/utoipa/macro.schema.html

PR #7953, opened 2024-07-04
core: change pathfinding errors on rolling-stock constraints

### LINKED ISSUE #7719: Return list of incompatible RS constraints in path finding (back)
Back part of https://github.com/osrd-project/osrd-confidential/issues/309

bougue-pe:
Change interface for pathfinding errors on rolling-stock constraints and project RS constraints on the relaxed path.
Not available for STDCM, only for operational studies.

⚠️ this breaks core interface and will be handled in another PR by editoast.
So: to be squashed with later work.

TODO:
- [x] test with holes between blocked ranges
- [x] test incompatible gauge
- [x] check if validate PR response is/should be done

Fix: #7719


core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlockResponse.kt
```diff
@@ -39,41 +41,21 @@ class NotFoundInRoutes(
...
-open class IncompatibleConstraint(
-    val blocks: List<String>,
-    val routes: List<String>,
-    @Json(name = "track_section_ranges") val trackSectionRanges: List<TrackRange>,
-    val length: Length<Path>,
-    @Json(name = "incompatible_ranges")
-    val incompatibleRanges: List<List<Offset<Path>>>, // List of pairs
+class IncompatibleConstraintsPathResponse(
+    @Json(name = "relaxed_constraints_path") val relaxedConstraintsPath: PathfindingBlockSuccess,
+    val constraints: IncompatibleConstraints
```
Erashin: I'd rather just name it `incompatibleConstraints`.

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingV2Test.kt
```diff
@@ -96,6 +101,7 @@ class PathfindingV2Test : ApiTest() {
                 .act(RqFake("POST", "/v2/pathfinding/blocks", requestBody))
         val response = TakesUtils.readBodyResponse(rawResponse)
         val parsed =
-            (pathfindingResponseAdapter.fromJson(response) as? IncompatibleElectrification)!!
+            (pathfindingResponseAdapter.fromJson(response)
+                as? IncompatibleConstraintsPathResponse)!!
```
Erashin, line 124: You could check right now that the `incompatibleElectrifiations` is equal to the default value you put in the endpoint (that way it'll fail when you actually put the correct values later on).

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlockResponse.kt
bougue-pe: Done (amended commit): https://github.com/OpenRailAssociation/osrd/compare/0a2d394be4e8afa67d5252bcab9022c61100a769..f0a49f08065c5926544b0fa61ada07b9be7e6958#diff-2322092101a8cb6588d8c9145f722ce6338e7bb9b0139d2d30bc02fcf042ad63R64

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingV2Test.kt
bougue-pe: Indeed, I started to improve test on this (amended commit): https://github.com/OpenRailAssociation/osrd/compare/0a2d394be4e8afa67d5252bcab9022c61100a769..f0a49f08065c5926544b0fa61ada07b9be7e6958

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlockResponse.kt
sim51: The JSON name should also be  `incompatible_constraints` : 

```
@Json(name = "incompatible_constraints") val constraints: IncompatibleConstraints
```

bougue-pe: Indeed, thanks, amended again the first commit : https://github.com/OpenRailAssociation/osrd/compare/f0a49f0..c783c53

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -144,69 +145,56 @@ private fun computePaths(
...
-                            IncompatibleConstraintsPathResponse(
-                                relaxedPathResponse,
-                                IncompatibleConstraints(
-                                    listOf(),
-                                    listOf(Pathfinding.Range(Offset.zero(), Offset.zero())),
-                                    listOf()
-                                )
-                            )
-                        )
+    if (
```
Erashin, line 203: This looks like it deserves its own method now: `buildIncompatibleConstraintsResponse` or something ?

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -144,69 +145,56 @@ private fun computePaths(
...
+                    val range =
+                        Pathfinding.Range(
+                            travelledPathBlockStart +
+                                Offset.max(blockedRange.start, range.start).distance,
+                            travelledPathBlockStart +
+                                Offset.min(blockedRange.end, range.end).distance
                         )
+                    when (currentConstraint::class.java) {
+                        ElectrificationConstraints::class.java -> {
+                            elecRanges.add(RangeValue(range, "elec"))
```
Erashin: I don't remember if it was discussed in the ticket, but this means we potentially return ranges of constraints that could be merged, i.e.: [100, 200, "15000V], [200, 300, "elec"] => [100, 300, "elec"]. I feel like it would be useful to merge them when possible, but I'll let you check if it's required and a change we'd want.

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -144,69 +145,56 @@ private fun computePaths(
...
+        val elecRanges = mutableListOf<RangeValue<String>>()
+        val gaugeRanges = mutableListOf<Pathfinding.Range<TravelledPath>>()
+        val signalRanges = mutableListOf<RangeValue<String>>()
+        var travelledPathBlockStart =
+            Offset<TravelledPath>(
+                -possiblePathWithoutErrorNoConstraints.ranges.first().start.distance
+            )
+        for (range in possiblePathWithoutErrorNoConstraints.ranges) {
+            for (currentConstraint in constraints) {
+                for (blockedRange in currentConstraint.apply(range.edge)) {
```
Erashin: I wonder if it wouldn't be better if `PathfindingConstraint` could manage multiple edges
(i.e. `apply(ranges.map { it.edge })`) or something like that.
EDIT: good job, you already did it in the following commit with `ExplainedPathfindingConstraint`.

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingV2Test.kt
```diff
@@ -134,7 +134,20 @@ class PathfindingV2Test : ApiTest() {
             parsed.incompatibleConstraints ==
                 IncompatibleConstraints(
                     incompatibleElectrificationRanges =
-                        listOf(RangeValue(Pathfinding.Range(Offset.zero(), Offset.zero()), "elec")),
+                        listOf(
+                            RangeValue(
```
Erashin, line 131: Throwback to previous comment, feels a bit scuffed to not just have one unique `RangeValue` going from 125 to 10250.

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
bougue-pe: Done in fully amended second commit https://github.com/OpenRailAssociation/osrd/pull/7953/commits/8ee6553e87735672c30538f445ee6e66d877360e

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
bougue-pe: Yes merging is handled by the `DistanceRangeMap` in https://github.com/OpenRailAssociation/osrd/pull/7953/commits/8ee6553e87735672c30538f445ee6e66d877

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
bougue-pe: Actually I dropped the `ExplainedPathfindingConstraint` interface in amended https://github.com/OpenRailAssociation/osrd/pull/7953/commits/8ee6553e87735672c30538f445ee6e66d877
It may be discussed, but what we want to display is not always directly linked to the constraint, and info are not always available in the constraint itself.
So I kept it into `buildIncompatibleConstraintsResponse()`.

I resolve this and let you reopen an issue on amended commit if needed.

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingV2Test.kt
bougue-pe: Indeed, it was just a first try, better in amended https://github.com/OpenRailAssociation/osrd/pull/7953/commits/8ee6553e87735672c30538f445ee6e66d877

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -138,63 +147,192 @@ private fun computePaths(
...
+    infra: FullInfra,
+    blockList: List<BlockId>
+): DistanceRangeMap<String> {
+    val pathSignalingSystem = distanceRangeMapOf<String>()
+    var startBlockPathOffset = Distance.ZERO
+    for (block in blockList) {
+        val blockLength = infra.blockInfra.getBlockLength(block).distance
+        val blockSignalingSystemIdx = infra.blockInfra.getBlockSignalingSystem(block)
+        val blockSignalingSystemName =
+            infra.signalingSimulator.sigModuleManager.getName(blockSignalingSystemIdx)
```
bougue-pe, line 320: @Khoyo: not totally confident that this (conbined with [the add of `getName` to the interface](https://github.com/OpenRailAssociation/osrd/pull/7953/files#diff-481923950eaa8c032e916301d46efc041482a6364a59cfc1bab17d9019c33930R54)) is the best way to get a readable name for signaling system on a block. Your opinion is welcome on that 🙂

core/kt-osrd-utils/src/main/kotlin/fr/sncf/osrd/utils/DistanceRangeMap.kt
```diff
@@ -95,3 +96,20 @@ fun <T> mergeDistanceRangeMaps(
     // Build the whole map at once to avoid redundant computations.
     return distanceRangeMapOf(resEntries)
 }
+
+/**
+ * Filters the 'filtered' map, keeping only ranges also present in 'filter' map (values from
+ * 'filter' map are not considered)
+ */
+fun <T, R> filterIntersection(
+    filtered: DistanceRangeMap<T>,
```
Erashin: `mapToFilter`? which would allow you to rename `rangeFirst` into `filtered`?

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -138,79 +147,192 @@ private fun computePaths(
...
+    // It didn’t fail due to an incompatible constraint, no path exists
+    throw NoPathFoundException(NotFoundInBlocks(listOf(), Length(0.meters)))
+}
+
+private fun buildIncompatibleConstraintsResponse(
+    infra: FullInfra,
+    waypoints: ArrayList<Collection<PathfindingEdgeLocationId<Block>>>,
+    constraints: List<PathfindingConstraint<Block>>,
+    remainingDistanceEstimators: List<AStarHeuristicId<Block>>,
+    timeout: Double?
```
Erashin, line 190: Not sure we need the timeout. Timeout is used to prevent the algorithm to compute for too much time when it doesn't find a solution. Since we've arrived here, it means the constrained pathfinding has already taken less than 2min. As a result, there's no reason the unconstrained pathfinding should take more than 2min.

Edit: might be fine to actually keep the timeout here, but having it be only 2min - basePathfindingTime is harsh.

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -138,79 +147,192 @@ private fun computePaths(
     }
 
     // Handling errors
-    // Check if pathfinding failed due to constraints
+    // Check if pathfinding failed due to incompatible constraints
+    val elapsedSeconds = Duration.between(start, Instant.now()).toSeconds()
```
Erashin, line 158: I really don't like this. I see why you're doing it likeso: you want the pathfinding to timeout if constrained pathfinding + unconstrained pathfinding > 2min. But the whole point of the timeout was to interrupt requests that would take way too much time to find a solution. Here, we already know the constrained is <2min, so the unconstrained will take even less time. We risk throwing a timeout exception instead of waiting 10 more seconds to have more details about why the pathfinding failed, which is the goal of this feature.

Edit: same as above, unconstrained could take longer (more edges to explore). Question remains: do we want the details (goal of this feature), or timeout anyway after 2min (probably will need a different error to specify it timedout looking for the details, and there's indeed no path found).

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -138,79 +147,192 @@ private fun computePaths(
...
+        runPathfindingPostProcessing(infra, possiblePathWithoutErrorNoConstraints),
+        IncompatibleConstraints(
+            elecBlockedRangeValues,
+            gaugeBlockedRanges,
+            signalingSystemBlockedRangeValues
+        )
+    )
+}
+
+private fun getElecBlockedRangeValues(
```
Erashin: You could probably do only one method here, 3 methods takes a lot of place not gonna lie (and it also means we'd have to change all 3 of em' if we need to remodify this code). Would only require to precompute the `pathProps` and pass the corresponding DistanceRangeMap as an argument (which also means we don't compute it 3 times, yay), and change the output: either keep it as is and do a condition for electrification, or just return a `DistanceRangeMap` and postprocess it (both are fine, but I feel like the first one would have less code and doesn't make the method too complex). What do you think?

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/constraints/ElectrificationConstraints.kt
```diff
@@ -6,7 +6,11 @@ import com.google.common.collect.TreeRangeSet
 import fr.sncf.osrd.api.pathfinding.makePathProps
 import fr.sncf.osrd.graph.Pathfinding
 import fr.sncf.osrd.graph.PathfindingConstraint
-import fr.sncf.osrd.sim_infra.api.*
+import fr.sncf.osrd.sim_infra.api.Block
```
Erashin: When I got on the project, didn't much like the `import *` either. I changed it to 5 max to be coherent with everyone else and stop always modifying all the imports whenever I'd change a file (and someone else with different settings would remodify it as well).

core/kt-osrd-utils/src/main/kotlin/fr/sncf/osrd/utils/DistanceRangeMap.kt
bougue-pe: Done in https://github.com/OpenRailAssociation/osrd/pull/7953/commits/fe3b77920cf0753496d31eed0e1c0517fd9d2ffd

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
bougue-pe: Thanks for this, I improved this in https://github.com/OpenRailAssociation/osrd/pull/7953/commits/fe3b77920cf0753496d31eed0e1c0517fd9d2ffd

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/constraints/ElectrificationConstraints.kt
bougue-pe: Done in https://github.com/OpenRailAssociation/osrd/pull/7953/commits/fe3b77920cf0753496d31eed0e1c0517fd9d2ffd on the files of this PR

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
eckter: > Since we've arrived here, it means the constrained pathfinding has already taken less than 2min. As a result, there's no reason the unconstrained pathfinding should take more than 2min.

It can, the unconstrained pathfinding has more edges to explore.

bougue-pe: After first discussions: we want to be able to distinguish a timeout using initial (real) constraints and a timeout unconstrained if we have an "unconstrained" timeout.
Now the question is (@bloussou ?): What timeout do we want on unconstrained PF (launched to "explain" failure) ?
1. shared with the constrained PF
2. equal to the constrained PF
3. no timeout

bloussou: Is it triggered for each failing PF ?

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
Khoyo: LGTM ;)

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
Erashin: Yes. For each failing PF (except if it takes >2min and fails due to timeout), the unconstrained PF is triggered, to find out why it failed (ex: due to electrification constraints), the explanation is returned and will be displayed in the front end of the app (back end part of this feature is this PR).
If your question was if the timeout was triggered for every failing PF, the answer is no, it depends.

bougue-pe: We may not care about explaining PF failure (RS constraints) when using the timeout (BASIC import).
It would let more freedom and maybe aim at a quicker fail, so a mutualized timeout?

bloussou: okay lets go for a mutualized timeout I think

bougue-pe: OK, so the timeout handling didn't change, but a dedicated error type was added in https://github.com/OpenRailAssociation/osrd/pull/7953/commits/9a63a73b0cfadf86020072793530da2eb6b8e6e5

Erashin (approved):

front/public/locales/en/operationalStudies/manageTrainSchedule.json
```diff
@@ -60,6 +60,8 @@
   "maximumRunTime": "Maximum running time",
   "maximumRunTimeError": "Travel time must not exceed 12h",
   "No path could be found": "No path found",
+  "Pathfinding timed out: no path could be found": "Time limit exceeded, no path found",
+  "Relaxed pathfinding timed out: no path could be found, then relaxed path search timed out": "No path found, then exceeded time limit while searching unconstrained path",
```
Erashin: I didn't know we couldn't/weren't using keys for error matching in the front end (here, we do send `no_path_found:relaxed_path_timeout`), and instead use the entire string message sent from the back. I thought that was the plan. I'm a bit sad.

eckter: That's weird, it does work with error codes in other parts. Especially here: https://github.com/OpenRailAssociation/osrd/blob/dev/front/public/locales/en/errors.json#L11-L14

I guess it's not related to this specific PR, but something is off here

SharglutDev: How come errors key are not generated in errors.json here ?

If you think it can't be fixed in this PR, we should at least use a proper syntax in camelCase for these new keys, the other surrounding keys seem to come from a very old PR and we don't use this syntax anymore.

bougue-pe: To be discussed (probably with @sim51) but after digging a bit on that, this translation is probably not used (could trigger the timeout, but not translate it in front).

Here are some responses obtained from `api/v2/infra/<x>/pathfinding/blocks`:
```json
{"status":"pathfinding_failed","core_error":{"status":500,"type":"core:no_path_found:timeout","context":{},"message":"Pathfinding timed out: no path could be found"}}
{"status":"pathfinding_failed","core_error":{"status":500,"type":"core:no_path_found:relaxed_path_timeout","context":{},"message":"Relaxed pathfinding timed out: no path could be found, then relaxed path search timed out"}}
```
What's displayed in Operational studies PF is `An error occurred in pathfinding: No path found.` (possibly translated in 🇫🇷 if language configured).

Those error subtypes are useful for debugging and batch on endpoint directly, not for front (this may be OK to let it like that).

So my proposition would be:
* remove those lines (maybe also the one above)
* not dig more on this now, but maybe delegate it to another issue depending on the goal (https://github.com/OpenRailAssociation/osrd/issues/6685 ?)

Khoyo: The whole editoast error system should be rewritten anyway (necessary for the crate split).

As for the matching on strings and not keys, after a little digging through the git history I'm pretty sure it's just an historical remnant from darker times. We used to directly reply this string as a response...

shenriotpro (approved):
OK for tests/

front/public/locales/en/operationalStudies/manageTrainSchedule.json
sim51: Keys are used for errors defined in editoast. But here, we are with a core error which is wrapped into an editoast error (moreover for editoast, core's errors are dynamics).  
That's why we can't use the same pipe as the one defined for editoast. 
Definitively, we should do something about that on the frontend, but it's a generic task about how to handle core's error on the frontend

sim51 (approved):

SharglutDev (approved):

front/public/locales/en/operationalStudies/manageTrainSchedule.json
bougue-pe: After discussion with @sim51 & @SharglutDev:
* removed the outdated translations: https://github.com/OpenRailAssociation/osrd/pull/7953/commits/f62952842dad7d5e0a5b10c8db9be227a61ed668
* created a dedicated issue for the handling of those errors: https://github.com/OpenRailAssociation/osrd/issues/8246 (please fill in more info if relevant)

So I'm resolving the (and will merge to the dedicated branch probably by noon): please don't hesitate to ask for more as this is not yet into `dev` there is some time to improve.

PR #7958, opened 2024-07-05
editoast: add telemetry to core requests

woshilapin:
We don't have any information about how long it takes for each request to `core`. This telemetry should give us some information about it.


flomonster (approved):
LGTM. (Not tested)

PR #7961, opened 2024-07-05
front: set timezone for test

sim51:
Tests are failing locally because I'm not in the UTC timezone. So tests are plateform dependent, which is bad.
This commit sets the timezone when runnning vitest.


sim51:
Tested on linux & windows with WSL (thanks Sarah)

SharglutDev (approved):

PR #7962, opened 2024-07-05
Add xmas logo :santa: 

flomonster:
Winter is coming.


front/src/main/home.jsx
```diff
@@ -9,14 +9,20 @@ import operationalStudiesImg from 'assets/pictures/home/operationalStudies.svg';
...
 export default function Home() {
   const { t } = useTranslation('home/home');
 
-  const logo = new Date().getMonth() === 5 ? proudLogo : defaultLogo;
+  let logo = defaultLogo;
+  if (new Date().getMonth() === 5) {
+    logo = proudLogo;
+  } else if (new Date().getMonth() === 11) {
+    logo = xmasLogo;
+  }
```
woshilapin: How about doing this so we don't mixup the numbers (5 is June, really?).

```suggestion
  // 0-based months
  const JUNE = 5;
  const DECEMBER = 11;
  if (new Date().getMonth() === 5) {
    logo = proudLogo;
  } else if (new Date().getMonth() === 11) {
    logo = xmasLogo;
  }
```

woshilapin (approved):
Just missing the "save the planet" one and we should be good! 😝

theocrsb (approved):

front/src/main/home.jsx
SharglutDev: In the rest of the front, we would do something like this (you can move the const outside the `Home` function at l.16) : 

```suggestion
  enum MONTH_VALUES {
    JUNE: 5,
    DECEMBER: 11
  }
  
  if (new Date().getMonth() === MONTH_VALUES.JUNE) {
    logo = proudLogo;
  } else if (new Date().getMonth() === MONTH_VALUES.DECEMBER) {
    logo = xmasLogo;
  }
```

SharglutDev (approved):
Lgtm !

woshilapin (approved):

woshilapin (approved):

PR #7968, opened 2024-07-05
Bulk get cached projection

flomonster:
## Description

The idea is to bulk-get the cached projection to limit the number of queries performed to Redis.

## Benchmark

Condition:
- I performed the test when everything was cached.
- It was tested locally
- Timetable size: 823 trains
- Valid trains: 674

The `path_projection` performance doesn't seem to change (1.2s).

Queries performances:
- Before: 1 query took ~0,07ms (times 674 = 47ms)
- Now: 1 bulk query takes ~8ms

I think in production, because of queries overhead the difference will be far more important.

The number of Redis queries is reduced by the number of valid train schedule. (2211 -> 1532 queries).
Next, I will do the same to retrieve the path and simulation from the cache (but this implies code refactoring).

> [!IMPORTANT]
> There is a limitation with this. I didn't find how to set the expiration time of Redis entry with a bulk command. So now the Redis entry expires after 1 week even if they're used recently.

### DIFF
--- a/editoast/src/redis_utils.rs
+++ b/editoast/src/redis_utils.rs
@@ -77,6 +77,31 @@ impl RedisConnection {
         }
     }
 
+    /// Get a list of deserializable value from redis
+    #[tracing::instrument(name = "cache:get_bulk", skip(self), err)]
+    pub async fn json_get_bulk<T: DeserializeOwned, K: Debug + ToRedisArgs + Send + Sync>(
+        &mut self,
+        keys: &[K],
+    ) -> Result<Vec<Option<T>>> {
+        // Avoid mget to fail if keys is empty
+        if keys.is_empty() {
+            return Ok(vec![]);
+        }
+        let values: Vec<Option<String>> = self.mget(keys).await?;
+        values
+            .into_iter()
+            .map(|value| match value {
+                Some(v) => match serde_json::from_str::<T>(&v) {
+                    Ok(value) => Ok(Some(value)),
+                    Err(_) => {
+                        Err(RedisError::from((ErrorKind::TypeError, "Expected valid json")).into())
+                    }
+                },
+                None => Ok(None),
+            })
+            .collect()
+    }
+
     /// Get a deserializable value from redis with expiry time
     #[tracing::instrument(name = "cache:get_with_expiration", skip(self), err)]
     pub async fn json_get_ex<T: DeserializeOwned, K: Debug + ToRedisArgs + Send + Sync>(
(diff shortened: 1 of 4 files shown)


flomonster:
> There is a limitation with this. I didn't find how to set the expiration time of Redis entry with a bulk command. So now the Redis entry expires after 1 week even if they're used recently.

I found a solution. We can use a transaction or pipe to execute all the commands simultaneously. Example with the CLI:

```
$ MSET hey 42 hi 43
$ MULTI
$ EXPIRE hey 10
$ EXPIRE hi 10
$ EXEC
```

editoast/src/redis_utils.rs
```diff
@@ -77,6 +77,27 @@ impl RedisConnection {
         }
     }
 
+    /// Get a list of deserializable value from redis
+    #[tracing::instrument(name = "cache:get_bulk", skip(self), err)]
+    pub async fn json_get_bulk<T: DeserializeOwned, K: Debug + ToRedisArgs + Send + Sync>(
+        &mut self,
+        keys: K,
```
woshilapin: I'm a bit disturbed by the way this API is defined. Since it's a bulk, I would expect to find a collection here, not just a `K`. So how about changing this line to something like `&[K]`? It's a bit more restrictive but express in a better way the API? I believe it works both ways because `impl<T: ToRedisArgs> ToRedisArgs for &[T] {}`.

flomonster:
A better solution is to set up an `allkeys-lru` policy to let Redis handle this automatically.

editoast/src/views/v2/train_schedule/projection.rs
```diff
@@ -245,15 +245,32 @@ async fn project_path(
...
+            miss_cache.insert(train_id, train_details.clone());
         }
     }
+
+    // Reset expire time of hit cache
+    let mut pipe = redis::pipe();
+    for id in hit_cache.keys() {
+        pipe.expire(&trains_hash_values[id], CACHE_PROJECTION_EXPIRATION as i64);
+    }
+    pipe.query_async(&mut redis_conn).await?;
```
woshilapin: I feel like this is mostly an implementation detail, nothing really related with the endpoint's logic. Maybe it'd be better to put that away in `redis_utils`? Ideally, I'd say that business logic should just care about an available storage to get objects. The fact that these objects are backed up by a cache solution is not the concern of this function.

Castavo: I agree.
Maybe this could be a part of the `json_get_bulk` function ?

But if we move to redis's allkeys-lru strategy, is there still a need to manage expiration times ?

flomonster: > But if we move to redis's allkeys-lru strategy, is there still a need to manage expiration times ?

@ElysaSrc helped me setup the policy. So we don't need to manage this anymore :tada:

ghost (commented):
Reviewed redis part, lgtm

ghost (approved):

woshilapin (approved):
Thanks for the nice improvements 👍

PR #7971, opened 2024-07-07
editoast: train-simulation: return SimulationFailed on 500 from core

Castavo:
When `core.fetch` returns an Error made from a 500 from core, `train_simulation` returns `Ok(SimulationResponse::SimulationFailed)` rather than an `Error`.

This allows to still be able to call `train_schedule/simulation_summary` when one of the trains leads to an error in core (an assert error for example).

I left other error status codes fail because they'll likely be transient, what do you think ?


Castavo:
Would not work since 500 errors could come from core but also from diesel for example...

PR #7980, opened 2024-07-08
front: change margin logic

### LINKED ISSUE #7950: TSV2 : incorrect standard margin distribution in case of specified passing point
### What happened?

When we put a value of standard margin in the input table, the correspondent margin with this value is inserted till the following passing point 

When we put a value of standard margin in the input table, the correspondent margin with this value should be inserted till the following point where a new value of standard margin is specified  

Go to a scenario TSV2
Chose the rolling stock and a path with a passing point 
Put 5% value of standard margin
Look the margin distribution in to the speed space graph

- **Browser**: Edge
- **OS**: Windows
- **Env**: Recette

staging 4698a66

theocrsb:
close #7950

### DIFF
--- a/front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
+++ b/front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
@@ -1,21 +1,31 @@
 import type { Margin } from 'modules/trainschedule/components/ManageTrainSchedule/types';
 import type { PathStep } from 'reducers/osrdconf/types';
 
-const formatMargin = (pathSteps: PathStep[]): Margin => {
-  const margins: Margin = {
-    boundaries: [],
-    values: [],
-  };
+const formatMargin = (pathSteps: PathStep[]): Margin | undefined => {
+  const boundaries: string[] = [];
+  const values: string[] = [];
 
   pathSteps.forEach((step, index) => {
     if (index === 0) {
-      margins.values.push(step.theoreticalMargin || 'none');
-    } else if (step.theoreticalMargin !== pathSteps[index - 1].theoreticalMargin) {
-      margins.boundaries.push(step.id);
-      margins.values.push(step.theoreticalMargin || 'none');
+      values.push(step.theoreticalMargin || 'none');
+    } else if (index === pathSteps.length - 1) {
+      if (values.length === 1 && values[0] !== 'none') {
+        boundaries.push(step.id);
+        values.push('none');
+      }
+    }
+
+    // for the other steps, we add the margin if it's different from the previous one
+    else if (step.theoreticalMargin && step.theoreticalMargin !== values[values.length - 1]) {
+      boundaries.push(step.id);
+      values.push(step.theoreticalMargin);
     }
   });
-  return margins;
+
+  if (values.length === 1) {
+    return undefined;
+  }
+  return { boundaries, values };
 };
 
 export default formatMargin;


front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
```diff
@@ -1,20 +1,44 @@
...
+    // for the last step, if the origin has a margin and there isn't any via with one,
+    // we add 'none' add the end of the values because we need N+1 values compared to boundaries
+    else if (index === pathSteps.length - 1 && pathSteps[0].theoreticalMargin && !viaWithMargin) {
+      margins.boundaries.push(step.id);
+      margins.values.push('none');
     }
   });
+
+  // check if all margins are 'none'
+  if (!margins.values.some((value) => value !== 'none')) {
```
anisometropie: A bit easier without the 2 negations

```suggestion
  if (margins.values.every((value) => value === 'none')) {
```

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
```diff
@@ -1,20 +1,44 @@
...
+    ) {
+      viaWithMargin = true;
       margins.boundaries.push(step.id);
-      margins.values.push(step.theoreticalMargin || 'none');
+      margins.values.push(step.theoreticalMargin);
+    }
+
+    // for the last step, if the origin has a margin and there isn't any via with one,
+    // we add 'none' add the end of the values because we need N+1 values compared to boundaries
+    else if (index === pathSteps.length - 1 && pathSteps[0].theoreticalMargin && !viaWithMargin) {
```
anisometropie: you probably don’t need `viaWithMargin`, you never enter this `else if` if you went in the first `else if`

anisometropie (changes requested):

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
theocrsb: Yes, that works too. Done

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
theocrsb: Done

Uriel-Sautron (approved):
Tested nice !! Thank you

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
```diff
@@ -1,20 +1,40 @@
 import type { Margin } from 'modules/trainschedule/components/ManageTrainSchedule/types';
 import type { PathStep } from 'reducers/osrdconf/types';
 
-const formatMargin = (pathSteps: PathStep[]): Margin => {
+const formatMargin = (pathSteps: PathStep[]): Margin | undefined => {
   const margins: Margin = {
     boundaries: [],
     values: [],
   };
```
clarani: You can do this, I think it's more elegant :)
```suggestion
  const boundaries: string[] = [];
  const values: string[] = [];
```
Then you just have to update `boundaries` or `values`, and not `margins.boundaries` or `margins.values`

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
```diff
@@ -1,20 +1,40 @@
...
-      margins.values.push(step.theoreticalMargin || 'none');
+      margins.values.push(step.theoreticalMargin);
+    }
+
+    // for the last step, if the origin has a margin and there isn't any via with one,
+    // we add 'none' add the end of the values because we need N+1 values compared to boundaries
+    else if (index === pathSteps.length - 1 && pathSteps[0].theoreticalMargin) {
+      margins.boundaries.push(step.id);
+      margins.values.push('none');
     }
```
clarani: Move the case of last element just below the case `index === 0`, and instead of testing that the first pathSteps has a theoreticalMargin, you can check that `value` has only 1 element and that it is not 'none'
```suggestion
   else if (index === pathSteps.length - 1) {
      if (values.length === 1 && values[0] !== 'none') {
        boundaries.push(step.id);
        values.push('none');
      }
    }
```

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
```diff
@@ -1,20 +1,40 @@
...
       margins.values.push(step.theoreticalMargin || 'none');
-    } else if (step.theoreticalMargin !== pathSteps[index - 1].theoreticalMargin) {
+    }
+
+    // for the other steps, we add the margin if it's different from the previous one
+    else if (
+      step.theoreticalMargin &&
+      step.theoreticalMargin !== pathSteps[index - 1].theoreticalMargin &&
+      index !== pathSteps.length - 1
+    ) {
```
clarani: - since you moved the case of the last element above, you can remove the test `index !== pathSteps.length - 1`
- you should compare `step.theoreticalMagin` with the last value, and not the theoreticalMargin of the last pathStep (it could be undefined)
```suggestion
     else if (step.theoreticalMargin && step.theoreticalMargin !== values[values.length - 1]) {
```

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
```diff
@@ -1,20 +1,40 @@
...
+    // for the last step, if the origin has a margin and there isn't any via with one,
+    // we add 'none' add the end of the values because we need N+1 values compared to boundaries
+    else if (index === pathSteps.length - 1 && pathSteps[0].theoreticalMargin) {
+      margins.boundaries.push(step.id);
+      margins.values.push('none');
     }
   });
+
+  // check if all margins are 'none'
+  if (margins.values.every((value) => value === 'none')) {
```
clarani: You can simply check if the `values.length !== 1`
```suggestion
  if (values.length === 1) {
```

clarani (changes requested):

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
SharglutDev: It's easier to read but I think the point was not to have to iterate in the whole array to check this condition (as soon as we see a not "none" value, we can stop the loop)

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
theocrsb: Done

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
theocrsb: Done

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
theocrsb: Done

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
theocrsb: Done

clarani (approved):
LGTM, approved :white_check_mark:

anisometropie (approved):

PR #7981, opened 2024-07-08
Role-based authorization system

leovalais:
> [!IMPORTANT]
> Tracking issue: #8171 

# Implementation plan

- [X] diesel migration
- [X] Builtin role definition
- [X] Role config parsing
- [X] `struct Authorizer`
- [X] User registration using request headers
- [X] Role fetching from db
- [X] Role checking
- [X] Test using one endpoint (/health)
- [X] Role CRD operations
- [x] Rewrite the `PgAuthDriver` using `diesel::dsl`
- [x] Unit tests
    - [x] Role config parsing (in progress)
    - [x] Role checking function
    - [x] `PgAuthDriver`

# Finitions

- [x] Review naming
- [x] `mv modelsv2/authn.rs modelsv2/auth.rs`
- [ ] Documentation of the definitive parts of the API
- [x] Try to reduce (within the scope of the PR ofc) the truckload of new `TODO:` comments

### DIFF
--- a/editoast/editoast_authz/Cargo.toml
+++ b/editoast/editoast_authz/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "editoast_authz"
+license.workspace = true
+version.workspace = true
+edition.workspace = true
+
+[features]
+fixtures = []
+
+[dependencies]
+serde = { workspace = true, features = ["derive"] }
+strum.workspace = true
+thiserror.workspace = true
+toml_edit = { version = "0.22.16", features = ["parse", "serde"] }
+tracing.workspace = true
+
+[dev-dependencies]
+pretty_assertions.workspace = true
+tokio.workspace = true
--- a/editoast/editoast_authz/src/authorizer.rs
+++ b/editoast/editoast_authz/src/authorizer.rs
@@ -1,0 +1,294 @@
+use std::{collections::HashSet, future::Future};
+
+use tracing::debug;
+
+use crate::roles::{BuiltinRoleSet, RoleConfig};
+
+pub type UserIdentity = String;
+pub type UserName = String;
+
+#[derive(Debug)]
+pub struct UserInfo {
+    pub identity: UserIdentity,
+    pub name: UserName,
+}
+
+pub struct Authorizer<'config, S: StorageDriver> {
+    user: UserInfo,
+    user_id: i64,
+    roles_config: &'config RoleConfig<S::BuiltinRole>,
+    user_roles: Option<HashSet<S::BuiltinRole>>,
+    storage: S,
+}
+
+pub trait StorageDriver {
+    type BuiltinRole: BuiltinRoleSet + std::fmt::Debug;
+    type Error: std::error::Error;
+
+    fn ensure_user(&self, user: &UserInfo)
+        -> impl Future<Output = Result<i64, Self::Error>> + Send;
+
+    fn fetch_subject_roles(
+        &self,
+        subject_id: i64,
+        roles_config: &RoleConfig<Self::BuiltinRole>,
+    ) -> impl Future<Output = Result<HashSet<Self::BuiltinRole>, Self::Error>> + Send;
+
+    fn ensure_subject_roles(
+        &self,
+        subject_id: i64,
+        roles_config: &RoleConfig<Self::BuiltinRole>,
+        roles: HashSet<Self::BuiltinRole>,
+    ) -> impl Future<Output = Result<(), Self::Error>> + Send;
+
+    fn remove_subject_roles(
+        &self,
+        subject_id: i64,
+        roles_config: &RoleConfig<Self::BuiltinRole>,
+        roles: HashSet<Self::BuiltinRole>,
+    ) -> impl Future<Output = Result<HashSet<Self::BuiltinRole>, Self::Error>> + Send;
+}
+
+impl<'config, S: StorageDriver> Authorizer<'config, S> {
+    pub async fn try_initialize(
+        user: UserInfo,
+        roles_config: &'config RoleConfig<S::BuiltinRole>,
+        storage_driver: S,
+    ) -> Result<Self, S::Error> {
+        let user_id = storage_driver.ensure_user(&user).await?;
+        debug!(%user, %user_id, "user authenticated");
+        Ok(Self {
+            user,
+            user_id,
+            roles_config,
+            user_roles: None,
+            storage: storage_driver,
+        })
+    }
+
+    pub fn new_superuser(
+        roles_config: &'config RoleConfig<S::BuiltinRole>,
+        storage_driver: S,
+    ) -> Self {
+        debug_assert!(
+            roles_config.is_superuser(),
+            "Authorizer::new_superuser requires a superuser role config"
+        );
+        Self {
+            user: UserInfo {
+                identity: "superuser".to_string(),
+                name: "Super User".to_string(),
+            },
+            user_id: -1,
+            roles_config,
+            user_roles: None,
+            storage: storage_driver,
+        }
+    }
+
+    pub fn is_superuser(&self) -> bool {
+        self.roles_config.is_superuser()
+    }
+
+    /// Check that the user has all the required builting roles
+    #[tracing::instrument(skip_all, fields(user = %self.user, user_roles = ?self.user_roles, ?required_roles), ret, err)]
+    pub async fn check_roles(
+        &mut self,
+        required_roles: HashSet<S::BuiltinRole>,
+    ) -> Result<bool, S::Error> {
+        if self.is_superuser() {
+            tracing::debug!("role checking skipped for superuser");
+            return Ok(true);
+        }
+
+        let user_roles = self.get_user_roles().await?;
+        Ok(required_roles.is_subset(user_roles))
+    }
+
+    async fn get_user_roles(&mut self) -> Result<&HashSet<S::BuiltinRole>, S::Error> {
+        match self.user_roles {
+            None => {
+                tracing::info!("fetching user roles");
+                let user_roles = self
+                    .storage
+                    .fetch_subject_roles(self.user_id, self.roles_config)
+                    .await?;
+                tracing::debug!(roles = ?user_roles, "caching user roles");
+                Ok(self.user_roles.insert(user_roles))
+            }
+            Some(ref user_roles) => Ok(user_roles),
+        }
+    }
+}
+
+impl<'config, S: StorageDriver> std::fmt::Debug for Authorizer<'config, S> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Authorizer")
+            .field("user", &self.user)
+            .field("user_id", &self.user_id)
+            .field("roles_config", &self.roles_config)
+            .field("user_roles", &self.user_roles)
+            .finish()
+    }
+}
+
+impl std::fmt::Display for UserInfo {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{} ({})", self.identity, self.name)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::fixtures::*;
+    use pretty_assertions::assert_eq;
+    use std::{
+        borrow::BorrowMut,
+        collections::HashMap,
+        convert::Infallible,
+        sync::{Arc, Mutex},
+    };
+
+    #[derive(Debug, Clone, Default)]
+    struct MockStorageDriver {
+        users: Arc<Mutex<HashMap<UserIdentity, i64>>>,
+        user_roles: Arc<Mutex<HashMap<i64, HashSet<TestBuiltinRole>>>>,
+    }
+
+    #[tokio::test]
+    async fn superuser() {
+        let config = RoleConfig::new_superuser();
+        let storage = MockStorageDriver::default();
+        let mut authorizer = Authorizer::new_superuser(&config, storage);
+        assert!(authorizer.is_superuser());
+        // Check that the superuser has any role even if not explicitely granted
+        assert_eq!(
+            authorizer
+                .check_roles(HashSet::from([TestBuiltinRole::UserBan]))
+                .await,
+            Ok(true)
+        );
+    }
+
+    #[tokio::test]
+    async fn check_roles() {
+        let config = default_test_config();
+        let storage = MockStorageDriver::default();
+        let mut authorizer = Authorizer::try_initialize(
+            UserInfo {
+                identity: "toto".to_owned(),
+                name: "Sir Toto, the One and Only".to_owned(),
+            },
+            &config,
+            storage,
+        )
+        .await
+        .unwrap();
+
+        {
+            let users = authorizer.storage.users.lock().unwrap();
+            assert_eq!(
+                users.iter().next(),
+                Some((&"toto".to_owned(), &0)),
+                "new user should have been created"
+            );
+        }
+
+        // insert some mocked roles
+        {
+            let mut user_roles = authorizer.storage.user_roles.lock().unwrap();
+            user_roles.insert(
+                0,
+                HashSet::from([
+                    TestBuiltinRole::DocRead,
+                    TestBuiltinRole::UserAdd,
+                    TestBuiltinRole::UserBan,
+                ]),
+            );
+        }
+
+        assert!(authorizer
+            .check_roles(HashSet::from([TestBuiltinRole::DocRead]))
+            .await
+            .unwrap());
+        assert!(authorizer
+            .check_roles(HashSet::from([
+                TestBuiltinRole::UserAdd,
+                TestBuiltinRole::UserBan
+            ]))
+            .await
+            .unwrap());
+        assert!(authorizer
+            .check_roles(HashSet::from([
+                TestBuiltinRole::DocRead,
+                TestBuiltinRole::UserAdd,
+                TestBuiltinRole::UserBan
+            ]))
+            .await
+            .unwrap());
+
+        assert!(!authorizer
+            .check_roles(HashSet::from([TestBuiltinRole::DocEdit,]))
+            .await
+            .unwrap());
+        assert!(!authorizer
+            .check_roles(HashSet::from([
+                TestBuiltinRole::DocRead,
+                TestBuiltinRole::DocDelete,
+            ]))
+            .await
+            .unwrap());
+
+        assert!(authorizer.check_roles(HashSet::from([])).await.unwrap());
+    }
+
+    impl StorageDriver for MockStorageDriver {
+        type BuiltinRole = TestBuiltinRole;
+        type Error = Infallible;
+
+        async fn ensure_user(&self, user: &UserInfo) -> Result<i64, Self::Error> {
+            let mut users = self.users.lock().unwrap();
+            let next_id = users.len() as i64;
+            let user_id = users.entry(user.identity.clone()).or_insert(next_id);
+            Ok(*user_id)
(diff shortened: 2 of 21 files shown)


leovalais:
### On-site peer review with @woshilapin 

- [x] Prefer `toml_edit` over `toml`
- [ ] The generic parameter of `RoleConfig` may be unnecessary
- [x] The `config/config_version` in table `roles` is unnecessary and may cause maintenance issues
- [x] All application roles should be serialized in DB and garbage collected upon configuration changes. We can ensure the sync either at server boot or manually with a CLI command.

A small design discussion about the last two points will be necessary @Khoyo

leovalais:
Changes:

* builtin roles are now stored in DB with deduplication
* no more `config/config_version`
* simplified requests
* diesel query builder use
* lift some `StorageDriver` constraints
* `StorageDriver` functions now take a `subject_id` instead of an `identifier` as these operations may be applied to groups
* `StorageDriver` is now responsible for serializing and deserializing builtin roles

editoast/editoast_authz/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "editoast_authz"
+version = "0.1.0"
+edition = "2021"
+license = "LGPL-3.0"
```
woshilapin:
```suggestion
license.workspace = true
version.workspace = true
edition.workspace = true
```

editoast/editoast_authz/src/roles.rs
```diff
@@ -0,0 +1,246 @@
+use std::collections::{HashMap, HashSet};
+
+// TODO: improve general naming of stuff
+
+pub trait BuiltinRoleSet: Sized + Clone + std::hash::Hash + std::cmp::Eq + std::fmt::Debug {
+    fn parse_tag(tag: &str) -> Option<Self>;
```
woshilapin: It looks like implementing [`FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html) (modulo that `FromStr` returns a `Result`, not an `Option` but I think this is fine). How about having `BuiltinRoleSet` a super-trait of `FromStr`?

editoast/editoast_authz/src/roles.rs
```diff
@@ -0,0 +1,246 @@
...
+    fn tag(&self) -> &str;
```
woshilapin: When I see `tag()`, I expect something typed in return. How about naming that function `as_str`? Or even better, how about having `BuiltinRoleSet` a super-trait of [`Display`](https://doc.rust-lang.org/std/fmt/trait.Display.html)?

editoast/editoast_authz/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+                "doc_provider".to_string(),
+                HashSet::from([DocDelete, DocEdit, DocRead]),
+            ),
+            ("admin".to_string(), HashSet::from([UserAdd, UserBan])),
+            (
+                "dev".to_string(),
+                HashSet::from([UserAdd, UserBan, DocDelete, DocEdit, DocRead]),
+            ),
+        ]))
+    }
```
woshilapin: Instead of maintaining both the code and a comment of the same data structure, how about using directly the comment as input to your function, I believe it's going to be more maintainable that way. Something like this.

```rust
    pub fn default_test_config() -> roles::RoleConfig<TestBuiltinRole> {
        let roles = r#"
        [application_roles.doc_reader]
        implies = ["doc_read"]
        [application_roles.doc_provider]
        implies = ["doc_delete", "doc_edit", "doc_read"]
        [application_roles.admin]
        implies = ["user_add", "user_ban"]
        [application_roles.dev]
        implies = ["admin", "doc_provider"]
        "#;
        RoleConfig::load(roles).unwrap()
    }
```

editoast/editoast_authz/src/roles.rs
```diff
@@ -0,0 +1,246 @@
+use std::collections::{HashMap, HashSet};
+
+// TODO: improve general naming of stuff
```
woshilapin: See comments below, it might help 😉

editoast/editoast_authz/src/roles.rs
```diff
@@ -0,0 +1,246 @@
...
+
+pub trait BuiltinRoleSet: Sized + Clone + std::hash::Hash + std::cmp::Eq + std::fmt::Debug {
+    fn parse_tag(tag: &str) -> Option<Self>;
+    fn tag(&self) -> &str;
+    fn implies(&self) -> impl IntoIterator<Item = Self>;
```
woshilapin: I would expect some `_iter` suffix maybe 🤷

editoast/editoast_authz/src/roles.rs
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    pub fn is_superuser(&self) -> bool {
+        self.superuser
+    }
+
+    pub fn resolve<'r>(
+        &self,
+        app_roles: impl Iterator<Item = &'r RoleIdentifier>,
+    ) -> Result<HashSet<B>, &'r RoleIdentifier> {
```
woshilapin, line 47: I really like that elegant API about erroring (feel free to resolve the comment, obviously 😃).

editoast/editoast_authz/src/builtin_role.rs
```diff
@@ -0,0 +1,29 @@
+use crate::roles::BuiltinRoleSet;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum BuiltinRole {
+    OpsWrite,
+}
+
+// TODO: a derive macro would be nice
```
woshilapin: Please no more macro, especially here, it's not complex to write 😄 On top of that, if we use `FromStr` and `Display`, we will be able to leverage `strum` I believe (actually, with `strum`, we might even be able to convert to `&'static str` with [`IntoStaticStr`](https://docs.rs/strum_macros/latest/strum_macros/derive.IntoStaticStr.html) instead of implementing `Display`).

editoast/editoast_authz/src/roles.rs
```diff
@@ -0,0 +1,246 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum RoleConfigParsingError {
+    #[error(transparent)]
+    ParsingError(#[from] toml_edit::de::Error),
+    #[error("Invalid role identifier: '{0}'")]
+    InvalidRoleIdentifier(String),
+    #[error("Role identifier '{0}' is neither a builtin role or a declared application role")]
+    UndeclaredRoleIdentifier(RoleIdentifier),
+    #[error("Role identifier '{0}' is a reserved for a builtin role")]
+    ResevedRoleIdentifier(RoleIdentifier),
```
woshilapin:
```suggestion
    ReservedRoleIdentifier(RoleIdentifier),
```

editoast/editoast_authz/src/roles.rs
```diff
@@ -0,0 +1,246 @@
...
+
+#[derive(Debug, thiserror::Error)]
+pub enum RoleConfigParsingError {
+    #[error(transparent)]
+    ParsingError(#[from] toml_edit::de::Error),
+    #[error("Invalid role identifier: '{0}'")]
+    InvalidRoleIdentifier(String),
+    #[error("Role identifier '{0}' is neither a builtin role or a declared application role")]
+    UndeclaredRoleIdentifier(RoleIdentifier),
+    #[error("Role identifier '{0}' is a reserved for a builtin role")]
```
woshilapin:
```suggestion
    #[error("Role identifier '{0}' is a reserved builtin role")]
```

editoast/editoast_authz/src/roles.rs
```diff
@@ -0,0 +1,246 @@
...
+}
+
+impl<B: BuiltinRoleSet> std::fmt::Display for RoleConfig<B> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if self.superuser {
+            return write!(f, "RoleConfig(superuser)");
+        }
+        if self.superuser {
+            return write!(f, "RoleConfig(superuser)");
+        }
```
woshilapin: Why is there twice the same code? Mmm, copy-paste frenzy?

editoast/editoast_authz/src/roles.rs
```diff
@@ -0,0 +1,246 @@
...
+    #[error("Role identifier '{0}' is neither a builtin role or a declared application role")]
+    UndeclaredRoleIdentifier(RoleIdentifier),
+    #[error("Role identifier '{0}' is a reserved for a builtin role")]
+    ResevedRoleIdentifier(RoleIdentifier),
+}
+
+impl<B: BuiltinRoleSet> std::fmt::Display for RoleConfig<B> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if self.superuser {
+            return write!(f, "RoleConfig(superuser)");
```
woshilapin: Using `write!()` without `?` at the end probably ignore the result. I would rewrite the function like this.

```rust
if self.superuser {
  write!(f, "RoleConfig(superuser)")
] else {
  write!(f, "RoleConfig({} roles))", self.resolved_roles.len())
}
```

editoast/editoast_authz/src/roles.rs
```diff
@@ -0,0 +1,246 @@
...
+        assert!(matches!(
+            err,
+            RoleConfigParsingError::InvalidRoleIdentifier(_)
+        ));
+
+        let source = format!(
+            r#"
+            application_roles = {{ "{} batman!" = {{ implies = [] }} }}
+            "#,
+            "na".repeat(256)
```
woshilapin, line 244: 🤣

front/public/locales/fr/errors.json
```diff
@@ -16,6 +16,11 @@
     "attached": {
       "TrackNotFound": "Section de ligne {{track_id}} non trouvée"
     },
+    "authorization": {
+      "AuthError": "Erreur interne d'authentification ou d'authorisation, veuillez réessayer ou contacter le support",
```
Khoyo: autorisation

editoast/editoast_authz/src/authorizer.rs
```diff
@@ -0,0 +1,286 @@
...
+        if self.user_roles.is_none() {
+            tracing::info!("fetching user roles");
+            let user_roles = self
+                .storage
+                .fetch_subject_roles(self.user_id, self.roles_config)
+                .await?;
+            tracing::debug!(roles = ?user_roles, "caching user roles");
+            self.user_roles = Some(user_roles);
+        }
+        let user_roles = self.user_roles.as_ref().expect("roles should be in cache");
```
woshilapin: Instead of an `expect` here, can't we do something about using `self.user_roles.unwrap_or_else(|| /* filling up the cache */)`?

editoast/editoast_authz/src/authorizer.rs
```diff
@@ -0,0 +1,286 @@
...
+        user_roles: Arc<Mutex<HashMap<i64, HashSet<TestBuiltinRole>>>>,
+    }
+
+    #[tokio::test]
+    async fn superuser() {
+        let config = RoleConfig::new_superuser();
+        let storage = MockStorageDriver::default();
+        let mut authorizer = Authorizer::new_superuser(&config, storage);
+        assert!(authorizer.is_superuser());
+        // We don't grant any roles to the user
```
woshilapin: Do you really want to negate that sentence? Isn't the superuser supposed to be granted all the roles, no matter what?

woshilapin (commented):
A first series of comments about `editoast_authz`. The rest is coming soon, stay tuned.

editoast/editoast_authz/src/roles.rs
Khoyo: No it doesn't, we return the result - (using `?` we would need to wrap the expression in an `Ok()`). But sure, we can switch to a more functional style.

editoast/editoast_authz/src/authorizer.rs
Khoyo: I think it was a commentary about what the test is doing ^^"

editoast/editoast_authz/src/authorizer.rs
Khoyo: We insert it just above. `get_or_insert_with()` would be ideal if not for the fact that it cannot take an async closure :/

editoast/editoast_authz/src/roles.rs
leovalais: Used `IntoStaticStr` + helper function

editoast/editoast_authz/src/lib.rs
leovalais: I wrote it this way initially to decorrelate the fixture from the parsing code. I changed it though.

editoast/editoast_authz/src/builtin_role.rs
leovalais: On a maintenance perspective, less proc macros is a good thing I agree ^^
(I'll add my macro somewhere else then... 😞)

editoast/editoast_authz/src/authorizer.rs
Khoyo: We moved it to a helper function that perform explicit matching

editoast/editoast_authz/src/roles.rs
leovalais: I think naming should be a bit better now.

editoast/editoast_authz/src/roles.rs
```diff
@@ -0,0 +1,252 @@
...
+    str::FromStr,
+};
+
+pub trait BuiltinRoleSet:
+    FromStr + Into<&'static str> + Sized + Clone + std::hash::Hash + std::cmp::Eq + std::fmt::Debug
+{
+    fn implies_iter(&self) -> impl IntoIterator<Item = Self>;
+
+    fn as_str(&self) -> &'static str {
+        self.clone().into()
```
woshilapin: Mmm, sad that `strum` implements `From<T> for &'static str` and not `From<&T> for &'static str`... I wonder why... and looking for ideas, I found out about [this](https://github.com/Peternator7/strum/issues/291) and reading through comments, found about [this example](https://github.com/Peternator7/strum/blob/db553e3dcf642ccc3fba60190bae3c42d50ea803/strum_tests/tests/as_ref_str.rs#L73C1-L74C1). I believe, it should be possible to avoid `.clone()`.

editoast/src/modelsv2/authn.rs
```diff
@@ -0,0 +1,224 @@
...
+
+#[derive(Debug, thiserror::Error)]
+pub enum AuthDriverError {
+    #[error(transparent)]
+    DieselError(#[from] diesel::result::Error),
+    #[error(transparent)]
+    PoolError(#[from] editoast_models::db_connection_pool::DatabasePoolError),
+}
+
+// TODO: use load_stream instead of load to avoid useless allocations
```
woshilapin: Do we really worry about performance on loading this information? I don't believe we expect an important number of objects. I would rather not leave such `TODO` that probably will never be addressed and might not even be useful, what do you think?

editoast/src/modelsv2/authn.rs
```diff
@@ -0,0 +1,224 @@
...
+        conn.transaction(|conn| {
+            async {
+                let user_id = authn_user::table
+                    .select(authn_user::id)
+                    .filter(authn_user::identity_id.eq(&user.identity))
+                    .first::<i64>(conn)
+                    .await
+                    .optional()?;
+
+                if user_id.is_none() {
```
woshilapin: You can use a `match` instead, that will avoid an `.unwrap()` down the line.

editoast/src/modelsv2/authn.rs
```diff
@@ -0,0 +1,224 @@
...
+                authn_group_membership::table.on(authn_group_membership::user.eq(subject_id)),
+            )
+            .filter(authz_role::subject.eq(subject_id))
+            .or_filter(authz_role::subject.eq(authn_group_membership::group))
+            .load::<String>(&mut conn)
+            .await?
+            .into_iter()
+            .map(|role| {
+                Self::BuiltinRole::from_str(role.as_str())
+                    .ok()
```
woshilapin: Why the `.ok()`? How about using `.expect()` directly on the `Result`?

editoast/src/views/mod.rs
```diff
@@ -112,6 +116,56 @@ editoast_common::schemas! {
     work_schedules::schemas(),
 }
 
+pub type Roles = editoast_authz::roles::RoleConfig<BuiltinRole>;
+
+// This function will become a middleware once we switch to axum
```
woshilapin, line 121: It might be time to address that then? 😉 [Remember](https://github.com/OpenRailAssociation/osrd/pull/8047)

editoast/src/views/mod.rs
```diff
@@ -112,6 +116,56 @@ editoast_common::schemas! {
...
+            PgAuthDriver::<BuiltinRole>::new(db_pool.clone()),
+        ));
+    }
+    let Some(header) = headers.get("x-remote-user") else {
+        return Err(AuthorizationError::Unauthenticated);
+    };
+    let (identity, name) = header
+        .to_str()
+        .expect("unexpected non-ascii characters in x-remote-user")
+        .split_once('/') // FIXME: the gateway should inject two headers instead
```
woshilapin, line 139: Probably for a follow-up commit or PR, but leaving a comment to not forget about it.

editoast/src/views/projects.rs
```diff
@@ -144,9 +148,22 @@ impl ProjectWithStudyCount {
...
     Json(project_create_form): Json<ProjectCreateForm>,
 ) -> Result<Json<ProjectWithStudyCount>> {
+    let mut authorizer = make_authorizer(&headers, role_config.as_ref(), db_pool.clone()).await?;
+    let authorized = authorizer
+        .check_roles([BuiltinRole::OpsWrite].into())
+        .await
+        .map_err(AuthorizationError::AuthError)?;
+    if !authorized {
+        return Err(AuthorizationError::Unauthorized.into());
+    }
```
woshilapin, line 166: I suppose all of this would disappear with a middleware?

woshilapin (approved):
Huge PR, looks fine from my side even if I am no expert in authorization & co.

editoast/editoast_authz/src/roles.rs
Khoyo: This would require introducing a trait bound on `&Self`, so it wouldn't be a supertrait, so it would have to be repeated everywhere we use a `BuiltinRoleSet` generic bound :/

On the other hand, I dont think we ever expect those enums to hold data, so we could use a trait bound on Copy (or just expect the clone to be optimized away). @leovalais wdyt?

editoast/src/modelsv2/authn.rs
leovalais: Because `FromStr` doesn't require it's `Err` to be `Debug`

editoast/src/views/mod.rs
Khoyo: We'll do that in a follow-up, I'm opening an issue

editoast/src/views/projects.rs
Khoyo: I don't think so. `let mut authorizer = make_authorizer(&headers, role_config.as_ref(), db_pool.clone()).await?;` would, but the actual check would not. Unless we put all the rules in a centralized place, but that isn't the current design.

editoast/editoast_authz/src/roles.rs
leovalais: Using `derive(AsRefStr)` solves the issue.

editoast/src/modelsv2/authn.rs
leovalais: Seems like a good plan.

kmer2016 (approved):
LGTM

PR #7988, opened 2024-07-09
ci: add front code formatting check

emersion:
We run ESLint already at build-time, however we never run Prettier. As a result, non-TS files such as SCSS are not getting checked.

* * *

Of course, this isn't enough as-is because Prettier will trip on many SCSS files.


Yohh (approved):
nice! thanks :muscle:

woshilapin (approved):

emersion:
List of re-formatted files: https://github.com/OpenRailAssociation/osrd/actions/runs/9854031436/job/27205971674?pr=7988

I've pushed a commit to re-format everything but it needs more eyeballing. For instance, `src/reducers/osrdconf/infra_schema.json` is automatically generated (we don't want to fail CI when re-generating the file).

emersion:
I've added a `.prettierignore` file with machine-generated files.

emersion:
Ah, maybe we want to ignore map styles as well. These seem to be machine-generated. Edited the commits to do so.

theocrsb (approved):

front/.prettierignore
```diff
@@ -0,0 +1,7 @@
+src/common/ReleaseInformations/json/licenses.json
+src/modules/rollingStock/components/RollingStockSelector/consts/mlgtraffic.json
+src/modules/rollingStock/components/RollingStockSelector/consts/rollingstockDetailDB.json
+src/modules/trainschedule/components/ImportTrainSchedule/rollingstock_opendata2osrd.json
+src/reducers/osrdconf/infra_schema.json
+src/assets/mapstyles/*.json
+src/utils/__tests__/assets/*.json
```
Khoyo, line 10: I'd add `src/config/openapi-editoast-config.ts`  and `src/config/openapi-gateway-config.ts` too as they are generated

Khoyo (approved):
Otherwise LGTM

front/.prettierignore
emersion: Hm, I don't think these are generated? These are hand-written config files to configure the OpenAPI codegen, I believe?

anisometropie: yeah `src/config/openapi-editoast-config.ts` and `src/config/openapi-gateway-config.ts` are not generated, 
the ones to ignore are `src/common/api/generatedEditoastApi.ts` and `src/common/api/osrdGatewayApi.ts`.

They are also ignored by eslint (even though some eslint is run **once** when they are generated)

emersion: Right. Currently `scripts/generate-types.sh` runs `eslint` after generating the files, so it's not strictly necessary to add these files to the ignore list.

anisometropie: we don’t really care about the formatting of these files. eslint is run just once at generation.
otherwise they are ignored in the eslint patterns, I would ignore these too in prettier.

front/README.md
```diff
@@ -40,26 +40,27 @@ Launches end to end tests.
...
 - Running front with `docker compose up --build --detach front`
 
 Now you can run the test with `cd front/ && yarn e2e-tests`.
 
-> [!CAUTION]
-> If you try to run `yarn start` instead of running it through docker, you'll notice it doesn't
-> work because the gateway can't access your local port from inside a container. 2 solutions:
+> [!CAUTION] If you try to run `yarn start` instead of running it through docker, you'll notice it
+> doesn't work because the gateway can't access your local port from inside a container. 2
+> solutions:
```
flomonster: This is broken (you can check the GitHub preview). The line feed is required.
We should probably add this file to the `.prettierignore`.

flomonster (approved):
LGTM (not tested)

front/.prettierignore
Khoyo: Right, I copy pasted the wrong thing ^^'

emersion: @anisometropie That makes sense, I added these files for consistency.

emersion:
While resolving the comments from @anisometropie, I noticed that a few hand-written files were being ignored in the ESLint config. I've fixed this in a new commit: "front: tighten up ignoredPatterns in .eslintrc". See its commit message for details.

front/README.md
emersion: Eh :/ 

Oh well, I just added a rule to ignore `*.md` for now.

(An alternative would be to add two spaces at the end of the line probably, but we'd risk getting it removed by mistake by some editors stripping whitespace at end-of-lines automatically when saving a file.)

emersion:
Does this version still look good with the additional commit?

anisometropie (approved):
Great, thanks for tidying all that up.

clarani (approved):
LGTM

PR #7991, opened 2024-07-09
docker: add back build cache mounts

Khoyo:


emersion:
The CI in this PR takes more than 20min - do we expect subsequent CI runs to take significantly less time than that or not?

Khoyo:
~~I'm checking. This is mostly for local speedup, I wasn't expecting a speedup in the CI. It seems to go reasonably fast anyway when not rebuilding the whole backend (8min with a random editoast change) ~~

It is now disabled during CI builds.

ghost (approved):
Yes, that will help developers a lot, thanks !

emersion (approved):
LGTM!

flomonster (approved):
LGTM

PR #7992, opened 2024-07-09
front: hide provisional curve in the speed space chart

### LINKED ISSUE #7967: Double computation when setting margin + specific times
### What happened?

Asking for some margins + specific times:

image

I got this output:

image

The computed time is correct but it seems like the margins aren't. And the Space-Speed Chart is weird. Both problems seems related.

The requested time should prevail on the the requested margin. So the real margins should be increased.

And the chart should only have 2 curves anyways (one fastest, one real)

1. Reproduce the screenshots above

- **Browser**: Chrome
- **OS**: W11
- **Env**: dev

839f7ef

SarahBellaha:
Closes #7967

### DIFF
--- a/front/src/modules/simulationResult/components/SpeedSpaceChart/d3Helpers.ts
+++ b/front/src/modules/simulationResult/components/SpeedSpaceChart/d3Helpers.ts
@@ -283,17 +283,7 @@ export const drawSpeedSpaceTrain = (
       CHART_AXES.SPACE_SPEED,
       'speed'
     );
-    if (dataSimulation.standardMarginSpeedData) {
-      drawCurve(
-        chartLocal,
-        'speed margins',
-        dataSimulation.standardMarginSpeedData,
-        'speedSpaceChart',
-        'curveLinear',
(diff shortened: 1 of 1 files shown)


front/src/modules/simulationResult/components/SpeedSpaceChart/d3Helpers.ts
```diff
@@ -283,17 +283,18 @@ export const drawSpeedSpaceTrain = (
...
+    //   drawCurve(
+    //     chartLocal,
+    //     'speed margins',
+    //     dataSimulation.standardMarginSpeedData,
+    //     'speedSpaceChart',
+    //     'curveLinear',
+    //     CHART_AXES.SPACE_SPEED,
+    //     'margins_speed'
+    //   );
+    // }
```
flomonster: It's a bad practice to leave a deadcode like this. Ideally, it should be removed. If we need to retrieve what was done previously we can always use git :wink:

flomonster (approved):

SharglutDev (approved):

PR #7994, opened 2024-07-09
front: change itinerary tab title

### LINKED ISSUE #7819: TSV2 : Input table is not loaded as long as the user doesn't go over the itinerary tab
### What happened?

When creating a train for which I already have the path (for example, start creating a train, then leave the creation, then come back), I cannot open the "Times and stops" tab : 

image

But here, one can see that there is a path, since the length "36.718km" is shown on the "Route" tab

I expected to be able to open the "Time and stops" tab, since I have a path

as above

- **Browser**:  Chrome
- **OS**: W11
- **Env**: Staging internal

c57458e

SarahBellaha:
Closes #7819


theocrsb (approved):
LGTM and tested

Yohh (approved):

kmer2016 (approved):
LGTM, tested

PR #8001, opened 2024-07-10
Remove layer cache duration

flomonster:
Since Redis is configured with the lru policy we don't need to handle the expiration time.


Tguisnet (approved):
lgtm

PR #8004, opened 2024-07-10
Fix simulation_summary being called too often

### LINKED ISSUE #7989: Simulation summary endpoint is called unnecessarily
### What happened?

When a scenario is already open, if the user changes the project of the space-time chart the front re-calls the `simulation_summary` endpoint.

The `simulation_summary` endpoint should be called only when the timetable is loaded.

1. Go to a scenario with multiple trains
2. Open the browser debugger (network tab)
3. Change the train schedule used for the projection

- **Browser**: Firefox v127
- **OS**: Nixos v24.05
- **Env**: All

3a4069c

anisometropie:
- removing `train_schedulev2` tag from `postV2TrainScheduleProjectPath` and making the endpoint a query (codegen made it a mutation by defaut as it is a POST) 
- keeping `train_schedulev2` `postV2TrainScheduleSimulationSummary` => the endpoint should but called again if one train is updated (more refinement to come #8010 ). It should not if a train is created/removed, but we can handle that manually with `trainResultToFetch` logic

closes #7989

### DIFF
--- a/front/src/applications/stdcm/hooks/useStdcmResults.ts
+++ b/front/src/applications/stdcm/hooks/useStdcmResults.ts
@@ -50,7 +50,7 @@ const useStdcmResults = (
   const [postPathProperties] =
     osrdEditoastApi.endpoints.postV2InfraByInfraIdPathProperties.useMutation();
   const [projectTrainSchedules] =
-    osrdEditoastApi.endpoints.postV2TrainScheduleProjectPath.useMutation();
+    osrdEditoastApi.endpoints.postV2TrainScheduleProjectPath.useLazyQuery();
 
   const { data: otherSelectedTrainSchedule } =
     osrdEditoastApi.endpoints.getV2TrainScheduleById.useQuery(
@@ -176,7 +176,7 @@ const useStdcmResults = (
         stdcmResponse.rollingStock.length
       );
     }
-  }, [stdcmResponse]);
+  }, [stdcmResponse, infraId, timetable, projectTrainSchedules]);
 
   if (!infraId || !stdcmResponse || !selectedTrainSchedule) return null;
 
--- a/front/src/common/api/generatedEditoastApi.ts
+++ b/front/src/common/api/generatedEditoastApi.ts
@@ -925,7 +925,7 @@ const injectedRtkApi = api
         }),
         invalidatesTags: ['timetablev2', 'train_schedulev2'],
       }),
-      postV2TrainScheduleProjectPath: build.mutation<
+      postV2TrainScheduleProjectPath: build.query<
         PostV2TrainScheduleProjectPathApiResponse,
         PostV2TrainScheduleProjectPathApiArg
       >({
@@ -934,7 +934,7 @@ const injectedRtkApi = api
           method: 'POST',
           body: queryArg.projectPathForm,
         }),
-        invalidatesTags: ['train_schedulev2'],
+        providesTags: ['train_schedulev2'],
       }),
       postV2TrainScheduleSimulationSummary: build.query<
         PostV2TrainScheduleSimulationSummaryApiResponse,
--- a/front/src/config/openapi-editoast-config.ts
+++ b/front/src/config/openapi-editoast-config.ts
@@ -10,7 +10,11 @@ const config: ConfigFile = {
   tag: true,
   endpointOverrides: [
     {
-      pattern: ['postV2TrainSchedule', 'postV2TrainScheduleSimulationSummary'],
+      pattern: [
+        'postV2TrainSchedule',
+        'postV2TrainScheduleSimulationSummary',
+        'postV2TrainScheduleProjectPath',
+      ],
       type: 'query',
     },
   ],


emersion:
Instead of changing the OpenAPI definition side, we can override the auto-generated `providesTags`/`invalidatesTags` here: https://github.com/OpenRailAssociation/osrd/blob/dev/front/src/common/api/osrdEditoastApi.ts

editoast/src/views/v2/train_schedule/projection.rs
```diff
@@ -119,7 +119,6 @@ struct CachedProjectPathTrainResult {
 ///
 /// Train schedules that are invalid (pathfinding or simulation failed) are not included in the result
 #[utoipa::path(
-    tag = "train_schedulev2",
```
flomonster: This should be overridden from the frontend.

flomonster (changes requested):

flomonster (approved):
LGTM

front/src/common/api/osrdEditoastApi.ts
```diff
@@ -86,6 +86,9 @@ const osrdEditoastApi = generatedEditoastApi.enhanceEndpoints({
     deleteV2ProjectsByProjectIdStudiesAndStudyIdScenariosScenarioId: {
       invalidatesTags: ['scenariosv2', 'studies', 'projects'],
     },
+    postV2TrainScheduleProjectPath: {
+      invalidatesTags: [],
+    }
```
SharglutDev: I don't think we should remove the tag in this endpoint `invalidatesTags` but rather remove for now the `trainSchedulev2` in the `postV2TrainScheduleSimulationSummary` `providesTags` so that any of the other trainschedule mutation endpoints don't trigger summary.

The workflow in trainschedule v2 when adding/editing a train : `timetablev2` tag is invalidated, we get the new trainIds required to call `postV2TrainScheduleSimulationSummary` and `postV2TrainSchedule` that give us the train details to display in the timetable. So we don't really need to have any automate refetching other than the timetable one.

We plan one day to use more granular control to invalidate cache with rtk query (https://redux-toolkit.js.org/rtk-query/usage/automated-refetching#invalidating-cache-data)

emersion: rtk-query doesn't auto-generate any `providesTags` for mutations. It only generates `invalidatesTags`. See:

https://github.com/OpenRailAssociation/osrd/blob/de8ada893e20f4b16cee8f34f7b60e7f53d7625d/front/src/common/api/generatedEditoastApi.ts#L928-L938

emersion: TBH it's a bit weird to use a mutation for this request. Would be nicer to add it here so that it's generated as a query instead:

https://github.com/OpenRailAssociation/osrd/blob/de8ada893e20f4b16cee8f34f7b60e7f53d7625d/front/src/config/openapi-editoast-config.ts#L13

But that's kinda unrelated to the original bug.

anisometropie: > I don't think we should remove the tag in this endpoint invalidatesTags but rather remove for now the trainSchedulev2 in the postV2TrainScheduleSimulationSummary providesTags so that any of the other trainschedule mutation endpoints don't trigger summary.

> The workflow in trainschedule v2 when adding/editing a train : timetablev2 tag is invalidated, we get the new trainIds required to call postV2TrainScheduleSimulationSummary and postV2TrainSchedule that give us the train details to display in the timetable. So we don't really need to have any automate refetching other than the timetable one.

> We plan one day to use more granular control to invalidate cache with rtk query (https://redux-toolkit.js.org/rtk-query/usage/automated-refetching#invalidating-cache-data)

why does postV2TrainScheduleProjectPath need to invalidate that tag in the first place ?

anisometropie: > TBH it's a bit weird to use a mutation for this request. Would be nicer to add it here so that it's generated as a query instead:
> 
> https://github.com/OpenRailAssociation/osrd/blob/de8ada893e20f4b16cee8f34f7b60e7f53d7625d/front/src/config/openapi-editoast-config.ts#L13
> 
> But that's kinda unrelated to the original bug.

I think it considers all POST/PUT as mutations. `train_schedule/project_path` is creating or modifying something in DB ?

emersion: > I think it considers all POST/PUT as mutations.

Yup, that's the default behavior (can be changed by editing the referenced file).

> train_schedule/project_path is creating or modifying something in DB ?

I don't believe it does? (Correct me if I'm wrong)

anisometropie: train_schedule/project_path is confirmed not to be modifying anything, so it has no reason to invalidate a tag that we would need to refetch.

SharglutDev: Agreed for the projectpath but I still think we should also remove the trainSchedulev2 tag from summary endpoint `providesTags` because we will have the same issue when post/put a trainSchedule

emersion: Hm, what is your concern with `postV2TrainScheduleSimulationSummary`? When we create/edit/delete a train schedule, everything with the `train_schedulev2` tag gets invalidated, so the summary is pruned from the cache, I think?

SharglutDev: It might have been lost in the conversation, but in my first comment, I described the workflow used in the app in the timetable section : 
Capture d’écran 2024-07-12 à 15 04 27

To summarize, we will refetch summary anyway because add/put trainschedule invalidates timetable which cascades into summary.

anisometropie: I’m still a bit puzzled about completely removing the tag from summary.

if you modify the train start time, the result will be the same, no need to recall summary. 
but  if you modify the train path, it won’t, so summary needs to be recalled.

SharglutDev (changes requested):
We should maybe connect the trainResultToFetch state to the hook containing the summary endpoint to be sure that when we add a train, we fetch only the summary/getTrainSchedule of the new train and not the others that didn't changed.
Same when deleting a train.

SharglutDev:
Thank you for the changes. Regarding my last comment, we have fresh news.

A high priority [issue](https://github.com/OpenRailAssociation/osrd/issues/8055) has just came up. Most of the logic of timetable, simulation result will be changed (including `trainResultToFetch`) so don't bother to make these changes.

You just have the build to fix and we can approve this PR I think :)

SharglutDev (approved):
Lgtm, thank you !

PR #8007, opened 2024-07-10
Remove some unused dependencies in editoast

leovalais:
🔪 `cargo machete`


flomonster (approved):
:hocho:

flomonster:
It looks like `async-std` was used and cannot be removed.

leovalais:
According to the docs of `rstest`, we can customize which `test` attribute we want: https://github.com/la10736/rstest?tab=readme-ov-file#inject-test-attribute

Maybe replacing every `rstest` with `tokio::test` or injecting it when we use `#[case]` could help us get rid of that dependency? I'll look into it another day.

Issue #7585, opened 2024-06-03
TSv2: Multiple problems to create a train with 'arrival on stop signal' on TSV2


woshilapin:
### What happened?

Multiple different bugs happened in a relatively simple scenario. See the video and the description.

https://github.com/OpenRailAssociation/osrd/assets/2520723/204180e9-9bb9-4208-b957-8aaae355421c

1. I'm trying to create a train from the track above and an arrival point on the right. We can see that the pathfinding, highlighted in green on the track, doesn't go all the way to the arrival point but stops at the switch/node.
2. I'm adding a _via_ and then go on the tab for scheduled stops, I see only 2 lines.
3. When I click on the 'arrival on stop signal' checkbox for the last stop, it works. But when I try to click the same checkbox for the _via_, it doesn't get checked
4. Each time I click the checkbox for the _via_, the identifier on the left changes
5. When I go back to the itinerary tab, a new _via_ appeared

### What did you expect to happen?

- [x] I would expect the full pathfinding to be highlighted (https://github.com/OpenRailAssociation/osrd/pull/7630)
- [x] I would expect to see 3 lines, one for the departure, one for the via, then one for the arrival stop (https://github.com/OpenRailAssociation/osrd/pull/8008)
- [x] I would expect to either have the checkbox clicked, or at least having a message to indicate why I cannot click it.
- [x] I would expect the identifier for the _via_ to not change when I only change a configuration of that _via_ (https://github.com/OpenRailAssociation/osrd/pull/7775)

### How can we reproduce it (as minimally and precisely as possible)?

Here's a video to reproduce the steps. The TSV2 and new STDCM features are activated in the user profile.

https://github.com/OpenRailAssociation/osrd/assets/2520723/204180e9-9bb9-4208-b957-8aaae355421c

### What operating system, browser and environment are you using?

- **Browser**: Firefox v126.0.1
- **OS**: Windows
- **Env**: Local **and** internal dev deployment

### OSRD version (top right corner `Account` button > `Informations`)

625d185

woshilapin:
Note: the Station that appears in the tab for scheduled stops, is expected (even if it's surprising in terms of User Experience). When launching a pathfinding, the response will contain all Operational points on the way (the station is an Operational Point). I tested the functionality of 'arrival on stop signal' for Operational Point, and it works as expected. So only _via_ seems to create problems with my video-recorded scenario.

SharglutDev:
[...]

PR #8008, opened 2024-07-10
front: fix pathsteps sorting

theocrsb:
before, for the pathSteps added by a map click, we didn't handle the cases for origin and destination.

#7585

### DIFF
--- a/front/src/modules/pathfinding/utils.ts
+++ b/front/src/modules/pathfinding/utils.ts
@@ -87,10 +87,11 @@ export const getPathfindingQuery = ({
   return null;
 };
 
-export const upsertViasInOPs = (ops: SuggestedOP[], pathSteps: PathStep[]): SuggestedOP[] => {
+export const upsertPathStepsInOPs = (ops: SuggestedOP[], pathSteps: PathStep[]): SuggestedOP[] => {
   let updatedOPs = [...ops];
   pathSteps.forEach((step) => {
-    // We check only for vias added by map click
+    const { stopFor, arrival, onStopSignal, theoreticalMargin } = step;
+    // We check only for pathSteps added by map click
     if ('track' in step) {
       const formattedStep: SuggestedOP = {
         opId: step.id,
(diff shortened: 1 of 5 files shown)


front/src/modules/pathfinding/utils.ts
```diff
@@ -86,34 +86,47 @@ export const getPathfindingQuery = ({
...
         opId: step.id,
         positionOnPath: step.positionOnPath!,
         offsetOnTrack: step.offset,
         track: step.track,
         coordinates: step.coordinates,
+        stopFor,
+        arrival,
+        onStopSignal,
+        theoreticalMargin,
       };
```
Yohh, line 106:
```js
export const upsertPathStepsInOPs = (
  ops: SuggestedOP[],
  pathSteps: Array<PathStep & TrackOffset  & { secondary_code?: string | null; uic: number }>
): SuggestedOP[] => {
  let updatedOPs = [...ops];
  pathSteps.forEach((step) => {
    const {
      id,
      positionOnPath,
      offset,
      track,
      coordinates,
      stopFor,
      arrival,
      onStopSignal,
      theoreticalMargin,
      uic,
      ch,
      kp,
    } = step;
    // We check only for pathSteps added by map click
    if ('track' in step) {
      const formattedStep: SuggestedOP = {
        opId: id,
        positionOnPath: positionOnPath!,
        offsetOnTrack: offset,
        track,
        coordinates,
        stopFor,
        arrival,
        onStopSignal,
        theoreticalMargin,
      };
```

front/src/modules/pathfinding/utils.ts
```diff
@@ -86,34 +86,47 @@ export const getPathfindingQuery = ({
...
       // If it hasn't an uic, the step has been added by map click,
       // we know we have its position on path so we can insert it
       // at the good index in the existing operational points
       const index = updatedOPs.findIndex(
-        (op) => step.positionOnPath && op.positionOnPath >= step.positionOnPath
+        (op) => step.positionOnPath !== undefined && op.positionOnPath >= step.positionOnPath
```
Yohh, line 111:
```suggestion
        (op) => positionOnPath && op.positionOnPath >= positionOnPath
```

front/src/modules/pathfinding/utils.ts
```diff
@@ -86,34 +86,47 @@ export const getPathfindingQuery = ({
...
       );
-      updatedOPs = addElementAtIndex(updatedOPs, index, formattedStep);
+
+      // if index === -1, it's means that the position on path of the last step is bigger
+      // than the last operationnal point position.
+      // So we know this pathStep is the destionation and we want to add it at the end of the array.
```
Yohh:
```suggestion
      // if index === -1, it means that the position on path of the last step is bigger
      // than the last operationnal point position.
      // So we know this pathStep is the destination and we want to add it at the end of the array.
```

front/src/modules/pathfinding/utils.ts
```diff
@@ -86,34 +86,47 @@ export const getPathfindingQuery = ({
...
+      if (index !== -1) {
```
Yohh, line 117: can't it be anything else than -1?

front/src/modules/pathfinding/utils.ts
```diff
@@ -86,34 +86,47 @@ export const getPathfindingQuery = ({
...
+        updatedOPs = addElementAtIndex(updatedOPs, index, formattedStep);
+      } else {
+        updatedOPs.push(formattedStep);
+      }
     } else if ('uic' in step) {
```
Yohh, line 122: to avoid type never

```suggestion
    }
    if ('uic' in step) {
```

front/src/modules/pathfinding/utils.ts
```diff
@@ -86,34 +86,47 @@ export const getPathfindingQuery = ({
...
       updatedOPs = updatedOPs.map((op) => {
         if (op.uic === step.uic && op.ch === step.ch && op.kp === step.kp) {
```
Yohh, line 124:
```suggestion
        if (op.uic === uic && op.ch === ch && op.kp === kp) {
```

Yohh (commented):
nice, a few comments for first

spoiler, changing pathsteps type for `Array<PathStep & TrackOffset  & { secondary_code?: string | null; uic: number }>` add erros in `usePathfinding.ts`, `ManageTrainScheduleV2.tsx`, `operationalStudies/hooks.ts` and `useStdcmResults.ts`, we should have a talk about it and figure out if there is a better way to do it or if I totally loose my mind, perhaps I'm looking for a mistake where there is no one

front/src/modules/pathfinding/utils.ts
theocrsb: Done

front/src/modules/pathfinding/utils.ts
theocrsb: Find calls predicate once for each element of the array, in ascending order, until it finds one where predicate returns true. If such an element is found, findIndex immediately returns that element index. Otherwise, findIndex returns -1.

front/src/modules/pathfinding/utils.ts
theocrsb: The first element at index 0. So if we do not check that it is different from undefined 'findIndex' does not execute because 0 is falsy

Yohh (approved):
if we decide to do anything with PathStep type it's more like a refacto so we can keep it in mind for a better moment. So LGTM

SharglutDev (approved):
Lgtm and tested

PR #8011, opened 2024-07-11
front: fix wrong position of train times in timetable

### LINKED ISSUE #8009: Wrong position of train times in timetable
### What happened?

Capture d’écran 2024-07-11 à 09 40 02

Since this [PR](https://github.com/OpenRailAssociation/osrd/pull/7959), the is a gap on the right of train cards.
It comes from the `display: flex` in the `scenario-timetable-train-times` class.

The display: flex and justify-content should be on only when the .not-honored class is used.

There shouldn't be any spacing on the right of train times if no invalidity.

1. Go to a scenario with trains

- **Browser**: Firefox v126
- **OS**: macOS Sonoma 
- **Env**: Dev

dev 45b2718

theocrsb:
close #8009

### DIFF
--- a/front/src/styles/scss/applications/operationalStudies/_scenario.scss
+++ b/front/src/styles/scss/applications/operationalStudies/_scenario.scss
@@ -321,7 +321,7 @@
                 }
                 .scenario-timetable-train-times {
                   display: flex;
-                  justify-content: space-between;
+                  justify-content: right;
                   min-width: 20%;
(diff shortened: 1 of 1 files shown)


Akctarus (approved):
Tested, thank you for this fix !

emersion (approved):
LGTM and works

Math-R (approved):
tested and lgtm =)

Issue #7559, opened 2024-05-30
actions: make the build less resource intensive


multun:
### Description and goal

Currently, each build pushes a crazy amount of data to ghcr: the build steps of the rust and js projects produce several gigabytes (for a total of 8gb) of built dependencies

Storing this isn't really an issue for the ghcr setup, as the registry performs deduplication, but it becomes really bad when we can't use the registry, due to the PR coming from a fork: https://github.com/OpenRailAssociation/osrd/pull/7558

We identified multiple sources of size bloats, and devised a plan to fix this issue:
 - have a look at the buildx dependency graph, and run each build target cluster in a separate job: Core-build, front-build, editoast-build, gateway-build. This would require moving the bake-metadata step to a separate job
 - reduce the size of editoast-test and gateway-test by [building test binaries first, copying the resulting binaries to a separate image, and publish this one](https://github.com/OpenRailAssociation/osrd/pull/7554/files#diff-d03fc7a67d0a77c20bae12ca4154790c86e901ebb1bbb09b604137e493a14f06R37-R49). The linked WIP does not properly setup the test runner, which should have the same libc as the builder, plus runtime system libs
 - run cargo fmt, clippy and doc outside of docker, which should improve built times by increasing concurrency (we could run these without waiting for the build to complete)
 - attempt to reduce the size of the front-end by hunting down large dependencies

### Acceptance criteria

- [ ] build goes brrr

emersion:
Ref https://github.com/RSeidelsohn/license-checker-rseidelsohn/pull/112 which should help a bit.

emersion:
To get a feel of the culprits, see the artifacts here for instance: https://github.com/OpenRailAssociation/osrd/actions/runs/9729525596

(When building without a fork, we (ab)use the Docker registry and the images size isn't as clearly listed.)

flomonster:
Legacy

PR #8013, opened 2024-07-11
ci: split build job

emersion:
Split CI build job into multiple jobs. This should decrease build time (by using more workers in parallel for intensive tasks) and has the nice side-effect of making the build failure logs more readable (less noise to mentally filter out).

See https://github.com/OpenRailAssociation/osrd/issues/7559

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,6 +20,13 @@ jobs:
       stable_tags: ${{ steps.bake-metadata.outputs.stable_tags }}
       stable_version: ${{ steps.bake-metadata.outputs.stable_version }}
       output_method: ${{ steps.bake-metadata.outputs.output_method }}
+    strategy:
+      matrix:
+        targets:
+          - [core-build, core]
+          - [editoast, editoast-test]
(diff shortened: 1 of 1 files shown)


ghost (approved):
Looks good to me ! Thank you for doing it

woshilapin (approved):

Khoyo (approved):

emersion:
It seems like the "build" job is marked as required for merging in the GitHub settings, and blocks this PR. I think it could be removed because other build steps that depend on it are required anyways (all of the tests). Alternatively, one could replace "build" with every job of the matrix.

ghost:
> It seems like the "build" job is marked as required for merging in the GitHub settings, and blocks this PR. I think it could be removed because other build steps that depend on it are required anyways (all of the tests). Alternatively, one could replace "build" with every job of the matrix.

I think the rationale about removing the requirement makes sense. I can do it, I'll check with other DevOps before doing so and after I'll merge this PR.

Khoyo:
Done

emersion:
Thanks!

PR #8023, opened 2024-07-11
Use multiple dropdown for bug reporting template

flomonster:


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -39,16 +39,30 @@ body:
...
+    id: Browser
+    attributes:
+      label: On which browser the bug occurs?
+      multiple: true
+      options:
+        - Firefox
+        - Google Chrome
+        - Microsoft Edge
+        - Safari
+        - Opera
```
emersion, line 68: Maybe we could have an "Other" option here, just because there will always be That Browser That Isn't In The Big List Of All Browsers?

flomonster: Done

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -39,16 +39,32 @@ body:
...
+      label: On which environments the bug occurs?
+      multiple: true
+      options:
+        - Local
+        - Dev (DgexSol)
+        - Staging (DgexSol)
+        - Dev (SNCF)
+        - Recette (SNCF)
+        - Integration (SNCF)
+        - Other
```
ghost, line 56: Maybe we could add two more options :
- Custom (Docker compose)
- Custom (Using Kubernetes Helm Chart)

Since we will soon have osrdyne that will tighten some behavior with its execution environment it might be relevant?

ghost (approved):
Overall LGTM, one little remark that can be dismissed if not relevant

emersion (approved):

PR #8015, opened 2024-07-11
front: fix scenario timetable height

clarani:
closes #8012

Issue #8012, opened 2024-07-11
Conflicts are no longer displayed at the bottom of the timetable


Akctarus:
### What happened?

It happens since we remove `height: calc(var(--content-height) - 8rem);` from `_scenario.scss/.scenario/.scenario-sidemenu/.scenario-timetable`

<img width="1465" alt="Capture d’écran 2024-07-11 à 10 44 53" src="https://github.com/OpenRailAssociation/osrd/assets/80635161/4c34240e-4e1a-4635-a17a-745b678b9102">

### What did you expect to happen?

<img width="1469" alt="Capture d’écran 2024-07-11 à 11 00 39" src="https://github.com/OpenRailAssociation/osrd/assets/80635161/68909cc0-10c0-4e78-a9f8-bdb448267ea0">

### How can we reproduce it (as minimally and precisely as possible)?

Go to operational studies, create trains with conflicts

### What operating system, browser and environment are you using?

- **Browser**: Firefox v109
- **OS**: Nixos v22.11
- **Env**: Local

### OSRD version (top right corner `Account` button > `Informations`)

dev bea61b1

PR #7993, opened 2024-07-09
front: drop hardcoded scenario header height

emersion:
The scenario header size is hardcoded to 8rem. Replace these rules with a flexbox-based layout.

PR #8028, opened 2024-07-11
front: drop hardcoded scenario header height, second take

emersion:
_See individual commits._

This is a re-do of #7993, which got reverted in #8015 because of a bug when the timetable contains lots of trains.

- [x] Fix the bug described in #8012


Caracol3 (commented):
LGTM (tested)

kmer2016 (approved):
LGTM and tested !

emersion:
Dropped the first commit because similar changes have been made in Clara's lazy loading patches.

PR #8029, opened 2024-07-11
front: display restrictions reset warning when needed only

### LINKED ISSUE #8017: front: a warning is shown every time a pathfinding requests is sent
### What happened?

See:
image

Not have the popup "Les restrictions de puissance ont été réinitialisées."
(idea by @bougue-pe: Have the warning only when a manual power restriction was actually reset)

1. create a path (no need for a particular path)

- **Browser**: Firefox 

staging de8ada8

clarani:
closes #8017

### DIFF
--- a/front/src/modules/pathfinding/components/Itinerary/ItineraryV2.tsx
+++ b/front/src/modules/pathfinding/components/Itinerary/ItineraryV2.tsx
@@ -13,9 +13,11 @@ import { useOsrdConfActions, useOsrdConfSelectors } from 'common/osrdContext';
 import Tipped from 'common/Tipped';
 import PathfindingV2 from 'modules/pathfinding/components/Pathfinding/PathfindingV2';
 import TypeAndPathV2 from 'modules/pathfinding/components/Pathfinding/TypeAndPathV2';
+import { setWarning } from 'reducers/main';
 import { updateViewport } from 'reducers/map';
 import { getMap } from 'reducers/map/selectors';
 import { useAppDispatch } from 'store';
+import { isEmptyArray } from 'utils/array';
 
 import DestinationV2 from './DisplayItinerary/v2/DestinationV2';
 import OriginV2 from './DisplayItinerary/v2/OriginV2';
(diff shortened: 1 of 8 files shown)


theocrsb (approved):
LGTM and tested !

front/src/modules/pathfinding/components/Itinerary/ItineraryV2.tsx
```diff
@@ -57,13 +61,26 @@ const ItineraryV2 = ({
     }
   };
 
+  const toastRestrictionResetWarning = () => {
```
SharglutDev: I think it's better if a function start with a verb, is that ok for you ?

```suggestion
  const notifyPowerRestrictionsReset = () => {
```

front/src/modules/pathfinding/hook/usePathfinding.ts
```diff
@@ -248,20 +256,26 @@ export const usePathfindingV2 = (
                   positionOnPath: pathfindingResult.path_items_positions[i],
                   stopFor,
                   ...(correspondingOp && {
+                    name: correspondingOp.name,
+                    uic: correspondingOp.uic,
+                    ch: correspondingOp.ch,
```
SharglutDev, line 261: I'm not sure why you would want to erase these current op properties by the pathSteps ones ? They should be the same no ? 
We needed that for kp and coordinates because between two pathfindings, the path could sometimes use different tracks.

front/src/modules/pathfinding/components/Pathfinding/TypeAndPathV2.tsx
SharglutDev, line 1: Very nice clean, good catch !

SharglutDev (changes requested):
Thank you for this PR, nice cleaning ! Tested and it works great.

front/src/modules/pathfinding/components/Itinerary/ItineraryV2.tsx
clarani: Done

front/src/modules/pathfinding/hook/usePathfinding.ts
clarani: As discussed, we can keep this like this :+1: In the worst scenario, we only set again the props

SharglutDev (approved):
Lgtm !

PR #8030, opened 2024-07-11
front: localize missing editoast pathfinding error codes

emersion:
In #7883, core pathfinding errors have been localized. But I missed a few error codes that are generated by editoast instead of core: "invalid_path_item" and "rolling_stock_not_found". Localize these as well.

The editoast errors are defined here:
https://github.com/OpenRailAssociation/osrd/blob/769de63a18d74b6239b85dd2094ae713b56de2c1/editoast/src/core/v2/pathfinding.rs#L40


kmer2016 (approved):
LGTM

PR #8031, opened 2024-07-11
Specialize and simplify editoast_models error management

leovalais:
- `editoast_models` will become a big package, a single error type will become uselessly overcrowded
- the testing DbPoolV2 API can `.expect` more on errors since panics is the expected behavior
- prunes unused variants
- adds missing translation keys

> [!IMPORTANT]
> > And one more error for the front. That said, I'm not sure that we want such technical details be exposed to the final user. Maybe the message can be a little more generic? Something like "Service unavailable, please contact your admin". Front maintainers probably have a better opinion on this.
> https://github.com/OpenRailAssociation/osrd/pull/8031#pullrequestreview-2174136557


flomonster (approved):
LGTM (tested)

editoast/editoast_models/src/db_connection_pool.rs
```diff
@@ -262,41 +270,39 @@ impl DbConnectionPoolV2 {
     #[allow(unused)] // TEMPORARY
     pub fn iter_conn(
         &self,
-    ) -> impl Iterator<Item = impl Future<Output = Result<DbConnectionV2, EditoastModelsError>> + '_>
+    ) -> impl Iterator<Item = impl Future<Output = Result<DbConnectionV2, DatabasePoolError>> + '_>
     {
         std::iter::repeat(self).map(|p| p.get())
```
woshilapin: I know it's not part of the PR... but you can write this in one function call I believe.

```suggestion
        std::iter::repeat_with(|| self.get())
```

editoast/editoast_models/src/lib.rs
```diff
@@ -1,11 +1,9 @@
 use diesel_async::{pooled_connection::deadpool::Pool, AsyncPgConnection};
 
-mod db_connection_pool;
-mod error;
+pub mod db_connection_pool;
 
-pub use db_connection_pool::create_connection_pool;
+pub use db_connection_pool::DatabasePoolError;
 pub use db_connection_pool::DbConnectionPoolV2;
```
woshilapin: I'm not sure that I understand the logic about what should be publicly exposed from `lib.rs` and what should stay namespaced in `db_connection_pool`. If you make `db_connection_pool` public, then maybe it's not necessary to make public any of the underlying object?

For example, here, only `DatabasePoolError` is exposed... but in the diff of `editoast/src/error.rs`, there is the following code
```rust

use editoast_models::db_connection_pool::DatabasePoolBuildError;
use editoast_models::db_connection_pool::DatabasePoolError;
```

woshilapin (approved):
And one more error for the front. That said, I'm not sure that we want such technical details be exposed to the final user. Maybe the message can be a little more generic? Something like "Service unavailable, please contact your admin". Front maintainers probably have a better opinion on this.

Apart from that, thanks for the improvements, they're nice!

Wadjetz (approved):

editoast/editoast_models/src/db_connection_pool.rs
leovalais: It's not a big change and it's close to a diff of the PR so imo it's a relevant addition :)

editoast/editoast_models/src/lib.rs
leovalais: My logic here was that since `DbConnectionPoolV2` is one of the most commonly used type of `models`, it's convenient to expose it at the root of the package. Likewise `create_connection_pool` won't be used often, so it's fine to ask its caller to import the full module path.

`pub use DatabasePoolError` was a bad last minute change, thanks for pointing it out 😬

SharglutDev (approved):
Lgtm (not tested)

PR #8032, opened 2024-07-11
front: fix TSv2 times and stop table input

### LINKED ISSUE #8019: TSV2 : inserted input are not displayed in the table
### What happened?

We create a train. We try to enter some input values (arrival time, stopping time or theoretical margin). These values are not displayed on the table altought they are used in the calculation. If we look at simulation results, the values are taken into account.
If we edit the train, the input table is still empty. Sometimes, we manage to get the values ​​by reloading the page or if we try to create another train.

The values should be displayed 

Go to a scenario TSV2
Add a train
Go to times and stops table 
Put a margin value for example

- **Browser**: Edge
- **OS**: Windows
- **Env**: Recette

staging de8ada8

emersion:
_See individual commits._

Closes: https://github.com/OpenRailAssociation/osrd/issues/8019

### DIFF
--- a/front/src/modules/pathfinding/utils.ts
+++ b/front/src/modules/pathfinding/utils.ts
@@ -34,6 +34,23 @@ export const formatSuggestedOperationalPoints = (
     coordinates: getPointCoordinates(geometry, pathLength, op.position),
   }));
 
+export const matchPathStepAndOp = (step: PathStep, op: SuggestedOP) => {
+  if ('operational_point' in step) {
+    return step.operational_point === op.opId;
+  }
+  // We match the kp in case two OPs have the same uic+ch (can happen when the
+  // infra is imported)
+  if ('uic' in step) {
+    return step.uic === op.uic && step.ch === op.ch && step.kp === op.kp;
+  }
+  if ('trigram' in step) {
+    return step.trigram === op.trigram && step.ch === op.ch && step.kp === op.kp;
+  }
+  // TODO: we abuse the PathStep.id field here, the backend also sets it to an
+  // ID which has nothing to do with OPs
+  return step.id === op.opId;
+};
+
 export const getPathfindingQuery = ({
   infraId,
   rollingStock,
--- a/front/src/modules/timesStops/utils.ts
+++ b/front/src/modules/timesStops/utils.ts
@@ -1,5 +1,6 @@
 import type { TFunction } from 'i18next';
 
+import { matchPathStepAndOp } from 'modules/pathfinding/utils';
 import type { SuggestedOP } from 'modules/trainschedule/components/ManageTrainSchedule/types';
 import type { PathStep } from 'reducers/osrdconf/types';
 import { marginRegExValidation } from 'utils/physics';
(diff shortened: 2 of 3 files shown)


front/src/modules/timesStops/utils.ts
```diff
@@ -20,32 +21,25 @@ export const formatSuggestedViasToRowVias = (
       (op) => op.uic === origin.uic && op.ch === origin.ch && op.name === origin.name
     );
     // If the origin is in the ops and isn't the first operational point, we need to move it to the first position
-    if (originIndexInOps !== (-1 || 0)) {
+    if (originIndexInOps !== -1) {
```
SharglutDev: It seems like we should do the same test for destination.

On the screen below, I did a path between St Lazare and Dieppe by selecting both origin/via by map click, and the destination is not the last in the list as they both have the same `positionOnPath`

Capture d’écran 2024-07-12 à 09 00 06

front/src/modules/pathfinding/utils.ts
```diff
@@ -34,6 +34,21 @@ export const formatSuggestedOperationalPoints = (
...
 
+export const matchPathStepAndOp = (step: PathStep, op: SuggestedOP) => {
+  if ('operational_point' in step) {
+    return step.operational_point === op.opId;
+  }
+  if ('uic' in step) {
+    return step.uic === op.uic && step.ch === op.ch;
+  }
+  if ('trigram' in step) {
+    return step.trigram === op.trigram && step.ch === op.ch;
```
SharglutDev: I feel like we should check more thant that. Because of some imports issues, we can sometimes have multiple (2 most of the time) ops with same uic/ch. As it's done in other parts in the app, we could add a check for `kp` as well (between gaia and siph I think, the same op can be present 2 times with same uic/ch but two different `kp`) so we are sure we will edit the good one.

SharglutDev (changes requested):
Thank you for this nice fix and very well documented commits ! The fix works great, just left two comments.

flomonster (commented):
Tested and it works

front/src/modules/pathfinding/utils.ts
emersion: That makes sense to me in general. Question: is there a downside to always check the `kp`? Looking at `isVia()`, it seems like we sometimes check `kp` and sometimes not - in which situation do we want to ignore `kp`?

emersion: Also, should we match the `name`?

SharglutDev: If I remember correctly (should have better documented this), in TimesStops we prefer to check the isVia with kp because some of ops in the list have same uic/ch/name so name is not relevant to compare them.
In SuggestedViaModal, we checked for the name because we decided that the "duplicated" op should add or removed together. But it seems that now it's not iso with TimeStops anymore and it shouldn't be like that.

front/src/modules/timesStops/utils.ts
emersion: Fixed in https://github.com/OpenRailAssociation/osrd/pull/8037

front/src/modules/pathfinding/utils.ts
emersion: Alright, added the `kp` matching!

SharglutDev (approved):

PR #8034, opened 2024-07-11
front: do not modify rolling stock search filter and trim it

woshilapin:
When searching for rolling stock, the input does modify what is entered (making it lower case). It is a disturbing user experience to type a capital letter and see a lower case letter being displayed.

Also, if a leading space is typed, the search won't work. We might want to trim the search input before using it.

### DIFF
--- a/front/src/modules/rollingStock/hooks/useFilterRollingStock.tsx
+++ b/front/src/modules/rollingStock/hooks/useFilterRollingStock.tsx
@@ -55,7 +55,7 @@ function rollingStockPassesSearchedStringFilter(
     return true;
   }
   function includesSearchedString(str: string | undefined) {
-    return str && str.toLowerCase().includes(filters.text);
+    return str?.trim().toLowerCase().includes(filters.text.trim().toLowerCase());
   }
   return [
     name,
@@ -142,7 +142,7 @@ export default function useFilterRollingStock() {
     useState<LightRollingStockWithLiveries[]>(allRollingStocks);
 
   const searchRollingStock = (value: string) => {
-    setFilters({ ...filters, text: value.toLowerCase() });
+    setFilters({ ...filters, text: value });
     setSearchIsLoading(true);
   };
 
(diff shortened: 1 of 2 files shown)


Uriel-Sautron (approved):
LGTM, thank you for this refacto !! Tested.

flomonster (approved):
LGTM (tested)

emersion (approved):

kmer2016 (approved):
LGTM. Thanks for updating the tests too !

front/src/modules/rollingStock/hooks/useFilterRollingStock.tsx
```diff
@@ -55,7 +55,7 @@ function rollingStockPassesSearchedStringFilter(
     return true;
   }
   function includesSearchedString(str: string | undefined) {
-    return str && str.toLowerCase().includes(filters.text);
+    return str && str.trim().toLowerCase().includes(filters.text.trim().toLowerCase());
```
kmer2016:
```suggestion
    return str?.trim().toLowerCase().includes(filters.text.trim().toLowerCase());
```

emersion: I think this changes the type of the returned value: was `boolean` before, will be `boolean | undefined` with the suggested change. Maybe it doesn't matter too much though.

anisometropie: `str && str.trim()…` was already returning `boolean | undefined`.

`&&` returns the left operand `str` if it is falsy, and the right operand otherwise.

`undefined && (any expression)` evaluates as `undefined`

emersion: Hmmm, that's not what my devtools are saying? Am I missing something?

```
> undefined || true
true
> undefined || false
false 
```

emersion: Yup, indeed, you're right! Disregard my comment then.

woshilapin: Nice, I learned a few things with your comments. Thank you 🫶 Done in [bf7f62a](https://github.com/OpenRailAssociation/osrd/compare/4a93d831c73b848f20ce5f8d99032647b4fe2346..bf7f62a68289458a4423ea06ac8a9a840110d7a1). Since it's a small fix, I'm resolving the comment.

achrafmohye (commented):
Looking good 👍

PR #8035, opened 2024-07-11
Improve batch pathfinding performances

flomonster:
## Description

This PR tries to improve calls to `simulation_summary` when train schedules are **not cached** to redis.

- Fixing the `--no-cache` option
- Reducing the number of call perform to the database

## Benchmark

- Run `simulation_summary` endpoint on 30 train schedules timetable (all valid)
- `--no-cache` option is enabled

### Before

Time: 1.15s
Number of DB queries: 236 

begore

### After

Time: 937ms
Number of DB queries: 5 

after

### DIFF
--- a/editoast/src/views/v2/path/path_item_cache.rs
+++ b/editoast/src/views/v2/path/path_item_cache.rs
@@ -1,0 +1,256 @@
+use crate::core::v2::pathfinding::PathfindingResult;
+use crate::error::Result;
+use crate::modelsv2::TrackSectionModel;
+use crate::RetrieveBatchUnchecked;
+use editoast_schemas::infra::TrackOffset;
+use std::collections::HashMap;
+use std::collections::HashSet;
+
+use editoast_models::DbConnection;
+use editoast_schemas::train_schedule::PathItemLocation;
+
+use crate::modelsv2::OperationalPointModel;
+
+type TrackOffsetResult = std::result::Result<Vec<Vec<TrackOffset>>, PathfindingResult>;
+
+/// Gather information about several path items, factorizing db calls.
+pub struct PathItemCache {
+    uic_to_ops: HashMap<i64, Vec<OperationalPointModel>>,
+    trigram_to_ops: HashMap<String, Vec<OperationalPointModel>>,
+    ids_to_ops: HashMap<String, OperationalPointModel>,
+    existing_track_ids: HashSet<String>,
+}
+
+impl PathItemCache {
+    /// Load the path item cache from a list of pathfinding inputs
+    pub async fn load(
+        conn: &mut DbConnection,
+        infra_id: i64,
+        path_items: &[&PathItemLocation],
+    ) -> Result<PathItemCache> {
+        let (trigrams, ops_uic, ops_id) = collect_path_item_ids(path_items);
+        let uic_to_ops = retrieve_op_from_uic(conn, infra_id, &ops_uic).await?;
+        let trigram_to_ops = retrieve_op_from_trigrams(conn, infra_id, &trigrams).await?;
+        let ids_to_ops = retrieve_op_from_ids(conn, infra_id, &ops_id).await?;
+
+        let tracks = path_items.iter().filter_map(|item| match item {
+            PathItemLocation::TrackOffset(TrackOffset { track, .. }) => {
+                Some((infra_id, track.0.clone()))
+            }
+            _ => None,
+        });
+        let tracks = ids_to_ops
+            .values()
+            .chain(trigram_to_ops.values().flatten())
+            .chain(uic_to_ops.values().flatten())
+            .flat_map(|op| &op.parts)
+            .map(|part| (infra_id, part.track.0.clone()))
+            .chain(tracks);
+        let existing_track_ids =
+            TrackSectionModel::retrieve_batch_unchecked::<_, Vec<_>>(conn, tracks)
+                .await?
+                .into_iter()
+                .map(|track| track.obj_id)
+                .collect();
+
+        Ok(PathItemCache {
+            uic_to_ops,
+            trigram_to_ops,
+            ids_to_ops,
+            existing_track_ids,
+        })
+    }
+
+    /// Get the operational points associated with an identifier
+    pub fn get_from_id(&self, id: &str) -> Option<&OperationalPointModel> {
+        self.ids_to_ops.get(id)
+    }
+
+    /// Get the operational points associated with a trigram
+    pub fn get_from_trigram(&self, trigram: &str) -> Option<&Vec<OperationalPointModel>> {
+        self.trigram_to_ops.get(trigram)
+    }
+
+    /// Get the operational points associated with a UIC code
+    pub fn get_from_uic(&self, uic: i64) -> Option<&Vec<OperationalPointModel>> {
+        self.uic_to_ops.get(&uic)
+    }
+
+    /// Check if a track exists
+    pub fn track_exists(&self, track: &str) -> bool {
+        self.existing_track_ids.contains(track)
+    }
+
+    /// Extract locations from path items
+    pub fn extract_location_from_path_items(
+        &self,
+        path_items: &[&PathItemLocation],
+    ) -> TrackOffsetResult {
+        let mut result: Vec<Vec<_>> = Vec::default();
+        for (index, &path_item) in path_items.iter().enumerate() {
+            let track_offsets = match path_item {
+                PathItemLocation::TrackOffset(track_offset) => {
+                    vec![track_offset.clone()]
+                }
+                PathItemLocation::OperationalPointId { operational_point } => {
+                    match self.get_from_id(&operational_point.0) {
+                        Some(op) => op.track_offset(),
+                        None => {
+                            return Err(PathfindingResult::InvalidPathItem {
+                                index,
+                                path_item: path_item.clone(),
+                            });
+                        }
+                    }
+                }
+                PathItemLocation::OperationalPointDescription {
+                    trigram,
+                    secondary_code,
+                } => {
+                    let ops = self
+                        .get_from_trigram(&trigram.0)
+                        .cloned()
+                        .unwrap_or_default();
+                    let ops = secondary_code_filter(secondary_code, ops);
+                    if ops.is_empty() {
+                        return Err(PathfindingResult::InvalidPathItem {
+                            index,
+                            path_item: path_item.clone(),
+                        });
+                    }
+                    track_offsets_from_ops(&ops)
+                }
+                PathItemLocation::OperationalPointUic {
+                    uic,
+                    secondary_code,
+                } => {
+                    let ops = self
+                        .get_from_uic(i64::from(*uic))
+                        .cloned()
+                        .unwrap_or_default();
+                    let ops = secondary_code_filter(secondary_code, ops);
+                    if ops.is_empty() {
+                        return Err(PathfindingResult::InvalidPathItem {
+                            index,
+                            path_item: path_item.clone(),
+                        });
+                    }
+                    track_offsets_from_ops(&ops)
+                }
+            };
+
+            // Check if tracks exist
+            for track_offset in &track_offsets {
+                if !self.track_exists(&track_offset.track.0) {
+                    return Err(PathfindingResult::InvalidPathItem {
+                        index,
+                        path_item: path_item.clone(),
+                    });
+                }
+            }
+
+            result.push(track_offsets);
+        }
+        Ok(result)
+    }
+}
+
+/// Collect the ids of the operational points from the path items
+fn collect_path_item_ids(path_items: &[&PathItemLocation]) -> (Vec<String>, Vec<i64>, Vec<String>) {
+    let mut trigrams: Vec<String> = Vec::new();
+    let mut ops_uic: Vec<i64> = Vec::new();
+    let mut ops_id: Vec<String> = Vec::new();
+
+    for item in path_items {
+        match item {
+            PathItemLocation::OperationalPointDescription { trigram, .. } => {
+                trigrams.push(trigram.clone().0);
+            }
+            PathItemLocation::OperationalPointUic { uic, .. } => {
+                ops_uic.push(i64::from(*uic));
+            }
+            PathItemLocation::OperationalPointId {
+                operational_point, ..
+            } => {
+                ops_id.push(operational_point.clone().0);
+            }
+            _ => {}
+        }
+    }
+    (trigrams, ops_uic, ops_id)
+}
+
+/// Retrieve operational points from operational point uic codes
+async fn retrieve_op_from_uic(
+    conn: &mut DbConnection,
+    infra_id: i64,
+    ops_uic: &[i64],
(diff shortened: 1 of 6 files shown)


editoast/src/views/v2/path/path_item_cache.rs
```diff
@@ -0,0 +1,149 @@
...
+                ops_id.push(operational_point.clone().0);
+            }
+            _ => {}
+        }
+    }
+    (trigrams, ops_uic, ops_id)
+}
+
+/// Retrieve operational points from locations
+async fn retrieve_op_from_locations(
```
Castavo: This could be three separate functions

editoast/src/views/v2/path/path_item_cache.rs
```diff
@@ -0,0 +1,149 @@
...
+        self.trigram_to_ops.get(trigram)
+    }
+
+    /// Get the operational points associated with a UIC code
+    pub fn get_uic(&self, uic: i64) -> Option<&Vec<OperationalPointModel>> {
+        self.uic_to_ops.get(&uic)
+    }
+
+    /// Check if a track exists
+    pub fn is_track_valid(&self, track: &String) -> bool {
```
Castavo: Using the word `validity` seems confusing to me, its sense is less precise than existing or not
```suggestion
    pub fn track_exists(&self, track: &String) -> bool {
```

editoast/src/views/v2/path/path_item_cache.rs
```diff
@@ -0,0 +1,149 @@
...
+use editoast_schemas::infra::TrackOffset;
+use std::collections::HashMap;
+use std::collections::HashSet;
+
+use editoast_models::DbConnection;
+use editoast_schemas::train_schedule::PathItemLocation;
+
+use crate::modelsv2::OperationalPointModel;
+
+/// Contains information about the location of a path item.
```
Castavo: You could insist on the fact that it is meant to reduce the number of db calls
```suggestion
/// Gather information about several path items, factorizing db calls. 
```

editoast/src/views/v2/path/pathfinding.rs
```diff
@@ -52,7 +49,7 @@ editoast_common::schemas! {
 /// and a list of path waypoints
 #[derive(Deserialize, Clone, Debug, Hash, ToSchema)]
 #[schema(as = PathfindingInputV2)]
-struct PathfindingInput {
+pub struct PathfindingInput {
```
woshilapin: I believe you can remove this `pub`.

editoast/src/views/v2/path/pathfinding.rs
```diff
@@ -64,7 +61,7 @@ struct PathfindingInput {
     rolling_stock_supported_signaling_systems: Vec<String>,
     /// List of waypoints given to the pathfinding
     #[schema(inline)]
-    path_items: Vec<PathItemLocation>,
+    pub path_items: Vec<PathItemLocation>,
```
woshilapin: I believe you can remove this `pub`.

editoast/src/views/v2/path/pathfinding.rs
```diff
@@ -186,27 +191,28 @@ async fn pathfinding_blocks_batch(
...
     // Check if tracks exist
-    if let Err(err) = check_tracks_from_path_items(conn, infra.id, &track_offsets).await? {
-        return Ok(Err(err));
+    for tracks in &track_offsets {
+        for track_offset in tracks {
+            if !path_item_cache.is_track_valid(&track_offset.track.0) {
+                return Err(PathfindingResult::NotFoundInTracks);
+            }
+        }
     }
```
Castavo: This feels like it belongs in `extract_location_from_path_items`
If I'm not mistaken, it's no use checking that operational point's tracks do exist, since the infra would be invalid anyway.
This check could then be performed in the `PathItemLocation::TrackOffset` branch of the mentionned function.

editoast/src/views/v2/path/pathfinding.rs
```diff
@@ -392,19 +329,13 @@ impl From<TrackOffsetExtractionError> for PathfindingResult {
...
-    infra_id: i64,
-    path_items: &[PathItemLocation],
-) -> Result<TrackOffsetResult> {
-    let (trigrams, ops_uic, ops_id) = collect_path_item_ids(path_items);
-
-    let (uic_to_ops, trigrams_to_ops, ids_to_ops) =
-        retrieve_op_from_locations(conn, infra_id, &ops_uic, &trigrams, &ops_id).await?;
-
+/// Extract locations from path items
+pub fn extract_location_from_path_items(
```
Castavo: This could be a method of `PathItemCache`

Castavo (commented):
Nice optimization !

(This patch has 0% coverage 👀)

editoast/src/views/v2/path/path_item_cache.rs
Castavo: Oh this is a copy paste from an old function, refactoring is kind of out of scope then

editoast/src/views/v2/path/pathfinding.rs
flomonster: Actually, it's checking that operational point's tracks do exist. The infra would be invalid but I this way we trigger a clean error.

Castavo: I feel like the responsibility of checking whether operational point's track sections exist does not belong to this function. 
Maybe the track locations in the `PathItemCache` could be filtered to contain only existing tracks for each operational point ?
This is very subjective, feel free to just resolve this conversation if you disagree

editoast/src/views/v2/path/pathfinding.rs
flomonster: Done

editoast/src/views/v2/path/path_item_cache.rs
flomonster: Done

editoast/src/views/v2/path/path_item_cache.rs
flomonster: Done

editoast/src/views/v2/path/path_item_cache.rs
flomonster: Done

editoast/src/views/v2/path/path_item_cache.rs
```diff
@@ -0,0 +1,258 @@
...
+use crate::modelsv2::OperationalPointModel;
+
+type TrackOffsetResult = std::result::Result<Vec<Vec<TrackOffset>>, TrackOffsetExtractionError>;
+
+/// Gather information about several path items, factorizing db calls.
+pub struct PathItemCache {
+    uic_to_ops: HashMap<i64, Vec<OperationalPointModel>>,
+    trigram_to_ops: HashMap<String, Vec<OperationalPointModel>>,
+    ids_to_ops: HashMap<String, OperationalPointModel>,
+    valid_tracks: HashSet<String>,
```
Castavo: This could also be renamed
```suggestion
    existing_track_ids: HashSet<String>,
```

Castavo (approved):
Neat

editoast/src/views/v2/path/path_item_cache.rs
```diff
@@ -0,0 +1,258 @@
...
+        Ok(PathItemCache {
+            uic_to_ops,
+            trigram_to_ops,
+            ids_to_ops,
+            valid_tracks,
+        })
+    }
+
+    /// Get the operational points associated with an identifier
+    pub fn get_id(&self, id: &String) -> Option<&OperationalPointModel> {
```
woshilapin: It's weird to call it `get_id` because you don't get an identifier as a return value. So I suggest changing it like this.
```suggestion
    pub fn get_from_id(&self, id: &String) -> Option<&OperationalPointModel> {
```

editoast/src/views/v2/path/path_item_cache.rs
```diff
@@ -0,0 +1,258 @@
...
+        self.ids_to_ops.get(id)
+    }
+
+    /// Get the operational points associated with a trigram
+    pub fn get_trigram(&self, trigram: &String) -> Option<&Vec<OperationalPointModel>> {
```
woshilapin:
```suggestion
    pub fn get_from_trigram(&self, trigram: &String) -> Option<&Vec<OperationalPointModel>> {
```

editoast/src/views/v2/path/path_item_cache.rs
```diff
@@ -0,0 +1,258 @@
...
+        self.trigram_to_ops.get(trigram)
+    }
+
+    /// Get the operational points associated with a UIC code
+    pub fn get_uic(&self, uic: i64) -> Option<&Vec<OperationalPointModel>> {
```
woshilapin:
```suggestion
    pub fn get_from_uic(&self, uic: i64) -> Option<&Vec<OperationalPointModel>> {
```

editoast/src/views/v2/path/path_item_cache.rs
```diff
@@ -0,0 +1,258 @@
...
+        self.uic_to_ops.get(&uic)
+    }
+
+    /// Check if a track exists
+    pub fn track_exists(&self, track: &String) -> bool {
```
woshilapin: In general, always use `&str` instead of `&String` for an API, it accepts more (for example, it will accept both `&String::from("super")` and `"super"`).

```suggestion
    pub fn track_exists(&self, track: &str) -> bool {
```

editoast/src/views/v2/path/path_item_cache.rs
```diff
@@ -0,0 +1,258 @@
...
+        path_items: &[&PathItemLocation],
+    ) -> TrackOffsetResult {
+        let mut result: Vec<Vec<_>> = Vec::default();
+        for (index, &path_item) in path_items.iter().enumerate() {
+            let track_offsets = match path_item {
+                PathItemLocation::TrackOffset(track_offset) => {
+                    vec![TrackOffset {
+                        track: track_offset.track.clone(),
+                        offset: track_offset.offset,
+                    }]
```
woshilapin:
```suggestion
                    vec![track_offset.clone()]
```

editoast/src/views/v2/path/path_item_cache.rs
```diff
@@ -0,0 +1,258 @@
...
+                            path_item: path_item.clone(),
+                        });
+                    }
+                    track_offsets_from_ops(&ops)
+                }
+                PathItemLocation::OperationalPointUic {
+                    uic,
+                    secondary_code,
+                } => {
+                    let ops = self.get_uic(*uic as i64).cloned().unwrap_or_default();
```
woshilapin:
```suggestion
                    let ops = self.get_uic(i64::from(*uic)).cloned().unwrap_or_default();
```

editoast/src/views/v2/path/path_item_cache.rs
```diff
@@ -0,0 +1,258 @@
...
+    let mut ops_uic: Vec<i64> = Vec::new();
+    let mut ops_id: Vec<String> = Vec::new();
+
+    for item in path_items {
+        match item {
+            PathItemLocation::OperationalPointDescription { trigram, .. } => {
+                trigrams.push(trigram.clone().0);
+            }
+            PathItemLocation::OperationalPointUic { uic, .. } => {
+                ops_uic.push(*uic as i64);
```
woshilapin:
```suggestion
                ops_uic.push(i64::from(*uic));
```

editoast/src/views/v2/path/pathfinding.rs
```diff
@@ -186,27 +187,28 @@ async fn pathfinding_blocks_batch(
...
+    let path_items: Vec<_> = pathfinding_input.path_items.iter().collect();
     if path_items.len() <= 1 {
-        return Ok(Err(PathfindingResult::NotEnoughPathItems));
+        return Err(PathfindingResult::NotEnoughPathItems);
     }
-    let track_offsets = match extract_location_from_path_items(conn, infra.id, path_items).await? {
-        Ok(track_offsets) => track_offsets,
-        Err(e) => return Ok(Err(e.into())),
-    };
+    let track_offsets = path_item_cache.extract_location_from_path_items(&path_items)?;
```
woshilapin, line 199: I wonder if we should change something about the naming. We create a variable with "offsets" but the function talks about "location". Maybe "offset" is more used everywhere?

editoast/src/views/v2/path/pathfinding.rs
```diff
@@ -186,27 +187,28 @@ async fn pathfinding_blocks_batch(
...
     // Check if tracks exist
-    if let Err(err) = check_tracks_from_path_items(conn, infra.id, &track_offsets).await? {
-        return Ok(Err(err));
+    for tracks in &track_offsets {
+        for track_offset in tracks {
+            if !path_item_cache.track_exists(&track_offset.track.0) {
+                return Err(PathfindingResult::NotFoundInTracks);
+            }
+        }
     }
```
woshilapin: Something feels weird with that piece of code. You calculate `track_offsets` from `path_item_cache`. Then you iterate on all the track offsets to ask again to check if the track exists to `path_item_cache`. Shouldn't all of this be done directly in `extract_location_from_path_items`?

woshilapin (commented):
Mostly naming considerations. In general, this should help a lot to be more gentle on the DB, so nice work!

editoast/src/views/v2/path/pathfinding.rs
flomonster: It will no longer be possible to differentiate between an `InvalidPathItem` and a `NotFoundInTracks`.

woshilapin (approved):

editoast/src/views/v2/path/pathfinding.rs
flomonster: > I feel like the responsibility of checking whether operational point's track sections exist does not belong to this function. 

I agree but I don't know how to move things around to make it more readable.

editoast/src/views/v2/path/path_item_cache.rs
flomonster: Good catch

editoast/src/views/v2/path/pathfinding.rs
flomonster: Yes, it's an OSRD convention (it comes from core). The idea is to call:
- `location`: The association of a track with its offset
- `position`: An offset on a path

editoast/src/views/v2/path/pathfinding.rs
flomonster: I think it's linked to this comment: https://github.com/OpenRailAssociation/osrd/pull/8035#discussion_r1675588199

If we handle this in `extract_location_from_path_items`, we lose precision on the error type.

editoast/src/views/v2/path/pathfinding.rs
woshilapin: Then why is there a `TrackOffset`? Should it be named `Location`?

flomonster: Won't do it in this PR.

PR #8037, opened 2024-07-12
front: ensure destination is last in formatSuggestedViasToRowVias()

emersion:
See individual commits (some cleanups are also included).

See https://github.com/OpenRailAssociation/osrd/pull/8032#discussion_r1675411754

Note, the first commit has been moved from #8032 to this PR to avoid merge conflicts.

### DIFF
--- a/front/src/modules/timesStops/utils.ts
+++ b/front/src/modules/timesStops/utils.ts
@@ -16,17 +16,13 @@ export const formatSuggestedViasToRowVias = (
 ): PathWaypointColumn[] => {
   const formattedOps = [...operationalPoints];
   const origin = pathSteps[0];
-  if ('uic' in origin && 'ch' in origin) {
-    const originIndexInOps = operationalPoints.findIndex(
-      (op) => op.uic === origin.uic && op.ch === origin.ch && op.name === origin.name
-    );
-    // If the origin is in the ops and isn't the first operational point, we need to move it to the first position
-    if (originIndexInOps !== -1) {
-      [formattedOps[0], formattedOps[originIndexInOps]] = [
-        formattedOps[originIndexInOps],
-        formattedOps[0],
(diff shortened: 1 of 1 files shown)


emersion:
I tried isolating the logic in a function to avoid repetition, but ended up with something that was IMHO harder to read. (Maybe could have a function to do the `findIndex()` call?)

front/src/modules/timesStops/utils.ts
SharglutDev, line 1: I think your logic works great if origin/destination comes from the rocket but as we check only `if ('uic' in origin && 'ch' in origin)`, it won't work for an origin/destination added by map click (to test it, you can do a Paris St Lazare - Dieppe by clicking on map and see the destination at the wrong place).

SharglutDev (changes requested):

front/src/modules/timesStops/utils.ts
emersion: Do you think it would make sense to leverage `matchPathStepAndOp()` here? I was worried about breaking stuff since the checks are a bit different (`name` is checked, not `kp`).

SharglutDev: I thought about your new util too which should be the best way to change this behavior. I just tested by putting `kp` instead of `name` and it seems to be working fine as well.
An easy test with multiple ops at position 0, is using a path "PNO AS". If you comment the origin switch logic, you will see the origin at the wrong place in the table and both fix with kp or name works.

emersion: Done! See [c1f1b45](https://github.com/OpenRailAssociation/osrd/pull/8037/commits/c1f1b452d244bb6a78a203825d92afc916937547).

SharglutDev (approved):
Lgtm and tested

PR #7165, opened 2024-04-10
Fix pathfinding v2 using uic or trigrams

flomonster:

PR #8047, opened 2024-07-14
Bye actix

Khoyo:
5yvReCm

# To verify

- [x] The app works 🙃 
    * liveries import
    * timetable import 
- [x] Telemetry works
- [x] The deleted tests can indeed be removed

> [!NOTE]
> Commits are overly split to ease the review, but will be squashed together before merging ofc.

### Main motivation

* the actix execution model was a little strange: it created one thread per worker and each of these threads had a monothread async runtime
* axum's API is more user-friendly and better documented
* actix tends to redefine its entire stack, whereas axum tends to work with other popular libs (hyper, tower, etc.).
* axum is maintained by the tokio team, so we can expect good integration (same for tracing).

### DIFF
--- a/editoast/Cargo.toml
+++ b/editoast/Cargo.toml
@@ -55,23 +55,21 @@ serde_json = "1.0.120"
 strum = { version = "0.26.3", features = ["derive"] }
 tempfile = "3.10.1"
 thiserror = "1.0.63"
-tokio = { version = "1.39.1", features = ["macros", "rt-multi-thread"] }
+tokio = { version = "1.39.1", features = ["fs", "macros", "rt-multi-thread"] }
 tokio-postgres = "0.7.11"
 tracing = { version = "0.1.40", features = ["log"] }
 url = { version = "2.5.2", features = ["serde"] }
-utoipa = { version = "4.2.3", features = ["actix_extras", "chrono", "uuid"] }
+utoipa = { version = "4.2.3", features = ["chrono", "uuid"] }
 uuid = { version = "1.10.0", features = ["serde", "v4"] }
 
 [dependencies]
 # For batch dependency updates, see editoast/README.md
 
-actix-cors = "0.7.0"
-actix-files = "0.6.6"
-actix-http = "3.8.0"
-actix-multipart = "0.7.2"
-actix-web = "4.8.0"
-actix-web-opentelemetry = { version = "0.18.0", features = ["awc"] }
 async-trait = "0.1.81"
+axum = { version = "0.7.5", features = ["multipart"] }
+axum-extra = { version = "0.9.3", features = ["typed-header"] }
+axum-test = "15.3.0"
+axum-tracing-opentelemetry = "0.19.0"
 chashmap = "2.2.2"
 chrono.workspace = true
 clap = { version = "4.5.10", features = ["derive", "env"] }
@@ -90,10 +88,12 @@ enumset = "1.1.5"
 futures.workspace = true
 futures-util.workspace = true
 geos.workspace = true
+headers = "0.4.0"
 image = "0.25.2"
 inventory = "0.3"
 itertools.workspace = true
 json-patch = { version = "2.0.0", features = ["utoipa"] }
+mime = "0.3.17"
 mvt.workspace = true
 opentelemetry = "0.23.0"
 opentelemetry-datadog = { version = "0.11.0", features = ["reqwest-client"] }
@@ -112,18 +112,27 @@ redis = { version = "0.25.4", features = [
   "tokio-comp",
   "tokio-native-tls-comp",
 ] }
+regex = "1.10.5"
 # 0.12.0 to 0.12.4 have weird timeout issues https://github.com/seanmonstar/reqwest/issues/2283
 # This bug was introduced between 0.12.0 and 0.12.3.
 reqwest = { version = "0.11.27", features = ["json"] }
 serde.workspace = true
 serde_derive.workspace = true
 serde_json.workspace = true
-serde_qs = { version = "0.13.0", features = ["actix4"] }
+serde_qs = { version = "0.13.0", features = ["axum"] }
 serde_yaml = "0.9.34"
 sha1 = "0.10"
 strum.workspace = true
 thiserror.workspace = true
 tokio.workspace = true
+tokio-util = { version = "0.7.11", features = ["io", "tracing"] }
+tower = "0.4.13"
+tower-http = { version = "0.5.2", features = [
+  "cors",
+  "limit",
+  "normalize-path",
+  "trace",
+] }
 tracing.workspace = true
 tracing-opentelemetry = "0.24.0"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
@@ -134,6 +143,7 @@ validator = { version = "0.18.1", features = ["derive"] }
 
 [dev-dependencies]
 async-std = { version = "1.12.0", features = ["attributes", "tokio1"] }
+axum = { version = "0.7.5", features = ["macros", "multipart"] }
 editoast_models = { workspace = true, features = ["testing"] }
 pretty_assertions = "1.4.0"
 rstest.workspace = true
--- a/editoast/openapi.yaml
+++ b/editoast/openapi.yaml
@@ -7,7 +7,7 @@ info:
     url: https://www.gnu.org/licenses/lgpl-3.0.html
   version: 0.1.0
 paths:
-  /documents/:
+  /documents:
     post:
       tags:
       - documents
@@ -81,7 +81,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/InternalError'
-  /electrical_profile_set/:
+  /electrical_profile_set:
     get:
       tags:
       - electrical_profiles
@@ -118,7 +118,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ElectricalProfileSet'
-  /electrical_profile_set/{electrical_profile_set_id}/:
+  /electrical_profile_set/{electrical_profile_set_id}:
     get:
       tags:
       - electrical_profiles
@@ -163,7 +163,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/InternalError'
-  /electrical_profile_set/{electrical_profile_set_id}/level_order/:
+  /electrical_profile_set/{electrical_profile_set_id}/level_order:
     get:
       tags:
       - electrical_profiles
@@ -208,7 +208,7 @@ paths:
             text/plain:
               schema:
                 type: string
-  /infra/:
+  /infra:
     get:
       tags:
       - infra
@@ -374,7 +374,7 @@ paths:
               - 2500.5V
         '404':
           description: The infra was not found
-  /infra/{infra_id}/:
+  /infra/{infra_id}:
     get:
       tags:
       - infra
@@ -522,7 +522,7 @@ paths:
                   type: array
                   items:
                     type: string
-  /infra/{infra_id}/auto_fixes/:
+  /infra/{infra_id}/auto_fixes:
     get:
       tags:
       - infra
@@ -650,7 +650,7 @@ paths:
                         properties:
                           information:
                             $ref: '#/components/schemas/InfraError'
-  /infra/{infra_id}/lines/{line_code}/bbox/:
+  /infra/{infra_id}/lines/{line_code}/bbox:
     get:
       tags:
       - infra
@@ -752,7 +752,7 @@ paths:
           description: Duplicate object ids provided
         '404':
           description: Object ID or infra ID invalid
-  /infra/{infra_id}/pathfinding/:
+  /infra/{infra_id}/pathfinding:
     post:
       tags:
       - infra
@@ -1092,7 +1092,7 @@ paths:
               - 2500.5V
         '404':
           description: The infra was not found
-  /layers/layer/{layer_slug}/mvt/{view_slug}/:
+  /layers/layer/{layer_slug}/mvt/{view_slug}:
     get:
       tags:
       - layers
@@ -1160,7 +1160,7 @@ paths:
                     - http://localhost:7070/tile/track_sections/geo/{z}/{x}/{y}/?infra=1
                   type:
                     type: string
-  /layers/tile/{layer_slug}/{view_slug}/{z}/{x}/{y}/:
+  /layers/tile/{layer_slug}/{view_slug}/{z}/{x}/{y}:
     get:
       tags:
       - layers
@@ -1211,7 +1211,7 @@ paths:
               schema:
                 type: string
                 format: binary
-  /light_rolling_stock/:
+  /light_rolling_stock:
     get:
       tags:
       - rolling_stock
@@ -1241,7 +1241,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/PaginatedResponseOfLightRollingStockWithLiveries'
-  /light_rolling_stock/name/{rolling_stock_name}/:
+  /light_rolling_stock/name/{rolling_stock_name}:
     get:
       tags:
       - rolling_stock
@@ -1259,7 +1259,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/LightRollingStockWithLiveries'
-  /light_rolling_stock/{rolling_stock_id}/:
+  /light_rolling_stock/{rolling_stock_id}:
     get:
       tags:
       - rolling_stock
@@ -1278,7 +1278,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/LightRollingStockWithLiveries'
-  /pathfinding/:
+  /pathfinding:
     post:
       tags:
       - pathfinding
@@ -1296,7 +1296,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/PathResponse'
-  /pathfinding/{pathfinding_id}/:
+  /pathfinding/{pathfinding_id}:
     get:
       tags:
       - pathfinding
@@ -1408,7 +1408,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ElectrificationsOnPathResponse'
-  /projects/:
+  /projects:
     get:
       tags:
       - projects
@@ -1469,7 +1469,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ProjectWithStudies'
-  /projects/{project_id}/:
+  /projects/{project_id}:
     get:
       tags:
       - projects
@@ -1548,7 +1548,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/InternalError'
-  /projects/{project_id}/studies/:
+  /projects/{project_id}/studies:
     get:
       tags:
       - studies
@@ -1623,7 +1623,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/StudyResponse'
-  /projects/{project_id}/studies/{study_id}/:
+  /projects/{project_id}/studies/{study_id}:
     get:
       tags:
       - studies
@@ -1720,7 +1720,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/InternalError'
-  /projects/{project_id}/studies/{study_id}/scenarios/:
+  /projects/{project_id}/studies/{study_id}/scenarios:
     get:
       tags:
       - scenarios
@@ -1799,7 +1799,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ScenarioResponse'
-  /projects/{project_id}/studies/{study_id}/scenarios/{scenario_id}/:
+  /projects/{project_id}/studies/{study_id}/scenarios/{scenario_id}:
     get:
       tags:
       - scenarios
@@ -1913,7 +1913,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/InternalError'
-  /rolling_stock/:
+  /rolling_stock:
     post:
       tags:
       - rolling_stock
@@ -1937,7 +1937,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/RollingStock'
-  /rolling_stock/name/{rolling_stock_name}/:
+  /rolling_stock/name/{rolling_stock_name}:
     get:
       tags:
       - rolling_stock
@@ -1955,7 +1955,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/RollingStockWithLiveries'
-  /rolling_stock/power_restrictions/:
+  /rolling_stock/power_restrictions:
     get:
       tags:
       - rolling_stock
@@ -1969,7 +1969,7 @@ paths:
                 type: array
                 items:
                   type: string
-  /rolling_stock/{rolling_stock_id}/:
+  /rolling_stock/{rolling_stock_id}:
     get:
       tags:
       - rolling_stock
@@ -2040,7 +2040,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/RollingStockWithLiveries'
-  /rolling_stock/{rolling_stock_id}/livery/:
+  /rolling_stock/{rolling_stock_id}/livery:
     post:
       tags:
       - rolling_stock
@@ -2068,7 +2068,7 @@ paths:
                 $ref: '#/components/schemas/RollingStockLivery'
         '404':
           description: The requested rolling stock was not found
-  /rolling_stock/{rolling_stock_id}/locked/:
+  /rolling_stock/{rolling_stock_id}/locked:
     patch:
       tags:
       - rolling_stock
@@ -2171,7 +2171,7 @@ paths:
                 type: array
                 items:
                   $ref: '#/components/schemas/SearchResultItem'
-  /single_simulation/:
+  /single_simulation:
     post:
       summary: Runs a simulation with a single train, does not write anything to the database
       requestBody:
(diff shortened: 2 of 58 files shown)


.github/workflows/build.yml
```diff
@@ -71,6 +71,9 @@ jobs:
         run: |
           set -eo pipefail
 
+          # disable cache mounts as github cache is slow
+          find -name Dockerfile -print0 | xargs -0 sed -Ei 's/--mount=type=cache,target=[^[:blank:]]+//g'
+
```
multun: I think we did that before, and removed it as it causes the pushed images to be from a different source than what's in the tree: it messes up the build cache if you just pull

editoast/openapi.yaml
```diff
@@ -199,7 +199,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/InternalError'
-  /health:
+  /health/:
```
multun: Is that intended? Many routes have the same change, which doesn't make sense to me

editoast/src/error.rs
```diff
@@ -149,19 +147,19 @@ impl EditoastError for RedisError {
...
+//     }
+
+//     fn get_type(&self) -> &str {
+//         "editoast:JsonError"
+//     }
+
+//     fn context(&self) -> HashMap<String, Value> {
+//         [("cause".into(), json!(self.to_string()))].into()
+//     }
+// }
```
multun:
```suggestion
```

editoast/openapi.yaml
Khoyo: I believe that's a side-effect of the macro changes.

It doesn't really matter as we trim any trailing slash in a request anyway.

We should probably do it to when generating the openapi.yml

.github/workflows/build.yml
Khoyo: This is just a cherry-pick for convenience, the right PR is here https://github.com/OpenRailAssociation/osrd/pull/7991 ;)

You mean, people pulling CI-built images will not reuse the cached layers when they build locally? If so, I think that's an okay tradeoff.

editoast/openapi.yaml
leovalais: Done.

shenriotpro (approved):
LGTM for tests

editoast/Cargo.toml
```diff
@@ -55,23 +55,21 @@ serde_json = "1.0.120"
...
-actix-files = "0.6.6"
-actix-http = "3.8.0"
-actix-multipart = "0.7.2"
-actix-web = "4.8.0"
-actix-web-opentelemetry = { version = "0.18.0", features = ["awc"] }
 async-trait = "0.1.81"
+axum = { version = "0.7.5", features = ["multipart"] }
+axum-extra = { version = "0.9.3", features = ["typed-header"] }
+axum-test = "15.3.0"
+axum-tracing-opentelemetry = "0.19.0"
```
woshilapin, line 72: Not sure about the usefulness of this dependency, already tried on another project, and was not that satisfied with the level of maintenance, especially since it's might be relatively simple to do the same ourselves without too much hassle. We can try to do that together.

editoast/Cargo.toml
```diff
@@ -134,6 +143,8 @@ validator = { version = "0.18.1", features = ["derive"] }
 
 [dev-dependencies]
 async-std = { version = "1.12.0", features = ["attributes", "tokio1"] }
+axum = { version = "0.7.5", features = ["macros", "multipart"] }
+axum-macros = "0.4.1"
```
woshilapin: I'm pretty sure we don't need this dependency since we already have `axum` with the `macros` feature. All of the macros of `axum-macros` are re-exposed in `axum` directly (except `TypedPath` but we don't seem to use it).

editoast/src/main.rs
```diff
@@ -344,98 +391,46 @@ async fn runserver(
...
-    });
-
-    let server = match args.workers {
-        Some(workers) => server.workers(workers),
-        None => server,
-    };
+    // Configure the axum router
+    let router: Router<()> = axum::Router::<AppState>::new()
+        .merge(views::router())
+        .layer(OtelInResponseLayer)
```
woshilapin, line 429: I believe this one is useless, or at least, only useful in some sort of debugging purpose. From what I understand, it will insert the OpenTelemetry context (trace ID mostly) into the headers of the response. The front doesn't use that. See [this comment](https://github.com/davidB/tracing-opentelemetry-instrumentation-sdk/blob/b64946e5c37fdd0557c43af1286533fac2ac6853/axum-tracing-opentelemetry/src/middleware/response_injector.rs#L53).

editoast/src/main.rs
```diff
@@ -344,98 +391,46 @@ async fn runserver(
...
+        .layer(OtelAxumLayer::default())
+        .layer(request_payload_limit)
+        .layer(cors)
+        .layer(TraceLayer::new_for_http())
```
woshilapin, line 433: I'm wondering if this is not doing mostly what `OtelAxumLayer::default()` is doing already. Can you try to see if they create some sort of duplicate message? If removing one or the other still creates useful logs?

editoast/src/views/test_app.rs
```diff
@@ -96,47 +80,61 @@ impl TestAppBuilder {
...
+            map_layers: MapLayers::parse().into(),
+            map_layers_config: MapLayersConfig::default().into(),
+        };
+
+        // Configure the axum router
+        let router: Router<()> = axum::Router::<AppState>::new()
+            .merge(crate::views::router())
+            .layer(OtelInResponseLayer)
+            .layer(OtelAxumLayer::default())
+            .layer(TraceLayer::new_for_http())
```
woshilapin, line 133: See comments in `main.rs` about those that are useless, and those that might be useless.

editoast/src/views/documents.rs
```diff
@@ -48,17 +48,16 @@ pub enum DocumentErrors {
...
-    let document_key = document_key.into_inner();
+async fn get(
+    State(db_pool): State<DbConnectionPoolV2>,
+    Path(document_id): Path<i64>,
+) -> Result<impl IntoResponse> {
     let conn = &mut db_pool.get().await?;
-    let doc = Document::retrieve_or_fail(conn, document_key, || DocumentErrors::NotFound {
-        document_key,
+    let doc = Document::retrieve_or_fail(conn, document_id, || DocumentErrors::NotFound {
+        document_key: document_id,
```
woshilapin, line 57: Any reason to change it to `document_id`? If yes, maybe change also the `DocumentErrors::NotFound` variant to reflect that change. But it feels this shouldn't be a part of this PR (another PR maybe).

editoast/src/views/rolling_stocks/mod.rs
```diff
@@ -405,33 +421,81 @@ async fn get_rolling_stock_usage(
...
 }
 
+#[derive(ToSchema)]
+#[allow(unused)] // Schema only
+struct RollingStockLiveryCreateForm {
+    name: String,
+    images: Vec<Vec<u8>>,
+}
+
+async fn parse_mulipart_content(
```
woshilapin:
```suggestion
async fn parse_multipart_content(
```

editoast/src/views/rolling_stocks/mod.rs
```diff
@@ -405,33 +421,81 @@ async fn get_rolling_stock_usage(
...
-    rolling_stock_id: Path<i64>,
-    MultipartForm(form): MultipartForm<RollingStockLiveryCreateForm>,
+    State(db_pool): State<DbConnectionPoolV2>,
+    Path(rolling_stock_id): Path<i64>,
+    form: Multipart,
 ) -> Result<Json<RollingStockLivery>> {
     let conn = &mut db_pool.get().await?;
 
-    let rolling_stock_id = rolling_stock_id.into_inner();
+    let (name, images) = parse_mulipart_content(form)
```
woshilapin: Did you evaluate [`axum_typed_multipart`](https://docs.rs/axum_typed_multipart)?

editoast/src/views/rolling_stocks/mod.rs
```diff
@@ -598,26 +654,24 @@ pub mod tests {
...
+            self.get(format!("/rolling_stock/{rolling_stock_id}").as_str())
+        }
 
-    fn rolling_stock_delete_request(rolling_stock_id: i64) -> Request {
-        TestRequest::delete()
-            .uri(format!("/rolling_stock/{rolling_stock_id}").as_str())
-            .to_request()
+        // fn rolling_stock_delete_request(&self, rolling_stock_id: i64) -> axum_test::TestRequest {
+        //     self.delete(format!("/rolling_stock/{rolling_stock_id}").as_str())
+        // }
```
woshilapin: Is there any problem with this test? If yes, we should probably try to address it.

editoast/src/views/rolling_stocks/mod.rs
```diff
@@ -1041,88 +1077,13 @@ pub mod tests {
...
-    async fn delete_unexisting_rolling_stock_returns_not_found_v1() {
-        let app = create_test_service().await;
-        let delete_request = rolling_stock_delete_request(0);
-        let delete_response = call_service(&app, delete_request).await;
-
-        assert_eq!(delete_response.status(), StatusCode::NOT_FOUND);
-    }
-
-    #[rstest] // TODO test for rolling stock with DBConnectionPoolV1 and with TrainScheduleV1
-    async fn delete_used_rolling_stock_should_fail() {
```
woshilapin, line 1059: I understand why the previous test function has been removed (TSv1). But I don't understand why this one is. Isn't it a valid test for TSv2? It might need to be converted to the new testing framework.

editoast/src/views/sprites.rs
```diff
@@ -53,16 +57,52 @@ async fn signaling_systems() -> Result<Json<Vec<String>>> {
...
-    let (signaling_system, file_name) = path.into_inner();
+async fn sprites(
+    Path((signaling_system, file_name)): Path<(String, String)>,
+) -> Result<impl IntoResponse> {
     let sprite_configs = SpriteConfig::load();
     if !sprite_configs.contains_key(&signaling_system) {
         return Err(SpriteErrors::UnknownSignalingSystem { signaling_system }.into());
     }
+
+    let re = Regex::new(r#"[-_ @0-9A-Za-z]+\.(json|png|svg)"#).expect("failed to compile regex");
```
woshilapin: I believe you should be able to build that `Regex` only once for the entirety of the program (see [documentation](https://docs.rs/regex/latest/regex/#avoid-re-compiling-regexes-especially-in-a-loop)). You can try something like this (not tested, but I hope there is enough pointers to understand the point 😄)

```rust
static RE = std::cell::OnceLock::new();
let re = RE.get_or_init(|| Regex::new(r#"[-_ @0-9A-Za-z]+\.(json|png|svg)"#).expect("failed to compile regex"));
```

editoast/src/views/sprites.rs
```diff
@@ -53,16 +57,52 @@ async fn signaling_systems() -> Result<Json<Vec<String>>> {
         (status = 404, description = "Signaling system not found"),
     ),
 )]
-#[get("/{signaling_system}/{file_name:[-_ @0-9A-Za-z]+\\.(json|png|svg)}")]
-async fn sprites(path: Path<(String, String)>) -> Result<NamedFile> {
-    let (signaling_system, file_name) = path.into_inner();
+async fn sprites(
```
woshilapin, line 59: Do we have test that this function behaves like expected? Since we manually implemented some of it (which was before handled by `actix`), I believe it would be the right time to add a test.

editoast/src/views/search.rs
```diff
@@ -353,13 +352,22 @@ pub async fn search(
...
     let objects = query.load::<SearchDBResult>(conn).await?;
     let results: Vec<_> = objects.into_iter().map(|r| r.result).collect();
-    Ok(HttpResponse::Ok().json(results))
+    Ok(axum::response::Response::builder()
+        .status(StatusCode::OK)
+        .header(CONTENT_TYPE, "application/json")
+        .body(axum::body::Body::from(
+            serde_json::to_string(&results).expect("serde error"),
+        ))
+        .expect("response builder error"))
```
woshilapin, line 372: Can't you do something like this instead?

```suggestion
    Ok((StatusCode::Ok, Json(serde_json::to_string(&results).expect("serde error")))

File: editoast/src/views/v2/train_schedule.rs [hunk 17]
```diff
@@ -846,91 +831,4 @@ mod tests {
             update_train_schedule_form.train_schedule.rolling_stock_name
         )
     }
-
-    #[rstest]
-    #[ignore] // TODO: This test should be rewritten using mocks
-    async fn train_schedule_simulation(
```
woshilapin, line 859: Why is this test removed? Was ignored in https://github.com/OpenRailAssociation/osrd/pull/7165 but feels like we should do something about it (something else than removing it).

Same for the test below.

editoast/src/views/infra/railjson.rs
```diff
@@ -117,14 +115,22 @@ async fn get_railjson(
...
-        .content_type(ContentType::json())
-        .append_header(("x-infra-version", infra_meta.version))
-        .body(railjson))
+    Ok((
+        StatusCode::OK,
+        [
+            (
+                header::CONTENT_TYPE.as_str(),
+                headers::ContentType::json().to_string(),
+            ),
```
woshilapin, line 124: I believe you don't need this. If you wrap the variable `railjson` into a `Json(railjson)`, `axum` will automatically add the `Content-Type` header in the response.

woshilapin (commented):
Huge work, hope switching to `axum` will help us get the expected "return on investment". I've made a bunch of comments, but also some more general comments below.

# `route!()`

Do we really need this `route!()` macro? From what I understand, it allows us to not declare twice the routing: one for `utoipa` and one for `axum`. The maintenance of this macro, compared to what it brings, feels a bit high. Also, I'm wondering if we can automate this entirely, by setting up the `axum` router based on the OpenApi specifications through whatever `utoipa` implements (which interestingly, might also produce a piece of code hard to maintain 😝). 

# `panic!()`, `.expect()` and friends?

What is the effect of `panic` on a request with `axum`. So far, we've adopted an ok-ish attitude about `panic` because `actix` is basically handling `panic` in the actix worker (possibly restarting the worker when crashed. But I'm worried that `axum` might not react the same.

# Homogeneity
## Naming
I noticed sometimes `redis_client` is used, and sometimes `redis`, it might be nice to take the opportunity to do the same everywhere.

## destructuring or not destructuring, that is the question
In the handlers, to access the `AppState`,  sometimes destructuring is used, and sometimes not. It might be good to do the same everywhere.

editoast/src/views/infra/railjson.rs
Khoyo: Wouldn't it also serialize it? `railjson` is a String containing JSON, we want to pass it verbatim.

leovalais:
Thanks a lot for reviewing this long (and off-putting) PR 😄

# About `routes!`

I strongly agree. I believe generating the router from a tree of annotations using `nest()` might be my preference as it avoids duplication and enforce the necessity of the annotation.

# About `panic!`s

When an endpoint `panic!`s, the server doesn't stop and closes the connection. That results in a `502 Bad gateway` on the front-end side. `tower` provides a [middleware](https://docs.rs/tower-http/latest/tower_http/catch_panic/index.html) that converts `panic!`s into HTTP responses, but I'm not convinced that's a design we'd want to adopt.

# About destructuring

I'm strongly in favor of destructuring early for many reasons (we can discuss about that if needed). Wdyt about a follow up PR that fixes the inconsistency to avoid blocking this one?

# About improvements

I suggest opening issues classified as "enhancements" for the following list of improvements:

* removing `routes!`
* naming and homogeneity pass
* [implementing our own otel middleware](https://github.com/OpenRailAssociation/osrd/pull/8047#discussion_r1684411285)
* [testing signal sprite import](https://github.com/OpenRailAssociation/osrd/pull/8047#discussion_r1687996132)
* [evaluating `axum_typed_multipart` and test liveries import](https://github.com/OpenRailAssociation/osrd/pull/8047#discussion_r1687963958)
* gate `dry` parameter of `/search` on `#[cfg(test)]` and homogenize the return type of the endpoint
* rename `document_key` to `document_id` in `/document` APIs

All these are very pertinent but this PR is massive and conflicts a lot. I believe it's probably easier to do (and review) that incrementally once `axum` will be set up for everyone. Wdyt?

editoast/src/views/documents.rs
leovalais: We usually have the suffix `_id` for variables holding a row PK. We didn't change the error variant as this would impact the public api of editoast. (We made sure that we wouldn't break the API bu watching changes made to `openapi.yaml`. Changing that would introduce "noise" in the diff.)

I agree that the error needs to change though. I'd favor a follow-up PR.

editoast/src/views/rolling_stocks/mod.rs
leovalais: I don't believe we did 😬

However, since we want to try to reduce the number of dependencies editoast has, I'm not sure if it's worth adding that for a single use case. (If one day we write another endpoint for a multipart request, I'm all for using the lib though.)

Wdyt?

editoast/src/views/rolling_stocks/mod.rs
leovalais: That was related to the tests we then deleted. It's dead code now, thanks.

editoast/src/views/rolling_stocks/mod.rs
leovalais: These tests and the associated endpoint will be removed with TSV1. That feature was forgotten in the TSV2 implementation plan. There's an [issue](https://github.com/OpenRailAssociation/osrd/issues/8078) open to make its TSV2 version.

editoast/src/views/v2/train_schedule.rs
leovalais: There's an issue open on that : https://github.com/OpenRailAssociation/osrd/issues/8121. I would say that it's fine to delete them now and fetch them in the git history when we'll work on them again. Wdyt?

editoast/src/views/search.rs
leovalais: Sadly we can't because it's not the same type as the other `return` of the function : https://github.com/OpenRailAssociation/osrd/pull/8047/files/035803e3552389598686696a86e2892775cbcc4a..488e4c81dc2ff3140643a0e863293890b1ef0a8d#diff-b6a2310b76e9bd3bab165e74f48fbf50e3619c5d178c7c2bcfef596e2845f434R355-R358

That other branch used to return a `content-type: text`. But it sure we can return a JSON string instead.

editoast/src/views/rolling_stocks/mod.rs
```diff
@@ -598,26 +654,24 @@ pub mod tests {
...
-    use crate::views::tests::create_test_service;
 
-    fn rolling_stock_create_request(rolling_stock_form: &RollingStockForm) -> Request {
-        TestRequest::post()
-            .uri("/rolling_stock")
-            .set_json(rolling_stock_form)
-            .to_request()
-    }
+    impl TestApp {
+        fn rolling_stock_create_request(
```
hamz2a, line 661: Is there a reason to add this `impl TestApp`?
Why not just use this: 
1. `let request = app.post("/rolling_stock").json(rolling_stock_form);`
2.  `let request = app.get(format!("/rolling_stock/{rolling_stock_id}").as_str());`

editoast/src/views/projects.rs
```diff
@@ -301,19 +298,16 @@ impl From<ProjectPatchForm> for Changeset<Project> {
...
     )
 )]
-#[patch("")]
 async fn patch(
-    data: Json<ProjectPatchForm>,
-    project_id: Path<i64>,
-    db_pool: Data<DbConnectionPoolV2>,
+    Path(project_id): Path<i64>,
+    State(db_pool): State<DbConnectionPoolV2>,
+    Json(form): Json<ProjectPatchForm>,
```
hamz2a, line 304: It is better to have explicit name like `project_form`

editoast/src/views/study.rs
```diff
@@ -132,21 +128,19 @@ impl StudyCreateForm {
...
     )
 )]
-#[post("")]
 async fn create(
-    db_pool: Data<DbConnectionPoolV2>,
-    data: Json<StudyCreateForm>,
-    project: Path<i64>,
+    State(db_pool): State<DbConnectionPoolV2>,
+    Path(project_id): Path<i64>,
+    Json(data): Json<StudyCreateForm>,
```
hamz2a, line 142: It is better to have explicit name like `Json(study_form)`

editoast/src/views/study.rs
```diff
@@ -303,13 +294,11 @@ impl StudyPatchForm {
...
     )
 )]
-#[patch("")]
 async fn patch(
-    data: Json<StudyPatchForm>,
-    path: Path<(i64, i64)>,
-    db_pool: Data<DbConnectionPoolV2>,
+    Path((project_id, study_id)): Path<(i64, i64)>,
+    State(db_pool): State<DbConnectionPoolV2>,
+    Json(data): Json<StudyPatchForm>,
```
hamz2a, line 300: `Json(study_form)`

hamz2a (commented):
First of all, thank you for the PR; it's impressive. I have added some comments.

In several endpoints, we have forms in this format:
```rust
Json(data): Json<StudyCreateForm>
```
It is better to use explicit names like:
```rust
Json(study_form): Json<StudyCreateForm>
```

editoast/src/views/infra/mod.rs
```diff
@@ -651,10 +657,10 @@ pub mod tests {
     use editoast_schemas::infra::RAILJSON_VERSION;
     use editoast_schemas::primitives::ObjectType;
 
-    fn delete_infra_request(infra_id: i64) -> Request {
-        TestRequest::delete()
-            .uri(format!("/infra/{infra_id}").as_str())
-            .to_request()
+    impl TestApp {
```
hamz2a, line 660: Is there a reason to add this impl TestApp?

editoast/src/views/rolling_stocks/mod.rs
```diff
@@ -598,26 +654,24 @@ pub mod tests {
...
     use crate::views::test_app::TestAppBuilder;
-    use crate::views::tests::create_test_service;
 
-    fn rolling_stock_create_request(rolling_stock_form: &RollingStockForm) -> Request {
-        TestRequest::post()
-            .uri("/rolling_stock")
-            .set_json(rolling_stock_form)
-            .to_request()
-    }
+    impl TestApp {
```
hamz2a, line 662: Is there a reason to add this impl TestApp?
Why not just use this:
```rust
let request = app.post("/rolling_stock").json(rolling_stock_form);
let request = app.get(format!("/rolling_stock/{rolling_stock_id}").as_str());
```

hamz2a (commented):
First of all, thank you for the PR; it's impressive. I have added some comments.

In several endpoints, we have forms in this format:
```rust
Json(data): Json<StudyCreateForm>
```
It is better to use explicit names like:
```rust
Json(study_form): Json<StudyCreateForm>
```

editoast/src/views/rolling_stocks/mod.rs
Khoyo: We're trying to keep the tests structure as close as possible to the actix ones in order to lower the risk of introducing bugs.

editoast/src/views/infra/railjson.rs
woshilapin: I don't see how setting up or not setting up a header will change how the body is handled. If the current code works for you, it should be fine without the header too I guess 🤷

editoast/src/views/v2/train_schedule.rs
woshilapin: OK, if we track that somewhere else, fine by me!

editoast/src/views/search.rs
woshilapin: I'd say it's weird that the other return doesn't return the same type so I'd say maybe we need to address that in the first place. Returning the text as a JSON is probably a good idea?

editoast/src/views/documents.rs
woshilapin: It seems that we have a bunch of plans for follow-up PR, it'd be nice to make a list in the current PR of all the follow-up PR we expect, what do you think?

editoast/src/views/rolling_stocks/mod.rs
woshilapin: Yes, I almost didn't make the comment for that reason... ok fine, I'm resolving this comment.

editoast/src/views/infra/railjson.rs
leovalais: The `Json` extractor needs a `T: Serialize` but since `railjson` is already a string, we'd get a single JSON string. We do that to avoid the cost of serialization for big infras.

Khoyo: Wrapping a data structure into a `Json<>` tend to pass them to Serde for json serialization...

editoast/src/views/documents.rs
leovalais: I opened a tracking issue here: https://github.com/OpenRailAssociation/osrd/issues/8178

editoast/src/views/search.rs
leovalais: I've added to the list of #8178. This way we can make it a good first issue (we need these 😬).

editoast/src/main.rs
Khoyo: Tracked here: #8178

editoast/src/main.rs
Khoyo: Tracked here: #8178

flomonster (approved):
Tested, all open conversations are tracked [here](https://github.com/OpenRailAssociation/osrd/issues/8178)

Math-R (approved):

clarani (commented):
It would be nice to remove the `/` from `osrdEditoastApi.ts` too

PR #7103, opened 2024-04-04
WIP: OSRD goes async

ghost:
This PR holds the whole modifications for making OSRD async using a RabbitMQ. This is a work in progress, here is an overview:

- [X] Modify local environement to provide rabbitmq
- [X] Implement Osrdyne
  - [X] Implement Kubernetes driver
  - [X] Implement Docker driver
  - [X] Implement RabbitMQ driver
  - [X] Implement logic
- [X] Test the whole stack
  - [X] Against Docker
  - [X] Against Kubernetes
- [X] Modify Core to use queues
  - [X] Create a client library for enqueing/dequeing messages
  - [X] Use the said library
- [X] Modify Editoast to use queues
  - [X] Create a client library for enqueing/dequeing messages
  - [X] Use the said library

Closes #6679, #6680, #6681

Helm cart related PR : https://github.com/OpenRailAssociation/osrd-chart/pull/11

PR #8048, opened 2024-07-14
Scalable async RPC

### LINKED ISSUE #6117: [meta] Scalable async RPC between editoast and core
### Description and goal

To address the scalability and memory constraint issues faced by the stateful core workers, which require loading comprehensive infrastructure data to process requests, a new architecture using async RPC is proposed

In this refined approach, a custom-made core-controller component dynamically orchestrates core workers, ensuring that requests are directed a specialized Core that only works on this infrastructure. 

This method significantly enhances the system's ability to handle multiple infrastructure concurrently by leveraging a distributed processing model, where each core worker is specialized.

Those modifications aim to eliminate the bottleneck of single-worker memory limitations.

### Tasks

- https://github.com/osrd-project/osrd/issues/6679
- https://github.com/osrd-project/osrd/issues/6680
- https://github.com/osrd-project/osrd/issues/6681

ghost:
Fixes https://github.com/OpenRailAssociation/osrd/issues/6117.

_(This PR replaces the closed #7103 PR. Check the log of the previous PR for the reason)._

This PR implements the [Scalable Async RPC](https://osrd.fr/en/docs/reference/design-docs/scalable-async-rpc/) design.

## Changes

- [X] Modify local environment to provide RabbitMQ
- [X] Implement OSRDyne
  - [X] Implement Kubernetes driver
  - [X] Implement Docker driver
  - [X] Implement RabbitMQ driver
  - [X] Implement logic
- [X] Modify Core to use RabbitMQ queues
- [X] Modify Editoast to use RabbitMQ queues

## Tests

- [X] Against Docker
- [X] Against Kubernetes (we deployed a custom Kubernetes cluster to test it)


ghost:
This PR is still in draft because we have some rough edges and various fixes. I've opened the new early in order to facilitate early reviews.

osrdyne/Cargo.toml
woshilapin, line 1: Should we add telemetry on `osrdyne`? Can be done in another PR.

core/src/main/java/fr/sncf/osrd/cli/WorkerCommand.kt
```diff
@@ -0,0 +1,236 @@
...
+    private var editoastUrl: String? = null
+
+    @Parameter(
+        names = ["--editoast-authorization"],
+        description = "The HTTP Authorization header sent to editoast"
+    )
+    private var editoastAuthorization: String? = null
+
+    @Parameter(names = ["-j", "--threads"], description = "The number of threads to serve requests from")
+    private var threads: Int = 1
```
woshilapin: Don't we want to handle everything with environment variables (like what you have below with `WORKER_ID` & co.)? Won't this be easier for k8s configuration?

docker-compose.yml
```diff
@@ -35,29 +36,33 @@ services:
...
+    healthcheck:
+      test: ["CMD", "rabbitmqctl", "status"]
+      interval: 30s
+      timeout: 30s
+      retries: 3
+
   core:
     image: ghcr.io/openrailassociation/osrd-edge/osrd-core:${TAG-dev}
-    container_name: osrd-core
+    container_name: osrd-core-dummy
```
woshilapin, line 59: Why do we want to keep that one up?

docker/docker-bake.hcl
```diff
@@ -152,3 +154,23 @@ target "gateway-front" {
...
   }
 }
+
+###########
+# OSRDyne #
+###########
+
+target "base-osrdyne" {}
+target "osrdyne" {
+  inherits = ["base", "base-osrdyne"]
```
woshilapin, line 167: I'm not very familiar with `bake`, but what is the purpose of having this empty `base-osrdyne` which is used in `osrdyne`? Looks useless from a newbie eye 👀

osrdyne/src/config.rs
```diff
@@ -0,0 +1,56 @@
...
+            max_length: None,
+            max_length_bytes: None,
+            api_address: "0.0.0.0:4242".into(), // TODO: decide on the port
+        }
+    }
+}
+
+pub fn parse_config() -> Result<OsrdyneConfig, figment::Error> {
+    Figment::from(Serialized::defaults(OsrdyneConfig::default()))
+        .merge(Yaml::file("osrdyne.yml"))
```
woshilapin, line 54: Any reason to use YAML instead of another format like JSON or TOML? The reason I ask the question is that `figment::providers::Yaml` is backed up by `serde_yaml`, which is discontinued (https://github.com/dtolnay/serde-yaml). I don't see a big risk... but still, might be better to bet on a format that is less likely to lack a bit of support on the short/middle term. TOML in `figment` is backed up by `toml_edit` (used in `cargo`), maintained by Ed Page, one of the main maintainer of `cargo`.

osrdyne/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+FROM debian:bookworm-slim as running_env
+RUN apt update -yqq && \
+    apt install -yqq --no-install-recommends curl ca-certificates libjemalloc2 jq && \
+    apt-get purge -y --auto-remove -o APT::AutoRemove::RecommendsImportant=false && \
+    rm -rf /var/lib/apt/lists/*
+
+COPY --from=run_builder /usr/local/cargo/bin/osrdyne /usr/local/bin/osrdyne
+
+ARG OSRD_GIT_DESCRIBE
+ENV OSRD_GIT_DESCRIBE=${OSRD_GIT_DESCRIBE}
```
flomonster, line 56: This is not used. We could add a `/version` route to display the version of `osrd-dyne` in the frontend.

osrdyne/src/drivers/docker.rs
```diff
@@ -0,0 +1,196 @@
...
+            amqp_uri,
+        }
+    }
+}
+
+impl WorkerDriver for DockerDriver {
+    fn get_or_create_core_pool(
+        &self,
+        infra_id: Key,
+    ) -> Pin<Box<dyn Future<Output = Result<Uuid, DriverError>> + Send + '_>> {
```
flomonster: I got the impression from reading the doc that there was a desire to make it core agnostic.

```suggestion
    fn get_or_create_worker_pool(
        &self,
        worker_key: Key,
    ) -> Pin<Box<dyn Future<Output = Result<Uuid, DriverError>> + Send + '_>> {
```

osrdyne/src/drivers/docker.rs
```diff
@@ -0,0 +1,196 @@
...
+                env
+            };
+
+            let labels = {
+                let mut labels = std::collections::HashMap::new();
+                labels.insert(LABEL_MANAGED_BY.to_owned(), MANAGED_BY_VALUE.to_owned());
+                labels.insert(LABEL_CORE_ID.to_owned(), new_id.to_string());
+                labels.insert(LABEL_INFRA_ID.to_owned(), infra_id.to_string());
+                labels
+            };
```
flomonster:
```suggestion
            let labels = HashMap::from([
                (LABEL_MANAGED_BY.to_owned(), MANAGED_BY_VALUE.to_owned()),
                (LABEL_CORE_ID.to_owned(), new_id.to_string()),
                (LABEL_INFRA_ID.to_owned(), infra_id.to_string()),
            ]);
```

flomonster (commented):
I haven't finished my first review. But here are a few comments.

flomonster:
We don't have `check_osrdyne_tests` where we run tests, fmt and clippy.
There are currently several clippy warnings.

osrdyne/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+name = "osrdyne"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+im = "15.1.0"
+smol_str = "0.2.2"
+smallvec = { version = "1.13.2", features = ["serde"] }
```
woshilapin: It seems that `smol_str` is not used anywhere. But I'm not sure that I understand why `smallvec` is used instead of `smol_str` for the internal of `Key`? Was there a technical reason? (just feels more natural to handle a string with `smol_str` than with `smallvec`).

In any case, probably one of these 2 dependencies can be removed.

osrdyne/src/drivers/docker.rs
```diff
@@ -0,0 +1,196 @@
...
+        &self,
+        infra_id: Key,
+    ) -> Pin<Box<dyn Future<Output = Result<Uuid, DriverError>> + Send + '_>> {
+        Box::pin(async move {
+            let current_cores = self.list_core_pools().await?;
+            for core in current_cores {
+                if core.infra_id == infra_id {
+                    return Ok(core.core_id);
+                }
+            }
```
woshilapin: Does this means we can only have 1 instance of a `core` per infra? I thought the idea was also to be able to spawn multiple instance of `core` for one infrastructure.

osrdyne/src/drivers/kubernetes.rs
```diff
@@ -0,0 +1,326 @@
...
+    fn get_or_create_core_pool(
+        &self,
+        infra_id: Key,
+    ) -> Pin<Box<dyn Future<Output = Result<Uuid, DriverError>> + Send + '_>> {
+        Box::pin(async move {
+            let current_cores = self.list_core_pools().await?;
+            for core in current_cores {
+                if core.infra_id == infra_id {
+                    return Ok(core.core_id);
+                }
```
woshilapin: This might be implemented directly in the trait as a default method, since all of the implementation have the same behavior.

```rust
trait WorkerDriver {
    // Required
    fn create_core_pool(
        &self,
        infra_id: Key,
    ) -> Pin<Box<dyn Future<Output = Result<Uuid, DriverError>> + Send + '_>>;
    
    // Provided
    fn get_or_create_core_pool(
        &self,
        infra_id: Key,
    ) -> Pin<Box<dyn Future<Output = Result<Uuid, DriverError>> + Send + '_>> {
        Box::pin(async move {
            let current_cores = self.list_core_pools().await?;
            for core in current_cores {
                if core.infra_id == infra_id {
                    return Ok(core.core_id);
                }
           }
           self.create_pool(infra_id)
        })
    }
    // Other functions
}
```

Note that I didn't tested it and took some shortcuts about `Box::pin` & co. It might not be that straightforward to do...

osrdyne/src/main.rs
```diff
@@ -0,0 +1,166 @@
...
+    let target_tracker_config = TargetTrackerConfig::default();
+    let mut target_tracker = spawn(target_tracker_actor(
+        tracker_inbox,
+        init_keys,
+        target_tracker_config,
+    ));
+
+    // target_tracker_client
+    //     .require_worker_group(Key::new("test"), Duration::ZERO)
+    //     .await?;
```
woshilapin: Can be removed.

osrdyne/src/main.rs
```diff
@@ -0,0 +1,166 @@
...
+                driver,
+                config.worker_loop_interval,
+                worker_list_send.clone(),
+            )
+            .await?;
+
+        // wait for one of the jobs to complete
+        select! {
+            _ = signal::ctrl_c() => {
+                info!("received interupt, shutting down");
```
woshilapin:
```suggestion
                info!("received interrupt, shutting down");
```

osrdyne/src/drivers/mod.rs
woshilapin, line 1: I believe we're trying (at least in `editoast`) to move away of `mod.rs` and adopt more the abilities from Rust edition 2018 module system with `drivers.rs` with its submodules in `drivers/` (see https://github.com/OpenRailAssociation/osrd/pull/7939#discussion_r1666955223 and https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html).

woshilapin (commented):
In general, great work. Here is a first iteration of general remarks.

It seems that even if we have an abstract worker driver API, each of the implementations provide a different set of features (I can only have on core worker per infra for Docker, but autoscaling allow for more than one in k8s, I can setup the machine capacity for k8s, but not setup any limit for Docker). It's like implementation details leaks out of this API. Maybe number of workers per infra and capacities of these nodes should be part of the API?

ghost:
> It seems that even if we have an abstract worker driver API, each of the implementations provide a different set of features (I can only have on core worker per infra for Docker, but autoscaling allow for more than one in k8s, I can setup the machine capacity for k8s, but not setup any limit for Docker). It's like implementation details leaks out of this API. Maybe number of workers per infra and capacities of these nodes should be part of the API?

I disagree on this : we've considered integrating the scaling of cores in osrdyne but rejected it. It was clearly not easy to determine when go up or down, and it requires taking account metrics that are not easy to handle properly.

From my perspective, it makes way more sense that for a Kubernetes deployment the scaling happens at the orchestrator level and not the applicative level. We will probably, in the near future, add an option to leverage Keda instead of the HPA.

The Kubernetes Driver exposes features of the orchestrator, and for a Docker deployment we can only work with a single host so horizontally scaling the core makes little sense. We could add a driver for docker swarm clusters but they are marginal (and not what we use).

To finish, I'll say that I do not feel like the API is leaking, it just allows to pass options to the driver.

docker-compose.yml
ghost: It's not up per se, the command exits right away (true) but it trigger a build of the image locally to be used by osrdyne after

osrdyne/src/config.rs
ghost: Yes, the capability to pass structure from the helm chart through from the Values into the k8s.

Having them in toml would require to write a toml with inputs that will be just forwarded to k8s. Checkout the [PR on osrd-chart](https://github.com/OpenRailAssociation/osrd-chart/pull/11), the config map of osrdyne, you'll see the passthrough that ends up in the k8s driver.

osrdyne/src/drivers/docker.rs
ghost: On the docker driver, yes. It makes little sense, since it is running on a single host. Check my global response for more thoughts on this ^^

docker/docker-bake.hcl
ghost: Its primary use was for debugging iirc, not sure exactly how we ended up with this to be honest, but I'm almost sure it helps when debugging the bakefile locally. I will check this specifically later.

osrdyne/src/config.rs
woshilapin: Hmm, didn't see that coming. Indeed, this is a good reason👍

docker-compose.yml
woshilapin: I see, one of the perks to use `docker compose` as our development tool. Maybe just add your response as a comment in `docker-compose.yml` so the next one finding this understands why it's here.

ghost: Good point, will do.

woshilapin:
> From my perspective, it makes way more sense that for a Kubernetes deployment the scaling happens at the orchestrator level and not the applicative level.

This sentence did make me understand your point of view. And indeed, I think I do agree. It also made me realize that `osrdyne` doesn't even handle workers, only workers' ~pools~ groups. So let's make that even more obvious by renaming `WorkerMetadata` to ~`WorkerPoolMetadata`~ `WorkerGroupMetadata` and maybe the field `core_id` should be ~`worker_pool_id`~ `worker_group_id` (if I understand correctly, kubernetes returns the deployment, so one object per worker ~pool~ group)?

osrdyne/Cargo.toml
Khoyo: Keys aren't necessarily strings, they are byte sequences (that could happen to be text) as far as the rpc protocol is concerned.

Khoyo:
To be clear, we should use worker_group, not worker_pool (https://osrd.fr/en/docs/reference/design-docs/scalable-async-rpc/#conceps)

Currently osrdyne manages one worker pool, `core`, and starts worker groups (eg. `core-1`). The docker driver does that by directly starting a single worker, but the k8s driver can create autoscaled deployments.

(yes, fields referring to core are inherited from a previous design - core_controller - and should disappear)

ghost:
> > From my perspective, it makes way more sense that for a Kubernetes deployment the scaling happens at the orchestrator level and not the applicative level.
> 
> This sentence did make me understand your point of view. And indeed, I think I do agree. It also made me realize that `osrdyne` doesn't even handle workers, only workers' pools. So let's make that even more obvious by renaming `WorkerMetadata` to `WorkerPoolMetadata` and maybe the field `core_id` should be `worker_pool_id` (if I understand correctly, kubernetes returns the deployment, so one object per worker pool)?

Yes, that's a really good point that @flomonster also raised during its review so I think we're going to modify this. I'll commit something about this later today.

osrdyne/Cargo.toml
Khoyo: Moved here #8081

core/src/main/java/fr/sncf/osrd/cli/WorkerCommand.kt
Khoyo: Those parameters do come from env variables ;)

editoast/src/core/mod.rs
```diff
@@ -156,18 +172,41 @@ impl CoreClient {
...
                     Ok(Some(response)) => Ok(response),
                     Ok(None) => Err(CoreError::NoResponseContent.into()),
-                    Err(MockingError { bytes, status, url }) => Err(self.handle_error(
-                        &bytes,
-                        reqwest::StatusCode::from_u16(status.as_u16()).unwrap(),
-                        url,
-                    )),
+                    Err(MockingError { bytes, status, url }) => Err({
+                        let mut err = self.handle_error(
+                            &bytes, //reqwest::StatusCode::from_u16(status.as_u16()).unwrap(),
```
leovalais: dead code

editoast/src/core/mq_client.rs
```diff
@@ -0,0 +1,237 @@
...
+    /// Default timeout for the response
+    pub timeout: u64,
+}
+
+#[derive(Debug, Error, EditoastError)]
+#[editoast_error(base_id = "coreclient")]
+pub enum Error {
+    #[error("AMQP error: {0}")]
+    #[editoast_error(status = "500")]
+    Lapin(lapin::Error),
```
leovalais:
```suggestion
    Lapin(#[from] lapin::Error),
```

editoast/src/core/mq_client.rs
```diff
@@ -0,0 +1,237 @@
...
+
+        Ok(RabbitMQClient {
+            connection: Arc::new(connection),
+            exchange: format!("{}-req-xchg", options.worker_pool_identifier),
+            timeout: options.timeout,
+            hostname,
+        })
+    }
+
+    #[allow(dead_code)]
```
leovalais, line 72: Please comment why we want to keep this dead code.

editoast/src/core/mq_client.rs
```diff
@@ -0,0 +1,237 @@
...
+        Ok(())
+    }
+
+    pub async fn call_with_response<T>(
+        &self,
+        routing_key: String,
+        path: &str,
+        published_payload: &Option<T>,
+        mandatory: bool,
+        correlation_id: Option<String>,
```
leovalais: is this still necessary (discussed orally)?

osrdyne/src/main.rs
```diff
@@ -0,0 +1,168 @@
...
+                break 'reconnect_loop;
+            }
+            _ = tasks.join_next() => {
+                error!("one of the tasks stopped, reconnecting");
+            },
+        }
+
+        // stop workers. shutdown is not graceful, as the system is expected
+        // to be capable to handle interuptions and network partitions at any time.
+        // tasks.shutdown().await;
```
leovalais: why? probably try to shutdown gracefully with a small timeout

leovalais (commented):
LGTM for the most part. Haven't tested yet (might be useful to check the dev setup on macOS where no host networking is available).

Thanks for guiding me through the review and answering my questions :)

front/public/locales/fr/errors.json
```diff
@@ -33,6 +33,10 @@
       "ConnectionResetByPeer": "Core: réinitialisation de la connexion. Nouvelle tentative.",
       "CoreResponseFormatError": "Core: impossible d'analyser la réponse '{{msg}}'",
       "GenericCoreError": "Core: erreur {{raw_error}}",
+      "Lapin": "Core: file de message: erreur de protocole",
```
SharglutDev: I think I don't have the `Lapin` ref :)
It feels weird to read `file de message`, could we add `file d'attente` or we lose some signification with that ?
```suggestion
      "Lapin": "Core: file d'attente de message: erreur de protocole",
```

front/public/locales/fr/errors.json
```diff
@@ -33,6 +33,10 @@
...
+      "ResponseTimeout": "Core: temps d'attente écoulé",
+      "Serialization": "Core: impossible de serialiser la requête",
```
```suggestion
      "Serialization": "Core: impossible de sérialiser la requête",
```

front/public/locales/fr/errors.json
Khoyo: That's the name of the rust library we use to interface with our message queue (RabbitMQ) ;)

front/public/locales/fr/errors.json
```diff
@@ -33,6 +33,10 @@
       "ConnectionResetByPeer": "Core: réinitialisation de la connexion. Nouvelle tentative.",
       "CoreResponseFormatError": "Core: impossible d'analyser la réponse '{{msg}}'",
       "GenericCoreError": "Core: erreur {{raw_error}}",
+      "Lapin": "Core: file d'attente de message: erreur de protocole",
```
clarani, line 41: Why lapin haha ? :rabbit:

clarani (approved):

front/public/locales/fr/errors.json
SharglutDev: Check my previous comment :D

SharglutDev (approved):
Lgtm (not tested)

front/public/locales/fr/errors.json
flomonster: It's the rust library name to handle RabbitMQ protocole. :fr: :rabbit:  https://github.com/amqp-rs/lapin

editoast/src/core/mq_client.rs
```diff
@@ -0,0 +1,232 @@
...
+}
+
+pub struct MQResponse {
+    pub payload: Vec<u8>,
+    pub status: Vec<u8>,
+}
+
+impl RabbitMQClient {
+    pub async fn new(options: Options) -> Result<Self, Error> {
+        let connection = Connection::connect(&options.uri, ConnectionProperties::default())
```
woshilapin, line 57: Naive question, don't we want a pool of connection instead of a unique connection?

Khoyo: The preferred mechanism is using channels inside an unique connection (see https://www.rabbitmq.com/tutorials/amqp-concepts#amqp-channels). This gives use multiplexing at a protocol level.

However, it uses a single TCP connection, so it suffers from the usual problems (head of line blocking, etc). It don't think those will come up at our planned deployment scale.

docker/docker-compose.core-test.yml
```diff
@@ -0,0 +1,17 @@
+version: '3'
```
Erashin: This should be removed yes (the entire file)?

Erashin (approved):
LGTM. ApiServerCommand to be deleted at a later stage (+ hopefully at some point we'll drop the v1 endpoints).

osrdyne/Dockerfile
```diff
@@ -0,0 +1,60 @@
+# syntax=docker/dockerfile:1
+
+##############
+# Cargo chef #
+##############
+FROM lukemathwalker/cargo-chef:0.1.67-rust-latest AS chef
```
woshilapin: `grcov` doesn't compile on latest version on rust >1.80.0 (see https://github.com/mozilla/grcov/pull/1191). We pinned the version of Rust in `editoast` and `gateway` in https://github.com/OpenRailAssociation/osrd/pull/8214.

```suggestion
FROM lukemathwalker/cargo-chef:0.1.67-rust-1.79.0 AS chef
```

docker/docker-compose.core-test.yml
ghost: Yes, it should be removed. Not needed.

editoast/src/core/mq_client.rs
Khoyo: Mostly for completeness sake. We can remove it too, wdyt?

Khoyo (approved):
image

bloussou (approved):
Great work !! Tested on the dev environment it works 🎉  !!!

editoast/src/core/mq_client.rs
```diff
@@ -0,0 +1,232 @@
...
+
+        Ok(RabbitMQClient {
+            connection: Arc::new(connection),
+            exchange: format!("{}-req-xchg", options.worker_pool_identifier),
+            timeout: options.timeout,
+            hostname,
+        })
+    }
+
+    #[allow(dead_code)]
```
flomonster, line 72:
```suggestion
```

flomonster (approved):
LGTM

editoast/src/core/mq_client.rs
ghost: Yes, we've decided to keep it for completeness.

PR #8053, opened 2024-07-15
core: pathfinding: look for fastest path instead of shortest

### LINKED ISSUE #7058: core: pathfinding: find the fastest path (instead of the shortest)

EVERYBODY

When running the pathfinding, we return the *shortest* path in terms of distance.

There are situations (possibly in *all* situations) where we would want to use the fastest path instead. 

The returned paths would generally make more sense because tracks are slower in their opposite direction, this would be closer to the paths that are actually used. It would also make the high-speed lines prevail (_e.g._ Paris-Rennes)

When importing timetables, lots of trains can't go fast enough to match their scheduled times, I suspect this is because we use slower (but shorter) paths. This change would fix some error cases.

The implementation itself should be mostly straightforward (tweak both the cost function and heuristic).

(note: this would just account for the MRSP and ignore the speedup / slowdown phases)

The heuristic would be a little less efficient.

**For proper route-specific speed limits**, requires https://github.com/OpenRailAssociation/osrd/issues/6784 to be implemented first

---

**There are things to refine and decide beforehand**. Mostly if and how we want to change the API. Do we just change the global pathfinding behavior, or would we add an extra parameter? (Changing the behavior globally would be *much* easier, but we need to make sure it doesn't have any unexpected side-effects)

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/7058

Change summary:

1. Move code from `STDCMHeuristic` into a new class (anything related to estimating min MRSP time)
2. Add a pathfinding parameter to define a cost function
3. Set the cost function to use MRSP times
4. Remove heuristic implementation and tests (should be fixed before reintroduced)

Note: there's approximations made here. We don't actually run simulation and only use MRSPs. We also don't know the rolling stop speed or length as we'd need to add parameters to the request. Even if we add them, speed limits wouldn't carry over blocks as they should (to account for the RS length).

We're generally after approximate values there so I don't think it would cause any issue. 

I tested on a few trains on imported infras, path lengths go up and travel time go down.
Paris -> Miramas on one arbitrary rolling stock: 790km -> 808km, 12:52 -> 12:04


core/src/test/java/fr/sncf/osrd/utils/graph/PathfindingTests.kt
```diff
@@ -702,6 +707,48 @@ class PathfindingTests {
         }
     }
 
+    /**
+     * Check that the cost function is used, shortest path is longest. This test use STDCM tooling
```
Erashin:
```suggestion
     * Check that the cost function is used, shortest path is longest. This test uses STDCM tooling
```

Erashin (approved):

core/src/main/kotlin/fr/sncf/osrd/utils/CachedBlockMRSPBuilder.kt
```diff
@@ -0,0 +1,44 @@
...
+/** Used to compute block MRSPs and min time required to reach a point, with proper caching */
+data class CachedBlockMRSPBuilder(
+    val rawInfra: RawInfra,
+    val blockInfra: BlockInfra,
+    val rollingStock: PhysicsRollingStock?,
+) {
+    private val mrspCache = mutableMapOf<BlockId, Envelope>()
+
+    // 320km/h as default value (global max speed in France)
+    private val rsMaxSpeed = rollingStock?.maxSpeed ?: (320.0 / 3.6)
```
Erashin, line 23: Comment disappeared for some reason.
This could be a pb in the future for other countries. For example, germany's top train speed is 550hkm/h, which is a lot higher than France's. Not saying we could /should handle this right now, but we'll have to keep it in mind.

eckter: I don't think German's train lines are equipped for trains to go faster than 300km/h.

Ideally we'd change the editoast -> core API to include some simulation data, but that seems a little overkill here. The whole process does a lot of approximations anyway.

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/RemainingDistanceEstimator.kt
```diff
@@ -13,6 +13,13 @@ import fr.sncf.osrd.utils.units.Offset
 /**
  * This is a function object that estimates the remaining distance to the closest target point,
  * using geo data. It is used as heuristic for A*.
+ *
+ * **LEGACY**
+ *
+ * Not currently in used as the geodesic computations are too expensive to compute, resulting in a
+ * net performance loss.
+ *
+ * TODO: before reusing this class, it now needs to return a minimum *time* instead of a distance.
```
Khoyo: I don't really this the value in keeping this class anymore, shouldn't we just delete it?

Khoyo (approved):
Otherwise LGTM

SharglutDev (approved):
Lgtm (not tested)

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/RemainingDistanceEstimator.kt
eckter: It's supposed to be fixed and used soon-ish. But that's the second comment on this, so I'll just remove it and link the removal in the relevant issue.

PR #8057, opened 2024-07-15
front: retain margin boundaries when set but constant

### LINKED ISSUE #8022: TSV2 : theoretical margin distribution
### What happened?

We insert a theoretical margin value at a certain PR1. We insert the same theoretical margin value for a different PR2 on a following line of the table. The results are the same for the margin distribution, as if we had not set the second margin value.

In this case, the margin value should be distributed between the PR1 and the PR2, and between PR2 and the end of the path. This is not the same as distributing the value between the PR1 and the ned of the path. 

1. Go to a scenario TSV2
2. Create a train
3. Insert a theoretical margin value of 5% at a certain PR1 in the times and stops table
4. Insert the same theoretical margin value of 5% for a different PR2 on a following line of the table
5. Go to simulation results and compare with a train without the 5% in the line of PR2

- **Browser**: Edge
- **OS**: Windows
- **Env**: Recette

staging de8ada8

emersion:
If the user sets two consecutive margins at 5%, we need to send two ranges [5%, 5%] to the backend instead of merging them together into a single margin.

Closes: https://github.com/OpenRailAssociation/osrd/issues/8022

### DIFF
--- a/front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
+++ b/front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
@@ -13,10 +13,7 @@ const formatMargin = (pathSteps: PathStep[]): Margin | undefined => {
         boundaries.push(step.id);
         values.push('none');
       }
-    }
-
-    // for the other steps, we add the margin if it's different from the previous one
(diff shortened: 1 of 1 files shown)


theocrsb (approved):
LGTM and tested !

SharglutDev (approved):

PR #8072, opened 2024-07-16
front: fix width and height of a train schedule tile can be expanded

### LINKED ISSUE #8033: The height of a train schedule tile can be expanded
### What happened?

A train schedule tile can have its height expanded when:
-  The mouse is over it
-  Selection is enabled
- The train schedule doesn't honor its scheduled points

2024-07-11-200116_1916x1037_scrot

All tiles should have the same height.

1. Go to a scenario with a train schedule that doesn't honor its scheduled points
2. Enable the selection
3. Set your mouse over it

_No response_

Firefox

b4c553ea5063d7ccbcb15626e51d1128164bc4b1

### LINKED ISSUE #8054: The width of a train schedule tile can be expanded
### What happened?

The width of a train schedule name can be expanded
image

We can hide the year of the date and show it only on hover and if the train name is too long, add an ellipsis.

Try to create a train with name that contain 128 char.

Dev (DgexSol)

Google Chrome

769de63

theocrsb:
close #8033 
close #8054


Math-R (approved):
Lgtm & tested

PR #8073, opened 2024-07-16
editoast: move PathItemLocation to its own openapi type

### LINKED ISSUE #8062: editoast: introduce a proper PathItemLocation type in OpenAPI
### Description and goal

Right now `PathItemLocation` is inlined in the OpenAPI schema, which makes it tricky to work with from the frontend: the type definition is duplicated and spread out in multiple parent types.

### Acceptance criteria

.

flomonster:
Removes unnecessary inline annotation.

close #8062

### DIFF
--- a/editoast/editoast_schemas/src/train_schedule/path_item.rs
+++ b/editoast/editoast_schemas/src/train_schedule/path_item.rs
@@ -24,7 +24,6 @@ pub struct PathItem {
     #[serde(default)]
     pub deleted: bool,
     #[serde(flatten)]
-    #[schema(inline)]
     pub location: PathItemLocation,
 }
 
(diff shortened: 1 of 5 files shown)


Tguisnet (approved):

emersion (approved):
Ah damn, I thought just dropping the `#[schema(inline)]` would require a whole bunch of refactoring but it seems like it doesn't!

LGTM

woshilapin (approved):
`front` will be happy 🎉 Thank you!

SharglutDev (approved):
Lgtm, thank you for that.

Before merge, could you switch the type [here](https://github.com/OpenRailAssociation/osrd/blob/dev/front/src/reducers/osrdconf/types.ts#L74) ? And remove the one we created above (forgot that there was already a type for that in osrdEditoast)

PR #8075, opened 2024-07-16
editoast: refactor the remaining tests

hamz2a:
Part of https://github.com/OpenRailAssociation/osrd/issues/6980

I've made several commits to make the review easier.
I will fixup the commits once the review is done.
`Pathfinding` still uses `db_pool_v1`, so I added `{function}_v1` to the fixtures to distinguish them from those using `db_pool_v2`.

### DIFF
--- a/editoast/src/modelsv2/fixtures.rs
+++ b/editoast/src/modelsv2/fixtures.rs
@@ -1,6 +1,8 @@
 use std::io::Cursor;
 
 use chrono::Utc;
+use editoast_schemas::infra::Direction;
+use editoast_schemas::infra::DirectionalTrackRange;
 use editoast_schemas::infra::InfraObject;
 use editoast_schemas::infra::RailJson;
 use editoast_schemas::primitives::OSRDObject;
@@ -8,8 +10,21 @@ use editoast_schemas::train_schedule::TrainScheduleBase;
 use postgis_diesel::types::LineString;
 
 use crate::infra_cache::operation::create::apply_create_operation;
+use crate::models::train_schedule::Mrsp;
+use crate::models::train_schedule::SimulationOutput;
+use crate::models::train_schedule::TrainScheduleValidation;
+use crate::models::Create as _;
 use crate::models::Pathfinding;
 use crate::models::PathfindingChangeset;
+use crate::models::PathfindingPayload;
+use crate::models::ResultPosition;
+use crate::models::ResultStops;
+use crate::models::ResultTrain;
+use crate::models::RoutePath;
+use crate::models::Scenario as ScenarioV1;
+use crate::models::SimulationOutputChangeset;
+use crate::models::Timetable as TimetableV1;
+use crate::models::TrainSchedule as TrainScheduleV1;
 use crate::modelsv2::prelude::*;
 use crate::modelsv2::rolling_stock_livery::RollingStockLiveryModel;
 use crate::modelsv2::timetable::Timetable;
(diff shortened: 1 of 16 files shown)


editoast/src/modelsv2/fixtures.rs
```diff
@@ -1,12 +1,30 @@
...
 use editoast_schemas::infra::RailJson;
 use editoast_schemas::primitives::OSRDObject;
 use editoast_schemas::train_schedule::TrainScheduleBase;
+use postgis_diesel::types::LineString;
 
 use crate::infra_cache::operation::create::apply_create_operation;
+use crate::models::train_schedule::Mrsp;
+use crate::models::train_schedule::SimulationOutput;
+use crate::models::train_schedule::TrainScheduleValidation;
+use crate::models::Create as CreateV1;
```
woshilapin: Since it's not used, I'm not sure that this is useful to alias it as `CreateV1`. You might as well do `use crate::models::Create as _;` so it imports the trait, but doesn't occupy any name in the namespace. Same comment for all the others below.

woshilapin (commented):
I like the effort for splitting up the PR in smaller commits to ease the review, but beware of dragons: commits still needs to be atomic, semantically self-contained, and buildable. I believe this is not the case for the current PR. Here are a few examples:

- In f90ab4d7e51217ec378edf2671997957caf6469e and 092a298a128c538a9a92e5010e8d8472f9f32d79, the signature of functions is changed, but the calls to these functions are not, which means these commits doesn't even compile.
- In 93dcc4676448e787a0b528324160a7d9a2cd07e1, 01fd18900a5d3ca08ffe8d66975aae79defedd71 and 84e7a70ea2de8d9649cb07b8ea2ed51a551dd137, functions are Added, but never used anywhere... which means the CI will fail on these commits (the compiler warns on unused values and our CI fails on warnings)

Also, some commits titles could be improved, less vague, like [clean up tests](cfda8dc656e9fc94c3e5509981923e30ffbef255) and [fix lints](fa2a4238ddebec9df17e3c32c682a45b8a55f323). That said, the last one should probably be a fixup of other commits.

editoast/src/modelsv2/fixtures.rs
hamz2a: Done.

hamz2a:
> I like the effort for splitting up the PR in smaller commits to ease the review, but beware of dragons: commits still needs to be atomic, semantically self-contained, and buildable. I believe this is not the case for the current PR. Here are a few examples:
> 
> * In [f90ab4d](https://github.com/OpenRailAssociation/osrd/commit/f90ab4d7e51217ec378edf2671997957caf6469e) and [092a298](https://github.com/OpenRailAssociation/osrd/commit/092a298a128c538a9a92e5010e8d8472f9f32d79), the signature of functions is changed, but the calls to these functions are not, which means these commits doesn't even compile.
> * In [93dcc46](https://github.com/OpenRailAssociation/osrd/commit/93dcc4676448e787a0b528324160a7d9a2cd07e1), [01fd189](https://github.com/OpenRailAssociation/osrd/commit/01fd18900a5d3ca08ffe8d66975aae79defedd71) and [84e7a70](https://github.com/OpenRailAssociation/osrd/commit/84e7a70ea2de8d9649cb07b8ea2ed51a551dd137), functions are Added, but never used anywhere... which means the CI will fail on these commits (the compiler warns on unused values and our CI fails on warnings)
> 
> Also, some commits titles could be improved, less vague, like [clean up tests](cfda8dc656e9fc94c3e5509981923e30ffbef255) and [fix lints](fa2a4238ddebec9df17e3c32c682a45b8a55f323). That said, the last one should probably be a fixup of other commits.

Thank you for your feedback. I have fixed the commits and tested each one to ensure they are buildable.

woshilapin (approved):
Thanks for the rework on the commit history 😊

PR #8077, opened 2024-07-16
Fix timestops margins

### LINKED ISSUE #8066: Cannot add a margin of 0%
### What happened?

When adding a 0% or 0min/100km in a margin, right now we get an editoast error : 

Capture d’écran 2024-07-16 à 11 30 40

- It should be possible to add a 0% margin to reset any previous margin.
- Remove "none" from possible value in margin values (and so editoast now accepts 0% or 0min/100km)
- set the origin default's margin to 0% after a pathfinding

1. Go to "Times and stop" tab
2. Add a 0% margin

_No response_

_No response_

65e7f5b2411c04690f5457712de980b68d842a22

### LINKED ISSUE #8067: Cannot set margin on origin
### What happened?

When setting a margin on the origin point, it disappears after saving.

_No response_

1. Go to "Times and stops"
2. Add a margin of 5% to the origin
3. Save
4. Open again the "Times and stops" tab
5. Witness the disappearance of the margin

_No response_

_No response_

65e7f5b2411c04690f5457712de980b68d842a22

SharglutDev:
See commits for details

close #8066 
close #8067


front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
```diff
@@ -6,16 +6,18 @@ const formatMargin = (pathSteps: PathStep[]): Margin | undefined => {
...
-        boundaries.push(step.id);
+        boundaries.push(id);
         values.push('none');
       }
-    } else if (step.theoreticalMargin) {
-      boundaries.push(step.id);
-      values.push(step.theoreticalMargin);
+    } else if (theoreticalMargin) {
+      boundaries.push(id);
+      values.push(theoreticalMargin.startsWith('0') ? 'none' : theoreticalMargin);
```
emersion: Seems safer to use `=== '0%'` here, I think? The user may enter e.g. `05%`.

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
```diff
@@ -6,16 +6,18 @@ const formatMargin = (pathSteps: PathStep[]): Margin | undefined => {
   const values: string[] = [];
 
   pathSteps.forEach((step, index) => {
+    const { id, theoreticalMargin } = step;
```
emersion: We can destructure the object directly in the function arguments here:

```js
pathSteps.forEach(({ id, theoreticalMargin }, index) => {
```

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -24,6 +24,8 @@ import { ISO8601Duration2sec } from 'utils/timeManipulation';
 
 import type { ManageTrainSchedulePathProperties } from '../types';
 
+const checkMarginValue = (margin: string) => (margin !== 'none' ? margin : '0%');
```
emersion: Suggestion: maybe `normalizeMarginValue` would be a better name?

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
SharglutDev: I thought about that but if we do that we need also to check `=== '0km/100km'`

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
SharglutDev: Oh yes, good point

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
emersion: Yeah, I think I'd prefer to tighten up the checks and hit editoast errors in weird cases like `000%`, instead of silently changing the value the user entered in weird cases like `05%`.

SharglutDev: Done

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatMargin.ts
SharglutDev: Done

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
SharglutDev: Done

emersion:
Seems like there is a strange side-effect of this PR: when no margin is set, clicking on the margin cell for the origin in the table prefills with "none":

out

emersion:
Also, after saving a margin on the origin and re-editing the train, there is a "0%" one showing up on the destination:

out

SharglutDev:
Thank you for spotting these. Should be fixed.

emersion (approved):
LGTM!

editoast throws an error when setting a 0% margin on origin but this is being discussed separately.

shenriotpro (approved):
LGTM for tests/

leovalais (approved):
LGTM (editoast; not tested)

Math-R (dismissed):
lgtm & tested

flomonster (approved):
Go :tada:

Math-R (approved):
Find a bug unrelated to this PR, lgtm

PR #8080, opened 2024-07-16
front:timestops: fix operational point removal

### LINKED ISSUE #8079: Margin/stop time remains after deleting waypoint
### What happened?

When deleting a waypoint, margins and stop times remain in the table. This is just a UI bug: they are properly removed when saving the edited train.

out

_No response_

1. Edit a train
2. Go to "Times and stops"
3. Add a 5% margin for a waypoint
4. Save
5. Edit train again
6. Click the red cross to delete the waypoint

_No response_

_No response_

e725429b688bf44edb8c7c8001b212c16903716d

SharglutDev:
When an operational point in the timestops table was removed from via, it was removed from the store but not updated in the op list

close #8079

### DIFF
--- a/front/src/modules/timesStops/utils.ts
+++ b/front/src/modules/timesStops/utils.ts
@@ -28,7 +28,8 @@ export const formatSuggestedViasToRowVias = (
   }
   return formattedOps.map((op, i) => {
     const pathStep = pathSteps.find((step) => matchPathStepAndOp(step, op));
-    const { arrival, onStopSignal, name, stopFor, theoreticalMargin } = pathStep || op;
+    const { name } = pathStep || op;
+    const { arrival, onStopSignal, stopFor, theoreticalMargin } = pathStep || {};
 
     const isMarginValid = theoreticalMargin ? marginRegExValidation.test(theoreticalMargin) : true;
 


front/src/modules/timesStops/utils.ts
```diff
@@ -28,7 +28,13 @@ export const formatSuggestedViasToRowVias = (
...
   return formattedOps.map((op, i) => {
     const pathStep = pathSteps.find((step) => matchPathStepAndOp(step, op));
-    const { arrival, onStopSignal, name, stopFor, theoreticalMargin } = pathStep || op;
+    const { arrival, onStopSignal, name, stopFor, theoreticalMargin } = pathStep || {
+      ...op,
+      stopFor: undefined,
+      arrival: undefined,
+      onStopSignal: false,
+      theoreticalMargin: undefined,
+    };
```
emersion: Ah! If we don't want to grab some props from `op` then maybe we can split into two assignments?

```js
const { name } = pathStep || op;
const { arrival, onStopSignal, stopFor, theoreticalMargin } = pathStep || {};
```

theocrsb (approved):
LGTM and tested !

achrafmohye (approved):
LGTM :)

front/src/modules/timesStops/utils.ts
SharglutDev: Done

kmer2016 (approved):
LGTM !

PR #8084, opened 2024-07-16
Fix records query for mvt tile generation

leovalais:
XYZ coordinates were provided in the wrong order.


flomonster (approved):
LGTM (not tested)

PR #8085, opened 2024-07-17
core: take step planned timing data into account in block availability

### LINKED ISSUE #7688: core: stdcm: generate occupations for points with specified times
We need to add "fake" occupations to force the train to arrive in the correct time window.

It's not 100% as straightforward as it may sound: zone occupations affects the signals before them. That means that adding an new zone use that ends at 10:00 at the arrival point would make the train arrive *after* 10:00 by a few minutes. 

We may need to generalize `BlockAvailability` to handle different types of "unavailability".

Erashin:
Fixes #7688.

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/api/api_v2/stdcm/STDCMEndpointV2.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/api/api_v2/stdcm/STDCMEndpointV2.kt
@@ -25,6 +25,7 @@ import fr.sncf.osrd.standalone_sim.makeElectricalProfiles
 import fr.sncf.osrd.standalone_sim.makeMRSPResponse
 import fr.sncf.osrd.standalone_sim.result.ElectrificationRange
 import fr.sncf.osrd.standalone_sim.runScheduleMetadataExtractor
+import fr.sncf.osrd.stdcm.PlannedTimingData
 import fr.sncf.osrd.stdcm.STDCMResult
 import fr.sncf.osrd.stdcm.STDCMStep
 import fr.sncf.osrd.stdcm.graph.findPath
@@ -33,9 +34,12 @@ import fr.sncf.osrd.train.RollingStock
 import fr.sncf.osrd.train.TrainStop
 import fr.sncf.osrd.utils.Direction
 import fr.sncf.osrd.utils.units.Offset
+import fr.sncf.osrd.utils.units.TimeDelta
 import fr.sncf.osrd.utils.units.meters
 import fr.sncf.osrd.utils.units.seconds
+import java.time.Duration.between
 import java.time.Duration.ofMillis
+import java.time.ZonedDateTime
 import org.takes.Request
 import org.takes.Response
 import org.takes.Take
@@ -67,6 +71,7 @@ class STDCMEndpointV2(private val infraManager: InfraManager) : Take {
             val trainsRequirements =
                 parseRawTrainsRequirements(request.trainsRequirements, request.startTime)
             val spacingRequirements = trainsRequirements.flatMap { it.spacingRequirements }
+            val steps = parseSteps(infra, request.pathItems, request.startTime)
 
             // Run the STDCM pathfinding
             val path =
@@ -75,11 +80,12 @@ class STDCMEndpointV2(private val infraManager: InfraManager) : Take {
                     rollingStock,
                     request.comfort,
                     0.0,
-                    parseSteps(infra, request.pathItems),
+                    steps,
                     makeBlockAvailability(
                         infra,
                         spacingRequirements,
                         workSchedules = request.workSchedules,
+                        steps,
                         gridMarginBeforeTrain = request.timeGapBefore.seconds,
                         gridMarginAfterTrain = request.timeGapAfter.seconds
                     ),
@@ -173,7 +179,11 @@ class STDCMEndpointV2(private val infraManager: InfraManager) : Take {
     }
 }
 
-private fun parseSteps(infra: FullInfra, pathItems: List<STDCMPathItem>): List<STDCMStep> {
+private fun parseSteps(
+    infra: FullInfra,
+    pathItems: List<STDCMPathItem>,
+    startTime: ZonedDateTime
+): List<STDCMStep> {
     if (pathItems.last().stopDuration == null) {
         throw OSRDError(ErrorType.MissingLastSTDCMStop)
     }
@@ -182,7 +192,14 @@ private fun parseSteps(infra: FullInfra, pathItems: List<STDCMPathItem>): List<S
             STDCMStep(
                 findWaypointBlocks(infra, it.locations),
                 it.stopDuration?.seconds,
-                it.stopDuration != null
+                it.stopDuration != null,
+                if (it.stepTimingData != null)
+                    PlannedTimingData(
+                        TimeDelta(between(startTime, it.stepTimingData.arrivalTime).toMillis()),
+                        it.stepTimingData.arrivalTimeToleranceBefore,
+                        it.stepTimingData.arrivalTimeToleranceAfter
+                    )
+                else null
             )
         }
         .toList()
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
@@ -10,6 +10,7 @@ import fr.sncf.osrd.envelope_utils.DoubleBinarySearch
 import fr.sncf.osrd.sim_infra.api.Path
 import fr.sncf.osrd.sim_infra.api.RawSignalingInfra
 import fr.sncf.osrd.standalone_sim.result.ResultTrain.SpacingRequirement
+import fr.sncf.osrd.stdcm.STDCMStep
 import fr.sncf.osrd.stdcm.infra_exploration.InfraExplorerWithEnvelope
 import fr.sncf.osrd.stdcm.preprocessing.interfaces.BlockAvailabilityInterface
 import fr.sncf.osrd.utils.units.Distance
@@ -23,8 +24,8 @@ import org.slf4j.LoggerFactory
 val blockAvailabilityLogger: Logger = LoggerFactory.getLogger("BlockAvailability")
 
 data class BlockAvailability(
-    val fullInfra: FullInfra,
     val incrementalConflictDetector: IncrementalConflictDetector,
+    val plannedSteps: List<STDCMStep>,
     val gridMarginBeforeTrain: Double,
     val gridMarginAfterTrain: Double,
 ) : BlockAvailabilityInterface {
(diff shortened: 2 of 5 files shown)


core/src/main/kotlin/fr/sncf/osrd/api/api_v2/stdcm/STDCMEndpointV2.kt
```diff
@@ -173,16 +179,30 @@ class STDCMEndpointV2(private val infraManager: InfraManager) : Take {
...
+    infra: FullInfra,
+    pathItems: List<STDCMPathItem>,
+    startTime: ZonedDateTime
+): List<STDCMStep> {
     if (pathItems.last().stopDuration == null) {
         throw OSRDError(ErrorType.MissingLastSTDCMStop)
     }
+    if (pathItems.first().stepTimingData == null && pathItems.last().stepTimingData == null) {
+        throw OSRDError(ErrorType.MissingDepartureOrArrivalTime)
+    }
```
eckter: We only need at least one value to be set, not necessarily first or last

core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
```diff
@@ -34,13 +36,109 @@ data class BlockAvailability(
...
+    private fun getStepAvailability(
+        infraExplorer: InfraExplorerWithEnvelope,
+        startOffset: Offset<Path>,
+        endOffset: Offset<Path>,
+        startTime: Double
+    ): BlockAvailabilityInterface.Availability {
+        var maximumDelay = 0.0
+        first@ for (step in plannedSteps) {
+            for (location in step.locations) {
+                if (location.edge == infraExplorer.getCurrentBlock()) {
```
eckter: We shouldn't just check the current block, during post-processing we check the availability of the whole path

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/stdcm/STDCMEndpointV2.kt
Erashin: In the feature it's specified we should at least have the step timing data either the departure or arrival. Is that no longer needed?

eckter: With the planned front-end we can expect to have either the start or end time, but we only *need* at least one time value on *any* step (including intermediate steps)

core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
```diff
@@ -34,13 +36,153 @@ data class BlockAvailability(
...
+                    infraExplorer,
+                    startOffset,
+                    pathStartTime,
+                    shiftedStartTime,
+                    endTime
+                )
+            val availability =
+                getMostRestrictiveAvailability(stepAvailability, conflictAvailability)
+
+            if (availability !is BlockAvailabilityInterface.Unavailable) {
```
eckter:
```suggestion
            if (availability is BlockAvailabilityInterface.Available) {
```

core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
```diff
@@ -34,13 +36,153 @@ data class BlockAvailability(
...
+        pathStartTime: Double,
+        startTime: Double,
+    ): BlockAvailabilityInterface.Availability {
+        var maximumDelayToStayAvailable = Double.POSITIVE_INFINITY
+        var minimumDelayToBecomeAvailable = 0.0
+        var firstConflictOffset = Offset<TravelledPath>(Double.POSITIVE_INFINITY.meters)
+
+        val incrementalPath = infraExplorer.getIncrementalPath()
+        first@ for (step in plannedSteps) {
+            for (location in step.locations) {
```
eckter: Moving the inner loop into its own method would remove the need for a label

Generally speaking the method could be broken down into smaller ones

eckter (commented):
First pass, I haven't checked the tests yet

Looks pretty good

core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/BlockAvailabilityTests.kt
```diff
@@ -636,4 +636,330 @@ class BlockAvailabilityTests {
...
+                    null,
+                    false,
+                    PlannedTimingData(
+                        30.seconds,
+                        (stepMaximumTolerance / 2).seconds,
+                        (stepMaximumTolerance / 2).seconds
+                    )
+                )
+            )
+        val stepMaximumDelay = stepMaximumTolerance - timeAtStep
```
eckter: The semantics are a little weird in this sequence.


`val stepMaximumDelay = stepMaximumTolerance - timeAtStep` means that the last available time is `stepMaximumTolerance`. It's true, but not 100% obvious.

We could introduce a variable named `lastAvailableTime` and define it as 30s + tolerance / 2. Which does result in 60s = `stepMaximumTolerance`, but it would explain how we got there.

core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/BlockAvailabilityTests.kt
```diff
@@ -636,4 +636,330 @@ class BlockAvailabilityTests {
...
+        val plannedStepOffset = Offset<Block>(100.meters)
+        val timeAtStep = explorer.interpolateDepartureFromClamp(Offset(plannedStepOffset.distance))
+        val timeAtZoneEnd = explorer.interpolateDepartureFromClamp(explorer.getSimulatedLength())
+        val steps =
+            listOf(
+                STDCMStep(
+                    listOf(PathfindingEdgeLocationId(blocks[0], Offset(100.meters))),
+                    null,
+                    false,
+                    PlannedTimingData(timeAtZoneEnd.seconds, 0.seconds, 30.seconds)
```
eckter: If I get this right the train arrives as the step is available, it doesn't match the comments

core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/BlockAvailabilityTests.kt
```diff
@@ -636,4 +636,330 @@ class BlockAvailabilityTests {
...
+                )
+            )
+        val stepUnavailableDuration = timeAtZoneEnd.seconds.seconds - timeAtStep
+
+        val requirements =
+            listOf(
+                // Requirement ending just before step becomes available
+                SpacingRequirement(zoneNames[0], 0.0, stepUnavailableDuration, true),
+                // Requirement starting just after step becomes available
```
eckter:
```suggestion
                // Requirement starting just after step becomes unavailable
```

Either a typo or the code doesn't match

eckter (commented):
Unit tests look good as well

Careful though, there's an integration test that fails

core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/BlockAvailabilityTests.kt
Erashin: Actually, TimeAtZoneEnd is the end of the time at the end of the simulated length, which is the first zone. The end of the zone is not 100.meters, it's a lot larger. Time at zone end is like 160s, time at 100.meters is 3.3333s. Making the step available at timeAtZoneEnd is a bit overkill but ensures that the step is unavailable.

core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/BlockAvailabilityTests.kt
Erashin: I kinda went another way for the following tests, and should have modified this one. Will make it clearer ;)

core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/BlockAvailabilityTests.kt
Erashin: Hmm. Not exactly. The unavailable step becomes available if you delay it by `stepUnavailableDuration`. As a result, this conflict, starting at `timeAtZoneEnd + stepUnavailableDuration` does indeed start just after the step becomes available if we add delay. I could put `step becomes available after adding stepUnavailableDuration as delay`, would that help make it clearer?

core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
Erashin: There should be a way to make this better yeah (either that or using `when`). I'll modify this.

core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
```diff
@@ -77,6 +219,36 @@ data class BlockAvailability(
...
+            return secondAvailability
+        } else {
+            if (
+                (firstAvailability as BlockAvailabilityInterface.Available).maximumDelay <=
+                    (secondAvailability as BlockAvailabilityInterface.Available).maximumDelay
+            )
+                return firstAvailability
+            return secondAvailability
+        }
+    }
```
Khoyo: This looks like a good case for using when

Khoyo (approved):

core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/BlockAvailabilityTests.kt
```diff
@@ -636,4 +636,332 @@ class BlockAvailabilityTests {
...
+                    PlannedTimingData(timeAtZoneEnd.seconds, 0.seconds, 30.seconds)
+                )
+            )
+        val stepUnavailableDuration = timeAtZoneEnd.seconds.seconds - timeAtStep
+
+        val requirements =
+            listOf(
+                // Requirement ending just before step becomes available
+                SpacingRequirement(zoneNames[0], 0.0, stepUnavailableDuration, true),
+                // Requirement starting just after step becomes available after addinf
```
eckter:
```suggestion
                // Requirement starting just after step becomes available after adding
```

core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/BlockAvailabilityTests.kt
eckter: It still feels wrong, the step becomes available at `timeAtZoneEnd`, not `timeAtZoneEnd - timeAtStep`. This isn't related to the traversal time, it should be an absolute time (if the names match)

Erashin: Seen with @eckter. This test is quite hard to undertand, and the test case is already tested by `testPassingBeforePlannedStep`: test was removed.

eckter (approved):

PR #8087, opened 2024-07-17
Fix cleanup-db script deleting redis cache

### LINKED ISSUE #8082: scripts/load-backup: clear redis cache
### Description and goal

Run `REMOVE *` in the redis contain after loading a backup. Without this the cache contains stale data and the app misbehaves, see e.g. https://github.com/OpenRailAssociation/osrd/issues/7802.

### Acceptance criteria

.

flomonster:
close #8082


bougue-pe (approved):
Thanks, ~it's a good thing anyway, but it looks like a more general solution would be nice to have too (to be discussed aside).~

EDIT: this is a specific case when version number is stable (empty)

PR #8091, opened 2024-07-17
editoast: stdcm: pre-filter requirements that are not in the range

### LINKED ISSUE #7649: editoast: stdcm: pre-filter train requirements that can't affect the new train
When editoast build the stdcm request to be sent to core, it fetches all the train requirements for the scheduled trains. See [here](https://github.com/OpenRailAssociation/osrd/blob/dev/editoast/src/views/v2/timetable/stdcm.rs#L183-L197)

At this point, we know when the train is supposed to start and when it will end (based on max running time). We could easily pre-filter the train requirements to only include the ones in this window.

It would make the request smaller (they're currently around 20MB for an imported timetable), and it would speed some things up in core. We could do the filtering on the core side, but we'd miss out on the request size reduction. 

It's a prerequisite for stdcm requests in timetables that span over more than a few hours.

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/7649

I haven't run an actual benchmark (some of the tooling isn't up to date for tsv2), but I checked that the payload are much smaller when the timetable spans over a large time frame

### DIFF
--- a/editoast/src/views/v2/timetable/stdcm.rs
+++ b/editoast/src/views/v2/timetable/stdcm.rs
@@ -20,7 +20,7 @@ use utoipa::IntoParams;
 use utoipa::ToSchema;
 
 use crate::core::v2::pathfinding::PathfindingResult;
-use crate::core::v2::simulation::SimulationResponse;
+use crate::core::v2::simulation::{RoutingRequirement, SimulationResponse, SpacingRequirement};
 use crate::core::v2::stdcm::STDCMResponse;
 use crate::core::v2::stdcm::TrainRequirement;
 use crate::core::v2::stdcm::{STDCMPathItem, STDCMWorkSchedule, UndirectedTrackRange};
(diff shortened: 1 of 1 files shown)


editoast/src/views/v2/timetable/stdcm.rs
```diff
@@ -274,6 +263,79 @@ async fn stdcm(
...
+/// that overlap with the possible simulation times.
+fn build_train_requirements(
+    trains: Vec<TrainSchedule>,
+    simulations: Vec<(SimulationResponse, PathfindingResult)>,
+    departure_time: DateTime<Utc>,
+    latest_simulation_end: DateTime<Utc>,
+) -> HashMap<i64, TrainRequirement> {
+    let mut trains_requirements = HashMap::new();
+    for (train, sim) in trains.iter().zip(simulations) {
+        let (sim, _) = sim;
```
flomonster:
```suggestion
    for (train, (sim, _)) in trains.iter().zip(simulations) {
```

editoast/src/views/v2/timetable/stdcm.rs
```diff
@@ -274,6 +263,79 @@ async fn stdcm(
...
+                    req.begin_time,
+                    req.zones
+                        .iter()
+                        .map(|zone_req| zone_req.end_time)
+                        .max()
+                        .unwrap_or(req.begin_time),
+                )
+            })
+            .collect();
+        if (!spacing_requirements.is_empty()) || (!routing_requirements.is_empty()) {
```
flomonster: Should we add a pre-check that the train start simulation overlaps the `departure_time` and the `latest_simulation_end`. 

By doing so, we can remove this check and avoid filtering all resources occupied by trains.

flomonster (changes requested):
Great PR! Thanks.

editoast/src/views/v2/timetable/stdcm.rs
eckter: I've added a pre-check for each train.

It's not as straightforward: a scheduled train can start before the earliest time and still affect the new train (if it lasts long enough). I've kept the resource filtering for the same reason, sometimes only part of the resources are in the range.

But I could remove the last check (that there is at least one overlapping requirement)

flomonster (approved):
LGTM (reviewed + tested)

PR #8094, opened 2024-07-17
front: check input size of train name and update checkfields methods

### LINKED ISSUE #8051: front: no specific error message when input is too long
### What happened?

- No specific error message when input is too long in "train name" when creating a train. 
When the train name exceeds 128 characters there is an error from backend.

We can check before to send request if the train name exceeds 128 char.

For the error from backend. Try to create a train with a name that exceeds 128 char.

Dev (DgexSol)

Google Chrome

769de63

theocrsb:
close #8051


front/src/applications/operationalStudies/utils.ts
```diff
@@ -246,3 +246,6 @@ export const formatSpaceTimeData = (
     trainName: trainName || 'Train name not found',
   };
 };
+export const checkNameInvalidity = (name?: string | null) => (name && name.length > 128) || !name;
+export const checkFieldInvalidity = (length: number, field?: string | null) =>
```
clarani: - As the function returns a boolean, it is better to use a name like nameIsInvalid or stringIsInvalid ;) 
- I think you can move the stringIsInvalid in the file utils/string 
- Also you can rename the param length -> maxLengthAuthorized
```suggestion
export const nameIsInvalid = (name?: string | null) => (name && name.length > 128) || !name;
export const stringIsInvalid = (maxLengthAuthorized: number, field?: string | null) =>
```

front/src/modules/project/utils.ts
```diff
@@ -9,10 +11,10 @@ const checkProjectFields = (
   funders: boolean;
   budget: boolean;
 } => ({
-  name: !project.name || project.name.length > 128,
-  objectives: (project.objectives ?? '').length > 4096,
-  description: (project.description ?? '').length > 1024,
-  funders: (project.funders ?? '').length > 255,
+  name: checkNameInvalidity(project.name),
+  objectives: checkFieldInvalidity(4096, project.objectives),
```
clarani: Avoid using magic numbers. You can set for instance (in utils/strings.ts for instance):
```
export const TEXT_AREA_MAX_LENGTH = 4096;
export const TEXT_INPUT_MAX_LENGTH = 255;
export const SMALL_INPUT_MAX_LENGTH = 128;
```
and then:
```suggestion
  objectives: checkFieldInvalidity(TEXT_AREA_MAX_LENGTH, project.objectives),
```

front/public/locales/en/operationalStudies/manageTrainSchedule.json
```diff
@@ -34,6 +34,7 @@
     "noTrainStepInput": "Must be > 0",
     "softErrorInfra": "The application is unavailable, a server error has occurred, please try again later.",
     "The requested train couldn't reach its destination": "The train cannot reach its destination",
+    "trainScheduleNameInvalid": "Required field. The name must not exceed 128 characters",
```
clarani: Do you think you can add 2 messages ? One 'Required field' and another one 'The name must not exceed 128 characters' ?

theocrsb: Sure, it's done

front/src/modules/project/utils.ts
theocrsb: Done

front/src/applications/operationalStudies/utils.ts
theocrsb: Done

front/src/applications/operationalStudies/utils.ts
```diff
@@ -246,3 +247,5 @@ export const formatSpaceTimeData = (
     trainName: trainName || 'Train name not found',
   };
 };
+export const nameIsInvalid = (name?: string | null) =>
+  (name && name.length > SMALL_INPUT_MAX_LENGTH) || !name;
```
clarani: You can change the condition like this, it can be read like this: "the name is invalid if it is undefined (or empty string) OR if its length is bigger than the max length (implicitely, the name is not undefined)"
```suggestion
  !name || name.length > SMALL_INPUT_MAX_LENGTH;
```

front/src/utils/strings.ts
```diff
@@ -66,3 +66,11 @@ export function convertInputStringToNumber(str: string) {
...
   return uic.toString().replace(/^87/, '');
 }
+
+export const TEXT_AREA_MAX_LENGTH = 4096;
+export const SMALL_TEXT_AREA_MAX_LENGTH = 1024;
+export const TEXT_INPUT_MAX_LENGTH = 255;
+export const SMALL_INPUT_MAX_LENGTH = 128;
+
+export const stringIsInvalid = (maxLengthAuthorized: number, field?: string | null) =>
+  field ? field.length > maxLengthAuthorized : false;
```
clarani: The string is invalid if the field is defined (if it is undefined, then it is valid) AND if its length is too long
```suggestion
  field && field.length > maxLengthAuthorized;
```

front/src/modules/trainschedule/components/ManageTrainSchedule/TrainSettings.tsx
```diff
@@ -91,8 +94,8 @@ export default function TrainSettings() {
           id="trainSchedule-name"
           onChange={(e: React.ChangeEvent<HTMLInputElement>) => setName(e.target.value)}
           value={name}
-          isInvalid={!name}
-          errorMsg={t('errorMessages.mandatoryField')}
+          isInvalid={invalidName}
+          errorMsg={`${t('errorMessages.requiredField')} ${t('errorMessages.nameLengthLimit')}`}
```
clarani: Ah actually, I thought that you could differentiate the 2 cases:
- if the name is not defined, then we should display the requiredField
- if the name is defined but too long, we should display the nameLengthLimit message

front/src/applications/operationalStudies/utils.ts
theocrsb: Done

front/src/utils/strings.ts
theocrsb: Done

front/src/modules/trainschedule/components/ManageTrainSchedule/TrainSettings.tsx
theocrsb: Done

front/src/utils/strings.ts
theocrsb: Done,but I changed the function a bit like this 
`export const stringIsInvalid = (maxLengthAuthorized: number, field?: string | null): boolean =>
  !!field && field.length > maxLengthAuthorized;` 
 to return a boolean if field is falsy and not get a typescript error

clarani: Nice !

clarani (approved):

front/src/modules/trainschedule/components/ManageTrainSchedule/TrainSettings.tsx
```diff
@@ -77,6 +78,8 @@ export default function TrainSettings() {
     setStartTime(startTimeFromStore.substring(0, 19));
   }, [nameFromStore, departureTimeFromStore, initialSpeedFromStore]);
 
+  const invalidName = nameIsInvalid(name);
```
Uriel-Sautron:
```suggestion
  const isInvalidName = nameIsInvalid(name);
```
Maybe this name would be better because this variable is a boolean

Uriel-Sautron (approved):
LGTM !! Tested, good job.
Just a comment.

front/src/modules/trainschedule/components/ManageTrainSchedule/TrainSettings.tsx
theocrsb: Done

front/src/applications/operationalStudies/utils.ts
```diff
@@ -205,3 +206,6 @@ export const convertDepartureTimeIntoSec = (departureTime: string) => {
   const isoDateInSec = ms2sec(isoDateToMs(departureTime));
   return convertUTCDateToLocalDate(isoDateInSec);
 };
+
+export const isNameIsInvalid = (name?: string | null) =>
```
Uriel-Sautron:
```suggestion
export const isInvalidName = (name?: string | null) =>
```

front/src/utils/strings.ts
```diff
@@ -66,3 +66,11 @@ export function convertInputStringToNumber(str: string) {
 export function formatUicToCi(uic: number) {
   return uic.toString().replace(/^87/, '');
 }
+
+export const TEXT_AREA_MAX_LENGTH = 4096;
+export const SMALL_TEXT_AREA_MAX_LENGTH = 1024;
+export const TEXT_INPUT_MAX_LENGTH = 255;
+export const SMALL_INPUT_MAX_LENGTH = 128;
+
+export const isStringIsInvalid = (maxLengthAuthorized: number, field?: string | null): boolean =>
```
Uriel-Sautron:
```suggestion
export const isInvalidString= (maxLengthAuthorized: number, field?: string | null): boolean =>
```

front/src/applications/operationalStudies/utils.ts
theocrsb: Done

front/src/utils/strings.ts
theocrsb: Done

clarani (approved):
LGTM :white_check_mark:

Issue #6925, opened 2024-03-18
Design and implement a way to forward warnings from core to the editoast response


eckter:
EDIT: scope change: we don't have to forward it all the way to the user, including it in the editoast response + tracing is enough for automated checks. Impossible scheduled points are double-checked in the front-end anyway.

EDIT-2: do we actually want to rely on the front-end to detect core errors? That's highly unstable and error prone. It would make the APIs unusable by themselves. CLI tools are becoming unreliable and it won't be possible for someone to fork and make a different front-end.

### Who would benefit from this feature?

Everyone

### What is this feature about?

When something goes wrong in core, we either ignore it or return no result at all, there's no middle ground. But there are lots of cases where we could send a result that would be *a little off* while still being close enough (like being off by a few seconds in an allowance section).

### Why is this feature valuable?

Returning a valid result with a warning would let us detect bugs and inaccurate results without the user being completely blocked. It would let us both send warnings in small errors we ignored, and give slightly off results when we used to send an error. Currently, errors and assertions are overly strict and lots of them could be turned to warnings without too much trouble. 

Some checks are made in the front-end after the response is received, but this makes tests and sanity checks much harder (especially with the fuzzer)

### Anything else people should know?

We used to have something to log warnings in core, but it wasn't forwarded all the way to the front. I don't think it's used anymore. It should probably be remade differently (without having an extra parameter for every method). 

---

List of things that may be included as warnings (not exhaustive):

1. Impossible margins and scheduled points
    a. We could add a lot of context, such as the expected time, or how margin and points may be incompatible
    a. We can add warnings for things that aren't currently checked, such as overlapping scheduled points / zero-length margins
1. Infra errors
    a. It would need its own specific endpoint, but we currently generate *a lot* of non-blocking errors that just get ignored (and flood datadog)
1. Missing or invalid data (e.g. speed tag ranges, it's currently a mess)
2. STDCM:
    a. Stops that are longer than requested? Additional stops?

bloussou:
up to date no need to refine again

PR #8096, opened 2024-07-17
core: fix: follow scheduled point even if faster than margins

### LINKED ISSUE #8092: Scheduled point are not honored even if possible
### What happened?

Simulation tries to respect margins in priority even if scheduled points are not honored.

In practice, the function to compute how much time should be lost for the final_schedule is wrong. It should use the base simulation and add the calculated lost time, which may be smaller than the provisional time. 

Scheduled points should be honored in priority.
If the lost time is smaller than the provided margins a warning will be sent back to the user.

1. Create a scenario using small infra
2. Import [this timetable](https://github.com/user-attachments/files/16263930/train_schedules.json)
3. See the final schedule is the same as the provisional (but a scheduled point is not honored)

_No response_

_No response_

769de63

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/8092

We don't *quite* handle it perfectly yet: when we need to go faster than the margins, we ignore the time repartitions among different margin ranges (between the faulty pair of scheduled points).

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/standalone_sim/StandaloneSimulation.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/standalone_sim/StandaloneSimulation.kt
@@ -191,7 +191,9 @@ fun makeMRSPResponse(speedLimits: Envelope): MRSPResponse {
 
 /**
  * Build the final envelope from the max effort / provisional envelopes. The final envelope modifies
- * the margin ranges to match the scheduled points.
+ * the margin ranges to match the scheduled points. The added time is distributed over the different
+ * margin ranges, following a logic described in details on the OSRD website:
+ * https://osrd.fr/en/docs/reference/design-docs/timetable/#combining-margins-and-schedule
  */
 fun buildFinalEnvelope(
     maxEffortEnvelope: Envelope,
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/fr/sncf/osrd/standalone_sim/StandaloneSimulation.kt
```diff
@@ -217,24 +222,51 @@ fun buildFinalEnvelope(
...
+            // we just set the time for the scheduled point without more details. It will be easier
+            // to handle it properly when we'll have migrated to standalone sim v3.
+            val maxEffortSectionTime =
+                getMaxEffortEnvelopeTimeAt(point.pathOffset) -
+                    getMaxEffortEnvelopeTimeAt(prevFixedPointOffset)
+            val earliestPossibleArrival = prevFixedPointDepartureTime + maxEffortSectionTime
+            var maxEffortExtraTime = point.arrival.seconds - earliestPossibleArrival
+            if (maxEffortExtraTime < 0.0) {
+                standaloneSimLogger.warn("impossible scheduled point")
+                // TODO: raise warning: scheduled point isn't possible
```
Khoyo, line 253: I'm not sure what this TODO is, considering the above line

core/src/main/kotlin/fr/sncf/osrd/standalone_sim/StandaloneSimulation.kt
```diff
@@ -217,24 +222,51 @@ fun buildFinalEnvelope(
...
+                maxEffortExtraTime = 0.0
+            } else {
+                standaloneSimLogger.warn("scheduled point doesn't follow standard allowance")
+                // TODO: raise warning: scheduled point doesn't follow standard allowance
```
Khoyo, line 257: Ditto

Erashin (approved):

core/src/main/kotlin/fr/sncf/osrd/standalone_sim/StandaloneSimulation.kt
eckter: The above line is just logging, we want that info to be forwarded to the user. It's related to https://github.com/OpenRailAssociation/osrd/issues/6925

Khoyo (approved):

shenriotpro (approved):
LGTM for tests/

PR #8106, opened 2024-07-18
editoast: add endpoint to list configured speed_limit_tags

### LINKED ISSUE #7976: Editoast: add new endpoint to list the speedLimitTags in OSRD's config
To be used by front to enrich the list of available speedLimitTags (infra edition and simulation).

Output a list of speedLimitTags' short ids ~and their corresponding long names~
(EDIT: long names are not going to be used in the future).

Part of https://github.com/osrd-project/osrd-confidential/issues/364

bougue-pe:
To be used to complete list of possible speed-limit tags in simulation or infra editor.

Also:
* share (move) static assets between core and editoast (for `speed_limit_tags.yml`)
* rename dynamic assets path for explicitness in editoast
* little cleanup in bake file (unused `tests/data` when running editoast)

Fix: https://github.com/OpenRailAssociation/osrd/issues/7976


editoast/src/views/speed_limit_tags.rs
```diff
@@ -0,0 +1,24 @@
...
+
+#[utoipa::path(
+    tag = "speed_limit_tags",
+    responses(
+        (status = 200, description = "List of configured speed-limit tags", body = Vec<String>, example = json!(["V200", "MA80"])),
+    ),
+)]
+#[get("")]
+async fn speed_limit_tags() -> Result<Json<Vec<String>>> {
+    let speed_limit_tags_config = SpeedLimitTag::load();
```
woshilapin: `SpeedLimitTag` looks like a resource used by all actix workers. Loading it on each request looks weird. So my first thought was to load it as an Application data (see [here](https://github.com/OpenRailAssociation/osrd/blob/580140d68159ed3b858569e5278dc5022d0ec53a/editoast/src/main.rs#L416-L424)).

After that, you should be able to have it as a parameter of the current `speed_limit_tags()` handler like

```rust
async fn speed_limit_tags(speed_limit_tags: Data<SpeedLimitTag>) -> Result<Json<Vec<String>>> {
  // ...
}
```

editoast/src/generated_data/speed_limit_tags_config.rs
```diff
@@ -0,0 +1,22 @@
...
+pub struct SpeedLimitTag {
+    pub name: String,
+    #[serde(default)]
+    pub fallback_list: Vec<String>,
+    pub default_speed: f64,
+}
+
+impl SpeedLimitTag {
+    /// Get the speed-limit tags configuration
+    pub fn load() -> SpeedLimitTagsConfig {
```
woshilapin: I expect `SpeedLimitTag::load()` to return me a `SpeedLimitTag`, not a `HashMap<String, SpeedLimitTag>`. My guess is that should have implemented this instead.

```rust
impl SpeedLimitTagsConfig {
  pub fn load() -> SpeedLimitTagsConfig { todo!() }
}
```
... but you can't since `SpeedLimitTagsConfig` is only an alias on `HashMap` and you won't be able to implement something for `HashMap`. I see two options here (but would favor the first one).

# [New type pattern](https://rust-unofficial.github.io/patterns/patterns/behavioural/newtype.html)
Instead of

```rust
pub type SpeedLimitTagsConfig = HashMap<String, SpeedLimitTag>;
```
do this
```rust
pub struct SpeedLimitTagsConfig(HashMap<String, SpeedLimitTag>);
```

the drawback is that you would need to redefine all the functions of `HashMap` you need as a proxy (not necessarily bad as code architecture level, just more boilerplate).

# [Extension trait](http://xion.io/post/code/rust-extension-traits.html)
Since you can implement something if you either define the type or the trait, and in this case, you have an alias on the type so you don't define the type, let's make a trait.
```rust
trait Load {
  fn load() -> Self;
}
pub type SpeedLimitTagsConfig = HashMap<String, SpeedLimitTag>;
impl Load for SpeedLimitTagsConfig {
  fn load() -> Self { todo!() }
}
```
but I'm not really fond of this solution because it extends `HashMap` (although the implementation is constrained to a `K = String, V = SpeedLimitTag`).

editoast/src/views/speed_limit_tags.rs
```diff
@@ -0,0 +1,24 @@
...
+    let speed_limit_tags = speed_limit_tags_config.keys().cloned().collect();
```
woshilapin: I'm not sure that I understand, if only keys are used, why do we load in memory all the information. Will it be used later (in other endpoints)?

woshilapin (changes requested):
Looks like you had fun with all these docker shenanigans (and might still have some, considering what the CI is saying). But really nice, hope it ends up being a small fix.

I'd say the API of `SpeedLimitTagsConfig` has to be reworked.

docker/docker-bake.hcl
```diff
@@ -43,6 +43,7 @@ target "core-build" {
   target = "build_env"
   contexts = {
     test_data = "./tests/data"
+    static_assets = "./assets"
```
Khoyo, line 46: You need it in `target "core" {` core too (I can't comment there)

bougue-pe: Thanks, added in https://github.com/OpenRailAssociation/osrd/pull/8106/commits/e2ead0e450a800ab5df676b88272c85627a32f00 and fixed the CI 🙏

editoast/src/views/speed_limit_tags.rs
bougue-pe: Good idea, done in https://github.com/OpenRailAssociation/osrd/pull/8106/commits/e379f7c4c2816f42cddc3c79360b2d20e330beab
And tests added on the way.

Note that it could be extended to sprites (`/signaling_systems` endpoint), but I didn't dig that.
If you think I should, please shout 🙂

editoast/src/generated_data/speed_limit_tags_config.rs
bougue-pe: Done in https://github.com/OpenRailAssociation/osrd/pull/8106/commits/e379f7c4c2816f42cddc3c79360b2d20e330beab

editoast/src/views/speed_limit_tags.rs
bougue-pe: Indeed, restricted exposed api to `Vec<String>` in https://github.com/OpenRailAssociation/osrd/pull/8106/commits/e379f7c4c2816f42cddc3c79360b2d20e330beab
Kept the whole structure to "make sure" the file and deserializer are synced when reading it.

editoast/src/views/speed_limit_tags.rs
```diff
@@ -0,0 +1,47 @@
...
+    tag = "speed_limit_tags",
+    responses(
+        (status = 200, description = "List of configured speed-limit tags", body = Vec<String>, example = json!(["V200", "MA80"])),
+    ),
+)]
+#[get("")]
+async fn speed_limit_tags(
+    speed_limit_tag_ids: Data<SpeedLimitTagIds>,
+) -> Result<Json<SpeedLimitTagIds>> {
+    Ok(Json((**speed_limit_tag_ids).clone()))
```
woshilapin: Maybe this is less weird to understand?

```suggestion
    Ok(Json(speed_limit_tag_ids.get_ref().clone()))
```

woshilapin (approved):
Thank you for the improvements.

editoast/src/views/speed_limit_tags.rs
bougue-pe: Done in https://github.com/OpenRailAssociation/osrd/pull/8106/commits/bf8e3fecc408e6d398b2e581f4f49afa2e222d69

woshilapin (approved):
Thanks for the improvements. Good to merge on my side.

Erashin (approved):

Khoyo (approved):
LGTM

PR #8108, opened 2024-07-18
front: fix spacetimechart not updating

### LINKED ISSUE #8107: Front: Projection doesn't refresh when margins are applied multiple times
### What happened?

When modifying the margins of a train multiple times the SpaceTimeChart doesn't refresh every time

I expect the projection to be refresh on the graph every time the margin were updated

1- Select a train and apply some margin and save
2- Contast the projection changes
3- Reselect the train and change the previous margin an save
4- See that the projection doesnt change
5- Reload your browser, and see thaht the projection doesnt change neither
6- Select the train go to margin and see that the margin is still set
7- Click sur modify and the projection will change

Local

Google Chrome

bc52bb6908b78f4b609be9742559b0543e54b040

SharglutDev:
See commits description

close #8107

### DIFF
--- a/front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
+++ b/front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
@@ -132,8 +132,10 @@ const useSetupItineraryForTrainUpdate = (
                 margins: { boundaries: string[]; values: string[] }
               ) => {
                 // The first pathStep will never have its id in boundaries
-                if (stepIndex === 0)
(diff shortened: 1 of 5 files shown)


flomonster (changes requested):
This doesn't fix the bug

flomonster (changes requested):
It fixes the bug but I found a regression:

1. Create a timetable with multiple trains (having different path)
2. Try to change the projection path (it doesn't refresh the space time chart)

clarani:
The issue https://github.com/OpenRailAssociation/osrd/issues/8107 was solved by #8271

PR #8109, opened 2024-07-18
"Fix" stack overflows for large batch operations

### LINKED ISSUE #8090: Stack overflow when simulating trains
### What happened?

When I sent a medium sized payload (300 trains), and then called simulation_summary, editoast crashed with a stack overflow error

I expect editoast to never overflow its stack

1. Get a payload of 300 realistic trains (you can ask me for some)
2. Import them to a timetable
3. Call the summary with the ids of the 300 trains
4. Witness

Local

Other

f2d6885f0

leovalais:
Closes #8090

TW: hacks like this should be illegal, but here we are.

The stack overflow problem occurs in `OperationalPointModel::retrieve_batch_unchecked` when the list of IDs is large enough.

1. The problem likely happens either in `diesel` or in `libpq` since everything is fine before attempting to execute the query.
2. My first thought was that the `OrFilter` objects (produced by the `.or_filter()` expression method, cf. `retrieve_batch_impl.rs`) were nested beyond control that their recursive flattening would break the stack.
3. Replacing the `.or_filter()`s by a single `.filter()` with nested `.or()`s would have the same effect, so I didn't try that.
4. Using a raw boxed `sql_query` should have helped since the bindings are stored in a `Vec` internally. It didn't. (*)
5. Even a fully (unsafe) single raw SQL query made with `format!` and no bindings stack overflows. (*) So now I doubt that `diesel` is the problem, but googling `libpq stack overflow` isn't very helpful...
6. Copy pasting the query in a DBeaver (which I believe uses `libpq`, to double check) works.

(*) https://gist.github.com/leovalais/301ff4ae0117b0e644c76dd6850f8dac

If I had the time I'd do the following:

* try to reproduce the query using `libpq` directly,
* or try to reproduce the query with `diesel` and a MySQL backend (or anything but Postgres)
* if both of these were to work fine, I'd just burn my PC and become a beekeeper

I can't spend the rest of my sprint on this bug and this "fix" seems to work (even if it degrades the performance significantly, sorry 😞). If someone else with some time wants to pick it up from here, I'd be glad to help though.

### DIFF
--- a/editoast/editoast_derive/src/lib.rs
+++ b/editoast/editoast_derive/src/lib.rs
@@ -237,6 +237,10 @@ pub fn search_config_store(input: proc_macro::TokenStream) -> proc_macro::TokenS
 /// * `#[model(preferred = PREFERRED)]`: just like `#[model(preferred)]` for fields, but at the struct level.
 ///     Compound identifier syntax is supported. This option can be specified only once, including at field level.
 ///     It is NOT NECESSARY to also specify `#[model(identifier = PREFERRED)]`.
+/// * `#[model(batch_chunk_size_limit = usize)]` (default: [modelv2::DEFAULT_BATCH_CHUNK_SIZE_LIMIT]): there seem to be a bug from either diesel or libpq that causes
+///     a stack overflow for large batch chunk sizes. Until a better solution is found, this option allows to limit the
+///     size of each chunk on a per-model basis. Increasing this value could lead to stack overflows, decreasing it
+///     might degrade the performance of batch operations.
 ///
 /// ### Field-level options
 ///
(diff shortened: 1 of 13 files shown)


Castavo (commented):
This feels indeed hacky.

I don't know if I can judge on the tradeoff in performance

flomonster (approved):
I tested locally and it's working, but I couldn't reproduce the original bug :disappointed:

leovalais:
You probably need to generate a huge payload. It'd be interesting to binary search the minimum request size for 16G and 32G setups, that value may be useful to figure out what happens. (Or you can just remove a ram stick, its probably less work tbh. 🙃)

PR #8124, opened 2024-07-18
front: driver train schedule : fix display for base/eco simulation switch

### LINKED ISSUE #8105: Eco simulation is not available on the driver's sheet
### What happened?

When a train schedule has scheduled points but **no** margins. The driver sheet considers that there is no eco simulation to show.

image

We should see the switch when a margin **OR** a scheduled point is provided.

1. Go to a scenario
2. Create a train with a scheduled point
3. Go to the driver sheet
4. No switch available

Local

_No response_

67d50c4

Akctarus:
closes #8105

### DIFF
--- a/front/src/modules/trainschedule/components/DriverTrainScheduleV2/DriverTrainScheduleHeaderV2.tsx
+++ b/front/src/modules/trainschedule/components/DriverTrainScheduleV2/DriverTrainScheduleHeaderV2.tsx
@@ -46,7 +46,7 @@ const DriverTrainScheduleHeaderV2 = ({
     <>
       <div className="d-flex align-items-center">
         <h1 className="text-blue flex-grow-1">{train.train_name}</h1>
-        {isEco(simulatedTrain) && (
+        {isEco(train) && (
           <div className="text-uppercase">
             <OptionsSNCF
               name="driver-train-schedule-base-or-eco"
--- a/front/src/modules/trainschedule/components/DriverTrainScheduleV2/utils.ts
+++ b/front/src/modules/trainschedule/components/DriverTrainScheduleV2/utils.ts
@@ -2,10 +2,7 @@ import * as d3 from 'd3';
 import { uniq } from 'lodash';
 
 import type { TrackSectionEntity } from 'applications/editor/tools/trackEdition/types';
-import type {
-  PathPropertiesFormatted,
-  SimulationResponseSuccess,
-} from 'applications/operationalStudies/types';
+import type { PathPropertiesFormatted } from 'applications/operationalStudies/types';
 import { convertDepartureTimeIntoSec } from 'applications/operationalStudies/utils';
 import {
   osrdEditoastApi,
@@ -163,8 +160,13 @@ export function getAverageSpeed(posA: number, posB: number, speedList: PositionS
   return Math.round(averageSpeed * 3.6 * 10) / 10;
 }
 
-export const isEco = (simulatedTrain: SimulationResponseSuccess) =>
-  simulatedTrain.base.energy_consumption !== simulatedTrain.final_output.energy_consumption;
+export const isEco = (train: TrainScheduleBase) => {
+  const { schedule, margins } = train;
+  return (
+    (schedule && schedule.some((item) => item.arrival !== null)) ||
+    (margins && margins.values.some((value) => value !== '0%' && value !== '0min/100km'))
+  );
+};
 
 /**
  * Interpolate a speed or time value at a given position when the operational point's position
--- a/front/src/modules/trainschedule/components/DriverTrainScheduleV2/DriverTrainScheduleStopListV2.tsx
+++ b/front/src/modules/trainschedule/components/DriverTrainScheduleV2/DriverTrainScheduleStopListV2.tsx
@@ -3,25 +3,25 @@ import React, { useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 
 import type { SimulationResponseSuccess } from 'applications/operationalStudies/types';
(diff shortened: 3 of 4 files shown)


SharglutDev (commented):
I think that rather than removing the `isEco` function check, we should find a better way in this function to know if a train has an eco simulation.

The issue is that `final_output` is returned by the simulation endpoint even if there isn't any margin/schedule points added.
So when creating this function, my thought was that the energy_consumption would be different but that's not the case here.

@flomonster how would you check if the train is eco with the new model ?

flomonster:
> I think that rather than removing the `isEco` function check, we should find a better way in this function to know if a train has an eco simulation.
> 
> The issue is that `final_output` is returned by the simulation endpoint even if there isn't any margin/schedule points added. So when creating this function, my thought was that the energy_consumption would be different but that's not the case here.
> 
> @flomonster how would you check if the train is eco with the new model ?

Checking the energy consumption doesn't work if it has a point schedule that doesn't slow down the train. 
 
A way is to check the train schedule input as described in the issue:

> If the train schedule has a margin (higher than 0%) **OR** a scheduled point.

front/src/modules/trainschedule/components/DriverTrainScheduleV2/DriverTrainScheduleHeaderV2.tsx
```diff
@@ -46,7 +46,7 @@ const DriverTrainScheduleHeaderV2 = ({
     <>
       <div className="d-flex align-items-center">
         <h1 className="text-blue flex-grow-1">{train.train_name}</h1>
-        {isEco(simulatedTrain) && (
+        {isEco(simulatedTrain) || (train.schedule && train.schedule?.length > 1) ? (
```
flomonster: This doesn't check if there is a scheduled point.

You can have a non-empty `train.schedule` array but without a scheduled point:

```json
{
    // ...
    "schedule": [
      {
        "at": "id2544",
        "arrival": null, // <-- If arrival field is set to anything but `null` it's a scheduled point
        "stop_for": "P0D",
        "on_stop_signal": false,
        "locked": false
      }
    ]
}
```

flomonster (changes requested):

flomonster (approved):
LGTM

> [!IMPORTANT]
> Not tested

front/src/modules/trainschedule/components/DriverTrainScheduleV2/DriverTrainScheduleHeaderV2.tsx
```diff
@@ -46,7 +46,8 @@ const DriverTrainScheduleHeaderV2 = ({
     <>
       <div className="d-flex align-items-center">
         <h1 className="text-blue flex-grow-1">{train.train_name}</h1>
-        {isEco(simulatedTrain) && (
+        {(isEco(simulatedTrain) ||
+          (train.schedule && train.schedule.some((item) => item.arrival !== null))) && (
```
SharglutDev: The second check could be added inside `isEco()` function. Moreover, rather than having to pass 2 parameters (`train` and `simulatedTrain`) to the function, you can only pass `train` and do @flomonster proposal and checking inside if one the schedule arrivals is not null OR if one of the margin values is not 0% and not 0min/100km.

```suggestion
        {(isEco(train) && (
```

SharglutDev (changes requested):

front/src/modules/trainschedule/components/DriverTrainScheduleV2/utils.ts
```diff
@@ -163,8 +163,10 @@ export function getAverageSpeed(posA: number, posB: number, speedList: PositionS
   return Math.round(averageSpeed * 3.6 * 10) / 10;
 }
 
-export const isEco = (simulatedTrain: SimulationResponseSuccess) =>
-  simulatedTrain.base.energy_consumption !== simulatedTrain.final_output.energy_consumption;
+export const isEco = (train: TrainScheduleBase) =>
+  (train.schedule && train.schedule?.length > 1) ||
+  train.margins?.values.some((value) => Number(value.split('%')[0]) > 0) ||
+  train.margins?.values.some((value) => Number(value.split('min')[0]) > 0);
```
SharglutDev: I think you have some issues here : 
- for the schedules, you lost the fix from @Akctarus with the `.some()`
- when you use `margins?.values`, probably it will return undefined instead of a boolean and you don't want that
- for values, if you test `value > 0`, you have to iterate twice but you can just test that the value is different than 0% AND 0min/100km, all the other values will work as valid margin.
- the CI failing comes from the fact that `isEco` is used in 2 other components, you have to change the argument you pass to it in `DriverTrainScheduleV2.tsx` and for `DriverTrainScheduleStopListV2.tsx`, as its parent is `DriverTrainScheduleV2.tsx`, maybe you can change the props simulatedTrain to trainRegime and when passing the props in the parent do `trainRegime={baseOrEco === BaseOrEco.eco ? simulatedTrain.final_output : simulatedTrain.base}`

```suggestion
export const isEco = (train: TrainScheduleBase) => {
 const { schedule, margins } = train;
 return (schedule && schedule.some((item) => item.arrival !== null))) ||
 (margins && (margins.values.some((value) => value !== '0%' && value !== '0min/100km' ));
}
 
```

SharglutDev (changes requested):

front/src/modules/trainschedule/components/DriverTrainScheduleV2/utils.ts
theocrsb: Done

SharglutDev (approved):
Lgtm and tested, good job, thank you for finishing the PR @theocrsb

Issue #7928, opened 2024-07-03
integrate ui-core font in osrd


kmer2016:
Currently, we have two fonts in OSRD: one from UI-core, which is applied to components from OSRD-UI, and the older font in OSRD. The idea is to standardize the font by using the one from OSRD-UI, as it aligns with the new design system.

PR #8126, opened 2024-07-18
Display STDCM simulation table results 

### LINKED ISSUE #8125: STDCM new interface : Display good table results when a simulation is selected
When a simulation is selected from the list of available simulations, the results table must be updated with the correct data 

Mock-up : 

Image

### LINKED ISSUE #8128: STDCM new interface : Handle "No simulation found"
When a simulation fails, a message is displayed to inform the user. The inputs to this simulation must be stored with the simulation results, which will be of type “Error” or “Invalid” whatever. The failed simulation must appear in the list of simulations created.

- [x] The tab is displaying "No result", and the content of the tab is `No solution has been found for this configuration`. We can't display the reasons so far, this will be implemented with [this feature](https://github.com/osrd-project/osrd-confidential/issues/296).

Mock-up : 

image

- [x] The button to start another request must also be present in this tab. It opens a new tab in the navigator:
image

### LINKED ISSUE #8143: Match inputs with outputs : we need to display the inputs corresponding to the selected simulation

### LINKED ISSUE #8157: Selecting simulations that have failed
Simulations that fail for any reason lead to this display, with a box saying `The calculation has failed` and a short message `We are sorry that we were unable to process your request. The STDCM team has been notified and will take care of this shortly.`:

image

Akctarus:
closes #8125 
closes #8143
closes #8128
closes #8157 

This PR enables the launch of multiple simulations from the new STDCM view.

### Main Features:

- **Storage of Inputs and Outputs:** Stores the inputs and outputs of each simulation launch.
- **Simulation Navigation:** Allows users to navigate between different simulations. Display navigation button only when it is necessary.
- **Simulation Details Display:** When a simulation is selected, the user can see the inputs used and the outputs generated.
- **No result Handling:** Displays a message below the simulation selector banner when there are no results.
- **Simulation Retention:** Users can retain a simulation, preventing them from selecting another simulation.

### Testing the Case Where the Simulation Has No Result

To test the scenario where a simulation yields no result:
- Duplicate the rolling stock `VAPEUR 141R VAPEUR + TENDER FUEL-OIL`.
- Set the mass of the duplicated rolling stock to 10,000t.
- From the new STDCM view, select the new rolling stock.
- Set Chambéry-Poste-Bois as the origin and Modane as the destination.
- Launch the simulation and verify that no results are returned.


front/src/applications/stdcmV2/views/StdcmViewV2.tsx
SharglutDev, line 1: When launching a simulation for the first time, I have this warning in console : 

Capture d’écran 2024-07-25 à 10 10 33

front/src/applications/stdcmV2/components/StdcmResults/StdcmResults.tsx
SharglutDev, line 1: Capture d’écran 2024-07-25 à 10 15 07

I don't think we should still see the green block "Simulation Disponible"

SharglutDev, line 1: Capture d’écran 2024-07-25 à 10 25 17

Here we should display "Sans résultat" and still see the time of the request , see sketch below

Capture d’écran 2024-07-25 à 10 26 58

SharglutDev, line 1: Capture d’écran 2024-07-25 à 10 28 08

Here the `underline` should be below the border and not on top, see sketch below : 

Capture d’écran 2024-07-25 à 10 29 47

SharglutDev, line 1: Capture d’écran 2024-07-25 à 10 30 16

When the simulation didn't work, the `underline` is not displayed at the right place

SharglutDev, line 1: Capture d’écran 2024-07-25 à 10 31 04

When switching to a tab where the simulation failed, we shouldn't see the message from above anymore (only the one below)

SharglutDev, line 1: Capture d’écran 2024-07-25 à 10 32 17

Here we don't have a fading style when we don't see the entire title of the tab. See sketch below : 

Capture d’écran 2024-07-25 à 10 32 43

SharglutDev (changes requested):
Thank you for this PR, it starts to look very nice ! Didn't check the code yet but left some comments regarding the compatibility with the sketch.

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
kmer2016: Done

front/src/applications/stdcmV2/components/StdcmResults/StdcmResults.tsx
kmer2016: Done

Caracol3:
Good work !! It looks very nice 
I have one interrogation maybe for an enhancement later.
According to the sketch, when you choose a simulation, the other ones are loked and you can't see them anymore.
Shouldn't we keep the possibility to still see them ?
I think that the user who has made 20 simulations and want to download two of them (because he wants to do more than one proposition) will be disappointed to be only able to choose betwwen the two and not both.

kmer2016:
> Good work !! It looks very nice I have one interrogation maybe for an enhancement later. According to the sketch, when you choose a simulation, the other ones are loked and you can't see them anymore. Shouldn't we keep the possibility to still see them ? I think that the user who has made 20 simulations and want to download two of them (because he wants to do more than one proposition) will be disappointed to be only able to choose betwwen the two and not both.

Thank you !
I asked @thibautsailly about this, and he mentioned that it was designed to prevent users from retaining multiple simulations. However, I agree with you that users might feel frustrated if they cannot retain another simulation after retaining one.

front/src/applications/stdcmV2/components/StdcmResults/StdcmResults.tsx
kmer2016: Done

kmer2016: Done

front/src/applications/stdcmV2/components/StdcmConfig.tsx
```diff
@@ -71,24 +94,15 @@ const StdcmConfig = ({
...
-            disabled={isPending}
-            setCurrentSimulationInputs={setCurrentSimulationInputs}
-          />
-          <StdcmVias disabled={isPending} setCurrentSimulationInputs={setCurrentSimulationInputs} />
-          <StdcmDestination
-            disabled={isPending}
-            setCurrentSimulationInputs={setCurrentSimulationInputs}
-          />
+          <StdcmOrigin {...inputsProps} />
+          <StdcmVias {...inputsProps} />
```
SharglutDev, line 99: It looks like when I'm changing the stop duration of a via and leaving the focus, the pathfinding is relaunched but it shouldn't be the case. The stop duration is only necessary for the stdcm endpoint, not the pathfinding one.

SharglutDev, line 99: When changing the stop duration, the "launch simulation" button doesn't get displayed until I leave the focus of the input. I find it pretty confusing. Do you think, it could be possible to set `showBtnToLaunchSimulation` to true in the `onChange` and not in the `onBlur` ?

front/src/applications/stdcmV2/components/StdcmResults/StdcmResults.tsx
```diff
@@ -1,105 +1,130 @@
...
-                      operationalPointsList={operationalPointsList}
-                    />
-                  }
-                  fileName={`STDCM-${simulationReportSheetNumber}.pdf`}
-                >
-                  <Button label={t('simulation.results.downloadSimulationSheet')} />
-                </PDFDownloadLink>
+      )}
+
+      <StdcmSimulationBanner
```
SharglutDev: For some reason, when I switch between simulation, we briefly see the simulation button appearing and disappearing. And the viewport changes between simulation 4 and the others (simulation 4 has one more stop than the rest)

https://github.com/user-attachments/assets/4a7e4663-67ea-49ab-9dd1-be457270326b

front/src/applications/stdcmV2/components/StdcmResults/StdcmResults.tsx
SharglutDev: Same as the simulation button, I think this block should disappear as soon as we typed something in an input (not when leaving the input)

SharglutDev: When having a simulation not working, I get this : 

Capture d’écran 2024-07-30 à 10 31 08

`Sans résultat` is not displayed and when I switch between simulation and come back to the failing one. I still have the block on top displayed. Is that what @thibautsailly wants ?

SharglutDev: Looks like it's still on top

front/src/applications/stdcmV2/components/StdcmConfig.tsx
SharglutDev, line 1: I have these inputs setup and I would like to add a via before Mâcon. How can I do that ? Do I need to replace Macon by the other one and add Macon back ? How about when I have a lot of vias, do I need to start from scratch ? Could you please check this behavior with @thibautsailly ? :)

Capture d’écran 2024-07-30 à 10 39 33

front/src/applications/stdcmV2/components/StdcmConfig.tsx
```diff
@@ -97,8 +111,8 @@ const StdcmConfig = ({
               'pb-5': !pathfindingState.error,
             })}
           >
-            {currentSimulationInputs && (
-              <Button label={t('simulation.getSimulation')} onClick={clickOnSimulation} />
+            {showBtnToLaunchSimulation && (
+              <Button label={t('simulation.getSimulation')} onClick={startSimulation} />
```
SharglutDev, line 110: Sometimes when I click on the simulation button, the first click doesn't work (probably because the pathfinding isn't done). It's a little weird. Do you think we could add a loader or a message while we can't click on the button ? Maybe we need to check the behavior with @thibautsailly.

SharglutDev (changes requested):
Didn't start to review the code yet but here are some other comments (and left some previous openened)

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
```diff
@@ -5,60 +5,138 @@ import { useTranslation } from 'react-i18next';
...
+    stdcmV2Results,
+    pathProperties,
+  } = useStdcm();
   const { getScenarioID } = useOsrdConfSelectors();
   const scenarioID = useSelector(getScenarioID);
 
   const [currentSimulationInputs, setCurrentSimulationInputs] = useState<
-    StdcmSimulationResult['input'] | undefined
+    StdcmSimulationInputs | undefined
   >(undefined);
```
SharglutDev:
```suggestion
  const [currentSimulationInputs, setCurrentSimulationInputs] = useState<StdcmSimulationInputs>();
```

front/src/applications/stdcm/hooks/useStdcm.ts
```diff
@@ -43,6 +43,7 @@ const useStdcm = () => {
     STDCM_REQUEST_STATUS.idle
   );
   const [pathProperties, setPathProperties] = useState<ManageTrainSchedulePathProperties>();
+  const [isStdcmResultsEmpty, setIsStdcmResultsEmpty] = useState<boolean | undefined>(undefined);
```
SharglutDev: Can't this be false by default ? If I understand this state correctly, it's used to insert a failed simulation (only if path not found) in the list so initializing it with false shouldn't be an issue.

```suggestion
  const [isStdcmResultsEmpty, setIsStdcmResultsEmpty] = useState(false);
```

front/src/applications/stdcmV2/components/StdcmConfig.tsx
```diff
@@ -17,28 +18,40 @@ import StdcmDestination from './StdcmDestination';
...
+import type { StdcmSimulation, StdcmSimulationInputs } from '../types';
 
+/**
+ * inputs in different card inside the StdcmConfig component come from the stdcm redux store.
+ * currentSimulationInputs is the current simulation inputs which are not yet launched.
+ * selectedSimulation is the simulation that is currently selected from the list of simulations.
+ */
 type StdcmConfigProps = {
-  currentSimulationInputs: StdcmSimulationResult['input'] | undefined;
+  currentSimulationInputs: StdcmSimulationInputs | undefined;
```
SharglutDev:
```suggestion
  currentSimulationInputs?: StdcmSimulationInputs;
```

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
```diff
@@ -5,60 +5,138 @@ import { useTranslation } from 'react-i18next';
...
 
-  const [interactedResultsElements, setInteractedResultsElements] = useState(false);
+  const [simulationsList, setSimulationsList] = useState<StdcmSimulation[]>([]);
+  const [selectedSimulationIndex, setSelectedSimulationIndex] = useState<number>(-1);
+  const [showSuccessBanner, setShowSuccessBanner] = useState<boolean>(false);
```
SharglutDev:
```suggestion
  const [selectedSimulationIndex, setSelectedSimulationIndex] = useState(-1);
  const [showSuccessBanner, setShowSuccessBanner] = useState(false);
```

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
```diff
@@ -5,60 +5,138 @@ import { useTranslation } from 'react-i18next';
...
+    onInteractWithOutputs();
+  };
+
+  const startNewQuery = () => {
+    setSimulationsList([]);
+    setSelectedSimulationIndex(-1);
+    setRetainedSimulationIndex(-1);
+    setShowSuccessBanner(false);
+    dispatch(resetStdcmConfig());
+  };
```
SharglutDev: Feels like these could be moved in `StdcmResults` by changing some props and passing some states directly in it ? These 2 functions concern the result area so I think it would be better to have them close where they are used. What do you think ?

front/src/styles/scss/applications/stdcmV2/_results.scss
SharglutDev, line 1: When will we be able to use tailwind custom utility classes for colors ?

In the mean time, do you think it would be better to create some css variables for these colors so we can reuse them where we need ?

front/src/applications/stdcmV2/types.ts
```diff
@@ -1,28 +1,36 @@
...
-      tractionEngine: RollingStockWithLiveries | undefined;
-    };
+export type StdcmSimulationInputs = {
+  origin?: PathStep | null;
+  departureDate?: string;
+  departureTime?: string;
+  destination?: PathStep | null;
+  pathSteps?: (PathStep | null)[];
+  consist?: {
+    tractionEngine: RollingStockWithLiveries | undefined;
```
SharglutDev: Is there a reason why we don't use rollingStock for this property ? Is it because of the future implementation of the convoy ?

```suggestion
    tractionEngine?: RollingStockWithLiveries;
```

front/src/applications/stdcmV2/types.ts
```diff
@@ -1,28 +1,36 @@
...
-    consist?: {
-      tractionEngine: RollingStockWithLiveries | undefined;
-    };
+export type StdcmSimulationInputs = {
+  origin?: PathStep | null;
+  departureDate?: string;
+  departureTime?: string;
+  destination?: PathStep | null;
+  pathSteps?: (PathStep | null)[];
+  consist?: {
```
SharglutDev, line 10: We don't put speedlimittags here ?

front/src/applications/stdcmV2/components/StdcmConsist.tsx
```diff
@@ -76,10 +81,12 @@ const StdcmConsist = ({ setCurrentSimulationInputs, disabled = false }: StdcmCon
     } else {
       dispatch(updateRollingStockID(undefined));
     }
+    onInteractWithInputs();
   };
 
   const onSelectSuggestion = (option: StdcmSuggestionsConsistOption) => {
     dispatch(updateRollingStockID(option.value.id));
+    onInteractWithInputs();
```
SharglutDev: Should we do the same for speedlimittags ?

front/src/applications/stdcmV2/components/StdcmConfig.tsx
```diff
@@ -17,28 +18,40 @@ import StdcmDestination from './StdcmDestination';
...
 import StdcmOrigin from './StdcmOrigin';
 import StdcmVias from './StdcmVias';
-import type { StdcmSimulationResult } from '../types';
+import type { StdcmSimulation, StdcmSimulationInputs } from '../types';
 
+/**
+ * inputs in different card inside the StdcmConfig component come from the stdcm redux store.
+ * currentSimulationInputs is the current simulation inputs which are not yet launched.
+ * selectedSimulation is the simulation that is currently selected from the list of simulations.
+ */
```
SharglutDev:
```suggestion
/**
 * Inputs in different cards inside the StdcmConfig component come from the stdcm redux store.
 * CurrentSimulationInputs is the current simulation inputs which are not yet launched.
 * SelectedSimulation is the simulation that is currently selected from the list of simulations.
 */
```

front/src/applications/stdcmV2/components/StdcmResults/StdcmResultsTable.tsx
```diff
@@ -120,7 +112,7 @@ const StcdmResultsTable = ({
           })}
         </tbody>
       </table>
-      <div className="display-all" style={!isSimulationSelected ? {} : { borderRadius: '0' }}>
+      <div className="display-all" style={!isSimulationRetained ? {} : { borderRadius: '0' }}>
```
SharglutDev: We should rename this class with something like that 

```suggestion
      <div className="results-buttons" style={!isSimulationRetained ? {} : { borderRadius: '0' }}>
```

SharglutDev: Should we avoid using `style` and add a class `simulation-retained` and use it with `cx()` ?
```suggestion
      <div className="display-all" style={!isSimulationRetained ? {} : { borderRadius: '0' }}>
```

front/public/locales/fr/stdcm.json
```diff
@@ -10,18 +10,22 @@
...
   "simulation": {
-    "available": "Simulation disponible",
     "averageRequestTime": "Pour votre demande, le temps nécessaire est généralement de 90 secondes.",
     "calculatingSimulation": "Calcul en cours...",
     "getSimulation": "Obtenir la simulation",
+    "modifySearchCriteria": "Vous pouvez modifier vos critères de recherche pour trouver une solution.",
+    "notFound": "Désolé, aucune solution n'a été trouvée.",
     "pendingSimulation": "simulation en cours",
     "results": {
+      "available": "Simulation disponible",
```
SharglutDev: As it's not a big change : seems like sketch now use this text (need to change english as well)

```suggestion
      "available": "Calcul terminé",
```

front/src/applications/stdcmV2/components/StdcmResults/StdcmSimulationBanner.tsx
```diff
@@ -0,0 +1,75 @@
...
+      setSelectedSimulationIndex(index);
+      onInteractWithOutputs();
+    }
+  };
+
+  return (
+    <div className="simulation-banner">
+      <div className="simulation-item-wrapper">
+        {simulationsList?.map(({ creationDate }, index) => (
+          // eslint-disable-next-line jsx-a11y/no-static-element-interactions
```
SharglutDev: You can remove this line by adding a role="button" and tabIndex to the div

front/src/applications/stdcmV2/components/StdcmResults/StdcmSimulationBanner.tsx
```diff
@@ -0,0 +1,75 @@
...
+}: StdcmSimulationBannerProps) => {
+  const { t } = useTranslation('stdcm');
+  const shorterYear = true;
+
+  const onSelectSimulation = (index: number) => {
+    if (retainedSimulationIndex === -1) {
+      setSelectedSimulationIndex(index);
+      onInteractWithOutputs();
+    }
+  };
```
SharglutDev: Instead of updating PathSteps multiple times in all origin/vias/dest, we can dispatch here

```suggestion
  const onSelectSimulation = (index: number) => {
    if (retainedSimulationIndex === -1) {
      setSelectedSimulationIndex(index);
      onInteractWithOutputs();
      dispatch(updatePathSteps({ pathSteps: simulationsList[index].inputs.pathSteps! }));
    }
  };
```

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -15,6 +15,12 @@ export const stdcmConfSlice = createSlice({
   initialState: stdcmConfInitialState,
   reducers: {
     ...buildCommonConfReducers<OsrdStdcmConfState>(),
+    resetStdcmConfig(state: Draft<OsrdStdcmConfState>) {
```
SharglutDev, line 18: Should we reset speedlimittags as well ?

SharglutDev (changes requested):

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
```diff
@@ -5,60 +5,138 @@ import { useTranslation } from 'react-i18next';
...
 
+  const selectedSimulation = simulationsList[selectedSimulationIndex];
+
+  // hide the simulation available card when the user interacts with the outputs.
+  const onInteractWithOutputs = () => setShowSuccessBanner(false);
+
+  // hide the simulation available card when the user interacts with the inputs.
+  // Even if currentSimulationInputs changed each times inputs changed, we can't use it to hide the card.
+  // Because it can change when the user change the selected simulation too.
+  const onInteractWithInputs = () => setShowSuccessBanner(false);
```
SharglutDev: After a discussion with @thibautsailly, we noticed that we want to hide the success banner as soon as the user has a click interaction anywhere in the page. So it will save you a lot of prop drilling with these functions :)

SharglutDev (changes requested):

front/src/applications/stdcmV2/components/StdcmConfig.tsx
kmer2016: I added a new [enhancement](https://github.com/OpenRailAssociation/osrd/issues/8261) for this

front/src/applications/stdcmV2/components/StdcmResults/StdcmResults.tsx
kmer2016: Done. Show the block when the simulation calculation returns.

kmer2016: Done

kmer2016: Done

kmer2016: Done

front/src/applications/stdcmV2/components/StdcmConfig.tsx
kmer2016: Yes, that's because changing the stop duration triggers the `updateViaStopTimeV2` action, which looks like this:
```typescript
updateViaStopTimeV2(
      state: Draft<S>,
      action: PayloadAction<{ via: PathStep; duration: string }>
    ) {
      const {
        payload: { via, duration },
      } = action;
      state.pathSteps = state.pathSteps.map((pathStep) => {
        if (pathStep && pathStep.id === via.id) {
          return { ...pathStep, stopFor: duration };
        }
        return pathStep;
      });
    },
```
We can see that we modify the pathStep, so the pathfinding is relaunched.

kmer2016: I'm not sure if using `onChange` is the best approach to determine when to show or hide the button. Instead, we should base the decision on whether `selectedSimulation.inputs` is equal to `currentSimulationInputs` (the simulation being edited). Because, there are many where we can change the `currentSimulationInputs`

front/src/applications/stdcmV2/components/StdcmResults/StdcmResults.tsx
kmer2016: The button now appears and disappears as intended.

The viewport changes because the configuration form takes up more space in the last simulation compared to the first ones. So for me it's normal.

front/src/applications/stdcmV2/components/StdcmConfig.tsx
kmer2016: Done

front/src/applications/stdcmV2/components/StdcmResults/StdcmResults.tsx
kmer2016: The button now appears and disappears as intended.

The viewport changes because the configuration form takes up more space in the last simulation compared to the first ones. So for me it's normal.

front/src/applications/stdcmV2/components/StdcmConfig.tsx
kmer2016: I asked @thibautsailly about this, and we agreed that since pathfinding isn't long, the current behavior is acceptable.

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
kmer2016: Done

front/src/applications/stdcm/hooks/useStdcm.ts
kmer2016: Done

front/src/applications/stdcmV2/components/StdcmConfig.tsx
kmer2016: Done

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
kmer2016: Done

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
kmer2016: Thank you for your feedback.

I understand your point about moving the functions closer to where they are used in StdcmResults. However, I chose to define these actions in the View component for the following reasons:

- **Centralization of Logic**: Having all the state logic in the View component ensures that the state management is centralized, making it easier **to track and debug**.

- **Avoid Passing Multiple State Variables**: These functions manipulate multiple state variables. Keeping the functions in View allows us to pass a single function as a prop instead of multiple state variables, which simplifies the props passing and keeps the components cleaner.

- **Consistency**: By centralizing complex actions, we ensure that the logic is not scattered and is easier to manage.

While moving the functions to `StdcmResults` might make them appear closer to where they are used, it would require passing multiple state variables through several levels, making the code harder to maintain.

I hope this clarifies my approach. What do you think?

front/src/styles/scss/applications/stdcmV2/_results.scss
kmer2016: We will use tailwind when this [issue](https://github.com/OpenRailAssociation/osrd/issues/7928) will be merged.

It's a good idea to store colors in varaibles that we can then not only use where needed but also facilitate integration with tailwind. May be we can do that in another PR ?

front/src/applications/stdcmV2/types.ts
kmer2016: Yes, the form is not completed.

front/src/applications/stdcmV2/components/StdcmConsist.tsx
kmer2016: Since we don't use this anymore, we don't need ;)

front/src/applications/stdcmV2/components/StdcmConfig.tsx
kmer2016: Done, thank you !

front/src/applications/stdcmV2/components/StdcmResults/StdcmResultsTable.tsx
kmer2016: Done

kmer2016: Done

front/public/locales/fr/stdcm.json
kmer2016: Done

front/src/applications/stdcmV2/components/StdcmResults/StdcmSimulationBanner.tsx
kmer2016: Done

front/src/applications/stdcmV2/components/StdcmResults/StdcmSimulationBanner.tsx
kmer2016: You raise a very good point. However, I believe there's more to consider. We also need to push `rollingStockID`, `originDate`, and `originTime`. To handle all of these together efficiently, it's better to manage them in one place.

That's why I added a new action, resetStdcmConfigWithData, which I call from the StdcmView2 component. This approach allows us to centralize the logic, as explained in my previous comment. It simplifies the state management and keeps our code cleaner and more maintainable.

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
kmer2016: Done

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
```diff
@@ -1,66 +1,189 @@
...
+    }
+  }, [simulationsList]);
+
   return (
-    <div className="stdcm-v2">
+    <div
+      role="button"
+      tabIndex={0}
+      className="stdcm-v2"
+      onClick={(_) => setShowStatusBanner(false)}
```
SharglutDev:
```suggestion
      onClick={() => setShowStatusBanner(false)}
```

front/public/locales/en/stdcm.json
```diff
@@ -10,18 +10,29 @@
...
   "simulation": {
-    "available": "Simulation available",
     "averageRequestTime": "For your request, the time required is generally 90 seconds.",
     "calculatingSimulation": "Calculation in progress...",
     "getSimulation": "Get the simulation",
+    "modifySearchCriteria": "You can modify your search criteria to find a solution.",
     "pendingSimulation": "Simulation in progress",
     "results": {
+      "status": {
+        "completed": "Calcultaion completed",
```
```suggestion
        "completed": "Calculation completed",
```

front/public/locales/en/stdcm.json
```diff
@@ -10,18 +10,29 @@
...
+      "formatCreationDate": "{{month}}/{{day}}/{{year}} {{hours}}:{{minutes}}",
       "gesicoRequest": "and attach this document to your GESICO DSDM request.",
-      "selectThisSimulation": "Select this simulation",
-      "simulationNumber": "Simulation n°1",
+      "retainThisSimulation": "Retain this simulation",
+      "simulationName": {
+        "withOutputs": "Simulation n° {{simulationNumber}}",
+        "withoutOutputs": "Without result"
+      },
+      "notFound": "No solution had been found for this configuration.",
```
```suggestion
      "notFound": "No solution has been found for this configuration.",
```

front/public/locales/en/stdcm.json
SharglutDev, line 1: Could you order these 2 trad files (you can use `Sort JSON objects` vscode extension (if you don't have it already)

front/src/applications/stdcmV2/components/StdcmResults/StdcmSimulationNavigator.tsx
```diff
@@ -0,0 +1,105 @@
...
+        {showRightBtn && (
+          <div
+            className="scroll-btn right"
+            role="button"
+            tabIndex={0}
+            aria-label="Scroll right"
+            onClick={scrollRight}
+          >
+            <span>
+              <ChevronRight />
```
SharglutDev: Comparing to the sketch, I think we need to use the large icon (same for left chevron)

```suggestion
              <ChevronRight size='lg' />
```

front/src/styles/scss/applications/stdcmV2/_results.scss
```diff
@@ -0,0 +1,380 @@
...
+    line-height: 1.5rem;
+    margin-bottom: 2.25rem;
+    margin-top: 4.063rem;
+    .simulation-list-wrapper {
+      position: relative;
+      display: flex;
+      overflow: hidden;
+
+      .scroll-btn {
+        position: absolute;
```
SharglutDev, line 36: On the sketch (see screen below), it seems like the button is not placed according the end of the navigation (it goes beyond the horizontal bar) but with a 20px padding-right from the right edge of the screen.

Capture d’écran 2024-08-01 à 18 28 48

front/src/styles/scss/applications/stdcmV2/_results.scss
```diff
@@ -0,0 +1,380 @@
...
+        background-color: rgba(246, 248, 249, 1);
+        width: 3rem;
+        height: 3rem;
+        display: flex;
+        justify-content: center;
+        z-index: 1;
+
+        span {
+          color: rgb(24, 68, 239);
+          padding: 0.75rem;
```
SharglutDev: Looks like the the icon is not centered in the button

Capture d’écran 2024-08-01 à 18 20 24

front/src/applications/stdcmV2/components/StdcmResults/StdcmSimulationNavigator.tsx
```diff
@@ -0,0 +1,105 @@
...
+                  {t('formatCreationDate', formatDateToString(creationDate, true))}
+                </span>
+              </div>
+              {selectedSimulationIndex === index && (
+                <div className="selected-simulation-indicator" />
+              )}
+            </div>
+          ))}
+        </div>
+        {showRightBtn && (
```
SharglutDev, line 92: When I reach the end of the simulations list, the right button doesn't always disappear

front/src/applications/stdcmV2/components/StdcmConfig.tsx
SharglutDev: Ok might be a big struggle to change this behavior :)

front/src/applications/stdcmV2/components/StdcmResults/StdcmResults.tsx
SharglutDev: Ok got it. But sometimes, when switching between a simulation with few inputs to another with many, we totally lose the viewport of the simulation results (see video below) which I found weird because if I click on it, I want to see the results without having to scroll again. Do you know what I mean ?

https://github.com/user-attachments/assets/57a85401-414a-4f6a-8bd8-c4c737070c82

Not sure how to do it, but it would be nice, when we switch, to at least keep the viewport with the begining of the results. The very best would be to keep the viewport (the height position of the navigation where I click according to the whole page) at the exact same spot so we don't see the screen moving but only a glitch at the top if the form/map form have been moving.

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
SharglutDev: Ok, works for me :) Just wanted your reasons for this choice but I really like that the logic is grouped in this component !

front/src/styles/scss/applications/stdcmV2/_results.scss
SharglutDev: Ok, fine by me. Could you create an enhancement issue for that ?

front/src/applications/stdcmV2/components/StdcmResults/StdcmSimulationBanner.tsx
SharglutDev: Even better, very nice ! Just think the name is strange "resetting but with data". Maybe it could be `updateStdcmConfigWithData` or `loadSelectedSimulationInputs`. What do you think ?

front/src/applications/stdcm/hooks/useStdcm.ts
```diff
@@ -26,13 +26,22 @@ import { castErrorToFailure } from 'utils/error';
 import useStdcmResults from './useStdcmResults';
 import { checkStdcmConf, formatStdcmPayload } from '../utils/formatStdcmConfV2';
 
-const useStdcm = () => {
+/**
+ * hook to manage the stdcm request
```
SharglutDev:
```suggestion
 * Hook to manage the stdcm request
```

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
```diff
@@ -1,66 +1,189 @@
...
+    stdcmV2Results,
+    pathProperties,
+  } = useStdcm(false);
   const { getScenarioID } = useOsrdConfSelectors();
   const scenarioID = useSelector(getScenarioID);
 
-  const [currentSimulationInputs, setCurrentSimulationInputs] = useState<
-    StdcmSimulationResult['input'] | undefined
-  >(undefined);
+  const [currentSimulationInputs, setCurrentSimulationInputs] = useState<StdcmSimulationInputs>();
```
SharglutDev: To stay closer to the store behavior, do you think we could initiate the state like that ?
This way, you can : 
- remove the "?" in the type definition of the state for pathSteps
- remove the "undefined" checks where you update the state
- remove the check "where to replaceAtIndex" (it will always be at the end of the array as it's always defined) when updating destination

Optional : we could

```suggestion
  const [currentSimulationInputs, setCurrentSimulationInputs] = useState<StdcmSimulationInputs>({
    pathSteps: [null, null],
  });
```

front/src/styles/scss/applications/stdcmV2/_home.scss
```diff
@@ -1,12 +1,33 @@
 .stdcm-v2 {
+  /* 
+  We set the cursor to default to avoid the pointer cursor that appears due to role="button" on the div.
+  The role="button" makes the div appear clickable, which causes the cursor to change to a pointer on hover.
+  Using !important ensures that the cursor remains in its default state.
+*/
+  cursor: default !important;
```
SharglutDev, line 7: Sorry you had to do that :)

front/src/applications/stdcmV2/hooks/useHorizontalScroll.ts
```diff
@@ -0,0 +1,111 @@
...
+import { debounce } from 'lodash';
+
+const DEBOUNCE = 100;
+const useHorizontalScroll = (itemSelectorClassname: string, itemToShowCount: number) => {
+  const scrollableRef = useRef<HTMLDivElement>(null);
+  const [showLeftBtn, setShowLeftBtn] = useState(false);
+  const [showRightBtn, setShowRightBtn] = useState(false);
+  const selector = `.${itemSelectorClassname}`;
+
+  // Calculate the width of an item including its margin-right
```
SharglutDev: why not the margin left ? (maybe we can update the comment)

front/src/applications/stdcmV2/hooks/useHorizontalScroll.ts
SharglutDev, line 1: Insane work ! Wasn't a lib doing that to avoid you so much work ?

SharglutDev (changes requested):
Really nice improvements ! We're getting close :)

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
```diff
@@ -1,66 +1,189 @@
...
+  }, [selectedSimulation]);
 
   useEffect(() => {
-    setInteractedResultsElements(false);
+    setShowBtnToLaunchSimulation(!isEqual(currentSimulationInputs, selectedSimulation?.inputs));
   }, [currentSimulationInputs]);
 
+  useEffect(() => {
+    if (isPending) {
+      setShowBtnToLaunchSimulation(false);
```
kmer2016, line 84: use a const which follow the state instead of using an useEffect ? (seen with @SarahBellaha )

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
```diff
@@ -1,66 +1,189 @@
...
+    }
+  }, [isStdcmResultsEmpty]);
+
+  // select the last simulation in the list
+  useEffect(() => {
+    if (simulationsList.length > 0) {
+      setSelectedSimulationIndex(simulationsList.length - 1);
+    }
+  }, [simulationsList]);
+
```
kmer2016: can be merged seen with @SarahBellaha

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
kmer2016: Good point !

front/src/styles/scss/applications/stdcmV2/_home.scss
kmer2016: Don't worry , it much better than before like this.

front/src/applications/stdcmV2/hooks/useHorizontalScroll.ts
kmer2016: Yes, For more robustness we should add margin left too ;)

front/src/applications/stdcmV2/hooks/useHorizontalScroll.ts
kmer2016: I confess I didn't really look. :)

front/src/applications/stdcmV2/components/StdcmResults/StdcmSimulationBanner.tsx
kmer2016: Done

front/public/locales/en/stdcm.json
kmer2016: Done

front/public/locales/en/stdcm.json
kmer2016: Done

front/public/locales/en/stdcm.json
kmer2016: Done

front/src/applications/stdcmV2/components/StdcmResults/StdcmSimulationNavigator.tsx
kmer2016: Done

front/src/applications/stdcm/hooks/useStdcm.ts
kmer2016: Done

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
kmer2016: Done

front/src/applications/stdcmV2/hooks/useHorizontalScroll.ts
kmer2016: Done

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
kmer2016: Done

front/src/applications/stdcmV2/components/StdcmResults/StdcmResults.tsx
kmer2016: I haven't found an easy way to do it.

front/src/styles/scss/applications/stdcmV2/_results.scss
kmer2016: Done

front/src/styles/scss/applications/stdcmV2/_results.scss
```diff
@@ -0,0 +1,380 @@
...
+        }
+
+        &.left {
+          left: -0.75rem;
+        }
```
SharglutDev, line 53: I don't think you need this

```suggestion
```

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
```diff
@@ -1,66 +1,184 @@
...
-    <div className="stdcm-v2">
+    <div role="button" tabIndex={0} className="stdcm-v2" onClick={() => setShowStatusBanner(false)}>
       <StdcmHeader />
-      <div
-        className={cx('stdcm-container', {
-          'simulation-visible':
-            !isPending && !currentSimulationInputs && !interactedResultsElements,
-        })}
-      >
+      <div className={cx('stdcm-container', { 'simulation-visible': showStatusBanner })}>
```
SharglutDev: Is the style for class `.stdcm-container` defined somewhere ?

front/src/styles/scss/applications/stdcmV2/_results.scss
SharglutDev: It was because of an extra span around the chevron icon

front/src/applications/stdcmV2/components/StdcmResults/StdcmSimulationNavigator.tsx
SharglutDev: if (scrollable) {
      const totalItemWidth = calculateTotalItemsWidth();
      const contentExceedsVisibleWidth = totalItemWidth > scrollable.clientWidth;
      const diff = totalItemWidth - (scrollable.scrollLeft + scrollable.clientWidth);
      const canScrollRight =  diff > largeurBouton;
      console.log({
        contentExceedsVisibleWidth,
        canScrollRight,
        totalItemWidth,
        scrollableLeft: scrollable.scrollLeft,
        scrollableWidth: scrollable.clientWidth,
        diff,
      });
      setShowLeftBtn(scrollable.scrollLeft > 0);
      setShowRightBtn(contentExceedsVisibleWidth && canScrollRight);
    }

front/src/styles/scss/applications/stdcmV2/_results.scss
kmer2016: Done

front/src/applications/stdcmV2/components/StdcmResults/StdcmSimulationNavigator.tsx
kmer2016: Done

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
kmer2016: Nice catch !

front/src/styles/scss/applications/stdcmV2/_results.scss
kmer2016: I do need this. Just as I adjust the right scroll button, I also need to adjust the left scroll button to maintain symmetry.

without `left : -0.75rem;`
image

with `left : -0.75rem;`
image

front/src/styles/scss/applications/stdcmV2/_results.scss
kmer2016: yes https://github.com/OpenRailAssociation/osrd/issues/8337

front/src/applications/stdcmV2/components/StdcmResults/StdcmResults.tsx
SharglutDev: We will deal with that in a future PR if necessary.

front/public/locales/en/stdcm.json
```diff
@@ -10,19 +10,31 @@
...
       "displayMain": "Display main operational points",
       "downloadSimulationSheet": "Download simulation report sheet",
+      "formatCreationDate": "{{month}}/{{day}}/{{year}} {{hours}}:{{minutes}}",
       "gesicoRequest": "and attach this document to your GESICO DSDM request.",
-      "selectThisSimulation": "Select this simulation",
-      "simulationNumber": "Simulation n°1",
+      "notFound": "No solution has been found for this configuration.",
+      "retainThisSimulation": "Retain this simulation",
+      "simulationName": {
+        "withOutputs": "Simulation n° {{simulationNumber}}",
```
SharglutDev: (No space between `n°` and the number in the sketch)
```suggestion
        "withOutputs": "Simulation n°{{id}}",
```

front/src/styles/scss/applications/stdcmV2/_results.scss
SharglutDev: Oh yes my bad

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
```diff
@@ -1,66 +1,187 @@
...
+              <div className="simulation-status-banner">
+                <span className={cx('status', { failed })}>
+                  {failed ? t('failed') : t('completed')}
+                </span>
+              </div>
+            )} */}
+
+            {/* {showStatusBanner && true && (
+              <span className="error-message">{t('errorMessage')}</span>
+            )} */}
```
SharglutDev: Do we need this ?

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
```diff
@@ -1,66 +1,187 @@
...
+        setShowStatusBanner(true);
+      } else if (isStdcmResultsEmpty) {
+        setSimulationsList([
+          ...simulationsList,
+          {
+            id: newSmulationId,
+            creationDate: new Date(),
+            inputs: currentSimulationInputs,
+          },
+        ]);
```
SharglutDev: https://github.com/user-attachments/assets/7c562080-1823-4759-bc6b-64806c96b14e

Seems like something is wrong when the simulation fails. In the video, I have some simulations in the history, I just changed for a different input (chambery - modane), the simulation fails but when checking the result, I still have the one from the previous simulation.

Moreover, sometimes the simulation result is not displayed in the navigator and when launching another one, 2 simulations are popping together in the nav.

What's stranger is that it seems the proper simulation is saved but it's not properly displayed in the result, see below video : 

https://github.com/user-attachments/assets/d9edca7a-42b5-42b1-8633-3e5344fa4458

Maybe it comes from this useEffect refacto ?

front/public/locales/fr/stdcm.json
```diff
@@ -10,19 +10,31 @@
...
       "displayMain": "Afficher les jalons principaux",
       "downloadSimulationSheet": "Télécharger la fiche de simulation",
+      "formatCreationDate": "{{day}}/{{month}}/{{year}} {{hours}}:{{minutes}}",
       "gesicoRequest": "et joignez ce document à votre demande GESICO DSDM",
-      "selectThisSimulation": "Retenir cette simulation",
-      "simulationNumber": "Simulation n°1",
+      "notFound": "Aucune solution n’a été trouvée pour cette configuration",
+      "retainThisSimulation": "Retenir cette simulation",
+      "simulationName": {
+        "withOutputs": "Simulation n° {{id}}",
```
SharglutDev:
```suggestion
        "withOutputs": "Simulation n°{{id}}",
```

front/src/applications/stdcmV2/components/StdcmOrigin.tsx
```diff
@@ -31,12 +32,20 @@ const StdcmOrigin = ({ setCurrentSimulationInputs, disabled = false }: StdcmConf
...
   }, [origin, originDate, originTime]);
 
+  const onOriginDateInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    dispatch(updateOriginDate(e.target.value));
+  };
+
+  const onOriginTimeInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    dispatch(updateOriginTime(e.target.value));
+  };
+
```
SharglutDev, line 48: Maybe we can insert them directly in the `onChange` ?

front/src/applications/stdcmV2/components/StdcmVias.tsx
```diff
@@ -44,6 +44,15 @@ const StdcmVias = ({ disabled = false, setCurrentSimulationInputs }: StdcmConfig
     );
   };
 
+  const onClickDeleteVia = (index: number) => {
+    dispatch(deleteViaV2(index));
+  };
```
SharglutDev, line 49: Same here maybe we can put that directly in the `onClick` ?

SharglutDev (changes requested):

front/src/applications/stdcmV2/components/StdcmOrigin.tsx
kmer2016: I chose to do it this way to prepare for component testing. When the tests are implemented, these methods will be provided via a hook, so that the component is really a visual, dumb component devoid of logic and only declarative.
This hook will then be tested.

SharglutDev: Ok understood ! Should these be wrapped with a `useCallBack` then ?

front/public/locales/en/stdcm.json
kmer2016: Done

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
kmer2016: Done

front/src/applications/stdcmV2/components/StdcmOrigin.tsx
kmer2016: Yes, that's right! As recommended in the [react doc.](https://react.dev/reference/react/useCallback#optimizing-a-custom-hook)

SarahBellaha (approved):
Nice work 💪

front/src/styles/scss/applications/stdcmV2/_home.scss
```diff
@@ -92,244 +117,67 @@
...
+                0px 0px 0px 4px rgba(255, 255, 255, 1) inset,
+                0px 0px 0px 5px rgba(255, 104, 104, 1) inset,
+                0px 3px 7px -3px rgba(217, 28, 28, 0.4);
+            }
+          }
+
+          .error-message {
+            // display: flex;
+            // width: 34.125rem;
+            // margin: 1.375rem 0 3.438rem 21.125rem;
```
theocrsb: You can delete this commented code if you don't need it

SharglutDev (approved):
Lgtm and tested, great job !

PR #8140, opened 2024-07-19
front: fix blank page when train schedule with no arrival date

### LINKED ISSUE #8112: A train schedule breaks the frontend: dateTime.split is undefined
### What happened?

Using a specific train schedules breaks the frontend:

image

_No response_

1. Create a scenario
2. Import this [train_broken.json](https://github.com/user-attachments/files/16283415/train_broken.json)
3. The frontend fails

Local, Staging (DgexSol), Recette (SNCF)

Firefox

580140d

achrafmohye:
Closes #8112

### DIFF
--- a/front/src/modules/trainschedule/components/TimetableV2/utils.ts
+++ b/front/src/modules/trainschedule/components/TimetableV2/utils.ts
@@ -31,6 +31,3 @@ export const extractTagCode = (tag?: string | null) => {
 
 export const timetableHasInvalidTrain = (trains: TrainScheduleWithDetails[]) =>
   trains.some((train) => train.invalidReason);
-
-export const transformDateWithoutYear = (dateTime: string): string =>
-  `${dateTime.split('/')[0]}/${dateTime.split('/')[1]} ${dateTime.split('/')[2].slice(5)} `;
--- a/front/src/utils/date.ts
+++ b/front/src/utils/date.ts
@@ -1,5 +1,6 @@
 import dayjs from 'dayjs';
 import 'dayjs/locale/fr';
+import customParseFormat from 'dayjs/plugin/customParseFormat';
 import timezone from 'dayjs/plugin/timezone';
 import utc from 'dayjs/plugin/utc';
 
@@ -9,6 +10,7 @@ import { ISO8601Duration2sec } from './timeManipulation';
 
 dayjs.extend(utc);
 dayjs.extend(timezone);
+dayjs.extend(customParseFormat);
 
 export function timestampToHHMMSS(timestamp: number) {
   const date = new Date(timestamp * 1000);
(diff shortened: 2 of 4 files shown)


front/src/modules/trainschedule/components/TimetableV2/utils.ts
```diff
@@ -32,5 +32,8 @@ export const extractTagCode = (tag?: string | null) => {
 export const timetableHasInvalidTrain = (trains: TrainScheduleWithDetails[]) =>
   trains.some((train) => train.invalidReason);
 
-export const transformDateWithoutYear = (dateTime: string): string =>
-  `${dateTime.split('/')[0]}/${dateTime.split('/')[1]} ${dateTime.split('/')[2].slice(5)} `;
+export const transformDateWithoutYear = (dateTime: string): string => {
+  const parts = dateTime.split('/');
+  if (!parts[0] || !parts[1] || !parts[2]) return '';
+  return `${parts[0]}/${parts[1]} ${parts[2].slice(5)}`;
+};
```
flomonster: ## Structure

If I understand correctly, this function takes a date and formats it, ignoring the year.

Weirdly, the input is a string and not a `Date` type. We should use a standard library to do this kind of operation.
Date formatting is complicated and depends on the user's region. Doing it manually is tedious and error-prone.

Examples:
- Canada: `YYYY-MM-DD`
- France: `DD/MM/YYYY`
- Italy: `DD.MM.YY`
- US: `MM-DD-YYYY`

## The fix

- Returning an empty string when the input format is invalid is unpredictable and must be documented
- We should have unit tests to avoid regression

> [!NOTE]
> Ideally, changing the behavior of a function should necessarily lead to a modification of docs and tests.

flomonster (changes requested):
Thanks for your PR. 

I have some structural remarks on this code (before the PR) and on the fix (the PR itself).
It's important to revamp the logic of this code to make it more robust.

front/src/modules/trainschedule/components/TimetableV2/utils.ts
achrafmohye: Fixed, can you check behaviour and new implementation :)

theocrsb (approved):
LGTM and tested

front/src/utils/date.ts
```diff
@@ -162,3 +164,9 @@ export function convertIsoUtcToLocalTime(isoUtcString: string): string {
 export function addDurationToIsoDate(startTime: string, duration: string) {
   return dayjs(startTime).add(ISO8601Duration2sec(duration), 'second').format();
 }
+
+export function transformDateWithoutYear(dateTime: string) {
```
flomonster: > Weirdly, the input is a string and not a Date type.

Means that the function **MUST** take a Date and not a string.

```suggestion
export function serializeDateTimeWithoutYear(dateTime: Date) {
```

> [!NOTE]
> I suggest renaming the function name since this function doesn't transform the given argument but serializes it.

flomonster: > Returning an empty string when the input format is invalid is unpredictable and must be documented

Means that the function **MUST** be documented and should not have this unpredictable behavior.

> [!NOTE]
> If you take a `Date` as input you don't have to handle the case where the string is not a formated date.

flomonster: > We should have unit tests to avoid regression.

Is there any reason not to test it?
In our case, the affected function is particularly easy to test.

flomonster (changes requested):
Some comments were not taken into account.

eckter (approved):
LGTM, bug is fixed

flomonster (approved):
Looks okay.

> [!NOTE]
> The date management requires a code refactoring.
>
> The transformation process is weird, for example the `train.startTime` process is as follows 🤯:
> 1. String ISO8601
> 2. Dayjs obj
> 3. String custom  `DD/MM/YYYY HH:mm:ss`
> 4. Date obj
> 5. Dayjs obj
> 6. String custom `DD/MM HH:mm:ss`

clarani (approved):
LGTM, I created a enhancement to refacto this later : https://github.com/OpenRailAssociation/osrd/issues/8166

PR #8141, opened 2024-07-19
synchronize input table arrival/departure/stop

### LINKED ISSUE #7782: TSV2: input table: time synchronisation
# Description

Part of https://github.com/osrd-project/osrd-confidential/issues/535

# AC
- [x] In the input table, when the user has defined a requested arrival time, if the stop time is entered, the requested departure time is calculated (and vice versa).
- [x] The synchronization should be operational at all times between those three columns:
- If the departure time is edited, the stop time must be re-computed
- If the arrival time is edited, the stop time must be re-computed
- If the stop time is edited, the departure time should be re-computed

anisometropie:
closes [#7782](https://github.com/OpenRailAssociation/osrd/issues/7782)

### DIFF
--- a/front/src/common/types.ts
+++ b/front/src/common/types.ts
@@ -10,4 +10,7 @@ export const DATA_TYPES = {
   percent: 'PERCENT',
 };
 
-export type TimeString = string; // hh:mm:ss or hh:mm
+/**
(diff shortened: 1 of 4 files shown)


clarani (approved):

PR #8142, opened 2024-07-19
Adding path item times

### LINKED ISSUE #7990: Editoast + core: return the time at which the train crosses each waypoint in the train schedule
This is linked to #7738 and #7758

Instead of a flag `scheduled_points_honored`, the simulation output / summary should contain the time at which the train crosses waypoints that the train_schedule defined

flomonster:
close #7990 

## Description

Instead of having the `honored_scheduled_point` information, we return the simulated times at every path_item position.
The editoast API user is responsible for deducing if the simulation honored the scheduled points / respects the margins.


front/src/modules/trainschedule/components/TimetableV2/hooks.ts
```diff
@@ -205,6 +214,8 @@ const useTrainSchedulesDetails = (
     debouncedRollingstockFilter,
     validityFilter,
     scheduledPointsHonoredFilter,
+
+    // retour si margin non respecté ou si tolerance non respecté
```
clarani: Remaining comment ;)
```suggestion
```

front/src/applications/operationalStudies/utils.ts
```diff
@@ -209,3 +214,79 @@ export const convertDepartureTimeIntoSec = (departureTime: string) => {
...
+ * @param trainResult
+ * @param trainSummary
+ * @returns true if the scheduled points are not honored
+ */
+export const isScheduledPointsNotHonored = (
+  trainSchedule: TrainScheduleResult,
+  trainSummary: Extract<SimulationSummaryResult, { status: 'success' }>
+): boolean => {
+  if (trainSummary.path_item_times_final.length !== trainSchedule.path.length) {
+    throw new Error('Asserstion failed');
```
clarani: Typo. Maybe it would be useful to console.error a more detailed error
```suggestion
    throw new Error('Assertion failed');
```

theocrsb:
You can import this json on small infra to see trains that are not honored, too fast or normal 
[small_normal_notHonored_tooFast.json](https://github.com/user-attachments/files/16518676/small_normal_notHonored_tooFast.json)

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/standalone_sim/SimulationRequest.kt
```diff
@@ -88,6 +88,7 @@ class SimulationPath(
     val blocks: List<String>,
     val routes: List<String>,
     @Json(name = "track_section_ranges") val trackSectionRanges: List<TrackRange>,
+    @Json(name = "path_item_positions") val pathItemPositions: List<Offset<Path>>
```
Erashin, line 91: We talked and decided a while ago it'd be clearer in the apis if we used `Offset<TravelledPath>`. There's a couple other places where we need to do this change, so just leave it as it is for now, I'll do it later on in another PR :)

Erashin (approved):
Core part LGTM.

front/src/applications/operationalStudies/consts.ts
```diff
@@ -60,3 +60,5 @@ export const STUDY_TYPES = [
 export type StudyType = typeof STUDY_TYPES;
 
 export type PowerRestrictionRange = LinearMetadataItem<{ value: string }>;
+
+export const TOLERANCES: number = 1000;
```
clarani: You can use `ARRIVAL_TIME_ACCEPTABLE_ERROR_MS` from `modules/timesStops/const.ts` ;) 
```suggestion
```

front/src/applications/operationalStudies/utils.ts
```diff
@@ -209,3 +214,79 @@ export const convertDepartureTimeIntoSec = (departureTime: string) => {
...
+  }
+  if (!trainSchedule.schedule) return false;
+  return trainSchedule.schedule.some((schedule, index) => {
+    if (schedule.arrival === null) return false;
```
clarani: You can also test that schedule.arrival is not undefined, and you can remove the cast l237
```suggestion
    if (schedule.arrival === null || schedule.arrival === undefined) return false;
```
then: 
```
sToMs(ISO8601Duration2sec(schedule.arrival))
```

front/src/applications/operationalStudies/utils.ts
```diff
@@ -209,3 +214,79 @@ export const convertDepartureTimeIntoSec = (departureTime: string) => {
...
+    // need to add +1 because origin will never be in schedules
+    return (
+      Math.abs(
+        sToMs(ISO8601Duration2sec(schedule.arrival as string)) -
+          trainSummary.path_item_times_final[index + 1]
+      ) >= TOLERANCES
+    );
```
clarani: I think it will be easier to read if you set a const to store this computation ;)

```suggestion
    const arrivalTimeInMs = sToMs(ISO8601Duration2sec(schedule.arrival));
    return Math.abs(arrivalTimeInMs - trainSummary.path_item_times_final[index + 1]) >= TOLERANCES;
```

front/src/applications/operationalStudies/utils.ts
```diff
@@ -209,3 +214,79 @@ export const convertDepartureTimeIntoSec = (departureTime: string) => {
...
+    throw new Error('Assertion failed');
+  }
+
+  const toCheckPathItemIds = new Set<string>(trainSchedule.margins?.boundaries);
+  trainSchedule.schedule?.forEach((schedule) => {
+    if (schedule.arrival !== null || schedule.stop_for !== null) {
+      toCheckPathItemIds.add(schedule.at);
+    }
+  });
+  toCheckPathItemIds.add(trainSchedule.path[trainSchedule.path.length - 1].id);
```
clarani, line 277: Why do you add the last step of the path ?

front/src/applications/operationalStudies/utils.ts
```diff
@@ -209,3 +214,79 @@ export const convertDepartureTimeIntoSec = (departureTime: string) => {
...
+  return pathItemMap;
+};
+
+/**
+ * Check if the train is too fast with a trainSchedule and a trainSummary
+ * @param trainResult
+ * @param trainSummary
+ * @returns true if the train is too fast
+ */
+export const isToofast = (
```
clarani:
```suggestion
export const isTooFast = (
```

front/src/applications/operationalStudies/utils.ts
```diff
@@ -209,3 +214,79 @@ export const convertDepartureTimeIntoSec = (departureTime: string) => {
...
+  trainSchedule: TrainScheduleResult,
+  trainSummary: Extract<SimulationSummaryResult, { status: 'success' }>
+): boolean => {
+  if (
+    trainSummary.path_item_times_final.length !== trainSummary.path_item_times_provisional.length
+  ) {
+    throw new Error('Assertion failed');
+  }
+
+  const toCheckPathItemIds = new Set<string>(trainSchedule.margins?.boundaries);
```
clarani: You can remove the string here, typescript knows that it will be a set of strings ;)
```suggestion
  const toCheckPathItemIds = new Set(trainSchedule.margins?.boundaries);
```

front/src/applications/operationalStudies/utils.ts
```diff
@@ -209,3 +214,79 @@ export const convertDepartureTimeIntoSec = (departureTime: string) => {
...
+export const pathItemMapping = (trainResult: TrainScheduleResult) => {
+  const pathItemMap: { [key: string]: number } = {};
+
+  trainResult.path.forEach((pathItem, index) => {
+    if (pathItemMap[pathItem.id]) throw new Error('Assertion failed');
+    pathItemMap[pathItem.id] = index;
+  });
+
+  return pathItemMap;
+};
```
clarani: I think you can improve this function:
- rename it to make it more explicit
- use a reduce, the code is clearer
- import the type Dictionary from lodash or utils/types (once you have rebased)
```suggestion
export const getPathItemByIndexDict = (trainResult: TrainScheduleResult) =>
  trainResult.path.reduce((acc, pathItem, index) => {
    acc[pathItem.id] = index;
    return acc;
  }, {} as Dictionary<number>);
```

front/src/applications/operationalStudies/utils.ts
```diff
@@ -209,3 +214,79 @@ export const convertDepartureTimeIntoSec = (departureTime: string) => {
...
+
+  // eslint-disable-next-line no-restricted-syntax
+  for (const pathItemId of toCheckPathItemIds) {
+    const pathItemIndex = pathItemMap[pathItemId];
+    const pathItemTimeFinal = trainSummary.path_item_times_final[pathItemIndex];
+    const pathItemTimeProvisional = trainSummary.path_item_times_provisional[pathItemIndex];
+
+    if (pathItemTimeProvisional > pathItemTimeFinal + TOLERANCES) return true;
+  }
+  return false;
```
clarani: You can use the function some instead of this loop, like in isScheduledPointsNotHonored ;)
```suggestion
  return [...toCheckPathItemIds].some((pathItemId) => {
    const pathItemIndex = pathItemMap[pathItemId];
    const pathItemTimeFinal = trainSummary.path_item_times_final[pathItemIndex];
    const pathItemTimeProvisional = trainSummary.path_item_times_provisional[pathItemIndex];

    if (pathItemTimeProvisional > pathItemTimeFinal + TOLERANCES) return true;
    return false;
  });
```

front/src/applications/operationalStudies/utils.ts
```diff
@@ -209,3 +214,79 @@ export const convertDepartureTimeIntoSec = (departureTime: string) => {
...
+
+  const toCheckPathItemIds = new Set<string>(trainSchedule.margins?.boundaries);
+  trainSchedule.schedule?.forEach((schedule) => {
+    if (schedule.arrival !== null || schedule.stop_for !== null) {
+      toCheckPathItemIds.add(schedule.at);
+    }
+  });
+  toCheckPathItemIds.add(trainSchedule.path[trainSchedule.path.length - 1].id);
+
+  if (!toCheckPathItemIds.size) return false;
```
clarani: Maybe you can be more explicit here 
```suggestion
  if (toCheckPathItemIds.size === 0) return false;
```

front/src/modules/trainschedule/components/TimetableV2/hooks.ts
```diff
@@ -88,15 +89,14 @@ const useTrainSchedulesDetails = (
...
-            if (
-              (scheduledPointsHonoredFilter === 'honored' && !isHonored) ||
-              (scheduledPointsHonoredFilter === 'notHonored' && isHonored)
-            ) {
-              return false;
-            }
-          } else {
+          if (trainSummary.status !== 'success') return false;
+
+          const isHonored = isScheduledPointsNotHonored(trainSchedule, trainSummary);
```
clarani: I think isNotHonored would be a better name ;)
```suggestion
          const isNotHonored = isScheduledPointsNotHonored(trainSchedule, trainSummary);
```

front/src/applications/operationalStudies/utils.ts
```diff
@@ -209,3 +214,79 @@ export const convertDepartureTimeIntoSec = (departureTime: string) => {
...
+ */
+export const isToofast = (
+  trainSchedule: TrainScheduleResult,
+  trainSummary: Extract<SimulationSummaryResult, { status: 'success' }>
+): boolean => {
+  if (
+    trainSummary.path_item_times_final.length !== trainSummary.path_item_times_provisional.length
+  ) {
+    throw new Error('Assertion failed');
+  }
```
clarani, line 269: Why do we have the same number of items in path_item_times_final and in path_item_times_provisional ?

front/src/applications/operationalStudies/__tests__/sampleData.ts
clarani, line 1: Is it real data for the pathStep ? We should avoid pushing real data on the github ;) If it is, you can replace it by data from small infra if you want

front/src/applications/operationalStudies/__tests__/sampleData.ts
```diff
@@ -344,3 +345,194 @@ export const electrificationRangesLarge: ElectrificationRangeV2[] = [
...
+  status: 'success',
+  length: 145084000,
+  time: 4084949,
+  energy_consumption: 3729835446.4172735,
+  path_item_times_final: [0, 2567248, 4084949],
+  path_item_times_provisional: [0, 2567248, 4084949],
+  path_item_times_base: [0, 2567248, 4084949],
+};
+
+export const trainScheduleHonoredAndNotTooFast: TrainScheduleResult = {
```
clarani: Maybe trainScheduleHonored may be enough for the name :)
```suggestion
export const trainScheduleHonored: TrainScheduleResult = {
```

clarani (changes requested):
Nice job :) I left some comments

front/src/modules/trainschedule/components/TimetableV2/hooks.ts
theocrsb: Done :)

front/src/applications/operationalStudies/utils.ts
flomonster: Let's say we set one margin for our train schedule. We want to check that the provisional time is not smaller than the final time. To perform this check we have to test the last item in the table.

front/src/applications/operationalStudies/utils.ts
flomonster: >  if (pathItemTimeProvisional > pathItemTimeFinal + TOLERANCES) return true;
> return false;

Bad pattern:

```
return (pathItemTimeProvisional > pathItemTimeFinal + TOLERANCES);
```

What is the behavior of using `return` in a lambda?

front/src/applications/operationalStudies/__tests__/sampleData.ts
theocrsb: Done

front/src/modules/trainschedule/components/TimetableV2/hooks.ts
theocrsb: Done

front/src/applications/operationalStudies/utils.ts
theocrsb: Done

front/src/applications/operationalStudies/utils.ts
theocrsb: Done

front/src/applications/operationalStudies/utils.ts
theocrsb: Done

front/src/applications/operationalStudies/utils.ts
theocrsb: Done

front/src/applications/operationalStudies/utils.ts
theocrsb: Done

front/src/applications/operationalStudies/consts.ts
theocrsb: Done

front/src/applications/operationalStudies/utils.ts
theocrsb: Done

front/src/applications/operationalStudies/utils.ts
flomonster: Do we have optimization when using an iterator instead of a for loop?

Nitpick: In my opinion the first version is more readable :shrug:

clarani: > Bad pattern:
> 
> ```
> return (pathItemTimeProvisional > pathItemTimeFinal + TOLERANCES);
> ```
You're right about this, I was considering the for loop and wondering whether .some() would be better or not, and I missed it :/

From what I have read, a for loop is as efficient as .some(). I usually prefer to use the available functions of JavaScript, but @emersion also thinks that it is more readable like this, so we can keep it like this 👍

front/src/applications/operationalStudies/utils.ts
theocrsb: These lists have the same number of elements as in 'path_items'. So they have the same number of elements

front/src/applications/operationalStudies/utils.ts
emersion: TBH the speed here likely doesn't matter at all, we shouldn't be micro-optimizing :)

theocrsb: Done

front/src/applications/operationalStudies/utils.ts
theocrsb: Done

front/src/applications/operationalStudies/__tests__/sampleData.ts
theocrsb: You're right, I just changed it

front/src/applications/operationalStudies/utils.ts
clarani: But it won't already be in the margin boundaries ?

front/src/modules/trainschedule/components/TimetableV2/hooks.ts
```diff
@@ -91,11 +91,11 @@ const useTrainSchedulesDetails = (
...
           if (trainSummary.status !== 'success') return false;
 
-          const isHonored = isScheduledPointsNotHonored(trainSchedule, trainSummary);
+          const isNotHonored = isScheduledPointsNotHonored(trainSchedule, trainSummary);
 
           if (
-            (scheduledPointsHonoredFilter === 'honored' && !isHonored) ||
-            (scheduledPointsHonoredFilter === 'notHonored' && isHonored)
+            (scheduledPointsHonoredFilter === 'honored' && !isNotHonored) ||
+            (scheduledPointsHonoredFilter === 'notHonored' && isNotHonored)
```
clarani: Small bug spotted while testing
```suggestion
            (scheduledPointsHonoredFilter === 'honored' && isNotHonored) ||
            (scheduledPointsHonoredFilter === 'notHonored' && !isNotHonored)
```

clarani (changes requested):
Almost here

clarani (commented):
It's not very readable, can you fix this ? 
image

front/src/modules/trainschedule/components/TimetableV2/hooks.ts
theocrsb: Done !

theocrsb:
> It's not very readable, can you fix this ? image

Done

clarani (approved):

front/src/applications/operationalStudies/utils.ts
clarani: You're right, my bad :+1:

front/src/applications/operationalStudies/utils.ts
```diff
@@ -209,3 +216,74 @@ export const convertDepartureTimeIntoSec = (departureTime: string) => {
...
+ * @param trainResult
+ * @param trainSummary
+ * @returns true if the scheduled points are not honored
+ */
+export const isScheduledPointsNotHonored = (
+  trainSchedule: TrainScheduleResult,
+  trainSummary: Extract<SimulationSummaryResult, { status: 'success' }>
+): boolean => {
+  if (trainSummary.path_item_times_final.length !== trainSchedule.path.length) {
+    console.error();
```
emersion: Is this `console.error()` call without a message intentional?

front/src/applications/operationalStudies/utils.ts
```diff
@@ -209,3 +216,74 @@ export const convertDepartureTimeIntoSec = (departureTime: string) => {
...
+    throw new Error('Assertion failed');
+  }
+  if (!trainSchedule.schedule) return false;
+  return trainSchedule.schedule.some((schedule, index) => {
+    if (schedule.arrival === null || schedule.arrival === undefined) return false;
```
emersion: Nit: `if (!schedule.arrival)` should work here (a zero duration is represented as `P0D`)

front/src/applications/operationalStudies/utils.ts
```diff
@@ -209,3 +216,74 @@ export const convertDepartureTimeIntoSec = (departureTime: string) => {
...
+): boolean => {
+  if (
+    trainSummary.path_item_times_final.length !== trainSummary.path_item_times_provisional.length
+  ) {
+    throw new Error('Assertion failed');
+  }
+
+  const toCheckPathItemIds = new Set(trainSchedule.margins?.boundaries);
+  trainSchedule.schedule?.forEach((schedule) => {
+    if (schedule.arrival !== null || schedule.stop_for !== null) {
```
emersion: The above case handles `arrival === undefined` as well - is this something that can happen?

Tguisnet (approved):
LGTM for the editoast part.
Tested, there are few bugs spotted (with the output array) but it doesn't concern this PR.
There is no filter for the no respect of margins, only for honored_scheduled_points, but we could keep it like this, until there is a real need.
Before merging this PR, you should fix the checks that failed in the CI

front/src/applications/operationalStudies/utils.ts
theocrsb: I don't think so but with this condition I don't need to add the cast l237
[https://github.com/OpenRailAssociation/osrd/pull/8142/commits/464feb66f78a8ea158b4632afebe8a4c5799183d#r1705449975](url)

front/src/applications/operationalStudies/utils.ts
theocrsb: Done

front/src/applications/operationalStudies/utils.ts
theocrsb: I don't think so, but when in doubt I modified it

emersion (approved):
LGTM, thanks a lot!

PR #8144, opened 2024-07-22
ci: add telemetry and push to Honeycomb on our CI

woshilapin:
# Goal

Let's set up some telemetry on our CI to evaluate the time to build and test. I'm also hoping to track down the time it takes to build `editoast` and find which solution works when we try to reduce the build time.

I'm aware that we don't make a full `editoast` build most of the time, because we're heavily relying on caching Docker layers. It's true, and I don't yet know how we'll deal with this, but the setup of this solution is easy enough to get some first numbers and iterate from there (maybe the `release` build will give us the hints we need?).

I'm also aware that we don't have detailed information about one step of the workflow. For example, it might be interesting to have more detailed information about what time it takes to compile each `crate`, how long it takes to the linker. Again, let's start somewhere. I'm happy to drop this solution if we end up with something better later.

# Notes about the PR

`workflow_run` even only works on the default branch. Therefore, I tested with the following configuration, just to evaluate the Github action, and the connection with HoneyComb.

```yaml
name: Export workflow to Honeycomb

on:
  push:
    branches:
      - "wsl/ci/ci-telemetry"

jobs:
  otel-export-honeycomb:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - name: Export workflow to Honeycomb
        uses: corentinmusard/otel-cicd-action@v1
        with:
          otlpEndpoint: grpc://api.honeycomb.io:443/
          # Example value for HONEYCOMB_OTLP_HEADERS:
          # x-honeycomb-team=YOUR_API_KEY,x-honeycomb-dataset=YOUR_DATASET
          otlpHeaders: ${{ secrets.HONEYCOMB_OTLP_HEADERS }}
          githubToken: ${{ secrets.GITHUB_TOKEN }}
          runId: ${{ github.run_id }}
```

It does work and we can see the following on HoneyComb.

image

Note that I don't care if we use HoneyComb or anything else. It just happened there was a HoneyComb example in the documentation of `corentinmusard/otel-cicd-action` and I already had a HoneyComb account so it was the fastest to set up and experiment for me.

Thank you @Khoyo for the help setting up `secrets.HONEYCOMB_OTLP_HEADERS`.

### DIFF
--- a/.github/workflows/telemetry.yml
+++ b/.github/workflows/telemetry.yml
@@ -1,0 +1,23 @@
+name: Export workflow to Honeycomb
+
+on:
+  workflow_run:
+    workflows:
+      # The name of the workflow(s) that triggers the export
+      - "build"
+    types: [completed]
+
+jobs:
+  otel-export-honeycomb:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Export workflow to Honeycomb
+        uses: corentinmusard/otel-cicd-action@v1
+        with:
+          otlpEndpoint: grpc://api.honeycomb.io:443/
+          # Example value for HONEYCOMB_OTLP_HEADERS:
+          # x-honeycomb-team=YOUR_API_KEY,x-honeycomb-dataset=YOUR_DATASET
+          otlpHeaders: ${{ secrets.HONEYCOMB_OTLP_HEADERS }}
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          runId: ${{ github.event.workflow_run.id }}


ghost (commented):
Hello ! That's an interesting idea. 

After checking the documentation for the `corentinmusard/otel-cicd-action`, I had the feeling it just pushes information about the timing of each steps so the gain here would be having them displayed in a more convenient and queryable way, right ? (since this information is already visible on the build details).

If that's required and helping to improve build times we should go for it.

woshilapin:
> Hello ! That's an interesting idea.
> 
> After checking the documentation for the `corentinmusard/otel-cicd-action`, I had the feeling it just pushes information about the timing of each steps so the gain here would be having them displayed in a more convenient and queryable way, right ? (since this information is already visible on the build details).
> 
> If that's required and helping to improve build times we should go for it.

Yes, I understand exactly the same thing. The information is already available, just pushing them in a system that can query them and provide diagram facilities should help produces helpful boards.

Note that I still have a problem with the solution I provide. I'd like to provide a board that is open to everyone (open-source so maximum tranparency), but I have the feeling that HoneyComb only provide boards available to those who are subscribed. If I don't find a solution for this, that would be a no-go for HoneyComb to me (and I'd be looking for another vendor). What do you think?

Wadjetz (approved):

flomonster:
> > Hello ! That's an interesting idea.
> > After checking the documentation for the `corentinmusard/otel-cicd-action`, I had the feeling it just pushes information about the timing of each steps so the gain here would be having them displayed in a more convenient and queryable way, right ? (since this information is already visible on the build details).
> > If that's required and helping to improve build times we should go for it.
> 
> Yes, I understand exactly the same thing. The information is already available, just pushing them in a system that can query them and provide diagram facilities should help produces helpful boards.
> 
> Note that I still have a problem with the solution I provide. I'd like to provide a board that is open to everyone (open-source so maximum tranparency), but I have the feeling that HoneyComb only provide boards available to those who are subscribed. If I don't find a solution for this, that would be a no-go for HoneyComb to me (and I'd be looking for another vendor). What do you think?

I agree, if the honeycomb charts cannot be read by everyone it should be a no-go. In addition, the tool seems to me to be overkill for our needs.

Khoyo:
>> I'd like to provide a board that is open to everyone (open-source so maximum tranparency), but I have the feeling that HoneyComb only provide boards available to those who are subscribed. If I don't find a solution for this, that would be a no-go for HoneyComb to me (and I'd be looking for another vendor). What do you think?

> I agree, if the honeycomb charts cannot be read by everyone it should be a no-go. In addition, the took seems to me to be overkill for our needs.

Looking at their [pricing model](https://www.honeycomb.io/pricing), it seems to me it allows you to add as many people as you want to your board, so while the board can't be publicly readable, we could add all interested collaborators. It's not ideal, but I don't think passersby not having access to CI build-timing graphs history is a complete no go.

(«Number of Users: Unlimited» in the detailed comparison, seems to me the main factor for the pricing is the number of monthly event and triggers)

woshilapin:
I'm closing this since I'm pretty sure the ROI won't bring anything useful.

PR #8145, opened 2024-07-22
core: fix and reintroduce pathfinding heuristic

### LINKED ISSUE #7200: core: pathfinding heuristic is too expensive to compute
When running a pathfinding request, we spend most of the execution time evaluating the heuristic. It actually gets slower than no heuristic at all. 

Switching from geodesic calculations to spherical / planar *may* fix the problem entirely. If it doesn't there can be other improvements to be made:

1. `blockToPoint` isn't actually needed, we can use points at the edge of a chunk and compute the distance difference. This may even be more accurate.
2. We compute the minimum distance to the next step, which can be made of several locations. We can try to merge them, maybe computing a circle that contains all of them.

Whatever change we make, it's important that the heuristic **never overestimates the remaining distance**, it can't be pessimistic. 

In the meantime, we can actually turn off the heuristic (moving back to a Dijkstra from an A*). It's still useful for STDCM though, considering the graph exploration is much more expensive

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/7200

Change summary:

1. Revert the removal of the heuristic
2. Use remaining time instead of remaining distance
3. Use faster maths to compute geographic distance
4. Speed up the "block location -> geo point" method (see comments)
5. Remove old geo computation + external lib dependency

When running the benchmark, average process time per train goes from 40s to 25s. (note: this includes *more* than just the pathfinding). No extra error. 

While tested I compared the old and new distance methods. The new one is less accurate, but it's not off by more than 0.1%.

### DIFF
--- a/core/osrd-geom/src/main/java/fr/sncf/osrd/geom/Point.java
+++ b/core/osrd-geom/src/main/java/fr/sncf/osrd/geom/Point.java
@@ -1,6 +1,6 @@
 package fr.sncf.osrd.geom;
 
-import org.gavaghan.geodesy.*;
+import static java.lang.Math.*;
 
 public record Point(
         // Longitude
@@ -10,15 +10,17 @@ public record Point(
 
     /**
      * Returns the distance between this point and another in meters, assuming x = longitude and y =
-     * latitude
+     * latitude. Uses equirectangular distance approximation (very fast but not 100% accurate)
      */
     public double distanceAsMeters(Point other) {
-        GeodeticCalculator geoCalc = new GeodeticCalculator();
-        Ellipsoid reference = Ellipsoid.WGS84;
-        GlobalPosition thisPosition = new GlobalPosition(y, x, 0.0);
-        GlobalPosition otherPosition = new GlobalPosition(other.y, other.x, 0.0);
-        return geoCalc.calculateGeodeticCurve(reference, thisPosition, otherPosition)
-                .getEllipsoidalDistance();
+        final var earthRadius = 6_378_160;
+        var lon1 = toRadians(x);
+        var lon2 = toRadians(other.x);
+        var lat1 = toRadians(y);
+        var lat2 = toRadians(other.y);
+        var xDiff = (lon1 - lon2) * cos(0.5 * (lat1 + lat2));
+        var yDiff = lat1 - lat2;
+        return earthRadius * sqrt(xDiff * xDiff + yDiff * yDiff);
     }
 
     @Override
--- a/core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
@@ -22,6 +22,7 @@ import fr.sncf.osrd.sim_infra.utils.getNextTrackSections
 import fr.sncf.osrd.toDirection
 import fr.sncf.osrd.train.RollingStock
 import fr.sncf.osrd.utils.CachedBlockMRSPBuilder
+import fr.sncf.osrd.utils.CachedBlockMRSPBuilder.Companion.DEFAULT_MAX_ROLLING_STOCK_SPEED
 import fr.sncf.osrd.utils.indexing.*
 import fr.sncf.osrd.utils.units.Distance.Companion.fromMeters
 import fr.sncf.osrd.utils.units.Offset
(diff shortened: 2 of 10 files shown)


core/osrd-geom/src/main/java/fr/sncf/osrd/geom/Point.java
```diff
@@ -10,15 +10,17 @@ public record Point(
...
+     * latitude. Uses equirectangular distance approximation (very fast but not 100% accurate)
      */
     public double distanceAsMeters(Point other) {
-        GeodeticCalculator geoCalc = new GeodeticCalculator();
-        Ellipsoid reference = Ellipsoid.WGS84;
-        GlobalPosition thisPosition = new GlobalPosition(y, x, 0.0);
-        GlobalPosition otherPosition = new GlobalPosition(other.y, other.x, 0.0);
-        return geoCalc.calculateGeodeticCurve(reference, thisPosition, otherPosition)
-                .getEllipsoidalDistance();
+        var earthRadius = 6_378_160;
```
Erashin: I feel like this should set off a `warning: magic number`. Put it into a final static value?

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
```diff
@@ -224,16 +224,55 @@ fun runPathfinding(
     }
     val constraints = initConstraints(infra, rollingStocks!!)
 
-    // The heuristic is temporarily disabled, its implementation
-    // is currently too slow and the requests are actually slower.
-    // See https://github.com/OpenRailAssociation/osrd/issues/7200
-    val nullHeuristic: AStarHeuristicId<Block> = AStarHeuristic() { _, _ -> 0.0 }
-    val remainingDistanceEstimators = List(waypoints.size - 1) { nullHeuristic }
+    val remainingDistanceEstimators = makeHeuristics(infra, waypoints, 320.0 / 3.6)
```
Erashin: A bit off putting to see `320.0/3.6` coming out of nowhere. Make it into a variable? final static constant if you feel like it might be used elsewhere, if not at the minimum put it into a value just before method call?

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/RemainingDistanceEstimator.kt
```diff
@@ -0,0 +1,105 @@
...
+    private val remainingDistance: Distance
+) {
+    private val targets: MutableCollection<Point> = ArrayList()
+
+    init {
+        for (edgeLocation in edgeLocations) {
+            val point =
+                blockOffsetToPoint(blockInfra, rawInfra, edgeLocation.edge, edgeLocation.offset)
+            // Avoid adding duplicate geo points to the target list
+            val exists = targets.any { point.distanceAsMeters(it) < DISTANCE_THRESHOLD }
```
Erashin:
```suggestion
            val isDuplicate = targets.any { point.distanceAsMeters(it) < DISTANCE_THRESHOLD }
```

Erashin (approved):
LGTM. For other optional reviewers, I suggest comparing this commit with https://github.com/OpenRailAssociation/osrd/commit/1122145ef03fc15aac3516d009bf1a8ff13f98f3, where the remaining distance classes were previously removed.

core/osrd-geom/src/main/java/fr/sncf/osrd/geom/Point.java
eckter: We don't really have a magic number warning

I can make it final but not static without moving it out of the method (which would be weird, it's an "internal" aspect). I don't think this is a wrong use of magic number anyway, we do need the info and I don't expect it to change soon

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
eckter: Ideally we'd add a parameter to the pathfinding requests, I'll open an issue and add a comment

eckter: https://github.com/OpenRailAssociation/osrd/issues/8154

Comment added in the code

core/osrd-geom/src/main/java/fr/sncf/osrd/geom/Point.java
```diff
@@ -10,15 +10,17 @@ public record Point(
 
     /**
      * Returns the distance between this point and another in meters, assuming x = longitude and y =
-     * latitude
+     * latitude. Uses equirectangular distance approximation (very fast but not 100% accurate)
```
Khoyo, line 13: Do you know if the inaccuracies keep the heuristic admissible? Or are they small enough that they don't matter?

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
Khoyo: I'd really like it to become a (global) constant in the meantime since its used in multiple places

core/osrd-geom/src/main/java/fr/sncf/osrd/geom/Point.java
eckter: No, the heuristic isn't admissible anymore. But I believe that it's small enough to not be an issue. Though if we really want to make it admissible again, we could scale the value down by like 1%.

But since we now look for the fastest path instead of shortest, there's a fair amount of approximations anyway. I don't think it would make sense to be too rigid on finding the single "best possible path"

Khoyo (approved):
LGTM

PR #8152, opened 2024-07-23
core: prioritize paths closer to expected passage times

### LINKED ISSUE #7689: core: stdcm: update node comparaison functions to prioritize paths closer to the expected passage times
If the departure time can be modified, we use the best value of the interval. We also need to account for different tolerance (e.g. arrival window from 10:00 to 11:00 with best time at 10:15, then 10:25 is better than 10:05).

Following the expected departure time should be mostly straightforward: we can keep track of the ideal departure time delay. Same for intermediate points.

Following the expected arrival time at the last point is a little trickier, we only know the ideal departure time delay as we reach the destination. We shouldn't return "path found" just as it's explored.

When several points have been set, the expected behavior becomes ambiguous. I'd suggest that we don't actively think about this case and handle it in the way that's the most convenient. The front-end doesn't even let the user input that.

Erashin:
Fixes #7689.

Highest prio paths remain paths wth the minimum total run time. Afterwards, prioritize paths closer to expected passage times.

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMEdge.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMEdge.kt
@@ -54,6 +54,7 @@ data class STDCMEdge(
     /** Returns the node at the end of this edge */
     fun getEdgeEnd(graph: STDCMGraph): STDCMNode {
         var newWaypointIndex = waypointIndex
+        val previousPlannedNodeRelativeTimeDiff = getPreviousPlannedNodeRelativeTimeDiff()
         while (newWaypointIndex + 1 < graph.steps.size) {
             val nextStep = graph.steps[newWaypointIndex + 1]
             val endOffset = envelopeStartOffset + length.distance
@@ -78,12 +79,15 @@ data class STDCMEdge(
                 newWaypointIndex,
                 null,
                 null,
+                graph.steps[newWaypointIndex].plannedTimingData,
+                previousPlannedNodeRelativeTimeDiff,
                 timeSinceDeparture,
                 graph.remainingTimeEstimator.invoke(this, null, newWaypointIndex),
             )
         } else {
             // New edge on the same block, after a stop
-            val stopDuration = graph.getFirstStopAfterIndex(waypointIndex)!!.duration!!
+            val firstStopAfterIndex = graph.getFirstStopAfterIndex(waypointIndex)!!
+            val stopDuration = firstStopAfterIndex.duration!!
             val locationOnEdge = envelopeStartOffset + length.distance
             STDCMNode(
                 totalTime + timeStart + stopDuration,
@@ -95,9 +99,26 @@ data class STDCMEdge(
                 newWaypointIndex,
                 envelopeStartOffset + length.distance,
                 stopDuration,
+                firstStopAfterIndex.plannedTimingData,
+                previousPlannedNodeRelativeTimeDiff,
                 graph.remainingTimeEstimator.invoke(this, locationOnEdge, newWaypointIndex)
             )
         }
+    }
+
+    /**
+     * Computes the last planned node's previousPlannedNodeRelativeTimeDiff, taking the potentially
+     * new total departure time shift into account.
+     */
+    private fun getPreviousPlannedNodeRelativeTimeDiff(): Double? {
+        var previousPlannedNode = this.previousNode
+        while (previousPlannedNode != null) {
+            if (previousPlannedNode.plannedTimingData != null) {
+                return previousPlannedNode.getRelativeTimeDiff(totalDepartureTimeShift)
+            }
+            previousPlannedNode = previousPlannedNode.previousEdge.previousNode
+        }
+        return null
     }
 
     /**
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMEdge.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMEdge.kt
@@ -27,8 +27,8 @@ data class STDCMEdge(
     // Total delay we have added by shifting the departure time since the start of the
     // path
     val totalDepartureTimeShift: Double,
-    // Node located at the start of this edge, null if this is the first edge
-    val previousNode: STDCMNode?,
+    // Node located at the start of this edge
+    val previousNode: STDCMNode,
     // Offset of the envelope if it doesn't start at the beginning of the edge
     val envelopeStartOffset: Offset<Block>,
     // Index of the last waypoint passed by this train
@@ -79,7 +79,7 @@ data class STDCMEdge(
                 newWaypointIndex,
                 null,
                 null,
-                graph.steps[newWaypointIndex].plannedTimingData,
+                null,
                 previousPlannedNodeRelativeTimeDiff,
                 timeSinceDeparture,
                 graph.remainingTimeEstimator.invoke(this, null, newWaypointIndex),
@@ -111,12 +111,13 @@ data class STDCMEdge(
      * new total departure time shift into account.
      */
     private fun getPreviousPlannedNodeRelativeTimeDiff(): Double? {
-        var previousPlannedNode = this.previousNode
-        while (previousPlannedNode != null) {
+        var currentEdge: STDCMEdge? = this
+        while (currentEdge != null) {
+            val previousPlannedNode = currentEdge.previousNode
             if (previousPlannedNode.plannedTimingData != null) {
                 return previousPlannedNode.getRelativeTimeDiff(totalDepartureTimeShift)
             }
-            previousPlannedNode = previousPlannedNode.previousEdge.previousNode
+            currentEdge = previousPlannedNode.previousEdge
         }
         return null
     }
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMEdgeBuilder.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMEdgeBuilder.kt
@@ -21,6 +21,8 @@ internal constructor(
     private val infraExplorer: InfraExplorerWithEnvelope,
     /** STDCM Graph, needed for most operations */
     private val graph: STDCMGraph,
+    /** Previous node, used to compute the final path */
+    private var prevNode: STDCMNode,
     /** Start time of the edge */
     private var startTime: Double = 0.0,
 
@@ -40,9 +42,6 @@ internal constructor(
      * Sum of all the delay that has been added in the previous edges by shifting the departure time
      */
     private var prevAddedDelay: Double = 0.0,
-
-    /** Previous node, used to compute the final path */
-    private var prevNode: STDCMNode? = null,
 
     /** Envelope to use on the edge, if unspecified we try to go at maximum allowed speed */
     private var envelope: Envelope? = null,
@@ -89,12 +88,6 @@ internal constructor(
      */
     fun setPrevAddedDelay(prevAddedDelay: Double): STDCMEdgeBuilder {
         this.prevAddedDelay = prevAddedDelay
-        return this
-    }
-
-    /** Sets the previous node, used to compute the final path */
-    fun setPrevNode(prevNode: STDCMNode?): STDCMEdgeBuilder {
-        this.prevNode = prevNode
         return this
     }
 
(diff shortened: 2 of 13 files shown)


core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
```diff
@@ -32,10 +38,13 @@ data class STDCMNode(
...
     /**
-     * Defines the estimated better path between 2 nodes, in terms of total run time, then departure
-     * time, then number of reached targets. If the result is negative, the current node has a
-     * better path, and should be explored first. This method allows us to order the nodes in a
-     * priority queue, from the best path to the worst path. We then explore them in that order.
+     * Defines the estimated better path between 2 nodes, in the following priority
+     * - lowest total run time
+     * - earliest departure time
+     * - closest planned arrival time, using the current then the last planned timing data
+     * - highest number of reached targets If the result is negative, the current node has a better
```
Khoyo: Missing a line break there ;)

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
```diff
@@ -44,6 +53,25 @@ data class STDCMNode(
...
         return if (abs(runTimeEstimation - otherRunTimeEstimation) >= 1e-3)
             runTimeEstimation.compareTo(otherRunTimeEstimation)
+        // Minimise the difference with the planned arrival times
+        else if (
+            plannedTimingData != null &&
+                other.plannedTimingData != null &&
+                abs(
+                    (plannedTimingData.arrivalTime.seconds - time) -
+                        (other.plannedTimingData.arrivalTime.seconds - other.time)
+                ) > 1e-3
```
Khoyo: That epsilon should probably become a constant (or you could create a comparison helper function, it might make those comparison more readable)

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
Erashin: Core format keeps putting it back it seems x) I agree though.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
Erashin: I think I'll create a helper function. I don't want people to confuse this with the `TIME_EPSILON = 1e-2` in `TrainPhysicsIntegrator.java`, as we really want to have a precision of 1e-3 here in my opinion.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
```diff
@@ -59,10 +59,10 @@ data class STDCMNode(
...
             plannedTimingData != null &&
                 other.plannedTimingData != null &&
-                abs(
-                    (plannedTimingData.arrivalTime.seconds - time) -
-                        (other.plannedTimingData.arrivalTime.seconds - other.time)
-                ) > 1e-3
+                areTimesDifferent(
+                    plannedTimingData.arrivalTime.seconds - time,
+                    other.plannedTimingData.arrivalTime.seconds - other.time
+                )
```
Khoyo: You should use it in the first if too ;)

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
```diff
@@ -61,4 +91,8 @@ data class STDCMNode(
             waypointIndex
         )
     }
+
+    private fun areTimesDifferent(time1: Double, time2: Double): Boolean {
+        return abs(time1 - time2) > 1e-3
+    }
```
eckter: There's already `TrainPhysicsIntegrator.areTimesEqual`. Though the epsilon is different there (1e-2).

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
```diff
@@ -32,18 +38,42 @@ data class STDCMNode(
...
 
     /**
-     * Defines the estimated better path between 2 nodes, in terms of total run time, then departure
-     * time, then number of reached targets. If the result is negative, the current node has a
-     * better path, and should be explored first. This method allows us to order the nodes in a
-     * priority queue, from the best path to the worst path. We then explore them in that order.
+     * Defines the estimated better path between 2 nodes, in the following priority
+     * - lowest total run time
+     * - earliest departure time
+     * - closest planned arrival time, using the current then the last planned timing data
```
eckter: This should actually be relative to the tolerance on either side. If there a 10 minutes tolerance before and 30 minutes tolerance after, then being late by 2 minutes is better than being early by 1 minute.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
```diff
@@ -32,18 +38,42 @@ data class STDCMNode(
...
+            (plannedTimingData.arrivalTime.seconds - time).compareTo(
+                other.plannedTimingData.arrivalTime.seconds - other.time
+            )
+        // If not, minimise the difference with the planned arrival times at the last planned node
+        else if (
+            prevPlannedNodeTimeDiff != null &&
+                other.prevPlannedNodeTimeDiff != null &&
+                areTimesDifferent(prevPlannedNodeTimeDiff, other.prevPlannedNodeTimeDiff)
+        )
+            prevPlannedNodeTimeDiff.compareTo(other.prevPlannedNodeTimeDiff)
```
eckter: This block could be reorganized with extra variables and possibly extra methods. Complex `if` like that are difficult to read (especially with the reformating).

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMEdge.kt
```diff
@@ -95,11 +99,27 @@ data class STDCMEdge(
...
+            if (prevPlannedNode.plannedTimingData != null) {
+                return prevPlannedNode.plannedTimingData!!.arrivalTime.seconds -
+                    prevPlannedNode.time
+            } else if (prevPlannedNode.prevPlannedNodeTimeDiff != null) {
+                return prevPlannedNode.prevPlannedNodeTimeDiff
+            }
+            prevPlannedNode = prevPlannedNode.previousEdge.previousNode
+        }
+        return null
+    }
```
eckter: I don't think this is right. We need to account for the departure time shift to know the arrival time at that location

eckter (commented):
There's a couple extra difficulties but for the most part it looks good.

It could also use a few tests

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
Erashin: I've actually been thinking about it during the afternoon, after @Khoyo 's comment. Since we want to fill the priority queue in the correct order, it made sense to me to compare the times strictly and use 1e-3 (i.e. precision to the ms). But that level of precision seems somewhat over kill (especially since we now have a lot of levels of conditional priority). I'm down to just use the standard `areTimesEqual` with a precision of 1e-2. Are you ok with that ?

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
Erashin: Good catch. It also means I'll have to load a different value for the previous planned node, or just additional values.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMEdge.kt
Erashin: Yeah, we have to take into account the additional departure time shift added since that node. The previousPlanned node's actual arrival time should be equal to :
`prevPlannedNode.time - (currentNode.totalPrevAddedDelay - prevPlannedNode.totalPrevAddedDelay)`

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
```diff
@@ -61,4 +91,44 @@ data class STDCMNode(
...
+        return null
+    }
+
+    private fun getLeftOrRightRelativeTimeDiff(timeDiff: Double): Double {
+        return when {
+            timeDiff < 0 -> abs(timeDiff / plannedTimingData!!.arrivalTimeToleranceBefore.seconds)
+            timeDiff > 0 -> timeDiff / plannedTimingData!!.arrivalTimeToleranceAfter.seconds
+            else -> timeDiff
+        }
+    }
```
eckter: This could be a method of `PlannedTimingData`, it would avoid the awkward `!!`

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMEdgeBuilder.kt
```diff
@@ -294,13 +287,13 @@ internal constructor(
...
+        while (node.previousEdge != null) {
             if (
-                !node.previousEdge.endAtStop &&
-                    node.previousEdge.block == infraExplorer.getCurrentBlock()
+                !node.previousEdge!!.endAtStop &&
+                    node.previousEdge!!.block == infraExplorer.getCurrentBlock()
             )
                 return true
-            node = node.previousEdge.previousNode
+            node = node.previousEdge!!.previousNode
```
eckter: The combination of a null condition followed by `!!` isn't ideal

Maybe we could change the while to `while (true)`, followed by a `val prevEdge = node.previousEdge ?: return false` ? That removes the need for `!!`

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
```diff
@@ -16,14 +19,20 @@ data class STDCMNode(
...
+    // Edge that lead to this node, null if this is the first node
+    val previousEdge: STDCMEdge?,
     // Index of the last waypoint passed by the train
     val waypointIndex: Int,
     // Position on a block, if this node isn't on the transition between blocks (stop)
     val locationOnEdge: Offset<Block>?,
     // When the node is a stop, how long the train remains here
     val stopDuration: Double?,
+    // Planned arrival time at node, with its tolerance. Is null at least when the node is not a
+    // stop
```
eckter, line 32: Maybe we could check when parsing the request that there's no scheduled point without stop?

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
eckter: reopening: it's not just about the code, the comment should be updated as well

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
```diff
@@ -61,4 +91,44 @@ data class STDCMNode(
...
+            timeDiff > 0 -> timeDiff / plannedTimingData!!.arrivalTimeToleranceAfter.seconds
+            else -> timeDiff
+        }
+    }
+
+    /**
+     * Takes into account the real current departure time shift to return the real current time at
+     * which the train arrives at this node.
+     */
+    private fun getRealTime(currentTotalAddedDelay: Double): Double {
```
eckter, line 125: We could add a comment on `node.time` to explain that it's not *necessarily* the actual time, and that this method should be used to get it

eckter (commented):
Looks pretty good, mostly minor comments

I haven't checked the tests yet (second part of the review will arrive soon) (edit: actually tests just look good)

Oh and there's formatting errors

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
Erashin: Already done in https://github.com/OpenRailAssociation/osrd/pull/8152/commits/f5480be912dd971e2189ef7ea7a2c1989850d9f0 ;) Just wanted to make sure it was also noted down in the actual class.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
Erashin: Done in https://github.com/OpenRailAssociation/osrd/pull/8152/commits/d73732b06b00a061c175bb5d2a0db657753bf346.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
Erashin: Done in https://github.com/OpenRailAssociation/osrd/pull/8152/commits/d73732b06b00a061c175bb5d2a0db657753bf346.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMEdgeBuilder.kt
Erashin: Done in https://github.com/OpenRailAssociation/osrd/pull/8152/commits/1ccc85c71cb905902f407c757a3ddb24a15fee5a.

eckter (approved):

PR #8153, opened 2024-07-23
core: stdcm: add opentelemetry spans for large sections

eckter:
Spans are cool

I'm not 100% sure of how I'm doing it, is there something else to do with datadog? I only tested the opentelemetry setup

---

image

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/STDCMHeuristic.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/STDCMHeuristic.kt
@@ -1,5 +1,6 @@
 package fr.sncf.osrd.stdcm
 
+import datadog.trace.api.Trace
 import fr.sncf.osrd.envelope_sim.PhysicsRollingStock
 import fr.sncf.osrd.sim_infra.api.Block
 import fr.sncf.osrd.sim_infra.api.BlockId
@@ -11,6 +12,8 @@ import fr.sncf.osrd.utils.CachedBlockMRSPBuilder
 import fr.sncf.osrd.utils.indexing.StaticIdx
 import fr.sncf.osrd.utils.units.Offset
 import fr.sncf.osrd.utils.units.meters
+import io.opentelemetry.api.trace.SpanKind
+import io.opentelemetry.instrumentation.annotations.WithSpan
 import java.util.*
 import kotlin.math.max
 import kotlin.math.min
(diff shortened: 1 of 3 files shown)


woshilapin:
Yes, sadly, we need to do all the job twice. So each `@WithSpan` should also have a `@Trace`.

Note that you can manually add some more metadata or modify the name of the span. And by default, all arguments to the function that can be displayed (don't remember what's the standard interface/function in Java), will be added as metadata to the span if I recall correctly.

```kotlin
import datadog.trace.api.Trace
import io.opentracing.util.GlobalTracer

@Trace
fun do_something(arg1: String) {
  val span = GlobalTracer.get().activeSpan()
  span.setOperationName("the name of the span") // by default, it will be 'do_something'
  span.setTag("some tag", someValue)
}
```

The equivalent for OpenTelemetry also exists
```kotlin
import io.opentelemetry.api.trace.Span

@WithSpan(value = "the name of the span") // setting `value` here is equivalent to `.setOperationName` in Datadog
fun do_something(arg1: String) {
  val span = Span.current()
  val.updateName("the name of the span") // if you need something more dynamic to name the span, same as `.setOperationName`
  val.setAttribute("some tag", someValue)
}
```

Oh, and if you don't remember all of this in the future, it's all in [`TkDatadog.kt`](https://github.com/OpenRailAssociation/osrd/blob/8a11ce2f82dc45a04884c40a721f461a3a3d36ae/core/src/main/java/fr/sncf/osrd/cli/TkDataDog.kt) and [`TkOpenTelemetry.kt`](https://github.com/OpenRailAssociation/osrd/blob/8a11ce2f82dc45a04884c40a721f461a3a3d36ae/core/src/main/java/fr/sncf/osrd/cli/TkOpenTelemetry.kt) (which I realize are stored inside a `java/` folder and should probably be in a `kotlin/` one 😉).

eckter:
I've added traces for each span. But I'm not quite sure how to test it. Is there a local setup similar to `jaeger`?

woshilapin:
> I've added traces for each span. But I'm not quite sure how to test it. Is there a local setup similar to `jaeger`?

There is no easy local setup. For local testing, you would need a free account on Datadog, deploying a local datadog agent and configuring all of this.

Another possibility is to deploy a branch on our `dev` environment. I can help testing that if you want.

UPDATE: I was able to test that it works well in Datadog.

woshilapin (approved):
Nice to see that kind of PR 😍

Khoyo (approved):

Issue #7977, opened 2024-07-08
etl-basic (+core): switch to the use of short ids for speedLimitTags


bougue-pe:
In order to allow for clearer and shorter interface:
- [ ] Imports from BASIC (simulations) + GAIA (infra) to OSRD must now use short ids of speedLimitTags instead of the current long names.
  - [x] PR opened on dags: https://github.com/osrd-project/osrd-dags/pull/417
  - [x] [etl-basic's PR](https://gitlab-repo-res.apps.eul.sncf.fr/dsir/groupedxs-dsir/04735/etl_basic/-/merge_requests/22) introducing speed-limit tags is reviewed to use short ids
- [x] Create, test and share a new PG dump with this data.
- [ ] Clean OSRD's code: https://github.com/OpenRailAssociation/osrd/pull/8417
  - [x] Cleanup bloated core code from #7903 (currently allows both short id or long name).
  - [x] Remove long names from static config-file
  - [x] Maybe remove default-speed too if unused is most likely in the future

Part of https://github.com/osrd-project/osrd-confidential/issues/364

PR #8159, opened 2024-07-23
core: stop using default-speed when handling speed-limit tags

bougue-pe:
Use only given speed-limit tag + fallback tags.

Adapt tests accordingly

As stated in https://github.com/osrd-project/osrd-confidential/issues/364#issuecomment-2236185634:
The use of given tag + fallback-tags but without (super low) default-speed has the best chances to reach a consensus among different use-cases (circulation mass-imports, STDCM, operational studies).
The use of default-speed creates overly-hard constraints on MRSP when infrastructure data is not perfect (which is the case most of the time) and the issues in infrastructure data should still be visible once showing a warning that a range of the path is not covered by a speed-limit tag in speed-space chart.

To be decided later:

- [x] cleanup default-speed values in config file if no use is planned (in https://github.com/OpenRailAssociation/osrd/issues/7977)
- [x] offer different MRSP policies for speed-limit tag handling if we need to differentiate use-cases (place to track this is undecided) : tracked into a separate issue : https://github.com/osrd-project/osrd-confidential/issues/606

### DIFF
--- a/core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/RawInfraImpl.kt
+++ b/core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/RawInfraImpl.kt
@@ -562,19 +562,7 @@ class RawInfraImpl(
             // TODO: as stated for SpeedSection class, a refactor is required to respect
             //       specifications
             assert(infraTagSpeed == null || speedFromRoute == null) { "checked at parsing" }
-            val allowedSpeed =
-                if (speedFromRoute != null) {
-                    speedFromRoute
-                } else if (infraTagSpeed != null) {
-                    infraTagSpeed
-                } else if (
-                    trainSpeedLimitTagDescriptor?.defaultSpeed != null &&
-                        trainSpeedLimitTagDescriptor.defaultSpeed < speedSection.default
-                ) {
-                    trainSpeedLimitTagDescriptor.defaultSpeed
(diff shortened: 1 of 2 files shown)


core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/RawInfraImpl.kt
```diff
@@ -621,11 +567,6 @@ class RawInfraImpl(
                     speedFromRoute
                 } else if (infraTagSpeed != null) {
```
eckter: The whole `if` could now be changed to `val allowedSpeed = speedFromRoute ?: (infraTagSpeed ?: speedSection.default)`

It's not necessarily *better*, but it's an option

eckter (approved):
LGTM

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/RawInfraImpl.kt
bougue-pe: true, I didn't like it with the default-speed logic, but now it looks better.
Amended the commit: https://github.com/OpenRailAssociation/osrd/compare/f60a9905e7491cdb9c92060bc0a94d65d81c472f..ae39abaf5205346613c6c1b7f58b801b33b480da

Erashin (approved):

PR #8172, opened 2024-07-25
Fix disabled add train button

### LINKED ISSUE #8170: Can't add train anymore
### What happened?

In a scenario, when trying to add a train, the button to add the train is always disabled even when a rollingstock and a path are selected

image

I think It probably comes from this [PR](https://github.com/OpenRailAssociation/osrd/pull/8006) : `infraState` is not provided to `<TimetableManegeTrainSchedule/>` anymore : 

image

We should be able to add a train

Try to add a train in a scenario

Dev (DgexSol)

Firefox

dev 4070d78

achrafmohye:
Closes #8170


Uriel-Sautron (approved):
LGTM !! Tested

flomonster (approved):
LGTM

SharglutDev (approved):
Lgtm, thank you

PR #8181, opened 2024-07-25
Fix naming table

flomonster:
## Description

- [x] Rename column titles
- [ ] Adapt variable names
- [ ] Adapt row size


flomonster:
Abort this change

PR #8183, opened 2024-07-25
replace switch with track node

younesschrifi:
Close https://github.com/osrd-project/osrd-confidential/issues/231


younesschrifi:
Will be done in another PR.

PR #8184, opened 2024-07-25
front: send track offset in mm in stdcm request

### LINKED ISSUE #7646: front: stdcm: v2: track offsets are in m instead of mm
**This is the same bug as the one fixed in https://github.com/OpenRailAssociation/osrd/pull/7596, but in a different context**. This time I couldn't quite identify where the steps are initialized.

I believe we really should double-check that stuff in general. 

---

### What happened?

When the front-end sends an STDCM request (v2), the path offsets are in meters instead of millimeters. The resulting path is not aligned with origin / destination.

Offsets should be in mm, the displayed path should align with origin / destination

Run an stdcm request, place origin or destination at the end of a track. The path will start/end near the start of the track.

- **Env**: Local

8392e7bcdaeda9ef8f998304ff9a43ad9453d28c

### LINKED ISSUE #7955: TS v2, adding  points by clicking on the map creates bugged time stops
#### Add 1 starting point near Nézel and 1 end point near Épône by clicking on the map:

Screenshot_20240704_220730

#### 2 phantom vias are created when looking in the time stops input table

Screenshot_20240704_220755

#### After the train is created, the path shown the simulation result on a completely different point than the select starting point
The train goes from Beynes to Épône, even though we started at Nézel.

download

- Should just create the starting and ending point, no extra vias
- the starting point on simulation result should be the same as the one selected by the user

—

- **Browser**: chrome 125
- **OS**: debian
- **Env**: Local

e69ecb131cd1f909e4d42a66166da615fe228252

### LINKED ISSUE #8026: TSv2: Pathfinding from map seems to snap my origin/destination points to OP
### What happened?

I asked for a path somewhere random on the map, but the output seems to have snapped my origin and destination to the nearest operational point or something:
image

The output displays the new points, not the one that i selected;
image

Also the info in the train details aren't correct:
image

I'm expecting my path to remain the same after computation

1. Find any path that doesn't start nor end at an OP, using the map

- **Browser**: Chrome
- **OS**: W11
- **Env**: rec

de8ada8

clarani:
closes #7646
closes #7955
closes #8026 

When adding/updating a train or launching a stdcm request, the track offsets were sent in meters instead of mm.

### DIFF
--- a/front/src/applications/stdcm/utils/formatStdcmConfV2.ts
+++ b/front/src/applications/stdcm/utils/formatStdcmConfV2.ts
@@ -12,6 +12,7 @@ import type { InfraState } from 'reducers/infra';
 import { setFailure } from 'reducers/main';
 import type { OsrdStdcmConfState, StandardAllowance } from 'reducers/osrdconf/types';
 import { dateTimeToIso } from 'utils/date';
+import { mToMm } from 'utils/physics';
 import { ISO8601Duration2sec, sec2ms, time2sec } from 'utils/timeManipulation';
 
 import createMargin from './createMargin';
@@ -128,6 +129,7 @@ export const checkStdcmConf = (
       const {
         id,
         arrival,
+        deleted,
         locked,
         stopFor,
         positionOnPath,
@@ -141,10 +143,19 @@ export const checkStdcmConf = (
         ...stepLocation
       } = step;
 
-      const secondary_code = 'trigram' in stepLocation || 'uic' in stepLocation ? ch : undefined;
+      const duration = stopFor ? sec2ms(ISO8601Duration2sec(stopFor) || Number(stopFor)) : 0;
 
+      if ('track' in stepLocation) {
+        return {
+          duration,
+          location: { track: stepLocation.track, offset: mToMm(stepLocation.offset) },
+        };
+      }
+
+      const secondary_code = 'trigram' in stepLocation || 'uic' in stepLocation ? ch : undefined;
       return {
-        duration: stopFor ? sec2ms(ISO8601Duration2sec(stopFor) || Number(stopFor)) : 0,
+        duration,
+        // TODO DROP V1: we should store the offset in mm in the store
         location: { ...stepLocation, secondary_code },
       };
     }),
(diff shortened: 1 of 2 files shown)


front/src/applications/stdcm/utils/formatStdcmConfV2.ts
```diff
@@ -15,6 +15,7 @@ import { dateTimeToIso } from 'utils/date';
 import { ISO8601Duration2sec, sec2ms, time2sec } from 'utils/timeManipulation';
 
 import createMargin from './createMargin';
+import { mToMm } from 'utils/physics';
```
theocrsb: It seems to me that we need to move imports to solve the prettier problem.

`
import { mToMm } from 'utils/physics';
import { ISO8601Duration2sec, sec2ms, time2sec } from 'utils/timeManipulation';

import createMargin from './createMargin';
`

theocrsb (approved):
LGTM and tested

front/src/applications/stdcm/utils/formatStdcmConfV2.ts
clarani: You're right, thanks :+1:

clarani: Done !

front/src/applications/stdcm/utils/formatStdcmConfV2.ts
```diff
@@ -141,10 +143,19 @@ export const checkStdcmConf = (
...
+          duration,
+          location: { track: stepLocation.track, offset: mToMm(stepLocation.offset) },
+        };
+      }
 
+      const secondary_code = 'trigram' in stepLocation || 'uic' in stepLocation ? ch : undefined;
       return {
-        duration: stopFor ? sec2ms(ISO8601Duration2sec(stopFor) || Number(stopFor)) : 0,
+        duration,
+        // TODO drop TS V1: we should store the offset in mm in the store
```
SharglutDev: Could you rename the comment like that to harmonize with existing, I'm affraid we might miss these :)

```suggestion
        // TODO DROP V1: we should store the offset in mm in the store
```

SharglutDev (changes requested):
Thank you for the PR.

Tested the case from @anisometropie and the bug seems to be still here : 

https://github.com/user-attachments/assets/79875fe1-0468-482e-b698-58edc469b4a9

SharglutDev (approved):
Lgtm and tested. The related bug comes from the back-end and will be fix in another PR.

PR #8186, opened 2024-07-25
front: fix app crash when switching too quickly on the times and stops tab

### LINKED ISSUE #8114: Switching from "Route" to "Time and stops" tabs too fast breaks train creation in V2
### What happened?

If you click on "Route" then "Time and stops" tabs too fast, it breaks the train creation page, see my second attempt
https://github.com/user-attachments/assets/4c269af8-6afd-40b0-b54c-80970506fb97

We should see a loader while the pathfinding is loading.
- If the path loads successfully we display the input table
- If the path is not found we display the current error "No path loaded"

1. Create a train
2. Select a path (the longer the easier to reproduce the bug)
3. Quit train creation
4. Reopen train creation
5. Click real fast on path then time and stops
6. Witness the bug

Recette (SNCF)

Firefox

580140d

### LINKED ISSUE #8120: TSV2: Edit train break when click on "Time and stops" tab and the route is not loaded
### What happened?

Check Video : 

https://github.com/user-attachments/assets/a39fd2dc-2cee-4c33-a603-4004ade011ad

I don't get the erreur

1. Edit a train
2. Click on  "Time and stops" tab

Recette (SNCF)

Google Chrome

 580140d

clarani:
closes #8114 
closes #8120

The app crashes because in some cases, the app render more hooks than during the previous render. The early return in TimesStops need to be moved after all the hooks of the file.

### DIFF
--- a/front/src/modules/timesStops/TimesStops.tsx
+++ b/front/src/modules/timesStops/TimesStops.tsx
@@ -39,13 +39,6 @@ const TimesStops = ({
 }: TimesStopsProps) => {
   const isInputTable = tableType === TableType.Input;
   const { t } = useTranslation('timesStops');
-  if (!allWaypoints) {
-    return (
-      <div className="d-flex justify-content-center align-items-center h-100">
-        <p className="pt-1 px-5">{t('noPathLoaded')}</p>
-      </div>
-    );
-  }
 
   const dispatch = useAppDispatch();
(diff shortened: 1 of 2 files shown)


theocrsb (approved):
LGTM and tested

Math-R (approved):
LGTM and tested

PR #8190, opened 2024-07-25
front: fix scroll bar macro mode

### LINKED ISSUE #8187: Macro-micro switch : weird scroll bar in the macro mode
### What happened?

In the microscropic mode, there is a vertical scroll bar which should not be here : the NGE part should just be a little bit smaller to fit the page 

image

I expected to have no scrollbar

as above

Recette (SNCF)

Google Chrome

7faec4a

Uriel-Sautron:
close #8187

### DIFF
--- a/front/src/styles/scss/applications/operationalStudies/_scenario.scss
+++ b/front/src/styles/scss/applications/operationalStudies/_scenario.scss
@@ -32,7 +32,6 @@
       color: var(--white);
       border: unset;
       border-radius: 4px;
-      font-size: 1.5rem;
       padding: 0.5rem;
       z-index: 2;
     }


Yohh (approved):
LGTM tested

front/src/styles/scss/applications/operationalStudies/_scenario.scss
```diff
@@ -32,7 +32,6 @@
       color: var(--white);
       border: unset;
       border-radius: 4px;
-      font-size: 1.5rem;
```
anisometropie, line 35: can’t use a overflow: hidden or change the height ?

Yohh: with the font-size:
image

without:
image

it really seems better without

kmer2016 (approved):
LGTM and tested !

front/src/styles/scss/applications/operationalStudies/_scenario.scss
Uriel-Sautron: With osrd the icons are now the right size, no need for the font-size here.
overflow: hidden cuts the lower part of the macro component

anisometropie (approved):

PR #7855, opened 2024-06-27
Add output table to the simulation result

anisometropie:
closes #7757 
closes #7758

PR #8197, opened 2024-07-25
front: revert formatSuggestedViasToRowVias

anisometropie:
Some changed were lost due to a rebase in https://github.com/OpenRailAssociation/osrd/pull/7855

reverting to the state we had in https://github.com/OpenRailAssociation/osrd/pull/8037


SharglutDev (approved):
Lgtm and tested, thank you for the fix :)

PR #8207, opened 2024-07-25
core: change ignore electrical profile value in sim v2

### LINKED ISSUE #8003: Electrical profile set not used
### What happened?

When creating a V2 scenario with an eletrical profile set, they do not appear in the GEV. (I checked, they do not appear in simulation result query either)

https://github.com/OpenRailAssociation/osrd/assets/47494402/0c36fdd5-4e04-48b6-8187-d7775b82ab2f

 

I would expect the electrical profiles to appear in the GEV, the same way they did in V1

1. Create a scenario on small_infra with small_infra's electrical profile set
2. Create a train following the path I used
3. Witness that there is no electrical profile shown in V2 (you can repeat 1 and 2 on V1 to see it works there)

- **Browser**: Firefox v115
- **OS**: Windows 11

803e0d1

Castavo:
closes #8003


eckter (approved):
oh

We should probably change the "ignore" to "use" to avoid confusion. But that can wait

PR #8210, opened 2024-07-25
editoast: add instrumentation in diesel-async connections

woshilapin:
New version of `diesel-async` gives us the possibility to instrument different events of a DB connection.

We assume that a connection cannot be used concurrently: it cannot start a new query if another one is not already finished. In theory, it could happen if we're pipelining multiple queries into a single connection (but the current code will panic in that case).

See an example in Jaeger here.

image

### DIFF
--- a/editoast/editoast_models/src/db_connection_pool.rs
+++ b/editoast/editoast_models/src/db_connection_pool.rs
@@ -1,3 +1,5 @@
+mod tracing_instrumentation;
+
 use std::ops::Deref;
 use std::ops::DerefMut;
 use std::sync::Arc;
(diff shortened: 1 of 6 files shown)


editoast/editoast_models/src/db_connection_pool.rs
```diff
@@ -91,7 +93,11 @@ impl DbConnectionPoolV2 {
 
     #[cfg(not(feature = "testing"))]
     async fn get_connection(&self) -> Result<DbConnectionV2, DatabasePoolError> {
-        let connection = self.pool.get().await?;
+        use diesel_async::AsyncConnection as _;
+        use tracing_instrumentation::TracingInstrumentation;
+
+        let mut connection = self.pool.get().await?;
+        connection.set_instrumentation(TracingInstrumentation::default());
```
leovalais: Don't we want to log queries running in tests as well?

Wadjetz (approved):

leovalais (approved):
LGTM, tested.

In the "Finishing query" span, the statement seems to be logged twice:

Capture d’écran 2024-07-29 à 11 16 09

Additionally, wdyt about renaming the target to something shorter (like `db_connection`)? I feel like leaving the full module path takes a lot of space and won't help writing `RUST_LOG` filters.

editoast/editoast_models/src/db_connection_pool.rs
woshilapin: Done, not entirely sure how useful it will be though, but not harmful either.

PR #8211, opened 2024-07-26
front: fix time display, use utc for hour conversion

### LINKED ISSUE #8193: TSV2 :  incoherent calculated arrival time in the output table
### What happened?

When you create a train starting at a departur time from a point on the track, 
image
in the otuput table the calculated time does not correspond to the one on the time space graph and on the train sheet
image
image
In addition the line is orange as if the train does not respect the scheduled time. This is the first point of the table. In addition the first point of the table is the first opeartional point and not the waypoint I had chosen. 

- [x] The first point of the output table should be the waypoint I had chosen as departure point.  (fixed in https://github.com/OpenRailAssociation/osrd/pull/8858)
- [x] The output table should display the same arrival time as in the time space graph and in the inputs (fixed in https://github.com/OpenRailAssociation/osrd/pull/8211)

Activate TSV2
Go to the page of creation of a train
Add a train from a point near to Bordeaux St Jean and a point near to Marseille St Charles
look at output table on the page of results 

Recette (SNCF)

Google Chrome

staging 7faec4a

anisometropie:
fixes time display, wrong function was used in d3.
partially fixes #8193

### DIFF
--- a/front/src/modules/timesStops/helpers/__tests__/formatScheduleData.spec.ts
+++ b/front/src/modules/timesStops/helpers/__tests__/formatScheduleData.spec.ts
@@ -1,0 +1,22 @@
+import { describe, it, expect } from 'vitest';
+
+import formatScheduleData from '../formatScheduleData';
+
+describe('formatScheduleData', () => {
+  it('should compute simple arrival time in the correct timezone', () => {
+    const schedule = {
+      at: 'id325',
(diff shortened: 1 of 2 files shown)


SharglutDev (approved):
Lgtm and tested (but it doesn't fix all issues related in #8193)

theocrsb (approved):
LGTM and tested

PR #8214, opened 2024-07-26
ci: pin rust version for 'editoast' and 'gateway'

woshilapin:
`grcov` doesn't compile anymore on `rust@1.80.0` (see https://github.com/mozilla/grcov/pull/1191).


flomonster (approved):
Thanks, we need to do the same with osrdyne

younesschrifi (approved):

PR #8218, opened 2024-07-26
front: fix power restriction edition

### LINKED ISSUE #8021: Adding power restrictions sometimes causes pathfinding failure
### What happened?

Paris Montparnasse-Rennes with a 1TGVDUPL and these power restrictions;
image

image

Created an invalid train with pathfinding_not_found error:
image

And when editing it, it saw this path:
image

I'm expecting power restrictions not to impact my pathfinding, and especially not making it fail.

Reproduce the  description

- **Browser**: Chrome
- **OS**: W11
- **Env**: rec

de8ada8

clarani:
closes #8021

### DIFF
--- a/front/src/modules/powerRestriction/helpers/__tests__/createPathStep.spec.ts
+++ b/front/src/modules/powerRestriction/helpers/__tests__/createPathStep.spec.ts
@@ -1,0 +1,83 @@
+import type { TrackRange } from 'common/api/osrdEditoastApi';
+
+import { findTrackSectionOffset } from '../createPathStep';
+
+describe('findTrackOffset', () => {
+  it('should correctly find the track offset', () => {
+    const trackRangesLengthCumulativeSums = [1000, 2000, 3000, 4000, 5000];
+    const trackRanges = [
+      { track_section: 'track_0' },
+      { track_section: 'track_1' },
+      { track_section: 'track_2', begin: 0, end: 1000, direction: 'START_TO_STOP' },
+      { track_section: 'track_3' },
+      { track_section: 'track_4' },
+    ] as TrackRange[];
+
+    const offsetOnPath = 2060;
+    const result = findTrackSectionOffset(
+      offsetOnPath,
+      trackRangesLengthCumulativeSums,
+      trackRanges
+    );
+
+    expect(result).toEqual({ track: 'track_2', offset: 60 });
+  });
+
+  it('should correctly find the track offset if the track range does not cover the whole track section', () => {
+    const trackRangesLengthCumulativeSums = [1000, 2000, 3000, 4000, 5000];
+    const trackRanges = [
+      { track_section: 'track_0' },
+      { track_section: 'track_1' },
+      { track_section: 'track_2', begin: 600, end: 1600, direction: 'START_TO_STOP' },
+      { track_section: 'track_3' },
+      { track_section: 'track_4' },
+    ] as TrackRange[];
+
+    const offsetOnPath = 2300;
+    const result = findTrackSectionOffset(
+      offsetOnPath,
+      trackRangesLengthCumulativeSums,
+      trackRanges
+    );
+
+    expect(result).toEqual({ track: 'track_2', offset: 900 });
+  });
+
+  it('should correctly find the track offset if the track range direction is STOP_TO_START', () => {
+    const trackRangesLengthCumulativeSums = [1000, 2000, 3000, 4000, 5000];
+    const trackRanges = [
+      { track_section: 'track_0' },
+      { track_section: 'track_1' },
+      { track_section: 'track_2', begin: 1060, end: 60, direction: 'STOP_TO_START' },
+      { track_section: 'track_3' },
+      { track_section: 'track_4' },
+    ] as TrackRange[];
+
+    const offsetOnPath = 2010;
+    const result = findTrackSectionOffset(
+      offsetOnPath,
+      trackRangesLengthCumulativeSums,
+      trackRanges
+    );
+
+    expect(result).toEqual({ track: 'track_2', offset: 1050 });
+  });
+
+  it('should not find the track offset if past cumulative sums', () => {
+    const trackRangesLengthCumulativeSums = [1000, 2000, 3000];
+    const trackRanges = [
+      { track_section: 'track_0' },
+      { track_section: 'track_1' },
+      { track_section: 'track_2' },
+    ] as TrackRange[];
+
+    const offsetOnPath = 3001;
+    const result = findTrackSectionOffset(
+      offsetOnPath,
+      trackRangesLengthCumulativeSums,
+      trackRanges
+    );
+
+    expect(result).toEqual(null);
+  });
+});
--- a/front/src/modules/powerRestriction/helpers/createPathStep.ts
+++ b/front/src/modules/powerRestriction/helpers/createPathStep.ts
@@ -27,16 +27,19 @@ const findTrackSection = (
   return index !== -1 ? pathProperties.trackSectionRanges[index] : null;
 };
 
+/** Return the offset on the track section in mm*/
 const calculateOffset = (
   trackSectionRange: TrackRange,
   position: number, // in mm
   tracksLengthCumulativeSums: number[] // in meters
 ) => {
   const index = findTrackSectionIndex(position, tracksLengthCumulativeSums);
-  const inferiorSum = tracksLengthCumulativeSums[index];
-  return trackSectionRange.direction === 'START_TO_STOP'
-    ? inferiorSum - position
-    : position - inferiorSum;
+  const inferiorSum = index > 0 ? tracksLengthCumulativeSums[index - 1] : 0;
+  const offsetOnTrackRange =
+    trackSectionRange.direction === 'START_TO_STOP'
+      ? inferiorSum - position
+      : position - inferiorSum;
+  return trackSectionRange.begin + offsetOnTrackRange;
 };
 
 const createPathStep = (
--- a/front/src/modules/powerRestriction/helpers/createPathStep.ts
+++ b/front/src/modules/powerRestriction/helpers/createPathStep.ts
@@ -5,7 +5,7 @@ import type { TrackRange } from 'common/api/osrdEditoastApi';
 import type { IntervalItem } from 'common/IntervalsEditor/types';
 import type { PathStep } from 'reducers/osrdconf/types';
 import { getPointCoordinates } from 'utils/geometry';
-import { mToKm, mToMm } from 'utils/physics';
+import { mmToM, mToMm } from 'utils/physics';
 
 import { NO_POWER_RESTRICTION } from '../consts';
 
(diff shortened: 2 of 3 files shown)


woshilapin:
It would be nice to have some tests to avoid regression later. That said, it doesn't fix #8021. The problem is not only about wrong units (millimeters vs meters) but also about the calculation. See, the calculation of the second path item (first power restriction bound) for example, it is 3933000 but as explained in #8021, it should be 1616000.

front/src/modules/powerRestriction/helpers/createPathStep.ts
```diff
@@ -18,25 +18,27 @@ const findTrackSectionIndex = (position: number, tracksLengthCumulativeSums: num
...
-  const inferiorSum = tracksLengthCumulativeSums[index];
-  return trackSectionRange.direction === 'START_TO_STOP'
-    ? inferiorSum - position
-    : position - inferiorSum;
+  const index = findTrackSectionIndex(positionOnPath, tracksLengthCumulativeSums);
+  const trackRange = trackRangesOnPath[index];
+  if (index === -1 || !trackRange) return null;
+
+  // compute offset
+  const inferiorSum = (index && tracksLengthCumulativeSums[index - 1]) || 0;
```
emersion: Subjective nit: I'd personally find it more readable to (1) use `index > 0` instead of `index` to better reflect the intent and (2) use a ternary:

```ts
const inferiorSum = index > 0 ? tracksLengthCumulativeSums[index - 1] : 0;
```

front/src/modules/powerRestriction/helpers/__tests__/createPathStep.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    const tracksLengthCumulativeSums = [1000, 2000, 3000, 4000, 5000];
+    const trackRanges = [
+      { track_section: 'track_0' },
+      { track_section: 'track_1' },
+      { track_section: 'track_2', begin: 0, end: 1000, direction: 'START_TO_STOP' },
+      { track_section: 'track_3' },
+      { track_section: 'track_4' },
+    ] as TrackRange[];
+
+    const offsetOnPath = 2500;
```
emersion: I would recommend using a point which isn't in the middle of the track range for tests. That way, we're more likely to catch issues when operating from the wrong side (`begin + offset` vs. `end - offset` with `offset = (end - begin) / 2`).

front/src/modules/powerRestriction/helpers/__tests__/createPathStep.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    expect(result).toEqual({ track: 'track_2', offset: 500 });
+  });
+
+  it('should correctly find the track offset if the track range does not cover the whole track section', () => {
+    const tracksLengthCumulativeSums = [1000, 2000, 3000, 4000, 5000];
+    const trackRanges = [
+      { track_section: 'track_0' },
+      { track_section: 'track_1' },
+      { track_section: 'track_2', begin: 1000, end: 2000, direction: 'START_TO_STOP' },
```
emersion: Isn't this track range outside of the track section's bounds? According to `tracksLengthCumulativeSums`, each track section has a length of 1000.

Maybe in practice that doesn't matter for the purposes of the test, but would be nicer to have non-bogus test inputs.

Additionally, I'd recommend using `begin`/`end` values which aren't multiples of the track section lengths, to be more likely to catch math mixups between the two.

front/src/modules/powerRestriction/helpers/__tests__/createPathStep.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+      { track_section: 'track_4' },
+    ] as TrackRange[];
+
+    const offsetOnPath = 2500;
+    const result = findTrackOffset(offsetOnPath, tracksLengthCumulativeSums, trackRanges);
+
+    expect(result).toEqual({ track: 'track_2', offset: 1500 });
+  });
+
+  it('should correctly find the track offset if the track range direction is STOP_TO_START ', () => {
```
emersion: Nit: extra space at the end of the test description here

front/src/modules/powerRestriction/helpers/createPathStep.ts
```diff
@@ -18,25 +18,27 @@ const findTrackSectionIndex = (position: number, tracksLengthCumulativeSums: num
...
-  pathProperties: ManageTrainSchedulePathProperties
-) => {
-  const index = findTrackSectionIndex(position, tracksLengthCumulativeSums);
-  return index !== -1 ? pathProperties.trackSectionRanges[index] : null;
-};
-
-const calculateOffset = (
-  trackSectionRange: TrackRange,
-  position: number, // in mm
-  tracksLengthCumulativeSums: number[] // in meters
```
emersion, line 33: Big +1 for not mixing up units in arguments here!

front/src/modules/powerRestriction/helpers/__tests__/createPathStep.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+      { track_section: 'track_3' },
+      { track_section: 'track_4' },
+    ] as TrackRange[];
+
+    const offsetOnPath = 2500;
+    const result = findTrackOffset(offsetOnPath, tracksLengthCumulativeSums, trackRanges);
+
+    expect(result).toEqual({ track: 'track_2', offset: 2500 });
+  });
+});
```
emersion, line 83: Suggestion for a negative test:

```ts
  it('should not find the track offset if past cumulative sums', () => {
    const tracksLengthCumulativeSums = [1000, 2000, 3000];
    const trackRanges = [
      { track_section: 'track_0' },
      { track_section: 'track_1' },
      { track_section: 'track_2' },
    ] as TrackRange[];

    const offsetOnPath = 3001;
    const result = findTrackOffset(offsetOnPath, tracksLengthCumulativeSums, trackRanges);

    expect(result).toEqual(null);
  });
```

front/src/modules/powerRestriction/helpers/createPathStep.ts
```diff
@@ -18,25 +18,27 @@ const findTrackSectionIndex = (position: number, tracksLengthCumulativeSums: num
...
+  const trackRange = trackRangesOnPath[index];
+  if (index === -1 || !trackRange) return null;
+
+  // compute offset
+  const inferiorSum = (index && tracksLengthCumulativeSums[index - 1]) || 0;
+  const offsetOnTrackRange = positionOnPath - inferiorSum;
+  const offsetOnTrackSection =
+    trackRange.direction === 'START_TO_STOP'
+      ? trackRange.begin + offsetOnTrackRange
+      : trackRange.begin - offsetOnTrackRange;
```
emersion, line 40: Question: should this be `trackRange.end + offsetOnTrackRange` instead for `STOP_TO_START`? Disclaimer, I have no clue how these things work.

front/src/modules/powerRestriction/helpers/createPathStep.ts
```diff
@@ -18,25 +18,27 @@ const findTrackSectionIndex = (position: number, tracksLengthCumulativeSums: num
...
-};
-
-const calculateOffset = (
-  trackSectionRange: TrackRange,
-  position: number, // in mm
-  tracksLengthCumulativeSums: number[] // in meters
+/**
+ * Return the track offset corresponding to the given position on path, composed of the track section id
+ * and the offet on this track section in mm*/
+export const findTrackOffset = (
```
emersion: Maybe this could be named `findTrackSectionOffset`, to make it clear it's the offset in terms of track section (as opposed to track range)?

front/src/modules/powerRestriction/helpers/createPathStep.ts
```diff
@@ -18,25 +18,27 @@ const findTrackSectionIndex = (position: number, tracksLengthCumulativeSums: num
...
+
+  // compute offset
+  const inferiorSum = (index && tracksLengthCumulativeSums[index - 1]) || 0;
+  const offsetOnTrackRange = positionOnPath - inferiorSum;
+  const offsetOnTrackSection =
+    trackRange.direction === 'START_TO_STOP'
+      ? trackRange.begin + offsetOnTrackRange
+      : trackRange.begin - offsetOnTrackRange;
+
+  return { track: trackRangesOnPath[index].track_section, offset: offsetOnTrackSection };
```
emersion: Can be simplified to `track: trackRange.track_section`.

front/src/modules/powerRestriction/helpers/createPathStep.ts
```diff
@@ -18,25 +18,27 @@ const findTrackSectionIndex = (position: number, tracksLengthCumulativeSums: num
...
+  trackRangesOnPath: TrackRange[]
 ) => {
-  const index = findTrackSectionIndex(position, tracksLengthCumulativeSums);
-  const inferiorSum = tracksLengthCumulativeSums[index];
-  return trackSectionRange.direction === 'START_TO_STOP'
-    ? inferiorSum - position
-    : position - inferiorSum;
+  const index = findTrackSectionIndex(positionOnPath, tracksLengthCumulativeSums);
+  const trackRange = trackRangesOnPath[index];
+  if (index === -1 || !trackRange) return null;
```
emersion: Nit: I'd personally find it more readable to move the `index === -1` check above `trackRange = trackRangesOnPath[index]`. Or alternatively change the check to `if (!trackRange)` only. Reason being that we don't need two conditions here, a single one should be enough.

emersion (commented):
Here's a first pass of comments. I like the refactoring!

front/src/modules/powerRestriction/helpers/createPathStep.ts
clarani: Yep, done !

front/src/modules/powerRestriction/helpers/createPathStep.ts
clarani: Done

front/src/modules/powerRestriction/helpers/createPathStep.ts
clarani: Nice catch !

front/src/modules/powerRestriction/helpers/createPathStep.ts
clarani: You're right, I kept `if (!trackRange)`

front/src/modules/powerRestriction/helpers/__tests__/createPathStep.spec.ts
clarani: Actually `tracksLengthCumulativeSums` is `trackRangesLengthCumulativeSums` so the track range on track_2 is correct.
I can rename `tracksLengthCumulativeSums` and modify the inputs, tell me if you think it's better now :)

front/src/modules/powerRestriction/helpers/__tests__/createPathStep.spec.ts
clarani: Done

front/src/modules/powerRestriction/helpers/__tests__/createPathStep.spec.ts
clarani: Nice catch, thanks :)

front/src/modules/powerRestriction/helpers/__tests__/createPathStep.spec.ts
clarani: Nice, done !

front/src/modules/powerRestriction/helpers/createPathStep.ts
clarani: image
Here is a small diagram.

To compute the track range offset in this case, I think we have 2 options:
- `trackRange.end + trackRangeLength - offsetOnTrackRange` with `trackRangeLength = Math.abs(trackRange.end - trackRange.begin)`
- `trackRange.begin - offsetOnTrackRange`

I don't see a more general formula, do you ?
If we don't find any, I still prefer the second option, if it's fine for you

clarani:
Thanks @woshilapin for you review, I have fixed the problems you raised

front/src/modules/powerRestriction/helpers/createPathStep.ts
```diff
@@ -127,8 +124,7 @@ export const createCutAtPathStep = (
     id: nextId(),
     positionOnPath: cutAtPosition,
     coordinates: coordinatesAtCut,
-    track: trackSectionRangeAtCut.track_section,
-    offset: offsetAtCut,
+    ...trackOffset,
```
emersion, line 127: No need for `offset: mmToM(trackOffset.offset)` here? Seems like the inputs are the same as in `createPathStep()`, and the result is passed to `cutPowerRestrictionRangesV2()` which may insert the whole thing incl. `offset` into the store?

front/src/modules/powerRestriction/helpers/__tests__/createPathStep.spec.ts
emersion: Oh! Makes perfect sense.

front/src/modules/powerRestriction/helpers/createPathStep.ts
emersion: Thanks a lot for the explanation, I wasn't sure whether `offsetOnTrackRange` was relative to `begin` or `end` for `STOP_TO_START`. The current formula looks good to me!

front/src/modules/powerRestriction/helpers/createPathStep.ts
clarani: Yes you're right, I thought it was a different case, but after a check it was not ! Thanks :pray:

emersion (approved):
LGTM! :)

front/src/modules/powerRestriction/helpers/createPathStep.ts
```diff
@@ -18,25 +18,28 @@ const findTrackSectionIndex = (position: number, tracksLengthCumulativeSums: num
...
-    : position - inferiorSum;
+  const index = findTrackSectionIndex(positionOnPath, tracksLengthCumulativeSums);
+  const trackRange = trackRangesOnPath[index];
+  if (!trackRange) return null;
+
+  // compute offset
+  const inferiorSum = index > 0 ? tracksLengthCumulativeSums[index - 1] : 0;
+  const offsetOnTrackRange = positionOnPath - inferiorSum;
+  const offsetOnTrackSection =
+    trackRange.direction === 'START_TO_STOP'
```
Math-R, line 38: Is there is several direction it could be nice if we have an enum

emersion: This field is auto-generated from OpenAPI, so we can't manually define an enum. (I haven't looked whether it's possible to change the Rust counterpart to get an enum on the TypeScript side.)

clarani: There are only 2 directions, `'START_TO_STOP'` and `'STOP_TO_START'`.
As emersion said, the field is auto-generated, is it ok for you if we report this issue in a new ticket ?

Math-R (approved):
lgtm :)

PR #8219, opened 2024-07-26
front: fix electrification tags on power restriction selector

### LINKED ISSUE #8093: Electrification tags display is incorrect when zooming in the power restrictions selector
### What happened?

When we zoom in the power restrictions selector, the electrification values (voltages) are changing. 
Capture d’écran 2024-07-17 à 11 13 19

They should be responsive and keep their values after zooming in. 

Capture d’écran 2024-07-17 à 11 12 36

1. Go to any V2 scenario 
2. Find a path
3. Go to Simulation Settings
4. Add a power restriction between any interval
5. Zoom in the interval, notice the voltage has changed

Local, Dev (DgexSol), Staging (DgexSol), Dev (SNCF)

Firefox, Google Chrome

64f0491

clarani:
closes #8093 

See commits


Uriel-Sautron (approved):
LGTM !! Tested thank you

SharglutDev (approved):
Lgtm and tested, good job !

PR #8220, opened 2024-07-26
front: convert operations from NGE to OSRD

### LINKED ISSUE #7539: Implement a converter from NGE operations  to OSRD timetable update - Part 1
4th step of meta #7535 

## Acceptance criteria 

- [x] A timetable change implies an update of the NGE model
- [x] An action on trainruns* in NGE implies an update of the timetable
  - [x] Create trainrun
    - [x] Create valid trainrun with valid path => train schedule creation
  - [x] Update trainrun
    - [x] Update trainrun path (shorten and lengthen steps or only start, intermediary or end step) => train schedule updated 
  - [x] Delete trainrun
    - [x] Delete trainrun => delete train schedule

- [x] Information converted from NGE to OSRD : 
  - [x] Name
  - [x] Path

*a `trainrun` in NGE is the equivalent of a `train schedule` in OSRD (even tho it is not semantically 100% correct)

## Old acceptance criteria (when NGE was supposed to take the output of OSRD)

- [x] A timetable change implies an update of the NGE model
- [x] An action on trainruns* in NGE implies an update of the timetable
  - [ ] Create trainrun
    - [x] Create valid trainrun with valid path => train schedule creation (simulation not working, rolling stock must be entered in OSRD first) => once the rollingstock is given and simulation done => trainrun updated (running times, same path)
    - [ ] Create valid trainrun with incomplete path (should pass by intermediary stations) => train schedule creation (simulation not working, rolling stock must be entered in OSRD first) => once the rollingstock is given and simulation done => trainrun updated (running times, updated path)
    - [ ] Create invalid trainrun => train schedule creation (simulation not working, rolling stock must be entered in OSRD first) => once the rollingstock is given and simulation failed => trainrun deleted / updated (TBD)
  - [ ] Update trainrun
    - [x] Update trainrun path (shorten and lengthen steps or only start, intermediary or end step) => train schedule updated and simulation done => trainrun updated (updated running times, updated path)
    - [ ] Update trainrun departure time => train schedule updated and simulation done => trainrun updated (updated running times, updated path)
    - [ ] Update trainrun label => create/update/delete label of associated train schedule
  - [x] Delete trainrun
    - [x] Delete trainrun => delete train schedule

*a `trainrun` in NGE is the equivalent of a `train schedule` in OSRD (even tho it is not semantically 100% correct)

## Implementation plan

- [ ] [OSRD <- NGE] Implement events handler for NGE operations (in OSRD front) => requests to update timetable
  - [x] Enable [`eventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) for the NGE `operationX` events
  - [ ] Callback functions to update back-end (timetable)
  - [ ] Unitary and functionnal tests

## Description

OSRD-NGE Concept diagram

### Train schedules v2

Train schedules v2 is not released at the time on the refinement of this ticket, refinement taking in account TSv2 will have to be done during the development of this ticket.

### [POSTPONE] Train frequency

Train frequency is not a concept well used in OSRD as well as it is in NGE, however this is a key concept to understand how to work with these two level of detail.

A train frequency helper will have to be implemented on OSRD side, a service that will "translate" a trainrun in NGE into a serie of train schedules in OSRD.

For example, if a trainrun, with a trainrun frequency of 1 hour, is created, it will result in the creation of X (to be defined, maybe 24) train schedules in OSRD, with a very close name (i.e., "{trainrun.name}-{hourOfDeparture}), all shifted from 1 hour, using the same rolling stock, same minute of departure, same path etc...

For you information, in the other hand, when converting a scenario from OSRD to NGE, when a serie of train schedules that have the same attributes enough to be interpreted as frequenced train schedules (same rolling stock, slight variation of name (i.e., "{trainrun.name}-{hourOfDeparture}), same path, evenly shifted in time (every hour / 2 hours...)), it will have to let to display only a single trainrun in NGE (frequency depending on the frequence pattern observed) 

### Payloads

See [Trainrun data model](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/src/app/models/trainrun.model.ts) and [TrainrunSection data model](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/src/app/models/trainrunsection.model.ts)

**Create** `trainrun`:
```javascript
TrainrunSectionOperation {
    "type": "create",
    "trainrunSection": TrainrunSection
}
```

**Update** `trainrun`:
```javascript
TrainrunOperation {
    "type": "update",
    "trainrun": Trainrun
}
```

**Delete** `trainrun`:
```javascript
TrainrunOperation {
    "type": "delete",
    "trainrun": Trainrun
}
```

### Timetable update

The philosophy behind the timetable update is different from the PoC, since in the PoC the whole NGE model is computed and the result is compared to the whole timetable. Here we aim to deal with unitary action from NGE and request for a timetable update only once (atomic change instead of whole timetable comparison). In addition to that, the PoC has been implemented on TSv1. For these reasons, do not stick to the PoC algorithms (even tho it can help to understand).

### Labels

The creation or update of a label of a trainrun has to be persisted in OSRD train schedule.

To start with, trainrun labels can be set as train schedule labels, such as:
- `category` ("IC", "EC", ...)
- `frequency` (15min, 30min, ...)

## Definition of ready

- **PO**
  - [x] Acceptance criteria are clear and understandable

- **Technical**
  - [x] If required documentation is available on osrd.fr

- **General**
  - [x] WSJF values are filled
  - [x] Validated by the PM

Caracol3:
close #7539 

Implement create, update and delete operation from nge to osrd
  - create nge operation file
  - create ngeCreateOperation function
  - create ngeUpdateOperation function
  - change the name of the file ngeOperations.ts to export.ts
  - create in export.ts ngeOperations function
  
  - [x] Create trainrun
    - [x] Create valid trainrun with valid path => create invalid trains chedule rolling stock must be entered in OSRD

  - [x] Update trainrun
    - [x] Update trainrun path  => update trains chedule path

  - [x] Delete trainrun
    - [x] Delete trainrun => delete train schedule


front/src/applications/operationalStudies/components/MacroEditor/types.ts
```diff
@@ -65,6 +65,30 @@ export type TimeLock = {
   timeFormatter: null;
 };
 
+export type TrainrunSectionOperation = {
```
emersion: I think we can just use the existing `TrainrunSection` type here instead of re-defining a new one?

front/src/applications/operationalStudies/components/MacroEditor/ngeOperations.ts
emersion, line 1: Suggestion: maybe this file could be renamed to `export.ts` for consistency with the other side (`import.ts`)?

front/src/applications/operationalStudies/components/MacroEditor/NGE.tsx
```diff
@@ -0,0 +1,117 @@
...
+    const frame = frameRef.current!;
+
+    const handleFrameLoad = () => {
+      frame.removeEventListener('load', handleFrameLoad);
+
+      ngeRoot = frame.contentDocument!.createElement('sbb-root') as NGEElement;
+      frame.contentDocument!.body.appendChild(ngeRoot);
+
+      // listens to create, update and delete operations
+      ngeRoot.addEventListener('operation', async (event: Event) => {
```
emersion: Maybe we could introduce a function which takes an NGE operation and handles it inside `ngeOperations.ts`? Basically extract this callback and move it to that file, since it's specific to the NGE → OSRD conversion.

front/src/applications/operationalStudies/components/MacroEditor/NGE.tsx
```diff
@@ -0,0 +1,117 @@
...
+        console.log('Operation received', (event as CustomEvent).detail);
+
+        if ((event as CustomEvent).detail.type === 'create') {
```
emersion: This could probably use a `switch` to handle the different cases.

front/src/applications/operationalStudies/components/MacroEditor/ngeOperations.ts
```diff
@@ -0,0 +1,54 @@
+import type { Distribution, TrainScheduleBase } from 'common/api/generatedEditoastApi';
```
emersion: `osrdEditoastApi` should be used instead of `generatedEditoastApi`.

emersion: Ref https://github.com/OpenRailAssociation/osrd/pull/8250

front/src/applications/operationalStudies/components/MacroEditor/types.ts
Uriel-Sautron: We thought `TrainrunSection` was used for typed TrainrunSectionDTO.
We can therefore add keys to `TrainrunSection` ?

emersion: Eh, indeed, there are differences between [the DTO type](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/b83f0ec2af0c4658632d03ac4497c6a38807cb75/src/app/data-structures/business.data.structures.ts#L113) and [the model type](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/b83f0ec2af0c4658632d03ac4497c6a38807cb75/src/app/models/trainrunsection.model.ts#L19)...

Which additional fields do we need that are missing from the DTO? If we need these additional fields, I agree it's better to keep the types separate to not add extraneous fields which won't be taken into account by NGE to the DTO.

Uriel-Sautron: We need `targetNode`, `sourceNode`  and  `trainrun`

front/src/applications/operationalStudies/components/MacroEditor/ngeOperations.ts
Uriel-Sautron: Done

front/src/applications/operationalStudies/components/MacroEditor/NGE.tsx
Uriel-Sautron: Done

front/src/applications/operationalStudies/components/MacroEditor/NGE.tsx
Uriel-Sautron: Done

front/src/applications/operationalStudies/components/MacroEditor/ngeOperations.ts
Uriel-Sautron: Done

front/src/applications/operationalStudies/components/MacroEditor/types.ts
louisgreiner: Since the `operations` need more detail, we could stick to the model type, and create a new DTO type for `netzgrafikDto`

front/src/applications/operationalStudies/components/MacroEditor/NGE.tsx
```diff
@@ -22,29 +33,33 @@ const frameSrc = `
...
 
-const NGE = () => {
+const NGE = ({ dto }: { dto?: NetzgrafikDto }) => {
   const frameRef = useRef<HTMLIFrameElement>(null);
+  const dispatch = useAppDispatch();
+
+  const [ngeRootElement, setNgeRootElement] = useState<NGEElement | null>(null);
+
+  const { getTimetableID } = useOsrdConfSelectors();
+  const timeTableId = useSelector(getTimetableID);
```
emersion: Up until now, I've designed the `NGE` component to be completely self-contained. That is, it's not tied to anything OSRD-specific. The goal is to have a clear separation of concerns: keep the OSRD-specific logic outside of the component, and have the component only define an API to manage NGE without anything else. That way the low-level NGE details (iframe management, typing etc) are isolated away from the rest of the code.

To retain this design, we'd need to remove this and instead expose a `onOperation` prop that callers can set to a callback taking an `Operation` argument.

Do you think it's better to keep this design, or do you think there are good reason to diverge from it?

Uriel-Sautron: No you're right to want to separate NGE and OSRD.

emersion: Here's what I had in mind:

```ts
// types.ts
export type NGEEvent = NGECreateEvent | NGEDeleteEvent | NGEUpdateEvent;

// NGE.tsx
type NGEProps = {
  dto?: NetzgrafikDto;
  onOperation?: (event: NGEEvent);
};
// ...
ngeRoot.addEventListener('operation', async (event: Event) => {
  const customEvent = event as CustomEvent;
  const op = customEvent.detail as NGEEvent;
  if (onOperation) onOperation(op);
});

// ScenarioV2.tsx
const handleNGEOperation = (op: NGEEvent) =>
  ngeOperations(op, dispatch, timeTableId);
// ...
<NGE dto={ngeDto} onOperation={handlNGEOperation} />
```

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,105 @@
...
+  osrdEditoastApi,
+  type Distribution,
+  type TrainScheduleBase,
+} from 'common/api/osrdEditoastApi';
+import type { AppDispatch } from 'store';
+
+import type { NGEDeleteEvent, NGEEvent, NGEUpdateEvent, TrainrunSectionOperation } from './types';
+
+const defautlPayloads = {
+  constraint_distribution: 'STANDARD' as Distribution,
```
emersion: Nit: I believe `as Distribution` is unnecessary here.

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,105 @@
+import {
+  osrdEditoastApi,
+  type Distribution,
+  type TrainScheduleBase,
+} from 'common/api/osrdEditoastApi';
+import type { AppDispatch } from 'store';
+
+import type { NGEDeleteEvent, NGEEvent, NGEUpdateEvent, TrainrunSectionOperation } from './types';
+
+const defautlPayloads = {
```
emersion: - Typo: "default"
- Nit: should this be singular? There is an "s" at the end but this object represents a single payload
- I wonder if it should be UPPER_CASE instead? I'm not sure what the style rules are but I was under the impression that constants needed to be that way.

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,105 @@
...
+  constraint_distribution: 'STANDARD' as Distribution,
+  rolling_stock_name: '',
+  start_time: '2024-07-15T08:00:00+02:00',
```
emersion: In the future, we should grab this from the first NGE train run section. However, this is out-of-scope for this PR. Maybe we should add a TODO to make it clear the hardcoded value is temporary?

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,105 @@
...
+  };
+  return payloads;
+};
+
+const ngeOperations = async (event: NGEEvent, dispatch: AppDispatch, timeTableId: number) => {
+  switch (event.type) {
+    case 'create': {
+      dispatch(
+        osrdEditoastApi.endpoints.postV2TimetableByIdTrainSchedule.initiate({
+          id: timeTableId!,
```
emersion: Nit: `!` should be unnecessary here because `timeTableId` isn't nullable

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,105 @@
...
+    path,
+    train_name: firstTrainrunSection.trainrun.name,
+  };
+  return payloads;
+};
+
+const ngeOperations = async (event: NGEEvent, dispatch: AppDispatch, timeTableId: number) => {
+  switch (event.type) {
+    case 'create': {
+      dispatch(
```
emersion: We should probably `await` here, so that the function doesn't return before we're done sending the HTTP request?

front/src/applications/operationalStudies/components/MacroEditor/types.ts
emersion: Hm. How about we do this for now?

```ts
export type OperationTrainrunSection = TrainrunSection & {
  // extra props we need here
};
```

Nit: IMHO `OperationTrainrunSection` reads a _tad_ better than `TrainrunSectionOperation`: the latter makes it sound like the whole thing is an operation.

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,105 @@
...
+          id: timeTableId!,
+          body: ngeCreateOperation(event.trainrunSection),
+        })
+      ).unwrap();
+      break;
+    }
+    case 'delete': {
+      const { trainrun } = event as NGEDeleteEvent;
+      await dispatch(
+        osrdEditoastApi.endpoints.deleteV2TrainSchedule.initiate({ body: { ids: [trainrun.id] } })
```
emersion: Here we're assuming that `trainrun.id` is the OSRD train schedule ID. This works fine if the trainrun was imported from OSRD, but breaks when the trainrun was created inside NGE. Trainruns created inside NGE get an auto-incremented ID generated by NGE which don't match OSRD IDs.

(Same applies to `update`.)

I don't think we can easily fix this here, but it's potentially dangerous since we may delete completely unrelated trainruns (what NGE thinks trainrun 42 is may be a completely unrelated trainrun for OSRD).

(My plan to fix this completely would be to add a map of NGE <-> OSRD IDs.)

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,105 @@
...
+          id: `${sourceNode.id}`,
+          deleted: false,
+        },
+        {
+          trigram: targetTrigram,
+          secondary_code: targetSecondaryCode,
+          id: `${targetNode.id}`,
+          deleted: false,
+        },
+      ],
```
emersion: In order to not duplicate the logic to create the `path`, could we call the logic contained in `ngeUpdateOperation` from `ngeCreateOperation`? Basically the idea would be:

```ts
const payloads: TrainScheduleBase[] = [
  ngeUpdateOperation([trainrunSection]),
];
```

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,105 @@
...
+
+import type { NGEDeleteEvent, NGEEvent, NGEUpdateEvent, TrainrunSectionOperation } from './types';
+
+const defautlPayloads = {
+  constraint_distribution: 'STANDARD' as Distribution,
+  rolling_stock_name: '',
+  start_time: '2024-07-15T08:00:00+02:00',
+};
+
+const ngeCreateOperation = (trainrunSection: TrainrunSectionOperation) => {
```
emersion: Naming nit: `ngeCreateOperation` makes it sound like this is a variable storing an operation, but it's actually a function. Maybe `handleCreateOperation` would be a bit better? (Same applies to other functions.)

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,105 @@
...
+        },
+      ],
+      train_name: trainrunSection.trainrun.name,
+    },
+  ];
+  return payloads;
+};
+
+const ngeUpdateOperation = (trainrunSections: TrainrunSectionOperation[]) => {
+  const path = trainrunSections.flatMap((section, index) => {
```
emersion: This loop assumes that the trainrun sections have a specific format: [A B] [B C] [C D]. However this is not something guaranteed by NGE, we could get something like [B C] [A B] or even [A B] [C D].

I don't think this is necessarily something that we need to handle in this PR, but could we add a TODO about it?

emersion (commented):
This is better :) here are a few more comments. Let me know if something doesn't make sense!

emersion:
Also, this needs a rebase now that NGE import has been merged.

front/src/applications/operationalStudies/components/MacroEditor/export.ts
emersion: An, ahem, quick'n'dirty way to ~~address~~ hack this would be to use a different type to host OSRD IDs, that way we can tell them apart from NGE IDs and bail out if we encounter one. Typically convert the OSRD ID to a string at import time, check if the ID is a string here, and convert it back to a number before passing it to the backend.

Please don't credit me for this if you end up implementing it.

front/src/applications/operationalStudies/components/MacroEditor/export.ts
louisgreiner: Since non-continous trainruns (I believe you had that in mind) exist in NGE, we need to prevent that. However it is true that it is not our priority (but soon to be)

emersion: [A B] [C D] is a case of discontinuous trainruns (a single train run made up of two separate paths).

[B C] [A B] is a case of a continuous train run, but with the sections in an arbitrary order.

front/src/applications/operationalStudies/components/MacroEditor/types.ts
Uriel-Sautron: Done

front/src/applications/operationalStudies/components/MacroEditor/export.ts
Uriel-Sautron: Done, you're right about UPPER_CASE

front/src/applications/operationalStudies/components/MacroEditor/export.ts
Uriel-Sautron: We need to specify the type here, because when the payload is created typescript assigns the string type to the constraint_distribution

front/src/applications/operationalStudies/components/MacroEditor/export.ts
Uriel-Sautron: Done

front/src/applications/operationalStudies/components/MacroEditor/export.ts
Uriel-Sautron: Done

front/src/applications/operationalStudies/components/MacroEditor/export.ts
Uriel-Sautron: Done

front/src/applications/operationalStudies/components/MacroEditor/export.ts
Uriel-Sautron: Done

front/src/applications/operationalStudies/components/MacroEditor/export.ts
Uriel-Sautron: Done

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,87 @@
...
+    case 'delete': {
+      const { trainrun } = event as NGEDeleteEvent;
+      await dispatch(
+        osrdEditoastApi.endpoints.deleteV2TrainSchedule.initiate({ body: { ids: [trainrun.id] } })
+      ).unwrap();
+      break;
+    }
+    case 'update': {
+      const { trainrunSections } = event as NGEUpdateEvent;
+      const [firstTrainrunSection] = trainrunSections;
```
clarani: Maybe we should add a test here to check that `trainrunSections.length > 0`, otherwise the app will panick l76
```suggestion
      const [firstTrainrunSection] = trainrunSections;
```

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,87 @@
...
+
+const ngeOperations = async (event: NGEEvent, dispatch: AppDispatch, timeTableId: number) => {
+  switch (event.type) {
+    case 'create': {
+      await dispatch(
+        osrdEditoastApi.endpoints.postV2TimetableByIdTrainSchedule.initiate({
+          id: timeTableId,
+          body: handleCreateOperation(event.trainrunSection),
+        })
+      ).unwrap();
```
clarani: nit: what do you think of adding a const to create the payload here ? You could then remove the function `handleCreateOperation`
```suggestion
      const trainSchedulePayload = createPayloadOperation([event.trainrunSection]);
      await dispatch(
        osrdEditoastApi.endpoints.postV2TimetableByIdTrainSchedule.initiate({
          id: timeTableId,
          body: [trainSchedulePayload],
        })
      ).unwrap();
```

front/src/applications/operationalStudies/components/MacroEditor/export.ts
clarani: You can type DEFAULT_PAYLOAD like this to remove the Distribution ;)
```
const DEFAULT_PAYLOAD: Pick<
  TrainScheduleBase,
  'constraint_distribution' | 'rolling_stock_name' | 'start_time'
> = {
  constraint_distribution: 'STANDARD',
  rolling_stock_name: '',
  start_time: '2024-07-15T08:00:00+02:00',
};
```

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,87 @@
...
+import type { NGEDeleteEvent, NGEEvent, NGEUpdateEvent, OperationTrainrunSection } from './types';
+
+// TODO: add a dynamic values when available
+const DEFAULT_PAYLOAD = {
+  constraint_distribution: 'STANDARD' as Distribution,
+  rolling_stock_name: '',
+  start_time: '2024-07-15T08:00:00+02:00',
+};
+
+const createPayloadOperation = (trainrunSections: OperationTrainrunSection[]) => {
```
clarani: `createTrainSchedulePayload` would be a better name I think
```suggestion
const createTrainSchedulePayload = (trainrunSections: OperationTrainrunSection[]) => {
```

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,87 @@
...
+    const sourceNode = {
+      trigram: sourceTrigram,
+      secondary_code: sourceSecondaryCode,
+      id: `${section.sourceNode.id}-${index}`,
+    };
+    const targetNode = {
+      trigram: targetTrigram,
+      secondary_code: targetSecondaryCode,
+      id: `${section.targetNode.id}-${index}`,
+    };
```
clarani: nit: you can create a small function to create a pathItem from a node:
```
const createPathItemFromNode = (
  { betriebspunktName: nodeName, id: nodeId }: Node,
  index: number
) => {
  const [trigram, secondaryCode] = nodeName.split('/');
  return {
    trigram: trigram,
    secondary_code: secondaryCode,
    id: `${nodeId}-${index}`,
  };
};
```
and then 
```suggestion
     const sourceNode = createPathItemFromNode(section.sourceNode, index);
    const targetNode = createPathItemFromNode(section.targetNode, index);
```

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,87 @@
...
+    if (index === trainrunSections.length - 1) {
+      return [sourceNode, targetNode];
+    }
```
clarani: You can compute targetNode only in the if
```suggestion
    if (index === trainrunSections.length - 1) {
      const targetNode = ...
      return [sourceNode, targetNode];
    }
```

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,87 @@
...
+    return [sourceNode];
+  });
+  const [firstTrainrunSection] = trainrunSections;
```
clarani: I think this is more readable with `[0]`. Should we add a test to check that `trainrunSections.length > 0` ?
```suggestion
  const firstTrainrunSection = trainrunSections[0];
```

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,87 @@
...
+  const payloads: TrainScheduleBase = {
```
clarani: You can also directly return here and add the TrainScheduleBase type in the function signature
```suggestion
  const payload: TrainScheduleBase = {
```

front/src/applications/operationalStudies/components/MacroEditor/NGE.tsx
```diff
@@ -42,9 +48,15 @@ const NGE = ({ dto }: { dto?: NetzgrafikDto }) => {
       setNgeRootElement(ngeRoot);
 
       // listens to create, update and delete operations
-      // ngeRoot.addEventListener('operation', (event: Event) => {
-      //   console.log('Operation received', (event as CustomEvent).detail);
-      // });
+      ngeRoot.addEventListener('operation', async (event: Event) => {
```
emersion: Nit: this doesn't need to be `async` now

front/src/applications/operationalStudies/components/MacroEditor/NGE.tsx
```diff
@@ -42,9 +48,15 @@ const NGE = ({ dto }: { dto?: NetzgrafikDto }) => {
...
+        const customEvent = event as CustomEvent;
+        const op = customEvent.detail as NGEEvent;
+        if (onOperation) onOperation(op);
+      });
+
+      if (dto) {
+        ngeRoot.netzgrafikDto = dto;
+      }
```
emersion: This should be removed: it was part of the old logic, but is now handled by the second `useEffect` below.

(Probably remnant from a rebase.)

front/src/applications/operationalStudies/components/MacroEditor/export.ts
emersion: Oh right good point!

front/src/applications/operationalStudies/views/v2/ScenarioV2.tsx
```diff
@@ -165,6 +169,8 @@ const ScenarioV2 = () => {
     setCollapsedTimetable(isMacroMode);
   };
 
+  const handleNGEOperation = (op: NGEEvent) => ngeOperations(op, dispatch, timetableId!);
```
clarani: I think this function should be moved in `NGE.tsx`. You can give a new props `timetableId` to `NGE`, and get the dispatch with `useAppDispatch` directly in `NGE.tsx`

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,87 @@
...
+};
+
+const handleCreateOperation = (trainrunSection: OperationTrainrunSection) => [
+  createPayloadOperation([trainrunSection]),
+];
+
+const handleUpdateOperation = (trainrunSections: OperationTrainrunSection[]) =>
+  createPayloadOperation(trainrunSections);
+
+const ngeOperations = async (event: NGEEvent, dispatch: AppDispatch, timeTableId: number) => {
```
emersion: Can we also rename this to `handleOperation` for consistency?

front/src/applications/operationalStudies/views/v2/ScenarioV2.tsx
```diff
@@ -8,7 +8,13 @@ import { useSelector } from 'react-redux';
 import { useParams } from 'react-router-dom';
 
 import BreadCrumbs from 'applications/operationalStudies/components/BreadCrumbs';
+import ngeOperations from 'applications/operationalStudies/components/MacroEditor/export';
```
emersion: Ditto!

emersion (commented):
Some very minor comments, apart from these and https://github.com/OpenRailAssociation/osrd/pull/8220/files#r1705446045, looks good!

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,87 @@
...
+  };
+  return payloads;
+};
+
+const handleCreateOperation = (trainrunSection: OperationTrainrunSection) => [
+  createPayloadOperation([trainrunSection]),
+];
+
+const handleUpdateOperation = (trainrunSections: OperationTrainrunSection[]) =>
+  createPayloadOperation(trainrunSections);
```
emersion: Another nit: TBH, sounds like these are simple enough to just be inlined in the caller, WDYT?

front/src/applications/operationalStudies/components/MacroEditor/export.ts
emersion: Okay, last nit: this is a single payload so should be singular not plural :P

emersion: Damn it, @clarani ended up being faster than me: https://github.com/OpenRailAssociation/osrd/pull/8220/files#r1707204029

front/src/applications/operationalStudies/components/MacroEditor/export.ts
emersion: I believe this is guaranteed by NGE. Wouldn't hurt to add a `console.assert()` or throw an error out of caution though.

front/src/applications/operationalStudies/views/v2/ScenarioV2.tsx
emersion: The motivation for this to be out of `NGE.tsx` is explained here: https://github.com/OpenRailAssociation/osrd/pull/8220#discussion_r1701793216

front/src/applications/operationalStudies/components/MacroEditor/export.ts
emersion: Ref https://github.com/OpenRailAssociation/osrd/pull/8220/files#r1707093806 for the discussion about the length check.

Agreed that `[0]` is more readable!

front/src/applications/operationalStudies/components/MacroEditor/export.ts
clarani, line 1: `export` is not really understandable at first sight. Do you think `convertNgeOperationToOsrdOperation` is a better name ? or `ngeOperationToOsrdOperation` ?

front/src/applications/operationalStudies/components/MacroEditor/types.ts
```diff
@@ -153,3 +159,20 @@ export type NetzgrafikDto = {
...
 };
+
+type NGECreateEvent = {
+  type: 'create';
+  trainrunSection: OperationTrainrunSection;
+};
+
+export type NGEDeleteEvent = {
+  type: 'delete';
+  trainrun: Trainrun;
```
clarani, line 160: nit: you can only give the trainrunId to delete
```suggestion
  trainrunId: number;
```

clarani (commented):
Thanks for you PR :) I left some comments ;)

front/src/applications/operationalStudies/components/MacroEditor/NGE.tsx
Uriel-Sautron: Done

front/src/applications/operationalStudies/components/MacroEditor/NGE.tsx
Uriel-Sautron: Done

clarani:
I have this bug, I can't edit the name of a created train on NGE
[Capture vidéo du 2024-08-07 17-22-07.webm](https://github.com/user-attachments/assets/e3397f94-7073-4a10-9c28-c3a46d47b486)

front/src/applications/operationalStudies/components/MacroEditor/export.ts
emersion: `export` is for symmetry with `import`.

front/src/applications/operationalStudies/components/MacroEditor/export.ts
Uriel-Sautron: Done

emersion:
> I have this bug, I can't edit the name of a created train on NGE

I don't think we get an event from NGE for this action yet. This will be fixed once we upgrade NGE to include https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/pull/213.

@louisgreiner can you confirm?

front/src/applications/operationalStudies/components/MacroEditor/export.ts
Uriel-Sautron: Done

front/src/applications/operationalStudies/views/v2/ScenarioV2.tsx
Uriel-Sautron: Done

front/src/applications/operationalStudies/components/MacroEditor/export.ts
Uriel-Sautron: Done

front/src/applications/operationalStudies/views/v2/ScenarioV2.tsx
clarani: Hm I understand. I didn't want to add this function in Scenario because this component is already huge and handles a lot of high level things. But I also understand your point.
We can keep it here for the moment and move it to an NGE wrapper if more props need to be created for NGE in the future.

emersion: Right! We've tried to keep the `ScenarioV2` changes as small as possible.

front/src/applications/operationalStudies/components/MacroEditor/export.ts
Uriel-Sautron: I had already planned this refacto, but I'm waiting for this [PR](https://github.com/OpenRailAssociation/osrd/pull/8343) to be merged as there will be changes to make on it.

front/src/applications/operationalStudies/components/MacroEditor/export.ts
Uriel-Sautron: Done

front/src/applications/operationalStudies/components/MacroEditor/export.ts
Uriel-Sautron: Done

louisgreiner:
Other events than:
- create a trainrun
- update the PATH of a trainrun
- delete a trainrun 

are not implemented yet as mentionned above, but will be at the end of next week, @clarani

front/src/applications/operationalStudies/components/MacroEditor/export.ts
Uriel-Sautron: Done

front/src/applications/operationalStudies/components/MacroEditor/export.ts
Uriel-Sautron: This will be the case in the next commit when the NGE upgrade PR is merged

Uriel-Sautron:
Thank you for your review @clarani.
We don't have a name edition event yet.

front/src/applications/operationalStudies/components/MacroEditor/types.ts
emersion: We can't change this without changing NGE. These data structures describe values coming from NGE.

The full trainrun isn't useful to us, but might be to other NGE users.

clarani: OK got it, it's fine for me :+1:

front/src/applications/operationalStudies/components/MacroEditor/export.ts
clarani: `export` and `import` are both not really understandable at first sight.
I would prefer to have `ngeOperationToOsrdOperation` and `osrdOperationToNgeOperation`, or more simply `ngeToOsrd` and `osrdToNge`, what do you think ?

emersion: The "operation" bit is wording specific to the NGE → OSRD side. The "operation" data structures represent NGE changes/updates. So I don't think it makes sense to include in the name.

If you don't like import/export, I think ngeToOsrd/osrdToNge are great names!

front/src/applications/operationalStudies/components/MacroEditor/export.ts
louisgreiner: resolved by creating a map that gets the train schedule instead

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,114 @@
...
+      break;
+    }
+    case 'delete': {
+      const trainrunIdToDelete = CREATED_TRAINRUN.get(trainrun.id) || trainrun.id;
+      await dispatch(
+        osrdEditoastApi.endpoints.deleteV2TrainSchedule.initiate({
+          body: { ids: [trainrunIdToDelete] },
+        })
+      ).unwrap();
+      CREATED_TRAINRUN.delete(trainrun.id);
```
emersion: Note to self: even if the map keys are NGE-generated IDs, this should be fine to do unconditionally even with IDs coming from OSRD. Indeed, NGE will never generate an ID already in use by a trainrun imported from OSRD.

front/src/applications/operationalStudies/components/MacroEditor/export.ts
```diff
@@ -0,0 +1,114 @@
+import { osrdEditoastApi, type TrainScheduleBase } from 'common/api/osrdEditoastApi';
+import type { AppDispatch } from 'store';
+
+import type { NetzgrafikDto, NGETrainrunEvent, TrainrunSection, Node, Trainrun } from './types';
+
+const CREATED_TRAINRUN = new Map<number, number>();
```
emersion: Tiny nit: since this is a map that we mutate at runtime (adding/removing elements as NGE operations are handled), it's not really a constant value, so maybe it's unfair to put it in UPPER_CASE?

front/src/applications/operationalStudies/components/MacroEditor/export.ts
louisgreiner: We could indeed populate the map as the import (`osrdToNge`) is called, and then use this table only to communicate with train schedule operations

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -50,6 +50,17 @@ const useSetupItineraryForTrainUpdate = (
       const trainSchedule = await getTrainScheduleById({
         id: trainIdToEdit,
       }).unwrap();
+      if (!trainSchedule.rolling_stock_name) {
+        adjustConfWithTrainToModifyV2(
+          trainSchedule,
+          [],
```
clarani, line 56: Why do you reset the pathSteps here ?

emersion (approved):
With the remaining comments fixed, this LGTM!

front/src/applications/operationalStudies/components/MacroEditor/export.ts
Uriel-Sautron: Done

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/adjustConfWithTrainToModify.ts
```diff
@@ -9,8 +9,8 @@ import { msToKmh } from 'utils/physics';
 
 export function adjustConfWithTrainToModifyV2(
   trainSchedule: TrainScheduleResult,
-  pathSteps: PathStep[],
-  rollingStockId: number,
+  pathSteps: (PathStep | null)[],
```
clarani, line 12: ~~You don't need to change the type here I think~~
```suggestion
  pathSteps: PathStep[],
```

clarani: My bad, no you need to keep this sorry ! You can resolve this comment

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
clarani: If you want to reset the pathSteps, you need to give `[null, null]`

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -50,6 +50,17 @@ const useSetupItineraryForTrainUpdate = (
       const trainSchedule = await getTrainScheduleById({
         id: trainIdToEdit,
       }).unwrap();
+      if (!trainSchedule.rolling_stock_name) {
+        adjustConfWithTrainToModifyV2(
```
emersion, line 54: Note: this fixes the edit form not getting populated with the train schedule details in case there is no rolling stock, however there are more error cases below where we should populate the form but don't do it, for instance when pathfinding fails. We should really fix that but it's orthogonal to this PR.

clarani (approved):
LGTM :white_check_mark:

PR #8222, opened 2024-07-26
front: stop showing arrival warning due to rounding errors.

### LINKED ISSUE #8185: TSV2 : Differences due to calculation accuracy generate trains that do not respect scheduled points
### What happened?

When the difference between the requested arrival and the calculated arrival is not zero, the line of the table is orange to indicate that the train does not respect the scheduled point. This happens even if the difference is tiny. Some times, this difference is smaller than computational accurancy. As a consequence, all the scheduled points are considered as not honored. 

A theshold equal to 1 sec should be set to differenciate the real cases of not honored scheduled points from the accurancy errors. This means that:

- if the difference between the requested arrival time and the calculated arrival time is less than 1 sec the line of the batble is not orange (no warning)
- in the list of trains, the icon to indicate not honored scheduled points on this train is not displayed
- the calculated arrival time should be roundend to the value of the requested arrival time 

Activate TSV2
Go to the page to create a train
Fix a scheduled point
Go to the results page and look at the output table 

Dev (DgexSol)

Google Chrome, Microsoft Edge

dev 7faec4a

anisometropie:
the simulation result output table oftens was displaying scheduled arrival time and calculated arrival time with a difference of one second, due to rounding errors during interpolation.

Values such that | scheduleArrival - calculatedArrival | <= 1 second are now displayed both as scheduleArrival, and no warning is shown

fixes #8185


front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -89,11 +90,15 @@ function useOutputTableData(
             selectedTrainSchedule,
             pathStepId
           );
+          const calculatedArrival = secToHoursString(opPoint.time, true);
```
woshilapin: Why converting to a string, and then comparing typed datetime (don't know exactly what type it is but I guess `dayjs` return something typed)? Wouldn't it be more logical to compare first datetimes directly?

front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -89,11 +90,15 @@ function useOutputTableData(
...
+          const arrivalValuesAreClose =
+            Math.abs(compareTimeString(scheduleData.arrival, calculatedArrival)) <= 1000;
```
woshilapin: How about auto-documented code by naming this hardcoded number (without it, it looks like a non-explained arbitrary value). I'm open to any other naming, I just didn't came up with a good idea.

```suggestion
          const MAXIMAL_ROUNDING_ERROR_IN_MS = 1000;
          const arrivalValuesAreClose =
            Math.abs(compareTimeString(scheduleData.arrival, calculatedArrival)) <= 1000;
```

woshilapin (changes requested):
Nice work tackling that bug.

At the very least, it would be good to add some unit tests so this fix never regress.

front/src/modules/timesStops/hooks/useOutputTableData.ts
anisometropie: I’m now comparing the numbers directly before converting the TimeString.
added some tests as well

front/src/modules/timesStops/hooks/useOutputTableData.ts
anisometropie: added a named constant

clarani (approved):
LGTM

front/src/modules/timesStops/helpers/scheduleData.ts
```diff
@@ -0,0 +1,35 @@
...
+  const arrivalSeconds = schedule.arrival
+    ? startTimeSeconds + ISO8601Duration2sec(schedule.arrival)
+    : null;
+  const stopForSeconds = schedule.stop_for ? ISO8601Duration2sec(schedule.stop_for) : null;
+  const departure =
+    arrivalSeconds && stopForSeconds ? startTimeSeconds + arrivalSeconds + stopForSeconds : null;
+  return {
+    arrival: arrivalSeconds,
+    departure,
+    stopFor: stopForSeconds,
```
Yohh, line 25:
```suggestion
  const arrival = schedule.arrival
    ? startTimeSeconds + ISO8601Duration2sec(schedule.arrival)
    : null;
  const stopFor = schedule.stop_for ? ISO8601Duration2sec(schedule.stop_for) : null;
  const departure = arrival && stopFor ? startTimeSeconds + arrival + stopFor : null;
  return {
    arrival,
    departure,
    stopFor,
```

Yohh: it was maybe to explicitly show that there is seconds, but the function name seems enough

woshilapin (approved):
Thank you for addressing the comments and adding some tests.

front/src/modules/timesStops/helpers/scheduleData.ts
clarani: I would rather keep it like this, because otherwise we don't know at first sight what we are manipulating. Is it ok for you ?

Yohh (approved):

PR #8225, opened 2024-07-29
front: fix driver train schedules not updating when switching between…

### LINKED ISSUE #8203: TSV2 : the value of theoretical margin is not taken into account
### What happened?

We create a train in TSV2. The path is from a waypoint chosen on the map (not an operational point). We put a theoretical margin of 4min/100km. Cette valeur n'est pas prise en compte dans le calcul. 

the margin should be applied

Create a train in TSV2. 
Select a point as departure from the map and an arrival. 
Go to "Time and Stops" and put a theoretical margin of 4min/100km on the departure point
Go to simulation results and compare the arrival time by switching from ECO to BASE

Recette (SNCF)

Google Chrome

staging 7faec4a

SharglutDev:
… base or eco

Move the switch state in the custom hook where the operational points are formatted

close #8203


kmer2016 (approved):
LGTM and tested

emersion (approved):

PR #8228, opened 2024-07-29
front: fix crash when deleting route in train editor

### LINKED ISSUE #8216: TSV2: Error when deleting a route and the scenario is no longer accessible
### What happened?

Error when deleting a route, and the scenario is no longer accessible

https://github.com/user-attachments/assets/98809e19-cdb1-40c0-bc6f-82f1db2d34bd

Can successfully delete the route

Check video

Dev (DgexSol)

Google Chrome

dba683b

emersion:
To reproduce, open the form to add a new train, select the "Route" tab and click the "Delete the route" button.

pathSteps may be empty, causing this crash:

    Uncaught TypeError: right-hand side of 'in' should be an object, got undefined
        matchPathStepAndOp utils.ts:38
        originIndexInOps utils.ts:28
        formatSuggestedViasToRowVias utils.ts:28
        TimesStops TimesStops.tsx:51

Closes: https://github.com/OpenRailAssociation/osrd/issues/8216

### DIFF
--- a/front/src/modules/timesStops/helpers/utils.ts
+++ b/front/src/modules/timesStops/helpers/utils.ts
@@ -25,7 +25,9 @@ export const formatSuggestedViasToRowVias = (
   // If the origin is in the ops and isn't the first operational point, we need
   // to move it to the first position
   const origin = pathSteps[0];
-  const originIndexInOps = operationalPoints.findIndex((op) => matchPathStepAndOp(origin, op));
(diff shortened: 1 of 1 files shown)


SharglutDev (approved):
Lgtm and tested, thank you !

theocrsb (commented):
LGTM and tested

PR #8229, opened 2024-07-29
front: add mandatory fields in rollingstock editor

### LINKED ISSUE #6548: Add a `*` for mandatory fields when creating a rolling stock
There is a lot of fields to fill here :

Image

Only some are mandatory, we could maybe show them with a `*` Like `Name*` instead of `Name`, `Length*` instead of `Length`, `Mass*` instead of `Mass`, ...

RomainValls:
close #6548


theocrsb (approved):
LGTM and tested

clarani (commented):

While reviewing, I spotted another bug: the fields related to BP/CC should not be required and displayed if the rolling stock is a thermal one. We can fix this in another PR

clarani (approved):

PR #8230, opened 2024-07-29
scripts/cleanup-db: fix command to drop all keys from redis 

emersion:
_Also contains two cleanup commits._

`DEL *` doesn't work, it never deletes any key:

    127.0.0.1:6379> DBSIZE
    (integer) 31
    127.0.0.1:6379> DEL *
    (integer) 0
    127.0.0.1:6379> DBSIZE
    (integer) 31

Redis silently ignores any argument which doesn't match any key.
Seems like `DEL` doesn't support wildcards.

Use `FLUSHALL` instead.

Ref https://github.com/OpenRailAssociation/osrd/pull/8087

### DIFF
--- a/scripts/cleanup-db.sh
+++ b/scripts/cleanup-db.sh
@@ -57,7 +57,7 @@ docker exec osrd-postgres psql -f //tmp/init.sql > /dev/null
 
 # Clear Redis Cache
 echo "Deleting redis cache..."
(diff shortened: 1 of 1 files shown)


flomonster (approved):
Good catch!

Khoyo (approved):

PR #8232, opened 2024-07-29
front: fix unselect train on onblur

### LINKED ISSUE #8194: stdcm: when clicking out of the train selection tool, it removes the selected train
### What happened?

https://github.com/user-attachments/assets/d411d350-cece-41e9-ae00-8ab8507d08c4

In this video  I click on the gauge selector but anywhere else seems to work

I expect the train to not disappear if I click inside the field then out

1. Go to STDCM page
2. Pick a train
3. Click inside the train name field
4. Click out
5. Witness

Recette (SNCF)

Firefox

7faec4a

kmer2016:
close #8194


SarahBellaha (approved):
Tested, LGTM ✔️

Math-R (approved):
lgtm :)

PR #8233, opened 2024-07-29
front: fix scroll overflow in infra editor

### LINKED ISSUE #7317: Scroll down on the editor page
### What happened?

When we select an locked infra, we should not be able to scroll down on the editor main page. On some screens and if we select the tiny infra, we can even see a small 0 at the bottom of the page.
We should also check the html of the page and look for some 0.
image

- [ ] no scroll on the editor main page

1. Select a locked infra
2. Scroll on the page

- **Browser**: Firefox v109
- **OS**: Nixos v22.11
- **Env**: Local

staging 11da135

RomainValls:
close #7317

### DIFF
--- a/front/src/styles/scss/applications/editor/_editor.scss
+++ b/front/src/styles/scss/applications/editor/_editor.scss
@@ -103,12 +103,10 @@
   }
 
   .infra-locked {
-    width: 32.25rem;
-    overflow-y: auto;
-    height: auto;
+    flex-shrink: 0;
   }
 
-  .panel-box {
+  .panel-container {
     width: 32.25rem;
     height: 100%;
     overflow-y: auto;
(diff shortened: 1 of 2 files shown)


front/src/styles/scss/applications/editor/_editor.scss
SharglutDev, line 1: I think just by using the panel-container class instead of panel-box just did the trick you don't need any of the other changes : 
- `flex-shrink` in `.infra-locked`
- you van remove flex-grow and overflow in `.panel-box`
- you can remove flex and flex-column in `.panel-container`

Or did I miss something ?

SharglutDev (changes requested):
Thank you for the fix ! Everything works but probably you can simplify it, see comments

front/src/styles/scss/applications/editor/_editor.scss
RomainValls: You're right, done 👍 !

SharglutDev (approved):
Lgtm and tested

PR #8234, opened 2024-07-29
front: convert track offset to meters when stashing in the store

emersion:
_Includes a few cleanup commits before the actual bugfix._

The backend would choke with "java.lang.RuntimeException: The
waypoint is not located on the track section XXX" when creating
a train with an itinerary defined by clicking on tracks on the
map, and then editing that train without changing anything (edit
then save immediately).

https://github.com/OpenRailAssociation/osrd/commit/bc3c0028ca48ab2ce9611342115e87d83ed9ccab ("front: send track offset in mm in train schedule v2
requests") has modified our logic when sending path items to the
backend to convert meters to millimeters. However, when receiving
path items from the backend, we need to convert millimeters back
to meters as well. Adjust useSetupItineraryForTrainUpdate() to
perform this conversion.

### DIFF
--- a/front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
+++ b/front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
@@ -43,143 +43,141 @@ const useSetupItineraryForTrainUpdate = (
     osrdEditoastApi.endpoints.postV2InfraByInfraIdPathProperties.useMutation();
   useEffect(() => {
     const setupItineraryForTrainUpdate = async () => {
+      if (!infraId) {
+        return;
+      }
       const trainSchedule = await getTrainScheduleById({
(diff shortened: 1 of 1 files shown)


emersion:
Suggestions for reviewers: read commit by commit, and enable the "Hide whitespace" setting:

out

clarani (commented):
I have this bug: when selecting a path by clicking on the map, the pathfinding is not launched. You can reproduce here (near quiberon)
image

emersion:
Weird, still can't reproduce that bug with these waypoints…

Uriel-Sautron:
With the same points, I have no problems

Uriel-Sautron (approved):
LGTM !! Tested

clarani (approved):
LGTM, I can't reproduce the bug anymore :white_check_mark:

PR #8235, opened 2024-07-29
editoast, front: incompatible constraints in path finding

sim51:
Implements #7719 for editoast and frontend.

I discussed with Maelys about two things that have not be  done : 
- it's OK to display the path in green with the errors areas in orange
- it's OK if we can't see on the map the segment (blank tick separator), if we zoom in (don't know how to do that in map libre style with just LineString ...)


editoast/src/core/v2/pathfinding.rs
```diff
@@ -35,6 +40,36 @@ pub struct PathfindingRequest {
     pub rolling_stock_supported_signaling_systems: Vec<String>,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct RangeOffet {
```
younesschrifi:
```suggestion
pub struct RangeOffset {
```

editoast/src/core/v2/pathfinding.rs
```diff
@@ -35,6 +40,36 @@ pub struct PathfindingRequest {
...
+    start: u64,
+    end: u64,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct IncompatibleElectrification {
+    range: RangeOffet,
```
younesschrifi:
```suggestion
    range: RangeOffset,
```

younesschrifi (changes requested):
Thanks for this PR :)

editoast/src/core/v2/pathfinding.rs
sim51: thanks Younees !

front/public/locales/en/operationalStudies/manageTrainSchedule.json
```diff
@@ -53,6 +53,21 @@
...
+      "disable": "Hide constraints of type '{{name}}'"
+    },
+    "incompatible_electrification_ranges": "Incompatible electrification",
+    "incompatible_gauge_ranges": "Incompatible loading gauge",
+    "incompatible_signalisation_system_ranges": "Incompatible signaling system",
+    "info": "Pathfinding failed due to some constraints. Here you will find the list of constraints that failed on the shortest path, with for each, the type of constraint, the position of the start of the failure zone relative to the start of the path, and its length.",
+    "gotoMap": "See on the map",
+    "seeConstraintsOnMap": "See contraints on the map",
+    "title": "{{count}} incompatible contraints",
+    "title_on": "{{count}} on {{total}} incompatible contraintss"
```
bougue-pe:
```suggestion
    "title_on": "{{count}} on {{total}} incompatible constraints"
```

front/public/locales/en/operationalStudies/manageTrainSchedule.json
```diff
@@ -79,10 +94,8 @@
     "not_found_in_blocks": "Missing block",
     "not_found_in_routes": "Missing route",
     "not_found_in_tracks": "Missing track",
-    "incompatible_electrification": "Incompatible electrification",
-    "incompatible_loading_gauge": "Incompatible loading gauge",
-    "incompatible_signaling_system": "Incompatible signaling system",
     "not_enough_path_items": "Missing parameters for pathfinding: Departure, Arrival, Rolling stock",
+    "incompatible_constraints": "Incompatible constraints",
```
bougue-pe: nit: this could be bumped one line up to have the same order 🇫🇷 and 🇬🇧

front/public/locales/fr/operationalStudies/manageTrainSchedule.json
```diff
@@ -125,6 +138,7 @@
   "usingElectricalProfiles": "Utiliser les profils électriques",
   "via": "Étape",
   "vias": "Points de passage, arrêts",
+  "viewItineraryOnMap": "Voir l'iténéraire sur la carte",
```
bougue-pe:
```suggestion
  "viewItineraryOnMap": "Voir l'itinéraire sur la carte",
```

editoast/src/core/v2/pathfinding.rs
```diff
@@ -35,6 +40,36 @@ pub struct PathfindingRequest {
...
+pub struct RangeOffset {
+    start: u64,
+    end: u64,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct IncompatibleElectrification {
+    range: RangeOffset,
+    value: String,
+}
```
bougue-pe: Not sure if it's better for (front) typing, but how about just one structure like in core:
```rs
pub struct RangeValue {
    range: RangeOffset,
    value: Option<String>,
}
```

editoast/src/core/v2/pathfinding.rs
```diff
@@ -35,6 +40,36 @@ pub struct PathfindingRequest {
...
+
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct IncompatibleLoadingGauge {
```
bougue-pe: I'd add "Range" to the name (same for the 3)
```suggestion
pub struct IncompatibleLoadingGaugeRange {
```

editoast/src/core/v2/pathfinding.rs
sim51: Done

editoast/src/core/v2/pathfinding.rs
sim51: Done

editoast/src/core/v2/pathfinding.rs
```diff
@@ -35,6 +40,36 @@ pub struct PathfindingRequest {
     pub rolling_stock_supported_signaling_systems: Vec<String>,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct RangeOffset {
```
bougue-pe: nit (feel free to disregard and resolve): more 🇬🇧 IMO
```suggestion
pub struct OffsetRange {
```

editoast/src/core/v2/pathfinding.rs
```diff
@@ -35,6 +40,36 @@ pub struct PathfindingRequest {
...
+pub struct IncompatibleSignalingSystem {
+    range: RangeOffset,
+    value: String,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct IncompatibleConstraints {
+    incompatible_electrification_ranges: Option<Vec<IncompatibleElectrification>>,
+    incompatible_gauge_ranges: Option<Vec<IncompatibleLoadingGauge>>,
+    incompatible_signalisation_system_ranges: Option<Vec<IncompatibleSignalingSystem>>,
```
bougue-pe: Those are not nullable in core (empty list only): should we make it more consistent? And if so, should we change core or editoast (like in the suggestion)?
```suggestion
    incompatible_electrification_ranges: Vec<IncompatibleElectrification>,
    incompatible_gauge_ranges: Vec<IncompatibleLoadingGauge>,
    incompatible_signalisation_system_ranges: Vec<IncompatibleSignalingSystem>,
```

editoast/src/core/v2/pathfinding.rs
```diff
@@ -48,32 +83,9 @@ pub enum PathfindingResult {
...
-        #[schema(inline)]
-        blocks: Vec<Identifier>,
-        #[schema(inline)]
-        routes: Vec<Identifier>,
-        track_section_ranges: Vec<TrackRange>,
-        length: u64,
-        incompatible_ranges: Vec<(u64, u64)>,
+    IncompatibleConstraints {
+        relaxed_constraints_path: Box<PathfindingResultSuccess>,
+        incompatible_constraints: Box<IncompatibleConstraints>,
```
bougue-pe, line 81: It looks possible to avoid Boxing, no?
```suggestion
        relaxed_constraints_path: PathfindingResultSuccess,
        incompatible_constraints: IncompatibleConstraints,
```

front/public/locales/en/operationalStudies/manageTrainSchedule.json
```diff
@@ -53,6 +53,21 @@
...
+      "enable": "Display constraints of type'{{name}}'",
+      "disable": "Hide constraints of type '{{name}}'"
+    },
+    "incompatible_electrification_ranges": "Incompatible electrification",
+    "incompatible_gauge_ranges": "Incompatible loading gauge",
+    "incompatible_signalisation_system_ranges": "Incompatible signaling system",
+    "info": "Pathfinding failed due to some constraints. Here you will find the list of constraints that failed on the shortest path, with for each, the type of constraint, the position of the start of the failure zone relative to the start of the path, and its length.",
+    "gotoMap": "See on the map",
+    "seeConstraintsOnMap": "See contraints on the map",
+    "title": "{{count}} incompatible contraints",
```
bougue-pe:
```suggestion
    "title": "{{total}} incompatible constraints",
```

front/public/locales/en/operationalStudies/manageTrainSchedule.json
```diff
@@ -53,6 +53,21 @@
...
+      "title": "Filter on constraint's type",
+      "enable": "Display constraints of type'{{name}}'",
+      "disable": "Hide constraints of type '{{name}}'"
+    },
+    "incompatible_electrification_ranges": "Incompatible electrification",
+    "incompatible_gauge_ranges": "Incompatible loading gauge",
+    "incompatible_signalisation_system_ranges": "Incompatible signaling system",
+    "info": "Pathfinding failed due to some constraints. Here you will find the list of constraints that failed on the shortest path, with for each, the type of constraint, the position of the start of the failure zone relative to the start of the path, and its length.",
+    "gotoMap": "See on the map",
+    "seeConstraintsOnMap": "See contraints on the map",
```
bougue-pe:
```suggestion
    "seeConstraintsOnMap": "See constraints on the map",
```

front/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstraintsList.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { type FC, useMemo } from 'react';
+
+import { debounce, sortBy } from 'lodash';
+import { useMap } from 'react-map-gl/maplibre';
+
+import IncompatibleConstraintItem from './IncompatibleConstraintsItem';
+import type { IncompatibleConstraintItemEnhanced } from './type';
+
+interface IncompatibleContraintListProps {
```
bougue-pe:
```suggestion
interface IncompatibleConstraintListProps {
```

front/public/locales/fr/common/common.json
```diff
@@ -1,7 +1,9 @@
 {
   "actions": {
     "add": "Ajouter",
+    "collapse": "Reduire",
```
bougue-pe:
```suggestion
    "collapse": "Réduire",
```

editoast/src/core/v2/pathfinding.rs
bougue-pe: This doesn't match core and shadows the actual value (leading to always `null` list).
To be done in multiple places:
```suggestion
    incompatible_signaling_system_ranges: Option<Vec<IncompatibleSignalingSystem>>,
```

front/src/modules/pathfinding/components/IncompatibleConstraints/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+export function getIcon(type: string) {
+  switch (type) {
+    case 'incompatible_electrification_ranges':
+      return Zap;
+    case 'incompatible_gauge_ranges':
+      return Gauge;
+    case 'incompatible_signalisation_system_ranges':
+      return Signals;
+    default:
+      return Zap;
```
bougue-pe: How about a `?` here?

bougue-pe (changes requested):
Thanks for this work !
Tested on small-infra and France to get all the constraints (after suggested corrections it looks OK).
Did not review front.

editoast/src/core/v2/pathfinding.rs
sim51: Nop, I added them yesterday because at the compilation rust tells me to do something ...

editoast/src/core/v2/pathfinding.rs
sim51: I'm agree that those types should match at 100% the  core return type. 
So for now I remove the `Option` for now.

front/public/locales/en/operationalStudies/manageTrainSchedule.json
sim51: I reorder this section to follow the alpha one

editoast/src/core/v2/pathfinding.rs
sim51: Ok today I don't issue when I remove the box wrapper, so I removed it

sim51: It's not at the compilation, it's clippy the linter that says that :
```
error: the `Err`-variant returned from this function is very large
  --> src/views/v2/path/path_item_cache.rs:88:10
   |
88 |       ) -> TrackOffsetResult {
   |            ^^^^^^^^^^^^^^^^^
   |
  ::: src/core/v2/pathfinding.rs:86:5
   |
86 | /     IncompatibleConstraints {
87 | |         relaxed_constraints_path: PathfindingResultSuccess,
88 | |         incompatible_constraints: IncompatibleConstraints,
89 | |     },
   | |_____- the largest variant contains at least 176 bytes
   |
   = help: try reducing the size of `core::v2::pathfinding::PathfindingResult`, for example by boxing large elements or replacing it with `Box<core::v2::pathfinding::PathfindingResult>`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
   = note: `-D clippy::result-large-err` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::result_large_err)]`
  ```
  That's why I added the `Box` wrapper.

File: front/src/modules/pathfinding/components/IncompatibleConstraints/utils.ts (hunk 15 above)
Comment (sim51) [reply front/src/modules/pathfinding/components/IncompatibleConstraints/utils.ts]: I returned a default one because we need one (linter rule), and normally we should never use it. 
A `?` is not an Icon component

File: front/public/locales/en/operationalStudies/manageTrainSchedule.json [hunk 16]
```diff
@@ -53,6 +53,21 @@
...
   "inputOPTrigramsExample": "Ex: LSN CIG CCS",
+  "incompatibleConstraints": {
+    "filters": {
+      "title": "Filter on constraint's type",
+      "enable": "Display constraints of type'{{name}}'",
+      "disable": "Hide constraints of type '{{name}}'"
+    },
+    "incompatible_electrification_ranges": "Incompatible electrification",
+    "incompatible_gauge_ranges": "Incompatible loading gauge",
+    "incompatible_signalisation_system_ranges": "Incompatible signaling system",
```
bougue-pe: Same for 🇫🇷 and 🇬🇧 
```suggestion
    "incompatible_signaling_system_ranges": "Incompatible signaling system",
```

editoast/src/core/v2/pathfinding.rs
bougue-pe: Not sure what's best here, I will let an editoast maintainer respond (and possibly resolve), here are the ideas I have:
1. Let the Box as proposed
2. Box one level up in the enum (if doable)
3. Box in the 2 Results that trigger clippy (`src/views/v2/path/path_item_cache.rs:88` and `src/views/v2/path/pathfinding.rs:198`)
4. Allow those large Err in Results by adding `#[allow(clippy::result_large_err)]`

bougue-pe (approved):
Thanks, the feature is nice (not re-tested) !
No blocker for me anymore.

Curious about:
* the `Box<>` conversation ✅ 
* ~and if `yarn.lock` should be updated after the add of Question logo~ ✅ as discussed: no need, previously existing

younesschrifi (approved):
LGTM for editoast part :)

editoast/src/core/v2/pathfinding.rs
flomonster: Adding Box is a good option here. There's no reason why this enum should use the IncompatibleConstraints variant more often, so you should reduce its allocation.

flomonster (approved):
Editoast part looks good to me.

> [!IMPORTANT]
> The frontend has errors and the CI didn't catch it :thinking:

front/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstrainstFilters.tsx
```diff
@@ -0,0 +1,47 @@
...
+        .map(([name, type]) => {
+          const Icon = getIcon(name as IncompatibleConstraintTypes);
+          const label = t(`incompatibleConstraints.${name}`).toString();
+          return (
+            <button
+              key={name}
+              type="button"
+              title={
+                type.enabled
+                  ? t('incompatibleConstraints.filters.disable', { name: label })
```
SharglutDev: When I hover on an active filter, I see the related text. But the behavior of the filter changes depending of which filter are active. This message works only if I have all incompatible constraints on my path and if only 2 are active. For the other cases, clicking on the icon will actually keep only this filter (if all 3 are active) or activate the 2 others (if it's the last active filter).

Capture d’écran 2024-08-02 à 16 30 56

front/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstraintsItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const { t } = useTranslation(['operationalStudies/manageTrainSchedule']);
+  const Icon = getIcon(data.type);
+  let bgClass = 'bg-light';
+  if (isSelected === true) bgClass = 'bg-success text-white';
+  if (isHovered === true) bgClass = 'bg-info text-white';
+
+  return (
+    <div
+      className={cx(
+        'incompatible-constraints-item d-flex justify-content-around align-items-center my-2 rounded py-1 ',
```
SharglutDev, line 52: As discussed, even with the align-center, the text feels like it's more on top (probably to handle the "long letters"). Is this change ok for you ? It's also close to the behavior @thibautsailly tries to implement in the new osrd ui.

Capture d’écran 2024-08-02 à 16 45 36

```suggestion
        'incompatible-constraints-item d-flex justify-content-around align-items-center my-2 rounded pt-1 ',
```

front/src/common/Collapsable.tsx
```diff
@@ -0,0 +1,48 @@
...
+      className={cx(className, 'p-1')}
+      {...omit(props, ['children', 'defaultCollapsed', 'className'])}
+    >
+      <div className="collapse-header d-flex align-items-start">
+        <div className="flex-grow-1">{children[0]}</div>
+        <button
+          type="button"
+          className="flex-grow-0 px-1 pb-1 mx-1"
+          title={
+            isCollapsed ? t('common/common:actions.expand') : t('common/common:actions.collapse')
```
SharglutDev: Seems like the trad doesn't work with this syntax.

```suggestion
            isCollapsed ? t('actions.expand') : t('actions.collapse')
```

And `const { t } = useTranslation('common/common');` at the top

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -3443,6 +3443,22 @@ export type PathfindingResultSuccess = {
   /** Path description as track ranges */
   track_section_ranges: TrackRange[];
 };
+export type OffsetRange = {
+  end: number;
+  start: number;
+};
+export type IncompatibleOffsetRangeWithValue = {
+  range: OffsetRange;
+  value: string;
```
SharglutDev, line 3452: Capture d’écran 2024-08-02 à 17 27 47

Request for @bougue-pe : would it be possible to return "non-electrified" instead of an empty string when the range is non electrified ? In the `pathProperties` endpoint, in the `electrification` property, we have this : 

Capture d’écran 2024-08-02 à 17 30 48

It might be safer to do that instead of the front assuming that an empty string always means that it's a "non-electrified" behavior. What do you think ?

front/public/locales/en/operationalStudies/manageTrainSchedule.json
```diff
@@ -53,6 +53,21 @@
...
+      "title": "Filter on constraint's type",
+      "enable": "Display constraints of type'{{name}}'",
+      "disable": "Hide constraints of type '{{name}}'"
+    },
+    "incompatible_electrification_ranges": "Incompatible electrification",
+    "incompatible_gauge_ranges": "Incompatible loading gauge",
+    "incompatible_signaling_system_ranges": "Incompatible signaling system",
+    "info": "Pathfinding failed due to some constraints. Here you will find the list of constraints that failed on the shortest path, with for each, the type of constraint, the position of the start of the failure zone relative to the start of the path, and its length.",
+    "gotoMap": "See on the map",
+    "seeConstraintsOnMap": "See constraints on the map",
```
SharglutDev: Here you use `See` but for the itinerary l.140, you use `View`. Should we harmonize that ?

And maybe it should say `all constraints`.

front/public/locales/en/operationalStudies/manageTrainSchedule.json
```diff
@@ -53,6 +53,21 @@
   "infraLoading": "Infrastructure is loading…",
   "inputOPTrigrams": "Enter the trigrams of operational points you want to the path to go through, separated by a space.",
   "inputOPTrigramsExample": "Ex: LSN CIG CCS",
+  "incompatibleConstraints": {
+    "filters": {
+      "title": "Filter on constraint's type",
+      "enable": "Display constraints of type'{{name}}'",
```
SharglutDev:
```suggestion
      "enable": "Display constraints of type '{{name}}'",
```

front/public/locales/fr/operationalStudies/manageTrainSchedule.json
```diff
@@ -53,6 +53,21 @@
...
+  "incompatibleConstraints": {
+    "filters": {
+      "title": "Filtrer sur le type de contrainte",
+      "enable": "Afficher les contraintes de type '{{name}}'",
+      "disable": "Cacher les contraintes de type '{{name}}'"
+    },
+    "incompatible_electrification_ranges": "Électrification incompatible",
+    "incompatible_gauge_ranges": "Gabarit incompatible",
+    "incompatible_signaling_system_ranges": "Système de signalisation incompatible",
+    "info": "La recherche de chemin a échoué à cause de certaines contraintes. Vous trouverez ici la liste des contraintes ayant échoué sur le chemin le plus court, avec pour chacune, le type de contrainte, la position du début de la zone d’échec par rapport au début du chemin, et sa longueur. ​",
```
SharglutDev:
```suggestion
    "info": "La recherche de chemin a échoué à cause de certaines contraintes. Vous trouverez ici la liste des contraintes ayant échoué sur le chemin le plus court, avec pour chacune, le type de contrainte, la position du début de la zone d’échec par rapport au début du chemin, et sa longueur.",
```

front/public/locales/fr/operationalStudies/manageTrainSchedule.json
```diff
@@ -53,6 +53,21 @@
...
+    "gotoMap": "Voir sur la carte",
+    "seeConstraintsOnMap": "Voir la liste des contraintes incompatibles sur la carte",
```
SharglutDev: It's not really the list that we will see when clicking here.

```suggestion
    "seeConstraintsOnMap": "Voir toutes les contraintes incompatibles sur la carte",
```

Other suggestion (you can adapt english trad and the `center on itinerary` if you chose this one)

```suggestion
    "seeConstraintsOnMap": "Centrer sur toutes les contraintes incompatibles",
```

SharglutDev (changes requested):
Thank you for this great PR ! Didn't review the code yet, here are some comments after testing the feature.

front/src/common/api/generatedEditoastApi.ts
bougue-pe: This may be ambiguous no matter what string is chosen as the "no-electrification" representation.
Introducing an option would be the better choice here (so null in that case).
This raises 2 questions I think:
1. Improving this also comes with design questions: how to display clearly the "missing electrification" case?
2. Internally in core this is not a cost we want to pay in this PI (handling options internally might be an important issue with `DistanceRangeMap`). IMO, we should track it as an improvement issue, but I'm not sure that this is going to be judged reasonable in the end: opened https://github.com/OpenRailAssociation/osrd/issues/8327

front/public/locales/en/operationalStudies/manageTrainSchedule.json
sim51: It can't be "all constraints", because it focus on the not filtered constraints (ie. those that are displayed on the boxed). 
About the `view` vs `see` I'm agree that we need to choose one.

front/public/locales/fr/operationalStudies/manageTrainSchedule.json
sim51: You will see the filtered list on the map, not everything.  So for me it can't be `toutes les contraintes`. 
We can just say `Voir les contraintes`.

front/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstraintsItem.tsx
sim51: If you change the `py-1` by `pt-1`, the whole inner item will be changed, and not only the text.
I  did some changes, and it's better now

sim51:
@SharglutDev  I fixed all you comments
BTW, I fixed also the order in the list

About the `The frontend has errors and the CI didn't catch it ` , it's certainly due to some dependencies that have been updated in dev,  and  match the defined version in the package.json.
The CI build from scratch, so it has not this issue.

At the end of this review, I have a rebase to do with the dev branch

front/src/common/api/generatedEditoastApi.ts
bougue-pe: Reply to self 😅 :
1. The display is just "writing it" (from the mockup)
un-electrified

And as for the `"non_electrified"` string it is not the same as an option (or enum in the electrification properties).
We could change the interface, but the internal core model would still be dirty, which is more contained and similar to electrification property.
We could reuse the enum from pathProperties (maybe the same for other constraints if possible).
Leading to a bit heavier but more consistent responses and mutualized code.

bougue-pe: As discussed with @sim51 : No change in back for the moment, matching in front that empty string to display non-electrified ranges as expected.
Then the issue #8327 previously created would force migration of all this because of the type change.

SharglutDev: Ok fine by me :)

sim51:
About the "no electrification", I managed it in the front. If the value is an empty string, I display the "no electrification"  message

front/src/utils/strings.ts
```diff
@@ -66,3 +68,28 @@ export function convertInputStringToNumber(str: string) {
 export function formatUicToCi(uic: number) {
   return uic.toString().replace(/^87/, '');
 }
+
+/**
+ * Given a distance in millimeters, returns a human readable string.
+ * Exemples:
```
SharglutDev:
```suggestion
 * Examples:
```

front/src/utils/strings.ts
```diff
@@ -66,3 +68,28 @@ export function convertInputStringToNumber(str: string) {
...
+  if (Math.floor(meters) > 0) {
+    label = `${round(meters, 0)}m`;
+  }
+
+  const kms = distance / 1000 / 1000;
+  if (Math.floor(kms) > 0) {
+    label = `${round(kms, 2)}Km`;
+  }
+  return label;
+}
```
SharglutDev: It's not a heavy function, but maybe we can optimize it like that instead of sometimes passing two times in the condition by checking km first ? (I also added some utils we have for conversions)

```suggestion
export function distanceToHumanReadable(distance: number): string {
  if (!distance) return '0';

  const kms = mmToKm(distance);
  if (Math.floor(kms) > 0) return `${round(kms, 2)}Km`;

  const meters = mmToM(distance);
  if (Math.floor(meters) > 0) return `${round(meters, 0)}m`;

  return `${distance}mm`;
}
```

front/public/locales/fr/operationalStudies/manageTrainSchedule.json
```diff
@@ -53,6 +53,23 @@
   "infraLoading": "Infra en cours de chargement…",
   "inputOPTrigrams": "Entrez les trigrammes des points remarquables par lesquels vous voulez que l'itinéraire se fasse, séparés par un espace.",
   "inputOPTrigramsExample": "Ex : LSN CIG CCS",
+  "incompatibleConstraints": {
+    "filters": {
+      "title": "Filtrer sur le type de contrainte",
+      "enable": "Afficher les contraintes de type '{{name}}'",
+      "disable": "Cacher les contraintes de type '{{name}}'",
+      "focus": "Afficher unqieuement les contraintes de type '{{name}}'"
```
SharglutDev:
```suggestion
      "focus": "Afficher uniquement les contraintes de type '{{name}}'"
```

front/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstrainstFilters.tsx
SharglutDev: Better after your fix, but there is still one issue : 

Capture d’écran 2024-08-05 à 17 53 49

Here if I click on the signal icon, it will activate all filters again but it says "hide signaling filters"

front/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstrainstFilters.tsx
SharglutDev, line 1: Don't think in was on the sketch but the current number of constraints displayed is pretty nice !
Only one thing, when we pass from all filters to one filter and reverse, the list width is resizing, it's not very clean for the user I think. 

https://github.com/user-attachments/assets/a13ff3bf-6cb1-410b-a4d3-b11d85ee23a2

We could have a fix width based on the larger size (but might take too much area for nothing when the user don't filter) or do you prefer to leave it like that ?

front/src/common/api/generatedEditoastApi.ts
SharglutDev, line 3452: This could have been like this @bougue-pe ?

```suggestion
export type IncompatibleOffsetRangeWithValue = IncompatibleOffsetRange & {
  value: string;
};
```

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -3443,6 +3443,22 @@ export type PathfindingResultSuccess = {
   /** Path description as track ranges */
   track_section_ranges: TrackRange[];
 };
+export type OffsetRange = {
+  end: number;
+  start: number;
+};
```
SharglutDev, line 3449: @bougue-pe : not related to this PR but with this type, we could now use it in `ElectrificationRange` and `SimulationPowerRestrictionRange`.

front/src/common/Collapsable.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { useCallback, useEffect, useState, type FC, type HTMLAttributes } from 'react';
+
+import { ChevronDown, ChevronUp } from '@osrd-project/ui-icons';
+import cx from 'classnames';
+import { omit } from 'lodash';
+import { useTranslation } from 'react-i18next';
+
+interface CollapsableProps extends Omit<HTMLAttributes<unknown>, 'children' | 'collapsed'> {
```
SharglutDev, line 8: There can be a `collapsed` attribute in a regular html element ?

front/src/common/Collapsable.tsx
```diff
@@ -0,0 +1,46 @@
...
+        <button
+          type="button"
+          className="flex-grow-0 px-1 pb-1 mx-1"
+          title={isCollapsed ? t('actions.expand') : t('actions.collapse')}
+          onClick={toggle}
+        >
+          {isCollapsed ? <ChevronUp /> : <ChevronDown />}
+        </button>
+      </div>
+      {isCollapsed === false && <div className="collapse-header">{children[1]}</div>}
```
SharglutDev:
```suggestion
      {!isCollapsed && <div className="collapse-header">{children[1]}</div>}
```

front/src/common/Collapsable.tsx
```diff
@@ -0,0 +1,46 @@
...
+  }, []);
+
+  useEffect(() => {
+    setIsCollapsed(collapsed);
+  }, [collapsed]);
+
+  return (
+    <div
+      className={cx(className, 'p-1')}
+      {...omit(props, ['children', 'defaultCollapsed', 'className'])}
```
SharglutDev: Do we need `defaultCollapsed` ?

front/src/common/Collapsable.tsx
```diff
@@ -0,0 +1,46 @@
...
+import cx from 'classnames';
+import { omit } from 'lodash';
+import { useTranslation } from 'react-i18next';
+
+interface CollapsableProps extends Omit<HTMLAttributes<unknown>, 'children' | 'collapsed'> {
+  children: [JSX.Element, JSX.Element];
+  collapsed: boolean;
+}
+const Collapsable: FC<CollapsableProps> = (props) => {
+  const { children, collapsed, className } = props;
```
SharglutDev: Wouldn't it be possible to initialize it to false so we don't have to always pass it in props ?

```suggestion
  const { children, collapsed = false, className } = props;
```

front/src/common/Collapsable.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const { t } = useTranslation('common/common');
+  const [isCollapsed, setIsCollapsed] = useState(collapsed);
+
+  const toggle = useCallback(() => {
+    setIsCollapsed((prev) => !prev);
+  }, []);
+
+  useEffect(() => {
+    setIsCollapsed(collapsed);
+  }, [collapsed]);
```
SharglutDev, line 23: not sure why we need this ? Initializing `isCollapsed` state with `collapsed` is not enough ?

front/src/common/Collapsable.tsx
```diff
@@ -0,0 +1,46 @@
...
+  collapsed: boolean;
+}
+const Collapsable: FC<CollapsableProps> = (props) => {
+  const { children, collapsed, className } = props;
+  const { t } = useTranslation('common/common');
+  const [isCollapsed, setIsCollapsed] = useState(collapsed);
+
+  const toggle = useCallback(() => {
+    setIsCollapsed((prev) => !prev);
+  }, []);
```
SharglutDev, line 19: What is the difference between doing this or do this in the button's `onClick` ?

```suggestion
  onClick={() => {
    setIsCollapsed((prev) => !prev);
  }}
```

front/src/modules/pathfinding/hook/usePathfinding.ts
```diff
@@ -284,16 +301,33 @@ export const usePathfindingV2 = (
...
+                    allWaypoints,
+                    length: pathResult.length,
+                    incompatibleConstraints: pathfindingResult.incompatible_constraints,
+                    trackSectionRanges: pathResult.track_section_ranges,
+                  });
+                pathfindingDispatch({
+                  type: 'PATHFINDING_INCOMPATIBLE_CONSTRAINTS',
+                  message: `pathfindingErrors.${pathfindingResult.status}`,
+                });
+              }
```
SharglutDev: Maybe we could do like this instead of using two time the `setPathProperties` ?

```suggestion
              if (setPathProperties)
                setPathProperties({
                  electrifications,
                  geometry,
                  suggestedOperationalPoints,
                  allWaypoints,
                  length: pathResult.length,
                  incompatibleConstraints:
                    pathfindingResult.status === 'incompatible_constraints'
                      ? pathfindingResult.incompatible_constraints
                      : undefined,
                  trackSectionRanges: pathResult.track_section_ranges,
                });

              if (pathfindingResult.status === 'incompatible_constraints') {
                pathfindingDispatch({
                  type: 'PATHFINDING_INCOMPATIBLE_CONSTRAINTS',
                  message: `pathfindingErrors.${pathfindingResult.status}`,
                });
              } else {
                pathfindingDispatch({ type: 'PATHFINDING_FINISHED' });
              }
```

front/src/applications/operationalStudies/types.ts
```diff
@@ -95,6 +96,7 @@ export type ManageTrainSchedulePathProperties = {
   allWaypoints: SuggestedOP[];
   length: number;
   trackSectionRanges: NonNullable<PathfindingResultSuccess['track_section_ranges']>;
+  incompatibleConstraints?: Partial<IncompatibleConstraints>;
```
SharglutDev: Seems like the endpoint always returns all 3 constraints but they are empty arrays if constraint is not present

```suggestion
  incompatibleConstraints?: IncompatibleConstraints;
```

SharglutDev (changes requested):
Not finished yet, but here are a first batch of comments

flomonster:
I had this white page:

image

Console log:
```
The above error occurred in the <IncompatibleConstraintItem> component:

IncompatibleConstraintItem@http://localhost:4000/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstraintsItem.tsx:18:36
IncompatibleConstraintsList@http://localhost:4000/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstraintsList.tsx:15:37
div
div
div
Collapsable@http://localhost:4000/src/common/Collapsable.tsx:18:48
IncompatibleConstraints@http://localhost:4000/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx:27:33
div
div
Map3@http://localhost:4000/node_modules/.vite/deps/chunk-3ECQ5EEX.js?v=d24acd2c:1416:12
Map@http://localhost:4000/src/modules/trainschedule/components/ManageTrainSchedule/Map.tsx:60:13
div
div
Tab@http://localhost:4000/src/common/Tabs.tsx:13:13
div
div
Tabs@http://localhost:4000/src/common/Tabs.tsx:23:14
ManageTrainScheduleV2@http://localhost:4000/src/applications/operationalStudies/views/v2/ManageTrainScheduleV2.tsx:37:31
div
div
div
div
main
ScenarioV2@http://localhost:4000/src/applications/operationalStudies/views/v2/ScenarioV2.tsx:42:22
Scenario@http://localhost:4000/src/applications/operationalStudies/views/Scenario.tsx:18:49
RenderedRoute@http://localhost:4000/node_modules/.vite/deps/react-router-dom.js?v=2a33eb75:3563:7
Outlet@http://localhost:4000/node_modules/.vite/deps/react-router-dom.js?v=2a33eb75:3931:20
ModalProvider@http://localhost:4000/src/common/BootstrapSNCF/ModalSNCF/ModalProvider.tsx:75:34
OsrdContextLayout@http://localhost:4000/src/common/osrdContext.tsx:104:34
RenderedRoute@http://localhost:4000/node_modules/.vite/deps/react-router-dom.js?v=2a33eb75:3563:7
RenderErrorBoundary@http://localhost:4000/node_modules/.vite/deps/react-router-dom.js?v=2a33eb75:3520:5
DataRoutes@http://localhost:4000/node_modules/.vite/deps/react-router-dom.js?v=2a33eb75:4673:7
Router@http://localhost:4000/node_modules/.vite/deps/react-router-dom.js?v=2a33eb75:3945:7
RouterProvider@http://localhost:4000/node_modules/.vite/deps/react-router-dom.js?v=2a33eb75:4490:7
Suspense
App@http://localhost:4000/src/main/app.jsx:194:22
PersistGate2@http://localhost:4000/node_modules/.vite/deps/redux-persist_integration_react.js?v=02663751:90:20
Provider@http://localhost:4000/node_modules/.vite/deps/chunk-6HQ3C4LF.js?v=d24acd2c:1094:18
Container

React will try to recreate this component tree from scratch using the error boundary you provided, RenderErrorBoundary.
```

front/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstrainstFilters.tsx
sim51: In your scenario it is due to the change of the title.
I think we have the same issue on the error box on the editor. 

One solution can be to change the design, and make something like that 
```
4 incompatible constrainsts
on a total of 8 
```
But it changes the y axis this time, by adding a new line when you filter.

Don't know what is better. So I prefer to let it like that, and have a discussion with Thibaut later.

front/src/common/Collapsable.tsx
sim51: I don't know all HTML attributes, but I do know what I defined in my function. 
So I prefer to omit them (just to be sure)

front/src/common/Collapsable.tsx
sim51: Nop.  The init is just done at mount. 
If you change its value in the upper component, the value will not be changed.

front/src/common/Collapsable.tsx
sim51: If you do that, at every render you redefine the function, so you change its ref, so you retrigger the render of the children component.

I'm ok that here it's not really needed, the function is really tiny, and it's a `button` component.
But it cost nothing to do it like that.

front/src/common/Collapsable.tsx
sim51: You're right, I changed the name of the attribut, but if forgot to change it here.
Thanks !

front/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstrainstFilters.tsx
SharglutDev: Fine by me :)

front/src/common/api/generatedEditoastApi.ts
bougue-pe: After discussion, it probably means using `#[serde(flatten)]` on the range in rust, and maybe a change in core too.
I'll try to check it (today, but the first one trying just drops a word to avoid multiple work).

front/src/common/api/generatedEditoastApi.ts
bougue-pe: Yes, it's a step-by-step work, in both core and editoast, and it's spotted for core (MRSP probably changing soon), so any remark/reminder on that is welcome.
I would not require it in this PR (I will have a look to add those 2 also in the PR changing MRSP).

front/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstrainstFilters.tsx
sim51: Technically, when you click on it you disable it, but if  everything  is disabled,  it is equal to everything is enable.

We already have : 
- focus on it 
- disable it
- enable it

I'm not sure it will be more clear for the user if I add a fourth label saying "click here to enable all constraints".
What wording do you want to have here ?
We can also use  "toggle the filter 'X'"   ("Changer l'état du filtre 'X'") in all cases

front/src/common/api/generatedEditoastApi.ts
SharglutDev: Ok for me :)

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+    if (!pathProperties?.geometry || filteredConstraint.length === 0)
+      return featureCollection<LineString, { ids: string[] }>([]);
+    return getSegmentsConstraints(pathProperties?.geometry as LineString, filteredConstraint);
+  }, [pathProperties?.geometry, filteredConstraint]);
+
+  if (total === 0) return null;
+  return (
+    <>
+      <Collapsable
+        collapsed={false}
```
SharglutDev:
```suggestion
```

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+interface IncompatibleConstraintsProps {
+  pathProperties?: ManageTrainSchedulePathProperties;
+}
+
+const IncompatibleConstraints: FC<IncompatibleConstraintsProps> = ({ pathProperties }) => {
+  const { t } = useTranslation(['operationalStudies/manageTrainSchedule']);
+  const map = useMap();
+
+  const [types, setTypes] = useState<Record<string, { count: number; enabled: boolean }>>({});
```
SharglutDev: To easily identify the meaning of the state, could we rename it with something like that ?

I think we can also improve the type of the state for the keys (need some adapts in the rest of the file).

And maybe we can add a type for this state in types.ts as we reuse it elsewhere.

```suggestion
  const [incompatibleConstraintsDictionnary, setIncompatibleConstraintsDictionnary] = useState<Record<IncompatibleConstraint, { count: number; enabled: boolean }>>({});
```

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+    };
+  }, [map]);
+
+  // When pathProperties changes
+  //  => reset state
+  useEffect(() => {
+    const data =
+      pathProperties?.geometry && pathProperties?.incompatibleConstraints
+        ? pathProperties.incompatibleConstraints
+        : undefined;
```
SharglutDev: Instead of that, maybe we could just add at top of the useEffect : 
`if (!pathProperties?.geometry || !pathProperties?.incompatibleConstraints) return`

front/src/modules/pathfinding/components/IncompatibleConstraints/type.ts
SharglutDev, line 1: Could you rename the file with `types.ts`

front/src/modules/pathfinding/components/IncompatibleConstraints/type.ts
```diff
@@ -0,0 +1,25 @@
...
+  IncompatibleConstraints,
+  IncompatibleOffsetRange,
+  IncompatibleOffsetRangeWithValue,
+} from 'common/api/osrdEditoastApi';
+
+export type IncompatibleConstraintItemType =
+  | IncompatibleOffsetRange
+  | IncompatibleOffsetRangeWithValue;
+
+export type IncompatibleConstraintTypes = keyof IncompatibleConstraints;
```
SharglutDev: We try to avoid using `types` in a type name 

```suggestion
export type IncompatibleConstraint = keyof IncompatibleConstraints;
```

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
SharglutDev, line 1: Could you rename this file `IncompatibleConstraints` so we can find it easily with a ctrl+p. 
And then you can create an index.ts to facilitate the imports (see RollingStockSelector)

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+}
+
+const IncompatibleConstraints: FC<IncompatibleConstraintsProps> = ({ pathProperties }) => {
+  const { t } = useTranslation(['operationalStudies/manageTrainSchedule']);
+  const map = useMap();
+
+  const [types, setTypes] = useState<Record<string, { count: number; enabled: boolean }>>({});
+  const [selected, setSelected] = useState<Set<string>>(new Set([]));
+  const [hovered, setHovered] = useState<Set<string>>(new Set([]));
+  const [total, setTotal] = useState<number>(0);
```
clarani:
```suggestion
  const [total, setTotal] = useState(0);
```

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+  const [constraints, setConstraints] = useState<IncompatibleConstraintItemEnhanced[]>([]);
+
+  // Function to toggle a type for the filters
+  const toggleType = useCallback((name: string) => {
```
clarani: - either remove the comment or use a js comment
- maybe rename it `toggleTypeFilter`, do you think it is clearer ?
```suggestion
  // Function to toggle a type for the filters
  const toggleTypeFilter = useCallback((name: string) => {
```

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+  }, [map]);
+
+  // When pathProperties changes
+  //  => reset state
+  useEffect(() => {
+    const data =
+      pathProperties?.geometry && pathProperties?.incompatibleConstraints
+        ? pathProperties.incompatibleConstraints
+        : undefined;
+    const nextTypes: Record<string, { count: number; enabled: boolean }> = toPairs(
```
SharglutDev: Replace also the type with new one created in types.ts (the one from the state)

```suggestion
    const formattedIncompatibleConstraints: NewCreatedType = toPairs(
```

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+  const [constraints, setConstraints] = useState<IncompatibleConstraintItemEnhanced[]>([]);
+
+  // Function to toggle a type for the filters
+  const toggleType = useCallback((name: string) => {
+    setTypes((prev) => {
+      const result = { ...prev };
+
+      const filtersName = toPairs(prev)
+        .filter(([, value]) => value.count)
+        .map(([key]) => key);
```
clarani: You can move this in a useMemo outside the function, we don't need to compute it each time we click on a type

front/src/common/api/generatedEditoastApi.ts
bougue-pe: I had a look to see if it's easily/cleanly doable: yes for editoast (`#[serde(flatten)]`), no for core 🫤
In core we would need to either:
* re-declare the `range` part without sharing
* tweak the moshi serializer
* abandon the `data` class that can't be inherited
* ... didn't think of any other possibility?

All in all, it bloats core way more than the gain in editoast and front, so I'd let it like that

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+      (acc, [key, value]) => ({ ...acc, [key]: { count: (value ?? []).length, enabled: true } }),
+      {}
+    );
+    setTypes(nextTypes);
+
+    const turfLength = pathProperties?.geometry
+      ? length(feature(pathProperties?.geometry as LineString), { units: 'millimeters' })
+      : 0;
+    const ratio = turfLength / (pathProperties?.length || turfLength);
+    const nextConstraints = toPairs(data)
```
SharglutDev: We do two times `toPairs(data)` in this useEffect. Could you stock this in a const at the beginning of the useEffect ?

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+      .filter(([, value]) => !isNil(value))
```
SharglutDev: Related to the type we changed, constraints can't be null or undefined but will be empty array if a constraint isn't present. We should rather check for that in this filter.

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+  // Function to toggle a type for the filters
+  const toggleType = useCallback((name: string) => {
+    setTypes((prev) => {
+      const result = { ...prev };
+
+      const filtersName = toPairs(prev)
+        .filter(([, value]) => value.count)
+        .map(([key]) => key);
+      const nbEnabledBefore = toPairs(prev).filter(
```
clarani:
```suggestion
      const nbEnabledBefore = Object.values(prev).filter(
```

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+        ([, value]) => value.count > 0 && value.enabled === true
+      ).length;
+
+      // If everything is selected, then we only select the provided type
+      if (filtersName.length === nbEnabledBefore) {
+        filtersName
+          .filter((filter) => filter !== name)
+          .forEach((e) => {
+            result[e].enabled = false;
+          });
```
clarani: You can avoid a for loop like this
```suggestion
        filtersName.forEach((constraint) => {
          if (constraint !== name) {
            result[constraint].enabled = false;
          }
        });
```

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+      } else {
+        result[name].enabled = !result[name].enabled;
+      }
+
+      // if on the result, nothing is selcted, then everything is selected
```
clarani: Typo
```suggestion
      // if on the result, nothing is selected, then everything is selected
```

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+      const nbEnabledAfter = toPairs(result).filter(
```
clarani:
```suggestion
      const nbEnabledAfter = Object.keys(result).filter(
```

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+        ([, value]) => value.count > 0 && value.enabled === true
```
clarani: Maybe you can create a small utils function

SharglutDev (changes requested):
Would you mind to add some spacing between some blocks of code to help readability ? :)

(Still some files to check)

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
SharglutDev: Or maybe you could reuse nextTypes you just defined and filter constraints with `count > 0` ?

front/src/common/api/generatedEditoastApi.ts
SharglutDev: Ok I understand. So for the other cases where we use this syntax in openApi, it just depends on editoast ?

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+  pathProperties?: ManageTrainSchedulePathProperties;
+}
+
+const IncompatibleConstraints: FC<IncompatibleConstraintsProps> = ({ pathProperties }) => {
+  const { t } = useTranslation(['operationalStudies/manageTrainSchedule']);
+  const map = useMap();
+
+  const [types, setTypes] = useState<Record<string, { count: number; enabled: boolean }>>({});
+  const [selected, setSelected] = useState<Set<string>>(new Set([]));
+  const [hovered, setHovered] = useState<Set<string>>(new Set([]));
```
SharglutDev: To improve readability ?

```suggestion
  const [selectedConstraint, setSelectedConstraint] = useState<Set<string>>(new Set([]));
  const [hoveredConstraint, setHoveredConstraint] = useState<Set<string>>(new Set([]));
```

SharglutDev (changes requested):

front/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstrainstFilters.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { Filter } from '@osrd-project/ui-icons';
+import cx from 'classnames';
+import { toPairs } from 'lodash';
+import { useTranslation } from 'react-i18next';
+
+import type { IncompatibleConstraintTypes } from './type';
+import { getIcon } from './utils';
+
+interface IncompatibleConstraintsFiltersProps extends HTMLAttributes<unknown> {
+  data: Record<IncompatibleConstraintTypes, { count: number; enabled: boolean }>;
```
SharglutDev: To replace by the state type you created

```suggestion
  incompatibleConstraintsDictionnary: StateType;
```

front/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstrainstFilters.tsx
```diff
@@ -0,0 +1,56 @@
...
+  toggle: (type: string) => void;
```
SharglutDev:
```suggestion
  toggleFilter: (type: string) => void;
```

front/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstrainstFilters.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const getTitle = useCallback(
+    (label: string, enabled: boolean) => {
+      if (allSelected) {
+        return t('incompatibleConstraints.filters.focus', { name: label });
+      }
+      return enabled
+        ? t('incompatibleConstraints.filters.disable', { name: label })
+        : t('incompatibleConstraints.filters.enable', { name: label });
+    },
+    [t, allSelected]
```
SharglutDev: Maybe you can merge these. And here, we could also have check when only one filter is enable ?

```suggestion
  const allSelected = toPairs(data).every(([, e]) => e.enabled);

  const getTitle = useCallback(
    (label: string, enabled: boolean) => {
    const allSelected = Object.values(data).every((value) => value.enabled);
      if (allSelected) {
        return t('incompatibleConstraints.filters.focus', { name: label });
      }
      
      if (oneEnabled) return t('incompatibleConstraints.filters.activateAll'
      
      return enabled
        ? t('incompatibleConstraints.filters.disable', { name: label })
        : t('incompatibleConstraints.filters.enable', { name: label });
    },
    [t, data]
```

front/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstrainstFilters.tsx
SharglutDev: Maybe here we could only pass the data (constraints) for the one who have `count > 0` no ? The other are not displayed in the filter list and we can't select them anyway.

front/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstrainstFilters.tsx
```diff
@@ -0,0 +1,56 @@
...
+  );
+
+  return (
+    <div {...htmlAttrs}>
+      <Filter title={t('incompatibleConstraints.filters.title')} />
+      {toPairs(data)
+        .filter(([, type]) => type.count > 0)
```
SharglutDev: If you chose to pass to this component only the constraints with count > 0 (cf previous comment), you can use the props directly.

If not, I'm not sure if you think it's necessary but maybe we can extract this in a useMemo to avoid recomputing this each time the constraints list is rendered ?

```suggestion
const validFilters = useMemo(
    () => Object.entries(data).filter(([, type]) => type.count > 0),
    [data]
  );
```

front/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstraintsLayer.tsx
```diff
@@ -0,0 +1,64 @@
...
+): FilterSpecification | undefined {
+  return [
+    predicate,
+    ...Array.from(ids).map((id) => ['in', id, ['get', 'ids']]),
+  ] as FilterSpecification;
+}
+interface IncompatibleConstraintsLayerProps {
+  data: FeatureCollection<LineString, { ids: string[] }>;
+  selected: Set<string>;
+  hovered: Set<string>;
```
SharglutDev: To improve readability ?

```suggestion
  filteredConstraintsGeojson: FeatureCollection<LineString, { ids: string[] }>;
  selectedConstraintsIds: Set<string>;
  hoveredConstraintsIds: Set<string>;
```

front/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstraintsList.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import IncompatibleConstraintItem from './IncompatibleConstraintsItem';
+import type { IncompatibleConstraintItemEnhanced } from './type';
+
+interface IncompatibleConstraintListProps {
+  data: Array<IncompatibleConstraintItemEnhanced>;
+  hovered: Set<string>;
+  selected: Set<string>;
+  onHover: (itemId: string | null) => void;
+  onSelect: (itemId: string) => void;
```
SharglutDev: To improve readability

```suggestion
  formattedConstraints: IncompatibleConstraintItemEnhanced[];
  hoveredConstraintsIds: Set<string>;
  selectedConstraintsIds: Set<string>;
  onConstraintHover: (constraintId: string | null) => void;
  onConstraintSelect: (constraintId: string) => void;
```

front/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstraintsItem.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { getIcon } from './utils';
+
+interface IncompatibleConstraintItemProps {
+  data: IncompatibleConstraintItemEnhanced;
+  isHovered?: boolean;
+  isSelected?: boolean;
+  onEnter: () => void;
+  onLeave: () => void;
+  onClick: () => void;
+  gotoMap: () => void;
```
SharglutDev: Same renaming as in the List component for readability

SharglutDev (changes requested):
If you don't mind, could you also add a extra line between interfaces and the begining of the component in your new components :)

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
sim51: I renamed it `toggleFilter`

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
sim51: If I use a `useMemo`, it will depend on the filter state, which changes at each click, so we will have the same effect. 

What I should do is to add a new state variable to store it, and compute it in the "reset state" useEffect (~ L:130). 

This will add more complexity in the component just for specific "Object.keys()" function, which will be called on an object with 3 keys. I'm pretty  sure it will not be more performant.

front/src/common/api/generatedEditoastApi.ts
bougue-pe: For most of the cases I checked, yes (no direct serializing of something provided by core).
The only counter-example I found is `CompleteReportTrain` (`ReportTrainV2`) which uses inheritance in core, and wouldn't take advantage of a data class.

front/src/modules/pathfinding/components/IncompatibleConstraints/IncompatibleConstraintsList.tsx
sim51: Adding `constraints` everywhere for me it's no-sense. We are already on a folder with this name, with a component with it, ... 
On a button on react you don't have `onButtonClick` just `onClick`. The component name give you the context.
IMHO, it doesn't improve the readability but instead spams my vision with `Contrainst` everywhere.

front/src/utils/strings.ts
```diff
@@ -66,3 +69,24 @@ export function convertInputStringToNumber(str: string) {
...
+ * - 1927520 -> 1.92Km
+ *
+ * @param distance the distance in millimeters
+ * @returns A human readable string of the distance
+ */
+export function distanceToHumanReadable(distance: number): string {
+  if (!distance) return '0';
+
+  const kms = mmToKm(distance);
+  if (Math.floor(kms) > 0) return `${round(kms, 2)}Km`;
```
SharglutDev:
```suggestion
  if (Math.floor(kms) > 0) return `${round(kms, 2)}km`;
```

SharglutDev (changes requested):

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
sim51: If I do an early return here, I need to set all state variables twice, firstly here in the early return where I will put static default value (ex: `setTotal(0)`) , and also at the end of the useEffect.
Because it's the 'reset' useEffect, I prefer to have one place to set the initial state, even if you do some "loops" on empty arrays.

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
sim51: I removed the filter, indeed it should not be null. 
For others comments, `nextConstraints` is the flatten list of constraints, so there is no `count > 0` here. I don't understand.

front/src/modules/pathfinding/components/IncompatibleConstraints/index.tsx
SharglutDev: Ok got it

SharglutDev (approved):
Lgtm and tested, good job and good luck for the rebase !

PR #8236, opened 2024-07-29
front: prevent point selection on STDCM maps

### LINKED ISSUE #8201: stdcm: the page jumps if you click on the input map, if the output map is present
### What happened?

If you have both input and output map visible in the stdcm page, clicking somewhere on the input map makes the page jump to show this point but into the output map

https://github.com/user-attachments/assets/89fa0a31-5885-4992-984d-8b263d22f0d1

I expect the map not to jump to the output map, and stay focused on the input map

1. Go to STDCM page
2. Create a simulation successfully so the output map appears
3. Click somewhere on the input the map 
4. Witness

Recette (SNCF)

Firefox

7faec4a

kmer2016:
close #8201


SarahBellaha (approved):
LGTM ✔️

clarani (approved):

PR #8240, opened 2024-07-29
editoast: make core pathfinding/simulation errors non-fatal

emersion:
core may throw an HTTP error when a pathfinding or simulation fails, e.g. when a track offset is past the end of a track. In that case, don't fail the whole batch, only fail a single pathfinding/simulation result.

### DIFF
--- a/editoast/src/views/v2/path/pathfinding.rs
+++ b/editoast/src/views/v2/path/pathfinding.rs
@@ -177,7 +177,12 @@ async fn pathfinding_blocks_batch(
     let computed_paths: Vec<_> = futures::future::join_all(futures)
         .await
         .into_iter()
-        .collect::<Result<_>>()?;
+        .map(|res| match res {
+            Ok(pathfinding_result) => pathfinding_result,
+            // TODO: only make HTTP status code errors non-fatal
+            Err(core_error) => PathfindingResult::PathfindingFailed { core_error },
+        })
+        .collect();
 
     for (index, computed_path) in computed_paths.into_iter().enumerate() {
         let path_index = pathfinding_requests_index[index];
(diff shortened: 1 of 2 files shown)


woshilapin (approved):
Nice improvement.

Castavo:
Aren't there cases where the core client would give us an error and we would wish to bubble it up as a 500 ?

For example communication errors with the core server ?

emersion:
That's a good point. I thought that case would be handled separately but it turns out that's not the case.

Unfortunately that doesn't seem easy to fix: `CoreClient::handle_error` converts the original `StandardCoreError` into an `InternalError`. It doesn't seem like there is a good way to differentiate between a 500 sent by the core HTTP server and a 500 due to a network error (or really any other kind of error).

Maybe we could use `InternalError.context` here? Stash some info there in `CoreClient::handle_error`, check it in pathfidning/simulation? That said, it doesn't seem like this is a great approach reading the discussion in https://github.com/OpenRailAssociation/osrd/pull/8241#issuecomment-2258540157?

Any suggestions?

leovalais:
I don't know how we would do this in the current state of things. (There's an ongoing effort to remove `InternalError` only to keep this error format in the views.) When #8048 will be merged, the error cause field will be serialized (https://github.com/OpenRailAssociation/osrd/pull/8048/files#diff-8679b01de12fad7a910082070f3a8c83e80d7e444ed5a0063c579b1e45a32324). We may be able to use this field as an indicator to whether the error should be fatal or not (needs confirmation).

In the meantime, since we're not losing track of the error, I'd say it's fine to merge this :)

emersion:
Alright! I've added TODOs accordingly.

PR #8241, opened 2024-07-29
editoast: populate InternalError status/context when unparsable

emersion:
Add more metadata to the InternalError when it's not a proper JSON-encoded StandardCoreError.

### DIFF
--- a/editoast/src/core/mod.rs
+++ b/editoast/src/core/mod.rs
@@ -88,14 +88,15 @@ impl CoreClient {
         url: String,
     ) -> InternalError {
         // We try to deserialize the response as an StandardCoreError in order to retain the context of the core error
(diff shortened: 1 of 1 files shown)


woshilapin (approved):
Thank you.

leovalais:
Thanks. This module is being rewritten in #8048. We'll likely be able to keep this change in the new version, however since we'll communicate with core using message queues, the `url` part of the request has become an implementation detail. We probably don't want to rely on that field of the error. What are you needing this change for?

emersion:
Nothing in particular TBH, just manual inspection and extracting metadata from errors while debugging. Happy to close this PR if this chunk of code is on the chopping block anyways.

leovalais:
Oh no, it's fine. My point was that we shouldn't rely on the `url` programmatically, but like you said it's always useful for debugging! 😄

PR #8247, opened 2024-07-30
front: fix scheduled points and stops in imported trains

### LINKED ISSUE #8104: Edition of imported trains doesn't work properly
### What happened?

The times and stops table is not filled when editing an imported train schedule (from open data or a file).

Example: 
This train schedule contains a lot of **scheduled points** and **stops**.
image

The table should be filled with the values.

1. Create a scenario
2. Import this [example_train.json](https://github.com/user-attachments/files/16281391/example_train.json)
3. Edit the `831701` train
4. Go to times and stops
5. The table is empty

Local

_No response_

67d50c4166542bd74e2303c4aa285b858882047f

achrafmohye:
Closes #8104

### DIFF
--- a/front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
+++ b/front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
@@ -100,7 +100,8 @@ const useSetupItineraryForTrainUpdate = (
             (suggestedOp) =>
               'uic' in step &&
               suggestedOp.uic === step.uic &&
-              suggestedOp.ch === step.secondary_code
+              // When importing train from open data or from files, secondary_code might not always exist
+              (!step.secondary_code || suggestedOp.ch === step.secondary_code)
           );
 
           const correspondingSchedule = trainSchedule.schedule?.find(


RomainValls:
After importing the exampleTrain json, when I'm editing the train, the Time and Stops table is still empty.

SharglutDev (changes requested):
Thank you for the PR. When testing, the bug is still there.

When we edit a train, all the logic to update the store is in the hook `useSetupItineraryForTrainUpdate`, we don't call the `usePathfinding` hook.

So when updating pathSteps (l.90), you can adapt the `correspondingOp` check with something similar you used in this PR :

```suggestion
   const correspondingOp = suggestedOperationalPoints.find(
                  (suggestedOp) =>
                    'uic' in step &&
                    suggestedOp.uic === step.uic &&
                    (!step.secondary_code || suggestedOp.ch === step.secondary_code)
                );
```

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -92,7 +92,8 @@ const useSetupItineraryForTrainUpdate = (
                   (suggestedOp) =>
                     'uic' in step &&
                     suggestedOp.uic === step.uic &&
-                    suggestedOp.ch === step.secondary_code
+                    // If 'secondary_code' is not in step, return true; if 'secondary_code' is in step, check if suggestedOp.ch matches step.secondary_code
+                    (!step.secondary_code || suggestedOp.ch === step.secondary_code)
```
SharglutDev: Rather than explaining the code, I would suggest this, is that ok for you ?

```suggestion
                    // When importing train from open data or from files, secondary_code might not always exist
                    (!step.secondary_code || suggestedOp.ch === step.secondary_code)
```

SharglutDev (changes requested):

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
achrafmohye: Sounds good to me, Fixed.

RomainValls (approved):
LGTM and tested 👍

SharglutDev (approved):
Lgtm and tested. You just have one conflict to handle after rebasing (don't hesitate to ask if it's hard to fix)

PR #8250, opened 2024-07-30
front: add linter rule to prevent generatedEditoastApi imports

emersion:
The generated file is not meant to be used directly: we override a few auto-generated pieces in osrdEditoastApi.ts.

* * *

Alternatively, we could maybe leverage [`allowTypeImports`](https://typescript-eslint.io/rules/no-restricted-imports/#allowtypeimports) to allow type imports and disallow regular imports, since we don't override any type in `osrdEditoastApi.ts`.

### DIFF
--- a/front/.eslintrc
+++ b/front/.eslintrc
@@ -81,7 +81,11 @@
     "react/prefer-stateless-function": "off",
     "react/static-property-placement": 0,
     // disable vitest/prefer-to-be because it does not authorize toEqual for the floats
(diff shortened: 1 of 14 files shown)


clarani (approved):
LGTM :white_check_mark:

theocrsb (approved):
LGTM and tested !

SharglutDev (approved):
Lgtm

PR #8251, opened 2024-07-30
core: change error to 400 when waypoint has invalid offset

eckter:


core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -264,12 +264,9 @@ private fun getTrackSectionChunkOnWaypoint(
...
         waypointOffset in startChunk..endChunk
     }
-        ?: throw RuntimeException(
-            String.format(
-                "The waypoint is not located on the track section %s",
-                rawInfra.getTrackSectionName(trackSectionId)
-            )
-        )
+        ?: throw OSRDError(ErrorType.InvalidWaypointLocation)
+            .withContext("track", rawInfra.getTrackSectionName(trackSectionId))
```
Erashin, line 268: What happens on the editoast side if you throw this exception without adding the context as you do here? I have a feeling there will be a bug, because data will be missing, and I'm assuming that's why we create specific methods such as `newInvalidRollingStockError`. Could you check?

eckter: This data isn't currently used, but it's forwarded up to the front-end. If we want to display a nicely formatted and translated message that includes this info, we can (but I don't think it's worth the trouble here). It would also be visible when checking the response content.

It's just info that seems relevant, nothing really happens if it's missing

Erashin (approved):

PR #8256, opened 2024-07-30
stdcm: simulation report map is not showing the path

### LINKED ISSUE #8016: stdcm: simulation report map is not showing the path
### What happened?

I have found a solution but the path is not shown in the pdf:
image

See the path in the map screenshot

1. create a stdcm path
2. download the report

- **Browser**: Firefox 

staging de8ada8

### LINKED ISSUE #8198: stcdm result pdf: map no included
### What happened?

[STDCM-0724-586-828.pdf](https://github.com/user-attachments/files/16378710/STDCM-0724-586-828.pdf)

https://github.com/user-attachments/assets/63771bff-9fd6-4c91-98de-007647a07d69

If you're moving the map while the result arrives, you get no map inside the result pdf

I expect what I'm doing with map to not impact what's included in the pdf

1. Go to stdcm page
2. Pick any path
3. Launch a simulation and move the map (keep the click held) while the result isn't there
4. Download the pdf 
5. Witness

Recette (SNCF)

Firefox

7faec4a

### LINKED ISSUE #8620: Path is missing in STDCM V2 simulation sheet map
### What happened?

The map isn't showing the path:
[STDCM-0824-672-610.pdf](https://github.com/user-attachments/files/16800045/STDCM-0824-672-610.pdf)

It should show the path in yellow-ish

.

Recette (SNCF)

Google Chrome

80973c5

SarahBellaha:
Closes #8016
Closes #8198 
Closes #8620

### DIFF
--- a/front/src/applications/stdcmV2/views/StdcmViewV2.tsx
+++ b/front/src/applications/stdcmV2/views/StdcmViewV2.tsx
@@ -136,7 +136,6 @@ const StdcmViewV2 = () => {
           <StdcmConfig
             selectedSimulation={selectedSimulation}
             currentSimulationInputs={currentSimulationInputs}
-            pathProperties={pathProperties}
             isPending={isPending}
             showBtnToLaunchSimulation={showBtnToLaunchSimulation}
             retainedSimulationIndex={retainedSimulationIndex}
--- a/front/src/modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/ItineraryMarkersV2.tsx
+++ b/front/src/modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/ItineraryMarkersV2.tsx
@@ -33,6 +33,11 @@ type MarkerInformation = {
     }
 );
 
+type ItineraryMarkersProps = {
+  map: Map;
+  simulationPathSteps?: PathStep[];
+};
+
 const formatPointWithNoName = (
   lineCode: number,
   lineName: string,
@@ -78,11 +83,14 @@ const extractMarkerInformation = (pathSteps: (PathStep | null)[]) =>
     return acc;
   }, [] as MarkerInformation[]);
 
-const ItineraryMarkers = ({ map }: { map: Map }) => {
+const ItineraryMarkers = ({ map, simulationPathSteps }: ItineraryMarkersProps) => {
   const { getPathSteps } = useOsrdConfSelectors();
   const pathSteps = useSelector(getPathSteps);
 
-  const markersInformation = useMemo(() => extractMarkerInformation(pathSteps), [pathSteps]);
+  const markersInformation = useMemo(
+    () => extractMarkerInformation(simulationPathSteps || pathSteps),
+    [simulationPathSteps, pathSteps]
+  );
 
   const getMarkerDisplayInformation = useCallback(
     (markerInfo: MarkerInformation) => {
--- a/front/src/modules/trainschedule/components/ManageTrainSchedule/Map.tsx
+++ b/front/src/modules/trainschedule/components/ManageTrainSchedule/Map.tsx
@@ -55,6 +55,7 @@ import RenderPopup from 'modules/trainschedule/components/ManageTrainSchedule/Ma
 import { updateViewport } from 'reducers/map';
 import type { Viewport } from 'reducers/map';
 import { getMap, getTerrain3DExaggeration } from 'reducers/map/selectors';
+import type { PathStep } from 'reducers/osrdconf/types';
 import { useAppDispatch } from 'store';
 import { getMapMouseEventNearestFeature } from 'utils/mapHelper';
 
@@ -64,17 +65,23 @@ import ItineraryMarkersV2 from './ManageTrainScheduleMap/ItineraryMarkersV2';
 type MapProps = {
   pathProperties?: ManageTrainSchedulePathProperties;
   setMapCanvas?: (mapCanvas: string) => void;
+  isReadOnly?: boolean;
   hideAttribution?: boolean;
   hideItinerary?: boolean;
   preventPointSelection?: boolean;
+  mapId?: string;
+  simulationPathSteps?: PathStep[];
 };
 
 const Map: FC<PropsWithChildren<MapProps>> = ({
   pathProperties,
   setMapCanvas,
+  isReadOnly = false,
   hideAttribution = false,
   hideItinerary = false,
   preventPointSelection = false,
+  mapId = 'map-container',
+  simulationPathSteps,
   children,
 }) => {
   const mapBlankStyle = useMapBlankStyle();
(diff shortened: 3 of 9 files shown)


front/src/modules/trainschedule/components/ManageTrainSchedule/Map.tsx
```diff
@@ -56,26 +57,38 @@ import ItineraryMarkersV2 from './ManageTrainScheduleMap/ItineraryMarkersV2';
 type MapProps = {
   pathProperties?: ManageTrainSchedulePathProperties;
   setMapCanvas?: (mapCanvas: string) => void;
+  /** prop to display a lite version of the map for stdcm simulation results,
+   meaning that we won't have any interaction and that the map will be centered on the path */
+  isLiteVersion?: boolean;
```
emersion: Maybe `isReadOnly` would be a more descriptive name?

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
```diff
@@ -41,7 +47,7 @@ const StdcmViewV2 = () => {
         {scenarioID && (
           <StdcmConfig
             currentSimulationInputs={currentSimulationInputs}
-            pathProperties={pathProperties}
+            // pathProperties={pathProperties}
```
emersion: Remnant from the past?

front/src/styles/scss/applications/stdcmV2/_home.scss
```diff
@@ -169,16 +176,26 @@
...
       justify-content: space-between;
+      .map-results {
+        border: 0.063rem solid rgba(255, 255, 255, 1);
+        border-radius: 0.5rem;
+        box-shadow:
+          0rem 0rem 0rem 0.125rem rgba(255, 255, 255, 0.75) inset,
+          0rem 0rem 0rem 0.063rem rgba(0, 0, 0, 0.25) inset;
+        height: 33.75rem;
+        margin: 2rem 2rem 0 0.85rem;
+        width: 100%;
```
emersion: Would probably be a tad better to use `flex-grow: 1` instead of `width: 100%` since the parent is a flex container.

emersion: TBH it's a bit strange to me to make the map width adaptive, I would've expected the table on the left side to be `flex-grow: 1` and the map to have a fixed size?

front/src/applications/stdcmV2/components/StdcmConfig.tsx
```diff
@@ -45,7 +43,8 @@ const StdcmConfig = ({
   const { updateGridMarginAfter, updateGridMarginBefore, updateStdcmStandardAllowance } =
     useOsrdConfActions() as StdcmConfSliceActions;
 
-  const { pathfindingState } = usePathfindingV2(undefined, pathProperties);
+  const [pathProperties, setPathProperties] = useState<ManageTrainSchedulePathProperties>();
+  const { pathfindingState } = usePathfindingV2(setPathProperties, pathProperties);
```
emersion: Hm, I think this state is too high up: the path properties may be different for each simulation result. Maybe there are 3 simulation results and each have their own different origin/destination.

(Of course, in this branch there is only a single simulation result, but I think it'd make sense to put the state at the correct level from the start.)

emersion (commented):
Seems like this TypeScript error pops up:

```
 [TypeScript] Type '{ hideAttribution: true; hideItinerary: true; preventPointSelection: true; pathProperties: ManageTrainSchedulePathProperties | undefined; }' is not assignable to type 'IntrinsicAttributes & MapProps'. Property 'preventPointSelection' does not exist on type 'IntrinsicAttributes & MapProps'.

/app/src/applications/stdcmV2/components/StdcmConfig.tsx:115:44
```

By spamming clicks on the map markers (when the cursor appears as a hand), I can get it in a weird state where the markers are offset:

out

(Could maybe be fixed with `pointer-events: none`?)

front/src/modules/trainschedule/components/ManageTrainSchedule/Map.tsx
```diff
@@ -64,27 +65,39 @@ import ItineraryMarkersV2 from './ManageTrainScheduleMap/ItineraryMarkersV2';
 type MapProps = {
   pathProperties?: ManageTrainSchedulePathProperties;
   setMapCanvas?: (mapCanvas: string) => void;
+  isReadOnly?: boolean;
```
SharglutDev, line 68: What is the difference with `preventPointSelection` ? Maybe we can replace it with `isReadOnly` ?

front/src/modules/trainschedule/components/ManageTrainSchedule/Map.tsx
```diff
@@ -206,60 +217,63 @@ const Map: FC<PropsWithChildren<MapProps>> = ({
...
         attributionControl={false} // Defined below
-        onClick={onFeatureClick}
-        onResize={(e) => {
-          updateViewportChange({
-            width: e.target.getContainer().offsetWidth,
-            height: e.target.getContainer().offsetHeight,
-          });
-        }}
+        {...(!isReadOnly && {
+          dragPan: true,
```
SharglutDev, line 265: Works perfectly (and I also agree that there is no point to play with the map on this view) but as it was not really what @Castavo asked to solve the issue, maybe you should check with him that this fix is ok :)

front/src/modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/ItineraryMarkersV2.tsx
```diff
@@ -78,11 +78,20 @@ const extractMarkerInformation = (pathSteps: (PathStep | null)[]) =>
...
   }, [] as MarkerInformation[]);
 
-const ItineraryMarkers = ({ map }: { map: Map }) => {
+const ItineraryMarkers = ({
+  map,
+  simulationPathSteps,
+}: {
+  map: Map;
+  simulationPathSteps?: PathStep[];
+}) => {
```
SharglutDev: As we know have two props, could you extract them in a `ItineraryMarkersProps` type ? :)

front/src/modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/ItineraryMarkersV2.tsx
```diff
@@ -78,11 +78,20 @@ const extractMarkerInformation = (pathSteps: (PathStep | null)[]) =>
...
   const { getPathSteps } = useOsrdConfSelectors();
   const pathSteps = useSelector(getPathSteps);
 
-  const markersInformation = useMemo(() => extractMarkerInformation(pathSteps), [pathSteps]);
+  const markersInformation = useMemo(
+    () => extractMarkerInformation(simulationPathSteps || pathSteps),
```
SharglutDev, line 91: Not sure to understand why we might need `simulationPathSteps` ? We don't pass it to the map from StdcmConfig but informations seem to be the same as in StdcmResults.

SharglutDev (changes requested):
Thank you for this PR ! Tested and it works great. Left some comments.

front/src/modules/trainschedule/components/ManageTrainSchedule/Map.tsx
SharglutDev: isReadOnly also blocks the map interactivity for zoom and moving on it so we don't want that for the map in the config.

front/src/modules/trainschedule/components/ManageTrainSchedule/Map.tsx
SharglutDev: Nevermind for that, it works as expected

front/src/modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/ItineraryMarkersV2.tsx
SharglutDev: After discussions, we need it in case we change some inputs in the config so it doesn't affect the markers in the results map

SharglutDev (approved):
Lgtm and tested, good job for the fix !

emersion (approved):
LGTM, thanks! (Sorry for the delay!)

PR #8257, opened 2024-07-30
editoast: add stdcm search environment model and endpoints

### LINKED ISSUE #8168: editoast: create STDCM search env data model and CRUD
## Data Model
```rust
struct StdcmSearchEnvironment{
    infra_id: i64,
    work_schedule_group_id: Option<i64>,
    timetable_id: i64,
    simulation_window_begin: NaiveDateTime,
    simulation_window_end: NaiveDateTime
}
```

## CRUD
* POST, for creation, delete existing envs.
* GET, no id, returns the latest object that exists, based on simulation_window_begin and end.
   If no object exists, returns a 500

We add a CLI to create an environment, and we add scripts like for railjson import and such.

## How do we handle the fact that the env references other objects?
In the DB: forbid the deletion of the infra and timetable that are associated.
People will get 500s, it's okay, in the end it will be prevented by permissions added on objects if all goes well.

Castavo:
close #8168


editoast/src/modelsv2/stdcm_search_environment.rs
```diff
@@ -0,0 +1,199 @@
...
+        use crate::tables::stdcm_search_environment::dsl::*;
+        diesel::delete(stdcm_search_environment)
+            .execute(conn)
+            .await?;
+        Ok(())
+    }
+}
+
+impl StdcmSearchEnvironmentChangeset {
+    pub async fn over_write(self, conn: &mut DbConnection) -> Result<StdcmSearchEnvironment> {
```
woshilapin: I believe overwrite is one word.

```suggestion
    pub async fn overwrite(self, conn: &mut DbConnection) -> Result<StdcmSearchEnvironment> {
```

editoast/src/views/stdcm/stdcm_search_environment.rs
```diff
@@ -0,0 +1,243 @@
...
+
+#[utoipa::path(
+    post, path = "",
+    tag = "stdcm_search_environment",
+    request_body = StdcmSearchEnvironmentCreateForm,
+    responses(
+        (status = 201, body = StdcmSearchEnvironment),
+    )
+)]
+async fn over_write(
```
woshilapin: Same here, I believe `overwrite` is one word, not 2. That said, if the only way to insert a new `StdcmSearchEnvironment` always overwrite the existing one, I don't see why we have a `retrieve_latest` which would suppose we could e able to get another one?

woshilapin (approved):
I'm not sure that I understand completely how it will be used. In the description of the migration, it's mentioned that the object will almost be a singleton. And most of the code does seems to reflect that. But the name of the function `retrieve_latest` almost gives you a hint that there might be other instances of the object ("if there is a latest, it's relative to others"). It might just be a naming problem.

Apart from that, very clean PR with lot of tests, thank you 🎉

Castavo:
@woshilapin when designing this implem with @flomonster we considered making the sql table an actual singleton. 
However, there is no builtin in way to achieve this, the most commonly found way is to add a column, add a UNIQUE constraint on it, and a value constraint so it can take only one value. 

So we opted to not enforce the singleton-ness.

Thus, if by some manual manipulation, the database has several rows, the "retrieve" function handles it gracefully.

What do you think ? 
If we stay with this current model, I can add a clearer docstring on why `retrieve_latest` isn't called `retrieve`

Khoyo:
From the original goal of https://github.com/osrd-project/osrd-confidential/issues/481, don't we want to support multiple environments and select the correct one based on the requested date ? This seems to override all  the envs every time a POST is made

Castavo:
@Khoyo with @flomonster we discussed that there actually probably wasn't a need to keep history. We could decide to actually keep history and add indexes on the time windows fields.

However, given that there are no "ON DELETE" clauses on the fields (so the env can't be accidentally deleted), all objects referenced by the history wouldn't be deletable.

PR #8260, opened 2024-07-30
core: delay departure time in stdcm to make the train pass closer to first planned step arrival time

### LINKED ISSUE #7690: core: stdcm: post-processing: when we can ajust departure time, move it to the specified times
When we reach the destination with a "max departure time delay" higher than 0, we can adjust the departure time within an interval. When that happens, we should look for expected passage times for all points and pick the time that fits the most.

When there are several points with specified times, we may go for the simplest solution

Erashin:
Fixes #7690.

Once an stdcm solution is found, evaluate the different planned steps: the first planned step where the train arrives before planned arrival time, delay departure time as much as possible to make it pass through planned step at the closest time to planned arrival time.

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
@@ -106,16 +106,17 @@ data class STDCMNode(
         // ------------[-----|-----------|----------------|----------]------------
         //               currentTime            currentTimeWithMaxDelay
         if (plannedTimingData != null) {
-            val timeDiff =
-                getRealTime(currentTotalAddedDelay) - plannedTimingData.arrivalTime.seconds
+            val realTime = getRealTime(currentTotalAddedDelay)
+            val timeDiff = plannedTimingData.getTimeDiff(realTime)
             val relativeTimeDiff = plannedTimingData.getBeforeOrAfterRelativeTimeDiff(timeDiff)
             // If time diff is positive, adding delay won't decrease relative time diff: return
             // relativeTimeDiff
             if (timeDiff >= 0) return relativeTimeDiff
 
+            val maxTime = getRealTime(currentMaximumDelay)
             val maxTimeDiff =
                 min(
-                    getRealTime(currentMaximumDelay) - plannedTimingData.arrivalTime.seconds,
+                    plannedTimingData.getTimeDiff(maxTime),
                     plannedTimingData.arrivalTimeToleranceAfter.seconds
                 )
             val relativeMaxTimeDiff =
@@ -134,7 +135,7 @@ data class STDCMNode(
      * Takes into account the real current departure time shift to return the real current time at
      * which the train arrives at this node.
      */
-    private fun getRealTime(currentTotalAddedDelay: Double): Double {
+    fun getRealTime(currentTotalAddedDelay: Double): Double {
         assert(currentTotalAddedDelay >= totalPrevAddedDelay)
         return time + (currentTotalAddedDelay - totalPrevAddedDelay)
     }
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
@@ -25,6 +25,8 @@ import fr.sncf.osrd.utils.units.meters
 import io.opentelemetry.api.trace.SpanKind
 import io.opentelemetry.instrumentation.annotations.WithSpan
 import java.util.*
+import kotlin.math.abs
+import kotlin.math.min
 
 /**
  * This class contains all the static methods used to turn the raw pathfinding result into a full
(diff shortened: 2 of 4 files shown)


core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
```diff
@@ -130,6 +129,15 @@ data class STDCMNode(
         return null
     }
 
+    /**
+     * Get time difference between the time at which the train really passes through the node and
+     * the planned time at which it should have arrived at the node.
+     */
+    fun getTimeDiffWithPlannedArrival(currentTotalAddedDelay: Double): Double {
+        assert(plannedTimingData != null)
```
eckter: Same as `getBeforeOrAfterRelativeTimeDiff`: could be moved to `PlaneedTimingData` to avoid the explicit null check here

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
```diff
@@ -120,106 +122,142 @@ class STDCMPostProcessing(private val graph: STDCMGraph) {
...
+        return nodes
+    }
+
+    private fun getPlannedNodes(nodes: List<STDCMNode>): List<STDCMNode> {
+        return nodes.filter { it.plannedTimingData != null }
+    }
+
+    private fun getMaxNodeDelay(nodes: List<STDCMNode>): Double {
+        return nodes.minOfOrNull { it.maximumAddedDelay + it.totalPrevAddedDelay }!!
+    }
```
eckter, line 176: That method returns the maximum *total* added delay (compared to the min start time). But it's used to cap the delay *on top* of the one that's already added to avoid conflicts

It should probably just be `nodes.last().maximumAddedDelay` given how it's used

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
```diff
@@ -120,106 +122,142 @@ class STDCMPostProcessing(private val graph: STDCMGraph) {
...
+        var delayForPlannedNodes = 0.0
+        for (plannedNode in plannedNodes) {
+            val timeDiff = plannedNode.getTimeDiffWithPlannedArrival(addedDelay)
+            // Train passes before planned node: adding delay can make the train pass closer to
+            // planned arrival time
+            if (timeDiff < 0.0) {
+                delayForPlannedNodes = abs(timeDiff)
+                break
+            }
+        }
```
eckter: Do we really need a loop? We can just pick any value here

If `timeDiff >= 0.0` for any node we can just not do anything (adding delay would add difference for at least one node)

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
```diff
@@ -120,106 +122,142 @@ class STDCMPostProcessing(private val graph: STDCMGraph) {
...
-    while (i < edges.size) {
-        val edge = edges[i]
-        val start = edge.envelopeStartOffset
-        var length = edge.length
-        while (i + 1 < edges.size) {
-            val nextEdge = edges[i + 1]
-            if (edge.block != nextEdge.block) break
-            length += nextEdge.length.distance
+    /** Builds the list of block ranges, merging the ranges on the same block */
+    private fun makeBlockRanges(edges: List<STDCMEdge>): List<PathfindingEdgeRangeId<Block>> {
```
eckter, line 204: The indentation change makes it hard to review: I'll assume that nothing change in methods that aren't directly relevant. If there's other changes please let me know

core/src/test/kotlin/fr/sncf/osrd/stdcm/STDCMPathfindingTests.kt
```diff
@@ -672,4 +672,77 @@ class STDCMPathfindingTests {
...
+                    PlannedTimingData(plannedStepArrivalTime.seconds, 10.seconds, 10.seconds)
+                )
+                .setEndLocations(setOf(EdgeLocation(blocks[0], Offset(100.meters))))
+                .setUnavailableTimes(occupancyGraph)
+                .run()!!
+        assertEquals(maxPossibleDelay, res.departureTime)
+    }
+
+    /**
+     * Train goes through start window without added delay, after planned arrival time: no delay
```
eckter, line 730: We could add a test with a lot of added delay right from the start

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
Erashin: Nope just moved 'em into the class, no other changes. Didn't see why we'd need the private methods outside of the class tbh.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
Erashin: Would work as well. Just to be sure, you're proposing the following: we take the first planned node, if `timeDiff >= 0` we add nothing (even if further along the way we arrive before a planned node's arrival time)?
I'll also add a corresponding unit test if so.

core/src/test/kotlin/fr/sncf/osrd/stdcm/STDCMPathfindingTests.kt
Erashin: Don't quite get this.
Is this a new test? If so, what behaviour would it be testing?
Or just the same test, because you'd rather have an occupancy at the start rather than a hack negative planned arrival time? (honestly thought it was fine, we have a bit of leeway with test set up in my opinion).

eckter: New test. I'm commenting there because I noticed that every comment starts with `without added delay`.

Just checking that everything behaves properly when we already have delay and then add some more. Especially regarding the `getMaxNodeDelay` comments, if we're wrong there the test would likely show it

eckter: In fact I have a specific test case in mind, I could write it

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
Erashin: This should be correct. This value gives the maximum possible delay that can be added to the start time without encountering any sort of conflict. I compare this value to the edge delay added + additional delay i want to add for the planned timing data.

eckter: Nevermind I misread line 141

core/src/test/kotlin/fr/sncf/osrd/stdcm/STDCMPathfindingTests.kt
Erashin: Oh you mean check that we actually add the edge delay and the planned delay. Yeah ok I see the test. It's fine i can prolly write it ;)

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
Khoyo: BTW, there is an option to hide whitespace changes on github ;)

image

eckter (approved):

PR #8266, opened 2024-07-31
Use new speed limit tags endpoint

### LINKED ISSUE #7975: Front: add speedLimitTags from description file to propositions (simulation & infra edition)
After the creation of an endpoint `/speed_limit_tags` to expose the list of known speedLimitTags in OSRD's config (https://github.com/OpenRailAssociation/osrd/issues/7976), it has to be used by the front.

In order of priority, enrich (merge with the list from infra) the list of possible speedLimitTags:
1. when creating a simulation (operation studies or STDCM)
2. when editing speedLimits in an infrastructure
3. when picking speedLimitTag for map layers in map viewer and infra editor

EDIT: alphabetical sorting should be applied everywhere if possible.

NB: **No** deduplication of "identical" long-name/short-id is expected (long-names will be removed soon)

Side-note: So far, no use of the long names (probably available in the endpoint) is done by the front, so they are not exposed in the endpoint.

Part of https://github.com/osrd-project/osrd-confidential/issues/364

Uriel-Sautron:
### Use the new endpoint `/speed_limit_tags` in front:

  - Operational Studies
  - STDCM
  - Map
  - Infra Editor

Close #7975

### DIFF
--- a/front/src/applications/editor/components/LayersModal.tsx
+++ b/front/src/applications/editor/components/LayersModal.tsx
@@ -1,6 +1,17 @@
 import React, { type FC, useEffect, useMemo, useState } from 'react';
 
-import { groupBy, mapKeys, mapValues, sum, isString, isArray, uniq, isNil } from 'lodash';
+import {
+  groupBy,
+  mapKeys,
+  mapValues,
+  sum,
+  isString,
+  isArray,
+  uniq,
+  isNil,
+  concat,
+  compact,
+} from 'lodash';
 import { useTranslation } from 'react-i18next';
 import { GiElectric, GiUnplugged } from 'react-icons/gi';
 import { MdSpeed } from 'react-icons/md';
@@ -69,13 +80,19 @@ const LayersModal: FC<LayersModalProps> = ({
   const { t } = useTranslation();
   const { layersSettings } = useSelector(getMap);
   const [selectedLayers, setSelectedLayers] = useState<Set<Layer>>(initialLayers);
+
   const infraID = useInfraID();
 
-  const { data: speedLimitTags } =
+  const { data: speedLimitTagsByInfraId } =
     osrdEditoastApi.endpoints.getInfraByInfraIdSpeedLimitTags.useQuery(
       { infraId: infraID as number },
       { skip: isNil(infraID) }
     );
+  const { data: speedLimitTags } = osrdEditoastApi.endpoints.getSpeedLimitTags.useQuery();
+
+  const allSpeedLimitTags = compact(concat(speedLimitTags, speedLimitTagsByInfraId));
+  const allSpeedLimitTagsOrdered = useMemo(() => allSpeedLimitTags.sort(), [allSpeedLimitTags]);
+
   const DEFAULT_SPEED_LIMIT_TAG = useMemo(() => t('map-settings:noSpeedLimitByTag'), [t]);
   const selectionCounts = useMemo(
     () =>
@@ -125,8 +142,8 @@ const LayersModal: FC<LayersModalProps> = ({
   );
 
   const speedLimitOptions = useMemo(
-    () => uniq([DEFAULT_SPEED_LIMIT_TAG, ...(speedLimitTags || [])]),
-    [speedLimitTags]
+    () => uniq([DEFAULT_SPEED_LIMIT_TAG, ...allSpeedLimitTagsOrdered]),
+    [allSpeedLimitTagsOrdered]
   );
 
   return (
@@ -193,7 +210,7 @@ const LayersModal: FC<LayersModalProps> = ({
             id="speedLimitTag"
             className="form-control"
             value={layersSettings.speedlimittag || DEFAULT_SPEED_LIMIT_TAG}
-            disabled={!isArray(speedLimitTags) || !selectedLayers.has('speed_sections')}
+            disabled={!isArray(allSpeedLimitTags) || !selectedLayers.has('speed_sections')}
             onChange={(e) => {
               const newTag = e.target.value !== DEFAULT_SPEED_LIMIT_TAG ? e.target.value : null;
               dispatch(
--- a/front/src/applications/editor/tools/rangeEdition/components/RangeEditionLeftPanel.tsx
+++ b/front/src/applications/editor/tools/rangeEdition/components/RangeEditionLeftPanel.tsx
@@ -1,6 +1,6 @@
-import React, { useContext, useEffect, useState } from 'react';
+import React, { useContext, useEffect, useMemo, useState } from 'react';
 
-import { cloneDeep, isEmpty, isEqual, last, pick, uniqWith } from 'lodash';
+import { cloneDeep, compact, concat, isEmpty, isEqual, last, pick, uniq, uniqWith } from 'lodash';
 import { useTranslation } from 'react-i18next';
 
 import EntityError from 'applications/editor/components/EntityError';
@@ -107,13 +107,18 @@ const RangeEditionLeftPanel = () => {
     { skip: !infraID }
   );
 
-  const { data: speedLimitTags } =
+  const { data: speedLimitTagsByInfraId } =
     osrdEditoastApi.endpoints.getInfraByInfraIdSpeedLimitTags.useQuery(
       {
         infraId: infraID as number,
       },
       { skip: !infraID }
     );
+
+  const { data: speedLimitTags } = osrdEditoastApi.endpoints.getSpeedLimitTags.useQuery();
+
+  const allSpeedLimitTags = uniq(concat(speedLimitTags, speedLimitTagsByInfraId));
+  const allSpeedLimitTagsOrdered = useMemo(() => allSpeedLimitTags.sort(), [allSpeedLimitTags]);
 
   const updateSpeedSectionExtensions = (
     extensions: SpeedSectionEntity['properties']['extensions']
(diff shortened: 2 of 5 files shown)


bougue-pe:
The previous list of speed-limit tag values from infra/RS has to be kept, but enriched (merged/deduplicated) with the new one.
Some values may be present only in the infra and we want to offer them (case of foreign infra, or exotic use).

front/src/common/SpeedLimitByTagSelector/useStoreDataForSpeedLimitByTagSelector.ts
```diff
@@ -15,21 +16,22 @@ export const useStoreDataForSpeedLimitByTagSelector = () => {
...
-  const infraID = useInfraID();
 
   const { updateSpeedLimitByTag } = useOsrdConfActions();
   const dispatchUpdateSpeedLimitByTag = (newTag: string | null) => {
     dispatch(updateSpeedLimitByTag(newTag));
   };
+  const infraID = useInfraID();
 
-  const { data: speedLimitsByTags = [], error } =
+  const { data: speedLimitTags } = osrdEditoastApi.endpoints.getSpeedLimitTags.useQuery();
```
bougue-pe, line 28: I get those errors:  `[TypeScript] Type '(string | undefined)[]' is not assignable to type 'string[]'.`
Looks like this should never be undefined, but not sure.

bougue-pe (commented):
Thanks for the changes 🙏 

---

There are some duplicate bugs on a local test on a small_infra copy (unlocked), docker, Firefox, WSL:
* edit infra, select existing speed-limit
* add an Additional speed_limit by tag `E30C`

Then in operational studies (or STDCM) use that infra and create a train
* in the composition code list, `E30C` appears twice.
* if using the search on that list it looks like when typing `E30C`, then backspace the list of E30C gets longer and longer (maybe a previously existing bug).
* the list gets back to 2 duplicates only when stopping edit on the train then back to edit

The same goes for the selector in infra speed-limit editor, actually

front/src/applications/editor/components/LayersModal.tsx
```diff
@@ -125,8 +130,8 @@ const LayersModal: FC<LayersModalProps> = ({
   );
 
   const speedLimitOptions = useMemo(
-    () => uniq([DEFAULT_SPEED_LIMIT_TAG, ...(speedLimitTags || [])]),
-    [speedLimitTags]
+    () => uniq([DEFAULT_SPEED_LIMIT_TAG, ...(allSpeedLimitTags || [])]),
```
Yohh: don't forget to `compact()`

Yohh (commented):
Nice, one comment

front/src/applications/editor/components/LayersModal.tsx
Uriel-Sautron: Done, thank you.

Yohh (approved):
LGTM

front/src/common/SpeedLimitByTagSelector/useStoreDataForSpeedLimitByTagSelector.ts
bougue-pe: New tests are OK ✅

bougue-pe (approved):
Retested on the same cases, works fine, thank you for this 🙂

I didn't think of the speed-limit tag picker when displaying the infra: I don't know if it's better with all the values (including absent ones) or only the present ones... But both are OK IMO, so it's not a blocker for me.

EDIT: after discussion, let's use the same (extended) list for map layers, so current choices are good ✅

front/src/common/SpeedLimitByTagSelector/useStoreDataForSpeedLimitByTagSelector.ts
```diff
@@ -15,21 +16,22 @@ export const useStoreDataForSpeedLimitByTagSelector = () => {
 
   const { getSpeedLimitByTag } = useOsrdConfSelectors();
   const speedLimitByTag = useSelector(getSpeedLimitByTag);
-  const infraID = useInfraID();
 
   const { updateSpeedLimitByTag } = useOsrdConfActions();
   const dispatchUpdateSpeedLimitByTag = (newTag: string | null) => {
     dispatch(updateSpeedLimitByTag(newTag));
   };
+  const infraID = useInfraID();
```
SharglutDev: Sorry for the request but could you move it back with the other selectors ? :)

front/src/common/Map/Settings/MapSettingsSpeedLimits.tsx
```diff
@@ -44,6 +45,9 @@ const FormatSwitch = ({ name, icon: IconComponent, color = '', disabled }: Forma
       skip: !infraID,
     }
   );
+  const { data: speedLimitTags } = osrdEditoastApi.endpoints.getSpeedLimitTags.useQuery();
+
+  const allSpeedLimitTags = uniq(concat(speedLimitTags, speedLimitsTagsByInfraId));
```
SharglutDev: Maybe you could rather reuse the logic you did in `useStoreDataForSpeedLimitByTagSelector` with the concatenated list of tags. In the hook, you can also expose the `error` and `isError` from tagsByInfraId to use them here.

```suggestion
  const { speedLimitsByTags, speedLimitByTagsByInfraIdError, isErrorSpeedLimitByTagsByInfraId } = useStoreDataForSpeedLimitByTagSelector();
```

front/src/common/SpeedLimitByTagSelector/useStoreDataForSpeedLimitByTagSelector.ts
```diff
@@ -39,6 +41,7 @@ export const useStoreDataForSpeedLimitByTagSelector = () => {
     }
   }, [error]);
 
+  const speedLimitsByTags = compact(uniq(concat(speedLimitTags, speedLimitsTagsByInfraId)));
```
SharglutDev: Did you have any acceptance criteria for how to sort the tags ? Like leave them like that where we can guess their endpoints origin or merge all of them and sort them alphabetically ?

front/src/applications/editor/tools/rangeEdition/components/RangeEditionLeftPanel.tsx
```diff
@@ -107,14 +107,18 @@ const RangeEditionLeftPanel = () => {
...
       {
         infraId: infraID as number,
       },
       { skip: !infraID }
     );
 
+  const { data: speedLimitTags } = osrdEditoastApi.endpoints.getSpeedLimitTags.useQuery();
+
+  const allSpeedLimitTags = uniq(concat(speedLimitTags, speedLimitTagsByInfraId));
+
```
SharglutDev: Same here you could use the custom hook.

front/src/applications/editor/components/LayersModal.tsx
```diff
@@ -69,13 +80,18 @@ const LayersModal: FC<LayersModalProps> = ({
...
-  const { data: speedLimitTags } =
+  const { data: speedLimitTagsByInfraId } =
     osrdEditoastApi.endpoints.getInfraByInfraIdSpeedLimitTags.useQuery(
       { infraId: infraID as number },
       { skip: isNil(infraID) }
     );
+  const { data: speedLimitTags } = osrdEditoastApi.endpoints.getSpeedLimitTags.useQuery();
+
+  const allSpeedLimitTags = concat(speedLimitTags, speedLimitTagsByInfraId);
+
```
SharglutDev: And same here for the custom hook

front/src/common/SpeedLimitByTagSelector/useStoreDataForSpeedLimitByTagSelector.ts
SharglutDev: Maybe you could inject directly here the `DEFAULT_SPEED_LIMIT_TAG` in the list instead of doing that multiple times in other components. What do you think ?

It seems the only component where you don't need the default value is in the infra editor so over there you could just `shift` the array to remove the first element.

SharglutDev (changes requested):
Thank you for this PR, left some comments.

Also it's strange that we don't use the `SpeedLimitByTagSelector` everywhere we need to select a speedlimit tag. I understand that it's not related to your PR but maybe you could create an enhancement to harmonize that ?

front/src/common/SpeedLimitByTagSelector/useStoreDataForSpeedLimitByTagSelector.ts
Uriel-Sautron: Done

front/src/common/Map/Settings/MapSettingsSpeedLimits.tsx
Uriel-Sautron: This hook cannot be reused here, it calls up properties of the store that are outside the scope of the map

front/src/common/SpeedLimitByTagSelector/useStoreDataForSpeedLimitByTagSelector.ts
Uriel-Sautron: No specific acceptance criteria

front/src/applications/editor/tools/rangeEdition/components/RangeEditionLeftPanel.tsx
Uriel-Sautron: Same case as for MapSettingsSpeedLimits

front/src/common/SpeedLimitByTagSelector/useStoreDataForSpeedLimitByTagSelector.ts
Uriel-Sautron: Done

bougue-pe: After discussion with the POs, alphabetical order is better everywhere.
I'll add it to the issue.

Uriel-Sautron: Done

SharglutDev (approved):
Lgtm and tested, good job !

PR #8279, opened 2024-08-01
front: remove trailing slashes in osrdEditoastApi URLs

leovalais:
Refs https://github.com/OpenRailAssociation/osrd/pull/8047#pullrequestreview-2199001346

### DIFF
--- a/front/src/common/api/osrdEditoastApi.ts
+++ b/front/src/common/api/osrdEditoastApi.ts
@@ -33,21 +33,11 @@ const osrdEditoastApi = generatedEditoastApi.enhanceEndpoints({
       }),
     },
     deleteV2TrainSchedule: {
-      query: (queryArg) => ({
-        url: `/v2/train_schedule/`,
-        method: 'DELETE',
-        body: queryArg.body,
-      }),
       // As we always use all get trainschedule v2 endpoints after updating the timetable,
       // we don't want to invalidate the trainschedulev2 tag here to prevent multiple calls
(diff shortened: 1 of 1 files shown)


emersion:
I don't think we even need to set the `url` field here, it's already in `generatedEditoastApi` and we don't need to override it. Same for `method`.

emersion:
Hm, my bad, it's a callback so ~~probably~~ [definitely](https://github.com/reduxjs/redux-toolkit/blob/75b4f698e1c58efa6b19096abe5850da0a431da0/packages/toolkit/src/query/createApi.ts#L329) won't work…

front/src/common/api/osrdEditoastApi.ts
```diff
@@ -34,7 +34,7 @@ const osrdEditoastApi = generatedEditoastApi.enhanceEndpoints({
     },
     deleteV2TrainSchedule: {
       query: (queryArg) => ({
-        url: `/v2/train_schedule/`,
+        url: `/v2/train_schedule`,
         method: 'DELETE',
         body: queryArg.body,
       }),
```
emersion: Upon further inspection, it seems like `query` is identical to the one in `generatedEditoastApi` so can just be dropped altogether.

Same for `postV2TimetableByIdTrainSchedule`.

emersion (approved):
LGTM!

SharglutDev (approved):
Lgtm

PR #8281, opened 2024-08-01
editoast: revert "make core pathfinding/simulation errors non-fatal"

flomonster:
This reverts commit 1ab7fa8635cc861b8bb487ada2309a6691851355.

We consider core 500 errors as unexpected so we don't want to cache the result of this query.

Example:
- Let's say the infra is not loaded but a pathfinding query is performed.
- The "Infra not loaded" error will be transformed as PathfindingResult and cached
- Once the infra is loaded if the same query is performed the cached response is sent (instead of re-computing the pathfinding)


leovalais (approved):
Oh, you're right. I forgot about the caching, my bad. Thanks for the fix :)

PR #8282, opened 2024-08-01
front: add import trains e2e tests

### LINKED ISSUE #8040: front: add e2e tests import des trains
- [x] Create 20 trains with varying status, routes, stops, rolling stock, and simulation settings in a JSON file.
- [x]  Import the JSON file using train_schedule API
- [x] Ensure that each train is imported with the correct status
- [x] Confirm that the simulation results for each train are displayed accurately.

Maymanaf:
Closes #8040
Blocked by #8350 & #8654


front/tests/global-teardown.ts
```diff
@@ -2,24 +2,40 @@ import { test as setup } from '@playwright/test';
...
-  );
-  const dualModeRollingStock: RollingStock = await rollingStocks.results.find(
-    (r: RollingStock) => r.name === 'dual-mode_rollingstock_test_e2e'
-  );
-  await deleteApiRequest(`/api/infra/${infra.id}/`);
-  await deleteApiRequest(`/api/projects/${project.id}/`);
-  await deleteApiRequest(`/api/rolling_stock/${electricRollingStock.id}/`);
-  await deleteApiRequest(`/api/rolling_stock/${dualModeRollingStock.id}/`);
+  // Fetch rolling stocks
+  const rollingStocks = await getApiRequest('/api/light_rolling_stock/', { page_size: 500 });
```
theocrsb, line 18: Can you explain me why we use 500 here ?

front/tests/utils/rollingStock-utils.ts
Akctarus, line 1: I think formatRollingStock.ts would be a better name.

front/tests/utils/trainSchedule-utils.ts
Akctarus, line 1: Same here, formatTrainSchedule.ts seems better

front/tests/utils/index.ts
Akctarus, line 1: can you create some unit tests for theses functions ?

front/tests/utils/trainSchedule-utils.ts
```diff
@@ -0,0 +1,51 @@
+import type { APIResponse, APIRequestContext } from '@playwright/test';
+
+import { postApiRequest, getApiContext } from './api-setup';
+
```
Akctarus: It's just a suggestion, but I think it clearer to extract the handleApiResponse function : 

```suggestion
// Helper function to handle API error responses
async function handleApiResponse(response: APIResponse, errorMessage: string): Promise<void> {
  if (!response.ok()) {
    throw new Error(`${errorMessage}: ${response.status()} ${response.statusText()}`);
  }
}
```

front/tests/utils/trainSchedule-utils.ts
```diff
@@ -0,0 +1,51 @@
...
+        'Content-Type': 'application/json',
+      },
+    }
+  );
+
+  if (!trainScheduleResponse.ok()) {
+    throw new Error(
+      `Failed to send train schedule: ${trainScheduleResponse.status()} ${trainScheduleResponse.statusText()}`
+    );
+  }
```
Akctarus:
```suggestion
  await handleApiResponse(trainScheduleResponse, 'Failed to send train schedule');
```

front/tests/utils/trainSchedule-utils.ts
```diff
@@ -0,0 +1,51 @@
...
+  if (trainIds && trainIds.length > 0) {
+    const simulationResponse = await postApiRequest(`/api/v2/train_schedule/simulation_summary/`, {
+      ids: trainIds,
+      infra_id: infraId,
+    });
+    if (!simulationResponse.ok()) {
+      throw new Error(
+        `Failed to post simulation: ${simulationResponse.status()} ${simulationResponse.statusText()}`
+      );
+    }
```
Akctarus:
```suggestion
    await handleApiResponse(simulationResponse, 'Failed to post simulation');
```

front/tests/008-train-schedule.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    /* eslint-disable no-restricted-syntax, no-await-in-loop */
+    for (const filter of compositionFilters) {
+      await opTimetablePage.clickCodeCompoTrainFilterButton(
+        selectedLanguage,
+        filter.code,
+        filter.count
+      );
+    }
+    /* eslint-enable no-restricted-syntax, no-await-in-loop */
```
Akctarus: I found this solution to get rid of the warnings, but it may not be the best one

```suggestion
await compositionFilters.reduce(async (promise, filter) => {
      // Wait for the previous promise to complete
      await promise;
      // Now perform the current asynchronous operation
      return opTimetablePage.clickCodeCompoTrainFilterButton(
        selectedLanguage,
        filter.code,
        filter.count
      );
    }, Promise.resolve());
```

Akctarus (commented):
Left few comments, all works but we need the bug #8350 to be fix before merging this PR.

front/tests/utils/index.ts
Maymanaf: We don't add unit tests for E2E test functions because it's redundant and adds unnecessary complexity

front/tests/utils/trainSchedule-utils.ts
Maymanaf: The file contains utility functions beyond just formatting.
 I will change the name to` trainSchedule.ts` as the file is under the `utils `folder

front/tests/utils/rollingStock-utils.ts
Maymanaf: The file contains utility functions beyond just formatting.
 I will change the name to` rollingStock.ts` as the file is under the `utils `folder

front/tests/global-teardown.ts
Maymanaf: The default value of `page_size` is 25, which prevents listing all the rolling stock on the same page and makes parsing more difficult.

front/tests/utils/rollingStock.ts
```diff
@@ -0,0 +1,24 @@
...
+
+// Add a rolling Stock
+export async function addRollingStock(rollingStockName: string, rollingStockJson: JSON) {
+  await postApiRequest('/api/rolling_stock/', {
+    ...rollingStockJson,
+    name: rollingStockName,
+  });
+}
+// Find and delete rolling stock with the given name
+export async function findAndDeleteRollingStocks(rollingStockNames: string[]) {
```
clarani:
```suggestion
export async function findAndDeleteRollingStocksByName(rollingStockNames: string[]) {
```

front/tests/utils/trainSchedule.ts
```diff
@@ -0,0 +1,43 @@
+import type { APIResponse, APIRequestContext } from '@playwright/test';
+
+import { postApiRequest, getApiContext } from './api-setup';
+import { handleApiResponse } from './index';
+
+// Function to import train schedules and return the API response
+export async function sendTrainSchedule(timetableId: number, body: object): Promise<APIResponse> {
```
clarani:
```suggestion
export async function sendTrainSchedules(timetableId: number, body: object): Promise<APIResponse> {
```

front/tests/utils/trainSchedule.ts
```diff
@@ -0,0 +1,43 @@
...
+  const apiContext: APIRequestContext = await getApiContext();
+  const trainScheduleResponse: APIResponse = await apiContext.post(
```
clarani:
```suggestion
  const trainSchedulesResponse: APIResponse = await apiContext.post(
```

front/tests/utils/index.ts
```diff
@@ -137,3 +43,15 @@ export async function extractNumberFromString(input: string): Promise<number> {
...
+
+// Helper function to handle API error responses
+export async function handleApiResponse(
+  response: APIResponse,
+  errorMessage: string
+): Promise<void> {
+  if (!response.ok()) {
+    throw new Error(`${errorMessage}: ${response.status()} ${response.statusText()}`);
+  }
+}
```
clarani: Some questions:
- why is this function async ? I think the request has already finished
- can't we return the response.json() ? This way we don't have to manipulate APIResponse everywhere

```suggestion
export function handleApiResponse(
  response: APIResponse,
  errorMessage: string
) {
  if (!response.ok()) {
    throw new Error(`${errorMessage}: ${response.status()} ${response.statusText()}`);
  }
  return response.json()
}
```

front/tests/utils/trainSchedule.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+  );
+
+  await handleApiResponse(trainScheduleResponse, 'Failed to send train schedule');
+  return trainScheduleResponse;
+}
+
+// Function to extract train IDs from the train schedules
+export async function getTrainIds(
+  trainScheduleResponse: APIResponse
```
clarani: Can you modify the code so the 1st arg of the function is not an APIResponse ?

front/tests/utils/trainSchedule.ts
```diff
@@ -0,0 +1,43 @@
...
+): Promise<number[] | undefined> {
+  if (Array.isArray(trainScheduleResponse)) {
+    return trainScheduleResponse.map((item: { id: number }) => item.id);
+  }
+  return undefined;
+}
+
+// Function to post simulation using extracted train IDs
+export async function postSimulation(response: APIResponse, infraId: number): Promise<void> {
```
clarani: Can you modify the code so the 1st arg of the function is not an APIResponse ?

front/tests/utils/api-setup.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export const deleteApiRequest = async (url: string) => {
+  const apiContext = await getApiContext();
+  const response = await apiContext.delete(url);
+  return response;
+};
+
+// API calls for beforeAll setup in tests
+
+export const findOneInResults = <T extends { name: string }>(results: T[], name: string) =>
```
clarani: No need to export this function
```suggestion
const findOneInResults = <T extends { name: string }>(results: T[], name: string) =>
```

front/tests/global-teardown.ts
clarani, line 1: Nice function :+1:

front/tests/pages/op-timetable-page-model.ts
```diff
@@ -0,0 +1,217 @@
...
+class OperationalStudiesTimetablePage {
+  readonly page: Page;
+
+  readonly trainCountText: Locator;
+
+  readonly invalidTrainsMessage: Locator;
+
+  readonly scenarioTimetableTrains: Locator;
+
+  readonly selectedScenarioTimetableTrain: Locator;
```
clarani: nit: Maybe you can remove the Scenario in the name ?
```suggestion
  readonly timetableTrains: Locator;

  readonly selectedTimetableTrain: Locator;
```

front/tests/pages/op-timetable-page-model.ts
```diff
@@ -0,0 +1,217 @@
...
+    this.page = page;
+    this.trainCountText = page.getByTestId('train-count');
+    this.invalidTrainsMessage = page.getByTestId('invalid-trains-message');
+    this.scenarioTimetableTrains = page.getByTestId('scenario-timetable-train');
+    this.selectedScenarioTimetableTrain = page.locator(
+      '[data-testid="scenario-timetable-train"].selected'
+    );
+    this.simulationBar = page.locator('.osrd-simulation-sticky-bar');
+    this.manchetteSpaceTimeChart = page.locator('.manchette-space-time-chart-wrapper');
+    this.spaceTimeChart = page.locator('#container-SpeedSpaceChart');
```
clarani, line 41:
```suggestion
    this.speedSpaceChart = page.locator('#container-SpeedSpaceChart');
```

front/tests/pages/op-timetable-page-model.ts
```diff
@@ -0,0 +1,217 @@
...
+    this.timeStopsDatasheet = page.locator('.time-stops-datasheet');
+    this.spaceCurvesSlopesChart = page.locator('#container-SpaceCurvesSlopes');
+    this.simulationMap = page.locator('.osrd-simulation-map');
+    this.simulationDriverTrainSchedule = page.locator('.simulation-driver-train-schedule');
+    this.timetableFilterButton = page.getByTestId('timetable-filter-button');
+    this.validTrainsFilterButton = page.locator('#train-validityvalid');
```
clarani: The locator `#train-validityvalid` seems a bit weird

front/tests/pages/op-timetable-page-model.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  readonly spaceCurvesSlopesChart: Locator;
+
+  readonly simulationMap: Locator;
+
+  readonly simulationDriverTrainSchedule: Locator;
+
+  readonly timetableFilterButton: Locator;
+
+  readonly validTrainsFilterButton: Locator;
```
clarani: I think this locator is not used
```suggestion
```

front/tests/pages/op-timetable-page-model.ts
```diff
@@ -0,0 +1,217 @@
...
+    await this.verifyTrainCount(expectedTrainCount, selectedLanguage);
+  }
+
+  // Iterate over each train element and verify the visibility of simulation results
+  async verifyEachTrainSimulation(): Promise<void> {
+    const trainCount = await this.scenarioTimetableTrains.count();
+    let currentTrainIndex = 0;
+
+    /* eslint-disable no-await-in-loop */
+    while (currentTrainIndex < trainCount) {
```
clarani: Why do you use a while loop here ? I think a for loop does the same thing and is more standard

front/tests/008-train-schedule.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+let study: Study;
+let scenario: Scenario;
+let timetableResult: TimetableResult;
+
+const trainSchedulesJson = readJsonFile('./tests/assets/trainSchedule/train_schedules.json');
+test.beforeAll(async () => {
+  // Fetch infrastructure, project, study, and timetable result
+  smallInfra = (await getInfra()) as Infra;
+  project = await getProject();
+  study = await getStudy(project.id);
```
clarani: Maybe it would be nice to create a utils function to directly get the study or directly post the scenario

front/tests/008-train-schedule.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+    ...scenarioData,
+    name: `${scenarioData.name} ${uuidv4()}`,
+    study_id: study.id,
+    infra_id: smallInfra.id,
+    timetable_id: timetableResult.timetable_id,
+  });
+
+  // Post train schedule import and simulation
+  const response = await sendTrainSchedule(scenario.timetable_id, trainSchedulesJson);
+  await postSimulation(response, smallInfra.id);
```
clarani: Why do you need to post the simulation here ? Is it to cache the result ?
Remark: I think the request will fail if the infra is not loaded

front/tests/utils/index.ts
Maymanaf: This function handles only the error case of the API response to prevent code duplication. It's part of a larger `async `function, and adding `async` ensures consistency and allows for awaiting the response when the function is called

front/tests/008-train-schedule.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+    ];
+
+    await compositionFilters.reduce(async (promise, filter) => {
+      await promise;
+      return opTimetablePage.clickCodeCompoTrainFilterButton(
+        selectedLanguage,
+        filter.code,
+        filter.count
+      );
+    }, Promise.resolve());
```
clarani: Using a reduce here is a bit weird since you are not creating a accumulator. Why not use a for loop ?

clarani (commented):
Very nice test, thanks :)

front/tests/pages/op-timetable-page-model.ts
Maymanaf: I had the same remark, but I don't change the IDs in the source code only data-testid

front/tests/pages/op-timetable-page-model.ts
Maymanaf: I awlays prefer using for loop but to avoid eslint messages I changed the function to while loop

front/tests/008-train-schedule.spec.ts
Maymanaf: The postSimulation is only used in these tests and needs to be run just once in `beforeAll`
The infra is loaded in `global-setup.ts`, in this test, we only use the ID of the loaded infra

front/tests/008-train-schedule.spec.ts
Maymanaf: I was using a `for `loop, but it generated ESLint errors. Following Theo's suggestion to avoid those messages, I switched to using `reduce `instead.

front/tests/pages/op-timetable-page-model.ts
clarani: I think for loop is better yes, can you locally deactivate eslint for the for loops ?

front/tests/008-train-schedule.spec.ts
clarani: Why do you need to run the postSimulation in the beforeAll ? Is it to cache the result in the backend ? If that's the case, a comment to explain it would be nice :)

Maymanaf: The postSimulation in the beforeAll is used to cache the results for fast loading the results of each train and to avoid timeouts

front/tests/pages/op-timetable-page-model.ts
Maymanaf: Done

front/tests/008-train-schedule.spec.ts
Maymanaf: Changed back to for loop

front/tests/utils/index.ts
Maymanaf: - I removed the async function 
- The response.json() is used only once, so I didn't add it to the method (The response is handled directly within an if condition the second time).

clarani (approved):
LGTM :white_check_mark:

front/public/locales/en/operationalStudies/scenario.json
Akctarus, line 1: Can you fix this line by deleting the "s" at the end of "trains" ? 

"trainCount_one": "1/{{ totalCount }} trains",

front/public/locales/fr/operationalStudies/scenario.json
Akctarus, line 1: Same here. Can you fix this line by deleting the "s" at the end of "trains" ?

"trainCount_one": "1/{{ totalCount }} trains",

front/public/locales/fr/operationalStudies/scenario.json
```diff
@@ -70,7 +70,7 @@
   "toggleTimetable": "Basculer l'affichage de la grille horaires",
   "trainCount_one": "1/{{ totalCount }} trains",
   "trainCount_other": "{{ count }}/{{ totalCount }} trains",
-  "trainCount_zero": "0/{{ totalCount }} train",
+  "trainCount_zero": "0/{{ totalCount }} trains",
```
Akctarus, line 73:
```suggestion
  "trainCount_zero": "0/{{ totalCount }} train",
```

In French, when the number is zero, the singular is used

Akctarus (approved):
LGTM, just left 2 typo comments

front/public/locales/fr/operationalStudies/scenario.json
Maymanaf: The use of plural should apply based on the totalCount, not the filtered results of 0 or 1. 
For example, if we have 10 trains in total and the filter shows 0 invalid trains, it should display '0/10 trains' . 
'trainCount_zero' and 'trainCount_one' need to be adjusted to align with the total count, rather than just the filtered values

Akctarus: Oh ok, so no change needed, good job for this PR !

PR #8284, opened 2024-08-01
core: stdcm: rework visited nodes to include time range

### LINKED ISSUE #8268: core: stdcm: find a better way to identify visited edges (in time)
Related to https://github.com/OpenRailAssociation/osrd/issues/6273, but because the notes taken there mostly concern the train *location* and we have the same problem for *time*, I'm splitting this issue

When two different paths lead to the same location (in space), we also check if we've already visited the location at similar times.

The way it currently works is by rounding the passage time and doing an equality check. (I'm simplifying a little but that's the idea). This is limited: visiting the same location every 5 minutes doesn't necessarily bring much value, and may result in infinite loops.

---

My first intuition was to handle time ranges for each location. The range would be defined as [arrival time, arrival time + max delay]. We consider that the location is visited if the new range is fully enclosed in the visited ranges.

But there's a problem: when two paths differ by a small amount, there may be a lot of overlap between ranges but there's always a little extra unvisited range. We can add some rounding and tolerance, but we're mostly back to the previous version. See this drawing, the little red area is always unvisited:

image

Now here's my new intuition: **these two paths are in fact equivalent** (in most cases). 

The red segment is actually available from the green path. We can't reach it by moving the departure time, but we can *add a margin* to make the train slow down. A margin is not as good as it makes travel time longer, so we can't just mark it as visited by the green path. 

But the blue path is shifted because *it took longer* to get there. Which is equivalent to a margin. If the solution were in the red segment, blue path + no margin is precisely as good as green path + margin. So we can in fact mark it as visited, but in a different and "weaker" way.

Now I need to figure out how that could be actually implemented.

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/8268

First commit is unrelated (debugging utils)

### Performances:

On average it barely makes a measurable difference and actually slows things down (by 1%). But std goes down by 25%.

On some extreme cases that caused timeouts (and in practice infinite loops / OOM), it's down to a reasonable time (22s).

No extra error or "path not found" when running the benchmark.

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMGraph.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMGraph.kt
@@ -15,6 +15,8 @@ import fr.sncf.osrd.train.RollingStock.Comfort
 import fr.sncf.osrd.utils.units.meters
 import java.lang.Double.isFinite
 import java.lang.Double.isNaN
+import kotlin.math.max
+import kotlin.math.min
 
 /**
  * This is the class that encodes the STDCM problem as a graph on which we can run our pathfinding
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMGraph.kt
```diff
@@ -98,36 +99,71 @@ class STDCMGraph(
...
+        // We look for the 20km before the node (very rough estimation of a distance that lets the
+        // train slow down to a stop and speed up). We return the max delay that can be added after
+        // the train in all of those edges, on top of the start delay
+        var node = inputNode
+        var remainingDistance = 20_000.meters
+        var maxTime = Double.POSITIVE_INFINITY
+        while (true) {
+            val edge = node.previousEdge ?: return 0.0
+
+            val endTime = edge.timeStart + edge.totalTime + edge.maximumAddedDelayAfter
```
Erashin: It doesn't change the math, since we substract the following values, but i feel like it'd be clearer: the real start time is `edge.timeStart + edge.addedDelay`. Same with `edge.timeNextOccupancy + edge.addedDelay`.

eckter: What I'm interested in here is the time we can add with engineering allowances, *after applying the maximum departure time shift*.

`endTime` is the last time on the edge (including the time spent on it), with max added delay

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMGraph.kt
```diff
@@ -158,7 +159,7 @@ class STDCMGraph(
             maxTime = min(maxTime, maxDelayAddedOnEdge)
 
             remainingDistance -= edge.length.distance
-            if (edge.beginSpeed == 0.0 || remainingDistance <= 0.meters) return maxTime
+            if (edge.beginSpeed == 0.0 || remainingDistance <= 0.meters) return max(0.0, maxTime)
```
Erashin: I'd rather you did this on `maxDelayAddedOnEdge`, it makes more sense for it to always be positive.

core/src/test/kotlin/fr/sncf/osrd/stdcm/VisitedNodesTests.kt
```diff
@@ -0,0 +1,92 @@
...
+        val params1 =
+            VisitedNodes.Parameters(
+                fingerprint = fingerprint,
+                startTime = 0.0,
+                duration = 100.0,
+                maxMarginDuration = 100.0,
+                baseNodeCost = 0.0,
+            )
+        val params2 = params1.copy(startTime = 120.0, baseNodeCost = 15.0)
+        // At t=120, it's covered by the first parameters at cost=20
```
Erashin, line 87: This looks a lot like the same setup as the previous test, but the outcome is the opposite. Might need more details in the comment.

Erashin (approved):

PR #8286, opened 2024-08-01
front: fix timestops output using wrong simulation

### LINKED ISSUE #8280: Output table not always displaying final_output schedules
### What happened?

In the simulation results, the output table is synchronized with the simulation used in the driver trainSchedule but it shouldn't.

Output table should always use the final_output simulation

Go to a scenario and add train with margins

Dev (DgexSol)

Firefox

dev 1ab7fa8

SharglutDev:
close #8280 

- fix the bug
- optimize the switch in driver trainSchedule so it doesn't call the endpoint each time we switch (resulting with a small loading)

### DIFF
--- a/front/src/modules/trainschedule/components/DriverTrainScheduleV2/utils.ts
+++ b/front/src/modules/trainschedule/components/DriverTrainScheduleV2/utils.ts
@@ -2,7 +2,10 @@ import * as d3 from 'd3';
 import { uniq } from 'lodash';
 
 import type { TrackSectionEntity } from 'applications/editor/tools/trackEdition/types';
-import type { PathPropertiesFormatted } from 'applications/operationalStudies/types';
+import type {
+  PathPropertiesFormatted,
+  SimulationResponseSuccess,
+} from 'applications/operationalStudies/types';
 import { convertDepartureTimeIntoSec } from 'applications/operationalStudies/utils';
 import {
   osrdEditoastApi,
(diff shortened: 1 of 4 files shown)


front/src/modules/trainschedule/components/DriverTrainScheduleV2/utils.ts
```diff
@@ -214,21 +220,42 @@ export const formatOperationalPoints = async (
...
+    let finalOutputTime = 0;
+    let finalOutputSpeed = 0;
+    // Get time
+    if (matchingFinalOutPutReportTrainIndex !== -1) {
+      finalOutputTime = final_output.times[matchingFinalOutPutReportTrainIndex];
+      finalOutputSpeed = final_output.speeds[matchingFinalOutPutReportTrainIndex];
+    } else {
+      finalOutputTime = interpolateValue(final_output, op.position, 'times');
+      finalOutputSpeed = interpolateValue(final_output, op.position, 'speeds');
     }
```
kmer2016: Since we applied the same operation to both base and finalOutput, we should consider extracting this logic into a function and applying it to both variables. What do you think?

kmer2016 (approved):
LGTM

front/src/modules/trainschedule/components/DriverTrainScheduleV2/utils.ts
SharglutDev: Done

PR #8307, opened 2024-08-01
Permission-based authorization system

leovalais:
wip


leovalais:
Too far behind, and the design will probably radically change.

PR #8313, opened 2024-08-02
editoast: make core pathfinding/simulation errors non-fatal, take 2

emersion:
Commit 1ab7fa8635cc ("editoast: make core pathfinding/simulation errors non-fatal") got reverted in #8281 because it ended up caching HTTP errors, which isn't desirable since some of these errors are transient.

Fix this by only caching successful HTTP responses, retaining previous behavior.


flomonster (approved):
LGTM

PR #8316, opened 2024-08-02
front: ignore errors for route entry/exit in getAdditionalEntities()

### LINKED ISSUE #8304: Can't edit invalid routes
### What happened?

When a route has an invalid reference we can't select it and edit it.

I suspect this [commit](https://github.com/OpenRailAssociation/osrd/commit/4070d784c756e38dfcad3cd4ac8adbec09fdcbd7) to be the source of the issue.

We should be able to edit an invalid route. If the geometry cannot be computed we should still be able to see the edition form.

1. Download and import this infra: [infra_route_bug.json](https://github.com/user-attachments/files/16457315/infra_route_bug.json)
2. Go to the editor
3. Click on "5 issues"
4. Select the invalid reference
5. It's loading forever

You can also craft url to access the invalid route: https://osrd.staging.dgexsol.fr/editor/INFRA_ID?selection=Route%7Ert.DA4-%3Ebuffer_stop.2

Local

Firefox

0a34296

emersion:
_See individual commits._

Also contains a collection of totally gratuitous cleanups.

Closes: https://github.com/OpenRailAssociation/osrd/issues/8304

### DIFF
--- a/front/src/applications/editor/components/EntitySumUp.tsx
+++ b/front/src/applications/editor/components/EntitySumUp.tsx
@@ -327,31 +327,21 @@ const EntitySumUp: FC<
     const fetchEntities = async () => {
       setState({ type: 'loading' });
 
-      if (entity) {
-        const additionalEntities = await getAdditionalEntities(infraID as number, entity, dispatch);
-        setState({
-          type: 'ready',
-          entity,
-          additionalEntities,
-        });
-      } else {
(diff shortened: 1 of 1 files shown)


theocrsb (approved):
LGTM and tested
Thank you for the detailed commits!

SharglutDev (approved):
Lgtm and tested. Thank you for the cleaning !

PR #8317, opened 2024-08-02
Add roles to endpoints

Khoyo:


woshilapin (commented):
Do we still need this draft? It seems that builtin-role are added as we go, which seems fine and easier to review. Or do I miss something?

PR #8319, opened 2024-08-02
editoast: implement Authorizer creation middleware

### LINKED ISSUE #8276: editoast: make an authorizer request extractor
_Originally posted by @woshilapin in https://github.com/OpenRailAssociation/osrd/pull/7981#discussion_r1698707373_

leovalais:
Half of the axum server is in `main.rs` and the rest in `views/mod.rs`. We'll need to homogenize that.

We removed the lazy role fetching and caching as this would require internal mutability since the `Authorizer` now lives in an `Arc`. Instead, we now fetch the roles when the `Authorizer` is initialized.

Closes #8276

### DIFF
--- a/editoast/editoast_authz/src/authorizer.rs
+++ b/editoast/editoast_authz/src/authorizer.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashSet, future::Future};
+use std::{collections::HashSet, future::Future, sync::Arc};
 
 use tracing::debug;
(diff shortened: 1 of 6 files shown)


Wadjetz (approved):

hamz2a (approved):
LGTM, thanks !

flomonster (approved):
LGTM

PR #8320, opened 2024-08-02
editoast: silence dependency_on_unit_never_type_fallback

Khoyo:
This warning only occurs on nightly, as a results of using the `tracing::instrument` macro. We can't really do anything about it, so we should stop polluting our compiler output.


leovalais (approved):
Works on my machine™️, if the CI passes on `stable` then 🚀

PR #8332, opened 2024-08-05
front: fix problem many zeros after setting the initial velocity in a…

### LINKED ISSUE #7271: Many zeros after setting the initial velocity in a scenario
### What happened?

The value setted for the field inital velocity has many zeros
image

Round the value at 0.1 km/h

1. Go to a scenario
2. Click sur "Add"
3. Select a rolling stock and set the Route
4. Set a name for the train
5. Set the value of "initial velocity" field to 120
6. Click on "Add one or several trains"
7. Click on "Go back to simulation results"
8. Click on "Edit" for the added train and observe the field "initial velocity"

- **Browser**: Chrome
- **OS**: W11
- **Env**: Rec

b0b88cf

theocrsb:
close #7271

### DIFF
--- a/front/src/utils/__tests__/numbers.spec.ts
+++ b/front/src/utils/__tests__/numbers.spec.ts
@@ -1,4 +1,4 @@
-import { budgetFormat, isFloat, stripDecimalDigits } from 'utils/numbers';
+import { budgetFormat, isFloat, isInvalidFloatNumber, stripDecimalDigits } from 'utils/numbers';
 import { NARROW_NO_BREAK_SPACE, NO_BREAK_SPACE } from 'utils/strings';
 
 describe('budgetFormat', () => {
@@ -78,5 +78,31 @@ describe('isFloat', () => {
 
   it('should return false if the number is NaN', () => {
     expect(isFloat(NaN)).toBe(false);
+  });
+});
+
+describe('isInvalidFloatNumber', () => {
+  it('should return true if the number has more decimal places than allowed', () => {
+    expect(isInvalidFloatNumber(17.12345, 1)).toBe(true);
+  });
+
+  it('should return false if the number has the same number of decimal places than allowed', () => {
+    expect(isInvalidFloatNumber(17.1, 1)).toBe(false);
+  });
+
+  it('should return false if the number has less decimal places than allowed', () => {
+    expect(isInvalidFloatNumber(17.1, 2)).toBe(false);
+  });
+
+  it('should return false if the number is NaN', () => {
+    expect(isInvalidFloatNumber(NaN, 1)).toBe(false);
+  });
+
+  it('should return false if the number is a float and decimal number is NaN', () => {
+    expect(isInvalidFloatNumber(10.5, NaN)).toBe(false);
+  });
+
+  it('should return false if the number is NaN and decimal number is NaN', () => {
+    expect(isInvalidFloatNumber(NaN, NaN)).toBe(false);
   });
 });
--- a/front/public/locales/en/operationalStudies/manageTrainSchedule.json
+++ b/front/public/locales/en/operationalStudies/manageTrainSchedule.json
@@ -16,6 +16,7 @@
     "allowance_convergence": "We could not go slow enough in this setting to match the given allowance time",
     "error": "An error has occurred",
     "hardErrorInfra": "The infrastructure contains errors and cannot be used without first being corrected.",
+    "invalidInitialSpeed": "The initial speed must be rounded to the nearest tenth",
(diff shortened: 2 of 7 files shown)


SharglutDev (changes requested):
With your solution, the value entered by the user will always be rounded even if he doesn't want to.

The issue related by the bug comes from the fact that the endpoint to add a train requires a speed in m/s and not in km/h. So when the conversion is done when adding (save) and getting the train (for edit), some extra decimals can pop.

I think it would be better to handle this logic in the `adjustConfWithTrainToModifyV2` function, l.151 when we are getting the initialSpeed from the train and update the store with it by converting the value in km/h.

```suggestion
const convertedInitialSpeed = initial_speed ? Math.round(msToKmh(initial_speed) * 10) / 10 : 0)
dispatch(updateInitialSpeed(convertedtedInitialSpeed));
```
NB : 
- Maybe you can add the const before all the dispatch
- Maybe we need to check with a PO how many decimal max we want

theocrsb:
> With your solution, the value entered by the user will always be rounded even if he doesn't want to.
> 
> The issue related by the bug comes from the fact that the endpoint to add a train requires a speed in m/s and not in km/h. So when the conversion is done when adding (save) and getting the train (for edit), some extra decimals can pop.
> 
> I think it would be better to handle this logic in the `adjustConfWithTrainToModifyV2` function, l.151 when we are getting the initialSpeed from the train and update the store with it by converting the value in km/h.
> 
> ```
> const convertedInitialSpeed = initial_speed ? Math.round(msToKmh(initial_speed) * 10) / 10 : 0)
> dispatch(updateInitialSpeed(convertedtedInitialSpeed));
> ```
> 
> NB :
> 
> * Maybe you can add the const before all the dispatch
> * Maybe we need to check with a PO how many decimal max we want

After discussion with a PO, it was decided to keep only one digit after the decimal point.

front/src/modules/trainschedule/components/ManageTrainSchedule/hooks/isValidInitalSpeed.ts
```diff
@@ -0,0 +1,6 @@
+const isInvalidInitalSpeed = (speed: number): boolean => {
+  const speedString = speed.toString();
+  return speedString.includes('.') && speedString.split('.')[1].length > 1;
+};
```
SharglutDev: This util could be more generic and added in utils/number. We also have an util to check for float number already. If you could also add a doc for the function and a few tests (with no decimals, with 0, with regular decimal etc). 

```suggestion
const isInvalidFloatNumber = (value: number, numberOfDecimal: number): boolean => {
  if (!isFloat(value)) return false
  const stringifyValue = value.toString();
  return stringifyValue.split('.')[1].length > numberOfDecimal;
};
```

SharglutDev (changes requested):

front/src/modules/trainschedule/components/ManageTrainSchedule/hooks/isValidInitalSpeed.ts
theocrsb: You're right. It's done

front/src/utils/__tests__/numbers.spec.ts
```diff
@@ -80,3 +80,21 @@ describe('isFloat', () => {
...
+    expect(isInvalidFloatNumber(17.1, 1)).toBe(false);
+  });
+
+  it('should return false if the number has less decimal places than allowed', () => {
+    expect(isInvalidFloatNumber(17.1, 2)).toBe(false);
+  });
+
+  it('should return false if the number is NaN', () => {
+    expect(isInvalidFloatNumber(NaN, 1)).toBe(false);
+  });
```
SharglutDev, line 99: Just add these tests and lgtm :)

```suggestion
  it('should return false if the number is NaN', () => {
    expect(isInvalidFloatNumber(NaN, 1)).toBe(false);
  });
  
  it('should return false if the number is a float and decimal number is NaN', () => {
    expect(isInvalidFloatNumber(10.5, NaN)).toBe(false);
  });
  
  it('should return false if the number is NaN and decimal number is NaN', () => {
    expect(isInvalidFloatNumber(NaN, NaN)).toBe(false);
  });
```

SharglutDev (changes requested):

front/src/utils/__tests__/numbers.spec.ts
theocrsb: Done !

SharglutDev (approved):
Lgtm and tested

PR #8334, opened 2024-08-05
table, trains after midnight

### LINKED ISSUE #7781: TSV2: input table: support trains that go over midnight
# Description

Part of https://github.com/osrd-project/osrd-confidential/issues/535

# AC

- [x] If the user asks for a time that would require the train to go over midnight (time inferior to a previous via point), "D+1" is automatically displayed in the "Requested arrival" or "Requested departure" column (depending on which one the user has modified).
- [x] Only one "D+1" should be displayed, and it must be in the cell closest to midnight.
- [x] "D+1" should not be displayed in any other time columns
- [x] Same behavior for "D+2", "D+3", etc

---
We need a format to represent the number of day since departure.
using` { time: '23:50:00', daySinceDeparture: 1 }`
- [x] change format for input table (only using the pathSteps in store)
- [x] change format for the output table (using data from GET /train_schedule)

anisometropie:
closes #7781 

We now use a different format to represent `arrival` and `departure` in `SuggestedOP` and `PathStep`, an ISO duration string. The previous time string format hh:mm:ss was insufficient for distinguishing times spanning more than 24 hours

These two types are used in different places beyond the scope of just the input/outpt, we should test everything that uses these two types

### DIFF
--- a/front/public/locales/en/timesStops.json
+++ b/front/public/locales/en/timesStops.json
@@ -2,6 +2,7 @@
   "arrivalTime": "Requested arrival Time",
   "calculatedArrivalTime": "Calculated arrival time",
   "calculatedDepartureTime": "Calculated departure time",
+  "dayCounter": "D+{{count}}",
   "departureTime": "Requested departure Time",
   "diffMargins": "Margins diff.",
   "name": "Name",
--- a/front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
+++ b/front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
@@ -22,7 +22,6 @@ import { setFailure } from 'reducers/main';
 import type { OperationalStudiesConfSliceActions } from 'reducers/osrdconf/operationalStudiesConf';
 import type { PathStep } from 'reducers/osrdconf/types';
 import { useAppDispatch } from 'store';
-import { addDurationToIsoDate } from 'utils/date';
 import { castErrorToFailure } from 'utils/error';
 import { getPointCoordinates } from 'utils/geometry';
 import { mmToM } from 'utils/physics';
@@ -70,9 +69,7 @@ const computeBasePathSteps = (trainSchedule: TrainScheduleResult) =>
       ...stepWithoutSecondaryCode,
       ch: 'secondary_code' in step ? step.secondary_code : undefined,
       name,
-      arrival: arrival
-        ? addDurationToIsoDate(trainSchedule.start_time, arrival).substring(11, 19)
-        : arrival,
+      arrival, // ISODurationString
       stopFor: stopFor ? ISO8601Duration2sec(stopFor).toString() : stopFor,
       locked,
       onStopSignal,
--- a/front/src/common/types.ts
+++ b/front/src/common/types.ts
@@ -15,6 +15,21 @@ export const DATA_TYPES = {
  */
 export type TimeString = string;
 
+/**
+ * A string with the complete iso format
+ *
+ * @example "2024-08-08T10:12:46.209Z"
+ * @example "2024-08-08T10:12:46Z"
+ * @example "2024-08-08T10:12:46+02:00"
+ */
+export type IsoDateTimeString = string;
+
+/**
+ * A ISO 8601 duration string
+ * @example "PT3600S"
+ */
+export type IsoDurationString = string;
+
 export type RangedValue = {
   begin: number;
   end: number;
--- a/front/src/modules/timesStops/ReadOnlyTime.tsx
+++ b/front/src/modules/timesStops/ReadOnlyTime.tsx
@@ -1,0 +1,23 @@
+import type { CellProps } from 'react-datasheet-grid/dist/types';
+import { useTranslation } from 'react-i18next';
+
+import { NO_BREAK_SPACE } from 'utils/strings';
+
+import type { TimeExtraDays } from './types';
+
+type ReadOnlyTimeProps = CellProps<TimeExtraDays | undefined, string>;
+
+const ReadOnlyTime = ({ rowData }: ReadOnlyTimeProps) => {
+  const { time, daySinceDeparture, dayDisplayed } = rowData || {};
+  if (!time) {
+    return null;
+  }
+  const { t } = useTranslation('timesStops');
+  const fullString =
+    daySinceDeparture !== undefined && dayDisplayed
+      ? `${time}${NO_BREAK_SPACE}${t('dayCounter', { count: daySinceDeparture })}`
+      : time;
+  return <div className="read-only-time">{fullString}</div>;
+};
+
+export default ReadOnlyTime;
--- a/front/src/modules/timesStops/TimeInput.tsx
+++ b/front/src/modules/timesStops/TimeInput.tsx
@@ -1,15 +1,17 @@
 import { useRef, useState, useEffect } from 'react';
 
+import cx from 'classnames';
 import type { CellProps } from 'react-datasheet-grid/dist/types';
+import { useTranslation } from 'react-i18next';
 
-const TimeInput = ({
-  focus,
-  rowData,
-  active,
-  setRowData,
-}: CellProps<string | null | undefined, string>) => {
+import type { TimeExtraDays } from './types';
+
+type TimeInputProps = CellProps<TimeExtraDays | undefined, string>;
+
+const TimeInput = ({ focus, rowData, active, setRowData }: TimeInputProps) => {
+  const { t } = useTranslation('timesStops');
   const ref = useRef<HTMLInputElement>(null);
-  const [tempTimeValue, setTempTimeValue] = useState<string | null | undefined>(rowData);
+  const [tempTimeValue, setTempTimeValue] = useState<TimeExtraDays | undefined>(rowData);
 
   useEffect(() => {
     if (active) {
@@ -26,8 +28,9 @@ const TimeInput = ({
     setTempTimeValue(rowData);
   }, [rowData]);
 
-  return (
+  const input = (
     <input
+      // className from react-datasheet-grid library
       className="dsg-input"
       type="time"
       tabIndex={-1}
@@ -37,9 +40,9 @@ const TimeInput = ({
         pointerEvents: focus ? 'auto' : 'none',
         opacity: rowData || active ? undefined : 0,
       }}
-      value={tempTimeValue ?? ''}
+      value={tempTimeValue?.time ?? ''}
       onChange={(e) => {
-        setTempTimeValue(e.target.value);
+        setTempTimeValue((prev) => ({ ...prev, time: e.target.value }));
       }}
       onBlur={() => {
         // To prevent the operational point to be transformed into a via if we leave the cell empty after focusing it
@@ -49,8 +52,22 @@ const TimeInput = ({
       }}
     />
   );
-};
 
-TimeInput.displayName = 'TimeInput';
+  if (tempTimeValue?.daySinceDeparture && tempTimeValue.dayDisplayed) {
+    return (
+      <div className="time-input-container">
+        {input}
+        <span
+          className={cx('extra-text', {
+            'extra-text-firefox': navigator.userAgent.search('Firefox') !== -1,
+          })}
+        >
+          {t('dayCounter', { count: tempTimeValue.daySinceDeparture })}
+        </span>
+      </div>
+    );
+  }
+  return input;
+};
 
 export default TimeInput;
--- a/front/src/modules/timesStops/TimesStops.tsx
+++ b/front/src/modules/timesStops/TimesStops.tsx
@@ -12,7 +12,7 @@ import { useAppDispatch } from 'store';
 import { time2sec } from 'utils/timeManipulation';
 
 import { marginRegExValidation } from './consts';
-import { formatSuggestedViasToRowVias } from './helpers/utils';
+import { formatSuggestedViasToRowVias, transformRowDataOnChange } from './helpers/utils';
 import { useTimeStopsColumns } from './hooks/useTimeStopsColumns';
 import { TableType } from './types';
 import type { PathWaypointRow } from './types';
@@ -78,31 +78,11 @@ const TimesStops = ({
         if (!isInputTable) {
           return;
         }
-        const rowData = { ...row[op.fromRowIndex] };
-        const previousRowData = rows[op.fromRowIndex];
-        if (
-          rowData.departure &&
-          rowData.arrival &&
-          (rowData.arrival !== previousRowData.arrival ||
-            rowData.departure !== previousRowData.departure)
-        ) {
-          rowData.stopFor = String(time2sec(rowData.departure) - time2sec(rowData.arrival));
-        }
-        if (!rowData.stopFor && op.fromRowIndex !== allWaypoints.length - 1) {
-          rowData.onStopSignal = false;
-        }
-        if (rowData.theoreticalMargin && !marginRegExValidation.test(rowData.theoreticalMargin)) {
-          rowData.isMarginValid = false;
+        const newRowData = transformRowDataOnChange(row[op.fromRowIndex], rows[op.fromRowIndex], op, allWaypoints.length);
+        if (!newRowData.isMarginValid) {
           setRows(row);
         } else {
-          rowData.isMarginValid = true;
-          if (op.fromRowIndex === 0) {
-            rowData.arrival = null;
-            // As we put 0% by default for origin's margin, if the user removes a margin without
-            // replacing it to 0% (undefined), we change it to 0%
-            if (!rowData.theoreticalMargin) rowData.theoreticalMargin = '0%';
-          }
-          dispatch(upsertViaFromSuggestedOP(rowData as SuggestedOP));
+          dispatch(upsertViaFromSuggestedOP(newRowData as SuggestedOP));
         }
       }}
       stickyRightColumn={stickyRightColumn}
--- a/front/src/modules/timesStops/TimesStops.tsx
+++ b/front/src/modules/timesStops/TimesStops.tsx
@@ -1,18 +1,23 @@
 import { useState, useEffect } from 'react';
 
 import cx from 'classnames';
+import { isEqual } from 'lodash';
 import { DynamicDataSheetGrid, type DataSheetGridProps } from 'react-datasheet-grid';
 import { useTranslation } from 'react-i18next';
 
 import { useOsrdConfActions } from 'common/osrdContext';
+import type { IsoDateTimeString } from 'common/types';
 import { isVia } from 'modules/pathfinding/utils';
 import type { SuggestedOP } from 'modules/trainschedule/components/ManageTrainSchedule/types';
 import type { PathStep } from 'reducers/osrdconf/types';
 import { useAppDispatch } from 'store';
-import { time2sec } from 'utils/timeManipulation';
 
-import { marginRegExValidation } from './consts';
-import { formatSuggestedViasToRowVias, transformRowDataOnChange } from './helpers/utils';
+import {
+  formatSuggestedViasToRowVias,
+  updateRowTimesAndMargin,
+  updateDaySinceDeparture,
+  durationSinceStartTime,
+} from './helpers/utils';
 import { useTimeStopsColumns } from './hooks/useTimeStopsColumns';
 import { TableType } from './types';
 import type { PathWaypointRow } from './types';
@@ -22,7 +27,7 @@ export const WITH_KP = true;
 type TimesStopsProps = {
   allWaypoints?: SuggestedOP[];
   pathSteps?: PathStep[];
-  startTime?: string;
+  startTime?: IsoDateTimeString;
   tableType: TableType;
   cellClassName?: DataSheetGridProps['cellClassName'];
   stickyRightColumn?: DataSheetGridProps['stickyRightColumn'];
@@ -42,7 +47,7 @@ const TimesStops = ({
   const { t } = useTranslation('timesStops');
 
   const dispatch = useAppDispatch();
-  const { upsertViaFromSuggestedOP } = useOsrdConfActions();
+  const { upsertSeveralViasFromSuggestedOP } = useOsrdConfActions();
 
   const [rows, setRows] = useState<PathWaypointRow[]>([]);
 
@@ -55,7 +60,7 @@ const TimesStops = ({
         startTime,
         tableType
       );
-      setRows(suggestedOPs);
+      setRows(updateDaySinceDeparture(suggestedOPs, startTime, true));
     }
   }, [allWaypoints, pathSteps, startTime]);
 
@@ -74,15 +79,38 @@ const TimesStops = ({
       className="time-stops-datasheet"
       columns={columns}
       value={rows}
-      onChange={(row: PathWaypointRow[], [op]) => {
+      onChange={(newRows: PathWaypointRow[], [op]) => {
         if (!isInputTable) {
           return;
         }
-        const newRowData = transformRowDataOnChange(row[op.fromRowIndex], rows[op.fromRowIndex], op, allWaypoints.length);
-        if (!newRowData.isMarginValid) {
-          setRows(row);
+        let updatedRows = [...newRows];
+        updatedRows[op.fromRowIndex] = updateRowTimesAndMargin(
+          newRows[op.fromRowIndex],
+          rows[op.fromRowIndex],
+          op,
+          allWaypoints.length
+        );
+        updatedRows = updateDaySinceDeparture(updatedRows, startTime);
+        if (!updatedRows[op.fromRowIndex].isMarginValid) {
+          newRows[op.fromRowIndex].isMarginValid = false;
+          setRows(newRows);
+        } else if (
+          !rows[op.fromRowIndex].isMarginValid &&
+          updatedRows[op.fromRowIndex].isMarginValid
+        ) {
+          newRows[op.fromRowIndex].isMarginValid = true;
+          setRows(newRows);
         } else {
-          dispatch(upsertViaFromSuggestedOP(newRowData as SuggestedOP));
+          const newVias = updatedRows
+            .filter((row, index) => !isEqual(row, rows[index]))
+            .map(
+              (row) =>
+                ({
+                  ...row,
+                  ...(row.arrival && { arrival: durationSinceStartTime(startTime, row.arrival) }),
+                }) as SuggestedOP
+            );
+          dispatch(upsertSeveralViasFromSuggestedOP(newVias));
         }
       }}
       stickyRightColumn={stickyRightColumn}
@@ -94,7 +122,7 @@ const TimesStops = ({
           activeRow:
             rowIndex === 0 ||
             rowIndex === allWaypoints.length - 1 ||
-            isVia(pathSteps || [], rowData, WITH_KP),
+            isVia(pathSteps || [], rowData, { withKP: true }),
         })
       }
       cellClassName={cellClassName}
--- a/front/src/modules/timesStops/TimesStopsInput.tsx
+++ b/front/src/modules/timesStops/TimesStopsInput.tsx
@@ -1,6 +1,7 @@
 /* eslint-disable react/jsx-no-useless-fragment */
 
 import { useOsrdConfActions } from 'common/osrdContext';
+import { isVia } from 'modules/pathfinding/utils';
 import type { SuggestedOP } from 'modules/trainschedule/components/ManageTrainSchedule/types';
 import type { PathStep } from 'reducers/osrdconf/types';
 import { useAppDispatch } from 'store';
@@ -16,11 +17,18 @@ type ClearButtonProps = {
   pathSteps: PathStep[];
 };
 
-const createClearViaButton = ({ removeVia, rowIndex, rowData, allWaypoints }: ClearButtonProps) => {
+const createClearViaButton = ({
+  removeVia,
+  rowIndex,
+  rowData,
+  allWaypoints,
+  pathSteps,
+}: ClearButtonProps) => {
   const isClearBtnShown =
     allWaypoints &&
     rowIndex > 0 &&
     rowIndex < allWaypoints.length - 1 &&
+    isVia(pathSteps || [], rowData, { withKP: true }) &&
     (rowData.stopFor !== undefined ||
       rowData.theoreticalMargin !== undefined ||
       rowData.arrival !== undefined ||
@@ -79,7 +87,6 @@ const TimesStopsinput = ({ allWaypoints, startTime, pathSteps }: TimesStopsInput
     });
     dispatch(updatePathSteps({ pathSteps: updatedPathSteps }));
   };
-
   return (
     <TimesStops
       allWaypoints={allWaypoints}
--- a/front/src/modules/timesStops/TimesStopsOutput.tsx
+++ b/front/src/modules/timesStops/TimesStopsOutput.tsx
@@ -53,8 +53,8 @@ const TimesStopsOutput = ({
       tableType={TableType.Output}
       cellClassName={({ rowData: rowData_ }) => {
         const rowData = rowData_ as PathWaypointRow;
-        const arrivalScheduleNotRespected = rowData.arrival
-          ? rowData.calculatedArrival !== rowData.arrival
+        const arrivalScheduleNotRespected = rowData.arrival?.time
+          ? rowData.calculatedArrival !== rowData.arrival.time
           : false;
         const negativeDiffMargins = Number(rowData.diffMargins?.split(NO_BREAK_SPACE)[0]) < 0;
         return cx({
--- a/front/src/modules/timesStops/helpers/__tests__/utils.spec.ts
+++ b/front/src/modules/timesStops/helpers/__tests__/utils.spec.ts
@@ -1,75 +1,610 @@
-import type { PathWaypointRow } from 'modules/timesStops/types';
-import { transformRowDataOnChange } from '../utils';
+import { describe, it, expect } from 'vitest';
 
+import { type PathWaypointRow } from 'modules/timesStops/types';
 
-describe('transformRowDataOnChange', () => {
+import {
+  updateRowTimesAndMargin,
+  updateDaySinceDeparture,
+  durationSinceStartTime,
+  calculateStepTimeAndDays,
+} from '../utils';
+
+describe('updateRowTimesAndMargin', () => {
   const whateverOperation = { fromRowIndex: 2 };
 
   describe('arrival is set, departure just changed', () => {
     it('should update stop duration from the arrival and departure', () => {
       const rowData = {
-        opId: 'd94a2af4-6667-11e3-89ff-01f464e0362d',
+        opId: 'd94a2af4',
         name: 'Gr',
-        arrival: '23:40:00',
-        departure: '23:50:00',
+        arrival: { time: '23:40:00' },
+        departure: { time: '23:50:00' },
         stopFor: '300', // no longer correct, not yet updated by the function
       } as PathWaypointRow;
       const previousRowData = {
-        opId: 'd94a2af4-6667-11e3-89ff-01f464e0362d',
+        opId: 'd94a2af4',
         name: 'Gr',
-        arrival: '23:40:00',
-        departure: '23:45:00',
+        arrival: { time: '23:40:00' },
+        departure: { time: '23:45:00' },
         stopFor: '300',
       } as PathWaypointRow;
-      const result = transformRowDataOnChange(rowData, previousRowData, whateverOperation, 4);
+      const result = updateRowTimesAndMargin(rowData, previousRowData, whateverOperation, 4);
       expect(result).toEqual({
-        opId: 'd94a2af4-6667-11e3-89ff-01f464e0362d',
+        opId: 'd94a2af4',
         name: 'Gr',
-        arrival: '23:40:00',
-        departure: '23:50:00',
+        arrival: { time: '23:40:00' },
+        departure: { time: '23:50:00' },
         stopFor: '600', // now correct with the new arrival and departure
         isMarginValid: true,
       });
     });
   });
-  describe('theoritical margin is incorrect', () => {
+  describe('theoretical margin is incorrect', () => {
     it('should set isMarginValid flag to false', () => {
       const rowData = {
-        opId: 'd94a2af4-6667-11e3-89ff-01f464e0362d',
+        opId: 'd94a2af4',
         name: 'Gr',
         theoreticalMargin: '10',
       } as PathWaypointRow;
       const previousRowData = {
-        opId: 'd94a2af4-6667-11e3-89ff-01f464e0362d',
+        opId: 'd94a2af4',
         name: 'Gr',
       } as PathWaypointRow;
-      const result = transformRowDataOnChange(rowData, previousRowData, whateverOperation, 4);
+      const result = updateRowTimesAndMargin(rowData, previousRowData, whateverOperation, 4);
       expect(result.isMarginValid).toBe(false);
     });
   });
-  describe('user removed first row theoritical margin', () => {
-    it('should set the theoritical margin back to 0%', () => {
+  describe('user removed first row theoretical margin', () => {
+    it('should set the theoretical margin back to 0%', () => {
       const rowData = {
-        opId: 'd94a2af4-6667-11e3-89ff-01f464e0362d',
+        opId: 'd94a2af4',
         name: 'Gr',
       } as PathWaypointRow;
       const previousRowData = {
-        opId: 'd94a2af4-6667-11e3-89ff-01f464e0362d',
+        opId: 'd94a2af4',
         name: 'Gr',
         theoreticalMargin: '10%',
       } as PathWaypointRow;
       const operation = {
         fromRowIndex: 0,
       };
-      const result = transformRowDataOnChange(rowData, previousRowData, operation, 4);
+      const result = updateRowTimesAndMargin(rowData, previousRowData, operation, 4);
       expect(result).toEqual({
-        opId: 'd94a2af4-6667-11e3-89ff-01f464e0362d',
+        opId: 'd94a2af4',
         name: 'Gr',
-        arrival: null,
+        arrival: undefined,
         isMarginValid: true,
         onStopSignal: false,
         theoreticalMargin: '0%',
       });
+    });
+  });
+  describe('arrival is before midnight, departure after midnight', () => {
+    it('should compute the stopFor correctly', () => {
+      const rowData = {
+        opId: 'd94a2af4',
+        name: 'Gr',
+        arrival: { time: '23:40:00' },
+        departure: { time: '00:20:00' },
+        stopFor: '300',
+      } as PathWaypointRow;
+      const previousRowData = {
+        opId: 'd94a2af4',
+        name: 'Gr',
+        arrival: { time: '23:40:00' },
+        departure: { time: '23:45:00' },
+        stopFor: '300',
+      } as PathWaypointRow;
+      const result = updateRowTimesAndMargin(rowData, previousRowData, whateverOperation, 4);
+      expect(result).toEqual({
+        opId: 'd94a2af4',
+        name: 'Gr',
+        arrival: { time: '23:40:00' },
+        departure: { time: '00:20:00' },
+        stopFor: '2400',
+        isMarginValid: true,
+      });
+    });
+  });
+  describe('arrival, departure & stopFor are set, arrival gets erased', () => {
+    it('should keep stopFor and remove departure', () => {
+      const rowData = {
+        opId: 'd94a2af4',
+        name: 'Gr',
+        arrival: undefined,
+        departure: { time: '00:20:00' },
+        stopFor: '600',
+      } as PathWaypointRow;
+      const previousRowData = {
+        opId: 'd94a2af4',
+        name: 'Gr',
+        arrival: { time: '00:10:00' },
+        departure: { time: '00:20:00' },
+        stopFor: '600',
+      } as PathWaypointRow;
+      const result = updateRowTimesAndMargin(rowData, previousRowData, whateverOperation, 4);
+      expect(result).toEqual({
+        opId: 'd94a2af4',
+        name: 'Gr',
+        arrival: undefined,
+        departure: undefined,
+        stopFor: '600',
+        isMarginValid: true,
+      });
+    });
+    it('should keep stopFor and remove departure (double click + delete button version', () => {
+      const rowData = {
+        opId: 'd94a2af4',
+        name: 'Gr',
+        arrival: { time: '' },
+        departure: { time: '00:20:00' },
+        stopFor: '600',
+      } as PathWaypointRow;
+      const previousRowData = {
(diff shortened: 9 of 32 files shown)


front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -62,18 +66,26 @@ const computeBasePathSteps = (trainSchedule: TrainScheduleResult) =>
...
       stopFor: stopFor ? ISO8601Duration2sec(stopFor).toString() : stopFor,
       locked,
       onStopSignal,
     } as PathStep;
   });
+  // console.log(
+  //   'computeBasePathSteps',
+  //   basePathSteps,
+  //   updateDaySinceDeparture(basePathSteps, trainSchedule.start_time)
+  // );
```
theocrsb: You can delete this commented code

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -185,6 +197,7 @@ const useSetupItineraryForTrainUpdate = (
         });
       }
 
+      // console.log('updatedPathSteps', trainSchedule, updatedPathSteps);
```
theocrsb: Same here

theocrsb:
image
When I try to add 1 hour to 23 hours the requested departure time is not displayed. It is displayed when it is not midnight.

theocrsb:
image
I don't know if this use case should be possible, but when I try to start the train 2 days later (172800 seconds), the requested departure time only shows J+1.

front/src/modules/timesStops/ReadOnlyTime.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import type { TimeExtraDays } from 'modules/trainschedule/components/ManageTrainSchedule/types';
+
+type ReadOnlyTimeProps = CellProps<TimeExtraDays | null | undefined, string>;
+
+const ReadOnlyTime = ({ rowData }: ReadOnlyTimeProps) => {
+  const { time, daySinceDeparture, dayDisplayed } = rowData || {};
+  if (time) {
+    const fullString =
+      daySinceDeparture !== undefined && dayDisplayed ? `${time} J+${daySinceDeparture}` : time;
```
theocrsb: Perhaps this ‘J+’ should be included in the translation files. To get ‘D+’ in English

front/src/modules/timesStops/TimeInput.tsx
```diff
@@ -49,6 +48,16 @@ const TimeInput = ({
       }}
     />
   );
+
+  if (tempTimeValue?.daySinceDeparture && tempTimeValue.dayDisplayed) {
+    return (
+      <div className="time-input-container">
+        {input}
+        <span className="extra-text">J+{tempTimeValue.daySinceDeparture}</span>
```
theocrsb: Same here for translation files

theocrsb (dismissed):

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -62,18 +66,27 @@ const computeBasePathSteps = (trainSchedule: TrainScheduleResult) =>
       name = step.operational_point;
     }
 
+    const fomattedArrival: TimeExtraDays | undefined = arrival
```
kmer2016:
```suggestion
    const formattedArrival: TimeExtraDays | undefined = arrival
```

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
anisometropie: done

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
anisometropie: done

anisometropie:
> image I don't know if this use case should be possible, but when I try to start the train 2 days later (172800 seconds), the requested departure time only shows J+1.

It’s not a case we currently choose to handle

front/src/modules/timesStops/ReadOnlyTime.tsx
anisometropie: done

front/src/modules/timesStops/TimeInput.tsx
anisometropie: done

front/src/modules/timesStops/ReadOnlyTime.tsx
```diff
@@ -0,0 +1,25 @@
+import type { CellProps } from 'react-datasheet-grid/dist/types';
+import { useTranslation } from 'react-i18next';
+
+import { NO_BREAK_SPACE } from 'utils/strings';
+
+import type { TimeExtraDays } from './types';
+
+type ReadOnlyTimeProps = CellProps<TimeExtraDays | undefined | undefined, string>;
```
kmer2016:
```suggestion
type ReadOnlyTimeProps = CellProps<TimeExtraDays | undefined, string>;
```

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -1,24 +1,38 @@
...
-import { TableType } from '../types';
-import type { PathStepOpPointCorrespondance, PathWaypointRow } from '../types';
+import {
+  TableType,
+  type TimeExtraDays,
+  type PathStepOpPointCorrespondance,
+  type PathWaypointRow,
+} from '../types';
 
 export const formatSuggestedViasToRowVias = (
```
kmer2016, line 32: Should be nice If we test this function, not necessary in this PR

kmer2016 (commented):
Nice work ! Thank you for unit test ! Just few comments

front/src/common/types.ts
```diff
@@ -15,8 +15,23 @@ export const DATA_TYPES = {
...
   begin: number;
   end: number;
   value: string;
 };
+
+/**
+ * A ISO 8601 duration string
+ * @example "PT3600S"
+ */
+export type IsoDurationString = string;
```
SharglutDev, line 31: Maybe we could keep all these types together and move it before `RangedValue` :)

front/src/reducers/osrdconf/helpers.ts
```diff
@@ -142,3 +145,47 @@ export const updateDestinationPathStep = (
   destination: Partial<PathStep> | null,
   replaceCompletely: boolean = false
 ) => updatePathStepAtIndex(pathSteps, pathSteps.length - 1, destination, replaceCompletely);
+
+/**
+ * modifies the array statePathSteps in place
```
SharglutDev: Maybe to clarify why we chose to mutate the parameter

```suggestion
 * Modifies the array statePathSteps in place in the reducer
```

front/src/reducers/osrdconf/helpers.ts
```diff
@@ -142,3 +145,47 @@ export const updateDestinationPathStep = (
...
+      'theoreticalMargin',
+    ]),
+    id: nextId(),
+    ...(op.uic
+      ? { uic: op.uic }
+      : {
+          track: op.track,
+          offset: op.offsetOnTrack,
+        }),
+  } as PathStep;
```
SharglutDev: Maybe it's better not to force the type

```suggestion
  let newVia: PathStep = {
    ...pick(op, [
      'coordinates',
      'positionOnPath',
      'name',
      'ch',
      'kp',
      'stopFor',
      'arrival',
      'locked',
      'deleted',
      'onStopSignal',
      'theoreticalMargin',
    ]),
    id: nextId(),
    ...(op.uic
      ? { uic: op.uic }
      : {
          track: op.track,
          offset: op.offsetOnTrack,
        }),
  };
```

front/src/reducers/osrdconf/helpers.ts
SharglutDev: Should we add a test for it ?

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -71,9 +70,7 @@ const computeBasePathSteps = (trainSchedule: TrainScheduleResult) =>
       ...stepWithoutSecondaryCode,
       ch: 'secondary_code' in step ? step.secondary_code : undefined,
       name,
-      arrival: arrival
-        ? addDurationToIsoDate(trainSchedule.start_time, arrival).substring(11, 19)
-        : arrival,
+      arrival,
```
SharglutDev: To keep track at what kind of arrival we have here ?

```suggestion
      arrival, // as ISODurationString
```

front/src/modules/timesStops/TimesStopsInput.tsx
SharglutDev, line 1: Could we keep the spacing for readability ?

front/src/modules/timesStops/types.ts
```diff
@@ -2,8 +2,16 @@ import type { TrainScheduleBase, TrainScheduleResult } from 'common/api/osrdEdit
 import type { SuggestedOP } from 'modules/trainschedule/components/ManageTrainSchedule/types';
 import type { ArrayElement } from 'utils/types';
 
-export type PathWaypointRow = SuggestedOP & {
+export type TimeExtraDays = {
+  time: string | null | undefined;
```
SharglutDev: Should it be a TimeString to know what we manipulate or it can be something else ?

```suggestion
  time?: TimeString | null;
```

front/src/modules/timesStops/types.ts
```diff
@@ -2,8 +2,16 @@ import type { TrainScheduleBase, TrainScheduleResult } from 'common/api/osrdEdit
...
+  daySinceDeparture?: number;
+  dayDisplayed?: boolean;
+};
+
+export type PathWaypointRow = Omit<SuggestedOP, 'arrival' | 'departure'> & {
   isMarginValid: boolean;
+  arrival?: TimeExtraDays | undefined; // value asked by user
+  departure?: TimeExtraDays | undefined; // value asked by user
```
SharglutDev:
```suggestion
  arrival?: TimeExtraDays; // value asked by user
  departure?: TimeExtraDays; // value asked by user
```

front/src/modules/timesStops/TimesStopsOutput.tsx
```diff
@@ -53,8 +53,8 @@ const TimesStopsOutput = ({
       tableType={TableType.Output}
       cellClassName={({ rowData: rowData_ }) => {
         const rowData = rowData_ as PathWaypointRow;
-        const arrivalScheduleNotRespected = rowData.arrival
-          ? rowData.calculatedArrival !== rowData.arrival
+        const arrivalScheduleNotRespected = rowData?.arrival?.time
```
SharglutDev: If you cast on the upper line, you shouldn't need this ?

```suggestion
        const arrivalScheduleNotRespected = rowData.arrival?.time
```

front/src/modules/timesStops/TimeInput.tsx
SharglutDev, line 1: There is a weird display here (input table)

image

front/src/modules/timesStops/TimesStopsOutput.tsx
SharglutDev, line 1: I added a constraint on this OP in the input table (so the row becomes bold and a path step is added in store) but here in the output table, its not bold

image

front/src/modules/timesStops/TimesStops.tsx
```diff
@@ -74,35 +79,30 @@ const TimesStops = ({
...
-          }
-          dispatch(upsertViaFromSuggestedOP(rowData as SuggestedOP));
+          const newVias = updatedRows
+            .filter((row, index) => !isEqual(row, rows[index]))
+            .map((row) => {
+              return {
+                ...row,
+                ...(row.arrival && { arrival: durationSinceStartTime(startTime, row.arrival) }),
+              } as SuggestedOP;
+            });
```
SharglutDev: To remove the eslint warning

```suggestion
            .map(
              (row) =>
                ({
                  ...row,
                  ...(row.arrival && { arrival: durationSinceStartTime(startTime, row.arrival) }),
                }) as SuggestedOP
            );
```

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -125,3 +139,145 @@ export function disabledTextColumn(
...
+  isoDuration?: IsoDurationString | null
+): TimeExtraDays | undefined {
+  if (!startTime || !isoDuration) {
+    return undefined;
+  }
+
+  const start = dayjs(startTime);
+  const duration = dayjs.duration(isoDuration);
+
+  const endTime = start.add(duration);
```
SharglutDev: Maybe it's more understandable like this ?

```suggestion
  const waypointArrivalTime = start.add(duration);
```

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -125,3 +139,145 @@ export function disabledTextColumn(
...
+  const daySinceDeparture = endTime.diff(start, 'day');
+  const time = endTime.format('HH:mm:ss');
```
SharglutDev: To know what we manipulate

```suggestion
  const time: TimeString = endTime.format('HH:mm:ss');
```

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -56,21 +70,21 @@ export const formatSuggestedViasToRowVias = (
...
-    if (stopFor) {
-      if (i === 0) {
-        departure = startTime
-          ? secToHoursString(datetime2sec(new Date(startTime)) + Number(stopFor), true)
-          : undefined;
-      } else if (arrival) {
-        departure = secToHoursString(time2sec(arrival) + Number(stopFor), true);
-      }
-    }
+    const roundedArrivalTime = i === 0 ? 'PT0S' : arrival;
```
SharglutDev: Maybe it's clearer like that

```suggestion
    const durationArrivalTime = i === 0 ? 'PT0S' : arrival;
```

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -56,21 +70,21 @@ export const formatSuggestedViasToRowVias = (
...
+    const arrivalInSeconds = roundedArrivalTime ? time2sec(roundedArrivalTime) : null;
+
+    const formattedArrival = calculateStepTimeAndDays(startTime, roundedArrivalTime);
+
+    const departureTime =
+      stopFor && arrivalInSeconds
+        ? secToHoursString(arrivalInSeconds + Number(stopFor), true)
```
SharglutDev: Instead of just passing true maybe it's clearer like that (and we add a `const WITH_SECONDS = true;` at the top of the file)

```suggestion
        ? secToHoursString(arrivalInSeconds + Number(stopFor), WITH_SECONDS)
```

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -56,21 +70,21 @@ export const formatSuggestedViasToRowVias = (
...
+        : undefined;
+    const formattedDeparture = departureTime ? { time: departureTime } : undefined;
```
SharglutDev:
```suggestion
    const formattedDeparture: TimeExtraDays | undefined = departureTime
      ? { time: departureTime }
      : undefined;
```

front/src/modules/timesStops/helpers/scheduleData.ts
```diff
@@ -7,31 +8,31 @@ import type { ComputedScheduleEntry, ScheduleEntry } from '../types';
  * @param schedule for a given operational point
  * @param startTime time of departure from the very beginning of the route
```
SharglutDev:
```suggestion
```

front/src/modules/timesStops/helpers/__tests__/utils.spec.ts
```diff
@@ -0,0 +1,503 @@
...
+    const result = durationSinceStartTime(startTime, stepTimeDays);
+
+    expect(result).toBe('PT90000S');
+  });
+});
+
+describe('calculateStepTimeDays', () => {
+  it('should return correct time and daySinceDeparture', () => {
+    const startTime = '2023-09-01T10:00:00Z';
+    const isoDuration = 'PT36000S';
```
SharglutDev:
```suggestion
    const isoDuration = 'PT36000S'; // 10 hours
```

front/src/modules/timesStops/helpers/__tests__/utils.spec.ts
```diff
@@ -0,0 +1,503 @@
...
+
+    expect(result).toEqual({
+      time: '20:00:00',
+      daySinceDeparture: 0,
+    });
+  });
+
+  it('should return correct time and daySinceDeparture, daySinceDeparture 1', () => {
+    const startTime = '2023-09-01T10:00:00Z';
+    const isoDuration = 'PT122400S';
```
SharglutDev:
```suggestion
    const isoDuration = 'PT122400S'; // 1 day 10 hours
```

front/src/modules/timesStops/helpers/__tests__/utils.spec.ts
```diff
@@ -0,0 +1,503 @@
...
+        arrival: { time: '23:40:00' },
+        departure: { time: '00:20:00' },
+        stopFor: '2400',
+        isMarginValid: true,
+      });
+    });
+  });
+});
+
+describe('updateTimeAndDays', () => {
```
SharglutDev: Maybe it's better to put the name of the function here like we usually do

```suggestion
describe('updateDaySinceDeparture', () => {
```

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -125,3 +139,145 @@ export function disabledTextColumn(
...
+
+/**
+ * This function goes through the whole array of path waypoints
+ * and updates the number of days since departure.
+ */
+export function updateDaySinceDeparture<T extends Pick<PathWaypointRow, 'arrival' | 'stopFor'>>(
+  pathWaypointRows: T[],
+  startTime?: IsoDateTimeString,
+  keepFirstIndexArrival?: boolean
+): T[] {
```
SharglutDev: Why can't we just do it like that since everywhere we use this function we pass it a `PathWaypointRow` array ?

```suggestion
export function updateDaySinceDeparture(
  pathWaypointRows: PathWaypointRow[],
  startTime?: IsoDateTimeString,
  keepFirstIndexArrival?: boolean
): PathWaypointRow[] {
```

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -125,3 +139,145 @@ export function disabledTextColumn(
...
+      }
+      previousTime = arrivalInSeconds;
+    }
+
+    let formattedDeparture: TimeExtraDays | undefined;
+    if (stopFor && arrivalInSeconds) {
+      const departureInSeconds = (arrivalInSeconds + Number(stopFor)) % SECONDS_IN_A_DAY;
+      const isAfterMidnight = departureInSeconds < previousTime;
+      if (isAfterMidnight) {
+        currentDaySinceDeparture += 1;
```
SharglutDev, line 235: image

Probably not gonna happen in real life but I used a stop_for > 1 day but it still increment by one only

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -125,3 +139,145 @@ export function disabledTextColumn(
...
+        formattedDeparture = {
+          time: secToHoursString(departureInSeconds, true),
+          daySinceDeparture: currentDaySinceDeparture,
+          dayDisplayed: true,
+        };
+      } else {
+        formattedDeparture = {
+          time: secToHoursString(departureInSeconds, true),
```
SharglutDev: Same as previous comment

```suggestion
          time: secToHoursString(departureInSeconds, WITH_SECONDS),
          daySinceDeparture: currentDaySinceDeparture,
          dayDisplayed: true,
        };
      } else {
        formattedDeparture = {
          time: secToHoursString(departureInSeconds, WITH_SECONDS),
```

front/src/modules/timesStops/helpers/utils.ts
SharglutDev, line 235: Another (unlikely) edge case : 
image

It seems a `stop_for` making the day change is not taken into account (should be a D+2)

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -125,3 +139,145 @@ export function disabledTextColumn(
...
+          daySinceDeparture: currentDaySinceDeparture,
+        };
+      }
+      previousTime = departureInSeconds;
+    }
+
+    return {
+      ...pathWaypoint,
+      arrival: keepFirstIndexArrival || index > 0 ? formattedArrival : undefined,
```
SharglutDev, line 252: Not sure what `keepFirstIndexArrival` does ?

front/src/modules/timesStops/helpers/__tests__/utils.spec.ts
```diff
@@ -0,0 +1,503 @@
...
+        opId: 'd94a2af4',
+        name: 'Gr',
+        arrival: { time: '23:40:00' },
+        departure: { time: '23:50:00' },
+        stopFor: '600', // now correct with the new arrival and departure
+        isMarginValid: true,
+      });
+    });
+  });
+  describe('theoritical margin is incorrect', () => {
```
SharglutDev:
```suggestion
  describe('theoretical margin is incorrect', () => {
```

front/src/modules/timesStops/helpers/__tests__/utils.spec.ts
```diff
@@ -0,0 +1,503 @@
...
+      const previousRowData = {
+        opId: 'd94a2af4',
+        name: 'Gr',
+      } as PathWaypointRow;
+      const result = updateRowTimesAndMargin(rowData, previousRowData, whateverOperation, 4);
+      expect(result.isMarginValid).toBe(false);
+    });
+  });
+  describe('user removed first row theoritical margin', () => {
+    it('should set the theoritical margin back to 0%', () => {
```
SharglutDev:
```suggestion
  describe('user removed first row theoretical margin', () => {
    it('should set the theoretical margin back to 0%', () => {
```

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -125,3 +139,145 @@ export function disabledTextColumn(
     disabled: true,
   };
 }
+
+/**
+ * Synchronizes arrival, departure and stop times.
+ * updates onStopSignal
+ * updates isMarginValid and theoreticalMargin
+ */
+export function updateRowTimesAndMargin(
```
SharglutDev, line 151: I don't think it comes from your PR but if I start by typing a departure time in an empty row, it's not displayed. But if I click on the departure cell, the value is still there and typing an arrival time or a stopFor doesn't display it either.

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -125,3 +139,145 @@ export function disabledTextColumn(
...
+): PathWaypointRow {
+  const newRowData = { ...rowData };
+  if (
+    newRowData.departure &&
+    newRowData.arrival &&
+    (!isEqual(newRowData.arrival, previousRowData.arrival) ||
+      !isEqual(newRowData.departure, previousRowData.departure))
+  ) {
+    newRowData.stopFor = String(
+      durationInSeconds(
```
SharglutDev: In this util, you can replace the hard coded `86400` by your new const `SECONDS_IN_A_DAY`

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -125,3 +139,145 @@ export function disabledTextColumn(
...
+        time2sec(newRowData.arrival.time || ''),
+        time2sec(newRowData.departure.time || '')
```
SharglutDev: Is it possible to go in the fallback even though we checked lines 156/157 that departure and arrival are defined ? (if yes in what usecase ?)

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -125,3 +139,145 @@ export function disabledTextColumn(
...
+    newRowData.theoreticalMargin && !marginRegExValidation.test(newRowData.theoreticalMargin)
+  );
+  if (newRowData.isMarginValid && op.fromRowIndex === 0) {
+    newRowData.arrival = undefined;
+    // As we put 0% by default for origin's margin, if the user removes a margin without
+    // replacing it to 0% (undefined), we change it to 0%
+    if (!newRowData.theoreticalMargin) {
+      newRowData.theoreticalMargin = '0%';
+    }
+  }
```
SharglutDev, line 192: I have a weird thing going one with margins. If I enter "10" it doesn't update the store and it's not displayed as "invalid" (in red). The store is updated only if I enter a valid value.

[Capture vidéo du 11-09-2024 17:58:15.webm](https://github.com/user-attachments/assets/8885f0f7-0832-470f-9403-0f9482f60a48)

front/src/modules/timesStops/TimesStops.tsx
```diff
@@ -74,35 +79,30 @@ const TimesStops = ({
...
-            // replacing it to 0% (undefined), we change it to 0%
-            if (!rowData.theoreticalMargin) rowData.theoreticalMargin = '0%';
-          }
-          dispatch(upsertViaFromSuggestedOP(rowData as SuggestedOP));
+          const newVias = updatedRows
+            .filter((row, index) => !isEqual(row, rows[index]))
+            .map((row) => {
+              return {
+                ...row,
+                ...(row.arrival && { arrival: durationSinceStartTime(startTime, row.arrival) }),
```
SharglutDev: Maybe it's clearer like that ?

```suggestion
                arrival: row.arrival && durationSinceStartTime(startTime, row.arrival),
```

front/src/modules/timesStops/helpers/utils.ts
SharglutDev, line 151: Another strange thing : 

[Capture vidéo du 11-09-2024 18:09:28.webm](https://github.com/user-attachments/assets/3a3eab44-9d5a-464b-b708-773988e0389c)

SharglutDev (changes requested):
Thank you for this PR, didn't have time to check all files but here are some comments, I'm opened to discussion if needed

emersion:
TBH I am not a huge fan of these new types which don't improve type safety (since they're aliases)…

SharglutDev:
> TBH I am not a huge fan of these new types which don't improve type safety (since they're aliases)…

I agree but it doesn't add to much logic in the code. When the refacto for that will be done it can be adapted ? Or do you prefer to keep `string`everywhere in the meantime ?

front/public/locales/en/timesStops.json
```diff
@@ -2,6 +2,7 @@
   "arrivalTime": "Requested arrival Time",
   "calculatedArrivalTime": "Calculated arrival time",
   "calculatedDepartureTime": "Calculated departure time",
+  "d+": "D+",
```
SharglutDev: The key is not very explicit. Maybe we could do that ? How about passing the count inside also ?

```suggestion
  "dayCounter": "D+",
```

front/src/modules/timesStops/helpers/scheduleData.ts
```diff
@@ -7,31 +8,31 @@ import type { ComputedScheduleEntry, ScheduleEntry } from '../types';
...
     stopFor: stopForSeconds,
   };
 }
 
-export function formatScheduleData(scheduleData: ComputedScheduleEntry) {
+export function formatScheduleData(
+  scheduleData: ComputedScheduleEntry
+): Pick<SuggestedOP, 'arrival' | 'departure' | 'stopFor'> {
+  const arrival = scheduleData.arrival ? formatDurationAsISO8601(scheduleData.arrival) : null;
+  const departure = scheduleData.departure ? formatDurationAsISO8601(scheduleData.departure) : null;
```
SharglutDev: To keep track of the kind of time we use

```suggestion
  const arrival: IsoDurationString | null = scheduleData.arrival
    ? formatDurationAsISO8601(scheduleData.arrival)
    : null;
  const departure: IsoDurationString | null = scheduleData.departure
    ? formatDurationAsISO8601(scheduleData.departure)
    : null;
```

front/src/modules/timesStops/ReadOnlyTime.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const ReadOnlyTime = ({ rowData }: ReadOnlyTimeProps) => {
+  const { t } = useTranslation('timesStops');
+  const { time, daySinceDeparture, dayDisplayed } = rowData || {};
+  if (time) {
+    const fullString =
+      daySinceDeparture !== undefined && dayDisplayed
+        ? `${time}${NO_BREAK_SPACE}${t('d+')}${daySinceDeparture}`
+        : time;
+    return <div className="read-only-time">{fullString}</div>;
```
SharglutDev: Do we need this class if you put just a padding inside ?

front/src/modules/timesStops/ReadOnlyTime.tsx
```diff
@@ -0,0 +1,25 @@
...
+  }
+  return null;
```
SharglutDev: Even though the component is small, we could keep the early return like we do in other components : `if (!times) return null`

front/src/modules/timesStops/ReadOnlyTime.tsx
```diff
@@ -0,0 +1,25 @@
...
+};
+
+ReadOnlyTime.displayName = 'ReadOnlyTime';
```
SharglutDev: I'm not familiar with this, how does it work ? It seems like it's needed if we want to use a name different than the component one ?

front/src/modules/timesStops/TimeInput.tsx
```diff
@@ -26,7 +27,7 @@ const TimeInput = ({
     setTempTimeValue(rowData);
   }, [rowData]);
 
-  return (
+  const input = (
     <input
       className="dsg-input"
```
SharglutDev, line 34: What is `dsg` ?

front/src/modules/timesStops/ReadOnlyTime.tsx
```diff
@@ -0,0 +1,25 @@
...
+import type { TimeExtraDays } from './types';
+
+type ReadOnlyTimeProps = CellProps<TimeExtraDays | undefined | undefined, string>;
+
+const ReadOnlyTime = ({ rowData }: ReadOnlyTimeProps) => {
+  const { t } = useTranslation('timesStops');
+  const { time, daySinceDeparture, dayDisplayed } = rowData || {};
+  if (time) {
+    const fullString =
+      daySinceDeparture !== undefined && dayDisplayed
```
SharglutDev: We don't check `daySinceDeparture > 0` ? If somehow we change the startTime, some of the times might go from `daySinceDeparture: 1` to 0. But will `dayDisplayed` will pass to false automatically ?

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
anisometropie: done

front/src/modules/timesStops/ReadOnlyTime.tsx
anisometropie: oops, done !

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/__tests__/formatSchedule.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+        {
+          id: 'id331',
+          deleted: false,
+          uic: 8706,
+          ch: 'BV',
+          kp: '130+538',
+          name: 'G',
+          positionOnPath: 0,
+        },
+      ] as PathStep[];
```
SharglutDev: Same for others

```suggestion
      const pathSteps: PathStep[] = [
        {
          id: 'id331',
          deleted: false,
          uic: 8706,
          ch: 'BV',
          kp: '130+538',
          name: 'G',
          positionOnPath: 0,
        },
      ];
```

SharglutDev (changes requested):

front/src/modules/timesStops/helpers/utils.ts
anisometropie: sure, opened https://github.com/OpenRailAssociation/osrd/issues/8835

front/src/common/types.ts
anisometropie: done

front/src/reducers/osrdconf/helpers.ts
anisometropie: sure. It was existing code that I extracted from the reducer, I’ll do it in another PR.

I opened a ticket https://github.com/OpenRailAssociation/osrd/issues/8836

anisometropie: changed the tsdoc :)

front/src/reducers/osrdconf/helpers.ts
anisometropie: done

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
anisometropie: sure done

front/src/modules/timesStops/TimesStopsInput.tsx
anisometropie: yes

front/src/modules/timesStops/types.ts
anisometropie: it’s only ever a TimeString, done

front/src/modules/timesStops/types.ts
anisometropie: done

front/src/modules/timesStops/TimesStopsOutput.tsx
anisometropie: It’s not really to comply with the type, it’s a runtime safety. I think I head some crashes, but it doesn’t seem to happen if i remove it. 
let’s remove for now.

front/src/modules/timesStops/TimesStopsInput.tsx
anisometropie: actually https://github.com/OpenRailAssociation/osrd/actions/runs/10827960551/job/30042250527?pr=8334#step:6:363

front/src/modules/timesStops/TimesStopsOutput.tsx
anisometropie: was broken before: https://github.com/OpenRailAssociation/osrd/pull/8762

front/src/modules/timesStops/TimeInput.tsx
anisometropie: fixed

front/src/modules/timesStops/TimesStops.tsx
anisometropie: done

front/src/modules/timesStops/helpers/utils.ts
anisometropie: done

front/src/modules/timesStops/helpers/utils.ts
anisometropie: done

front/src/modules/timesStops/TimesStops.tsx
anisometropie: Here I don’t want to add the key at all. This would create an `arrival: undefined` entry

anisometropie:
> TBH I am not a huge fan of these new types which don't improve type safety (since they're aliases)…

@emersion 

They are meant to define clearly what they are, and give move details when you hover your mouse over them. 

I don’t want to lose the info and go back to a `string`. they can easily be changed with better types if we know where they are

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/__tests__/formatSchedule.spec.ts
anisometropie: done

front/src/modules/timesStops/ReadOnlyTime.tsx
anisometropie: The function `updateDaySinceDeparture` is responsible to provide sound data, it recompute daySinceDepature and dayDisplayed everytime a value change in the table. I don’t think we need to over check here

front/src/modules/timesStops/TimeInput.tsx
anisometropie: I think it was made so that `react-datasheet-grid` can use it, it’s classes used inside the library.

https://github.com/nick-keller/react-datasheet-grid/blob/master/src/style.css

front/src/modules/timesStops/ReadOnlyTime.tsx
anisometropie: it’s use for debugging and to have a name in react dev tools


Screenshot_20240913_121437

but it actually works without it here, i’ll remove it

@Uriel-Sautron

front/src/modules/timesStops/ReadOnlyTime.tsx
anisometropie: sure

front/src/modules/timesStops/ReadOnlyTime.tsx
anisometropie: there is no 10px padding in bootstrap, do you want to use inline style ?

front/src/modules/timesStops/helpers/scheduleData.ts
anisometropie: done

front/public/locales/en/timesStops.json
anisometropie: done

front/src/modules/timesStops/helpers/utils.ts
anisometropie: we don’t handle that case 
@axrolld

front/src/modules/timesStops/helpers/utils.ts
anisometropie: done

front/src/modules/timesStops/helpers/utils.ts
anisometropie: I’m a bit on the fence regarding that. I don’t think we want to create a variable for every single function that takes a boolean flag. 

It’s just a mouse hover away to get the info.

Screenshot_20240913_164345

@clarani @emersion what do you think?

front/src/modules/timesStops/helpers/utils.ts
anisometropie: done

front/src/modules/timesStops/helpers/scheduleData.ts
anisometropie: done

front/src/modules/timesStops/helpers/__tests__/utils.spec.ts
anisometropie: done

front/src/modules/timesStops/helpers/__tests__/utils.spec.ts
anisometropie: done

front/src/modules/timesStops/helpers/__tests__/utils.spec.ts
anisometropie: yes, that was forgotten when function was renamed

front/src/modules/timesStops/helpers/utils.ts
anisometropie: done; the function received two kind of object with arrival and departure, but it’s changed since the store changed.

front/src/modules/timesStops/helpers/utils.ts
anisometropie: https://github.com/OpenRailAssociation/osrd/pull/8334#discussion_r1759003784

front/src/modules/timesStops/helpers/utils.ts
anisometropie: updateDaySinceDeparture should not put an arrival value in the store for `index` `0`, but should when it’s formatting data from the store to be used by the table, we want to show a value at `index 0`

front/src/modules/timesStops/helpers/__tests__/utils.spec.ts
anisometropie: done

front/src/modules/timesStops/helpers/__tests__/utils.spec.ts
anisometropie: done

front/src/modules/timesStops/helpers/utils.ts
anisometropie: this needs to be discussed with @axrolld 
a departure time doesn’t makes sense on its own
since the departure doesn’t exist in DB, it’s stored stopFor value (computed using the arrival time) and back.

front/src/modules/timesStops/helpers/utils.ts
anisometropie: done

front/src/modules/timesStops/helpers/utils.ts
anisometropie: I think the TimeExtraDay with `{ time: undefined }` doesn’t make sense. we only need the case where the whole thing is `undefined`.
I changed `TimeExtraDay` so we don’t have the issue here.

```
TimeExtraDays = {
  time: TimeString;
  daySinceDeparture?: number;
  dayDisplayed?: boolean;
};
```

front/src/modules/timesStops/helpers/utils.ts
anisometropie: I think the original design decision to only put valid values in the store shows its limits.
I fixed it to update locally.
When you put an invalid value, it’s only going to update `rows` locally,  but when we update another cell that’s valid, it’s going to be update the entire table from the new data from the store. as the invalid cell was not in store, it’s going to get erase.
but it might be an acceptable behavior.

@axrolld we can check that together ?

front/src/modules/timesStops/helpers/utils.ts
anisometropie: This case is now handled (added a unit test)

front/public/locales/en/timesStops.json
SharglutDev: Thx. And for the `count` ? I think by putting it in the t function, there is less stuff in the jsx

front/src/modules/timesStops/ReadOnlyTime.tsx
SharglutDev: Ok if it needs to be specifically 10px, we can leave it like that. But in scss file, it needs to be translated in rem.

front/src/modules/timesStops/helpers/utils.ts
emersion: In general I'd prefer options objects e.g.

```ts
secToHoursString(arrivalInSeconds + Number(stopFor), { withSeconds: true })
```

since it's an idiomatic design pattern for JS APIs and doesn't require readers to have a full IDE set up (useful for reading on git(1) or GitHub, and for weird minimalists like me).

clarani: I think it would be better to have a new utils function secToHhmmss which calls secToHoursString with true, what do you think ?
Also we could rename the function secToHoursString into secToHhmmString, but maybe in another PR

clarani: I also agree with @emersion ;)

emersion: TBQH, I think the utils date API is due for a big overhaul to improve/unify all exposed functions. It's quite inconsistent at the moment.

front/src/modules/timesStops/helpers/utils.ts
SharglutDev: Hmm I still have the issue

front/src/modules/timesStops/helpers/utils.ts
SharglutDev: I also like @emersion suggestion

anisometropie: going for @emersion’s suggestion as we seem to all agree :)

front/src/modules/timesStops/helpers/utils.ts
axrolld: Yep, stop aren't meant to last more than 24h. I don't know if we should set a limit in the stop time column to prevent from this, what do you think ?

axrolld: Yes if a train arrives somewhere at 23h50 and stops for 20mn the day should be incremented

front/src/modules/timesStops/helpers/utils.ts
axrolld: Could we consider that a departure time on its own is equivalent to a passing time ? I'm saying this because a use case i see would be if the user wants to set a stop, but he/she starts with the departure time input, then a stop time for example. In my opinion the departure time should be displayed if it is set by the user, and used either as a passing time if it's on its own, of used normally if it comes with an arrival time or a stop time

axrolld: If we don't do this, setting a stop becomes dependant on the right order of the cells filling, which to me isn't great from a UX perspective

front/src/modules/timesStops/helpers/utils.ts
SharglutDev: I don't think it's needed but it can be done in another PR if we think it's important

front/src/modules/timesStops/helpers/utils.ts
SharglutDev: Ok maybe we can create an issue for this usecase to be solved as it doesn't really concern this PR

front/src/modules/timesStops/helpers/utils.ts
SharglutDev: Thank you for the change. The current behavior seems ok to me.

front/src/modules/timesStops/helpers/utils.ts
anisometropie: We’re gonna leave it like this for now, we created this ticket https://github.com/OpenRailAssociation/osrd/issues/8894

front/src/modules/timesStops/ReadOnlyTime.tsx
anisometropie: as discussed, leaving the 10px to conform with the other classes of the lib

front/public/locales/en/timesStops.json
anisometropie: done

front/src/modules/timesStops/helpers/utils.ts
anisometropie: added one more test (case where we get { time: '' } when we double clic first on the cell and erase only one part of the time (like `--:00:00`)

SharglutDev (approved):
Lgtm and tested, great work :)

PR #8336, opened 2024-08-05
front: using config stored in db for stdcm v2

### LINKED ISSUE #8239: front: update stdcm page to use the stdcm search env configured in the db
## Description

> [!warning]
> This issue depends on #8168 with PR #8257

### Workflow

The `stdcmV2` page queries the stdcm search env data set through the crud of #8168 (i.e. with something like GET /api/stdcm/search_environment). 

This data is then used to fill the attributes of the POST /api/stdcm request, and to limit the date times fields (see [the mockup](https://www.sketch.com/s/15e0909c-f17d-4200-a770-b9fc1154fa7e/a/9ljpLwx)). 
The overall design of the page isn't changed except for the scenario selector disappearing.

If there is no data (the GET /api/stdcm/search_environment returned a 204), then a page explaining that the application isn't properly configured should be displayed, giving instructions on how to alert maintainers of the problem.
[see the mockup (FR)](https://www.sketch.com/s/15e0909c-f17d-4200-a770-b9fc1154fa7e/a/3oqEreG) [see the mockup (EN)](https://www.sketch.com/s/15e0909c-f17d-4200-a770-b9fc1154fa7e/a/nQzxjpg)
(side note: ignore the hyperlink specified in the mockups, users already know who to contact (not primarily us))

### Acceptance criteria

- [x] The front uses the data set in the db
  - [x] The scenario selector is gone
  - [x] The date and time selectors are bound by the window set in the stdcm search env
- [x] An absence of stdcm search env data is handled gracefully
- [ ] (Could be done later) In debug version, the user can either use the original scenario selector or the  stdcm search env configured in the db
- [ ] (Ideally) the debug page allows a user to set the stdcm search env
- [ ] (Even more ideally), the debug page allows a user to set stdcm search env fields with helpful selectors, and not just plain text fields

sim51:
Fix #8239

To test it, you need to insert a row in the table `stdcm_search_environment`

### DIFF
--- a/editoast/src/views/v2/timetable/stdcm.rs
+++ b/editoast/src/views/v2/timetable/stdcm.rs
@@ -42,6 +42,8 @@ use crate::RetrieveBatch;
 use editoast_models::DbConnection;
 use editoast_models::DbConnectionPoolV2;
 
+use super::SelectionSettings;
+
 crate::routes! {
     "/stdcm" => stdcm,
 }
@@ -76,6 +78,8 @@ pub struct STDCMRequestPayload {
     start_time: Option<DateTime<Utc>>,
     steps: Vec<PathfindingItem>,
     rolling_stock_id: i64,
+    electrical_profile_set_id: Option<i64>,
+    work_schedule_group_id: Option<i64>,
     comfort: Comfort,
     /// By how long we can shift the departure time in milliseconds
     /// Deprecated, first step data should be used instead
@@ -190,7 +194,7 @@ async fn stdcm(
         core_client.clone(),
         &trains,
         &infra,
-        None,
+        data.electrical_profile_set_id,
     )
     .await?;
 
@@ -245,13 +249,19 @@ async fn stdcm(
         time_gap_after: data.time_gap_after,
         margin: data.margin,
         time_step: Some(2000),
-        work_schedules: build_work_schedules(
-            db_pool.get().await?.deref_mut(),
-            departure_time,
-            data.maximum_departure_delay,
-            maximum_run_time,
-        )
-        .await?,
+        work_schedules: match data.work_schedule_group_id {
+            Some(work_schedule_group_id) => {
+                build_work_schedules(
+                    db_pool.get().await?.deref_mut(),
+                    departure_time,
+                    data.maximum_departure_delay,
+                    maximum_run_time,
+                    work_schedule_group_id,
+                )
+                .await?
+            }
+            None => vec![],
+        },
     }
     .fetch(core_client.as_ref())
     .await?;
@@ -468,9 +478,12 @@ async fn build_work_schedules(
     time: DateTime<Utc>,
     maximum_departure_delay: u64,
     maximum_run_time: u64,
+    work_schedule_group_id: i64,
 ) -> Result<Vec<STDCMWorkSchedule>> {
     let maximum_simulation_time = maximum_run_time + maximum_departure_delay;
-    let res = Ok(WorkSchedule::list(conn, Default::default())
+    let selection_setting: SelectionSettings<WorkSchedule> = SelectionSettings::new()
+        .filter(move || WorkSchedule::WORK_SCHEDULE_GROUP_ID.eq(work_schedule_group_id));
+    let res = Ok(WorkSchedule::list(conn, selection_setting)
         .await?
         .iter()
         .map(|ws| {
--- a/front/src/applications/stdcm/utils/formatStdcmConfV2.ts
+++ b/front/src/applications/stdcm/utils/formatStdcmConfV2.ts
@@ -11,7 +11,7 @@ import type {
 import type { InfraState } from 'reducers/infra';
 import { setFailure } from 'reducers/main';
 import type { OsrdStdcmConfState, StandardAllowance } from 'reducers/osrdconf/types';
-import { dateTimeToIso } from 'utils/date';
+import { dateTimeFormatting, dateTimeToIso } from 'utils/date';
 import { mToMm } from 'utils/physics';
 import { ISO8601Duration2sec, sec2ms, time2sec } from 'utils/timeManipulation';
 
@@ -31,6 +31,8 @@ type ValidStdcmConfig = {
   margin?: StandardAllowance;
   gridMarginBefore?: number;
   gridMarginAfter?: number;
+  workScheduleGroupId?: number;
+  electricalProfileSetId?: number;
 };
 
 export const checkStdcmConf = (
@@ -52,6 +54,9 @@ export const checkStdcmConf = (
     originUpperBoundTime,
     originDate,
     originTime,
+    searchDatetimeWindow,
+    workScheduleGroupId,
+    electricalProfileSetId,
   } = osrdconf;
   let error = false;
   if (pathSteps[0] === null) {
@@ -117,6 +122,23 @@ export const checkStdcmConf = (
         message: t('operationalStudies/manageTrainSchedule:errorMessages.noOriginTime'),
       })
     );
+  } else if (searchDatetimeWindow) {
+    const startDatetime = new Date(startTime);
+    if (startDatetime < searchDatetimeWindow.begin || searchDatetimeWindow.end < startDatetime) {
+      error = true;
+      dispatch(
+        setFailure({
+          name: t('operationalStudies/manageTrainSchedule:errorMessages.trainScheduleTitle'),
+          message: t(
+            'operationalStudies/manageTrainSchedule:errorMessages.originTimeOutsideWindow',
+            {
+              low: dateTimeFormatting(searchDatetimeWindow.begin),
+              high: dateTimeFormatting(searchDatetimeWindow.end),
+            }
+          ),
+        })
+      );
+    }
   }
 
   if (error) return null;
@@ -167,6 +189,8 @@ export const checkStdcmConf = (
     gridMarginBefore,
     gridMarginAfter,
     lastestStartTime: originUpperBoundTime!,
+    workScheduleGroupId,
+    electricalProfileSetId,
   };
 };
 
@@ -194,6 +218,8 @@ export const formatStdcmPayload = (
       steps: validConfig.path,
       time_gap_after: toMsOrUndefined(validConfig.gridMarginBefore),
       time_gap_before: toMsOrUndefined(validConfig.gridMarginAfter),
+      work_schedule_group_id: validConfig.workScheduleGroupId,
+      electrical_profile_set_id: validConfig.electricalProfileSetId,
     },
   };
 };
--- a/front/src/applications/stdcmV2/components/StdcmEmptyConfigError.tsx
+++ b/front/src/applications/stdcmV2/components/StdcmEmptyConfigError.tsx
@@ -1,0 +1,21 @@
+import React from 'react';
+
+import { Tools } from '@osrd-project/ui-icons';
+import { useTranslation } from 'react-i18next';
+
+import 'applications/stdcmV2/styles.scss';
+
+const StdcmEmptyConfigError = () => {
+  const { t } = useTranslation('stdcm');
+  return (
+    <div className="stdcm-config-error">
+      <span className="mb-3 icon">
+        <Tools size="lg" />
+      </span>
+      <h2 className="my-3 mx-0">{t('noConfigurationFound.title')}</h2>
+      <p>{t('noConfigurationFound.text')}</p>
+    </div>
+  );
+};
+
+export default StdcmEmptyConfigError;
--- a/front/src/applications/stdcmV2/components/StdcmOrigin.tsx
+++ b/front/src/applications/stdcmV2/components/StdcmOrigin.tsx
@@ -1,5 +1,6 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useMemo } from 'react';
 
+import { isNil } from 'lodash';
 import { useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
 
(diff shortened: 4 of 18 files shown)


editoast/src/views/v2/timetable/stdcm.rs
```diff
@@ -468,9 +473,16 @@ async fn build_work_schedules(
     time: DateTime<Utc>,
     maximum_departure_delay: u64,
     maximum_run_time: u64,
+    work_schedule_group_id: Option<i64>,
```
flomonster: If the work_schedule_group provided in the request is `None` we should not retrieve the work schedule at all.

```suggestion
    work_schedule_group_id: i64,
```

flomonster (changes requested):
Thanks for this PR.

editoast/src/views/v2/timetable/stdcm.rs
sim51: I did that to keep the previous behavior (ie to take all work_schedule if nothing is provided).
I just pushed some changes to do what you want, and I had to change `[tests/tests/test_stdcm.py](https://github.com/OpenRailAssociation/osrd/pull/8336/files#diff-ab7effc3c3bbcd58d98213d18d546f1d6c8044a23a25c9b4e6b3d240d6187645)
@flomonster Can you check if it's ok for you ?

front/src/reducers/osrdconf/osrdConfCommon/index.ts
```diff
@@ -176,11 +178,17 @@ export function buildCommonConfReducers<S extends OsrdConfState>(): CommonConfRe
     updateTimetableID(state: Draft<S>, action: PayloadAction<S['timetableID']>) {
       state.timetableID = action.payload;
     },
-    updateElectricalProfileSetId(
+    updateElectricalProfileSetID(
```
clarani: Can you keep updateElectricalProfileSetId ? We would like to remove the `ID` in maj of the code base

front/src/applications/stdcmV2/components/StdcmEmptyConfigError.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { type FC } from 'react';
+
+import { Tools } from '@osrd-project/ui-icons';
+import { useTranslation } from 'react-i18next';
+
+const StdcmEmptyConfigError: FC<unknown> = () => {
+  const { t } = useTranslation('stdcm');
+  return (
+    <div className="d-flex flex-column align-items-center" style={{ marginTop: '5em' }}>
```
clarani: In the application, we don't use inline style. Create a style file, in `src/styles/scss/applications/stdcmV2`, to move your style here

front/src/applications/stdcmV2/components/StdcmEmptyConfigError.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { type FC } from 'react';
+
+import { Tools } from '@osrd-project/ui-icons';
+import { useTranslation } from 'react-i18next';
+
+const StdcmEmptyConfigError: FC<unknown> = () => {
```
clarani: Outside of the editor, we don't use FC. Use a simple const to be consistent with the rest of the code base
```suggestion
const StdcmEmptyConfigError = () => {
```

front/src/applications/stdcmV2/hooks/useStdmcEnv.tsx
```diff
@@ -0,0 +1,53 @@
...
+      dispatch(updateInfraID(data.infra_id));
+      dispatch(updateTimetableID(data.timetable_id));
+      dispatch(updateElectricalProfileSetID(data.electrical_profile_set_id));
+      dispatch(updateWorkScheduleGroupID(data.work_schedule_group_id));
+      dispatch(
+        updateSearchDatetimeWindow([
+          new Date(data.search_window_begin),
+          new Date(data.search_window_end),
+        ])
+      );
```
clarani: Can you create a unique action `loadStdcmEnvironment` (you can change the name) to update all the store at the same time ?

front/src/applications/stdcmV2/hooks/useStdmcEnv.tsx
clarani, line 1: Rename the file useStdcmEnv (typo)

front/src/reducers/osrdconf/types.ts
```diff
@@ -26,7 +26,9 @@ export interface OsrdConfState extends InfraState {
   scenarioID?: number;
   pathfindingID?: number;
   timetableID?: number;
-  electricalProfileSetId?: number;
+  electricalProfileSetID?: number;
+  workScheduleGroupID?: number;
+  searchDatetimeWindow?: [Date, Date];
```
clarani: nit: I prefer to have an object with begin and end (or low and high), I think it's more explicit when we use it afterwards
```suggestion
  searchDatetimeWindow?: { begin: Date, end: Date };
```

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
```diff
@@ -120,11 +121,16 @@ const StdcmViewV2 = () => {
     }
   }, [simulationsList]);
 
+  // If we've got an error during the loading of the stdmc env which is not the "no config error" message,
```
clarani:
```suggestion
  // If we've got an error during the loading of the stdcm env which is not the "no config error" message,
```

front/src/applications/stdcmV2/views/StdcmViewV2.tsx
```diff
@@ -120,11 +121,16 @@ const StdcmViewV2 = () => {
...
+  // we let the error boundary manage it
+  if (error && error.message !== NO_CONFIG_FOUND_MSG) throw error;
+
   return (
     <div role="button" tabIndex={0} className="stdcm-v2" onClick={() => setShowStatusBanner(false)}>
       <StdcmHeader />
-      <div>
-        {scenarioID && (
+
+      {isNil(error) ? (
```
clarani: nit: I think it would be easier to read if you write it like this: 
```
if error
  then display error
else
  display the body
```
because the errorComponent is way smaller than the main component.

clarani (changes requested):

front/src/reducers/osrdconf/osrdConfCommon/index.ts
sim51: Ho, by looking at the code base, I was thinking the opposite, that's why I changed it. 
I will reverse it, thanks.

front/src/applications/stdcm/utils/formatStdcmConfV2.ts
```diff
@@ -118,6 +123,22 @@ export const checkStdcmConf = (
       })
     );
   }
+  const startDatetime = new Date(`${startTime}`);
+  if (
+    searchDatetimeWindow &&
```
clarani: Use a else if (), like this you know that startTime is not null
```suggestion
  } else if (searchDatetimeWindow) {
    const startDatetime = new Date(startTime);
  if (
```

front/src/applications/stdcmV2/components/StdcmOrigin.tsx
```diff
@@ -46,6 +50,19 @@ const StdcmOrigin = ({ setCurrentSimulationInputs, disabled = false }: StdcmConf
     dispatch(updateOriginTime(e.target.value));
   };
 
+  const error = useMemo(() => {
+    if (!isNil(originDate) && !isNil(searchDatetimeWindow)) {
```
clarani: Do you have a specific reason to use isNil here ? Or can we write it like this ?
```suggestion
    if (originDate && searchDatetimeWindow) {
```

front/src/applications/stdcmV2/components/StdcmOrigin.tsx
```diff
@@ -77,10 +95,16 @@ const StdcmOrigin = ({ setCurrentSimulationInputs, disabled = false }: StdcmConf
               label={t('trainPath.time')}
               id="originTime"
               onChange={onOriginTimeInputChange}
+              isInvalid={!isNil(error)}
               value={originTime}
               disabled={disabled}
             />
           </div>
+          {!isNil(error) && (
```
clarani: I think `isNil` is not compulsory here
```suggestion
          {error && (
```

front/src/applications/stdcmV2/hooks/useStdmcEnv.tsx
```diff
@@ -0,0 +1,53 @@
...
+    updateWorkScheduleGroupID,
+    updateSearchDatetimeWindow,
+  } = useOsrdConfActions();
+  const [getStdcmSearchEnvironment] =
+    osrdEditoastApi.endpoints.getStdcmSearchEnvironment.useLazyQuery();
+
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState<null | Error>(null);
+
+  const loadStdcmEnvironment = useCallback(async () => {
```
clarani: It's only a question but do we need to create this async function or do you think we could use a simple useQuery (with the state loading and error of the request) ?

I'm fine with this solution too

front/src/applications/stdcmV2/hooks/useStdmcEnv.tsx
```diff
@@ -0,0 +1,53 @@
...
+      dispatch(updateElectricalProfileSetID(data.electrical_profile_set_id));
+      dispatch(updateWorkScheduleGroupID(data.work_schedule_group_id));
+      dispatch(
+        updateSearchDatetimeWindow([
+          new Date(data.search_window_begin),
+          new Date(data.search_window_end),
+        ])
+      );
+    } catch (e) {
+      setError(e as Error);
```
clarani: Should we reset the store ?

front/src/reducers/osrdconf/osrdConfCommon/selectors.ts
```diff
@@ -39,7 +39,9 @@ const buildCommonConfSelectors = <ConfState extends OsrdConfState>(
     getScenarioID: makeOsrdConfSelector('scenarioID'),
     getPathfindingID: makeOsrdConfSelector('pathfindingID'),
     getTimetableID: makeOsrdConfSelector('timetableID'),
-    getElectricalProfileSetId: makeOsrdConfSelector('electricalProfileSetId'),
+    getElectricalProfileSetId: makeOsrdConfSelector('electricalProfileSetID'),
+    getWorkScheduleGroupID: makeOsrdConfSelector('workScheduleGroupID'),
```
clarani: This selector is not used
```suggestion
```

front/src/applications/stdcmV2/hooks/useStdmcEnv.tsx
sim51: We can use the `useQuery` and check the error, the loading and do the dispatch.
But in this case all the code of this hook should be included in the `StdcmViewv2`  component.

I did it to centralized the logic in one file.

front/src/applications/stdcmV2/hooks/useStdmcEnv.tsx
sim51: If there is an error it comes from the API call, not from the dispatch, so the store should not be touched.
And with which values you want to reset the store ?

front/src/applications/stdcmV2/components/StdcmEmptyConfigError.tsx
clarani, line 1: You still have some inline styles in this component

front/src/reducers/osrdconf/osrdConfCommon/__tests__/utils.ts
```diff
@@ -155,15 +155,15 @@ const testCommonConfReducers = (slice: OperationalStudiesConfSlice | StdcmConfSl
   });
 
   it('should handle updateElectricalProfileSetId with number', () => {
-    const newElectricalProfileSetId = 1;
-    defaultStore.dispatch(slice.actions.updateElectricalProfileSetId(newElectricalProfileSetId));
+    const newElectricalProfileSetID = 1;
```
clarani: You still have some unwanted changes in this file
```suggestion
    const newElectricalProfileSetId = 1;
```

front/src/reducers/osrdconf/osrdConfCommon/index.ts
```diff
@@ -78,6 +78,8 @@ interface CommonConfReducers<S extends OsrdConfState> extends InfraStateReducers
   ['updatePathfindingID']: CaseReducer<S, PayloadAction<S['pathfindingID']>>;
   ['updateTimetableID']: CaseReducer<S, PayloadAction<S['timetableID']>>;
   ['updateElectricalProfileSetId']: CaseReducer<S, PayloadAction<S['electricalProfileSetId']>>;
+  ['updateWorkScheduleGroupId']: CaseReducer<S, PayloadAction<S['workScheduleGroupId']>>;
+  ['updateSearchDatetimeWindow']: CaseReducer<S, PayloadAction<S['searchDatetimeWindow']>>;
```
clarani: You can remove these 2 actions

front/src/applications/stdcmV2/hooks/useStdmcEnv.tsx
clarani: Undefined I believe. We can keep it like this :+1:

front/src/applications/stdcmV2/hooks/useStdmcEnv.tsx
clarani: You can define the function inside the useEffect directly

front/src/applications/stdcmV2/components/StdcmEmptyConfigError.tsx
sim51: Sorry, I must have been tired yesterday ...

clarani: No problem :)

clarani (approved):

bloussou (approved):
test part is ✅

sim51:
I squashed my commits

flomonster (approved):
LGTM!

PR #8345, opened 2024-08-06
editoast: add roles check to `views::infra`

hamz2a:
- [x] `/infra` GET
- [x] `/infra` POST
- [x] `/infra/refresh`
- [x] `/infra/voltages`
- [x] `/infra/{infra_id}` GET
- [x] `/infra/{infra_id}` DELETE
- [x] `/infra/{infra_id}` PUT
- [x] `/infra/load` 
- [x] `/infra/clone` 
- [x] `/infra/lock` 
- [x] `/infra/unlock` 
- [x] `/infra/speed_limit_tags` 
- [x] `/infra/voltages` 
- [x] `/infra/switch_types`

### DIFF
--- a/editoast/editoast_authz/src/builtin_role.rs
+++ b/editoast/editoast_authz/src/builtin_role.rs
@@ -18,7 +18,7 @@ impl BuiltinRoleSet for BuiltinRole {
     fn implies_iter(&self) -> impl IntoIterator<Item = Self> {
         use BuiltinRole::*;
         match self {
-            OpsWrite => vec![],
+            OpsWrite => vec![InfraRead],
             InfraRead => vec![],
             InfraWrite => vec![InfraRead],
         }
--- a/editoast/src/views/infra/mod.rs
+++ b/editoast/src/views/infra/mod.rs
@@ -14,6 +14,8 @@ use axum::extract::Query;
 use axum::extract::State;
 use axum::http::StatusCode;
 use axum::response::IntoResponse;
+use axum::Extension;
+use editoast_authz::BuiltinRole;
 use editoast_derive::EditoastError;
 use serde::Deserialize;
 use serde::Serialize;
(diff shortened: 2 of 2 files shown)


editoast/editoast_authz/src/builtin_role.rs
```diff
@@ -5,14 +5,22 @@ use crate::roles::BuiltinRoleSet;
...
+    #[strum(serialize = "infra:write")]
+    InfraWrite,
 }
 
 impl BuiltinRoleSet for BuiltinRole {
     fn implies_iter(&self) -> impl IntoIterator<Item = Self> {
+        use BuiltinRole::*;
         match self {
-            Self::OpsWrite => vec![],
+            OpsWrite => vec![],
```
flomonster: I think we must be able to read infra to write operational studies. 
I'm not 100% sure, what do you think?

```suggestion
            OpsWrite => vec![InfraRead],
```

editoast/src/views/infra/mod.rs
```diff
@@ -119,8 +123,17 @@ struct RefreshResponse {
 )]
 async fn refresh(
     app_state: State<AppState>,
+    Extension(authorizer): AuthorizerExt,
     Query(query_params): Query<RefreshQueryParams>,
 ) -> Result<Json<RefreshResponse>> {
+    let authorized = authorizer
+        .check_roles([BuiltinRole::InfraWrite].into())
```
flomonster, line 129: The frontend never uses this. This endpoint is used to force refresh. It's exclusively used by developers (ops)

flomonster (approved):
LGTM

editoast/src/views/infra/mod.rs
hamz2a: ok, Do you want me to delete the check role?

flomonster: Ideally, this endpoint should be available only for superusers.

hamz2a: As discussed, we'll keep the check for now.

editoast/editoast_authz/src/builtin_role.rs
hamz2a: It sounds logical.

Issue #8215, opened 2024-07-26
Pathfinding geometry does not match the infra geometry


flomonster:
### What happened?

The pathfinding geometry does not match the infrastructure geometry.

![image](https://github.com/user-attachments/assets/627d35ad-8708-449b-bb47-51a2c5b40fa8)

> [!NOTE]
> This is only the case with tracks that have a very large geometry.

### How can we reproduce it (as minimally and precisely as possible)?

1. Use this [infrastructure.json](https://github.com/user-attachments/files/16390022/Debug.pos.path.id452.railjson.json)
2. Create a scenario and do pathfinding

### On which environments the bug occurs?

Local

### On which browser the bug occurs?

Firefox

### OSRD version (top right corner `Account` button > `Informations`)

dba683be39981510e0024ce973280049e030f25c

sim51:
I did some investigation, and for me it comes from the `LineString` class in `osrd-geom`.
I wrote a simple test :

```
public void slice_should_work() {
        LineString line = LineString.make(
            new Point(2.412869556,48.865639558),
            new Point(4.435524213221072,49.526087172096595)
        );
        LineString sliced_line = line.slice(0.1, 0.9);
        assertEquals(Math.round(0.8*line.getLength() *100000), Math.round(sliced_line.getLength()*100000));
 }
 ```
 
 Distance are OK, but if you build the geojson from the `line` and `sliced_line` and display them, you have the same issue than yours : 
 
![image](https://github.com/user-attachments/assets/083922dc-d310-4b25-90f1-9f193d2de3bc)

sim51:
FYI, turf is not better : https://runkit.com/embed/yl291r6mwt5i

PR #8346, opened 2024-08-06
 refacto: core: rename occurences of "x/y" to "latitude/longitude"

### LINKED ISSUE #8344: refacto: core: rename occurences of "x/y" to "latitude/longitude"
The classes we have in `core` to handle geographic data are named "x/y", but used as latitude / longitude. Figuring out which is which is the source of many headaches and errors. We should rename them.

This also highlights the few places where we still naively handle these values as cartesian points

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/8344

3 parts and 3 commits:

1. Remove what should be the last occurrence of cartesian maths
    1. Doesn't fix the one in https://github.com/OpenRailAssociation/osrd/issues/8215 as it will be in another PR, but it's commented
    1. Includes test updates, as this slightly changes the results
3. Rename "x" and "y" to "longitude" and "latitude"
4. Swap longitudes and latitudes in parameter orders (lat/lon is more common than lon/lat)


Erashin (approved):

shenriotpro (approved):
Thanks, LGTM for tests/

PR #8351, opened 2024-08-07
core, editoast: return speed-limit source (tag) of MRSP

### LINKED ISSUE #7717: Return speed-limit tag applied in simulation (back)
Last part (back) of https://github.com/osrd-project/osrd-confidential/issues/364

Output the speed-limit tag applied by simulation/STDCM (and error/warning if the case arises) in the speed-space graph backend output.

Back : return an enum with 4 variants:
* given speed-limit-tag from request(string)
* fallback speed-limit-tag(string)
* ~default-speed(float)~ (dismissed at least for now)
* unknown-tag # tag not found, missing from both infra and config
* none # tag not relevant (no train-tag provided or route speed-limit is used or using RS max speed, etc.)

Nice to have: if no trainTag was provided, return an empty list (or even no list at all)

bougue-pe:
* transport `SpeedLimitSource` via `EnvelopePart.attr`
  * add and use `utils.SelfTypeHolder` in place of `envelope.EnvelopeAttr`
* add backend integration tests
* use `BlockPathRangeValues` for MRSP (core, editoast, front)
  * group speed-limit and source
  * "fix" type of offset in simulation response, as it is a `BlockPath` (not a `TravelledPath`)

🔍 review by commit should be easier

closes: https://github.com/OpenRailAssociation/osrd/issues/7717


core/kt-osrd-utils/src/main/kotlin/fr/sncf/osrd/utils/json/UnitJsonAdapters.kt
```diff
@@ -49,6 +47,26 @@ class OffsetAdapter<T> : JsonAdapter<Offset<T>?>() {
...
+        if (reader.peek() == JsonReader.Token.NULL) {
+            reader.skipValue()
+            return null
+        }
+        return reader.nextDouble().metersPerSecond
+    }
+
+    @ToJson
+    override fun toJson(writer: JsonWriter, value: Speed?) {
+        writer.value(if (value != null) Speed.toMetersPerSecond(value) else null)
```
bougue-pe: This choice is debatable: serialize in meters per second (float), while internal value is in millimeters per second (unsigned int).
This eases the use in front (unchanged).

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/standalone_sim/SimulationResponse.kt
```diff
@@ -19,8 +20,9 @@ class SimulationSuccess(
     val base: ReportTrain,
     val provisional: ReportTrain,
     @Json(name = "final_output") val finalOutput: CompleteReportTrain,
-    val mrsp: MRSPResponse,
-    @Json(name = "electrical_profiles") val electricalProfiles: RangeValues<ElectricalProfileValue>,
+    val mrsp: BlockPathRangeValues<SpeedLimitProperty>,
+    @Json(name = "electrical_profiles")
+    val electricalProfiles: BlockPathRangeValues<ElectricalProfileValue>,
```
bougue-pe: Using "BlockPath" ranges here as it relates to the (Block)`Path`, not the `TravelledPath` if I'm correct.

front/src/modules/trainschedule/components/DriverTrainScheduleV2/DriverTrainScheduleStopListV2.tsx
```diff
@@ -35,9 +35,9 @@ const DriverTrainScheduleStopListV2 = ({
 
   const formattedMrsp = useMemo(
     () =>
-      mrsp.positions.map((_position, index) => ({
+      mrsp.boundaries.map((_position, index) => ({
```
clarani: I think it would be better to map on the values rather than on the boundaries, otherwise you might miss the last value.
Moreover, I think you need to -1 to get the correct boundary
```suggestion
      mrsp.values.map(({ speed }, index) => ({
        position: index === 0 ? 0 :  mrsp.boundaries[index-1],
        speed
      })
```

front/src/modules/simulationResult/components/SpeedSpaceChart/__tests__/sampleData.ts
```diff
@@ -57,8 +57,25 @@ export const simulation: SimulationResponseSuccess = {
...
+      },
+      {
+        speed: 10,
+        source: null,
+      },
+      {
+        speed: 20,
+        source: { speed_limit_source_type: 'unknown_tag' },
+      },
+    ],
```
clarani: Here we have 5 boundaries and 4 values, but I thought we should have `n` boudaries and `n+1` values :thinking:

clarani (changes requested):

front/src/modules/trainschedule/components/DriverTrainScheduleV2/DriverTrainScheduleStopListV2.tsx
bougue-pe: Logic to be explored, but at least the last boundary doesn't have a corresponding value
```suggestion
      mrsp.boundaries.slice(0,-1).map((_position, index) => ({
```
Same on `exportDriverScheduleCSV.ts` below

Thanks @RomainValls
EDIT: ✅ Handled in a more complex way, as required.

core/kt-osrd-utils/src/main/kotlin/fr/sncf/osrd/utils/json/UnitJsonAdapters.kt
eckter:
```suggestion
        writer.value(value?.metersPerSecond)
```

Not tested, but that's how it's done with other units

eckter: Looks fine for me, these classes are used to decorrelate the internal representations from the ones in the api

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/standalone_sim/SimulationResponse.kt
eckter: That feels wrong, the "block path" (the one where offsets always start at the border of a block) should only be an internal core representation. We've had confusions with `Path` / `TravelledPath` in the schemas (as the name `Offset<Path>` feels closer to what we want to describe), but that confusion should only be a naming thing. I don't think any actual `Offset<Path>` should be returned.

core/src/main/kotlin/fr/sncf/osrd/utils/JavaInteroperabilityTools.kt
```diff
@@ -2,16 +2,18 @@ package fr.sncf.osrd.utils
...
 import fr.sncf.osrd.utils.indexing.mutableStaticIdxArrayListOf
-import fr.sncf.osrd.utils.units.Speed
 
 /**
  * Java can't handle generics with value classes, it's impossible to extract the underlying speed
  * from a `RangeMapEntry` in pure java. This method handles this specific case.
  */
-fun rangeMapEntryToSpeed(entry: DistanceRangeMap.RangeMapEntry<Speed>): Long {
-    return entry.value.millimetersPerSecond.toLong()
+fun rangeMapEntryToSpeedLimitProperty(
```
eckter, line 13: If handling java classes is too much of a pain, they can be automatically converted to kotlin at any point. We just haven't bothered converting classes that don't change much anymore.

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/standalone_sim/SimulationResponse.kt
bougue-pe: As discussed, I was mistaken on that, I tried to document/rename a bit around that: https://github.com/OpenRailAssociation/osrd/pull/8351/commits/b4c6f6f013b673a29e906c167d7bc2d24f5f92a0#diff-4c2dee00411e3f70730c838ef56323cdd8c1272c53725ec2db95cabc701e134eR23
This is part of a greater effort to be made step-by-step 🙂

front/src/modules/trainschedule/components/DriverTrainScheduleV2/DriverTrainScheduleStopListV2.tsx
bougue-pe: The handling of the response was indeed wrong.
Yet, the path boundaries are not useful, so I chose to work with the actual MRSP in https://github.com/OpenRailAssociation/osrd/pull/8351/commits/b4c6f6f013b673a29e906c167d7bc2d24f5f92a0#diff-e7728b2c01bcf96a9b19c23cff25682e233aea4a7dcee40f4615cb88e38d28e9R38-R39 (same for CSV to mutualize)

front/src/modules/simulationResult/components/SpeedSpaceChart/__tests__/sampleData.ts
bougue-pe: As discussed, my bad on that format, changed in https://github.com/OpenRailAssociation/osrd/pull/8351/commits/b4c6f6f013b673a29e906c167d7bc2d24f5f92a0

core/kt-osrd-utils/src/main/kotlin/fr/sncf/osrd/utils/json/UnitJsonAdapters.kt
bougue-pe: Done in https://github.com/OpenRailAssociation/osrd/pull/8351/commits/b4c6f6f013b673a29e906c167d7bc2d24f5f92a0

core/src/main/kotlin/fr/sncf/osrd/utils/JavaInteroperabilityTools.kt
bougue-pe: It would be nicer, but not mandatory here, so I let it like that.

front/src/reducers/osrdsimulation/types.ts
```diff
@@ -72,6 +72,12 @@ export type PositionSpeedTime<Time = number> = Position<Time> & {
   speed: number;
 };
 export type ConsolidatedPositionSpeedTime = PositionSpeedTime<Date>;
+export interface RangeSpeeds {
```
clarani: Use a type instead of an interface

front/src/reducers/osrdsimulation/types.ts
```diff
@@ -72,6 +72,12 @@ export type PositionSpeedTime<Time = number> = Position<Time> & {
...
+  // List of `n` internal boundaries of the ranges along the path (excluding start and end bounds).
+  internal_boundaries: number[];
```
clarani: use camelCase ;)

front/src/reducers/osrdsimulation/types.ts
clarani: SpeedRanges should be a better name I think :thinking:

front/src/modules/trainschedule/components/DriverTrainScheduleV2/exportDriverScheduleCSV.ts
```diff
@@ -81,12 +71,14 @@ const overloadSteps = (
...
-    position: mmToM(speedLimit.position),
-    speed: interpolateValue(trainRegime, speedLimit.position, 'speeds'),
-    speedLimit: speedLimit.speed,
-    time: interpolateValue(trainRegime, speedLimit.position, 'times'),
-  }));
+  const speedsAtSpeedLimitChange = speedLimits.speeds.map((_speed, index) => {
+    const boundary = index > 0 ? speedLimits.internal_boundaries[index - 1] : 0;
+    return {
+      position: boundary,
+      speed: interpolateValue(trainRegime, mToMm(boundary), 'speeds'),
```
clarani, line 78: I'm a bit confused, but technically, boundary is already in mm (according to the description in `generatedEditoastApi`). So either the description is wrong, or the calculation is a bit weird here

front/src/modules/trainschedule/components/DriverTrainScheduleV2/exportDriverScheduleCSV.ts
```diff
@@ -81,12 +71,14 @@ const overloadSteps = (
...
-  const speedsAtSpeedLimitChange = speedLimits.map((speedLimit) => ({
-    position: mmToM(speedLimit.position),
-    speed: interpolateValue(trainRegime, speedLimit.position, 'speeds'),
-    speedLimit: speedLimit.speed,
-    time: interpolateValue(trainRegime, speedLimit.position, 'times'),
-  }));
+  const speedsAtSpeedLimitChange = speedLimits.speeds.map((_speed, index) => {
+    const boundary = index > 0 ? speedLimits.internal_boundaries[index - 1] : 0;
+    return {
+      position: boundary,
```
clarani, line 77: In the previous version, the position was in meter, but here I think you are giving it in mm

front/src/modules/trainschedule/components/DriverTrainScheduleV2/utils.ts
```diff
@@ -87,6 +78,26 @@ export function getActualPositionLeft(givenPosition: number, speed: PositionSpee
...
+  // If exactly on a speed-limit change, use the minimal value of both side
+  if (
+    vmaxUpperBoundIndex > 0 &&
+    vmax.internal_boundaries[vmaxUpperBoundIndex - 1] === givenPosition
+  )
+    return Math.min(vmax.speeds[vmaxUpperBoundIndex], vmax.speeds[vmaxUpperBoundIndex - 1]);
+  return vmax.speeds[vmaxUpperBoundIndex];
+}
+
+export function getActualVmax(givenPosition: number, vmax: RangeSpeeds) {
```
clarani: I think you can merge this function with getActualVmaxMetersPerSecond

front/src/modules/trainschedule/components/DriverTrainScheduleV2/utils.ts
```diff
@@ -87,6 +78,26 @@ export function getActualPositionLeft(givenPosition: number, speed: PositionSpee
...
+  return Math.round(getActualVmaxMetersPerSecond(givenPosition, vmax) * 3.6);
```
clarani: Don't use magic numbers, you can use instead a conversion function, for instance msToKmh
```suggestion
  return Math.round(msToKmh(getActualVmaxMetersPerSecond(givenPosition, vmax)));
```

front/src/modules/trainschedule/components/DriverTrainScheduleV2/DriverTrainScheduleStopListV2.tsx
```diff
@@ -33,12 +33,13 @@ const DriverTrainScheduleStopListV2 = ({
...
-  const formattedMrsp = useMemo(
-    () =>
-      mrsp.positions.map((_position, index) => ({
-        position: _position,
-        speed: mrsp.speeds[index],
-      })),
+  const formattedMrsp: RangeSpeeds = useMemo(
+    () => ({
+      internal_boundaries: mrsp.values.map((_, index) =>
+        index === 0 ? 0 : mrsp.boundaries[index - 1]
```
clarani: Why do you add this 0 at the beginning of the array ?

front/src/reducers/osrdsimulation/types.ts
```diff
@@ -72,6 +72,12 @@ export type PositionSpeedTime<Time = number> = Position<Time> & {
   speed: number;
 };
 export type ConsolidatedPositionSpeedTime = PositionSpeedTime<Date>;
+export interface RangeSpeeds {
+  // List of `n` internal boundaries of the ranges along the path (excluding start and end bounds).
```
clarani: Can you specify whether the boundaries are in meter or in mm ?

front/src/modules/trainschedule/components/DriverTrainScheduleV2/utils.ts
clarani: In the comment of the function, specify in which unit givenPosition must be (m or mm)

Tguisnet (approved):
LGTM for the editoast part, but I can't test now, so you should check that an other reviewer could test the PR.

front/src/modules/trainschedule/components/DriverTrainScheduleV2/exportDriverScheduleCSV.ts
Tguisnet: According to @Erashin, the description seems to be right.

front/src/modules/trainschedule/components/DriverTrainScheduleV2/DriverTrainScheduleStopListV2.tsx
RomainValls: Should we just remove this check as the origin is induced ?

Erashin (approved):
Core LGTM.

front/src/modules/trainschedule/components/DriverTrainScheduleV2/exportDriverScheduleCSV.ts
clarani: Ok,discussed with @RomainValls, the internalBoundary is in m (while the boundary from the backend in mm)

clarani (commented):
LGTM for the code, you can rebase while I try to test :)

clarani (approved):
LGTM :white_check_mark:

tests/tests/test_timetable_v2.py
```diff
@@ -139,3 +139,70 @@ def test_scheduled_points_with_incompatible_margins(
...
+        "boundaries": [3680000, 4580000],
+        "values": [
+            {"speed": 39.444, "source": {"speed_limit_source_type": "unknown_tag"}},
+            {"speed": 31.111, "source": {"speed_limit_source_type": "unknown_tag"}},
+            {"speed": 80, "source": None},
+        ],
+    }
+
+
+def get_train_schedule_simulation_response(infra, timetable_v2_id, train_schedule_payload):
```
bloussou:
```suggestion
def _get_train_schedule_simulation_response(infra: Infra, timetable_v2_id: int, train_schedule_payload: Dict[str, Any]):
```

tests/tests/test_stdcm.py
```diff
@@ -155,3 +155,57 @@ def test_work_schedules(small_scenario_v2: Scenario, fast_rolling_stock: int):
...
+        "boundaries": [3680000, 4580000],
+        "values": [
+            {"speed": 39.444, "source": {"speed_limit_source_type": "unknown_tag"}},
+            {"speed": 31.111, "source": {"speed_limit_source_type": "unknown_tag"}},
+            {"speed": 80, "source": None},
+        ],
+    }
+
+
+def get_stdcm_response(infra, stdcm_payload, timetable_v2_id):
```
bloussou:
```suggestion
def _get_stdcm_response(infra: Infra, stdcm_payload: Dict[str, Any], timetable_v2_id: int):
```

bloussou (approved):
Thanks :)

eckter (approved):

tests/tests/test_stdcm.py
bougue-pe: Done in https://github.com/OpenRailAssociation/osrd/pull/8351/commits/228d679c54799f5835bfe6d8f0a6e090c1c9b2c4 (change the order of params to match TS helper)

tests/tests/test_timetable_v2.py
bougue-pe: Done in https://github.com/OpenRailAssociation/osrd/pull/8351/commits/228d679c54799f5835bfe6d8f0a6e090c1c9b2c4 (this is a batch endpoint, so it's actually a list of TS)

front/src/modules/trainschedule/components/DriverTrainScheduleV2/__tests__/utils.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  it('should return the min Vmax when givenPosition is equal to a boundary (min before)', () => {
+    const result = findActualVmax(3400, vMax);
+    expect(result).toEqual(100);
+  });
+
+  it('should return the min Vmax when givenPosition is equal to a boundary (min after)', () => {
+    const result = findActualVmax(5300, vMax);
+    expect(result).toEqual(100);
```
bougue-pe, line 28: @RomainValls Thanks for those tests!
Added a test on both types of boudaries (min before or after), and also put first speed at `10` to avoid cases where `0` value would be a hidden bug (as it may be the default value).
Done in https://github.com/OpenRailAssociation/osrd/pull/8351/commits/228d679c54799f5835bfe6d8f0a6e090c1c9b2c4

RomainValls: LGTM

editoast/Cargo.toml
```diff
@@ -16,7 +16,6 @@ members = [
 ]
 
 [workspace.lints.rust]
-dependency_on_unit_never_type_fallback = "allow" # this is caused by tracing, we can't do anything about it
```
leovalais, line 19: This warning was disabled because it occurs on nightly and doesn't seem to be fixable on our end. Did it disappear on recent builds?

bougue-pe: ~Yes, I got a warning on this lint not being recognized for rust 1.80.1~
```
warning[E0602]: unknown lint: `dependency_on_unit_never_type_fallback`
```

Forget it, didn't try on nightly

leovalais: Yep, it generates a warning on `stable`. Resolving since it's our target, not nightly.

PR #8354, opened 2024-08-07
Draft test : Aba/e2e map firefox

Maymanaf:
**DO NOT REVIEW**   : This is a draft PR for testing


.github/workflows/build.yml
```diff
@@ -817,6 +816,43 @@ jobs:
           docker load --input ./osrd-core.tar
           docker load --input ./osrd-gateway-standalone.tar
           docker load --input ./osrd-front-nginx.tar
+          - name: Install Firefox and dependencies
+          run: |
+            sudo apt-get update
+            sudo apt-get install -y firefox
+            sudo apt-get install -y \
```
emersion: Do we really need to explicitly list all of Firefox' dependencies here? These should be pulled in automatically with Firefox no? (The list of Firefox dependencies may change over time so it's a bit annoying to maintain. Also, as the Wayland maintainer I can't accept explicitly installing X11 libraries. 😂)

Maymanaf: No it's just a draft for testing on the CI

PR #8355, opened 2024-08-07
front: handle the two date and time formats

### LINKED ISSUE #8296: When creating a train the arrival and departure dates/times in the train list are not displayed
### What happened?

When creating a train the arrival and departure dates/times in the train list are not displayed

When creating a train the arrival and departure dates/times in the train list are displayed
image

Create a train.
Check the train list.
image

Recette (SNCF)

Google Chrome

0a34296

achrafmohye:
Closes: #8296

### DIFF
--- a/front/src/utils/__tests__/date.spec.ts
+++ b/front/src/utils/__tests__/date.spec.ts
@@ -55,6 +55,12 @@ describe('parseDateTime', () => {
     expect(result).toBeInstanceOf(Date);
   });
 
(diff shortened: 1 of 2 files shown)


theocrsb (approved):
LGTM and tested !

clarani (commented):
LGTM and tested :white_check_mark: Thanks for the fix :)

Can you add a test where the datetime string has only 1 number for the day ?

clarani (approved):
LGTM :white_check_mark:

PR #8356, opened 2024-08-07
Display pathfinding constraints on shortest path

sim51:
Implements #7719

This PR is just the fusion (+ rebase) of :

* #7953 the one from PE for the code in core 
* [8235](https://github.com/OpenRailAssociation/osrd/pull/8235) : for the code of editoast & front


flomonster (approved):
LGTM (tested + already reviewed in another PR).

A brief comment on the behavior. When the user centers on a constraint element, the camera flies to the location. But there is a user parameter called "Smooth travel" that configures this behavior for search. I think we should use this parameter in our case too.

image

It's not blocking to merge this PR.

SharglutDev (approved):
Lgtm (for front part) and tested !

eckter (approved):
I only skimmed through the core changes as it's already been reviewed, but I tested and it's really nice :+1:

Erashin (approved):

PR #8362, opened 2024-08-08
editoast: add roles check to rolling stocks

Wadjetz:


hamz2a (approved):
LGTM.

editoast/src/views/light_rolling_stocks.rs
leovalais, line 1: missing `get` endpoint

leovalais (changes requested):

leovalais (approved):

Issue #7712, opened 2024-06-14
front: stdcm: use ui-core components


SarahBellaha:
## Description

Replace the actual components by their ui-core versions

## Acceptance criteria

The following ui-core components have been implemented
- [x] datePicker
- [x] timePicker
- [x] tolerancePicker
- [x] select
- [x] #8706

PR #8364, opened 2024-08-08
front: stdcm scheduled steps 7693

### LINKED ISSUE #7693: front: stdcm: scheduled times and tolerance for stdcm steps
# Description

It must be possible to specify desired times at the various via points during an STDCM simulation.

# AC
- [ ] The departure and arrival points have a selector to indicate the timetable indicating either
    - as soon as possible 
    - precise time
- [ ] If precise time is selected, inputs appear for a date (mandatory), a desired time (mandatory) and a +/- tolerance (optional, in minutes).

Image

- [ ] By default, the departure is timed but the date, time and tolerance fields are empty.
- [ ] Departure and arrival cannot be timed simultaneously: if a precise time is entered for the departure, and the user selects ‘precise time’ for the destination, a warning is displayed in the ‘Get simulation’ button, with two buttons for respecting the departure or arrival time. When either of these two buttons is clicked, the corresponding selector switches back to ‘as soon as possible’, and a message to restart the simulation is displayed. Any inputs entered for the date, time and tol fields are saved in case the user decides to re-select ‘precise time’.

Image

- [ ] If the origin and destination are set to ‘as soon as possible’, a warning is displayed for the ‘Get simulation’ button.

Image

- [ ] For intermediate points, there is no timetable selector because these points cannot be timetabled.
- [ ] Intermediate waypoints must have an input field for the stop time (_optional_).

### LINKED ISSUE #8261: Adding a Via at a specific position in the New STDCM Interface
### Description and goal

Currently, in the new STDCM interface, the user can add a via by clicking on a button displayed just above the destination card. The new via is added immediately before the destination card, making it the last via. However, the user cannot add a via at a specific position.

One way to solve the problem is to add the button which allow to add via between each via.

### Acceptance criteria

- [x] User can insert a via anywhere in the list of vias

SarahBellaha:
Closes #7693 
Closes #8261 
Resolves some part of  #7712


front/src/reducers/osrdconf/osrdConfCommon/index.ts
```diff
@@ -97,6 +102,20 @@ interface CommonConfReducers<S extends OsrdConfState> extends InfraStateReducers
...
+  ['updateOriginArrivalType']: CaseReducer<S, PayloadAction<ArrivalTimeTypes>>;
+  ['updateDestinationArrivalType']: CaseReducer<S, PayloadAction<ArrivalTimeTypes>>;
+  ['updateOriginTolerances']: CaseReducer<
+    S,
+    PayloadAction<{ toleranceBefore: number; toleranceAfter: number }>
+  >;
+  ['updateDestinationTolerances']: CaseReducer<
+    S,
+    PayloadAction<{ toleranceBefore: number; toleranceAfter: number }>
+  >;
```
clarani: These actions are only used in stdcm, so you can move them to `stdcmConf/index.ts` ;)

front/src/reducers/osrdconf/helpers.ts
```diff
@@ -113,3 +114,46 @@ export const insertViaFromMap = (
...
+
+export const updateDestinationPathStep = (
+  state: OsrdConfState,
+  destination: Partial<PathStep> | null,
+  replaceCompletely: boolean = false
+) => {
+  updatePathStepAtIndex(state, state.pathSteps.length - 1, destination, replaceCompletely);
+};
+
+export const getPathStepArrivalData = (arrivalTime: PathStep['arrival']) => ({
```
clarani: I'm not sure we should do that :thinking: 
I think that if the arrivalTime is modified, we should only update it

front/src/reducers/osrdconf/helpers.ts
clarani, line 1: Nice functions :)

front/src/applications/stdcmV2/types.ts
```diff
@@ -30,3 +30,19 @@ export type StdcmConfigCardProps = {
   setCurrentSimulationInputs: React.Dispatch<React.SetStateAction<StdcmSimulationInputs>>;
   disabled?: boolean;
 };
+
+export enum ArrivalTimeTypes {
+  PRECISE_TIME = 'preciseTime',
+  ASAP = 'asSoonAsPossible',
+}
+
+export enum StdcmErrorTypes {
```
clarani: I think `StdcmConfigErrorTypes` would be a better name because later we will also have to manage the response errors
```suggestion
export enum StdcmConfigErrorTypes {
```

front/src/applications/stdcmV2/components/StdcmWarningBox.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const StdcmWarningBox = ({
+  errorInfos: { errorType, errorDetails },
+  resetOriginArrival,
+  resetDestinationArrival,
+}: {
+  errorInfos: { errorType: StdcmErrorTypes; errorDetails?: StdcmConfigErrors['errorDetails'] };
+  resetOriginArrival: () => void;
+  resetDestinationArrival: () => void;
+}) => {
```
clarani: Maybe you can define props above, to improve the code readability

front/src/applications/stdcmV2/components/StdcmVias.tsx
```diff
@@ -48,8 +48,8 @@ const StdcmVias = ({ disabled = false, setCurrentSimulationInputs }: StdcmConfig
     dispatch(deleteViaV2(index));
   };
 
-  const onClickOnCard = () => {
-    const newPathSteps = addElementAtIndex(pathSteps, pathSteps.length - 1, null);
+  const onClickOnCard = (pathStepId: number) => {
```
clarani: Maybe addPathStep or addVia would be a better name ? Or addViaOnClick ?
```suggestion
  const addPathStep = (pathStepId: number) => {
```

clarani: Also I think it's pathStepIndex
```suggestion
  const onClickOnCard = (pathStepIndex: number) => {
```

front/src/applications/stdcmV2/components/StdcmVias.tsx
```diff
@@ -66,43 +66,52 @@ const StdcmVias = ({ disabled = false, setCurrentSimulationInputs }: StdcmConfig
         intermediatePoints.map((pathStep, index) => {
           const pathStepId = index + 1;
```
clarani: Don't use `pathStepId`, the pathStep's id is `pathStep.id`

front/src/applications/stdcm/utils/formatStdcmConfV2.ts
```diff
@@ -123,7 +125,15 @@ export const checkStdcmConf = (
       })
     );
   } else if (searchDatetimeWindow) {
-    const startDatetime = new Date(startTime);
+    const startDatetime = (() => {
```
clarani: I think it would be clearer to handle this like this:
```
let startDatetime: Date | undefined;
if (stdcmV2Activated) {
  ...
} else {
  if (!startTime) {
    ...
  } else {
    startDatetime = new Date(startTime);
  }
}

if (
  startDatetime &&
  searchDatetimeWindow &&
  (startDatetime < searchDatetimeWindow.begin || searchDatetimeWindow.end < startDatetime)
) {
  ...
}
```

front/src/applications/stdcm/utils/formatStdcmConfV2.ts
```diff
@@ -166,29 +179,41 @@ export const checkStdcmConf = (
       } = step;
 
       const duration = stopFor ? sec2ms(ISO8601Duration2sec(stopFor) || Number(stopFor)) : 0;
+      const timing_data =
```
clarani: Snake case

front/src/applications/stdcm/utils/formatStdcmConfV2.ts
```diff
@@ -151,6 +161,9 @@ export const checkStdcmConf = (
       const {
         id,
         arrival,
+        arrivalType,
```
clarani: Do you think you can compute the path outside the return ? (l155)

front/src/applications/stdcmV2/components/StdcmConfig.tsx
```diff
@@ -55,29 +58,87 @@ const StdcmConfig = ({
...
     if (isPending) {
       loaderRef?.current?.scrollIntoView({ behavior: 'smooth' });
     }
   }, [isPending]);
 
+  const checkErrors = () => {
+    if (pathfindingState.error) {
+      setFormErrors({ errorType: StdcmErrorTypes.PATHFINDING_FAILED });
+    }
+    if (origin && destination) {
```
clarani: It should be a `else if` here no ? Otherwise, the other error will erase the pathfinding error

front/src/applications/stdcmV2/components/StdcmConfig.tsx
```diff
@@ -55,29 +58,87 @@ const StdcmConfig = ({
...
   };
 
   const startSimulation = () => {
-    if (pathfindingState.done) {
+    if (pathfindingState.done && !formErrors) {
       launchStdcmRequest();
     }
   };
 
+  const resetOriginArrival = () => {
```
clarani: Maybe we can find a better name for resetOriginArrival and resetDestinationArrival :thinking: Since we're only changing the arrivalTime type no ?
Maybe removeOriginArrivalTime ?

front/src/applications/stdcmV2/components/StdcmOrigin.tsx
```diff
@@ -1,67 +1,74 @@
...
-  const originDate = useSelector(getOriginDate);
-  const originTime = useSelector(getOriginTime);
-  const searchDatetimeWindow = useSelector(getSearchDatetimeWindow);
 
-  const updateOriginV2Point = (pathStep: PathStep | null) => {
-    dispatch(updateOriginV2(pathStep));
-  };
+  const { originArrival, originToleranceValues } = useMemo(
+    () => ({
+      originArrival: origin?.arrival ? extractDateAndTimefromISO(origin?.arrival) : undefined,
```
clarani:
```suggestion
      originArrival: origin?.arrival ? extractDateAndTimefromISO(origin.arrival) : undefined,
```

front/src/utils/date.ts
```diff
@@ -184,5 +202,29 @@ export function serializeDateTimeWithoutYear(date: Date): string {
...
 }
 
-export const isoUtcStringToLocaleDateString = (isoUtcString: string) =>
-  dayjs(isoUtcString).local().format();
+/**
+ * Convert an ISO date into a string formatted as 'DD/MM/YYYY' and extract the numeric values for hours and minutes.
+ * @param {string} arrivalTime - Arrival time at which the train should arrive at the location. (Format: 'YYYY-MM-DDTHH:mm:ss+HH:mm')
+ * @returns {object} An object containing the parsed date and time.
+ */
+export function extractDateAndTimefromISO(arrivalTime: string) {
```
clarani, line 210: From what I discussed last week with Florian, it would be better to convert the string into a date and then to use the librairies to extract the date, the time etc. You can have a look to the function formatDateToString

front/package.json
```diff
@@ -10,10 +10,10 @@
     "@nivo/line": "^0.80.0",
     "@nivo/tooltip": "^0.80.0",
     "@openapi-contrib/openapi-schema-to-json-schema": "^5.1.0",
-    "@osrd-project/ui-core": "^0.0.31",
-    "@osrd-project/ui-icons": "^0.0.31",
+    "@osrd-project/ui-core": "^0.0.34",
+    "@osrd-project/ui-icons": "^0.0.34",
```
kmer2016: You haven't used the `0.0.35` with the reference Date fix ?

front/src/applications/stdcmV2/components/StdcmConfig.tsx
SarahBellaha: Nice catch, thanks.

front/src/utils/date.ts
```diff
@@ -184,5 +202,29 @@ export function serializeDateTimeWithoutYear(date: Date): string {
...
+    arrivalDate: parsedArrival[0], // ISO date part
+    arrivalTime: parsedArrival[1].substring(0, 5), // ISO time part
+    arrivalTimehours: Number(hours),
+    arrivalTimeMinutes: Number(minutes),
+  };
+}
+
+export function isArrivalDateInSearchTimeWindow(
+  arrivalTime: string,
+  searchDatetimeWindow: { begin: Date; end: Date } | undefined
```
kmer2016:
```suggestion
  searchDatetimeWindow?: { begin: Date; end: Date }
```

front/src/utils/date.ts
```diff
@@ -75,6 +86,13 @@ export const formatToIsoDate = (date: number | string, formatDate: boolean = fal
   return dayjs.tz(date, userTimeZone).format(format);
 };
 
+/**
+ * Transform a milliseconds locale date to an ISO 8601 date
+ * @param msDate milliseconds date (elapsed from January 1st 1970)
+ * @return an ISO 8601 date (e.g. 2024-04-25T08:30:00+02:00)
+ */
+export const formatLocaleDateToIsoDate = (msDate: number | string) => dayjs.tz(msDate).format();
```
kmer2016: It would better if we use `Date` instead of number/string as parameter. By passing a `Date`, we are sure it's a date that's passed, but with a `number/string`, you can't be sure.

front/src/applications/stdcmV2/components/StdcmOperationalPoint.tsx
```diff
@@ -148,13 +150,15 @@ const StdcmOperationalPoint = ({ updatePoint, point, disabled }: StdcmOperationa
           disabled={disabled}
         />
       </div>
-      <div className="suggestions stdcm-v2-ch-selector w-100 py-2 col-3">
-        <SelectSNCF
+      <div className="suggestions stdcm-v2-ch-selector w-100 pl-1 pb-2 col-3">
+        <Select
           label={t('trainPath.ch')}
           id="ch"
```
kmer2016: As the component is used many times
```suggestion
          id={`${point?.id}-ch`}
```

Same for `StdcmSuggestions` at line 143

front/src/reducers/osrdconf/stdcmConf/stdcmConfReducers.spec.ts
```diff
@@ -113,3 +115,116 @@ describe('stdcmConfReducers', () => {
...
+      })
+    );
+    const state = store.getState()[stdcmConfSlice.name];
+    expect(last(state.pathSteps)).toEqual({
+      ...newDestination,
+      arrivalToleranceBefore: 300,
+      arrivalToleranceAfter: 300,
+    });
+  });
+});
```
kmer2016: Could you move this code inside `describe('stdcmConfReducers'`? Just before `testCommonConfReducers(stdcmConfSlice);`. This way, when the tests are run, it will be clear that these new cases are related to the stdcmConfReducers.

kmer2016 (changes requested):
Thank you for the PR !

front/package.json
SarahBellaha: I did, maybe you checked an old commit ?

front/src/applications/stdcmV2/components/StdcmVias.tsx
```diff
@@ -64,45 +64,55 @@ const StdcmVias = ({ disabled = false, setCurrentSimulationInputs }: StdcmConfig
...
+            <>
+              <StdcmDefaultCard
+                hasTip
+                text={t('trainPath.addVia')}
+                Icon={<Location size="lg" variant="base" />}
+                onClick={() => addViaOnClick(pathStepIndex)}
+                disabled={disabled}
+              />
+              <StdcmCard
+                key={pathStepIndex}
```
clarani: I think that for the key, you should use the pathStep.id actually, and not the pathStepIndex :) Especially because you're adding some elements in the array

front/src/applications/stdcmV2/components/StdcmConfig.tsx
```diff
@@ -55,29 +58,86 @@ const StdcmConfig = ({
...
   useEffect(() => {
     if (isPending) {
       loaderRef?.current?.scrollIntoView({ behavior: 'smooth' });
     }
   }, [isPending]);
 
+  const checkErrors = () => {
+    if (pathfindingState.error) {
+      setFormErrors({ errorType: StdcmConfigErrorTypes.PATHFINDING_FAILED });
+    } else if (origin && destination) {
```
clarani: I think it would be easier to do if you add early returns after each setFormErrors, what do you think ?

front/src/applications/stdcm/utils/formatStdcmConfV2.ts
```diff
@@ -113,82 +115,112 @@ export const checkStdcmConf = (
...
-      setFailure({
-        name: t('operationalStudies/manageTrainSchedule:errorMessages.trainScheduleTitle'),
-        message: t('operationalStudies/manageTrainSchedule:errorMessages.noOriginTime'),
-      })
-    );
-  } else if (searchDatetimeWindow) {
-    const startDatetime = new Date(startTime);
-    if (startDatetime < searchDatetimeWindow.begin || searchDatetimeWindow.end < startDatetime) {
+
+  const startDateTime = (() => {
```
clarani: It's a bit weird to instantiate this function in order to call it directly after. You don't need this function, use a `let startDatetime: Date | undefined` and add the code below :)

front/src/applications/stdcm/utils/formatStdcmConfV2.ts
```diff
@@ -113,82 +115,112 @@ export const checkStdcmConf = (
...
+        message: t('operationalStudies/manageTrainSchedule:errorMessages.originTimeOutsideWindow', {
+          low: dateTimeFormatting(searchDatetimeWindow.begin),
+          high: dateTimeFormatting(searchDatetimeWindow.end),
+        }),
+      })
+    );
+  }
 
-      const duration = stopFor ? sec2ms(ISO8601Duration2sec(stopFor) || Number(stopFor)) : 0;
+  const path = compact(osrdconf.pathSteps).map((step) => {
```
clarani, line 160: Compute the path only after the early return `if (error) return null`

clarani (commented):
I have not reviewed all the files but LGTM for the files I've reviewed

kmer2016 (approved):
Nice work, LGTM and tested ! Good job !

PR #8366, opened 2024-08-08
core: add margin for planned time steps in stdcm

Erashin:
Time window for a planned time step should be [arrivalTime - toleranceBefore - timeStep, arrivalTime + toleranceAfter + timeStep]

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/api/api_v2/stdcm/STDCMEndpointV2.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/api/api_v2/stdcm/STDCMEndpointV2.kt
@@ -91,9 +91,10 @@ class STDCMEndpointV2(private val infraManager: InfraManager) : Take {
                         workSchedules = request.workSchedules,
                         steps,
                         gridMarginBeforeTrain = request.timeGapBefore.seconds,
-                        gridMarginAfterTrain = request.timeGapAfter.seconds
+                        gridMarginAfterTrain = request.timeGapAfter.seconds,
+                        timeStep = request.timeStep!!.seconds
                     ),
-                    request.timeStep!!.seconds,
+                    request.timeStep.seconds,
                     request.maximumDepartureDelay!!.seconds,
                     request.maximumRunTime.seconds,
                     request.speedLimitTag,
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
@@ -28,6 +28,7 @@ data class BlockAvailability(
     val plannedSteps: List<STDCMStep>,
     val gridMarginBeforeTrain: Double,
     val gridMarginAfterTrain: Double,
+    val internalMarginForSteps: Double,
 ) : BlockAvailabilityInterface {
 
     override fun getAvailability(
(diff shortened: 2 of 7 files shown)


core/src/main/kotlin/fr/sncf/osrd/api/api_v2/stdcm/STDCMEndpointV2.kt
```diff
@@ -91,9 +91,10 @@ class STDCMEndpointV2(private val infraManager: InfraManager) : Take {
                         workSchedules = request.workSchedules,
                         steps,
                         gridMarginBeforeTrain = request.timeGapBefore.seconds,
-                        gridMarginAfterTrain = request.timeGapAfter.seconds
+                        gridMarginAfterTrain = request.timeGapAfter.seconds,
+                        timeStep = request.timeStep!!.seconds
                     ),
-                    request.timeStep!!.seconds,
+                    request.timeStep.seconds,
```
eckter, line 97: `request.timeStep` probably shouldn't be optional

core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
```diff
@@ -181,7 +183,10 @@ data class BlockAvailability(
                         if (plannedMinTimeAtStep > timeAtStep) {
                             // Train passes through planned timing data before it is available
                             return AvailabilityProperties(
-                                plannedMinTimeAtStep - timeAtStep,
+                                max(
+                                    plannedMinTimeAtStep - timeAtStep - internalMarginForSteps,
+                                    0.0
+                                ),
```
eckter: I'd prefer if we update `plannedMinTimeAtStep` instead of offsetting it here. It *feels* like we'd have weird edge cases otherwise, like we'd consider that we're too early/late but then actually not

core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
```diff
@@ -200,7 +205,7 @@ data class BlockAvailability(
                         return AvailabilityProperties(
                             0.0,
                             Offset(0.meters),
-                            plannedMaxTimeAtStep - timeAtStep,
+                            plannedMaxTimeAtStep - timeAtStep + internalMarginForSteps,
```
eckter: Same as above

core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
```diff
@@ -340,6 +345,7 @@ fun makeBlockAvailability(
     steps: List<STDCMStep> = listOf(),
     gridMarginBeforeTrain: Double = 0.0,
     gridMarginAfterTrain: Double = 0.0,
+    timeStep: Double = 2.0,
```
eckter, line 345: I'd prefer if it's renamed here, it doesn't function as a time step. Maybe something like `plannedTimingTolerance`?

eckter (approved):
It could use a test. Testing with `BlockAvailability` is a pain but we can make it work with railjson infras (see `FullSTDCMTests.kt`)

Other than that and the few comments, LGTM (early approval because I'm leaving for the week and I trust you with the fixes)

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/stdcm/STDCMEndpointV2.kt
Erashin: This is from the api, so I don't really wanna change that rn. Defaults to 2.0 as well. We'll circle back to it once you're back from vacation.

PR #8367, opened 2024-08-08
front: fix editing a train add stops at timed points

### LINKED ISSUE #8291: TSV2 : Editing a train with a point with time constraint (but no stopping time) creates a 0s stopping time constraint
### What happened?

1.I created a train with a time constraint (but no stopping time constraint) in Lantenay
image
2. The train is created, and there is no stop in Lantenay (correct)
image
3. I click on "edit" to open the edition mode and then on "edit train" (without changing anything) -> The train now stops in Lantenay : a 0s stopping time was added (problem)
image

If I reopen the editition mode, I can see that a 0s stop was indeed added : 
image

I expected no stop to be added

as above

Recette (SNCF)

Google Chrome

0a34296

clarani:
closes #8291

### DIFF
--- a/front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatSchedule.ts
+++ b/front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatSchedule.ts
@@ -1,4 +1,4 @@
-import { compact, isNaN } from 'lodash';
+import { compact, isNaN, isNil } from 'lodash';
 
 import type { TrainScheduleBase } from 'common/api/osrdEditoastApi';
 import type { PathStep } from 'reducers/osrdconf/types';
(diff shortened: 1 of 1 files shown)


front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/formatSchedule.ts
```diff
@@ -32,9 +32,10 @@ const formatSchedule = (
         arrival: formatArrival ?? undefined,
         locked: step.locked,
         on_stop_signal: step.onStopSignal,
-        stop_for: isNaN(Number(step.stopFor))
-          ? undefined
-          : formatDurationAsISO8601(Number(step.stopFor)),
+        stop_for:
+          isNil(step.stopFor) || isNaN(Number(step.stopFor))
```
theocrsb, line 36: Maybe you can just add this condition ?

```suggestion
          !step.stopFor
```

Uriel-Sautron: I think we should keep `isNaN`

clarani: I think we should keep the isNan, it's to prevent bad behaviours from the user (if he tries to enter 'abc' in the cell for instance)

clarani:
I added some details in the commit message :+1:

theocrsb (approved):
LGTM and tested !

SharglutDev (approved):
Lgtm and tested

Uriel-Sautron (approved):
LGTM, tested !!!

PR #8372, opened 2024-08-08
front: save NGE node positions in localStorage

emersion:
Short-term fix for setup persistance for node positions in localStorage

See long-term solution : https://github.com/osrd-project/osrd-confidential/issues/597

### DIFF
--- a/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
@@ -1,3 +1,5 @@
+import { debounce } from 'lodash';
+
 import { osrdEditoastApi, type TrainScheduleBase } from 'common/api/osrdEditoastApi';
 import type { AppDispatch } from 'store';
 
@@ -114,6 +116,11 @@ const handleTrainrunOperation = async ({
   }
 };
 
+const handleUpsertNode = debounce((timeTableId: number, node: Node) => {
+  const { betriebspunktName: trigram, positionX, positionY } = node;
+  nodeStore.set(timeTableId, { trigram, positionX, positionY });
+}, 500);
+
 const handleNodeOperation = ({
   type,
   node,
@@ -123,15 +130,14 @@ const handleNodeOperation = ({
   node: Node;
   timeTableId: number;
 }) => {
-  const { betriebspunktName: trigram, positionX, positionY } = node;
   switch (type) {
     case 'create':
     case 'update': {
-      nodeStore.set(timeTableId, { trigram, positionX, positionY });
+      handleUpsertNode(timeTableId, node);
       break;
     }
     case 'delete': {
-      nodeStore.delete(timeTableId, trigram);
+      nodeStore.delete(timeTableId, node.betriebspunktName);
       break;
     }
     default:
--- a/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
@@ -159,7 +159,7 @@ const handleOperation = async ({
   const { type } = event;
   switch (event.objectType) {
     case 'node':
-      handleNodeOperation({ type, node: event.node!, timeTableId });
+      handleNodeOperation({ type, node: event.node, timeTableId });
       break;
     case 'trainrun':
       await handleTrainrunOperation({
--- a/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
@@ -116,10 +116,10 @@ const handleTrainrunOperation = async ({
   }
 };
 
-const handleUpsertNode = debounce((timeTableId: number, node: Node) => {
+const handleUpdateNode = debounce((timeTableId: number, node: Node) => {
   const { betriebspunktName: trigram, positionX, positionY } = node;
   nodeStore.set(timeTableId, { trigram, positionX, positionY });
-}, 500);
+}, 2000);
 
 const handleNodeOperation = ({
   type,
@@ -133,7 +133,7 @@ const handleNodeOperation = ({
   switch (type) {
     case 'create':
     case 'update': {
-      handleUpsertNode(timeTableId, node);
+      handleUpdateNode(timeTableId, node);
       break;
     }
     case 'delete': {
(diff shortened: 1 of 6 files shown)


louisgreiner (approved):
First version of node position persistance, to allow user to work on its browser with NGE.

Remains to do : 
- [ ] save position at node creation (for now it only works after going back to Micro then back to Macro)
- [ ] handle non-trigram nodes

These enhancements can be done in a later work. Howerver, the current feature is highly needed for user friendliness and a good fix for now.

Later, this persistance layer will be implemented in the database, see https://github.com/osrd-project/osrd-confidential/issues/597

louisgreiner:
Bugs noticed while testing:
- entries in `localStorage` are not removed when a node disappear while Macro->Micro. For example, if 2 nodes are connected by only 1 trainrun and that we delete the trainrun, the nodes will disappear after doing Macro->Micro, but not their entries in `localStorage`
- when editing the `Operational point` name, an update event is sent at EVERY change, and thus creating a new entry for EACH in `localStorage`. For example, if you enter "PLY", you'll have an entry for "P", "PL", and "PLY". Worst case scenario : you already have an entry for "PL", which ends up being rewrite with the position of "PLY". **EDIT**: fixed by adding a debounce time of 2000ms, but the long term fix of that should be on NGE side, and then revert this one

front/src/applications/operationalStudies/components/MacroEditor/types.ts
```diff
@@ -154,8 +154,9 @@ export type NetzgrafikDto = {
...
 };
 
-export type NGETrainrunEvent = {
+export type NGEEvent = {
   type: 'create' | 'delete' | 'update';
-  objectType: 'trainrun';
-  trainrun: Trainrun;
+  objectType: 'node' | 'trainrun';
+  trainrun?: Trainrun;
+  node?: Node;
```
clarani: Use a type like this, this way we know in the type that if the objectType is node, then we have a Node and same for the trainrun
```suggestion
  } & (
  | {
      objectType: 'trainrun';
      trainrun: Trainrun;
    }
  | { objectType: 'node'; node: Node }
);
```

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
```diff
@@ -113,4 +114,59 @@ const handleOperation = async ({
...
+}: {
+  event: NGEEvent;
+  dispatch: AppDispatch;
+  timeTableId: number;
+  netzgrafikDto: NetzgrafikDto;
+}) => {
+  const { type } = event;
+  switch (event.objectType) {
+    case 'node':
+      handleNodeOperation({ type, node: event.node!, timeTableId });
```
clarani: Once you have modified the NGEEvent, you can remove the `!` after `node` l156 and `trainrun` line161

louisgreiner: true !

front/src/applications/operationalStudies/components/MacroEditor/nodeStore.ts
```diff
@@ -0,0 +1,49 @@
...
+    entries = JSON.parse(rawNodes).map((node: Node) => [node.trigram, node]);
+  }
+
+  nodesCache = new Map(entries);
+  return nodesCache;
+};
+
+const saveNodes = (nodes: Map<string, Node>) => {
+  localStorage.setItem(NODES_LOCAL_STORAGE_KEY, JSON.stringify([...nodes.values()]));
+};
```
Uriel-Sautron: When the page is refreshed, our saved nodes no longer have their saved positions.

```suggestion
const loadNodes = () => {
  if (nodesCache) return nodesCache;

  let entries = [];
  const rawNodes = localStorage.getItem(NODES_LOCAL_STORAGE_KEY);
  if (rawNodes) {
    entries = JSON.parse(rawNodes).map((node: Node) =>  node);
  }

  nodesCache = new Map(entries);
  return nodesCache;
};

const saveNodes = (nodes: Map<string, Node>) => {
  localStorage.setItem(NODES_LOCAL_STORAGE_KEY, JSON.stringify([...nodes]));
};
```

louisgreiner: done thanks

Uriel-Sautron (commented):
Tested, just a comment, otherwise good work

Uriel-Sautron (approved):

front/src/applications/operationalStudies/components/MacroEditor/nodeStore.ts
```diff
@@ -0,0 +1,49 @@
...
+
+let nodesCache: Map<string, Node> | null;
+
+const loadNodes = () => {
+  if (nodesCache) return nodesCache;
+
+  let entries = [];
+  const rawNodes = localStorage.getItem(NODES_LOCAL_STORAGE_KEY);
+  if (rawNodes) {
+    entries = JSON.parse(rawNodes).map((node: Node) => node);
```
clarani, line 17:
```suggestion
    entries = JSON.parse(rawNodes);
```

front/src/applications/operationalStudies/components/MacroEditor/nodeStore.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  nodesCache = new Map(entries);
+  return nodesCache;
+};
+
+const saveNodes = (nodes: Map<string, Node>) => {
+  localStorage.setItem(NODES_LOCAL_STORAGE_KEY, JSON.stringify([...nodes]));
```
clarani, line 25:
```suggestion
  localStorage.setItem(NODES_LOCAL_STORAGE_KEY, JSON.stringify(nodes));
```

front/src/applications/operationalStudies/components/MacroEditor/nodeStore.ts
clarani:
```suggestion
    entries = JSON.parse(rawNodes).map((node: [string, Node]) => node);
```

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
```diff
@@ -113,4 +116,63 @@ const handleOperation = async ({
...
+}) => {
+  const { type } = event;
+  switch (event.objectType) {
+    case 'node':
+      handleNodeOperation({ type, node: event.node, timeTableId });
+      break;
+    case 'trainrun':
+      await handleTrainrunOperation({
+        type,
+        trainrun: event.trainrun!,
```
clarani: You can remove the ! here
```suggestion
        trainrun: event.trainrun,
```

clarani (approved):
You can resolve the 2 remaining comments on your own :)

emersion:
Thanks for finishing this up! Next time, please squash commits together.

PR #8373, opened 2024-08-08
osrdyne: fix missing options on k8s driver

ghost:


bloussou (approved):
thanks

Khoyo (approved):
s/core/workers/, otherwise LGTM

PR #8374, opened 2024-08-08
docker: fix restart policy type for core

emersion:
In YAML, "no" is a boolean. This makes podman-compose choke on the docker-compose.yml, because it expects a string:

    Error: running container create option: "False" is not a valid restart policy: invalid argument

Use a string instead, just like wait-healthy already does.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -63,7 +63,7 @@ services:
       additional_contexts:
         test_data: tests/data
         static_assets: assets
-    restart: no
+    restart: "no"
     command: "true"
 
   front:


ghost (approved):
LGTM, is it podman specific ? I do not remember we've encountered an issue while developing.

emersion:
Yeah, likely docker-compose doesn't actually care about the type but podman does.

ghost:
> Yeah, likely docker-compose doesn't actually care about the type but podman does.

Sorry about that, we didn't test it with podman. I'll think about this next time.

emersion:
For the spec, see https://github.com/compose-spec/compose-spec/blob/main/05-services.md#restart

> Sorry about that, we didn't test it with podman. I'll think about this next time.

No worries, I'm very aware that this is totally unsupported and all on me.

Khoyo (approved):
YAML really is such a reasonable markup language

PR #8378, opened 2024-08-08
front: fix NGE trainrun update overwriting other fields

emersion:
When updating a trainrun in NGE, we'd send a request to update the matching train schedule in OSRD, but the payload always contained default values for start_time and rolling_stock_name. These end up overwriting any previous values the user might've entered. Moreover, all other fields would be reset to null.

Fix this by fetching the train schedule we're about to update before sending the request to update.

Note, the schedule array is reset because we don't yet support converting NGE times into OSRD. We can't copy over the original `schedule` field because the user might've edited the train path in an arbitrary fashion inside NGE so we cannot match up reliably old path steps and new ones. We overwrite the path steps with our own, each with a fresh ID, so the IDs in `schedule` are stale (and are rejected by the backend). This will be fixed naturally when we implement the NGE → OSRD conversion for stop times.

### DIFF
--- a/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
@@ -34,7 +34,7 @@ const createTrainSchedulePayload = (
   trainrunSections: TrainrunSection[],
   nodes: Node[],
   trainrun: Trainrun
-): TrainScheduleBase => {
+) => {
   // TODO: check that the trainrunSections format is still compatible
   const path = trainrunSections.flatMap((section, index) => {
     const sourceNode = getNodeById(nodes, section.sourceNodeId);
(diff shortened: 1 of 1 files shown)


front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
```diff
@@ -96,14 +100,20 @@ const handleOperation = async ({
     }
     case 'update': {
       const trainrunIdToUpdate = createdTrainrun.get(trainrun.id) || trainrun.id;
+      const trainSchedule = await dispatch(
```
clarani, line 103: For your information, later, we will have already fetched all the trainSchedules of the timetable in the parent component and we may pass the trainSchedules directly in props.

clarani (approved):

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
emersion: My hope was that the automagic rtk-query cache would already contain the train schedule. But I haven't verified whether in practice it works that way.

PR #8379, opened 2024-08-08
front: fix output table

### LINKED ISSUE #8369: front: in the outputTable, a honored point is in orange
### What happened?

I created a train with a scheduled point. 
The train card in the timetable component shows that the train honored all its scheduled points, but when having a look at the output table, the row of the schedule point is in orange and the arrival time has not been rounded

image

The row background should not be in orange

1. Create a train between PLY MSC with a TGVA, departure time at 10h00 and add a schedule point at Pasilly at 10H50
2. Look at the output table

_No response_

_No response_

dev

clarani:
closes #8369


clarani:
Not relevant anymore since the issue was duplicated and the problem was solved by another PR :+1:

PR #8381, opened 2024-08-08
front: fix nge node position calculation

louisgreiner:
Fix NGE node position when dividing width or height by 0


SharglutDev (approved):
Lgtm and tested

PR #8383, opened 2024-08-08
front: fix edit form for invalid trains

### LINKED ISSUE #8375: Can't produce PathSteps when pathfinding fails (i.e., no rolling stock defined)
### What happened?

Train Schedules v2 should allow a train schedule to exist with no rolling stock defined, and still be able to set its PathSteps.

image

Being able to create a train schedule with no rolling stock, persist its pathItems and recover them as pathSteps

1. Open a scenario in Operational Studies -> Macroscopic
2. Create 2 nodes (NES and MES if on small infra let's say)
3. Create a trainrun between them
4. Go back to Microscopic and edit the train schedule
5. Notice absence of PathSteps in Route window

-> Some work has been already done [here](https://github.com/OpenRailAssociation/osrd/tree/lgr/fix-train-schedule-without-rs-path-steps) to fix this issue

Local, Dev (DgexSol)

Firefox

-

emersion:
_See individual commits._

To create an invalid train schedule, one can go to the macro editor and create a new trainrun between 2 nodes there.

Closes: https://github.com/OpenRailAssociation/osrd/issues/8375

### DIFF
--- a/front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
+++ b/front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
@@ -8,6 +8,7 @@ import {
   type PathItemLocation,
   type PostV2InfraByInfraIdPathPropertiesApiArg,
   type PostV2InfraByInfraIdPathfindingBlocksApiArg,
+  type TrainScheduleResult,
 } from 'common/api/osrdEditoastApi';
 import { useOsrdConfActions, useOsrdConfSelectors } from 'common/osrdContext';
 import { formatSuggestedOperationalPoints, upsertPathStepsInOPs } from 'modules/pathfinding/utils';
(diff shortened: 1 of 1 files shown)


front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -43,23 +92,14 @@ const useSetupItineraryForTrainUpdate = (
...
+      trainSchedule: TrainScheduleResult
+    ): Promise<ItineraryForTrainUpdate | null> => {
       if (!infraId) {
-        return;
+        return null;
       }
-      const trainSchedule = await getTrainScheduleById({
-        id: trainIdToEdit,
-      }).unwrap();
       if (!trainSchedule.rolling_stock_name) {
```
kmer2016, line 101: How do you think about merge the first conditions ?
```typescript
if (!infraId || !trainSchedule.rolling_stock_name) {
        return null;
}
```

kmer2016 (commented):
tested ! Just a small suggestion

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
emersion: The first `if (!infraId)` is before the `await getTrainScheduleById()`, and we can only check for `if (!trainSchedule.rolling_stock_name)` after the HTTP request. I prefer to not send an HTTP request if we can avoid it, so that's why I left it that way. (I'd be fine moving both `if`s after the HTTP request if you prefer.)

achrafmohye (approved):
LGTM and tested

achrafmohye:
Well articulated comments that provide excellent context. Thanks 👍

kmer2016 (approved):
LGTM and tested

PR #8384, opened 2024-08-08
front: upgrade NGE to forked v2.7.5

emersion:
This small update contains two bugfixes:

- A fix for the arrival time input: it would previously change to a value different from the one entered by the user.
- The occupancy diagram has been hidden because it still shows return travels.


SharglutDev (approved):
Lgtm and tested

RomainValls (approved):
Lgtm and tested 🙂

PR #8403, opened 2024-08-09
front: refacto time stops table props type

### LINKED ISSUE #8402: front: refacto the timesStopsTable props type
### Description and goal

The timesStops table takes a array of SuggestedOp as main props, but it doesn't need all the fields of the SuggestedOp. This is quite painful because the Input table indeed needs whole SuggestedOp, but not the output one.

We should simplify the timesStops table props type and use a generic type so each table (input or output) can give to the timesStops component only what it needs.

### Acceptance criteria

- [ ] the code is clearer and more easily maintainable

clarani:
closes #8402

### DIFF
--- a/front/.eslintrc
+++ b/front/.eslintrc
@@ -44,6 +44,7 @@
     "@typescript-eslint/no-shadow": "error",
     "@typescript-eslint/no-use-before-define": "error",
     "@typescript-eslint/no-unnecessary-type-assertion": "error",
+    "@typescript-eslint/no-unsafe-return": "error",
 
     "@typescript-eslint/ban-types": [
       "error",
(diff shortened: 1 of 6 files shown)


front/src/modules/timesStops/TimesStopsInput.tsx
```diff
@@ -66,24 +58,68 @@ const TimesStopsinput = ({ allWaypoints, startTime, pathSteps }: TimesStopsInput
...
+    ) {
+      row.stopFor = String(time2sec(row.departure) - time2sec(row.arrival));
+    }
+    if (!row.stopFor && op.fromRowIndex !== rows.length - 1) {
+      row.onStopSignal = false;
+    }
+
+    if (row.theoreticalMargin && !marginRegExValidation.test(row.theoreticalMargin!)) {
+      row.isMarginValid = false;
+      setRows(rows);
```
anisometropie: if we can a way to keep the setRows inside `TimesStops` that would work. but that’s not good design to allow users to know implementation details of an api. we can’t let TimesStopsInput know about setRows, let alone act on it.

A solution would be to call setRows inside `TimesStops` no matter what, and update the store on conditions.

I made a lot of changes here, in #8334 I’d prefer to do this refacto after it’s done

PR #8415, opened 2024-08-09
Fix track offset calculation when the track range direction is STOP_TO_START

### LINKED ISSUE #8306: Failed to add more than one power restriction
### What happened?

https://github.com/user-attachments/assets/ee22e0bd-1b4e-4d9e-b75a-f2975be83038

Can add the power restriction successfully 

1. Edit a train 
2. Add more than one power restriction 
3. Click on "Edit train"
https://rec-osrd.reseau.sncf.fr/operational-studies/projects/39/studies/142/scenarios/264

Recette (SNCF)

Google Chrome

0a34296

achrafmohye:
Closes #8306

### DIFF
--- a/front/src/modules/powerRestriction/helpers/__tests__/createPathStep.spec.ts
+++ b/front/src/modules/powerRestriction/helpers/__tests__/createPathStep.spec.ts
@@ -48,7 +48,7 @@ describe('findTrackOffset', () => {
     const trackRanges = [
       { track_section: 'track_0' },
       { track_section: 'track_1' },
-      { track_section: 'track_2', begin: 1060, end: 60, direction: 'STOP_TO_START' },
+      { track_section: 'track_2', begin: 60, end: 1060, direction: 'STOP_TO_START' },
       { track_section: 'track_3' },
       { track_section: 'track_4' },
     ] as TrackRange[];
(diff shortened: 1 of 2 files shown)


emersion:
Ref https://github.com/OpenRailAssociation/osrd/pull/8218#discussion_r1698473643 which has a discussion about this.

front/src/modules/powerRestriction/helpers/__tests__/createPathStep.spec.ts
```diff
@@ -60,7 +60,7 @@ describe('findTrackOffset', () => {
       trackRanges
     );
 
-    expect(result).toEqual({ track: 'track_2', offset: 1050 });
+    expect(result).toEqual({ track: 'track_2', offset: 50 });
```
clarani: You need to change the input, for the track_2, the begin must be 60 and the end 1060 (as we have seen that the begin < end in the backend response)

clarani (changes requested):

front/src/modules/powerRestriction/helpers/__tests__/createPathStep.spec.ts
achrafmohye: Fixed :)

clarani (approved):
LGTM :white_check_mark:

RomainValls (approved):
LGTM

PR #8416, opened 2024-08-09
front: convert times from NGE to OSRD

### LINKED ISSUE #8389: Converting times from NGE to OSRD
### Description and goal

Part of ticket https://github.com/OpenRailAssociation/osrd/issues/7539

The goal of this issue is to have the conversion of times written in NGE to write them in the `train_schedule` in OSRD

### Acceptance criteria

- [ ] The time of the first node of the `trainrun` in NGE is converted as the start time in OSRD (NGE gives minutes only, so the hour is arbitrarly chosen to 8h)
- [ ] The time of intermediate nodes of the `trainrun` in NGE are converted as time contraints on the corresponding points in OSRD :  
    - [ ] arrival time at the node is the arrival time for the point in OSRD
    - [ ] if there is no stop in NGE, there is no stop in OSRD
    - [ ] if there is a stop in NGE, there is a stop in OSRD and the stopping time is computed as the difference between the departure and the arrival time
- [ ] The time of the last node of the `trainrun` in NGE is converted as a time constraint on the last point in OSRD : 
    - [ ] arrival time at the node is the arrival time for the point in OSRD
    - [ ] the stopping time is 0  

### Implementation plan

- [ ] The time of the first node of the `trainrun` in NGE is written in the `start_time` attribute in OSRD
- [ ] For other nodes
  - [ ] the arrival is written in `arrival` of the `path_item` for the attribute `schedule` in OSRD
  - [ ] if there is no stop, the `stopFor` attribute of the `path_item` for `schedule` in OSRD is `None`
  - [ ] if there is a stop, the `stopFor` attribute of the `path_item` for `schedule` is the difference between the departure time and the arrival time
- [ ] The `consecutiveTime` from NGE must be used to know which hour to write in the `arrival` attribute (5 means 8h05 but 65 means 9h05 for example)

emersion:
- [x] `isNonStopTransit` makes NGE blow up with `NaN` when editing times
- [x] Somehow times are updated on the backend but the frontend refuses to display the new information? Yes, see: https://github.com/OpenRailAssociation/osrd/pull/8555

Closes: https://github.com/OpenRailAssociation/osrd/issues/8389

### DIFF
--- a/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
@@ -7,9 +7,19 @@ import {
   type TrainScheduleResult,
 } from 'common/api/osrdEditoastApi';
 import type { AppDispatch } from 'store';
+import { formatToIsoDate } from 'utils/date';
+import { calculateTimeDifferenceInSeconds, formatDurationAsISO8601 } from 'utils/timeManipulation';
 
 import nodeStore from './nodeStore';
-import type { NetzgrafikDto, NGEEvent, TrainrunSection, Node, Trainrun, Label } from '../NGE/types';
+import type {
+  NetzgrafikDto,
+  NGEEvent,
+  TrainrunSection,
+  Node,
+  TimeLock,
+  Trainrun,
+  Label,
+} from '../NGE/types';
 
 const createdTrainrun = new Map<number, number>();
 
@@ -90,14 +100,9 @@ const getTrainrunSectionsByTrainrunId = (netzgrafikDto: NetzgrafikDto, trainrunI
   return orderedSections;
 };
 
-// TODO: add a dynamic values when available
-const DEFAULT_PAYLOAD: Pick<
-  TrainScheduleBase,
-  'constraint_distribution' | 'rolling_stock_name' | 'start_time'
-> = {
+const DEFAULT_PAYLOAD: Pick<TrainScheduleBase, 'constraint_distribution' | 'rolling_stock_name'> = {
   constraint_distribution: 'STANDARD',
   rolling_stock_name: '',
-  start_time: '2024-07-15T08:00:00+02:00',
 };
 
 const createPathItemFromNode = async (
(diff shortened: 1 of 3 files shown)


front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
```diff
@@ -106,26 +110,69 @@ const createPathItemFromNode = (node: Node, index: number) => {
...
-      const destinationPathItem = createPathItemFromNode(targetNode, index);
+      const destinationPathItem = createPathItemFromNode(targetNode, index + 1);
       return [originPathItem, destinationPathItem];
     }
     return [originPathItem];
   });
 
+  // The departure time of the first section is guaranteed to be non-null
+  const startTimeLock = trainrunSections[0].sourceDeparture;
+  const startDate = new Date(oldStartDate);
```
SharglutDev, line 195: Why do we need to `new Date` the date which already is a `new Date` ?

louisgreiner:
It looks like changing the source arrival time for the first section also changes the departure time (caused directly by NGE times propagation).
It then schedules the start and next stop of the train schedule in OSRD.
In addition, the travel time is always `0`, don't we want to play with it to adjust the differents times ?

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
emersion: `startDate` is mutated on the next line, and we don't want to mutate `oldStartDate`.

SharglutDev: Yep I get it but what I meant was do we have to setup the `oldStartDate` outside of the `createTrainSchedulePayload` ? Can't we initialize it directly inside it ?

emersion: It's a different value whether we're creating a new train or we're updating one.

louisgreiner:
`isNonStopTransit` sets the times at `0`. It would be nice to set 2' of stopping time at this point, but since we don't deal with simulation results yet, i wonder how to deal with that :( 

Before:
image

After:
image
image

emersion:
Right, `isNonStopTransit` is not handled in this PR at all (left as a TODO).

SharglutDev (commented):
Another weird thing : when creating a train with a via in NGE and asking the train to stop at this via, the output table is properly updated with the stop but not the input table.

louisgreiner:
image

Interacting with NGE times triggers a `POST v2/trainschedule/` that leads to a `404`, I think it is related with `UpsertedTrainSchedules` feature. Is it in the scope of that PR ? I'm not sure to understand what it does and if the error is important

emersion:
> Interacting with NGE times triggers a POST v2/trainschedule/ that leads to a 404, I think it is related with UpsertedTrainSchedules feature. Is it in the scope of that PR ? I'm not sure to understand what it does and if the error is important

I believe this bug happens without this PR as well.

louisgreiner (approved):
Tested and LGTM. Thanks a lot for this precious feature :100:

SharglutDev (approved):
Lgtm and tested. The input table issue will be done in a different PR.

PR #8422, opened 2024-08-12
editoast: add roles check to `views::pathfinding`

hamz2a:


hamz2a:
It's going to be removed with the TSV1.

PR #8423, opened 2024-08-12
front: show 0s on output table when 0s has been added in input table

### LINKED ISSUE #8290: TSV2 : Output table doesn't show 0s stopping times
### What happened?

I created a train with a stopping time of 0s.
image

The train is created but I don't see my stopping time in the output table
image

I expected to see the 0s stopping time in the output table

Create a train with a stopping time of 0s at any point

Recette (SNCF)

Google Chrome

0a34296

theocrsb:
close #8290

### DIFF
--- a/front/src/modules/timesStops/helpers/scheduleData.ts
+++ b/front/src/modules/timesStops/helpers/scheduleData.ts
@@ -16,7 +16,9 @@ export function computeScheduleData(schedule: ScheduleEntry, startTime: string) 
   const arrivalSeconds = schedule.arrival
     ? startTimeSeconds + ISO8601Duration2sec(schedule.arrival)
     : null;
+
   const stopForSeconds = schedule.stop_for ? ISO8601Duration2sec(schedule.stop_for) : null;
+
   const departure =
     arrivalSeconds && stopForSeconds ? startTimeSeconds + arrivalSeconds + stopForSeconds : null;
   return {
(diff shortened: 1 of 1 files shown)


front/src/modules/timesStops/helpers/scheduleData.ts
```diff
@@ -30,6 +32,9 @@ export function formatScheduleData(scheduleData: ComputedScheduleEntry) {
   return {
     arrival: scheduleData.arrival ? secToHoursString(scheduleData.arrival, true) : '',
     departure: scheduleData.departure ? secToHoursString(scheduleData.departure, true) : '',
-    stopFor: scheduleData.stopFor ? String(scheduleData.stopFor) : '',
+    stopFor:
+      scheduleData.stopFor !== null && scheduleData.stopFor !== undefined
```
anisometropie: `stopFor` is `number | null` you shouldn’t need to check for `undefined`

theocrsb: You're right. It's done

front/src/modules/timesStops/helpers/scheduleData.ts
```diff
@@ -30,6 +32,6 @@ export function formatScheduleData(scheduleData: ComputedScheduleEntry) {
   return {
     arrival: scheduleData.arrival ? secToHoursString(scheduleData.arrival, true) : '',
     departure: scheduleData.departure ? secToHoursString(scheduleData.departure, true) : '',
-    stopFor: scheduleData.stopFor ? String(scheduleData.stopFor) : '',
+    stopFor: scheduleData.stopFor !== null ? String(scheduleData.stopFor) : '',
```
Uriel-Sautron, line 35: This solution solves the bug of 0s not being displayed in the output table, but we still have the problem we discussed together.
1. On the input table enter a value in one of the stop for.
2. Delete this value.
3. Add your train.
4. In the output table we find 0s in the stop for when it shouldn't be

Uriel-Sautron (approved):
LGTM, tested !!! Good work.

clarani (approved):

PR #8425, opened 2024-08-12
editoast: add roles checks to documents and electrical profiles api

Wadjetz:


hamz2a (approved):
LGTM.

editoast/editoast_authz/src/builtin_role.rs
```diff
@@ -12,6 +12,16 @@ pub enum BuiltinRole {
...
+
+    #[strum(serialize = "document:read")]
+    DocumentRead,
+    #[strum(serialize = "document:write")]
+    DocumentWrite,
+
+    #[strum(serialize = "electrical_profile:read")]
+    ElectricalProfileRead,
+    #[strum(serialize = "electrical_profile:write")]
+    ElectricalProfileWrite,
```
leovalais: electrical profiles don't need their roles since it doesn't make sense for a user to manipulate them if they're not allowed to manage infras. You can just use `InfraRead/Write`

cf. https://github.com/OpenRailAssociation/osrd/pull/8317

leovalais (changes requested):

leovalais (approved):
LGTM, thanks!

PR #8427, opened 2024-08-12
front: fix rounding of arrival time

### LINKED ISSUE #8426: Rounding of arrival time broke again
### What happened?

https://github.com/OpenRailAssociation/osrd/pull/8222 fixed the issue originally, but a more recent change in the way operationalPoint time is returned from the backend broke the fix.

Screenshot_20240812_193129

rounding should work again

- 

_No response_

_No response_

-

anisometropie:
closes #8426


clarani (approved):

Issue #8437, opened 2024-08-13
The ❌ icon is replaced by an eraser in input table


theocrsb:
### What is this feature about?

We want to change the icon to an "eraser" icon

![image](https://github.com/user-attachments/assets/0a064feb-f951-436c-9afa-6cd3e4f86dbc)

### Why is this feature valuable?

The ❌  in the timetable has changed its behaviour, so it must also change its visual appearance

PR #8438, opened 2024-08-13
front: reset the pathStep properties instead of deleting the pathStep

### LINKED ISSUE #8164: Schedule table: replace ❌ icon to an eraser and change it's behavior
## Description

Today using the ❌ on the schedule table delete the path item wich can impact the pathfinding.
We should instead of removing the path item, clear schedule input (stop, schedule point, margin).
Since the button change its meaning we want to change the icon to an "eraser" icon

image

## Acceptance criteria

- [ ] The ❌icon is replaced by an eraser
- [ ] Clicking on this button erases the information written for this line of the input table (arrival time, stopping time, margin, closed signal checkbox)
- [ ] Clicking on this button DOES NOT remove the point from the path list

theocrsb:
Close #8164 but we don't have an 'eraser' logo. I have opened a new [ticket](https://github.com/OpenRailAssociation/osrd/issues/8437).
And this PR should also fix this [comment](https://github.com/OpenRailAssociation/osrd/issues/7585#issuecomment-2249770837)

### DIFF
--- a/front/src/modules/timesStops/TimesStopsInput.tsx
+++ b/front/src/modules/timesStops/TimesStopsInput.tsx
@@ -2,16 +2,14 @@
 import React from 'react';
 
 import { useOsrdConfActions } from 'common/osrdContext';
-import { isVia } from 'modules/pathfinding/utils';
 import type { SuggestedOP } from 'modules/trainschedule/components/ManageTrainSchedule/types';
 import type { PathStep } from 'reducers/osrdconf/types';
 import { useAppDispatch } from 'store';
-import { removeElementAtIndex } from 'utils/array';
 
-import TimesStops, { WITH_KP } from './TimesStops';
+import TimesStops from './TimesStops';
 import { TableType, type PathWaypointRow } from './types';
 
-type DeleteButtonProps = {
+type ClearButtonProps = {
   removeVia: () => void;
   rowIndex: number;
   rowData: PathWaypointRow;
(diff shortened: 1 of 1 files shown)


front/src/modules/timesStops/TimesStopsInput.tsx
```diff
@@ -56,13 +55,31 @@ const TimesStopsinput = ({ allWaypoints, startTime, pathSteps }: TimesStopsInput
       if ('uic' in step) {
```
Akctarus: For readability, maybe you can use something like that, so the extracted functions can be used in `findIndex`;

```
const isMatchingUICStep = (step: PathStep) =>
      'uic' in step &&
      step.uic === rowData.uic &&
      step.ch === rowData.ch &&
      step.name === rowData.name;

    const isMatchingTrackStep = (step: PathStep) =>
      'track' in step &&
      step.track === rowData.track &&
      step.positionOnPath === rowData.positionOnPath &&
      step.offset === rowData.offsetOnTrack &&
      step.id === rowData.opId;

    const index = pathSteps.findIndex(
      (step) => isMatchingUICStep(step) || isMatchingTrackStep(step)
    );
```

Akctarus (commented):
Tested, works well. For the bug to which this PR is linked, this closes it well since it is no longer reproducible.

front/src/modules/timesStops/TimesStopsInput.tsx
theocrsb: You're right. It's done.
I also rename `removeVia` in `clearPathStep` because I find this name more appropriate.

kmer2016 (commented):
The code looks fine to me. However, I'm not sure if it's intentional, but when the pathStep is erased, the information is deleted, yet the icon remains. Was this the intended behavior?

theocrsb:
> The code looks fine to me. However, I'm not sure if it's intentional, but when the pathStep is erased, the information is deleted, yet the icon remains. Was this the intended behavior?

After talking with @Tguisnet. We think it's better to remove the cross when everything is clear.

kmer2016 (approved):
LGTM and tested

PR #8444, opened 2024-08-14
Adapt new speed limit source to speedspacechart

### LINKED ISSUE #8328: Front: branch speed-limit source to speed-space chart in OpStudies and STDCM
The goal is to plug back output to `core-ui` once all is available, to have the feature fully implemented.

Depends on https://github.com/OpenRailAssociation/osrd/issues/7717 and https://github.com/OpenRailAssociation/osrd/issues/7732 (part of https://github.com/osrd-project/osrd-confidential/issues/364)

AC: In the end, we should have the speed-limit sources (tags) available in osrd as designed in https://github.com/osrd-project/osrd-confidential/issues/364 (except maybe for the long names that should disappear with https://github.com/OpenRailAssociation/osrd/issues/7977)

Related idea (postponed, to be decided if relevant): https://github.com/OpenRailAssociation/osrd-ui/issues/428

RomainValls:
close #8328

### DIFF
--- a/front/src/modules/simulationResult/components/SpeedSpaceChart/helpers.ts
+++ b/front/src/modules/simulationResult/components/SpeedSpaceChart/helpers.ts
@@ -14,9 +14,76 @@ import type {
   SimulationResponseSuccess,
 } from 'applications/operationalStudies/types';
 import type { ReportTrain } from 'common/api/osrdEditoastApi';
+import { TAG_COLORS } from 'modules/simulationResult/consts';
+import type { SpeedLimitTagValue } from 'modules/simulationResult/types';
 import { mmToKm, msToKmh, mToKm } from 'utils/physics';
 
 import { electricalProfilesDesignValues } from './consts';
+
+const getTag = (source?: SpeedLimitTagValue['source']): { name: string; color: string } => {
+  let name: string;
+  let color: string;
+
+  switch (source?.speed_limit_source_type) {
+    case 'given_train_tag':
+      name = source.tag;
+      color = TAG_COLORS.GIVEN_TRAIN;
+      break;
+    case 'fallback_tag':
+      name = source.tag;
+      color = TAG_COLORS.FALLBACK;
+      break;
+    case 'unknown_tag':
+      name = 'incompatible';
+      color = TAG_COLORS.INCOMPATIBLE;
+      break;
+    default:
+      name = 'missing_from_train';
+      color = TAG_COLORS.MISSING;
+  }
+
+  return { name, color };
+};
+
+export const formatSpeedLimitTags = (
+  rawSpeedLimitTags: SimulationResponseSuccess['mrsp'],
+  pathLength: number
+): LayerData<SpeedLimitTagValues>[] =>
+  rawSpeedLimitTags.values.reduce((mergedPositions, value, index) => {
+    const { name, color } = getTag(value.source);
+
+    if (name === 'missing_from_train') return mergedPositions;
+
+    const start = index === 0 ? 0 : rawSpeedLimitTags.boundaries[index - 1];
+    const end =
+      index === rawSpeedLimitTags.values.length - 1
+        ? pathLength
+        : rawSpeedLimitTags.boundaries[index];
+
+    const newPosition = {
+      position: {
+        start: mmToKm(start),
+        end: mmToKm(end),
+      },
+      value: {
+        tag: name,
+        color,
+      },
+    };
+
+    const lastPosition = mergedPositions.at(-1);
+    if (
+      lastPosition &&
+      lastPosition.value.tag === newPosition.value.tag &&
+      lastPosition.value.color === newPosition.value.color &&
+      lastPosition.position.end === newPosition.position.start
+    ) {
+      lastPosition.position.end = newPosition.position.end;
+    } else {
+      mergedPositions.push(newPosition);
+    }
+    return mergedPositions;
+  }, [] as LayerData<SpeedLimitTagValues>[]);
 
 export const formatSpeeds = (simulation: ReportTrain) => {
   const { positions, speeds } = simulation;
@@ -139,8 +206,10 @@ export const formatData = (
     selectedTrainPowerRestrictions;
   const electricalProfiles: LayerData<ElectricalPofilelValues>[] | undefined =
     formatElectricalProfiles(simulation.electrical_profiles, electrifications, pathLength);
-  const speedLimitTags: LayerData<SpeedLimitTagValues>[] | undefined = undefined;
-
+  const speedLimitTags: LayerData<SpeedLimitTagValues>[] | undefined = formatSpeedLimitTags(
+    simulation.mrsp,
+    pathLength
+  );
   return {
     speeds,
     ecoSpeeds,
--- a/front/src/modules/simulationResult/consts.ts
+++ b/front/src/modules/simulationResult/consts.ts
@@ -32,3 +32,10 @@ export const LIST_VALUES = {
 
 export type ListValues = ValueOf<typeof LIST_VALUES>;
 export type AllListValues = ArrayElement<ListValues>;
+
+export const TAG_COLORS = {
+  MISSING: '#94918E',
+  GIVEN_TRAIN: '#494641',
+  FALLBACK: '#216482',
+  INCOMPATIBLE: '#EAA72B',
+};
--- a/front/src/modules/simulationResult/types.ts
+++ b/front/src/modules/simulationResult/types.ts
@@ -1,5 +1,9 @@
-import type { PositionData } from 'applications/operationalStudies/types';
+import type {
+  PositionData,
+  SimulationResponseSuccess,
+} from 'applications/operationalStudies/types';
 import type { HeightPosition } from 'reducers/osrdsimulation/types';
+import type { ArrayElement } from 'utils/types';
 
 import type { MergedBlock } from './components/ChartHelpers/ChartHelpers';
 import type { ReportTrainData } from './components/SpeedSpaceChart/types';
@@ -23,3 +27,5 @@ export type SpaceCurvesSlopesDataV2 = {
   slopesCurve: HeightPosition[];
   curvesHistogram: PositionData<'radius'>[];
 };
+
+export type SpeedLimitTagValue = ArrayElement<SimulationResponseSuccess['mrsp']['values']>;


front/src/modules/simulationResult/components/SpeedSpaceChart/helpers.ts
```diff
@@ -14,10 +14,75 @@ import type {
...
+      return '#EAA72B'; // Red color for incompatible tags (alerts)
+    case 'UU':
+      return '#D91C1C';
+    default:
+      return '#94918E'; // Default color for missing or unknown tags
+  }
+};
+
+export const formatSpeedLimitTags = (
+  input: SpeedLimitTagInput,
```
clarani: You can use the type from editoast like this ;)
```suggestion
  input: SimulationResponseSuccess['mrsp'],
```

clarani: `input` is not a very good name, what about rawSpeedLimitTags ?

front/src/modules/simulationResult/components/SpeedSpaceChart/helpers.ts
```diff
@@ -14,10 +14,75 @@ import type {
...
+  pathLength: number
+): LayerData<SpeedLimitTagValues>[] => {
+  const positions: LayerData<SpeedLimitTagValues>[] = [];
+
+  for (let i = 0; i < input.values.length; i += 1) {
```
clarani: Use a forEach instead
```suggestion
  rawSpeedLimitTags.values.forEach((value, index) => {
```

front/src/modules/simulationResult/components/SpeedSpaceChart/helpers.ts
```diff
@@ -14,10 +14,75 @@ import type {
...
 
+type SpeedLimitTagInput = {
+  boundaries: number[];
+  values: {
+    speed: number;
+    source: { speed_limit_source_type: string; tag?: string } | null;
+  }[];
+};
+
+const getTagName = (source: { speed_limit_source_type: string; tag?: string } | null): string => {
```
clarani: - You can use the real type from editoast ;)
- then you will have to remove the extra `''` (in each return), check that source is not undefined and remove the impossible case

```suggestion
type SpeedLimitTagValue = ArrayElement<SimulationResponseSuccess['mrsp']['values']>;

const getTagName = (source: SpeedLimitTagValue['source']): string => {
```

front/src/modules/simulationResult/components/SpeedSpaceChart/helpers.ts
```diff
@@ -14,10 +14,75 @@ import type {
...
+    case 'unknown_tag':
+      return 'incompatible';
+    case 'UU':
+      return source.tag || '';
+    default:
+      return 'missing_from_train'; // Fallback for any unknown types
+  }
+};
+
+const getTagColor = (sourceType: string): string => {
```
clarani: - allow undefined param
- l79, you can remove the `| ''`, because now undefined is handled
```
const getTagColor = (sourceType?: string): string => {
```

front/src/modules/simulationResult/components/SpeedSpaceChart/helpers.ts
```diff
@@ -14,10 +14,75 @@ import type {
...
+  const positions: LayerData<SpeedLimitTagValues>[] = [];
+
+  for (let i = 0; i < input.values.length; i += 1) {
+    const start = i === 0 ? 0 : input.boundaries[i - 1];
+    const end = i === input.values.length - 1 ? pathLength : input.boundaries[i];
+
+    positions.push({
+      position: {
+        start: mToKm(start),
+        end: mToKm(end),
```
clarani: I think the unit is in mm no ? In which unit do you want to convert the positions ? km or m ?

front/src/modules/simulationResult/components/SpeedSpaceChart/helpers.ts
```diff
@@ -14,10 +14,75 @@ import type {
...
+      },
+      value: {
+        tag: getTagName(input.values[i].source), // This will be 'incompatible' for unknown_tag
+        color: getTagColor(input.values[i].source?.speed_limit_source_type || ''), // This should set the correct color
```
clarani: You can remove the comments on these 2 lines

clarani: I think you can merge the two functions getTagName and getTagColor => getTag which returns the name and the color, what do you think ?

front/src/modules/simulationResult/components/SpeedSpaceChart/helpers.ts
RomainValls: I think we need to convert mm into km here

front/src/modules/simulationResult/components/SpeedSpaceChart/helpers.ts
```diff
@@ -15,9 +15,71 @@ import type {
 } from 'applications/operationalStudies/types';
 import type { ReportTrainV2 } from 'common/api/osrdEditoastApi';
 import { mmToKm, msToKmh, mToKm } from 'utils/physics';
+import type { ArrayElement } from 'utils/types';
 
 import { electricalProfilesDesignValues } from './consts';
 
+type SpeedLimitTagValue = ArrayElement<SimulationResponseSuccess['mrsp']['values']>;
+const getTag = (source?: SpeedLimitTagValue['source']): { name: string; color: string } => {
```
SharglutDev: Maybe you can put this type in the types.ts file ?

front/src/modules/simulationResult/components/SpeedSpaceChart/helpers.ts
```diff
@@ -15,9 +15,71 @@ import type {
...
+  let name: string;
+  let color: string;
+
+  if (!source) {
+    name = 'missing_from_train';
+    color = '#94918E';
+  } else {
+    switch (source.speed_limit_source_type) {
+      case 'given_train_tag':
+        name = source.tag || 'missing_from_train';
```
SharglutDev: Not sure why we need the fallback here. Can `source.tag`equal an empty string ?

```suggestion
        name = source.tag || 'missing_from_train';
```

front/src/modules/simulationResult/components/SpeedSpaceChart/helpers.ts
```diff
@@ -15,9 +15,71 @@ import type {
...
 import { electricalProfilesDesignValues } from './consts';
 
+type SpeedLimitTagValue = ArrayElement<SimulationResponseSuccess['mrsp']['values']>;
+const getTag = (source?: SpeedLimitTagValue['source']): { name: string; color: string } => {
+  let name: string;
+  let color: string;
+
+  if (!source) {
+    name = 'missing_from_train';
+    color = '#94918E';
```
SharglutDev: Instead of "magin colors", do you think you could add an object inside consts.ts like this : 

```suggestion
const TAG_COLORS = {
  MISSING: '#94918E',
  ... (same for other tags)
}
```

front/src/modules/simulationResult/components/SpeedSpaceChart/helpers.ts
```diff
@@ -15,9 +15,71 @@ import type {
...
+        name = source.tag || 'missing_from_train';
+        color = '#216482';
+        break;
+      case 'unknown_tag':
+        name = 'incompatible';
+        color = '#EAA72B';
+        break;
+      default:
+        name = 'missing_from_train';
+        color = '#94918E';
```
SharglutDev: We use two times these lines (when no source and default case), maybe we could simplify like this (if there is no source it goes to default) : 

```suggestion
const getTag = (source?: SpeedLimitTagValue['source']): { name: string; color: string } => {
  let name: string;
  let color: string;

  switch (source?.speed_limit_source_type) {
    case 'given_train_tag':
      name = source.tag || 'missing_from_train';
      color = '#494641';
      break;
    case 'fallback_tag':
      name = source.tag || 'missing_from_train';
      color = '#216482';
      break;
    case 'unknown_tag':
      name = 'incompatible';
      color = '#EAA72B';
      break;
    default:
      name = 'missing_from_train';
      color = '#94918E';
  }

  return { name, color };
};
```

If you don't like the fact that we still check all of `source_types` if there is no `source`, we can do like this (I think I prefer this one) : 

```suggestion
const getTag = (source?: SpeedLimitTagValue['source']): { name: string; color: string } => {
  let name = 'missing_from_train';
  let color = '#94918E';

  if (source) {
    name = source.speed_limit_source_type === 'unknown_tag' ? 'incompatible' : source.tag;

    if (source.speed_limit_source_type === 'given_train_tag') {
      color = '#494641';
    } else if (source.speed_limit_source_type === 'fallback_tag') {
      color = '#216482';
    } else if (source.speed_limit_source_type === 'unknown_tag') {
      color = '#EAA72B';
    }
  }

  return { name, color };
};
```

SharglutDev (changes requested):
Thank you for the PR, I added some comments

front/src/modules/simulationResult/components/SpeedSpaceChart/helpers.ts
RomainValls: You're right, I think we can just remove the fallback and only keep the source.tag

front/src/modules/simulationResult/components/SpeedSpaceChart/helpers.ts
```diff
@@ -14,10 +14,83 @@ import type {
...
+  return { name, color };
+};
+
+export const formatSpeedLimitTags = (
+  rawSpeedLimitTags: SimulationResponseSuccess['mrsp'],
+  pathLength: number
+): LayerData<SpeedLimitTagValues>[] => {
+  const mergedPositions: LayerData<SpeedLimitTagValues>[] = [];
+
+  rawSpeedLimitTags.values.forEach((value, index) => {
```
woshilapin: How about using a `reduce` instead, or a real `for` loop? I find `.forEach` to have all the disadvantages: tries to be more 'functional programming' (so might be harder to read for those not used to it), but not fully embrace it.

```ts
return rawSpeedLimitTags.values.reduce([], (mergedPositions, value, index) => {
  /* same function */
});
```

or full `for` loop.
```ts
const mergedPositions: LayerData<SpeedLimitTagValues>[] = [];
for (const [value, index] of rawSpeedLimitTags.values.entries) {
  /* same function */
}
return mergedPositions;
```

RomainValls: I decided to use the reduce. The speedLimitTags that were blinking seem to be much better thank you 👍

woshilapin (approved):
Thanks.

front/src/modules/simulationResult/components/SpeedSpaceChart/helpers.ts
```diff
@@ -14,10 +14,78 @@ import type {
...
+          start: mmToKm(start),
+          end: mmToKm(end),
+        },
+        value: {
+          tag: name,
+          color,
+        },
+      };
+
+      const lastPosition = mergedPositions[mergedPositions.length - 1];
```
SharglutDev: I find it cleaner like that (and it forces us to check the undefined case, as you already did)

```suggestion
      const lastPosition = mergedPositions.at(-1);
```

front/src/modules/simulationResult/components/SpeedSpaceChart/helpers.ts
```diff
@@ -14,10 +14,78 @@ import type {
...
+        lastPosition.position.end === newPosition.position.start
+      ) {
+        lastPosition.position.end = newPosition.position.end;
+      } else {
+        mergedPositions.push(newPosition);
+      }
+
+      return mergedPositions;
+    }, [] as LayerData<SpeedLimitTagValues>[])
+    .filter((position) => position.value.tag !== 'missing_from_train');
```
SharglutDev: Instead of looping one more time in the array, maybe after l.60, you can put this : 

```
if (name === 'missing_from_train') return mergedPositions;
```
And maybe, we can start by that in the reduce before defined start and end.

Don't know how to test if it returns the same array at the end (how can we get "missing_from_train" ?)

SharglutDev (changes requested):

SharglutDev (approved):
Lgtm and tested, good job !

PR #8448, opened 2024-08-14
front: origin departure time not displayed even with stopping time

### LINKED ISSUE #8305: Departure time of the origin is never displayed, even after some inputs
### What happened?

The departure time isn't display after the train creation, which is normal, but:
- when entering a value in the departure time cell
- when adding a stop at the origin

the depature time is never displayed:
image

I'm expecting it to be displayed.

.

Recette (SNCF)

Google Chrome

0a34296

achrafmohye:
Closes #8305

### DIFF
--- a/front/src/modules/timesStops/helpers/utils.ts
+++ b/front/src/modules/timesStops/helpers/utils.ts
@@ -8,7 +8,7 @@ import type { SuggestedOP } from 'modules/trainschedule/components/ManageTrainSc
 import type { PathStep } from 'reducers/osrdconf/types';
 import { extractHHMMSS } from 'utils/date';
 import { NO_BREAK_SPACE } from 'utils/strings';
-import { secToHoursString, time2sec } from 'utils/timeManipulation';
+import { datetime2sec, secToHoursString, time2sec } from 'utils/timeManipulation';
 
 import { marginRegExValidation, MarginUnit } from '../consts';
 import { TableType } from '../types';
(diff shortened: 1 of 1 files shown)


theocrsb (approved):
LGTM and tested

front/src/modules/timesStops/TimesStops.tsx
```diff
@@ -97,7 +97,6 @@ const TimesStops = ({
         } else {
           rowData.isMarginValid = true;
           if (op.fromRowIndex === 0) {
-            rowData.arrival = null;
```
achrafmohye: Not exactly sure what this line does, if anyone have some clarifications

clarani (changes requested):
:x: Found several bugs:
- I can't save my train if I set up a stop duration for the first pathStep now
- if I update the time start of the train (in the field "heure de départ"), its departure time is not updated in the table

front/src/modules/timesStops/TimesStops.tsx
clarani: Editoast does not accept any arrival time for the first pathStep. We should handle this when we format the payload for the backend instead

achrafmohye: That's clearer, thanks

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -58,11 +58,14 @@ export const formatSuggestedViasToRowVias = (
     const isMarginValid = theoreticalMargin ? marginRegExValidation.test(theoreticalMargin) : true;
     const departure =
       stopFor && arrival ? secToHoursString(time2sec(arrival) + Number(stopFor), true) : undefined;
+    const scheduledDepartureTime = startTime
```
clarani: If I understand correctly, `scheduledDepartureTime` is used only for the departure.

I think it would be better to do smth like this: 
```
    let departure: string | undefined;
    if (stopFor) {
      if (i === 0) {
        departure = ....
      } else {
        departure = ...
      }
    }
```
This way, we have only 1 const `departure`

clarani (approved):

PR #8454, opened 2024-08-14
editoast: fix max_run_time, departure_time and max_departure_delay for stdcm v2 payload

Erashin:
departure_time is the train's earliest possible departure_time: if there is a tolerance before a step, it should be taken into account, and departure_time = arrival_time - tolerance_before.
max_run_time and max_departure_delay should be dynamic and take the tolerance into account. See examples below (1rst case is with a fixed arrival time, 2nd case is with a fixed departure time).

~EDIT: TBDetermined: commit completely remove fields from payload which are used in the debug (v1) stdcm interface: meeting this afternoon to see if we should keep on maintaining it~ => fields will be kept untill v1 debugging interface is removed and v2 interface supports a debugging mode with more inputs/outputs.

image

### DIFF
--- a/editoast/src/views/v2/timetable/stdcm.rs
+++ b/editoast/src/views/v2/timetable/stdcm.rs
@@ -83,10 +83,9 @@ pub struct STDCMRequestPayload {
     comfort: Comfort,
     /// By how long we can shift the departure time in milliseconds
     /// Deprecated, first step data should be used instead
-    #[serde(default = "default_maximum_departure_delay")]
-    #[schema(default = default_maximum_departure_delay)]
-    maximum_departure_delay: u64,
+    maximum_departure_delay: Option<u64>,
     /// Specifies how long the total run time can be in milliseconds
+    /// Deprecated, first step data should be used instead
     maximum_run_time: Option<u64>,
     /// Train categories for speed limits
     speed_limit_tags: Option<String>,
@@ -126,11 +125,6 @@ struct StepTimingData {
     arrival_time_tolerance_before: u64,
     /// The train may arrive up to this duration after the expected arrival time
     arrival_time_tolerance_after: u64,
-}
-
-const TWO_HOURS_IN_MILLISECONDS: u64 = 2 * 60 * 60 * 60;
-const fn default_maximum_departure_delay() -> u64 {
-    TWO_HOURS_IN_MILLISECONDS
 }
 
 #[derive(Debug, Default, Clone, Serialize, Deserialize, IntoParams, ToSchema)]
@@ -198,8 +192,8 @@ async fn stdcm(
     )
     .await?;
 
-    // 2. Compute the earliest start time and maximum running time
-    let maximum_run_time_result = get_maximum_run_time(
+    // 2. Compute the earliest start time, maximum running time and maximum departure delay
+    let simulation_run_time_result = get_simulation_run_time(
         db_pool.clone(),
         redis_client.clone(),
         core_client.clone(),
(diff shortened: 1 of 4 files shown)


shenriotpro:
It looks like departure_time is no longer in the response, but that's used in tests.

editoast/src/views/v2/timetable/stdcm.rs
```diff
@@ -83,10 +83,9 @@ pub struct STDCMRequestPayload {
     comfort: Comfort,
     /// By how long we can shift the departure time in milliseconds
     /// Deprecated, first step data should be used instead
-    #[serde(default = "default_maximum_departure_delay")]
-    #[schema(default = default_maximum_departure_delay)]
-    maximum_departure_delay: u64,
+    maximum_departure_delay: Option<u64>,
     /// Specifies how long the total run time can be in milliseconds
+    /// Deprecated, first step data should be used instead
```
leovalais, line 88: Unless I'm wrong, it's not really deprecated but reserved to developers for debugging situations (ie. `request_user.role == Role::Dev` when we'll have role checking 😁).

editoast/src/views/v2/timetable/stdcm.rs
```diff
@@ -209,18 +203,26 @@ async fn stdcm(
...
     };
+    let earliest_step_tolerance_window = get_earliest_step_tolerance_window(&data);
+    let maximum_departure_delay =
+        get_maximum_departure_delay(&data, simulation_run_time, earliest_step_tolerance_window);
+    let maximum_run_time_without_tolerance = 2 * simulation_run_time + get_total_stop_time(&data);
+    let maximum_run_time = get_maximum_run_time(
+        &data,
+        maximum_run_time_without_tolerance,
+        earliest_step_tolerance_window,
+    );
```
leovalais, line 222: There's quite a lot going on between the parsing of the request, the default values and the debug overrides. I'd suggest to wrap all these values in a dedicated structure like:

```rust
struct StdcmTimings {
    maximum_run_time: u64,
    maximum_run_time_without_tolerance: u64,
    ...
}
```

and `let timings = StdcmTimings::from(&data)`. This would:

* gather all these operations in one place, shortening the endpoint code
* make explicit the values that are computed / overridden
* create a place to document the whole process with your schema
* allow us to use `StdcmTimings::has_debug_overrides(&data)` in order to check whether we need to ensure the `StdcmDebug` user role

Of course since this PR is rather urgent this can be left ~~as an exercise for the reader~~ for another PR (which would be another core-related good first issue, maybe for Loup?).

leovalais (approved):
LGTM. I can't really test it rn since `osrdyne` kinda broke the dev setup for Macs 😭. I'm working on it, but you probably want this PR merged before that... 😬

editoast/src/views/v2/timetable/stdcm.rs
Erashin: This should actually be removed at the same time as `start_time` and `maximum_departure_delay` parameters, following the removal of the v1 debugging interface and the update of the v2 interface to allow a debugging mode with more inputs/outputs. Hence the `Deprecated`. I'd rather keep it likeso if it's all the same with you :)

editoast/src/views/v2/timetable/stdcm.rs
Erashin: I'll see if i have the time to fit it in, if not, I'll create a corresponding issue :)

editoast/src/views/v2/timetable/stdcm.rs
leovalais: Fine by me, resolving :)

editoast/src/views/v2/timetable/stdcm.rs
Erashin: It is a good first issue, and I'd rather do it while not hard pressed with time. Issue created: https://github.com/OpenRailAssociation/osrd/issues/8507.

PR #8457, opened 2024-08-14
front: fix train schedule input table

### LINKED ISSUE #8377: Front crash when edit input table for a specific scenario
### What happened?

I get this error when edit the table : 

https://github.com/user-attachments/assets/46ed2768-db28-4e5d-b952-c0545d881bda

The front don't crash

1. Import the [train_schedules.json](https://github.com/user-attachments/files/16549370/train_schedules.json)
2. Edit a train
3. Set any value in Times and stops table 
4. Save the modification 

Recette (SNCF)

Google Chrome

02781ca

### LINKED ISSUE #8455: input table edition missing lines
### What happened?

when creating a train with several steps (here output table)

Screenshot_20240814_160701

going back to input table to edit they are gone

Screenshot_20240814_160737

git bisect seems to point to commit db3ce245158b055f93e30563c784195151c5e2de

anisometropie:
when the list of suggested Operational points contained 2 points with the same uic but different ch, the function was not checking for ch and thus was choosing the wrong op.

because of this the input table was displaying the data of some lines on wrong lines

closes #8455 

#8377 thought to be related, but not fixed by this PR

### DIFF
--- a/front/src/applications/operationalStudies/hooks/__tests__/useSetupItineraryForTrainUpdate.spec.ts
+++ b/front/src/applications/operationalStudies/hooks/__tests__/useSetupItineraryForTrainUpdate.spec.ts
@@ -1,0 +1,294 @@
+import type { PathfindingResult } from 'common/api/osrdEditoastApi';
+
+import { updatePathStepsFromOperationalPoints } from '../useSetupItineraryForTrainUpdate';
+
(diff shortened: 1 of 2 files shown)


theocrsb (approved):
LGTM and tested

kmer2016 (approved):
LGTM and tested !

Issue #8446, opened 2024-08-14
Add roles check to endpoints


hamz2a:
### Description and goal

views:
- [x]  `documents.rs` #8425
- [x]  `electrical_profiles.rs` #8425
- [x]  `layers.rs` #8479
- [x]  `light_rolling_stocks.rs` #8362
-  ~~`mod.rs`~~
- [x]  `projects.rs` #8401
- ~~`scenario.rs`~~ **(tsv1)**
- [x]  `search.rs` #8469
-  ~~`single_simulation.rs`~~ **(tsv1)**
- [x]  `speed_limit_tags.rs` #8468
- [x]  `sprites.rs` #8510
- [x]  `study.rs` #8401
- ~`timetable.rs`~ **(tsv1)**
- [x]  `work_schedules.rs` #8466
- [x] `stdcm_search_environment.rs` https://github.com/OpenRailAssociation/osrd/pull/8641

views/infra:
- [x]  `attached.rs` #8361
- [x]  `edition.rs` #8361
- [x]  `errors.rs` #8361
- [x]  `lines.rs` #8361
- [x]  `mod.rs` #8345 
- [x]  `objects.rs` #8361
- [x]  `pathfinding.rs` #8361
- [x]  `railjson.rs` #8361
- [x]  `routes.rs` #8361
- [x]  `auto_fixes/mod.rs` #8361

views/v2:
- [x]  `scenario.rs` #8401
- [x]  `timetable.rs` #8445
- [x]  `train_schedule.rs` #8436

views/v2/path:
- [x]  `pathfinding.rs` #8449

views/v2/timetable:
- [x]  `stdcm.rs` #8424

views/v2/train_schedule:
- [x]  `projection.rs` #8447

views/rolling_stocks:
- [x]  `mod.rs` #8362

---
~~views/pathfinding~~: **(tsv1)**
- `electrical_profiles.rs`
- `electrifications.rs`
- `mod.rs`

~~views/stdcm~~:
- `mod.rs` **(tsv1)**

~~views/timetable:~~ **(tsv1)**
- `import.rs`

~~views/train_schedule:~~ **(tsv1)**
- `mod.rs`
- `projection.rs`
- `simulation_report.rs`

### Acceptance criteria

add roles check to all endpoints

PR #8468, opened 2024-08-16
editoast: add roles check to `views::speed_limit_tags`

hamz2a:
Part of https://github.com/OpenRailAssociation/osrd/issues/8446

### DIFF
--- a/editoast/src/views/speed_limit_tags.rs
+++ b/editoast/src/views/speed_limit_tags.rs
@@ -1,8 +1,12 @@
 use axum::extract::State;
+use axum::Extension;
 use axum::Json;
+use editoast_authz::BuiltinRole;
 
 use crate::error::Result;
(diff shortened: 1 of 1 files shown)


editoast/editoast_authz/src/builtin_role.rs
```diff
@@ -12,6 +12,9 @@ pub enum BuiltinRole {
     InfraRead,
     #[strum(serialize = "infra:write")]
     InfraWrite,
+
+    #[strum(serialize = "speed_limit_tags:read")]
+    SpeedLimitTagsRead,
```
leovalais: Use RollingStockCollection* roles, it does't make much sense to work with speed limit tags while being denied access to rolling stocks.

leovalais (changes requested):

leovalais (approved):

PR #8469, opened 2024-08-16
editoast: add roles check to `views::search`

hamz2a:
Part of #8446

### DIFF
--- a/editoast/src/views/search.rs
+++ b/editoast/src/views/search.rs
@@ -203,6 +203,7 @@ use axum::extract::State;
 use axum::http::header::CONTENT_TYPE;
 use axum::http::StatusCode;
 use axum::response::IntoResponse;
+use axum::Extension;
 use chrono::NaiveDateTime;
 use diesel::pg::Pg;
 use diesel::sql_query;
@@ -210,6 +211,7 @@ use diesel::sql_types::Jsonb;
 use diesel::sql_types::Text;
 use diesel::QueryableByName;
 use diesel_async::RunQueryDsl;
+use editoast_authz::BuiltinRole;
 use editoast_common::geometry::GeoJsonPoint;
 use editoast_derive::EditoastError;
 use editoast_derive::Search;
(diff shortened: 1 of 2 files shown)


editoast/src/views/search.rs
```diff
@@ -333,10 +337,19 @@ struct SearchDBResult {
...
     Json(SearchPayload { object, query, dry }): Json<SearchPayload>,
 ) -> Result<impl IntoResponse> {
+    let authorized = authorizer
+        .check_roles([BuiltinRole::InfraRead].into())
+        .await
+        .map_err(AuthorizationError::AuthError)?;
+    if !authorized {
+        return Err(AuthorizationError::Unauthorized.into());
+    }
+
```
leovalais: It's more complicated than that, the search engine is dynamic and isn't restricted to infra object search. Use `object` to determine the correct roles to check.

leovalais (changes requested):

editoast/src/views/search.rs
```diff
@@ -333,10 +338,31 @@ struct SearchDBResult {
...
+    match object.as_str() {
+        "track" | "operationalpoint" | "signal" => roles_to_check.insert(BuiltinRole::InfraRead),
+        "project" | "study" | "scenario" => roles_to_check.insert(BuiltinRole::OpsRead),
+        &_ => {
+            return Err(SearchApiError::ObjectType {
+                object_type: object.to_owned(),
+            }
+            .into())
+        }
+    };
```
leovalais: I think you can write directly:

```suggestion
    let roles: &[BuiltinRole] = match object.as_str() {
        "track" | "operationalpoint" | "signal" => &[BuiltinRole::InfraRead],
        "project" | "study" | "scenario" => &[BuiltinRole::OpsRead],
        _ => {
            return Err(SearchApiError::ObjectType {
                object_type: object.to_owned(),
            }
            .into())
        }
    };
```

leovalais (approved):
A minor change to avoid mutability, otherwise lgtm

editoast/src/views/search.rs
hamz2a: as discussed, we will use compile-time defined strings to check roles based on the `object` field in the `SearchPayload`.

PR #8470, opened 2024-08-16
osrd-schemas: tidy type hints

Castavo:
I removed NewType ([as documented by pydantic](https://docs.pydantic.dev/2.6/concepts/types/#named-type-aliases)).
This allows instantiating the classes with `id="example"` without Pylance yielding an error. 
I could have used a `TypeAliasType` but then that would have changed the json schema: it would no longer inline the constraints. This could be desirable, in another PR.

I corrected/added type hints, to fix all errors Pylance would otherwise yield.

I removed the direct calls to `Identifier` because they were unnecessary for constraint checking (and caused Pylance to yield an error).


shenriotpro (approved):
Thanks

leovalais (approved):

PR #8471, opened 2024-08-16
misc: ignore .vscode in root .gitignore

Castavo:
I offer that we move all of those in once place. What do you think ?
This also ignores .vscode folders created in python/ subfolders by the way

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .DS_Store
 *.backup
 __pycache__
+.vscode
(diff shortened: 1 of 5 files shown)


ghost (approved):

Khoyo (changes requested):
You missed `core/.gitignore`

Castavo:
@Khoyo thanks, my bad
(Who uses vscode with core ?)

Khoyo (approved):

clarani (approved):

leovalais (approved):

PR #8476, opened 2024-08-19
docker: use dev profile instead of release for docker-compose

emersion:
docker-compose is used for development. Use the "dev" profile instead of "release" to speed up builds.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -102,6 +102,8 @@ services:
       additional_contexts:
         test_data: tests/data
         static_assets: assets
+      args:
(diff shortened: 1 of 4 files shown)


emersion:
Err, it sounds like `cargo install` re-builds in `release` mode…

emersion:
There, fixed.

younesschrifi (approved):
LGTM ! 👍🏽

Khoyo (approved):

leovalais (approved):
Thanks! (not tested)

PR #8478, opened 2024-08-19
front: fix localStorage.setItem() in macro node store

### LINKED ISSUE #8477: NGE can't load if no entry is in nodeStore
### What happened?

If `nodeStore from `localStorage` is empty, NGE fails to load the microscopic timetable, and load the default netzgrafik from NGE instead.

This issue is critical because NGE can't be use in this case

image

Initialize `nodeStore` if it's empty

1. Clear `localStorage`
2. Open Macroscopic

Local

Firefox

-

emersion:
JSON.stringify() will always return "{}" for a Map. Convert the Map to an array of entries before encoding these to JSON.

Closes: https://github.com/OpenRailAssociation/osrd/issues/8477

### DIFF
--- a/front/src/applications/operationalStudies/components/MacroEditor/nodeStore.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/nodeStore.ts
@@ -22,7 +22,8 @@ const loadNodes = () => {
 };
 
 const saveNodes = (nodes: Map<string, Node>) => {
-  localStorage.setItem(NODES_LOCAL_STORAGE_KEY, JSON.stringify(nodes));
+  const entries = [...nodes.entries()];
+  localStorage.setItem(NODES_LOCAL_STORAGE_KEY, JSON.stringify(entries));
 };
 
 const getNodeKey = (timetableId: number, trigram: string) => `${timetableId}:${trigram}`;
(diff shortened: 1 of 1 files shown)


front/src/applications/operationalStudies/components/MacroEditor/nodeStore.ts
```diff
@@ -14,7 +14,10 @@ const loadNodes = () => {
   let entries = [];
   const rawNodes = localStorage.getItem(NODES_LOCAL_STORAGE_KEY);
   if (rawNodes) {
-    entries = JSON.parse(rawNodes);
+    const parsedNodes = JSON.parse(rawNodes);
+    if (Array.isArray(parsedNodes)) {
+      entries = parsedNodes;
+    }
```
emersion, line 24: Can we log an error if it ends up not being an array? Just to let the dev know that we didn't actually load anything from the localStorage.

emersion:
Your changes LGTM. Can you squash your two commits together?

theocrsb:
Closes #8477

theocrsb (approved):
LGTM and tested

Uriel-Sautron (approved):
LGTM!! Tested

clarani (approved):
LGTM :white_check_mark: not tested

PR #8480, opened 2024-08-19
docker: add override for Docker socket path

emersion:
Some systems (e.g. podman) have a non-standard socket path. Allow developers to override the default via a .env file.

I've tried using compose.override.yml but it doesn't seem to work.


younesschrifi (approved):
LGTM :)

PR #8483, opened 2024-08-19
core: stdcm: step availability: avoid n² complexity

eckter:
The whole function has a diff because of indentation change (github can be config to ignore it)

We now:

1. Iterate backwards
2. Break when going too far back in the path


core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
```diff
@@ -158,55 +158,67 @@ data class BlockAvailability(
...
-                        // Planned timing data respected
-                        return AvailabilityProperties(
-                            0.0,
-                            Offset(0.meters),
-                            plannedMaxTimeAtStep - timeAtStep,
-                            plannedMaxTimeAtStep
-                        )
-                    }
+        // Iterate over all the predecessor blocks + current block
+        // Iterate backwards, as the range is generally toward the end
```
Erashin:
```suggestion
        // Iterate backwards, as the range is generally towards the end
```

Erashin (approved):
Thanks for that!

PR #8484, opened 2024-08-19
front: update display in rolling stock editor

### LINKED ISSUE #8370: Small bugs on '*' for mandatory fields when creating a rolling stock
### What happened?

When I create a rolling stock, I have the two following behaviours : 
image
- In black : the speed-effort curves are mandatory but there is no '*'
- In red : when the name of the field is too large, the '*' goes on the next line

I would expect : 
- Black : to have a '*'
- Red : to have the '*' on the same line

as above

Recette (SNCF)

Google Chrome

theocrsb:
close #8370


kmer2016 (approved):
LGTM and tested !

RomainValls (approved):
LGTM and Tested ✅

PR #8486, opened 2024-08-19
front: fix requested arrival time not displayed in times table

emersion:
To reproduce, create a train and set a scheduled time for the final destination.

### DIFF
--- a/front/src/modules/timesStops/hooks/useOutputTableData.ts
+++ b/front/src/modules/timesStops/hooks/useOutputTableData.ts
@@ -79,18 +79,20 @@ function useOutputTableData(
         );
         const pathStepKey = `${sugOpPoint.uic}-${sugOpPoint.ch}`;
 
-        if (pathStepKey in pathStepsByUic && nextOpPoint) {
+        if (pathStepKey in pathStepsByUic) {
           const pathStepId = pathStepsByUic[pathStepKey].id || '';
           const schedule = scheduleByAt[pathStepId];
           const scheduleData = computeScheduleData(schedule, selectedTrainSchedule.start_time);
           const formattedScheduleData = formatScheduleData(scheduleData);
-          const marginsData = computeMargins(
-            simulatedTrain,
-            opPoint,
-            nextOpPoint,
-            selectedTrainSchedule,
-            pathStepId
-          );
+          const marginsData = nextOpPoint
+            ? computeMargins(
+                simulatedTrain,
+                opPoint,
+                nextOpPoint,
+                selectedTrainSchedule,
+                pathStepId
+              )
+            : null;
           const calculatedArrival = checkAndFormatCalculatedArrival(scheduleData, opPoint.time);
 
           return {


front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -79,18 +79,20 @@ function useOutputTableData(
...
           const scheduleData = computeScheduleData(schedule, selectedTrainSchedule.start_time);
           const formattedScheduleData = formatScheduleData(scheduleData);
-          const marginsData = computeMargins(
-            simulatedTrain,
-            opPoint,
-            nextOpPoint,
-            selectedTrainSchedule,
-            pathStepId
-          );
+          const marginsData = nextOpPoint
```
anisometropie, line 87: Can you do that inside the computeMargins function and make a new unit test for this new case ?

anisometropie (dismissed):

front/src/modules/timesStops/hooks/useOutputTableData.ts
emersion: Hm, does it really make sense? `computeMargins()` is non-sensical without a `nextOpPoint` (would return immediately with an empty object), or is it?

anisometropie: I think it would make more sense to let the function handle that case when there is no `nextOpPoint` rather than having a ternary at the call site here

emersion: Can you expand why? To me, it's the contrary since the function can't do anything without a `nextOpPoint` :P

anisometropie: well :P
- mostly because it’s impossible to test here in this hook, and it’s really easy to test the function directly.
- In the way I view function design, as a general principle, it would be the function’s responsibility to handle the case where their input is undefined: It looks a lot messier to have this 
`nextOpPoint ? myFunction(nextOpPoint) : null;` rather than hide it inside `myFunction(nextOpPoint)`.

I know  the function is called `computeMargins`, but it’s kinda implied it « computes margins if possible, or return null if it can’t »

emersion: > mostly because it’s impossible to test here in this hook, and it’s really easy to test the function directly.

I don't see any substantial value in testing that a function returns `null` when one of its parameters is `null`. In general I'm not in love with trivial unit tests (these tend to be costly on the long term maintenance-wise and won't catch any real bug/regression). It would be worthwhile to test `useOutputTableData` instead, but that's a much bigger endeavor.

> In the way I view function design, as a general principle, it would be the function’s responsibility to handle the case where their input is undefined: It looks a lot messier to have this
nextOpPoint ? myFunction(nextOpPoint) : null; rather than hide it inside myFunction(nextOpPoint).

That sounds backwards to me, especially since the null check is trivial.

I don't feel like I've anything else to add to the discussion, I'll leave it up to someone else to weight in.

kmer2016: I agree more with @emersion  approach. The `null` check is really trivial. It makes more sense to test the whole hook, but we can discuss that later.

kmer2016 (approved):
LGTM and tested

PR #8487, opened 2024-08-19
front: don't debounce NGE node events

emersion:
The current approach doesn't work when editing two nodes in succession: edit node A and before 2s elapse edit node B. The debounce will only invoke the callback for node B, dropping changes for node A. Moreover, switching to the micro tab will not immediately reflect changes.

Instead, [NGE has been patched][1] to only trigger the node update event once the input field is unfocused.

[1]: https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/pull/248

### DIFF
--- a/front/package.json
+++ b/front/package.json
@@ -83,7 +83,7 @@
     "lodash": "^4.17.21",
     "maplibre-gl": "^4.0.0",
     "match-sorter": "^6.3.3",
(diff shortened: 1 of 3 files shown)


emersion:
Depends on https://github.com/OpenRailAssociation/osrd/pull/8478 for proper testing.

louisgreiner (approved):
Tested, thanks for the fixes from both sides :+1:

Uriel-Sautron (approved):
LGTM!! Tested

kmer2016 (approved):
LGTM

PR #8495, opened 2024-08-20
front: fix train run section ordering in NGE to OSRD converter

### LINKED ISSUE #8392: NGE bug : trainrunSections are not always interpreted in the right order by OSRD
### What happened?

When one adds a trainrunSection at the beginning of a trainrun in NGE, the order is not well interpreted and the new `path_item` in OSRD is not added at the beginning

One expects the order in NGE to be respected in OSRD

Create a trainrun in NGE, and extend it from the beginning of the section

Dev (SNCF)

Google Chrome

3e8f772

emersion:
To reproduce:

- Open a scenario in macro mode with 3 nodes A, B and C (A      B      C)
- Create a new train run section starting at B and ending at C (A       B ---> C)
- Extend the train run by adding a new section starting at A and ending at B (A ---> B ---- C)

The old logic would interpret the sections in the wrong order. Walk the graph to re-order the sections.

Closes: https://github.com/OpenRailAssociation/osrd/issues/8392

### DIFF
--- a/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
@@ -10,11 +10,82 @@ import type { NetzgrafikDto, NGEEvent, TrainrunSection, Node, Trainrun } from '.
 
 const createdTrainrun = new Map<number, number>();
 
-const getTrainrunSectionsByTrainrunId = (trainrunSections: TrainrunSection[], trainrunId: number) =>
-  trainrunSections.filter((section) => section.trainrunId === trainrunId);
-
 const getNodeById = (nodes: Node[], nodeId: number | string) =>
   nodes.find((node) => node.id === nodeId);
+
+const getTrainrunSectionsByTrainrunId = (netzgrafikDto: NetzgrafikDto, trainrunId: number) => {
+  // The sections we obtain here may be out-of-order. For instance, for a path
+  // A → B → C, we may get two sections B → C and then A → B. We need to
+  // re-order the section A → B before B → C.
+  const sections = netzgrafikDto.trainrunSections.filter(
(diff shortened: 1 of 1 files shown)


front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
```diff
@@ -8,12 +8,71 @@ import type { NetzgrafikDto, NGEEvent, TrainrunSection, Node, Trainrun } from '.
...
+  // A → B → C, we may get two sections B → C and then A → B. We need to
+  // re-order the section A → B before B → C.
+  const sections = netzgrafikDto.trainrunSections.filter(
+    (section) => section.trainrunId === trainrunId
+  );
+
+  // Sections are linked together with transitions and ports. A section in the
+  // middle of a path will have a dedicated target port connected to a
+  // transition on one side, and on the other side connected to a source port
+  // dedicated to the next section. Build a map of sections keyed by their
```
clarani: I'm not sure, but I think the description would be clearer like this 
```suggestion
  // transition on one side, and on the other side connected to a source port
  // dedicated to the transition in the next node. Build a map of sections keyed by their
```

clarani (approved):

SarahBellaha (approved):
LGTM :)

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
emersion: Hm, I don't think that's quite correct: I'm only looking at a single node here and trying to explain that it goes like section → port → transition → port → section.

I've tried to make the paragraph clearer by including a small diagram - let me know if it's better or if you have ideas to improve it!

clarani (approved):

clarani (approved):
Thanks, it's way clearer with the diagram !

PR #8500, opened 2024-08-20
editoast: increase default message queue timeout

Castavo:
With the default value, importing a reasonably-sized timetable (400 trains) results in a timeout for the last trains.

Increasing the default value would allow running our benchmark directly from freshly-pulled images.

### DIFF
--- a/editoast/src/client/mod.rs
+++ b/editoast/src/client/mod.rs
@@ -141,8 +141,8 @@ pub struct RunserverArgs {
     #[derivative(Default(value = r#""amqp://osrd-rabbitmq:5672/%2f".into()"#))]
     #[clap(long, env = "OSRD_MQ_URL", default_value_t = String::from("amqp://osrd-rabbitmq:5672/%2f"))]
     pub mq_url: String,
-    #[derivative(Default(value = "60"))] // TODO: find the currently used timeout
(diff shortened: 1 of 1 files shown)


Khoyo (approved):

editoast/src/client/mod.rs
```diff
@@ -141,8 +141,8 @@ pub struct RunserverArgs {
     #[derivative(Default(value = r#""amqp://osrd-rabbitmq:5672/%2f".into()"#))]
     #[clap(long, env = "OSRD_MQ_URL", default_value_t = String::from("amqp://osrd-rabbitmq:5672/%2f"))]
     pub mq_url: String,
-    #[derivative(Default(value = "60"))] // TODO: find the currently used timeout
-    #[clap(long, env = "EDITOAST_CORE_TIMEOUT", default_value_t = 60)]
+    #[derivative(Default(value = "180"))] // TODO: find the currently used timeout
```
leovalais: Is that comment still relevant?

leovalais (approved):

editoast/src/client/mod.rs
Castavo: I don't know, it was added by @Khoyo 
Since he didn't say anything here, maybe it's good ?

Khoyo: We did find out and adjusted the prod value, you can remove it.

PR #8505, opened 2024-08-20
front: removed rocket component after pathfinding search

### LINKED ISSUE #8440: Clear rocket input after looking for a path with it

End users

This component could be removed after finding a path. It's taking quite a lot of space.
image

It would make the map much more visible, especially on small screens

_No response_

Caracol3:
close #8440

### DIFF
--- a/front/src/modules/pathfinding/components/Pathfinding/TypeAndPathV2.tsx
+++ b/front/src/modules/pathfinding/components/Pathfinding/TypeAndPathV2.tsx
@@ -54,8 +54,11 @@ function OpTooltips({ opList }: { opList: SearchResultItemOperationalPoint[] }) 
     </div>
   );
 }
+type typeAndPathV2Props = {
+  setDisplayTypeAndPath: React.Dispatch<React.SetStateAction<boolean>>;
+};
 
-const TypeAndPathV2 = () => {
+const TypeAndPathV2 = ({ setDisplayTypeAndPath }: typeAndPathV2Props) => {
   const dispatch = useAppDispatch();
   const [inputText, setInputText] = useState('');
   const [opList, setOpList] = useState<SearchResultItemOperationalPoint[]>([]);
@@ -168,6 +171,7 @@ const TypeAndPathV2 = () => {
         }));
 
       dispatch(updatePathSteps({ pathSteps }));
+      setDisplayTypeAndPath(false);
     }
   };
 
@@ -233,7 +237,7 @@ const TypeAndPathV2 = () => {
     <>
       <div
         className="type-and-path mb-2"
-        style={{ minWidth: `${monospaceOneCharREMWidth * inputText.length + 5.5}rem` }} // To grow input field & whole div along text size
+        style={{ minWidth: `${monospaceOneCharREMWidth * inputText.length + 5.5}rem` }}
         data-testid="type-and-path-container"
       >
         <div className="help">{opList.length === 0 && tManageTrainSchedule('inputOPTrigrams')}</div>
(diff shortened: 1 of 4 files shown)


theocrsb (commented):
I think you can add a check in the parent `ItineraryV2.tsx` to display or not `TypeAndPathV2` instead of creating another variable. What do you think?

```
      {displayTypeAndPath && (!pathProperties || !origin || !destination) && (
        <div className="mb-2">
          <TypeAndPathV2 />
        </div>
      )}
```

kmer2016:
@theocrsb 
> I think you can add a check in the parent `ItineraryV2.tsx` to display or not `TypeAndPathV2` instead of creating another variable. What do you think?

The problem with your solution is that if the user clicks on the rocket button after selecting an itinerary, the module can't be opened again.

front/src/modules/pathfinding/components/Pathfinding/TypeAndPathV2.tsx
```diff
@@ -224,75 +226,79 @@ const TypeAndPathV2 = () => {
   }, []);
```
kmer2016: It can be more readable to make early check, insteadd wrap all the component return with `isVisible`. What do you think ?
```suggestion
if (!isVisible) {
    return null;
  }

```

kmer2016 (changes requested):

front/src/modules/pathfinding/components/Pathfinding/TypeAndPathV2.tsx
Caracol3: Indeed, the effect is the same and it's easier to read.
Thank's ;-)

clarani (changes requested):
In the parent component, you already have a state const which allows you to display the component or not (`displayTypeAndPath`).
You should use this one instead of creating a new one ;) (you can give `setDisplayTypeAndPath` as a props in TypeAndPathV2 in `ItineraryV2` and use it instead of `setIsVisible`)

Caracol3:
> In the parent component, you already have a state const which allows you to display the component or not (`displayTypeAndPath`). You should use this one instead of creating a new one ;) (you can give `setDisplayTypeAndPath` as a props in TypeAndPathV2 in `ItineraryV2` and use it instead of `setIsVisible`)

I missed that, thanks. It's done :-)

node_modules/.yarn-integrity
clarani, line 1: You should not commit this file :)

clarani (commented):

Can you remove the file in node_modules ?

theocrsb (approved):
LGTM and tested ! (same for the file in node_modules)

node_modules/.yarn-integrity
Caracol3: done

clarani (approved):
LGTM :white_check_mark:

kmer2016 (approved):
LGTM

front/tests/pages/operational-studies-page-model.ts
```diff
@@ -231,15 +231,16 @@ class OperationalStudiesPage {
...
-    );
-    expect(await this.destinationInfo.innerText()).toEqual(
-      await this.getDestinationLocatorByTrigram(destinationTrigram).innerText()
-    );
+    // expect(await this.originInfo.innerText()).toEqual(
+    //   await this.getOriginLocatorByTrigram(originTrigram).innerText()
+    // );
+    // expect(await this.destinationInfo.innerText()).toEqual(
+    //   await this.getDestinationLocatorByTrigram(destinationTrigram).innerText()
+    // );
```
theocrsb: This commented code should be deleted

Maymanaf (approved):
LGTM for e2e tests

PR #8514, opened 2024-08-21
editoast: mark documents as cacheable

emersion:
Without this, rolling stock images are re-fetched each time they are displayed.

### DIFF
--- a/editoast/src/views/documents.rs
+++ b/editoast/src/views/documents.rs
@@ -1,7 +1,7 @@
 use axum::body::Bytes;
 use axum::extract::Path;
 use axum::extract::State;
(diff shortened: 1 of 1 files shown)


leovalais (approved):
lgtm, good idea! :)

emersion:
Trivial rebase and `cargo fmt` fix.

clarani (approved):
Thanks :pray:

How can I check that it works correctly in the front ? It seems that it still sends a request for each train card

emersion:
In the Network tab of the Firefox devtools, the "transferred" column should be set to "cached" the second time a document is fetched:

out

clarani:
Nice !! Thanks :)

PR #8515, opened 2024-08-21
front: stdcm result table, round times to the closest minute

### LINKED ISSUE #8502: front: stdcm: times should be rounded
In this screenshot, it *seems* like we haven't matched the requested time (9:59 instead of 10:00). But we're actually a less than a second away from 10:00. Rounding the values should clear that up. 

I don't know by how much we should round though. The simulation can be "wrong" by up to a timestep (2s) but that feels magic-number-y. More than a few seconds could be misleading. 

---

image

anisometropie:
closes https://github.com/OpenRailAssociation/osrd/issues/8502

All the times in the table were just floored down to the minute. We need round to the closest minute

### DIFF
--- a/front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
+++ b/front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
@@ -83,7 +83,7 @@ function getTimeAtPosition(
 ): string {
   const index = trainPositions.findIndex((pos) => pos >= trainPosition);
   const timeInMillis = trainTimes[index];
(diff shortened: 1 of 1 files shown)


eckter (approved):
Tested, it works :+1: 

Eventually we should check how much rounding we actually want (+-30s may be a bit much). But this at least avoids misleading outputs.

emersion (approved):

clarani (approved):
LGTM :white_check_mark: not tested

PR #8516, opened 2024-08-21
front: fix height of rollingstock editor spreadsheet for matching wit…

### LINKED ISSUE #7315: Unnecessary vertical scroll in speed-effort curves tab
### What happened?

When editing a rolling stock curve with at least 15 rows, an unwanted vertical scroll appears.

image

 

It should not scroll vertically

1. Go to rolling stock editor
2. Edit a train
3. Go to speed-effort curves tab
4. Have at least 15 rows

You can observe that the `Confirm` button is hidden (you have to scroll)

- **Browser**: Firefox v125
- **OS**: Nixos v23.11
- **Env**: DgexSol

11da135

theocrsb:
closes #7315

### DIFF
--- a/front/src/modules/rollingStock/components/RollingStockEditor/CurveSpreadsheet.tsx
+++ b/front/src/modules/rollingStock/components/RollingStockEditor/CurveSpreadsheet.tsx
@@ -126,6 +126,7 @@ const CurveSpreadsheet = ({
         addRowsComponent={false}
         onBlur={handleBlur}
         onSelectionChange={handleBlur}
+        height={332}
       />
     </div>
   );
--- a/front/src/styles/scss/common/components/_tabs.scss
+++ b/front/src/styles/scss/common/components/_tabs.scss
@@ -1,35 +1,45 @@
 .tabs-container {
   display: flex;
   flex-direction: column;
+  overflow: auto;
+
   &.full-width {
     width: 100%;
+
     .tabs.pills {
       overflow: auto;
       justify-content: space-between;
       min-height: 2.35rem;
+
       .tab {
         flex-grow: 1;
       }
     }
   }
+
   &.full-height {
     height: 100%;
+
     .tab-content {
       height: 100%;
+
       .tab-pane {
         height: 100%;
       }
     }
   }
+
   .tabs {
     display: flex;
     margin-bottom: 0.5rem;
     gap: 1rem;
+
     .tab {
       padding-bottom: 0.5rem;
       position: relative;
       border-bottom: 3px solid transparent;
       transition: text-shadow 0.3s;
+
       &::after {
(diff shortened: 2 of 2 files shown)


Akctarus (approved):
Tested, works well !

kmer2016:
You are solution works with 15 rows. However, if we have more rows, like 35, a scroll bar appears. From what I understand about the [issue](https://github.com/OpenRailAssociation/osrd/issues/7315) the main problem is that the confirm button is hidden when there are many rows in the grid. So, I believe a better solution would be to move the container of the confirm/cancel buttons to a non-scrollable area.

kmer2016 (changes requested):
You are solution works with 15 rows. However, if we have more rows, like 35, a scroll bar appears. Generally with overflow problem, add the height/width don't solve the problem, it deplaceFrom what I understand about the [issue](https://github.com/OpenRailAssociation/osrd/issues/7315) the main problem is that the confirm button is hidden when there are many rows in the grid. 
Generally, when dealing with overflow issues like this, simply increasing the height or width doesn't actually solve the problem; it just shifts it. So, I believe that a better solution would be to place the confirm/cancel buttons in a non-scrollable container, ensuring they remain visible regardless of the number of rows and apply the `overflow` on the area which can be expanded.

kmer2016:
>You are solution works with 15 rows. However, if we have more rows, like 35, a scroll bar appears. Generally with overflow problem, add the height/width don't solve the problem, it deplaceFrom what I understand about the https://github.com/OpenRailAssociation/osrd/issues/7315 the main problem is that the confirm button is hidden when there are many rows in the grid.
Generally, when dealing with overflow issues like this, simply increasing the height or width doesn't actually solve the problem; it just shifts it. So, I believe that a better solution would be to place the confirm/cancel buttons in a non-scrollable container, ensuring they remain visible regardless of the number of rows and apply the overflow on the area which can be expanded.

I have checked the buttons' separate container, and I think you can simply apply overflow: auto to the tab container.

kmer2016 (approved):
LGTM and tested !

theocrsb:
> > You are solution works with 15 rows. However, if we have more rows, like 35, a scroll bar appears. Generally with overflow problem, add the height/width don't solve the problem, it deplaceFrom what I understand about the #7315 the main problem is that the confirm button is hidden when there are many rows in the grid.
> > Generally, when dealing with overflow issues like this, simply increasing the height or width doesn't actually solve the problem; it just shifts it. So, I believe that a better solution would be to place the confirm/cancel buttons in a non-scrollable container, ensuring they remain visible regardless of the number of rows and apply the overflow on the area which can be expanded.
> 
> I have checked the buttons' separate container, and I think you can simply apply overflow: auto to the tab container.

You're right, it's done

PR #8523, opened 2024-08-21
front: fix stdcm GET departure time

### LINKED ISSUE #8310: GET: 2h offset with the trains in stdcm GET
### What happened?

I created a scenario with only one train, leaving at 12:09+:
image

I asked for a STDCM leaving between 08:00 and 10:00:
image

Which lead to a train leaving at 08:00 cause the grid's almost empty:
image

But i got a GET with `Train 1` leavin at 10:09 and `Train 2` leaving at 06:00:
image

It seems like there's a 2h difference

.

.

Recette (SNCF)

Google Chrome

0a34296

Akctarus:
closes #8310

### DIFF
--- a/front/src/applications/operationalStudies/helpers/upsertNewProjectedTrains.ts
+++ b/front/src/applications/operationalStudies/helpers/upsertNewProjectedTrains.ts
@@ -3,7 +3,7 @@ import { type ProjectPathTrainResult, type TrainScheduleResult } from 'common/ap
 import { convertIsoUtcToLocalTime } from 'utils/date';
 
 // TODO: feed the spaceTimeChart with date instead of formatted string
-const formatDatetimeForSpaceTimeChart = (departureTime: string) =>
+export const formatDatetimeForSpaceTimeChart = (departureTime: string) =>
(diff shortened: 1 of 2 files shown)


front/src/utils/date.ts
```diff
@@ -177,6 +177,13 @@ export function convertIsoUtcToLocalTime(isoUtcString: string): string {
   return dayjs(isoUtcString).local().format();
 }
 
+export function convertToLocalTime(isoString: string): string {
```
emersion: I'd prefer not to add this function to `date.ts`: it forces UTC without converting. It's not something that one should do in general (although here it's necessary to workaround a space time chart bug).

bloussou: Yes and there should be no need to convert all the trains since only the stdcm is off

Akctarus: You're right, there was a much simpler way to fix this bug

bloussou (approved):
Tested !

emersion (approved):

kmer2016 (approved):
LGTM

PR #8526, opened 2024-08-22
Remove v2 suffix

### LINKED ISSUE #8265: DROP TSV1: drop "v2" from editoast

younesschrifi:
Close #8265


shenriotpro (approved):
I don't really mind the suffix, but LGTM for tests/, assuming they pass

front/src/modules/scenario/components/ScenarioExplorer/ScenarioExplorer.tsx
```diff
@@ -44,8 +44,8 @@ const ScenarioExplorer = ({
       { skip: !globalProjectId && !globalStudyId }
     );
 
-  const { currentData: scenarioV2, isSuccess: isScenarioSuccess } =
-    osrdEditoastApi.endpoints.getV2ProjectsByProjectIdStudiesAndStudyIdScenariosScenarioId.useQuery(
+  const { currentData: Scenario, isSuccess: isScenarioSuccess } =
```
SharglutDev: We don't use capital case for consts (it's not used as a type here)

```suggestion
  const { currentData: scenario, isSuccess: isScenarioSuccess } =
```

SharglutDev (changes requested):
Thank you for this PR, v1 almost gone :) Just one comment and it's good to go for me !

editoast/src/main.rs
```diff
@@ -243,6 +243,7 @@ async fn trains_export(
     args: ExportTimetableArgs,
     db_pool: Arc<DbConnectionPoolV2>,
 ) -> Result<(), Box<dyn Error + Send + Sync>> {
+    dbg!("ok");
```
woshilapin: Can be removed.

SharglutDev (approved):
Lgtm !

editoast/src/views/path/pathfinding.rs
```diff
@@ -28,15 +28,15 @@ use crate::modelsv2::Retrieve;
...
+use crate::views::path::path_item_cache::PathItemCache;
+use crate::views::path::PathfindingError;
 use crate::views::AuthorizationError;
 use crate::views::AuthorizerExt;
 use crate::AppState;
 use editoast_models::DbConnection;
 
 crate::routes! {
-    "/v2/infra/{infra_id}/pathfinding/blocks" => post,
+    "/infra/{infra_id}/pathfinding/blocks" => post,
```
woshilapin, line 39: Do we really want to drop the V2 in the path. It's actually a quiet nice thing to version our API and help when migrating to new ones. Any opinion @flomonster @leovalais @Tguisnet?

woshilapin (approved):
I left one comment, not blocking since we do not keep the v1 around anyway. Apart from that, nice clean up, thanks for your hard work about cleaning up TSV1 🤩

editoast/src/views/path/pathfinding.rs
flomonster: I agree with @woshilapin. It's standard to have a version number in the API path.

Tguisnet: At first I was not convinced (especially for the consistency of the app), however, after discussing with @woshilapin, it could be ok to have the v2 in the path. And I agree with him to do this only when there are changes in the endpoint.

leovalais: Theoretically, I agree with you, but in the case of tsv2, I'm not entirely convinced. We have removed support for V1, so any client that was using those endpoints will have to adapt their API calls anyway. If we set up a system to fully version the editoast API and include deprecation notices, I'm all for it. But as it stands, we would have some endpoints associated with the "v2" schema, all others with the schema "none" and the "v1" path prefix leads to nowhere. Personally, I'd prefer to plan the implementation of a proper versioning system that would apply our entire API and consider the tsv2 API as being the "v1" schema.

woshilapin: I thought a bit more about this and here is what I had in mind.

1. Usually, `v2` is introduced for a new API for an entire system: this is not our case since only a subset of the API, TSv2, is concerned.
2. Usually, a `v2` is introduced to live next to the **still existing** `v1`: this is not our case since `v1` completely disappeared.
3. Usually, we define a proper scheme of versioning for the API because we have external clients on which we do not have control (so we can't know when they will migrate): this is not our case since the only clients are the `front` and our research team that have some scripts.
4. Usually, having the new API under a new prefix avoid old documentation/code to refer to existing endpoints that completely changed their API: this might be our case, although I doubt we have a lot of documentation/code that might both refer to old endpoints, and then, would end up being misleading for a potential reader.

All that said, only the 4th item would force us to do a proper migration (meaning, we would keep `v2`, but also migrate/alias all the existing non-changed endpoint so all the API surface of `editoast` would be behind `v2`). However, documentation might be the least important item in this list.

But as mentioned by @leovalais, moving to `v2` might mean more work and reflection for a ROI probably very small (again, we are the only clients of our own API at the moment). So maybe the pragmatic choice here is not to go for a standard way (prefixing everything with `v2`), but just to keep what @younesschrifi did in the first place (removing `v2` entirely)? What do you think @flomonster? In any case, having only a part (TSv2) of our API behind `v2` is not satisfactory, we should either remove it entirely, or move the entirety of our endpoints behind `v2`.

Tguisnet (approved):
Lgtm

we resolved @woshilapin comment.

editoast/src/views/path/pathfinding.rs
woshilapin: Discussed orally with @younesschrifi @leovalais @Tguisnet, let's keep the PR as-is: meaning no `v2`. We can always expose the entire API under a `v2` (or even `v1` in this case) prefix through the `gateway` which might even be better.

flomonster (approved):
LGTM

PR #8527, opened 2024-08-22
front: stdcm: fix op comparison before saving op

### LINKED ISSUE #8501: front: stdcm: when clicking the "ci" field, the date switches to "precise time"
### What happened?

In the STDCM page (new UI), when doing any interaction with the CI field, the date switches to "precise time" for this step.

It can be quite annoying as we need only one step to be in "precise time" mode.

The "precise time" / "as soon as possible" button shouldn't change unless interacted with

Go to the stdcm page (new UI). One step should be in "as soon as possible". Click on the CI field, then away from it. 

Local

Firefox

a777019e2e7ecb71bb90f8b4bc84167e9e268fd6

SarahBellaha:
Closes #8501


kmer2016 (approved):
LGTM ! and tested

Issue #8410, opened 2024-08-09
Display infrastructure status in the frontend


bloussou:
# Description :

This is not working anymore: 
![image](https://github.com/user-attachments/assets/7b901e5c-d16f-4895-86e9-f4d8a101a677)

# Implementation plan

Frontend part is already done, calls to editoast to get infra status is just responding `ready` atm.

1. osrdyne must be able to tell if the pod has loaded an infra or not (adapt `/status` to information of pod readiness)
2. editoast must call `/status`

- 1 sprint to know if the infrastructure is `none` or `live`
- 1 sprint to have all the statuses

# Acceptance Criteria : 

- [ ] infra has 5 states `none`, `scheduled`, `pending`, `running`, `live`
- [ ] one `live` pod for one infrastructure is enough to tell that infrastructure is `live`

## Definition of ready

- **PO**
  - [x] Acceptance criteria are clear and understandable

- **Technical**
  - [x] Implementation plan is detailed and validated by maintainers

- **General**
  - [x] WSJF values are filled
  - [x] Validated by the PM

celine-sncf:
Scope reduction proposal: instead of having 5 statuses that are not very clear to the end user, have only 3 statuses  `None, Loading, Ready`

louisgreiner:
## Tested in Recette (26/09/24)

- 0 to 7 scale: 
  - it still displays the loading on a scale from 0 to 7 (and not 3 statuses as [mentioned above](https://github.com/OpenRailAssociation/osrd/issues/8410#issuecomment-2371278697))
  - the hover displays `INITIALIZING` (or `CACHED`) (which is neither `None, Loading, Ready`, shouldn't it be translated ?
![Image](https://github.com/user-attachments/assets/30f04878-7052-40c1-8b83-aa261725a2a8)
  - 1 to 7 takes a long time then instantly finishes
- the editor is well disabled when infra is loading.

![Image](https://github.com/user-attachments/assets/a1978468-0bd6-4c8e-898c-e0a459948872)

PR #8528, opened 2024-08-22
osrdyne: report worker status to editoast

Khoyo:
First part of #8410 - it doesn't track the kubernetes pod startup sequence, only the infra loading status

Also include the bump to rustc 1.81, as it seems to hit the e2e tests timeout and this PR is dependent on it - but it is also the one that fixes the timeout


SharglutDev (approved):

osrdyne/src/api.rs
```diff
@@ -43,11 +66,41 @@ pub async fn create_server(
 
 async fn app_state_updater(
     state: AppState,
-    mut known_workers: tokio::sync::watch::Receiver<Arc<Vec<WorkerMetadata>>>,
+    mut known_workers_recv: watch::Receiver<Arc<Vec<WorkerMetadata>>>,
+    mut worker_status_recv: watch::Receiver<Arc<HashMap<Vec<u8>, WorkerStatus>>>,
 ) {
```
flomonster, line 71: It feels like a lot of things will have to be adapted once multiple workers could be running for one infra.

flomonster (approved):
Editoast part lgtm.

I was wondering, we tried our best not to have a link between osrdyne and editoast. This PR breaks this principle because now editoast can communicate with osrdyne. Shouldn't we just let the frontend communicate with osrdyne directly (through the gateway)?

osrdyne/src/api.rs
Khoyo: This is the next step indeed

editoast/editoast_osrdyne_client/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+
+pub struct OsrdyneClient {
+    inner: OsrdyneClientInternal,
+}
+
+enum OsrdyneClientInternal {
+    HTTPClient(HTTPClient),
+    #[cfg(any(test, feature = "mock_client"))]
+    MockClient(MockClient),
+}
```
woshilapin: Why not something like this instead? Seems weird to expose an `enum` because we will only use on XOR the other, never both in the same compilation, right?

```rust
#[cfg(not(any(test, feature = "mock_client")))]
struct OsrdyneClientInternal(HttpClient);
#[cfg(any(test, feature = "mock_client"))]
struct OsrdyneClientInternal(MockClient);
```

Khoyo:
> Shouldn't we just let the frontend communicate with osrdyne directly (through the gateway)?

I'd rather not, there can be some complications on this (eg. related to listing all infra when using the noop drivers, as osrdyne doesn't have the list of infras). I'd rather keep the front-exposed api as foolproof as possible.

Plus, there is not problem having editoast communicating with osrdyne to retrieve status information. We tried very hard for it to not be required when submitting requests (as in, signal that we need to start a worker, etc), but having the gateway and the front do it would have been problematic in the same way anyway.

editoast/editoast_osrdyne_client/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+    statuses: HashMap<String, WorkerStatus>,
+}
+
+#[cfg(any(test, feature = "mock_client"))]
+#[derive(Default)]
+pub struct MockClientBuilder {
+    statuses: HashMap<String, WorkerStatus>,
+}
+
+#[cfg(any(test, feature = "mock_client"))]
```
woshilapin: Instead of having multiple `#[cfg(any(test, feature = "mock_client"))]`, you could move all of that code in a module and have only one `#[cfg]`.

editoast/src/views/test_app.rs
```diff
@@ -34,19 +35,23 @@ use super::{authorizer_middleware, Roles};
...
+#[cfg(test)]
 pub(crate) struct TestAppBuilder {
     db_pool: Option<DbConnectionPoolV2>,
     core_client: Option<CoreClient>,
     roles: Option<Roles>,
+    osrdyne_client: Option<OsrdyneClient>,
     db_pool_v1: bool,
 }
 
+#[cfg(test)]
```
woshilapin: Instead of putting it here, can we put it on the entire module in `src/views.rs`? (and maybe reexpose the exception if they need it).

editoast/src/views/test_app.rs
```diff
@@ -64,6 +69,12 @@ impl TestAppBuilder {
         self
     }
 
+    pub fn osrdyne_client(mut self, osrdyne_client: OsrdyneClient) -> Self {
+        assert!(self.osrdyne_client.is_none());
```
woshilapin, line 71: Why is that important to assert that? Wouldn't you want to just replace it?

osrdyne/src/main.rs
```diff
@@ -98,8 +101,20 @@ async fn main() -> Result<(), anyhow::Error> {
     ));
 
     let (worker_list_send, woker_list_recv) = tokio::sync::watch::channel(Arc::new(vec![]));
```
woshilapin: Is it a typo (not from your PR but can be easily fixed)?

```suggestion
    let (worker_list_send, worker_list_recv) = tokio::sync::watch::channel(Arc::new(vec![]));
```

woshilapin (approved):
Nothing major, and CI is green on first try, which is a good sign considering our recent CI frequent failures.

editoast/src/views/test_app.rs
Khoyo: Consistency with the other methods... I agree though

flomonster:
> [!NOTE]
> This [PR](https://github.com/OpenRailAssociation/osrd/pull/8675) should also be merged to have rustc v1.81.0 for nix users :wink:

editoast/editoast_osrdyne_client/src/lib.rs
Khoyo: In principle yes, but that would mean `new()` must either create a mock client when compiling for tests/checking/etc (which may lead to unintended consequences), or be disabled too - which would have knock on effects on editoast main.

Additionally, this enum is not exposed and zero cost when not compiling with the mock client

I guess this is why CoreClient uses an enum too (even before the amqp changes).

eckter (approved):
LGTM for core (not tested)

Erashin (approved):
Core lgtm

editoast/src/views/infra/mod.rs
```diff
@@ -255,7 +255,18 @@ pub enum InfraState {
...
 
-#[derive(Debug, Clone, Serialize, ToSchema)]
+impl From<editoast_osrdyne_client::WorkerStatus> for InfraState {
+    fn from(status: editoast_osrdyne_client::WorkerStatus) -> Self {
+        match status {
+            editoast_osrdyne_client::WorkerStatus::Unscheduled => InfraState::NotLoaded,
+            editoast_osrdyne_client::WorkerStatus::Started => InfraState::Initializing,
+            editoast_osrdyne_client::WorkerStatus::Ready => InfraState::Cached,
+            editoast_osrdyne_client::WorkerStatus::Error => InfraState::Error,
+        }
```
leovalais, line 266: You can remove the outdated `InfraState` variants

editoast/src/views/infra/mod.rs
```diff
@@ -727,12 +738,26 @@ async fn load(
...
+pub enum InfraStateError {
+    #[error("Failed to fetch infra state: {0}")]
+    #[editoast_error(status = 500)]
+    FetchError(#[from] editoast_osrdyne_client::Error),
+}
+
+pub async fn fetch_infra_state(
+    infra_id: i64,
+    osrdyne: &OsrdyneClient,
+) -> Result<InfraStateResponse> {
```
leovalais: `InfraStateResponse` is a `core_client` definition, which is not relevant anymore. You can delete `src/core/infra_state.rs` and return directly the `InfraState`.

leovalais (approved):
LGTM. A few old things to remove. I'll test later (unless I forget).

editoast/src/views/infra/mod.rs
Khoyo: If that's okay with you, I'd rather do this all at once when we actually integrate the new loading states (from k8s)

ghost (approved):
LGTM, not tested. Will be affected by the KEDA support coming soon.

PR #8529, opened 2024-08-22
editoast: mark sprites as cacheable

emersion:
Same goal as #8514 but for sprites.

Since the data comes from the filesystem, we can just use ServeFile instead of hand-rolling our own logic. We get Last-Modified for free (allows user agents to cache requests).


leovalais (approved):
Didn't know about `ServeFile`, lgtm!

PR #8531, opened 2024-08-22
front: allow to edit invalid trains without rolling stock name

### LINKED ISSUE #8530: Input fields of invalid train are not updated
### What happened?

I edit a valid train -> OK
I go to the edition mode of an invalid train -> all the fields are filled with those of the valid train so I cannot edit my invalid train without changing everything
I change the name in the field of the invalid train (without saving, juste to change what is in the store)
I go to the edition mode of the valid train -> I see for 1s the new name, but then the real name of the valid train comes back : I would expect the same thing for invalid trains

https://github.com/user-attachments/assets/f9502669-fa3d-4a44-9b8d-7435a734d5c9

I expect the fields to be displayed with their value and not with the store

As in the video

Dev (SNCF)

Google Chrome

769e7c8

clarani:
The user couldn't see the configuration of an invalid train without rolling stock name because of an early return.

closes #8530

### DIFF
--- a/front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
+++ b/front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
@@ -237,32 +237,34 @@ const useSetupItineraryForTrainUpdate = (
         return;
       }
       const trainSchedule = await getTrainScheduleById({ id: trainIdToEdit }).unwrap();
-      if (!trainSchedule.rolling_stock_name) {
-        return;
-      }
 
-      let itinerary = null;
-      try {
-        const rollingStock = await getRollingStockByName({
-          rollingStockName: trainSchedule.rolling_stock_name,
-        }).unwrap();
-        itinerary = await computeItineraryForTrainUpdate(trainSchedule, rollingStock);
-        const { pathSteps, pathProperties } = itinerary || {};
+      let rollingStock: RollingStockWithLiveries | null = null;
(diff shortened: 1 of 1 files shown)


front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -237,32 +237,36 @@ const useSetupItineraryForTrainUpdate = (
...
 
       let itinerary = null;
-      try {
-        const rollingStock = await getRollingStockByName({
-          rollingStockName: trainSchedule.rolling_stock_name,
-        }).unwrap();
-        itinerary = await computeItineraryForTrainUpdate(trainSchedule, rollingStock);
-        const { pathSteps, pathProperties } = itinerary || {};
+      let rollingStock: RollingStockWithLiveries | null = null;
+      let pathSteps: (PathStep | null)[] | undefined;
```
emersion, line 242: Nit: we don't really need this variable. In the `adjustConfWithTrainToModifyV2()` call we can use `itinerary?.pathSteps` instead.

emersion (approved):
LGTM and works

Feel free to resolve the comment

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
clarani: Yes you're right, I remove it :+1:

kmer2016 (approved):
LGTM and tested

PR #8532, opened 2024-08-22
editoast: stdcm: add end stop to preprocessing request

### LINKED ISSUE #8522: stdcm: the preprocessing simulation request doesn't end with a stop
### What happened?

When trying an stdcm request with a really weak rolling stock, I had a preprocessing error saying that the 4.5m/100km margin was impossible. Which should be impossible, but the preprocessing simulation doesn't end with a stop. It had a fixed end speed > 0, which was impossible with the margin.

It also means that our preprocessing time estimation is too optimistic. 

_No response_

Check the request sent to core for the preprocessing simulation

Local

Firefox

b58bc6b4c2c5130a822aebcda1ffd2a69b2236cc

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/8522

The lack of stop there would mean that the preprocessing simulation would end with a speed >0. This isn't what we'd want here, and it sometimes even fail to match the margin.

### DIFF
--- a/editoast/src/views/v2/timetable/stdcm.rs
+++ b/editoast/src/views/v2/timetable/stdcm.rs
@@ -7,9 +7,9 @@ use chrono::Utc;
 use chrono::{DateTime, Duration, NaiveDateTime, TimeZone};
 use editoast_authz::BuiltinRole;
 use editoast_derive::EditoastError;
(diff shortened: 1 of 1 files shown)


leovalais (approved):
LGTM

PR #8536, opened 2024-08-22
front: drop axios dependency

emersion:
Replace axios with the [fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).

_See individual commits._

Motivations:

- The fetch API is a Web standard (developers are more likely to be familiar with the APIs implemented by browsers, rather than libraries).
- The fetch API is about as ergonomic as axios, so we don't really loose anything.
- One less dependency (less code to fetch/build/bundle, no need to keep it up-to-date, no need to worry about vulnerabilities, etc).

Also fix a small bug while at it.

To test:

- Create a scenario with a custom uploaded image.
- Search for train stations and train schedules with the import tool.


theocrsb (approved):
LGTM and tested

clarani (approved):
LGTM and tested :white_check_mark: Thanks for the cleaning :broom:

PR #8542, opened 2024-08-22
front: open-data import: keep stop at end of path

### LINKED ISSUE #8541: No more stop at the end of open data imported trains
### What happened?

I imported trains between RES and QR and none of them as a stop at the end. This is due to the fact that this stop used to be automatically added at each circulation but it's not the case anymore.

I'm expecting the impported trains to trop at the end.

1. Import open data trains

Recette (SNCF)

Google Chrome

c885d29

Castavo:
close #8541

### DIFF
--- a/front/src/modules/trainschedule/components/ImportTrainSchedule/generateTrainSchedulesPayloads.ts
+++ b/front/src/modules/trainschedule/components/ImportTrainSchedule/generateTrainSchedulesPayloads.ts
@@ -20,12 +20,15 @@ export function generateV2TrainSchedulesPayloads(
             train.departureTime,
             step.arrivalTime
           );
-          const isoDuration = formatDurationAsISO8601(timeDifferenceInSeconds);
-          acc.schedule.push({
+          const schedulePoint = {
             at: stepId,
-            arrival: isoDuration,
-            ...(step.duration ? { stop_for: `PT${step.duration}S` } : {}),
-          });
+            arrival: formatDurationAsISO8601(timeDifferenceInSeconds),
+            stop_for: null as string | null,
+          };
+          if (step.duration || index === train.steps.length - 1) {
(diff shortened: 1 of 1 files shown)


emersion:
Ah. Is this something the front is supposed to do, or is this something the backend is supposed to handle?

Castavo:
@emersion If you don't specify a stop at the end of a path, then the simulated train won't stop. It's the expected behavior of the backend.
You need to have the stop (even for 0s) at the end of the path if you want the train to stop

front/src/modules/trainschedule/components/ImportTrainSchedule/generateTrainSchedulesPayloads.ts
```diff
@@ -20,12 +20,15 @@ export function generateV2TrainSchedulesPayloads(
...
+          const schedulePoint = {
             at: stepId,
-            arrival: isoDuration,
-            ...(step.duration ? { stop_for: `PT${step.duration}S` } : {}),
-          });
+            arrival: formatDurationAsISO8601(timeDifferenceInSeconds),
+            stop_for: null as string | null,
+          };
+          if (step.duration || index === train.steps.length - 1) {
+            schedulePoint.stop_for = `PT${step.duration}S`;
```
clarani: You should set the value 0 by default, for if the step.duration is undefined ;)
```suggestion
            schedulePoint.stop_for = `PT${step.duration || 0}S`;
```

RomainValls (approved):
LGTM and tested ✅

front/src/modules/trainschedule/components/ImportTrainSchedule/generateTrainSchedulesPayloads.ts
Castavo: Indeed

clarani (approved):

PR #8543, opened 2024-08-22
tests: update fuzzer

eckter:
1. Fix a few outdated elements in the payload format
2. Add scheduled points for stdcm payloads instead of global departure time
3. Remove v1 fuzzer and remove `v2` from script name

The two commits may be reviewed separately as the renaming v2 messes up with the diff


shenriotpro:
It looks like there's an issue with timetable creation

eckter:
I forgot about this PR...

The file has diverged too much from the main branch, I'll close this and maybe open another one if relevant

PR #8550, opened 2024-08-22
front: add button to set STDCM search environment

emersion:


Castavo (changes requested):
This button should probably be right next to the scenario selector.
Also, it seems to be the same colour as the background, making it look like just text.

The search window is used to limit the date at which the stdcm request can be done. It should be centered on the trains in the timetable associated rather than on the current date.

This does not allow setting the work schedule group... Most importantly, it unsets it so a manual request has to be done to set it back when needed

In fact, (I could have thought about this earlier, sorry), what about race conditions ? Since this is a shared state, how would this behave when different persons interact with it ?

emersion:
This view is getting nuked so no point in adding developer features to it.

PR #8555, opened 2024-08-22
front: output table does not display schedule for op via

### LINKED ISSUE #8453: Imported trains requested times are not visible in the output table
### What happened?

I imported trains from open data between RES and QR, and none of them have the information of the requested times, but most have the yellow clock icon idincating that some times have not been matched:

image

Even though they clearly try to respect specific times if we look at the space-speed-chart:

image

I'm expecting the requested times to be visible in the output table, in order to understand what hour has not been respected.

1. Import trains on France between RES and QR

Recette (SNCF)

Google Chrome

ab7706e

clarani:
closes #8453

Because of the condition `if (pathStepKey in pathStepsByUic && nextOpPoint)` (l95 in `useOutputData`), the schedules of the pathSteps whose location was not a uic were ignored.

Now, we handle the pathSteps first, then the suggestedOp, and finally we merge them using the positionOnPath.

### DIFF
--- a/front/src/modules/timesStops/hooks/useOutputTableData.ts
+++ b/front/src/modules/timesStops/hooks/useOutputTableData.ts
@@ -6,7 +6,7 @@ import type {
   PathPropertiesFormatted,
   SimulationResponseSuccess,
 } from 'applications/operationalStudies/types';
-import type { TrainScheduleResult } from 'common/api/osrdEditoastApi';
+import type { PathfindingResultSuccess, TrainScheduleResult } from 'common/api/osrdEditoastApi';
 import { formatSuggestedOperationalPoints } from 'modules/pathfinding/utils';
 import type { OperationalPointWithTimeAndSpeed } from 'modules/trainschedule/components/DriverTrainScheduleV2/types';
 import type { SuggestedOP } from 'modules/trainschedule/components/ManageTrainSchedule/types';
@@ -23,20 +23,28 @@ function useOutputTableData(
   pathProperties: PathPropertiesFormatted,
   operationalPoints: OperationalPointWithTimeAndSpeed[],
   selectedTrainSchedule: TrainScheduleResult,
-  pathLength?: number
+  path?: PathfindingResultSuccess
 ) {
   const scheduleByAt: Record<string, ScheduleEntry> = keyBy(selectedTrainSchedule.schedule, 'at');
   const suggestedOperationalPoints: SuggestedOP[] = useMemo(
     () =>
-      pathLength
+      path?.length
         ? formatSuggestedOperationalPoints(
             pathProperties.operationalPoints,
             pathProperties.geometry,
-            pathLength
+            path.length
           )
         : [],
     [pathProperties.operationalPoints, pathProperties.geometry]
   );
+
+  const pathStepsWithPositionOnPath = useMemo(() => {
+    if (!path || !selectedTrainSchedule) return [];
+    return selectedTrainSchedule.path.map((pathStep, index) => ({
+      ...pathStep,
+      positionOnPath: path.path_item_positions[index],
+    }));
+  }, [path, selectedTrainSchedule]);
 
   const pathStepsWithOpPointIndices = useMemo(
     () =>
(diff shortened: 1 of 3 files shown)


bloussou:
Tested ✅

bloussou (approved):
Tested ✅

emersion:
Needs to be rebased on top of https://github.com/OpenRailAssociation/osrd/commit/f0a220a785869208208e26324110725ea9aba048 (sorry!).

Akctarus (approved):
Tested and LGTM

front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -65,51 +73,61 @@ function useOutputTableData(
     (opPoint) => `${opPoint.name}-${opPoint.ch}`
```
emersion, line 73: Unrelated drive-by comment: we probably shouldn't use "name" as a key, it's not guaranteed to be unique.

front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -65,51 +73,61 @@ function useOutputTableData(
...
-        return { ...sugOpPoint, calculatedArrival: secToHoursString(opPoint.time, true) };
-      }),
-    [simulatedTrain, pathProperties, operationalPoints, selectedTrainSchedule, pathLength]
-  );
+      return { ...sugOpPoint, calculatedArrival: secToHoursString(opPoint.time, true) };
+    });
+
+    return suggestedOpRows.map((sugOpPoint) => {
+      const matchingPathStep = pathStepRows.find(
+        (pathStepRow) => sugOpPoint.positionOnPath === pathStepRow.positionOnPath
```
emersion, line 120: I must admit I'm not a fan of using the position on path as a kind of unique identifier… The backend doesn't guarantee that the `positionOnPath` are unique for each OP. Ideally we should also match the trigram/UIC/etc with e.g. `matchPathStepAndOp()` when we're hitting duplicate `positionOnPath` values. But it sounds unlikely to happen with the France infrastructure, so maybe we can just leave this as a TODO or something…

Edit: turns out it does happen quite often in practice.

front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -65,51 +73,61 @@ function useOutputTableData(
...
-            ...formattedScheduleData,
-            onStopSignal: schedule?.on_stop_signal || '',
-            calculatedArrival,
-            calculatedDeparture:
-              opPoint.duration > 0 ? secToHoursString(opPoint.time + opPoint.duration, true) : '',
-            ...marginsData,
-          } as SuggestedOP;
-        }
+      if (pathStepKey in pathStepsByUic) {
+        const pathStepId = pathStepsByUic[pathStepKey].id || '';
```
emersion, line 96: Any reason why we're still using the uic+ch here to fish the `pathStepId` (and then the `schedule`) instead of using a logic similar to the one below?

front/src/modules/timesStops/hooks/useOutputTableData.ts
clarani: Yes, I agree :/ I have a refacto in mind, can we leave it like this for now ?

front/src/modules/timesStops/hooks/useOutputTableData.ts
clarani: I didn't change the logic here. I can leave it like this for now if it's ok for you ?

front/src/modules/timesStops/hooks/useOutputTableData.ts
emersion: Sure

front/src/modules/timesStops/hooks/useOutputTableData.ts
emersion: Sure

emersion (approved):
Still quite fragile, but it's an improvement.

front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -65,51 +73,61 @@ function useOutputTableData(
...
-          operationalPoints,
-          pathStepsWithOpPointIndices,
-          sugOpIndex
-        );
-        const pathStepKey = `${sugOpPoint.uic}-${sugOpPoint.ch}`;
+  const outputTableData = useMemo(() => {
+    const pathStepRows = pathStepsWithPositionOnPath.map((pathStep) => {
+      const schedule = scheduleByAt[pathStep.id];
+      const scheduleData = computeScheduleData(schedule, selectedTrainSchedule.start_time);
+      return { ...pathStep, ...formatScheduleData(scheduleData) };
```
SharglutDev, line 80: We use this combination two times in the file (computeScheduleData then format the result). Maybe we could do the format inside the compute as format is never used without computing first ?

SharglutDev (approved):
Lgtm (just one comment that you can resolve if you don't agree)

front/src/modules/timesStops/hooks/useOutputTableData.ts
clarani: Yes I agree, but I would like to do it separately if it's ok for you :)

PR #8563, opened 2024-08-23
Just a pr so the images are built

Castavo:


bloussou:
marking ready for review to have a build

PR #8564, opened 2024-08-23
Add rolling stock related schedules list endpoint

### LINKED ISSUE #8078: Create an endpoint that lists the train schedules a rolling stock is used in
### Description and goal

We used to have such a mechanism that prevented a rolling stock from being deleted if it was used in some train schedules. We did that because `TrainSchedule` V1 used to reference a rolling stock using a foreign key. So removing it also deleted the associated train schedules.

Since TSV2, rolling stocks are loosely referenced using their name. If the rolling stock doesn't exist, the train schedule is deemed invalid but still exists. We still want to warn the user that deleting a rolling stock might invalidate some train schedules though.

We therefore need a new endpoint `GET /rolling_stock/:id/usage` that can retrieve the list of `TrainSchedule` (v2) in which the rolling stock is used. The response could be typed like this:

```ts
type Response = {
    project_id: number,
    project_name: string,
    study_id: number,
    study_name: number,
    scenario_id: number,
    scenario_name: string
}[]
```

> [!NOTE]
> The front-end part can be done separately.

### Acceptance criteria

- [ ] The endpoint exists and works with Train Schedule V2
- [ ] Unit tests

Sh099078:
Fixes #8078

### DIFF
--- a/editoast/openapi.yaml
+++ b/editoast/openapi.yaml
@@ -1779,6 +1779,29 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/RollingStock'
+  /rolling_stock/{rolling_stock_id}/usage:
+    get:
+      tags:
+      - rolling_stock
+      summary: List the scenarios (and their respective studies and projects) which use a given rolling stock.
+      parameters:
+      - name: rolling_stock_id
+        in: path
+        required: true
+        schema:
+          type: integer
+          format: int64
+      responses:
+        '200':
+          description: A list of the associated scenarios and their respective studies and projects.
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/ScenarioReference'
+        '404':
+          description: The requested rolling stock was not found
   /search:
     post:
       tags:
@@ -8185,6 +8208,31 @@ components:
           allOf:
           - $ref: '#/components/schemas/Tags'
           nullable: true
+    ScenarioReference:
+      type: object
+      required:
+      - project_id
+      - project_name
+      - study_id
+      - study_name
+      - scenario_id
+      - scenario_name
+      properties:
+        project_id:
+          type: integer
+          format: int64
+        project_name:
+          type: string
+        scenario_id:
+          type: integer
+          format: int64
+        scenario_name:
+          type: string
+        study_id:
+          type: integer
+          format: int64
+        study_name:
+          type: string
     ScenarioResponseV2:
       allOf:
       - $ref: '#/components/schemas/ScenarioV2'
--- a/editoast/src/modelsv2/rolling_stock_model.rs
+++ b/editoast/src/modelsv2/rolling_stock_model.rs
@@ -23,6 +23,9 @@ use validator::ValidationErrors;
 
 use crate::modelsv2::prelude::*;
 
+mod schedules_from_rolling_stock;
+pub use schedules_from_rolling_stock::ScenarioReference;
+
 editoast_common::schemas! {
     RollingStockModel,
     PowerRestriction,
--- a/editoast/src/modelsv2/rolling_stock_model/schedules_from_rolling_stock.rs
+++ b/editoast/src/modelsv2/rolling_stock_model/schedules_from_rolling_stock.rs
@@ -1,0 +1,68 @@
+use diesel::prelude::*;
+use diesel_async::RunQueryDsl;
+use itertools::Itertools;
+#[cfg(test)]
+use serde::Deserialize;
+use serde::Serialize;
+use utoipa::ToSchema;
+
+use crate::error::Result;
+use crate::modelsv2::rolling_stock_model::RollingStockModel;
+use editoast_models::db_connection_pool::DbConnectionV2;
+use editoast_models::tables::{project, rolling_stock, scenario_v2, study, train_schedule_v2};
+
+#[derive(Debug, Serialize, ToSchema)]
+#[cfg_attr(test, derive(PartialEq, Eq, PartialOrd, Ord, Deserialize))]
+pub struct ScenarioReference {
+    pub project_id: i64,
+    pub project_name: String,
+    pub study_id: i64,
+    pub study_name: String,
+    pub scenario_id: i64,
+    pub scenario_name: String,
+}
+
+impl From<SchedulesFromRollingStock> for ScenarioReference {
+    fn from(value: SchedulesFromRollingStock) -> Self {
+        let (project_id, project_name, study_id, study_name, scenario_id, scenario_name) = value;
+        ScenarioReference {
+            project_id,
+            project_name,
+            study_id,
+            study_name,
+            scenario_id,
+            scenario_name,
+        }
+    }
+}
+
+type SchedulesFromRollingStock = (i64, String, i64, String, i64, String);
+
+impl RollingStockModel {
+    pub async fn get_usage(&self, conn: &mut DbConnectionV2) -> Result<Vec<ScenarioReference>> {
+        let schedules: Vec<_> = train_schedule_v2::table
+            .inner_join(
+                rolling_stock::table
+                    .on(train_schedule_v2::rolling_stock_name.eq(rolling_stock::name)),
+            )
+            .inner_join(
+                (scenario_v2::table
+                    .on(scenario_v2::timetable_id.eq(train_schedule_v2::timetable_id)))
+                .inner_join(study::table.inner_join(project::table)),
+            )
+            .select((
+                project::id,
+                project::name,
+                study::id,
+                study::name,
+                scenario_v2::id,
+                scenario_v2::name,
+            ))
+            .filter(rolling_stock::id.eq(self.id))
+            .filter(train_schedule_v2::id.is_not_null())
+            .load::<SchedulesFromRollingStock>(conn)
+            .await?;
+        let schedules = schedules.into_iter().map_into().collect();
+        Ok(schedules)
+    }
+}
--- a/editoast/src/views/rolling_stocks/mod.rs
+++ b/editoast/src/views/rolling_stocks/mod.rs
@@ -35,6 +35,7 @@ use crate::error::InternalError;
 use crate::error::Result;
 use crate::modelsv2::prelude::*;
 use crate::modelsv2::rolling_stock_livery::RollingStockLiveryModel;
+use crate::modelsv2::rolling_stock_model::ScenarioReference;
 use crate::modelsv2::rolling_stock_model::TrainScheduleScenarioStudyProject;
 use crate::modelsv2::Document;
 use crate::modelsv2::RollingStockModel;
@@ -53,6 +54,7 @@ crate::routes! {
             delete,
             "/locked" => update_locked,
             "/livery" => create_livery,
+            "/usage" => get_usage,
         },
     },
 }
@@ -65,6 +67,7 @@ editoast_common::schemas! {
     RollingStockError,
     RollingStockKey,
     RollingStockWithLiveries,
+    ScenarioReference,
     light_rolling_stock::schemas(),
 }
 
@@ -626,6 +629,34 @@ async fn create_livery(
     Ok(Json(rolling_stock_livery))
 }
 
+/// List the scenarios (and their respective studies and projects) which use a given rolling stock.
+#[utoipa::path(
+    get, path = "",
+    tag = "rolling_stock",
+    params(RollingStockIdParam),
+    responses(
+        (status = 200, description = "A list of the associated scenarios and their respective studies and projects.", body = Vec<ScenarioReference>),
+        (status = 404, description = "The requested rolling stock was not found"),
+    )
+)]
+pub async fn get_usage(
+    State(db_pool): State<DbConnectionPoolV2>,
+    Path(rolling_stock_id): Path<i64>,
+) -> Result<Json<Vec<ScenarioReference>>> {
+    let mut conn = db_pool.get().await?;
+
+    let rolling_stock = RollingStockModel::retrieve_or_fail(&mut conn, rolling_stock_id, || {
+        RollingStockError::KeyNotFound {
+            rolling_stock_key: RollingStockKey::Id(rolling_stock_id),
+        }
+    })
+    .await?;
+
+    let related_train_schedules = rolling_stock.get_usage(&mut conn).await?;
+
+    Ok(Json(related_train_schedules))
+}
+
 /// Retrieve a rolling stock by id or by name
 pub async fn retrieve_existing_rolling_stock(
     conn: &mut DbConnection,
@@ -742,19 +773,14 @@ pub mod tests {
     use pretty_assertions::assert_eq;
     use rstest::rstest;
     use serde_json::json;
+    use uuid::Uuid;
 
     use crate::error::InternalError;
 
-    use crate::modelsv2::fixtures::create_fast_rolling_stock;
-
-    use crate::modelsv2::fixtures::create_rolling_stock_with_energy_sources;
-
-    use crate::modelsv2::fixtures::fast_rolling_stock_changeset;
-    use crate::modelsv2::fixtures::fast_rolling_stock_form;
-    use crate::modelsv2::fixtures::get_rolling_stock_with_invalid_effort_curves;
-    use crate::modelsv2::fixtures::rolling_stock_with_energy_sources_form;
-    use crate::modelsv2::prelude::*;
+    use super::*;
+    use crate::modelsv2::fixtures::*;
     use crate::modelsv2::rolling_stock_model::RollingStockModel;
+    use crate::modelsv2::train_schedule::TrainSchedule;
     use crate::views::rolling_stocks::rolling_stock_form::RollingStockForm;
     use crate::views::test_app::TestApp;
     use crate::views::test_app::TestAppBuilder;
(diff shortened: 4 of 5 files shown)


editoast/src/modelsv2/rolling_stock_model/schedules_from_rolling_stock.rs
```diff
@@ -0,0 +1,112 @@
...
+use utoipa::ToSchema;
+
+use crate::error::Result;
+use editoast_models::db_connection_pool::DbConnectionV2;
+
+#[derive(Debug, Serialize, Deserialize, ToSchema)]
+pub struct RollingStockUsage {
+    #[schema(inline)]
+    pub train_schedule: TrainScheduleReference,
+    #[schema(inline)]
```
leovalais:
```suggestion
    #[schema(inline, required)]
```

editoast/src/modelsv2/rolling_stock_model/schedules_from_rolling_stock.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+#[derive(Debug, Serialize, Deserialize, ToSchema)]
+pub struct TrainScheduleReference {
+    pub id: i64,
+    pub name: String,
+}
+
+#[derive(Debug, Serialize, Deserialize, ToSchema)]
+pub struct StudyReference {
```
leovalais:
```suggestion
pub struct ScenarioReference {
```

editoast/src/modelsv2/rolling_stock_model/schedules_from_rolling_stock.rs
```diff
@@ -0,0 +1,112 @@
...
+            ) => Some(StudyReference {
+                project_id,
+                project_name,
+                study_id,
+                study_name,
+                scenario_id,
+                scenario_name,
+            }),
+            _ => None,
+        };
```
leovalais: For `match`es like this one, there's the `if let` construct. You can also avoid repeating the `Some` in the destructuring pattern by using `Option::zip`.

editoast/src/modelsv2/rolling_stock_model/schedules_from_rolling_stock.rs
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+}
+
+impl RollingStockUsage {
+    pub async fn get_train_schedules_from_rolling_stock(
+        conn: &mut DbConnectionV2,
+        rs_id: i64,
+    ) -> Result<Vec<RollingStockUsage>> {
+        let schedules = sql_query(include_str!("sql/get_related_train_schedules.sql"))
```
leovalais: It'd be nice to rewrite this request using `diesel`'s DSL. It allows checking the consistency of the request with the database schema at compile-time. That really helps when refactoring. There's examples at https://diesel.rs and don't hesitate to ping if the type checker gets annoying :D

editoast/src/modelsv2/rolling_stock_model/schedules_from_rolling_stock.rs
```diff
@@ -0,0 +1,112 @@
...
+            train_schedule: TrainScheduleReference { id, name },
+            op_study: study_reference,
+        }
+    }
+}
+
+impl RollingStockUsage {
+    pub async fn get_train_schedules_from_rolling_stock(
+        conn: &mut DbConnectionV2,
+        rs_id: i64,
```
leovalais:
```suggestion
impl RollingStockModel {
    pub async fn get_usage(
        &self,
        conn: &mut DbConnectionV2,
```

editoast/src/views/rolling_stocks/mod.rs
```diff
@@ -781,6 +817,68 @@ pub mod tests {
...
+
+        let request = app.get(&format!("/rolling_stock/{stock_id}/usage"));
+        let related_schedules: Vec<RollingStockUsage> =
+            app.fetch(request).assert_status(StatusCode::OK).json_into();
+        assert_eq!(
+            related_schedules
+                .iter()
+                .map(|stock_usage| stock_usage.train_schedule.id)
+                .collect::<Vec<_>>(),
+            vec!(schedule_1.id, schedule_2.id)
```
leovalais:
```suggestion
            vec![schedule_1.id, schedule_2.id]
```

leovalais (changes requested):
Great! A few remarks. Don't hesitate to answer if something's not clear or if you disagree with something :)

Not tested yet.

editoast/src/modelsv2/rolling_stock_model/schedules_from_rolling_stock.rs
Sh099078: I could do something along the lines of :
```
let study_reference = if let Some((project_id, (project_name, (study_id, (study_name, (scenario_id, scenario_name)))))) = 
    project_id.zip(project_name.zip(study_id.zip(study_name.zip(scenario_id.zip(scenario_name))))) {
            Some(ScenarioReference {
                project_id,
                project_name,
                study_id,
                study_name,
                scenario_id,
                scenario_name,
            })
    } else { None };
```

It avoids repeating the `Some` everywhere but we are now dealing with nested tuples due to the multiple calls to `Option::zip`.

Am I missing on a cleaner way to rewrite this part ?

leovalais: Admittedly it's not much better. I like it slightly more as it spreads over less lines but that's aesthetics and it forces us to deal with nested tuples. I'm not sure there's a cleaner standard way. We have [itertools](https://docs.rs/itertools/latest/itertools/) already in our deps, maybe something there would be useful? (`Option::iter` returns an iterator with 0 or 1 value.)

In any case, `zip` was a suggestion more than anything else, just keep the version you prefer :)

leovalais: Maybe https://docs.rs/itertools/latest/itertools/macro.izip.html?

Sh099078: It feels like `Option` could use some sort of `zip_multiple` macro for such cases that would allow to build from `n` options in one go an option containing a n-tuple equivalent to calling `Option::zip` n times and then flattening the result.

I agree the `Option::zip` approach is a bit more concise, I will go with it. Thanks for mentioning `Option::zip`, I did not know about it !

leovalais: FYI there's one there:

https://github.com/OpenRailAssociation/osrd/blob/b4cfac0e0659d6cbe3735d32d919eeba68500b13/editoast/editoast_derive/src/modelv2/utils.rs#L4-L17

However I'm not in favor of using it in editoast itself as it's both non-standard and not coming from a "popular" crate. (Ie. let's not invent our own stdlib.) It's useful in that module though because 1. the manual alternative is (much) worse, 2. the macro is only involved locally, 3. it's used **extensively**.

woshilapin: Maybe something like this?

```rust
let scenario_reference = project_id
    .and_then(|project_id| Some(ScenarioReference {
        project_id,
        project_name: project_name?,
        study_id: study_id?,
        study_name: study_name?,
        scenario_id: scenario_id?,
        scenario_name: scenario_name?,
    }))
```

Or more cursed if you're not ok with `project_id` being the only one treated differently...

```rust
let scenario_reference = None
    .and_then(|()| Some(ScenarioReference {
        project_id: project_id?,
        project_name: project_name?,
        study_id: study_id?,
        study_name: study_name?,
        scenario_id: scenario_id?,
        scenario_name: scenario_name?,
    }))
```

Note the empty tuple `()` that will help the type inference of `None` to avoid [turbofish](https://turbo.fish/) on `None::<()>`. It's also nice to use the empty tuple because it will destructure so you don't need to specify the type of the closure's parameter (no need to type `|_: ()|`).

editoast/src/modelsv2/rolling_stock_model.rs
```diff
@@ -23,6 +23,9 @@ use validator::ValidationErrors;
 
 use crate::modelsv2::prelude::*;
 
+pub mod schedules_from_rolling_stock;
```
woshilapin: Is there anything else in the module that is exposed. I believe you can reduce the API here.

```suggestion
mod schedules_from_rolling_stock;
```

woshilapin (approved):
Apart from the expert comments from @leovalais 😗, the rest looks good. Thank you.

editoast/src/modelsv2/rolling_stock_model/schedules_from_rolling_stock.rs
leovalais: That's much better indeed :) I'm not sure about the second part though, `and_then` is defined like:

```rust
    pub fn and_then<U, F>(self, f: F) -> Option<U>
    where
        F: FnOnce(T) -> Option<U>,
    {
        match self {
            Some(x) => f(x),
            None => None,
        }
    }
```

Sh099078: Thank you for your comments !

I also find the `project_id.and_then(...)` option to be the cleanest one. I will go with it by default.

editoast/src/modelsv2/rolling_stock_model/schedules_from_rolling_stock.rs
Sh099078: Should the response field be renamed from `op_study` to `scenario` aswell ?

leovalais: I'd say yes.

editoast/src/modelsv2/rolling_stock_model/schedules_from_rolling_stock.rs
woshilapin: Note, after discussion with @leovalais, `None.and_then(|()| todo!())` will not execute the closure, but `Some(()).and_then(|_| todo!())` will. Maybe we want to write it like this?

Khoyo: We should probably just call the closure ourselves...

```rust
let scenario_reference = (|| Some(ScenarioReference {
        project_id: project_id?,
        project_name: project_name?,
        study_id: study_id?,
        study_name: study_name?,
        scenario_id: scenario_id?,
        scenario_name: scenario_name?,
    }))();
```

Sh099078: cease-fire

editoast/src/modelsv2/rolling_stock_model/schedules_from_rolling_stock.rs
```diff
@@ -0,0 +1,96 @@
...
+use crate::tables::{project, rolling_stock, scenario_v2, study, train_schedule_v2};
+use editoast_models::db_connection_pool::DbConnectionV2;
+
+#[derive(Debug, Serialize, Deserialize, ToSchema)]
+pub struct RollingStockUsage {
+    #[schema(inline, required)]
+    pub train_schedule: TrainScheduleReference,
+    #[schema(inline)]
+    pub scenario: Option<ScenarioReference>,
+}
```
flomonster: After discussing with @leovalais : 
- We should only return `Usage` timetables used by a scenario(s)
    - Note that multiple scenarios can use the same timetable. In this case, all the scenarios should be returned
 - No need to return the `TrainScheduleReference`. In the future, we could put scenario filters in the URL so the user could directly see the train schedules that are using the rolling stock.

Sh099078: Updated the endpoint accordingly in https://github.com/OpenRailAssociation/osrd/pull/8564/commits/6eefdddba70a32c74936c9fcca757ec774c7ad10.

editoast/openapi.yaml
```diff
@@ -2102,6 +2102,52 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/RollingStock'
+  /rolling_stock/{rolling_stock_id}/usage:
+    get:
+      tags:
+      - rolling_stock
+      summary: List the train schedules a rolling stock is used in
```
leovalais: Has to be updated

editoast/src/modelsv2/rolling_stock_model/schedules_from_rolling_stock.rs
```diff
@@ -0,0 +1,70 @@
...
+use diesel_async::RunQueryDsl;
+use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
+
+use crate::error::Result;
+use crate::modelsv2::rolling_stock_model::RollingStockModel;
+use crate::tables::{project, rolling_stock, scenario_v2, study, train_schedule_v2};
+use editoast_models::db_connection_pool::DbConnectionV2;
+
+#[derive(Debug, Serialize, Deserialize, ToSchema, PartialEq, Eq, PartialOrd, Ord)]
```
leovalais:
```suggestion
#[derive(Debug, Serialize, ToSchema)]
#[cfg_attr(test, derive(PartialEq, Eq, PartialOrd, Ord, Deserialize)]
```

editoast/src/modelsv2/rolling_stock_model/schedules_from_rolling_stock.rs
```diff
@@ -0,0 +1,70 @@
...
+    #[schema(inline, required)]
+    pub project_name: String,
+    #[schema(inline, required)]
+    pub study_id: i64,
+    #[schema(inline, required)]
+    pub study_name: String,
+    #[schema(inline, required)]
+    pub scenario_id: i64,
+    #[schema(inline, required)]
+    pub scenario_name: String,
```
leovalais: You don't need any `#[schema(inline, required)]` in this struct:

* `inline` instructs utoipa to inline a sub-schema instead of adding a `$ref` to a `component/schemas` schema. Ints, strings, and other primitive types are always inlined.
* `required` is primarily useful when we know the key of an `Option` is always in the response. Basically, the TypeScript type of an `Option<String>` is:
  - without `#[schema(inline)]`: `foo: string | null | undefined`
  - with `#[schema(inline)]`: `foo: string | null`

editoast/src/views/rolling_stocks/mod.rs
```diff
@@ -556,6 +558,34 @@ async fn create_livery(
     Ok(Json(rolling_stock_livery))
 }
 
+/// List the train schedules a rolling stock is used in
+#[utoipa::path(
+    get, path = "",
+    tag = "rolling_stock",
+    params(RollingStockIdParam),
+    responses(
+        (status = 200, description = "A list of the associated scenarios and their respective studies and projects.", body = inline(Vec<ScenarioReference>)),
```
leovalais:
```suggestion
        (status = 200, description = "A list of the associated scenarios and their respective studies and projects.", body =Vec<ScenarioReference>),
```

Likewise

editoast/src/modelsv2/rolling_stock_model/schedules_from_rolling_stock.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub async fn get_usage(&self, conn: &mut DbConnectionV2) -> Result<Vec<ScenarioReference>> {
+        let schedules: Vec<_> = train_schedule_v2::table
+            .inner_join(
+                rolling_stock::table
+                    .on(train_schedule_v2::rolling_stock_name.eq(rolling_stock::name)),
+            )
+            .inner_join(
+                (scenario_v2::table
+                    .on(scenario_v2::timetable_id.eq(train_schedule_v2::timetable_id)))
+                .inner_join((study::table).inner_join(project::table)),
```
leovalais:
```suggestion
                .inner_join(study::table.inner_join(project::table)),
```

editoast/src/modelsv2/rolling_stock_model/schedules_from_rolling_stock.rs
```diff
@@ -0,0 +1,70 @@
...
+                study::id,
+                study::name,
+                scenario_v2::id,
+                scenario_v2::name,
+            ))
+            .filter(rolling_stock::id.eq(self.id))
+            .filter(train_schedule_v2::id.is_not_null())
+            .load::<SchedulesFromRollingStock>(conn)
+            .await?;
+        let schedules = schedules.into_iter().map(|flat| flat.into()).collect();
```
leovalais:
```suggestion
        let schedules = schedules.into_iter().map_into().collect();
```

leovalais (changes requested):
A few last changes. Otherwise tested and it works great! 😁

leovalais:
Sad to see the result of our `Option` bikeshed go away though 😞

editoast/src/modelsv2/rolling_stock_model/schedules_from_rolling_stock.rs
Sh099078: I added PartialEq / Eq / PartialOrd / Ord to be able to use `Vec::sort` in a test assert instead of adding some bloating in the test to do the same thing manually. Would you prefer to the test to be refactored and the Ord related traits to be removed ?

leovalais: The [`cfg_attr`](https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg_attr-attribute) attribute allows conditional compilation. Since these traits (`PartialEq, Eq, PartialOrd, Ord, Deserialize `) are only used in tests (i.e. if `cfg!(test)`), we can omit them when we're not testing.

Maybe I missed something but I don't think you'll have to change the tests to change these `derive(..)`s.

editoast/openapi.yaml
leovalais: It doesn't list train schedules.

leovalais (approved):
LGTM, thank you (I unresolved a comment).

> [!NOTE]
> You may want to squash your commits.

editoast/openapi.yaml
Sh099078: My bad, fixed it

PR #8567, opened 2024-08-23
front: drop trainIdsToFetch state

emersion:
We don't actually need a state here: all setTrainIdsToFetch() calls were overwriting any previously stored value.

The state makes the whole thing harder to reason about, because a state update will (asynchronously) invoke any useEffect() that depends on that state, so it's harder to figure out what exact side effects will be caused by setting the state. A simple callback is easier to follow.

No behavior change intended.

### DIFF
--- a/front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
+++ b/front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
@@ -38,7 +38,6 @@ const useLazyLoadTrains = ({ infraId, path, trainSchedules }: UseLazyLoadTrainsP
   const [trainScheduleSummariesById, setTrainScheduleSummariesById] = useState<
     Map<number, TrainScheduleWithDetails>
   >(new Map());
-  const [trainIdsToProject, setTrainIdsToProject] = useState<number[]>([]);
   const [allTrainsLoaded, setAllTrainsLoaded] = useState(false);
 
   const [postTrainScheduleSimulationSummary] =
@@ -47,13 +46,11 @@ const useLazyLoadTrains = ({ infraId, path, trainSchedules }: UseLazyLoadTrainsP
   const { data: { results: rollingStocks } = { results: [] } } =
     osrdEditoastApi.endpoints.getLightRollingStock.useQuery({ pageSize: 1000 });
 
-  const { projectedTrainsById, setProjectedTrainsById } = useLazyProjectTrains({
+  const { projectedTrainsById, setProjectedTrainsById, projectTrains } = useLazyProjectTrains({
     infraId,
-    trainIdsToProject,
     path,
     trainSchedules,
     moreTrainsToCome: !allTrainsLoaded,
-    setTrainIdsToProject,
   });
 
   const loadTrains = async (_trainSchedules: TrainScheduleResult[]) => {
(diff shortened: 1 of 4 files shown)


front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
```diff
@@ -66,51 +58,44 @@ const useLazyLoadTrains = ({
...
-
-      setTrainIdsToFetch([]);
-      setAllTrainsLoaded(true);
-    };
-
-    if (infraId && trainIdsToFetch && trainIdsToFetch.length > 0) {
-      getTrainScheduleSummaries(infraId, trainIdsToFetch);
+  const loadTrainIds = async (trainIds: number[]) => {
+    setAllTrainsLoaded(false);
+
```
anisometropie: if we have no scenario we don’t get an infraId from `useScenarioData`. 
in that case we need not do anything and just return early.

```suggestion
  const loadTrainIds = async (trainIds: number[]) => {
    if (!infraId) {
      return;
    }

File: front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts [hunk 2]
```diff
@@ -66,51 +58,44 @@ const useLazyLoadTrains = ({
...
+    for (let i = 0; i < trainIds.length; i += BATCH_SIZE) {
+      const packageToFetch = getBatchPackage(i, trainIds, BATCH_SIZE);
+
+      const rawSummaries = await postV2TrainScheduleSimulationSummary({
+        body: {
+          infra_id: infraId!,
```
anisometropie:
```suggestion
          infra_id: infraId,
```

anisometropie (changes requested):

clarani (commented):
When I try to add trains in a timetable which already have some, the new trains are not refetched :/

emersion:
There is a race between the summaries endpoint and the train schedule endpoint. If `trainScheduleSummaries` is populated with the new data before `trainSchedules` is updated, `formatTrainScheduleSummaries()` will not include the new train.

emersion:
I've fixed the race. It effectively reverts 6d020d86930318a09b9ca7b0d947ce1678055e65.

front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
emersion: Is this something which can really happen? Can we have a list of trains to load from a timetable without an infra?

If it can't happen, I'd rather fail hard if we hit this case, so that we can debug the root cause instead of silently doing nothing.

emersion:
Couldn't resist the urge to add in one more refactor commit.

front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
anisometropie: alright, I think I’m okay to consider we should have a studyId defined at this point

anisometropie (commented):
unfortunately it breaks the fix I did. The assert fails again in `isScheduledPointsNotHonored`, when you modify a train by adding a path step

Screenshot_20240910_104846

emersion:
Superseded by https://github.com/OpenRailAssociation/osrd/pull/11390

PR #8568, opened 2024-08-23
front: fix incorrect time in conflict card

### LINKED ISSUE #8548: Conflict times are not correct
### What happened?

I have a conflict here between 5h and 6h:
image

But my freight train leaves between 6 and 7:
image

Trains are always on time, so i'm guessing the conflict should follow that time

1. create a scenario with conflicting trains

Recette (SNCF)

Google Chrome

c885d29

RomainValls:
close #8548 

In formatToIsoDate function, we now convert the date with dayjs first, then apply the userTimeZone. 
It avoids double application of timezone offsets.

### DIFF
--- a/editoast/src/views/electrical_profiles.rs
+++ b/editoast/src/views/electrical_profiles.rs
@@ -6,14 +6,18 @@ use axum::extract::Query;
 use axum::extract::State;
 use axum::http::StatusCode;
 use axum::response::IntoResponse;
+use axum::Extension;
+use editoast_authz::BuiltinRole;
 use editoast_derive::EditoastError;
-use serde::Deserialize;
-use thiserror::Error;
-use utoipa::IntoParams;
 use editoast_models::DbConnectionPoolV2;
 use editoast_schemas::infra::ElectricalProfileSetData;
 use editoast_schemas::infra::LevelValues;
+use serde::Deserialize;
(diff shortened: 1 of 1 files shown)


front/src/utils/date.ts
```diff
@@ -82,8 +82,14 @@ export const dateTimeToIso = (inputDateTime: string) => {
...
   const userTimeZone = dayjs.tz.guess(); // Format : 'Europe/Paris'
+
+  let parsedDate = dayjs(date);
+
+  parsedDate = parsedDate.tz(userTimeZone);
+
   const format = formatDate ? 'D/MM/YYYY HH:mm:ss' : '';
-  return dayjs.tz(date, userTimeZone).format(format);
+
+  return parsedDate.format(format);
```
SharglutDev: I think you can do it shorter like that (and removed parsedDate logic) : 

```suggestion
  return dayjs(date).tz(userTimeZone).format(format);
```

SharglutDev (commented):
Thank you for the fix. It works great but left one suggestion so the code can be shorter.

front/src/utils/date.ts
RomainValls: Indeed, done 👍 !

SharglutDev (approved):
Lgtm :)

Issue #8178, opened 2024-07-25
Tracking issue • Post-axum migration improvements


leovalais:
Tracking issues for problems in the `axum` migration PR #8047.

We decided to solve these problems afterwards in other PRs as #8047 is large and conflicts with everyone. It also blocks #8171.

- [ ] Design a replacement for the `routes!` macro
- [ ] Naming and homogeneity pass (good first issue)
- [ ] [Implementing our own otel middleware](https://github.com/OpenRailAssociation/osrd/pull/8047#discussion_r1684411285) to stop using `axum-tracing-opentelemetry` and check the need for `TraceLayer`
- [ ] [evaluating `axum_typed_multipart` and test liveries import](https://github.com/OpenRailAssociation/osrd/pull/8047#discussion_r1687963958) (good first issue)
- [x] #8673
- [ ] Rename `document_key` to `document_id` in `/document` APIs (good first issue)
- [x] #8674

PR #8598, opened 2024-08-28
editoast: hide 'datadog' and 'opentelemetry' behind feature flags

woshilapin:
The goal is to reduce the default compile time for developers.
Datadog is mostly used in specific deployment, not necessary to run any
test or any useful runtime functionalities unless communicating with
a Datadog agent.

Some summary about the total number of dependencies depending on which
feature is activated:

|   activated features   | number of dependencies |
| ---------------------- | ---------------------- |
|         (None)         |           566          |
|         datadog        |           582          |
|      opentelemetry     |           596          |
| datadog, opentelemetry |           604          |

### DIFF
--- a/editoast/src/main.rs
+++ b/editoast/src/main.rs
@@ -38,7 +38,6 @@ use modelsv2::{
     train_schedule::TrainScheduleChangeset,
 };
 use modelsv2::{Changeset, RollingStockModel};
-use opentelemetry_datadog::DatadogPropagator;
 use opentelemetry_sdk::propagation::TraceContextPropagator;
 use tower::Layer as _;
 use tower_http::cors::{Any, CorsLayer};
@@ -109,6 +108,7 @@ fn init_tracing(mode: EditoastMode, telemetry_config: &client::TelemetryConfig) 
     // https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/index.html#runtime-configuration-with-layers
     let telemetry_layer = match telemetry_config.telemetry_kind {
         client::TelemetryKind::None => None,
+        #[cfg(feature = "datadog")]
         client::TelemetryKind::Datadog => {
             let datadog_tracer = opentelemetry_datadog::new_pipeline()
                 .with_service_name(telemetry_config.service_name.as_str())
@@ -118,7 +118,9 @@ fn init_tracing(mode: EditoastMode, telemetry_config: &client::TelemetryConfig) 
             let layer = tracing_opentelemetry::layer()
                 .with_tracer(datadog_tracer)
                 .boxed();
-            opentelemetry::global::set_text_map_propagator(DatadogPropagator::default());
+            opentelemetry::global::set_text_map_propagator(
+                opentelemetry_datadog::DatadogPropagator::default(),
+            );
             Some(layer)
         }
         client::TelemetryKind::Opentelemetry => {
--- a/editoast/src/main.rs
+++ b/editoast/src/main.rs
@@ -17,7 +17,7 @@ use crate::modelsv2::Infra;
 use crate::views::OpenApiRoot;
 use axum::extract::FromRef;
 use axum::{Router, ServiceExt};
-use axum_tracing_opentelemetry::middleware::{OtelAxumLayer, OtelInResponseLayer};
+use axum_tracing_opentelemetry::middleware::OtelAxumLayer;
 use chashmap::CHashMap;
 use clap::Parser;
 use client::{
@@ -431,7 +431,6 @@ async fn runserver(
             app_state.clone(),
             authorizer_middleware,
         ))
-        .layer(OtelInResponseLayer)
         .layer(OtelAxumLayer::default())
         .layer(request_payload_limit)
         .layer(cors)
--- a/editoast/Cargo.toml
+++ b/editoast/Cargo.toml
@@ -123,7 +123,7 @@ opentelemetry = { version = "0.23.0", default-features = false, features = [
 opentelemetry-datadog = { version = "0.11.0", default-features = false, features = [
   "intern-std",
   "reqwest-client",
-] }
+], optional = true }
 opentelemetry-otlp = { version = "0.16.0", default-features = false, features = [
   "grpc-tonic",
   "trace",
@@ -188,6 +188,9 @@ pretty_assertions.workspace = true
 rstest.workspace = true
 serial_test = "3.1.1"
 tempfile.workspace = true
+
+[features]
+datadog = ["dep:opentelemetry-datadog"]
 
 [lints]
 workspace = true
--- a/editoast/Cargo.toml
+++ b/editoast/Cargo.toml
@@ -64,7 +64,10 @@ tempfile = "3.12.0"
 thiserror = "1.0.63"
 tokio = { version = "1.40.0", features = ["fs", "macros", "rt-multi-thread"] }
 tokio-postgres = "0.7.11"
-tracing = { version = "0.1.40", default-features = false, features = ["log"] }
+tracing = { version = "0.1.40", default-features = false, features = [
+  "attributes",
+  "log",
+] }
 url = { version = "2.5.2", features = ["serde"] }
 utoipa = { version = "4.2.3", features = ["chrono", "uuid"] }
 uuid = { version = "1.10.0", features = ["serde", "v4"] }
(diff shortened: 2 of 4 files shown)


editoast/src/main.rs
```diff
@@ -426,14 +426,17 @@ async fn runserver(
...
         .route_layer(axum::middleware::from_fn_with_state(
             app_state.clone(),
             authorizer_middleware,
-        ))
-        .layer(OtelInResponseLayer)
-        .layer(OtelAxumLayer::default())
+        ));
+    #[cfg(any(feature = "opentelemetry", feature = "datadog"))]
+    let router = router
+        .layer(axum_tracing_opentelemetry::middleware::OtelInResponseLayer)
```
leovalais: I recall we guessed that this layer might not be necessary since we don't need open telemetry headers in responses. Did that change?

leovalais (approved):
LGTM (editoast, not tested)

flomonster:
Do you have stats on the compile time with/without these dependencies? I feel it won't change much since most of the compile time is spent on osrd (not the dependencies). 

We might add complexity without real gain. From my point of view, the pain comes from the rustanalyzer runtime rather than the build time.

editoast/src/main.rs
woshilapin: It seems it was not cleaned-up during the [big `axum` PR](https://github.com/OpenRailAssociation/osrd/pull/8047). I thought it was tracked in the [follow-up ticket](https://github.com/OpenRailAssociation/osrd/issues/8178), but it's not. I'm just going to do it in that PR (separate commit, see dbfcb062b).

leovalais: You're right, but I figured that since the line is in the diff, we might as well delete it now 🤷

woshilapin:
> Do you have stats on the compile time with/without these dependencies? I feel it won't change much since most of the compile time is spent on osrd (not the dependencies).
> 
> We might add complexity without real gain. From my point of view, the pain comes from the rustanalyzer runtime rather than the build time.

I agree that for rust-analyzer it might not be a big impact. With these 2 benchmarks, it shows no effect (if you check the entire workspace or not).

```
❯ hyperfine --warmup 1 --parameter-list branch dev,wsl/editoast/build/no-default-datadog --setup 'git switch {branch}' --prepare "fd '.*\.rs' --exec touch '{}' ';'" 'cargo check --workspace'

Benchmark 1: cargo check --workspace (branch = dev)
  Time (mean ± σ):     22.991 s ±  3.734 s    [User: 23.251 s, System: 1.138 s]
  Range (min … max):   18.889 s … 31.894 s    10 runs

Benchmark 2: cargo check --workspace (branch = wsl/editoast/build/no-default-datadog)
  Time (mean ± σ):     23.035 s ±  3.541 s    [User: 23.241 s, System: 1.091 s]
  Range (min … max):   18.356 s … 28.058 s    10 runs

Summary
  cargo check --workspace (branch = dev) ran
    1.00 ± 0.22 times faster than cargo check --workspace (branch = wsl/editoast/build/no-default-datadog)

❯ hyperfine --warmup 1 --parameter-list branch dev,wsl/editoast/build/no-default-datadog --setup 'git switch {branch}' --prepare "fd '.*\.rs' --exec touch '{}' ';'" 'cargo check'

Benchmark 1: cargo check (branch = dev)
  Time (mean ± σ):     22.497 s ±  2.661 s    [User: 22.300 s, System: 1.032 s]
  Range (min … max):   18.330 s … 26.341 s    10 runs

Benchmark 2: cargo check (branch = wsl/editoast/build/no-default-datadog)
  Time (mean ± σ):     21.864 s ±  2.106 s    [User: 21.662 s, System: 1.033 s]
  Range (min … max):   17.721 s … 24.563 s    10 runs

Summary
  cargo check (branch = wsl/editoast/build/no-default-datadog) ran
    1.03 ± 0.16 times faster than cargo check (branch = dev)
```

But rust-analyzer is not the only case where this might matters. For example, we often run (and therefore compile) the tests... which means there is linking happening. Therefore, less dependencies might be important in this case. Here are the numbers (notice the `--no-run` in order to really measure the compile time).

```
 ❯ hyperfine --warmup 1 --parameter-list branch dev,wsl/editoast/build/no-default-datadog --setup 'git switch {branch}' --prepare "fd '.*\.rs' --exec touch '{}' ';'" 'cargo test --no-run'

Benchmark 1: cargo test --no-run (branch = dev)
  Time (mean ± σ):     46.629 s ±  2.917 s    [User: 54.822 s, System: 2.596 s]
  Range (min … max):   42.925 s … 50.927 s    10 runs

Benchmark 2: cargo test --no-run (branch = wsl/editoast/build/no-default-datadog)
  Time (mean ± σ):     44.674 s ±  4.304 s    [User: 52.171 s, System: 2.437 s]
  Range (min … max):   36.253 s … 53.162 s    10 runs

Summary
  cargo test --no-run (branch = wsl/editoast/build/no-default-datadog) ran
    1.04 ± 0.12 times faster than cargo test --no-run (branch = dev)
```

We see there is no significant improvement (the difference in time is smaller than the variance).

⚠️ UPDATED [31/08/2024]: all benchmarks were wrongly executed on an outdated `dev` branch. The numbers above have been updated with a better benchmark.

leovalais:
My speedup factor is less surprising, but 3% just for the telemetry still seems quite a lot tbh.
 
```
❯ hyperfine --warmup 1 --parameter-list branch 'dev,wsl/editoast/build/no-default-datadog' --setup 'git switch {branch}' --prepare "fd '.*\.rs' --exec touch '{}' ';'" 'cargo test --no-run'

Benchmark 1: cargo test --no-run (branch = dev)
  Time (mean ± σ):     23.636 s ±  1.564 s    [User: 42.207 s, System: 6.771 s]
  Range (min … max):   19.469 s … 25.034 s    10 runs

Benchmark 2: cargo test --no-run (branch = wsl/editoast/build/no-default-datadog)
  Time (mean ± σ):     23.000 s ±  0.493 s    [User: 39.789 s, System: 6.363 s]
  Range (min … max):   22.442 s … 24.086 s    10 runs

Summary
  cargo test --no-run (branch = wsl/editoast/build/no-default-datadog) ran
    1.03 ± 0.07 times faster than cargo test --no-run (branch = dev)
```

woshilapin:
I updated [comment](https://github.com/OpenRailAssociation/osrd/pull/8598#issuecomment-2321098421) since the result were plainly wrong. In the end, there is no situation in which removing some dependencies has an impact on the performance. Therefore, the [comment of @flomonster](https://github.com/OpenRailAssociation/osrd/pull/8598#issuecomment-2320620133) is completely relevant.

Even the [results obtained by @leovalais](https://github.com/OpenRailAssociation/osrd/pull/8598#issuecomment-2321622753) don't show a significant result, the variance is greater than the time difference. Just to be sure, I increased the number of calculation in a benchmark and even remove compilation of sub-crates from the equation (just `touch` file in `editoast/src` instead of all the rust files in sub-crates). See the results, it's a plain 0% improvement.

```
❯ hyperfine --min-runs 42 --warmup 1 --parameter-list branch dev,wsl/editoast/build/no-default-datadog --setup 'git switch {branch}' --prepare "fd '.*\.rs' src/ --exec touch '{}' ';'" 'cargo test --no-run'

Benchmark 1: cargo test --no-run (branch = dev)
  Time (mean ± σ):     44.930 s ±  4.062 s    [User: 52.078 s, System: 1.966 s]
  Range (min … max):   33.290 s … 53.586 s    42 runs

Benchmark 2: cargo test --no-run (branch = wsl/editoast/build/no-default-datadog)
  Time (mean ± σ):     44.949 s ±  4.553 s    [User: 51.683 s, System: 1.960 s]
  Range (min … max):   32.626 s … 58.047 s    42 runs

Summary
  cargo test --no-run (branch = dev) ran
    1.00 ± 0.14 times faster than cargo test --no-run (branch = wsl/editoast/build/no-default-datadog)
``` 

Now, I'm ok to drop the `opentelemetry` feature as it brings significant complexity for no improvement. However, I'd be ok to discuss keeping the `datadog` feature for the following 2 reasons:
- the commit to make Datadog optional is much more simple than for OpenTelemetry (see for yourself 7760192cfc9decc9f098e770f454cdaf095c61ce).
- Datadog being some non-standard proprietary solution, only needed in some specific deployment situations, I'd say it reasonable to deactivate it by default in an open-source project

I'm ok either way. Note that this PR still also contains 2 small fix commits anyway so even if we remove the entire featurization of Datadog and OpenTelemetry, we might want to merge it still (with an edit of the title maybe 🤷).

leovalais:
Too bad it doesn't lead to any improvement. Agreed about the datadog feature.

flomonster:
Same as @leovalais. Another reason to keep a feature flag for `datadog` is to make it easier to remove datadog soon (at least I hope so).

PR #8622, opened 2024-08-29
template: bugfix in bug report template

axrolld:
Sorry about that !

PR #8626, opened 2024-08-29
github: fix bug report template

emersion:
GitHub complains with "default must be of type Integer".

Fixes: 83f8fb54d73e ("template: remove dropdown values and add default values")


Castavo:
Duplicate of #8622

PR #8630, opened 2024-08-29
front: collapse scenario description

emersion:
If the scenario description is empty, or shorter than 4 lines, collapse it instead of eating up screen real estate.

https://github.com/OpenRailAssociation/osrd/pull/8028 made this possible.


Caracol3 (approved):
LGTM (tested)

kmer2016 (approved):
LGTM and tested

PR #8631, opened 2024-08-29
front: make ch check more robust in updatePathStepsFromOperationalPoints()

emersion:
The PathItemLocation type definition allows for null secondary_code such as:

    { uic: 42, secondary_code: null }

and this is equivalent to:

    { uic: 42 }

Make the logic more future-proof by checking whether the property is truthy instead of checking if step has a secondary_code property.


SharglutDev (approved):
Lgtm ! Not related to your PR but I just noticed this `condition` const, I think the term is a bit weird.

anisometropie (approved):

PR #8632, opened 2024-08-29
front: fix dropped ch when running pathfinding

emersion:
The ch was taken into account for UICs but not for trigrams.


RomainValls (approved):
LGTM

SharglutDev (approved):
Lgtm

PR #8633, opened 2024-08-29
front: drop PathStep.ch

emersion:
This information is duplicated in PathItemLocation.secondary_code. This has several downsides.

First, the field secondary_code is inherited in the PathStep type event if it's never set (we omit() it when receiving the train schedule path from the backend). As a result writing "pathStep.secondary_code" in code is valid even if it would never contain anything. Additionally, the field needs to be converted back and forth from/to ch when sending HTTP requests.

Second, the field is always defined even for path steps for which a ch doesn't make sense. For instance, if the path step is an operational_point or is a TrackOffset, the ch property would still be defined (and always be empty).

All of this makes it easy to trip over and use "ch" or "secondary_code" in contexts where it doesn't make sense. While this is easily discovered in testing when initially writing a feature, any subsequent refactoring may subtly break the rest of the code.

To fix this, drop the "ch" field and only use the auto-generated "secondary_code" field. This forces any code making use of ch to verify (with the "in" keyword) that accessing the field makes sense.

Depends on:

- https://github.com/OpenRailAssociation/osrd/pull/9756
- https://github.com/OpenRailAssociation/osrd/pull/9761
- https://github.com/OpenRailAssociation/osrd/pull/9763
- https://github.com/OpenRailAssociation/osrd/pull/9764


theocrsb (approved):
LGTM

clarani (approved):
LGTM :white_check_mark:

PR #8637, opened 2024-08-30
front: enable source maps in production

emersion:
Source maps are separate .map files which contain debug information and allows user agents to display a proper stack trace on crash.

vite's defaults are tuned for enterprise™ proprietary© code: source map files contain function and variable names thus they are Super Secret and shouldn't get exposed. We are an open-source project so we don't need to hide anything.

This patch should make it easier to debug crashes in production.

### DIFF
--- a/front/vite.config.mts
+++ b/front/vite.config.mts
@@ -48,6 +48,7 @@ export default defineConfig(({ mode }) => {
     ],
     build: {
       outDir: 'build',
+      sourcemap: true,
     },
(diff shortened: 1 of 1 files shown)


flomonster (approved):

woshilapin (approved):

Uriel-Sautron (approved):
Thank's !!!

clarani (approved):
LGTM

PR #8638, opened 2024-08-30
Add maintainers and governance files

flomonster:


loic-hamelin (approved):
lgtm

GOVERNANCE.md
```diff
@@ -0,0 +1,41 @@
...
+
+**3.5. Written procedures**. This governance document and other materials documenting this project's development process shall be available to any interested person.
+
+## 4. No Confidentiality.
+
+Information disclosed in connection with any Project activity, including but not limited to meetings, contributions, and submissions, is not confidential, regardless of any markings or statements to the contrary.
+
+## 5. Amendments.
+
+Amendments to this governance policy may be made by affirmative vote of 2/3 of all Maintainers, with approval by project admins.
```
SergeCroise, line 37: just a little question:
why only 2/3 of all...,
why not all Maintainers (or e.g., 83%, 89%, or 97% of all) ?

emersion (approved):

GOVERNANCE.md
emersion: This comes from MVG: https://github.com/github/MVG/blob/main/project-docs/GOVERNANCE.md#6-amendments

flomonster: It comes from the MVG [here](https://github.com/github/MVG/blob/main/project-docs/GOVERNANCE.md). The idea is to allow modifications a small group of maintainers agree without having to check with everyone.

SergeCroise: Ok, 2/3 is a minimum according to MVG. Do not know how important amendments or governance is, but the idea is:
for the first version of this governance file, we could use 
**97% of all Maintainers**,
and later (as needed) if all Maintainers agree, we could reduce the rate to **67%** (2/3), or e.g.,
71%, 73%, 79%, 83%,...

```suggestion
Amendments to this governance policy may be made by affirmative vote of 97% of all Maintainers, with approval by project admins.
```

SergeCroise (approved):

GOVERNANCE.md
flomonster: I don't think that's a critical point. I prefer to have a static reasonable percentage rather than changing it when needed.

bloussou (approved):
Thanks

PR #8640, opened 2024-08-30
front: upgrade license-checker-rseidelsohn

emersion:
Test files have been dropped from this package's tarball, so the size of the on-disk node_modules/license-checker-rseidelsohn directory goes from 148.1 MiB down to 3.1 MiB.

Also move to a dev dependency since it's only used at build-time.


kmer2016 (approved):
LGTM

Akctarus (approved):
tested, LGTM

emersion:
Blocked on https://github.com/RSeidelsohn/license-checker-rseidelsohn/pull/119

PR #8644, opened 2024-08-30
Fix actions make bake metadata endgroup

flomonster:
Fix this: https://github.com/OpenRailAssociation/osrd/actions/runs/10618838075/job/29435015961#step:3:574

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,6 +30,7 @@ jobs:
 
           echo ::group::Bake metadata
           .github/scripts/bake-metadata.py | tee bake-metadata.json
+          echo
           echo ::endgroup::
 
       - name: Set up Docker Buildx


.github/workflows/release.yml
```diff
@@ -30,7 +30,7 @@ jobs:
 
           echo ::group::Bake metadata
           .github/scripts/bake-metadata.py | tee bake-metadata.json
-          echo ::endgroup::
+          echo '\n::endgroup::'
```
emersion:
```
$ echo '\n abc'
\n abc
```

Maybe adding a `echo ''` before would be the simplest way to fix it. (Alternatively, `echo -e` interprets `\n`)

Khoyo (approved):

emersion (approved):

PR #8653, opened 2024-08-30
rvs/front-fix-incorrect-time-in-conflict-cards

### LINKED ISSUE #8548: Conflict times are not correct

RomainValls:
close #8548


SharglutDev (approved):

Uriel-Sautron (approved):
LGTM, tested !!

PR #8655, opened 2024-08-30
editoast: move tables.rs into editoast_models

leovalais:
We'll need to do that eventually. I might need it now to experiment with applying migrations programmatically (I'm working on test isolation).


flomonster (approved):
LGTM.

> [!IMPORTANT]
> Not tested just reviewed.

PR #8656, opened 2024-08-30
front: add stop type in stdcm

### LINKED ISSUE #8681: Add selector for stop type in STDCM
The selector must appear after a PR has been selected.

- [x] Each milestone (excluding origin and destination) has a selector for the type of stop. Single choice among the 3 following choices :
    - Service stop
    - Passage (`stop_for = None`)
    - Driver lift
- [x] The “Passage stop” type is selected by default when a PR is added.
- [x] For all intermediate waypoints, the user cannot set the time for the waypoint: “As soon as possible” is selected.
- [x] If the user selects the “Service stop” or “Driver switch” type, the field for entering the duration of the stop is displayed (mandatory).
- [x] If “Driver switch” is selected, the default stop time is set to 3 minutes. This time can be modified by the user, with a warning if it falls below 3mn.
- [x] In case of “Service stop” type, the stop time is not pre-filled.

Image

Akctarus:
closes #8681


Caracol3:
Nice work ! :-) 
I spotted one thing. When you  have selected 'service stop' and then 'driver switch' or the opposite, there is no reset of the duration. I think we should reajust the duration of the default stop when changing a type.

front/src/applications/stdcmV2/components/StdcmVias.tsx
SharglutDev, line 1: Not coming from your PR but there is a warning for a missing `key`. We probably should convert the `fragment` l.88 to a `div` and move the key l.97 inside

front/public/locales/en/stdcm.json
```diff
@@ -77,8 +77,15 @@
     "origin": "Origin",
     "preciseTime": "precise time",
     "stopFor": "Minimum stop time",
+    "stopType": {
+      "driverSwitch": "driver switch",
+      "passageTime":"passage time",
```
```suggestion
      "passageTime": "passage time",
```

front/src/styles/scss/applications/stdcmV2/_card.scss
```diff
@@ -49,21 +49,29 @@
   }
 
   .stdcm-v2-card__body {
-    padding: 14px 21px 21px 21px;
+    padding: 14px 16px 21px 16px;
```
SharglutDev: Padding seems wrong here for the default card (see mockup below) and we should use rem. Maybe there should a specific class for this component. I'm fine if it's done in another PR as it doesn't concern yours but it would be great if you could create the issue and add it to the Meta interface stdcm in confidentials :)

Capture d’écran 2024-09-24 à 22 44 16

front/src/styles/scss/applications/stdcmV2/_card.scss
SharglutDev, line 1: I know it wasn't strictly follow in stdcm but we try to use `rem` instead of `px`. If you could convert at least the one you changed here :) (if you don't have it there is a nice vs code extension : "px to rem" to help you)

front/src/styles/scss/applications/stdcmV2/_card.scss
```diff
@@ -49,21 +49,29 @@
...
     }
 
     .feed-back {
-      padding-left: 0rem;
       padding-right: 0.3rem;
     }
 
+    .selector {
+      padding-left: 3px;
+      padding-right: 4px;
```
SharglutDev: You can use `padding-inline` instead of these too

front/src/applications/stdcmV2/components/StdcmOpSchedule.tsx
```diff
@@ -85,50 +85,46 @@ const StdcmOpSchedule = ({
...
+            onArrivalTypeChange(e as ArrivalTimeTypes);
+          }
+        }}
+        options={['preciseTime', 'asSoonAsPossible']}
+        getOptionLabel={(option) => t(`trainPath.${option}`)}
+        getOptionValue={(option) => option}
+        disabled={disabled}
+      />
       {opScheduleTimeType === 'preciseTime' && (
         <div className="d-flex">
```
SharglutDev, line 95: Spacing between the 3 elements doesn't seem to respect to respect the mockup (16px). If it was like that before your changes, could you create an issue add it to the meta in confidential ?

Capture d’écran 2024-09-24 à 23 10 07

front/src/applications/stdcmV2/components/StdcmOpSchedule.tsx
```diff
@@ -85,50 +85,46 @@ const StdcmOpSchedule = ({
...
-              onArrivalTypeChange(e as ArrivalTimeTypes);
-            }
-          }}
-          options={['preciseTime', 'asSoonAsPossible']}
-          getOptionLabel={(option) => t(`trainPath.${option}`)}
-          getOptionValue={(option) => option}
-          disabled={disabled}
-        />
-      </div>
+    <div className="d-flex flex-column selector">
```
SharglutDev, line 80: The fact that we need to use this new class to adjust element position is not very good and maintainable.

I'm not sure why we have default padding in select and input elements but it's really difficult to integrate them in a complex UI. @kmer2016 do you know the reason ? Maybe we should open an issue for that.

front/src/styles/scss/applications/stdcmV2/_card.scss
```diff
@@ -49,21 +49,29 @@
...
-    padding: 14px 21px 21px 21px;
+    padding: 14px 16px 21px 16px;
     // TODO: remove this style after ui-core component is implemented
     .suggestions > div:first-child {
-      padding-left: 0;
+      padding-left: 8px;
     }
     .suggestions * label,
     .stdcm-v2-via-stop-for * label {
       margin-bottom: 0;
```
SharglutDev, line 71: This should be 2px (in rem)

front/src/applications/stdcmV2/components/StdcmInputVia.tsx
```diff
@@ -20,27 +24,68 @@ const StdcmInputVia = ({
...
+    if (isPassageTime && pathStepStopTime !== '0') {
+      newStopTime = '0';
+    } else if (isDriverSwitch) {
+      const stopTimeValue = Number(pathStepStopTime);
+      if (stopTimeValue === 0) {
+        newStopTime = '3';
+      } else if (stopTimeValue < 3) {
+        warning = true;
+      }
+    } else if (isServiceStop && pathStepStopTime <= '0') {
```
SharglutDev: Why do we check inferior to 0 ?

front/src/applications/stdcmV2/components/StdcmInputVia.tsx
```diff
@@ -20,27 +24,68 @@ const StdcmInputVia = ({
...
+
+    if (isPassageTime && pathStepStopTime !== '0') {
+      newStopTime = '0';
+    } else if (isDriverSwitch) {
+      const stopTimeValue = Number(pathStepStopTime);
+      if (stopTimeValue === 0) {
+        newStopTime = '3';
+      } else if (stopTimeValue < 3) {
+        warning = true;
+      }
```
SharglutDev: Maybe we can avoid one nested if with we config warning like that at the beginning

```suggestion
   const warning = isDriverSwitch && stopTimeValue < 3
```

front/src/applications/stdcmV2/components/StdcmInputVia.tsx
```diff
@@ -20,27 +24,68 @@ const StdcmInputVia = ({
     pathStep.stopFor ? `${ISO8601Duration2sec(pathStep.stopFor) / 60}` : ''
   );
 
+  const [showWarning, setShowWarning] = useState(false);
```
SharglutDev: Maybe we could rather have a const like this and avoid having a state for the warning

```suggestion
  const stopWarning = stopType === StdcmStopTypes.DRIVER_SWITCH && Number(pathStepStopTime) < 3;
```

front/src/applications/stdcmV2/components/StdcmVias.tsx
```diff
@@ -27,10 +29,19 @@ const StdcmVias = ({ disabled = false, setCurrentSimulationInputs }: StdcmConfig
...
+  );
+
   const intermediatePoints = useMemo(() => pathSteps.slice(1, -1), [pathSteps]);
 
   const updatePathStepsList = (pathStep: PathStep | null, index: number) => {
-    const newPathSteps = replaceElementAtIndex(pathSteps, index, pathStep);
+    if (!pathStep) return;
+    const updatedPathStep = {
+      ...pathStep,
+    };
```
SharglutDev: Why do we need that ?

front/src/applications/stdcmV2/components/StdcmVias.tsx
```diff
@@ -27,10 +29,19 @@ const StdcmVias = ({ disabled = false, setCurrentSimulationInputs }: StdcmConfig
     useOsrdConfActions() as StdcmConfSliceActions;
   const pathSteps = useSelector(getPathSteps);
 
+  const [stopTypes, setStopTypes] = useState<StdcmStopTypes[]>(
+    pathSteps.map(() => StdcmStopTypes.PASSAGE_TIME)
+  );
```
SharglutDev: It might be safer to group them by pathStep ids instead of their indexes in case one of the via is removed ? What do you think ?

```suggestion
  const [stopTypes, setStopTypes] = useState<Record<string, StdcmStopTypes>>(
    compact(pathSteps).reduce(
      (acc, cur) => {
        acc[cur.id] = StdcmStopTypes.PASSAGE_TIME;
        return acc;
      },
      {} as Record<string, StdcmStopTypes>
    )
  );
```

front/src/applications/stdcmV2/components/StdcmInputVia.tsx
```diff
@@ -20,27 +24,68 @@ const StdcmInputVia = ({
...
+      newStopTime = '1';
+    }
+
+    if (newStopTime !== pathStepStopTime) {
+      setPathStepStopTime(newStopTime);
+      updatePathStepStopTime(newStopTime);
+    }
+
+    setShowWarning(warning);
+  }, [stopType, pathStepStopTime]);
```
SharglutDev: Not sure it's healthy to have in the dependency array the state and update with it's setter inside, it might go through some loops

front/src/applications/stdcm/components/SimulationReportSheet.tsx
```diff
@@ -134,14 +134,18 @@ const SimulationReportSheet = ({
                   <View style={styles.convoyAndRoute.stopTableStartWidth}>
                     <TD>{t('startStop')}</TD>
                   </View>
-                  <View style={styles.convoyAndRoute.stopTableMotifWidth}>
-                    <TD>{t('motif')}</TD>
+                  <View style={styles.convoyAndRoute.stopTableStopTypeWidth}>
+                    <TD>{t('stopType')}</TD>
```
SharglutDev, line 139: The stop type should be capitalized

SharglutDev (commented):

front/public/locales/en/stdcm.json
Akctarus: done

front/src/styles/scss/applications/stdcmV2/_card.scss
Akctarus: Fixed, it's a little hard since we can't see the arrow size

front/src/styles/scss/applications/stdcmV2/_card.scss
Akctarus: ty ! done

front/src/applications/stdcmV2/components/StdcmOpSchedule.tsx
Akctarus: I agree, but do I leave it like that until I find a better solution?

front/src/styles/scss/applications/stdcmV2/_card.scss
Akctarus: We need to have a 24px padding between edge of card and input, we already have a 16px padding with another class, so 8px (0.5rem) is good

front/src/applications/stdcmV2/components/StdcmInputVia.tsx
Akctarus: fixed

front/src/applications/stdcmV2/components/StdcmInputVia.tsx
Akctarus: fixed

front/src/applications/stdcmV2/components/StdcmVias.tsx
Akctarus: an oversight, well seen, fixed

front/src/applications/stdcm/components/SimulationReportSheet.tsx
Akctarus: fixed

front/src/styles/scss/applications/stdcmV2/_card.scss
Akctarus: done !

front/src/applications/stdcmV2/components/StdcmOpSchedule.tsx
Akctarus: Moved to META  : https://github.com/OpenRailAssociation/osrd/issues/8991

front/src/applications/stdcmV2/components/StdcmVias.tsx
Akctarus: done

front/src/applications/stdcmV2/components/StdcmInputVia.tsx
Akctarus: Changed

front/src/applications/stdcmV2/components/StdcmInputVia.tsx
Akctarus: I had to keep it but I moved lot of the logic in parent file.

Caracol3:
Good job !
I have just one  thing with the behavior of the selector when we hover it after selection.
There is a big arrow instead of our icon.
image

Akctarus:
> Good job ! I have just one thing with the behavior of the selector when we hover it after selection. There is a big arrow instead of our icon. image

It has been fixed in another PR

Caracol3 (approved):
lgtm (tested)

SharglutDev (approved):
Lgtm and tested, good job ! Some UI issues remain (not from the PR), they have been added in the meta issue

PR #8659, opened 2024-08-30
front: deactivate platforms layer in editor

### LINKED ISSUE #7492: Platforms layer cannot be deactivated in the infrastructure editor
### What happened?

I tried to deactivate the platforms layer in the infrastructure editor : the platforms disappear for a very short moment (<1s) and then come back.

I expected the platforms not to be displayed.

As above

- **Browser**: Chrome
- **OS**: W11
- **Env**: Staging

staging 3f7ff32

clarani:
close #7492


SharglutDev (approved):
Lgtm and tested, good job ! Note for later : we might need to decide how we want to harmonize maps with platforms layer (on or off by default).

PR #8661, opened 2024-08-30
front: improve osrdEditoastApi to deserialiaze dates as Date and not string

### LINKED ISSUE #8660: front: modify osrdEditoastApi to receive the dates from editoast in Date and not string
### Description and goal

We should receive dates instead of string. Keeping the dates in string is very painful and causes a lot of bugs

### Acceptance criteria

.

clarani:
closes #8660

### DIFF
--- a/front/src/utils/rtk.ts
+++ b/front/src/utils/rtk.ts
@@ -1,0 +1,33 @@
+import type {
+  MutationDefinition,
+  OverrideResultType,
+  QueryDefinition,
+} from '@reduxjs/toolkit/query';
+
+export type OverrideQueryArgType<Definition, NewQueryArgType> =
+  Definition extends QueryDefinition<
+    any,
+    infer BaseQuery,
+    infer TagTypes,
+    infer ResultType,
+    infer ReducerPath
+  >
+    ? QueryDefinition<NewQueryArgType, BaseQuery, TagTypes, ResultType, ReducerPath>
+    : Definition extends MutationDefinition<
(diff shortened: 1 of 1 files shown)


front/src/utils/rtk.ts
```diff
@@ -0,0 +1,33 @@
...
+          infer ResultType,
+          infer ReducerPath
+        >
+      ? MutationDefinition<NewQueryArgType, BaseQuery, TagTypes, ResultType, ReducerPath>
+      : never;
+
+export type OverrideQueryArgResultType<Definition, NewQueryArgType, NewResultType> =
+  OverrideQueryArgType<OverrideResultType<Definition, NewResultType>, NewQueryArgType>;
+
+export type PartiallyOverride<Type, Find, Repl> = {
```
emersion, line 29: Maybe `RecursiveOverride` would be a tad better?

front/src/utils/rtk.ts
```diff
@@ -0,0 +1,33 @@
+import type {
+  MutationDefinition,
+  OverrideResultType,
+  QueryDefinition,
+} from '@reduxjs/toolkit/query';
+
+export type OverrideQueryArgType<Definition, NewQueryArgType> =
```
emersion, line 7: This handles both queries and mutation, so we could leave out `Query` from the name (would be more consistent with the existing `OverrideResultType` helper from upstream).

flomonster:
Can we close this PR?

PR #8662, opened 2024-08-30
Remove old pagination system of editoast

leovalais:
We had to wait for the removal of TSV1 and here we are.


Castavo (approved):
Looks good to me

Wadjetz (approved):

flomonster (approved):
LGTM

PR #8671, opened 2024-08-31
editoast: fix std_search_env tests for cases when the table initially contains data

### LINKED ISSUE #8574: editoast: stdcm_search_environment unit tests fail when run in a DB with an existing environment
### What happened?

When a search environment already exists in the DB the tests are run with, unit tests fail.
There's a pre-set search env in the shared dumps.

Not fail.

1. Apply the latest dump.
2. Run the tests `cargo test -- stdcm_search`
3. Contemplate the multiple failures
4. Cry

Local

Other

dev

leovalais:
closes #8574 

> [!NOTE]
> It's fine to delete everything from the table as changes are never committed.


Castavo (approved):
Yup, indeed, my bad, thanks

flomonster (approved):
LGTM

PR #8672, opened 2024-08-31
Simplify LightRollingStock management & modules reorganisation

leovalais:
Removes the `LightRollingStock` model as it can be derived from `RollingStock` (effort curves simplification).

> [!NOTE]
> The endpoint `/light_rolling_stock/name/{lolz}` is not used anywhere in the frontend. Is it ok to remove it / Is it used by another service?

> [!CAUTION]
> * `modelsv2::fixtures` references `views::rolling_stock::RollingStockForm`. Since that may change the test data, I figured a separate PR would be better for that.
> * `s/RollingStockModel/RollingStock`


Wadjetz (approved):

flomonster:
> Note
> 
> The endpoint `/light_rolling_stock/name/{lolz}` is not used anywhere in the frontend. Is it ok to remove it / Is it used by another service?

Currently, the frontend calls `/light_rolling_stock?page_size=1000`. This is why it does not use the `/light_rolling_stock/name/{lolz}` endpoint. I'm not sure if it's the expected behavior though :shrug:.

flomonster (approved):
LGTM

leovalais:
> > Note
> > The endpoint `/light_rolling_stock/name/{lolz}` is not used anywhere in the frontend. Is it ok to remove it / Is it used by another service?
> 
> Currently, the frontend calls `/light_rolling_stock?page_size=1000`. This is why it does not use the `/light_rolling_stock/name/{lolz}` endpoint. I'm not sure if it's the expected behavior though 🤷.

Okay, there's some work to do on that part of the API with the frontend then (there's already the pagination to enforce). I'll leave the endpoint there for now.

PR #8673, opened 2024-08-31
Disable `/search` `dry` parameter in non-debug builds & harmonize endpoint results

leovalais:
The `dry` parameter is useful for debugging the query being generated by the search expression. It's not necessary in production. Moreover we spotted during the `axum` migration that the `/search` endpoint sometimes returns json (search results) or text (`dry`). This PR harmonizes that.

### DIFF
--- a/editoast/openapi.yaml
+++ b/editoast/openapi.yaml
@@ -8337,7 +8337,10 @@ components:
       properties:
         dry:
           type: boolean
-          description: Whether to return the SQL query instead of executing it
+          description: |-
+            Whether to return the SQL query instead of executing it
+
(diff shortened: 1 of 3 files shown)


flomonster (commented):
Sounds good to me, but I'm not sure I understand the purpose of PR...
Can you add some context to the PR description?

leovalais:
> Sounds good to me, but I'm not sure I understand the purpose of PR... Can you add some context to the PR description?

Good point, sorry. Updated. While doing that I recalled we needed to change the response of the endpoint. This PR now also includes that.

flomonster (approved):
LGTM

PR #8677, opened 2024-09-02
front: be able to select a via on the map even when pf failed

### LINKED ISSUE #8117: Waypoint button should be disabled on the map if origin or destination aren't set

End users

When selecting a path on the map, when you hover a track this popup appears
image

Today, when you click the blue middle button, if you haven't set an origin or a destination, the popup simply closes without anything happening.

It would be great if the button was disabled so it does not look like something should happen when clicking on it

This feature matters because it:
- makes the workflow for path creation clearer

_No response_

### LINKED ISSUE #8497: Adding a via is impossible when no path was found
### What happened?

I created a train for which no path could be found (using the right track for example) and tried to add a via in the middle (using the map), but nothing happens (no request is sent) 

image

I expected to be able to add a via even without any path, since we must be able to create and update invalid trains

Create a train with no path and try to add a via

Dev (SNCF)

Google Chrome

672ec62

SarahBellaha:
Closes #8497
Closes #8117

### DIFF
--- a/front/src/modules/pathfinding/components/Itinerary/DisplayItinerary/v2/ViasV2.tsx
+++ b/front/src/modules/pathfinding/components/Itinerary/DisplayItinerary/v2/ViasV2.tsx
@@ -4,6 +4,7 @@ import { XCircle } from '@osrd-project/ui-icons';
 import cx from 'classnames';
 import type { Position } from 'geojson';
 import { DragDropContext, Droppable, Draggable } from 'react-beautiful-dnd';
+import { useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
 
 import { useOsrdConfActions, useOsrdConfSelectors } from 'common/osrdContext';
@@ -18,6 +19,7 @@ type DisplayViasV2Props = {
 };
 
 const ViasV2 = ({ zoomToFeaturePoint, shouldManageStopDuration }: DisplayViasV2Props) => {
+  const { t } = useTranslation('operationalStudies/manageTrainSchedule');
   const { getViasV2 } = useOsrdConfSelectors();
   const dispatch = useAppDispatch();
   const vias = useSelector(getViasV2());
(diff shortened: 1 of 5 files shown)


Caracol3 (approved):
Works fine for me. Good job !

front/src/modules/pathfinding/components/Itinerary/DisplayItinerary/v2/ViasV2.tsx
```diff
@@ -61,7 +63,7 @@ const ViasV2 = ({ zoomToFeaturePoint, shouldManageStopDuration }: DisplayViasV2P
                       >
                         <small className="font-weight-bold text-muted mr-1">{index + 1}</small>
                         <small data-testid="via-dropped-name" className="mr-1 text-nowrap">
-                          {`${via.name || `KM ${via.positionOnPath && (Math.round(via.positionOnPath) / 1000000).toFixed(3)}`}`}
+                          {`${via.name || (via.positionOnPath && `KM ${(Math.round(via.positionOnPath) / 1000000).toFixed(3)}`) || `${t('requestedPoint')} ${index + 1}`}`}
```
SharglutDev: Maybe we can add the via counter inside the trad with a `{{ count }}` ?

SharglutDev (commented):
Thank you for the PR, left just one comment. Also it seems there is an issue with the trads in the editoast checks (maybe you can try a rebase).

front/src/modules/pathfinding/components/Itinerary/DisplayItinerary/v2/ViasV2.tsx
SarahBellaha: As seen with @thibautsailly, it will now be "distance indisponible" / "unavailable distance" instead of "point demandé" / "requested point"

SharglutDev (approved):
Lgtm and tested

PR #8678, opened 2024-09-02
front: move NGE logic to dedicated component directory

emersion:
Move low-level NGE details to a separate directory to make sure it stays clearly split from the OSRD-specific logic.


kmer2016 (approved):
LGTM

front/src/applications/operationalStudies/views/v2/ScenarioV2.tsx
```diff
@@ -7,14 +7,11 @@ import { GiElectric } from 'react-icons/gi';
...
 import BreadCrumbs from 'applications/operationalStudies/components/BreadCrumbs';
-import NGE from 'applications/operationalStudies/components/MacroEditor/NGE';
 import handleOperation from 'applications/operationalStudies/components/MacroEditor/ngeToOsrd';
 import importTimetableToNGE from 'applications/operationalStudies/components/MacroEditor/osrdToNge';
-import type {
-  NetzgrafikDto,
-  NGEEvent,
-} from 'applications/operationalStudies/components/MacroEditor/types';
 import MicroMacroSwitch from 'applications/operationalStudies/components/MicroMacroSwitch';
+import NGE from 'applications/operationalStudies/components/NGE/NGE';
```
theocrsb, line 13: Perhaps the file name should be changed to avoid having `/NGE/NGE`

emersion: Yup, I agree that would be nicer! Though, everywhere else in OSRD we always duplicate the directory name and the filename. So I'd argue this is something we ought to change throughout the whole codebase (I would rather keep everything consistent).

Uriel-Sautron (approved):
LGTM !!! Tested.

theocrsb (approved):
LGTM

PR #8691, opened 2024-09-02
front: fix infinite loop in timestops

### LINKED ISSUE #8690: "Maximum update depth exceeded" in console in simulation results view
### What happened?

In local environment, in the simulation results view, there is a warning in console for a probably infinite loop in a `useEffect`.

image

It probably comes from the `useEffect` in `TimeStops.tsx`. We have both `allwaypoints` and `pathSteps` in the dependency array but `allwaypoints` already uses `pathSteps` to be computed.

No more warning

1. Go on a scenario with a train
2. Open the console

Local

Firefox

xxx

SharglutDev:
The useEffect dependency array was using allwaypoints and pathSteps but allwaypoints was already using pathSteps to be computed

fix #8690

### DIFF
--- a/front/src/modules/timesStops/TimesStops.tsx
+++ b/front/src/modules/timesStops/TimesStops.tsx
@@ -31,7 +31,7 @@ type TimesStopsProps = {
 
 const TimesStops = ({
   allWaypoints,
-  pathSteps = [],
+  pathSteps,
   startTime,
   tableType,
   cellClassName,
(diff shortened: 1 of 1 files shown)


emersion:
Hm, out of curiosity, why does this cause an infinite loop?

If there is a loop, it means the outputs of this `useEffect` are used as a dependency of `pathSteps` somehow?

I don't think the `pathSteps` dependency here is the root cause? (it is a real dep, and maybe in the future `allWaypoints` won't depend on `pathSteps` anymore)

emersion:
The root cause seems to be the default parameter for `TimesStops`:

https://github.com/OpenRailAssociation/osrd/blob/e4e273841ebe448ef0450e640b5873ef8c5c0c30/front/src/modules/timesStops/TimesStops.tsx#L34

JavaScript evaluates default parameters each time the function is called. In other words, each function call a new Array object is created and triggers a re-render.

emersion (approved):
LGTM :)

kmer2016 (approved):
LGTM and tested

PR #8692, opened 2024-09-02
front: converting train labels between osrd and nge

### LINKED ISSUE #8397: Converting labels between OSRD and NGE
### Description and goal

The goal of this feature is to share the labels of trains and nodes between OSRD and NGE

### Acceptance criteria

Trains:
- [x] If a label is added in OSRD, it is added in NGE
- [x] If a label is deleted in OSRD, it is deleted in NGE
- [x] If a label is added in NGE, it is added in OSRD
- [x] If a label is deleted in NGE, it is deleted in OSRD
- [x] If a label is updated in NGE, it is updated in OSRD

Macro nodes:
- [ ] If a label is added in OSRD, it is added in NGE
- [ ] If a label is deleted in OSRD, it is deleted in NGE
- [ ] If a label is added in NGE, it is added in OSRD
- [ ] If a label is deleted in NGE, it is deleted in OSRD
- [ ] If a label is updated in NGE, it is updated in OSRD

Uriel-Sautron:
first part of #8397

### DIFF
--- a/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
@@ -1,3 +1,5 @@
+import { compact } from 'lodash';
+
 import {
   osrdEditoastApi,
   type SearchResultItemOperationalPoint,
@@ -7,7 +9,7 @@ import {
 import type { AppDispatch } from 'store';
 
 import nodeStore from './nodeStore';
-import type { NetzgrafikDto, NGEEvent, TrainrunSection, Node, Trainrun } from '../NGE/types';
+import type { NetzgrafikDto, NGEEvent, TrainrunSection, Node, Trainrun, Label } from '../NGE/types';
 
 const createdTrainrun = new Map<number, number>();
 
@@ -142,7 +144,8 @@ const createTrainSchedulePayload = async (
   nodes: Node[],
   trainrun: Trainrun,
   infraId: number,
-  dispatch: AppDispatch
+  dispatch: AppDispatch,
+  labels: Label[]
 ) => {
   // TODO: check that the trainrunSections format is still compatible
   const pathPromise = trainrunSections.map(async (section, index) => {
@@ -164,9 +167,14 @@ const createTrainSchedulePayload = async (
 
   const path = await Promise.all(pathPromise);
 
+  const trainrunLabels = trainrun.labelIds.map(
+    (labelId) => labels.find((label) => label.id === labelId)?.label
+  );
+
   return {
     path: path.flat(),
     train_name: trainrun.name,
+    labels: compact(trainrunLabels),
   };
 };
 
@@ -189,8 +197,7 @@ const handleTrainrunOperation = async ({
   addUpsertedTrainSchedules: (trainSchedules: TrainScheduleResult[]) => void;
   addDeletedTrainIds: (trainIds: number[]) => void;
 }) => {
-  const { nodes } = netzgrafikDto;
-
+  const { nodes, labels } = netzgrafikDto;
   switch (type) {
     case 'create': {
       const trainrunSectionsByTrainrunId = getTrainrunSectionsByTrainrunId(
@@ -208,7 +215,8 @@ const handleTrainrunOperation = async ({
                 nodes,
                 trainrun,
                 infraId,
-                dispatch
+                dispatch,
+                labels
               )),
             },
           ],
@@ -250,7 +258,8 @@ const handleTrainrunOperation = async ({
               nodes,
               trainrun,
               infraId,
-              dispatch
+              dispatch,
+              labels
             )),
             // TODO: convert NGE times to OSRD schedule
             schedule: [],
--- a/front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
@@ -19,6 +19,8 @@ import type {
   TrainrunFrequency,
   TrainrunTimeCategory,
   NetzgrafikDto,
+  Label,
+  LabelGroup,
 } from '../NGE/types';
 import { PortAlignment } from '../NGE/types';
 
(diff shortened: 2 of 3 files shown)


Caracol3 (approved):
lgtm (tested)

front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
```diff
@@ -285,15 +288,35 @@ const importTimetable = async (
...
+    }
+    return {
+      id: trainSchedule.id,
+      name: trainSchedule.train_name,
+      categoryId: DEFAULT_TRAINRUN_CATEGORY.id,
+      frequencyId: DEFAULT_TRAINRUN_FREQUENCY.id,
+      trainrunTimeCategoryId: DEFAULT_TRAINRUN_TIME_CATEGORY.id,
+      labelIds: formatedLabels.map((label) => label.id),
+    };
+  });
```
SharglutDev: To avoid doing another map at the end, maybe we can directly stock the ids in `formatedLabels` ?
I also feel like we have a lot of nesting loops. Maybe we can improve that ?

```suggestion
  const trainruns: Trainrun[] = trainSchedules.map((trainSchedule) => {
    let formatedLabels: (string | number)[] = [];
    if (trainSchedule.labels) {
      formatedLabels = trainSchedule.labels.map((label) => {
        const DTOLabel = DTOLabels.find((DTOlabel) => DTOlabel.label === label);
        if (DTOLabel) {
          return DTOLabel.id;
        }
        const newDTOLabel: NGELabel = {
          id: uuid(),
          label,
          labelGroupId: 1,
          labelRef: 'Trainrun',
        };
        DTOLabels.push(newDTOLabel);
        return newDTOLabel.id;
      });
    }
    return {
      id: trainSchedule.id,
      name: trainSchedule.train_name,
      categoryId: DEFAULT_TRAINRUN_CATEGORY.id,
      frequencyId: DEFAULT_TRAINRUN_FREQUENCY.id,
      trainrunTimeCategoryId: DEFAULT_TRAINRUN_TIME_CATEGORY.id,
      labelIds: formatedLabels,
    };
  });
```

SharglutDev (commented):
Thank you for the PR, left just one comment

louisgreiner:
Seems like there is something missing for the synchronization when we create a label from OSRD and then try to visualize it in NGE
image

front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
Uriel-Sautron: Done

front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
```diff
@@ -285,15 +288,35 @@ const importTimetable = async (
...
+  const trainruns: Trainrun[] = trainSchedules.map((trainSchedule) => {
+    let labelIds: (string | number)[] = [];
+    if (trainSchedule.labels) {
+      labelIds = trainSchedule.labels.map((label) => {
+        const DTOLabel = DTOLabels.find((DTOlabel) => DTOlabel.label === label);
+        if (DTOLabel) {
+          return DTOLabel.id;
+        }
+        const newDTOLabel: NGELabel = {
+          id: uuid(),
```
emersion: Could we use an auto-incrementing ID (like other objects) instead of a string? That'd better match NGE's data type.

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
```diff
@@ -123,9 +138,14 @@ const createTrainSchedulePayload = (
     return [originPathItem];
   });
 
+  const trainrunLabels = trainrun.labelIds.map(
+    (labelId) => labels.find((label) => label.id === labelId)?.label
```
emersion, line 171: There should always be a match here, no?

Uriel-Sautron: No, a train doesn't always have a label.

emersion: But all label IDs in train runs must always have an entry in the dto labels array?

front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
Uriel-Sautron: Done

SharglutDev (approved):
Lgtm (ok for merge after @emersion approval :)

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
Uriel-Sautron: Yes, there will always be a match.
But i still have a TS error if I don't put "?"

front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
```diff
@@ -285,15 +286,37 @@ const importTimetable = async (
...
+          return DTOLabel.id;
+        }
+        const newDTOLabel: NGELabel = {
+          id: labelId,
+          label,
+          labelGroupId: 1,
+          labelRef: 'Trainrun',
+        };
+        DTOLabels.push(newDTOLabel);
+        labelId += 1;
```
louisgreiner:
```suggestion
          id: labelId++,
          label,
          labelGroupId: 1,
          labelRef: 'Trainrun',
        };
        DTOLabels.push(newDTOLabel);
```

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
emersion: I'd suggest using `!` then, that way we'll get an error (which we can debug) in case this assumption ends up being broken, instead of silently ignoring the issue.

front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
emersion: `++` is forbidden in our style guide (and I'm not a fan of using `++` as an expression instead of a statement, it's less clear whether the previous or next value ends up being used).

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
Uriel-Sautron: Ok, this solution also seems the best thank you for this catch.

Uriel-Sautron: Done

front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
SharglutDev: I think eslint doesn't link  de `++` :)

louisgreiner: Good to know ! :)

louisgreiner:
image

/!\ Labels are not displayed anymore on train schedules, even tho they still work (through filtering)

PR #8698, opened 2024-09-02
front: manchette: display waypoints added with map click

### LINKED ISSUE #8174: front: display the via added by click on the map in the manchette
### What happened?

> When the train path doesn't cover at least two operational points, nothing is displayed. It's an edge case but it feels wrong

This is because the vias added by clicking on the map are not displayed on the chart. We should add them with the operational points.

- [ ] the via added by clicking on the map are displayed
- [ ] if a train doesn't cover 2 operational points, the space time chart is displayed

1. Go to a scenario v2
2. Add a train by clicking on the map, set the origin and destination on the same track (try to choose a track without any op)
3. Visualize the results of the simulation: the space time chart is not displayed

_No response_

_No response_

.

SharglutDev:
fix #8174

### DIFF
--- a/front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
+++ b/front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
@@ -1,24 +1,29 @@
 import { useEffect, useState } from 'react';
 
+import { useTranslation } from 'react-i18next';
+
 import { STDCM_TRAIN_ID } from 'applications/stdcm/consts';
 import {
   osrdEditoastApi,
   type PathProperties,
-  type PostInfraByInfraIdPathPropertiesApiArg,
+  type TrainScheduleResult,
 } from 'common/api/osrdEditoastApi';
+import { addElementAtIndex } from 'utils/array';
 
 const useGetProjectedTrainOperationalPoints = (
+  trainScheduleUsedForProjection?: TrainScheduleResult,
   trainIdUsedForProjection?: number,
   infraId?: number
 ) => {
+  const { t } = useTranslation('simulation');
   const [operationalPoints, setOperationalPoints] = useState<
     NonNullable<PathProperties['operational_points']>
   >([]);
 
   const { data: pathfindingResult } = osrdEditoastApi.endpoints.getTrainScheduleByIdPath.useQuery(
     {
-      id: trainIdUsedForProjection as number,
-      infraId: infraId as number,
+      id: trainIdUsedForProjection!,
+      infraId: infraId!,
     },
     {
       skip: !trainIdUsedForProjection || !infraId || trainIdUsedForProjection === STDCM_TRAIN_ID,
@@ -30,23 +35,60 @@ const useGetProjectedTrainOperationalPoints = (
 
   useEffect(() => {
     const getOperationalPoints = async () => {
-      if (infraId && pathfindingResult && pathfindingResult.status === 'success') {
-        const pathPropertiesParams: PostInfraByInfraIdPathPropertiesApiArg = {
+      if (infraId && trainScheduleUsedForProjection && pathfindingResult?.status === 'success') {
+        const { operational_points } = await postPathProperties({
           infraId,
           props: ['operational_points'],
           pathPropertiesInput: {
             track_section_ranges: pathfindingResult.track_section_ranges,
           },
-        };
-        const { operational_points } = await postPathProperties(pathPropertiesParams).unwrap();
+        }).unwrap();
 
-        setOperationalPoints(
-          operational_points as NonNullable<PathProperties['operational_points']>
-        );
+        let operationalPointsWithAllWaypoints = operational_points!;
+
+        // Check if there are vias added by map click and insert them in the operational points
+        let waypointCounter = 1;
+
+        trainScheduleUsedForProjection.path.forEach((step, i) => {
+          if (!('track' in step)) return;
+
+          const positionOnPath = pathfindingResult.path_item_positions[i];
+          const indexToInsert = operationalPointsWithAllWaypoints.findIndex(
+            (op) => op.position >= positionOnPath
+          );
+
+          const formattedStep: NonNullable<PathProperties['operational_points']>[number] = {
+            id: step.id,
+            extensions: {
+              identifier: {
+                name: t('requestedPoint', { count: waypointCounter }),
+                uic: 0,
+              },
+            },
+            part: { track: step.track, position: step.offset },
+            position: positionOnPath,
+          };
+
+          waypointCounter += 1;
+
+          // If we can't find any op position greater than the current step position, we add it at the end
+          if (indexToInsert === -1) {
+            operationalPointsWithAllWaypoints.push(formattedStep);
+            return;
+          }
+
+          operationalPointsWithAllWaypoints = addElementAtIndex(
+            operationalPointsWithAllWaypoints,
+            indexToInsert,
+            formattedStep
+          );
+        });
+
+        setOperationalPoints(operationalPointsWithAllWaypoints);
       }
     };
     getOperationalPoints();
-  }, [pathfindingResult, infraId]);
+  }, [pathfindingResult, infraId, t]);
 
   return operationalPoints;
 };
--- a/front/public/locales/en/simulation.json
+++ b/front/public/locales/en/simulation.json
@@ -34,6 +34,7 @@
     "used": "used"
   },
   "projectionInProgress": "Projection in progress: {{ count }}/{{ total }} trains",
+  "requestedPoint": "requested point ({{ count }})",
   "reset": "Reset the chart",
   "resetTimer": "Reset timer",
   "rotate": "Rotate the chart",
--- a/front/public/locales/fr/simulation.json
+++ b/front/public/locales/fr/simulation.json
@@ -34,6 +34,7 @@
     "used": "utilisée"
   },
   "projectionInProgress": "Projection en cours: {{ count }}/{{ total }} trains",
+  "requestedPoint": "point demandé ({{ count }})",
   "reset": "Réinitialiser le graphique",
   "resetTimer": "Réinitialiser le compteur",
   "rotate": "Pivoter le graphique",
--- a/front/src/applications/operationalStudies/hooks/useScenarioData.ts
+++ b/front/src/applications/operationalStudies/hooks/useScenarioData.ts
@@ -92,6 +92,11 @@ const useScenarioData = () => {
     [projectedTrainsById]
   );
 
+  const trainScheduleUsedForProjection = useMemo(
+    () => trainSchedules?.find((trainSchedule) => trainSchedule.id === trainIdUsedForProjection),
+    [trainIdUsedForProjection, trainSchedules]
+  );
+
   useEffect(() => {
     if (!rawTrainSchedules) {
       setTrainSchedules(undefined);
@@ -193,6 +198,7 @@ const useScenarioData = () => {
         infra: { infra, isInfraLoaded, reloadCount },
         trainScheduleSummaries,
         trainSchedules,
+        trainScheduleUsedForProjection,
         trainIdUsedForProjection,
         projectedTrains,
         simulationResults,
--- a/front/src/applications/operationalStudies/views/v2/ScenarioV2.tsx
+++ b/front/src/applications/operationalStudies/views/v2/ScenarioV2.tsx
@@ -97,6 +97,7 @@ const Scenario = () => {
     infra: { infra, isInfraLoaded, reloadCount },
     trainScheduleSummaries,
     trainSchedules,
+    trainScheduleUsedForProjection,
     trainIdUsedForProjection,
     projectedTrains,
     simulationResults,
@@ -308,6 +309,7 @@ const Scenario = () => {
                       collapsedTimetable={collapsedTimetable}
                       spaceTimeData={projectedTrains}
                       simulationResults={simulationResults}
+                      trainScheduleUsedForProjection={trainScheduleUsedForProjection}
                       trainIdUsedForProjection={trainIdUsedForProjection}
                       infraId={infraId}
                       timetableTrainNb={timetable.train_ids.length}
--- a/front/src/applications/operationalStudies/views/v2/SimulationResultsV2.tsx
+++ b/front/src/applications/operationalStudies/views/v2/SimulationResultsV2.tsx
@@ -8,6 +8,7 @@ import { useSelector } from 'react-redux';
 import { Rnd } from 'react-rnd';
 
 import type { SimulationResults, TrainSpaceTimeData } from 'applications/operationalStudies/types';
+import type { TrainScheduleResult } from 'common/api/osrdEditoastApi';
 import SimulationWarpedMap from 'common/Map/WarpedMap/SimulationWarpedMap';
 import { getScaleDomainFromValuesV2 } from 'modules/simulationResult/components/ChartHelpers/getScaleDomainFromValues';
 import SimulationResultsMapV2 from 'modules/simulationResult/components/SimulationResultsMapV2';
@@ -33,6 +34,7 @@ type SimulationResultsV2Props = {
   collapsedTimetable: boolean;
   spaceTimeData?: TrainSpaceTimeData[];
   infraId?: number;
+  trainScheduleUsedForProjection?: TrainScheduleResult;
   trainIdUsedForProjection?: number;
   simulationResults: SimulationResults;
   timetableTrainNb: number;
@@ -42,6 +44,7 @@ const SimulationResultsV2 = ({
   collapsedTimetable,
   spaceTimeData,
   infraId,
+  trainScheduleUsedForProjection,
   trainIdUsedForProjection,
   simulationResults: {
     selectedTrainSchedule,
@@ -88,6 +91,7 @@ const SimulationResultsV2 = ({
   );
 
   const projectedOperationalPoints = useGetProjectedTrainOperationalPoints(
+    trainScheduleUsedForProjection,
     trainIdUsedForProjection,
     infraId
   );


front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
```diff
@@ -1,20 +1,34 @@
...
 ) => {
+  const { t } = useTranslation('simulation');
   const [operationalPoints, setOperationalPoints] = useState<
     NonNullable<PathProperties['operational_points']>
   >([]);
 
+  const { data: trainScheduleUsedForProjection } =
+    osrdEditoastApi.endpoints.getV2TrainScheduleById.useQuery(
+      {
+        id: trainIdUsedForProjection as number,
```
emersion: Nit: in general I prefer to avoid `as` casts if possible, because it makes the compiler ignore any type mismatch. If `trainIdUsedForProjection` happens to be a `string` for instance because there is a mistake in the code, this cast would hide the issue because the type-checker would remain silent.

Here `trainIdUsedForProjection` is a `number | undefined` so we should be able to use `trainIdUsedForProjection!` instead of `trainIdUsedForProjection as number` to retain the type check.

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
```diff
@@ -40,13 +59,44 @@ const useGetProjectedTrainOperationalPoints = (
...
         const { operational_points } = await postPathProperties(pathPropertiesParams).unwrap();
 
-        setOperationalPoints(
-          operational_points as NonNullable<PathProperties['operational_points']>
-        );
+        let operationalPointsWithAllWaypoints = operational_points as NonNullable<
+          PathProperties['operational_points']
+        >;
+        // Check if there are vias added by map click and insert them in the operational points
+        if (trainScheduleUsedForProjection.path.some((step) => 'track' in step)) {
```
emersion: Do we really need this `if`? The `forEach` already has a check to ignore steps which aren't track offsets no?

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
```diff
@@ -40,13 +59,44 @@ const useGetProjectedTrainOperationalPoints = (
...
+          trainScheduleUsedForProjection.path.forEach((step, i) => {
+            if ('track' in step) {
```
emersion: To avoid growing the indentation, an early return could be used:

```ts
if (!('track' in step)) {
  return;
}
```

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
```diff
@@ -40,13 +59,44 @@ const useGetProjectedTrainOperationalPoints = (
...
+              const indexToInsert = operationalPointsWithAllWaypoints.findIndex(
+                (op) => op.position >= positionOnPath
+              );
+
+              const formattedStep: NonNullable<PathProperties['operational_points']>[number] = {
+                id: step.id,
+                extensions: {
+                  identifier: {
+                    name: t('requestedPoint', { count: indexToInsert }),
+                    uic: 0,
```
emersion: Hm. This sounds like a bit of an abuse of the `PathProperties.operational_points` type...

Would we be able to perform this conversion a bit later so that we're dealing with `PathStep` instead of this type? That way we wouldn't have to fake an OP object for something which isn't an OP?

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
```diff
@@ -1,20 +1,34 @@
...
   trainIdUsedForProjection?: number,
   infraId?: number
 ) => {
+  const { t } = useTranslation('simulation');
   const [operationalPoints, setOperationalPoints] = useState<
     NonNullable<PathProperties['operational_points']>
   >([]);
 
+  const { data: trainScheduleUsedForProjection } =
+    osrdEditoastApi.endpoints.getV2TrainScheduleById.useQuery(
```
emersion: Instead of making a new query, should we take the train schedule object as input parameter of this hook? I think that would be simpler and less reliant on the rtk-query cache.

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
```diff
@@ -40,13 +59,44 @@ const useGetProjectedTrainOperationalPoints = (
         };
         const { operational_points } = await postPathProperties(pathPropertiesParams).unwrap();
 
-        setOperationalPoints(
-          operational_points as NonNullable<PathProperties['operational_points']>
-        );
+        let operationalPointsWithAllWaypoints = operational_points as NonNullable<
+          PathProperties['operational_points']
+        >;
```
emersion: Orthogonal to this PR since it was like this in the original code, but: can't this cast be simplified to a `operational_points!`?

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
```diff
@@ -40,13 +59,44 @@ const useGetProjectedTrainOperationalPoints = (
...
+              const positionOnPath = pathfindingResult.path_item_positions[i];
+              const indexToInsert = operationalPointsWithAllWaypoints.findIndex(
+                (op) => op.position >= positionOnPath
+              );
+
+              const formattedStep: NonNullable<PathProperties['operational_points']>[number] = {
+                id: step.id,
+                extensions: {
+                  identifier: {
+                    name: t('requestedPoint', { count: indexToInsert }),
```
emersion: `indexToInsert` is a bit weird to display to the user IMHO, since:

- It's 0-based, so can result in "requested point 0", which is scary for non-programmers.
- It's not the index in the list of waypoints requested by the user, it's an index in the path properties ranges? Not exactly sure.

Maybe `i + 1` would be a tad more informative. But maybe "point on track XXX" would be even better?

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
```diff
@@ -40,13 +59,44 @@ const useGetProjectedTrainOperationalPoints = (
...
+          PathProperties['operational_points']
+        >;
+        // Check if there are vias added by map click and insert them in the operational points
+        if (trainScheduleUsedForProjection.path.some((step) => 'track' in step)) {
+          trainScheduleUsedForProjection.path.forEach((step, i) => {
+            if ('track' in step) {
+              const positionOnPath = pathfindingResult.path_item_positions[i];
+              const indexToInsert = operationalPointsWithAllWaypoints.findIndex(
+                (op) => op.position >= positionOnPath
+              );
```
emersion: What if the point on the map is the last one? We wouldn't find any OP whose position is greater than `positionOnPath` right? We should insert at the end of the list in that case I think?

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
SharglutDev: I used that syntax because it's the one we use in most of the app but I'm not specially attached to it, so I can change it :)

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
SharglutDev: That was to avoid looping through all `trainSchedule.path` with the forEach if there isn't at least one step with `track` but you right it's probably overkill considering the size of the array.

emersion: `some()` also loops on all elements, so it doesn't actually result in any performance benefit. (And yeah micro-optimizations blah blah blah)

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
emersion: Yeah. I'd really like to scan all of the frontend and drop any unnecessary cast that can be simplified.

For me as a reviewer, `as` is a red flag shouting "we think we're smarter than the compiler" and I always wonder if I'm really smarter :P (sometimes it's really necessary, sometimes not)

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
SharglutDev: Done

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
SharglutDev: I don't like that either. But `later` would be inside `Manchette` component in osrd-ui, we give that array directly to the component after performing this hook.

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
SharglutDev: Yeah I thought it was a bit overkill too... The only other place where we have the data is when we fetch the batch trainSchedules in `ScenarioV2` component. I could make a `useMemo` watching for `trainIdUsedForProjection` and keeping track of the related `trainSchedules` and pass it to `SimulationResults` and then this custom hook.
What do you think ?

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
SharglutDev: Oh yep good point

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
SharglutDev: Actually I discussed with @thibautsailly about that and he doesn't want to display the index among all ops but only from the "waypoint added by map click" list

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
SharglutDev: Hmm good point. And if I create a train with an OD added by map click, I get a white screen. There is an issue in the space time chart. Do you have it too ?

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
emersion: Oh my. So, the bad news is that I now remember that ui-manchette just duplicates the editoast types as-is… And the good news is that the result of this function is only consumed by ui-manchette and nothing else (so no need to refactor a whole lot of OSRD files when we decide to change the return type), and that we've planned to refactor ui-manchette to stop doing this: https://github.com/osrd-project/osrd-confidential/issues/637

Let's leave it at that for now :)

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
SharglutDev: Totally agreed :)

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
emersion: Hm, don't we have the train schedule in `selectedTrainSchedule` in `SimulationResultsV2`? Or is this somehow sometimes not the same train as `trainIdUsedForProjection`? Or is there another reason to not use that one?

SharglutDev: Yep it's not that easy :) The selectedTrainSchedule (green background) is not always the same than the one used for projection (blue square around the train on the left of the card)

emersion: Right that makes sense indeed…! If it's not too much of a hassle I'd indeed prefer to grab the projected train schedule from `trainSchedules` in `ScenarioV2` as you suggested since that'd save an HTTP request.

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
SharglutDev: Done

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
SharglutDev: Done

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
emersion: Hm I don't see this change in the diff I think?

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
SharglutDev: Done, it also fixes the crash.

front/src/modules/simulationResult/components/SpaceTimeChart/SpaceTimeChartHeader.tsx
```diff
@@ -0,0 +1,18 @@
+import { KebabHorizontal } from '@osrd-project/ui-icons';
```
emersion: Seems like this was added in this PR by mistake?

emersion (approved):
With that extraneous file removed, LGTM! Thanks for the fix!

SharglutDev:
> With that extraneous file removed, LGTM! Thanks for the fix!

Yep my bad -_-

front/src/modules/simulationResult/components/SpaceTimeChart/SpaceTimeChartHeader.tsx
SharglutDev: Yep it comes from another PR, not sure how this file ended up here. I just amended the last commit :)

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
```diff
@@ -30,7 +36,12 @@ const useGetProjectedTrainOperationalPoints = (
 
   useEffect(() => {
     const getOperationalPoints = async () => {
-      if (infraId && pathfindingResult && pathfindingResult.status === 'success') {
+      if (
+        infraId &&
+        trainScheduleUsedForProjection &&
+        pathfindingResult &&
+        pathfindingResult.status === 'success'
```
Math-R: you can combine those in `pathfindingResult?.status === 'success'`

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
```diff
@@ -40,13 +51,51 @@ const useGetProjectedTrainOperationalPoints = (
         };
         const { operational_points } = await postPathProperties(pathPropertiesParams).unwrap();
```
Math-R: i think you can optimize operational_points set doing 

```
const { operational_points } = await postPathProperties({
        infraId,
        props: ['operational_points'],
      }).unwrap();
```

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
```diff
@@ -40,13 +51,51 @@ const useGetProjectedTrainOperationalPoints = (
...
 
-        setOperationalPoints(
-          operational_points as NonNullable<PathProperties['operational_points']>
-        );
+        let operationalPointsWithAllWaypoints = operational_points!;
```
Math-R, line 47: Are you sure that `operational_points` can't be null or undefined ?

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
```diff
@@ -40,13 +51,51 @@ const useGetProjectedTrainOperationalPoints = (
...
+            return;
+          }
+
+          operationalPointsWithAllWaypoints = addElementAtIndex(
+            operationalPointsWithAllWaypoints,
+            indexToInsert,
+            formattedStep
+          );
+
+          waypointCounter += 1;
```
emersion: Something was bugging me about this code, originally I thought it was inconsequential so ignored it but during a shower I realized it might matter after all…

What if the last two waypoints are track offsets? In that case, we'd take the `push()` route, but we still want the `waypointCounter` to be incremented to avoid having these share the same number when presented to the user. So I think we should move this line up before `if (indexToInsert === -1)`?

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
SharglutDev: Hmm we need also `pathPropertiesInput` for the request. I think it's more clearer to leave the object before and pass it to the endpoint. Is that ok for you ?

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
SharglutDev: Yep. As long as we ask for a data (ops, curves...), the endpoint will return them (the pathfinding has already been done with the previous endpoint : pathfindingblocks)

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
SharglutDev: Wow very nice shower inspiration, you were right ! Fixed :)

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
SharglutDev: Done

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
Math-R: Yes sorry the code was wrapped but still, if pathPropertiesParams is not to be reused, it's not useful. You should just put args in the request

SharglutDev: Ok done :)

Math-R (approved):
LGTM !

PR #8712, opened 2024-09-03
Better encapsulate diesel's connection to ease the split of editoast_models

Wadjetz:
Create an abstraction of the diesel database connection and transactions

part of the #6980

How to test this PR:

- [ ] Make sure everything work as before
- [ ] No deadlocks in tests

### DIFF
--- a/editoast/editoast_derive/src/model.rs
+++ b/editoast/editoast_derive/src/model.rs
@@ -84,7 +84,7 @@ fn create_functions(config: &Config) -> TokenStream {
     quote! {
         #[async_trait::async_trait]
         impl crate::models::Create for #model_name {
-            async fn create_conn(self, conn: &mut editoast_models::DbConnection) -> crate::error::Result<Self> {
+            async fn create_conn(self, conn: &mut editoast_models::DieselConnection) -> crate::error::Result<Self> {
                 use #table::table;
                 use diesel_async::RunQueryDsl;
 
@@ -126,7 +126,7 @@ fn retrieve_functions(config: &Config) -> TokenStream {
     quote! {
         #[async_trait::async_trait]
         impl crate::models::Retrieve for #model_name {
-            async fn retrieve_conn(conn: &mut editoast_models::DbConnection, obj_id: i64) -> crate::error::Result<Option<Self>> {
+            async fn retrieve_conn(conn: &mut editoast_models::DieselConnection, obj_id: i64) -> crate::error::Result<Option<Self>> {
                 use #table::table;
                 use #table::dsl;
                 use diesel_async::RunQueryDsl;
@@ -174,7 +174,7 @@ fn delete_functions(config: &Config) -> TokenStream {
     quote! {
         #[async_trait::async_trait]
         impl crate::models::Delete for #model_name {
-            async fn delete_conn(conn: &mut editoast_models::DbConnection, obj_id: i64) -> crate::error::Result<bool> {
+            async fn delete_conn(conn: &mut editoast_models::DieselConnection, obj_id: i64) -> crate::error::Result<bool> {
                 use #table::table;
                 use #table::dsl;
                 use diesel_async::RunQueryDsl;
@@ -207,7 +207,7 @@ fn update_functions(config: &Config) -> TokenStream {
     quote! {
         #[async_trait::async_trait]
         impl crate::models::Update for #model_name {
-            async fn update_conn(self, conn: &mut editoast_models::DbConnection, obj_id: i64) -> crate::error::Result<Option<Self>> {
+            async fn update_conn(self, conn: &mut editoast_models::DieselConnection, obj_id: i64) -> crate::error::Result<Option<Self>> {
                 use #table::table;
 
                 match diesel::update(table.find(obj_id)).set(&self).get_result(conn).await
--- a/editoast/editoast_models/src/db_connection_pool.rs
+++ b/editoast/editoast_models/src/db_connection_pool.rs
@@ -1,3 +1,5 @@
+use std::ops::Deref;
+use std::ops::DerefMut;
 use std::sync::Arc;
 
 use diesel::sql_query;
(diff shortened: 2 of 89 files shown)


editoast/editoast_models/src/db_connection_pool.rs
```diff
@@ -17,21 +21,81 @@ use openssl::ssl::SslMethod;
...
+                let mut handle = self.write().await;
+                match TxManager::rollback_transaction(handle.deref_mut()).await {
+                    Ok(()) | Err(diesel::result::Error::BrokenTransactionManager) => {
+                        Err(callback_error)
+                    }
+                    Err(rollback_error) => Err(rollback_error.into()),
+                }
+            }
+        }
+    }
```
woshilapin, line 89: I believe this piece of code takes inspiration from the existing implementation inside `diesel` crate (see [`TransactionManager::transaction()`](https://docs.rs/diesel/2.1.6/src/diesel/connection/transaction_manager.rs.html#50-71)). It's probably a good idea to keep a link to it and refer to it so the next developer that needs to maintain this function know it should also take inspiration on the `diesel` crate. Maybe also add a comment on the reason why we have to reimplement it ourselves and we can't use the existing one.

editoast/src/modelsv2/fixtures.rs
```diff
@@ -193,9 +194,6 @@ pub fn rolling_stock_livery_changeset(
     rolling_stock_id: i64,
     compound_image_id: i64,
 ) -> Changeset<RollingStockLiveryModel> {
-    // let rolling_stock = named_fast_rolling_stock(&rs_name, db_pool.clone()).await;
-    // let image = document_example(db_pool.clone()).await;
```
woshilapin, line 197: Can be removed.

editoast/src/modelsv2/fixtures.rs
```diff
@@ -310,3 +308,15 @@ pub async fn create_work_schedule_group(conn: &mut DbConnection) -> WorkSchedule
...
+    include_str!("../tests/train_schedules/simple_array.json")
+}
+
+pub fn get_fast_rolling_stock_schema(name: &str) -> RollingStock {
+    let mut rolling_stock_form: RollingStock =
+        serde_json::from_str(include_str!("../tests/example_rolling_stock_1.json"))
+            .expect("Unable to parse");
+    rolling_stock_form.name = name.to_string();
+    rolling_stock_form
+}
```
woshilapin: Why do we need these new functions in a PR that only change the `DbConnection`?

woshilapin (changes requested):
Big work, will be much better. Thank you 🎉 

I request a change on the documentation of the re-implementation of `transaction` since it's a central but non-trivial function of this new `DbConnection`. I believe it's important to give as much context as possible to ease future maintenance.

editoast/src/modelsv2/fixtures.rs
woshilapin: Removed in https://github.com/OpenRailAssociation/osrd/pull/8712/commits/e8c0350489cfc4c6543056d5fb81ba53621c7e21.

editoast/editoast_models/src/db_connection_pool.rs
woshilapin: Proposition of documentation in https://github.com/OpenRailAssociation/osrd/pull/8712/commits/9fef9a6a13826c3c8cf2c5ba809dc1aa61b2314c. @leovalais What do you think?

editoast/editoast_derive/src/model.rs
leovalais, line 1: Unless I misunderstood something, `DieselConnection` is supposed to be an `editoast_models` internal type. I'd say that these functions should take a `DbConnection` just like before instead.

editoast/editoast_models/src/lib.rs
```diff
@@ -1,16 +1,23 @@
...
 /// Useful for functions which only points of failure are the DB calls.
 #[derive(Debug, thiserror::Error)]
-#[error("an error occured while querying the database: {0}")]
-pub struct DatabaseError(#[from] diesel::result::Error);
+pub enum DatabaseError {
+    #[error("an error occured while querying the database: {0}")]
+    DieselError(#[from] diesel::result::Error),
+    #[error("an error occured while retrieving a connection from the pool: {0}")]
+    DatabasePoolError(#[from] DatabasePoolError),
+}
```
leovalais: I don't understand this change. There was two different types `DatabaseError` and `DatabasePoolError` that

1. serve different purposes, cf. their error messages
2. allows `editoast_models` to forward errors without exposing diesel's errors

Besides, `DatabasePoolError` also forwards `diesel::result::Error`. Its seems to me that the error types don't require any modification.

editoast/editoast_models/src/lib.rs
```diff
@@ -1,16 +1,23 @@
...
 
 pub mod db_connection_pool;
 pub mod tables;
 
+pub use db_connection_pool::DbConnection;
 pub use db_connection_pool::DbConnectionPoolV2;
 
-pub type DbConnection = AsyncPgConnection;
-pub type DbConnectionPool = Pool<DbConnection>;
+pub type DieselConnection = AsyncPgConnection;
```
leovalais: Why is this type still `pub`? We'd like to only expose `DbConnection` outside of the crate. Am I missing something?

editoast/src/modelsv2/auth.rs
```diff
@@ -39,49 +41,50 @@ impl<B: BuiltinRoleSet + Send + Sync> StorageDriver for PgAuthDriver<B> {
...
-                                authn_user::name.eq(&user.name),
-                            ))
-                            .execute(conn)
-                            .await?;
-
-                        Ok(id)
+        self.pool
+            .get()
+            .await?
+            .clone()
```
leovalais: `clone`?

leovalais (changes requested):
Huge work! I have a few comments about the API, especially for errors.

That's work for another PR since this one is already massive:

1. Most functions don't require a `conn: &mut DbConnection` but rather just a `conn: DbConnection` which became cheap to clone.
2. Maybe we could implement `Future` for `DbConnection` so that `conn.await` <=> `conn.write().await.deref_mut()`
3. Some sync test util functions might declutter the boilerplate the `RwLock` introduced.

editoast/editoast_derive/src/model.rs
leovalais: Oops scratch that... it's Model V1

leovalais (approved):
Approving in order not to block the PR, but since I made the modifications myself, this doesn't account for much 🤣

woshilapin (approved):

PR #8713, opened 2024-09-03
editoast: update all 'opentelemetry' related dependencies

woshilapin:
All Opentelemetry dependencies must be updated all at once (Dependabot can't really understand that). Here is the PR.

### DIFF
--- a/editoast/Cargo.lock
+++ b/editoast/Cargo.lock
@@ -18,6 +18,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "adler2"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


woshilapin:
Seems like upgrading these dependencies means also that we need to upgrade minimum version of Rust

```
error: rustc 1.79.0 is not supported by the following package:
  axum-tracing-opentelemetry@0.19.1 requires rustc 1.80.0
Either upgrade rustc or select compatible dependency versions with
`cargo update <name>@<current-ver> --precise <compatible-ver>`
where `<compatible-ver>` is the latest version supporting rustc 1.79.0
```

Are we ok with such a change ?

woshilapin:
Argh, installation of `grcov` doesn't work with Rust 1.80 because they didn't update `time` version... Brrrrr, should we start thinking of moving away of `grcov` ([no update on the subject since June](https://github.com/mozilla/grcov/issues/1187))? Until we decide, I believe removing `--locked` will do the trick...

Khoyo (approved):

Khoyo:
>  no update on the subject since June? 

No commits on master since almost a year ago

flomonster (approved):
LGTM

Khoyo (approved):

woshilapin:
I'm closing this PR since it doesn't work and it's already out of date anyway (see #9005 which will need to be fixed too).

PR #8714, opened 2024-09-03
front: add webkit browser to e2e tests

### LINKED ISSUE #8645: front: Add webkit for e2e tests
# Description and goal

Add webkit for e2e test for them to work on Safari.

### Acceptance criteria

- [x] All e2e tests work on WebKit

Maymanaf:
Closes #8645

### DIFF
--- a/front/playwright.config.ts
+++ b/front/playwright.config.ts
@@ -35,10 +35,10 @@ export default defineConfig({
     /* Base URL to use in actions like `await page.goto('/')`. */
     baseURL: process.env.BASE_URL || 'http://localhost:4000',
 
-    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
-    trace: 'on-first-retry',
+    /* Collect trace and video when retrying the first failed test. See https://playwright.dev/docs/trace-viewer */
+    trace: process.env.CI ? 'on-first-retry' : 'retain-on-failure',
+    video: process.env.CI ? 'on-first-retry' : 'retain-on-failure',
     locale: 'fr',
-    video: 'retain-on-failure',
   },
   reporter: process.env.CI ? 'github' : [['line'], ['html']],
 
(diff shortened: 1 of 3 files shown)


front/playwright.config.ts
```diff
@@ -63,5 +63,15 @@ export default defineConfig({
...
+    {
+      name: 'webkit',
+      use: {
+        browserName: 'webkit',
+        launchOptions: {
+          slowMo: 500, // Slows down WebKit interactions by 500 milliseconds
+        },
+      },
+      dependencies: ['setup'],
+    },
```
Yohh, line 75: shouldn't we only use test.slow() for specifics tests? I'm afraid that the actual delay could hide some performance issues

Yohh (commented):
nice! I'm glad to finally see it!

front/playwright.config.ts
Maymanaf: WebKit is around 1.7 times slower than Firefox and Chromium, according to feedback from the Playwright community. Therefore, adding `slowMo` ( only 500ms) limits the number of tests on which we apply `test.slow()`.

Yohh (approved):
LGTM

we should be carefull, tests are a lot flaky on macOs if others apps are open

SharglutDev (approved):
A lot of tests are failing for me too in local (macOS)

Maymanaf:
> A lot of tests are failing for me too in local (macOS)

@SharglutDev @Yohh Locally, it's much slower than on the CI, and the application's performance on an unloaded infrastructure is not optimal. That's why it's better to limit the number of workers and increase the timeout. I tested several builds and solutions on the CI before submitting the PR, and it works well now. Only the 003 tests remain unstable, but they've been giving this result for a while. I'm going to have to modify the logic behind them

SharglutDev:
Maybe we should add a readme somewhere or complete the osrd documentation to explain the best way to launch the e2e tests in local.

Yohh:
> Maybe we should add a readme somewhere or complete the osrd documentation to explain the best way to launch the e2e tests in local.

or in this specific case, we should limit the workers to 1 by default, continue to write tests parallelize and let the user increase the workers if he wants

if the tests are too huge to work parallelized we shouldn't let the workers to 50%

SharglutDev:
If we do that, the tests will take longer on the CI or it can be set only for local env ?

Yohh:
yes we can keep 50% for CI, 1 for local

PR #8721, opened 2024-09-03
front: fix dropped ch in NGE when missing from infra

emersion:
To test:

- Create a node "SAVA/XD" in NGE
- Create a train run from this new node to any other node
- Go to micro mode
- Go back to NGE

Without this patch, the node would show up as "SAVA" in NGE. With this patch, it should show up as "SAVA/XD".

This bug is only triggered for trigrams which can't be found in the infrastructure.


Math-R (approved):
lgtm 👍🏻

Caracol3 (approved):
Works fine for me (tested)

PR #8722, opened 2024-09-03
front: update useMemo() dependencies in useOutputTableData()

emersion:
pathProperties and path aren't consumed by this function, so drop them. pathStepsWithOpPointIndices was consumed but was missing from the dependencies.


kmer2016 (approved):
LGTM

Uriel-Sautron (approved):
LGTM !!

PR #8724, opened 2024-09-03
editoast: activate 'datadog' feature on production images

woshilapin:
Our default deployment need Datadog so let's make it an option for building production images.

### DIFF
--- a/editoast/Dockerfile
+++ b/editoast/Dockerfile
@@ -46,13 +46,14 @@ COPY --from=planner /editoast/recipe.json recipe.json
 COPY --from=planner /editoast/editoast_derive/ editoast_derive
 ENV RUSTFLAGS="-C link-arg=-fuse-ld=mold"
 ARG CARGO_PROFILE=release
+ARG CARGO_FEATURES
 RUN --mount=type=cache,target=/usr/local/cargo/registry \
     --mount=type=cache,target=/editoast/target \
-    cargo chef cook --profile ${CARGO_PROFILE} --recipe-path recipe.json
+    cargo chef cook --features="${CARGO_FEATURES}" --profile="${CARGO_PROFILE}" --recipe-path recipe.json
 COPY . .
 RUN --mount=type=cache,target=/usr/local/cargo/registry \
     --mount=type=cache,target=/editoast/target \
-    cargo install --profile ${CARGO_PROFILE} --locked --path .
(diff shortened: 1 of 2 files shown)


editoast/Dockerfile
```diff
@@ -46,13 +46,14 @@ COPY --from=planner /editoast/recipe.json recipe.json
 COPY --from=planner /editoast/editoast_derive/ editoast_derive
 ENV RUSTFLAGS="-C link-arg=-fuse-ld=mold"
 ARG CARGO_PROFILE=release
+ARG CARGO_FEATURES
 RUN --mount=type=cache,target=/usr/local/cargo/registry \
     --mount=type=cache,target=/editoast/target \
-    cargo chef cook --profile ${CARGO_PROFILE} --recipe-path recipe.json
+    cargo chef cook --features ${CARGO_FEATURES} --profile ${CARGO_PROFILE} --recipe-path recipe.json
```
emersion: Hm. `${}` is docker-style expansion, so I'm worried this gets expanded to `cook --features  --profile debug` for local builds. Does this work fine still?

woshilapin: Oh, right... why didn't I thought about that (thank you peer review 😄). It probably doesn't work indeed. Should be fix in the last push.

emersion (approved):

leovalais (approved):

flomonster (approved):

ghost (approved):
Lgtm but not tested

younesschrifi (approved):

Khoyo (approved):
LGTM. Is the feature unnecessary to compute the recipe.json file?

woshilapin:
> LGTM. Is the feature unnecessary to compute the recipe.json file?

The whole point of the recipe is to build as many dependencies as possible to put in a image layer as a cache so we don’t compile it later. So I thought yes, it makes sense to compile this optional dependency in the recipe. Do you have something different in mind?

Khoyo:
> > LGTM. Is the feature unnecessary to compute the recipe.json file?
> 
> The whole point of the recipe is to build as many dependencies as possible to put in a image layer as a cache so we don’t compile it later. So I thought yes, it makes sense to compile this optional dependency in the recipe. Do you have something different in mind?

I was wondering if cargo chef included all feature-gated dependencies in the recipe.json file or not.

PR #8737, opened 2024-09-04
front: fix time extraction for single-digit days

### LINKED ISSUE #8736: Hour minute second display bug
### What happened?

The `extractHHMMSS` and `extractHHMM` functions do not give the expected result if the day does not contain 2 digits. 

Example : 
 -12/09/2024 12:00:00 => 12:00
 -2/09/2024 07:00:00 => 7:00:

The `extractHHMMSS` and `extractHHMM` functions must return hours minutes seconds in the correct format. 

Expect =>  
-12/09/2024 12:00:00 => 12:00
-2/09/2024 07:00:00 => 07:00

Use function `extractHHMM` with date `2/09/2024 07:00:00`

Local

Google Chrome

dd98ab5

theocrsb:
closes #8736

### DIFF
--- a/front/src/utils/date.ts
+++ b/front/src/utils/date.ts
@@ -18,22 +18,29 @@ const userTimeZone = dayjs.tz.guess(); // Format : 'Europe/Paris'
  * @param dateTimeString date string in ISO format
  * @returns string "HH:MM:SS"
  */
-export function extractHHMMSS(dateTimeString?: string) {
+export function extractHHMMSS(dateTimeString?: string): string {
   if (!dateTimeString) {
     return '';
   }
-  return dateTimeString.substring(11, 19);
+  const date = new Date(dateTimeString);
+  const hours = date.getHours().toString().padStart(2, '0');
+  const minutes = date.getMinutes().toString().padStart(2, '0');
+  const seconds = date.getSeconds().toString().padStart(2, '0');
+  return `${hours}:${minutes}:${seconds}`;
 }
 
 /**
  * @param dateTimeString date string in ISO format
  * @returns string "HH:MM"
  */
-export function extractHHMM(dateTimeString?: string) {
+export function extractHHMM(dateTimeString?: string): string {
   if (!dateTimeString) {
     return '';
   }
-  return dateTimeString.substring(11, 16);
+  const date = new Date(dateTimeString);
+  const hours = date.getHours().toString().padStart(2, '0');
+  const minutes = date.getMinutes().toString().padStart(2, '0');
+  return `${hours}:${minutes}`;
 }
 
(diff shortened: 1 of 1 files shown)


front/src/utils/date.ts
Akctarus, line 1: Maybe, to reduce redundancy, since both functions share common logic for extracting hours and minutes, you can create a function to handle that part : 

function formatTime(date: Date, includeSeconds: boolean = false): string {
  const hours = date.getHours().toString().padStart(2, '0');
  const minutes = date.getMinutes().toString().padStart(2, '0');
  const seconds = date.getSeconds().toString().padStart(2, '0');
  return includeSeconds ? `${hours}:${minutes}:${seconds}` : `${hours}:${minutes}`;
}

export function extractHHMMSS(dateTimeString?: string): string {
  if (!dateTimeString) {
    return '';
  }
  const date = new Date(dateTimeString);
  return formatTime(date, true);
}

export function extractHHMM(dateTimeString?: string): string {
  if (!dateTimeString) {
    return '';
  }
  const date = new Date(dateTimeString);
  return formatTime(date);
}

Akctarus (commented):
Tested, works well, I left a comment to discuss a small refactor

front/src/utils/date.ts
```diff
@@ -18,22 +18,29 @@ const userTimeZone = dayjs.tz.guess(); // Format : 'Europe/Paris'
  * @param dateTimeString date string in ISO format
  * @returns string "HH:MM:SS"
  */
-export function extractHHMMSS(dateTimeString?: string) {
+export function extractHHMMSS(dateTimeString?: string): string {
```
anisometropie, line 21: When you have specific cases you want to handle, you need to write some unit tests that give these specific inputs to the function (like the one you gave in the issue), this helps to document the function behavior.

-12/09/2024 12:00:00 => 12:00
-2/09/2024 07:00:00 => 7:00:


Most importantly, the function is supposed to take _date string in ISO format_, it is not supposed to take `12/09/2024 12:00:00` `2/09/2024 07:00:00` as input.

If we want this function to handle different formats, it needs to be more robust and fully tested

anisometropie (changes requested):

front/src/utils/date.ts
anisometropie: if we handle this format you need to use dayjs instead of js `Date` api
Screenshot_20240904_124630


dayjs('2/09/2024 07:00:00', ['DD/MM/YYYY HH:mm:ss', 'D/MM/YYYY HH:mm:ss'])

theocrsb:
PR no longer necessary. Workaround added in another [PR](https://github.com/OpenRailAssociation/osrd/pull/8765)

PR #8739, opened 2024-09-04
front: simplify OP matching in useOutputTableData()

emersion:
We were using a map keyed by the name and ch of OPs. This is more complicated than it needs to be and isn't robust (there is no guarantee that names+ch are unique).

Instead, just use the OP ID.

Original discussion: https://github.com/OpenRailAssociation/osrd/pull/8555/files#r1735948569


kmer2016 (approved):
LGTM

Akctarus (approved):
Tested, LGTM

PR #8740, opened 2024-09-04
core: stdcm: refactor time data into a new data class

### LINKED ISSUE #8587: core: stdcm: rework timing data along edges and nodes
Timing data is currently handled as a set of variables on each edge and node, which is quite messy:

* `time` (which is very poorly named)
* `total_prev_added_delay`
* `maximum_added_delay`
* `time_since_departure`

There's also issues around finding out basic values such as:

* finding out when we're supposed to reach a given point (in post-processing)
* finding out the time since departure at any given point (in post-processing)

We should do a small refactoring that merge these values together in a new object, which may have extra data for more straightforward computations.

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/8587

**Now is the time for reviews about these time fields, their name, their structure, their docstrings, and so on**

Change summary:

1. Move time-related data from edges and nodes into a new struct
2. Improve how these values are forwarded between nodes and edges
3. Add comments for what still needs to be changed for variable stop times

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/TimeData.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/TimeData.kt
@@ -67,4 +67,43 @@ data class TimeData(
      * lengthening stop durations.
      */
     val timeSinceDeparture = totalRunningTime + totalStopTime
+
+    /** Returns a copy of the current instance, with added travel / stop time. */
+    fun withAddedTime(
+        extraTravelTime: Double,
+        extraStopTime: Double,
+    ): TimeData {
+        return copy(
+            earliestReachableTime = earliestReachableTime + extraTravelTime + extraStopTime,
+            totalRunningTime = totalRunningTime + extraTravelTime,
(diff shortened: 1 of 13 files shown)


core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/TimeData.kt
```diff
@@ -0,0 +1,113 @@
...
+            totalRunningTime = totalRunningTime + extraTravelTime,
+            totalStopTime = totalStopTime + extraStopTime
+        )
+    }
+
+    /**
+     * Return a copy of the current instance, with "shifted" time values. Used to create new edges.
+     *
+     * @param timeShift by how much we delay the new element compared to the earliest possible time.
+     *   A value >0 means that we want to arrive later (to avoid a conflict)
```
Erashin: This is completely nitpicky, I don't really care.
```suggestion
     *   A value > 0 means that we want to arrive later (to avoid a conflict)
```

Erashin (approved):
Looks good :)

PR #8741, opened 2024-09-04
osrdyne: support keda autoscaling for k8s driver

### LINKED ISSUE #8408: osrdyne: Horizontal scaling for one infrastructure
# Description :

If a queue is full of messages for an infrastructure then new instances for this ijnfrastructure should be provided to dequeue messages faster ! 

# Implementation plan

⚠ complex work has it is a fine tuning ⚠ 

1. Implement rust types for kube/keda api
2. Add scaling object and defined a scaling policy (depending on the number of messages waiting in the queue)
4. tests tests tests

# Acceptance Criteria : 

- [ ] Be able to spawn multiple pods for an infrastructure instance if messages are not processed fast enough

## Definition of ready

- **PO**
  - [x] Acceptance criteria are clear and understandable

- **Technical**

- **General**
  - [x] WSJF values are filled
  - [x] Validated by the PM

ghost:
Closes #8408

### DIFF
--- a/osrdyne/src/drivers/kubernetes.rs
+++ b/osrdyne/src/drivers/kubernetes.rs
@@ -15,19 +15,70 @@ use k8s_openapi::{
     },
     apimachinery::pkg::apis::meta::v1::LabelSelector,
 };
+use keda::MetricType;
 use kube::{api::ObjectMeta, Client};
+use log::debug;
 use serde::{Deserialize, Serialize};
-use std::{collections::BTreeMap, future::Future, pin::Pin};
+use std::{
+    collections::{BTreeMap, HashMap},
+    future::Future,
+    pin::Pin,
+};
 use uuid::Uuid;
 
+mod keda;
+
 #[derive(Serialize, Deserialize, Clone, Debug)]
-pub struct AutoscalingOptions {
+pub struct HPAOptions {
     /// The minimum number of replicas
     pub min_replicas: i32,
+
     /// The maximum number of replicas
     pub max_replicas: i32,
+
     /// The target CPU utilization percentage
     pub target_cpu_utilization_percentage: i32,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct KedaOptions {
+    /// Polling Interval in seconds
+    pub polling_interval: i32,
+
+    /// Cooldown Period in seconds
+    pub cooldown_period: i32,
+
+    /// Min Replicas for the HPA of the ScaledObject
+    pub min_replicas: i32,
+
+    /// Max Replicas for the HPA of the ScaledObject
+    pub max_replicas: i32,
+
+    /// Amqp Host
+    pub amqp_host: String,
+
+    /// Mode
+    pub mode: String,
+
+    /// Value
+    pub value: String,
+
+    /// Activation Value
+    pub activation_value: String,
+
+    /// Use cached metrics
+    pub use_cached_metrics: bool,
+
+    /// Metric Type
+    pub metric_type: MetricType,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+#[serde(tag = "type")]
+pub enum AutoscalingOptions {
+    Hpa(HPAOptions),
+    Keda(KedaOptions),
+    NoScaling,
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug)]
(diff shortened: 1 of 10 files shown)


eckter (approved):
LGTM for core

Not tested, but I trust you on this

osrdyne/src/drivers/kubernetes.rs
```diff
@@ -15,21 +15,72 @@ use k8s_openapi::{
...
     /// The maximum number of replicas
     pub max_replicas: i32,
+
     /// The target CPU utilization percentage
     pub target_cpu_utilization_percentage: i32,
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct KedaOptions {
+    /// The name of the ScaledJob
```
flomonster: Wrong doc

osrdyne/src/drivers/kubernetes.rs
```diff
@@ -15,21 +15,72 @@ use k8s_openapi::{
...
+    pub polling_interval: i32,
+
+    /// The name of the ScaledJob
+    pub cooldown_period: i32,
+
+    /// The name of the ScaledJob
+    pub min_replicas: i32,
+
+    /// The name of the ScaledJob
+    pub max_replicas: i32,
```
bloussou: Documentation of fields here
`/// The name of the ScaledJob`

osrdyne/src/drivers/kubernetes.rs
```diff
@@ -15,21 +15,72 @@ use k8s_openapi::{
...
+mod keda;
+
 #[derive(Serialize, Deserialize, Clone, Debug)]
-pub struct AutoscalingOptions {
+pub struct HPAOptions {
     /// The minimum number of replicas
     pub min_replicas: i32,
+
     /// The maximum number of replicas
     pub max_replicas: i32,
```
flomonster, line 37: Small question, currently osrdyne deploy an horizontal pod autoscaller with the capabilities to schedule multiple pods for the same infra. Why isn't this scale method good enough for us?

bloussou (commented):
Can't wait to test this !!

osrdyne/src/drivers/kubernetes.rs
```diff
@@ -208,53 +265,125 @@ impl WorkerDriver for KubernetesDriver {
...
+            debug!("Creating deployment: {:?}", deployment);
 
             // Create the deployment
             kube::api::Api::<Deployment>::namespaced(self.client.clone(), &self.options.namespace)
                 .create(&kube::api::PostParams::default(), &deployment)
                 .await
                 .map_err(super::worker_driver::DriverError::KubernetesError)?;
 
+            // Create the autoscaler if needed
+            match &self.options.autoscaling {
```
woshilapin, line 387: This is a huge `match` (more than a hundred lines). Maybe splitting it into functions would help see where is each block (because indentation on a hundred lines is not enough).

flomonster (approved):

osrdyne/src/drivers/kubernetes.rs
```diff
@@ -208,53 +265,125 @@ impl WorkerDriver for KubernetesDriver {
...
+                    debug!("Creating Keda ScaledObject: {:?}", scaled_object);
+
+                    kube::api::Api::namespaced(self.client.clone(), &self.options.namespace)
+                        .create(&kube::api::PostParams::default(), &scaled_object)
+                        .await
+                        .map_err(super::worker_driver::DriverError::KubernetesError)?;
+                }
+
+                // No autoscaler configured
+                _ => {}
```
woshilapin: Don't you want the compiler to yell a bit if a new variant is added to `AutoscalingOptions`? I would rather be explicit on this last branch.

```suggestion
                Some(AutoscalingOptions::NoScaling) | None => {}
```

osrdyne/src/drivers/kubernetes.rs
ghost: Corrected

woshilapin (approved):
Thanks. Only minor comments.

osrdyne/src/drivers/kubernetes.rs
ghost: Basically, Keda is able to use information from within the RabbitMQ. On the backend it creates an HPA that uses a metrics that is provided by Keda itself (instead of CPU/MEM that we could theorically uses).

The rationale behind the usage of Keda instead of CPU/RAM is that the load average for Core is not a good indicator of a required scaling, the latency in message dequeuing will be a more accurate indicator of the current load since it will directly be related to the latency experience by the users.

bloussou (approved):
Let's go !

Khoyo (approved):
LGTM

The amount of inline hasmap/btreemap construction we are starting to do makes me wonder if we should use maplit

ghost:
> maplit

Will do in another PR later in time

PR #8745, opened 2024-09-04
editoast: add cli to manage stdcm search environments

### LINKED ISSUE #8272: editoast: add cli and script for stdcm search env
### Description and goal

> [!warning]
> Depends on #8168

* Add a cli command to post an stdcm search env
* Add a script which uses its cli in editoast's docker
* Add the use of this script in the readme along the use of scripts importing infras and rolling stocks

### Acceptance criteria

- [ ] A new user can follow the readme and be able to run STDCM queries in the STDCM v2 interface

Castavo:
close #8272 

Add 3 commands:
* show
* set-from-scratch
* set-from-scenario

I don't know if set-from-scratch will ever be useful...

I also chose to make the search window 1 day bigger than the start of the last train, in order to be able to ask for an STDCM arrival time which arrives later than the departure of the last train in the timetable. I don't know if that makes a lot of sense, what do you think ?

### DIFF
--- a/editoast/src/client/stdcm_search_env_commands.rs
+++ b/editoast/src/client/stdcm_search_env_commands.rs
@@ -1,0 +1,454 @@
+use crate::models::electrical_profiles::ElectricalProfileSet;
+use crate::models::stdcm_search_environment::StdcmSearchEnvironment;
+use crate::models::timetable::Timetable;
+use crate::models::work_schedules::WorkScheduleGroup;
+use crate::models::Infra;
+use crate::models::Scenario;
+use crate::CliError;
+use crate::Exists;
+use crate::Model;
+use crate::Retrieve;
+use chrono::Duration;
+use chrono::NaiveDateTime;
+use clap::Args;
+use clap::Subcommand;
+use editoast_models::DbConnection;
+use editoast_models::DbConnectionPoolV2;
+use std::error::Error;
+
+#[derive(Subcommand, Debug)]
+pub enum StdcmSearchEnvCommands {
+    SetFromScenario(SetSTDCMSearchEnvFromScenarioArgs),
+    SetFromScratch(SetSTDCMSearchEnvFromScratchArgs),
+    Show,
+}
+
+pub async fn handle_stdcm_search_env_command(
+    command: StdcmSearchEnvCommands,
+    db_pool: DbConnectionPoolV2,
+) -> Result<(), Box<dyn Error + Send + Sync>> {
+    let conn = &mut db_pool.get().await?;
+    match command {
+        StdcmSearchEnvCommands::SetFromScenario(args) => {
+            set_stdcm_search_env_from_scenario(args, conn).await
+        }
+        StdcmSearchEnvCommands::SetFromScratch(args) => {
+            set_stdcm_search_env_from_scratch(args, conn).await
+        }
+        StdcmSearchEnvCommands::Show => show_stdcm_search_env(conn).await,
+    }
+}
+
+async fn check_exists<T: Exists<i64>>(
+    conn: &mut DbConnection,
+    object_id: i64,
+    readable_name: &str,
+) -> Result<(), Box<dyn Error + Send + Sync>> {
+    if !T::exists(conn, object_id).await? {
+        let err_msg = format!("❌ {0} not found, id: {1}", readable_name, object_id);
+        return Err(Box::new(CliError::new(1, err_msg)));
+    }
+    Ok(())
+}
+
+#[derive(Args, Debug)]
+#[command(
+    about,
+    long_about = "Set the current STDCM search env by copying most data from a scenario"
+)]
+pub struct SetSTDCMSearchEnvFromScenarioArgs {
+    pub scenario_id: i64,
+    #[arg(long)]
+    pub work_schedule_group_id: Option<i64>,
+    /// If omitted, set to the earliest train start time in the timetable
+    #[arg(long)]
+    pub search_window_begin: Option<NaiveDateTime>,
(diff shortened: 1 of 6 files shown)


editoast/src/client/stdcm_search_env_commands.rs
woshilapin, line 1: Minor: In the `enum`, `FromScenario` is first then `FromScratch` but for the rest of the file, it's the other way. Might be nice to keep the same order.

editoast/src/client/stdcm_search_env_commands.rs
```diff
@@ -0,0 +1,448 @@
...
+        let start_times = Timetable::gather_start_times(timetable_id, conn).await?;
+
+        if start_times.is_empty() {
+            let error_msg =
+                "❌ Timetable specified contains no train. Please fully specify search window.";
+            return Err(Box::new(CliError::new(1, error_msg)));
+        }
+
+        let begin = search_window_begin
+            .unwrap_or_else(|| start_times.iter().min().expect("Vec.len() > 0").naive_utc());
```
woshilapin: The `expect` error message might not help too much an end-user? How about something like "Needs at least one train in timetables".

Maybe you can even avoid entirely the `expect` by changing the above `if` by something like

```rust
let (Some(min), Some(max)) = (start_times.iter().min().naive_utc(), start_times.iter().max().naive_utc()) else {
    let error_msg =
        "❌ Timetable specified contains no train. Please fully specify search window.";
    return Err(Box::new(CliError::new(1, error_msg)));
};
```

woshilapin (approved):
Will be much better with this utility. Thank you. Only minor comments, nothing is a blocker.

editoast/src/client/stdcm_search_env_commands.rs
```diff
@@ -0,0 +1,444 @@
...
+            set_stdcm_search_env_from_scenario(args, conn).await
+        }
+        StdcmSearchEnvCommands::SetFromScratch(args) => {
+            set_stdcm_search_env_from_scratch(args, conn).await
+        }
+        StdcmSearchEnvCommands::Show => show_stdcm_search_env(conn).await,
+    }
+}
+
+macro_rules! exists_or_return {
```
emersion: Instead of a macro, could we use a simple function taking an `Exists<K>` from `src/modelsv2/prelude/retrieve.rs`?

Castavo: Why do you want to use a function instead ?
To me, in a function, having 
```rust
check_existence(WorkScheduleGroup, work_schedule_group_id, conn)?
``` 
seems as readable as 
```rust
exists_or_return!(WorkScheduleGroup, work_schedule_group_id, conn)
``` 

If I'm not mistaken, with a function I wouldn't benefit from `stringify!`, right ?
I also feel like a function returning a Result made only for being `?`ed would have a purpose a bit less clear than the macro

WDYT ?

Castavo: Okay, while re-reading, it does seem a bit more readable since it seems more rustic.

However I feel like the last two points still outweigh the readability improvement (especially `stringify`)

emersion: I dislike using metaprogramming unless it's actually required. The macro itself is less readable than a regular function, it requires juggling with more complicated constructs, error messages are tricky to decipher, it slows down the build, etc etc.

Castavo: Could you please provide me an example on how to retrieve the class's name without the macro ? I found [`type_name`](https://doc.rust-lang.org/std/any/fn.type_name.html) but fully qualified struct names aren't desirable.

I don't see how to have a meaningful error message inside the function, so I would guess I could  just put the code from the macro everywhere it was used, but that's an awful lot of boilerplate

woshilapin: You could always split at `:` with `type_name`.

```rust
std::any::type_name::<Option<bool>>().rsplit_once(':').expect("type_name always return a full path containing ':'").1
```

Note that [per documentation](https://doc.rust-lang.org/std/any/fn.type_name.html#note), it's not ensured that the full path will be output from `type_name()`

> For example, amongst the strings that `type_name::<Option<String>>()` might return are `"Option<String>"` and `"std::option::Option<std::string::String>"`.

But I'd say it's relatively unlikely that `std` will change on that, and even if it does, we'll panic so it'll be easy to notice and fix.

Castavo: Isn't it preferable to have boilerplate rather than a potential panic ?

woshilapin: I'll add that we do use technical `struct` names for user facing content: it's probably wrong. Instead of `❌ ElectricalProfileSet not found, id: 42`, a user might want a sentence with English words, so something like `❌ electrical profile not found, id '42'`.

Don't get me wrong, I'm fine with displaying the type like you propose (we don't take the necessary time to consider a real user exposed to these messages anyway). But with that argument in mind, I'm not sure that `ElectricalProfileSet` is objectively much better than `editoast::modelsv2::electrical_profiles::ElectricalProfileSet` both are technical anyway (although one is arguably shorter so maybe more aesthetic and less noisy, but that starts to be more subjective).

emersion: If we want to display a human-readable name for the error message, we could pass it as an argument of the function?

Castavo: Fine

woshilapin (approved):

emersion (approved):
Thank you!

PR #8748, opened 2024-09-04
front: osrd recognizes secondary code as BV, 00, or an empty string if no secondary code is provided by the user in macroscopic mode.

### LINKED ISSUE #8396: When creating a node is NGE without secondary_code, recognize it in OSRD with secondary codes in this order : "BV">"00">""
### Description and goal

When one creates a node in NGE with a trigram (for example "PLY"), we would like OSRD to recognize this node as a  BV (which can correspond to secondary code of BV, 00 or nothing)

### Acceptance criteria

- [x] If a node without secondary_code is created in NGE, OSRD tries to recognize it to an OP with the same trigram, and BV as secondary code
- [x] If the first point fails, OSRD tries to recognize it to an OP with the same trigram, and 00 as secondary code
- [x] If the second point fails, OSRD tries to recognize it to an OP with the same trigram, with an empty secondary code
- [ ] If the third point fails, OSRD does not recognize the OP

Caracol3:
close #8396

### DIFF
--- a/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
@@ -1,5 +1,6 @@
 import {
   osrdEditoastApi,
+  type SearchResultItemOperationalPoint,
   type TrainScheduleBase,
   type TrainScheduleResult,
 } from 'common/api/osrdEditoastApi';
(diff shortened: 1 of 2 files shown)


front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
```diff
@@ -97,34 +98,70 @@ const DEFAULT_PAYLOAD: Pick<
...
+    };
+    let searchResults = (await dispatch(
+      osrdEditoastApi.endpoints.postSearch.initiate({
+        searchPayload,
+      })
+    ).unwrap()) as SearchResultItemOperationalPoint[];
+
+    let opFound = searchResults.find((op) => op.ch === 'BV' || op.ch === '00');
+    if (!opFound) opFound = searchResults.find((op) => op.ch === '');
+    finalSecondaryCode = opFound?.ch;
```
kmer2016: Is it necessary to split the `find` operation into two parts? What do you think to simplify it by merging them like this:

```suggestion
   finalSecondaryCode = searchResults.find((op) => op.ch === 'BV' || op.ch === '00' || op.ch === '')?.ch;
```

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
```diff
@@ -97,34 +98,70 @@ const DEFAULT_PAYLOAD: Pick<
...
+) => {
   const [trigram, secondaryCode] = node.betriebspunktName.split('/');
-  return {
+  let finalSecondaryCode: string | undefined;
+  if (!secondaryCode) {
+    const searchPayload = {
+      object: 'operationalpoint',
+      query: ['and', ['=', ['infra_id'], infraId], ['=', ['trigram'], trigram]],
+    };
+    let searchResults = (await dispatch(
```
kmer2016:
```suggestion
    const searchResults = (await dispatch(
```

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
Caracol3: Yes, I think it's necessary. We want to be able to prioritize the secondary_code, and writing it this way allows us to ensure we have a BV or 00 before the empty string.

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
Caracol3: done :-)

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
```diff
@@ -97,34 +98,70 @@ const DEFAULT_PAYLOAD: Pick<
...
+const createPathItemFromNode = async (
+  node: Node,
+  index: number,
+  infraId: number,
+  dispatch: AppDispatch
+) => {
   const [trigram, secondaryCode] = node.betriebspunktName.split('/');
-  return {
+  let finalSecondaryCode: string | undefined;
+  if (!secondaryCode) {
```
achrafmohye: Thanks for the PR ! One suggestion: you could use an **early return** when secondaryCode is present. While it’s not mandatory this will reduce the nesting of if statements and make the code cleaner by simplifying the logic before the final return.
```suggestion
    if (secondaryCode) {
      return {
          trigram: trigram,               
          secondary_code: secondaryCode,
          id: `${node.id}-${index}`,    
      };
  }
// continue If no secondaryCode, prepare the search payload
```

Caracol3: Thanks for the suggestions !
Its done.

achrafmohye (approved):
LGTM !

kmer2016 (approved):
LGTM !

PR #8751, opened 2024-09-04
core: pin jar versions in Dockerfile

emersion:
Automatic version upgrades can cause regressions and make the resulting container non-deterministic. Pin to a specific version and make version upgrades explicit.


emersion:
@Khoyo wants to add a GitHub action to open PRs to update the version numbers. I don't have interest in this, so will leave this up to someone else. :)

PR #8756, opened 2024-09-04
front: Import trainSchedules from Viriato file

### LINKED ISSUE #8718: Import trainSchedules from a Viriato file
### Description and goal

Related to https://github.com/osrd-project/osrd-confidential/issues/570
We should be able to import trainSchedules in the OSRD timetable, from a Viriato file. 

### Acceptance criteria

- [x]  Pouvoir déposer un fichier Viriato à l'aide du bouton d'import de circulation déjà présent dans les scénarios OSRD (le même bouton sera utilisé pour les imports json et xml, et le format sera reconnu pour savoir de quel type d'import il s'agit)
- [x] Les circulations importées depuis Viriato sont présentes dans OSRD, avec (cf plan d'implem pour savoir quel attribut du fichier xml utiliser) :
Le nom
L'heure de départ
La fréquency (sous forme de label) : cette fréquence est lue dans Viriato comme l'écart entre les 2 premiers trains du trainGroup. S'il n'y en a qu'un, on met 1h par défaut
La catégorie (sous forme de label)
Les points de passage
Les horaires aux points de passage
Les temps d'arrêt aux points de passage

RomainValls:
closes #8718 

To test the feature, download the folder containing the different import files.
We're testing the imports "Etoile de Veynes", "Etude Medoc", "MSC-AIX_LNPCA", "RP Trame" and "Test_Vir4".
_Link to the files :_
[https://sncf.sharepoint.com/sites/OSRD644GrpO365/Documents%20partages/Forms/AllItems.aspx?id=%2Fsites%2FOSRD644GrpO365%2FDocuments%20partages%2FDonn%C3%A9es%2F3%2DFichiers%20Viriato&viewid=21cdca8b%2D1763%2D4a6e%2Da9e3%2Dc9c036f6a510](url)

**Many trains are invalid :** 
In most cases, it's due to the fact that the rolling stock is not found : 

- Edit the train
- Check the error popup message with the rollingstock name
- Try changing the rollingstock with the correct one

When we have a rollingstock, sometimes we have errors : 
- Pathfinding works but the simulation fails
- Pathfinding doesn't work (most of the time -> Canton manquant)

What would be nice, for a future feature, is to be able to multi edit the invalid trains, and affect them correct rollingstocks simultaneously.

### DIFF
--- a/front/src/common/uploadFileModal.tsx
+++ b/front/src/common/uploadFileModal.tsx
@@ -11,25 +11,51 @@ import { ModalContext } from 'common/BootstrapSNCF/ModalSNCF/ModalProvider';
 interface UploadFileModalProps {
   handleSubmit: (file: File) => void;
 }
+
 const UploadFileModal = ({ handleSubmit }: UploadFileModalProps) => {
-  const { t } = useTranslation(['translation']);
+  const { t } = useTranslation(['operationalStudies/importTrainSchedule']);
   const { closeModal } = useContext(ModalContext);
   const [selectedFile, setSelectedFile] = useState<File | undefined>(undefined);
   const [isValid, setIsValid] = useState<string | undefined>(undefined);
 
+  const parseXML = (xmlString: string) => {
+    try {
+      const parser = new DOMParser();
+      const xmlDoc = parser.parseFromString(xmlString, 'application/xml');
+      const parserError = xmlDoc.getElementsByTagName('parsererror');
+      if (parserError.length > 0) {
+        throw new Error('Invalid XML');
+      }
+      return undefined;
+    } catch (error) {
+      return t('errorMessages.errorInvalidXMLFormat').toString();
+    }
+  };
+  // TODO : create the translation keys
   const validateFile = useCallback(
     async (fileToValidate: File): Promise<string | undefined> => {
-      if (fileToValidate.type !== 'application/json') {
-        return t('jsonUpload.notJSONFormat').toString();
-      }
       if (fileToValidate.size === 0) {
-        return t('jsonUpload.emptyFile').toString();
+        return t('errorMessages.errorEmptyFile').toString();
       }
-      try {
-        JSON.parse(await fileToValidate.text());
-      } catch (e) {
-        return t('jsonUpload.badJSON').toString();
+      if (fileToValidate.type === 'application/json') {
+        try {
+          JSON.parse(await fileToValidate.text());
+        } catch (e) {
+          return t('errorMessages.errorInvalidJSONFormat').toString();
+        }
+      } else if (
+        fileToValidate.type === 'application/railml' ||
+        fileToValidate.name.endsWith('.railml')
+      ) {
+        const fileContent = await fileToValidate.text();
+        const xmlError = parseXML(fileContent);
+        if (xmlError) {
+          return xmlError;
+        }
+      } else {
+        return t('errorMessages.errorUnsupportedFileType').toString();
       }
+
       return undefined;
     },
     [t]
@@ -47,7 +73,7 @@ const UploadFileModal = ({ handleSubmit }: UploadFileModalProps) => {
           <input
             type="file"
             name="file"
-            accept=".json"
+            accept=".json,.xml,.railml"
             onChange={async (e) => {
               if (e.target.files && e.target.files.length > 0) {
                 const error = await validateFile(e.target.files[0]);
@@ -73,7 +99,7 @@ const UploadFileModal = ({ handleSubmit }: UploadFileModalProps) => {
                 className="btn btn-block btn-sm btn-secondary"
                 onClick={closeModal}
               >
-                {t('common.cancel')}
+                {t('cancel')}
               </button>
             </div>
             <div className="col-6">
@@ -85,7 +111,7 @@ const UploadFileModal = ({ handleSubmit }: UploadFileModalProps) => {
                   if (selectedFile) handleSubmit(selectedFile);
                 }}
               >
-                {t('common.download')}
+                {t('download')}
               </button>
             </div>
           </div>
--- a/front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
+++ b/front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
@@ -3,13 +3,16 @@ import { useState, useContext } from 'react';
 import { Download, Search } from '@osrd-project/ui-icons';
 import { isEmpty } from 'lodash';
 import { useTranslation } from 'react-i18next';
+import nextId from 'react-id-generator';
 
 import type {
   ImportedTrainSchedule,
   TrainScheduleImportConfig,
+  Step,
+  CichDictValue,
 } from 'applications/operationalStudies/types';
 import { getGraouTrainSchedules } from 'common/api/graouApi';
-import type { TrainScheduleBase } from 'common/api/osrdEditoastApi';
+import { type TrainScheduleBase } from 'common/api/osrdEditoastApi';
 import InputSNCF from 'common/BootstrapSNCF/InputSNCF';
 import { ModalContext } from 'common/BootstrapSNCF/ModalSNCF/ModalProvider';
 import StationCard, { type ImportStation } from 'common/StationCard';
@@ -18,6 +21,13 @@ import StationSelector from 'modules/trainschedule/components/ImportTrainSchedul
 import { setFailure } from 'reducers/main';
 import { useAppDispatch } from 'store';
 import { formatIsoDate } from 'utils/date';
+
+import {
+  handleFileReadingError,
+  handleUnsupportedFileType,
+  processJsonFile,
+  processXmlFile,
+} from '../ManageTrainSchedule/helpers/handleParseFiles';
 
 interface ImportTrainScheduleConfigProps {
   setTrainsList: (trainsList: ImportedTrainSchedule[]) => void;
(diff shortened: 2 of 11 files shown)


front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
```diff
@@ -41,6 +47,69 @@ const ImportTrainScheduleConfig = ({
...
+  };
+
+  const overloadDict = async (localCichDict: Record<string, any>): Promise<Record<string, any>> => {
+    const newDict = { ...localCichDict };
+
+    for (const key of Object.keys(newDict)) {
+      const { ciCode, chCode } = newDict[key];
+
+      // Call the search function with ciCode and chCode
+      const searchResult = await searchOperationalPoints(ciCode, chCode);
```
emersion: This sends one OP search HTTP request per cich. Would be nicer to send a single HTTP request for all cichs. See the OSRD → NGE converter for an example.

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
```diff
@@ -41,6 +47,69 @@ const ImportTrainScheduleConfig = ({
...
+    chCode?: string
+  ): Promise<SearchResultItemOperationalPoint[]> => {
+    const isSearchingByTrigram = !Number.isInteger(+ciCode) && ciCode.length < 4;
+    const searchQuery = isSearchingByTrigram
+      ? ['or', ['=i', ['trigram'], ciCode], ['search', ['name'], ciCode]]
+      : ['or', ['search', ['name'], ciCode], ['like', ['to_string', ['uic']], `%${ciCode}%`]];
+
+    const payload = {
+      object: 'operationalpoint',
+      query: ['and', searchQuery],
```
emersion: Should probably narrow the search to the current infra?

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
```diff
@@ -144,17 +213,235 @@ const ImportTrainScheduleConfig = ({
...
 
-    const text = await file.text();
-    const importedTrainSchedules: TrainScheduleBase[] = JSON.parse(text);
+    const fileName = file.name.toLowerCase();
+    const fileExtension = fileName.split('.').pop();
 
-    if (!isEmpty(importedTrainSchedules)) {
-      setTrainsJsonData(importedTrainSchedules);
-    }
+    const reader = new FileReader();
```
emersion: Could probably use `await file.text()` like we do in the validation function instead of `FileReader`?

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
```diff
@@ -41,6 +47,69 @@ const ImportTrainScheduleConfig = ({
...
+      ? ['or', ['=i', ['trigram'], ciCode], ['search', ['name'], ciCode]]
+      : ['or', ['search', ['name'], ciCode], ['like', ['to_string', ['uic']], `%${ciCode}%`]];
+
+    const payload = {
+      object: 'operationalpoint',
+      query: ['and', searchQuery],
+    };
+
+    try {
+      const results = await postSearch({ searchPayload: payload, pageSize: 101 }).unwrap();
```
emersion: It's not super great to make an import process depend on the current infra… This makes it stateful. If the infra contains errors, it's not possible to fix it to address the issues - need to re-import the XML file as well. IOW, we loose information during the import.

Can't we generate `PathItemLocation` straight from the railml? Why do we need to search the infra?

front/src/common/uploadFileModal.tsx
```diff
@@ -11,25 +11,52 @@ import { ModalContext } from 'common/BootstrapSNCF/ModalSNCF/ModalProvider';
...
   const { closeModal } = useContext(ModalContext);
   const [selectedFile, setSelectedFile] = useState<File | undefined>(undefined);
   const [isValid, setIsValid] = useState<string | undefined>(undefined);
 
+  const parseXML = (xmlString: string) => {
+    try {
+      const parser = new DOMParser();
+      const xmlDoc = parser.parseFromString(xmlString, 'application/xml');
+      const parserError = xmlDoc.getElementsByTagName('parsererror');
+      console.log(xmlDoc, 'xmlDoc');
```
emersion: Nit: remnant from a debugging session?

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
RomainValls: I'm not sure we can generate PathItemLocation, as in the XML file we only have ci / ch codes. Is it enough ? 
I rely on those codes to find associated uic or trigram.

emersion: It seems like it's possible to convert between CI and UIC, there's just a "86" to add/remove.

One could argue that converting CI to UIC this way would make the import SNCF-specific. However I'd argue that it's already SNCF-specific by assuming a specific "ci/ch" format for the `code` XML attribute.

RomainValls: Done, we'll use ```87{cicode}``` to find operational points. However, this solution will break if the import files contain steps outside France

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleTrainsList.tsx
```diff
@@ -47,19 +47,40 @@ const ImportTrainScheduleTrainsList = ({
...
+          const normalizedRollingStock = rollingStock.toUpperCase().replace(/[_\W]/g, ''); // Normalize the rollingStock name
+
+          const validTrainNameKey = Object.keys(rollingstockOpenData2OSRD).find((key) => {
+            // Normalize key and value
+            const normalizedKey = key.toUpperCase().replace(/[_\W]/g, '');
+            const normalizedValue = rollingstockOpenData2OSRD[
+              key as keyof typeof rollingstockOpenData2OSRD
+            ]
+              .toUpperCase()
+              .replace(/[_\W]/g, '');
```
Math-R: I think you can extract the "normalize" logic in a dedicated fuinction

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleTrainsList.tsx
```diff
@@ -47,19 +47,40 @@ const ImportTrainScheduleTrainsList = ({
...
+          }) as RollingstockOpenData2OSRDKeys | undefined;
+
+          const validTrainName = validTrainNameKey
+            ? rollingstockOpenData2OSRD[validTrainNameKey as keyof typeof rollingstockOpenData2OSRD]
+            : rollingStock;
+
+          return { ...train, rollingStock: validTrainName };
+        }
+
+        return { ...train, rollingStock: '' }; // Ensure rollingStock is always present
```
Math-R: You can use an early return

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
```diff
@@ -144,16 +147,221 @@ const ImportTrainScheduleConfig = ({
...
+          const id = ocp.getAttribute('id');
+          const code = ocp.getAttribute('code');
+
+          if (id && code) {
+            const { ciCode, chCode } = extractCiChCode(code);
+            localCichDict[id] = { ciCode, chCode };
+          }
+        });
+      });
+    });
```
Math-R: As we just use ocp is it possible to just access ocp without using nested loops ?

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
```diff
@@ -144,16 +147,221 @@ const ImportTrainScheduleConfig = ({
...
+        console.error('Unsupported file type');
+        dispatch(
+          setFailure({
+            name: t('errorMessages.error'),
+            message: t('errorMessages.errorUnsupportedFileType'),
+          })
+        );
+      }
+    } catch (error) {
+      console.error('File reading error:', error);
```
Math-R: There is a lot of  nested try catch, i think you should extract the management of different file types into dedicated functions. And also have a function to dispatch errors to avoid repetition.
This could give a tree structure like this and simplify reading

```
const importFile = async (file: File) => {
  closeModal();
  setTrainsList([]);

  const fileName = file.name.toLowerCase();
  const fileExtension = fileName.split('.').pop();

  try {
    const fileContent = await file.text();

    if (fileExtension === 'json') {
      processJsonFile(fileContent);
    } else if (fileExtension === 'xml' || fileExtension === 'railml') {
      processXmlFile(fileContent);
    } else {
      handleUnsupportedFileType();
    }
  } catch (error) {
    handleFileReadingError(error);
  }
};
```

Math-R (changes requested):

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/handleParseFiles.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export const handleXmlParsingError = (error: unknown) => {
+  console.error('Error parsing XML/RailML:', error);
+  dispatch(
+    setFailure({
+      name: t('errorMessages.error'),
+      message: t('errorMessages.errorInvalidXMLFormat'),
+    })
+  );
+};
```
Math-R: Can you use a proper erro type

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/trainNameHelper.ts
```diff
@@ -9,4 +13,25 @@ const trainNameWithNum = (name: string, actualTrainCount: number, total: number)
...
+
+  return Object.keys(rollingstockOpenData2OSRD).find((key) => {
+    const normalizedKey = normalizeString(key);
+    const normalizedValue = normalizeString(
+      rollingstockOpenData2OSRD[key as keyof typeof rollingstockOpenData2OSRD]
+    );
+
+    return (
+      normalizedRollingStock.includes(normalizedKey) ||
+      normalizedRollingStock.includes(normalizedValue)
```
Math-R: can you replace name as follow 
```
key => rollingstock
normalizedKey => normalizedRollingstock
noralizedValue => normalizedRollingStockValue
```
without the file stucture on the side it will be more comprehensible

RomainValls: I let key as key, changed the others.

clarani:
Tested, the test file can be imported :white_check_mark:

Math-R (approved):
lgtm =)

emersion:
Can all of the commits be squashed together (or re-split into logical commits)?

emersion:
Thanks for that!

PR #8758, opened 2024-09-04
core, editoast: return conflict requirements

### LINKED ISSUE #8680: core, editoast: add requirements field to conflict list
### Description and goal

Return a new requirements field in the conflict list:

```yaml
- conflict_type: Spacing
  train_ids: [0, 1]
  start_time: "2024-08-07T08:50:17.421Z"
  duration: 120_000 # ms
  requirements:
    - time_range: [0, 50_000]
      # Core uses routing_requirements and spacing_requirements to deduce zones
      zones: ["zone_1", "zone_2"]
    - time_range: [45_000, 120_000]
      zones: ["zone_3"]
```

### Acceptance criteria

.

emersion:
Closes: https://github.com/OpenRailAssociation/osrd/issues/8680

There are two TODOs left which are open questions for reviewers.

### DIFF
--- a/core/src/main/java/fr/sncf/osrd/api/ConflictDetectionEndpoint.java
+++ b/core/src/main/java/fr/sncf/osrd/api/ConflictDetectionEndpoint.java
@@ -90,11 +90,31 @@ public class ConflictDetectionEndpoint implements Take {
             @Json(name = "conflict_type")
             public final ConflictType conflictType;
 
-            public Conflict(Collection<Long> trainIds, double startTime, double endTime, ConflictType conflictType) {
+            public final transient Collection<ConflictRequirement> requirements;
+
+            public Conflict(
+                    Collection<Long> trainIds,
+                    double startTime,
+                    double endTime,
+                    ConflictType conflictType,
+                    Collection<ConflictRequirement> requirements) {
                 this.trainIds = trainIds;
                 this.startTime = startTime;
                 this.endTime = endTime;
                 this.conflictType = conflictType;
+                this.requirements = requirements;
+            }
+        }
+
+        public static class ConflictRequirement {
+            public final String zone;
+            public final double startTime;
+            public final double endTime;
+
+            public ConflictRequirement(String zone, double startTime, double endTime) {
+                this.zone = zone;
+                this.startTime = startTime;
+                this.endTime = endTime;
             }
         }
 
--- a/core/src/main/java/fr/sncf/osrd/conflicts/Conflicts.kt
+++ b/core/src/main/java/fr/sncf/osrd/conflicts/Conflicts.kt
@@ -4,6 +4,7 @@ import com.carrotsearch.hppc.IntArrayList
 import com.squareup.moshi.Json
 import fr.sncf.osrd.api.ConflictDetectionEndpoint.ConflictDetectionResult.Conflict
 import fr.sncf.osrd.api.ConflictDetectionEndpoint.ConflictDetectionResult.Conflict.ConflictType
+import fr.sncf.osrd.api.ConflictDetectionEndpoint.ConflictDetectionResult.ConflictRequirement
 import fr.sncf.osrd.standalone_sim.result.ResultTrain.RoutingRequirement
 import fr.sncf.osrd.standalone_sim.result.ResultTrain.SpacingRequirement
 import kotlin.math.max
@@ -171,12 +172,15 @@ class IncrementalConflictDetectorImpl(trainRequirements: List<TrainRequirements>
         // look for requirement times overlaps.
         // as spacing requirements are exclusive, any overlap is a conflict
         val res = mutableListOf<Conflict>()
-        for (requirements in spacingZoneRequirements.values) {
-            for (conflictGroup in detectRequirementConflicts(requirements) { _, _ -> true }) {
+        for (entry in spacingZoneRequirements) {
+            for (conflictGroup in detectRequirementConflicts(entry.value) { _, _ -> true }) {
                 val trains = conflictGroup.map { it.trainId }
                 val beginTime = conflictGroup.minBy { it.beginTime }.beginTime
                 val endTime = conflictGroup.maxBy { it.endTime }.endTime
-                res.add(Conflict(trains, beginTime, endTime, ConflictType.SPACING))
+                val conflictReq = ConflictRequirement(entry.key, beginTime, endTime)
+                res.add(
+                    Conflict(trains, beginTime, endTime, ConflictType.SPACING, listOf(conflictReq))
+                )
             }
         }
         return res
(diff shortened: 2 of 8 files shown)


woshilapin (approved):
Looks good, I leave you in the expert hands of core maintainers for your questions.

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/conflicts/ConflictDetectionResponse.kt
```diff
@@ -18,6 +18,14 @@ class Conflict(
     @Json(name = "end_time") val endTime: ZonedDateTime,
     @Json(name = "conflict_type")
     val conflictType: ConflictDetectionEndpoint.ConflictDetectionResult.Conflict.ConflictType,
+    // TODO: would "zones" be a better name?
+    @Json(name = "requirements") val requirements: Collection<ConflictRequirement>,
```
flomonster: We should keep the word “requirements”, since we use `spacing_requirements` and `routing_requirements` as input. 

> [!NOTE]
> `zone_requirements` is a good option for me

flomonster (approved):

flomonster (approved):
LGTM. 

I'm just nitpicking that the hyperlinks have the full url as text.

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/conflicts/ConflictDetectionResponse.kt
eckter: Is it just the list of all requirements, or specifically the conflicting ones? Is it the raw requirements in full or just the conflicting ranges (intersection of requirement times)?

I can probably figure it out from the context, but the name (or maybe a comment) could be more precise

core/src/main/java/fr/sncf/osrd/conflicts/Conflicts.kt
```diff
@@ -447,6 +452,12 @@ fun mergeMap(
             events.add(Event(EventType.END, conflict.endTime))
         }
 
+        // TODO: accumulate in the for loop below somehow?
+        val conflictReqs = mutableListOf<ConflictRequirement>()
+        for (conflict in conflicts) {
+            conflictReqs.addAll(conflict.requirements);
+        }
```
eckter: Seems fine to me, accumulating in the loop wouldn't change the result and this is more straightforward.

That's assuming that we *do* want to list every requirement that appear in the conflict and group them in the same list. When two trains are fully overlapping we will end up with the full requirement list of both trains. But it sounds correct in this context.

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/conflicts/ConflictDetectionResponse.kt
emersion: It's only the list of conflicting requirements, not the full list of all requirements. It's the intersection of requirement times, not the full requirements.

I've added a comment in editoast's type to try to make this more explicit.

core/src/main/java/fr/sncf/osrd/conflicts/Conflicts.kt
emersion: My main worry here is that the requirements are duplicated across all `Conflict` objects created inside the loop, even if some of the requirements may not apply to all of these. We could possibly use `eventBeginning`/`event.time` to filter out requirements which don't apply when creating the merged `Conflict`? Or keep track of all accumulated conflicts in-between a `BEGIN` and `END` and then merge their requirements?

core/src/main/java/fr/sncf/osrd/api/ConflictDetectionEndpoint.java
```diff
@@ -90,11 +90,26 @@ public enum ConflictType {
             @Json(name = "conflict_type")
             public final ConflictType conflictType;
 
-            public Conflict(Collection<Long> trainIds, double startTime, double endTime, ConflictType conflictType) {
+            public final Collection<ConflictRequirement> requirements;
```
Erashin: You'll need to add `transient` here. Otherwise, this will be serialised in the tsv1 conflict detection json output, and we don't want that :)

emersion: Good point, done!

core/src/main/java/fr/sncf/osrd/conflicts/Conflicts.kt
Erashin: Honestly, filtering by `event.begin` and `event.time` at line 471 when creating the new conflict seems fine. So if you wanna diminish the duplicated requirements try that, and test it to check that we still have the full list of conflict requirements over all the created objects.

You should probably move line 458 to line 453, it's the same for loop.

emersion: Requirement lists are now more fine-grained.

Erashin: Only thing left is a unit test then, thx for that :)

emersion:
Is this the kind of test that you had in mind? I just kind of abused the first test here, but could easily move it to another test (maybe one with multiple conflicts? which one? I also wouldn't want to hardcode tons of expected output fields...).

core/src/test/java/fr/sncf/osrd/standalone_sim/ConflictDetectionTest.java
```diff
@@ -131,8 +135,31 @@ public void headToHeadRoutingConflict() throws Exception {
         {
             var conflicts = ConflictsKt.detectConflicts(List.of(
                     convertRequirements(0L, 0.0, simResultA.train), convertRequirements(1L, 0.0, simResultB.train)));
-            assertTrue(conflicts.stream().anyMatch((conflict) -> conflict.conflictType == ROUTING));
             assertFalse(conflicts.stream().anyMatch((conflict) -> conflict.conflictType == SPACING));
+
+            var routingConflicts = conflicts.stream()
```
Erashin: This works but I'd rather have a whole other unit test testing conflict requirements (as you said in a comment above), which would alleviate this test's complexity, and it would be easier to understand why the test fails if someone breaks this feature. What's more, that would allow you to have a better setup to avoid having a negative `expectedStartTime` (which is quite weird) and not have all the conflict requirements starting and ending at the same time (which is certainly not always the case).
As for the test case, well you could check an intersection of 3 conflicts on the same zone: A & B conflict , B & C conflict, A & C don't conflict. The resulting conflicts should have all 3 requirements (and other non related conflicts on the same zone should not). You could also test A & B & C conflict.

Erashin (approved):
Apart from the fix needed on the unit test, this looks good. Thanks :)

core/src/test/java/fr/sncf/osrd/standalone_sim/ConflictDetectionTest.java
emersion: This is much better, avoiding `simpleSim()` is great to have much more sensible expected test result data. Thanks for the suggestion!

I've written the test before seeing your comment edit - and it's not quite the same as you describe. I think my test covers all of the new logic I've added, did I miss something?

Erashin: It's fine. You got the main things right. The rest is finer details. Good job, lgtm :)

PR #8759, opened 2024-09-04
Rename ModelV2 by Model

### LINKED ISSUE #4184: [META] editoast: enhance the `Model` macro
# Enhance the `Model` macro

## Current state

Currently, when applied to a struct, the `Model` macro can generate a bunch of
implementations to factorize common DB operations:

* access: `trait Retrieve`
* deletion: `trait Delete`
* creation: `trait Create`
* update: `trait Update`

There is two approaches on how to use it:

### Application on one struct (most common in editoast)

```rust
#[derive(Debug, Default, Identifiable, Insertable, Model, Queryable)]
#[model(table = "osrd_infra_document")]
#[model(create, delete, retrieve, update)]
#[diesel(table_name = osrd_infra_document)]
pub struct Document {
    #[diesel(deserialize_as = i64)]
    pub id: Option<i64>,
    #[diesel(deserialize_as = String)]
    pub content_type: Option<String>,
    #[diesel(deserialize_as = Vec<u8>)]
    pub data: Option<Vec<u8>>,
}

let mut d: Document = Document::retrieve(42, conn).await?.unwrap();
println("ct: {}", d.content_type.unwrap());
d.content_type = Some("x".to_owned());
d.update(conn).await?;
let d2 = Document { id: None, content_type: d.content_type, data: d.data };
d2.create(conn).await?; // clone doc 42
assert_eq!(d2.id, 43);
```

Pros:

* only one struct per object to represent
* straightforward
* how `Model` was originally meant to be used

Cons:

* every field is an `Option<>`
  - otherwise columns with default values (such as `id`, `creation_date`, etc.) cannot be used properly
  - actually optional columns must be represented either by `Option<Option<>>` or by blurring the distinction (when possible)
  - each field must be annotated with `deserialize_as`
  - `unwrap()`s everywhere...
  - aaaaaaaaahhhhh........
* one same struct represents two very different things:
  - a valid document that represents something meaningful
  - a set of (column, value) pairs that define a partial document by leaving some columns to their default values, which is just a low level Rust way to represent the values for an `INSERT` or an `UPDATE`
* copy issues: default values have to be excluded instead of only including changes, etc.
* ID issues
* weakens high-level algorithms by making it `unwrap()` everywhere (bis), and let's not talk about the pain it is to review such code where you have to wonder whether each `unwrap()` is normal and never fails or not.

### Application on two structs (for pathfinding and maybe train schedules)

https://github.com/DGEXSolutions/osrd/blob/c8a1770ca54769c13e51c60931707d364e1ec781/editoast/src/models/pathfinding.rs#L21-L100

Pros:

* better separation between a valid pathfinding and an incomplete one for creation/insertion
* the AsChangeSet approach is what Diesel promotes
* `Option<>` fields only where they truly represent an optional value
* less `deserialize_as` and `unwrap()`

Cons:

* a lot of duplication
* unbearably verbose
* poorly maintainable
* copy issues: default values have to be excluded instead of only including changes, etc.
* `PathfindingChangeset::create` and `PathfindingChangeset::update` return a `PathfindingChangeset` instead of a `Pathfinding`
* no one wants to write code like that

## Proposal

I'd like to write something like that instead (:santa:):

```rust
#[derive(Debug, Default, Identifiable, Model, Queryable)]
#[model(table = "osrd_infra_document")]
#[model(create, delete, retrieve)]
#[diesel(table_name = osrd_infra_document)]
pub struct Document {
    pub id: i64,
    pub content_type: String,
    pub data: Vec<u8>,
}

let mut d: Document = Document::retrieve(42, conn).await?.unwrap();
println("ct: {}", d.content_type);
d.changeset()
 .content_type("x".to_owned())
 .update(conn)
 .await?;
assert_eq!(&d.content_type, "x"); // (mut d) should also be updated
let d2: Document = Document::changeset()
    .content_type(d.content_type)
    .data(d.data)
    .create(conn) // clone doc 42
    .await?;
assert_eq!(d2.id, 43);
```

While the actual API here is just a proposal, there are, I believe, a few
essential points that an enhanced `Macro` model should at least cover:

* no more unnecessary `Option` and cluttering
* create and update are patches of some columns instead of a full row replacement that are easy to write/spot/type/debug
* the model struct declaration should be as tight as possible, without duplication

## `Model`ling infra objects

In the case of infra objects, only a few models are implemented right now (atm only catenaries, OPs and tracks) but more will come :mage: and the implementation can be made more satisfactory ⚙️.

I.e.: for track sections, atm:

https://github.com/DGEXSolutions/osrd/blob/c8a1770ca54769c13e51c60931707d364e1ec781/editoast/src/models/infra_objects/track_section.rs#L9-L20

There are several problems:

* the same struct `(id, infra_id, obj_id, data)` is duplicated for each model
* infra objects are usually identified by the couple `(infra_id, obj_id)` rather that by row ID, but the traits `Retrieve`, etc. do not support that identification
* `retrieve`, `create`, etc. all work using a `infra_model::TrackSectionModel` (the boilerplate) and not using a `schema::TrackSection` (which is the struct we'd like to use)

### Proposal

* Get rid of `models::infra_objects` altogether and focus on what's defined in `schema`.
* Reuse the `InfraModel` derive macro so that it also:
  1. Generates the struct `(id, infra_id, obj_id, data)` that `derives(Model)` and references the schema
  2. Provides an implementation of `Retrieve`, `Create` and `Delete` (not `Update`, see below) directly for the schema, but using `(infra_id, obj_id)` instead of the row id directly.
  3. These implementation are responsible for making the equivalence row id <=> `(infra_id, obj_id)` and to forward the call to the actual `Model`

As far as I can tell, infra objects are only updated in `schema::operation` (osrd editor requests) that updates the DB directly.
Hence, for now I suppose there is no need to design a way to update infra models in a Rust fashion. But should we need to,
the separation `InfraModel`(schema) <-> `Model`(DB row) could allow us to design an elaborate schema changeset builder
where chained method calls would map into a set of (JSON path, update value), who knows?

Likewise, I've pinned two ways infra objects are created. Using the `persist_batch` function (current behaviour of `InfraModel`)
during infra import and in unit tests. `persist_batch` directly queries the DB, I think it should stay that way (I can elaborate if that's necessary :smile:).
In unit tests, it's convenient to have a `create()` method, but there is no need to handle the case of autogenerated column
values for infra objects (we only need to provide the whole schema). So here as well we don't need to bother with changesets and elaborate builders.

```rust
// The declaration is left unchanged (InfraModel already exists)
#[derive(Debug, Derivative, Clone, Deserialize, Serialize, PartialEq, InfraModel)]
#[serde(deny_unknown_fields)]
#[infra_model(table = "crate::tables::osrd_infra_tracksectionmodel")]
#[derivative(Default)]
pub struct TrackSection {
    pub id: Identifier,
    #[derivative(Default(value = "100."))]
    pub length: f64,
    pub slopes: Vec<Slope>,
    pub curves: Vec<Curve>,
    #[serde(default)]
    pub loading_gauge_limits: Vec<LoadingGaugeLimit>,
    #[derivative(Default(value = "Geometry::new(LineString(vec![]))"))]
    pub geo: Geometry,
    #[derivative(Default(value = "Geometry::new(LineString(vec![]))"))]
    pub sch: Geometry,
    #[serde(default)]
    pub extensions: TrackSectionExtensions,
}

// API proposal:
// The TrackSectionModel (mapping of the SQL table) management becomes transparent here
let mut ts = TrackSection::retrieve((3, "TD0"), conn).await?.unwrap();
assert_eq!(ts.length, 25000.); // lol
ts.length *= 2;
let ts2 = ts.create(conn).await?; // No builder to design (hopefully)
assert_eq!(ts2.length, 50000.); // lol^2
ts.delete(conn).await?;
```

### Copy

A copy function that generates a changeset with pre-filled values may prove useful. Any thoughts?

```rust
let d: Document = ...;
d.duplicate()
 .content_type("json")
 .create(conn)
 .await?;
```

## Batch operations

```rust
#[derive(Debug, Default, Identifiable, Model, Queryable)]
#[model(create, delete, retrieve, batch_retrieve, batch_delete, batch_update)] // <<<
pub struct Document { ... }

let docs: Vec<Document> = Document::batch_retrieve(conn, [1, 2, 3, 4].iter()).await?; // docs.len() may != 4
let _: HashMap<i64, Document> = Document::batch_retrieve_map(conn, [1, 2, 3, 4].iter()).await?;

// checks that all 4 documents could be retrieved or returns an error with the list
// of ids that do not exist
let docs: Vec<Document> = Document::batch_retrieve_all(conn, [1, 2, 3, 4].iter()).await?;
assert_eq!(docs.len(), 4);
let hdocs: HashMap<i64, Document> = Document::batch_retrieve_all_map(conn, [1, 2, 3, 4].iter()).await?;
assert_eq!(hdocs.len(), 4);

let _: Vec<Document> = Document::delete_batch(conn, docs.iter().map(|d| d.id)).await?;
let _: HashMap<i64, Document> = Document::delete_batch_map(conn, docs.iter().map(|d| d.id)).await?;
// no _all variant for delete

// locks infras 1, 2, 3
let infras: Vec<Infra> = Infra::changeset()
    .locked(true)
    .update_batch(conn, [1, 2, 3].iter())
    .await?;
// likewise for update_batch_map, update_batch_all, update_batch_all_map
```

---

Before starting to make a list of all traits + fns prototypes that the macros described here should generate, I'd like to hear your thoughts on both the general ideas and the API proposal above.

@flomonster

### LINKED ISSUE #5056: Editoast delete the old model v1 derive macro
Part of #4184

> [!CAUTION]
> Blocked by the end of TSV2

Once every model is using the new macro `ModelV2`, delete the old `Model` and s/ModelV2/Model. Better document how ModelV2 works.

leovalais:
closes https://github.com/OpenRailAssociation/osrd/issues/5056
closes https://github.com/OpenRailAssociation/osrd/issues/4184

Another refactor that celebrated its birthday 🥳


editoast/openapi.yaml
```diff
@@ -8527,7 +8527,7 @@ components:
       description: |-
         A search result item for a query with `object = "signal"`
 
-        **IMPORTANT**: Please note that any modification to this struct should be reflected in [crate::modelsv2::infra::Infra::clone]
+        **IMPORTANT**: Please note that any modification to this struct should be reflected in [crate::models::infra::Infra::clone]
```
woshilapin: Should these comments be shown in the OpenAPI? Maybe we can change some `///` into some `//`?

woshilapin (approved):

editoast/openapi.yaml
leovalais: I'm waiting for https://github.com/OpenRailAssociation/osrd/pull/8526 to be merged so that @younesschrifi doesn't have to deal with the conflicts. I'll probably have to redo the last commit entirely, I'll change the `///` at the same time.

younesschrifi: Thank you ! #8526 is merged 🙂

leovalais: Ill wait for https://github.com/OpenRailAssociation/osrd/pull/8712 as well.

woshilapin (approved):

emersion:
This needs to be rebased on top of `dev` so that the DCO status check is triggered.

Removed from the merge queue because it blocks other PRs.

emersion:
Apologies, it seems like the DCO check was missing only from the merge queue status checks, not in this PR checks.

PR #8762, opened 2024-09-05
front: restore vias shown in bold in output table

### LINKED ISSUE #8761: Missing bold vias
### What happened?

Vias in output table are no longer displayed in bold since befdbe1e94c789a6b280379d4c084abee7848e0f

Screenshot_20240905_093820

2817459389242205e867caf274c696782bd87940

anisometropie:
Restoring waypoint displayed in bold in the output table, effectively bringing peace and order back to the Galaxy.

boolean `isWaypoint` is built in two different ways in the input table and the output table.

when formatSuggestedViasToRowVias is called in the output table, operationalPoints already have `isWaypoint`, but in the input table, `isWaypoint` is computed in formatSuggestedViasToRowVias itself.

so we need to make sure not to override it when we are in the output table case.

closes #8761

no longer reverts befdbe1e94c789a6b280379d4c084abee7848e0f.

### DIFF
--- a/front/src/modules/timesStops/helpers/utils.ts
+++ b/front/src/modules/timesStops/helpers/utils.ts
@@ -31,7 +31,7 @@ import {
 } from '../types';
 
 export const formatSuggestedViasToRowVias = (
-  operationalPoints: SuggestedOP[],
+  operationalPoints: (SuggestedOP & { isWaypoint?: boolean })[],
   pathSteps: PathStep[],
   t: TFunction<'timesStops', undefined>,
   startTime?: IsoDateTimeString,
@@ -95,7 +95,7 @@ export const formatSuggestedViasToRowVias = (
       name: name || t('waypoint', { id: filteredOp.opId }),
       stopFor,
       theoreticalMargin,
-      isWaypoint: pathStep !== undefined,
+      isWaypoint: op.isWaypoint || pathStep !== undefined,
     };
   });
 };


emersion:
If path steps are _really_ required here, using path steps from the store is incorrect. The store contains stale information. The path steps should come from the train schedule instead.

emersion:
Note that I don't think path steps are really required here. It would be better to mark in bold steps with either non-null `arrival` or non-null `stopFor`.

SharglutDev:
> Note that I don't think path steps are really required here. It would be better to mark in bold steps with either non-null `arrival` or non-null `stopFor`.

Maybe it needs to be clarified by a PO but in the input table, ops are bold as long as they are a waypoint/pathStep from `trainSchedule.path` (with constraint or not) so I think it should be the same in the output. If that's the case, using a non `arrival` or `stop_for` wouldn't work.

emersion:
Right. Then add a new bool field in path steps to indicate whether it's a waypoint?

emersion (approved):
Nice, LGTM!

emersion:
The caller casts to `SuggestedOP` which is too narrow, fixed that in https://github.com/OpenRailAssociation/osrd/pull/9016.

achrafmohye (approved):
LGTM !

Akctarus (approved):
Tested, LGTM

SharglutDev (approved):
Lgmt and tested.

Maybe in `useOutputTableData.ts` l.145, we should change the cast for : 
`allWaypoints as (SuggestedOP & { isWaypoint: boolean })[],`

emersion:
> Maybe in useOutputTableData.ts l.145, we should change the cast for :
allWaypoints as (SuggestedOP & { isWaypoint: boolean })[],

The cast is dropped completely in https://github.com/OpenRailAssociation/osrd/pull/9016.

PR #8766, opened 2024-09-05
core: stdcm: simplify stdcm edge builder

eckter:
Turns out, starting the path with a node instead of an edge can make things much easier than before


Erashin (approved):
Yaaayyyy

PR #8769, opened 2024-09-05
core: pathfinding: log a little extra info about failures

eckter:
We barely logged anything on pathfindings. We now log:

1. success
2. no path found at all, including the last reached waypoint index
3. no path found because of [constraint type / error type]

Not much, but it can still help

This PR also fixes the "couldn't find route" error, it had an irrelevant type.

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/graph/Pathfinding.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/graph/Pathfinding.kt
@@ -1,6 +1,7 @@
 package fr.sncf.osrd.graph
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings
+import fr.sncf.osrd.api.api_v2.pathfinding.pathfindingLogger
 import fr.sncf.osrd.api.pathfinding.constraints.ConstraintCombiner
 import fr.sncf.osrd.reporting.exceptions.ErrorType
 import fr.sncf.osrd.reporting.exceptions.OSRDError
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/fr/sncf/osrd/graph/Pathfinding.kt
```diff
@@ -184,15 +186,23 @@ class Pathfinding<NodeT : Any, EdgeT : Any, OffsetType>(
...
         }
+        var maxReachedTarget = 0
         val start = Instant.now()
         while (true) {
             if (Duration.between(start, Instant.now()).toSeconds() >= timeout)
                 throw OSRDError(ErrorType.PathfindingTimeoutError)
-            val step = queue.poll() ?: return null
+            val step = queue.poll()
+            if (step == null) {
+                pathfindingLogger.info(
```
bougue-pe, line 196: Should we downgrade log level to debug for this (and maybe some others on happy path)?
I wonder if it'll spam logs (maybe up the aws costs a bit, but mostly make logs less readable).

bougue-pe (approved):
LGTM (feel free to resolve comment)

Erashin (approved):

core/src/main/kotlin/fr/sncf/osrd/graph/Pathfinding.kt
eckter: The logs in this PR can only display something once per request, which doesn't feel excessive. (per line, maybe 3 lines total per request max over the PR)

If we want to be more careful about what we log, maybe we should start with the warnings when loading an infra... Actually I'll open an issue for this

eckter: https://github.com/OpenRailAssociation/osrd/issues/8773

PR #8772, opened 2024-09-05
front: turn on eslint recommended type checked hints

emersion:
See https://typescript-eslint.io/getting-started/typed-linting/

- [x] `no-unsafe-return` (#8884)
- [ ] `no-unsafe-member-access` (43 errors)
- [ ] `no-unsafe-argument` (#10234)
- [ ] `no-unsafe-assignment` (71 errors)
- [x] `no-unsafe-call` (3 errors, #8923)
- [ ] `no-misused-promises`
- [ ] `no-floating-promises`
- [ ] `restrict-template-expressions`
- [ ] `only-throw-error`
- [ ] `require-await`
- [ ] More?

### DIFF
--- a/front/.eslintrc
+++ b/front/.eslintrc
@@ -12,7 +12,7 @@
     "plugin:import/recommended",
     "plugin:import/typescript",
     "prettier",
-    "plugin:@typescript-eslint/base"
+    "plugin:@typescript-eslint/recommended-type-checked"
   ],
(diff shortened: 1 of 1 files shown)


emersion:
We've switched to oxlint, so this PR is no longer relevant. We still need to enable all recommended lints, I've opened https://github.com/OpenRailAssociation/osrd/issues/17424.

Issue #8628, opened 2024-08-29
The waypoint multi-select modal is dynamic


SharglutDev:
### Description and goal

Now that we can display the waypoints modal, we should be able to save the changes and update the manchette accordingly.

### Acceptance criteria

- [x] By default, the manchette should only display waypoints with these CH codes : "BV", "00", "" and vias added by map click (no ch)
- [x] In proportional mode, hiding a waypoint shouldn't change their position in the manchette
- [x] In linear mode, hiding a waypoint moves the following ones towards the previous one
- [x] The configuration should be saved in local storage so that when coming back to the same scenario/train, the displayed waypoints are the same
- [x] When deleting a (or multiple) project/study/scenario, the localStorage should be cleaned for keys containing the corresponding timetable id's
- [x] The space time chart should still resize according how many waypoints are displayed in the manchette

Tguisnet:
The main work is done,
however there is still a strange behavior for the linear case, hiding the first or last points (required),
and the warning when we only have 2 points could hide some points (optionnal)

Tguisnet:
To be more precise : 
If we hide the first, or the last points, we should truncate the results.
It means we should have a normal GET on the path of the manchette (with the beginning and the end of the line, with the train name, ...)
We should not see a line in the GET going outside the path of the manchette.
The behavior should be the same in linear or proportionnal mode.

emersion:
Eh, is this really the intended behavior? Even if previous OPs are hidden, it's not like the train starts at the first visible OP. Makes much more sense to make this clear by showing the continuation of the path past the first/last visible OP.

Castavo:
After #9513 and #9592, it is done 😁

Issue #8627, opened 2024-08-29
Add multi-select modal for waypoints in manchette


SharglutDev:
### Description and goal

In the manchette, we want to be able to select/unselect multiple waypoints to hide/display them from the component.

This first issue needs to add the possibility to open this modal and display all the waypoints even if they are not displayed in the current manchette.

### Acceptance criteria

- [x] The modal should take the whole viewport
- [x] Pressing escape on keyboard should also close the modal (but not clicking outside of it)
- [x] The backdrop style will be re-used on other components and need to be a variable 
- [x] The modal, for each waypoint, should contain their offset on path, name, ch code and a checkbox on the left
- [x] There should be a button to select/unselect all waypoints at once
- [x] This multi-select icon should change depending on how many waypoints are checked (none > `checked: false`, at least one > `indeterminate` or all > `checked`)
- [x] If the multi-select icon is on `indeterminate`, a click selects all waypoints 
- [x] Pressing shift + click should allow the user to select multiple waypoints at once
- [x] The modal design should match the [sketch](https://www.sketch.com/s/a26d84be-5fb2-41e4-9e84-96d90f46a997/a/elLZPzm)
- [x] When opening the modal, only the waypoints presents in the current manchette should be checked
- [x] If there is less than 2 waypoints checked, a [warning message](https://www.sketch.com/s/a26d84be-5fb2-41e4-9e84-96d90f46a997/a/aM5Orwz) should be displayed at the bottom of the modal 
- [x] Clicking on "Validate" on the form when the warning is displayed should make it wizz (same as stdcm behavior)
- [x] The modal in this issue is "static", validating it won't update the manchette

PR #8775, opened 2024-09-05
front: add option menu to manchette

### LINKED ISSUE #8687: Manchette : add an options menu
### Description and goal

The manchette will grow in complexity and we need to have a menu to toggle different options that will come in the future.

### Acceptance criteria

- [ ] We should display a button at the top of the manchette that toggles an options menu
- [ ] The only option in this menu should be the one to open the waypoints modal
- [ ] The menu design should match the [sketch](https://www.sketch.com/s/a26d84be-5fb2-41e4-9e84-96d90f46a997/a/4oAMRkz)

SharglutDev:
close #8687 

As requested by @thibautsailly, the manchette menu button should remain hidden until the waypoint modal is completed (#8627 and #8628)

To test the component, uncomment it


front/src/utils/hooks/useModalFocusTrap.ts
```diff
@@ -14,10 +14,10 @@ export default function useModalFocusTrap(
     const focusableElements = modalElement?.querySelectorAll(
       // last declaration stands for all elements not natively focusable like li
       'input, button, [tabindex]:not([tabindex="-1"])'
-    ) as NodeListOf<Element>;
+    );
 
-    const firstElement = focusableElements[0] as HTMLElement;
-    const lastElement = focusableElements[focusableElements?.length - 1] as HTMLElement;
+    const firstElement = focusableElements?.[0];
+    const lastElement = focusableElements?.[focusableElements?.length - 1];
```
kmer2016, line 20:
```suggestion
    const firstElement = focusableElements?.at(0);
    const lastElement = focusableElements?.at(-1);
```

front/src/utils/hooks/useModalFocusTrap.ts
```diff
@@ -14,10 +14,10 @@ export default function useModalFocusTrap(
     const focusableElements = modalElement?.querySelectorAll(
       // last declaration stands for all elements not natively focusable like li
       'input, button, [tabindex]:not([tabindex="-1"])'
-    ) as NodeListOf<Element>;
+    );
 
-    const firstElement = focusableElements[0] as HTMLElement;
-    const lastElement = focusableElements[focusableElements?.length - 1] as HTMLElement;
```
kmer2016, line 20: Just to know, any reason to prefer to move the HTMLElement cast from here in the definition to where the variables are used (lines 32 & 35) ?

front/public/locales/en/simulation.json
```diff
@@ -60,6 +63,7 @@
   "stopTime": "Arrival",
   "timetable": "Timetable",
   "toggleElectricalProfileLegend": "Toggle electrical profiles legend",
+  "toggleManchetteMenu": "Toggle manchette menu",
```
Yohh, line 67:
```suggestion
  "toggleManchetteMenu": "Toggle the manchette menu Display",
```

front/public/locales/en/simulation.json
```diff
@@ -20,6 +20,9 @@
   "isWorking": "Generating simulation...",
   "labels": "Tags",
   "loadingTrain": "Loading train",
+  "manchetteSettings": {
+    "waypointsVisibility": "show/hide OP"
```
Yohh, line 24:
```suggestion
    "waypointsVisibility": "Change the visibility of OPs"
```
**or below and**
```suggestion
    "waypointsVisibility": "Show/hide OP"
```

front/public/locales/fr/simulation.json
```diff
@@ -20,6 +20,9 @@
   "isWorking": "Calcul de la simulation en cours…",
   "labels": "Étiquettes",
   "loadingTrain": "Chargement du train",
+  "manchetteSettings": {
+    "waypointsVisibility": "Modifier la visibilité des PRs"
+  },
```
Yohh, line 25:
```suggestion
  "manchetteSettings": {
    "waypointsVisibility": "Afficher/masquer les PRs"
  },
```
**or above**

Yohh (commented):
nice, a few suggestions first

front/src/utils/hooks/useModalFocusTrap.ts
SharglutDev: As I now add the `?`, it was to avoid writing `HTMLElement | undefined`. But maybe I can it for a `!` instead of the cast as recommanded by @emersion.

front/public/locales/en/simulation.json
SharglutDev: For the `show/hide OP`, it was a request from @thibautsailly (even though I also don't like it very much)

front/public/locales/fr/simulation.json
SharglutDev: `Modifier la visibilité des PRs` is the text on the sketch

Yohh: so this one seems better

@thibautsailly ?

front/public/locales/en/simulation.json
SharglutDev: I followed how it was done by a wise man for the key `"toggleSpeedSpaceSettings": "Toggle chart settings"` :)

SharglutDev:
Also I was wondering about the reusability of the menu : 
The UI from the manchette menu and the one from an op click on the manchette (see [sketch](https://www.sketch.com/s/a26d84be-5fb2-41e4-9e84-96d90f46a997/a/OeJV9yG#Inspect)) are very close. Do you think it would be better to have a "common" component for these cases and some custom props ? Like the width, items etc
But if yes, wouldn't it be better to have a component like this in osrd-ui ? (even though we didn't have a request for that)

front/public/locales/en/simulation.json
Yohh: this guy isn't wise at all!

Yohh: seriously, I think "Toggle chart settings" should be modified too

Yohh:
> Also I was wondering about the reusability of the menu : The UI from the manchette menu and the one from an op click on the manchette (see [sketch](https://www.sketch.com/s/a26d84be-5fb2-41e4-9e84-96d90f46a997/a/OeJV9yG#Inspect)) are very close. Do you think it would be better to have a "common" component for these cases and some custom props ? Like the width, items etc But if yes, wouldn't it be better to have a component like this in osrd-ui ? (even though we didn't have a request for that)

I'm a bit confused, in this case why don't we create a component in OSRD-UI? witch could be called as a child of manchette and used for each case

SharglutDev:
What I meant was : 

The manchette menu 
Capture d’écran 2024-09-06 à 14 14 58

and the OP menu
Capture d’écran 2024-09-06 à 14 14 33

are looking exactly the same except the items inside and the width of the menu. Maybe it should be better to have only one generic component for that to avoid code duplication.

Second question is : do we want it in osrd-ui or do we consider that it's purely an osrd component and we keep in osrd code base.

front/src/utils/hooks/useModalFocusTrap.ts
SharglutDev: Oups it seems that `NodeListOf<Element>` isn't associated as an `Array` so I can't use `.at()`on it

front/public/locales/en/simulation.json
SharglutDev: Changed

front/public/locales/en/simulation.json
SharglutDev: Changed it with an uppercase first

front/public/locales/fr/simulation.json
SharglutDev: @thibautsailly wants to keep the `Modifer la vi...`

emersion:
I've taken the liberty to rename the PR, I hope that's fine.

front/src/common/OSRDMenu.tsx
```diff
@@ -0,0 +1,88 @@
...
+  icon?: React.ReactNode;
+};
+
+type OSRDMenuProps = {
+  width: string;
+  openerRef: React.RefObject<HTMLButtonElement>;
+  items: MenuOption[];
+  onItemClick: (label: string) => void;
+  isMenuOpen: boolean;
+  setIsMenuOpen: (isOpen: boolean) => void;
```
Yohh:
```suggestion
  setIsMenuOpen: React.Dispatch<React.SetStateAction<boolean>>;
```

front/src/styles/scss/common/components/_osrdMenu.scss
```diff
@@ -0,0 +1,45 @@
...
+  position: absolute;
+  left: -0.1875rem;
+  z-index: 10;
+  background-color: var(--grey5);
+  outline: 0.0625rem solid var(--grey30);
+  border-radius: 0.5rem;
+  box-shadow:
+    0px 6px 21px -5px rgba(101, 169, 232, 0.341810533216783),
+    0px 16px 30px -5px rgba(0, 0, 0, 0.2),
+    0px 3px 5px -2px rgba(0, 0, 0, 0.1);
```
Yohh:
```suggestion
    0 0.375rem 1.3125rem -0.3125rem rgba(101, 169, 232, 0.341810533216783),
    0 1rem 1.875rem -0.3125rem rgba(0, 0, 0, 0.2),
    0 0.1875rem 0.3125rem -0.125rem rgba(0, 0, 0, 0.1);
```

are you sure about the alpha `0.341810533216783` ?

front/src/utils/hooks/useModalFocusTrap.ts
```diff
@@ -26,13 +26,13 @@ export default function useModalFocusTrap(
...
-      if (event.key === 'Tab') {
+      if (event.key === 'Tab' && firstElement && lastElement) {
         if (event.shiftKey && document.activeElement === firstElement) {
           event.preventDefault();
-          lastElement.focus();
+          (lastElement as HTMLElement).focus();
         } else if (!event.shiftKey && document.activeElement === lastElement) {
           event.preventDefault();
-          firstElement.focus();
+          (firstElement as HTMLElement).focus();
```
Yohh, line 35: you can set upper:
```js
    const focusableElements = modalElement?.querySelectorAll(
      // last declaration stands for all elements not natively focusable like li
      'input, button, [tabindex]:not([tabindex="-1"])'
    ) as NodeListOf<HTMLElement> | undefined;
```

then remove aliases in lines 32 & 35

front/src/utils/hooks/useModalFocusTrap.ts
SharglutDev: As @Yohh asked for it too, I put the cast back at the top.

front/src/utils/hooks/useModalFocusTrap.ts
SharglutDev: Done

front/src/utils/hooks/useModalFocusTrap.ts
SharglutDev: Ok new change. As eslint was warning for the cast, I added the `HTMLElement` directly on `querySelectorAll`.

front/src/styles/scss/common/components/_osrdMenu.scss
SharglutDev: Good point for the rems. All this box-shadow comes from the mockup so I just kept the original values.

front/src/common/OSRDMenu.tsx
SharglutDev: When possible I prefer to use the setter with `(state) => void` than the actual type. I use it only when need to do `(prev) => ***` inside (and I think we do like that in a lot of components already).

front/src/common/OSRDMenu.tsx
```diff
@@ -0,0 +1,88 @@
...
+  label: string;
+  disabled?: boolean;
+  icon?: React.ReactNode;
+};
+
+type OSRDMenuProps = {
+  width: string;
+  openerRef: React.RefObject<HTMLButtonElement>;
+  items: MenuOption[];
+  onItemClick: (label: string) => void;
```
kmer2016: I believe it would be better for each item to handle its own action rather than using a global handler for all items. Since each option has a distinct action, keeping a global handler forces us to manage different cases in the same function, making the code less flexible and less readable (e.g., using if or switch statements).

kmer2016 (commented):
In the mock-up, the button for displaying the menu is located on the Header. In the implementation, however, it's outside. Are you waiting for the new version of the headline?

image

image

SharglutDev:
> In the mock-up, the button for displaying the menu is located on the Header. In the implementation, however, it's outside. Are you waiting for the new version of the headline?
> 
> image
> 
> image

Yes that's exactly why. For now the space time chart header is inside the osrd-ui components and I can't access it but the other issue will be done, I would be able to create it on osrd side and put the button at the right place.

front/src/common/OSRDMenu.tsx
SharglutDev: That's much better, you're right. Done.

front/src/modules/simulationResult/components/SpaceTimeChart/ManchetteMenuButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+  useModalFocusTrap(menuRef, closeMenu);
+
+  useEffect(() => {
+    const handleClickOutside = (event: MouseEvent) => {
+      if (
+        menuRef.current &&
+        menuButtonRef.current &&
+        !menuRef.current.contains(event.target as Node) &&
+        !menuButtonRef.current.contains(event.target as Node)
+      ) {
```
kmer2016:
```suggestion
      if (
        menuRef.current?.contains(event.target as Node) &&
        menuButtonRef.current?.contains(event.target as Node)
      ) {
```

front/src/modules/simulationResult/components/SpaceTimeChart/ManchetteMenuButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+        onClick: () => {
+          closeMenu(); // TODO : in #8628, change this to open the waypoints modal
+        },
+      },
+    ],
+    [t]
+  );
+
+  useModalFocusTrap(menuRef, closeMenu);
+
```
kmer2016, line 31: I see that we already have a custom hook to handle the click ouside. What do you think to refactor to allow to use it in this case too ? Can be done in another PR

kmer2016 (commented):
Just small suggestions

front/src/modules/simulationResult/components/SpaceTimeChart/ManchetteMenuButton.tsx
Yohh: at least a TODO should be enough

front/src/modules/simulationResult/components/SpaceTimeChart/ManchetteMenuButton.tsx
```diff
@@ -15,6 +17,19 @@ const SpaceTimeChartHeader = () => {
...
+      {
+        title: t('manchetteSettings.waypointsVisibility'),
+        icon: <Eye />,
+        onClick: () => {
+          closeMenu(); // TODO : in #8628, change this to open the waypoints modal
+        },
+      },
+    ],
+    [t]
+  );
```
Yohh: do we really need to memoize?

front/src/modules/simulationResult/components/SpaceTimeChart/ManchetteMenuButton.tsx
SharglutDev: Done, thank you (the `!` at the beginning of both lines were missing)

front/src/modules/simulationResult/components/SpaceTimeChart/ManchetteMenuButton.tsx
SharglutDev: Oh yes, good point. It will indeed need a refacto as I need to pass 2 refs instead of one.
I added a TODO.

front/src/modules/simulationResult/components/SpaceTimeChart/ManchetteMenuButton.tsx
SharglutDev: You're probably right. As we pass it in props to the menu, I was afraid it might result to some unwant rerenders if something else in the page changes. What do you think @kmer2016 ?

kmer2016: I quite agree with @Yohh , I don't think `useMemo` is necessary here. The `menuItems` array is relatively simple and contains elment witth simple property. Removing `useMemo` would simplify the code without significantly impacting performance.

kmer2016 (approved):

Yohh (approved):
LGTM

PR #8785, opened 2024-09-06
front: turn on @typescript-eslint/no-unnecessary-type-assertion lint

emersion:
First part of https://github.com/OpenRailAssociation/osrd/pull/8772

Enable typed hinting so that ESLint has information about types
and can provide more feedback. Only enable a single typed lint
rule for now (rest will be a lot more work to fix).

The rule @typescript-eslint/no-unnecessary-type-assertion prevents
unnecessary type assertions (e.g. casting a string to a string).
Reducing type assertions is nice because type assertions override
type checking and thus may hide type mismatches.

### DIFF
--- a/front/src/applications/editor/components/EntitySumUp.tsx
+++ b/front/src/applications/editor/components/EntitySumUp.tsx
@@ -328,12 +328,7 @@ const EntitySumUp = ({ entity, id, objType, classes, status, error }: EntitySumU
       setState({ type: 'loading' });
 
       if (!entity) {
-        entity = await getEntity(
-          infraID as number,
-          id as string,
-          objType as EditoastType,
-          dispatch
-        );
+        entity = await getEntity(infraID as number, id, objType as EditoastType, dispatch);
       }
 
       const additionalEntities = await getAdditionalEntities(infraID as number, entity, dispatch);
--- a/front/.eslintrc
+++ b/front/.eslintrc
@@ -11,11 +11,13 @@
     "airbnb",
     "plugin:import/recommended",
     "plugin:import/typescript",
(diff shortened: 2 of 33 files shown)


front/src/applications/editor/components/EntitySumUp.tsx
```diff
@@ -328,12 +328,7 @@ const EntitySumUp = ({ entity, id, objType, classes, status, error }: EntitySumU
       setState({ type: 'loading' });
 
       if (!entity) {
-        entity = await getEntity(
-          infraID as number,
-          id as string,
-          objType as EditoastType,
-          dispatch
-        );
+        entity = await getEntity(infraID as number, id, objType as EditoastType, dispatch);
```
kmer2016, line 331: I think there's a type mismatch on `id`. `getEntity` expects `id` as a `string`, but id can be `undefined`.

emersion: The compiler doesn't complain at least.

I believe the compiler is smart enough to understand that `entity === undefined` since we're inside the if, and that's only possible if the argument was `{ id: string; objType: EditoastType; entity?: undefined }` (as opposed to `{ entity: EditorEntity; id?: undefined; objType?: undefined }`).

Yohh (approved):

front/src/applications/editor/components/EntitySumUp.tsx
kmer2016: Ah, it's my fault. My editor complained, and it misled me. But you're right.

kmer2016 (approved):
LGTM !

PR #8786, opened 2024-09-06
core, editoast: add work schedules to v2 conflict detection endpoint

### LINKED ISSUE #8589: core: conflict detection: take work schedules into account in V2 conflict detection endpoint
# Description

Take into account work schedules in V2 conflict detection endpoint.

# Goal

Being able to send those info to the front end.

# AC
- [x] The work schedules have been added to conflict detection endpoint request payload
- [x] The train/train and train/work_schedule conflicts are sent

Erashin:
Fixes #8589.

Add work shedule list to conflict detection endpoint. Put their ids into the payload to be able to send them back to editoast in the corresponding conflict groups.
If a conflict group is found on a zone between 2 times, but all the conflicts are work schedules, ignore it.

### DIFF
--- a/core/src/main/java/fr/sncf/osrd/conflicts/Conflicts.kt
+++ b/core/src/main/java/fr/sncf/osrd/conflicts/Conflicts.kt
@@ -10,6 +10,8 @@ import fr.sncf.osrd.standalone_sim.result.ResultTrain.SpacingRequirement
 import kotlin.math.max
 import kotlin.math.min
 
+const val DEFAULT_WORK_SCHEDULE_ID: Long = -1
+
 interface SpacingTrainRequirement {
     val trainId: Long
     val spacingRequirements: Collection<SpacingRequirement>
@@ -25,6 +27,7 @@ interface ResourceRequirement {
     val endTime: Double
 }
 
+// TODO: when dropping v1, remove these structs and directly use Requirements
 class TrainRequirements(
     @Json(name = "train_id")
     override val trainId: Long, // Not the usual RJS ids, but an actual DB id
@@ -33,6 +36,23 @@ class TrainRequirements(
     @Json(name = "routing_requirements")
     override val routingRequirements: Collection<RoutingRequirement>,
 ) : SpacingTrainRequirement, RoutingTrainRequirement
+
+class Requirements(
+    val id: RequirementId,
+    val spacingRequirements: Collection<SpacingRequirement>,
+    val routingRequirements: Collection<RoutingRequirement>,
+)
+
+data class RequirementId(
+    // Either a train db id or a work schedule db id
+    val id: Long,
+    val type: RequirementType,
+)
+
+enum class RequirementType {
+    TRAIN,
+    WORK_SCHEDULE
+}
 
 data class ConflictProperties(
     // If there are conflicts, minimum delay that should be added to the train so that there are no
@@ -46,7 +66,11 @@ data class ConflictProperties(
 )
 
 fun detectConflicts(trainRequirements: List<TrainRequirements>): List<Conflict> {
-    val res = incrementalConflictDetector(trainRequirements).checkConflicts()
+    return detectRequirementConflicts(convertTrainRequirements(trainRequirements))
+}
+
+fun detectRequirementConflicts(requirements: List<Requirements>): List<Conflict> {
+    val res = incrementalConflictDetectorFromRequirements(requirements).checkConflicts()
     return mergeConflicts(res)
 }
 
@@ -88,10 +112,16 @@ interface IncrementalConflictDetector {
 fun incrementalConflictDetector(
     trainRequirements: List<TrainRequirements>
 ): IncrementalConflictDetector {
-    return IncrementalConflictDetectorImpl(trainRequirements)
+    return IncrementalConflictDetectorImpl(convertTrainRequirements(trainRequirements))
 }
 
-class IncrementalConflictDetectorImpl(trainRequirements: List<TrainRequirements>) :
+fun incrementalConflictDetectorFromRequirements(
+    requirements: List<Requirements>
+): IncrementalConflictDetector {
+    return IncrementalConflictDetectorImpl(requirements)
+}
+
+class IncrementalConflictDetectorImpl(requirements: List<Requirements>) :
     IncrementalConflictDetector {
     private val spacingZoneRequirements =
         mutableMapOf<String, MutableList<SpacingZoneRequirement>>()
@@ -99,22 +129,22 @@ class IncrementalConflictDetectorImpl(trainRequirements: List<TrainRequirements>
         mutableMapOf<String, MutableList<RoutingZoneRequirement>>()
 
     init {
-        generateSpacingRequirements(trainRequirements)
-        generateRoutingRequirements(trainRequirements)
+        generateSpacingRequirements(requirements)
+        generateRoutingRequirements(requirements)
     }
 
     data class SpacingZoneRequirement(
-        val trainId: Long,
+        val id: RequirementId,
         override val beginTime: Double,
         override val endTime: Double,
     ) : ResourceRequirement
 
-    private fun generateSpacingRequirements(trainRequirements: List<SpacingTrainRequirement>) {
+    private fun generateSpacingRequirements(requirements: List<Requirements>) {
         // organize requirements by zone
-        for (req in trainRequirements) {
+        for (req in requirements) {
             for (spacingReq in req.spacingRequirements) {
                 val zoneReq =
-                    SpacingZoneRequirement(req.trainId, spacingReq.beginTime, spacingReq.endTime)
+                    SpacingZoneRequirement(req.id, spacingReq.beginTime, spacingReq.endTime)
                 spacingZoneRequirements.getOrPut(spacingReq.zone!!) { mutableListOf() }.add(zoneReq)
             }
         }
@@ -134,10 +164,10 @@ class IncrementalConflictDetectorImpl(trainRequirements: List<TrainRequirements>
         val config: RoutingZoneConfig,
     ) : ResourceRequirement
 
-    private fun generateRoutingRequirements(trainsRequirements: List<RoutingTrainRequirement>) {
+    private fun generateRoutingRequirements(requirements: List<Requirements>) {
         // reorganize requirements by zone
-        for (trainRequirements in trainsRequirements) {
-            val trainId = trainRequirements.trainId
+        for (trainRequirements in requirements) {
+            val trainId = trainRequirements.id.id
             for (routeRequirements in trainRequirements.routingRequirements) {
                 val route = routeRequirements.route!!
                 var beginTime = routeRequirements.beginTime
@@ -174,12 +204,28 @@ class IncrementalConflictDetectorImpl(trainRequirements: List<TrainRequirements>
         val res = mutableListOf<Conflict>()
         for (entry in spacingZoneRequirements) {
             for (conflictGroup in detectRequirementConflicts(entry.value) { _, _ -> true }) {
-                val trains = conflictGroup.map { it.trainId }
                 val beginTime = conflictGroup.minBy { it.beginTime }.beginTime
                 val endTime = conflictGroup.maxBy { it.endTime }.endTime
+                // If there are only conflicting work schedules, skip conflict group
+                if (conflictGroup.all { it.id.type == RequirementType.WORK_SCHEDULE }) {
+                    continue
+                }
+                val trains =
+                    conflictGroup.filter { it.id.type == RequirementType.TRAIN }.map { it.id.id }
+                val workSchedules =
+                    conflictGroup
+                        .filter { it.id.type == RequirementType.WORK_SCHEDULE }
+                        .map { it.id.id }
                 val conflictReq = ConflictRequirement(entry.key, beginTime, endTime)
                 res.add(
-                    Conflict(trains, beginTime, endTime, ConflictType.SPACING, listOf(conflictReq))
+                    Conflict(
+                        trains,
+                        workSchedules,
+                        beginTime,
+                        endTime,
+                        ConflictType.SPACING,
+                        listOf(conflictReq)
+                    )
                 )
             }
         }
@@ -225,17 +271,24 @@ class IncrementalConflictDetectorImpl(trainRequirements: List<TrainRequirements>
         for (otherReq in requirements) {
             val beginTime = max(req.beginTime, otherReq.beginTime)
             val endTime = min(req.endTime, otherReq.endTime)
-            val conflictReq = ConflictRequirement(req.zone, beginTime, endTime)
-            if (beginTime < endTime)
+            if (beginTime < endTime) {
+                val trainIds = mutableListOf<Long>()
+                val workScheduleIds = mutableListOf<Long>()
+                if (otherReq.id.type == RequirementType.WORK_SCHEDULE)
+                    workScheduleIds.add(otherReq.id.id)
+                else trainIds.add(otherReq.id.id)
+                val conflictReq = ConflictRequirement(req.zone, beginTime, endTime)
                 res.add(
                     Conflict(
-                        listOf(otherReq.trainId),
+                        trainIds,
+                        workScheduleIds,
                         beginTime,
                         endTime,
                         ConflictType.SPACING,
                         listOf(conflictReq)
                     )
                 )
+            }
         }
 
         return res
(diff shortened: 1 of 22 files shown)


editoast/src/core/conflict_detection.rs
```diff
@@ -49,19 +61,29 @@ pub struct Conflict {
...
 }
 
-#[derive(Debug, Clone, Copy, Serialize, Deserialize, ToSchema)]
+#[derive(Debug, Clone, Deserialize, Serialize, ToSchema)]
+pub struct ConflictWithWorkSchedules {
+    #[serde(flatten)]
+    pub conflict: Conflict,
+    /// List of work schedule ids involved in the conflict
+    pub work_schedule_ids: Vec<i64>,
+}
```
woshilapin: This creates a type that might be incoherent (`conflict.conflict_type` must be `WorkSchedule` when `work_schedule_ids` is not empty). It feels that `ConflictType` could be modified differently (something like this).

```rust
pub enum ConflictType {
  Spacing,
  Routing,
  WorkSchedule(Vec<i64>),
}
```

That said, to support this kind of structure, we might need to do the deserialization more manually 🤷 I could try to help/work on this but wanted to know other opinions first (cc @leovalais @flomonster @Tguisnet).

editoast/src/views/timetable.rs
```diff
@@ -338,14 +339,29 @@ async fn conflicts(
...
+        .conflicts
+        .into_iter()
+        .filter_map(|conflict_with_work_schedules| {
+            if conflict_with_work_schedules.conflict.conflict_type != ConflictType::WorkSchedule {
+                Some(conflict_with_work_schedules.conflict)
+            } else {
+                None
+            }
+        })
+        .collect();
```
woshilapin: I'm really not sure that we should be filtering in this endpoint. Work schedules are associated with an infra, this endpoint takes a timetable as an input, which is itself associated with an infra: semantically, this endpoint should return work schedule conflicts as well. Only the `front` knows the context in which it will be used and has the ability to decide if work schedule conflicts should be displayed or not. What do you think?

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/RequirementsParser.kt
```diff
@@ -66,3 +79,89 @@ fun parseRoutingRequirements(
...
+    workSchedulesRequest: WorkSchedulesRequest,
+    startTime: ZonedDateTime
+): Collection<Requirements> {
+    val delta = TimeDelta(between(startTime, workSchedulesRequest.startTime).toMillis())
+    return convertWorkSchedules(infra, workSchedulesRequest.workScheduleRequirements, delta)
+}
+
+/**
+ * Convert work schedules into timetable spacing requirements. This is not entirely semantically
+ * correct, but it lets us avoid work schedules like any other kind of time-bound constraint
```
woshilapin: Why do you say "This is not entirely semantically correct"? I believe a work schedule is really similar to a spacing requirement (reserving some zone during a certain time). Is there something else you have in mind with your comment?

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/RequirementsParser.kt
```diff
@@ -66,3 +79,89 @@ fun parseRoutingRequirements(
...
+    }
+    return res
+}
+
+/**
+ * Convert work schedules into timetable spacing requirements, without taking work schedule id into
+ * account. This is not entirely semantically correct, but it lets us avoid work schedules like any
+ * other kind of time-bound constraint
+ */
+fun convertWorkSchedules(
```
woshilapin: Is it intended to have a function with the exact same name than the one above, but with different input arguments? I know I'm not used to overloading (not available in the languages I deal with), but yet, I find this extremely misleading since both function don't even return the same type.

woshilapin (commented):
Not approving yet since I've made some comments that might need some feedback.

editoast/src/views/timetable.rs
Erashin: I'll let @flomonster answer this, but the reasoning was that it was for sure not planned to use this endpoint directly from the frontend with work_schedules, which is why we filter it in post-processing. Personally don't mind either way, adds very few modifications.

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/RequirementsParser.kt
Erashin: They're very similar, but they're not truly spacing requirements, so it's technically a hack, albeit it is 99% justified. Nothing more I had in mind with this. Do you think it's unneeded?

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/RequirementsParser.kt
Erashin: It was indeed intended. They technically perform the same operation, albeit for different structures : one is a collection of work_chedules, so it only outputs one Requirements, the other is a map of work_schedules, hence it outputs a collection of Requirements. Which is why I kept the same name for both. It also is used for the same reason : one converts the work_schedules in stdcm, the other in conflict_detection. Honestly think it's fine, especially since I don't really think there's a simple naming alternative. Don't hesitate to challenge this if you don't find the explanation satisfying.

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/RequirementsParser.kt
woshilapin: When talking with @Caracol3, he mentioned that when he was working on that kind of stuff, he really considered occupation of a track section by a train or by a work schedule the same way. So I would say the comment is not necessary.

Erashin: Works for me. I'll change that then.

core/src/main/java/fr/sncf/osrd/conflicts/Conflicts.kt
```diff
@@ -174,11 +207,26 @@ class IncrementalConflictDetectorImpl(trainRequirements: List<TrainRequirements>
...
         for (entry in spacingZoneRequirements) {
             for (conflictGroup in detectRequirementConflicts(entry.value) { _, _ -> true }) {
-                val trains = conflictGroup.map { it.trainId }
                 val beginTime = conflictGroup.minBy { it.beginTime }.beginTime
                 val endTime = conflictGroup.maxBy { it.endTime }.endTime
+                // If there are only conflicting work schedules, skip conflict group
+                if (conflictGroup.all { it.id.type == RequirementType.WORK_SCHEDULE }) {
+                    conflictLogger.warn(
+                        "Conflict between work schedules between $beginTime and $endTime"
+                    )
```
eckter: It's a little misleading to log this here. Because we can still have conflicting work schedules without ever printing this.

I'd prefer if we either don't log it at all, or log any conflict between two work schedules (and not just when there's no train involved)

core/src/main/java/fr/sncf/osrd/conflicts/Conflicts.kt
```diff
@@ -174,11 +207,26 @@ class IncrementalConflictDetectorImpl(trainRequirements: List<TrainRequirements>
...
+                }
+                val trains =
+                    conflictGroup.filter { it.id.type == RequirementType.TRAIN }.map { it.id.id }
+                val workSchedules =
+                    conflictGroup
+                        .filter { it.id.type == RequirementType.WORK_SCHEDULE }
+                        .map { it.id.id }
+                val conflictType =
+                    if (workSchedules.isNotEmpty()) ConflictType.WORK_SCHEDULE
+                    else ConflictType.SPACING
```
eckter: What if there's both spacing issues and work schedules in the same group? We seem to ignore this case for less complexity, but are we sure we can do that?

core/src/main/java/fr/sncf/osrd/conflicts/Conflicts.kt
```diff
@@ -221,11 +269,25 @@ class IncrementalConflictDetectorImpl(trainRequirements: List<TrainRequirements>
...
+                if (otherReq.id.type == RequirementType.WORK_SCHEDULE) {
+                    workScheduleIds.add(otherReq.id.id)
+                    conflictType = ConflictType.WORK_SCHEDULE
+                }
+                val conflictReq = ConflictRequirement(req.zone, beginTime, endTime)
                 res.add(
-                    Conflict(listOf(otherReq.trainId), beginTime, endTime, ConflictType.SPACING, listOf(conflictReq))
+                    Conflict(
+                        listOf(otherReq.id.id),
+                        workScheduleIds,
```
eckter: This isn't consistent, `trainIds` would contain the id even if it's a work schedule

core/src/main/java/fr/sncf/osrd/conflicts/Conflicts.kt
```diff
@@ -221,11 +269,25 @@ class IncrementalConflictDetectorImpl(trainRequirements: List<TrainRequirements>
         for (otherReq in requirements) {
             val beginTime = max(req.beginTime, otherReq.beginTime)
             val endTime = min(req.endTime, otherReq.endTime)
-            val conflictReq = ConflictRequirement(req.zone, beginTime, endTime)
-            if (beginTime < endTime)
+            if (beginTime < endTime) {
```
eckter, line 274: This method is only used in stdcm context, but we already have another way to handle work schedules. ~~Is it removed? (maybe I'll get the answer when reading the rest of the PR)~~

edit: it is indeed replaced

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/RequirementsParser.kt
eckter: I'll reopen because I still find it weird. 

~~The names could be different at least, if the difference is how many schedules it handles one should be singular and not plural.~~

edit: wait there's two plural versions and one singular version, this is confusing

I'm not completely convinced, we could easily name them differently. `convertWorkScheduleMap` / `convertWorkScheduleCollection`? Or `requirementFromWorkSchedules` / `requirementsFromWorkScheduleMap`?

Oh and the docstring seem to be mostly duplicated, but it doesn't quite match the version used in conflict detection

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/RequirementsParser.kt
```diff
@@ -66,3 +79,89 @@ fun parseRoutingRequirements(
...
+    for (range in workSchedule.trackRanges) {
+        val track = rawInfra.getTrackSectionFromName(range.trackSection) ?: continue
+        for (chunk in rawInfra.getTrackSectionChunks(track)) {
+            val chunkStartOffset = rawInfra.getTrackChunkOffset(chunk)
+            val chunkEndOffset = chunkStartOffset + rawInfra.getTrackChunkLength(chunk).distance
+            if (chunkStartOffset > range.end || chunkEndOffset < range.begin) continue
+            val zone = rawInfra.getTrackChunkZone(chunk)
+            if (zone == null) {
+                requirementsParserLogger.info(
+                    "Skipping part of work schedule [${workSchedule.startTime}; ${workSchedule.endTime}] because it is on a track not fully covered by routes: $track",
```
eckter: This line doesn't fit on my screen, it could be split on several lines

core/src/main/java/fr/sncf/osrd/conflicts/Conflicts.kt
Erashin: I'll probably remove the logs then. Theoretically, you could have two work_schedules not directly conflicting with each other in the same conflict_group (work_schedule conflicts with following train which conflicts with following work_schedule).

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/RequirementsParser.kt
Erashin: These are okay alternatives. Especially since it's confusing for 2 people already, I'll definitely do the change.

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/RequirementsParser.kt
Erashin: After further talk, this should also mean that there shouldn't be any distinction in the conflict type bewteen a spacing conflict and a work schedule conflict. I'm down to remove the distinction (i'll do it in a separate commit to revert in case you don't agree, but i think it'll be better).

core/src/main/java/fr/sncf/osrd/conflicts/Conflicts.kt
Erashin: > After further talk, this should also mean that there shouldn't be any distinction in the conflict type bewteen a spacing conflict and a work schedule conflict. I'm down to remove the distinction (i'll do it in a separate commit to revert in case you don't agree, but i think it'll be better).

This should handle it :)

core/src/main/java/fr/sncf/osrd/conflicts/Conflicts.kt
Erashin: Yep, I'll just put an empty list if it conflicts with a work schedule.

editoast/src/views/timetable.rs
Erashin: > After further talk, this should also mean that there shouldn't be any distinction in the conflict type bewteen a spacing conflict and a work schedule conflict. I'm down to remove the distinction (i'll do it in a separate commit to revert in case you don't agree, but i think it'll be better).

This should also resolve this problem :)

editoast/src/core/conflict_detection.rs
Erashin: > After further talk, this should also mean that there shouldn't be any distinction in the conflict type bewteen a spacing conflict and a work schedule conflict. I'm down to remove the distinction (i'll do it in a separate commit to revert in case you don't agree, but i think it'll be better).

This should also resolve this problem :)

editoast/src/views/timetable.rs
flomonster: Actually we provide an empty list of work schedule. Having workschedule conflict is completly unexpected here. We should either remove this filter or assert core response :shrug:

editoast/src/core/conflict_detection.rs
woshilapin: Well not really, we still can define a `Routing` with a non-empty `Vec` with is incorrect. Should we do something like the following?
```rust
pub enum ConflictType {
  Spacing(Vec<i64>),
  Routing,
}
```

editoast/src/views/timetable.rs
```diff
@@ -338,14 +338,22 @@ async fn conflicts(
...
+    let conflict_detection_response = conflict_detection_request.fetch(&core_client).await?;
 
-    Ok(Json(conflicts.conflicts))
+    let conflicts = conflict_detection_response
+        .conflicts
+        .into_iter()
+        .map(|c| c.conflict)
+        .collect();
+
+    Ok(Json(conflicts))
```
woshilapin: Is there a reason for transforming into an iterator to recollect just after? Can't you just do the following?

```suggestion
    Ok(Json(conflict_detection_response.conflicts))
```

edit: I believe it's an artifact of the evolution of the code since your last fixup commit.

Erashin: You might have missed it, but with this, I'm removing the work_schedule_ids data to not have to send an empty list to the frontend for no reason (`ConflictDetectionResponse` contains `conflicts: Vec<ConflictWithWorkSchedule>`, and I remove the `work_schedule_ids` field to only send a `Vec<Conflict>` to the frontend).

eckter (approved):

editoast/src/core/conflict_detection.rs
woshilapin: We discussed with @Erashin and TLDR, concluded that all solutions seems to bring considerable complexity so we'll keep a simple data structure that might be corrupted but hey 🤷. 

First of all, putting `work_schedule_ids` in the enum `ConflictType`, when `train_ids` involved in the conflict are already in `Conflict` doesn't make sense: both categories of ids should probably live in the same place. If we were to do something. That means we could go to something like this.

```rust
pub enum ConflictType {
  Spacing {
    train_ids: Vec<i64>,
    workschedule_ids: Vec<i64>,
  },
  Routing {
    train_ids: Vec<i64>,
  },
}
```
but that would mean a more important refactoring (pretty large and not in the scope of this PR).

On top of that, it's not entirely clear that OpenAPI would support this data structure, and even more, not even sure that codegen from `utoipa` to OpenAPI, then from OpenAPI to Typescript would work well. Also, this would probably means to manually implement the deserializer for converting what `core` is sending into this new Rust structure (and implementing manually the deserializer is no fun, it adds complexity).

Having said all that, we took the pragmatic solution to just move the `work_schedule_ids` next to `train_ids` in `Conflict`. That means that the data structure may contain corrupted data (i.e. it's possible to encode a `ConflictType::Routing` with some `work_schedule_ids`).

woshilapin (approved):
Thanks for the evolutions.

flomonster (approved):
LGTM (editoast part)

PR #8790, opened 2024-09-06
front: use <img loading="lazy"> instead of react-lazy-load-image-component

emersion:
There is a standard way to defer image loading until they become visible. Use that instead of a React plugin.

More details: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#loading


theocrsb (approved):
LGTM

emersion:
Gentle ping @Math-R

anisometropie (approved):
Good stuff !

Math-R (approved):
lgtm :)

PR #8794, opened 2024-09-06
editoast: work schedule projection endpoint

### LINKED ISSUE #8642: Implement editoast endpoint for work schedules projection
# Decription

```yaml
- type: CATENARY
  start_date_time: "2023-12-21T08:51:00+00:00"
  end_date_time: "2023-12-21T10:30:00+00:00"
  # MUST NOT be empty
  path_position_ranges:
    # Ranges in mm
    - [20_000, 100_000]
    - [150_000, 300_000]
- type: TRACK
  start_date_time: "2023-12-21T08:51:00+00:00"
  end_date_time: "2023-12-21T10:30:00+00:00"
  path_position_ranges:
    - [20_000, 100_000]
    - [150_000, 300_000]
```

### Endpoint

Create a new [endpoint](https://github.com/osrd-project/osrd-confidential/issues/531#issuecomment-2258547362)

anisometropie:
closes [#8642](https://github.com/OpenRailAssociation/osrd/issues/8642)

### DIFF
--- a/editoast/openapi.yaml
+++ b/editoast/openapi.yaml
@@ -2748,6 +2748,72 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/WorkScheduleCreateResponse'
+  /work_schedules/project_path:
+    post:
+      tags:
+      - work_schedules
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
+              required:
+              - work_schedule_group_id
+              - path_track_ranges
+              properties:
+                path_track_ranges:
+                  type: array
+                  items:
+                    $ref: '#/components/schemas/TrackRange'
+                work_schedule_group_id:
+                  type: integer
+                  format: int64
+        required: true
+      responses:
+        '201':
+          description: Returns a list of work schedules whose track ranges intersect the given path
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: object
+                  description: Represents the projection of a work schedule on a path.
+                  required:
+                  - type
+                  - start_date_time
+                  - end_date_time
+                  - path_position_ranges
+                  properties:
+                    end_date_time:
+                      type: string
+                      format: date-time
+                      description: The date and time when the work schedule ends.
+                    path_position_ranges:
+                      type: array
+                      items:
+                        type: array
+                        items:
+                          allOf:
+                          - type: integer
+                            format: int64
+                            minimum: 0
+                          - type: integer
+                            format: int64
+                            minimum: 0
+                      description: |-
+                        a list of intervals `(a, b)` that represent the projections of the work schedule track ranges:
+                        - `a` is the distance from the beginning of the path to the beginning of the track range
+                        - `b` is the distance from the beginning of the path to the end of the track range
+                    start_date_time:
+                      type: string
+                      format: date-time
+                      description: The date and time when the work schedule takes effect.
+                    type:
+                      type: string
+                      enum:
+                      - CATENARY
+                      - TRACK
 components:
   schemas:
     AddOperation:
--- a/editoast/src/models/fixtures.rs
+++ b/editoast/src/models/fixtures.rs
@@ -13,13 +13,14 @@ use editoast_schemas::primitives::OSRDObject;
 use editoast_schemas::rolling_stock::RollingStock;
 use editoast_schemas::train_schedule::TrainScheduleBase;
 use postgis_diesel::types::LineString;
+use serde_json::Value;
 
 use crate::infra_cache::operation::create::apply_create_operation;
 use crate::models::prelude::*;
 use crate::models::rolling_stock_livery::RollingStockLiveryModel;
 use crate::models::timetable::Timetable;
 use crate::models::train_schedule::TrainSchedule;
-use crate::models::work_schedules::WorkScheduleGroup;
+use crate::models::work_schedules::{WorkSchedule, WorkScheduleGroup};
 use crate::models::Document;
 use crate::models::Infra;
 use crate::models::Project;
@@ -307,4 +308,20 @@ pub async fn create_work_schedule_group(conn: &mut DbConnection) -> WorkSchedule
         .create(conn)
         .await
         .expect("Failed to create empty work schedule group")
+}
+
+pub async fn create_work_schedules_fixture_set(
+    conn: &mut DbConnection,
+    work_schedules: Vec<Changeset<WorkSchedule>>,
+) -> (WorkScheduleGroup, Vec<WorkSchedule>) {
+    let work_schedule_group = create_work_schedule_group(conn).await;
+    let work_schedules_changesets = work_schedules
+        .into_iter()
+        .map(|ws| ws.work_schedule_group_id(work_schedule_group.id))
+        .collect::<Vec<_>>();
+    let work_schedules = WorkSchedule::create_batch(conn, work_schedules_changesets)
+        .await
+        .expect("Failed to create work test schedules");
+
+    (work_schedule_group, work_schedules)
 }
--- a/editoast/src/views/work_schedules.rs
+++ b/editoast/src/views/work_schedules.rs
@@ -6,6 +6,7 @@ use chrono::Utc;
 use derivative::Derivative;
 use editoast_authz::BuiltinRole;
 use editoast_derive::EditoastError;
+use editoast_models::DbConnectionPoolV2;
 use serde::de::Error as SerdeError;
 use serde::Deserialize;
 use serde::Serialize;
@@ -13,19 +14,24 @@ use std::result::Result as StdResult;
 use thiserror::Error;
 use utoipa::ToSchema;
 
+use crate::core::pathfinding::TrackRange as CoreTrackRange;
 use crate::error::InternalError;
 use crate::error::Result;
 use crate::models::prelude::*;
 use crate::models::work_schedules::WorkSchedule;
 use crate::models::work_schedules::WorkScheduleGroup;
 use crate::models::work_schedules::WorkScheduleType;
+use crate::views::path::projection::PathProjection;
 use crate::views::AuthorizationError;
 use crate::views::AuthorizerExt;
 use crate::AppState;
-use editoast_schemas::infra::TrackRange;
+use editoast_schemas::infra::{Direction, TrackRange};
 
 crate::routes! {
-    "/work_schedules" => create,
+    "/work_schedules" => {
+        create,
+        "/project_path" => project_path,
+    },
 }
 
 editoast_common::schemas! {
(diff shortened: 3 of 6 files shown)


editoast/src/views/work_schedules.rs
```diff
@@ -13,25 +13,34 @@ use std::result::Result as StdResult;
...
 use crate::views::AuthorizerExt;
 use crate::AppState;
-use editoast_schemas::infra::TrackRange;
+use editoast_schemas::infra::{Direction, TrackRange};
 
 crate::routes! {
-    "/work_schedules" => create,
+    "/work_schedules" => {
+        create,
+        "/project" => project,
```
woshilapin: The existing endpoint for `train_schedule` is named `project_path`. It would be a good idea to keep the same naming. On top of that, we also have a project object in our model (in studies) so `project` alone can be misleading.

```suggestion
        "/project_path" => project,
```

woshilapin (changes requested):
Nice work. Just a small comment.

editoast/src/views/work_schedules.rs
```diff
@@ -175,15 +184,107 @@ async fn create(
...
+struct WorkScheduleProjectForm {
+    work_schedule_group_id: i64,
+    #[schema(value_type = Vec<TrackRange>)]
+    path_track_ranges: Vec<CoreTrackRange>,
+}
+
+#[derive(Serialize, Deserialize, ToSchema, PartialEq, Debug)]
+struct WorkScheduleProjectResponse {
+    projections: Vec<WorkScheduleProjection>,
+}
```
flomonster: This struct is not required. We can respond directly a `Vec<WorkScheduleProjection>`

```suggestion
```

editoast/src/views/work_schedules.rs
```diff
@@ -175,15 +184,107 @@ async fn create(
...
+
+#[derive(Serialize, Deserialize, ToSchema, PartialEq, Debug)]
+struct WorkScheduleProjection {
+    #[serde(rename = "type")]
+    pub work_schedule_type: WorkScheduleType,
+    pub start_date_time: NaiveDateTime,
+    pub end_date_time: NaiveDateTime,
+    pub path_position_ranges: Vec<(u64, u64)>,
+}
```
flomonster: Missing documentation. It is strongly recommended to document structures derived from `ToSchema`. This will fill in the openapi doc, and the TS frontend doc too.

editoast/src/views/work_schedules.rs
```diff
@@ -175,15 +184,107 @@ async fn create(
...
+    let db_pool = app_state.db_pool_v2.clone();
+    let conn = &mut db_pool.get().await?;
+    let settings: SelectionSettings<WorkSchedule> = SelectionSettings::new()
+        .filter(move || WorkSchedule::WORK_SCHEDULE_GROUP_ID.eq(work_schedule_group_id));
+    let work_schedules = WorkSchedule::list(conn, settings).await?;
+
+    let projections = work_schedules
+        .into_iter()
+        .filter_map(|ws| {
+            let ws_track_ranges: Vec<CoreTrackRange> = ws
```
flomonster: Nitpick

```suggestion
            let ws_track_ranges: Vec<_> = ws
```

editoast/src/views/work_schedules.rs
```diff
@@ -175,15 +184,107 @@ async fn create(
...
+                    track_section: tr.track,
+                    begin: (tr.begin * 1000.0) as u64,
+                    end: (tr.end * 1000.0) as u64,
+                    direction: Direction::StartToStop,
+                })
+                .collect();
+
+            let path_projection = PathProjection::new(&ws_track_ranges);
+            // project this work_schedule on the path
+            let path_position_ranges = path_projection.get_intersections(&path_track_ranges);
```
flomonster: Actually, I have some doubts as to whether `get_intersections` is the function we need.


There's a slight difference between the train schedule projection and the work schedule projection:
- A train schedule path is linear, a work schedule track ranges are not
- A PathProjection required unique tranck sections, which might not be the case for work schedule track ranges

------------------

I think we could adapt the get_intersections function to make it work in the general case (work schedule). This could be done in a later PR.

flomonster (changes requested):

editoast/src/views/work_schedules.rs
anisometropie: Here we are projecting a work_schedule on a path, it’s the length of the path that’s defining the different intervals returned.
what do you think of `project_on_path` ?

editoast/src/views/work_schedules.rs
anisometropie: done

editoast/src/views/work_schedules.rs
anisometropie: done

editoast/src/views/work_schedules.rs
anisometropie: alright, I made a ticket https://github.com/OpenRailAssociation/osrd/issues/8864

editoast/src/views/work_schedules.rs
anisometropie: done

editoast/src/views/work_schedules.rs
woshilapin: That's not a part of the job I know the most. Can you explain the difference with `train_schedule/project_path`? I thought they were for similar purposes.

Note that in any case, I believe that anything that is not just `project` will be better.

anisometropie: both are very similar, you have a “main” path and you project track ranges on this main path.

`train_schedule/project_path` we project the path of one train on the path of another train (which is being displayed on the GET, y-axis is the length of the path)

here we are projecting a list of track ranges (closed for railroad work)  on a path being displayed on the GET

woshilapin: So both are projecting on the path of train. I'd say it'd be better to also keep consistency in the naming. So keep `project_path` for both. What do you think?

editoast/openapi.yaml
```diff
@@ -2748,6 +2748,25 @@ paths:
...
+      - work_schedules
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/WorkScheduleProjectForm'
+        required: true
+      responses:
+        '201':
+          description: Returns a list of work schedule whose track ranges intersect the given path
```
leovalais:
```suggestion
          description: Returns a list of work schedules whose track ranges intersect the given path
```

editoast/src/models/fixtures.rs
```diff
@@ -308,3 +310,27 @@ pub async fn create_work_schedule_group(conn: &mut DbConnection) -> WorkSchedule
...
+
+pub struct WorkSchedulesFixtureSet {
+    pub work_schedule_group: WorkScheduleGroup,
+    pub work_schedules: Vec<WorkSchedule>,
+}
+
+pub async fn create_work_schedules_fixture_set(
+    conn: &mut DbConnection,
+    work_schedules: Vec<WorkScheduleItemForm>,
+) -> WorkSchedulesFixtureSet {
```
leovalais: We used to need `XFixtureSet` because we had to wrap that into `TestFixture<T>`, which is gone now. You can just return a tuple.

```suggestion
pub async fn create_work_schedules_fixture_set(
    conn: &mut DbConnection,
    work_schedules: Vec<WorkScheduleItemForm>,
) -> (WorkScheduleGroup, Vec<WorkSchedule>) {
```

editoast/src/models/fixtures.rs
```diff
@@ -29,6 +30,7 @@ use crate::models::Study;
 use crate::models::Tags;
 use crate::views::rolling_stock::form::RollingStockForm;
 use crate::views::train_schedule::TrainScheduleForm;
+use crate::views::work_schedules::WorkScheduleItemForm;
```
leovalais: In prevision of the editoast split in crates, we don't want to use `crate::views` in `models` (it's the opposite we want). The other `use crate::views` shouldn't be there and will need to be removed at some point. Even conceptually, the `XyzForm`-like structs represent views exchange format and don't have anything to do in the `models` module.

Just use a `Changeset<WorkSchedule>` instead.

editoast/src/views/work_schedules.rs
```diff
@@ -56,7 +64,7 @@ pub fn map_diesel_error(e: InternalError, name: impl AsRef<str>) -> InternalErro
 }
 
 #[derive(Serialize, Derivative, ToSchema)]
-struct WorkScheduleItemForm {
+pub struct WorkScheduleItemForm {
```
```suggestion
struct WorkScheduleItemForm {
```

editoast/src/views/work_schedules.rs
```diff
@@ -13,25 +13,33 @@ use std::result::Result as StdResult;
...
+    },
 }
 
 editoast_common::schemas! {
     WorkScheduleCreateForm,
     WorkScheduleCreateResponse,
+    WorkScheduleProjectForm,
+    WorkScheduleProjection,
     WorkScheduleItemForm,
+    WorkScheduleType,
```
```suggestion
```

You can `inline` this one as it's a small enum with a single ref. (Actually you can inline all the schemas you created since they're only used once.)

editoast/src/views/work_schedules.rs
```diff
@@ -175,15 +183,107 @@ async fn create(
...
+    responses(
+        (
+            status = 201,
+            body = Vec<WorkScheduleProjection>,
+            description = "Returns a list of work schedule whose track ranges intersect the given path"
+        ),
+    )
+)]
+async fn project(
+    State(app_state): State<AppState>,
```
```suggestion
    State(db_pool): State<DbConnectionPoolV2>,
```

editoast/src/views/work_schedules.rs
```diff
@@ -175,15 +183,107 @@ async fn create(
...
+    // get all work_schedule of the group
+    let db_pool = app_state.db_pool_v2.clone();
+    let conn = &mut db_pool.get().await?;
+    let settings: SelectionSettings<WorkSchedule> = SelectionSettings::new()
+        .filter(move || WorkSchedule::WORK_SCHEDULE_GROUP_ID.eq(work_schedule_group_id));
+    let work_schedules = WorkSchedule::list(conn, settings).await?;
+
+    let projections = work_schedules
+        .into_iter()
+        .filter_map(|ws| {
```
leovalais: This `filter_map` is big, I'd rather have it split into a 
```rust
.map({
  ...;
  path_projection.get_intersections(&path_track_ranges)
})
.filter_map(...)
```

leovalais (changes requested):
Great job! A few comments, mostly cosmetic. Thanks for the tests!

editoast/src/views/work_schedules.rs
flomonster: It does not seem to be done. The `WorkScheduleProjectResponse` is still there.

editoast/src/views/work_schedules.rs
anisometropie: done

editoast/src/views/work_schedules.rs
anisometropie: done

editoast/src/views/work_schedules.rs
anisometropie: done

editoast/src/views/work_schedules.rs
anisometropie: done

editoast/src/views/work_schedules.rs
anisometropie: done

editoast/src/models/fixtures.rs
anisometropie: alright, done

editoast/src/models/fixtures.rs
anisometropie: done !

editoast/openapi.yaml
anisometropie: done

editoast/src/views/work_schedules.rs
anisometropie: alright

anisometropie: done

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -811,6 +811,17 @@ const injectedRtkApi = api
         }),
         invalidatesTags: ['work_schedules'],
       }),
+      postWorkSchedulesProjectPath: build.mutation<
```
SharglutDev: If I understand correctly, it's a post request because of the body but we will use it like a "get".
If that's the case, we should add a special rule for this endpoint like we did for others in the file `openapi-editoast-config.ts`.

anisometropie: good point, changed it

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -1479,6 +1490,19 @@ export type PostWorkSchedulesApiResponse =
 export type PostWorkSchedulesApiArg = {
   workScheduleCreateForm: WorkScheduleCreateForm;
 };
+export type PostWorkSchedulesProjectPathApiResponse =
+  /** status 201 Returns a list of work schedules whose track ranges intersect the given path */ {
+    end_date_time: string;
+    path_position_ranges: (number & number)[][];
```
SharglutDev, line 1500: This syntax seems strange, shouldn't this be `[number, number][]`?

anisometropie: it should, we can’t really control what’s generated

woshilapin (approved):
Thanks for the follow-up 👍

front/src/common/api/generatedEditoastApi.ts
emersion: Sounds like a bug in the Rust OpenAPI generator. It should use [`prefixItems`](https://json-schema.org/draft/2020-12/json-schema-core#section-10.3.1.1) instead of two identical items in `allOf`.

A workaround would be to use a struct instead of a tuple in the Rust types. (Maybe that's better anyways since it'd be more explicit what the two values are.)

I'd recommend searching/opening an issue on the Rust OpenAPI generator anyways.

emersion: Seems to be https://github.com/juhaku/utoipa/issues/901.

editoast/src/views/work_schedules.rs
```diff
@@ -175,15 +181,109 @@ async fn create(
...
+/// relatively to the beginning of the path.
+#[derive(Serialize, Deserialize, ToSchema, PartialEq, Debug)]
+struct WorkScheduleProjection {
+    #[serde(rename = "type")]
+    #[schema(inline)]
+    pub work_schedule_type: WorkScheduleType,
+    pub start_date_time: NaiveDateTime,
+    pub end_date_time: NaiveDateTime,
+    pub path_position_ranges: Vec<(u64, u64)>,
+}
```
flomonster: Rather than documenting each field in the structure's documentation. The fields should be documented directly. 
You can run `cargo doc --open` to see how it looks.

```suggestion
/// A work schedule projected on a path
struct WorkScheduleProjection {
    /// The work schedule type
    #[serde(rename = "type")]
    #[schema(inline)]
    pub work_schedule_type: WorkScheduleType,
    /// The date and time when the work schedule takes effect.
    pub start_date_time: NaiveDateTime,
    /// The date and time at which the work schedule begins to apply
    pub end_date_time: NaiveDateTime,
    /// The date and time at which the work schedule ends.
    pub path_position_ranges: Vec<(u64, u64)>,
}
```

flomonster (approved):
LGTM. Thanks for this PR.

front/src/common/api/generatedEditoastApi.ts
anisometropie: thanks for digging up the dirt !

woshilapin: This might help? https://github.com/OpenRailAssociation/osrd/pull/8897

leovalais (approved):
LGTM, thanks!

editoast/src/views/work_schedules.rs
anisometropie: done

front/src/common/api/generatedEditoastApi.ts
anisometropie: Resolving to close here, better to fix this in next PR

SharglutDev (approved):
Lgtm

front/src/common/api/generatedEditoastApi.ts
woshilapin: For information, the other PR got merged.

PR #8801, opened 2024-09-09
editoast: add `.cargo` in `.gitignore`

hamz2a:


woshilapin (approved):
I'm not entirely sure about that PR since we might enforce project configuration through `.carg/config.toml` at some point in the future. Is there some use case you have that you can't handle differently:
1. you can use `~/.cargo/config.toml` (a global configuration on your machine)
2. using [`direnv`](https://github.com/direnv/direnv) combined with [`cargo` environment variables](https://doc.rust-lang.org/cargo/reference/config.html#environment-variables)

In any case, I'm not planning on blocking this PR since it will be easy to revert or do differently the day we need something different. Just wanted to point out alternative solutions for whatever use case you might have. Feel free to merge or discard the PR, whatever suits you.

hamz2a:
Thanks for the review. I'll close the PR and use the global `~/.cargo/config.toml` option instead.

PR #8803, opened 2024-09-09
front: add eslintcache, envrc and direnv to gitignore

RomainValls:
add eslintcache to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,7 @@
 *.backup
 __pycache__
 .vscode
+
+# Direnv
+.envrc
+.direnv
--- a/front/.gitignore
+++ b/front/.gitignore
@@ -34,6 +34,9 @@ yarn-error.log*
 /playwright-report/
 /playwright/.cache/
 
+# eslint cache
+.eslintcache
+
 # .env is built from .env.local and .env.local.defaults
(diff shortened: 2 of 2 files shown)


emersion (approved):

theocrsb (approved):

emersion:
I'm not familiar with envrc and direnv. If it's used by just a single person, maybe this approach should be used instead? https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

If it's used by more than one person, and if it's not frontend-specific, should it be in the toplevel gitignore?

anisometropie (approved):

Yohh:
> I'm not familiar with envrc and direnv. If it's used by just a single person, maybe this approach should be used instead? https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/
> 
> If it's used by more than one person, and if it's not frontend-specific, should it be in the toplevel gitignore?

we are many to use it, in both frontend and backend

SharglutDev (approved):

emersion:
> we are many to use it, in both frontend and backend

Then probably the rule should be moved to the root gitignore file?

anisometropie (approved):

Issue #8757, opened 2024-09-04
the label of the select are blue instead of black


kmer2016:
The label color of the Select component is blue in the STDCM implementation instead of black.

Mockup
![image](https://github.com/user-attachments/assets/d8a14737-e6c1-4371-b2fc-aa47e54755d8)

Implementation
![image](https://github.com/user-attachments/assets/fef48865-f0f3-4746-b43d-b5f62e40b1e8)

Issue #8731, opened 2024-09-03
Fix the arrow icon in the Select dropdown for changing CH code


kmer2016:
The icon of CH code Select is different between the mockup and the implementation of STDCM interface.

The mockup
![image](https://github.com/user-attachments/assets/8c45db3d-d4fd-4326-8aff-d807da53a565)

The implementation
![image](https://github.com/user-attachments/assets/34ce3aaf-c554-4483-88b9-9d9de965a43e)

Issue #8730, opened 2024-09-03
Set a fixed width for the Consist and Itinerary sections as defined in the mockup


kmer2016:
The consist and Itinerary sections have fixed width in [mockup](https://www.sketch.com/s/15e0909c-f17d-4200-a770-b9fc1154fa7e/p/CA8C9E9A-B795-4C87-9CF5-508DD0B3634A/canvas). while the map takes up the remaining space.

PR #8804, opened 2024-09-09
Adjustment of STDCM interface (Part 1)

kmer2016:
There are some differences between [mockup](https://www.sketch.com/s/15e0909c-f17d-4200-a770-b9fc1154fa7e/p/CA8C9E9A-B795-4C87-9CF5-508DD0B3634A/canvas) and the STDCM interface implementation. This PR fix three of them
- [x] #8730 (first commit)
- [x] #8731 (second commit)
- [x] #8757 (second commit)


front/src/styles/scss/_uiCoreIntegration.scss
```diff
@@ -68,3 +68,61 @@ img {
     }
   }
 }
+
+// __________________ Select  __________________
```
SharglutDev, line 72: Maybe we can add a todo here to remove this when the bootstrap/tailwind conflict will be resolved ?

front/src/styles/scss/_uiCoreIntegration.scss
SharglutDev, line 1: The position of the items list when opening the select seems to be wrong for me :/ (both on chrome and firefox)

mockup : 
Capture d’écran 2024-09-10 à 09 49 25

chrome :
Capture d’écran 2024-09-10 à 09 47 52

firefox : 
Capture d’écran 2024-09-10 à 09 44 41

SharglutDev (commented):
Thank you for the PR ! Consist/itinerary and blue>black label fixes are good for me but still have some issues with the select one.

front/src/styles/scss/_uiCoreIntegration.scss
SharglutDev, line 1: Oh and for the arrow fix, when opening the select and closing it by clicking outside or on an item, the arrow seems to be the wrong one : 
Capture d’écran 2024-09-10 à 09 53 19

SharglutDev: Ok turns out it comes from a macOS internal config. This behavior is normal on a linux/windows device.

Caracol3:
image

image


If the width should be fixed, are we sure to want to keep the icon ? When french langage is on, it prevents to put 'code de composition on one line'. Or maybe we should put it under the icon ?

If we keep it like that, i think it would be good to at least add a padding between the icon and the text. For me, the two are too close.

EDIT: in the sketch its called 'code compo.' , this solves the problem.

SharglutDev (approved):
Lgtm and tested

Caracol3:
In consist component, like 'tracion engine', 'speed limits' and 'composition code' should be bold.

kmer2016:
Thank you for your review, @Caracol3. I will address your feedback in a separate [PR](https://github.com/OpenRailAssociation/osrd/issues/8820) where I will directly implement the Code Compo. component.

Caracol3 (approved):
lgtm

PR #8805, opened 2024-09-09
editoast: remove duplicate query results using `DISTINCT ON (object_id)`

hamz2a:
The query in the `Infra::get_objects()` function was returning duplicate rows due to multiple matches in the `LEFT JOIN` when joining `infra_layer_{object_type}` and `infra_object_{object_type}`. This caused the same object to be returned multiple times.


woshilapin (approved):
Nice!

PR #8807, opened 2024-09-09
front: custom help modal

### LINKED ISSUE #8710: Add support for custom "Help" button inside the application menu
We should create a solution that leverages two elements:
1. We should modify the https://github.com/OpenRailAssociation/osrd-chart in order to support injecting a plain Markdown file within the static folder served by the gateway
2. We should check for this file on application boot and if it exists, add a Help menu item to the application menu on the top right corner. It should open a popup that will displays the content of the markdown, rendered.

ghost:
This PR adds a menu item which displays a modal with custom information. Closes #8710 

In development, to test it, add the following file at `front/public/help.md`:

```
**File with markdown rendering**
```

In production, the helm chart will be modified to serve the static file from within the static folder of the gateway using a configmap and a volume mount.

For other deployments that would not use our helm-chart (docker compose etc...) the markdown file can be mounted on the required path easily.

You can find the corresponding PR on osrd-chart [here](https://github.com/OpenRailAssociation/osrd-chart/pull/14).

### DIFF
--- a/front/public/locales/en/home/navbar.json
+++ b/front/public/locales/en/home/navbar.json
@@ -2,6 +2,7 @@
   "about": "About",
   "attributions": "Attributions",
   "disconnect": "Sign out",
+  "help": "Help",
   "informations": {
     "application": "Application",
     "releaseInformations": "Libraries & licenses",
--- a/front/src/common/HelpModal/HelpModal.tsx
+++ b/front/src/common/HelpModal/HelpModal.tsx
@@ -1,0 +1,44 @@
+import { useState, useEffect } from 'react';
+
+import { useTranslation } from 'react-i18next';
+import ReactMarkdown from 'react-markdown';
+import remarkGfm from 'remark-gfm';
+
+import ModalBodySNCF from 'common/BootstrapSNCF/ModalSNCF/ModalBodySNCF';
+import ModalHeaderSNCF from 'common/BootstrapSNCF/ModalSNCF/ModalHeaderSNCF';
+
+const DEFAULT_TEXT = '# contact@osrd.fr';
+
+const HelpModal = () => {
+  const { t } = useTranslation('home/navbar');
+  const [helpModalText, setHelpModalText] = useState<string>('');
+
+  useEffect(() => {
+    const checkHelpMarkdown = async () => {
+      try {
+        const response = await fetch('/help.md');
+
+        if (!response.ok || response.headers.get('Content-Type') !== 'text/markdown') {
+          setHelpModalText(DEFAULT_TEXT);
+          return;
+        }
+
+        setHelpModalText(await response.text());
+      } catch {
+        setHelpModalText(DEFAULT_TEXT);
+      }
+    };
+    checkHelpMarkdown();
+  }, []);
+
+  return (
+    <div className="informations-modal">
+      <ModalHeaderSNCF withCloseButton>{t('help')}</ModalHeaderSNCF>
+      <ModalBodySNCF>
+        <ReactMarkdown remarkPlugins={[remarkGfm]}>{helpModalText}</ReactMarkdown>
+      </ModalBodySNCF>
+    </div>
+  );
+};
+
+export default HelpModal;
--- a/front/public/locales/fr/home/navbar.json
+++ b/front/public/locales/fr/home/navbar.json
@@ -2,6 +2,7 @@
   "about": "À propos",
   "attributions": "Attributions",
   "disconnect": "Déconnexion",
+  "help": "Aide",
   "informations": {
     "application": "Application",
     "releaseInformations": "Librairies & licences",
--- a/front/src/common/BootstrapSNCF/NavBarSNCF.tsx
+++ b/front/src/common/BootstrapSNCF/NavBarSNCF.tsx
@@ -1,6 +1,6 @@
 import { type ReactElement } from 'react';
 
-import { Gear, Info, ShieldCheck, SignOut } from '@osrd-project/ui-icons';
+import { Gear, Info, Report, ShieldCheck, SignOut } from '@osrd-project/ui-icons';
 import getUnicodeFlagIcon from 'country-flag-icons/unicode';
 import i18n from 'i18next';
 import { useTranslation } from 'react-i18next';
@@ -8,6 +8,7 @@ import { useSelector } from 'react-redux';
 import { Link } from 'react-router-dom';
 
 import ChangeLanguageModal from 'common/ChangeLanguageModal';
+import HelpModal from 'common/HelpModal/HelpModal';
 import ReleaseInformations from 'common/ReleaseInformations/ReleaseInformations';
 import UserSettings from 'common/UserSettings';
 import { getUserSafeWord } from 'reducers/user/userSelectors';
@@ -77,6 +78,17 @@ const LegacyNavBarSNCF = ({ appName, logo = getLogo() }: Props) => {
                   <Info />
                 </span>
                 {t('about')}
+              </button>,
+              <button
+                type="button"
+                className="btn-link text-reset"
+                onClick={() => openModal(<HelpModal />, 'lg')}
+                key="help"
+              >
+                <span className="mr-2">
+                  <Report />
+                </span>
+                {t('help')}
               </button>,
               <button
                 type="button"


front/src/common/HelpModal/HelpModal.tsx
```diff
@@ -0,0 +1,51 @@
...
+import ModalHeaderSNCF from 'common/BootstrapSNCF/ModalSNCF/ModalHeaderSNCF';
+
+function HelpModal() {
+  const { t } = useTranslation('home/navbar');
+  const [helpModalText, setHelpModalText] = useState<string>('');
+
+  useEffect(() => {
+    const checkHelpMarkdown = async () => {
+      const expectedFileFirstLine = '<!-- OSRD Help Markdown File -->';
+      const defaultText = '# contact@osrd.fr';
```
SarahBellaha:
```suggestion
```

front/src/common/HelpModal/HelpModal.tsx
```diff
@@ -0,0 +1,51 @@
+import { useState, useEffect } from 'react';
+
+import { useTranslation } from 'react-i18next';
+import ReactMarkdown from 'react-markdown';
+import remarkGfm from 'remark-gfm';
+
+import ModalBodySNCF from 'common/BootstrapSNCF/ModalSNCF/ModalBodySNCF';
+import ModalHeaderSNCF from 'common/BootstrapSNCF/ModalSNCF/ModalHeaderSNCF';
+
```
SarahBellaha, line 9:
```suggestion

      const EXPECTED_FILE_FIRST_LINE = '<!-- OSRD Help Markdown File -->';
      const DEFAULT_TEXT = '# contact@osrd.fr';
```
It would be better to declare these constants at the top of the file and use them throughout the component.

SarahBellaha (approved):
Just a small suggestion, but LGTM :)

front/src/common/HelpModal/HelpModal.tsx
```diff
@@ -0,0 +1,51 @@
...
+        const response = await fetch('/help.md');
+
+        if (!response.ok) {
+          setHelpModalText(defaultText);
+          return;
+        }
+
+        const text = await response.text();
+        setHelpModalText(
+          text.startsWith(expectedFileFirstLine)
```
emersion: Why do we discard the file if it doesn't start with `expectedFileFirstLine`?

front/src/common/HelpModal/HelpModal.tsx
```diff
@@ -0,0 +1,51 @@
...
+            ? text.replace('<!-- OSRD Help Markdown File -->', '')
```
emersion: Is it necessary to strip the comment? HTML comments should get ignored in markdown no?

front/src/common/HelpModal/HelpModal.tsx
ghost: We always respond with the index file to every url that doesn't match a disk file in production (standard behavior for the path to be passed to the front router). I've detailed it in the PR text

front/src/common/HelpModal/HelpModal.tsx
ghost: I first thought, then I discovered our markdown renderer doesn't support this

front/src/common/HelpModal/HelpModal.tsx
emersion: Oh my bad! Makes sense for a SPA.

emersion: If we wanted to workaround this, we could either:

- Have a dedicated `/assets/` directory for static assets, the SPA index is never served there.
- Disallow paths ending with a file extension for SPA paths.
- Add `Accept: text/markdown` so that the server refuses to serve a non-Markdown file.

But these are larger changes.

front/src/common/HelpModal/HelpModal.tsx
emersion: Bummer, makes sense!

front/src/common/HelpModal/HelpModal.tsx
ghost: All these changes would require modifications of the gateway and they would need to create a service that would encapsulate the static_file service that is currently used.

I was not a real fan of modifying the gateway for this, it would make the PR way larger.

front/src/common/HelpModal/HelpModal.tsx
ghost: Done, thanks.

front/src/common/HelpModal/HelpModal.tsx
emersion: The last one could maybe not require any gateway changes -- or we could even check the `Content-Type` returned by the server instead of having a magic header. Anyways, the current approach is fine by me as well.

ghost: > The latter could maybe not require any gateway changes -- or we could even check the `Content-Type` returned by the server instead of having a magic header. Anyways, the current approach is fine by me as well.

That's true, I thought the Gateway might not implement this by default but it might be the case. I'm currently checking it.

ghost: image

Unfortunately, that's what I thought : we need to modify the gateway too if we want this behavior.

front/public/locales/en/home/navbar.json
SharglutDev, line 1: Could you insert the key with alphabetical order ? (same for fr file)

front/src/common/HelpModal/HelpModal.tsx
```diff
@@ -0,0 +1,52 @@
...
+const expectedFileFirstLine = '<!-- OSRD Help Markdown File -->';
+const defaultText = '# contact@osrd.fr';
+
+function HelpModal() {
+  const { t } = useTranslation('home/navbar');
+  const [helpModalText, setHelpModalText] = useState<string>('');
+
+  useEffect(() => {
+    const checkHelpMarkdown = async () => {
+
```
SharglutDev: There is an eslint warning here for an extra empty line that will break the CI. You can run `yarn lint` to detect these warnings and `yarn lint-fix` to fix small ones like this.

front/src/common/HelpModal/HelpModal.tsx
emersion: Right, but would my suggestion in the previous comment work? Could we replace the `text.startsWith(expectedFileFirstLine)` check with `response.headers.get('Content-Type') === 'text/markdown'`?

front/src/common/HelpModal/HelpModal.tsx
```diff
@@ -0,0 +1,52 @@
...
+import ReactMarkdown from 'react-markdown';
+import remarkGfm from 'remark-gfm';
+
+import ModalBodySNCF from 'common/BootstrapSNCF/ModalSNCF/ModalBodySNCF';
+import ModalHeaderSNCF from 'common/BootstrapSNCF/ModalSNCF/ModalHeaderSNCF';
+
+const expectedFileFirstLine = '<!-- OSRD Help Markdown File -->';
+const defaultText = '# contact@osrd.fr';
+
+function HelpModal() {
```
SharglutDev: We try to declare our components with arrow functions (we should probably add this in osrd website)

```suggestion
const HelpModal = () => {
```

front/src/common/HelpModal/HelpModal.tsx
ghost: Oh like that, sorry, I thought you meant server side check. I guess that would work yes. I'll modify and test it.

front/src/common/HelpModal/HelpModal.tsx
```diff
@@ -0,0 +1,52 @@
...
+      } catch {
+        setHelpModalText(defaultText);
+      }
+    };
+
+    checkHelpMarkdown();
+  }, []);
+
+  return (
+    <div className="informations-modal h-100">
```
SharglutDev: Not sure the `height: 100%` here will work as there is a`max-height: 75vh` in the `informations-modal` class

SharglutDev (commented):
Thank for your PR ! Left some comments (mostly front conventions).

front/src/common/HelpModal/HelpModal.tsx
```diff
@@ -0,0 +1,44 @@
+import { useState, useEffect } from 'react';
+
+import { useTranslation } from 'react-i18next';
+import ReactMarkdown from 'react-markdown';
+import remarkGfm from 'remark-gfm';
+
+import ModalBodySNCF from 'common/BootstrapSNCF/ModalSNCF/ModalBodySNCF';
+import ModalHeaderSNCF from 'common/BootstrapSNCF/ModalSNCF/ModalHeaderSNCF';
+
+const defaultText = '# contact@osrd.fr';
```
emersion: One last front convention: we use UPPER_SNAKE_CASE for global constants.

front/src/common/HelpModal/HelpModal.tsx
emersion: Looks good, thank you!

emersion (approved):
LGTM!

SharglutDev (approved):
Lgtm and tested :)

PR #8815, opened 2024-09-10
core: stdcm: implement variable stop durations

### LINKED ISSUE #7691: core: stdcm: variable stop times
* We currently use a value that describes by how much we can shift departure time at any given point. It would now describe how much we can shift the passage time, by *either* shifting the departure time *or* lengthening stops. 
* When we pass by a stop that can be lengthened, the value is increased (still accounting for possible conflicts later on)
* Post-processing: we allocate that extra time to the right stops / departure time.

We *may* have to explicitly keep track of the total running time for any given edge, instead of computing it. 

We should probably look into grouping some edge/node variables into new structs, with a small refactor for a better handling of timing data

Prerequisite:

- [x] #8587

eckter:
Fix #7691 

Change summary: 

1. Implement variable stop duration
2. Improve the concept of "visited" nodes to handle different criteria
    1. We had a binary concept of "visited range" and "visited with some extra allowance time"
    2. This concept was generalized and includes "visited with some extra stop duration"
3. Manage stops in the infra explorer with envelopes
    1. For the conflict detection, we had the stops directly in the incremental envelope
    2. We now explicitly keep the list and wrap the envelop on any access
4. Improve logging when postprocessing fail
    1. It's always very painful to figure out where the mismatch happens between exploration and postprocessing
    2. It's still painful but easier

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
@@ -33,9 +33,10 @@ data class STDCMNode(
 
     /**
      * Defines the estimated better path between 2 nodes, in the following priority:
-     * - lowest total run time
+     * - lowest total run time, excluding stops
      * - closest planned arrival time, taking the tolerance into account, using the current node's
-     *   planned timing data, then the last planned node's timing data
+     * - planned timing data, then the last planned node's timing data
+     * - total run time, including stops
      * - earliest departure time
      * - highest number of reached targets
      *
@@ -44,43 +45,44 @@ data class STDCMNode(
      * path. We then explore them in that order.
      */
     override fun compareTo(other: STDCMNode): Int {
-        val runTimeEstimation = timeData.timeSinceDeparture + remainingTimeEstimation
-        val otherRunTimeEstimation =
-            other.timeData.timeSinceDeparture + other.remainingTimeEstimation
-        val plannedRelativeTimeDiff =
-            getRelativeTimeDiff(
-                timeData.totalDepartureDelay,
-                timeData.maxDepartureDelayingWithoutConflict
-            )
-        val otherPlannedRelativeTimeDiff =
-            other.getRelativeTimeDiff(
-                other.timeData.totalDepartureDelay,
-                other.timeData.maxDepartureDelayingWithoutConflict
-            )
-        // Firstly, minimize the total run time: highest priority node takes the least time to
-        // complete the path
-        // TODO: with variable stop times, stop duration and running time should be separated
-        return if (!areTimesEqual(runTimeEstimation, otherRunTimeEstimation))
-            runTimeEstimation.compareTo(otherRunTimeEstimation)
-        // Minimise the difference with the planned arrival times
-        else if (
+        val runTimeEstimation = timeData.totalRunningTime + remainingTimeEstimation
+        val otherRunTimeEstimation = other.timeData.totalRunningTime + other.remainingTimeEstimation
+        // First, minimize the total run time:
+        // highest priority node takes the least time to complete the path
+        if (!areTimesEqual(runTimeEstimation, otherRunTimeEstimation))
+            return runTimeEstimation.compareTo(otherRunTimeEstimation)
+
+        val plannedRelativeTimeDiff = getRelativeTimeDiff(timeData)
+        val otherPlannedRelativeTimeDiff = other.getRelativeTimeDiff(other.timeData)
+
+        // If equal, minimise the difference with the planned arrival times
+        return if (
             plannedRelativeTimeDiff != null &&
                 otherPlannedRelativeTimeDiff != null &&
                 plannedRelativeTimeDiff != otherPlannedRelativeTimeDiff
         )
             (plannedRelativeTimeDiff).compareTo(otherPlannedRelativeTimeDiff)
-        // If not, minimise the difference with the planned arrival times at the last planned node
+
+        // If equal, minimise the difference with the planned arrival times at the last planned node
         else if (
             previousPlannedNodeRelativeTimeDiff != null &&
                 other.previousPlannedNodeRelativeTimeDiff != null &&
                 previousPlannedNodeRelativeTimeDiff != other.previousPlannedNodeRelativeTimeDiff
         )
             previousPlannedNodeRelativeTimeDiff.compareTo(other.previousPlannedNodeRelativeTimeDiff)
-        // If not, take the train which departs first, as it is the closest to the demanded
-        // departure time
+
+        // If equal, take the train which has the smallest time since its departure.
+        // Unlike the first check, this includes stop time.
+        else if (!areTimesEqual(timeData.timeSinceDeparture, other.timeData.timeSinceDeparture))
+            return timeData.timeSinceDeparture.compareTo(other.timeData.timeSinceDeparture)
+
+        // If equal, take the train which departs first
         else if (timeData.earliestReachableTime != other.timeData.earliestReachableTime)
             timeData.earliestReachableTime.compareTo(other.timeData.earliestReachableTime)
-        // In the end, prioritize the highest number of reached targets
+
+        // In the end, prioritize the highest number of reached targets.
+        // This doesn't define the priority between different paths,
+        // it just minimizes the chance of evaluating redundant nodes
         else other.waypointIndex - waypointIndex
     }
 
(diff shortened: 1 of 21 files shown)


core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/TimeData.kt
```diff
@@ -102,12 +108,60 @@ data class TimeData(
         maxDepartureDelayingWithoutConflict: Double,
     ): TimeData {
         assert(timeShift >= delayAddedToLastDeparture)
```
Erashin, line 119: We should probably add an `assert stopTimeData.isNotEmpty()`.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/TimeData.kt
```diff
@@ -102,12 +108,60 @@ data class TimeData(
...
+        var newStopData = stopTimeData
+        var newDepartureTime = departureTime
+        if (delayAddedToLastDeparture > 0) {
+            if (newStopData.isEmpty()) {
+                newDepartureTime += delayAddedToLastDeparture
+            } else {
+                val stopDataCopy = newStopData.toMutableList()
+                val oldStopData = stopDataCopy[stopDataCopy.size - 1]
```
Erashin: The point of this method is to shift the last stop of the list. I'm down to clarify that in the method description as well as in a comment here? (the current description could make you think the method shifts all of the time data). I woulda changed the var name to `lastStopData`, but `oldStopData` works pretty well with the subsequent `newStopData`, so idk.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/TimeData.kt
```diff
@@ -102,12 +108,60 @@ data class TimeData(
...
+
+    /**
+     * When we have moved further down the path, this method gives an updated estimation of the
+     * earliest reachable time. This accounts for any extra delay that may have been added to any
+     * departure time. "updatedTimeData" is the latest available TimeData instance.
+     */
+    fun getUpdatedEarliestReachableTime(updatedTimeData: TimeData): Double {
+        val addedDepartureDelay = updatedTimeData.departureTime - departureTime
+
+        // Ignore any stop that haven't been reached in the current instance
```
Erashin: `hasn't` or `Ignore all of the stops that haven't`.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
```diff
@@ -100,49 +102,71 @@ data class STDCMNode(
...
+     * further down the path. The tricky part is identifying how stop durations may be adjusted to
+     * locally change passage times without conflict.
+     */
+    private fun computeMaxAddedDelay(updatedTimeData: TimeData): Double {
+        var maxAddedDelay = Double.POSITIVE_INFINITY
+
+        // List of stops that haven't been reached on this node
+        var nextStopIndex = timeData.stopTimeData.size - 1
+        if (stopDuration == null) nextStopIndex++
+        val nextStops =
```
Erashin, line 148: Same here, if list is empty, this'll throw an exception. Maybe add a return 0.0 if list is empty?

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
```diff
@@ -133,32 +136,117 @@ class STDCMPostProcessing(private val graph: STDCMGraph) {
...
+        edges: List<STDCMEdge>,
+    ): TimeData {
+        // Make stop list mutable (locally)
+        val nodes = getNodes(edges)
+        val mutableStopData = nodes.last().timeData.stopTimeData.toMutableList()
+        var timeData = nodes.last().timeData.copy(stopTimeData = mutableStopData)
+
+        // Find the index of the first planned node, and matching stop index
+        val firstPlannedNodeIndex = nodes.indexOfFirst { it.plannedTimingData != null }
+        if (firstPlannedNodeIndex < 0) return timeData
```
Erashin, line 153: This should probably be an exception, as it means the request contained 0 planned timing data, but we should have caught it at the request's parsing, so I guess ignoring it is fine, but we should probably add an exception at the initial parsing?

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
```diff
@@ -133,32 +136,117 @@ class STDCMPostProcessing(private val graph: STDCMGraph) {
...
+        val node = nodes[firstPlannedNodeIndex]
+        val lastStopIndexBeforeNode =
```
Erashin, line 156: I thought there was a pb at first, but it is indeed the last stop index in the mutableStopData list. It's not 100% clear, I was confused at first, might need a comment?

Erashin (commented):
This is for the first commit :)

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/TimeData.kt
eckter: It can be empty, in which case it's the global departure time that's being delayed

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMNode.kt
eckter: It shouldn't throw, `subList` can take an empty list and return an empty list. In which case the method would return `min(POSITIVE_INFINITY, updatedTimeData.maxDepartureDelayingWithoutConflict)`

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
eckter: Timing data is required for endpoint requests, but there's still many unit tests that don't define it.

We could define timing data in all unit tests and add an error here, but I don't think it's worth the trouble. The function has a well-defined behavior that works without timing data.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/TimeData.kt
eckter: Fixed in https://github.com/OpenRailAssociation/osrd/pull/8815/commits/c171c03867dea76b67b482056837f74e67e74807 (extra comment in docstring)

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/TimeData.kt
eckter: Fixed in https://github.com/OpenRailAssociation/osrd/pull/8815/commits/c171c03867dea76b67b482056837f74e67e74807

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
eckter: Fixed in https://github.com/OpenRailAssociation/osrd/pull/8815/commits/c171c03867dea76b67b482056837f74e67e74807

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
Erashin: Ok, works for me :)

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/VisitedNodes.kt
```diff
@@ -28,86 +25,151 @@ import kotlin.Double.Companion.POSITIVE_INFINITY
...
  * The time check is where most complexity lies: we often pass by the same places at different
- * times, but in ways that have been fully covered by previously seen nodes. For each location, we
- * keep track of two things:
- * 1. When have we *visited* this place? This considers the ranges that have been covered by
- *    previously seen nodes, possibly by shifting the departure time
- * 2. When could we reach this place, including engineering margins? This isn't just a "visited"
- *    flag as it would increase the path cost, but it can avoid a lot of redundant computations
+ * times, but in ways that have been fully covered by previously seen nodes. We have a range map of
+ * "conditionally visited ranges". We may consider that a range is already visited if, for example,
+ * we need to add more than 42 seconds of stop duration to reach it.
```
Erashin, line 30: 42. Nice :D

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/VisitedNodes.kt
```diff
@@ -28,86 +25,151 @@ import kotlin.Double.Companion.POSITIVE_INFINITY
...
-        val baseNodeCost: Double,
-    )
+        val remainingTimeEstimation: Double = 0.0,
+    ) {
+        val nodeCost = timeData.totalRunningTime + remainingTimeEstimation
+    }
+
+    /** Any class that implements this interface may be added to the visited ranges. */
+    private sealed interface ConditionallyVisitedRange {
+        /** Is the given (sub)range is covered by `this`? */
```
Erashin: `Is the given (sub)range covered by`

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/VisitedNodes.kt
```diff
@@ -28,86 +25,151 @@ import kotlin.Double.Companion.POSITIVE_INFINITY
...
+        override fun mergeWith(other: ConditionallyVisitedRange?): ConditionallyVisitedRange {
+            return this
         }
     }
 
     /**
-     * For each location, the time ranges that have been visited. We know they can always be
-     * ignored.
+     * The given range is already visited by adding stop durations. The other candidates are visited
+     * if they add more stop time than this.
```
Erashin: The difference with the next class is that the other candidates are only visited if their cost is lower. Might be good to have it written in the main class description (i know you also wrote it a bit further down).

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/VisitedNodes.kt
```diff
@@ -28,86 +25,151 @@ import kotlin.Double.Companion.POSITIVE_INFINITY
...
+        val subMap = visitedRanges.subRangeMap(visitingRange)
+
+        // Check that the range is fully covered by the map
+        val uncovered = TreeRangeSet.create<Double>()
+        uncovered.add(visitingRange)
+        for (entry in subMap.asMapOfRanges()) uncovered.remove(entry.key)
+        if (!uncovered.isEmpty) return false
+
+        // Check that every range is considered "visited" with the given parameters
+        for (entry in subMap.asMapOfRanges()) {
```
Erashin, line 173: You could probably add this check in the previous loop in a boolean, and do:
`if (!uncovered.isEmpty || !isVisited) return false`
If it muddles the code too much, keeping it as is is fine as well.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/VisitedNodes.kt
```diff
@@ -116,26 +178,58 @@ data class VisitedNodes(val minDelay: Double) {
...
+        val endRangeExtraTravelTime = endRangeExtraStopTime + parameters.maxMarginDuration
+
+        fun putRange(start: Double, end: Double, value: ConditionallyVisitedRange) {
+            if (start < end) {
+                val range = Range.closedOpen(start, end)
+                visitedRanges.merge(range, value) { a, b -> a.mergeWith(b) }
+            }
         }
+
+        putRange(
```
Erashin, line 214: These 3 method calls might need a comment to clarify that we're marking the 3 types of classes as visited with their corresponding parameters.

core/src/test/kotlin/fr/sncf/osrd/stdcm/VisitedNodesTests.kt
```diff
@@ -79,17 +128,120 @@ class VisitedNodesTests {
...
         // The duration here is longer than the previous test: it's visited from t=120 to t=125
         // (above the cost function), then unvisited from t=125. Some parts are unvisited => false
         // is expected
         visitedNodes.markAsVisited(params1)
         assertFalse { visitedNodes.isVisited(params2) }
     }
+
+    @ParameterizedTest
+    @ValueSource(booleans = [true, false])
+    fun testStopTimes(lowerStopDuration: Boolean) {
```
Erashin, line 194: I feel like we're missing some test cases ? I don't think we're testing out the visited with margins case ?

Erashin (commented):
Commit numero dos !

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/TimeData.kt
```diff
@@ -81,11 +84,10 @@ data class TimeData(
                 )
             )
             newStopData = stopDataCopy
-            maxDepartureDelayingWithoutConflict = Double.POSITIVE_INFINITY
+            maxDepartureDelayingWithoutConflict = maxAdditionalStopTime!!
```
Erashin, line 90: `maxDepartureDelayingWithoutConflict = maxAdditionalStopTime ?: Double.POSITIVE_INFINITY`?

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
```diff
@@ -62,6 +70,68 @@ data class InfraExplorerWithEnvelopeImpl(
         return this
     }
 
+    override fun withNewEnvelope(envelope: Envelope): InfraExplorerWithEnvelope {
+        return copy(
+            envelopes =
+                appendOnlyLinkedListOf(
+                    LocatedEnvelope(envelope, 0.0, 0.0),
+                ),
+            spacingRequirementAutomaton =
```
Erashin, line 79: Worth to create a `spacingRequirementAutomaton.cloneOrSomething(IncrementalRequirementEnvelopeAdapter, IncrementalPath)` or something of the sort?

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
```diff
@@ -62,6 +70,68 @@ data class InfraExplorerWithEnvelopeImpl(
...
+                TrainStop(
+                    position,
+                    stopDuration,
+                    true,
+                )
+            )
+            envelopeCache = null
+            spacingRequirementsCache = null
+        } else {
+            assert(stops.isEmpty() || stops.last().duration == stopDuration)
```
Erashin: Shouldn't we just throw if `stopDuration < stops.last.duration`, and do `steps.last().duration = stopDuration` otherwise?

Erashin (commented):
Commit number 3!
More generally on this commit, as we discussed, might be worth it to investigate if we can make it so that the stop logic doesn't overstep into the infra explorer class. Would probably make things easier to understand (TBD).

EDIT: Nope. Riperoni pepperoni.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/VisitedNodes.kt
eckter: Fixed in 0c57be6390f29f51b60b1078138c60211d4dfd1f

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/VisitedNodes.kt
eckter: Fixed in 0c57be6390f29f51b60b1078138c60211d4dfd1f (new phrasing)

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/VisitedNodes.kt
eckter: Fixed in 0c57be6390f29f51b60b1078138c60211d4dfd1f

There's also an extra corner case identified with the new unit test handled here

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/VisitedNodes.kt
eckter: Fixed in 0c57be6390f29f51b60b1078138c60211d4dfd1f

core/src/test/kotlin/fr/sncf/osrd/stdcm/VisitedNodesTests.kt
eckter: Fixed in 0c57be6390f29f51b60b1078138c60211d4dfd1f

It did find some weird stuff, not related to the margin specifically but I hadn't try with 0-length visiting ranges

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/TimeData.kt
eckter: It should always be non-null, adding a stop should define both stop duration and max additional stop time

But I can add a comment

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
eckter: We *should* reset it entirely in this case, there's nothing to clone.

The name of the method is misleading though, I'll rename it.

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
eckter: It should be identical, we tolerate duplicates but only if there's no change. But the first part of the assert is useless, it's covered in the `if`

core/src/test/kotlin/fr/sncf/osrd/stdcm/VisitedNodesTests.kt
Erashin: Nice catch :)

Erashin (approved):
YYYAAAYYY

PR #8818, opened 2024-09-10
ci: add dco check

EthanPERRUZZA:


emersion:
Could we maybe merge this new check with the existing `check_commits` job? It's pretty similar and it would avoid spinning up two containers.

emersion (approved):
LGTM :)

younesschrifi (approved):
LGTM ! 😁

.github/workflows/build.yml
```diff
@@ -285,6 +285,10 @@ jobs:
 
         commit_titles | TERM=xterm-color .github/scripts/check-commit-titles.sh
 
+    - name: Check DCO
+      uses: christophebedard/dco-check@0.5.0
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
Khoyo: We should exclude dependabot from the check (eg using `--exclude-emails '49699333+dependabot[bot]@users.noreply.github.com'`)

Khoyo (changes requested):

.github/workflows/build.yml
EthanPERRUZZA: done ;)

Khoyo (approved):

PR #8819, opened 2024-09-10
front: ensure NGE's Node.id is always a number

emersion:
We were abusing this field to store the OSRD OP ID. However, NGE's type expects this value to always be an integer. Using anything else is quite fragile and might result in fireworks (e.g. when NGE tries to perform math on string values to auto-increment the ID).

Stop storing the OP ID, instead just use a boring auto-incremented integer like we do for other object IDs.


kmer2016 (approved):
LGTM !

PR #8822, opened 2024-09-10
front: replace some type assertions with non-null assertions

emersion:
In general I prefer to avoid `as` type assertions as much as possible, because it makes the compiler ignore any type mismatch. The non-null assertion operator `!` is a bit better: if the types don't match, the compiler will error out.

Use it in a bunch of places.

See https://github.com/OpenRailAssociation/osrd/pull/8698#discussion_r1745144151

### DIFF
--- a/front/src/applications/editor/tools/pointEdition/components/PointEditionLeftPanel.tsx
+++ b/front/src/applications/editor/tools/pointEdition/components/PointEditionLeftPanel.tsx
@@ -72,24 +72,22 @@ const PointEditionLeftPanel = <Entity extends EditorEntity>({
 
   useEffect(() => {
     const firstLoading = trackState.type === 'idle';
-    const trackId = state.entity.properties.track as string | undefined;
+    const trackId: string | undefined = state.entity.properties.track;
 
     if (trackId && trackState.id !== trackId) {
       setTrackState({ type: 'isLoading', id: trackId });
-      getEntity<TrackSectionEntity>(infraID as number, trackId, 'TrackSection', dispatch).then(
-        (track) => {
-          setTrackState({ type: 'ready', id: trackId, track });
+      getEntity<TrackSectionEntity>(infraID!, trackId, 'TrackSection', dispatch).then((track) => {
+        setTrackState({ type: 'ready', id: trackId, track });
 
-          if (!firstLoading) {
-            const { position } = state.entity.properties;
-            const turfPosition =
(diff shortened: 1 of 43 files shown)


front/src/applications/editor/tools/pointEdition/components/PointEditionLeftPanel.tsx
```diff
@@ -72,24 +72,22 @@ const PointEditionLeftPanel = <Entity extends EditorEntity>({
 
   useEffect(() => {
     const firstLoading = trackState.type === 'idle';
-    const trackId = state.entity.properties.track as string | undefined;
+    const trackId = state.entity.properties.track;
```
SharglutDev: we lose track of the type here. Is that ok ?

SharglutDev (approved):
Lgtm (just one question comment)

front/src/applications/editor/tools/pointEdition/components/PointEditionLeftPanel.tsx
emersion: Hm, what do you mean? Do you mean that the type is now inferred instead of being explicit? If we want to be explicit, we could go for

```ts
const trackId: string | undefined = state.entity.properties.track;
```

but I'm not sure I understand why we should be explicit here specifically? (We use type inference almost everywhere)

SharglutDev: Yes that's not a big deal. With some few `any`'s in the app, I feel scared to pass an any value to an object :) For me the cast was there to attest that we checked/tested the data from this const and we know it can be one of these 2.

But I understand why you would prefer to "break" the app if we misjudged the type (especially if it changes later and we forget to update this part).

emersion: Hm, I hadn't noticed that `state.entity.properties.track` had the type `any`… Then yeah I agree being explicit here is a good idea. Changed.

Note that there are many other places where we use `entity.properties.XXX` and don't cast or anything (so `any` spreads via type inference). We should probably enable the `no-unsafe-assignment` lint to avoid these footguns (see #8772).

SharglutDev: Yeah lots of `any`s in `geojson` and stuff... :)

SarahBellaha (approved):
Lgtm

PR #8827, opened 2024-09-11
front: fix missing rolling stock images in train schedule list

### LINKED ISSUE #8723: Train image disappears after reload
### What happened?

image
image

The bug is present on the local environment but not on the recipe environment.

The train image must be displayed even after a reload

-Go to operational studies. 
-In a scenario, create a new train. 
-See train image
-Reload
-The image is no longer displayed

Local, Dev (SNCF)

Firefox, Google Chrome

dev 53bd1e9

emersion:
We were calling formatTrainScheduleSummaries() before the rolling stock list got fetched. As a result all rolling stocks in the formatted train schedule summaries were null on first load.

Wait for the rolling stock list to be available before kicking off the train schedule summaries lazy loading.

To reproduce:

- Go to a scenario with less than 10 trains, wait for it to load completely (to ensure core is properly spin up)
- Reload the page
- Before this patch, rolling stock images would be missing

Closes: https://github.com/OpenRailAssociation/osrd/issues/8723

### DIFF
--- a/front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
+++ b/front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
@@ -52,7 +52,7 @@ const useLazyLoadTrains = ({
   const [postTrainScheduleSimulationSummary] =
     osrdEditoastApi.endpoints.postTrainScheduleSimulationSummary.useLazyQuery();
 
-  const { data: { results: rollingStocks } = { results: [] } } =
+  const { data: { results: rollingStocks } = { results: null } } =
     osrdEditoastApi.endpoints.getLightRollingStock.useQuery({ pageSize: 1000 });
 
   const trainSchedulesById = useMemo(() => mapBy(trainSchedules, 'id'), [trainSchedules]);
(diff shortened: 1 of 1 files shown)


theocrsb:
This PR will close #8723

theocrsb (approved):
Thanks for this PR! LGTM and tested

front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
```diff
@@ -52,7 +52,7 @@ const useLazyLoadTrains = ({
   const [postTrainScheduleSimulationSummary] =
     osrdEditoastApi.endpoints.postTrainScheduleSimulationSummary.useLazyQuery();
 
-  const { data: { results: rollingStocks } = { results: [] } } =
+  const { data: { results: rollingStocks } = { results: null } } =
```
SharglutDev, line 55: You prefer to default with `null` to oblige us to deal with this case or it's for another reason ?

SharglutDev (approved):
Lgtm and tested, thank you ! (you can close the comment after answering it)

front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
emersion: The `null` default is to figure out whether or not the rolling stocks have been loaded. `null` means not yet fetched from the server, an array means the list has been fetched.

PR #8828, opened 2024-09-11
front: add a 'packageManager' version

woshilapin:
When following the [official procedure](https://yarnpkg.com/getting-started/install) to install `yarn`,
 `corepack` yells about not finding a `packageManager` defined in `package.json`,
and then choose one to add it.

```
! Corepack is about to download https://registry.yarnpkg.com/yarn/-/yarn-1.22.22.tgz
? Do you want to continue? [Y/n] Y

! The local project doesn't define a 'packageManager' field. Corepack will now add one referencing yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e.
! For more details about this field, consult the documentation at https://nodejs.org/api/packages.html#packagemanager
```

:warning: It is to be noted that `yarn` documents installation through `corepack` which itself is still [experimental](https://nodejs.org/api/corepack.html#corepack).

### DIFF
--- a/front/package.json
+++ b/front/package.json
@@ -238,5 +238,6 @@
       "text",
       "cobertura"
     ]
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
 }


emersion:
Does dependabot support upgrading this field?

woshilapin:
> Does dependabot support upgrading this field?

[Not yet](https://github.com/dependabot/dependabot-core/issues/4830)

kmer2016 (approved):
Thank you for this PR !

emersion:
I wonder if we should just live with the warning instead of hardcoding a version which will quickly become outdated.

emersion:
By the way, an outdated package manager is not great because that program is responsible for doing a bunch of network requests, unpacking and writing files to disk. A vulnerability in there would have pretty unfortunate consequences, especially if the tool is pinned to a vulnerable version.

woshilapin:
Note that this is not just a warning, it also modify the `package.json` without asking. Even if we could turn it off, that would mean that each time you use `yarn`, it goes through the `corepack` shim and would display the warning. Not sure how to go from there... It seems both solution are particularly bad.

woshilapin:
OK, found a way to silence it in [the documentation](https://github.com/nodejs/corepack/blob/main/README.md#environment-variables). Just need a `COREPACK_ENABLE_AUTO_PIN=0` that I set up with `direnv`. Works for me. I'll close the PR.

emersion:
Possible workarounds:

- `export COREPACK_ENABLE_AUTO_PIN=0`
- `npm install -g yarn` instead of using corepack

EDIT: ah, seems like you've found the first one already :P

PR #8830, opened 2024-09-11
editoast: add default log directive to display all HTTP calls

woshilapin:
As a webservice, it's convenient that the default logs show all the request that goes in. For each request coming in, we should now see something like this.

```
  2024-09-11T04:58:17.332243Z DEBUG tower_http::trace::on_request: started processing request
    at /home/me/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tower-http-0.5.2/src/trace/on_request.rs
    in tower_http::trace::make_span::request with method: GET, uri: /, version: HTTP/1.1

  2024-09-11T04:58:17.332532Z DEBUG tower_http::trace::on_response: finished processing request, latency: 0 ms, status: 404
    at /home/me/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tower-http-0.5.2/src/trace/on_response.rs
    in tower_http::trace::make_span::request with method: GET, uri: /, version: HTTP/1.1
```

### DIFF
--- a/editoast/Cargo.toml
+++ b/editoast/Cargo.toml
@@ -76,7 +76,10 @@ uuid = { version = "1.10.0", features = ["serde", "v4"] }
 # For batch dependency updates, see editoast/README.md
 
 async-trait = "0.1.82"
-axum = { version = "0.7.5", default-features = false, features = ["multipart"] }
+axum = { version = "0.7.5", default-features = false, features = [
+  "multipart",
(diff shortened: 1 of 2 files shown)


flomonster (approved):
LGTM

I find the 'tower' logs a little less readable than with actix, but it's better than nothing.

anisometropie (approved):
thanks !

PR #8831, opened 2024-09-11
core: fix distance_range_map.to_range_map

Castavo:
Without this fix, we can't simulate trains with more than 1 power restriction range.

I added a test case.

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/standalone_sim/StandaloneSimulation.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/standalone_sim/StandaloneSimulation.kt
@@ -414,6 +414,7 @@ fun getStopPositions(schedule: List<SimulationScheduleItem>): List<Double> {
     return schedule.filter { it.stopFor != null }.map { it.pathOffset.distance.meters }
 }
 
+// TODO: Get rid of this function, by propagating DistanceRangeMap to the whole codebase
 /**
  * Converts a DistanceRangeMap<T> into a legacy RangeMap<Double, T>. Distances are converted to
  * floats (m).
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/fr/sncf/osrd/standalone_sim/StandaloneSimulation.kt
```diff
@@ -422,7 +422,7 @@ private fun <T> DistanceRangeMap<T>.toRangeMap(): RangeMap<Double, T> {
     val res = ImmutableRangeMap.builder<Double, T>()
     for (entry in this) {
         if (entry.value != null)
-            res.put(Range.closed(entry.lower.meters, entry.upper.meters), entry.value!!)
+            res.put(Range.closedOpen(entry.lower.meters, entry.upper.meters), entry.value!!)
```
bougue-pe, line 426: It does the job, but it makes me think of getting rid of this legacy `RangeMap` and work only with `DistanceRangeMap` (at least in `getElectrificationMap()`).
But it's a bit more work, so depending on the emergency and availability, maybe just add a TODO (and a `good_first_issue`?) ?

bougue-pe (approved):
Thanks for the test too!

eckter (approved):
Oof. I was almost certain that building with closed intervals worked, with the old value being overwritten at the transition. I don't know if it was changed but it's kind of bad that we didn't notice it.

Though there's technically no value at the very end with this change. But I don't think it matters.

core/src/main/kotlin/fr/sncf/osrd/standalone_sim/StandaloneSimulation.kt
Castavo: I added the todo see [force push](https://github.com/OpenRailAssociation/osrd/compare/914f29ebb9d8cc1959b8f4b77ffc68f3acfccdb5..62c072a1333e93c5e82aea7cbc79ca36aec22e2c)

Castavo: I created the issue here: https://github.com/OpenRailAssociation/osrd/issues/8857

Castavo:
@eckter I believe the overwriting works when using a MutableRangeMap rather than an ImmutableRangeMap

Erashin (approved):
Yay :)

Issue #8636, opened 2024-08-30
Editoast: being able to identify unrecognized PR in the infrastructure


hamz2a:
If one or more PRs are missing in the infrastructure during pathfinding, an error will be sent to the front-end containing the list of unrecognized PRs.

PR #8837, opened 2024-09-12
editoast: moved utility functions to helpers.rs

hamz2a:
Moved utility functions to `helpers.rs` for reusability across multiple files. This refactor prepares the functions to be used in other parts of the project (#8636).

### DIFF
--- a/editoast/src/views/path.rs
+++ b/editoast/src/views/path.rs
@@ -3,6 +3,9 @@ pub mod pathfinding;
 pub mod projection;
 mod properties;
 
+use std::collections::HashMap;
+
+use editoast_schemas::train_schedule::PathItemLocation;
 pub use pathfinding::pathfinding_from_train_batch;
 
 use editoast_derive::EditoastError;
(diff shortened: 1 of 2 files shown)


woshilapin (approved):
Seems ok. I wonder if we should just put the functions directly in `path.rs`. I know sometimes we create `utils.rs` or `helpers.rs` but they usually contains a bunch of stuff that we can't name correctly (hence `utils` or `helpers` in lack of a better name). In this case, it's only 3 functions, `path.rs` doesn't contain much yet. What do you think?

flomonster (commented):
Same comment as @woshilapin. If we generalize this practice, I'm afraid we'll end up with dozens of helpers.rs files all over the place.

leovalais:
I share the same concerns about this practice as @woshilapin and @flomonster. Maybe turning these functions into static methods of the related models could avoid creating `helpers`, `utils` and such? Wdyt?

hamz2a:
I put all the functions in `path.rs`

flomonster (approved):

woshilapin (approved):
Thank you.

leovalais (approved):
Thanks!

hamz2a:
We no longer need these changes.

Issue #8648, opened 2024-08-30
front: add e2e tests operational studies - full operations


Maymanaf:
### Description and goal

What's this task about, and what is it for?

### Acceptance criteria

A list of conditions which define when this task shall be considered complete

Issue #8647, opened 2024-08-30
[META] add e2e tests operational studies - simulation settings


Maymanaf:
### Description and goal

Write an e2e test to cover https://github.com/osrd-project/osrd-tests/issues/14

- [x] #9239
- [x] #9240
- [x] #9241
- [ ] ~~#9242~~
- [x] #9243

PR #8838, opened 2024-09-12
front: integrate ui-core select component in stdcm

### LINKED ISSUE #8820: Use the new Select for Speed Limit Tag selector
Replace 
image
 By
image

achrafmohye:
Closes #8820

### DIFF
--- a/front/src/applications/stdcmV2/components/StdcmOpSchedule.tsx
+++ b/front/src/applications/stdcmV2/components/StdcmOpSchedule.tsx
@@ -6,6 +6,7 @@ import { useSelector } from 'react-redux';
 
 import { useOsrdConfSelectors } from 'common/osrdContext';
 import { isArrivalDateInSearchTimeWindow } from 'utils/date';
+import { createStringSelectOptions } from 'utils/uiCoreHelpers';
 
 import type { ArrivalTimeTypes, ScheduleConstraint } from '../types';
 
@@ -77,20 +78,21 @@ const StdcmOpSchedule = ({
     }, [opTimingData, opToleranceValues, searchDatetimeWindow]);
 
   return (
-    <div className="d-flex flex-column selector">
-      <Select
-        id={`select-${opId}`}
-        value={opScheduleTimeType}
-        onChange={(e) => {
-          if (e) {
-            onArrivalTypeChange(e as ArrivalTimeTypes);
-          }
-        }}
-        options={['preciseTime', 'asSoonAsPossible']}
-        getOptionLabel={(option) => t(`trainPath.${option}`)}
-        getOptionValue={(option) => option}
-        disabled={disabled}
-      />
+    <div className="d-flex flex-column">
+      <div className="col-12 pr-1">
+        <Select
+          id={`select-${opId}`}
+          value={opScheduleTimeType}
+          onChange={(e) => {
+            if (e) {
+              onArrivalTypeChange(e as ArrivalTimeTypes);
+            }
+          }}
+          {...createStringSelectOptions(['preciseTime', 'asSoonAsPossible'])}
+          getOptionLabel={(option) => t(`trainPath.${option}`)}
+          disabled={disabled}
+        />
+      </div>
       {opScheduleTimeType === 'preciseTime' && (
         <div className="d-flex">
           {/* TODO: Remove empty onChange events once we fix the warning on ui-core side */}
--- a/front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
+++ b/front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
@@ -1,11 +1,11 @@
 import { useMemo } from 'react';
 
+import { Select } from '@osrd-project/ui-core';
 import cx from 'classnames';
 import { isEmpty } from 'lodash';
 import { useTranslation } from 'react-i18next';
 
-import icon from 'assets/pictures/components/speedometer.svg';
-import SelectImprovedSNCF from 'common/BootstrapSNCF/SelectImprovedSNCF';
+import { createStringSelectOptions } from 'utils/uiCoreHelpers';
 
 type SpeedLimitByTagSelectorProps = {
   condensed?: boolean;
@@ -30,6 +30,7 @@ export default function SpeedLimitByTagSelector({
   );
 
   if (!speedLimitsTagsList.length) return null;
+
   return (
     <div className="osrd-config-item mb-2">
       <div
@@ -37,22 +38,20 @@ export default function SpeedLimitByTagSelector({
           'd-flex align-items-center gap-10': condensed,
         })}
       >
-        <img width="32px" src={icon} alt="speedometer" />
-        <span className="text-muted">{t('speedLimitByTag')}</span>
-        <SelectImprovedSNCF
-          sm
+        <Select
           disabled={disabled}
-          withSearch
-          dataTestId="speed-limit-by-tag-selector"
-          value={speedLimitByTag || t('noSpeedLimitByTag').toString()}
-          options={speedLimitsTagsList}
+          id="speed-limit-by-tag-selector"
+          value={speedLimitByTag || ''}
+          label={t('speedLimitByTagAbbrev')}
+          placeholder={t('noSpeedLimitByTag')}
           onChange={(e) => {
-            if (e && e !== t('noSpeedLimitByTag')) {
+            if (e) {
               dispatchUpdateSpeedLimitByTag(e);
             } else {
               dispatchUpdateSpeedLimitByTag(null);
             }
           }}
+          {...createStringSelectOptions(speedLimitsTagsList)}
         />
       </div>
     </div>
--- a/front/src/common/SpeedLimitByTagSelector/useStoreDataForSpeedLimitByTagSelector.ts
+++ b/front/src/common/SpeedLimitByTagSelector/useStoreDataForSpeedLimitByTagSelector.ts
@@ -14,8 +14,6 @@ export const useStoreDataForSpeedLimitByTagSelector = () => {
   const dispatch = useAppDispatch();
   const { t } = useTranslation(['operationalStudies/manageTrainSchedule']);
 
-  const DEFAULT_SPEED_LIMIT_TAG = useMemo(() => t('noSpeedLimitByTag'), [t]);
-
   const { getSpeedLimitByTag } = useOsrdConfSelectors();
   const speedLimitByTag = useSelector(getSpeedLimitByTag);
   const infraID = useInfraID();
@@ -43,12 +41,14 @@ export const useStoreDataForSpeedLimitByTagSelector = () => {
     }
   }, [error]);
 
-  let speedLimitsByTags = compact(uniq(concat(speedLimitTags, speedLimitsTagsByInfraId)));
+  const speedLimitsByTags = compact(uniq(concat(speedLimitTags, speedLimitsTagsByInfraId)));
   const speedLimitsByTagsOrdered = useMemo(() => speedLimitsByTags.sort(), [speedLimitsByTags]);
 
-  speedLimitsByTags = [DEFAULT_SPEED_LIMIT_TAG, ...speedLimitsByTagsOrdered];
-
-  return { speedLimitByTag, speedLimitsByTags, dispatchUpdateSpeedLimitByTag };
+  return {
+    speedLimitByTag,
+    speedLimitsByTags: speedLimitsByTagsOrdered,
+    dispatchUpdateSpeedLimitByTag,
+  };
 };
 
 export default useStoreDataForSpeedLimitByTagSelector;
--- a/front/src/utils/uiCoreHelpers.ts
+++ b/front/src/utils/uiCoreHelpers.ts
@@ -1,0 +1,24 @@
+// TODO: Remove undefined when osrd-ui's value type is optional
+type SelectFixedOption =
+  | {
+      label: string;
+      id: string;
+    }
+  | undefined;
+
+// Utility to create standardized options for <Select> components.
+// `getOptionLabel`: Displays the option or a placeholder if empty.
+// `getOptionValue`: Uses the option itself as the value.
+export const createStringSelectOptions = (opts: string[]) => ({
+  options: opts,
+  getOptionLabel: (option: string) => option,
+  getOptionValue: (option: string) => option,
+});
+
+// Utility function for handling fixed label-value object arrays
+// structure may vary, and custom label/value mappings are needed.
+export const createFixedSelectOptions = (opts: SelectFixedOption[]) => ({
+  options: opts,
+  getOptionLabel: (option: SelectFixedOption) => option?.label ?? '',
+  getOptionValue: (option: SelectFixedOption) => option?.id ?? '',
+});
--- a/front/tests/005-operational-studies.spec.ts
+++ b/front/tests/005-operational-studies.spec.ts
@@ -74,15 +74,6 @@ test.describe('Testing if all mandatory elements simulation configuration are lo
 
     await scenarioPage.checkPathfindingDistance('34.000 km');
 
-    // TODO: move this test in his own file
-    // ***************** Test Composition Code *****************
-    await scenarioPage.openTabByDataId('tab-simulation-settings');
-    await expect(scenarioPage.getSpeedLimitSelector).toBeVisible();
-    await scenarioPage.getSpeedLimitSelector.click();
-    await scenarioPage.getSpeedLimitSelector.locator('input').fill('32');
-    await scenarioPage.getSpeedLimitSelector.getByRole('button', { name: 'E32C' }).click();
-    expect(await scenarioPage.getSpeedLimitSelector.textContent()).toMatch(/E32C/i);
-
     // ***************** Test Add Train Schedule *****************
     await scenarioPage.addTrainSchedule();
     await scenarioPage.page.waitForSelector('.dots-loader', { state: 'hidden' });
--- a/front/tests/pages/scenario-page-model.ts
+++ b/front/tests/pages/scenario-page-model.ts
@@ -13,8 +13,6 @@ class ScenarioPage extends BasePage {
 
   readonly getRollingStockSelector: Locator;
 
-  readonly getSpeedLimitSelector: Locator;
-
   readonly getItineraryModule: Locator;
 
   readonly getItineraryOrigin: Locator;
@@ -101,7 +99,6 @@ class ScenarioPage extends BasePage {
     this.getScenarioDeleteConfirmBtn = page
       .locator('#modal-content')
       .getByText('Supprimer', { exact: true });
-    this.getSpeedLimitSelector = page.getByTestId('speed-limit-by-tag-selector');
     this.getItineraryModule = page.getByTestId('itinerary');
     this.getItineraryOrigin = this.getItineraryModule
       .getByTestId('display-itinerary')
--- a/front/public/locales/en/operationalStudies/manageTrainSchedule.json
+++ b/front/public/locales/en/operationalStudies/manageTrainSchedule.json
@@ -116,6 +116,7 @@
   "saving": "Saving... please wait",
   "simulation": "Simulation",
   "speedLimitByTag": "Composition code",
+  "speedLimitByTagAbbrev": "Code compo.",
   "stopSimulation": "Stop",
   "tabs": {
     "rollingStock": "Rolling stock",
--- a/front/public/locales/fr/operationalStudies/manageTrainSchedule.json
+++ b/front/public/locales/fr/operationalStudies/manageTrainSchedule.json
@@ -117,6 +117,7 @@
   "saving": "Enregistrement en cours... veuillez patienter",
   "simulation": "Simulation",
   "speedLimitByTag": "Code de composition",
+  "speedLimitByTagAbbrev": "Code compo.",
   "stopSimulation": "Stopper",
   "tabs": {
     "rollingStock": "Matériel",
--- a/front/src/applications/stdcmV2/components/StdcmConsist.tsx
+++ b/front/src/applications/stdcmV2/components/StdcmConsist.tsx
@@ -144,7 +144,6 @@ const StdcmConsist = ({ setCurrentSimulationInputs, disabled = false }: StdcmCon
           <Input id="tonnage" label={t('consist.tonnage')} trailingContent="t" />
           <Input id="length" label={t('consist.length')} trailingContent="m" />
         </div>
-        <p className="stdcm-v2-consist__title">{t('translation:Editor.nav.speed-limits')}</p>
         <SpeedLimitByTagSelector
           disabled={disabled}
           selectedSpeedLimitByTag={speedLimitByTag}
--- a/front/src/applications/stdcmV2/components/StdcmOperationalPoint.tsx
+++ b/front/src/applications/stdcmV2/components/StdcmOperationalPoint.tsx
@@ -7,6 +7,7 @@ import nextId from 'react-id-generator';
 import type { SearchResultItemOperationalPoint } from 'common/api/osrdEditoastApi';
 import useSearchOperationalPoint from 'common/Map/Search/useSearchOperationalPoint';
 import type { PathStep } from 'reducers/osrdconf/types';
+import { createFixedSelectOptions } from 'utils/uiCoreHelpers';
 
 import StdcmSuggestions from './StdcmSuggestions';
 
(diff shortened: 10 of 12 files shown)


front/tests/pages/scenario-page-model.ts
```diff
@@ -101,7 +101,7 @@ class ScenarioPage extends BasePage {
     this.getScenarioDeleteConfirmBtn = page
       .locator('#modal-content')
       .getByText('Supprimer', { exact: true });
-    this.getSpeedLimitSelector = page.getByTestId('speed-limit-by-tag-selector');
+    this.getSpeedLimitSelector = page.locator('speed-limit-by-tag-selector');
     this.getItineraryModule = page.getByTestId('itinerary');
```
achrafmohye, line 102: Taking a look on the tests 🧐

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
```diff
@@ -39,11 +39,10 @@ export default function SpeedLimitByTagSelector({
       >
         <img width="32px" src={icon} alt="speedometer" />
         <span className="text-muted">{t('speedLimitByTag')}</span>
-        <SelectImprovedSNCF
-          sm
+        <Select
+          small
```
SharglutDev: In the mockup, the component seems to have the same font size as the rest of the form

```suggestion
```

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
```diff
@@ -53,6 +52,8 @@ export default function SpeedLimitByTagSelector({
               dispatchUpdateSpeedLimitByTag(null);
             }
           }}
+          getOptionLabel={(option) => option}
+          getOptionValue={(option) => option}
```
SharglutDev: Still find pretty weird that we will have to do this each time we use this component

achrafmohye: Would seem to agree, but we have to accept an array of strings for the suggestions or have a specific format for them !

kmer2016: The purpose of these two props is to allow users to pass their data in its raw form without needing to transform it beforehand. You provide the data as-is and pass two callbacks to specify how to display and identify each element in the list. The benefit becomes more apparent when dealing with more complex data than simple strings.

Maybe we could make these props optional if the list of select options consists of strings?

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
```diff
@@ -39,11 +39,10 @@ export default function SpeedLimitByTagSelector({
...
         <span className="text-muted">{t('speedLimitByTag')}</span>
-        <SelectImprovedSNCF
-          sm
+        <Select
+          small
           disabled={disabled}
-          withSearch
-          dataTestId="speed-limit-by-tag-selector"
+          id="speed-limit-by-tag-selector"
           value={speedLimitByTag || t('noSpeedLimitByTag').toString()}
```
emersion: Would be nicer to not use strings intended for humans (such as translations) as `<select>` values (which are meant to be unique and are not displayed to the user). A simple solution is to move `|| t('noSpeedLimitByTag').toString()` to `getOptionLabel()`. Then we can get rid of the `e !== t('noSpeedLimitByTag')` check.

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
emersion: I agree that allowing `<Select>` users to pass their own data type is worthwhile, and I also agree that for many cases it's cumbersome to pass trivial `getOptionLabel`/`getOptionValue` callbacks. I don't recommend adding more knobs to `<Select>` props because it would increase the complexity of its API. Instead, a separate helper function could be introduced. Here's one of the possible solutions:

```tsx
type SelectFixedOption = {
  label: string;
  value: string;
};

const withFixedOptions = (options: SelectFixedOption[]) => {
  return {
    options,
    getOptionLabel: (option) => option.label,
    getOptionValue: (option) => option.value,
  };
};

// usage:
const options = speedLimitsTagsList.map((tag) => ({ label: tag, value: tag }));
<Select {...withFixedOptions(options)} />
```

Many variations of this are possible as well, e.g. only providing `getOptionLabel`/`getOptionValue` helpers for `string[]` (but wouldn't work if you want to use a label different from the value, which is quite common):

```tsx
const withStringArray = (options: string[]) => {
  return {
    options,
    getOptionLabel: (option) => option,
    getOptionValue: (option) => option,
  };
};

// usage:
<Select {...withStringArray(speedLimitsTagsList)} />
```

or replacing `SelectOption.value` with an `onSelect` callback:

```tsx
type SelectFixedOption = {
  label: string;
  onSelect: () => void;
};

const withFixedOptions = (options: SelectFixedOption[]) => {
  const optionsMap = new Map(options.map((option, index) => [option, index.toString()]));
  return {
    options,
    getOptionLabel: (option) => option.label,
    getOptionValue: (option) => optionsMap.get(option),
    onChange: (option) => option?.onSelect(),
  };
};

// usage:
const options = speedLimitsTagsList.map((tag) =>
  ({ label: tag, onSelect: () => dispatchUpdateSpeedLimitByTag(tag) })
);
<Select {...withFixedOptions(options)} />
```

emersion: TBH… I'm not even sure why `<Select>` requires callers to pass a `value` at all. I think we could just auto-generate a `value` for each option based on their index. We wouldn't loose the current selection on component re-renders (e.g. when the option list changes) because internally `<Select>` does `value={selectedOption ? getOptionValue(selectedOption) : undefined}`.

emersion: I now realize that if we want to drop `value`, we wouldn't even need to generate them based on indexes: we can just grab the selected index via [`selectedIndex`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/selectedIndex).

achrafmohye: i like the solution of having `withFixedOptions` but if have to change the internal functioning of the `Select`   component we can create an issue in` ui-core`

kmer2016: > I agree that allowing `<Select>` users to pass their own data type is worthwhile, and I also agree that for many cases it's cumbersome to pass trivial `getOptionLabel`/`getOptionValue` callbacks. I don't recommend adding more knobs to `<Select>` props because it would increase the complexity of its API. Instead, a separate helper function could be introduced. Here's one of the possible solutions:
> 
> ```tsx
> type SelectFixedOption = {
>   label: string;
>   value: string;
> };
> 
> const withFixedOptions = (options: SelectFixedOption[]) => {
>   return {
>     options,
>     getOptionLabel: (option) => option.label,
>     getOptionValue: (option) => option.value,
>   };
> };
> 
> // usage:
> const options = speedLimitsTagsList.map((tag) => ({ label: tag, value: tag }));
> <Select {...withFixedOptions(options)} />
> ```
> 
> Many variations of this are possible as well, e.g. only providing `getOptionLabel`/`getOptionValue` helpers for `string[]` (but wouldn't work if you want to use a label different from the value, which is quite common):
> 
> ```tsx
> const withStringArray = (options: string[]) => {
>   return {
>     options,
>     getOptionLabel: (option) => option,
>     getOptionValue: (option) => option,
>   };
> };
> 
> // usage:
> <Select {...withStringArray(speedLimitsTagsList)} />
> ```
> 
> or replacing `SelectOption.value` with an `onSelect` callback:
> 
> ```tsx
> type SelectFixedOption = {
>   label: string;
>   onSelect: () => void;
> };
> 
> const withFixedOptions = (options: SelectFixedOption[]) => {
>   const optionsMap = new Map(options.map((option, index) => [option, index.toString()]));
>   return {
>     options,
>     getOptionLabel: (option) => option.label,
>     getOptionValue: (option) => optionsMap.get(option),
>     onChange: (option) => option?.onSelect(),
>   };
> };
> 
> // usage:
> const options = speedLimitsTagsList.map((tag) =>
>   ({ label: tag, onSelect: () => dispatchUpdateSpeedLimitByTag(tag) })
> );
> <Select {...withFixedOptions(options)} />
> ```

I totally agree with you. Thank you for your helpful feedback and examples!

kmer2016: > I now realize that if we want to drop `value`, we wouldn't even need to generate them based on indexes: we can just grab the selected index via [`selectedIndex`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/selectedIndex).

I see your point about auto-generating values based on the index. However, I believe keeping the value prop is important. Without it, we wouldn’t be able to initialize the Select component with a value from a Redux store. This makes it harder to set the correct selected option when loading from the store.
Do you have any suggestions on how to handle this situation with your proposal?

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
achrafmohye: @emersion can you check the fix and if it's done properly

front/tests/005-operational-studies.spec.ts
```diff
@@ -93,15 +93,6 @@ test.describe('Testing if all mandatory elements simulation configuration are lo
 
     await scenarioPage.checkPathfindingDistance('34.000 km');
 
-    // TODO: move this test in his own file
```
achrafmohye, line 77: @Maymanaf like discussed can you check the test part :)

Maymanaf: This part of the tests covers the old component and is no longer needed, aas we will have two tests that address the new behavior:

- A new test: https://github.com/OpenRailAssociation/osrd/issues/8647
- A refactor of test 005: https://github.com/OpenRailAssociation/osrd/issues/8648

Maymanaf (approved):

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
```diff
@@ -39,20 +41,14 @@ export default function SpeedLimitByTagSelector({
...
+          value={speedLimitByTag || ''}
           onChange={(e) => {
-            if (e && e !== t('noSpeedLimitByTag')) {
-              dispatchUpdateSpeedLimitByTag(e);
-            } else {
-              dispatchUpdateSpeedLimitByTag(null);
-            }
+            dispatchUpdateSpeedLimitByTag(e || null);
           }}
+          {...createStringSelectOptions(speedLimitsTagsList, t('noSpeedLimitByTag').toString())}
```
SharglutDev: Don't think you need to do that, the default one is already added in the list in `useStoreDataForSpeedLimitByTagSelector`

```suggestion
          {...createStringSelectOptions(speedLimitsTagsList)}
```

front/src/utils/uiCoreHelpers.ts
```diff
@@ -0,0 +1,6 @@
+// eslint-disable-next-line import/prefer-default-export
+export const createStringSelectOptions = (opts: string[], labelPlaceholder?: string) => ({
```
SharglutDev: Maybe a comment of why we need to have this util :)

SharglutDev: Should the function be more generic in case `options` are a bit more complex ?

achrafmohye: I can introduce a second function for a usage of non string cases where the options are objects ?

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
achrafmohye: Done

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
```diff
@@ -30,29 +30,23 @@ export default function SpeedLimitByTagSelector({
...
       <div
         className={cx('osrd-config-item-container', {
           'd-flex align-items-center gap-10': condensed,
         })}
       >
-        <img width="32px" src={icon} alt="speedometer" />
-        <span className="text-muted">{t('speedLimitByTag')}</span>
-        <SelectImprovedSNCF
-          sm
+        <Select
```
SharglutDev, line 41: This will also change the UI for the selector in operational studies when adding/editing a train, resulting to a "zombified" interface. Do we have the ok from PO's or @thibautsailly ?
Moreover, in operational studies, the label will dislay "Code compo" instead of "Code de composition". Is it a change we want ?

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
```diff
@@ -30,29 +30,23 @@ export default function SpeedLimitByTagSelector({
...
+          id="speed-limit-by-tag-selector"
+          value={speedLimitByTag || ''}
+          label={t('speedLimitByTagAbbrev')}
           onChange={(e) => {
-            if (e && e !== t('noSpeedLimitByTag')) {
-              dispatchUpdateSpeedLimitByTag(e);
-            } else {
-              dispatchUpdateSpeedLimitByTag(null);
-            }
+            dispatchUpdateSpeedLimitByTag(e || null);
```
SharglutDev: We no longer filter the `noSpeedLimitByTag` value, meaning when we select it, the store is updated with that and when adding/editing the train, we send this non consistent value instead of null. I'm afraid it might result to some weird things in the simulation. Maybe we should leave it like it was before ?

front/src/utils/uiCoreHelpers.ts
```diff
@@ -0,0 +1,23 @@
...
+      id: string;
+    }
+  | undefined;
+
+// Utility to create standardized options for <Select> components.
+// `getOptionLabel`: Displays the option or a placeholder if empty.
+// `getOptionValue`: Uses the option itself as the value.
+export const createStringSelectOptions = (opts: string[]) => ({
+  options: opts,
+  getOptionLabel: (option: string) => option || '',
```
SharglutDev: Don't think we need that

```suggestion
  getOptionLabel: (option: string) => option,
```

front/src/applications/stdcmV2/components/StdcmOpSchedule.tsx
```diff
@@ -95,9 +96,8 @@ const StdcmOpSchedule = ({
               onArrivalTypeChange(e as ArrivalTimeTypes);
             }
           }}
-          options={['preciseTime', 'asSoonAsPossible']}
+          {...createStringSelectOptions(['preciseTime', 'asSoonAsPossible'])}
```
SharglutDev, line 91: Wasn't the point of using `createStringSelectOptions` not to have to pass `getOptionLabel` ?

front/src/utils/uiCoreHelpers.ts
SharglutDev: Would have been better to be able to have one function but it seems complicated

front/src/applications/stdcmV2/components/StdcmOpSchedule.tsx
achrafmohye: Yes but it's a special case where we have a translation as label don't think that we want a function for each case ?

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
achrafmohye: It's the label suggested by @thibautsailly in the sketch, going to ask him if it's the the good label to have

achrafmohye: Here is the respone of @thibautsailly `To respect the width of the component, we don't really have a choice but to shorten the term, so it's indeed what is on the mockup that needs to be written, in English, no need to translate, you can leave it as 'code compo.'`

front/src/applications/stdcmV2/components/StdcmOpSchedule.tsx
SharglutDev: Hmm ok :)

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
achrafmohye: @thibautsailly is ok with having ui-core component in Operational studies, showed him a before and after screens of the component used simulation settings to make sure :)

SharglutDev: Ok for @thibautsailly

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
emersion: > we wouldn’t be able to initialize the Select component with a value from a Redux store

What's in the Redux store is completely orthogonal to the props passed to `Select` I believe. The current selected option in `<Select value=…>` is of type `T`, it's _not_ a `string` compared to `getOptionValue`.

In other words, I don't think dropping `getOptionValue` in `Select` would make any difference: in OSRD we would just drop it and everything would still continue to work as before.

See https://github.com/OpenRailAssociation/osrd-ui/pull/538

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
```diff
@@ -30,29 +30,27 @@ export default function SpeedLimitByTagSelector({
...
       >
-        <img width="32px" src={icon} alt="speedometer" />
-        <span className="text-muted">{t('speedLimitByTag')}</span>
-        <SelectImprovedSNCF
-          sm
+        <Select
           disabled={disabled}
-          withSearch
-          dataTestId="speed-limit-by-tag-selector"
-          value={speedLimitByTag || t('noSpeedLimitByTag').toString()}
```
emersion, line 47: I think we loose the ability to clear the speed limit by tag selection?

Could be fixed with something like `placeholder={t('noSpeedLimitByTag')}`.

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
emersion: Yeah, sounds good!

front/src/utils/uiCoreHelpers.ts
```diff
@@ -0,0 +1,23 @@
+type SelectFixedOption =
+  | {
+      label: string;
+      id: string;
+    }
+  | undefined;
```
emersion, line 7: Why allow `undefined` here?

front/src/utils/uiCoreHelpers.ts
```diff
@@ -0,0 +1,23 @@
...
+  getOptionValue: (option: string) => option,
+});
+
+// Utility function for handling fixed label-value object arrays
+// structure may vary, and custom label/value mappings are needed.
+export const createFixedSelectOptions = (opts: SelectFixedOption[]) => ({
+  options: opts,
+  getOptionLabel: (option: SelectFixedOption) => option?.label ?? '',
+  getOptionValue: (option: SelectFixedOption) => option?.id ?? '',
+});
```
emersion, line 24: It would probably be best to have these helpers in osrd-ui?

front/src/utils/uiCoreHelpers.ts
achrafmohye: To accommodate the the `value` and `options` of the select in `StdcmOperationalPoint` since the value can be undefined

front/src/utils/uiCoreHelpers.ts
achrafmohye: Had this discussion with @kmer2016, we talked about having a default behaviour at least for the case where the `options` are simple strings and for complex types keep the getOptionLabel optional used when needed

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
achrafmohye: a placeholder can be a good enhancement

front/src/utils/uiCoreHelpers.ts
achrafmohye: in matter fact the default value of the `Select` is already added in `useStoreDataForSpeedLimitByTagSelector` so i don't think it is necessary since it's the first value of the options passed to the `Select`

achrafmohye: @emersion what do you think ?

clarani (dismissed):
- can you rebase so we can see the new inputs added by Sarah ?
- the inputs are not aligned
image

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
emersion: I mean, we're loosing functionality here, it's a regression, no?

front/src/utils/uiCoreHelpers.ts
emersion: Not sure I understand, but it sounds a bit too automagical/surprising to me…

front/src/utils/uiCoreHelpers.ts
emersion: `sortedChOptions` can never contain an `undefined` item. Sorry, I don't understand the justification.

`undefined` items don't make sense to me, because `Select` has a separate mechanism to unset the selected option: `placeholder`.

front/src/utils/uiCoreHelpers.ts
kmer2016: I suggest we keep the `utils` in osrd for now, and I'll prepare an issue on the osrd-ui side to move them later. Does that sound good to you?

emersion: Sure!

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
emersion: To expand on this a bit: before a user could select a speed limit by tag and then unselect it. Now there is no way to unselect it. Once selected, the user is stuck with the speed limit by tag.

front/src/utils/uiCoreHelpers.ts
kmer2016: ok, here the [issue](https://github.com/OpenRailAssociation/osrd-ui/issues/555)

front/src/utils/uiCoreHelpers.ts
emersion: Actually that's not possible because of an osrd-ui bug: https://github.com/OpenRailAssociation/osrd-ui/pull/557

kmer2016 (approved):
LGTM !

emersion:
Can you squash the commits together?

emersion:
Ow wait, is it intentional that this PR changes the playwright config?

emersion:
diff also contains other unrelated changes, such as ESLint commands to disable some lints. However I don't see why this PR touches these unrelated parts of the code and why we should add these when the lints are already globally disabled in `.eslintrc`…

front/src/common/SpeedLimitByTagSelector/useStoreDataForSpeedLimitByTagSelector.ts
```diff
@@ -43,12 +41,14 @@ export const useStoreDataForSpeedLimitByTagSelector = () => {
     }
   }, [error]);
 
-  let speedLimitsByTags = compact(uniq(concat(speedLimitTags, speedLimitsTagsByInfraId)));
+  const speedLimitsByTags = compact(uniq(concat(speedLimitTags, speedLimitsTagsByInfraId)));
   const speedLimitsByTagsOrdered = useMemo(() => speedLimitsByTags.sort(), [speedLimitsByTags]);
```
emersion, line 45: Random drive-by comment: this `useMemo` never memorizes anything because its input changes at each render.

(But that's unrelated to this PR.)

emersion:
The two comments above about unrelated changes making their way into this PR are now resolved.

Maymanaf (approved):
LGTM for e2e tests

emersion (approved):
LGTM (not tested)

PR #8852, opened 2024-09-12
front: fix spacetime chart not displayed if no op on path

### LINKED ISSUE #8174: front: display the via added by click on the map in the manchette

SharglutDev:
Trying to push something in an empty array was considered as mutating the array and ended up with an error in console.

fix #8174

### DIFF
--- a/front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
+++ b/front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
@@ -8,7 +8,6 @@ import {
   type PathProperties,
   type TrainScheduleResult,
 } from 'common/api/osrdEditoastApi';
-import { addElementAtIndex } from 'utils/array';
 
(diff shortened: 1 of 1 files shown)


emersion:
Since we're copying the array up-front, we don't need to use `addElementAtIndex()` anymore, we can use `operationalPointsWithAllWaypoints.splice(indexToInsert, 0, formattedStep)`.

emersion (approved):
LGTM :)

clarani (approved):
LGTM :white_check_mark:

achrafmohye (approved):
LGTM :)

PR #8856, opened 2024-09-12
editoast, schemas: add 'freight_compatible' to the rolling stock model

woshilapin:
Classify rolling stock between those that can be used as fret, and those who don't (or those we don't know). A new field `freight_compatible` is added to the rolling stock schema (both python `osrd_schemas` and `editoast`), then also to the rolling stock model (DB), and finally in all API exposed struct like `LightRollingStock` and `RollingStockForm`.

Since there is still a lot of rolling stock for which we don't know any information, the new field is nullable.

⚠️ First thing to review: what do you think of the naming `freight_compatible`. If you don't like it, have other ideas, don't spend time reviewing the rest of the PR, let's discuss that first.

I also added a little icon for freight compatible train in the rolling stock editor.

image

### DIFF
--- a/editoast/migrations/2024-09-06-114952_add-rolling-stock-freight-compatible/up.sql
+++ b/editoast/migrations/2024-09-06-114952_add-rolling-stock-freight-compatible/up.sql
@@ -1,0 +1,3 @@
+-- Your SQL goes here
+ALTER TABLE rolling_stock
+ADD freight_compatible BOOLEAN;
--- a/editoast/migrations/2024-09-06-114952_add-rolling-stock-freight-compatible/up.sql
+++ b/editoast/migrations/2024-09-06-114952_add-rolling-stock-freight-compatible/up.sql
@@ -1,3 +1,3 @@
 -- Your SQL goes here
 ALTER TABLE rolling_stock
-ADD freight_compatible BOOLEAN;
+ADD freight_compatible BOOLEAN NOT NULL DEFAULT FALSE;
--- a/python/osrd_schemas/osrd_schemas/rolling_stock.py
+++ b/python/osrd_schemas/osrd_schemas/rolling_stock.py
@@ -226,6 +226,9 @@ class RollingStock(BaseModel, extra="forbid"):
         ge=0,
     )
     supported_signaling_systems: List[str] = Field(default_factory=list)
+    freight_compatible: Optional[bool] = Field(
+        description="The type of the train, wether it's a locomotive or something else", default=None
+    )
 
 
 if __name__ == "__main__":
--- a/python/osrd_schemas/osrd_schemas/rolling_stock.py
+++ b/python/osrd_schemas/osrd_schemas/rolling_stock.py
@@ -12,7 +12,7 @@ from pydantic import (
 
 from .infra import LoadingGaugeType
 
-RAILJSON_ROLLING_STOCK_VERSION_TYPE = Literal["3.2"]
+RAILJSON_ROLLING_STOCK_VERSION_TYPE = Literal["3.3"]
 RAILJSON_ROLLING_STOCK_VERSION = get_args(RAILJSON_ROLLING_STOCK_VERSION_TYPE)[0]
 
 
--- a/tests/data/rolling_stocks/electric_rolling_stock.json
+++ b/tests/data/rolling_stocks/electric_rolling_stock.json
@@ -977,5 +977,6 @@
     "BAPR",
     "TVM300",
     "TVM430"
-  ]
+  ],
+  "freight_compatible": false
 }
--- a/editoast/README.md
+++ b/editoast/README.md
@@ -94,7 +94,7 @@ To make sure it is always valid a CI check has been set up. To update the
 OpenApi when a change has been made to an endpoint, run the following command:
 
 ```sh
-cargo run openapi > openapi.yaml
+cargo run -- openapi > openapi.yaml
 ```
 
 ## Batch dependency updates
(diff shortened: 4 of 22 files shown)


editoast/migrations/2024-09-06-114952_add-rolling-stock-freight-compatible/up.sql
```diff
@@ -0,0 +1,3 @@
+-- Your SQL goes here
+ALTER TABLE rolling_stock
+ADD freight_compatible BOOLEAN;
```
leovalais, line 3: Instead of making it nullable, wdyt about making it `DEFAULT FALSE`? (In the models only, having it `null | undefined` in the schemas and API can be convenient.)

python/osrd_schemas/osrd_schemas/rolling_stock.py
shenriotpro, line 1: Please update the version number

tests/data/rolling_stocks/electric_rolling_stock.json
```diff
@@ -977,5 +977,6 @@
     "BAPR",
     "TVM300",
     "TVM430"
-  ]
+  ],
+  "freight_compatible": false
```
shenriotpro, line 981: Why not make this one compatible?

Wadjetz (approved):

leovalais (approved):
LGTM. I'm fine with `freight_compatible`. About its nullability: do we plan to make the field non-nullable eventually or is there a benefit to encode "we're not sure" in the application?

tests/data/rolling_stocks/electric_rolling_stock.json
woshilapin: Mostly because this one is electric, and I don't think there is any electric train for fret at the moment. Also because it felt it would be good to have different kind of values for test data. What do you thint?

python/osrd_schemas/osrd_schemas/rolling_stock.py
woshilapin: Done in e79759dee. It was this version you were talking about?

shenriotpro: Sorry, I was talking about this version https://github.com/OpenRailAssociation/osrd/blob/dev/python/osrd_schemas/pyproject.toml#L3

tests/data/rolling_stocks/electric_rolling_stock.json
shenriotpro: Alright, can't argue with physics.

flomonster:
I would have preferred an enum field like `transport_type`.

With these possible values:

- `freight`
- `passenger`
- `construction` for stuff like [tamping machines](https://en.wikipedia.org/wiki/Tamping_machine)
- ...

woshilapin:
> I would have preferred an enum field like `transport_type`.
> 
> With these possible values:
> 
>     * `freight`
> 
>     * `passenger`
> 
>     * `construction` for stuff like [tamping machines](https://en.wikipedia.org/wiki/Tamping_machine)
> 
>     * ...

@flomonster I see, but I don't think that is ok, because it's not an XOR. A locomotive can be both freight and passenger compatible. A better modeling would be some sort of list of tags... but that might mean another DB table and maybe a bit more design. What do you think?

python/osrd_schemas/osrd_schemas/rolling_stock.py
woshilapin: Done in [5c3faac](https://github.com/OpenRailAssociation/osrd/pull/8856/commits/5c3faacc323ecb3a2c50b48c9580b31c880a83d6). I left the other modification too since it looks fine to bump it too.

shenriotpro (approved):
Thanks

flomonster:
> I see, but I don't think that is ok, because it's not an XOR. A locomotive can be both freight and passenger compatible. A better modeling would be some sort of list of tags... but that might mean another DB table and maybe a bit more design. What do you think?

I see, why do we need this information attached to the rolling stock? Is it only for the icon?

woshilapin:
> I see, why do we need this information attached to the rolling stock? Is it only for the icon?

@flomonster No, we're going to use it to filter what is presented in the STDCM form, see https://github.com/OpenRailAssociation/osrd/pull/8921 for the following PR. The icon was just a quick win to display this information somewhere.

flomonster:
> > I see, why do we need this information attached to the rolling stock? Is it only for the icon?
> 
> No, we're going to use it to filter what is presented in the STDCM form, see #8921 for the following PR. The icon was just a quick win to display this information somewhere.

I think the best way to deal with this in the long term is to have groups of rolling stock (necessary for permissions). Then we could assign a group of rolling stock to the STDCM search environment.

It feels weird to add a flaky field and limit usage of STDCM depending of this field.

> [!NOTE]
> Since we want this feature quickly and we can't wait for rolling stock groups, should we add an hotfix in the frontend? It seems easier to revert than changing our model.

woshilapin:
> LGTM. I'm fine with `freight_compatible`. About its nullability: do we plan to make the field non-nullable eventually or is there a benefit to encode "we're not sure" in the application?

@leovalais I actually went back and forth on this idea, first thinking, if we don't know, we just don't know, it's not necessary false.  But then I realize that on STDCM form, we're going to have to take a decision about "do we display rolling stocks that are `null` or not?" and the answer is not clear. So I made the change to simplify the code in [78b1852](https://github.com/OpenRailAssociation/osrd/pull/8856/commits/78b18527f84b10ad5bdcb8085ae49e5fd7de647b).

clarani (approved):
LGTM for the front part :white_check_mark:

woshilapin:
Finally closing this PR. After discussion with multiple actors (@clarani @flomonster @axrolld), we realized that the vision is not yet clear and that we can achieve our short-term objective differently: we can have much less technical debt by encoding what we need directly in `front` without modifying our internal backend model.

PR #8860, opened 2024-09-13
editoast: migration: migrate speed-limit tags to short ids

bougue-pe:


bougue-pe:
Deployed on integration yesterday, working fine (but no use having this un codebase).

PR #8865, opened 2024-09-13
schemas: add from_dict to RollingResistance and Gamma

elise-chin:
To simplify the process of creating instances from a dictionary, instead of manually extracting each attribute and passing them to the class constructor.


Castavo (changes requested):
That's already something you can achieve with pydantic models using `<class>.model_validate(your_dict)`

elise-chin:
> That's already something you can achieve with pydantic models using `<class>.model_validate(your_dict)`

Thanks! I didn't know. Since we don't need to change anything, I will close the PR.

PR #8876, opened 2024-09-16
front: drop unused dependencies

emersion:
Some packages were included in package.json but weren't actually used anymore. Drop them.

Also move @vitejs/plugin-react-swc, @vitest/coverage-v8 and ts-node to devDependencies, since it's build tooling.


flomonster (approved):
:pray:  Thanks for this cleanup.

clarani (approved):
Wow thanks :pray:

PR #8880, opened 2024-09-16
front: fix the warning display above the GET

### LINKED ISSUE #8878: front: the warning above the GET doesn't disappear if a timetable has some invalid trains
### What happened?

When the GET is loading, a warning is displayed above the GET to inform the user that the data is loading.
However, if the timetable has some invalid trains, then the warning never disappear.
image

The warning should disappear once all the trains are projected

1. Open a scenario with some valid and invalid trains
2. Select a valid train and project on a valid train
3. See the GET

_No response_

_No response_

.

clarani:
closes #8878 

Add a new boolean allTrainsProjected and display the warning "Projecting..." only if this boolean is not truthy.

### DIFF
--- a/front/src/modules/simulationResult/components/SpaceTimeChart/useLazyProjectTrains.ts
+++ b/front/src/modules/simulationResult/components/SpaceTimeChart/useLazyProjectTrains.ts
@@ -21,11 +21,11 @@ const BATCH_SIZE = 10;
 
 type useLazyLoadTrainsProp = {
   infraId?: number;
-  trainIdsToProject: number[];
+  trainIdsToProject: Set<number>;
   path?: PathfindingResultSuccess;
   trainSchedules?: TrainScheduleResult[];
   moreTrainsToCome?: boolean;
-  setTrainIdsToProject: Dispatch<SetStateAction<number[]>>;
+  setTrainIdsToProject: Dispatch<SetStateAction<Set<number>>>;
 };
 
 /**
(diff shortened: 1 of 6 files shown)


emersion:
Could we instead use a `useMemo` with a callback checking `shouldProjectIds.length === 0`?

anisometropie:
it doesn’t go away after I just created 100 trains

Screenshot_20240918_134915

anisometropie:
weirdly the too figures are not synchronized 40 on the left 30 on the right
Screenshot_20240918_135020

clarani:
@anisometropie thanks for your review :pray: 
- thanks for the bug, I fixed it in the 3rd commit
- for your 2nd remark, actually it's normal that the counters are de-synchronized, since the loading of the trains and the projection are done in parallel and projection is done after the loading

emersion:
Another way to fix this without a desync would be to subtract the number of invalid trains to the total number of simulated trains. But the current approach here looks fine to me as well.

anisometropie (approved):
thanks!

front/src/modules/simulationResult/components/SpaceTimeChart/useLazyProjectTrains.ts
```diff
@@ -121,7 +122,7 @@ const useLazyProjectTrains = ({
     if (
       !moreTrainsToCome &&
       trainIdsToProject.length > 0 &&
-      requestedProjectedTrainIds.current.size === trainIdsToProject.length
+      requestedProjectedTrainIds.current.size === removeDuplicates(trainIdsToProject).length
```
emersion: Would be nicer to use a `Set` for `trainIdsToProject`.

emersion (approved):
LGTM regardless

kmer2016 (approved):
LGTM

PR #8881, opened 2024-09-16
front: space-curve-slope: rotate the unit on the right

### LINKED ISSUE #8095: front: rotate the unit m on the right on the space curves slopes chart
### What happened?

image

Rotate the unit on the right

No unit not rotated

Look at the screen

_No response_

_No response_

.

clarani:
closes #8095 

Before:
image

After:
image

### DIFF
--- a/front/src/modules/simulationResult/components/SpaceCurvesSlopes/utils.ts
+++ b/front/src/modules/simulationResult/components/SpaceCurvesSlopes/utils.ts
@@ -35,9 +35,9 @@ export const drawAxisTitle = (chart: Chart, slopes?: SpaceCurvesSlopesData['slop
       .append('text')
       .attr('class', 'axis-unit')
       .attr('text-anchor', 'end')
-      .attr('transform', 'rotate(0)')
-      .attr('x', chart.width - 10)
-      .attr('y', 30)
+      .attr('transform', 'rotate(-90)')
+      .attr('x', -30)
+      .attr('y', chart.width - 10)
       .text('m');
   }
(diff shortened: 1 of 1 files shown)


flomonster:
Since we already have slopes and curves information in the speed space chart. Should we not remove this component?

kmer2016:
Is it normal to have two units on the right? The first at the top seems to be correctly positioned but the second at the bottom has not been rotated.

SharglutDev:
> Since we already have slopes and curves information in the speed space chart. Should we not remove this component?

I think it had been decided to keep it because of the altitude not present in the speed space chart... (But @thibautsailly recently confirmed that it should be removed with the new scenario results UI)

flomonster:
> > Since we already have slopes and curves information in the speed space chart. Should we not remove this component?
> 
> I think it had been decided to keep it because of the altitude not present in the speed space chart... (But @thibautsailly recently confirmed that it should be removed with the new scenario results UI)

I agree with @thibautsailly. It's not a big loss.

clarani:
> Is it normal to have two units on the right? The first at the top seems to be correctly positioned but the second at the bottom has not been rotated.

The second one is the unit for the x axis ;) 

> I agree with @thibautsailly. It's not a big loss.

Yes it will be removed soon, but we can still fix it, since it is still in the interface for now and the fix is quite small ;)

Akctarus (approved):
Tested, LGTM !

kmer2016 (approved):
LGTM

PR #8884, opened 2024-09-16
front: enable @typescript-eslint/no-unsafe-return

emersion:
Part of #8772

Disallow returning `any` values, to avoid having these spread to other parts of the codebase due to type inference.

See https://typescript-eslint.io/rules/no-unsafe-return/

### DIFF
--- a/front/src/applications/editor/data/utils.ts
+++ b/front/src/applications/editor/data/utils.ts
@@ -134,6 +134,7 @@ export function nestEntity(entity: EditorEntity, type: EditoastType): EditorEnti
 
       if (isLast) props[k] = oldProperties[key];
       else props[k] = props[k] || {};
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
       return props[k];
     }, newProperties);
   });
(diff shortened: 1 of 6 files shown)


front/src/common/Map/Layers/GeoJSONs.tsx
```diff
@@ -418,7 +418,7 @@ interface EditorSourceProps {
 export const EditorSource = ({ id, data, layers, layerOrder }: EditorSourceProps) => {
   const dataFingerPrint =
     data.type === 'FeatureCollection'
-      ? data.features.map((f) => f.properties?.id).concat()
+      ? data.features.map((f) => f.properties?.id as unknown).concat()
```
emersion, line 421: Here `f.properties` is `any`. We don't actually care about the real type so we can just use `unknown`.

front/src/applications/editor/data/utils.ts
```diff
@@ -134,6 +134,7 @@ export function nestEntity(entity: EditorEntity, type: EditoastType): EditorEnti
 
       if (isLast) props[k] = oldProperties[key];
       else props[k] = props[k] || {};
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
       return props[k];
```
emersion, line 138: Here we're dealing with a `Record` with `any` values, and we need to pass-through `any` values when merging.

clarani (approved):
LGTM :white_check_mark:

anisometropie (approved):

PR #8886, opened 2024-09-16
front: imported trains are loaded automatically without refresh

### LINKED ISSUE #8679: Scenario page must be refreshed after importing trains
### What happened?

When importing trains from open data or files the train list/space time chart is not updated.
A refresh is required to see the new imported trains.

We should not have to refresh the page to see the imported trains.

1. Create a scenario
2. Import trains from open data
3. The train list is still empty 🐛 
4. Refresh the page (the trains should be visible)

Local

Firefox

36d375899a49e561e517a0ce3f199790408cd894

clarani:
closes #8679

Once the trains are imported, the new trainSchedules need to be refetched manually with the `upsertTrainSchedules` function


SharglutDev (approved):
Lgtm and tested, good job !

Math-R (approved):
nice job :) lgtm

PR #8895, opened 2024-09-17
tests: remove legacy v1 test data

eckter:
The files were still there but not used in the test suite. I've also removed the legacy field from the remaining jsons, and the filter in the test file collection.


shenriotpro (approved):
Thanks

PR #8896, opened 2024-09-17
tests: update fuzzer

eckter:
There were a couple small elements that weren't up to date with some api changes:

* `id` -> `timetable_id`
* `None` isn't a valid margin option anymore


shenriotpro (approved):
Thanks

PR #8897, opened 2024-09-17
editoast: make intersection as a type

woshilapin:
Might help https://github.com/OpenRailAssociation/osrd/pull/8794#discussion_r1762606010. And in general, explicit typing is not necessarily bad, just more verbose.

### DIFF
--- a/editoast/src/views/path/projection.rs
+++ b/editoast/src/views/path/projection.rs
@@ -5,6 +5,10 @@ use std::collections::HashMap;
 
 use super::TrackRange;
 
+editoast_common::schemas! {
+    Intersection,
+}
+
 /// This object is useful to:
 /// - Get the position in the path given a location (track section and offset).
 /// - Get the location (track section and offset) given a position in a path.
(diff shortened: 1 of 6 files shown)


emersion:
FWIW I think it's completely fine to use tuples in internal editoast APIs. I don't think it's a big deal if we need to convert them before sending them over HTTP.

woshilapin:
> FWIW I think it's completely fine to use tuples in internal editoast APIs. I don't think it's a big deal if we need to convert them before sending them over HTTP.

Well, to me, `(u64, u64)` doesn't give a clue about what is each integer. But a `Intersection { start: u64, end: u64 }` does give a hint about what you can expect, and as code, not as comment (so always up-to-date). So I do believe it's often an improvement.

leovalais (approved):
Nice improvement:)

flomonster (approved):
LGTM, thanks for this PR

editoast/src/views/path/projection.rs
```diff
@@ -206,10 +206,32 @@ impl<'a> PathProjection<'a> {
     }
 }
 
+#[cfg_attr(test, derive(Debug, PartialEq))]
+pub struct Intersection {
```
anisometropie, line 215: If this is meant to be used specifically here, to represent the intersection of a track range and a path perhaps add some documentation:

- specify that reference to a path
- the 2 bounds _start_ and _end_  are distances relatively to the beginning of the path

woshilapin: Does something like [08a956c](https://github.com/OpenRailAssociation/osrd/pull/8897/commits/08a956cf2333c3f5caee8372ba9d9a0a82081782) answer your comment?

I didn't document it because it is not (yet?) exposed on the API. But yet, the entire idea of this PR has been for #8794, and also, even if not exposed in the API, still good to document, especially since "from the beginning of the path" is quite important and not necessarily obvious.

anisometropie: thanks, looking good

anisometropie (approved):

PR #8900, opened 2024-09-17
front: use slice for simulation store

### LINKED ISSUE #8412: Refacto simulation store with slice
### Description and goal

Currently, the simulation store (reducers/osrdsimulation) uses a custom version.

It should be harmonized with the osrdConfCommon store that uses rtk `slice`

### Acceptance criteria

- [ ] store simulation now uses `slice` from rtk

clarani:
closes #8412 

See commits


Akctarus (approved):
Tested in operational studies and stdcm, works well

kmer2016 (commented):
Thanks so much for this PR, it’s great to finally tackle this debt! 

I was wondering if the `chart` field in the simulation slice is still necessary. There doesn’t seem to be a selector for it, so maybe this could be a good moment to consider removing it? Also, if possible, adding tests for the new slice would be awesome.

RomainValls (approved):
LGTM, tested ✅

clarani:
@kmer2016 I considered removing the chart props too, but actually it is used in dead code in the Warped map which is quite complex and which we might reuse very soon, so I chose to keep it for now.

What do you think ? Can we keep it, even if it's not very clean, in case we need it soon ?

kmer2016:
> I considered removing the chart props too, but actually it is used in dead code in the Warped map which is quite complex and which we might reuse very soon, so I chose to keep it for now.
> 
> What do you think ? Can we keep it, even if it's not very clean, in case we need it soon ?

Oh ok, in that case, we can keep it.

kmer2016 (approved):

PR #8909, opened 2024-09-18
front: use the new map style

### LINKED ISSUE #8908: front: use the new map style

End users

Use the new style of the map, designed by Thibaut

It will harmonize the app UI

Thibaut

clarani:
closes #8908 

The new mapStyle was provided by Thibaut

image

### DIFF
--- a/front/src/assets/mapstyles/OSMStyle.json
+++ b/front/src/assets/mapstyles/OSMStyle.json
@@ -1,1676 +1,828 @@
 [
-  {
-    "id": "landuse-cemetery",
-    "type": "fill",
-    "source": "openmaptiles",
-    "source-layer": "landuse",
-    "filter": [
-      "all",
-      [
-        "==",
-        "$type",
-        "Polygon"
-      ],
-      [
-        "==",
-        "class",
-        "cemetery"
-      ]
-    ],
-    "layout": {
-      "visibility": "visible"
+    {
+      "id": "background",
+      "type": "background",
+      "paint": {"background-color": "rgba(246, 245, 241, 1)"}
     },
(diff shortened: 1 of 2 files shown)


thibautsailly:
Would it be possible to see the map for the Bretagne area, and for the whole of France? It was not possible to view rail tracks at this scale in the tool I used to edit the theme (Maputnik). Thank you!

clarani:
Yes of course: 
image

Caracol3:
I don't know if it's intentional, but now, at initialization and at least up to zoom level 2, we have a less detailed map than before with the same display settings.
At first glance, I find it less visually appealing without an additional level of detail. It seems a bit empty. But this is subjective. Maybe the decision was made to focus on displaying tracks rather than forests (which would be a logical choice 😆).

on dev at initialization:
image

On your branch same zoom level:
image(1)

thibautsailly:
It was absolutely intentional to remove details from the map so it's more focused on the infrastructure. That doesn't mean we can't work on the less appealing aspect (the green areas are not forests but national parks). We should talk about it together in front of the map when we can.
It seems that the background color isn't the same, here is a screenshot of the theme in Maputnik:
image
background is `rgb(246, 245, 241)`

clarani:
I changed the color of the background:
image

SarahBellaha (approved):

SarahBellaha (approved):
Lgtm :)

Caracol3 (approved):
lgtm

SharglutDev (approved):
Lgtm and tested ! Really nice :) But we should wait after the demo to merge I think (checked with @flomonster)

PR #8912, opened 2024-09-18
front: update nge label

Uriel-Sautron:
end part trainrun label of https://github.com/OpenRailAssociation/osrd/issues/8397

### DIFF
--- a/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
@@ -11,20 +11,20 @@ import { formatToIsoDate } from 'utils/date';
 import { calculateTimeDifferenceInSeconds, formatDurationAsISO8601 } from 'utils/timeManipulation';
 
 import nodeStore from './nodeStore';
-import { DEFAULT_TRAINRUN_FREQUENCY } from './osrdToNge';
-import {
-  type NetzgrafikDto,
-  type NGEEvent,
-  type TrainrunSection,
-  type Node,
-  type TimeLock,
-  type Trainrun,
-  type Label,
+import { DEFAULT_TRAINRUN_FREQUENCIES, DEFAULT_TRAINRUN_FREQUENCY } from './osrdToNge';
+import type {
+  NetzgrafikDto,
+  NGEEvent,
+  TrainrunSectionDto,
+  NodeDto,
+  TimeLockDto,
+  TrainrunDto,
+  LabelDto,
 } from '../NGE/types';
 
 const createdTrainrun = new Map<number, number>();
 
-const getNodeById = (nodes: Node[], nodeId: number | string) =>
+const getNodeById = (nodes: NodeDto[], nodeId: number | string) =>
   nodes.find((node) => node.id === nodeId);
 
 const getTrainrunSectionsByTrainrunId = (netzgrafikDto: NetzgrafikDto, trainrunId: number) => {
@@ -53,8 +53,8 @@ const getTrainrunSectionsByTrainrunId = (netzgrafikDto: NetzgrafikDto, trainrunI
   // Build a map of sections keyed by their previous section's target port ID.
   // Find the departure section: it's the one without a transition for its
   // source port.
-  let departureSection: TrainrunSection | undefined;
-  const sectionsByPrevTargetPortId = new Map<number, TrainrunSection>();
+  let departureSection: TrainrunSectionDto | undefined;
+  const sectionsByPrevTargetPortId = new Map<number, TrainrunSectionDto>();
   // eslint-disable-next-line no-restricted-syntax
   for (const section of sections) {
     const sourceNode = getNodeById(netzgrafikDto.nodes, section.sourceNodeId)!;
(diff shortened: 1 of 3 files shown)


louisgreiner:
wait for https://github.com/OpenRailAssociation/osrd/pull/8946 to be merged (and rebase this branch) to be able to test this feature

Yohh (commented):
`ngeToOsrd.ts` and `osrdToNge.ts` are really, really, REALLY huge (I missed a really), can't they be separated in smaller ones in dedicated folders?

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
```diff
@@ -246,6 +246,51 @@ const createTrainSchedulePayload = async ({
...
+          dispatch,
+          labels,
+          oldStartDate: startDate,
+          trainSchedule,
+        })),
+        // Reset margins because they contain references to path items
+        margins: undefined,
+      },
+    })
+  ).unwrap();
```
Yohh:
```suggestion
const trainSchedulePayload = await createTrainSchedulePayload({
    trainrunSections: trainrunSectionsByTrainrunId,
    nodes,
    trainrun,
    infraId,
    dispatch,
    labels,
    oldStartDate: startDate,
    trainSchedule,
  });
  const newTrainSchedule = await dispatch(
    osrdEditoastApi.endpoints.putTrainScheduleById.initiate({
      id: trainrunIdToUpdate,
      trainScheduleForm: {
        ...trainSchedule,
        ...trainSchedulePayload,
        // Reset margins because they contain references to path items
        margins: undefined,
      },
    })
  ).unwrap();
```

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
```diff
@@ -376,6 +396,44 @@ const handleNodeOperation = ({
...
+          infraId,
+          addUpsertedTrainSchedules,
+        });
+      });
+      break;
+    }
+    default: {
+      break;
+    }
+  }
```
Yohh, line 438: an `if` seems enough, do we wait for more cases?

Uriel-Sautron: For maybe other cases in the future and harmonized with other oparation handle.

louisgreiner:
@Yohh, I think @emersion already talked about this topic on a previous issue, but i can't find it. The reason is that currently, we use `const` pretty much everywhere for our variables, which needs to be to in the same file (but I miss the details)

If you have some suggestions to refactor, we can start think about it

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
Uriel-Sautron: Good suggestion thanks

Uriel-Sautron: Done

emersion:
> ngeToOsrd.ts and osrdToNge.ts are really, really, REALLY huge (I missed a really), can't they be separated in smaller ones in dedicated folders?

500 lines of code isn't huge in my opinion (my rule of thumb is < 1k). In general, having to open and juggle between a myriad of files when working on something isn't super great.

(I don't really mind either way.)

> I think @emersion already talked about this topic on a previous issue, but i can't find it. The reason is that currently, we use const pretty much everywhere for our variables, which needs to be to in the same file (but I miss the details)

`osrdToNge` uses helper functions defined inside functions to avoid boilerplate and long argument lists.

Yohh:
> @Yohh, I think @emersion already talked about this topic on a previous issue, but i can't find it. The reason is that currently, we use `const` pretty much everywhere for our variables, which needs to be to in the same file (but I miss the details)
> 
> If you have some suggestions to refactor, we can start think about it

yes that's what @Uriel-Sautron told earlier, I'll think about it

Yohh:
> > ngeToOsrd.ts and osrdToNge.ts are really, really, REALLY huge (I missed a really), can't they be separated in smaller ones in dedicated folders?
> 
> 500 lines of code isn't huge in my opinion (my rule of thumb is < 1k). In general, having to open and juggle between a myriad of files when working on something isn't super great.
> 
> (I don't really mind either way.)
> 
> > I think @emersion already talked about this topic on a previous issue, but i can't find it. The reason is that currently, we use const pretty much everywhere for our variables, which needs to be to in the same file (but I miss the details)
> 
> `osrdToNge` uses helper functions defined inside functions to avoid boilerplate and long argument lists.

ok fine

SarahBellaha (approved):
Tested, LGTM :)

louisgreiner (changes requested):
It is working very well (tested), I would just like if you could rename theses types (from `NGE/types.ts`) and add "Dto" at the end of their type names :
`NodeDto, TrainrunSectionDto, TrainrunDto, ResourceDto, PortDto, TransitionDto, LabelGroupDto, LabelDto, TimeLockDto`

The reason is that we actually use the same types than NGE ones:
- https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/src/app/data-structures/technical.data.structures.ts
- https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/src/app/data-structures/business.data.structures.ts

And especially because in OSRD, we interact with [`netzgrafikDto`](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/b76ed6ab75283b0da21d80cd211524b10f81977b/src/app/data-structures/business.data.structures.ts#L309) in particular.

So, to make sure we talk about the same things, the names should be consistent.

louisgreiner (approved):
Good work, thanks (tested)

Math-R (approved):
lgtm nice work !

PR #8913, opened 2024-09-18
front: update-via-selection-for-010-test

Maymanaf:
The test `010-op-route-tab.spec.ts:67:7 › Verifying that all elements in the route tab are loaded correctly › should correctly add waypoints in a route for an operational study` was failing when waypoints were added rapidly, so I added a timeout and a verification step to ensure that each waypoint is added correctly


clarani (approved):
LGTM :white_check_mark: 

Thank you !

Uriel-Sautron (approved):
LGTM !!! Thank you

PR #8914, opened 2024-09-18
Script used to run the OSRD instance from a given PR in parallel with the development instance (for review purposes).

### LINKED ISSUE #8709: Create a PR local testing script
Create a script that:
  - Pull images from GitHub
  - Setup the volume with the given backup
  - Run containers (with unused ports)
  - Everything should be clean after use

Example usage:

```sh
$ ./scripts/test-pr.sh 4242 PATH_BACKUP_OSRD
```

EthanPERRUZZA:
This script enables the deployment of a dedicated app instance using the following inputs:
* **PR Number**: Specify the pull request number for the instance you wish to run.
* **Infrastructure Backup**: Provide a backup of the infrastructure to load into the app.

Closes [#8709](https://github.com/OpenRailAssociation/osrd/issues/8709)

## Key Benefits:
**Isolated Environment**: The new instance operates independently of the development environment, allowing both to run in parallel without interference.
**Faster Reviews**: The script pulls Docker images generated by CI, eliminating the need to rebuild everything from scratch, which accelerates the review process.

### DIFF
--- a/docker/osrdyne-pr-tests.yml
+++ b/docker/osrdyne-pr-tests.yml
@@ -1,14 +1,16 @@
 worker_driver:
   type: "DockerDriver"
-  worker_image: "ghcr.io/openrailassociation/osrd-edge/osrd-core:dev"
-  network: "osrd_default"
+  network: "osrd_pr_test"
   container_prefix: "osrd"
   container_command: ["sh", "-c", "exec java $JAVA_OPTS -ea -jar /app/osrd_core.jar worker"]
   default_env:
-      - "CORE_EDITOAST_URL=http://osrd-editoast"
+      - "CORE_EDITOAST_URL=http://osrd-editoast-pr-tests:8091"
       - "JAVA_TOOL_OPTIONS=-javaagent:/app/opentelemetry-javaagent.jar"
       - "CORE_MONITOR_TYPE=opentelemetry"
       - "OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=grpc"
-      - "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://jaeger:4317"
+      - "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://osrd-jaeger-pr-tests:4319"
       - "OTEL_METRICS_EXPORTER=none"
       - "OTEL_LOGS_EXPORTER=none"
+amqp_uri: "amqp://osrd:password@osrd-rabbitmq-pr-tests:5673/%2f"
+management_port: 15673
+api_address: "0.0.0.0:4243"
--- a/scripts/load-backup.sh
+++ b/scripts/load-backup.sh
@@ -36,17 +36,25 @@ else
   exit 2
 fi
 
+# These variables are necessary to load the infra on the correct instance (the pr-infra or the dev one)
+OSRD_POSTGRES="osrd-postgres"
+OSRD_POSTGRES_PORT=5432
+if [ "$PR_TEST" -eq 1 ]; then
+  OSRD_POSTGRES="osrd-postgres-pr-tests"
+  OSRD_POSTGRES_PORT=5433
+fi
+
 $(dirname "$0")/cleanup-db.sh # Cleanup and init db (no migration)
 
 # Copy needed files to the container
-docker cp "$BACKUP_PATH" osrd-postgres:tmp/backup-osrd
+docker cp "$BACKUP_PATH" "$OSRD_POSTGRES:tmp/backup-osrd"
 
 # restoring the backend can partialy fail, and that's sometimes ok
 echo "Restore backup..."
-docker exec osrd-postgres pg_restore --if-exists -c -d osrd -x //tmp/backup-osrd > /dev/null
+docker exec "$OSRD_POSTGRES" pg_restore --if-exists -p "$OSRD_POSTGRES_PORT" -c -d osrd -x //tmp/backup-osrd > /dev/null
 
 # analyze db for performances
 echo "Analyze for performances..."
-docker exec osrd-postgres psql -d osrd -c 'ANALYZE;' > /dev/null
+docker exec "$OSRD_POSTGRES" psql -p "$OSRD_POSTGRES_PORT" -d osrd -c 'ANALYZE;' > /dev/null
 
 echo "Load backup done!"
--- a/scripts/pr-infra-compose.sh
+++ b/scripts/pr-infra-compose.sh
@@ -1,0 +1,68 @@
+#!/bin/sh
+
+# For more information on that script, you can have a look at the documentation
+# following the link below :
+# https://osrd.fr/en/docs/guides/contribute/code-review/#script-for-testing-a-pr
+
+
+set -e
+
+# Usage
+
+if  {
+        [ "$#" -eq 3 ] &&
+        [ "$2" != "up-and-load-backup" ]
+    } ||
+    {
+        [ "$#" -eq 2 ] &&
+        {   [ "$2" != "up" ] &&
+            [ "$2" != "down" ] &&
+            [ "$2" != "down-and-clean" ]
+        }
+    }; then
+
+    echo "Usage: $0 pr-number (up | up-and-load-backup [osrd.backup] | down | down-and-clean )" >&2
+    exit 1
+
+fi
+
+# Change ports in needed files
+export PR_NB="pr-$1"
+
+if [ "$2" = "up" ] || [ "$2" = "up-and-load-backup" ]; then
+
+    if [ "$2" = "up-and-load-backup" ]; then
+
+        docker compose \
+            -p "osrd-pr-tests" \
+            -f "docker/docker-compose.pr-tests.yml" \
+            up -d postgres redis
+
+        # Load backup
+        export PR_TEST=1
+        ./scripts/load-backup.sh "$3"
+    fi
+
+    # Load the docker instance
+    docker compose \
+        -p "osrd-pr-tests" \
+        -f "docker/docker-compose.pr-tests.yml" \
+        up -d
+
+elif [ "$2" = "down" ]; then
+
+    # Shutdown the docker instance
+    docker compose \
+        -p "osrd-pr-tests" \
+        -f "docker/docker-compose.pr-tests.yml" \
+        down
+
+else
+
+    # Shutdown and clean the docker instance
+    docker compose \
+        -p "osrd-pr-tests" \
+        -f "docker/docker-compose.pr-tests.yml" \
+        down -v
+
+fi
--- a/docker/docker-compose.host.yml
+++ b/docker/docker-compose.host.yml
@@ -44,6 +44,7 @@ services:
     network_mode: host
     environment:
       OSRDYNE__WORKER_DRIVER__HOST_NETWORKING: "true"
+      OSRDYNE__WORKER_DRIVER__NETWORK: ""
       OSRDYNE__WORKER_DRIVER__DEFAULT_ENV: > 
         [
         "CORE_EDITOAST_URL=http://127.0.0.1:8090",
--- a/docker/docker-compose.pr-tests.yml
+++ b/docker/docker-compose.pr-tests.yml
@@ -1,0 +1,169 @@
+## This Docker Compose file is used by the pr-infra-compose.sh script.
+# More explanation on its purpose can be found in the script.
+#
+# This Docker Compose file is very similar to the one used for the development
+# environment, except that no services are built (all images are pulled from DockerHub),
+# and all ports have been adjusted (by adding +1 or +2 if the +1 port was already in use)
+# so that both instances can run concurrently on the same computer.
+
+volumes:
+  psql_data:
+  redis_data:
+  rabbitmq_data:
+
+networks:
+  pr-tests:
+    name: osrd_pr_test
+
+services:
+  postgres:
+    image: ${OSRD_POSTGIS_IMAGE:-postgis/postgis:16-3.4-alpine}
+    container_name: osrd-postgres-pr-tests
+    user: postgres
+    restart: unless-stopped
+    ports: ["5433:5433"]
+    environment:
+      POSTGRES_PASSWORD: "password"
+    command: -c port=5433
+    volumes:
+      - "psql_data:/var/lib/postgresql/data"
+      - "../docker/init_db.sql:/docker-entrypoint-initdb.d/init.sql"
+    healthcheck:
+      test: ["CMD", "pg_isready", "-d", "postgres://osrd:password@postgres:5433/osrd"]
+      start_period: 4s
+      interval: 5s
+    networks:
+      - pr-tests
+
+  redis:
+    image: valkey/valkey:alpine
+    container_name: osrd-redis-pr-tests
+    restart: unless-stopped
+    ports: ["6380:6380"]
+    volumes:
+      - "redis_data:/data"
+      - "../docker/valkey.conf:/etc/valkey.conf"
+    command: "valkey-server /etc/valkey.conf --port 6380 --save 30 1 --loglevel warning"
+    healthcheck:
+      test: ["CMD", "valkey-cli", "-p" , "6380", "ping"]
+      start_period: 4s
+      interval: 5s
+    networks:
+      - pr-tests
+
+  rabbitmq:
+    image: rabbitmq:3-management
+    container_name: osrd-rabbitmq-pr-tests
+    ports:
+      - "5673:5673"
+      - "15673:15673"
+    environment:
+      RABBITMQ_DEFAULT_USER: osrd
+      RABBITMQ_DEFAULT_PASS: password
+      RABBITMQ_DIST_PORT: 25673
+      RABBITMQ_CONFIG_FILE: "/etc/rabbitmq/rabbitmq.conf"
+    volumes:
+      - rabbitmq_data:/var/lib/rabbitmq
+      - ../docker/rabbitmq-pr-tests.conf:/etc/rabbitmq/rabbitmq.conf
+    healthcheck:
+      test: ["CMD", "rabbitmqctl", "status"]
+      interval: 15s
+      timeout: 30s
+      retries: 3
+    networks:
+      - pr-tests
+
+  core:
+    image: ghcr.io/openrailassociation/osrd-edge/osrd-core:${PR_NB}
+    container_name: osrd-core-dummy-pr-tests
+    restart: "no"
+    command: "true"
+    networks:
+      - pr-tests
+
+  editoast:
+    image: ghcr.io/openrailassociation/osrd-edge/osrd-editoast:${PR_NB}
+    container_name: osrd-editoast-pr-tests
+    depends_on:
+      postgres: {condition: service_healthy}
+      redis: {condition: service_healthy}
+      rabbitmq: {condition: service_healthy}
+    restart: unless-stopped
+    ports: ["8091:8091"]
+    environment:
+      EDITOAST_PORT: 8091
+      ROOT_URL: "http://localhost:8091"
+      REDIS_URL: "redis://redis:6380"
+      DATABASE_URL: "postgres://osrd:password@postgres:5433/osrd"
+      TELEMETRY_KIND: "opentelemetry"
+      TELEMETRY_ENDPOINT: "http://jaeger:4319"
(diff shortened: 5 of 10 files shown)


scripts/load-backup.sh
```diff
@@ -9,8 +9,9 @@
 set -e
 
 # Check arguments
-if [ "$#" -ne 1 ]; then
-  echo "usage: $0 osrd.backup"
+if [ "$#" -ne 1 ] &&
+   { [ "$#" -ne 2 ] || [ "$2" != "--pr-tests" ]; }; then
+  echo "usage: $0 osrd.backup [--pr-tests]"
   exit 1
 fi
```
Khoyo: You should set a variable here

scripts/load-backup.sh
```diff
@@ -38,15 +39,23 @@ fi
 
 $(dirname "$0")/cleanup-db.sh # Cleanup and init db (no migration)
 
+# These variables are necessary to load the infra on the correct instance (the pr-infra or the dev one)
+OSRD_POSTGRES="osrd-postgres"
+OSRD_POSTGRES_PORT=5432
+if [ "$#" -eq 2 ]; then
+  OSRD_POSTGRES="osrd-postgres-pr-tests"
+  OSRD_POSTGRES_PORT=5433
+fi
```
Khoyo: And use it here.

docker/rabbitmq.conf
```diff
@@ -0,0 +1,4 @@
+listeners.tcp.port = 5673
+management.tcp.port = 15673
+distribution.listener.port_range.min = 25673
+distribution.listener.port_range.max = 25673
```
Khoyo: If this is only used for pr-tests, it should be reflected in the filename.

scripts/pr-infra-compose.sh
```diff
@@ -0,0 +1,45 @@
...
+
+if [ "$#" -ne 3 ] ||
+    { [ "$3" != "up" ] && [ "$3" != "down" ]; }; then
+    echo "Usage: $0 pr-number osrd.backup (up | down)" >&2
+    exit 1
+fi
+
+# Change ports in needed files
+sed -i "s/{{pr-nb}}/pr-$1/g" ./docker/docker-compose.pr-tests.yml
+sed -i "s/{{pr-nb}}/pr-$1/g" ./docker/osrdyne-pr-tests.yml
```
Khoyo: Can we do this via environment variables? It would be much better.

**If not**, we shouldn't modify these files in place:

- It creates unwanted changes in the working tree
- Multiple invocation of `pr-infra-compose.sh` will fail once one replacement is done (unless people roll it back, but isn't the goal to make PR testing as seamless as possible?)

We should instead move `docker/docker-compose.pr-tests.yml` to  `docker/docker-compose.pr-tests.template.yml`, add `/docker/docker-compose.pr-tests.yml` to the `.gitignore`, and have sed generate it. (Ditto for `docker/osrdyne-pr-tests.yml`)

ghost: Env variables can be used on this.

ghost (changes requested):
We should only use env variable and not sed a template.

Erashin (commented):
I do believe you may have failed your rebase :D good luck!

scripts/pr-infra-compose.sh
EthanPERRUZZA: I could switch to env vars for everything but not for the osrdyne.yml config file unfortunately

scripts/pr-infra-compose.sh
```diff
@@ -0,0 +1,47 @@
...
+
+if [ "$3" = "up" ]; then
+
+    # Load the docker instance
+    sed -i "s/{{pr-nb}}/${PR_NB}/g" ./docker/osrdyne-pr-tests.yml
+    docker compose \
+        -p "osrd-pr-tests" \
+        -f "docker/docker-compose.pr-tests.yml" \
+        up -d
+    git restore ./docker/osrdyne-pr-tests.yml
```
woshilapin: Would it be a interesting modification to modify `osrdyne` so it does evaluate environment variables before deserializing the file? This way, we could avoid this weird dance of `sed` and `git restore`.

woshilapin (commented):
I'm generally not in favor on stacking up layers of complexity on top of layers of already complex setup as it brings difficulty to maintain and debug any time a problem arise. That said, if we are to follow-up with this, I made one small comment (but might be not trivial to address) and also:
- please edit the title of the PR so it's not the default branch name
- please provide a description of the PR so we know what's it's about (now but also in the future)

scripts/pr-infra-compose.sh
Khoyo: I'm pretty sure it does that already 😆 

https://github.com/OpenRailAssociation/osrd/blob/748e7cbf034fa11330cddb5ee01f264eac55b04d/osrdyne/src/config.rs#L64

EthanPERRUZZA:
> I'm generally not in favor on stacking up layers of complexity on top of layers of already complex setup as it brings difficulty to maintain and debug any time a problem arise. That said, if we are to follow-up with this, I made one small comment (but might be not trivial to address) and also:
> 
> * please edit the title of the PR so it's not the default branch name
> * please provide a description of the PR so we know what's it's about (now but also in the future)

Hey @woshilapin ,

I added a description to the PR as well as in the script files, so we don't have to search for the PR later.

I don't think this script will be hard to maintain since it pulls the Docker images from the CI. The most time-consuming part was writing it, as we had to modify the ports used by all the services.

Moreover, I believe it adds great value to the development experience, as it will allow for faster and easier reviews.

scripts/pr-infra-compose.sh
EthanPERRUZZA: I didn’t know I could combine the two methods. I’ve updated the script accordingly. ;)

Thanks!

scripts/pr-infra-compose.sh
EthanPERRUZZA: Thanks to Khoyo I was able to pass it using an env var too.

woshilapin:
> I don't think this script will be hard to maintain since it pulls the Docker images from the CI. The most time-consuming part was writing it, as we had to modify the ports used by all the services.

I don't only talk about maintaining the script. I also talk about all the other files that have been modified to allow this behavior. It brings complexity a bit more everywhere. And maintainability is not the only issue: to me, debuggability is too. Now I need to understand that everything is on another port, that I have 2 instances of everything running. Each taken independently might not be much, but the Docker stack in itself is already not trivial: when everything works, it's fine, but when it doesn't, the developers have to dive in that complexity.

Another different question: I'm a bit worried about the number of containers that we have to run with this proposition. I believe a simple OSRD stack is already a bunch of containers that can slow down computers. We almost spawn twice the number of containers with this proposition if I understand correctly.

ghost:
> > I don't think this script will be hard to maintain since it pulls the Docker images from the CI. The most time-consuming part was writing it, as we had to modify the ports used by all the services.
> 
> I don't only talk about maintaining the script. I also talk about all the other files that have been modified to allow this behavior. It brings complexity a bit more everywhere. And maintainability is not the only issue: to me, debuggability is too. Now I need to understand that everything is on another port, that I have 2 instances of everything running. Each taken independently might not be much, but the Docker stack in itself is already not trivial: when everything works, it's fine, but when it doesn't, the developers have to dive in that complexity.

I would tend to disagree on this point, could you pinpoint the complexity that is added for you that should be avoided ? Imho, the docker files are not too much altered, the code modified in Osrdyne and such is good to have since it permits more configuration (and will avoid issues with strange setups in the future).

> Another different question: I'm a bit worried about the number of containers that we have to run with this proposition. I believe a simple OSRD stack is already a bunch of containers that can slow down computers. We almost spawn twice the number of containers with this proposition if I understand correctly.

No, we do not. The goal here is to have a facilitator script that would help do the review of a given PR by pulling the CI-made container images. If someone's computer is too slow to handle the load of two OSRD at the mean time, they can shutdown their dev env.

The goal was having a separated workflow for testing a PR in order to avoid messing up with database (imagine a PR with migrations) or with other modifications.

All in all, having them separated just gives the option for the developer to run the PR built container as they see fit.

docker/osrdyne-pr-tests.yml
```diff
@@ -0,0 +1,16 @@
+worker_driver:
+  type: "DockerDriver"
+  network: "PR_Test_Network"
```
bougue-pe: for consistency:
```suggestion
  network: "pr_test_network"
```

scripts/pr-infra-compose.sh
```diff
@@ -0,0 +1,58 @@
...
+if [ "$#" -ne 3 ] ||
+    { [ "$3" != "up" ] && [ "$3" != "down" ]; }; then
+    echo "Usage: $0 pr-number osrd.backup (up | down)" >&2
+    exit 1
+fi
+
+# Change ports in needed files
+export PR_NB="pr-$1"
+
+if [ "$3" = "up" ]; then
```
bougue-pe: It would be nice to bring this script closer to `osrd-compose.sh` (and `docker compose`):
1. have `up` do the same thing everywhere, so rename this to `up-and-load-backup`?
2. really name the shutdown part like `down-and-clean`
3. maybe allow for a "classic" compose use for user willing to tweak their "2nd stack" a bit

Then, maybe mutualize with osrd-compose.sh and just allow for different project-name and file?

docker/osrdyne-pr-tests.yml
Khoyo:
```suggestion
  network: "osrd_pr_test"
```

scripts/pr-infra-compose.sh
EthanPERRUZZA: I added the first suggestions.

I'm not sure that we should combine the script with `osrd-compose.sh` as they don't have the same purpose, it would make the `osrd-compose.sh` a lot more complex for not so much added value (imo)
Or maybe I didn't understand your last suggestion ?

EthanPERRUZZA: Nevermind, I messed up with git, the change isn't on the upstream yet

EthanPERRUZZA: It is now ;)

scripts/pr-infra-compose.sh
```diff
@@ -0,0 +1,80 @@
...
+        -p "osrd-pr-tests" \
+        -f "docker/docker-compose.pr-tests.yml" \
+        up -d
+
+    if [ "$2" = "up-and-load-backup" ]; then
+
+        # Load backup
+        export PR_TEST=1
+        ./scripts/load-backup.sh "$3"
+        export PR_TEST=0
```
Khoyo: Why are you doing this? You could leave it set to 1

scripts/pr-infra-compose.sh
```diff
@@ -0,0 +1,80 @@
...
+# Change ports in needed files
+export PR_NB="pr-$1"
+
+if [ "$2" = "up" ] || [ "$2" = "up-and-load-backup" ]; then
+
+    # Load the docker instance
+    docker compose \
+        -p "osrd-pr-tests" \
+        -f "docker/docker-compose.pr-tests.yml" \
+        up -d
```
Khoyo: You should only do this up when doing load backup, and should should only up postgres

scripts/pr-infra-compose.sh
```diff
@@ -0,0 +1,80 @@
...
+
+        # Load backup
+        export PR_TEST=1
+        ./scripts/load-backup.sh "$3"
+        export PR_TEST=0
+
+        docker compose \
+            -p "osrd-pr-tests" \
+            -f "docker/docker-compose.pr-tests.yml" \
+            up  -d
```
Khoyo: This one should be done regardless of up/up-and-load-backup

Khoyo (commented):
Otherwise it starts to look good, except for the name clash of the osrdyne created worker containers

scripts/pr-infra-compose.sh
EthanPERRUZZA: my bad, I thought it was exported all the way up to the current shell

scripts/pr-infra-compose.sh
EthanPERRUZZA: I had to add redis for the loading script to work

scripts/pr-infra-compose.sh
ghost: > Then, maybe mutualize with osrd-compose.sh and just allow for different project-name and file?

I feel osrd-compose was more a hack to have the host mode accessible more easily (its name is kinda bad, I should not have named this script osrd-compose in the first place).

woshilapin:
> I would tend to disagree on this point, could you pinpoint the complexity that is added for you that should be avoided ? Imho, the docker files are not too much altered, the code modified in Osrdyne and such is good to have since it permits more configuration (and will avoid issues with strange setups in the future).

To me, I'm not sure what this PR brings compared to `TAG=pr-1234 scripts/osrd-compose.sh pull && TAG=pr-1234 scripts/osrd-compose.sh up --detach` that I can already do today. So because I can already do that, all additional setup is additional complexity, not really needed. It's additional code, it's also additional scripts that developers have to know and understand how to use. It's not **much more**, I agree, but yet, it's always more than adding nothing.
 
>> Another different question: I'm a bit worried about the number of containers that we have to run with this proposition. I believe a simple OSRD stack is already a bunch of containers that can slow down computers. We almost spawn twice the number of containers with this proposition if I understand correctly.
> 
> No, we do not. The goal here is to have a facilitator script that would help do the review of a given PR by pulling the CI-made container images. If someone's computer is too slow to handle the load of two OSRD at the mean time, they can shutdown their dev env.

Acknowledged, my worry disappeared.

> The goal was having a separated workflow for testing a PR in order to avoid messing up with database (imagine a PR with migrations) or with other modifications.

If this is the goal, why not addressing just this goal? How about only duplicating the DB instead of the entire stack? We could add a POSTGRES_PORT variable that would be used everywhere in the `docker-compose.yaml` (I didn't thoroughly think about the solution, so this one might be dumb, but yet, my main point is: we might have a lighter approach to the problem that is being solved here?)
 
> All in all, having them separated just gives the option for the developer to run the PR built container as they see fit.

I acknowledge one advantage: you would be able to actually compare live 2 different stacks. I never had such a use case, but I can more easily imagine that to be useful for `front` developers, comparing an interface.

scripts/pr-infra-compose.sh
bougue-pe: Thanks for the rename, the use should be less confusing that way 🙏 

@ElysaSrc : renaming (and changing doc if necessary) osrd-compose.sh to `host-compose.sh` may be more consistent indeed.

Finally, as discussed a bit yesterday, my idea was to "intercept" `up-and-load-backup` and `down-and-clean`, and in all other cases fallback to just `docker compose "$@"` (maybe not so easy to document and create a `--help`).
So enrich the docker compose options in a script, and then from this script offer host-compose and pr-infra-compose that would just provide appropriate `-p project` and `-f file`.
That way we would have one way of doing all and also allow for `up-and-load-backup`|`down-and-clean` on a "classic" or on a "host" setup.
This was my idea, but it's only a nice-to-have IMO, so feel free to let it aside.

scripts/pr-infra-compose.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/sh
+
+# This script enables the deployment of a dedicated app instance using the
```
bougue-pe: It would be nice to add a bit of doc for this.
I'd go for an update of the website https://osrd.fr/en/docs/guides/deploy/docker-compose ~(the localhost access part of components there should probably be updated too, mentioning that it's only available when using `docker/docker-compose.host.yml` file)~.

EthanPERRUZZA: Hey, I'll update the doc indeed, you can still access (nearly) everything from loccalhost, it's just that the port you need to use has been moved (using a +1 operation, or +2 on two cases when the +1 was already taken).
The only thing that is not accessible is the front as it is served directly by the gateway and not as a service on it's own.

EthanPERRUZZA: However, I would say that it's bet to talk about the script here https://osrd.fr/en/docs/guides/contribute/code-review/ as it is intended for code review usage only.
That what I'll do for now but we can always debate on this and move it later if you think it would be best otherwhere.

bougue-pe: Good idea to at least mention it in code-review part, yes.
As long as things are documented only once, it's fine for me (the place and the links are easier to maintain/add).

As for the update I mentioned, about being able to request any component: it was a mistake of mine, sorry, I mixed things and it's already good. I edit my first message.

EthanPERRUZZA: I made a PR on that, I'll ping you in it so you can have a look.
https://github.com/OpenRailAssociation/osrd-website/pull/241

bougue-pe (approved):
LGTM once all code comments are considered/resolved

ghost:
@woshilapin

> To me, I'm not sure what this PR brings compared to `TAG=pr-1234 scripts/osrd-compose.sh pull && TAG=pr-1234 scripts/osrd-compose.sh up --detach` that I can already do today. So because I can already do that, all additional setup is additional complexity, not really needed. It's additional code, it's also additional scripts that developers have to know and understand how to use. It's not **much more**, I agree, but yet, it's always more than adding nothing.

This is not the same thing. For instance, the gateway setup will not be the same (reversing to a front container vs having a gateway with the front embeded).
If you do this, you will reverse the local copy of your front and not the one compiled within the PR. You'll have to remember that "you must checkout the right branch" or your review will be broken, or even worse, miss an important bug.

> If this is the goal, why not addressing just this goal? How about only duplicating the DB instead of the entire stack? We could add a POSTGRES_PORT variable that would be used everywhere in the `docker-compose.yaml` (I didn't thoroughly think about the solution, so this one might be dumb, but yet, my main point is: we might have a lighter approach to the problem that is being solved here?)

For me, there are issues with this setup :
1. You do not have the correct network setup (you need to modify the front, the gateway).
2. You use some components that are in common between a PR and a work you could have locally, trigerring strange bugs (Redis, RabbitMQ).
3. You make the development stack more complex and less straightforward : having to add vars on every part of the Docker Compose (that is already a huge file, overriden with some helper scripts like `osrd-compose` for host mode etc...) will make it less manageable in all way.
4. I'm not even sure we could just use the override capabilities of docker compose to create such behavior, and we would need to add a script on top of it for testing PR, making it even less practical. 

Having a seperated yml file and script will duplicate code, yes. But I'd rather have a duplicated yml file than put the mental burden on the developer to diagnose a more complicated docker compose stack.

> I acknowledge one advantage: you would be able to actually compare live 2 different stacks. I never had such a use case, but I can more easily imagine that to be useful for `front` developers, comparing an interface.

To conclude, I'll say that having 2 seperated environements will be more convienient and avoid a wide range of strange bugs that are yet to be found and, even if we duplicate things and change some ports, will be easier to navigate on a daily basis than tweak once again the docker compose.

ghost (approved):
LGTM.

@woshilapin Do you think the points you're raising are blocking ? If so, we can organize a meeting a discuss more thoroughly the subject, otherwise I think we can ask for more review from the devs and merge it as it is.

docker/osrdyne-pr-tests.yml
```diff
@@ -0,0 +1,16 @@
+worker_driver:
+  type: "DockerDriver"
+  network: "osrd_pr_test"
+  container_prefix: "osrd"
```
ghost, line 4: Change the prefix so it doesn't conflict. Maybe `pr-osrd` ?

```suggestion
  container_prefix: "pr-osrd"
```

woshilapin:
> @woshilapin Do you think the points you're raising are blocking ? If so, we can organize a meeting a discuss more thoroughly the subject, otherwise I think we can ask for more review from the devs and merge it as it is.

Your [detailed answer](https://github.com/OpenRailAssociation/osrd/pull/8914#issuecomment-2390815002) helped to confirm some of what I understand. I don't entirely agree as basically, we hide behind one more layer how our developer's tools works? I do agree about putting the burden of understanding why the Docker stack might not work at some point (Redis, Postgres, wrong version of an image, etc.).

That said, I'm not alone on this project, and I spent the last few days discussing the topic around, with some of the concerned developers (backend as well as frontend): it seems that more people are interested in such a solution than not. So if it brings more to developers, who am I to block it (and also, it's been more than a week and nobody else did comment in the same direction than me).

So no, I won't be blocking. Thanks for the hard work @EthanPERRUZZA and the detailed answer @ElysaSrc.

PR #8917, opened 2024-09-18
front: convert frequency between osrd and nge

### LINKED ISSUE #8399: Converting NGE frequencies as labels in OSRD
### Description and goal

The goal of this feature is to be able to save the frequencies of trainruns in NGE in OSRD, in order to persist this information. This is persisted in the `labels` attribute, with a frequency prefix. 
Example : `frequency::60`

### Acceptance criteria

- [x] If the user sets a frequency on a trainrun in NGE, the frequency is saved in the labels in OSRD
- [x] If the user changes the frequency in NGE, the corresponding label is changed in OSRD (if several labels with frequency prefix were present in OSRD, NGE keeps only one)
- [x] If the user creates two labels in OSRD with the frequency prefix : NGE uses the one with the smallest frequency (can be discussed)
- [x] If the user deletes the labels with the frequency prefix in OSRD, NGE interprets this as a 1h frequency (can be discussed)
- [x] The list of possible frequencies presented in NGE when editing a train is the union of : 
  - [x] A default list (15, 20, 30, 60, 120) (this is the default list from NGE) EDIT: list of 30, 60, 120
  - [ ] The frequencies found in the trains of the scenario (for example if one create a label "frequency::45" in OSRD, the frequency 45 should be presented in NGE when editing a train

Uriel-Sautron:
close #8399 
close https://github.com/osrd-project/osrd-confidential/issues/663

## OSRD -> NGE
 - Tag a train in OSRD with `frequency::30` or `frequency::120`.
 - In the NGE tab, check that the trainrun frequency is 30 or 120.
 - Without a tag, the default frequency is 60.

## NGE -> OSRD
 - Change frequency on a run train.
 - 120 we will have the tag `frequency::120` and 30 the tag `frequency::30`.

/!\ At the moment we only manage 3 frequencies 30, 60, and 120

### DIFF
--- a/front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
@@ -1,3 +1,5 @@
+import { compact } from 'lodash';
+
 import { osrdEditoastApi } from 'common/api/osrdEditoastApi';
 import type {
   SearchResultItemOperationalPoint,
@@ -53,16 +55,38 @@ const DEFAULT_TRAINRUN_CATEGORY: TrainrunCategory = {
   sectionHeadway: 0,
 };
 
-const DEFAULT_TRAINRUN_FREQUENCY: TrainrunFrequency = {
-  id: 3, // In NGE, Trainrun.DEFAULT_TRAINRUN_FREQUENCY
-  order: 0,
-  frequency: 60,
-  offset: 0,
-  name: 'Default',
-  /** Short name, needs to be unique */
-  shortName: 'D',
-  linePatternRef: '60',
-};
+const DEFAULT_TRAINRUN_FREQUENCIES: TrainrunFrequency[] = [
+  {
+    id: 2,
+    order: 0,
+    frequency: 30,
+    offset: 0,
+    name: 'Half-hourly',
+    shortName: '30',
+    linePatternRef: '30',
+  },
+  {
+    id: 3, // default NGE frequency takes id 3
+    order: 1,
+    frequency: 60,
+    offset: 0,
+    name: 'Hourly',
+    /** Short name, needs to be unique */
+    shortName: '60',
+    linePatternRef: '60',
+  },
+  {
+    id: 4,
+    order: 2,
+    frequency: 120,
+    offset: 0,
+    name: 'Two-hourly',
+    shortName: '120',
+    linePatternRef: '120',
+  },
+];
+
+export const DEFAULT_TRAINRUN_FREQUENCY: TrainrunFrequency = DEFAULT_TRAINRUN_FREQUENCIES[1];
 
 const DEFAULT_TRAINRUN_TIME_CATEGORY: TrainrunTimeCategory = {
   id: 0, // In NGE, Trainrun.DEFAULT_TRAINRUN_TIME_CATEGORY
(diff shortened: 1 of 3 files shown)


louisgreiner:
Changes of scope :
- do not display any `frequency::xx` in NGE

Bug to fix:
- ensure OSRD labels are not overwritten by NGE ones, but rather `OSRD labels = unicity({OSRD labels} + {NGE labels} + {NGE trainrun frequency})`

louisgreiner:
The state for train schedule label is bugged if you go back and forth NGE and OSRD, after creating a label in OSRD

image

It feels it happens only when switching to macro mode

kmer2016:
image
I've noticed a bug where "non-frequency" tags get duplicated after modifying the frequency tag in NGE. Here's a step-by-step to reproduce:

- I create the tags toto, tata, and frequency::30 in OSRD.
- I switch to NGE and modify frequency::30 to frequency::120.
- When I return to OSRD, the tags toto and tata appear duplicated.

Uriel-Sautron:
> image J'ai remarqué un bug où les balises « non-fréquence » sont dupliquées après modification de la balise de fréquence dans NGE. Voici une procédure étape par étape pour reproduire ce bug :
> 
> * Je crée les balises toto, tata et frequency::30 dans OSRD.
> * Je passe en NGE et modifie la fréquence::30 en fréquence::120.
> * Lorsque je reviens sur OSRD, les tags toto et tata apparaissent en double.

Thank you, bug fix !!

front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
```diff
@@ -293,9 +317,20 @@ const importTimetable = async (
   // Create one NGE train run per OSRD train schedule
   let labelId = 0;
   const trainruns: Trainrun[] = trainSchedules.map((trainSchedule) => {
-    let formatedLabels: Label[] = [];
+    let formatedLabels: (Label | undefined)[] = [];
+    let trainrunFrequency: TrainrunFrequency | undefined;
     if (trainSchedule.labels) {
       formatedLabels = trainSchedule.labels.map((label) => {
+        if (label.includes('frequency')) {
```
kmer2016, line 325: The code which handle labels looks fine overall. However, I think that using `forEach` loop might be more appropriate than map in this case. Since not all labels need to be transformed into a new Object, especially the `frequency` labels, `forEach` would allow processing each label without returning a value, avoiding the insertion of `undefined` into the list of labels.

That said, it work as-is, so I'm fine with leaving it as is if preferred.

Uriel-Sautron: You're right, I'll change it.

Uriel-Sautron: Done

front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
```diff
@@ -333,7 +334,7 @@ const importTimetable = async (
         }
         const DTOLabel = DTOLabels.find((DTOlabel) => DTOlabel.label === label);
         if (DTOLabel) {
-          return DTOLabel;
+          formatedLabels.push(DTOLabel);
         }
```
kmer2016, line 348: Isn't there an `else` missing here? Before, there was a return, so the iteration ended in the if.

Uriel-Sautron: Done

kmer2016 (approved):
LGTM !

PR #8918, opened 2024-09-19
front: stdcm: fix arrival check

### LINKED ISSUE #8915: stdcm: unnecessary check on unscheduled point triggers an error
### What happened?

When you try to run a simulation with a scheduled destination, the date validity check is also performed on the starting point, which is not necessary.

The departure point doesn't need to be checked if it's the destination we want to schedule.

1. Go to STDCM (new interface)
2. Set arrival time for departure point to 00:00 then change to “as soon as possible”.
3. Set an arrival time for destination.
4. Run the request and see the error pop-up.

Recette (SNCF)

Firefox

6ea7e56

SarahBellaha:
Closes #8915


Uriel-Sautron (approved):
LGTM !! Tested

kmer2016 (approved):
LGTM !

PR #8920, opened 2024-09-19
front: bump osrd-ui to 0.0.42

kmer2016:


SharglutDev (approved):
Lgtm and tested (if its normal that the back always returns a date in June)

Capture d’écran 2024-09-19 à 10 37 37

SharglutDev:
Just one thing that could be clarified with @thibautsailly : it would be nice that the calendar modal closes when we selected a date (currently we need to click outside of the modal)

PR #8923, opened 2024-09-19
front: turn on @typescript-eslint/no-unsafe-call lint

emersion:
Part of https://github.com/OpenRailAssociation/osrd/pull/8772

See https://typescript-eslint.io/rules/no-unsafe-call/

### DIFF
--- a/front/src/applications/editor/components/LinearMetadata/FormComponent.tsx
+++ b/front/src/applications/editor/components/LinearMetadata/FormComponent.tsx
@@ -41,7 +41,9 @@ export type FormContext = {
   isCreation: boolean;
 };
 
-const IntervalEditorComponent = (props: FieldProps<GeoJsonProperties, RJSFSchema, FormContext>) => {
+const IntervalEditorComponent = (
+  props: FieldProps<LinearMetadataItem[], RJSFSchema, FormContext>
+) => {
   const { name, formContext, formData, schema, onChange, registry } = props;
   const { openModal, closeModal } = useModal();
   const { t } = useTranslation();
@@ -421,7 +423,7 @@ const IntervalEditorComponent = (props: FieldProps<GeoJsonProperties, RJSFSchema
   );
 };
 
-export const FormComponent = (props: FieldProps<GeoJsonProperties, RJSFSchema, FormContext>) => {
+export const FormComponent = (props: FieldProps<unknown, RJSFSchema, FormContext>) => {
   const { name, formContext, schema, registry } = props;
   const Fields = getDefaultRegistry().fields;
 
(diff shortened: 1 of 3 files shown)


front/src/applications/editor/components/LinearMetadata/FormComponent.tsx
```diff
@@ -421,7 +423,7 @@ const IntervalEditorComponent = (props: FieldProps<GeoJsonProperties, RJSFSchema
   );
 };
 
-export const FormComponent = (props: FieldProps<GeoJsonProperties, RJSFSchema, FormContext>) => {
+export const FormComponent = (props: FieldProps<unknown, RJSFSchema, FormContext>) => {
```
emersion, line 426: Couldn't find a better way to write this. Tried using two type variants (`IntervalFieldProps | ArrayFieldProps`) or generics, no dice. `FieldProps` is a reaaally complicated type so TypeScript is quite confused. At least with this version it's explicit that a cast is happening below with `as`, so the footguns are no longer hidden.

front/src/modules/simulationResult/components/ChartHelpers/enableInteractivity.tsx
```diff
@@ -282,8 +282,9 @@ export const enableInteractivity = <
     ])
     .wheelDelta(wheelDelta)
     // Allows evenements to be triggered on zoom and drag interactions for all graphs
-    .on('zoom', (event) => {
-      event.sourceEvent.preventDefault();
+    .on('zoom', (event: D3ZoomEvent<Element, ConsolidatedRouteAspect>) => {
+      const { sourceEvent }: { sourceEvent: Event } = event;
```
emersion, line 286: Would be nicer to write `const sourceEvent: Event = event.sourceEvent`, unfortunately an ESLint rule forces destructuring.

front/src/applications/editor/components/LinearMetadata/FormComponent.tsx
```diff
@@ -41,7 +41,9 @@ export type FormContext = {
   isCreation: boolean;
 };
 
-const IntervalEditorComponent = (props: FieldProps<GeoJsonProperties, RJSFSchema, FormContext>) => {
```
emersion, line 44: Sounds like the type I've written [here](https://github.com/OpenRailAssociation/osrd/pull/8884/files#diff-66b0673a768e8de8c885355417a57d1e5b21e151f544e934741984f87a9f589dR44) was incorrect, but I missed this because `GeoJsonProperties` is a record of `any` values so no compiler error was generated by `formData?.filter(notEmpty)` below.

front/src/modules/simulationResult/components/ChartHelpers/enableInteractivity.tsx
SharglutDev: Seems like the zoom event is broken (if we zoom on space curve slopes, the position synchro with train details is lost) with the but it was the case before your PR :)

SharglutDev (approved):
Lgtm and tested

Akctarus (approved):
LGTM

PR #8937, opened 2024-09-19
front: stdcm map adjustments

SarahBellaha:
Closes : [STDCM: Map adjustments](https://github.com/osrd-project/osrd-confidential/issues/673)

This PR also resolves an issue with updating the STDCM map viewport, which only refreshed once the simulation was launched.


front/src/modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/ItineraryLayer.tsx
```diff
@@ -9,25 +9,29 @@ type ItineraryLayerProps = {
...
   geometry?: GeoJsonLineString;
   hideItineraryLine?: boolean;
+  showStdcmAssets: boolean;
 };
 
 export default function ItineraryLayer({
   layerOrder,
   geometry,
   hideItineraryLine = false,
+  showStdcmAssets,
```
clarani, line 19: You can set it by default to false ;) This way, you only need to specify it in the 2 stdcm maps
```suggestion
  showStdcmAssets = false,
```

front/src/applications/stdcmV2/components/StdcmConfig.tsx
```diff
@@ -71,7 +69,11 @@ const StdcmConfig = ({
   const origin = useSelector(getOrigin);
   const destination = useSelector(getDestination);
 
-  const { pathfindingState } = usePathfinding(undefined, pathProperties);
+  const [pathfindingProperties, setPathfindingProperties] = useState<
+    ManageTrainSchedulePathProperties | undefined
+  >(undefined);
```
clarani: You don't need to specify that the undefined here ;) 
```suggestion
    ManageTrainSchedulePathProperties
  >();
```

clarani (commented):
Almost good to me !
I think that the ui of the main map has changed on the Eex, can you check that ? 
image

I think the 1 should be in white and more centered :thinking: 
How it looks like on dev right now: 
image

Caracol3:
Nice work  :-)
LGTM (tested)

Caracol3 (approved):
Nice work ! :-)
LGTM (tested)

front/src/modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/ItineraryLayer.tsx
SarahBellaha: This value is optional in the “Map” component, before being passed to the “ItineraryLayer” component. It is already set to “false” by default and specified only for Stdcm maps :)

front/src/applications/stdcmV2/components/StdcmConfig.tsx
```diff
@@ -71,7 +69,11 @@ const StdcmConfig = ({
   const origin = useSelector(getOrigin);
   const destination = useSelector(getDestination);
 
-  const { pathfindingState } = usePathfinding(undefined, pathProperties);
+  const [pathfindingProperties, setPathfindingProperties] = useState<
+    ManageTrainSchedulePathProperties | undefined
```
clarani: You can remove this undefined too
```suggestion
    ManageTrainSchedulePathProperties
```

clarani (approved):
LGTM :white_check_mark:

PR #8942, opened 2024-09-20
front: stdcm: persist user-entered date and time in the interface

SarahBellaha:
Closes [STDCM: Persist user-entered times & dates after a pathstep changed](https://github.com/osrd-project/osrd-confidential/issues/676)

### DIFF
--- a/front/src/applications/stdcmV2/components/StdcmDestination.tsx
+++ b/front/src/applications/stdcmV2/components/StdcmDestination.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useMemo } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 
 import { useTranslation } from 'react-i18next';
 
@@ -8,12 +8,12 @@ import type { StdcmConfSliceActions } from 'reducers/osrdconf/stdcmConf';
 import type { PathStep } from 'reducers/osrdconf/types';
 import { useAppDispatch } from 'store';
 import { replaceElementAtIndex } from 'utils/array';
-import { extractDateAndTimefromISO } from 'utils/date';
+import { extractDateAndTimefromISO, generateISODateFromDateTime } from 'utils/date';
 
 import StdcmCard from './StdcmCard';
 import StdcmOperationalPoint from './StdcmOperationalPoint';
 import StdcmOpSchedule from './StdcmOpSchedule';
-import { ArrivalTimeTypes, type StdcmConfigCardProps } from '../types';
+import { ArrivalTimeTypes, type ScheduleConstraint, type StdcmConfigCardProps } from '../types';
 
 const StdcmDestination = ({
   setCurrentSimulationInputs,
@@ -24,6 +24,11 @@ const StdcmDestination = ({
 }) => {
   const { t } = useTranslation('stdcm');
   const dispatch = useAppDispatch();
+
+  const [arrivalScheduleConstraint, setArrivalScheduleConstraint] = useState<
+    ScheduleConstraint | undefined
+  >();
+
   const {
     updateDestination,
     updateDestinationArrival,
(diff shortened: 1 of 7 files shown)


front/src/utils/date.ts
```diff
@@ -229,3 +229,18 @@ export function isArrivalDateInSearchTimeWindow(
...
+
+/**
+ * Generates an ISO arrival date string based on the provided date, hours, and minutes.
+ * @param {Date} date - The base date.
+ * @param {{ hours: number; minutes: number }} - An object containing the hours and the minutes of the arrival time.
+ * @returns {string} The ISO formatted arrival date string.
+ */
+export const generateISOArrival = (
+  date: Date,
+  { hours, minutes }: { hours: number; minutes: number }
```
Yohh: not really an issue, just a suggestion: we could create an explicit type instead of repeating those parameters

front/src/reducers/osrdconf/osrdConfCommon/index.ts
```diff
@@ -302,19 +302,21 @@ export function buildCommonConfReducers<S extends OsrdConfState>(): CommonConfRe
       state.startTime = action.payload;
     },
     updateOrigin(state: Draft<S>, action: PayloadAction<ArrayElement<S['pathSteps']>>) {
+      const prevOriginArrivalType = state.pathSteps[0]?.arrivalType;
```
kmer2016:
```suggestion
      const prevOriginArrivalType = state.pathSteps.at(0)?.arrivalType;
```

front/src/reducers/osrdconf/osrdConfCommon/index.ts
```diff
@@ -302,19 +302,21 @@ export function buildCommonConfReducers<S extends OsrdConfState>(): CommonConfRe
...
         ? {
             ...action.payload,
-            arrivalType: ArrivalTimeTypes.PRECISE_TIME,
+            arrivalType: prevOriginArrivalType || ArrivalTimeTypes.PRECISE_TIME,
           }
         : null;
       state.pathSteps = updateOriginPathStep(state.pathSteps, newPoint, true);
     },
     updateDestination(state: Draft<S>, action: PayloadAction<ArrayElement<S['pathSteps']>>) {
+      const prevDestinationArrivalType = state.pathSteps[state.pathSteps.length - 1]?.arrivalType;
```
kmer2016:
```suggestion
      const prevDestinationArrivalType = state.pathSteps.at(- 1)?.arrivalType;
```

front/src/utils/date.ts
```diff
@@ -229,3 +229,18 @@ export function isArrivalDateInSearchTimeWindow(
...
   return arrivalDate >= searchDatetimeWindow.begin && arrivalDate <= searchDatetimeWindow.end;
 }
+
+/**
+ * Generates an ISO arrival date string based on the provided date, hours, and minutes.
+ * @param {Date} date - The base date.
+ * @param {{ hours: number; minutes: number }} - An object containing the hours and the minutes of the arrival time.
+ * @returns {string} The ISO formatted arrival date string.
+ */
+export const generateISOArrival = (
```
kmer2016: I think the function could be used in contexts other than just arrival dates, so I suggest renaming it to something more generic. What do you think?
```suggestion
export const generateISODateFromDateTime = (
```

front/src/applications/stdcmV2/components/StdcmOrigin.tsx
```diff
@@ -50,10 +55,21 @@ const StdcmOrigin = ({
 
   const updateOriginPoint = (pathStep: PathStep | null) => {
     dispatch(updateOrigin(pathStep));
+    if (arrivalDateInput && arrivalTimeInput) {
+      const newOpArrival = generateISOArrival(arrivalDateInput, arrivalTimeInput);
+      dispatch(updateOriginArrival(newOpArrival));
+    }
```
kmer2016: To avoid dispatching two actions when arrivalDate and arrivalTime are defined, what do you think about dispatching a single `updateOrigin(pathStep)` with the pathStep already having the new arrival time set?

The same logic can be applied for `updateDestinationPoint`

```suggestion
const updateOriginPoint = (pathStep: PathStep | null) => {
    if (!pathStep || !arrivalDateInput || !arrivalTimeInput) {
      return dispatch(updateOrigin(pathStep));
    }
    dispatch(
      updateOrigin({
        ...pathStep,
        arrival: generateISOArrival(arrivalDateInput, arrivalTimeInput),
      })
    );
  };
```

front/src/applications/stdcmV2/components/StdcmDestination.tsx
```diff
@@ -24,6 +24,13 @@ const StdcmDestination = ({
 }) => {
   const { t } = useTranslation('stdcm');
   const dispatch = useAppDispatch();
+
+  const [arrivalDateInput, setArrivalDateInput] = useState<Date | undefined>(undefined);
+  const [arrivalTimeInput, setArrivalTimeInput] = useState<{ hours: number; minutes: number }>({
+    hours: 0,
+    minutes: 0,
+  });
+
```
kmer2016: I noticed that these states are duplicated in `StdcmDestination` and `StdcmOrigin`, so we might want to move them up to the parent `StdcmConfig`. Then, we can pass the necessary data to `StdcmDestination` and `StdcmOrigin`. 

To simplify things, I suggest unifying the variables. Instead of having `arrivalDateInput` and `arrivalTimeInput`, we could use a single variable called `scheduleConstraint` (you can suggest another name). This could have the following type:

```ts
type ScheduleConstraint = {
    date: Date;
    hour: number;
    minutes: number;
}
```

Then, `StdcmConfig` can manage this new state variable (`scheduleConstraint`) and pass it to `StdcmDestination` and `StdcmOrigin`.

kmer2016 (approved):
LGTM !

PR #8944, opened 2024-09-20
front: stdcm: display only CH codes associated with the selected PR

### LINKED ISSUE #8938: STDCM: The selected value is cleared for certain CHs
### What happened?

https://github.com/user-attachments/assets/6ad0b32c-f30d-4392-9b7f-5cb15201a86f

The selected value remains intact

1. Select a CI
2. Change the CH from the dropdown list (preferably choosing a value near the end of the list)

Recette (SNCF)

Google Chrome

fbf8e3e

SarahBellaha:
Closes #8938

### DIFF
--- a/front/src/applications/stdcmV2/components/StdcmOperationalPoint.tsx
+++ b/front/src/applications/stdcmV2/components/StdcmOperationalPoint.tsx
@@ -58,18 +58,20 @@ const StdcmOperationalPoint = ({
 
   const sortedChOptions = useMemo(
     () =>
-      sortedSearchResults.reduce(
-        (acc, pr) => {
-          const newObject = {
-            label: formatChCode(pr.ch),
-            id: pr.ch,
-          };
-          const isDuplicate = acc.some((option) => option.label === newObject.label);
-          if (!isDuplicate) acc.push(newObject);
-          return acc;
-        },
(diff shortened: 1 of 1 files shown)


front/src/applications/stdcmV2/components/StdcmOperationalPoint.tsx
```diff
@@ -58,18 +58,20 @@ const StdcmOperationalPoint = ({
...
+      sortedSearchResults
+        .filter((pr) => pr.name === searchTerm)
+        .reduce(
+          (acc, pr) => {
+            const newObject = {
+              label: formatChCode(pr.ch),
+              id: pr.ch,
+            };
+            const isDuplicate = acc.some((option) => option.label === newObject.label);
+            if (!isDuplicate) acc.push(newObject);
```
clarani, line 70: I think you can try to use a set, so you don't have to bother checking if it's a duplicate or not

clarani: Actually, it does not work because Sets compare the ref of the object and not the content. Let's keep it like this for now !

clarani (approved):
LGTM :white_check_mark:

RomainValls (approved):
LGTM

PR #8950, opened 2024-09-20
front: refacto TimesStops props

### LINKED ISSUE #8402: front: refacto the timesStopsTable props type

clarani:
closes #8402

### DIFF
--- a/front/src/modules/timesStops/TimesStops.tsx
+++ b/front/src/modules/timesStops/TimesStops.tsx
@@ -1,72 +1,33 @@
-import { useState, useEffect } from 'react';
-
 import cx from 'classnames';
-import { isEqual } from 'lodash';
 import { DynamicDataSheetGrid, type DataSheetGridProps } from 'react-datasheet-grid';
+import type { Operation } from 'react-datasheet-grid/dist/types';
 import { useTranslation } from 'react-i18next';
 
-import { useOsrdConfActions } from 'common/osrdContext';
-import type { IsoDateTimeString } from 'common/types';
-import { isVia } from 'modules/pathfinding/utils';
-import type { SuggestedOP } from 'modules/trainschedule/components/ManageTrainSchedule/types';
-import type { PathStep } from 'reducers/osrdconf/types';
-import { useAppDispatch } from 'store';
-
-import {
-  formatSuggestedViasToRowVias,
-  updateRowTimesAndMargin,
-  updateDaySinceDeparture,
-  durationSinceStartTime,
-} from './helpers/utils';
 import { useTimeStopsColumns } from './hooks/useTimeStopsColumns';
-import { TableType } from './types';
-import type { PathWaypointRow } from './types';
+import { type TableType, type TimeStopsRow } from './types';
 
-export const WITH_KP = true;
-
-type TimesStopsProps = {
-  allWaypoints?: SuggestedOP[];
-  pathSteps?: PathStep[];
-  startTime?: IsoDateTimeString;
+type TimesStopsProps<T extends TimeStopsRow> = {
+  rows: T[];
   tableType: TableType;
   cellClassName?: DataSheetGridProps['cellClassName'];
   stickyRightColumn?: DataSheetGridProps['stickyRightColumn'];
   headerRowHeight?: number;
+  onChange?: (newRows: T[], operation: Operation) => void;
 };
 
-const TimesStops = ({
-  allWaypoints,
-  pathSteps,
-  startTime,
+const TimesStops = <T extends TimeStopsRow>({
+  rows,
   tableType,
   cellClassName,
   stickyRightColumn,
   headerRowHeight,
-}: TimesStopsProps) => {
-  const isInputTable = tableType === TableType.Input;
+  onChange,
+}: TimesStopsProps<T>) => {
   const { t } = useTranslation('timesStops');
 
-  const dispatch = useAppDispatch();
-  const { upsertSeveralViasFromSuggestedOP } = useOsrdConfActions();
+  const columns = useTimeStopsColumns(tableType, rows);
 
-  const [rows, setRows] = useState<PathWaypointRow[]>([]);
-
-  useEffect(() => {
-    if (allWaypoints) {
-      const suggestedOPs = formatSuggestedViasToRowVias(
-        allWaypoints,
-        pathSteps || [],
-        t,
-        startTime,
-        tableType
-      );
-      setRows(updateDaySinceDeparture(suggestedOPs, startTime, true));
-    }
-  }, [allWaypoints, pathSteps, startTime]);
-
-  const columns = useTimeStopsColumns(tableType, allWaypoints);
-
-  if (!allWaypoints) {
+  if (!rows) {
     return (
       <div className="d-flex justify-content-center align-items-center h-100">
         <p className="pt-1 px-5">{t('noPathLoaded')}</p>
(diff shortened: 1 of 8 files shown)


front/src/modules/timesStops/types.ts
```diff
@@ -9,12 +9,29 @@ export type TimeExtraDays = {
...
+  onStopSignal?: boolean;
+  theoreticalMargin?: string; // value asked by user
+
+  theoreticalMarginSeconds?: string;
+  calculatedMargin?: string;
+  diffMargins?: string;
+  calculatedArrival?: string | null;
+  calculatedDeparture?: string | null;
+
+  isMarginValid?: boolean;
```
SharglutDev, line 30: Does the output table need this information ? Should this rather be in the other type ?

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -198,10 +199,10 @@ export function updateRowTimesAndMargin(
  * and updates the number of days since departure.
  */
 export function updateDaySinceDeparture(
-  pathWaypointRows: PathWaypointRow[],
+  pathWaypointRows: TimesStopsInputRow[],
   startTime?: IsoDateTimeString,
   keepFirstIndexArrival?: boolean
```
SharglutDev: I know it doesn't come from your PR but we could harmonize this with what we decided for `secToHoursString`

```suggestion
  { keepFirstIndexArrival?: boolean }
```

front/src/modules/timesStops/TimesStops.tsx
```diff
@@ -1,72 +1,33 @@
...
-  allWaypoints?: SuggestedOP[];
-  pathSteps?: PathStep[];
-  startTime?: IsoDateTimeString;
+type TimesStopsProps<T extends TimeStopsRow> = {
+  rows?: T[];
   tableType: TableType;
   cellClassName?: DataSheetGridProps['cellClassName'];
   stickyRightColumn?: DataSheetGridProps['stickyRightColumn'];
   headerRowHeight?: number;
+  onChange?: (newRows: T[], op: Operation) => void;
```
SharglutDev: Maybe we shouldn't use `op`as there can be points added by map click in the table.
We could use `row` or `waypoint` for example ?

front/src/modules/timesStops/TimesStopsInput.tsx
```diff
@@ -87,11 +101,53 @@ const TimesStopsinput = ({ allWaypoints, startTime, pathSteps }: TimesStopsInput
     });
     dispatch(updatePathSteps({ pathSteps: updatedPathSteps }));
   };
+
+  const onChange = (newRows: TimesStopsInputRow[], op: Operation) => {
```
SharglutDev: Should maybe be a  `useCallback

front/src/modules/timesStops/types.ts
```diff
@@ -9,12 +9,29 @@ export type TimeExtraDays = {
   dayDisplayed?: boolean;
 };
 
-export type PathWaypointRow = Omit<SuggestedOP, 'arrival' | 'departure'> & {
-  isMarginValid: boolean;
+export type TimeStopsRow = {
```
anisometropie, line 12:
```
type TimeStopsRow = Pick<
  SuggestedOP,
  | 'opId'
  | 'name'
  | 'ch'
  | 'stopFor'
  | 'onStopSignal'
  | 'theoreticalMargin'
  | 'theoreticalMarginSeconds'
  | 'diffMargins'
  | 'calculatedArrival'
  | 'calculatedDeparture'
  | 'calculatedMargin'
> & {
  arrival?: TimeExtraDays; // value asked by user
  departure?: TimeExtraDays; // value asked by user
  isWaypoint: boolean;
  isMarginValid?: boolean;
};
```

front/src/modules/timesStops/types.ts
```diff
@@ -9,12 +9,29 @@ export type TimeExtraDays = {
...
+  theoreticalMarginSeconds?: string;
+  calculatedMargin?: string;
+  diffMargins?: string;
+  calculatedArrival?: string | null;
+  calculatedDeparture?: string | null;
+
+  isMarginValid?: boolean;
 };
 
+export type TimesStopsInputRow = Omit<SuggestedOP, 'arrival' | 'departure'> & TimeStopsRow;
```
anisometropie, line 33: do you need the full `suggestedOP` here, all the other keys ?

front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -10,21 +11,33 @@ import type { PathfindingResultSuccess, TrainScheduleResult } from 'common/api/o
...
+} from '../helpers/utils';
+import {
+  type TrainScheduleBasePathWithUic,
+  type ScheduleEntry,
+  type TimeStopsRow,
+  TableType,
+} from '../types';
 
-function useOutputTableData(
+const useOutputTableData = (
```
anisometropie: I don’t see the point.
arrow functions were made to simplify writing callbacks like `setTimeout(() => { doStuff() }, 2000)` or higher level function
() => {} were never meant to replace `function`

if you remove every single function and replace by `const` you completely lose the visual cues, you have no more distinction between const functions and const everything else.

front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -116,19 +129,29 @@ function useOutputTableData(
...
-        return {
-          ...sugOpPoint,
-          ...matchingPathStep,
-        };
-      }
-      return sugOpPoint;
+      return {
+        ...sugOpPoint,
+        ...(matchingPathStep || {}),
+        isWaypoint: matchingPathStep !== undefined,
```
anisometropie, line 139: good idea !

front/src/modules/timesStops/TimesStops.tsx
```diff
@@ -1,72 +1,33 @@
...
-  const isInputTable = tableType === TableType.Input;
+  onChange,
+}: TimesStopsProps<T>) => {
   const { t } = useTranslation('timesStops');
 
-  const dispatch = useAppDispatch();
-  const { upsertSeveralViasFromSuggestedOP } = useOsrdConfActions();
-
-  const [rows, setRows] = useState<PathWaypointRow[]>([]);
+  const columns = useTimeStopsColumns(tableType, rows);
```
anisometropie, line 28: you have a 
`Argument of type 'T[] | undefined' is not assignable to parameter of type 'SuggestedOP[] | undefined'.
`
perhaps make useTimeStopColumns a generic function too

anisometropie (changes requested):
Great job !
really good to have `[rows, setRows]` and `onChange` only in the input table, it’s much neater and easy to use

front/src/modules/timesStops/TimesStops.tsx
```diff
@@ -79,50 +40,18 @@ const TimesStops = ({
...
       height={600}
       headerRowHeight={headerRowHeight}
-      rowClassName={({ rowData, rowIndex }) =>
+      rowClassName={({ rowData }) =>
         cx({
-          activeRow:
-            rowIndex === 0 ||
-            rowIndex === allWaypoints.length - 1 ||
-            isVia(pathSteps || [], rowData, { withKP: true }),
+          activeRow: rowData.isWaypoint,
```
SharglutDev, line 54: For a train imported with open data, it seems like the destination is not considered as a PathStep (or it can't find it), it's not bolded

front/src/modules/timesStops/TimesStops.tsx
clarani: Actually, op stood for operation :) I used this new name

front/src/modules/timesStops/TimesStopsInput.tsx
clarani: Done

front/src/modules/timesStops/types.ts
clarani: You're right, the input table does not need this props, I moved it in TimesStopsInputRow

front/src/modules/timesStops/types.ts
clarani: Hm I don't want to remove props here since the data is then dispatched in the store (see the OnChange callback in TimesStopsInput).
I think we should refacto the suggestedOp type later, in another PR. I would rather keep TimesStopsInputRow like this for now, to avoid breaking something by mistake

front/src/modules/timesStops/types.ts
clarani: I don't really want to tie TimeStopRow to SuggestedOp since SuggestedOp a bit messy (I think we should refacto it too).
Is it okay for you if I keep it like this ?

front/src/modules/timesStops/hooks/useOutputTableData.ts
clarani: I've changed it back. We can discuss it later if you want, maybe I should read some documentation about it

front/src/modules/timesStops/TimesStops.tsx
clarani: Done :)

front/src/modules/timesStops/TimesStops.tsx
clarani: Ah actually I have an error after `useTimeStopsColumns`. I think I will try to fix that in another PR, if that's ok for you ?

front/src/modules/timesStops/helpers/utils.ts
clarani: Done :)

front/src/modules/timesStops/types.ts
anisometropie: I guess my only concern was to have all these fields duplicated with the exact same type. 
if we change a field in one we might forget to change the other.
we can leave it like this but the refacto job should include coming back to TimeStopsRow after SuggestedOp

front/src/modules/timesStops/TimesStops.tsx
clarani: Discussed, we can accept this small regression since it's only for imported trains. I will open a issue to manage more globally the imported trains (their vias are not in bold right now on dev)

front/src/modules/timesStops/types.ts
anisometropie: works for me. can you create a ticket to keep a record of what we discussed here and [just above](https://github.com/OpenRailAssociation/osrd/pull/8950#discussion_r1774907994)

front/src/modules/timesStops/hooks/useOutputTableData.ts
anisometropie: alright, added it to next front meeting

anisometropie (approved):

clarani:
@anisometropie I have created this issue to refactor SuggestedOp : https://github.com/OpenRailAssociation/osrd/issues/8996

SharglutDev (approved):
Lgtm and tested, good job :)

PR #8953, opened 2024-09-20
editoast: add endpoint to check if locations exist

### LINKED ISSUE #8636: Editoast: being able to identify unrecognized PR in the infrastructure

hamz2a:
closes [#8636](https://github.com/OpenRailAssociation/osrd/issues/8636)

### DIFF
--- a/editoast/editoast_schemas/src/infra/operational_point.rs
+++ b/editoast/editoast_schemas/src/infra/operational_point.rs
@@ -94,6 +94,12 @@ impl OSRDIdentified for OperationalPoint {
     }
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum ParseError {
+    #[error("Invalid JSON format: {0}")]
+    JsonError(#[from] serde_json::Error),
+}
+
 impl OperationalPoint {
     pub fn track_offset(&self) -> Vec<TrackOffset> {
         self.parts
@@ -104,6 +110,17 @@ impl OperationalPoint {
                 offset: (el.position * 1000.0) as u64,
             })
             .collect()
+    }
+
+    pub fn from_railjson(railjson: serde_json::Value) -> Result<Self, ParseError> {
+        serde_json::from_value(railjson).map_err(ParseError::from)
+    }
+
+    pub fn matches_ch(&self, secondary_code: &str) -> bool {
+        if let Some(extensions) = &self.extensions.sncf {
+            return extensions.ch == secondary_code;
+        }
+        false
     }
 }
 
--- a/editoast/src/models/infra/object_queryable.rs
+++ b/editoast/src/models/infra/object_queryable.rs
@@ -3,6 +3,7 @@ use std::ops::DerefMut;
 use diesel::sql_query;
 use diesel::sql_types::Array;
 use diesel::sql_types::BigInt;
+use diesel::sql_types::Double;
 use diesel::sql_types::Jsonb;
 use diesel::sql_types::Nullable;
 use diesel::sql_types::Text;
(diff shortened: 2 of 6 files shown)


editoast/src/views/infra/objects.rs
```diff
@@ -20,8 +22,12 @@ use crate::views::AuthorizationError;
 use crate::views::AuthorizerExt;
 use crate::Retrieve;
 
+const CH_BV: &str = "BV";
+const CH_00: &str = "00";
+
 crate::routes! {
     "/objects/{object_type}" => get_objects,
+    "/check-locations" => check_locations,
```
flomonster, line 30: I would rename the endpoint since what it really does is returning path steps coordinates.

editoast/editoast_schemas/src/infra/operational_point.rs
```diff
@@ -105,6 +111,17 @@ impl OperationalPoint {
             })
             .collect()
     }
+
+    pub fn from_railjson(railjson: serde_json::Value) -> Result<Self, ParseError> {
+        serde_json::from_value(railjson).map_err(ParseError::from)
+    }
```
flomonster, line 117: Why not letting diesel deserialize to OP?

editoast/src/models/infra/object_queryable.rs
flomonster, line 1: We already have something that transform a list of `PathStep` into a `Vec<Vec<TrackOffset>>`. This also checked that all path steps are valids.

It's important that this logic is shared between both endpoints.

> [!IMPORTANT]
> Small changes are required. This function raise an error when it encounter the first invalid path step.
> It should instead continue and list all valids/invalid.

[The function I'm talking about](https://github.com/OpenRailAssociation/osrd/blob/0a8fa8019a7b99fd006aa80e87ff6fffd1c22beb/editoast/src/views/path/path_item_cache.rs#L85)

flomonster (changes requested):
First round of review. This new endpoint should share code with the PF one. Check my comments bellow.

editoast/src/models/infra/object_queryable.rs
```diff
@@ -72,4 +73,83 @@ impl Infra {
...
+    ) -> Result<Vec<ObjectQueryable>> {
+        let query = format!("
+            SELECT DISTINCT ON (object_table.obj_id)
+                object_table.obj_id as obj_id,
+                object_table.data as railjson,
+                ST_AsGeoJSON(ST_Transform(geographic, 4326))::jsonb as geographic
+            FROM {} AS object_table
+            LEFT JOIN {} AS geometry_table ON object_table.obj_id = geometry_table.obj_id AND object_table.infra_id = geometry_table.infra_id
+            WHERE object_table.infra_id = $1
+            AND (data->'extensions'->'identifier'->'uic')::integer = $2",
```
woshilapin, line 144: Since there are some not trivial SQL queries in this file, and we might benefit from syntax highlighting to maintain them, how about moving them into an independent file? You can already find a bunch of examples if you look for `include_str!("sql/something.sql")`.

editoast/src/views/infra/objects.rs
```diff
@@ -114,17 +120,150 @@ async fn get_objects(
...
+        return Err(AuthorizationError::Unauthorized.into());
+    }
+
+    let infra_id = infra_id_param.infra_id;
+    let infra = Infra::retrieve_or_fail(&mut db_pool.get().await?, infra_id, || {
+        InfraApiError::NotFound { infra_id }
+    })
+    .await?;
+
+    let mut results = HashMap::new();
```
woshilapin, line 152: I don't think you need to define a mutable map here. A mutable empty `HashMap` will have the problem of growing item by item with regular re-allocation. By streaming through function chaining, there is a good chance that `.size_hint()` will be used to allocate the entire capacity only once in `.collect()` (see [implementation of `FromIterator` of `hashbrown`](https://docs.rs/hashbrown/latest/src/hashbrown/map.rs.html#6437-6453), the underlying implementation of `std`).

```rust
path_item_locations.into_iter()
  .map(|location|
    match location {
      /* almost the same code except returns tuple instead of results.push() */
    }
  )
  .collect::<HashMap<_, _>>()
```

editoast/src/views/infra/objects.rs
```diff
@@ -114,17 +120,150 @@ async fn get_objects(
...
+    } else if let Some(op_obj) =
+        filter_operational_point_by_ch(operational_point_objects.clone(), CH_BV)
+    {
+        Some(op_obj)
+    } else if let Some(op_obj) =
+        filter_operational_point_by_ch(operational_point_objects.clone(), CH_00)
+    {
+        Some(op_obj)
+    } else {
+        Some(operational_point_objects.first().unwrap().clone())
```
woshilapin, line 241: You could avoid the `.unwrap()` here by replacing it by something like this.
```suggestion
        Some(operational_point_objects.first()?.clone())
```
Technically, you then wouldn't need the early return at the beginning of the function testing for the emptyness of the `Vec`... but it might still be useful because it's an early-return, happening before a bunch of function calls.

editoast/src/views/infra/objects.rs
```diff
@@ -246,4 +385,226 @@ mod tests {
...
+                    secondary_code: None,
+                },
+                PathItemLocation::OperationalPointUic {
+                    uic: 4,
+                    secondary_code: None,
+                },
+            ]);
+
+        let res: HashMap<String, Option<ObjectQueryable>> =
+            app.fetch(request).assert_status(StatusCode::OK).json_into();
```
woshilapin, line 416: There is something I don't understand: the endpoint returns `HashMap<String, ObjectQueryable>` but here, there is a `HashMap<String, Option<ObjectQueryable>>`, how does the test even compile? Did I miss something?

editoast/src/views/infra/objects.rs
```diff
@@ -246,4 +385,226 @@ mod tests {
...
+            })]);
+
+        let res: HashMap<String, Option<ObjectQueryable>> =
+            app.fetch(request).assert_status(StatusCode::OK).json_into();
+
+        assert!(res.is_empty());
+    }
+
+    #[rstest]
+    async fn check_locations_return_none_when_op_id_does_not_exist() {
```
woshilapin, line 457: When I see those tests, and that nothing is returned when location is not found, I realize how the proposition you showed me (having an `enum LocationChecked { Found(ObjectQueryable), NotFound }` makes sense: returning a `HashMap<String, LocationChecked>` instead of `HashMap<String, ObjectQueryable>`. As a consumer of the API, it is indeed weird to not return anything about something asked. I'm sorry I didn't fully understand the entire context when we talked about it. Let's ask @flomonster and/or @leovalais what's they're opinion before addressing that.

woshilapin (changes requested):
Really nice work and with plenty of tests which is really enjoyable.

I'm putting a Request for changes because I'd like some feedback about returning or not errors. The other comments are minor and not blocking.

leovalais:
@woshilapin We decided with @flomonster and @hamz2a to change the way this feature would be implemented yesterday evening.

We don't need a new endpoint because:

1. The pathfinding endpoint is called as soon as the front displays the schedule.
2. This endpoint already checks the locations are consistent within the infra of the scenario.
3. If they're not, the pathfinding returns the first errored location.

Basically, the only thing needed on our end is to improve the pathfinding algorithm to collect all invalid locations instead of just returning the first one.

We also want the frontend to display the subset of locations that are valid on the map. To achieve that, the `/search` endpoint already provides a way to select a few operational points and return their geometry. But it's now up to the frontend to call the `/search` endpoint when the pathfinding fails.

Additionally, improving the pathfinding error reporting will likely be useful to diagnose train schedule import failures. (And it's always a good thing for API quality.)

PR #8954, opened 2024-09-20
editoast: fix train schedule projection cache

### LINKED ISSUE #8952: front: missing trains in space-time chart
### What happened?

See attached video: there's supposed to be 4 trains in the space-time chart, but some trains can be missing. It changes from one request to the next.

Note: the response from the back-end doesn't seem to be missing any train (though I haven't checked if the content is valid)

We can also see https://github.com/OpenRailAssociation/osrd/issues/8935 in the video, I don't know if it's related

[Screencast from 20-09-2024 14:56:07.webm](https://github.com/user-attachments/assets/14f081d3-0e7f-4b99-a97d-94b3fbca2ae0)

All 4 trains should always be displayed in this case.

I haven't tested for the exact conditions for this bug to happen. It seems to happen *at least* on the old stdcm page when several trains are show in the chart (with a smaller path than the new train).  Having several trains that share the same path seems important. 

Local

Firefox

68e0513349a18e7c9faeed6724919ec161fcb046

flomonster:
The train projections retrieve in redis were associated with a random train id.

close  #8952

### DIFF
--- a/editoast/src/views/train_schedule/projection.rs
+++ b/editoast/src/views/train_schedule/projection.rs
@@ -6,6 +6,7 @@ use chrono::Utc;
 use editoast_authz::BuiltinRole;
 use editoast_schemas::primitives::Identifier;
 use futures::join;
+use itertools::izip;
 use serde::Deserialize;
 use serde::Serialize;
 use std::collections::hash_map::DefaultHasher;
@@ -193,10 +194,10 @@ async fn project_path(
     .await?;
 
     // 1. Retrieve cached projection
-    let mut trains_hash_values = HashMap::new();
-    let mut trains_details = HashMap::new();
+    let mut trains_hash_values = vec![];
+    let mut trains_details = vec![];
 
-    for (train, (sim, pathfinding_result)) in trains.iter().zip(simulations) {
+    for (sim, pathfinding_result) in simulations {
         let track_ranges = match pathfinding_result {
             PathfindingResult::Success(PathfindingResultSuccess {
                 track_section_ranges,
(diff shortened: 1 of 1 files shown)


editoast/src/views/train_schedule/projection.rs
```diff
@@ -234,17 +234,18 @@ async fn project_path(
...
+        redis_conn.json_get_bulk(&trains_hash_values).await?;
 
     let mut hit_cache: HashMap<i64, CachedProjectPathTrainResult> = HashMap::new();
     let mut miss_cache = HashMap::new();
-    for ((train_id, train_details), projection) in
-        trains_details.into_iter().zip(cached_projections)
+    for ((train_details, projection), train_id) in trains_details
+        .into_iter()
+        .zip(cached_projections)
+        .zip(trains.iter().map(|t| t.id))
```
woshilapin: Just wondering if something like this is not more readable? (I know, it was already like this before the PR, just this PR is chaining it one more layer).

```rust
for (index, train_details) in trains_details.enumerate() {
  let train_id = trains[index].id;
  if let Some(cached) = cached_projections[index] {
    /* ... */
  }
  /* ... */
}
```

One drawback is that we used the `[]` indexing which can panic. But technically, it might be better than `.zip()` which will just silently short-circuit if one of the iterator is shorter.

woshilapin (approved):
Ahhhh, the randomness of `HashMap`, always a little trick that comes back to bite you. Note: would a different `Hasher` (see https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#method.with_hasher) avoid this randomness, so you can keep the previous version of the code?

editoast/src/views/train_schedule/projection.rs
leovalais: [izip!](https://docs.rs/itertools/latest/itertools/macro.izip.html) could work as well

flomonster:
> Ahhhh, the randomness of `HashMap`, always a little trick that comes back to bite you. Note: would a different `Hasher` (see https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#method.with_hasher) avoid this randomness, so you can keep the previous version of the code?

I can probably find another hash function to avoid randomness. However, I prefer to use indexing for readability reasons.

editoast/src/views/train_schedule/projection.rs
flomonster: fixed

PR #8956, opened 2024-09-21
front: schedule table : fix 00:00:00 stop input

### LINKED ISSUE #8925: `00:00:00` doesn't seem to be a valid time input
### What happened?

This input:
image

Leads to this after enter:
image

And the info isn't taken into account in the computation

I'm expecting `00:00:00` to be interpreted as midnight, therefore having a J+1 next to it most likely

.

Recette (SNCF)

Google Chrome

68e0513

Akctarus:
fix #8925

### DIFF
--- a/front/src/modules/timesStops/helpers/__tests__/utils.spec.ts
+++ b/front/src/modules/timesStops/helpers/__tests__/utils.spec.ts
@@ -286,6 +286,47 @@ describe('updateDaySinceDeparture', () => {
       ];
       expect(result).toEqual(expected);
     });
+    it('should handle departure exactly at midnight', () => {
+      const pathWaypointRows = [
+        {
+          opId: 'd9a382bc',
+          name: 'St',
+          uic: 75,
+          ch: 'BV',
+          arrival: { time: '00:00:00' },
+        },
+        {
+          opId: 'd9b38600',
+          name: 'Ge',
+          uic: 86,
+          ch: 'BX',
+          arrival: { time: '01:00:00' },
+        },
+      ] as TimesStopsInputRow[];
+      const startTime = '2024-08-13T00:00:00';
+      const result = updateDaySinceDeparture(pathWaypointRows, startTime, {
+        keepFirstIndexArrival: true,
+      });
+      const expected = [
+        {
+          opId: 'd9a382bc',
+          name: 'St',
+          uic: 75,
+          ch: 'BV',
+          arrival: { time: '00:00:00', daySinceDeparture: 0 },
+          departure: undefined,
+        },
+        {
+          opId: 'd9b38600',
+          name: 'Ge',
+          uic: 86,
+          ch: 'BX',
+          arrival: { time: '01:00:00', daySinceDeparture: 0 },
+          departure: undefined,
+        },
+      ];
+      expect(result).toEqual(expected);
+    });
   });
   describe('2 day span', () => {
     it('should add day 1 field', () => {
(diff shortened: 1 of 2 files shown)


front/src/modules/timesStops/TimesStopsInput.tsx
```diff
@@ -85,12 +88,34 @@ const TimesStopsinput = ({ allWaypoints, startTime, pathSteps }: TimesStopsInput
       }
       return step;
     });
+
     dispatch(updatePathSteps({ pathSteps: updatedPathSteps }));
   };
+
+  // Handle 00:00:00 time input interpretation as J+1 (next day)
+  const timeWithDayAdjustment = useMemo(
```
anisometropie: I think you just need to use the existing `updateDaySinceDeparture`

anisometropie: write a few tests like this one 3fa20ae043ac03b0ab0f911b25a7d2e0aa369113
with the departure point at 00:00:00, a waypoint in the middle at 00:00:00
and over 2 day/3 day span

Akctarus: done !

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -228,14 +230,18 @@ export function updateDaySinceDeparture(
...
 
     let formattedDeparture: TimeExtraDays | undefined;
-    if (stopFor && arrivalInSeconds) {
+
+    if (stopFor && arrivalInSeconds !== null) {
       const departureInSeconds = (arrivalInSeconds + Number(stopFor)) % SECONDS_IN_A_DAY;
       const isAfterMidnight = departureInSeconds < previousTime;
-      if (isAfterMidnight) {
+      const isDepartureMidnight =
+        secToHoursString(departureInSeconds, { withSeconds: true }) === '00:00:00';
```
emersion: Wouldn't it be easier to check `departureInSeconds === 0`?

front/src/modules/timesStops/helpers/__tests__/utils.spec.ts
```diff
@@ -329,6 +329,215 @@ describe('updateDaySinceDeparture', () => {
...
+          name: 'Fr',
+          uic: 87,
+          ch: 'BY',
+          arrival: { time: '01:30:00', daySinceDeparture: 1 },
+          departure: undefined,
+        },
+      ];
+      expect(result).toEqual(expected);
+    });
+    it('should handle a two-day span with midnight', () => {
```
anisometropie: Actually all the tests are already all 2-day span since they go over midnight once. This one doesn’t really test anything new.

Perhaps make this test go over 3 different days? with one more `00:00:00` and move it inside of the `describe('3 day span')`.

anisometropie (changes requested):

front/src/modules/timesStops/helpers/utils.ts
Akctarus: done !

front/src/modules/timesStops/helpers/__tests__/utils.spec.ts
Akctarus: done :)

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -248,9 +253,8 @@ export function updateDaySinceDeparture(
           daySinceDeparture: currentDaySinceDeparture,
         };
       }
-      previousTime = departureInSeconds;
+      previousTime = isDepartureMidnight ? 0 : departureInSeconds;
```
emersion: `isDepartureMidnight` is true if and only if `departureInSeconds === 0` so I think we can just keep the old logic here?

kmer2016 (approved):

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -228,14 +230,17 @@ export function updateDaySinceDeparture(
...
-      previousTime = arrivalInSeconds;
+      previousTime = isMidnight ? 0 : arrivalInSeconds;
     }
 
     let formattedDeparture: TimeExtraDays | undefined;
-    if (stopFor && arrivalInSeconds) {
+
+    if (stopFor && arrivalInSeconds !== null) {
       const departureInSeconds = (arrivalInSeconds + Number(stopFor)) % SECONDS_IN_A_DAY;
       const isAfterMidnight = departureInSeconds < previousTime;
```
emersion, line 240: I was wondering if this is really the right way to figure out whether the day changed. It would be nicer to strip the time from the date+time, and then compare that with the previous one. Then we don't need to special-case midnight.

(Kind of orthogonal to this PR I suppose?)

front/src/modules/timesStops/helpers/utils.ts
Akctarus: you're right, I changed it

front/src/modules/timesStops/helpers/utils.ts
Akctarus: This is indeed an orthogonal suggestion but also a good idea, maybe we could create an enhancement/refactor ticket ?

front/src/modules/timesStops/helpers/__tests__/utils.spec.ts
```diff
@@ -329,6 +329,144 @@ describe('updateDaySinceDeparture', () => {
...
+        },
+        {
+          opId: 'd9b38600',
+          name: 'Ge',
+          uic: 86,
+          ch: 'BX',
+          arrival: { time: '01:00:00', daySinceDeparture: 1 },
+          departure: undefined,
+        },
+      ];
```
anisometropie: Missed that earlier, but this test is not correct. both steps should be in day 0 and not display D+

```suggestion
      const expected = [
        {
          opId: 'd9a382bc',
          name: 'St',
          uic: 75,
          ch: 'BV',
          arrival: { time: '00:00:00', daySinceDeparture: 0 },
          departure: undefined,
        },
        {
          opId: 'd9b38600',
          name: 'Ge',
          uic: 86,
          ch: 'BX',
          arrival: { time: '01:00:00', daySinceDeparture: 0 },
          departure: undefined,
        },
      ];
```

Screenshot_20241002_140303

anisometropie: You can move the test in the 1-day span section

anisometropie (changes requested):
thanks for earlier changes :) Just one last case that’s not handled correctly

front/src/modules/timesStops/helpers/utils.ts
emersion: Sure thing

front/src/modules/timesStops/helpers/__tests__/utils.spec.ts
Akctarus: Done + I've modified the function to take this case into account

anisometropie (approved):
Thanks !

PR #8973, opened 2024-09-24
front: remove all occurences of console.log

anisometropie:
I think we should reserve console.log for debugging in development branches, it should not exist in production.
also, when doing a search for console.log to see if none were left when cleaning a branch, they always show up.

- replaced console.log by warn/error/info in different scripts
- removed old debugging/timing logs that were always popping during development.


clarani (approved):
LGTM :white_check_mark:

Math-R (approved):

PR #8975, opened 2024-09-24
core: stdcm: fix invalid offset in stop locations

### LINKED ISSUE #8960: core: `STDCM result is conflicting with the scheduled timetable`
### What happened?

This assertion error happened.

More context: this is a "sanity check" where we compare the output to the timetable using the normal standalone sim + conflict detection pipeline. It is quite unusual because this check **should** be entirely redundant with stdcm post-processing. 

_No response_

Seems to be reproductible with the fuzzer on France, though it's very uncommon

Local, Recette (SNCF)

Firefox

a184fb531d9deac1afda9cde07959a9a2bf42298

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/8960

Includes a fuzzer-generated regression test on small_infra


Erashin (approved):

shenriotpro (approved):
LGTM for tests/

PR #8977, opened 2024-09-24
core: signaling: pick the right signaling system for next signal

eckter:
We'd use the last of the path signals to get the least restricting aspect for the next signal. But the `.evaluate` call simulates whole blocks. If there a sig system change after the end of the path but before the end of the last block, we'd get an error.


Khoyo (approved):
Nice catch!

PR #8983, opened 2024-09-24
editoast, core, front, tests: introduce ReceptionSignal enum

### LINKED ISSUE #8809: Editoast: add "short slip distance" as enum (with closed-signal)
Part of https://github.com/osrd-project/osrd-confidential/issues/305

Replace the existing flag for reception on closed signal by an enum "open-signal", "closed-signal", "short-slip-closed-signal".
Mainly used by operational studies, but should be the default value for STDCM when stop duration is null (!= 0).

Probably already add it to core interface should be easier.

bougue-pe:
Possible values are OPEN, STOP & SHORT_SLIP_STOP.
No functional change anywhere, just changing interfaces/types everywhere.

Fix: https://github.com/OpenRailAssociation/osrd/issues/8809


editoast/openapi.yaml
```diff
@@ -8372,11 +8378,8 @@ components:
         locked:
           type: boolean
           description: Whether the schedule item is locked (only for display purposes)
-        on_stop_signal:
-          type: boolean
-          description: |-
-            Whether the next signal is expected to be blocking while stopping
-            Can be true only if `stop_for` is `Some`
+        reception_signal:
+          $ref: '#/components/schemas/ReceptionSignal'
```
bougue-pe, line 8385: The description is not forwarded, I tried to dig different solutions, any idea is welcome.

front/src/modules/timesStops/TimesStops.tsx
```diff
@@ -103,13 +104,12 @@ const TimesStops = ({
...
-              (row) =>
-                ({
-                  ...row,
-                  ...(row.arrival && { arrival: durationSinceStartTime(startTime, row.arrival) }),
-                }) as SuggestedOP
-            );
+            .map(({ onStopSignal, arrival, departure, ...row }) => ({
+              ...row,
+              arrival: durationSinceStartTime(startTime, arrival),
+              departure: durationSinceStartTime(startTime, departure),
```
bougue-pe: 🔍 Out-of-scope change: adding `departure` management on the way to avoid "bad" cast as `SuggestedOP` it looks fine 🤞.

front/src/modules/timesStops/TimesStopsInput.tsx
```diff
@@ -29,10 +31,10 @@ const createClearViaButton = ({
     rowIndex > 0 &&
     rowIndex < allWaypoints.length - 1 &&
     isVia(pathSteps || [], rowData, { withKP: true }) &&
-    (rowData.stopFor !== undefined ||
+    (!isNil(rowData.stopFor) ||
```
bougue-pe, line 44: 🔍 Out-of-scope change: `null` is now also handled.

editoast/openapi.yaml
leovalais: From https://swagger.io/docs/specification/using-ref/

> $ref and Sibling Elements
Any sibling elements of a $ref are ignored. This is because $ref works by replacing itself and everything on its level with the definition it is pointing at.

I guess that's why utoipa ignores it.

Wdyt about moving the description of the `ReceptionSignal` to the enum itself, and the `stop_for = null` warning to the `stop_for` field description?

editoast/src/views/train_schedule.rs
```diff
@@ -563,7 +563,7 @@ fn build_simulation_request(
                 .stop_for
                 .as_ref()
                 .map(|t| t.num_milliseconds() as u64),
-            on_stop_signal: schedule_item.on_stop_signal,
+            reception_signal: schedule_item.reception_signal.clone(),
```
leovalais: Wdyt about making the enum derive `Copy`?

leovalais (approved):
LGTM

editoast/openapi.yaml
bougue-pe: Indeed, done in https://github.com/OpenRailAssociation/osrd/pull/8983/commits/daf83e3461ef97280de9964d6ba7202e7260bfd3, thanks!

editoast/src/views/train_schedule.rs
bougue-pe: Done in https://github.com/OpenRailAssociation/osrd/pull/8983/commits/daf83e3461ef97280de9964d6ba7202e7260bfd3

core/src/main/kotlin/fr/sncf/osrd/api/stdcm/STDCMEndpoint.kt
```diff
@@ -136,8 +137,8 @@ fun makeTrainSchedule(
     trainStops: List<TrainStop>
 ): StandaloneTrainSchedule {
     val mutTrainStops = ArrayList(trainStops)
-    // Force the train to end its path at speed=0 and on closed signal
-    mutTrainStops.add(TrainStop(endPos, 0.1, true))
+    // Force the train to end its path at speed=0 and on closed signal with short slip distance
+    mutTrainStops.add(TrainStop(endPos, 0.1, SHORT_SLIP_STOP))
```
eckter, line 141: Does it make sense as "default value" for a generic end-of-path stop? `STOP` would feel more standard

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
```diff
@@ -269,8 +271,8 @@ class STDCMPostProcessing(private val graph: STDCMGraph) {
                     TrainStop(
                         offset.meters,
                         timeData.stopTimeData[stopIndex++].currentDuration,
-                        // TODO: forward and use onStopSignal param from request
-                        isTimeStrictlyPositive(prevNode.stopDuration)
+                        // TODO: forward and use receptionSignal param from request
+                        if (isTimeStrictlyPositive(prevNode.stopDuration)) SHORT_SLIP_STOP else OPEN
```
eckter, line 275: Same as the other comment, I don't know which is the most "generic" option for stops

eckter (approved):
LGTM for core (with minor comments)

Erashin (approved):
Core lgtm!

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -83,13 +85,14 @@ export const formatSuggestedViasToRowVias = (
...
+    const { receptionSignal: _opReceptionSignal, ...filteredOp } = op;
     return {
-      ...op,
+      ...filteredOp,
       isMarginValid,
       arrival: formattedArrival,
       departure: formattedDeparture,
-      onStopSignal: onStopSignal || false,
-      name: name || t('waypoint', { id: op.opId }),
+      onStopSignal: receptionSignalToOnStopSignal(receptionSignal),
```
woshilapin, line 94: Why do we need to keep `onStopSignal` here? Shouldn't this be changed to a receptionSignal too?

woshilapin (approved):
Small API change, huge diff, looks good 😄

front/src/modules/timesStops/helpers/utils.ts
bougue-pe: There is still only a check box in the form, and its meaning is still "onStopSignal".
And the plan is to keep this checkbox and pop another one when "onStopSignal" is ✔️  to trigger short slip distance.

So the form part keeps its semantic and the app state migrates to "receptionSignal", with a translation in between.

core/src/main/kotlin/fr/sncf/osrd/api/stdcm/STDCMEndpoint.kt
bougue-pe: As discussed, this was refined to use the most restrictive, and this will be documented after this PR: https://github.com/OpenRailAssociation/osrd-website/pull/240/files

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
bougue-pe: See https://github.com/OpenRailAssociation/osrd/pull/8983#discussion_r1774971041 (resolving this conversation for now).

front/src/modules/timesStops/TimesStopsInput.tsx
clarani: Thanks :)

front/src/modules/timesStops/TimesStops.tsx
clarani: Thanks :pray:

Wadjetz (approved):

clarani (approved):

achrafmohye (approved):
LGTM 👍

Issue #8744, opened 2024-09-04
RBAC: CLI interface


leovalais:
### Description and goal

Refs #8171

Implement a CLI interface to ease role management.

CLI:

- [ ] `editoast roles add SUBJECT ROLE [ROLE+]`
- [ ] `editoast roles remove SUBJECT ROLE [ROLE+]`
- [ ] `editoast roles list SUBJECT`
- [ ] `editoast roles list-roles`

where:
* `SUBJECT` can be either a subject id (u64) or a user identity as provided by the gateway (String)
* `ROLE` is a builtin role identifier
* `list-roles` lists the builtin roles tag

### Acceptance criteria

The commands are implemented.

> [!NOTE]
> There will be a CLI to manage users and groups as well. Subject IDs will be obtainable from there.

flomonster:
- Why using `authz` in the cmd?
- Since we removed the application role only builtin should be listed and use
  - Special case: `admin` and `superuser` both grant all builtin roles to the subject
- We should have a command to list groups
- We should have a command to add a user to a group

leovalais:
> Why using authz in the cmd?

Just to scope things. I'll remove it.

> Special case: admin and superuser both grant all builtin roles to the subject

The more I think about it, the more fitting a builtin role `Superuser` or `Admin` that short-circuits all checks looks to me. This way we won't have to deal with migrations or the mutual-exclusiveness of builtin roles.

> We should have a command to list groups
We should have a command to add a user to a group

The other half is there: https://github.com/OpenRailAssociation/osrd/issues/8754

PR #8984, opened 2024-09-24
Add role management endpoints

### LINKED ISSUE #8743: RBAC: implement role management endpoints
### Description and goal

Part of #8171

Implement the following routes:

## No role control required

- [ ] `GET /authz/list_roles` ➡️ all application roles declared in the current configuration
- [ ] `GET /authz/roles` ➡️ the roles of the issuer of the request (available to every authenticated user)

### On role = `admin`

- [ ] `POST /authz/subject/:id/roles` ➡️ grants some roles to a subject
- [ ] `DELETE /authz/subject/:id/roles` ➡️ removes some roles of a subject

### Acceptance criteria

The endpoints exist.

leovalais:
> [!TIP]
> Review each commit individually.

Refs #8743

This PR lacks the following elements:

- [ ] Unit tests
    - [ ] Authorizer logic
    - [ ] PgAuthDriver
    - [ ] Endpoints
- [ ] Documentation

However I suggest we merge it like this if its content is good enough to allow @kmer2016 to work on the frontend part as soon as possible. Consequently, to also sum up the comments of this PR, we'll need to:

1. create a CLI command to grant builtin roles to a subject: #8744
2. remove application roles which are more than superfluous until we have a proper admin panel: https://github.com/OpenRailAssociation/osrd/pull/8984#issuecomment-2378510671
3. work on tests and documentation
4. improve `DbConnection` API: https://github.com/OpenRailAssociation/osrd/pull/8984#discussion_r1775160726

### DIFF
--- a/editoast/editoast_authz/src/authorizer.rs
+++ b/editoast/editoast_authz/src/authorizer.rs
@@ -1,9 +1,10 @@
+use itertools::Itertools as _;
 use std::{collections::HashSet, future::Future, sync::Arc};
 
 use tracing::debug;
 use tracing::Level;
 
-use crate::roles::{BuiltinRoleSet, RoleConfig};
+use crate::roles::{BuiltinRoleSet, RoleConfig, RoleIdentifier};
 
 pub type UserIdentity = String;
 pub type UserName = String;
@@ -27,6 +28,16 @@ pub struct Authorizer<S: StorageDriver> {
 pub trait StorageDriver: Clone {
     type BuiltinRole: BuiltinRoleSet + std::fmt::Debug;
     type Error: std::error::Error;
+
+    fn get_user_id(
+        &self,
+        user_info: &UserInfo,
+    ) -> impl Future<Output = Result<Option<i64>, Self::Error>> + Send;
+
+    fn get_user_info(
+        &self,
+        user_id: i64,
+    ) -> impl Future<Output = Result<Option<UserInfo>, Self::Error>> + Send;
 
     fn ensure_user(&self, user: &UserInfo)
         -> impl Future<Output = Result<i64, Self::Error>> + Send;
@@ -90,8 +101,21 @@ impl<S: StorageDriver> Authorizer<S> {
         }
     }
 
+    pub fn user_id(&self) -> i64 {
+        self.user_id
+    }
+
     pub fn is_superuser(&self) -> bool {
         self.roles_config.is_superuser()
+    }
+
+    /// Returns whether a user with some id exists
+    #[tracing::instrument(skip_all, fields(user_id = %user_id), ret(level = Level::DEBUG), err)]
+    pub async fn user_exists(&self, user_id: i64) -> Result<bool, S::Error> {
+        self.storage
+            .get_user_info(user_id)
+            .await
+            .map(|x| x.is_some())
     }
 
     /// Check that the user has all the required builting roles
@@ -106,6 +130,71 @@ impl<S: StorageDriver> Authorizer<S> {
         }
 
         Ok(required_roles.is_subset(&self.user_roles))
+    }
+
+    #[tracing::instrument(skip_all, fields(user_id, auth_user = %self.user, user_roles = ?self.user_roles), ret(level = Level::DEBUG), err)]
+    pub async fn infer_application_roles(
+        &self,
+        user_id: i64,
+    ) -> Result<Vec<RoleIdentifier>, S::Error> {
+        if self.is_superuser() {
+            return Ok(self.roles_config.application_roles().cloned().collect_vec());
+        }
+
+        let resolved_roles = &self.roles_config.resolved_roles;
+        let user_roles = self
+            .storage
+            .fetch_subject_roles(user_id, &self.roles_config)
+            .await?;
+
+        let app_roles = resolved_roles
+            .iter()
+            .filter(|(_, builtins)| user_roles.is_superset(builtins))
+            .map(|(app_role, _)| app_role)
+            .cloned()
+            .collect_vec();
+
+        Ok(app_roles)
+    }
+
+    #[tracing::instrument(skip_all, fields(user_id, auth_user = %self.user, user_roles = ?self.user_roles), ret(level = Level::DEBUG), err)]
+    pub async fn user_builtin_roles(
+        &self,
+        user_id: i64,
+    ) -> Result<HashSet<S::BuiltinRole>, S::Error> {
+        let user_roles = self
+            .storage
+            .fetch_subject_roles(user_id, &self.roles_config)
+            .await?;
+        Ok(user_roles.clone())
+    }
+
+    #[tracing::instrument(skip_all, fields(user_id, auth_user = %self.user, ?roles, role_config = ?self.roles_config), ret(level = Level::DEBUG), err)]
+    pub async fn grant_roles(
+        &mut self,
+        user_id: i64,
+        roles: HashSet<S::BuiltinRole>,
+    ) -> Result<(), S::Error> {
+        self.storage
+            .ensure_subject_roles(user_id, &self.roles_config, roles.clone())
+            .await?;
+        self.user_roles.extend(roles);
+        Ok(())
+    }
+
+    #[tracing::instrument(skip_all, fields(user_id, auth_user = %self.user, ?roles, role_config = ?self.roles_config), ret(level = Level::DEBUG), err)]
+    pub async fn strip_roles(
+        &mut self,
+        user_id: i64,
+        roles: HashSet<S::BuiltinRole>,
+    ) -> Result<(), S::Error> {
+        let removed_roles = self
+            .storage
+            .remove_subject_roles(user_id, &self.roles_config, roles.clone())
+            .await?;
+        tracing::debug!(?removed_roles, "removed roles");
+        self.user_roles.retain(|r| !roles.contains(r));
+        Ok(())
     }
 }
 
@@ -277,6 +366,22 @@ mod tests {
                 .cloned()
                 .collect();
             Ok(removed_roles)
+        }
+
+        async fn get_user_id(&self, user_info: &UserInfo) -> Result<Option<i64>, Self::Error> {
+            Ok(self.users.lock().unwrap().get(&user_info.identity).copied())
+        }
+
+        async fn get_user_info(&self, user_id: i64) -> Result<Option<UserInfo>, Self::Error> {
+            let users = self.users.lock().unwrap();
+            let user_info = users
+                .iter()
+                .find(|(_, id)| **id == user_id)
+                .map(|(identity, _)| UserInfo {
+                    identity: identity.clone(),
+                    name: "Mocked User".to_owned(),
+                });
+            Ok(user_info)
         }
     }
 }
--- a/editoast/editoast_authz/src/authorizer.rs
+++ b/editoast/editoast_authz/src/authorizer.rs
@@ -8,22 +8,23 @@ use crate::roles::{BuiltinRoleSet, RoleConfig};
 pub type UserIdentity = String;
 pub type UserName = String;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct UserInfo {
     pub identity: UserIdentity,
     pub name: UserName,
 }
 
+#[derive(Clone)]
 pub struct Authorizer<S: StorageDriver> {
     user: UserInfo,
     user_id: i64,
-    roles_config: Arc<RoleConfig<S::BuiltinRole>>,
+    pub roles_config: Arc<RoleConfig<S::BuiltinRole>>,
     user_roles: HashSet<S::BuiltinRole>,
     #[allow(unused)] // will be used soon
     storage: S,
 }
 
-pub trait StorageDriver {
+pub trait StorageDriver: Clone {
     type BuiltinRole: BuiltinRoleSet + std::fmt::Debug;
     type Error: std::error::Error;
 
--- a/editoast/openapi.yaml
+++ b/editoast/openapi.yaml
@@ -3454,6 +3454,25 @@ components:
           type: string
           enum:
           - editoast:authz:AuthError
+    EditoastAuthorizationErrorDbError:
+      type: object
+      required:
+      - type
+      - status
+      - message
+      properties:
+        context:
+          type: object
+        message:
+          type: string
+        status:
+          type: integer
+          enum:
+          - 500
+        type:
+          type: string
+          enum:
+          - editoast:authz:DbError
     EditoastAuthorizationErrorUnauthenticated:
       type: object
       required:
@@ -3984,6 +4003,7 @@ components:
       - $ref: '#/components/schemas/EditoastAppHealthErrorTimeout'
       - $ref: '#/components/schemas/EditoastAttachedErrorTrackNotFound'
       - $ref: '#/components/schemas/EditoastAuthorizationErrorAuthError'
+      - $ref: '#/components/schemas/EditoastAuthorizationErrorDbError'
       - $ref: '#/components/schemas/EditoastAuthorizationErrorUnauthenticated'
       - $ref: '#/components/schemas/EditoastAuthorizationErrorUnauthorized'
       - $ref: '#/components/schemas/EditoastAuthzErrorAuthz'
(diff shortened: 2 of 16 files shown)


editoast/editoast_authz/src/authorizer.rs
```diff
@@ -36,8 +48,8 @@ pub trait StorageDriver {
         roles_config: &RoleConfig<Self::BuiltinRole>,
     ) -> impl Future<Output = Result<HashSet<Self::BuiltinRole>, Self::Error>> + Send;
 
-    fn ensure_subject_roles(
-        &self,
+    fn ensure_subject_roles<'a>(
+        &'a self,
```
woshilapin: I don't understand why you would need to be explicit about the lifetimes here since it's basically what the ["lifetime elision rules"](https://doc.rust-lang.org/reference/lifetime-elision.html) already does. Maybe I'm missing something?

editoast/editoast_authz/src/authorizer.rs
```diff
@@ -89,23 +101,101 @@ impl<S: StorageDriver> Authorizer<S> {
...
+
     /// Check that the user has all the required builting roles
     #[tracing::instrument(skip_all, fields(user = %self.user, user_roles = ?self.user_roles, ?required_roles), ret(level = Level::DEBUG), err)]
     pub async fn check_roles(
         &self,
         required_roles: HashSet<S::BuiltinRole>,
     ) -> Result<bool, S::Error> {
         if self.is_superuser() {
-            tracing::debug!("role checking skipped for superuser");
+            tracing::warn!("role checking skipped for superuser");
```
woshilapin: Are we sure it won't get too noisy? You can always configure `RUST_LOG=info,editoast_authz::authorizer::check_roles=debug` to get them in DEBUG mode if you ever want to work on this specific part.

editoast/editoast_authz/src/authorizer.rs
```diff
@@ -277,5 +367,26 @@ mod tests {
...
             Ok(removed_roles)
         }
+
+        async fn get_user_id(&self, user_info: &UserInfo) -> Result<Option<i64>, Self::Error> {
+            self.users
+                .lock()
+                .unwrap()
+                .get(&user_info.identity)
+                .map(|id| Ok(Some(*id)))
+                .unwrap_or_else(|| Ok(None))
```
woshilapin: Isn't this equivalent to this?

```suggestion
            Ok(self.users
                .lock()
                .unwrap()
                .get(&user_info.identity)
                .copied())
```

editoast/openapi.yaml
```diff
@@ -3349,11 +3473,68 @@ components:
         status:
           type: integer
           enum:
-          - 401
+          - 403
+        type:
+          type: string
+          enum:
+          - editoast:authz:Unauthorized
+    EditoastAuthzErrorAuthError:
```
woshilapin: Do we really want to keep such convoluted names? Special mention for `EditoastAuthzErrorError` below...

editoast/openapi.yaml
```diff
@@ -4240,6 +4445,30 @@ components:
           type: string
           enum:
           - editoast:infra:railjson:WrongRailjsonVersionProvided
+    EditoastNoSuchUserErrorNoSuchUser:
```
woshilapin, line 4462: This name is not bad either 🤯

editoast/src/models/auth.rs
```diff
@@ -39,20 +41,44 @@ impl<B: BuiltinRoleSet + Send + Sync> StorageDriver for PgAuthDriver<B> {
...
     type Error = AuthDriverError;
 
-    #[tracing::instrument(skip_all, fields(%user), ret, err)]
+    #[tracing::instrument(skip_all, fields(%user_info), ret(level = Level::DEBUG), err)]
+    async fn get_user_id(&self, user_info: &UserInfo) -> Result<Option<i64>, Self::Error> {
+        let conn = self.pool.get().await?;
+        let id = authn_user::table
+            .select(authn_user::id)
+            .filter(authn_user::identity_id.eq(&user_info.identity))
+            .first::<i64>(conn.write().await.deref_mut())
```
woshilapin: I'm not sure that I understand why we need a `conn.write()` here, shouldn't `get_user_id` be a read-only operation?

editoast/src/models/auth.rs
```diff
@@ -39,20 +41,44 @@ impl<B: BuiltinRoleSet + Send + Sync> StorageDriver for PgAuthDriver<B> {
...
+        Ok(id)
+    }
+
+    #[tracing::instrument(skip_all, fields(%user_id), ret(level = Level::DEBUG), err)]
+    async fn get_user_info(&self, user_id: i64) -> Result<Option<UserInfo>, Self::Error> {
+        let conn = self.pool.get().await?;
+        let info = authn_user::table
+            .select((authn_user::identity_id, authn_user::name))
+            .filter(authn_user::id.eq(user_id))
+            .first::<(String, Option<String>)>(conn.write().await.deref_mut())
```
woshilapin: Same here, I'm not sure why this a `conn.write()`?

editoast/src/views/authz.rs
```diff
@@ -0,0 +1,229 @@
...
+        return Err(AuthorizationError::Unauthorized.into());
+    }
+
+    if !authorizer
+        .user_exists(user_id)
+        .await
+        .map_err(AuthzError::from)?
+    {
+        return Err(NoSuchUserError::NoSuchUser { user_id }.into());
+    }
```
woshilapin: Shouldn't we extract that into a function since all three above functions have the same boilerplate to check authorization then validity of the `user_id`?

front/public/locales/fr/errors.json
```diff
@@ -16,10 +16,16 @@
...
-    "authorization": {
+    "authz": {
       "AuthError": "Erreur interne d'authentification ou d'autorisation, veuillez réessayer ou contacter le support",
       "Unauthenticated": "Utilisateur non authentifié",
-      "Unauthorized": "Accès refusé"
+      "Unauthorized": "Accès refusé",
+      "DbError": "Erreur interne de base de données, veuillez réessayer ou contacter le support",
+      "Error": "Erreur interne d'authentification ou d'autorisation, veuillez réessayer ou contacter le support",
+      "role": {
+        "Invalid": "Libéllé de rôle inconnu"
```
woshilapin:
```suggestion
        "Invalid": "Libellé de rôle inconnu"
```

woshilapin (commented):
I hope the entire PR will be squashed because

```rust
#[cfg(test)]
mod test_app;
```
is removed in e08d2f4305321817e1bf8cf193ff6456d7bccabd and only comes back 3 commits later in 6c32f1a95c92f8fd2436044387c24ca6c670ac7c. Similar inconsistencies happen when `utoipa` annotations are added/modified but `openapi.yaml` is updated in a subsequent commit.

editoast/src/views/authz.rs
```diff
@@ -0,0 +1,229 @@
...
+    ),
+)]
+async fn list_roles(State(AppState { role_config, .. }): State<AppState>) -> Json<Vec<String>> {
+    let roles = role_config.application_roles().cloned().collect();
+    Json(roles)
+}
+
+#[derive(serde::Serialize, utoipa::ToSchema)]
+struct Roles {
+    builtin: Vec<String>,
```
leovalais: Should return a `Vec<BuiltinRole>`

editoast/src/views/authz.rs
```diff
@@ -0,0 +1,229 @@
...
+    application: Vec<String>,
```
woshilapin: As discussed, I would rather not expose `application` (yet?) as there is no obvious use case for it.

editoast/src/views/authz.rs
```diff
@@ -0,0 +1,229 @@
...
+    #[error("No user with ID {user_id} found")]
+    #[editoast_error(status = 404)]
+    NoSuchUser { user_id: i64 },
+}
+
+#[utoipa::path(
+    get, path = "",
+    tag = "authz",
+    responses(
+        (status = 200, description = "List of supported application roles", body = Vec<String>, example = json!(["Superuser", ""])),
```
flomonster:
```suggestion
        (status = 200, description = "List of supported application roles", body = Vec<String>, example = json!(["Superuser"])),
```

flomonster (commented):
Great PR.

I think it's going to be a pain not to have a “role config” in this repository but in the deployment repository. It's great that it can be overridden, but knowing that you're likely to want to update it regularly (until the whole thing has stabilized and settled down), it's error prone.

editoast/src/views/authz.rs
leovalais: I factorized the `user_exists` part but not the `check_roles` header to maintain consistency with the other handlers (at least until we have the middleware). I let you resolve if it's ok.

editoast/editoast_authz/src/authorizer.rs
leovalais: Yep, I fiddled with the parameters a bit and forgot to remove the lifetime... Done in https://github.com/OpenRailAssociation/osrd/pull/8984/commits/5c3222d8c34b931d34674bb1ef3a7b1ff5022cdd

editoast/editoast_authz/src/authorizer.rs
leovalais: I didn't mean to commit that, reverted in https://github.com/OpenRailAssociation/osrd/pull/8984/commits/5c3222d8c34b931d34674bb1ef3a7b1ff5022cdd

editoast/openapi.yaml
leovalais: It's not ideal to say the least. The schema is generated by the editoast error introspection in order for the front tooling to list the required translation keys. But the schema name is never used. The whole process will change once we settle on how we want to treat errors from now on.

This particular variant will also disappear if we get the role-checking middleware before the error revamp.

For these reasons I'm not sure it's worth doing something about it. Wdyt?

editoast/openapi.yaml
leovalais: https://github.com/OpenRailAssociation/osrd/pull/8984#discussion_r1775625670

editoast/src/models/auth.rs
leovalais: This is how we access the inner `&mut AsyncPgConnection` that's now wrapped in an `RwLock` in our now opaque `DbConnection`: cf. https://github.com/OpenRailAssociation/osrd/pull/8712.

I suggested implementing `Future` for the connection so that we only have to write `conn.await` instead.

editoast/src/models/auth.rs
leovalais: https://github.com/OpenRailAssociation/osrd/pull/8984#discussion_r1775641503

leovalais:
> I hope the entire PR will be squashed

It will. I just split it to help with the review. I may keep a few commits (especially for the stdcm env part), but I'll split it right this time 😁 

> I think it's going to be a pain not to have a “role config” in this repository but in the deployment repository. It's great that it can be overridden, but knowing that you're likely to want to update it regularly (until the whole thing has stabilized and settled down), it's error prone.

Agreed. I am becoming increasingly convinced that we don't need application roles after all. (Or at least until we have a proper administration panel.) Maybe it wouldn't be so bad to remove them since we are the ones who will grant roles anyway.

flomonster:
> Agreed. I am becoming increasingly convinced that we don't need application roles after all. (Or at least until we have a proper administration panel.) Maybe it wouldn't be so bad to remove them since we are the ones who will grant roles anyway.

Another argument is that with the current implementation, changing the scope of an application role won't change user roles. (Except if we do a proper migration). It reduces the usefulness of this object in my opinion.

editoast/openapi.yaml
woshilapin: I understand it would probably imply significant modification, not in the scope of this PR. I'm wondering if there isn't an easy path forward, renaming all enums' variants that are named something like `AuthorizationError::AuthError` into `AuthorizationError::Auth` since we most likely will never use the variant without the enum's name and therefore, we will know it's about an error in the code?

editoast/src/models/auth.rs
woshilapin: My point being, why not `conn.read()`? We might have a design problem that is really misleading in understanding what's going on. If I understand correctly, the `write` basically forward to the `write_owned()` of `RwLock` so I guess there is a possibility to do the same for a `read()`. That said, I didn't test that a read-only would allow us to correctly query the database. If that is not possible, maybe we have a naming problem in the function `write()` 🤷

editoast/src/views/authz.rs
woshilapin: I would really factorize the entirety of it because these 3 functions will actually always require the same code and we do want that, no? My point being if we change it at some point in the future, we might forget to change the 2 others and I feel there is a higher chance that we do indeed want to update all 3 every time, what do you think?

leovalais: Alright, done in [`01fe017` (#8984)](https://github.com/OpenRailAssociation/osrd/pull/8984/commits/01fe0173f5317adea09f4564805f231ef67f9c1a)

leovalais: As discussed, reverted since the required roles slightly differ in these endpoints.

editoast/openapi.yaml
leovalais: Removed the `Error` suffix.

editoast/openapi.yaml
leovalais: Not sure what I can do about this though (`NoSuchUser::NoUserIdeed`??)

editoast/src/models/auth.rs
leovalais: I'm not sure we can get an `&mut conn` from an `OwnedReadLockGuard`. But we'll likely never use an immutable reference to the connection so maybe renaming `DbConnection::write` into `DbConnection::use` would be less confusing?

In any case this is work for another PR since it will impact many unrelated files. I still believe that the most convenient form would be `conn.await?: &mut AsyncPgConnection` since it's shorter and doesn't rely on the internals of our wrapper.

woshilapin: OK, agreed with everything you said. Since it's more a naming problem or a refactor problem, I'm resolving.

editoast/openapi.yaml
woshilapin: Do we need a `enum` since it's only one variant. Won't having a `struct` instead solve the problem? If not, feel free to resolve anyway, I don't see an easy solution.

leovalais: `derive(EditoastError)` only works on `enums` 😢
It's new version will definitely have to though.

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -1508,6 +1577,26 @@ export type PostWorkSchedulesProjectPathApiArg = {
     work_schedule_group_id: number;
   };
 };
+export type BuiltinRole =
```
SharglutDev, line 1580: `Write` means also update/delete or it will come later ?

leovalais: Yes `Write` = `create + update + delete`

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -826,6 +864,37 @@ const injectedRtkApi = api
...
+  };
+export type GetAuthzRolesMeApiArg = void;
+export type GetAuthzRolesByUserIdApiResponse = /** status 200 List the roles of a user */ {
+  builtin: BuiltinRole[];
+};
+export type GetAuthzRolesByUserIdApiArg = {
+  /** A user ID (not to be mistaken for its identity, cf. editoast user model documentation) */
+  userId: number;
+};
+export type PostAuthzRolesByUserIdApiResponse = unknown;
```
SharglutDev, line 882: Don't we want a success message for when we will have an admin interface ?

leovalais: I'd say `204 NO CONTENT` is enough of a response in this case since the sole purpose of this endpoint is to grant a bunch of roles. In any case this whole API is likely to change when we'll implement the admin panel. But in the meantime, I figured the status code would be enough especially since the frontend will never call the endpoint (roles will be granted manually by the ops/support). Wdyt?

SharglutDev: Totally fine yes :)

flomonster (changes requested):
Tested and not working. No user is created when using the application.

flomonster (approved):
LGTM (tested).
Before using it in prod we need:
- Add tests
- Remove application role
- Add CLI

SharglutDev (approved):
Lgtm (not tested)

PR #8985, opened 2024-09-24
Editoast: import and send to core temporary speed limits (LTVs)

### LINKED ISSUE #8684: Update OSRD database with LTV data
Update the database schema so that it contains the relevant LTVs information for core to take them into account when building MRSPs.

### LINKED ISSUE #8685: Pass LTV data from editoast to core
Add to STDCM requests sent by editoast to core the list of temporary speed limits which are active between the earliest possible departure time and the latest possible arrival time (with margins).

Sh099078:
- Add a temporary speed limit table to the database.
- Add a create endpoint for temporary speed limits.
- Update STDCM requests to include the list of temporary speed limits which are active between the query earliest possible departure time and its latest possible arrival time, margins included.

Fixes https://github.com/OpenRailAssociation/osrd/issues/8684
Fixes https://github.com/OpenRailAssociation/osrd/issues/8685


editoast/src/core/stdcm.rs
```diff
@@ -90,6 +93,15 @@ pub struct LightWorkSchedule {
     pub track_ranges: Vec<UndirectedTrackRange>,
 }
 
+/// Lighter description of a work schedule with only the informations relevant for core
+#[derive(Debug, Serialize)]
+pub struct LightTemporarySpeedLimit {
+    // Speed limitation in km/h
+    pub speed_limit: i64,
```
flomonster: - Use `///` for documentation.
- Speed unit is `m/s` using `f64`

```suggestion
    /// Speed limitation in m/s
    pub speed_limit: f64,
```

editoast/src/models/temporary_speed_limits.rs
```diff
@@ -0,0 +1,29 @@
...
+#[derive(Debug, Serialize, Clone, Model)]
+#[model(table = temporary_speed_limit)]
+#[model(gen(ops = cr, batch_ops = c, list))]
+pub struct TemporarySpeedLimit {
+    pub id: i64,
+    pub start_date_time: NaiveDateTime,
+    pub end_date_time: NaiveDateTime,
+    pub speed_limit: i64,
+    #[model(json)]
+    pub track_ranges: Vec<DirectionalTrackRange>,
```
flomonster, line 27: I'm not sure if we should use `DirectionalTrackRange` or `TrackRange`

flomonster (changes requested):
We should add a table `speed_limit_group` like `work_schedule_group`. So speed limits are grouped behind a single ID.

editoast/src/models/temporary_speed_limits.rs
Sh099078: ~~I have not imported the track ranges yet but the speed limitation signals have a ¨sens de lecture¨ which is either C (croissant ?) or D (décroissant ?). Do you think temporary speed limitations should be directional @axrolld ?~~

Got confused with the structure names

editoast/src/core/stdcm.rs
```diff
@@ -56,6 +57,8 @@ pub struct STDCMRequest {
     pub margin: Option<MarginValue>,
     /// List of planned work schedules
     pub work_schedules: Vec<LightWorkSchedule>,
+    // List of applicable temporary speed limits between the train departure and arrival
+    pub temporary_speed_limits: Vec<LightTemporarySpeedLimit>,
```
woshilapin: Seeing this, I would expect the code on `core` to change too, notably `STDCCMRequestV2.kt` at the very least.

editoast/src/core/stdcm.rs
```diff
@@ -90,6 +93,15 @@ pub struct LightWorkSchedule {
     pub track_ranges: Vec<UndirectedTrackRange>,
 }
 
+/// Lighter description of a work schedule with only the informations relevant for core
+#[derive(Debug, Serialize)]
+pub struct LightTemporarySpeedLimit {
```
woshilapin: Not sure of the name here. To be honest, we already use `Light` in `editoast` to return something with less information because the original object is too huge (see `RollingStock` which contains `EffortCurve`). But I was trying to understand at first why we would need to do that in API's `editoast`. In this case, it's not really a light version, it's just the exact object for `core` (which happens to be a subset of the one exposed in `editoast` API). So I would just name `TemporarySpeedLimit` since we are in `core::stdcm` module and that we are describing the API to `core`, not something relative to `editoast` API. What do you think?

editoast/src/views/timetable/stdcm.rs
```diff
@@ -568,6 +650,29 @@ async fn build_work_schedules(
...
+    let selection_settings: SelectionSettings<TemporarySpeedLimit> = SelectionSettings::new();
+    let applicable_speed_limits: Vec<LightTemporarySpeedLimit> =
+        TemporarySpeedLimit::list(conn, selection_settings)
+            .await?
+            .into_iter()
+            .filter(|speed_limit| {
+                !(end_date_time <= speed_limit.start_date_time.and_utc()
+                    || speed_limit.end_date_time.and_utc() <= start_date_time)
+            })
+            .map(|speed_limit| speed_limit.into())
```
woshilapin: If possible, it's usually better to be more explicit. As a developer, reading `.into()` just make me think "into, ok, but into what?".

```suggestion
            .map(LightTemporarySpeedLimit::from)
```

It will also have the advantage to help type inference and so, you won't need `: Vec<LightTemporarySpeedLimit>` I believe.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -410,6 +425,60 @@ fn get_earliest_departure_time(
...
+        ..
+    }) = get_last_step_with_timing(data)
+    {
+        Some(
+            timing.arrival_time
+                + Duration::milliseconds(timing.arrival_time_tolerance_after as i64),
+        )
+    } else {
+        None
+    };
```
woshilapin: Since you have `Option` all along the transformations, it might be better to express that series of operations in a more functional way (not tested so I might have miss something)?

```suggestion
    let result_from_last_step: Option<DateTime<Utc>> = get_last_step_with_timing(data)
      .and_then(|pathfinding_item| pathfinding_item.timing_data)
      .map(|timing| timing.arrival_time + Duration::milliseconds(timing.arrival_time_tolerance_after as i64));
```

Same below for `result_from_first_step`.

editoast/src/views/temporary_speed_limits.rs
```diff
@@ -0,0 +1,216 @@
...
+    "/temporary_speed_limits" => create,
+}
+
+editoast_common::schemas! {
+    TemporarySpeedLimitCreateForm,
+    TemporarySpeedLimitCreateResponse,
+    TemporarySpeedLimitItemForm,
+}
+
+#[derive(Serialize, Derivative, ToSchema)]
```
woshilapin: It doesn't seem to be used so better remove it.

```suggestion
#[derive(Serialize, ToSchema)]
```

editoast/src/views/temporary_speed_limits.rs
```diff
@@ -0,0 +1,216 @@
...
+        TemporarySpeedLimit::changeset()
+            .start_date_time(self.start_date_time)
+            .end_date_time(self.end_date_time)
+            .track_ranges(self.track_ranges)
+            .speed_limit(self.speed_limit)
+            .obj_id(self.obj_id)
+    }
+}
+
+impl<'de> Deserialize<'de> for TemporarySpeedLimitItemForm {
```
woshilapin, line 67: That's unfortunate that we have to manually implement `Deserialize`, adding that much more code, only to check `end_date_time > start_date_time`. I found a way (not tested) to reduce the quantity of code with [this trick](https://stackoverflow.com/a/70293354) (which I honestly don't completely understand). I don't know what you think about it? In any case, feel free to resolve either way.

```rust
#[derive(Serialize, Deserialize, ToSchema)]
#[serde(remote = "Self")]
struct TemporarySpeedLimitItemForm {
    start_date_time: NaiveDateTime,
    end_date_time: NaiveDateTime,
    track_ranges: Vec<DirectionalTrackRange>,
    speed_limit: i64,
    obj_id: String,
}

impl<'de> Deserialize<'de> for TemporarySpeedLimitItemForm {
    fn deserialize<D>(deserializer: D) -> StdResult<TemporarySpeedLimitItemForm, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        let this = Self::deserialize(deserializer)?;
        // Check dates
        if this.end_date_time > this.start_date_time {
            Ok(this)
        } else {
            Err(SerdeError::custom(format!(
                "The temporary_speed_limit start date '{}' must be before the end date '{}'",
                this.start_date_time, this.end_date_time
            )))
        }
    }
}
```

editoast/src/views/temporary_speed_limits.rs
```diff
@@ -0,0 +1,216 @@
...
+}
+
+#[derive(Serialize, Deserialize, ToSchema)]
+struct TemporarySpeedLimitCreateForm {
+    speed_limits: Vec<TemporarySpeedLimitItemForm>,
+}
+
+#[derive(Serialize, Deserialize, ToSchema)]
+struct TemporarySpeedLimitCreateResponse {
+    pub created_speed_limits_ids: Vec<i64>,
```
woshilapin: Prefixing with `created_` is a pattern we don't have in any of our API, I would drop it. In fact, I would probably just put `temporary_speed_limits_ids` or even just `ids` (since we know which endpoint is called anyway). I don't mind either of the two latter solutions (I believe we have kind of both in our API anyway).

editoast/src/views/temporary_speed_limits.rs
```diff
@@ -0,0 +1,216 @@
...
+        let speed_limit_id = response
+            .created_speed_limits_ids
+            .first()
+            .expect("Temporary speed limit query did not return any temporary speed limit ids");
+
+        let speed_limit = TemporarySpeedLimit::retrieve(&mut pool.get_ok(), *speed_limit_id)
+            .await
+            .expect("Failed to retrieve temporary speed limit");
+
+        assert!(speed_limit.is_some());
```
woshilapin: I would at least assert the `obj_id` is the one we send (rather unique so if we assert that, we're sure it's not some other objects, I don't trust that we can't have data races between our tests since they're not isolated and use the same DB).

woshilapin (changes requested):
Great job. I left a bunch of comments, mostly suggestions, naming and nitpick. But at the very least, I believe the `core` part should also be provided as part of this PR (not the entire implementation but at least the update of the exposed API).

editoast/src/views/temporary_speed_limits.rs
Sh099078: Thank you for your suggestions. Following a change request from @flomonster, the endpoint will return the id of the newly created `TemporarySpeedLimitGroup` instead: I am updating this conversation to resolved. Otherwise I would have followed your last suggestion and renamed the identifiers vector to `ids` since the nature of the identifiers is indeed implied by the context.

editoast/src/core/stdcm.rs
Sh099078: Thank you for your comment. The `LightTemporarySpeedLimit` is indeed the exact object that core needs but it is also a lighter version of the `TemporarySpeedLimit` struct in `models`: whether it is a light version or not depends if we take `core` or `editoast` point of view. If we rename it to `TemporarySpeedLimit` I think it would lead to confusion and would require to make aliases for one of the two structs named this way that are used in `stdcm.rs`: we might aswell keep them named differently.  Maybe we could rename it to something like `CoreTemporarySpeedLimit` ? That name would be more explicit than `LightTemporarySpeedLimit` about the fact that this is the full `core` representation of a temporary speed limit. If we rename it we should also rename in the same manner the `LightWorkSchedule` just above it in stdcm.rs. What do you think ?

editoast/src/core/stdcm.rs
Khoyo: We do not `deny_unknown_fields` in core, so this should work anyway... ^^'

editoast/src/core/stdcm.rs
Khoyo: The fact that it is core::TemporarySpeedLimit makes it clear that it is exactly what core expects. If the name clash is to inconvenient, maybe then core::CoreTemporarySpeedLimit should be OK, but it don't think it's necessary.

(There is a `core::STDCMRequest`, not a `core::CoreSTDCMRequest`)

editoast/src/views/temporary_speed_limits.rs
Khoyo: I would be really stumped if this trick does work... (My first instinct is that rustc will complain about conflicting impls of Deserialize)

I also think we should write some guidelines about how to perform input data validation in editoast...

editoast/src/core/stdcm.rs
woshilapin: Didn't notice for `LightWorkSchedule` so yes, indeed, if you'd be kind enough to change both, that'd be nice. For the naming, `CoreTemporarySpeedLimit` or `TemporarySpeedLimit` both work, even if I believe that the latter is better. Aliasing is a problem of the consumer, but I don't think that should drive the design of the API. Also, `clippy` tends to reject naming things like `core::CoreTemporarySpeedLimit` (see [`module_name_repetitions`](https://rust-lang.github.io/rust-clippy/master/index.html#/module_name_repetitions) but it is a `pedantic` category).

editoast/migrations/2024-09-06-153144_add_ltv_table/down.sql
woshilapin, line 1: You didn't update the `down.sql` migration since https://github.com/OpenRailAssociation/osrd/pull/8985/commits/ab8c4c202340396a24398d024f708ad5025d0186. Probably need to drop the indexes and the new `temporary_speed_limit_group`.

editoast/src/views/timetable/stdcm.rs
Sh099078: My bad, the if let + pattern matching version is hard on the eyes indeed. Switched to mappings.

editoast/migrations/2024-09-06-153144_add_ltv_table/down.sql
Sh099078: My bad, I added the `temporary_speed_limit_group` drop to `down.sql`. The indexes should be deleted automatically when their table is dropped.

editoast/src/views/temporary_speed_limits.rs
Sh099078: I think I'll leave it as is until there is a consensus on the way we parse / validate data.

editoast/src/models/temporary_speed_limits.rs
Sh099078: Renamed core `TrackRange` into `DirectionalTrackRange` and used it in `core` `TemporarySpeedLimit`.

editoast/src/core/stdcm.rs
Sh099078: If if does not break core I would rather update the core part in another branch that updates STDCM with the temporary speed limits. Are you fine with me resolving that conversation ?

editoast/src/core/stdcm.rs
```diff
@@ -55,7 +56,9 @@ pub struct STDCMRequest {
     /// Margin to apply to the whole train
     pub margin: Option<MarginValue>,
     /// List of planned work schedules
-    pub work_schedules: Vec<LightWorkSchedule>,
+    pub work_schedules: Vec<WorkSchedule>,
+    // List of applicable temporary speed limits between the train departure and arrival
```
flomonster: You can do a small check: `git diff dev | grep '[^/]//[^/]'`.

```suggestion
    /// List of applicable temporary speed limits between the train departure and arrival
```

editoast/src/views/timetable/stdcm.rs
```diff
@@ -91,7 +93,7 @@ pub struct STDCMRequestPayload {
     /// Specifies how long the total run time can be in milliseconds
     /// Deprecated, first step data should be used instead
     maximum_run_time: Option<u64>,
-    /// Train categories for speed limits
+    /// Train categories for speed limit
```
flomonster: Note: We should rename the field too.
We can leave a TODO and fix it in another PR.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -234,17 +236,29 @@ async fn stdcm(
...
+        earliest_departure_time,
+        latest_simulation_end,
+    );
 
     // 4. Parse stdcm path items
     let path_items = parse_stdcm_steps(conn, &stdcm_request, &infra).await?;
 
+    // 5. Get applicable temporary speed limits
+    let temporary_speed_limits =
+        build_temporary_speed_limits(conn, earliest_departure_time, latest_arrival_time).await?;
```
flomonster: Like the `build_work_schedule_group_id`, we should filter given an **optional** group id.

> [!NOTE]
> In a following PR, we must add the `speed_limit_group_id` to the `STDCMSearchEnvironment`.

editoast/migrations/2024-09-06-153144_add_ltv_table/up.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE temporary_speed_limit_group(
+    id int8 PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+    creation_date timestamptz NOT NULL,
+    name varchar(255) NOT NULL UNIQUE
+);
+
+CREATE TABLE temporary_speed_limit (
+    id int8 PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+    start_date_time timestamptz NOT NULL,
+    end_date_time timestamptz NOT NULL,
```
leovalais, line 10: You can add a `CHECK` constraint for `end_date_time > start_date_time`. https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-CHECK-CONSTRAINTS

flomonster (changes requested):
A few things to fix and I think we're good.

editoast/src/views/temporary_speed_limits.rs
```diff
@@ -0,0 +1,268 @@
...
+    speed_limits: Vec<TemporarySpeedLimitItemForm>,
+}
+
+#[derive(Serialize, Deserialize, ToSchema)]
+struct TemporarySpeedLimitCreateResponse {
+    pub group_id: i64,
+}
+
+impl TemporarySpeedLimitItemForm {
+    pub fn into_temporary_speed_limit_changeset(
```
leovalais:
```suggestion
    fn into_temporary_speed_limit_changeset(
```

editoast/src/views/temporary_speed_limits.rs
leovalais: > The first approach doesn't quite work as is - conflicting implementations of trait `Deserialize<'_>` for type `Params` – 
[Julian](https://stackoverflow.com/users/252723/julian)
 Commented Jan 26 at 21:02

leovalais: There's multiple ways we validate input rn. The manual implementation of Deserialize is the preferred approach in `editoast_schemas`. In other places, it's more a "up to discretion" kind of thing :D

editoast/src/views/temporary_speed_limits.rs
```diff
@@ -0,0 +1,268 @@
...
+            track_ranges,
+            speed_limit,
+            obj_id,
+        })
+    }
+}
+
+#[derive(Debug, Error, EditoastError)]
+#[editoast_error(base_id = "temporary_speed_limit")]
+pub enum WorkScheduleError {
```
leovalais:
```suggestion
enum WorkScheduleError {
```

editoast/src/views/temporary_speed_limits.rs
```diff
@@ -0,0 +1,268 @@
...
+    #[error("Name '{name}' already used")]
+    #[editoast_error(status = 400)]
+    NameAlreadyUsed { name: String },
+}
+
+pub fn map_diesel_error(e: InternalError, name: impl AsRef<str>) -> InternalError {
```
leovalais:
```suggestion
fn map_diesel_error(e: InternalError, name: impl AsRef<str>) -> InternalError {
```

Once we decide how to handle errors in editoast, we'll have to rethink how we forward pg errors in ModelV2's API. Ideally the constraint name matching should be done in `editoast_models`. Nothing you can do about it rn though, feel free to resolve.

editoast/src/views/temporary_speed_limits.rs
```diff
@@ -0,0 +1,268 @@
...
+#[utoipa::path(
+    post, path = "",
+    tag = "temporary_speed_limits",
+    request_body = TemporarySpeedLimitCreateForm,
+    responses(
+        (status = 201, body = TemporarySpeedLimitCreateResponse, description = "The id of the created temporary speed limit group." ),
+    )
+)]
+async fn create(
+    State(app_state): State<AppState>,
```
leovalais:
```suggestion
    State(db_pool): State<DbConnectionPoolV2>,
```

editoast/src/views/temporary_speed_limits.rs
```diff
@@ -0,0 +1,268 @@
...
+
+#[utoipa::path(
+    post, path = "",
+    tag = "temporary_speed_limits",
+    request_body = TemporarySpeedLimitCreateForm,
+    responses(
+        (status = 201, body = TemporarySpeedLimitCreateResponse, description = "The id of the created temporary speed limit group." ),
+    )
+)]
+async fn create(
```
leovalais: The endpoint needs documentation: it's not straightforward from its path that it also always creates a new speed limit group.  Wdyt about `POST /temporary_speed_limit_group`?

editoast/src/views/temporary_speed_limits.rs
```diff
@@ -0,0 +1,268 @@
...
+                }]
+            }
+        ));
+
+        // Speed limit group checks
+
+        let TemporarySpeedLimitCreateResponse { group_id } = app
+            .fetch(request)
+            .assert_status(StatusCode::OK)
+            .json_into::<TemporarySpeedLimitCreateResponse>();
```
leovalais:
```suggestion
            .json_into();
```

editoast/src/views/work_schedules.rs
```diff
@@ -14,7 +14,7 @@ use std::result::Result as StdResult;
 use thiserror::Error;
 use utoipa::ToSchema;
 
-use crate::core::pathfinding::TrackRange as CoreTrackRange;
+use crate::core::pathfinding::DirectionalTrackRange as CoreTrackRange;
```
```suggestion
use crate::core::pathfinding::DirectionalTrackRange;
```

nitpick: the aliasing is there to avoid conflict with `editoast_schemas::TrackRange`. Now that the names naturally don't conflict anymore the `as` becomes superfluous.

editoast/src/views/temporary_speed_limits.rs
```diff
@@ -0,0 +1,268 @@
...
+
+crate::routes! {
+    "/temporary_speed_limits" => create,
+}
+
+editoast_common::schemas! {
+    TemporarySpeedLimitCreateForm,
+    TemporarySpeedLimitCreateResponse,
+    TemporarySpeedLimitItemForm,
+}
```
leovalais: These schemas are only used in one place, so you can `schema(inline)` them.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -234,17 +236,29 @@ async fn stdcm(
...
+
     // 5. Build STDCM request
```
```suggestion
    // 6. Build STDCM request
```

;p

editoast/src/views/timetable/stdcm.rs
```diff
@@ -411,6 +426,47 @@ fn get_earliest_departure_time(
     )
 }
 
+// Return the latest time at which the train may arrive, margins included.
```
```suggestion
/// Return the latest time at which the train may arrive, margins included.
```

`///` are doc comments, `//` and `/* */` are regular comments

editoast/src/views/timetable/stdcm.rs
```diff
@@ -569,6 +638,28 @@ async fn build_work_schedules(
...
+    }
+    let selection_settings: SelectionSettings<TemporarySpeedLimit> = SelectionSettings::new();
+    let applicable_speed_limits = TemporarySpeedLimit::list(conn, selection_settings)
+        .await?
+        .into_iter()
+        .filter(|speed_limit| {
+            !(end_date_time <= speed_limit.start_date_time.and_utc()
+                || speed_limit.end_date_time.and_utc() <= start_date_time)
+        })
+        .map(CoreTemporarySpeedLimit::from)
```
leovalais: `.map_into()` may work here if rust can infer the item type

leovalais (commented):
Great work! A few comments:

editoast/src/views/timetable/stdcm.rs
Sh099078: Apparently several speed limits can apply to a train category so it seems we should keep the `limits` plural. However it might need to be renamed to `train category` singular if a train can only have at most one category ? That's not linked to my PR: I am reverting my change and adding a TODO if that's fine with you.

flomonster: A "code composition" is modeled by a speed limit tag. A train has only one speed limit tag, but a speed limit tag can be missing and in this case a fallback system is used.

editoast/src/views/temporary_speed_limits.rs
Sh099078: LGTM. I copied the naming style used for the work schedules endpoint, do you want me to rename it aswell ?

leovalais: You mean rename the handler function? Sure, `create` is a bit vague.

editoast/src/views/work_schedules.rs
Khoyo: Nah I believe the names will conflict again... The types were leaked to the front, so the renaming becomes a bit involved and better reserved for another PR

Sh099078: I reverted the renaming of `core::pathfinding::TrackRange` to `DirectionalTrackRange` due to the build errors it caused in `front` (after updating the api and generating the typescript types succeeded). Currently the aliasing is still needed to distinguish `core` track ranges and `schemas` track ranges. Should I resolve this conversation ?

leovalais: > I reverted the renaming of core::pathfinding::TrackRange to DirectionalTrackRange due to the build errors it caused in front

Too bad. Resolving this then.

editoast/src/views/temporary_speed_limits.rs
Sh099078: I meant that another similar endpoint exists for the work schedules. Should I rename them both this way ?

leovalais: I'd rather not. The front-end doesn't use it I believe but you'll then have to update the work schedule import. Fine by me, but in another PR.

editoast/src/views/temporary_speed_limits.rs
```diff
@@ -0,0 +1,313 @@
...
+#[derive(Serialize, Deserialize, ToSchema)]
+struct TemporarySpeedLimitCreateForm {
+    speed_limit_group_name: String,
+    #[schema(inline)]
+    speed_limits: Vec<TemporarySpeedLimitItemForm>,
+}
+
+#[derive(Serialize, Deserialize, ToSchema)]
+struct TemporarySpeedLimitCreateResponse {
+    pub group_id: i64,
```
leovalais:
```suggestion
    group_id: i64,
```

flomonster (approved):
Great PR! LGTM

woshilapin (approved):
Thanks for all the work and follow-up.

leovalais (approved):
Thanks for the modifications!

clarani (approved):
LGTM :white_check_mark:

PR #2969, opened 2023-01-24
docker: front: Run dependabot only for security vulnerabilities

bloussou:
As explained in https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit, `. Use open-pull-requests-limit to change this limit. This also provides a simple way to temporarily disable version updates for a package manager. This option has no impact on security updates, which have a separate, internal limit of ten open pull requests.`

close https://github.com/DGEXSolutions/osrd/issues/2968

PR #8990, opened 2024-09-25
front: re-enable dependabot for direct dependencies

emersion:
In #2969, dependabot got disabled because it was too spammy. However, disabling dependency updates makes us fall behind pretty hard and makes version upgrades quite painful. Instead of gradually integrating new dependency upgrades and finding bugs as they arise, we're forced to make a huge commit upgrading many packages at once. Regressions are harder to narrow down and humans have to remember and volunteer to perform this unrewarding upgrade task.

As a middle ground, re-enable dependabot updates for direct dependencies only, and group all updates in a single weekly PR. This should make it less spammy than enabling it for all dependencies (transitive dependencies included), while at the same time allowing us to keep up with new versions of libraries we're using.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -23,12 +23,14 @@ updates:
   - package-ecosystem: "npm"
     directory: "/front/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
+    groups:
+      all:
+        patterns:
+          - "*"
     commit-message:
       prefix: "front:"
-    # Security updates only:
-    # https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates#overriding-the-default-behavior-with-a-configuration-file
-    open-pull-requests-limit: 0
+    open-pull-requests-limit: 100
     labels:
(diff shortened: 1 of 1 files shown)


anisometropie:
isn’t there a way to batch all the bumps into one single PR, like every week one big bump PR with all the new cool stuff?

woshilapin (approved):
How about also changing the frequency to "weekly" instead of "daily", this might also help with the "too noisy" problem and still be acceptable?

emersion:
Switched to weekly updates grouped in a single PR.

emersion:
If it's a single weekly PR, maybe we can re-enable updates for transitive dependencies.

Khoyo (approved):

anisometropie (approved):

anisometropie:
> If it's a single weekly PR, maybe we can re-enable updates for transitive dependencies.

I’d say go for it

emersion:
Alrighty

flomonster (approved):

Wadjetz (approved):

Issue #8868, opened 2024-09-13
core: output speed-limit tag warning even when train-speed is used


bougue-pe:
When the RS speed limit is below the one provided by the infrastructure and a speed-limit tag is associated to the simulation, no information about the tag used to compute infra max-speed (Given, Fallback, Unknown) is provided.
Which is a problem, especially for the warning (Unknown) case that gets hidden.

The MRSP value shouldn't change, but the source should be provided (maybe rename to "infra speed-limit source"?).

EDIT: after discussion with @loic-hamelin, only warning should be kept (see AC)

AC:

- [ ] When a speed-limit tag is provided on the train, but not present on a portion of the path, a warning should appear in the speed-speed chart in all cases:
  - [ ] when the RS max speed is the limiting one used in the MRSP (KO for now)
  - [ ] when the max speed limiting is from the infra (already working)

Issue #8959, opened 2024-09-23
Margin bugs and input row deletion bugs


louisgreiner:
### What happened?

I.
Margin of `0min/100km` still computes margin

![image](https://github.com/user-attachments/assets/193564e7-524f-45cf-be7f-00c7b54765a4)

II.
No margin defined (input table):
![image](https://github.com/user-attachments/assets/30cea75c-d403-4b98-8cd8-f15bd36f75dd)

But still, some margins are somehow displayed (output table):
![image](https://github.com/user-attachments/assets/43aefef3-c6d2-402b-bd3d-d5993363e9f2)

III.
Input departure time is bugged when try to update an existing value

https://github.com/user-attachments/assets/fdbf2d51-40d1-4ffe-9405-1d462d3c2f47

IV.
Multi selected fields modification is bugged

https://github.com/user-attachments/assets/74c09e44-5942-4329-967c-8b7c26090056

V.
Delete the value of a stopping time field behaves weirdly
(cf. previous video)

### How can we reproduce it (as minimally and precisely as possible)?

1. open input table
2. play with it
3. open output table

### On which environments the bug occurs?

Recette (SNCF)

### On which browser the bug occurs?

Firefox

### OSRD version (top right corner `Account` button > `Informations`)

staging fbf8e3e

Synar:
Issue II solved in #9375

Edit : One exception is for the departure, where a theoretical margin of 0% seems to be assumed by the sim, despite not being defined in the input table. Is this what we want?

Edit edit: Actually when deleted, the theoretial margin for the departure is autoset to 0%, but values of 0% are then hidden. Since there is still a defined value for the th margin, the other margins are then displayed. I don't know whether this is working as intended or not.

PR #8993, opened 2024-09-25
core: operational-studies: add safety speed to mrsp

### LINKED ISSUE #8810: Core: operational-studies: amend MRSP to consider safety speed of approach & short slip distance
Main part of https://github.com/osrd-project/osrd-confidential/issues/305

Depends of #8809 

Lower MRSP speed to 30 km/h 200m before signal if on closed signal.
Also lower to 10 km/h 100m before signal if short slip distance is triggered.

~Trigger the 2 limitations automatically before a buffer stop.~ see #9283 

+ tests

eckter:
Fix #8810 & #8868

1. refactor how MRSP attributes are handled. The main difference is the introduction of the `HasMissingTag` attribute, which is kept even if the speed limit comes from a different source (#8868).
1. Compute safety speeds on the new path and use them as parameters for the MRSP (#8810)
2. Adapt and add unit tests

### DIFF
--- a/core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
+++ b/core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
@@ -1,0 +1,100 @@
+package fr.sncf.osrd.standalone_sim
+
+import fr.sncf.osrd.api.FullInfra
+import fr.sncf.osrd.api.api_v2.standalone_sim.SimulationScheduleItem
+import fr.sncf.osrd.conflicts.TravelledPath
+import fr.sncf.osrd.railjson.schema.schedule.RJSTrainStop
+import fr.sncf.osrd.sim_infra.api.*
+import fr.sncf.osrd.sim_infra.impl.ChunkPath
+import fr.sncf.osrd.utils.DistanceRangeMap
+import fr.sncf.osrd.utils.distanceRangeMapOf
+import fr.sncf.osrd.utils.indexing.StaticIdx
+import fr.sncf.osrd.utils.indexing.StaticIdxList
+import fr.sncf.osrd.utils.units.Offset
+import fr.sncf.osrd.utils.units.Speed
+import fr.sncf.osrd.utils.units.kilometersPerHour
+import fr.sncf.osrd.utils.units.meters
+
+/**
+ * Compute safety speed ranges, areas where the train has a lower speed limit because of a scheduled
+ * stop. For details, see https://osrd.fr/en/docs/reference/design-docs/timetable/#modifiable-fields
+ * (or google "VISA (VItesse Sécuritaire d'Approche)" for resources in French)
+ */
+fun makeSafetySpeedRanges(
+    infra: FullInfra,
+    chunkPath: ChunkPath,
+    routes: StaticIdxList<Route>,
+    schedule: List<SimulationScheduleItem>
+): DistanceRangeMap<Speed> {
+    val rawInfra = infra.rawInfra
+    val zonePaths = routes.flatMap { rawInfra.getRoutePath(it) }
+    val zonePathStartOffset = getRoutePathStartOffset(rawInfra, chunkPath, zonePaths)
+    val signalOffsets = getSignalOffsets(infra, zonePaths, zonePathStartOffset)
+
+    val stopsWithSafetySpeed = schedule.filter { it.receptionSignal.isStopOnClosedSignal }
+
+    val res = distanceRangeMapOf<Speed>()
+    for (stop in stopsWithSafetySpeed) {
+        // Currently, safety speed is applied to the next signal no matter the sight distance
+        val nextSignalOffset =
+            signalOffsets.firstOrNull { it >= stop.pathOffset }?.distance ?: chunkPath.length
+        res.put(
+            lower = nextSignalOffset - 200.meters,
+            upper = nextSignalOffset,
+            value = 30.kilometersPerHour,
+        )
+        if (stop.receptionSignal == RJSTrainStop.RJSReceptionSignal.SHORT_SLIP_STOP) {
+            res.put(
+                lower = nextSignalOffset - 100.meters,
+                upper = nextSignalOffset,
+                value = 10.kilometersPerHour,
+            )
+        }
+    }
+    // Safety speed areas may extend outside the path
+    return res.subMap(0.meters, chunkPath.length)
+}
+
+/** Return the offsets of block-delimiting signals on the path. */
+fun getSignalOffsets(
+    infra: FullInfra,
+    zonePaths: List<StaticIdx<ZonePath>>,
+    pathStartOffset: Offset<Path>,
+): List<Offset<TravelledPath>> {
+    val res = mutableListOf<Offset<TravelledPath>>()
+    val rawInfra = infra.rawInfra
+    val signalingInfra = infra.loadedSignalInfra
+    var prevZonePathsLength = 0.meters
+    for (zonePath in zonePaths) {
+        val signalPositions = rawInfra.getSignalPositions(zonePath)
+        val signals = rawInfra.getSignals(zonePath)
+        for ((signal, signalPosition) in signals zip signalPositions) {
+            val isDelimiter =
+                signalingInfra.getLogicalSignals(signal).any { signalingInfra.isBlockDelimiter(it) }
+            if (isDelimiter) {
+                res.add(
+                    Offset(prevZonePathsLength + signalPosition.distance - pathStartOffset.distance)
+                )
+            }
+        }
+        prevZonePathsLength += rawInfra.getZonePathLength(zonePath).distance
+    }
+    return res.filter { it.distance >= 0.meters }
+}
+
+/** Returns the offset where the train actually starts, compared to the start of the first route. */
+fun getRoutePathStartOffset(
+    infra: RawInfra,
+    chunkPath: ChunkPath,
+    zonePaths: List<ZonePathId>
+): Offset<Path> {
+    var prevChunksLength = Offset<Path>(0.meters)
+    val routeChunks = zonePaths.flatMap { infra.getZonePathChunks(it) }
+    for (chunk in routeChunks) {
+        if (chunk == chunkPath.chunks.first()) {
+            return prevChunksLength + chunkPath.beginOffset.distance
+        }
+        prevChunksLength += infra.getTrackChunkLength(chunk.value).distance
+    }
+    throw RuntimeException("Unreachable (couldn't find first chunk in route list)")
+}
--- a/core/src/main/java/fr/sncf/osrd/standalone_sim/StandaloneSim.java
+++ b/core/src/main/java/fr/sncf/osrd/standalone_sim/StandaloneSim.java
@@ -3,6 +3,7 @@ package fr.sncf.osrd.standalone_sim;
 import static fr.sncf.osrd.api.pathfinding.PathPropUtilsKt.makeChunkPath;
 import static fr.sncf.osrd.envelope_sim_infra.MRSPKt.computeMRSP;
 import static fr.sncf.osrd.sim_infra.api.PathPropertiesKt.makePathPropertiesWithRouteNames;
+import static fr.sncf.osrd.utils.DistanceRangeMapKt.distanceRangeMapOf;
 
 import fr.sncf.osrd.DriverBehaviour;
 import fr.sncf.osrd.api.FullInfra;
@@ -61,8 +62,10 @@ public class StandaloneSim {
                 var rollingStock = trainSchedule.rollingStock;
 
                 // MRSP & SpeedLimits
-                var mrsp = computeMRSP(trainPath, rollingStock, true, trainSchedule.tag);
-                var speedLimits = computeMRSP(trainPath, rollingStock, false, trainSchedule.tag);
+                var mrsp = computeMRSP(
+                        trainPath, rollingStock, true, trainSchedule.tag, distanceRangeMapOf(new ArrayList<>()));
+                var speedLimits = computeMRSP(
+                        trainPath, rollingStock, false, trainSchedule.tag, distanceRangeMapOf(new ArrayList<>()));
                 mrsp = driverBehaviour.applyToMRSP(mrsp);
                 cacheSpeedLimits.put(trainSchedule, ResultEnvelopePoint.from(speedLimits));
 
--- a/core/src/main/java/fr/sncf/osrd/standalone_sim/StandaloneSim.java
+++ b/core/src/main/java/fr/sncf/osrd/standalone_sim/StandaloneSim.java
@@ -3,7 +3,6 @@ package fr.sncf.osrd.standalone_sim;
 import static fr.sncf.osrd.api.pathfinding.PathPropUtilsKt.makeChunkPath;
 import static fr.sncf.osrd.envelope_sim_infra.MRSPKt.computeMRSP;
 import static fr.sncf.osrd.sim_infra.api.PathPropertiesKt.makePathPropertiesWithRouteNames;
-import static fr.sncf.osrd.utils.DistanceRangeMapKt.distanceRangeMapOf;
 
 import fr.sncf.osrd.DriverBehaviour;
 import fr.sncf.osrd.api.FullInfra;
@@ -62,10 +61,8 @@ public class StandaloneSim {
                 var rollingStock = trainSchedule.rollingStock;
 
                 // MRSP & SpeedLimits
-                var mrsp = computeMRSP(
-                        trainPath, rollingStock, true, trainSchedule.tag, distanceRangeMapOf(new ArrayList<>()));
-                var speedLimits = computeMRSP(
-                        trainPath, rollingStock, false, trainSchedule.tag, distanceRangeMapOf(new ArrayList<>()));
+                var mrsp = computeMRSP(trainPath, rollingStock, true, trainSchedule.tag, null);
+                var speedLimits = computeMRSP(trainPath, rollingStock, false, trainSchedule.tag, null);
                 mrsp = driverBehaviour.applyToMRSP(mrsp);
                 cacheSpeedLimits.put(trainSchedule, ResultEnvelopePoint.from(speedLimits));
 
--- a/core/src/main/kotlin/fr/sncf/osrd/envelope_sim_infra/MRSP.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/envelope_sim_infra/MRSP.kt
@@ -7,7 +7,6 @@ import fr.sncf.osrd.envelope_sim.EnvelopeProfile
 import fr.sncf.osrd.envelope_sim.PhysicsRollingStock
 import fr.sncf.osrd.sim_infra.api.PathProperties
 import fr.sncf.osrd.sim_infra.api.SpeedLimitProperty
-import fr.sncf.osrd.sim_infra.api.SpeedLimitSource
 import fr.sncf.osrd.sim_infra.api.SpeedLimitSource.UnknownTag
 import fr.sncf.osrd.utils.DistanceRangeMap
 import fr.sncf.osrd.utils.SelfTypeHolder
@@ -37,7 +36,7 @@ fun computeMRSP(
     rollingStock: PhysicsRollingStock,
     addRollingStockLength: Boolean,
     trainTag: String?,
-    safetySpeedRanges: DistanceRangeMap<Speed> = distanceRangeMapOf(),
+    safetySpeedRanges: DistanceRangeMap<Speed>? = null,
 ): Envelope {
     return computeMRSP(
         path,
@@ -66,7 +65,7 @@ fun computeMRSP(
     rsLength: Double,
     addRollingStockLength: Boolean,
     trainTag: String?,
-    safetySpeedRanges: DistanceRangeMap<Speed> = distanceRangeMapOf(),
+    safetySpeedRanges: DistanceRangeMap<Speed>? = null,
 ): Envelope {
     val builder = MRSPEnvelopeBuilder()
     val pathLength = toMeters(path.getLength())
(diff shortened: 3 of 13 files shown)


core/src/test/kotlin/fr/sncf/osrd/envelope_sim_infra/MRSPTest.kt
```diff
@@ -313,8 +314,20 @@ class MRSPTest {
...
                 Envelope.make(
                     EnvelopeTestUtils.makeFlatPart(
-                        LimitKind.TRAIN_LIMIT,
+                        listOf(LimitKind.TRAIN_LIMIT, CONSTANT_SPEED, HasMissingSpeedTag),
                         0.0,
+                        POSITION1,
+                        TestTrains.MAX_SPEED
+                    ),
+                    EnvelopeTestUtils.makeFlatPart(
+                        LimitKind.TRAIN_LIMIT,
```
bougue-pe, line 323: Don't we expect a CONSTANT_SPEED too? (looks like a lot are missing if we want this)

core/src/main/kotlin/fr/sncf/osrd/envelope_sim_infra/MRSP.kt
```diff
@@ -93,5 +97,89 @@ fun computeMRSP(
...
+
+    // Add safety speeds
+    if (safetySpeedRanges != null) {
+        for (range in safetySpeedRanges) {
+            val speed = range.value
+            val attrs =
+                mutableListOf<SelfTypeHolder?>(
+                    EnvelopeProfile.CONSTANT_SPEED,
+                    MRSPEnvelopeBuilder.LimitKind.SPEED_LIMIT,
+                    SpeedLimitSource.SafetyApproachSpeed,
```
bougue-pe: How about moving `SafetyApproachSpeed` to `LimitKind`?
Semantic looks a bit better, although things are (already) mixed between `LimitKind` and `SpeedLimitSource`.
```suggestion
                    MRSPEnvelopeBuilder.LimitKind.SAFETY_APPROACH_SPEED,
```
Not sure about all this, but we already have the mechanism to have multiple tags (and enforce unicity, which may not be much needed) and it looks like we are making a mess of it (and this is the case prior to this work).

core/src/main/kotlin/fr/sncf/osrd/envelope_sim_infra/MRSP.kt
```diff
@@ -84,6 +87,7 @@ fun computeMRSP(
         if (speedLimitProp.source != null) {
             attrs.add(speedLimitProp.source)
         }
+        if (attrs.any { it is UnknownTag }) attrs.add(HasMissingSpeedTag)
```
bougue-pe, line 89: Pretty much parts of the same debate about `attrs`: not fan of the redundancy here.

core/src/main/kotlin/fr/sncf/osrd/envelope_sim_infra/MRSP.kt
```diff
@@ -93,5 +97,89 @@ fun computeMRSP(
...
+): DistanceRangeMap<List<SelfTypeHolder?>> {
+    val result: DistanceRangeMap<List<SelfTypeHolder?>> =
+        distanceRangeMapOf(listOf(DistanceRangeMap.RangeMapEntry(lower, upper, baseAttrs)))
+
+    // Add important attributes from the old speed ranges
+    val attrsWithMissingRange = baseAttrs.toMutableList()
+    attrsWithMissingRange.add(HasMissingSpeedTag)
+    for (oldRange in propertyRangeMap.subMap(lower, upper)) {
+        if (oldRange.value.source is UnknownTag) {
+            result.put(oldRange.lower, oldRange.upper, attrsWithMissingRange)
```
bougue-pe, line 175: To me, it's not the responsibility of a safety approach speed section (or a train limit speed section) to support that a SpeedLimitTag is unknown.
It would feel more natural to have the final builder pick the interesting parts on all those, no?

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
```diff
@@ -0,0 +1,94 @@
...
+import fr.sncf.osrd.utils.indexing.StaticIdxList
+import fr.sncf.osrd.utils.units.Offset
+import fr.sncf.osrd.utils.units.Speed
+import fr.sncf.osrd.utils.units.kilometersPerHour
+import fr.sncf.osrd.utils.units.meters
+
+/**
+ * Compute safety speed ranges, areas where the train has a lower speed limit because of a scheduled
+ * stop. For details, see https://osrd.fr/en/docs/reference/design-docs/timetable/#modifiable-fields
+ * (or google "VISA (VItesse Sécuritaire d'Approche)" for resources in French)
```
bougue-pe, line 21: Maybe we should add (move) this to the website doc too?

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
```diff
@@ -0,0 +1,94 @@
...
+    routes: StaticIdxList<Route>,
+    schedule: List<SimulationScheduleItem>
+): DistanceRangeMap<Speed> {
+    val zonePaths = routes.flatMap { infra.getRoutePath(it) }
+    val zonePathStartOffset = getRoutePathStartOffset(infra, chunkPath, zonePaths)
+    val signalOffsets = getSignalOffsets(infra, zonePaths, zonePathStartOffset)
+
+    val stopsWithSafetySpeed =
+        schedule.filter {
+            it.stopFor != null && it.receptionSignal != RJSTrainStop.RJSReceptionSignal.OPEN
```
bougue-pe: 2 suggestions here:
```suggestion
            it.receptionSignal.isStopOnClosedSignal()
```
1. There are checks that `receptionSignal` matches `stopFor` before that, so I would only protect for crashes, not for functionnality here (so no check on `stopFor`). Maybe raise an error before in core if not trusting the client?
2. `isStopOnClosedSignal()` is probably more future-proof if enum evolves.

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+    val res = distanceRangeMapOf<Speed>()
+    for (stop in stopsWithSafetySpeed) {
+        val nextSignalOffset =
+            signalOffsets.firstOrNull { it >= stop.pathOffset }?.distance ?: chunkPath.length
+        res.put(
```
bougue-pe, line 41: If we get the signalSightDistance, we can add a condition before all that, no?
Not sure if we should check the type of signal here.

core/src/test/kotlin/fr/sncf/osrd/envelope_sim_infra/MRSPTest.kt
eckter: It's added by default in this context, see [here](https://github.com/OpenRailAssociation/osrd/blob/dev/core/envelope-sim/src/testFixtures/java/fr/sncf/osrd/envelope/EnvelopeTestUtils.java#L97)

It's not very clear, but I didn't really change this (the git diff is just confused with the similar `Arguments.of` calls

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
eckter: The "google it" part? It's not entirely ideal, it doesn't point toward any specific resource (and it's mostly third-party stuff, I didn't find any SNCF document for this).

Ideally we should have a dedicated website page that describes the concept, similar to what we have on the issue with the whiteboard pictures. But this is better than nothing, the method would lack context otherwise

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
eckter: I'm leaving this unresolved until we talk with @axrolld, I don't know the expected behavor there. Apparently there *can* be stops on closed signals even if it's not seen, but I'm not sure what the expected behavior would be.

core/src/main/kotlin/fr/sncf/osrd/envelope_sim_infra/MRSP.kt
eckter: The builder just assembles parts that are immutable, I don't think it has the responsibility of cherry-picking attributes.

I don't think attributes were a good system for this data. But changing that would be a very large refactoring, on a system that's supposed to be legacy (envelopes). We can try to think of a better solution but I don't know if there is any

core/src/test/kotlin/fr/sncf/osrd/envelope_sim_infra/MRSPTest.kt
bougue-pe: Indeed, those interfaces are confusing for test, OK to let it like that and avoid mixing things 👍 
And OK too if someone ever wants to extend the helper (or remove it).

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
bougue-pe: To be confirmed, after discussion with @Yohh and @Uriel-Sautron:
- [x] it looks like there is no realistic counter-example for ignoring signal sight (at least for 🇫🇷🚂).
The 200m safety speed of approach is supposed to be respected even when the sight distance is less than 200m (cases of turns, fog, ...)
Cases of stops out of the sight of signal are also most probably security stops, not planned, and the behavior in those case is even more restrictive: safety speed all the way to the next block-signal.

- [x] As for the type, it looks like the type of the signal where the safety speed of approach must be applied is the next signal that is delimiting a block, so this should be handled when confirmed.

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
bougue-pe: Yes, I think that it indeed helps here, and it would help "more" to mention on the website that the French notion for this is named VItesse Securitaire d'Approche (and the result of this web search is currently not so bad I have to say).

Anyway, it's nitpicking, so I resolve it and if I have the opportunity I may add this to the website.

core/src/main/kotlin/fr/sncf/osrd/envelope_sim_infra/MRSP.kt
bougue-pe: Understood and OK.
Maybe worth adding a comment/TODO to acknowledge that refactoring is a good idea if usage/maintenance justifies it.
(I let you resolve that one when you choose to do it or not).

eckter: I have added a TODO

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
eckter: I've added a filter for the signal type, we now only consider signals that delimit blocks.

core/src/main/java/fr/sncf/osrd/standalone_sim/StandaloneSim.java
```diff
@@ -61,8 +61,8 @@ public static StandaloneSimResult run(
                 var rollingStock = trainSchedule.rollingStock;
 
                 // MRSP & SpeedLimits
-                var mrsp = computeMRSP(trainPath, rollingStock, true, trainSchedule.tag);
-                var speedLimits = computeMRSP(trainPath, rollingStock, false, trainSchedule.tag);
+                var mrsp = computeMRSP(trainPath, rollingStock, true, trainSchedule.tag, null);
```
Erashin, line 64: Unneeded since `safetySpeedRanges` had a default `null` value?

eckter: This is java code, there's no default value.

This is also why the default value is `null` instead of an empty range map. I didn't feel like adding `distanceRangeMapOf(new ArrayList())` to every call.

Erashin: Yaaaay jaaavvvvaaa :D

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
```diff
@@ -0,0 +1,99 @@
...
+    chunkPath: ChunkPath,
+    routes: StaticIdxList<Route>,
+    schedule: List<SimulationScheduleItem>
+): DistanceRangeMap<Speed> {
+    val rawInfra = infra.rawInfra
+    val zonePaths = routes.flatMap { rawInfra.getRoutePath(it) }
+    val zonePathStartOffset = getRoutePathStartOffset(rawInfra, chunkPath, zonePaths)
+    val signalOffsets = getSignalOffsets(infra, zonePaths, zonePathStartOffset)
+
+    val stopsWithSafetySpeed = schedule.filter { it.receptionSignal.isStopOnClosedSignal }
```
bougue-pe, line 34: It looks like the part "stops before a buffer-stop automatically trigger short-slip stop's speed (no matter the input)" is not managed.
* We can probably always trigger that, no matter if there's a stop or not 🙃
* We can probably only manage it at the end of the path if it's easier (not backtracking currently)
* We should look for "RouteWaypoints" (detectors) in that case (and especially look after the end of the path)
* It can probably be handled in a separate PR if more comfortable (someone else can take over too)

EDIT: See https://github.com/OpenRailAssociation/osrd/pull/8993#discussion_r1793789653 for test on this

core/src/main/kotlin/fr/sncf/osrd/envelope_sim_infra/MRSP.kt
bougue-pe: From the current implem it's probably the better tradeoff between semantics and work required to refactor.
So I resolve that one.

core/src/main/kotlin/fr/sncf/osrd/standalone_sim/StandaloneSimulation.kt
```diff
@@ -202,12 +204,15 @@ fun makeMRSPResponse(speedLimits: Envelope): RangeValues<SpeedLimitProperty> {
...
-                part.getAttr(SpeedLimitSource::class.java)
-            )
-        )
+        var attr = part.getAttr(SpeedLimitSource::class.java)
+        if (part.getAttr(HasMissingSpeedTag::class.java) != null) {
+            // HasMissingSpeedTag is a special flag that enforces `UnknownTag` even if the speed
+            // limit comes from a defined source (e.g. rolling stock max speed). This ensures we
+            // never lose that important piece of information.
+            // TODO: change the API with better semantics to properly handle missing tags
+            attr = SpeedLimitSource.UnknownTag()
```
bougue-pe, line 213: nit: maybe it would make sense to move this function in `MRSP.kt`, to contain the logic in one place.

core/src/test/kotlin/fr/sncf/osrd/standalone_sim/StandaloneSimulationTest.kt
```diff
@@ -286,6 +285,77 @@ class StandaloneSimulationTest {
...
+        val expected =
+            distanceRangeMapOf(
+                listOf(
+                    DistanceRangeMap.RangeMapEntry(0.meters, 50.meters, 30.kilometersPerHour),
+                    DistanceRangeMap.RangeMapEntry(50.meters, 150.meters, 10.kilometersPerHour),
+                    DistanceRangeMap.RangeMapEntry(
+                        10_200.meters,
+                        10_400.meters,
+                        30.kilometersPerHour
+                    ),
```
bougue-pe, line 329: If buffer-stops are automatically short-slip stops, this test should move (maybe also try to change the reception signal to `OPEN`)
```suggestion
                    DistanceRangeMap.RangeMapEntry(
                        10_200.meters,
                        10_300.meters,
                        30.kilometersPerHour
                    ),
                    DistanceRangeMap.RangeMapEntry(
                        10_300.meters,
                        10_400.meters,
                        10.kilometersPerHour
                    ),
```

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
```diff
@@ -0,0 +1,99 @@
...
+
+    val res = distanceRangeMapOf<Speed>()
+    for (stop in stopsWithSafetySpeed) {
+        val nextSignalOffset =
```
bougue-pe, line 39: nit: how about adding a comment to explicit business logic choices?
```suggestion
        // Currently, safety speed is applied to the next signal no matter the sight distance
        val nextSignalOffset =
```

front/tests/008-train-schedule.spec.ts
```diff
@@ -66,9 +66,9 @@ test.describe('Verifying that all elements in the train schedule are loaded corr
     await opTimetablePage.verifyTrainCount(20);
     await opTimetablePage.filterValidityAndVerifyTrainCount(selectedLanguage, 'Invalid', 4);
     await opTimetablePage.filterValidityAndVerifyTrainCount(selectedLanguage, 'All', 20);
-    await opTimetablePage.filterHonoredAndVerifyTrainCount(selectedLanguage, 'Honored', 12);
-    await opTimetablePage.filterValidityAndVerifyTrainCount(selectedLanguage, 'Valid', 12);
-    await opTimetablePage.filterHonoredAndVerifyTrainCount(selectedLanguage, 'Not honored', 4);
+    await opTimetablePage.filterHonoredAndVerifyTrainCount(selectedLanguage, 'Honored', 13);
+    await opTimetablePage.filterValidityAndVerifyTrainCount(selectedLanguage, 'Valid', 13);
+    await opTimetablePage.filterHonoredAndVerifyTrainCount(selectedLanguage, 'Not honored', 3);
```
bougue-pe, line 71: I don't understand those changes in e2e tests (same for margin), maybe we should discuss it aside and just report here the explanation.

bougue-pe (commented):
Thanks for the work on all this (close to merge once e2e tests are clear)!
* Depending on the discussions about buffer-stop's short-slip safety speed, I'll create a new issue to track this part of the feature.
* Anyway, a separate issue should be created to track the STDCM part (not sure when it will be handled).

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
eckter: Oh I actually forgot about this, good catch.

We can look for the exit point at the end of the last route and check if it's a buffer stop. It's the only place where we can find a relevant buffer stop.

It shouldn't take too long but the feature freeze is getting close, I'll leave the comment there for now and either fix it now or open an issue.

core/src/main/kotlin/fr/sncf/osrd/standalone_sim/StandaloneSimulation.kt
eckter: Seems fine to me, but I probably should comment the logic there a bit further.

The logic was to separate the "speed limit source" tag, which does what its name suggest: indicates where does this part of the MRSP come from. And then the `HasMissingSpeedTag` is different, it has nothing to do with the source of the speed limit. It's a warning.

And then when we build the response, we have to adapt to the expected format. The info that we're missing a tag is expected to be filled here.

I'll write a comment that explains all that. We could also open an issue as IMO the API is problematic, but there's no clear idea of what we'd want instead. (maybe [warnings](https://github.com/OpenRailAssociation/osrd/issues/6925) though)

eckter: https://github.com/OpenRailAssociation/osrd/issues/9281 + comment in https://github.com/OpenRailAssociation/osrd/pull/8993/commits/4fb2465eb61d1b21527e04ae2e3d35026d626344

front/tests/008-train-schedule.spec.ts
eckter: @Maymanaf could probably explain it better than me, I'm still a little confused

Apparently there was a train that was already really close to the limit, and lowering the speed limit made it possible to follow the margins

shenriotpro (approved):
LGTM for tests/

eckter:
> Anyway, a separate issue should be created to track the STDCM part (not sure when it will be handled).

https://github.com/OpenRailAssociation/osrd/issues/9282

> Depending on the discussions about buffer-stop's short-slip safety speed, I'll create a new issue to track this part of the feature.

https://github.com/OpenRailAssociation/osrd/issues/9283

I'll start working right away but on a different branch

front/tests/008-train-schedule.spec.ts
bougue-pe: As far as I understood, there are also probably rounding bugs and changing the base speed-profile may have "solved" the bug for this precise case.
We have some issue tracking those problems and possible evolutions: internal https://github.com/osrd-project/osrd-confidential/issues/680 and also bugs https://github.com/OpenRailAssociation/osrd/issues/9179 + https://github.com/OpenRailAssociation/osrd/issues/8959

I'm resolving it 👍

front/tests/assets/operationStudies/timesAndStops/expectedOutputsCellsData.json
```diff
@@ -25,8 +25,8 @@
     "margin": {
       "theoretical": "1 min/100km",
       "theoreticalS": "9 s",
-      "actual": "167 s",
-      "difference": "159 s"
+      "actual": "114 s",
```
bougue-pe, line 28: Understood that one: closed-signal reception leads to a slower base speed-profile (so longer base duration to reach the next stop) and the actual margin is the diff between that base duration and the duration of the actual speed-profile used to target both margin and provided arrival-time.
So that diff is lowered ✅

core/src/test/kotlin/fr/sncf/osrd/standalone_sim/StandaloneSimulationTest.kt
eckter: Will be part of https://github.com/OpenRailAssociation/osrd/issues/9283

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
eckter: Will be part of https://github.com/OpenRailAssociation/osrd/issues/9283

Erashin (approved):

clarani (approved):
LGTM :white_check_mark:

bougue-pe (commented):
✅ just tested OK on small infra & France.
🎉 thanks for all this !

PR #9024, opened 2024-09-26
editoast: add 'locked' to the schema definition of rolling stock

woshilapin:
In the schema definition in Python, the [field `locked` exists](https://github.com/OpenRailAssociation/osrd/blob/a5e9ca26575d885868a4e4a947479fea37c7299e/python/osrd_schemas/osrd_schemas/rolling_stock.py#L197). However, it doesn't exist in the corresponding `editoast` implementation.

### DIFF
--- a/editoast/editoast_schemas/src/rolling_stock.rs
+++ b/editoast/editoast_schemas/src/rolling_stock.rs
@@ -50,6 +50,7 @@ pub const ROLLING_STOCK_RAILJSON_VERSION: &str = "3.2";
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
 pub struct RollingStock {
     pub name: String,
+    pub locked: bool,
     pub effort_curves: EffortCurves,
     pub base_power_class: Option<String>,
     pub length: f64,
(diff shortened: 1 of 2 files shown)


leovalais (approved):
LGTM.

However I'm not sure it makes a lot of sense to have a `locked` field in the schemas at all. The field (which also exists for infrastructures) exists to prevent modification of the resource. When the permission system will be up these fields are likely to disappear since they're functionally equivalent to a share at the `READ` privilege level. That's why from a separation of concerns perspective I'm not sure we want to expose our current lack of resource control in the schemas. Maybe @flomonster has a better overview of the subject.

In any case, there shouldn't be any mismatch between `osrd_schemas` and `editoast_schemas` one way or the other.

flomonster:
> However I'm not sure it makes a lot of sense to have a `locked` field in the schemas at all. The field (which also exists for infrastructures) exists to prevent modification of the resource. When the permission system will be up these fields are likely to disappear since they're functionally equivalent to a share at the `READ` privilege level. That's why from a separation of concerns perspective I'm not sure we want to expose our current lack of resource control in the schemas. Maybe @flomonster has a better overview of the subject.

You are right, the locked field will disappear with the permission system.

The reason we did not expose this field is because the RS editor does not permit to edit it. It is only used during the import.

woshilapin:
Does that mean that in the meantime, we're ok with the difference between `osrd_schemas` and `editoast_schemas`, until permission system appears and then, we'll remove it everywhere from `osrd_schemas`?

flomonster:
> Does that mean that in the meantime, we're ok with the difference between `osrd_schemas` and `editoast_schemas`, until permission system appears and then, we'll remove it everywhere from `osrd_schemas`?

I prefer to merge this PR and try to make `osrd_schemas` match `editoast_schemas`.

PR #9027, opened 2024-09-26
front: fix trainrun deletion from NGE

### LINKED ISSUE #9015: nge: cannot delete trains from NGE interface
### What happened?

In the NGE interface I cannot delete a train with `del` or by moving the link out of an OP.

I expect it to be possible and see a request in the network of my browser

1. create a train
2. switch to macroscopic
3. try to delete your train

Recette (SNCF)

Firefox

staging a5e9ca2

emersion:
There is no trainrun in the DTO when it's been deleted. Instead, only use the trainrun from the DTO in the create/update handlers.

Closes: https://github.com/OpenRailAssociation/osrd/issues/9015


Uriel-Sautron (approved):
LGTM, tested !!

clarani (approved):

PR #9029, opened 2024-09-26
front: convert base64 encoded map image to jpeg

### LINKED ISSUE #8978: Make STDCM simulation sheet lighter

STDCM simulation sheet weights ~500ko/1Mo. This might be a bit heavy, it would be nice to be able to reduce this wigth by at least one order of magnitude.

This can be done by working on the map image, trying to make it lighter.
If it has a huge impact the map could pententially even be removed.

This feature matters because it will not overwhelm the train slot command software.

_No response_

Akctarus:
closes #8978 

The map image was just base64 encoded, now we convert it to jpeg and reduce its quality a little, saving up to 500KB

### DIFF
--- a/front/src/applications/stdcm/components/SimulationReportSheet.tsx
+++ b/front/src/applications/stdcm/components/SimulationReportSheet.tsx
@@ -1,3 +1,5 @@
+import { useEffect, useState } from 'react';
+
 import { Table, TR, TH, TD } from '@ag-media/react-pdf-table';
 import { Page, Text, Image, Document, View, Link } from '@react-pdf/renderer';
 import { useTranslation } from 'react-i18next';
@@ -9,7 +11,7 @@ import { capitalizeFirstLetter } from 'utils/strings';
 
 import styles from './SimulationReportStyleSheet';
 import type { SimulationReportSheetProps } from '../types';
-import { getStopDurationTime } from '../utils/formatSimulationReportSheet';
+import { base64ToJpeg, getStopDurationTime } from '../utils/formatSimulationReportSheet';
 
 const SimulationReportSheet = ({
   stdcmData,
@@ -31,6 +33,28 @@ const SimulationReportSheet = ({
     path_number1: 'n°XXXXXX',
     path_number2: 'n°YYYYYY',
   };
+
+  const [mapImageUrl, setMapImageUrl] = useState<string | null>(null);
+
+  // Convert image to JPEG
+  useEffect(() => {
+    if (mapCanvas) {
+      base64ToJpeg(mapCanvas, 0.8).then((blob) => {
+        const objectUrl = URL.createObjectURL(blob);
+        setMapImageUrl(objectUrl);
+      });
+    }
+  }, [mapCanvas]);
+
+  // Cleanup the object URL when the component is unmounted or before a new one is created
+  useEffect(
+    () => () => {
+      if (mapImageUrl) {
+        URL.revokeObjectURL(mapImageUrl);
+      }
+    },
+    [mapImageUrl]
+  );
 
   return (
     <Document>
@@ -345,7 +369,7 @@ const SimulationReportSheet = ({
         </View>
         {mapCanvas && (
           <View style={styles.map.map} id="simulationMap">
-            <Image src={mapCanvas} />
+            {mapImageUrl && <Image src={mapImageUrl} />}
           </View>
         )}
         <View style={styles.footer.warrantyBox}>
--- a/front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
+++ b/front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
@@ -142,3 +142,41 @@ export function getOperationalPointsWithTimes(
 
   return opResults;
 }
+
+/**
+ * Converts a base64 image into a JPEG blob while reducing quality.
+ * @param base64Image - Image in base64
+ * @param quality - Image quality (between 0 and 1, where 1 is the best quality)
+ * @returns {Promise<Blob>} - Return an optimised JPEG Blob
+ */
+export const base64ToJpeg = (base64Image: string, quality: number): Promise<Blob> =>
+  new Promise((resolve, reject) => {
+    const img = new Image();
+    img.src = base64Image;
+    img.onload = () => {
+      const canvas = document.createElement('canvas');
+      const ctx = canvas.getContext('2d');
+      if (!ctx) {
+        reject(new Error('Could not get canvas context'));
+        return;
+      }
+      canvas.width = img.width;
+      canvas.height = img.height;
+
+      ctx.drawImage(img, 0, 0);
+
+      canvas.toBlob(
+        (blob) => {
+          if (blob) {
+            resolve(blob);
+          }
+        },
+        'image/jpeg',
+        quality
+      );
+    };
+
+    img.onerror = (error) => {
+      reject(error);
+    };
+  });


front/src/applications/stdcm/components/SimulationReportSheet.tsx
```diff
@@ -32,6 +34,17 @@ const SimulationReportSheet = ({
     path_number2: 'n°YYYYYY',
   };
 
+  const [mapBlob, setMapBlob] = React.useState<Blob | null>(null);
+
+  // Convert image to JPEG
+  React.useEffect(() => {
```
SharglutDev:
```suggestion
  const [mapBlob, setMapBlob] = useState<Blob | null>(null);

  // Convert image to JPEG
 useEffect(() => {
```

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
```diff
@@ -142,3 +142,39 @@ export function getOperationalPointsWithTimes(
...
   return opResults;
 }
+
+/**
+ * Converts a base64 image into a JPEG blob while reducing quality.
+ * @param base64Image - Image in base64
+ * @param quality - Image quality (between 0 and 1, where 1 is the best quality)
+ * @returns {Promise<Blob | null>} - Return an optimised JPEG Blob
+ */
+export const base64ToJpeg = async (base64Image: string, quality: number): Promise<Blob | null> =>
```
emersion: There's no `await` inside this function so I don't think this needs to be `async`?

front/src/applications/stdcm/components/SimulationReportSheet.tsx
```diff
@@ -345,7 +358,7 @@ const SimulationReportSheet = ({
         </View>
         {mapCanvas && (
           <View style={styles.map.map} id="simulationMap">
-            <Image src={mapCanvas} />
+            {mapBlob && <Image src={URL.createObjectURL(mapBlob)} />}
```
emersion: Seems like this allocates resources each time the component is re-rendered: https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL_static#memory_management

Suggestions:

- Call `URL.createObjectURL()` from the `useEffect()` callback.
- Use [URL.revokeObjectURL()](https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL_static) when overwriting a previous URL with a fresh one

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
Akctarus: it's an oversight, well seen !

front/src/applications/stdcm/components/SimulationReportSheet.tsx
Akctarus: done :)

front/src/applications/stdcm/components/SimulationReportSheet.tsx
```diff
@@ -32,6 +34,33 @@ const SimulationReportSheet = ({
     path_number2: 'n°YYYYYY',
   };
 
+  const [mapBlob, setMapBlob] = useState<string | null>(null);
```
emersion: Nit: maybe `mapImageUrl` would be a more appropriate name now that this isn't a `Blob` anymore?

front/src/applications/stdcm/components/SimulationReportSheet.tsx
```diff
@@ -32,6 +34,33 @@ const SimulationReportSheet = ({
...
+          }
+          objectUrl = URL.createObjectURL(blob);
+          setMapBlob(objectUrl);
+        }
+      });
+    }
+
+    // Cleanup the object URL when the component is unmounted or before a new one is created
+    return () => {
+      if (objectUrl) {
```
emersion: Instead of a separate `objectUrl` variable, we could register a callback which runs on unmount in a separate `useEffect()` and uses the state variable. Could be simpler/more readable.

front/src/applications/stdcm/components/SimulationReportSheet.tsx
Akctarus: of course

front/src/applications/stdcm/components/SimulationReportSheet.tsx
Akctarus: Fixed

front/src/applications/stdcm/components/SimulationReportSheet.tsx
```diff
@@ -32,6 +34,31 @@ const SimulationReportSheet = ({
...
   };
 
+  const [mapImageUrl, setMapImageUrl] = useState<string | null>(null);
+
+  // Convert image to JPEG
+  useEffect(() => {
+    if (mapCanvas) {
+      base64ToJpeg(mapCanvas, 0.8).then((blob) => {
+        if (blob) {
+          // Revoke the previous object URL if it exists
```
emersion: Nit: this comment can be removed now

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
```diff
@@ -142,3 +142,39 @@ export function getOperationalPointsWithTimes(
...
   return opResults;
 }
+
+/**
+ * Converts a base64 image into a JPEG blob while reducing quality.
+ * @param base64Image - Image in base64
+ * @param quality - Image quality (between 0 and 1, where 1 is the best quality)
+ * @returns {Promise<Blob | null>} - Return an optimised JPEG Blob
+ */
+export const base64ToJpeg = (base64Image: string, quality: number): Promise<Blob | null> =>
```
emersion: Nit: `Blob | null` can be replaced with `Blob` only since the promise is rejected on error.

Akctarus: done

front/src/applications/stdcm/components/SimulationReportSheet.tsx
Akctarus: done

front/src/applications/stdcm/components/SimulationReportSheet.tsx
```diff
@@ -32,6 +34,30 @@ const SimulationReportSheet = ({
     path_number2: 'n°YYYYYY',
   };
 
+  const [mapImageUrl, setMapImageUrl] = useState<string | null>(null);
+
+  // Convert image to JPEG
+  useEffect(() => {
+    if (mapCanvas) {
+      base64ToJpeg(mapCanvas, 0.8).then((blob) => {
+        if (blob) {
```
emersion: Alright, one last nit: I think this `if` is unnecessary now that `blob` cannot be `null`.

emersion (approved):
LGTM with that nit fixed, but not tested! (feel free to resolve the discussion)

front/src/applications/stdcm/components/SimulationReportSheet.tsx
Akctarus: fixed, ty for your review :)

SarahBellaha (approved):
Tested, lgtm :)

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
```diff
@@ -142,3 +142,41 @@ export function getOperationalPointsWithTimes(
 
   return opResults;
 }
+
+/**
+ * Converts a base64 image into a JPEG blob while reducing quality.
```
SharglutDev, line 147: Do we accept this low quality or it was already like that before ? (open the img in full screen)

Capture d’écran 2024-10-04 à 11 19 54

SharglutDev (approved):
Lgtm and tested, nice improvement

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
Akctarus: Yes it used to be like that, it depends on the resolution of your screen, here's a screen I just took with 80% quality : 

Capture d’écran 2024-10-04 à 11 43 03

Akctarus: Anyways, once we have information on the maximum authorized weight, we'll be able to adapt the degree of quality reduction.

Issue #7861, opened 2024-06-27
[meta] front: manage TODOs in front/src


Yohh:
### Description and goal

we have a hundred TODO comments in front/src
half are tagged `TS2` or `DROP V1` and will be managed during the PI11
half are sometimes 15 months old
we have to managed those old TODOs

### Acceptance criteria

___

#### DROP V1

- [ ] common/Map/WarpedMap/SimulationWarpedMap.tsx:276:      // TODO DROP V1: remove this condition
- [ ] modules/simulationResult/components/SpeedSpaceChart/types.ts:70:// TODO DROP V1: remove this
- [ ] modules/simulationResult/components/SpeedSpaceChart/types.ts:99:// TODO DROP V1: remove this
- [ ] modules/simulationResult/components/SpeedSpaceChart/SpeedSpaceSettings.tsx:23:  electrificationRanges: ElectrificationRange[] | PathPropertiesFormatted['electrifications']; // TODO DROP V1 : remove ElectrificationRange type
- [ ] modules/simulationResult/components/SpeedSpaceChart/SpeedSpaceSettings.tsx:58:    electricRanges: (ElectrificationRange | ElectrificationRangeV2)[] // TODO DROP V1 : remove ElectrificationRange type
- [ ] modules/simulationResult/components/SpeedSpaceChart/utils.ts:33:  curves: RadiusPosition[] | PositionData<'radius'>[], // TODO DROPV1 : remove RadiusPosition type
- [ ] modules/simulationResult/components/SpeedSpaceChart/d3Helpers.ts:45:// TODO DROP V1 : remove this
- [ ] modules/simulationResult/components/SpeedSpaceChart/d3Helpers.ts:207:  trainSimulation: GevPreparedDataV2 | GevPreparedData, // TODO DROP V1 : remove GevPreparedData type
- [ ] modules/simulationResult/components/SpeedSpaceChart/prepareData.ts:11:// TODO DROP V1 : remove PositionSpeedTime type
- [ ] modules/simulationResult/components/SpeedSpaceChart/prepareData.ts:21:  slopesHistogram: GradientPosition[] | PositionData<'gradient'>[], // TODO DROP V1 : remove GradientPosition type
- [ ] modules/simulationResult/components/SpeedSpaceChart/prepareData.ts:31:// TODO DROP V1 : remove this
[...]

PR #9032, opened 2024-09-27
front: handles various TODOs

jacomyal:
This commit solves some TODO comments from the front codebase.

It relates to ticket #7861.

Details:
- Renames notification reducers
- Rewrites `checkCurrentConfig().path`
- Fixes speed limits selection count in the editor
- Cleans some TODOs, typos and dead code in the editor codebase
- Cleans up linear metadata code

### DIFF
--- a/front/src/applications/editor/components/EditorForm.tsx
+++ b/front/src/applications/editor/components/EditorForm.tsx
@@ -51,7 +51,7 @@ function EditorForm<T extends Omit<EditorEntity, 'objType'> & { objType: string 
 }: PropsWithChildren<EditorFormProps<T>>) {
   const [error, setError] = useState<string | null>(null);
   const [formData, setFormData] = useState<GeoJsonProperties>(data.properties);
-  const [submited, setsubmited] = useState<boolean>(false);
+  const [submitted, setSubmitted] = useState<boolean>(false);
   const { t } = useTranslation('infraEditor');
 
   const editorState = useSelector(getEditorState);
@@ -89,15 +89,15 @@ function EditorForm<T extends Omit<EditorEntity, 'objType'> & { objType: string 
 
   return (
     <div>
-      {submited && error !== null && (
+      {submitted && error !== null && (
         <div className="form-error mt-3 mb-3">
           <p className="mb-0">{error}</p>
         </div>
       )}
       <Form
         fields={{ ...fields, ...(overrideFields || {}) }}
-        liveValidate={submited}
-        showErrorList={submited ? 'top' : false}
+        liveValidate={submitted}
+        showErrorList={submitted ? 'top' : false}
         action={undefined}
         noHtml5Validate
         validator={validator}
@@ -110,7 +110,7 @@ function EditorForm<T extends Omit<EditorEntity, 'objType'> & { objType: string 
           length: data.properties?.length,
           isCreation: isNil(formData?.id) || formData?.id === NEW_ENTITY_ID,
         }}
-        onError={() => setsubmited(true)}
+        onError={() => setSubmitted(true)}
         onSubmit={async () => {
           try {
             setError(null);
@@ -118,7 +118,7 @@ function EditorForm<T extends Omit<EditorEntity, 'objType'> & { objType: string 
           } catch (e) {
             setError(getErrorMessage(e));
           } finally {
-            setsubmited(true);
+            setSubmitted(true);
           }
         }}
         onChange={(event) => {
--- a/front/src/applications/editor/components/EditorForm.tsx
+++ b/front/src/applications/editor/components/EditorForm.tsx
@@ -51,7 +51,7 @@ function EditorForm<T extends Omit<EditorEntity, 'objType'> & { objType: string 
 }: PropsWithChildren<EditorFormProps<T>>) {
   const [error, setError] = useState<string | null>(null);
   const [formData, setFormData] = useState<GeoJsonProperties>(data.properties);
-  const [submited, setSubmited] = useState<boolean>(false);
+  const [submited, setsubmited] = useState<boolean>(false);
   const { t } = useTranslation('infraEditor');
 
   const editorState = useSelector(getEditorState);
@@ -110,7 +110,7 @@ function EditorForm<T extends Omit<EditorEntity, 'objType'> & { objType: string 
           length: data.properties?.length,
           isCreation: isNil(formData?.id) || formData?.id === NEW_ENTITY_ID,
         }}
-        onError={() => setSubmited(true)}
+        onError={() => setsubmited(true)}
         onSubmit={async () => {
           try {
             setError(null);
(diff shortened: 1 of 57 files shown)


front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/checkCurrentConfig.ts
```diff
@@ -145,22 +146,28 @@ const checkCurrentConfig = (
...
-        ch,
-        metadata,
-        kp,
-        receptionSignal,
-        theoreticalMargin,
-        ...stepLocation
-      } = step;
+    path: compact(pathSteps).map((step: PathStep) => {
+      const { ch } = step;
+      const stepLocation = omit(
```
emersion, line 151: Shouldn't this be a `pick()` with `track` + `offset` + `trigram` + all other fields which can be in `PathItemLocation`, rather than `omit()`?

front/src/applications/editor/tools/switchEdition/components/TrackNodeTypeDiagram.tsx
```diff
@@ -6,40 +6,16 @@ import link from 'assets/pictures/trackNodeTypes/link.svg';
...
-      case TRACK_NODE_TYPES_ID.POINT_SWITCH:
-        source = pointSwitch;
-        break;
-      default:
-        source = undefined;
-    }
-    return source;
-  }, [trackNodeType]);
+const TrackNodeTypeDiagram = ({ trackNodeType }: { trackNodeType: string }) => {
+  const trackNodeTypeImage = useMemo(() => SOURCES[trackNodeType], [trackNodeType]);
```
emersion: Nit: since this is just a hashmap access, do we really need a `useMemo()` here?

front/src/applications/editor/components/EditorForm.tsx
```diff
@@ -51,7 +51,7 @@ function EditorForm<T extends Omit<EditorEntity, 'objType'> & { objType: string
 }: PropsWithChildren<EditorFormProps<T>>) {
   const [error, setError] = useState<string | null>(null);
   const [formData, setFormData] = useState<GeoJsonProperties>(data.properties);
-  const [submited, setSubmited] = useState<boolean>(false);
+  const [submited, setsubmited] = useState<boolean>(false);
```
emersion: Nit: missing capital "s" for "submited", and missing a "t".

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/checkCurrentConfig.ts
jacomyal: Well, both work. What is your argument in favor of `pick` rather than `omit`?

I chose to use `omit` because the common factor between `PathStep` and `TrainScheduleBase['path']` is `PathItemLocation`, which is a union type, so not something easy to construct properly using `pick`:

```typescript
export type PathItemLocation =
  | TrackOffset
  | {
      operational_point: string;
    }
  | {
      /** An optional secondary code to identify a more specific location */
      secondary_code?: string | null;
      trigram: string;
    }
  | {
      /** An optional secondary code to identify a more specific location */
      secondary_code?: string | null;
      /** The [UIC](https://en.wikipedia.org/wiki/List_of_UIC_country_codes) code of an operational point */
      uic: number;
    };
```

front/src/applications/editor/tools/switchEdition/components/TrackNodeTypeDiagram.tsx
jacomyal: Oh, you're right, I let the `useMemo`, but it's became useless. I'll remove it, thanks :+1:

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/checkCurrentConfig.ts
emersion: I think the old code was better than `omit()`, because in both cases we're listing all of the fields we don't want, and when using destructuring there's no need to cast. (As a bonus, it's vanilla JS.)

I'd prefer to use `pick()` because there are more fields we don't want than fields we want to keep, and we're more likely to add fields in `PathStep` then we are to add some in `PathItemLocation`.

> a union type, so not something easy to construct properly using pick

A union type should be easy to construct from `pick()` no? Basically just:

```ts
pick(pathStep, 'track', 'offset', 'secondary_code', 'trigram', 'uic')
```

Any field not present in the original value still won't be present in the final value.

front/src/common/IntervalsDataViz/data.ts
```diff
@@ -382,7 +357,6 @@ export function update<T>(
   if (diff.length === 0) return linearMetadata;
 
   // We take the first one
-  // TODO: an impovment can be to take the one in the middle if there are many
```
SharglutDev, line 385: Not needed anymore ?

front/src/common/IntervalsDataViz/data.ts
```diff
@@ -660,16 +589,11 @@ export function getClosestOperationalPoint(
 }
 
 /**
- * TODO: need to be check and tested (specially the underlying update function)
```
SharglutDev, line 663: No need for tests ?

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/checkCurrentConfig.ts
```diff
@@ -1,15 +1,16 @@
...
-import { setFailure } from 'reducers/main';
-import type { OsrdConfState } from 'reducers/osrdconf/types';
+import { notifyFailure } from 'reducers/main';
+import type { OsrdConfState, PathStep } from 'reducers/osrdconf/types';
 import { isInvalidFloatNumber } from 'utils/numbers';
 import { kmhToMs, mToMm } from 'utils/physics';
 
 import formatMargin from './formatMargin';
 import formatSchedule from './formatSchedule';
+import type { TrainScheduleBase } from '../../../../../common/api/generatedEditoastApi';
```
SharglutDev, line 13:
```suggestion
import type { TrainScheduleBase } from 'common/api/osrdEditoastApi';
```

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/checkCurrentConfig.ts
SharglutDev: I also think `pick` seems simpler bu not sure if that's why @jacomyal prefers `omit`, but as when using `pick` we don't have typescript auto completion for the `PathItemLocation` properties, it might result to some weird things like `pick(pathStep, 'track', 'offffset', 'secondaryCode', 'trigram', 'uic', 'blob')` without getting any error as we have to use a cast ?

emersion: To me `omit()` is the worst of both worlds: no upside over object destructuring and same downsides as `pick()`.

(Ideally I'd prefer to isolate the `PathItemLocation` in a prop of `PathStep`...)

flomonster:
Can we close this PR?

loic-hamelin:
Thanks Alexis !

PR #9033, opened 2024-09-27
Using isoDateWithTimezoneToSec for timezone-aware conversions

### LINKED ISSUE #8933: Train schedule created notification display wrong departure time
### What happened?

image

image

I created a train schedule that starts at 08:00:00 but the notification (bottom right) displays 06:00:00

 However, the created train schedule has the right departure time.

image

_No response_

1. Create train schedule
2. notice

Local

Firefox

staging fbf8e3e

theocrsb:
Closes #8933


Akctarus (dismissed):
This correction creates a new bug: the train created is set to `1/01/1970 01:00:00 `. I think the bug you're trying to correct is due to a miscalculation of the time zone, since the delay between the notification and the actual departure time is only 2 hours.

Akctarus (changes requested):
see last comment

theocrsb:
Ok I didn't see that I also modify the train creation date. I used the timezone only for the notification. Thanks for your feedback

Akctarus (approved):
Tested, LGTM :)

clarani (approved):

PR #9035, opened 2024-09-27
front: fix travel time >1h in OSRD → NGE conversion

### LINKED ISSUE #8862: Travel times (consecutiveTime) not set properly when > 60'
### What happened?

In macro : 
image

Switch macro -> micro -> macro :

image

90' became 30'

do not modulo travel time

1. set travel time > 60'
2. macro -> micro -> macro
3. notice modulo

Local

Firefox

c62c9db

emersion:
Multiple issues here:

- Source or target "time" may be zero or null, zero means it's set but top of the hour, null means unset/unknown. We still want to compute the travel time if it's zero.
- Travel "time" field doesn't need to be computed modulo 1h, because it's a duration, not an absolute time.
- Travel "consecutiveTime" field needs to be populated as well.

Closes: https://github.com/OpenRailAssociation/osrd/issues/8862

### DIFF
--- a/front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
@@ -464,8 +464,9 @@ const importTimetable = async (
         }
 
         const travelTime = { ...DEFAULT_TIME_LOCK };
-        if (targetArrival.time && sourceDeparture.time) {
-          travelTime.time = (targetArrival.time - sourceDeparture.time + 60) % 60;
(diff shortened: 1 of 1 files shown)


louisgreiner (approved):
Tested and it's working very well, thanks :)

For the code part, I still have some warning about linting

emersion:
> For the code part, I still have some warning about linting

Apologies, should be fixed now!

sim51 (approved):

emersion:
Small cleanup to get rid of unnecessary null assertions after discussing with @sim51.

sim51:
good for me !

SharglutDev (approved):
Lgtm and tested

PR #9040, opened 2024-09-27
front: fix comfort type not updated after rs change

### LINKED ISSUE #9020: The confort type doesn't change when the RS is changed
### What happened?

https://github.com/user-attachments/assets/97e57ebf-61eb-4c84-a496-1dc057a47295

The correct confort type should be displayed

1. Go to the **Rolling Stock** tab.
2. Select a rolling stock (RS) with multiple comfort types: **Standard**, and/or **Heating**, and/or **Air Conditioning**.
3. Modify the selected RS and choose a new one with only the **Standard** comfort type.

Recette (SNCF)

Google Chrome

a5e9ca2

RomainValls:
closes #9020

### DIFF
--- a/front/src/modules/rollingStock/components/RollingStockCard/RollingStockCardButtons.tsx
+++ b/front/src/modules/rollingStock/components/RollingStockCard/RollingStockCardButtons.tsx
@@ -68,10 +68,16 @@ const RollingStockCardButtons = ({
   }, [curvesComfortList]);
 
   useEffect(() => {
-    if (curvesComfortList.includes(currentComfortInStore)) {
-      setComfort(currentComfortInStore);
+    if (curvesComfortList.length === 0) {
+      setComfort('STANDARD');
+    } else {
+      setComfort(
+        curvesComfortList.includes(currentComfortInStore)
+          ? currentComfortInStore
+          : curvesComfortList[0]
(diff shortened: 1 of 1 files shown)


front/src/modules/rollingStock/components/RollingStockCard/RollingStockCardButtons.tsx
SharglutDev, line 1: Not sure id should be in the dependency array. Also If I open back the rs selector, it switches to standard even though I selected heating for example. I don't think it's what we want.

Here is a suggestion : 

```suggestion
  useEffect(() => {
    if (curvesComfortList.length === 0) {
      setComfort('STANDARD');
    } else {
      setComfort(
        curvesComfortList.includes(currentComfortInStore)
          ? currentComfortInStore
          : curvesComfortList[0]
      );
    }
  }, [curvesComfortList, currentComfortInStore]);
```

RomainValls: Thanks, it's better this way 👍 !

SharglutDev (approved):
Lgtm and tested, thank you :)

achrafmohye (approved):
LGTM :)

PR #9041, opened 2024-09-27
editoast, osrdyne: fix the multi infra status response

Khoyo:
- Build up list querys correctly on editoast side
- Aggregate the worker status results correctly on editoast side
- Use a by worker_id map (instead of a by worker_key) in osrdyne response

### DIFF
--- a/editoast/editoast_osrdyne_client/src/lib.rs
+++ b/editoast/editoast_osrdyne_client/src/lib.rs
@@ -54,6 +54,15 @@ pub enum WorkerStatus {
     Error,
 }
 
+fn combine_worker_status(a: WorkerStatus, b: WorkerStatus) -> WorkerStatus {
+    match (a, b) {
+        (WorkerStatus::Ready, _) | (_, WorkerStatus::Ready) => WorkerStatus::Ready,
+        (WorkerStatus::Started, _) | (_, WorkerStatus::Started) => WorkerStatus::Started,
+        (WorkerStatus::Error, _) | (_, WorkerStatus::Error) => WorkerStatus::Error,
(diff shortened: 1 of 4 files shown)


woshilapin (commented):
Can you put a description in the PR, I'm not entirely sure I understand what this is supposed to fix (is there any existing bug? why did the API changed? etc.)

Khoyo:
> is there any existing bug

Yes, the multi infra endpoint simply did not work (which is the reason for the request change on editoast side) :/ 

And we didn't support the case where there is multiple workers per key, which is coming soon with the work on using Keda, so I changed up the api a bit (we now return a map indexed by worker id, not worker key - except in the noop case where we have to make stuff up anyway)

woshilapin (approved):
Thanks for the explanations. Looks good.

flomonster (approved):

editoast/editoast_osrdyne_client/src/lib.rs
```diff
@@ -110,6 +119,12 @@ struct OsrdyneStatusResponse {
 #[derive(Deserialize)]
 struct OsrdyneWorkerState {
     status: OsrdyneWorkerStatus,
+    worker_metadata: OsrdyneWorkerMetadata,
+}
+
+#[derive(Deserialize)]
+pub struct OsrdyneWorkerMetadata {
+    pub worker_key: String,
```
leovalais:
```suggestion
struct OsrdyneWorkerMetadata {
    worker_key: String,
```

younesschrifi (approved):
LGTM 👍🏽

PR #9042, opened 2024-09-27
Fix editoast payload request body limit

### LINKED ISSUE #9034: can't import large timetable files from the front-end
### What happened?

When trying to import a timetable json with 2000 trains, I get an 502 error with `Broken pipe (os error 32)`. I assume that the file is too large and something breaks along the way. Max size seems to be between 350 and 500 trains per file. 

If the file size is a problem that we can't really avoid, we could split the file into different requests from the front-end.

_No response_

Try to import a train with 2000 trains (e.g. export from a daily basic import)

Local

Firefox

c1d34dd4be266e40e8932d789a3e7c73fef662b4

flomonster:
close #9034

Check axum documentation [here](https://docs.rs/axum/latest/axum/extract/struct.DefaultBodyLimit.html#method.disable).

> [!NOTE]
> We could adapt the `route!` macro to change the request size limit per endpoint (and not globally).

### DIFF
--- a/editoast/src/main.rs
+++ b/editoast/src/main.rs
@@ -14,6 +14,7 @@ mod views;
 use crate::core::CoreClient;
 use crate::models::Infra;
 use crate::views::OpenApiRoot;
+use axum::extract::DefaultBodyLimit;
 use axum::extract::FromRef;
 use axum::{Router, ServiceExt};
 use axum_tracing_opentelemetry::middleware::OtelAxumLayer;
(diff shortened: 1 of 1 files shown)


emersion (approved):
In general it's a good practice to have a global default limit, but all heavy-load requests are authenticated anyways, and there are many other simple ways to flood the server, so I don't think it matters much.

woshilapin (approved):

PR #9095, opened 2024-09-27
front: move sass, storybook, tslib and tsx to devDependencies

emersion:
These are not direct runtime dependencies. The difference is important for licensing and security (the code of runtime deps is bundled/distributed with OSRD, it's not for dev deps).


SharglutDev (approved):
Thank you for that :)

woshilapin (approved):

Akctarus (approved):
LGTM

PR #9099, opened 2024-09-27
front: upgrade @reduxjs/toolkit

emersion:
rtk-query now always returns the "data" field in the object, but it may be undefined:

```ts
{ data: PostInfraByInfraIdApiResponse; error?: undefined; } | { data?: undefined; error: ApiError | SerializedError; }
```

Thus `'data' in response` will always be true.

### DIFF
--- a/front/yarn.lock
+++ b/front/yarn.lock
@@ -2719,14 +2719,14 @@
     immutable "^4.3.4"
 
 "@reduxjs/toolkit@^2.1.0":
-  version "2.2.3"
-  resolved "https://registry.yarnpkg.com/@reduxjs/toolkit/-/toolkit-2.2.3.tgz#5ce71cbf162f98c5dafb49bd3f1e11c5486ab9c4"
-  integrity sha512-76dll9EnJXg4EVcI5YNxZA/9hSAmZsFqzMmNRHvIlzw2WS/twfcVX3ysYrWGJMClwEmChQFC4yRq74tn6fdzRA==
+  version "2.2.7"
+  resolved "https://registry.yarnpkg.com/@reduxjs/toolkit/-/toolkit-2.2.7.tgz#199e3d10ccb39267cb5aee92c0262fd9da7fdfb2"
+  integrity sha512-faI3cZbSdFb8yv9dhDTmGwclW0vk0z5o1cia+kf7gCbaCwHI5e+7tP57mJUv22pNcNbeA62GSrPpfrUfdXcQ6g==
   dependencies:
     immer "^10.0.3"
     redux "^5.0.1"
     redux-thunk "^3.1.0"
-    reselect "^5.0.1"
+    reselect "^5.1.0"
 
 "@remix-run/router@1.19.2":
   version "1.19.2"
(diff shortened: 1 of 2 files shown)


front/yarn.lock
```diff
@@ -2659,14 +2659,14 @@
     immutable "^4.3.4"
 
 "@reduxjs/toolkit@^2.1.0":
-  version "2.2.3"
-  resolved "https://registry.yarnpkg.com/@reduxjs/toolkit/-/toolkit-2.2.3.tgz#5ce71cbf162f98c5dafb49bd3f1e11c5486ab9c4"
-  integrity sha512-76dll9EnJXg4EVcI5YNxZA/9hSAmZsFqzMmNRHvIlzw2WS/twfcVX3ysYrWGJMClwEmChQFC4yRq74tn6fdzRA==
+  version "2.2.7"
+  resolved "https://registry.yarnpkg.com/@reduxjs/toolkit/-/toolkit-2.2.7.tgz#199e3d10ccb39267cb5aee92c0262fd9da7fdfb2"
+  integrity sha512-faI3cZbSdFb8yv9dhDTmGwclW0vk0z5o1cia+kf7gCbaCwHI5e+7tP57mJUv22pNcNbeA62GSrPpfrUfdXcQ6g==
```
Khoyo, line 2724: If this bump required code change, shouldn't it be also reflected in the `package.json` file?

emersion: Should it? Sorry, not sure I'm understanding the reasoning here!

Khoyo: My reasoning here is that the `package.json` file should ideally specify the versions required for the application to work...

After this change the `@reduxjs/toolkit@^2.1.0` specifier is wrong... (and rtk-query doesn't follow semver, which is sad too)

But this may not be the way things are done in JS land

emersion: Hm, I'm not sure this will work… How about the following situation:

- `yarn upgrade` updates foo dependency from v1.0.0 to v1.1.0. This doesn't require API changes, so `package.json` is left intact.
- Some time after, a developer starts using a foo function introduced in v1.1.0.
- The version in `package.json` is still v1.0.0 but the app requires v1.1.0 to work.

To me, the version is the lock file specifies the minimum version required for the app to work. If the exact same version is installed, you're guaranteed to have a working app¹.

<small>¹ wishful thinking</small>

Khoyo: IMO, this is wrong too, the `package.json` should be updated too (either when bumping the lockfile or when actually using the new features). Otherwise, what's the value of the versions mentioned in the `package.json`?

I agree that this is hard to check :/ There is some work around that in rust (it's the purpose of [-Z minimal-versions](https://github.com/rust-lang/cargo/issues/5657), I don't know about JS.

This was probably wishful thinking on my part, and I have other windmills to fight against :p

emersion: So you're saying we should always update `package.json` when we're doing any kind of dependency update? Doesn't seem like this is what `yarn upgrade`/`npm update` does...

The `package.json` file is important to specify the version range, ie. the maximum upgradeable version: `^1.0.0` means that `yarn upgrade` won't pick `2.0.0`. More involved constraints are possible e.g. `<1.2.0`.

emersion: Side note, as a Go developer I'm personally a fan of the minimum version selection algorithm :) just doesn't seem like this is what the JS ecosystem does.

Khoyo: > The package.json file is important to specify the version range, ie. the maximum upgradeable version: ^1.0.0 means that yarn upgrade won't pick 2.0.0. More involved constraints are possible e.g. <1.2.0.

I agree, but that means the specified version ranges *should* be valid. On the rust side, we bump our direct dependencies in the Cargo.toml file in addition to the lockfile. (See eg. https://github.com/OpenRailAssociation/osrd/commit/57ed9d9598bce8510d3d24573540f4d78d0a8c1c)

In any case, I agree that this is out of scope for this PR if this wasn't how things were done before.

Yohh: it's totally ok, each upgrade will be limited to v2.x.x, since there is no v3.x.x (or we don't want it) everything will work well

Yohh (approved):
LGTM

kmer2016 (approved):

Issue #8670, opened 2024-08-30
Modify STDCM endpoint to take a train consist instead of a rolling stock


woshilapin:
Today, the STDCM search request expect a rolling stock as an input information for the physics of the train. But we will need to make that endpoint evolve in order to provide:
- a locomotive
- a towed rolling stock
- a total length
- a total weight
- a "code composition"
- a gabarit

Once this new API is exposed, `editoast` must also transform all this information into the physical information that `core` need to resolve the STDCM request.

maelysLeratRosso:
https://github.com/OpenRailAssociation/osrd/pull/9383

axrolld:
Bug found on dev environment:
https://github.com/OpenRailAssociation/osrd/issues/9530

PR #8940, opened 2024-09-19
front: stdcm add consist inputs for tonnage and length

SarahBellaha:
Closes [STDCM: Add consist inputs for tonnage and length](https://github.com/osrd-project/osrd-confidential/issues/674)

PR #9106, opened 2024-09-30
editoast: stdcm simulation convoy parameters

Wadjetz:
part of #8670 and https://github.com/osrd-project/osrd-confidential/issues/368

- [x] Send simulation length and mass to Editoast.
- [x] If parameters are provided, send them to the core; otherwise, send the rolling stock mass and length.

How to test:
- Create a new STDCM simulation with parameters and check the results.

### DIFF
--- a/editoast/src/core/simulation.rs
+++ b/editoast/src/core/simulation.rs
@@ -4,6 +4,7 @@ use std::collections::HashMap;
 use editoast_schemas::rolling_stock::EffortCurves;
 use editoast_schemas::rolling_stock::Gamma;
 use editoast_schemas::rolling_stock::RollingResistance;
+use editoast_schemas::rolling_stock::RollingStock;
 use editoast_schemas::train_schedule::Comfort;
 use editoast_schemas::train_schedule::Distribution;
 use editoast_schemas::train_schedule::MarginValue;
@@ -17,7 +18,6 @@ use super::pathfinding::TrackRange;
 use crate::core::pathfinding::PathfindingResult;
 use crate::core::{AsCoreRequest, Json};
 use crate::error::InternalError;
-use crate::RollingStockModel;
 use derivative::Derivative;
 use editoast_schemas::primitives::Identifier;
 use std::hash::Hash;
@@ -66,25 +66,47 @@ pub struct PhysicsRollingStock {
     pub raise_pantograph_time: Option<u64>,
 }
 
-impl From<RollingStockModel> for PhysicsRollingStock {
-    fn from(value: RollingStockModel) -> Self {
+#[derive(Debug, Default)]
+pub struct SimulationParameters {
+    pub total_mass: Option<f64>,
+    pub total_length: Option<f64>,
+    pub max_speed: Option<f64>,
+}
+
+impl PhysicsRollingStock {
+    pub fn new(traction_engine: RollingStock, params: SimulationParameters) -> Self {
+        let traction_engine_length = traction_engine.length * 1000.0;
+        let length = params
+            .total_length
+            .map(|tl| tl * 1000.0)
+            .unwrap_or(traction_engine_length)
+            .round() as u64;
+
+        let traction_engine_mass = traction_engine.mass;
+        let mass = params.total_mass.unwrap_or(traction_engine_mass).round() as u64;
+
+        let max_speed = f64::min(
+            traction_engine.max_speed,
+            params.max_speed.unwrap_or(traction_engine.max_speed),
+        );
+
         Self {
-            effort_curves: value.effort_curves,
-            base_power_class: value.base_power_class,
-            length: (value.length * 1000.0).round() as u64,
-            max_speed: value.max_speed,
-            startup_time: (value.startup_time * 1000.0).round() as u64,
-            startup_acceleration: value.startup_acceleration,
-            comfort_acceleration: value.comfort_acceleration,
-            gamma: value.gamma,
-            inertia_coefficient: value.inertia_coefficient,
-            mass: value.mass.round() as u64,
-            rolling_resistance: value.rolling_resistance,
-            power_restrictions: value.power_restrictions.into_iter().collect(),
-            electrical_power_startup_time: value
+            effort_curves: traction_engine.effort_curves,
+            base_power_class: traction_engine.base_power_class,
+            length,
+            mass,
+            max_speed,
+            startup_time: (traction_engine.startup_time * 1000.0).round() as u64,
+            startup_acceleration: traction_engine.startup_acceleration,
+            comfort_acceleration: traction_engine.comfort_acceleration,
+            gamma: traction_engine.gamma,
+            inertia_coefficient: traction_engine.inertia_coefficient,
+            rolling_resistance: traction_engine.rolling_resistance,
+            power_restrictions: traction_engine.power_restrictions.into_iter().collect(),
+            electrical_power_startup_time: traction_engine
                 .electrical_power_startup_time
                 .map(|v| (v * 1000.0).round() as u64),
-            raise_pantograph_time: value
+            raise_pantograph_time: traction_engine
                 .raise_pantograph_time
                 .map(|v| (v * 1000.0).round() as u64),
         }
--- a/editoast/src/views/timetable/stdcm.rs
+++ b/editoast/src/views/timetable/stdcm.rs
@@ -28,6 +28,8 @@ use super::SelectionSettings;
 use crate::core::conflict_detection::TrainRequirements;
 use crate::core::pathfinding::InvalidPathItem;
 use crate::core::pathfinding::PathfindingResult;
+use crate::core::simulation::PhysicsRollingStock;
+use crate::core::simulation::SimulationParameters;
 use crate::core::simulation::{RoutingRequirement, SimulationResponse, SpacingRequirement};
 use crate::core::stdcm::STDCMResponse;
 use crate::core::stdcm::TemporarySpeedLimit as CoreTemporarySpeedLimit;
@@ -112,6 +114,22 @@ pub struct STDCMRequestPayload {
     #[serde(default)]
     #[schema(value_type = Option<String>, example = json!(["5%", "2min/100km"]))]
     margin: Option<MarginValue>,
+    /// Total mass of the consist in kg
+    total_mass: Option<f64>,
+    /// Total length of the consist in meters
+    total_length: Option<f64>,
+    /// Maximum speed of the consist in km/h
+    max_speed: Option<f64>,
+}
+
+impl STDCMRequestPayload {
+    pub fn simulation_parameters(&self) -> SimulationParameters {
+        SimulationParameters {
+            total_mass: self.total_mass,
+            total_length: self.total_length,
+            max_speed: self.max_speed,
+        }
+    }
 }
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone, ToSchema)]
@@ -173,6 +191,7 @@ async fn stdcm(
     let core_client = app_state.core_client.clone();
     let timetable_id = id;
     let infra_id = query.infra;
+    let simulation_parameters = stdcm_request.simulation_parameters();
 
     // 1. Retrieve Timetable / Infra / Trains / Simulation / Rolling Stock
     let timetable_trains = TimetableWithTrains::retrieve_or_fail(conn, timetable_id, || {
@@ -268,7 +287,6 @@ async fn stdcm(
     let stdcm_response = STDCMRequest {
         infra: infra.id,
         expected_version: infra.version,
-        rolling_stock: rolling_stock.clone().into(),
         rolling_stock_loading_gauge: rolling_stock.loading_gauge,
         rolling_stock_supported_signaling_systems: rolling_stock
             .supported_signaling_systems
@@ -297,6 +315,7 @@ async fn stdcm(
             None => vec![],
         },
         temporary_speed_limits,
+        rolling_stock: PhysicsRollingStock::new(rolling_stock.into(), simulation_parameters),
     }
     .fetch(core_client.as_ref())
     .await?;
@@ -641,4 +660,90 @@ async fn parse_stdcm_steps(
 enum SimulationTimeResult {
     SimulationTime { value: u64 },
     Error { error: Box<SimulationResponse> },
+}
+
+#[cfg(test)]
+mod tests {
+    use pretty_assertions::assert_eq;
+
+    use crate::views::rolling_stock::tests::fast_rolling_stock_form;
+
+    use super::*;
+
+    fn create_test_rolling_stock() -> RollingStockModel {
+        let rs_form = fast_rolling_stock_form("test_rolling_stock");
+        RollingStockModel {
+            id: 5656565,
+            name: rs_form.name,
+            loading_gauge: rs_form.loading_gauge,
+            supported_signaling_systems: rs_form.supported_signaling_systems,
+            version: 33,
+            base_power_class: rs_form.base_power_class,
+            comfort_acceleration: rs_form.comfort_acceleration,
+            inertia_coefficient: rs_form.inertia_coefficient,
+            startup_acceleration: rs_form.startup_acceleration,
+            startup_time: rs_form.startup_time,
+            effort_curves: rs_form.effort_curves,
+            electrical_power_startup_time: rs_form.electrical_power_startup_time,
+            raise_pantograph_time: rs_form.raise_pantograph_time,
+            energy_sources: rs_form.energy_sources,
+            gamma: rs_form.gamma,
+            locked: false,
+            metadata: rs_form.metadata,
+            power_restrictions: rs_form.power_restrictions,
+            railjson_version: "12".to_string(),
+            rolling_resistance: rs_form.rolling_resistance,
+            length: 140.0,   // m
+            mass: 150000.0,  // kg
+            max_speed: 20.0, // m/s
+        }
+    }
+
+    #[test]
+    fn simulation_with_parameters() {
+        let rolling_stock = create_test_rolling_stock();
+
+        let simulation_parameters = SimulationParameters {
+            total_mass: Some(123.0),
+            total_length: Some(455.0),
+            max_speed: Some(10.0), // m/s
+        };
+
+        let physics_rolling_stock: PhysicsRollingStock =
+            PhysicsRollingStock::new(rolling_stock.into(), simulation_parameters);
+
+        assert_eq!(physics_rolling_stock.mass, 123_u64);
+        assert_eq!(physics_rolling_stock.length, 455000_u64); // It should be converted in mm
+        assert_eq!(physics_rolling_stock.max_speed, 10_f64); // It should be in m/s
+    }
+
+    #[test]
+    fn simulation_without_parameters() {
+        let rolling_stock = create_test_rolling_stock();
+
+        let simulation_parameters = SimulationParameters::default();
+
+        let physics_rolling_stock: PhysicsRollingStock =
+            PhysicsRollingStock::new(rolling_stock.into(), simulation_parameters);
+
+        assert_eq!(physics_rolling_stock.mass, 150000_u64);
+        assert_eq!(physics_rolling_stock.length, 140000_u64); // It should be converted in mm
+        assert_eq!(physics_rolling_stock.max_speed, 20_f64);
+    }
+
+    #[test]
+    fn new_physics_rolling_stock_keeps_the_smallest_available_max_speed() {
+        let rolling_stock = create_test_rolling_stock();
+
+        let simulation_parameters = SimulationParameters {
+            total_mass: None,
+            total_length: None,
+            max_speed: Some(30.0), // m/s
+        };
+
+        let physics_rolling_stock: PhysicsRollingStock =
+            PhysicsRollingStock::new(rolling_stock.into(), simulation_parameters);
+
+        assert_eq!(physics_rolling_stock.max_speed, 20_f64);
+    }
 }
--- a/front/src/applications/stdcm/utils/formatStdcmConfV2.ts
+++ b/front/src/applications/stdcm/utils/formatStdcmConfV2.ts
@@ -12,7 +12,7 @@ import type { InfraState } from 'reducers/infra';
 import { setFailure } from 'reducers/main';
 import type { OsrdStdcmConfState, StandardAllowance } from 'reducers/osrdconf/types';
 import { dateTimeFormatting, dateTimeToIso } from 'utils/date';
-import { mToMm } from 'utils/physics';
+import { kmhToMs, mToMm, tToKg } from 'utils/physics';
 import { ISO8601Duration2sec, sec2ms, time2sec } from 'utils/timeManipulation';
 
 import createMargin from './createMargin';
@@ -26,6 +26,9 @@ type ValidStdcmConfig = {
   rollingStockComfort: TrainScheduleBase['comfort'];
   path: PathfindingItem[];
   speedLimitByTag?: string;
+  totalMass?: number;
+  totalLength?: number;
+  maxSpeed?: number;
   maximumRunTime?: number;
   startTime?: string; // must be a datetime
   latestStartTime?: string;
@@ -61,6 +64,9 @@ export const checkStdcmConf = (
     searchDatetimeWindow,
     workScheduleGroupId,
     electricalProfileSetId,
+    totalLength,
+    totalMass,
+    maxSpeed,
   } = osrdconf;
   let error = false;
   if (pathSteps[0] === null) {
@@ -224,6 +230,9 @@ export const checkStdcmConf = (
       maximumRunTime,
     }),
     speedLimitByTag,
+    totalMass,
+    totalLength,
+    maxSpeed,
     margin: standardStdcmAllowance,
     gridMarginBefore,
     gridMarginAfter,
@@ -246,6 +255,9 @@ export const formatStdcmPayload = (
     margin: createMargin(validConfig.margin),
     rolling_stock_id: validConfig.rollingStockId,
     speed_limit_tags: validConfig.speedLimitByTag,
+    total_mass: validConfig.totalMass ? tToKg(validConfig.totalMass) : undefined,
+    max_speed: validConfig.maxSpeed ? kmhToMs(validConfig.maxSpeed) : undefined,
+    total_length: validConfig.totalLength,
     ...(!stdcmV2Activated && {
       maximum_run_time: toMsOrUndefined(validConfig.maximumRunTime),
       maximum_departure_delay: sec2ms(
--- a/front/src/applications/stdcmV2/components/StdcmConsist.tsx
+++ b/front/src/applications/stdcmV2/components/StdcmConsist.tsx
@@ -10,11 +10,12 @@ import { useStoreDataForSpeedLimitByTagSelector } from 'common/SpeedLimitByTagSe
 import RollingStock2Img from 'modules/rollingStock/components/RollingStock2Img';
 import { useStoreDataForRollingStockSelector } from 'modules/rollingStock/components/RollingStockSelector/useStoreDataForRollingStockSelector';
 import useFilterRollingStock from 'modules/rollingStock/hooks/useFilterRollingStock';
-import type { StdcmConfSliceActions } from 'reducers/osrdconf/stdcmConf';
+import { type StdcmConfSliceActions } from 'reducers/osrdconf/stdcmConf';
 import { useAppDispatch } from 'store';
 
-import StdcmCard from './StdcmCard';
 import type { StdcmConfigCardProps } from '../types';
+import StdcmCard from './StdcmCard';
+import useStdcmConsist from '../hooks/useStdcmConsist';
 
 const ConsistCardTitle = ({
   rollingStock,
@@ -39,6 +40,15 @@ const StdcmConsist = ({ setCurrentSimulationInputs, disabled = false }: StdcmCon
   const dispatch = useAppDispatch();
 
   const { rollingStock } = useStoreDataForRollingStockSelector();
+
+  const {
+    totalMass,
+    onTotalMassChange,
+    totalLength,
+    onTotalLengthChange,
+    maxSpeed,
+    onMaxSpeedChange,
+  } = useStdcmConsist();
 
   const { filters, searchRollingStock, searchRollingStockById, filteredRollingStockList } =
     useFilterRollingStock({ isStdcm: true });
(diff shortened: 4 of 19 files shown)


editoast/src/core/simulation.rs
```diff
@@ -66,6 +66,47 @@ pub struct PhysicsRollingStock {
...
 
+pub struct SimulationParameters {
+    pub tonnage: Option<f64>,
+    pub total_length: Option<f64>,
+}
+
+impl PhysicsRollingStock {
+    /// Create a new PhysicsRollingStock from a RollingStockModel and SimulationParameters
+    pub fn new(traction_engine: RollingStockModel, params: SimulationParameters) -> Self {
+        let traction_engine_length = traction_engine.length * 1000.0; // RollingStockModel length is in km, convert to m
```
leovalais: The [schema doc of RollingStock](https://github.com/OpenRailAssociation/osrd/blob/78160a257541d772fb4df95ff92b3218b463ea2d/python/osrd_schemas/osrd_schemas/rolling_stock.py#L205) says that the length is in meters.
Additionally the documentation of `PhysicsRollingStock` states the the length is in millimeters: https://github.com/OpenRailAssociation/osrd/blob/78160a257541d772fb4df95ff92b3218b463ea2d/editoast/src/core/simulation.rs#L41-L42

I guess this is why the `* 1000` works but the comments are misleading.

editoast/src/core/simulation.rs
```diff
@@ -66,6 +66,47 @@ pub struct PhysicsRollingStock {
...
+                .electrical_power_startup_time
+                .map(|v| (v * 1000.0).round() as u64),
+            raise_pantograph_time: traction_engine
+                .raise_pantograph_time
+                .map(|v| (v * 1000.0).round() as u64),
+        }
+    }
+}
+
 impl From<RollingStockModel> for PhysicsRollingStock {
```
leovalais: This implementation should be removed since I don't believe it's used anymore. If it is, better call `::new(value, Default::default())` instead.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -245,9 +258,12 @@ async fn stdcm(
 
     // 5. Build STDCM request
     let stdcm_response = STDCMRequest {
+        rolling_stock: PhysicsRollingStock::new(
+            rolling_stock.clone(),
+            stdcm_request.clone().into(),
```
leovalais: Better build the object from the request instead of using `From` to avoid needlessly cloning the whole request.

Moving the `rolling_stock: ...` a few lines below may also help getting rid of the `rolling_stock.clone()`.

leovalais (changes requested):
Thanks! A few comments:

editoast/src/core/simulation.rs
Wadjetz: yes, it looks like a typo in the comment

editoast/src/views/timetable/stdcm.rs
```diff
@@ -107,6 +109,19 @@ pub struct STDCMRequestPayload {
     #[serde(default)]
     #[schema(value_type = Option<String>, example = json!(["5%", "2min/100km"]))]
     margin: Option<MarginValue>,
+    /// Total tonnage of the consist in tons
+    tonnage: Option<f64>,
```
flomonster: Even if the user interface request tons. We should keep use consistent unit over the API.

```suggestion
    /// Total mass of the consist in kg
    mass: Option<f64>,
```

flomonster (changes requested):
Thanks for this PR

editoast/src/core/simulation.rs
leovalais: Well if the simulation's length unit is actually meters as you corrected in the comment, I don't get the `* 1000` at all. Take `2TGVA` which `length` in the database is `476`. I hardly believe that's kilometers 😅

editoast/src/views/timetable/stdcm.rs
```diff
@@ -258,7 +272,7 @@ async fn stdcm(
         trains_requirements,
         maximum_departure_delay,
         maximum_run_time,
-        speed_limit_tag: stdcm_request.speed_limit_tags,
+        speed_limit_tag: stdcm_request.speed_limit_tags.clone(),
```
leovalais: Is this `clone()` necessary?

leovalais (commented):
Thanks for the modifications!

front/src/applications/stdcmV2/components/StdcmConsist.tsx
```diff
@@ -35,9 +37,28 @@ const StdcmConsist = ({ setCurrentSimulationInputs, disabled = false }: StdcmCon
...
+  const { updateRollingStockID, updateTotalMass, updateTotalLength } =
+    useOsrdConfActions() as StdcmConfSliceActions;
   const dispatch = useAppDispatch();
 
+  const { getTotalMass, getTotalLength } = useOsrdConfSelectors();
+  const totalMass = useSelector(getTotalMass);
+  const totalLength = useSelector(getTotalLength);
+  const onTotalMassChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    // TODO better handling of number in input component
+    const totalMassValue = parseInt(e.target.value, 10);
```
kmer2016: If the field can be a decimal number, use `Number(e.target.value)` to cast the value. The current approach with `parseInt` does not allow the user to enter decimal values.

front/src/applications/stdcmV2/components/StdcmConsist.tsx
```diff
@@ -35,9 +37,28 @@ const StdcmConsist = ({ setCurrentSimulationInputs, disabled = false }: StdcmCon
...
+    if (!isNil(totalMassValue)) {
+      dispatch(updateTotalMass(totalMassValue));
+    }
+  };
+  const onTotalLengthChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    // TODO better handling of number in input component
+    const totalLengthValue = parseInt(e.target.value, 10);
+    if (!isNil(totalLengthValue)) {
+      dispatch(updateTotalLength(totalLengthValue));
```
kmer2016: same here

editoast/src/core/simulation.rs
Wadjetz: The * 1000 was there before my changes as well. I guess I didn’t understand it correctly

front/src/applications/stdcmV2/components/StdcmConsist.tsx
```diff
@@ -125,8 +146,24 @@ const StdcmConsist = ({ setCurrentSimulationInputs, disabled = false }: StdcmCon
...
         <div className="stdcm-v2-consist__properties d-flex justify-content-between">
-          <Input id="tonnage" label={t('consist.tonnage')} trailingContent="t" />
-          <Input id="length" label={t('consist.length')} trailingContent="m" />
+          <Input
+            id="tonnage"
+            label={t('consist.tonnage')}
+            trailingContent="t"
+            type="number"
+            min={0}
+            value={totalMass === null ? undefined : totalMass}
```
kmer2016: As `totalMass` has the type `number | undefined`, there is no need to check if it is equal to null.
```suggestion
            value={totalMass}
```

front/src/applications/stdcmV2/components/StdcmConsist.tsx
```diff
@@ -125,8 +146,24 @@ const StdcmConsist = ({ setCurrentSimulationInputs, disabled = false }: StdcmCon
...
+            onChange={onTotalMassChange}
+          />
+          <Input
+            id="length"
+            label={t('consist.length')}
+            trailingContent="m"
+            type="number"
+            min={0}
+            value={totalLength === null ? undefined : totalLength}
```
kmer2016:
```suggestion
            value={totalLength}
```

kmer2016 (dismissed):
Thank you! I have some comments.

For the changes related to the store, could you please move these to the appropriate store section? 

The `stdcm` and `operational studies` apps share many store fields, but there are also fields that are not shared. The store in `osrdConfCommon` contains the parts common to both apps. Since your changes mostly concern the `stdcm` app, they should be in the `stdcm` store section.

Could you please:
- Move the two new action creators `updateTotalMass` and `updateTotalLength` to the `stdcm` slice located in `src/reducers/osrdconf/stdcmConf/index.ts`.
- Move the selectors `getTotalMass` and `getTotalLength` to the `stdcm` selectors in `src/reducers/osrdconf/stdcmConf/selectors.ts`.
- Move the new fields `totalMass` and `totalLength` to the `OsrdStdcmConfState` type instead of `OsrdConfState`?
- Add some tests for the new action creators in `stdcmConfReducers.spec.ts` file?

editoast/src/views/timetable/stdcm.rs
woshilapin: On `front`, `tonnage` has been used. We should be consistent. See https://github.com/OpenRailAssociation/osrd/pull/8940#discussion_r1784422643.

editoast/src/core/simulation.rs
Wadjetz: Maybe @flomonster can help to undestand this ?

editoast/src/views/timetable/stdcm.rs
```diff
@@ -107,6 +109,19 @@ pub struct STDCMRequestPayload {
     #[serde(default)]
     #[schema(value_type = Option<String>, example = json!(["5%", "2min/100km"]))]
     margin: Option<MarginValue>,
+    /// Total mass of the consist in kg
+    total_mass: Option<f64>,
+    /// Total length of the consist in meters
+    total_length: Option<f64>,
```
woshilapin, line 120: To me, the form now also takes a `max_speed` and `composition_code` (or whatever its name in English). See the form.

353865595-5628e1f9-8373-4aed-8c15-dbd35dcf736e

editoast/src/views/timetable/stdcm.rs
```diff
@@ -275,6 +290,7 @@ async fn stdcm(
             }
             None => vec![],
         },
+        rolling_stock: PhysicsRollingStock::new(rolling_stock, simulation_parameters),
```
woshilapin: Would have been nice to keep the line where it was, the diff would have been easier to understand.

editoast/src/core/simulation.rs
```diff
@@ -66,25 +66,41 @@ pub struct PhysicsRollingStock {
...
+}
+
+impl PhysicsRollingStock {
+    /// Create a new consist with simulation parameters
+    pub fn new(traction_engine: RollingStockModel, params: SimulationParameters) -> Self {
+        let traction_engine_length = traction_engine.length * 1000.0;
+        let length = params
+            .total_length
+            .unwrap_or(traction_engine_length)
+            .round() as u64;
```
Khoyo: Why the *1000 only in a single case? Assuming it's correct, I'm not sure having naked floats with two different length units is a good idea...

editoast/src/core/simulation.rs
leovalais: Rolling stocks' length in editoast (schema and model) are definitely meters and the comment on `PhysicsRollingStock` states that core needs millimeters. So everything is likely fine with the `* 1000` (we'd have noticed that simulations are wrong otherwise ^^).

However since editoast is increasingly doing more calculations I believe it's time to think about managing units like core does. I think [uom](https://crates.io/crates/uom) is a good candidate. This has to be discussed, and it's a topic for another PR.

editoast/src/core/simulation.rs
leovalais: https://github.com/OpenRailAssociation/osrd/pull/9106#discussion_r1784504055

editoast/src/core/simulation.rs
Khoyo: Is `params.total_length` in meters or in millimeters? If millimeters, it should probably be an int, and if meters, the code is wrong (it needs to be multiplied too).

(I'm also in strong agreement with any effort to bring units into the type system)

editoast/src/views/timetable/stdcm.rs
woshilapin, line 1: Would be nice to add (or update) the tests on this endpoint.

woshilapin (changes requested):
Thanks for the full-stack work. I'm putting a "Request changes" for 2 reasons:
- it's missing `max_speed` and `composition_code`
- it's missing some tests

editoast/src/core/simulation.rs
flomonster: I agree, we should start using types with units. Whenever we are using `mm` it should be stored as `u64`

editoast/src/views/timetable/stdcm.rs
flomonster: You mean the ID of the `Input` or in the user interface?

> [!NOTE]
> You commented a closed PR :/

editoast/src/views/timetable/stdcm.rs
Wadjetz: sorry, we try to avoid .clone()

editoast/src/views/timetable/stdcm.rs
woshilapin: Yes, I know I did comment a closed PR, don't think it should be a problem to come back to something already merged.

About "tonnage" vs "mass", do we want something different displayed to the user that what we use internally in variable names and identifiers? If yes, is there a compelling reason to do so?

front/src/applications/stdcmV2/components/StdcmConsist.tsx
```diff
@@ -35,9 +37,24 @@ const StdcmConsist = ({ setCurrentSimulationInputs, disabled = false }: StdcmCon
...
+  const { updateRollingStockID, updateTotalMass, updateTotalLength } =
+    useOsrdConfActions() as StdcmConfSliceActions;
   const dispatch = useAppDispatch();
 
+  const { getTotalMass, getTotalLength } = useOsrdConfSelectors() as StdcmConfSelectors;
+
+  const totalMass = useSelector(getTotalMass);
+  const onTotalMassChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const totalMassValue = Number(e.target.value);
+    dispatch(updateTotalMass(totalMassValue === 0 ? undefined : Number(e.target.value)));
```
kmer2016:
```suggestion
    dispatch(updateTotalMass(totalMassValue === 0 ? undefined : totalMassValue));
```

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -25,6 +25,18 @@ export const stdcmConfSlice = createSlice({
...
+      action: PayloadAction<OsrdStdcmConfState['totalMass'] | null>
+    ) {
+      state.totalMass = action.payload === null ? undefined : action.payload;
+    },
+    updateTotalLength(
+      state: Draft<OsrdStdcmConfState>,
+      action: PayloadAction<OsrdStdcmConfState['totalLength'] | null>
+    ) {
+      state.totalLength = action.payload === null ? undefined : action.payload;
+    },
```
kmer2016: I think we don't need to specify `null` in the type. Simplifies the code
```suggestion
    updateTotalMass(
      state: Draft<OsrdStdcmConfState>,
      action: PayloadAction<OsrdStdcmConfState['totalMass']>
    ) {
      state.totalMass = action.payload;
    },
    updateTotalLength(
      state: Draft<OsrdStdcmConfState>,
      action: PayloadAction<OsrdStdcmConfState['totalLength']>
    ) {
      state.totalLength = action.payload;
    },
```

front/src/reducers/osrdconf/types.ts
```diff
@@ -29,6 +29,8 @@ export interface OsrdConfState extends InfraState {
   searchDatetimeWindow?: { begin: Date; end: Date };
   rollingStockID?: number;
   speedLimitByTag?: string;
+  totalMass?: number;
+  totalLength?: number;
```
kmer2016: These fields are specific to `stdcm`, so they don't need to be in `OsrdConfState`.
```suggestion
```

editoast/src/views/timetable/stdcm.rs
flomonster: > About "tonnage" vs "mass", do we want something different displayed to the user that what we use internally in variable names and identifiers? If yes, is there a compelling reason to do so?

The frontend is not the only user of our API. I prefer a consistent API, `kg` is what we're using for rolling stocks.

editoast/src/core/simulation.rs
Wadjetz: I looked into this library, and it's great for computations, but not very practical for API building and Serde integration is not easy. Maybe we should organize a workshop ?

editoast/src/core/simulation.rs
```diff
@@ -66,25 +66,45 @@ pub struct PhysicsRollingStock {
...
+#[derive(Debug, Default)]
+pub struct SimulationParameters {
+    pub total_mass: Option<f64>,
+    pub total_length: Option<f64>,
+    pub max_speed: Option<f64>,
+}
+
+impl PhysicsRollingStock {
+    /// Create a new consist with simulation parameters
+    pub fn new(traction_engine: RollingStockModel, params: SimulationParameters) -> Self {
```
leovalais:
```suggestion
    pub fn new(traction_engine: RollingStock, params: SimulationParameters) -> Self {
```

Can we use the schema `RollingStock` instead of the model?

editoast/src/views/timetable/stdcm.rs
woshilapin: Tests added, thank you, I'm resolving.

flomonster (approved):

leovalais (approved):
Thanks for the changes!

woshilapin (approved):
Thanks for the tests and other fields.

editoast/src/core/simulation.rs
```diff
@@ -66,25 +66,48 @@ pub struct PhysicsRollingStock {
...
-    fn from(value: RollingStockModel) -> Self {
+#[derive(Debug, Default)]
+pub struct SimulationParameters {
+    pub total_mass: Option<f64>,
+    pub total_length: Option<f64>,
+    pub max_speed: Option<f64>,
+}
+
+impl PhysicsRollingStock {
+    /// Create a new consist with simulation parameters
```
Sh099078: Maybe we should remove that comment ? What the `new` method is doing is quite self-explanatory and the current comment is unclear.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -642,3 +661,89 @@ enum SimulationTimeResult {
...
+            energy_sources: rs_form.energy_sources,
+            gamma: rs_form.gamma,
+            locked: false,
+            metadata: rs_form.metadata,
+            power_restrictions: rs_form.power_restrictions,
+            railjson_version: "12".to_string(),
+            rolling_resistance: rs_form.rolling_resistance,
+            length: 140.0,       // m
+            mass: 150000.0,      // kg
+            max_speed: 110000.0, // m/s
```
Sh099078: mm/s* ? 110 000 m/s seems a bit fast !

editoast/src/views/timetable/stdcm.rs
```diff
@@ -642,3 +661,89 @@ enum SimulationTimeResult {
...
+    }
+
+    #[test]
+    fn simulation_with_parameters() {
+        let rolling_stock = create_test_rolling_stock();
+
+        let simulation_parameters = SimulationParameters {
+            total_mass: Some(123.0),
+            total_length: Some(455.0),
+            max_speed: Some(100000.0), // m/s
```
Sh099078: Same comment as above: `mm/s` ?

editoast/src/views/timetable/stdcm.rs
```diff
@@ -642,3 +661,89 @@ enum SimulationTimeResult {
...
+    }
+
+    #[test]
+    fn max_speed_min_parameters() {
+        let rolling_stock = create_test_rolling_stock();
+
+        let simulation_parameters = SimulationParameters {
+            total_mass: None,
+            total_length: None,
+            max_speed: Some(130000.0), // m/s
```
Sh099078: `mm/s` ?

front/src/reducers/osrdconf/stdcmConf/index.ts
SharglutDev, line 1: Seems like the store isn't updated when we enter an input.
You need to initialize the store with some values after l.12

```suggestion
  totalMass: undefined,
  totalLength: undefined,
  maxSpeed: undefined,
```

editoast/src/views/timetable/stdcm.rs
```diff
@@ -642,3 +661,89 @@ enum SimulationTimeResult {
...
+        let physics_rolling_stock: PhysicsRollingStock =
+            PhysicsRollingStock::new(rolling_stock.into(), simulation_parameters);
+
+        assert_eq!(physics_rolling_stock.mass, 150000_u64);
+        assert_eq!(physics_rolling_stock.length, 140000_u64); // It should be converted in mm
+        assert_eq!(physics_rolling_stock.max_speed, 110000_f64);
+    }
+
+    #[test]
+    fn max_speed_min_parameters() {
```
Sh099078: The naming was a bit confusing to me. Maybe something along the lines of `new_physics_rolling_stock_keeps_the_smallest_available_max_speed` would help understand better the expected behavior that is tested without the need to dive into the code ? At the first read, the value in the `assert_eq` looked like a typo and I had to go into `create_test_rolling_stock` to find the `11000` and understand what the test was about.

Sh099078 (commented):
Great PR ! I left some comments (mostly typos)

front/src/styles/scss/applications/stdcmV2/_card.scss
SharglutDev, line 1: We should not change the size and spacing of these components, they don't match the mockup anymore.
Do you have access to sketch ?

image

front/src/applications/stdcmV2/components/StdcmConsist.tsx
```diff
@@ -129,22 +153,38 @@ const StdcmConsist = ({ setCurrentSimulationInputs, disabled = false }: StdcmCon
           id="tonnage"
           label={t('consist.tonnage')}
           trailingContent="t"
-          inputFieldWrapperClassname="weight"
+          type="number"
+          min={0}
```
SharglutDev, line 143: I can put negative number, we should do something about that probably. @thibautsailly how do you want to handle this ?

front/src/applications/stdcmV2/components/StdcmConsist.tsx
```diff
@@ -35,9 +37,31 @@ const StdcmConsist = ({ setCurrentSimulationInputs, disabled = false }: StdcmCon
...
+  const onTotalLengthChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const totalLengthValue = Number(e.target.value);
+    dispatch(updateTotalLength(totalLengthValue === 0 ? undefined : totalLengthValue));
+  };
+
+  const maxSpeed = useSelector(getMaxSpeed);
+  const onMaxSpeedChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const totalMaxSpeed = Number(e.target.value);
+    dispatch(updateMaxSpeed(totalMaxSpeed === 0 ? undefined : totalMaxSpeed));
+  };
```
SharglutDev: Maybe we could created a custom hook `useStdcmConsist` that returns all these values ? (at least for selector and actions)

front/src/applications/stdcm/utils/formatStdcmConfV2.ts
```diff
@@ -224,6 +227,9 @@ export const checkStdcmConf = (
       maximumRunTime,
     }),
     speedLimitByTag,
+    totalMass: osrdconf.totalMass,
+    totalLength: osrdconf.totalLength,
+    maxSpeed: osrdconf.maxSpeed,
```
SharglutDev: We destructure `osrdConf` l.67, you can do the same with these 3 values

front/src/applications/stdcmV2/components/StdcmConsist.tsx
```diff
@@ -129,22 +153,38 @@ const StdcmConsist = ({ setCurrentSimulationInputs, disabled = false }: StdcmCon
...
+        <SpeedLimitByTagSelector
+          disabled={disabled}
+          selectedSpeedLimitByTag={speedLimitByTag}
+          speedLimitsByTags={speedLimitsByTags}
+          dispatchUpdateSpeedLimitByTag={dispatchUpdateSpeedLimitByTag}
+        />
+        <Input
+          id="maxSpeed"
+          label={t('consist.maxSpeed')}
+          trailingContent="m/s"
```
SharglutDev: Mockup says `km/h`

front/src/applications/stdcm/utils/formatStdcmConfV2.ts
```diff
@@ -246,6 +252,9 @@ export const formatStdcmPayload = (
     margin: createMargin(validConfig.margin),
     rolling_stock_id: validConfig.rollingStockId,
     speed_limit_tags: validConfig.speedLimitByTag,
+    total_mass: validConfig.totalMass ? validConfig.totalMass * 1000 : undefined, // Api require total_mass in kg
+    max_speed: validConfig.maxSpeed ? validConfig.maxSpeed * 1000 : undefined, // The form is in km/s but the Api is in m/s
```
SharglutDev: The form is (should be, see other comment) in km/h. Use the util `kmhToMs` in `physics.ts`

front/src/applications/stdcm/utils/formatStdcmConfV2.ts
```diff
@@ -246,6 +252,9 @@ export const formatStdcmPayload = (
     margin: createMargin(validConfig.margin),
     rolling_stock_id: validConfig.rollingStockId,
     speed_limit_tags: validConfig.speedLimitByTag,
+    total_mass: validConfig.totalMass ? validConfig.totalMass * 1000 : undefined, // Api require total_mass in kg
```
SharglutDev: You can add and use a util in `physics.ts` : `tToKg`

SharglutDev (changes requested):

SharglutDev:
Also depending on who finishes first, you might have conflicts with this PR : https://github.com/OpenRailAssociation/osrd/pull/9268

editoast/src/views/timetable/stdcm.rs
Wadjetz: it in m/s, https://github.com/OpenRailAssociation/osrd/blob/78160a257541d772fb4df95ff92b3218b463ea2d/python/osrd_schemas/osrd_schemas/rolling_stock.py#L206

The frontend rolling stock form also converts it and sends it in m/s.

leovalais: I suppose this comment is about the chosen speed being unrealistically fast 🚀

front/src/applications/stdcmV2/components/StdcmConsist.tsx
Wadjetz: Yes, the min props is not a validator, but it works when you use the up and down keys.

front/src/applications/stdcm/utils/formatStdcmConfV2.ts
```diff
@@ -246,6 +255,9 @@ export const formatStdcmPayload = (
     margin: createMargin(validConfig.margin),
     rolling_stock_id: validConfig.rollingStockId,
     speed_limit_tags: validConfig.speedLimitByTag,
+    total_mass: validConfig.totalMass ? tToKg(validConfig.totalMass) : undefined, // Api require total_mass in kg
+    max_speed: validConfig.maxSpeed ? kmhToMs(validConfig.maxSpeed) : undefined, // The form is in km/s but the Api is in m/s
```
SharglutDev: Now that we use the utils, you can remove the comments

front/src/applications/stdcmV2/components/StdcmConsist.tsx
```diff
@@ -129,24 +140,75 @@ const StdcmConsist = ({ setCurrentSimulationInputs, disabled = false }: StdcmCon
...
-        selectedSpeedLimitByTag={speedLimitByTag}
-        speedLimitsByTags={speedLimitsByTags}
-        dispatchUpdateSpeedLimitByTag={dispatchUpdateSpeedLimitByTag}
-        className="speed-limit-by-tag-selector"
-      />
     </StdcmCard>
   );
 };
 
+function useStdcmConsist() {
```
SharglutDev: Thank you for the change. Could you move the hook in a `useStdcmConsist.ts` file in the "hooks" folder ? (in stdcmV2 folder)

front/src/applications/stdcmV2/components/StdcmConsist.tsx
```diff
@@ -129,24 +140,75 @@ const StdcmConsist = ({ setCurrentSimulationInputs, disabled = false }: StdcmCon
...
+        <SpeedLimitByTagSelector
+          disabled={disabled}
+          selectedSpeedLimitByTag={speedLimitByTag}
+          speedLimitsByTags={speedLimitsByTags}
+          dispatchUpdateSpeedLimitByTag={dispatchUpdateSpeedLimitByTag}
+        />
+        <Input
+          id="maxSpeed"
+          label={t('consist.maxSpeed')}
+          trailingContent="km/s"
```
SharglutDev:
```suggestion
          trailingContent="km/h"
```

front/src/styles/scss/applications/stdcmV2/_card.scss
SharglutDev: Thank you for the changes but we should still try to pixel perfect. The mass input width as shown above is 119px but as we can see in the below screen, the length input width is 127px. (you can convert both in rems)

image

front/src/applications/stdcmV2/hooks/useStdcmConsist.ts
```diff
@@ -0,0 +1,41 @@
+import { useSelector } from 'react-redux';
+
+import { useOsrdConfActions, useOsrdConfSelectors } from 'common/osrdContext';
+import { type StdcmConfSliceActions } from 'reducers/osrdconf/stdcmConf';
+import type { StdcmConfSelectors } from 'reducers/osrdconf/stdcmConf/selectors';
+import { useAppDispatch } from 'store';
+
+export default function useStdcmConsist() {
```
SharglutDev: We try to use `const` for hooks and components. You can either export like that and disable eslint rule or not export here and `export default` at the bottom of the file.

```suggestion
export const useStdcmConsist() {
```

front/src/styles/scss/applications/stdcmV2/_card.scss
```diff
@@ -114,16 +114,13 @@
       padding-bottom: 0.5rem;
 
       .stdcm-v2-consist__properties {
-        display: flex;
+        display: grid;
+        grid-template-columns: 7.938rem 7.938rem;
```
SharglutDev:
```suggestion
        grid-template-columns: 7.4375rem 7.938rem;
```

SharglutDev (approved):
Lgtm and tested, good job ! We can keep the negative input feature for another issue, as it needs another refacto of the style of this part of the form.

Issue here : https://github.com/OpenRailAssociation/osrd/issues/9374

PR #9109, opened 2024-09-30
front: upgrade prettier

emersion:
This requires a small change in InfraErrorsList.


Caracol3 (commented):
I thought we shouldn't push the yarn.lock. 
Is there a reason to do so ?

emersion:
The `yarn.lock` is already in Git, this PR doesn't change anything about this :P 

It's important to include `yarn.lock` in Git for reproducibility: thanks to that file we're all using the same version of each dependency. Hence, if a dependency upgrade introduces a bug or changes behavior (as seen in this PR: prettier output changes), we all see the same thing and our production deployment doesn't burn without us noticing.

Does this make sense?

Caracol3:
Totally ! Thanks, for the details 🙏

Caracol3 (approved):
lgtm (tested)

SharglutDev (approved):
Lgtm and tested, thank you. Maybe we should warn on matrix that we all need to "yarn install" to have last rules of prettier.

emersion:
Yeah. Restarting the front container is necessary after pulling any kind of dependency change.

PR #9111, opened 2024-09-30
add moon logo when the train passes after midnight

### LINKED ISSUE #9104: Add `moon` logo when train past midnight on list of trains

End users

image

See from the train list if a train is past midnight

theocrsb:
closes #9104

### DIFF
--- a/front/src/modules/trainschedule/components/Timetable/TimetableTrainCard.tsx
+++ b/front/src/modules/trainschedule/components/Timetable/TimetableTrainCard.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 
 import { Checkbox } from '@osrd-project/ui-core';
-import { Duplicate, Pencil, Trash, Clock, Flame } from '@osrd-project/ui-icons';
+import { Duplicate, Pencil, Trash, Clock, Flame, Moon } from '@osrd-project/ui-icons';
 import cx from 'classnames';
 import dayjs from 'dayjs';
 import { omit } from 'lodash';
@@ -131,6 +131,12 @@ const TimetableTrainCard = ({
     dispatch(updateTrainIdUsedForProjection(train.id));
   };
 
+  /* TODO: delete the format when the date management PR has been passed */
+  const isAfterMidnight = dayjs(train.arrivalTime, 'D/MM/YYYY').isAfter(
+    dayjs(train.startTime, 'D/MM/YYYY'),
+    'day'
+  );
+
   return (
     <div
       data-testid="scenario-timetable-train"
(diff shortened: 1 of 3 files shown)


front/src/modules/trainschedule/components/Timetable/TimetableTrainCard.tsx
```diff
@@ -131,6 +131,11 @@ const TimetableTrainCard = ({
     dispatch(updateTrainIdUsedForProjection(train.id));
   };
 
+  const isAfterMidnight = dayjs(train.arrivalTime, 'D/MM/YYYY').isAfter(
```
emersion, line 135: Could we maybe copy over the `/* TODO: delete the format when the date management `PR` has been passed */` just so that we don't forget about this line?

front/src/styles/scss/applications/operationalStudies/_scenario.scss
```diff
@@ -353,8 +353,16 @@
...
+                width: 1.5rem;
+                height: 1.5rem;
+                color: var(--info60);
+                display: flex;
+                justify-content: center;
+                align-items: center;
+                margin: 0 0.25rem;
+                span {
+                  padding-bottom: 0.25rem;
+                }
```
emersion: Could we unify this with `not-honored-or-too-fast`? Seems like it could share most of this?

theocrsb: You're right, I've pushed a solution

front/src/styles/scss/applications/operationalStudies/_scenario.scss
```diff
@@ -352,27 +352,29 @@
...
-                font-weight: 400;
-                font-size: 1rem;
-              }
-
-              .not-honored-or-too-fast {
+              .logo {
                 width: 1.5rem;
                 height: 1.5rem;
-                color: var(--warning60);
+                color: var(--info60);
```
emersion, line 358: Maybe we can just drop this line since it's overridden in both cases?

front/src/styles/scss/applications/operationalStudies/_scenario.scss
```diff
@@ -352,27 +352,29 @@
...
-                margin: 0 0.5rem;
-              }
-
-              .scenario-timetable-train-departure {
-                font-weight: 400;
-                font-size: 1rem;
-              }
-
-              .not-honored-or-too-fast {
+              .logo {
```
emersion: Naming nit: maybe "indicator" would be slightly better? Or "status-icon"?

I'm completely fine with keeping the current name, feel free to ignore.

emersion (approved):
LGTM otherwise, feel free to resolve conversations!

kmer2016 (approved):

front/src/styles/scss/applications/operationalStudies/_scenario.scss
theocrsb: Done

front/src/styles/scss/applications/operationalStudies/_scenario.scss
theocrsb: Done

PR #9129, opened 2024-10-01
front: separate StdcmPathStep and PathStep types

### LINKED ISSUE #9462: front: separate PathStep and StdcmPathStep types
Separate PathStep and StdcmPathStep types to simplify the code. The new StdcmPathStep type should be as following:
```
export type StdcmPathStep = {
  id: string;
  arrival?: Date;
  stopFor?: string;
  isVia: boolean;
  location?: PathItemLocation & {
    name: string;
    coordinates: [number, number];
  };
  arrivalToleranceBefore?: number;
  arrivalToleranceAfter?: number;
} & ({ isVia: true; stopType?: StdcmStopTypes } | { isVia: false; arrivalType: ArrivalTimeTypes });
```

PathStep should also be cleaned.

clarani:
closes #9462 

:warning: you will need to clean your local storage to test this PR since I modified the store (or modify all the fields). You may need to clean your local storage also when you will leave my branch


front/src/reducers/osrdconf/types.ts
```diff
@@ -91,6 +91,25 @@ export type PathStep = PathItemLocation & {
...
 
-export type StdcmPathStep = PathStep & {
-  tolerances?: { before: number; after: number };
-} & ({ isVia: true; stopType: StdcmStopTypes } | { isVia: false; arrivalType: ArrivalTimeTypes });
+export type UicPathItemLocation = {
+  uic: number;
+  secondary_code?: string | null;
+  name: string;
+  coordinates: [number, number];
+};
```
SharglutDev: Maybe we could find another name because a PathItemLocation in the editoast way doesn't have a name or coordinates, it might be misleading.

Also, the type could be more dynamic using this syntax : 

```suggestion
type SomeCoolName = Extract<PathItemLocation, { uic: number }> & {
  name: string;
  coordinates: [number, number];
};
```

We don't have a use for other cases for now but we could also do 
```suggestion
type SomeCoolName<T> = Extract<PathItemLocation, T> & {
  name: string;
  coordinates: [number, number];
};
```

clarani: You're right, it's done :)

Synar (approved):
LGTM

front/src/reducers/index.ts
```diff
@@ -57,12 +58,26 @@ const saveUserFilter = createFilter('user', userWhiteList);
 
 const saveMainFilter = createFilter('main', mainWhiteList);
 
+const dateTransform = createTransform(
+  null,
+  (outboundState: StdcmPathStep[]) =>
+    outboundState.map((step) => {
+      if (!step.isVia && step.arrival && isISOString(step.arrival as unknown as string)) {
```
emersion: Why do we need this check? Why do we need this cast?

Are we receiving somehow some values which aren't actually of the type `StdcmPathStep`?

front/src/reducers/osrdconf/types.ts
```diff
@@ -91,6 +91,23 @@ export type PathStep = PathItemLocation & {
...
+  location?: PathItemLocation & {
+    name: string;
+    coordinates: [number, number];
+  };
+} & (
+  | { isVia: true; stopType: StdcmStopTypes; stopFor?: number }
+  | {
+      isVia: false;
+      arrivalType: ArrivalTimeTypes;
+      arrival?: Date;
```
emersion, line 107: Nice!

front/src/applications/stdcm/components/StdcmResults/SimulationReportSheet.tsx
```diff
@@ -305,7 +311,13 @@ const SimulationReportSheet = ({
                 const prevStep = operationalPointsList[index - 1];
                 const isViaInSimulationPath = stdcmData.simulationPathSteps
                   .slice(1, -1)
-                  .some((pathStep) => pathStep.name === step.name && pathStep.ch === step.ch);
+                  .some(
+                    ({ location }) =>
+                      location &&
+                      location.name === step.name &&
+                      'uic' in location &&
+                      location.secondary_code === step.ch
```
emersion: Would be nicer to use `getSecondaryCode()` here instead.

front/src/applications/stdcm/utils/index.ts
```diff
@@ -0,0 +1,24 @@
+import type { MarkerInformation } from 'modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/ItineraryMarkers';
+import type { StdcmPathStep } from 'reducers/osrdconf/types';
+import { extractHHMM, extractDDMMYYYY } from 'utils/date';
+
+export const getTimesInfoFromDate = (date?: Date) =>
+  date
+    ? {
+        arrivalDate: extractDDMMYYYY(date), // ISO date part
+        arrivalTime: extractHHMM(date),
+        arrivalTimehours: date.getHours(),
```
emersion: Typo: "hours" with capital "H"

front/src/modules/pathfinding/helpers/getStepLocation.ts
```diff
@@ -1,8 +1,7 @@
 import type { PathItemLocation } from 'common/api/osrdEditoastApi';
-import type { PathStep, StdcmPathStep } from 'reducers/osrdconf/types';
 import { mToMm } from 'utils/physics';
 
-const getStepLocation = (step: PathStep | StdcmPathStep): PathItemLocation => {
+const getStepLocation = <T extends PathItemLocation>(step: T): PathItemLocation => {
```
emersion: I don't think we need a generic here. I think this works as well?

```ts
const getStepLocation = (step: PathItemLocation): PathItemLocation => {
```

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -324,6 +324,7 @@ export const usePathfinding = (
                     name: correspondingOp.name,
                     uic: correspondingOp.uic,
                     ch: correspondingOp.ch,
+                    secondary_code: correspondingOp.ch,
```
emersion: Hm. I'm worried these two fields go out-of-sync.

https://github.com/OpenRailAssociation/osrd/pull/8633 completely removes the `ch` field.

front/src/modules/pathfinding/utils.ts
```diff
@@ -52,7 +53,7 @@ export const matchPathStepAndOp = (
   return step.track === op.track && step.offset === op.offsetOnTrack;
 };
 
-export const getPathfindingQuery = ({
+export const getPathfindingQuery = <T extends PathItemLocation>({
```
emersion: Ditto: I don't think we need the generics here.

front/src/applications/stdcm/components/StdcmResults/SimulationReportSheet.tsx
clarani: Done

front/src/modules/pathfinding/helpers/getStepLocation.ts
clarani: Yes it works :) I was anticipating, but you're right, I should keep it simple

front/src/modules/pathfinding/utils.ts
clarani: Done

front/src/modules/pathfinding/hooks/usePathfinding.ts
clarani: I can remove this line and wait for your PR. This hook is not used in stdcm iirc

front/src/reducers/index.ts
clarani: We need this cast because typescript thinks that arrival is already a Date, but actually it's only a string at this point :/
So I needed to hack typescript. I can add a comment if you want ?

front/src/utils/date.ts
```diff
@@ -27,15 +26,16 @@ function extractHHMMSS(dateTimeString?: string) {
...
 }
 
-/**
- * @param dateTimeString date string in ISO format
- * @returns string "HH:MM"
- */
-export function extractHHMM(dateTimeString?: string) {
-  if (!dateTimeString) {
+/** given a Date, returns its time "HH:MM" */
+export function extractHHMM(date?: Date) {
```
SharglutDev: Maybe we could have a single function for that with a 2nd param where we pass the format like 'HH:MM' or 'DD/MM/YY' ?

front/src/utils/timeManipulation.ts
```diff
@@ -16,6 +16,10 @@ export function ms2sec(ms: number) {
   return ms / 1000;
 }
 
+export function minToMs(min: number) {
+  return min * 60 * 1000;
+}
+
```
SharglutDev, line 22: Instead of having n function for each possible conversion, do you think we could have for example a conversion schema like `CONVERSION_FACTORS_SCHEMA` in the rollingStock module but for each unit (time, weight, length etc) so we could have one function for each ?

front/src/modules/pathfinding/utils.ts
```diff
@@ -61,9 +62,9 @@ export const getPathfindingQuery = ({
 }: {
   infraId?: number;
   rollingStock?: RollingStockWithLiveries;
-  origin: PathStep | null;
-  destination: PathStep | null;
-  pathSteps: (PathStep | null)[];
+  origin: PathItemLocation | null;
+  destination: PathItemLocation | null;
+  pathSteps: (PathItemLocation | null)[];
```
SharglutDev: We change this but we still pass `PathStep` objects when calling this function is `usePathfinding` ?

front/src/modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/ItineraryMarkers.tsx
```diff
@@ -13,17 +14,27 @@ import stdcmOrigin from 'assets/pictures/mapMarkers/start.svg';
...
+
 enum MARKER_TYPE {
   ORIGIN = 'origin',
   VIA = 'via',
   DESTINATION = 'destination',
 }
 
-type MarkerInformation = {
-  marker: PathStep;
+type MarkerProps = {
```
SharglutDev: I find it confusing to use the `Props` keyword when not referring to react component props. Would you consider renaming it ?

front/src/modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/ItineraryMarkers.tsx
```diff
@@ -57,7 +68,7 @@ const formatPointWithNoName = (
   </>
 );
 
-const extractMarkerInformation = (pathSteps: (PathStep | null)[], showStdcmAssets: boolean) =>
+const extractMarkerInformation = (pathSteps: MarkerInformation[], showStdcmAssets: boolean) =>
```
SharglutDev, line 68: Here we also change the type but operational studies keep passing `PathStep` to this function, it's confusing :/

front/src/utils/timeManipulation.ts
clarani: As discussed this morning, maybe we can stay like this ?

front/src/reducers/osrdconf/types.ts
```diff
@@ -91,6 +91,23 @@ export type PathStep = PathItemLocation & {
...
+};
+
+export type StdcmPathStep = {
+  id: string;
+  location?: PathItemLocation & {
+    name: string;
+    coordinates: [number, number];
+  };
+} & (
+  | { isVia: true; stopType: StdcmStopTypes; stopFor?: number }
```
emersion: Suggestion: add `// in minutes` after the `stopType` field.

front/src/utils/date.ts
SharglutDev: After discussion with the dev front team, it has been decided to keep it like this for now waiting for a larger refacto of the units to be done

front/src/utils/timeManipulation.ts
SharglutDev: After discussion with the dev front team, it has been decided to keep it like this for now waiting for a larger refacto of the units to be done

front/src/modules/pathfinding/utils.ts
SharglutDev: As `PathStep` extends `PathItemLocation`, it can remain like this

front/src/reducers/osrdconf/types.ts
clarani: Done

front/src/modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/ItineraryMarkers.tsx
clarani: `MarkerProperties` is a good name for you ?

front/src/utils/date.ts
clarani: I think it's better to keep 2 separated functions, it will be easier to find it by name (`extractHHMMSS` or `extractDDMMYY`), if it's fine for you ?

SharglutDev: I can live with that :)

front/src/modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/ItineraryMarkers.tsx
SharglutDev: Same as other comment. As typescript allows it, we will stick with this behavior.

front/src/reducers/index.ts
emersion: Why is it a string? If it's a string, it's supposed to cause a type error on the side where it's set, no?

front/src/modules/pathfinding/hooks/usePathfinding.ts
emersion: Hm, well, I think if we remove this `getStepLocation()` will break. In other words, if half of the codebase uses `ch` and the other half uses `secondary_code`, we get fireworks.

front/src/reducers/osrdconf/operationalStudiesConf/simulationConfReducers.spec.ts
SharglutDev, line 1: nit: maybe we should rename this file ? It's not in the simulation conf anymore

front/src/reducers/osrdconf/types.ts
```diff
@@ -92,6 +92,23 @@ export type PathStep = PathItemLocation & {
...
-  tolerances?: { before: number; after: number };
-} & ({ isVia: true; stopType: StdcmStopTypes } | { isVia: false; arrivalType: ArrivalTimeTypes });
+export type PathItemLocationWithMetadata<T> = Extract<PathItemLocation, T> & {
+  name: string;
+  coordinates: [number, number];
+};
+
+export type StdcmPathStep = {
+  id: string;
+  location?: PathItemLocation & {
```
SharglutDev: In stdcm, if I'm correct, a path step will always have an `uic`.

Also since recently, `PathItemLocation` has changed a bit to add the `track_reference` and we know distinguish `PathItemLocation` between a `TrackOffset` and an `OperationalPointReference`.
Should we start to use this more specific key word for our types ?

For example, `location` property could be renamed `reference` or something like that.

Anyway, by doing this, you could avoid all your `'uic' in step` checks across the code.

```suggestion
  location?: PathItemLocationWithMetadata<{ uic: number }> & {
```

front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
```diff
@@ -24,12 +24,15 @@ function formatChCode(chCode: string) {
...
-const StdcmOperationalPoint = ({ point, opPointId, disabled }: StdcmOperationalPointProps) => {
+const StdcmOperationalPoint = ({ location, pathStepId, disabled }: StdcmOperationalPointProps) => {
   const { t } = useTranslation('stdcm');
   const dispatch = useAppDispatch();
 
   const { searchTerm, chCodeFilter, sortedSearchResults, setSearchTerm, setChCodeFilter } =
-    useSearchOperationalPoint({ initialSearchTerm: point.name, initialChCodeFilter: point.ch });
+    useSearchOperationalPoint({
+      initialSearchTerm: location?.name,
+      initialChCodeFilter: location?.secondary_code || undefined,
```
SharglutDev: This should also work I think

```suggestion
      initialChCodeFilter: location?.secondary_code,
```

front/src/reducers/index.ts
emersion: Turns out we get strings because this function gets data from local storage (where it's serialized as JSON).

Docs: https://github.com/rt2zz/redux-persist?tab=readme-ov-file#transforms

My suggestion: avoid using `StdcmPathStep` since the serialized objects don't satisfy this type.

```ts
// Deserialize date strings coming from local storage
const dateTransform = createTransform(
  null,
  (outboundState: { arrival?: string }[]) =>
    outboundState.map(({ arrival, ...step }) => {
      if (arrival) {
        return { ...step, arrival: new Date(arrival) };
      }
      return step;
    }),
  { whitelist: ['stdcmPathSteps'] }
);
```

(Also I don't think we need `isISOString`: local storage should only contain valid strings.)

front/src/modules/pathfinding/hooks/usePathfinding.ts
emersion: https://github.com/OpenRailAssociation/osrd/pull/8633 has been merged now, so this concern is alleviated :)

(Sorry about the conflicts…)

front/src/reducers/osrdconf/types.ts
```diff
@@ -92,6 +92,23 @@ export type PathStep = PathItemLocation & {
   isInvalid?: boolean;
 };
 
-export type StdcmPathStep = PathStep & {
-  tolerances?: { before: number; after: number };
-} & ({ isVia: true; stopType: StdcmStopTypes } | { isVia: false; arrivalType: ArrivalTimeTypes });
+export type PathItemLocationWithMetadata<T> = Extract<PathItemLocation, T> & {
+  name: string;
+  coordinates: [number, number];
+};
```
SharglutDev: Aren't we sure that we will have a `secondary_code` when searching it in the `Combobox` ? We could update the type or add a new one specific for that so you don't have to check everywhere if it exists.

front/src/applications/stdcm/components/StdcmResults/SimulationReportSheet.tsx
```diff
@@ -201,31 +211,27 @@ const SimulationReportSheet = ({
                       </View>
                       <View style={styles.convoyAndRoute.stopTableOpWidth}>
                         <TD style={styles.convoyAndRoute.stopTableOpColumn}>
-                          {step.name || 'Unknown'}
+                          {step.location?.name || 'Unknown'}
```
SharglutDev: Won't we always have a location here ?

front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
```diff
@@ -67,9 +68,12 @@ const StcdmResults = ({
...
 
-  const simulationPathSteps = hasSimulationResults
-    ? outputs.results.simulationPathSteps
-    : undefined;
+  const simulationPathSteps = useMemo(() => {
+    if (!hasSimulationResults) {
+      return undefined;
+    }
+    return extractMarkersInfo(outputs.results.simulationPathSteps);
+  }, []);
```
SharglutDev: You probably should add `hasSimulationResults` and `outputs` in the deps

front/src/reducers/osrdconf/types.ts
emersion: Maybe we can use `Extract<PathItemLocation, { uic: number }>`?

front/src/modules/pathfinding/hooks/usePathfinding.ts
Synar: Is there a reason why we can't rename ch into secondary code for every single type? Are there cases where there's a meaningful difference?

emersion: No real reason apart from this being a sizable amount of work (some of the types are generated from OpenAPI). But yeah, we should aim to gradually rename everything.

clarani: No pb for the conflicts ! :+1:

front/src/reducers/osrdconf/types.ts
clarani: Yes you're right :+1: 
I was anticipating, but it's not a good thing

front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
clarani: No it does not work because secondary_code can be null or '' :/

front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
clarani: You're right, thanks !

front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
SharglutDev: Oups the comments aren't in the good order. This one was related to the one below :)

clarani: Ah yes, now you're right :)

front/src/applications/stdcm/components/StdcmResults/SimulationReportSheet.tsx
```diff
@@ -2,19 +2,31 @@ import { useEffect, useState } from 'react';
...
 import { msToKmh } from 'utils/physics';
 import { capitalizeFirstLetter } from 'utils/strings';
 
 import styles from './SimulationReportStyleSheet';
 import type { SimulationReportSheetProps } from '../../types';
 import { base64ToJpeg, getStopDurationTime } from '../../utils/formatSimulationReportSheet';
 
+const getSecondaryCode = ({ location }: StdcmPathStep) => {
+  return location!.secondary_code;
+};
```
SharglutDev:
```suggestion
const getSecondaryCode = ({ location }: StdcmPathStep) =>  location!.secondary_code;
```

front/src/applications/stdcm/utils/index.ts
```diff
@@ -0,0 +1,24 @@
+import type { MarkerInformation } from 'modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/ItineraryMarkers';
+import type { StdcmPathStep } from 'reducers/osrdconf/types';
+import { dateToHHMMSS, dateToDDMMYYYY } from 'utils/date';
+
+export const getTimesInfoFromDate = (date?: Date) =>
```
SharglutDev, line 5: When I hit refresh on the stdcm page, I get a white screen with this error : 
`date.getHours is not a function`. (from the call in `useStdcmForm`)
Only a store reset allows me to load the page again.

Maybe it has something to do with the fact that when refreshing/rerendering, there is no time selected in the dateTime input.

front/src/applications/stdcm/components/StdcmResults/SimulationReportSheet.tsx
```diff
@@ -65,13 +77,10 @@ const SimulationReportSheet = ({
     [mapImageUrl]
   );
 
-  const getArrivalTime = (
-    step: PathStep,
-    { isFirstStep, isLastStep }: { isFirstStep?: boolean; isLastStep?: boolean }
-  ) => {
-    if (isFirstStep || isLastStep) {
+  const getArrivalTime = (step: StdcmPathStep) => {
```
SharglutDev: We could also extract this function outside the component like your 2 new ones.

front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
```diff
@@ -67,9 +68,12 @@ const StcdmResults = ({
     );
   }, [outputs]);
 
-  const simulationPathSteps = hasSimulationResults
-    ? outputs.results.simulationPathSteps
-    : undefined;
+  const simulationPathSteps = useMemo(() => {
```
SharglutDev: Easier to understand (and harmonized with `StdcmConfig`)
```suggestion
  const markersInfo = useMemo(() => {
```

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
```diff
@@ -1,26 +1,21 @@
...
 ): StdcmConfigErrors | undefined => {
   const isOneOpPointMissing = !origin || !destination;
   if (isOneOpPointMissing) {
     return undefined;
   }
 
-  if (pathSteps.some((step) => step === null || ('uic' in step && step.uic === -1))) {
-    return { errorType: StdcmConfigErrorTypes.MISSING_LOCATION };
-  }
-
```
SharglutDev: We don't have any warning box anymore if I add a via and launch the simulation without filling its inputs

front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
```diff
@@ -73,20 +74,21 @@ const StdcmOperationalPoint = ({ point, opPointId, disabled }: StdcmOperationalP
...
           [] as { label: string; id: string }[]
         ),
-    [point, sortedSearchResults]
+    [location, sortedSearchResults]
   );
 
   const dispatchNewPoint = (p?: SearchResultItemOperationalPoint) => {
-    if (p && 'uic' in point && p.ch === point.secondary_code && p.uic === point.uic) return;
-    const newPoint = p
+    if (p && location && p.ch === location.secondary_code && p.uic === location.uic)
```
SharglutDev: For all the checks like that we do, maybe we could have an util like `matchPathStepAndOp` but for stdcm ?

front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
```diff
@@ -73,20 +74,21 @@ const StdcmOperationalPoint = ({ point, opPointId, disabled }: StdcmOperationalP
           },
           [] as { label: string; id: string }[]
         ),
-    [point, sortedSearchResults]
+    [location, sortedSearchResults]
   );
 
   const dispatchNewPoint = (p?: SearchResultItemOperationalPoint) => {
```
SharglutDev, line 80: I know it doesn't come from you but would you mind ? :)
```suggestion
  const dispatchNewPoint = (searchedOp?: SearchResultItemOperationalPoint) => {
```

front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
```diff
@@ -124,20 +126,20 @@ const StdcmOperationalPoint = ({ point, opPointId, disabled }: StdcmOperationalP
   };
 
   useEffect(() => {
-    if (point) {
-      setSearchTerm(point.name || '');
-      setChCodeFilter(pointCh || '');
+    if (location) {
+      setSearchTerm(location.name || '');
+      setChCodeFilter(location.secondary_code || '');
```
SharglutDev:
```suggestion
      setSearchTerm(location.name);
      setChCodeFilter(location.secondary_code);
```

front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
SharglutDev, line 1: You can do a `npm lint` in this file

front/src/applications/stdcm/components/StdcmResults/SimulationReportSheet.tsx
clarani: Done :+1:

front/src/applications/stdcm/components/StdcmResults/SimulationReportSheet.tsx
clarani: No we can't, because there is a translation :/

I don't think it's a good idea to add a new param `t` and to extract the function

front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
clarani: I will do another PR to rename `simulationPathSteps` by `markersInfo` in `ManageTrainSchedule/Map` too :+1:

front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
clarani: I plan to do a PR to refacto a bit this component, can this wait until this PR ? :)

front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
clarani: Done

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
clarani: Nice catch ! Done :white_check_mark:

front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
clarani: Do you think this can wait for the refacto of this component ?

front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
clarani: Done

front/src/applications/stdcm/components/StdcmResults/SimulationReportSheet.tsx
SharglutDev: Isn't it what you did with `getStopType` ?

SharglutDev (approved):
Lgtm and tested, great refacto !

front/src/utils/date.ts
```diff
@@ -253,3 +242,11 @@ export const dateToHHMMSS = (date: Date, { withoutSeconds } = { withoutSeconds:
   const format = withoutSeconds ? 'HH:mm' : 'HH:mm:ss';
   return dayjs(date).local().format(format);
 };
+
+export const dateToDDMMYYYY = (date: Date) => dayjs(date).format('DD/MM/YYYY');
+
+/** Check if a string is a valid ISO string and can be converted into a date */
+export const isISOString = (val: string) => {
```
emersion: This function is unused.

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -82,15 +82,11 @@ const StdcmConfig = ({
 
   const disabled = isPending || retainedSimulationIndex > -1;
 
+  const markerInfos = useMemo(() => extractMarkersInfo(pathSteps), [pathSteps]);
```
emersion: Nit: `markersInfo` (pluralization)

emersion (approved):
Just two nits, looks good to me!

front/src/applications/stdcm/components/StdcmForm/StdcmVias.tsx
```diff
@@ -31,8 +30,12 @@ const StdcmVias = ({ disabled = false }: StdcmConfigCardProps) => {
...
 
   const updateStopType = (newStopType: StdcmStopTypes, pathStep: StdcmPathStep) => {
-    const defaultStopTime =
-      newStopType === StdcmStopTypes.DRIVER_SWITCH ? formatDurationAsISO8601(3 * 60) : '';
+    let defaultStopTime: number | undefined;
+    if (newStopType === StdcmStopTypes.DRIVER_SWITCH) {
+      defaultStopTime = 3;
+    } else if (newStopType === StdcmStopTypes.SERVICE_STOP) {
+      defaultStopTime = 0;
+    }
```
SharglutDev, line 38: Looks like the store is updated but not the form in the UI

clarani: Done

front/tests/assets/stdcm/stdcmAllStops.json
Maymanaf, line 1: why we need to have different values?

PR #9130, opened 2024-10-01
Fix no cache docker

flomonster:
Faking the ping command to pass the health check.

### How to test it?

Edit the `docker-compose.yml` file.
Add to the `editoast`, `environment` variable list the following item:

```yaml
NO_CACHE: "true"
```


Erashin (approved):
Tested. Works as intended: editoast now launches corectly, and no cache is kept. Thanks!

leovalais (approved):

PR #9139, opened 2024-10-01
editoast: provide search object list to the store explicitely

leovalais:
Search object configurations used to be collected in a static list for each `Search` macro invocation.
This list is then used in `derive(SearchConfigStore)` to generate the store logic.

On top of being a Rust antipattern, it now triggers some warnings on
nightly 1.83.0 about mixing immutable and mutable references on static
data (which is indeed a bad idea). This static list also confuses
rust-analyzer. Also getting rid of it removes our only use of unsafe.


EthanPERRUZZA (commented):
Hello,

I checked the requests on my computer and it still works :)

It indeed looks a lot better than using `unsafe` and your code looks clean to me. However, I don't have a lot of experience working with Rust so I may not notice if something is off.

Thank you for the ping, I'll base my work on yours ;)

flomonster (approved):
Thanks for this PR.
No more unsafe :tada:

PR #9144, opened 2024-10-01
osrdyne: exit if the provided configuration file doesn't exist

leovalais:
`Yaml::file_exact` fails if the path doesn't point to a file.
`Yaml::file` does not (by design).

Yet another "what is wrong with my setup again?" and its just `osrdyne.yml` instead of `osrdyne.yaml`...


Khoyo (approved):

Khoyo:
But I am appalled that `Yaml::file_exact()` panics instead of reporting an error...

leovalais:
It does not `panic` but since the error is propagated up to the `main()` invocation the result is just like a panic. I reworded the commit as I can see how misleading the description was...

PR #9151, opened 2024-10-02
editoast: make the health check timeout configurable

Khoyo:
Also add a bit of debug logging, useful when debugging timeout issues.


leovalais (approved):
LGTM (not tested)

But either change the logging level to `trace!` or disable the entries by default to avoid spamming plz: https://github.com/OpenRailAssociation/osrd/blob/a327b3bdb376097ef2689144845e296fb2ee9247/editoast/src/main.rs#L95-L104

Khoyo:
> But either change the logging level to trace! or disable the entries by default to avoid spamming plz: 

Done. (switched to trace)

PR #9153, opened 2024-10-02
core: use infra explorer in pathfinding endpoint

### LINKED ISSUE #6784: core: refacto: use `InfraExplorer` in the block pathfinding
# BH

- [x] Fix https://github.com/osrd-project/osrd-confidential/issues/333
- [x] Operational studies already got this bug multiple times in their first study
- [x] Cleaner implementation of the pathfinding
- [x] Easier and less ambiguous post-processing

# AC

- [ ] The bug does not happen again

# Technical conception

The current pathfinding looks for a path in blocks while ignoring the routes, we then try to find a route that matches the block list.
There are two problems with this approach: finding that list is *a pain*, and it may not even exist. It causes a bug that may appear on our base infrastructures, and sometimes happens on edited infras (see https://github.com/osrd-project/osrd-confidential/issues/333)

The solution to this problem would be to explore paths while considering both routes and blocks. 

To do this, we may re-use the tooling that is *currently* being implemented on the STDCM side on [this branch](https://github.com/osrd-project/osrd/pull/6171) (`InfraExplorer`). The class handles the combination of routes and blocks, and will soon handle the constraints as well (see https://github.com/osrd-project/osrd/issues/6335).

## Definition of ready

- **Technical**

- **General**
  - [x] WSJF values are filled
  - [x] Validated by the PM

Erashin:
Fixes #6784.

Use infra explorer in pathfinding endpoint. This makes us explore the graph using both blocks and routes, ensuring that if we do find a block path, the corresponding routes exist.

As intended, corrects https://github.com/osrd-project/osrd-confidential/issues/333, a path is found between Metz and Nancy:
image

Benchmark run on same infra: results are slightly better (up by 0.3% for a payload insaaannee), and basically we're getting less repeated_tracks errors (instead the path is just not found):
image

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/graph/PathfindingGraph.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/graph/PathfindingGraph.kt
@@ -1,0 +1,45 @@
+package fr.sncf.osrd.graph
+
+import fr.sncf.osrd.sim_infra.api.Block
+import fr.sncf.osrd.stdcm.graph.extendLookaheadUntil
+import fr.sncf.osrd.stdcm.infra_exploration.InfraExplorer
+import java.util.Objects
+
+data class PathfindingEdge(val infraExplorer: InfraExplorer) {
+    val block = infraExplorer.getCurrentBlock()
+    val length = infraExplorer.getCurrentBlockLength()
+
+    override fun equals(other: Any?): Boolean {
+        return if (other !is PathfindingEdge) false
+        else
+            this.infraExplorer.getLastEdgeIdentifier() ==
+                other.infraExplorer.getLastEdgeIdentifier()
+    }
+
+    override fun hashCode(): Int {
+        return Objects.hash(infraExplorer.getLastEdgeIdentifier())
+    }
+}
+
+class PathfindingGraph : Graph<PathfindingEdge, PathfindingEdge, Block> {
+    override fun getEdgeEnd(edge: PathfindingEdge): PathfindingEdge {
+        return edge
+    }
+
+    override fun getAdjacentEdges(node: PathfindingEdge): Collection<PathfindingEdge> {
+        val res = ArrayList<PathfindingEdge>()
+        val extended = mutableListOf<InfraExplorer>()
+        if (node.infraExplorer.getLookahead().size > 0) {
+            extended.add(node.infraExplorer.clone())
+        } else {
+            extended.addAll(extendLookaheadUntil(node.infraExplorer, 1))
+        }
+        for (newPath in extended) {
+            if (newPath.getLookahead().size == 0) continue
+            newPath.moveForward()
+            val newEdge = PathfindingEdge(newPath)
+            res.add(newEdge)
+        }
+        return res
+    }
+}
--- a/core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/BlockInfraImpl.kt
+++ b/core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/BlockInfraImpl.kt
@@ -218,7 +218,7 @@ private fun buildBlockName(
     val detectors = mutableListOf<DirDetectorId>()
     detectors.add(rawInfra.getZonePathEntry(descriptor.path[0]))
     detectors.add(rawInfra.getZonePathExit(descriptor.path.last()))
-    val detectorStr = detectors.map { "${rawInfra.getDetectorName(it.value)}" }
+    val detectorStr = detectors.map { rawInfra.getDetectorName(it.value) }
     val rawStringId = "$signals;$detectorStr;$trackNodeConfigNames"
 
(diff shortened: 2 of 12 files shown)


core/src/main/kotlin/fr/sncf/osrd/graph/PathfindingGraph.kt
```diff
@@ -0,0 +1,60 @@
...
+        return edge.getEdgeEnd()
+    }
+
+    override fun getAdjacentEdges(node: PathfindingNode): Collection<PathfindingEdge> {
+        val res = ArrayList<PathfindingEdge>()
+        val extended = mutableListOf<InfraExplorer>()
+        if (node.infraExplorer.getLookahead().size > 0) {
+            extended.add(node.infraExplorer)
+        } else {
+            extended.addAll(extendLookaheadUntil(node.infraExplorer, 3))
```
eckter:
```suggestion
            extended.addAll(extendLookaheadUntil(node.infraExplorer, 1))
```

core/src/main/kotlin/fr/sncf/osrd/graph/PathfindingGraph.kt
```diff
@@ -0,0 +1,60 @@
...
+import fr.sncf.osrd.utils.units.Offset
+import fr.sncf.osrd.utils.units.meters
+
+data class PathfindingNode(
+    val infraExplorer: InfraExplorer,
+    val previousEdge: PathfindingEdge,
+    val locationOnEdge: Offset<Block>?,
+)
+
+data class PathfindingEdge(
```
eckter: Needs a good `.equals` and `.hash`, we don't actually mark anything as visited there

(using `getLastEdgeIdentifier`)

core/src/main/kotlin/fr/sncf/osrd/graph/PathfindingGraph.kt
```diff
@@ -0,0 +1,60 @@
...
+    val infraExplorer: InfraExplorer,
+    val previousNode: PathfindingNode?,
```
eckter: It shouldn't be *necessary* as the pathfinding class already builds the list. But if it makes things easier, why not

Erashin: I don't think it's all that useful either, so i'll probably scrap it.

core/src/main/kotlin/fr/sncf/osrd/graph/PathfindingGraph.kt
Erashin: See https://github.com/OpenRailAssociation/osrd/pull/9153/commits/f62e41b838b6b83281149893d500d2f2796da6d6

core/src/main/kotlin/fr/sncf/osrd/graph/PathfindingGraph.kt
Erashin: See https://github.com/OpenRailAssociation/osrd/pull/9153/commits/f62e41b838b6b83281149893d500d2f2796da6d6

core/src/main/kotlin/fr/sncf/osrd/graph/PathfindingGraph.kt
Erashin: See https://github.com/OpenRailAssociation/osrd/pull/9153/commits/f62e41b838b6b83281149893d500d2f2796da6d6

core/src/main/kotlin/fr/sncf/osrd/graph/PathfindingGraph.kt
```diff
@@ -0,0 +1,47 @@
+package fr.sncf.osrd.graph
+
+import fr.sncf.osrd.sim_infra.api.Block
+import fr.sncf.osrd.stdcm.graph.extendLookaheadUntil
+import fr.sncf.osrd.stdcm.infra_exploration.InfraExplorer
+import java.util.Objects
+
+data class PathfindingEdge(
```
Erashin: Theorically, could put the `equals()` and `hashcode()` methods into the `InfraExplorer` and implement them for both `InfraExplorerImpl` and `InfraExplorerWithEnvelopeImpl`, completely trash this object and just create a `PathfindingGraph : Graph<InfraExplorer, InfraExplorer, Block>`.
Personnally, i think it's clearer likeso: states clearly that it's an edge, and we might not want such a loose `equals()` and `hashcode()` methods for `InfraExplorerImpl`. It's not a strong opinion though, feel free to give me your pov.

eckter: I wouldn't be a big fan of the semantics of having that `.equals` in `InfraExplorer`, two `InfraExplorers` aren't equal just because they lead to the same place

Erashin: We agree on this then :)

eckter (approved):

PR #9157, opened 2024-10-02
Remove application roles

leovalais:
Follow up #8984

> [!TIP]
> Review commits individually.

- Adds the `Superuser` builtin role
- Adds the `--superuser` cmd flag
- Removes the `RoleConfig`
- Harmonizes the serialization of builtin role: we use the serde representation for everything. It's easier to manage, less error-prone, and the OpenAPI description of the endpoints becomes much more ergonomic.


Khoyo:
> Adds the --superuser cmd flag

Shouldn't it rather be named `--single-user` ? Since it doesn't give you a superuser role, it treats all users as superusers (which is what single user mode usually is)

Or maybe something more explicit, like `--disable-access-control` or `--disable-authorization`

editoast/src/client/mod.rs
```diff
@@ -134,38 +132,33 @@ pub struct MapLayersConfig {
...
-    /// the superuser role is granted to all users.
-    #[clap(long, env = "EDITOAST_ROLES_CONFIG")]
-    pub roles_config: Option<PathBuf>,
-    #[derivative(Default(value = r#""http://127.0.0.1:4242/".into()"#))]
+    /// If this option is set, any role and permission check will be bypassed. Even if no user is
+    /// provided by the request headers of if the provided user doesn't have the required privileges.
+    // TODO: once the whole role system will be deployed, the default value of this option should
+    // be set to false. It's currently set to true in order to pass integration tests, which otherwise
+    // only recieve 401 responses.
+    #[clap(long, env = "EDITOAST_DISABLE_AUTHORIZATION", default_value_t = true)]
```
flomonster, line 157: Let's keep true for now. We might want to turn it false later.

flomonster (approved):
Great PR! :pray:

SharglutDev (approved):

PR #9159, opened 2024-10-02
 ci: label PRs automatically based on changed files 

Khoyo:


emersion:
Missing osrdyne?

Apart from this, LGTM! Thanks a lot for splitting the re-formatting in a separate commit.

ghost (approved):
lgtm

emersion:
Oh, one more thing: seems like the workflow should be configured to run on `pull_request_target` only?

```yaml
on:
- pull_request_target
```

We might need a separate workflow file.

.github/workflows/build.yml
```diff
@@ -977,3 +977,12 @@ jobs:
...
         uses: fsfe/reuse-action@v4
+
+  label_pr:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Label PR
+        uses: actions/labeler@v5
```
woshilapin: How about putting this in another workflow as advertised [in the documentation](https://github.com/actions/labeler?tab=readme-ov-file#create-workflow)? I believe it'd be better because we typically don't need this action to run on `push` events on `dev` for example.

woshilapin (approved):
Thanks, you were faster than me 😅

woshilapin (approved):
I'd add the `sync-labels` option too. Thanks for the PR.

.github/workflows/pr.yml
```diff
@@ -0,0 +1,14 @@
+name: pr
+on:
+  pull_request:
```
emersion, line 3: Shouldn't this be `pull_request_target`? The docs seem to imply that `pull_request` should only be used temporarily for testing upgrades…

Khoyo: We use `pull_request` in the build.yml file, since `pull_request_target` is runs in the base branch context. `pull_request_target` also has *interesting* security implications.

I don't think this matters for here anyways, but I can change it if you see how this might break

emersion: I'm not sure what the exact implications are TBH, just noticed the deviation from the docs. I'm completely OK with just leaving this as-is and wait-and-see if there are any issues with it.

PR #9162, opened 2024-10-02
editoast: provide details when pathfinding failed

### LINKED ISSUE #8038: editoast: error pathfinding_not_found has not the details provided by core
### What happened?

The error is not has precise has the one provided by core

We should be able to have this :
image

1. not sure

Recette (SNCF)

Firefox

staging de8ada8

younesschrifi:
Close #8038

### DIFF
--- a/editoast/src/core/pathfinding.rs
+++ b/editoast/src/core/pathfinding.rs
@@ -14,10 +14,11 @@ editoast_common::schemas! {
     IncompatibleConstraints,
     IncompatibleOffsetRangeWithValue,
     IncompatibleOffsetRange,
-    PathfindingResult,
     PathfindingResultSuccess,
     OffsetRange,
     TrackRange,
+    PathfindingInputError,
+    PathfindingNotFound,
 }
 
 #[derive(Debug, Serialize)]
@@ -73,9 +74,9 @@ pub struct InvalidPathItem {
     pub path_item: PathItemLocation,
 }
 
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, ToSchema)]
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 #[serde(tag = "status", rename_all = "snake_case")]
-pub enum PathfindingResult {
+pub enum PathfindingCoreResult {
     Success(PathfindingResultSuccess),
     NotFoundInBlocks {
         track_section_ranges: Vec<TrackRange>,
@@ -91,14 +92,13 @@ pub enum PathfindingResult {
         incompatible_constraints: Box<IncompatibleConstraints>,
     },
     InvalidPathItems {
-        #[schema(inline)]
         items: Vec<InvalidPathItem>,
     },
     NotEnoughPathItems,
     RollingStockNotFound {
         rolling_stock_name: String,
     },
-    PathfindingFailed {
+    InternalError {
         core_error: InternalError,
     },
 }
@@ -119,6 +119,39 @@ pub struct PathfindingResultSuccess {
     /// The path offset in mm of each path item given as input of the pathfinding
     /// The first value is always `0` (beginning of the path) and the last one is always equal to the `length` of the path in mm
     pub path_item_positions: Vec<u64>,
+}
+
+// Enum for input-related errors
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, ToSchema)]
+#[serde(tag = "error_type", rename_all = "snake_case")]
+pub enum PathfindingInputError {
+    InvalidPathItems {
+        #[schema(inline)]
+        items: Vec<InvalidPathItem>,
+    },
+    NotEnoughPathItems,
+    RollingStockNotFound {
+        rolling_stock_name: String,
+    },
+}
+
+// Enum for not-found results and incompatible constraints
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, ToSchema)]
+#[serde(tag = "error_type", rename_all = "snake_case")]
+pub enum PathfindingNotFound {
+    NotFoundInBlocks {
+        track_section_ranges: Vec<TrackRange>,
+        length: u64,
+    },
+    NotFoundInRoutes {
+        track_section_ranges: Vec<TrackRange>,
+        length: u64,
+    },
+    NotFoundInTracks,
+    IncompatibleConstraints {
+        relaxed_constraints_path: Box<PathfindingResultSuccess>,
+        incompatible_constraints: Box<IncompatibleConstraints>,
+    },
 }
 
 /// An oriented range on a track section.
@@ -216,7 +249,7 @@ impl TrackRangeOffset<'_> {
     }
 }
 
-impl AsCoreRequest<Json<PathfindingResult>> for PathfindingRequest {
+impl AsCoreRequest<Json<PathfindingCoreResult>> for PathfindingRequest {
     const METHOD: reqwest::Method = reqwest::Method::POST;
     const URL_PATH: &'static str = "/v2/pathfinding/blocks";
 
--- a/editoast/src/views/path/pathfinding.rs
+++ b/editoast/src/views/path/pathfinding.rs
@@ -13,14 +13,19 @@ use editoast_schemas::rolling_stock::LoadingGaugeType;
 use editoast_schemas::train_schedule::PathItemLocation;
 use ordered_float::OrderedFloat;
 use serde::Deserialize;
+use serde::Serialize;
 use tracing::debug;
 use tracing::info;
 use utoipa::ToSchema;
 
+use crate::core::pathfinding::PathfindingCoreResult;
+use crate::core::pathfinding::PathfindingInputError;
+use crate::core::pathfinding::PathfindingNotFound;
 use crate::core::pathfinding::PathfindingRequest;
-use crate::core::pathfinding::PathfindingResult;
+use crate::core::pathfinding::PathfindingResultSuccess;
 use crate::core::AsCoreRequest;
 use crate::core::CoreClient;
+use crate::error::InternalError;
 use crate::error::Result;
 use crate::models::train_schedule::TrainSchedule;
 use crate::models::Infra;
@@ -41,6 +46,8 @@ crate::routes! {
 
 editoast_common::schemas! {
     PathfindingInput,
+    PathfindingFailure,
+    PathfindingResult,
 }
 
 /// Path input is described by some rolling stock information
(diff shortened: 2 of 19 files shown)


woshilapin (approved):

editoast/src/core/pathfinding.rs
```diff
@@ -91,18 +95,32 @@ pub enum PathfindingResult {
...
     InvalidPathItems {
-        #[schema(inline)]
         items: Vec<InvalidPathItem>,
     },
     NotEnoughPathItems,
     RollingStockNotFound {
         rolling_stock_name: String,
     },
-    PathfindingFailed {
+    PathfindingCoreError {
```
woshilapin: `PathfindingCoreResult::PathfindingCoreError` is a bit redundant. Not really necessary to repeat the name of the type in the variant. If we're renaming it, let's make it better? If I understand well, it's a proxy for error we don't understand? Maybe be `UnknownError` or `InternalError` is fine?

woshilapin (approved):
In general, I realize that every (or almost every) `struct` in the module `pathfinding` are prefixed with `Pathfinding` which is somewhat redundant. We might want to fix that but in another PR (hence I put a comment that don't need to be resolved). cc @leovalais 

Apart from that, thank you for this PR, I'm always amazed by what this application will bring to debuggability of pathfinding (and simulation) errors.

editoast/src/core/pathfinding.rs
leovalais: Agreed. Also, following a discussion we had on the `editoast` channel on matrix we agreed on not including `Core` in the names of definitions in the `core` module as it's redundant. Wdyt of `PathfindingResult::InternalError` instead?

editoast/src/core/pathfinding.rs
leovalais, line 1: Is `derive(Serialize)` really necessary for all these definitions?

editoast/src/core/pathfinding.rs
```diff
@@ -91,18 +95,32 @@ pub enum PathfindingResult {
...
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, ToSchema)]
+#[serde(tag = "failed_status", rename_all = "snake_case")]
+pub enum PathfindingFailed {
+    InputError(PathfindingInputError),
+    NotFound(PathfindingNotFound),
+    CoreError(PathfindingCoreError),
+}
+
```
leovalais: I feel like these definitions are more related to the endpoints API rather than core. I'd suggest moving these definitions and the conversion in the `views` module.

I'm also a bit weirded out by the past tense here (but maybe it's just me 😅), wdyt about `PathfindingFailure` instead?

editoast/src/views/path/pathfinding.rs
```diff
@@ -278,15 +287,22 @@ pub async fn pathfinding_from_train_batch(
     train_schedules: &[TrainSchedule],
     rolling_stocks: &HashMap<String, RollingStockModel>,
 ) -> Result<Vec<PathfindingResult>> {
-    let mut results = vec![PathfindingResult::NotEnoughPathItems; train_schedules.len()];
+    let mut results = vec![
+        PathfindingResult::Failed(PathfindingFailed::InputError(
+            PathfindingInputError::NotEnoughPathItems
+        ));
+        train_schedules.len()
+    ];
```
leovalais: I believe keeping a `Vec<PathfindingInputError>` and performing the conversion into a `PathfindingResult` just before returning would be easier to read. Wdyt?

leovalais (changes requested):
Thanks for this PR! LGTM on the `views` side, but I think there's too much logic in the `core` module. I agree with @woshilapin about the redundancy of the prefix, and I'm fine with changing that in another PR. I'll test this later.

front/public/locales/en/operationalStudies/scenario.json
```diff
@@ -53,8 +53,8 @@
     "filterLabel": "Name, label",
     "importTrainSchedule": "Import",
     "invalid": {
-      "pathfinding_not_found": "Pathfinding not found",
-      "pathfinding_failed": "Pathfinding failed",
+      "not_found": "Pathfinding not found",
+      "failed": "Pathfinding failed",
```
Akctarus: Why did you rename them ? Since we have `simulation_failed` and `rolling_stock_not_found` below, I think it's more readable to keep the `pathfinding` prefix

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -228,7 +228,9 @@ export const usePathfinding = (
 
           if (
             pathfindingResult.status === 'success' ||
-            pathfindingResult.status === 'incompatible_constraints'
+            (pathfindingResult.status === 'failed' &&
+              pathfindingResult.failed_status === 'not_found' &&
+              pathfindingResult.error_type === 'incompatible_constraints')
```
Akctarus: Since you use this condition 2 times, you can create a `const` for it, and use this instead

Akctarus (commented):
Tested, left few comments

editoast/src/views/path/pathfinding.rs
leovalais: As discussed, it doesn't bring much and adds a new conversion step. Resolving.

Akctarus (approved):
LGTM for the front part, not tested yet

Akctarus (approved):
LGTM

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -2449,46 +2433,74 @@ export type PathItemLocation =
...
+      rolling_stock_name: string;
+    };
+export type OffsetRange = {
+  end: number;
+  start: number;
+};
+export type IncompatibleOffsetRangeWithValue = {
+  range: OffsetRange;
+  value: string;
+};
```
SharglutDev, line 2458: Is there any reason why this type can't extend `IncompatibleOffsetRange` with the `value` property ?

front/public/locales/fr/operationalStudies/scenario.json
```diff
@@ -52,10 +52,16 @@
...
+      "not_found_in_tracks": "Voie manquante",
+      "incompatible_constraints": "Contraintes incompatibles",
       "pathfinding_not_found": "Chemin introuvable",
-      "pathfinding_failed": "Calcul non abouti",
+      "pathfinding_failure": "Calcul non abouti",
       "simulation_failed": "Simulation impossible",
-      "rolling_stock_not_found": "MR non trouvé"
+      "rolling_stock_not_found": "MR non trouvé",
+      "not_enough_path_items": "Point(s) de passage(s) manquants",
+      "invalid_path_items": "Point(s) de passage(s) invalides"
```
SharglutDev: Could you order the keys alphabetically ? :)

SharglutDev:
```suggestion
      "not_enough_path_items": "Point(s) de passage(s) manquant(s)",
      "invalid_path_items": "Point(s) de passage(s) invalide(s)"
```

front/public/locales/en/operationalStudies/scenario.json
```diff
@@ -53,10 +53,16 @@
...
+      "not_found_in_tracks": "Missing track",
+      "incompatible_constraints": "Incompatible constraints",
       "pathfinding_not_found": "Pathfinding not found",
-      "pathfinding_failed": "Pathfinding failed",
+      "pathfinding_failure": "Pathfinding failed",
       "simulation_failed": "Simulation failed",
-      "rolling_stock_not_found": "RS not found"
+      "rolling_stock_not_found": "RS not found",
+      "not_enough_path_items": "Not enough path items",
+      "invalid_path_items": "Invalid path items"
```
SharglutDev: Same here for key order

front/public/locales/en/operationalStudies/manageTrainSchedule.json
```diff
@@ -104,7 +104,7 @@
     "not_found_in_routes": "Missing route",
     "not_found_in_tracks": "Missing track",
     "not_enough_path_items": "Missing parameters for pathfinding: Departure, Arrival, Rolling stock",
-    "pathfinding_failed": "No path found",
+    "pathfinding_core_error": "No path found",
```
SharglutDev: Do we use this key somewhere ?

front/src/modules/trainschedule/components/Timetable/consts.ts
```diff
@@ -7,8 +7,13 @@ export const specialCodeDictionary: { [key: string]: string } = {
 export const invalidTrainValues: {
```
SharglutDev: Seems like we don't use that `const` anywhere in the app ? If we don't need it, you can remove it as well with the `InvalidReason` type which is used only here.

SharglutDev (changes requested):
Tested and it works great ! Left some comment

Erashin (approved):
Core part was very complex, good job navigating through it, thanks for that.. :D

front/src/common/api/generatedEditoastApi.ts
younesschrifi: We can not do that in editoast.

SharglutDev: When you do that, it's the same thing as I'm thinking
image

This could be 
```suggestion
export type IncompatibleOffsetRangeWithValue = IncompatibleOffsetRange & {
  value: string;
};
```

But if it's complicated, it's not very important :slightly_smiling_face:

front/public/locales/en/operationalStudies/scenario.json
```diff
@@ -53,8 +53,14 @@
...
+      "invalid_path_items": "Invalid path items",
+      "not_enough_path_items": "Not enough path items",
+      "not_found_in_blocks": "Missing block",
+      "not_found_in_routes": "Missing route",
+      "not_found_in_tracks": "Missing track",
+      "pathfinding_failure": "Pathfinding failed",
       "pathfinding_not_found": "Pathfinding not found",
-      "pathfinding_failed": "Pathfinding failed",
       "simulation_failed": "Simulation failed",
       "rolling_stock_not_found": "RS not found"
```
SharglutDev: rollingstock before simulation :)

SharglutDev (approved):
Lgtm and tested, good job. You can resolve the last 2 comments when needed :)

editoast/src/core/pathfinding.rs
```diff
@@ -9,15 +9,18 @@ use utoipa::ToSchema;
 
 use crate::core::{AsCoreRequest, Json};
 use crate::error::InternalError;
+use crate::views::path::pathfinding::PathfindingFailure;
+use crate::views::path::pathfinding::PathfindingResult;
```
leovalais: We want `views` to depend on `core` for the split in crates, can we move the code that uses these definitions in `views` instead?

editoast/src/core/pathfinding.rs
```diff
@@ -73,9 +76,9 @@ pub struct InvalidPathItem {
     pub path_item: PathItemLocation,
 }
 
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, ToSchema)]
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 #[serde(tag = "status", rename_all = "snake_case")]
-pub enum PathfindingResult {
+pub enum PathfindingCoreResult {
```
leovalais, line 79: We already in the `core` module ;)

```suggestion
pub enum PathfindingResult {
```

editoast/src/views/path/pathfinding.rs
```diff
@@ -66,6 +72,21 @@ struct PathfindingInput {
     rolling_stock_length: OrderedFloat<f64>,
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, ToSchema)]
+#[serde(tag = "status", rename_all = "snake_case")]
+pub enum PathfindingResult {
+    Success(PathfindingResultSuccess),
+    Failed(PathfindingFailure),
```
leovalais:
```suggestion
    Failure(PathfindingFailure),
```

for consistency

leovalais (approved):
LGTM, thanks for the changes! A few last modifications 🙏:

front/src/common/api/generatedEditoastApi.ts
Khoyo: This is a generated file from Rust code. There are no such things as inheritance or type extensions in Rust. We can use tricks to have utoipa generate the openapi with it, but IMO it's not very useful since typescript doesn't care (and it adds complexity in the rust code)

editoast/src/core/pathfinding.rs
younesschrifi: PathfindingResult is already defined :(

PR #9166, opened 2024-10-03
docker, core, editoast: add mode single-worker for all infra

### LINKED ISSUE #8599: Core debug: allow one core to handle multiple infra
### Description and goal

The goal is to be able to easily debug and use core locally 

There is a bit of doc for that.

Possibilities (sorted by preference, discussed with @Khoyo and @eckter ):
* trigger osrdyne to send all in one `all` queue (and subscribe core to that queue)
* allow for the "default" use of the web API on editoast
* some other options around rabbitmq (duplicate queues, something with a unique queue and multiple topics to be able to filter)

### Acceptance criteria

It is possible, documented and easy to use a unique local core and have everything working (unit tests, integration tests, regular use, ...)

bougue-pe:
Also:
- editoast: remove http core client
- cleanup unavailable envvar `OSRD_BACKEND_URL`

🔍 please review by commit + one more developper should test it (integration + e2e + regular use)
This work may probably be improved (doc, script, etc.) any suggestion is welcome.

Fix: https://github.com/OpenRailAssociation/osrd/issues/8599


flomonster:
I did not reviewed it yet but I have a small question.

This features seems useful for debuging purpose (see #8599). I believe we're talking of core? If so I don't see why we should handle this feature using docker compose.

core/src/main/java/fr/sncf/osrd/cli/WorkerCommand.kt
```diff
@@ -44,10 +44,12 @@ class WorkerCommand : CliCommand {
...
     val WORKER_REQUESTS_QUEUE: String
     val WORKER_ACTIVITY_EXCHANGE: String
+    val ALL_INFRA: Boolean
 
     init {
-        WORKER_ID = System.getenv("WORKER_ID")
-        WORKER_KEY = System.getenv("WORKER_KEY")
+        ALL_INFRA = !System.getenv("ALL_INFRA").isNullOrEmpty()
+        WORKER_ID = System.getenv("WORKER_ID") ?: (if (ALL_INFRA) "all_infra_worker" else null)
+        WORKER_KEY = System.getenv("WORKER_KEY") ?: (if (ALL_INFRA) "all" else null)
```
eckter: This sets the priority as "env variable > all infra > null". I'd prefer if we used "all infra > env variable > null". It's quite minor but that way we could change a single flag to enable the "all infra" mode, instead of having to also unset the worker key/id.

```suggestion
        WORKER_ID =  if (ALL_INFRA) "all_infra_worker" else System.getenv("WORKER_ID")
        WORKER_KEY =  if (ALL_INFRA) "all" else System.getenv("WORKER_KEY")
```

eckter (commented):
LGTM for core with one minor comment.

Not tested yet, I'll probably come back to it later.

bougue-pe:
> I did not reviewed it yet but I have a small question.
> 
> This features seems useful for debuging purpose (see #8599). I believe we're talking of core? If so I don't see why we should handle this feature using docker compose.

Compose file here is more of a helper/documentation to plug all together, as `editoast` also needs some specific param.
And I tried to put the params/envvar at the best place (IMO 😅) to enable mutualization when reusing `core` in another compose.

core/src/main/java/fr/sncf/osrd/cli/WorkerCommand.kt
bougue-pe: I went for the safety on regular use, but it's fine to change.
Any reaction to that is welcome to make a "definitive" choice: current choice is to go for @eckter's suggestion.

docker/docker-compose.host.yml
```diff
@@ -39,6 +38,13 @@ services:
     ports: []
     network_mode: host
 
+  core:
+    ports: []
+    network_mode: host
+    environment:
+      CORE_EDITOAST_URL: "http://127.0.0.1:8090"
```
woshilapin, line 45: I wondered why we still have an HTTP link between `editoast` and `core`? I'm guessing that loading the infra is still working that way? Shouldn't `osrdyne` have change that behavior too? Was it even addressed as part of the design? Maybe it's fine... 🤷

editoast/src/core/mod.rs
```diff
@@ -84,11 +84,17 @@ impl CoreClient {
         Self::Direct(client)
     }
 
-    pub async fn new_mq(uri: String, worker_pool_identifier: String, timeout: u64) -> Result<Self> {
+    pub async fn new_mq(
+        uri: String,
+        worker_pool_identifier: String,
+        timeout: u64,
+        single_worker: bool,
+    ) -> Result<Self> {
```
woshilapin: Would it make sense that this function takes directly a `mq_client::Options` as an input argument? I'm just triggered by the increasing number of arguments.

younesschrifi (commented):
LGTM ! Didn't tested it yet. I'm waiting for your rebase to do so and then I'll approve the PR👍🏽

woshilapin (approved):
Looks good although, as discussed, it will make us not really use the full complexity of `osrdyne` on local, might not help to detect possible bugs in it. That said, all for making the developer experience smoother.

⚠️ not tested since I still didn't find the time to handle `osrdyne` that depends on a Docker daemon that I don't have (using podman, daemon less).

docker/docker-compose.host.yml
eckter: `core` still need to fetch infra data from `editoast`. I don't think it should go through the queue or anything related to `osrdyne`, there's only one `editoast` instance. HTTP requests are straightforward.

core/src/main/java/fr/sncf/osrd/cli/WorkerCommand.kt
bougue-pe: Ah, I remember, I also wanted to let a possible choice of the `worker_key` name, which is not consistent with the "all infra" naming here.

The namings using "all" are about the targeted usage more than what it actually does/permit: no pre-load + work attached to just a grouping key, no matter the infra(s) being used.
Not sure it is useful for now or in the future to change/rename so I went for your suggestion in https://github.com/OpenRailAssociation/osrd/pull/9166/commits/302e4efbd7668b18a046b4ff58ea5f6bcb20e2ce after upvotes.

editoast/src/core/mod.rs
bougue-pe: Not convinced by the benefits: on the only caller place, we would just embed the `mq_client::Options` creation, revealing more of the internals (and adding little more code/import).
Not strictly opposed to it, but not so fond of the idea here.

leovalais: I'll be helpful to settle the debate: I'm fine with both :p

leovalais (approved):
Lgtm, not tested

editoast/src/core/mod.rs
woshilapin: Is revealing the internal really a problem? It will make calling the function more explicit. Compare these, for example:
```rust
new_mq("foobar", "yolo", 42, false);
```

As a reader of the code, I will therefore need to go consult the function to understand what each of these parameters mean (which in Github might not be trivial, so I may need to pull the code, to benefit from `rust-analyzer`).

Compared to this:
```rust
new_mq(mq_client::Options {
  uri: "foobar",
  worker_pool_identifier: "yolo",
  timeout: 42,
  single_worker: false
});
```

Sure it's more verbose, but more verbose and more explicit it's not necessarily bad if you think that code is read more often than it's written. In some languages, you can name the parameter when calling but sadly, Rust doesn't allow that ([there has been discussion about it](https://internals.rust-lang.org/t/pre-rfc-named-arguments/16413/1)).

⚠️ This comment is long, because I wanted to be explicit about why I tend to be averse to a lot of arguments in a function. For functions with only few arguments, usually, the name of the function is enough to understand what the 1 or 2 arguments are, without naming them. That said, 4 is not yet **a lot of arguments**, just enough to trigger me. So I'm fine to keep it as it is, feel free to resolve whatever you do.

ghost:
I do not have anything to say about the PR per se, but I'm curious on why we would want to make this change the default behavior ?

What's the rationale behind this ?

bougue-pe:
@ElysaSrc there is a bit more info in the related issue https://github.com/OpenRailAssociation/osrd/issues/8599 but you may already have had a look and I'm not sure about your question.

The main purpose is to ease debug of core (mainly without having to spawn a new core and track the id of the infra which can be painful, like for integration tests).
The solution provided is OK for the business logic and hijacks part of the ops/architecture part, so it is limited in that regard (but a bit less than switching back to the web api).

Then the rationale of changing the noopdyne compose file is to consider that using dyne as noop is (almost?) only for that use case, so we might as well add the "changes" to editoast and core conf on the way to ease/document that case.
So the use case (which I didn't test completly, as I realized few minutes ago) would be:
1. compose `up` the whole stack
2. `down` the component(s) that I want to debug/replace locally
3. spawn locally those components
4. `<do actual test/use>`

Hope that it answers your question, and maybe it's worth adding a bit of doc and a bit of description (will try to after more test, but any suggestion is welcome).

ghost (approved):

ghost:
> @ElysaSrc there is a bit more info in the related issue https://github.com/OpenRailAssociation/osrd/issues/8599 but you may already have had a look and I'm not sure about your question.
> 
> The main purpose is to ease debug of core (mainly without having to spawn a new core and track the id of the infra which can be painful, like for integration tests).
> The solution provided is OK for the business logic and hijacks part of the ops/architecture part, so it is limited in that regard (but a bit less than switching back to the web api).
> 
> Then the rationale of changing the noopdyne compose file is to consider that using dyne as noop is (almost?) only for that use case, so we might as well add the "changes" to editoast and core conf on the way to ease/document that case.
> So the use case (which I didn't test completly, as I realized few minutes ago) would be:
> 1. compose `up` the whole stack
> 2. `down` the component(s) that I want to debug/replace locally
> 3. spawn locally those components
> 4. `<do actual test/use>`
> 
> Hope that it answers your question, and maybe it's worth adding a bit of doc and a bit of description (will try to after more test, but any suggestion is welcome).

Thanks for the detailed response !

Erashin (approved):
Tested for core debug purposes. Works as intended. Good job :)

core/src/main/java/fr/sncf/osrd/cli/WorkerCommand.kt
```diff
@@ -44,10 +47,12 @@ class WorkerCommand : CliCommand {
     val WORKER_POOL: String
     val WORKER_REQUESTS_QUEUE: String
     val WORKER_ACTIVITY_EXCHANGE: String
+    val ALL_INFRA: Boolean
 
     init {
-        WORKER_ID = System.getenv("WORKER_ID")
-        WORKER_KEY = System.getenv("WORKER_KEY")
+        ALL_INFRA = !System.getenv("ALL_INFRA").isNullOrEmpty()
+        WORKER_ID = if (ALL_INFRA) "all_infra_worker" else System.getenv("WORKER_ID")
```
Khoyo: pretty sure this will conflict, could you rebase?

docker-compose.yml
```diff
@@ -55,14 +55,26 @@ services:
...
         static_assets: assets
+    environment:
+      # Should match the reference, see. ./docker/osrdyne.yml
+      CORE_EDITOAST_URL: "http://osrd-editoast"
+      JAVA_TOOL_OPTIONS: "-javaagent:/app/opentelemetry-javaagent.jar"
+      CORE_MONITOR_TYPE: "opentelemetry"
+      OTEL_EXPORTER_OTLP_TRACES_PROTOCOL: "grpc"
+      OTEL_EXPORTER_OTLP_TRACES_ENDPOINT: "http://jaeger:4317"
+      OTEL_METRICS_EXPORTER: "none"
+      OTEL_LOGS_EXPORTER: "none"
```
Khoyo: I think directly providing the reference in ./docker/osrdyne.yml is a better idea... Or at least make clear that those are just for documentation, and that `./docker/osrdyne.yml` is not the reference, but the actually used values.

Khoyo (commented):
Great work!

Should we also rename `noopdyne` as `single-worker`? It wasn't a great name, and I mostly named it that way because it *wasn't* a single-worker mode ^^'

core/src/main/java/fr/sncf/osrd/cli/WorkerCommand.kt
bougue-pe: It did, ended up refactoring a bit through a `getBooleanEnvvar` fun https://github.com/OpenRailAssociation/osrd/pull/9166/commits/52ffad3e68fc252d8e036c9ec76ba718383749ad#diff-7f5b2e605ea8cd19fd586f8d299a533f45c5419298dc3b16aca328645dee196aR75

editoast/src/core/mod.rs
bougue-pe: Convinced by the "named params": https://github.com/OpenRailAssociation/osrd/pull/9166/commits/52ffad3e68fc252d8e036c9ec76ba718383749ad#diff-09ec5a58d69869bd41dba8612f5c3973f7bbe0147510aa5f2a1a44632ef30903R417-R422

docker-compose.yml
bougue-pe: Hopefully improved in https://github.com/OpenRailAssociation/osrd/pull/9166/commits/52ffad3e68fc252d8e036c9ec76ba718383749ad#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R70-R71

bougue-pe:
Renamed in https://github.com/OpenRailAssociation/osrd/pull/9166/commits/2329db4747b9e586320a10ede86401552d65a019 & https://github.com/OpenRailAssociation/osrd/pull/9166/commits/d60f674fc44a54dc4183bb9d409721045ff35a6c:
* `osrd-compose.sh` to `host-compose.sh` 
* `docker-compose.noopdyne.yml` to `docker-compose.single-worker.yml` 

Also added a dedicated `single-worker-compose.sh` script to wrap this up with a bit more doc.

flomonster (approved):
LGTM (README change)

Khoyo (approved):

Erashin (approved):
Gj again :) The scale option is a huge plus :)

emersion (approved):
`front/` changes LGTM (not tested)

In general huge +1 on the idea, although I haven't looked at all at the impl :P

clarani (approved):

PR #9169, opened 2024-10-03
front: switch from yarn to npm

### LINKED ISSUE #9143: front: switch from yarn to npm
### Description and goal

we use actually yarn, in order to integrate osrd-ui repo in osrd we need to switch to npm

### Acceptance criteria

- [ ] remove `yarn.lock` and `node-modules` (convert yarn.lock??)
- [ ] update `scripts` in `package.json`
- [ ] update yarn commands in `dockerfile`
- [ ] install libraries via `npm install` in order to retrieve `node-modules` & `package-lock.json`

Yohh:
close #9143


emersion:
I had success running https://github.com/imsnif/synp

I think we'd be less likely to run into issues if we can use this tool instead of crossing fingers.

emersion:
- Running `npm install` before renaming `resolutions` to `overrides` in `package.json` makes npm use `@rtk-query/oazapfts-patched` instead of our fork.
- Running `npm install` after renaming `resolutions` to `overrides` in `package.json` makes npm die with ".git can't be found" in `husky install`.
- Trying to upgrade `oazapfts` to the latest version won't work because it uses a monorepo and Git repo dependencies can't target a subdir.

shenriotpro (approved):
Thanks for the README update

front/docker/Dockerfile.devel
```diff
@@ -19,4 +19,4 @@ WORKDIR /app
 # Start the app
 COPY docker/dev-entrypoint.sh /
 ENTRYPOINT ["/dev-entrypoint.sh"]
-CMD ["sh", "-c", "id && yarn install && exec yarn start-container"]
+CMD ["sh", "-c", "id && rm -rf node_modules && npm install && exec npm run start-container"]
```
emersion: I think `rm -rf node_modules` is no longer necessary now.

front/package.json
```diff
@@ -125,7 +126,7 @@
     "@vitejs/plugin-react-swc": "^3.6.0",
     "@vitest/coverage-v8": "^2.0.0",
     "dotenv": "^16.4.5",
-    "eslint": "^8.56.0",
+    "eslint": "^8.57.1",
```
emersion: Hm, why did the version constraints in `package.json` change? I'd expect the `package.json` to stay as-is except for the `overrides` rename.

front/docker/Dockerfile.devel
Yohh: yep, missed that, it's done

front/scripts/generate-types.sh
```diff
@@ -3,9 +3,9 @@ if ! npx @rtk-query/codegen-openapi src/config/openapi-editoast-config.ts; then
     echo "npx @rtk-query/codegen-openapi src/config/openapi-editoast-config.ts command failed. Exit the script"
     exit 1
 fi
-yarn eslint --fix src/common/api/generatedEditoastApi.ts --no-ignore
+npm exec eslint -- --fix src/common/api/generatedEditoastApi.ts --no-ignore
```
emersion: Nit: could use `npx` here instead to avoid having to use `--`. (But `npm exec` is fine as well.)

front/tsconfig.json
```diff
@@ -53,7 +53,7 @@
     "suppressExcessPropertyErrors": false,
     "suppressImplicitAnyIndexErrors": false,
     "typeRoots": ["node_modules/@types"],
-    "types": ["vite/client"],
+    "types": ["vite/client", "node"],
```
emersion: This exposes `node` specific types in the global scope... Is this necessary when switching to NPM?

front/package.json
Yohh: done

front/scripts/generate-types.sh
Yohh: nice, done

front/tsconfig.json
Yohh: it was about missing types while trying to migrate successfully, just forgot to remove it, it's done

Khoyo:
```
[khoyo@odin osrd-hotfix]$ git grep yarn
front/README.md:- Install playwright dependencies `cd ./front/ && yarn playwright install --with-deps`
front/README.md:this script as you would to `yarn e2e-tests` or `yarn playwright test`.
front/docker/Dockerfile.playwright:COPY front/package.json front/yarn.lock /app/front/
front/docker/Dockerfile.playwright:RUN yarn install --frozen-lockfile
front/package-lock.json:        "yarn": "^1.7.0"
front/package-lock.json:        "yarn": ">=1"
front/package-lock.json:        "yarn": ">=1"
front/package-lock.json:        "yarn": "*"
front/package-lock.json:    "node_modules/yarn": {
front/package-lock.json:      "resolved": "https://registry.npmjs.org/yarn/-/yarn-1.22.22.tgz",
front/package-lock.json:        "yarn": "bin/yarn.js",
front/package-lock.json:        "yarnpkg": "bin/yarn.js"
scripts/run-front-playwright-container.sh:VERSION=$(yarn list --pattern playwright --json | jq -r '.data.trees[].name | split("@")[-1]' | sort -u)
scripts/run-front-playwright-container.sh:    osrd-playwright:latest yarn playwright test "${args[@]}"
```

The dependency in the lockfile is probably valid, the rest probably isn't.

Yohh:
> ```
> [khoyo@odin osrd-hotfix]$ git grep yarn
> front/README.md:- Install playwright dependencies `cd ./front/ && yarn playwright install --with-deps`
> front/README.md:this script as you would to `yarn e2e-tests` or `yarn playwright test`.
> front/docker/Dockerfile.playwright:COPY front/package.json front/yarn.lock /app/front/
> front/docker/Dockerfile.playwright:RUN yarn install --frozen-lockfile
> front/package-lock.json:        "yarn": "^1.7.0"
> front/package-lock.json:        "yarn": ">=1"
> front/package-lock.json:        "yarn": ">=1"
> front/package-lock.json:        "yarn": "*"
> front/package-lock.json:    "node_modules/yarn": {
> front/package-lock.json:      "resolved": "https://registry.npmjs.org/yarn/-/yarn-1.22.22.tgz",
> front/package-lock.json:        "yarn": "bin/yarn.js",
> front/package-lock.json:        "yarnpkg": "bin/yarn.js"
> scripts/run-front-playwright-container.sh:VERSION=$(yarn list --pattern playwright --json | jq -r '.data.trees[].name | split("@")[-1]' | sort -u)
> scripts/run-front-playwright-container.sh:    osrd-playwright:latest yarn playwright test "${args[@]}"
> ```
> 
> The dependency in the lockfile is probably valid, the rest probably isn't.

you're right, I removed the useless mentions in `package-lock.json` and updated the playwright script, thank you

emersion:
The `package-lock.json` references should be retained. In general this file shouldn't be edited by hand. It has "yarn" references because some of our dependencies depend on yarn.

front/package.json
emersion: Hm, unless I'm missing something, I still see these `package.json` changes?

Yohh: are you sure your branch is up to date? I don"t see it anymore...

edit: ok I missed some packages

Yohh: done

emersion (commented):
The changes look good to me!

My only remaining worry is that during the various iterations of this PR, the `package-lock.json` file has evolved and contains unwanted changes. For instance, at some point some dependencies were upgraded in `package.json`. To ensure we keep the NPM dependency tree as similar as possible to yarn's (and resolve the `yarn.lock` conflicts with the dev branch), I'd suggest we re-generate `package-lock.json` from `yarn.lock` by running `npm install` on the dev branch and then using the resulting `package-lock.json` file as-is in this branch. Wdyt?

emersion (approved):
I've applied my suggestion above, rebased, and also had to add `--package-lock-only` for `npm list` to make it work without a prior `npm install`.

This LGTM, but it would be best if another maintainer could review as well!

ghost (approved):
Not tested but LGTM

emersion:
Some platform-specific optional dependencies were missing from `package-lock.json` due to https://github.com/npm/cli/issues/4828, causing breakage on platforms other than Linux amd64. To fix it, dependencies with platform-specific sub-packages need to be re-installed to force npm to not look at `node_modules/`. I did the following:

```sh
npm install --save-dev rollup@4.23.0 @swc/core@1.7.26
git restore package.json
npm install
```

Yohh:
everything works fine for me on macos and linux

Yohh:
everything works fine for me on macos and linux

Math-R (approved):
Lgtm and tested !

clarani (approved):
LGTM :white_check_mark:

PR #9177, opened 2024-10-03
editoast: fix max running time default value

### LINKED ISSUE #9164: stdcm: maximum running time parameter is set too high
### What happened?

When running stdcm request with editoast figuring out the best "max running time" parameter, the value is too high. The departure time tolerance is added to the result.

_No response_

Run stdcm requests with large departure time tolerance and compare max running time

_No response_

_No response_

efce7b7a4bb0342c8f3ba1d05eb64326c893c013

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/9164

With some cleaning up

There was a misconception around the "max run time" parameter: it describes the maximum duration between the first train departure and its last arrival, stops included. But it was computed and used as if it were the duration between the earliest *possible* train departure and the latest possible arrival at the last stop.

(e.g. if we want the train to run for 2h max but leave at any time between 10:00AM and 10:00PM, we should set it to 2h, but we used to set it to 2+12h)

### DIFF
--- a/editoast/src/views/timetable/stdcm.rs
+++ b/editoast/src/views/timetable/stdcm.rs
@@ -205,6 +205,7 @@ async fn stdcm(
     .await?;
 
     // 2. Compute the earliest start time, maximum running time and maximum departure delay
+    // Simulation time without stop duration
     let simulation_run_time_result = get_simulation_run_time(
         db_pool.clone(),
         redis_client.clone(),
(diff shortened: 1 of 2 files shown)


younesschrifi (approved):
LGTM ! 👍🏽

woshilapin (approved):
Thank you very much for the PR's description, helped a lot 😅.

Would it make sense to add/modify some tests to avoid a regression on that topic?

Erashin (approved):
Thanks for the fix :)

eckter:
> Would it make sense to add/modify some tests to avoid a regression on that topic?

Yes, most likely :see_no_evil: 

It's a little tedious to properly test this, I've checked the parameters that are forwarded by core but it's not that easy to automate. But it should be doable with "planches travaux". 

I just need to figure out what's the proper way to write tests in editoast. I vaguely remember that the python integration tests are supposed to be replaced by the editoast test suite?

shenriotpro (approved):
Thanks

eckter:
@woshilapin I've added an integration test

woshilapin (approved):
Thank you very much for adding a test (with beautiful ASCII art 😍).

Issue #9124, opened 2024-10-01
Search on train_schedule table


EthanPERRUZZA:
### Issue: 

**Endpoint**: `/search`

**Objective**: Add a new object to the search functionality. Currently, we can search on items like rolling stock, etc. The new object to add is: **`train_schedule`**.

---

### Things to Know

- **Software for testing requests easily**: Postman works well for testing.
- You can import from OpenAPI.
- There's an environment system where you can define variables for testing in local/dev environments easily.
- `page` and `page-size`: These are pagination options commonly found on websites, but they are already set up, so no need to worry about them.
- To test, I can make a request and check the **network** tab.
- Requests are done in **Reverse Polish Notation**.

---

### Search AST Language

- **like**: for regex-style searches.
- **=**: for exact word matching.
- **search**: for smart search (case and accent insensitive).

---

### TODOs

1. **Understand what's in `train_schedule`**: Refer to the SQL table and docs here: [Train Schedule Documentation](https://osrd.fr/en/docs/reference/design-docs/timetable/).
   
2. **Check if the table is indexed**: Ensure the `train_schedule` table is indexed (in SQL) by name, and create the index if necessary.

3. **Define search parameters and results**: 
   - Declare which parameters we can search by.
   - Specify what the search will return.

---

### Search Declaration

- The table name is `train_schedule`, no need for a join.
- Define the columns available for search (e.g., **name**).
- The structure underneath should represent the search results (ideally almost everything).

---

### Other Considerations

- Make sure we can search by `timetable_id` as well.
- There's a **timetable** table and a **train_schedule** table, linked by `timetable_id` (primary key, which we don’t need to focus on, but it should be returned).
- There may be a test timetable, and we need to query the correct one.

---

### Out of Scope

- No new tables.
- No migration needed.
- No need to handle fuzzy search.

---

### Acceptance Criteria

- At a minimum, we should be able to search by **name**.

celine-sncf:
@flomonster should we keep this issue open to add the date filtering? (or open another front issue ?)

flomonster:
> should we keep this issue open to add the date filtering? (or open another front issue ?)

No, we decided that the date filtering should be done by the frontend for now.

PR #9178, opened 2024-10-03
editoast: add train_schedule object to search endpoint

EthanPERRUZZA:
#9124 

Bellow is a test using the below request :
Based on the openapi request : Returns all infra objects of some type according to a hierarchical query.

POST : 
{{baseUrl}}/search?page=1&page_size=25
Body :
```json 
{
    "object": "trainschedule",
    "query": [
        "and",
        [
            "search",
            [
                "train_name"
            ],
            "10222"
        ],
        [
            "=",
            [
                "timetable_id"
            ],
            1
        ]
    ]
}
```

Answer : 
```json
[
    {
        "comfort": 0,
        "constraint_distribution": 0,
        "id": 1,
        "initial_speed": 0,
        "labels": [
            "all-stops",
            "H3C4"
        ],
        "margins": {
            "boundaries": [],
            "values": [
                "0%"
            ]
        },
        "options": {
            "use_electrical_profiles": true
        },
        "path": [
            {
                "deleted": false,
                "id": "0",
                "operational_point": "d9e37c1e-6667-11e3-89ff-01f464e0362d"
            },
            {
                "deleted": false,
                "id": "1",
                "operational_point": "d9a37fbe-6667-11e3-89ff-01f464e0362d"
            },
            {
                "deleted": false,
                "id": "2",
                "operational_point": "d97702c4-6667-11e3-89ff-01f464e0362d"
            }
        ],
        "power_restrictions": [],
        "rolling_stock_name": "69400US",
        "schedule": [
            {
                "arrival": "PT6720S",
                "at": "1",
                "locked": true,
                "reception_signal": "OPEN",
                "stop_for": "PT1680S"
            },
            {
                "arrival": "PT10200S",
                "at": "2",
                "locked": true,
                "reception_signal": "OPEN",
                "stop_for": "P0D"
            }
        ],
        "speed_limit_tag": "ME100",
        "start_time": "2024-09-17T10:03:00+00:00",
        "timetable_id": 1,
        "train_name": "102224"
    }
]
```

### DIFF
--- a/editoast/openapi.yaml
+++ b/editoast/openapi.yaml
@@ -8770,6 +8770,7 @@ components:
       - $ref: '#/components/schemas/SearchResultItemProject'
       - $ref: '#/components/schemas/SearchResultItemStudy'
       - $ref: '#/components/schemas/SearchResultItemScenario'
+      - $ref: '#/components/schemas/SearchResultItemTrainSchedule'
       description: A search result item that depends on the query's `object`
     SearchResultItemOperationalPoint:
       type: object
@@ -9002,6 +9003,72 @@ components:
           type: integer
           format: int64
         line_name:
+          type: string
+    SearchResultItemTrainSchedule:
+      type: object
+      description: A search result item for a query with `object = "trainschedule"`
+      required:
+      - id
+      - train_name
+      - labels
+      - rolling_stock_name
+      - timetable_id
+      - start_time
+      - schedule
+      - margins
+      - initial_speed
+      - comfort
+      - path
+      - constraint_distribution
+      - power_restrictions
+      - options
+      properties:
+        comfort:
+          type: integer
+          format: int64
+        constraint_distribution:
+          type: integer
+          format: int64
+        id:
+          type: integer
+          format: int64
+          minimum: 0
+        initial_speed:
+          type: number
+          format: double
+        labels:
+          type: array
+          items:
+            type: string
+            nullable: true
+        margins:
+          $ref: '#/components/schemas/Margins'
+        options:
+          $ref: '#/components/schemas/TrainScheduleOptions'
+        path:
+          type: array
+          items:
+            $ref: '#/components/schemas/PathItem'
+        power_restrictions:
+          type: array
+          items:
+            $ref: '#/components/schemas/PowerRestrictionItem'
+        rolling_stock_name:
+          type: string
+        schedule:
+          type: array
+          items:
+            $ref: '#/components/schemas/ScheduleItem'
+        speed_limit_tag:
+          type: string
+          nullable: true
+        start_time:
+          type: string
+          format: date-time
+        timetable_id:
+          type: integer
+          format: int64
+        train_name:
           type: string
     Side:
       type: string
@@ -10206,7 +10273,7 @@ components:
             format: int64
             description: Timetable attached to the train schedule
             nullable: true
-    TrainScheduleOptionsV2:
+    TrainScheduleOptions:
       type: object
       properties:
         use_electrical_profiles:
--- a/editoast/src/views/search.rs
+++ b/editoast/src/views/search.rs
@@ -203,7 +203,9 @@ use axum::extract::Json;
 use axum::extract::Query;
 use axum::extract::State;
 use axum::Extension;
+use chrono::DateTime;
 use chrono::NaiveDateTime;
+use chrono::Utc;
 use diesel::pg::Pg;
 use diesel::sql_query;
 use diesel::sql_types::Jsonb;
@@ -215,6 +217,11 @@ use editoast_common::geometry::GeoJsonPoint;
 use editoast_derive::EditoastError;
 use editoast_derive::Search;
 use editoast_derive::SearchConfigStore;
+use editoast_schemas::train_schedule::Margins;
+use editoast_schemas::train_schedule::PathItem;
+use editoast_schemas::train_schedule::PowerRestrictionItem;
+use editoast_schemas::train_schedule::ScheduleItem;
+use editoast_schemas::train_schedule::TrainScheduleOptions;
 use editoast_search::query_into_sql;
 use editoast_search::SearchConfigStore as _;
 use editoast_search::SearchError;
@@ -345,6 +352,7 @@ async fn search(
 ) -> Result<Json<serde_json::Value>> {
     let roles: HashSet<BuiltinRole> = match object.as_str() {
         "track" | "operationalpoint" | "signal" => HashSet::from([BuiltinRole::InfraRead]),
+        "trainschedule" => HashSet::from([BuiltinRole::TimetableRead]),
         "project" | "study" | "scenario" => HashSet::from([BuiltinRole::OpsRead]),
         _ => {
             return Err(SearchApiError::ObjectType {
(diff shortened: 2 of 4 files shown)


editoast/src/views/search.rs
```diff
@@ -344,7 +344,9 @@ async fn search(
     Json(SearchPayload { object, query, dry }): Json<SearchPayload>,
 ) -> Result<Json<serde_json::Value>> {
     let roles: HashSet<BuiltinRole> = match object.as_str() {
-        "track" | "operationalpoint" | "signal" => HashSet::from([BuiltinRole::InfraRead]),
+        "track" | "operationalpoint" | "signal" | "trainschedule" => {
```
leovalais: Check for `BuiltinRole::TimetableRead` instead.

editoast/src/views/search.rs
```diff
@@ -683,6 +685,47 @@ pub(super) struct SearchResultItemScenario {
...
+    column(name = "timetable_id", data_type = "integer"),
+    column(name = "train_name", data_type = "string")
+)]
+#[allow(unused)]
+/// A search result item for a query with `object = "trainschedule"`
+pub(super) struct SearchResultItemTrainSchedule {
+    #[search(sql = "train_schedule.id")]
+    id: u64,
+    #[search(sql = "train_schedule.train_name")]
+    train_name: String, // useless puisqu'on en a besoin pour la request?
```
leovalais:
```suggestion
    train_name: String,
```

We comment in English 🫘 ;)

I'd suggest to leave it because 1) all other search responses also include their criteria 2) it makes the API more convenient and "stateless" to use as we don't have to "remember" the data each request was made with.

editoast/src/views/search.rs
```diff
@@ -683,6 +685,47 @@ pub(super) struct SearchResultItemScenario {
...
+    #[search(sql = "train_schedule.path")]
+    path: Vec<String>,
+    #[search(sql = "train_schedule.constraint_distribution")]
+    constraint_distribution: i16,
+    #[search(sql = "train_schedule.speed_limit_tag")]
+    speed_limit_tag: Vec<String>,
+    #[search(sql = "train_schedule.power_restrictions")]
+    power_restrictions: Vec<String>,
+    #[search(sql = "train_schedule.options")]
+    options: Vec<String>,
```
leovalais: Many of these `Vec<String>` aren't actually string arrays.

* `margins`
* `options`
* `path`
* `power_restrictions`
* `schedule`

You can find the actual type they deserialize to in `models::train_schedule::TrainSchedule`.

leovalais (changes requested):
Nice job! Tested and it works great, a few comments on documentation though.

Side note: since the `train_name` column isn't indexed to support fuzzy matching, the `search` operator is likely much more inefficient than a plain `=`.

woshilapin (changes requested):
Thanks for giving an example in the PR description. I'd say it's perfect for actually adding that as a test case of the feature 😈 😜

editoast/src/views/search.rs
EthanPERRUZZA: Thank you, I updated the code according to suggestions

EthanPERRUZZA:
> Thanks for giving an example in the PR description. I'd say it's perfect for actually adding that as a test case of the feature 😈 😜

:sob: 
I'll have a look but I'm not sure that it's even possible as you need to load data from a backup in the database to make the search. I'll try to look if there is tests for the other sarch objects.

I'm also unsure if it's necessary to test that feature as it's not meant to change or be broken by anything unless we change the database structure and/or the request result content.

leovalais:
@woshilapin We didn't test the search endpoint because there's a bunch of things to setup and we didn't have the test infrastructure have today. 

@EthanPERRUZZA for train schedules it should be simpler though, especially since there's no cache. You have in `models::fixtures` a test train schedule you can add to the DB to test your requests. You don't have to cleanup the DB after the test, this is done automatically.

flomonster (commented):
LGTM. I agree with @leovalais we can add a test for this search object.

[Here an example](https://github.com/OpenRailAssociation/osrd/blob/a08b1f5054378be0531d42be868fb91154eb0d5b/editoast/src/views/train_schedule.rs#L827) on how create a train schedule in a test.

> [!NOTE]
> For your test, you don't need an infra or a scenario. You just need a timetable and a train schedule.

editoast/openapi.yaml
```diff
@@ -8983,6 +8984,72 @@ components:
...
+      - initial_speed
+      - comfort
+      - path
+      - constraint_distribution
+      - power_restrictions
+      - options
+      properties:
+        comfort:
+          type: integer
+          format: int64
```
woshilapin, line 9028: As discussed, we might want to expose that as an enum, but the implications is not in the scope of this PR. Let's keep it as an integer for the purpose of this PR. @leovalais let's discuss that.

woshilapin (approved):
Looks good, thank you the test. Maybe you can add another simple one to test no results?

editoast/src/views/search.rs
```diff
@@ -692,5 +743,73 @@ pub(super) struct SearchResultItemScenario {
...
+                [
+                    "=",
+                    [
+
+                        "timetable_id"
+                    ],
+                    timetable_id
+                ]
+            ],
+        }));
```
leovalais:
```suggestion
        let request = app.post("/search").json(&json!({
            "object": "trainschedule",
            "query": ["and", ["=", ["train_name"], train_name],
                             ["=", ["timetable_id"], timetable_id]],
        }));
```

If rustfmt allows it, it's probably more readable like so. Also the `search` operator is only meant to be used with columns indexed to allow fuzzy matching.

editoast/src/views/search.rs
```diff
@@ -692,5 +743,73 @@ pub(super) struct SearchResultItemScenario {
...
+        }
+        let train_schedule_info: TrainScheduleInfo =
+            serde_json::from_str(include_str!("../tests/train_schedules/simple.json"))
+                .expect("Unable to parse");
+        let train_name = train_schedule_info.train_name;
+
+        // Add a train_schedule in the database
+        let train_schedule: TrainSchedule =
+            create_simple_train_schedule(&mut pool.get_ok(), timetable_id).await;
+        assert_eq!(train_schedule.train_name, train_name);
```
leovalais:
```suggestion
        // Add a train_schedule in the database
        let train_schedule: TrainSchedule =
            create_simple_train_schedule(&mut pool.get_ok(), timetable_id).await;
```

I believe `create_simple_train_schedule` already reads this file. I'd say it's not the responsibility of this test to `assert` whether the fixture works as expected or not and that it's fine to assume it does. Wdyt?

editoast/src/views/search.rs
```diff
@@ -657,6 +665,7 @@ pub(super) struct SearchResultItemStudy {
 )]
 #[allow(unused)]
 /// A search result item for a query with `object = "scenario"`
+#[derive(serde::Deserialize)]
```
leovalais: Is this necessary?

editoast/src/views/search.rs
```diff
@@ -683,6 +692,48 @@ pub(super) struct SearchResultItemScenario {
...
 
+#[derive(Search, Serialize, ToSchema)]
+#[search(
+    table = "train_schedule",
+    column(name = "timetable_id", data_type = "integer"),
+    column(name = "train_name", data_type = "string")
+)]
+#[allow(unused)]
+/// A search result item for a query with `object = "trainschedule"`
+#[derive(serde::Deserialize)]
```
leovalais:
```suggestion
#[derive(serde::Deserialize)]
/// A search result item for a query with `object = "trainschedule"`
```

leovalais (commented):
Thanks for adding the test! Just a few more details:

editoast/openapi.yaml
leovalais: Agreed on exposing this as an enum. This would require a conversion step after searching a train schedule. I can see multiple ways to achieve that:

1. Add a post-processing step when searching for a ts
2. Change the macro to allow some form of mapping, like it's done in ModelV2
3. Figure out a way to use what's already implemented in ModelV2 and make the search response optionally interface itself with `(trait Model)::Row`.

In order to decide which is best (if any), how severe is exposing this `int` is? Will the frontend even use it? If not, maybe removing it from the response solves our problem. Otherwise do we see another use case?

editoast/src/views/search.rs
EthanPERRUZZA: It thought it was a bad thing to hardcode the expected value as we might want to change the JSON test file one day ? but it surely does look less complicated and in fact this is not the responsability of this script.

editoast/src/views/search.rs
EthanPERRUZZA: No, I don't know how it ended up there, I must have not seen it wasn't the right place. I'll delete it :)

editoast/src/views/search.rs
```diff
@@ -692,5 +742,74 @@ pub(super) struct SearchResultItemScenario {
...
+
+        // Add a train_schedule in the database
+        create_simple_train_schedule(&mut pool.get_ok(), timetable_id).await;
+
+        let train_name = "ABC3615";
+
+        // The body
+        let request = app.post("/search").json(&json!({
+            "object": "trainschedule",
+            "query": ["and", ["=", ["train_name"], train_name],
```
leovalais:
```suggestion
        let train = create_simple_train_schedule(&mut pool.get_ok(), timetable_id).await;

        // The body
        let request = app.post("/search").json(&json!({
            "object": "trainschedule",
            "query": ["and", ["=", ["train_name"], train.name],
```

Wouldn't this work?

editoast/src/views/search.rs
```diff
@@ -683,6 +691,48 @@ pub(super) struct SearchResultItemScenario {
...
 }
 
+#[derive(Search, Serialize, ToSchema)]
+#[search(
+    table = "train_schedule",
+    column(name = "timetable_id", data_type = "integer"),
+    column(name = "train_name", data_type = "string")
+)]
+#[allow(unused)]
+#[derive(serde::Deserialize)]
```
leovalais: style nitpick:

```suggestion
#[derive(Search, Serialize, Deserialize, ToSchema)]
#[search(
    table = "train_schedule",
    column(name = "timetable_id", data_type = "integer"),
    column(name = "train_name", data_type = "string")
)]
#[allow(unused)]
```

or:

```suggestion
#[derive(Search, Serialize, ToSchema)]
#[cfg_attr(test, derive(serde::Deserialize))]
#[search(
    table = "train_schedule",
    column(name = "timetable_id", data_type = "integer"),
    column(name = "train_name", data_type = "string")
)]
#[allow(unused)]
```

(only derives the implementation when `test` is enabled since we don't need it otherwise)

leovalais (approved):
Just a few nitpicks, otherwise LGTM thank you! You have to squash your commits before merging otherwise the CI won't pass.

PR #9187, opened 2024-10-03
editoast: remove duplicate `TrainRequirement` struct

hamz2a:


leovalais (approved):
LGTM thanks for spotting that!

PR #9195, opened 2024-10-04
core: fix "negative" routing requirements + add requirement margin for stop on closed signal (conflict detection)

### LINKED ISSUE #8813: Core: start conflict detection on closed signal 20s before train restart
Related to https://github.com/osrd-project/osrd-confidential/issues/568

After work done in https://github.com/OpenRailAssociation/osrd/pull/7322, an improvement must be done: start resources reservation 20s earlier than train restart.

bougue-pe:
* Fix routing requirements cases where 'begin' was later than 'end' (stop on closed signal only).
* Also a 20 s anticipation for the start of resource requirement when stopping on closed signal.

Fix: https://github.com/OpenRailAssociation/osrd/issues/8813

### DIFF
--- a/core/src/main/java/fr/sncf/osrd/standalone_sim/ScheduleMetadataExtractor.kt
+++ b/core/src/main/java/fr/sncf/osrd/standalone_sim/ScheduleMetadataExtractor.kt
@@ -366,21 +366,22 @@ fun routingRequirements(
             ) ?: return null
         val limitingBlock = blockPath[limitingSignalSpec.blockIndex]
         val signal = blockInfra.getBlockSignals(limitingBlock)[limitingSignalSpec.signalIndex]
-        val limitingSignalOffset =
+        val limitingSignalOffsetInBlock =
             blockInfra.getSignalsPositions(limitingBlock)[limitingSignalSpec.signalIndex].distance
 
-        val blockOffset = blockOffsets[limitingSignalSpec.blockIndex]
-        val sightDistance = rawInfra.getSignalSightDistance(rawInfra.getPhysicalSignal(signal))
+        val limitingBlockOffset = blockOffsets[limitingSignalSpec.blockIndex]
+        val signalSightDistance =
+            rawInfra.getSignalSightDistance(rawInfra.getPhysicalSignal(signal))
 
         // find the location at which establishing the route becomes necessary
-        var criticalPos = blockOffset + limitingSignalOffset - sightDistance
+        val criticalPos = limitingBlockOffset + limitingSignalOffsetInBlock - signalSightDistance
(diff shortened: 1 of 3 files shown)


eckter (approved):

core/src/main/java/fr/sncf/osrd/standalone_sim/ScheduleMetadataExtractor.kt
```diff
@@ -40,6 +40,8 @@ import mu.KotlinLogging
 
 private val logger = KotlinLogging.logger {}
 
+const val CLOSED_SIGNAL_RESERVATION_MARGIN = 20.0
```
Khoyo, line 44: Could you add some documentation on that constant?

bougue-pe (commented):
~✅ just tested OK on small infra & France.
🎉 thanks for all this !~
EDIT: Wrong PR

core/src/main/java/fr/sncf/osrd/standalone_sim/ScheduleMetadataExtractor.kt
bougue-pe: Done in https://github.com/OpenRailAssociation/osrd/pull/9195/commits/d4e146bff6000da1b381f62fefdb24c80a92f703

bougue-pe:
Please review https://github.com/OpenRailAssociation/osrd/pull/9195/commits/d4e146bff6000da1b381f62fefdb24c80a92f703: I fixed a little bit the processing of critical distance, as it could lead to use a stop that was before the signalSight in routing requirement (too late `if`): https://github.com/OpenRailAssociation/osrd/pull/9195/commits/d4e146bff6000da1b381f62fefdb24c80a92f703#diff-8078a2f57f53541732c3008fa2e15551402baba5959c846d63fa84e9d9637e4bL395
Also checked that we are not requiring before the signalSight if stop on closed-signal is shorter than 20s and close to signal-sight position.

bougue-pe:
@Khoyo @eckter please re-review as there is an important fix in https://github.com/OpenRailAssociation/osrd/pull/9195/commits/8e3010d5236fa6a2f031ff0a3027447aa7635af5 (and a rename to clarify in [previous commit](https://github.com/OpenRailAssociation/osrd/pull/9195/commits/4510843e92409bf35ce7c94d46f0f351df65441c)).
This bug actually prevented to trigger the bug fixed in https://github.com/OpenRailAssociation/osrd/commit/d4e146bff6000da1b381f62fefdb24c80a92f703.

Tested locally that `assertRequirementsPeriodsConsistency` does break previous to that commit.

Khoyo (approved):

Erashin (approved):

PR #9202, opened 2024-10-04
editoast, schemas: add towed rolling stock

### LINKED ISSUE #8666: Update DB schema for train consist
DB schema must be updated with:
- add a new table for towed rolling stock
- add a new flag in existing rolling stock table for multiple unit

### LINKED ISSUE #8669: Add a new endpoint for GET towed rolling stocks
In STDCM interface, we will need to provide towed rolling stocks for the form. `editoast` need to implement that new endpoint.
- Define the JSON format of the API
- Define the new endpoint

woshilapin:
Resolves #8666 
Resolves #8669 

Define a new object in the model to describe "towed rolling stock". Towed rolling stock is a subset of a rolling stock, without traction capacity.

`python/osrd_schemas` has been modified to introduce this new type of rolling stock. The RAILJSON version has been bumped to 3.3.

In `editoast`, a few things have been done:
- define the new schema (corresponding to `python/osrd_schemas`)
- define the new model (for interacting with the DB)
- create the DB migrations to create the table
- add multiple new endpoints:
  - GET /towed_rolling_stock: to list the towed rolling stock
  - POST /towed_rolling_stock: to create a new towed rolling stock
  - GET /towed_rolling_stock/{id}: to get a specific towed rolling stock
  - PATCH /towed_rolling_stock/{id}: to update a specific towed rolling stock
  - PATCH /towed_rolling_stock/{id}/locked: to lock/unlock a specific towed rolling stock

⚠️ ~A unresolved question, I've used the same `BuiltinRole` than for rolling stock: should we create specific roles for towed rolling stock?~ **SOLVED**, not necessary for the moment
⚠️ ~I made a proposition in #9208, that I implemented in the current PR. If the other proposition is rejected, the current PR need to be modified accordingly.~ **SOLVED**, other PR has been merged so let's keep it in the current PR too.


flomonster:
> ⚠️ A unresolved question, I've used the same BuiltinRole than for rolling stock: should we create specific roles for towed rolling stock?

It's fine for now.

editoast/editoast_schemas/src/rolling_stock.rs
```diff
@@ -45,7 +48,7 @@ editoast_common::schemas! {
     loading_gauge_type::schemas(),
 }
 
-pub const ROLLING_STOCK_RAILJSON_VERSION: &str = "3.2";
+pub const ROLLING_STOCK_RAILJSON_VERSION: &str = "3.3";
```
flomonster: Why bumbing this value? Towed rolling stocks is another objects.

> [!NOTE]
> Bumping this value requires to create a migration to apply this new version to the RS in DB. It also requires to adapt core.

This property is only here to do a small check when importing / exporting objects. We haven't set up this system everywhere (train schedules). I'm not sure if we should still use it or not.

flomonster (commented):
Sorry for the late review.

LGTM. We just have to take a decision for the `railjson_version`.

editoast/editoast_schemas/src/rolling_stock.rs
woshilapin: So do I drop that change? Or do you think we should make a different versioning constant for towed rolling stock?

flomonster: I think we can drop this field for towed RS. And maybe we should discuss keeping it for other objects later?

woshilapin: Reversed in 06ead7376a9534b76d3abc5be79517d823ee99d3.

editoast/src/views/rolling_stock.rs
```diff
@@ -1,5 +1,6 @@
 pub mod form;
 pub mod light;
+pub mod towed;
```
leovalais:
```suggestion
mod towed;
```

I think this doesn't need to be `pub`, could be totally wrong though.

editoast/src/views/rolling_stock/towed.rs
```diff
@@ -0,0 +1,610 @@
...
+        },
+    },
+}
+
+editoast_common::schemas! {
+    TowedRollingStock,
+    TowedRollingStockCountList,
+    TowedRollingStockError,
+    TowedRollingStockForm,
+    TowedRollingStockIdParam,
```
leovalais:
```suggestion
```

Query parameters are always inlined.

editoast/src/views/rolling_stock/towed.rs
```diff
@@ -0,0 +1,610 @@
...
+            patch_by_id,
+            "/locked" => patch_by_id_locked,
+        },
+    },
+}
+
+editoast_common::schemas! {
+    TowedRollingStock,
+    TowedRollingStockCountList,
+    TowedRollingStockError,
```
leovalais:
```suggestion
```

Errors are added to the OpenAPI using another mechanism.

editoast/src/views/rolling_stock/towed.rs
```diff
@@ -0,0 +1,610 @@
...
+            length: towed_rolling_stock.length,
+            comfort_acceleration: towed_rolling_stock.comfort_acceleration,
+            startup_acceleration: towed_rolling_stock.startup_acceleration,
+            inertia_coefficient: towed_rolling_stock.inertia_coefficient,
+            rolling_resistance: towed_rolling_stock.rolling_resistance,
+            gamma: towed_rolling_stock.gamma,
+        }
+    }
+}
+#[derive(Debug, Error, EditoastError, ToSchema, Serialize)]
```
leovalais:
```suggestion
#[derive(Debug, Error, EditoastError)]
```

editoast/src/views/rolling_stock/towed.rs
```diff
@@ -0,0 +1,610 @@
...
+    let authorized = authorizer
+        .check_roles([BuiltinRole::RollingStockCollectionRead].into())
+        .await
+        .map_err(AuthorizationError::AuthError)?;
+    if !authorized {
+        return Err(AuthorizationError::Unauthorized.into());
+    }
+    let settings = page_settings
+        .validate(1000)?
+        .warn_page_size(100)
```
leovalais:
```suggestion
        .validate(50)?
```

Since this is a new endpoint I'd suggest truly enforcing the pagination (50 seems like a good max to me, but feel free to change that). Wdyt?

editoast/src/views/rolling_stock/towed.rs
```diff
@@ -0,0 +1,610 @@
...
+    Ok(Json(TowedRollingStockCountList {
+        results: towed_rolling_stocks
+            .into_iter()
+            .map(TowedRollingStock::from)
+            .collect(),
+        stats,
+    }))
+}
+
+#[derive(Debug, IntoParams, ToSchema)]
```
leovalais:
```suggestion
#[derive(Debug, IntoParams)]
```

editoast/src/views/rolling_stock/towed.rs
```diff
@@ -0,0 +1,610 @@
...
+#[allow(unused)]
```
leovalais: You can avoid the `#[allow(unused)]` by replacing in each handlers:

```rust
Path(towed_rolling_stock_id): Path<i64>
```

by:

```rust
Path(TowedRollingStockIdParam { towed_rolling_stock_id }): Path<TowedRollingStockIdParam>
```

but you'll have to make `TowedRollingStockIdParam` derive `Deserialize`.

I don't have any strong opinion on this, just saying this option also exists. Feel free to resolve.

editoast/src/views/rolling_stock/towed.rs
```diff
@@ -0,0 +1,610 @@
...
+            .json_into();
+
+        assert!(!towed_rolling_stocks.results.is_empty());
+    }
+
+    #[rstest]
+    async fn get_unknown_towed_rolling_stock() {
+        let app = TestAppBuilder::default_app();
+
+        let id: i64 = rand::random();
```
leovalais, line 413: Not a huge fan of unbounded randomness in tests. What about
```suggestion
        let id: i64 = i64::MAX;
```

Because a failure can be hard to diagnose if `id` occasionally designates a towed rolling stock that already exists in the DB.

leovalais (approved):
LGTM. Sorry about the time I took to get to review this 😞. Just a few nitpicks:

editoast/editoast_schemas/src/rolling_stock.rs
flomonster: I was also thinking of removing the `railjson_version` field from the towed rolling stock. We can keep it that way if you want to be consistent with rolling stocks :shrug:

flomonster (approved):
Great PR!

editoast/src/views/rolling_stock/towed.rs
woshilapin: Done in d2f22638571f1a0811ffa691555eebff3d594878.

editoast/src/views/rolling_stock/towed.rs
woshilapin: Done in d2f22638571f1a0811ffa691555eebff3d594878.

editoast/src/views/rolling_stock/towed.rs
woshilapin: Done in d2f22638571f1a0811ffa691555eebff3d594878.

editoast/src/views/rolling_stock/towed.rs
woshilapin: Done in d2f22638571f1a0811ffa691555eebff3d594878.

editoast/src/views/rolling_stock/towed.rs
woshilapin: Good idea, I just copied/pasted but indeed. Done in 9e25169d761385f34419b82e843d4d08e3b8f8f8.

editoast/src/views/rolling_stock.rs
woshilapin: You're right, done in 54d9f7c3ae23bc027ae0d56b39ed1a83dd4d65f8.

editoast/src/views/rolling_stock/towed.rs
woshilapin: Thanks for the hint. I did use it in 8880e2d7e0c358714555fa93aee4b55172f14a05.

editoast/src/views/rolling_stock/towed.rs
woshilapin: Mmmm, I'd say we should be pretty safe. A lot of systems actually use UUID as identifier in production (not in tests), assuming there will never be a collision (the chance is so low). From what I understand in [this documentation](https://en.wikipedia.org/w/index.php?title=Universally_unique_identifier&oldid=755882275#Random_UUID_probability_of_duplicates), we should be able to calculate the probability of collision for an `i64` after generating a million of rolling stocks (which is already way more than we will ever have).

$`\frac{{1000000}^2}{2 • 2^{64}} = 2.710 • 10^{-8}`$

I agree `i64` is still way less than an UUID (which is 128 bits), but still, I'd say it's OK.

Also, I don't see `i64::MAX` as a good solution either because what if I want to write another test? In another file? It's going to start to be unmaintainable to find identifiers.

What do you think?

editoast/editoast_schemas/src/rolling_stock.rs
woshilapin: I'll keep it for consistency.

editoast/src/views/rolling_stock/towed.rs
leovalais: Sure, you're right. I just got a bit triggered by having randomness in a non property-based test, but factually the probability of things going wrong is indeed very low.

> Also, I don't see i64::MAX as a good solution either because what if I want to write another test?

Our tests are isolated in their own transaction so there shouldn't be conflicting IDs coming from the tests. ID conflicts are still possible with data that's already there from before the execution of the test though. 

Resolving this (especially since we have bigger flakiness issues on our hands already 🙃).

editoast/src/views/rolling_stock/towed.rs
```diff
@@ -0,0 +1,610 @@
...
+            },
+        });
+
+        let request = app
+            .post("/towed_rolling_stock")
+            .json(&towed_rolling_stock_json);
+
+        let towed_rolling_stock: TowedRollingStock =
+            app.fetch(request).assert_status(StatusCode::OK).json_into();
+        assert_eq!(towed_rolling_stock.name, name);
```
Sh099078: Maybe we could create a towed rolling stock fixture or some `create_towed_rolling_stock()` method in the test module to avoid copy pasting the rolling stock create endpoint call in other tests that are not about this specific endpoint ?

editoast/src/views/rolling_stock/towed.rs
```diff
@@ -0,0 +1,610 @@
...
+        assert_eq!(created_towed_rolling_stock.name, name);
+
+        let id = created_towed_rolling_stock.id;
+
+        let get_towed_rolling_stock: TowedRollingStock = app
+            .fetch(app.get(&format!("/towed_rolling_stock/{id}")))
+            .assert_status(StatusCode::OK)
+            .json_into();
+
+        assert_eq!(get_towed_rolling_stock.name, name);
```
Sh099078: Shouldn't we explicitly test all the fields in the creation endpoint against an expected result ? I think for this one, the equality check should be more throughout so that it catches mistakes like inverting two columns of the same type in the request like `mass` and `length`.

editoast/src/views/rolling_stock/towed.rs
```diff
@@ -0,0 +1,610 @@
...
+        .await
+        .map_err(AuthorizationError::AuthError)?;
+    if !authorized {
+        return Err(AuthorizationError::Unauthorized.into());
+    }
+
+    let conn = &mut db_pool.get().await?;
+
+    // FIXME: check that the towed rolling stock exists (the Option<RollingStockModel> is ignored here)
+    // should return a 404 instead of a 204 in case the identifier doesn't exist.
```
Sh099078: Maybe we should create an issue for that one (or fix it) ?

Sh099078 (changes requested):
Great PR ! I left a few comments :)

Wadjetz (approved):

editoast/src/views/rolling_stock/towed.rs
woshilapin: Good idea, done in 9d700232408b9c45f92956150fe5c25f62ceb0fb.

editoast/src/views/rolling_stock/towed.rs
woshilapin: Done in be1d9c704c644fde51c1a716157514262bd8471d.

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -3082,6 +3160,34 @@ export type TrainScheduleResult = TrainScheduleBase & {
...
+  comfort_acceleration: number;
+  gamma: Gamma;
+  inertia_coefficient: number;
+  length: number;
+  locked: boolean;
+  mass: number;
+  name: string;
+  rolling_resistance: RollingResistance;
+  startup_acceleration: number;
+};
```
SharglutDev, line 3239: Instead of having two types almost identical, could this one have been an Omit<TowedRollingStock, ["id", "railjson_version"]> ? (in rust equivalent)

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -1514,6 +1564,34 @@ export type PostTimetableByIdTrainScheduleApiArg = {
...
+export type PostTowedRollingStockApiResponse =
+  /** status 200 The created towed rolling stock */ TowedRollingStock;
+export type PostTowedRollingStockApiArg = {
+  towedRollingStockForm: TowedRollingStockForm;
+};
+export type GetTowedRollingStockByTowedRollingStockIdApiResponse =
+  /** status 200 The requested towed rolling stock */ TowedRollingStock;
+export type GetTowedRollingStockByTowedRollingStockIdApiArg = {
+  towedRollingStockId: number;
```
SharglutDev, line 1582: For the "standard" rolling stock, we use mostly "name" instead of "id" now. Won't we want to do the same with the towed rolling stock ?

SharglutDev (commented):
Thank you for your PR, left just two questions :)

editoast/src/views/rolling_stock/towed.rs
woshilapin: OK, finally, by asking around, there was a function to fix it. See e0209eb5c751a1c8708c6ad9bb8363251aaf60fc.

front/src/common/api/generatedEditoastApi.ts
woshilapin: I've seen that for rolling stocks. It's possible to do, but it does add some complexity in the code. Do you know for sure it will be needed? If not, I would propose to stay on the simple version until we know we need it. To my knowledge, it will only be needed in STDCM form at the moment (an input field in the form that will get the list of towed rolling stocks and send back one of them, an ID I guess, to `/stdcm` endpoint). What do you think?

front/src/common/api/generatedEditoastApi.ts
woshilapin: I do not think this is possible in Rust (or at least, I don't see an obvious solution). Note that I mostly reproduce what already exist on `RollingStock` and `RollingStockForm` (but that is not a good argument to not look for improvement, noted 😄).

If I understand correctly, it's some kind of inheritance. The OpenApi format does support inheritance, but I looked for inheritance in `utoipa` (the tool in `editoast` that generates OpenApi) without finding anything.

SharglutDev: Yes you're right it's not the only spot where it could be changed. Was just to pointing that out but I understand it might be difficult to target that behavior with open api conversion :) We can keep that in mind for later if we find a better solution.

front/src/common/api/generatedEditoastApi.ts
SharglutDev: This change from `id` to `name` has been done in operational studies for the trainschedule endpoints (using `TrainScheduleBase`). But we will probably keep needing the endpoints with `id` for the rolling stock editor and maybe we will want to use towed rs in the editor as well in the future ?

Anyway you argument is totally legit as this is used only for stdcm where we don't use `trainschedule` :)

SharglutDev (approved):
Lgtm (not tested), good job :)

front/src/common/api/generatedEditoastApi.ts
woshilapin: ok, let's leave it for when we implement the rolling stock editor part., if ever.

front/src/common/api/generatedEditoastApi.ts
woshilapin: That said, in both cases, it's 2 different types, so writing it with `Omit` or in full version is the same if I understand well. And we could argue it's more code to maintain... but this code is auto-generated, so it doesn't even apply here.

Sh099078 (approved):

PR #9207, opened 2024-10-04
editoast: bump rolling stock version in a transaction

woshilapin:
Updating then saving a new version based on the result of the update should be done in a transaction to avoid race conditions.

### DIFF
--- a/editoast/src/views/rolling_stock.rs
+++ b/editoast/src/views/rolling_stock.rs
@@ -13,6 +13,7 @@ use axum::extract::State;
 use axum::http::StatusCode;
 use axum::response::IntoResponse;
 use axum::Extension;
+use diesel_async::scoped_futures::ScopedFutureExt as _;
 use editoast_authz::BuiltinRole;
 use editoast_derive::EditoastError;
 use editoast_models::DbConnection;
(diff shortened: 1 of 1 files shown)


editoast/src/views/rolling_stock.rs
```diff
@@ -370,21 +371,32 @@ async fn update(
...
+                let mut new_rolling_stock =
+                    Into::<Changeset<RollingStockModel>>::into(rolling_stock_form)
+                        .update(&mut conn.clone(), rolling_stock_id)
+                        .await
+                        .map_err(|e| map_diesel_error(e, name.clone()))?
+                        .ok_or(RollingStockError::KeyNotFound {
+                            rolling_stock_key: RollingStockKey::Id(rolling_stock_id),
+                        })?;
+
+                if new_rolling_stock != previous_rolling_stock {
```
emersion, line 389: Shouldn't we grab `previous_rolling_stock` from inside the transaction as well? Otherwise we might be working with a stale `previous_rolling_stock`.

woshilapin: Indeed, I believe you're right. Thanks for catching that. Done in [dc3b78b](https://github.com/OpenRailAssociation/osrd/pull/9207/commits/dc3b78b710a2023514c7fdf0c26a7214c0285f86).

emersion (approved):

flomonster (approved):

PR #9208, opened 2024-10-04
editoast: locked rolling stocked returns a 409

woshilapin:
When a rolling stock is locked and we try to update it, we return a [400 BAD_REQUEST](https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.1). I don't think the request is malformed in case of a locked rolling stock.

Based on the description of what is a 409, the behavior of our system in case a rolling stock is already locked is more a [409 CONFLICT](https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.8).

⚠️ This PR is a proposition. If other developers disagree, we can just close it.

### DIFF
--- a/editoast/openapi.yaml
+++ b/editoast/openapi.yaml
@@ -4979,7 +4979,7 @@ components:
         status:
           type: integer
           enum:
-          - 400
+          - 409
         type:
           type: string
           enum:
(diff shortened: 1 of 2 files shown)


leovalais (approved):
Seems more correct to me indeed.

flomonster (approved):

PR #9211, opened 2024-10-04
Add roles CLI

### LINKED ISSUE #8744: RBAC: CLI interface

leovalais:
closes #8744


editoast/src/client/roles.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+pub fn list_roles() {
+    BuiltinRole::iter().for_each(|role| println!("{role}"));
+}
+
+async fn parse_and_fetch_subject_id(
+    subject: &String,
+    driver: &PgAuthDriver<BuiltinRole>,
+) -> anyhow::Result<i64> {
```
flomonster: We could return more information to have more user friendly prompt:

```diff
$ editoast roles list 2
- Subject 2 has no roles assigned
+ Subject flomonster (2) has no roles assigned
```

```suggestion
) -> anyhow::Result<UserInfo> {
```

editoast/src/client/roles.rs
```diff
@@ -0,0 +1,136 @@
...
+pub async fn add_roles(
+    AddArgs { subject, roles }: AddArgs,
+    conn: DbConnection,
+) -> anyhow::Result<()> {
+    let driver = PgAuthDriver::<BuiltinRole>::new(conn);
+    let subject_id = parse_and_fetch_subject_id(&subject, &driver).await?;
+    let roles = roles
+        .iter()
+        .map(|role| {
+            role.parse::<BuiltinRole>()
```
flomonster: This auto implementation is case sensitive which is not ideal. We could:
- Implement manually `std::str::FromStr` to be case insensitive.
- Add a lowercase version of all variant and lower case the input role.

flomonster (approved):
LGTM.

I came across two irritating little things while testing the CLI.

editoast/src/client/roles.rs
leovalais: I'm a bit hesitant to add such DoF to the builtin roles serialization at the API level just for this use case.

I've implemented this behavior in the parsing step directly (admittedly it's a O(n^2) operation but it's in the CLI honestly who cares). I'll let you resolve if it's ok.

woshilapin (changes requested):
Tested and might be missing some information on the CLI. Here is an example (I removed all `tracing` logs).

```
❯ export RUST_LOG=none 
❯ editoast roles list 1
morane#1 (Bob Morane) has no roles assigned

❯ editoast roles add 1 SuperUser StdcmAdmin

❯ editoast roles list 1
Superuser
StdcmAdmin

❯ editoast roles remove 1 SuperUser StdcmAdmin

```

A few things to note that could be improved.
- When adding roles, there is no message confirming everything went fine.
- When removing roles, there is no message confirming everything went fine.
- When adding or removing roles of a user that doesn't exist, there is also no message.
- When listing roles and the user has no roles, the user and its ID are clearly identified... but not when the user has roles, only the roles are displayed: maybe having also the user printed would be nice

Another problem I see is that I would expect `add` to append roles, but it actually overwrites the existing list. See for example.

```
❯ editoast roles add 1 SuperUser StdcmAdmin
❯ editoast roles list 1
Superuser
StdcmAdmin
❯ editoast roles add 1 OpsWrite
❯ editoast roles list 1
OpsWrite
```
I would expect the last command to output 3 values.

Would it make sense to add some tests about it?

leovalais:
As discussed:

> Another problem I see is that I would expect add to append roles, but it actually overwrites the existing list. 

Can't reproduce.

> A few things to note that could be improved.
> When adding roles, there is no message confirming everything went fine.
>When removing roles, there is no message confirming everything went fine.
>When adding or removing roles of a user that doesn't exist, there is also no message.
>When listing roles and the user has no roles, the user and its ID are clearly identified... but not when the user has roles, only the roles are displayed: maybe having also the user printed would be nice

My logic for choosing between logging and `println!` was that logging is only meant to be informative and provide context while `println!` is an output that can be used in a script (for piping into another process for ex.). This is debatable and can be changed ofc.

@woshilapin Wdyt?

woshilapin:
> As discussed:
> 
> > Another problem I see is that I would expect add to append roles, but it actually overwrites the existing list.
> 
> Can't reproduce.

Can't reproduce either, I might have messed up somewhere, sorry.

> My logic for choosing between logging and `println!` was that logging is only meant to be informative and provide context while `println!` is an output that can be used in a script (for piping into another process for ex.). This is debatable and can be changed ofc.
> 
> @woshilapin Wdyt?

I agree, that was a point of view I didn't have and a totally valid one. Fine with me.

woshilapin (approved):
Thank you. Let's merge that 😄

flomonster (approved):
Thanks!

PR #9213, opened 2024-10-04
front: upgrade vitest to v2

emersion:
Release notes:
https://github.com/vitest-dev/vitest/releases/tag/v2.0.0

No actions needed on our end.

Also configure dependabot to upgrade all vitest packages together because they need to be kept in sync.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -46,6 +46,10 @@ updates:
       turf:
         patterns:
           - "@turf/*"
+      vitest:
+        patterns:
+          - "vitest"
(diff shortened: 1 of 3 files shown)


SarahBellaha (approved):
Lgtm :)

SharglutDev (approved):
Lgtm. 

Why do we sometimes use `describe` and `it` from `vitest` and sometimes we don't ? Should we do something about that ?

flomonster (approved):

ghost (approved):
LGTM, not tested, trusting you on this one.

PR #9219, opened 2024-10-07
front: scenario page size adjustment

### LINKED ISSUE #9218: Unnecessary spacing in the scenario page
### What happened?

image

Go to scenario page
See spacing at bottom of page

Local, Dev (SNCF), Recette (SNCF)

Google Chrome

87a6547

theocrsb:
closes #9218

### DIFF
--- a/front/src/modules/conflict/styles/_conflictList.scss
+++ b/front/src/modules/conflict/styles/_conflictList.scss
@@ -5,6 +5,7 @@
   height: auto;
   min-height: 43px;
   overflow: hidden;
+  margin-bottom: 0.5rem;
   .conflicts-container {
     padding: 0.2rem;
     display: none;
--- a/front/src/styles/scss/applications/operationalStudies/_scenario.scss
+++ b/front/src/styles/scss/applications/operationalStudies/_scenario.scss
@@ -1,6 +1,6 @@
 .scenario {
-  --navbar-height: 4.0625rem;
-  --content-height: calc(100vh - var(--navbar-height) - 0.5rem);
+  --navbar-height: 3.75rem;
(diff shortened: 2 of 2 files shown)


emersion:
Seems like the conflict infobox doesn't stick to the bottom when there are only a few trains. See the blank space at the bottom:

out

emersion:
And with a lot of trains it seems there is a missing padding at the bottom:

out

theocrsb:
> Seems like the conflict infobox doesn't stick to the bottom when there are only a few trains. See the blank space at the bottom:
> 
> out

This behavior must be fixed by this [PR](https://github.com/OpenRailAssociation/osrd/pull/9219)

theocrsb:
> And with a lot of trains it seems there is a missing padding at the bottom:

image
I didn't put a `padding bottom` to prepare the next PR on the redesign of the conflict list.

emersion:
> I didn't put a `padding bottom` to prepare the next PR on the redesign of the conflict list. 

I see. I wonder if we really want to have a weird-looking UI during the transition though. (I suspect we will need to do a bunch more work anyways to adapt to the new UI, e.g. reorganizing the HTML.)

theocrsb:
> > I didn't put a `padding bottom` to prepare the next PR on the redesign of the conflict list.
> 
> I see. I wonder if we really want to have a weird-looking UI during the transition though. (I suspect we will need to do a bunch more work anyways to adapt to the new UI, e.g. reorganizing the HTML.)

You're right, it's done.

emersion (approved):
LGTM and works!

SharglutDev (approved):
Lgtm and tested

PR #9221, opened 2024-10-07
front: prevent the via card to be demounted and remounted several times

### LINKED ISSUE #8986: CH input blinking when changing stop duration
### What happened?

In stdcm, when changing th stop duration of a via, the ch component is blinking. There seems to be a lot of rerenders in the page when changing an input.

[Capture vidéo du 25-09-2024 10:07:22.webm](https://github.com/user-attachments/assets/471f8687-0c41-4326-9f1d-4bab3827da92)

The CH input shouldn't be blinking

1. Go to stdcm 
2. Add a via
3. Change the stop duration

Local

_No response_

.

### LINKED ISSUE #9107: STDCM: An empty "Intermediate OP" can't be deleted
### What happened?

https://github.com/user-attachments/assets/faadb449-a158-4850-9d31-dc123f508ac0

The empty "Intermediate OP" is deleted 

1. In STDCM Interface add a "Intermediate OP"
2. Cliquer on Delete button

Dev (SNCF), Recette (SNCF)

Google Chrome

a09a974

clarani:
closes #8986 
closes #9107

The via card was remounted because its key changed at each re-render (the key was nextId()). To avoid this, I modified the behaviour to create a via and give by default an id and a uic (= -1). This way, we can use the via.id for the key. It solves the main issue.

To handle the "fake" uic and avoid sending invalid payload to the pathfinding endpoint, I created a new hook usePathfindingIsPossible which filter the invalid uic and launch a pathfinding.
This hook is simpler than the existing usePathfinding hook, but I had to do some refacto in the map props (to give the path geometry separately from the path properties). This will be useful for a later refacto.

### DIFF
--- a/front/src/applications/stdcm/views/StdcmConfig.tsx
+++ b/front/src/applications/stdcm/views/StdcmConfig.tsx
@@ -1,3 +1,4 @@
+// /!\ This file will be removed soon, don't update it /!\
 import { useEffect, useMemo, useState } from 'react';
 
 import { ChevronDown, ChevronUp } from '@osrd-project/ui-icons';
--- a/front/src/applications/stdcmV2/components/StdcmConfig.tsx
+++ b/front/src/applications/stdcmV2/components/StdcmConfig.tsx
@@ -5,9 +5,7 @@ import cx from 'classnames';
 import { useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
 
-import type { ManageTrainSchedulePathProperties } from 'applications/operationalStudies/types';
 import { useOsrdConfActions, useOsrdConfSelectors } from 'common/osrdContext';
-import { usePathfinding } from 'modules/pathfinding/hooks/usePathfinding';
 import { Map } from 'modules/trainschedule/components/ManageTrainSchedule';
 import type { StdcmConfSliceActions } from 'reducers/osrdconf/stdcmConf';
 import { useAppDispatch } from 'store';
@@ -19,6 +17,7 @@ import StdcmOrigin from './StdcmOrigin';
 import StdcmSimulationParams from './StdcmSimulationParams';
 import StdcmVias from './StdcmVias';
 import StdcmWarningBox from './StdcmWarningBox';
+import useStaticPathfinding from '../hooks/useStaticPathfinding';
 import { ArrivalTimeTypes, StdcmConfigErrorTypes } from '../types';
 import type { StdcmConfigErrors, StdcmSimulation, StdcmSimulationInputs } from '../types';
 import checkStdcmConfigErrors from '../utils/checkStdcmConfigErrors';
@@ -65,10 +64,7 @@ const StdcmConfig = ({
   const studyID = useSelector(getStudyID);
   const scenarioID = useSelector(getScenarioID);
 
-  const [pathfindingProperties, setPathfindingProperties] =
-    useState<ManageTrainSchedulePathProperties>();
-
-  const { pathfindingState } = usePathfinding(setPathfindingProperties, pathfindingProperties);
+  const pathfinding = useStaticPathfinding();
 
   const [formErrors, setFormErrors] = useState<StdcmConfigErrors>();
 
@@ -81,7 +77,7 @@ const StdcmConfig = ({
   };
 
   const startSimulation = () => {
-    if (pathfindingState.done && !formErrors) {
+    if (pathfinding?.status === 'success' && !formErrors) {
       launchStdcmRequest();
     }
   };
@@ -94,7 +90,7 @@ const StdcmConfig = ({
   };
 
   useEffect(() => {
-    const isPathfindingFailed = pathfindingState.error !== '';
+    const isPathfindingFailed = !!pathfinding && pathfinding.status !== 'success';
     let formErrorsStatus = checkStdcmConfigErrors(isPathfindingFailed, origin, destination);
     if (formErrorsStatus?.errorType === StdcmConfigErrorTypes.BOTH_POINT_SCHEDULED) {
       formErrorsStatus = {
@@ -110,7 +106,7 @@ const StdcmConfig = ({
       };
     }
     setFormErrors(formErrorsStatus);
-  }, [origin, destination, pathfindingState.error]);
+  }, [origin, destination, pathfinding]);
 
   // TODO: DROP STDCMV1: set those values by default in the store when <StdcmAllowances/> is not used anymore.
   useEffect(() => {
@@ -144,7 +140,7 @@ const StdcmConfig = ({
             {/* <StdcmDefaultCard text="Indiquer le sillon postérieur" Icon={<ArrowDown size="lg" />} /> */}
             <div
               className={cx('stdcm-v2-launch-request', {
-                'wizz-effect': !pathfindingState.done || formErrors,
+                'wizz-effect': pathfinding?.status !== 'success' || formErrors,
               })}
             >
               {showBtnToLaunchSimulation && (
@@ -166,7 +162,7 @@ const StdcmConfig = ({
             hideAttribution
             hideItinerary
             preventPointSelection
-            pathProperties={pathfindingProperties}
+            pathGeometry={pathfinding?.geometry}
             showStdcmAssets
           />
         </div>
--- a/front/src/applications/stdcmV2/components/StdcmResults/StdcmResults.tsx
+++ b/front/src/applications/stdcmV2/components/StdcmResults/StdcmResults.tsx
@@ -131,7 +131,7 @@ const StcdmResults = ({
             hideAttribution
             showStdcmAssets
             setMapCanvas={setMapCanvas}
-            pathProperties={selectedSimulation.outputs?.pathProperties}
+            pathGeometry={selectedSimulation.outputs?.pathProperties.geometry}
             simulationPathSteps={selectedSimulation.outputs?.results.simulationPathSteps}
           />
         </div>
--- a/front/src/applications/stdcmV2/components/StdcmVias.tsx
+++ b/front/src/applications/stdcmV2/components/StdcmVias.tsx
@@ -97,7 +97,7 @@ const StdcmVias = ({ disabled = false, setCurrentSimulationInputs }: StdcmConfig
   };
 
   const addViaOnClick = (pathStepIndex: number) => {
-    const newPathSteps = addElementAtIndex(pathSteps, pathStepIndex, null);
+    const newPathSteps = addElementAtIndex(pathSteps, pathStepIndex, { id: nextId(), uic: -1 });
     dispatch(updatePathSteps({ pathSteps: newPathSteps }));
   };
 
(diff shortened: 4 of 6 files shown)


front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
```diff
@@ -119,7 +119,7 @@ const ManageTrainSchedule = ({ trainIdToEdit }: ManageTrainScheduleProps) => {
         <div className="floating-itinerary">
           <Itinerary pathProperties={pathProperties} setPathProperties={setPathProperties} />
         </div>
-        <Map pathProperties={pathProperties}>
+        <Map pathProperties={pathProperties} pathGeometry={pathProperties.geometry}>
```
SarahBellaha:
```suggestion
        <Map pathProperties={pathProperties} pathGeometry={pathProperties?.geometry}>
```
`pathProperties` is possibly `undefined` here :)

SarahBellaha (commented):
Thx for this PR 💪 
There is some warnings about the order of the imports in `StdcmConfig.tsx` & `usePathfindingIsPossible.ts`

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
```diff
@@ -0,0 +1,99 @@
...
+  const pathProperties = usePathProperties(
+    infra?.id,
+    pathfinding?.status === 'success' ? pathfinding : undefined,
+    ['geometry']
+  );
+
+  useEffect(() => {
+    const launchPathfinding = async () => {
+      setPathfinding(undefined);
+      if (!infra || infra.state !== 'CACHED' || !rollingStock) {
```
SharglutDev: It might probably work like that too ?

```suggestion
      if (infra?.state !== 'CACHED' || !rollingStock) {
```

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
```diff
@@ -0,0 +1,99 @@
...
+        return;
+      }
+
+      const pathItems = compact(pathStepsLocation).map((step) => {
```
SharglutDev: We do the same thing in `getPathfindingQuery`, should extract this logic in an util ?

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
```diff
@@ -0,0 +1,99 @@
...
+        }
+        return {
+          trigram: step.trigram,
+          secondary_code: step.secondary_code,
+        };
+      });
+
+      if (pathItems.length < 2) {
+        return;
+      }
```
SharglutDev: We could do this check before defining `pathItems` and use `pathStepsLocation`

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
```diff
@@ -0,0 +1,99 @@
...
+import { osrdEditoastApi, type PathfindingResult } from 'common/api/osrdEditoastApi';
+import { useOsrdConfSelectors } from 'common/osrdContext';
+
+import { useStoreDataForRollingStockSelector } from 'modules/rollingStock/components/RollingStockSelector/useStoreDataForRollingStockSelector';
+
+import useInfraStatus from 'modules/pathfinding/hooks/useInfraStatus';
+import { getSupportedElectrification, isThermal } from 'modules/rollingStock/helpers/electric';
+import usePathProperties from 'modules/pathfinding/hooks/usePathProperties';
+
+const usePathfindingIsPossible = () => {
```
SharglutDev: Since this usecase for a `uic === -1` is specific to stdcm, maybe we can rename this hook `useStdcmPathfinding` ? This way, we know we should use it elsewhere.

front/src/applications/stdcm/views/StdcmConfig.tsx
SharglutDev, line 1: We should stop maintaining these unused files :) Could we put some comments on top of the concerned files for that while we wait for @SarahBellaha 's cleaning PR ?

front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
SharglutDev: I find it pretty weird to pass two times the same data in the `Map` component. `geometry` is already included in `pathProperties` so we shouldn't have to pass it another time in the `pathGeometry` prop. And it the `Map component` we have to sources of truth, it might no be very good.

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
SharglutDev, line 1: Is this new hook really needed ? Couldn't we just add the `uic !== -1` logic in `usePathfinding` since it won't break anything even for operational studies ? (If needed, we could have an `isStdcm` check inside to avoid unwanted computations for stdcm. But it can be part of this hook refacto)

SharglutDev (changes requested):
Thank you for this fix ! Left some comments

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
clarani: Done

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
clarani: Done :white_check_mark:

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
clarani: Done

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
clarani: Yes, I introduced this hook now to avoid to have a too big PR for when I modify the PathStep type (and introduce a new StdcmPathStep type specific for stdcm).
Moreover, the hook usePathfinding is quite complex, and I would like to avoid use it again

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
clarani: Is it ok for you if we keep it like this ? Right now, this hook is only used in stdcm, but it could also be used in another part of the application

front/src/applications/stdcm/views/StdcmConfig.tsx
clarani: I have added 1 comment at the top, but I think Sarah is going to remove the files very soon, so maybe it is not very useful ? :)

front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
clarani: I have changed a bit the behaviour, now you can pass the path geometry only, the path properties or both :+1:

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
```diff
@@ -0,0 +1,77 @@
...
+      setPathfinding(undefined);
+      if (infra?.state !== 'CACHED' || !rollingStock || pathStepsLocations.length < 2) {
+        return;
+      }
+
+      const payload = getPathfindingQuery({
+        infraId: infra.id,
+        rollingStock,
+        origin: pathStepsLocations.at(0)!,
+        destination: pathStepsLocations.at(-1)!,
```
SharglutDev: We don't need to pass the `!` for origin and destination (are we sure that when calling this hook, they will be defined already ?), the function returns `null` if one of them is `null`.

front/src/applications/stdcm/views/StdcmConfig.tsx
SharglutDev: I had the same discussion with @Math-R last week and the files are still there ;)

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
```diff
@@ -0,0 +1,77 @@
...
+      setPathfinding(pathfindingResult);
+    };
+
+    launchPathfinding();
+  }, [pathStepsLocations, rollingStock, infra]);
+
+  const result = useMemo(
+    () =>
+      pathfinding
+        ? { status: pathfinding.status, geometry: pathProperties?.geometry || undefined }
```
SharglutDev: Probably not needed

```suggestion
        ? { status: pathfinding.status, geometry: pathProperties?.geometry }
```

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
SharglutDev: Ok fine but I still don't like the name very much :) We could do `useSimplifiedPathfinding` or `useStaticPathfinding` (meaning it doesn't update the store contrary to the other)

front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
SharglutDev: Ok. As you spotted it, we could add a small refacto issue to pass only `{ pathGeometry?, pathLength? }` instead of keeping both these props and change pathProperties to this new object until the store.

clarani: Maybe we can do this in a later PR ? :)

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
clarani: Done

front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
SharglutDev: Yes of course, I was just suggesting to create the issue so we don't forget :)

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
clarani: I had to replace the `!` by `|| null` because the getPathfindingQuery function expected `PathStep | null` and not `PathStep | undefined`

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
clarani: I renamed it useStaticPathfinding :white_check_mark:

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
SharglutDev: You left a `!` for the destination

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
clarani: Actually, I need this undefined, because otherwise, geometry can be defined, null or undefined (but in Map, it can only be defined or undefined)

front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
clarani: Here is the issue :) https://github.com/OpenRailAssociation/osrd/issues/9266
I also added to the board of the enhancements

front/src/applications/stdcmV2/hooks/usePathfindingIsPossible.ts
clarani: Done

SharglutDev (approved):
Lgtm and tested, great job !

Issue #7706, opened 2024-06-13
[META] Adding E2E tests


Maymanaf:
### Description and goal

Some functionalities are not tested with e2e-tests, we have to complete our battery of tests

### Configuration

- [x] Add Webkit ( Safari !) #8645 => Removed #10478
### Acceptance criteria

Tests to write:

- [x] Rolling stock editor 
  - [x] refacto creation #7763 
  - [x] update #7763 
  - [x] delete #7774 
  - [x] Filtering and searching #7826 
- [x] Operational studies 
  - [x] Rolling stock #7911
  - [x] RS +Route #7912
  - [x] Times & stops #8646
  - [x] Simulation Settings #8647
  - [x] Full operation : RS+ Route + Times & stops + Simulation Settings #8648
- [x] Import of trains ( infra et train de test)
  - [x] Simulation Results (The visibilty of each element) #8040
  - [x] Filtering  #8040
  - [x] Verify the data of each train (RS + Route + Times & stops + Simulation Settings) #8649
  - [x] Import / Export 
- [x] Last Minute Request
  - [x] #9265
  - [x] #9585
  - [x] Simulation sheet
  - [ ] ~~Debugging mode~~
- [x]  Simulation Results
  - [x] GET #9586
  - [x] GEV  #8651
  - [x] Output Table #8646 and #8647
  - [ ] ~~Simulation sheet (pdf)~~
  - [ ] ~~Export csv~~
  - [ ] ~~Carto~~ (complicated to test for the moment and lower priority)
- [x] Timetable management #8650
  - [x] Delete train
  - [x] Duplicate train
  - [x] Multiple selection
  - [x] Searching 
  - [x] Train details
- [ ] ~~Roles and permissions~~
  - [ ] ~~Roles #9588~~
  - [ ] ~~Permissions~~

PR #9227, opened 2024-10-07
front: fix the issue where the 'More details' button stopped working

### LINKED ISSUE #9181: `more details` button doesn't work anymore in the train list
### What happened?

image

.

.

Dev (SNCF)
Recette (SNCF)

Google Chrome

8795495

achrafmohye:
Closes #9181

### DIFF
--- a/front/src/modules/trainschedule/components/Timetable/Timetable.tsx
+++ b/front/src/modules/trainschedule/components/Timetable/Timetable.tsx
@@ -119,6 +119,7 @@ const Timetable = ({
       <div
         className={cx('scenario-timetable-trains', {
           expanded: conflictsListExpanded,
+          'with-details': showTrainDetails,
         })}
       >
         <TimetableToolbar


woshilapin (commented):
If this was a regression, maybe we need a test to avoid another regression later?

theocrsb (approved):
LGTM and tested

SarahBellaha (approved):
LGTM

clarani (approved):
LGTM :white_check_mark:

clarani:
@woshilapin I think the better way to test this is by an e2e test. The e2e test are planned and developed by @Maymanaf right now, so I think we won't add the test in this PR. But we can do it later, with @Maymanaf :)

Maymanaf:
> @woshilapin I think the better way to test this is by an e2e test. The e2e test are planned and developed by @Maymanaf right now, so I think we won't add the test in this PR. But we can do it later, with @Maymanaf :)

The timetable management tests are planned but not prioritized for now : You can check the order #7706

achrafmohye:
> If this was a regression, maybe we need a test to avoid another regression later?

Don't think it's a regression that needs testing, i guess it was rather an oversight

PR #9229, opened 2024-10-07
Remove invalid imports in models fixtures

leovalais:
Some `views` definitions are used for `models` fixtures. This is incorrect, and these are replaced by the schema instead.

There's also an import from `infra_cache`, which will be removed in another PR.


woshilapin (approved):
Nice catch, it's going to help for the split of crates I believe 🚀

PR #9237, opened 2024-10-08
front: enable authorization 

### LINKED ISSUE #8049: Show/hide  STDCM/Operational study application based on user role
From Home page, 
- [ ] wrap STDCM card with RoledBasedComponent and 
- [ ] wrap operational Studies card with RoledBasedComponent

kmer2016:
closes #8049 

> [!WARNING]
> There is currently an authorization bug between the core service and the Editoast service that prevents retrieving the `infraId`. This issue is being worked on and will be fixed soon. Because of this, the `ManageTrainSchedule` and `STDCM` views may have refresh problems.

**New Functionalities:**
- Allows specific parts of the interface to be accessible based on the user's role and the required role for access.
  - [x] When the user is **Railway Undertaking** , they are redirected directly to the **STDCM** page.
  - [x] When the user is **Operational Studies**, they are taken to the homepage, with the **STDCM** application grayed out.
  - [x] When the user is a **Superuser**, they have full access to everything.
  - [x] If the user has no role, they will have no access to any part of the app from the homepage.
  - [x] If a user tries to access an application directly via URL without the proper role, they are redirected to a **403** page. For instance, if an **Operational Studies** user tries to access anything other than **STDCM**, or if a user with no role tries to access any page, they will be blocked.
  - [x] The button to access the STDCM debug mode is now restricted to users with the **Superuser** role.
  - [x] The creation of a project, study, or scenario is limited to users with the **OpsWrite** role. (Both OpsWrite and OpsRead can open the Ops application. OpsWrite can create, OpsRead can list the project/studies/scenarios)

**Technical Updates:**
- User roles are now fetched and stored in the Redux store.
- Added the `REQUIRED_USER_ROLES_FOR` object, which describes the required roles for accessing various views, features and those associated with different user profiles.
- Improved the `RoleBasedComponent` wrapper by adding the `disabledIfUnauthorized` property, which grays out the wrapped component instead of hiding it when the user lacks proper authorization.
- [use the same componenet to handle project,study and scenario creation.](https://github.com/OpenRailAssociation/osrd/pull/9237/commits/3590532fb1ff74afd95bba0ba4105d27f49d27d0)

**How to Test:**
1. **Enable Authorization on Editoast:**
   Add the environment variable `EDITOAST_DISABLE_AUTHORIZATION: "false"` to the `docker-compose.yml` file, just below `OSRD_MQ_URL`.

or if you run editoast individually
`EDITOAST_DISABLE_AUTHORIZATION=false cargo run runserver`

2. **Rebuild the Editoast Docker Image:**
   Rebuild the Editoast Docker image and restart it.

3. **Apply the SQL Scripts:**
   Depending on the user profile and the feature access you want to test, apply the following SQL scripts:

Starting editoast with `EDITOAST_DISABLE_AUTHORIZATION=false`, you should have a subject `1` created in `authn_subject` table

#### Railway Undertaking profil
```sql
   DELETE FROM authz_role WHERE subject = 1;
   INSERT INTO authz_role (id, subject, role) VALUES 
   (1, 1, 'Stdcm'),
   (2, 1, 'RollingStockCollectionRead'),
   (3, 1, 'WorkScheduleRead'),
   (4, 1, 'InfraRead'),
   (5, 1, 'DocumentRead'),
   (6, 1, 'MapRead'),
   (7, 1, 'TimetableRead');
```

#### Operational Studies profil
```sql
DELETE FROM authz_role WHERE subject = 1;
INSERT INTO authz_role (id, subject, role) VALUES 
(1, 1, 'OpsRead'),
(2, 1, 'OpsWrite'),
(3, 1, 'InfraRead'),
(4, 1, 'InfraWrite'),
(5, 1, 'RollingStockCollectionRead'),
(6, 1, 'RollingStockCollectionWrite'),
(7, 1, 'WorkScheduleRead'),
(8, 1, 'TimetableRead'),
(9, 1, 'TimetableWrite'),
(10, 1, 'DocumentRead'),
(11, 1, 'DocumentWrite'),
(12, 1, 'MapRead');
```

#### Superuser
```sql
DELETE FROM authz_role WHERE subject = 1;
INSERT INTO authz_role (id, subject, role) VALUES 
(1, 1, 'Superuser');
```

#### No Authorization to create projet/study/scenario
```sql
DELETE FROM authz_role WHERE subject = 1;
INSERT INTO authz_role (id, subject, role) VALUES 
(1, 1, 'OpsRead'),
(3, 1, 'InfraRead'),
(5, 1, 'RollingStockCollectionRead'),
(6, 1, 'RollingStockCollectionWrite'),
(7, 1, 'WorkScheduleRead'),
(8, 1, 'TimetableRead'),
(9, 1, 'TimetableWrite'),
(10, 1, 'DocumentRead'),
(11, 1, 'DocumentWrite'),
(12, 1, 'MapRead');
```


RomainValls (commented):
Tested every SQL scripts for the permissions, it all seems correct.

leovalais:
> There is currently an authorization bug between the core service and the Editoast service that prevents retrieving the infraId. This issue is being worked on and will be fixed soon. Because of this, the ManageTrainSchedule and STDCM views may have refresh problems.

Was fixed in https://github.com/OpenRailAssociation/osrd/pull/9278, thanks for reporting it :)

front/src/utils/hooks/OsrdAuth.ts
```diff
@@ -1,35 +1,53 @@
...
   useEffect(() => {
     if (!isUserLogged && !isAuthenticateLoading) {
       login();
     }
   }, [isUserLogged]);
 
+  useEffect(() => {
+    if (isUserLogged) {
+      dispatch(setUserRoles({ userRoles: data?.builtin }));
+      setIsUserRolesFetched(true);
```
emersion: I don't think we need a separate state for this? Can't we check if `data` is null/undefined instead?

front/src/common/authorization/components/RoleBasedContent.tsx
```diff
@@ -1,17 +1,45 @@
...
+  children,
+  disableIfUnauthorized = false,
+}: RoleBasedContentProps) => {
   const isRoleAllowed = useUserRoleCheck(requiredRoles);
 
   if (isRoleAllowed) {
     return children;
   }
 
+  if (disableIfUnauthorized) {
```
emersion: Hm, that seems pretty specific to the cards on the homepage. Maybe it would be better to use `useUserRoleCheck()` inside `Home` and set a CSS class there on individual cards?

front/src/common/authorization/components/InitialRedirect.tsx
```diff
@@ -0,0 +1,13 @@
+import { useSelector } from 'react-redux';
+import { Navigate } from 'react-router-dom';
+
+import Home from 'main/home';
+import { getIsStdcmProfile } from 'reducers/user/userSelectors';
+
+const InitialRedirect = () => {
+  const isStdcmProfile = useSelector(getIsStdcmProfile);
```
emersion: This checks if the user has the `Stdcm` role. Wouldn't a super-user have all of the roles? In that case, I don't think they should be redirected?

Maybe we should instead check if the user only has the role to access STDCM, and has no permission to access any other tool?

front/src/reducers/user/index.ts
```diff
@@ -39,12 +44,25 @@ export const userSlice = createSlice({
     logoutSuccess() {
       return userInitialState;
     },
+    setUserRoles(state, action: PayloadAction<{ userRoles: BuiltinRole[] | undefined }>) {
+      state.userRoles = action.payload.userRoles ?? [];
+    },
     updateUserPreferences(state, action: PayloadAction<{ safeWord: string }>) {
       state.userPreferences = action.payload;
     },
+    switchStdcmV2Activated(state) {
```
emersion: This seems unused?

front/src/common/authorization/components/RoleBasedContent.tsx
emersion: It's also used in a few other places, but I'm a bit worried about having this kind of generic logic for pretty different use-cases…

Would suggest this

```tsx
<div
  className="col-6 col-md-5 col-lg-4 col-xl-3"
  {...(operationalStudiesAllowed && { 'aria-disabled': true })}
>
```

```css
.cardscontainer [aria-disabled] {
  pointer-events: none;
  opacity: 0.5;
}
```

and

```tsx
<div role="button" {...(newProjectStudyScenarioAllowed && { disabled: true })}>
```

front/src/reducers/user/index.ts
clarani: It has been removed by Sarah in this PR: https://github.com/OpenRailAssociation/osrd/pull/9325

front/src/common/authorization/components/InitialRedirect.tsx
RomainValls: I agree. But I'm not sure it's possible to access stdcm without additional permissions like mapread for example. 
Then, as we'll need those permissions, the mapViewer module would be accessible in Home? Should we redirect the user to stdcm even if he could get access to some of the modules inside Home ?

front/src/common/authorization/roleBaseAccessControl.ts
```diff
@@ -0,0 +1,58 @@
...
+    STDCM: BuiltinRole[];
+    OPERATIONAL_STUDIES: BuiltinRole[];
+  };
+  FEATURES: {
+    ACCESS_STDCM_DEBUG: BuiltinRole[];
+    CREATE_NEW_PROJECT_STUDY_SCENARIO: BuiltinRole[];
+  };
+};
+
+export const REQUIRED_USER_ROLES_FOR: RequiredUserRolesFor = {
```
flomonster, line 21: This feels overkill since we only want to check if the user is:
- Superuser
- STDCM
- Operational study
- Guest

> [!NOTE]
> The role system is temporary and will be drop soon.

This comment is not blocking to merge this PR.

docker-compose.yml
```diff
@@ -132,6 +132,7 @@ services:
       TELEMETRY_KIND: "opentelemetry"
       TELEMETRY_ENDPOINT: "http://jaeger:4317"
       OSRD_MQ_URL: "amqp://osrd:password@osrd-rabbitmq:5672/%2f"
+      EDITOAST_DISABLE_AUTHORIZATION: "true"
```
flomonster: The default value is already true.
```suggestion
```

front/src/common/AddNewCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const { openModal } = useModal();
+
+  const newProjectStudyScenarioAllowed = useUserRoleCheck(
+    REQUIRED_USER_ROLES_FOR.FEATURES.CREATE_NEW_PROJECT_STUDY_SCENARIO
+  );
+
+  return (
+    <div
+      data-testid={testId}
+      className={`${className} ${!newProjectStudyScenarioAllowed ? 'aria-disabled' : ''}`}
```
emersion: `aria-disabled` is not a class, it's an attribute. It's an accessibility feature to mark an element as disabled, useful for screen readers and similar. See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled

front/src/common/authorization/components/InitialRedirect.tsx
emersion: What about this: if the user has access to exactly all of the roles required for STDCM and nothing more, redirect to STDCM. If they have more permissions (e.g. operational studies), or less permissions (e.g. only map read), don't redirect.

front/src/reducers/user/index.ts
```diff
@@ -1,21 +1,26 @@
...
+import type { BuiltinRole } from 'common/api/osrdEditoastApi';
 
 export interface UserState {
   isLogged: boolean;
   loginError?: ApiError;
   username: string;
   userPreferences: { safeWord: string };
+  userRoles: BuiltinRole[];
   account: Record<string, string>;
+  stdcmV2Activated?: boolean;
```
emersion: Nit: this remnant still needs to be taken care of :P

front/src/common/authorization/components/RoleBasedContent.tsx
emersion: Looks good!

front/src/common/authorization/components/RoleBasedContent.tsx
```diff
@@ -1,17 +1,45 @@
+import React from 'react';
+
+import type { BuiltinRole } from 'common/api/osrdEditoastApi';
+
 import useUserRoleCheck from '../hooks/useUserRoleCheck';
 
 export type RoleBasedContentProps = {
-  requiredRoles: string[];
+  requiredRoles: BuiltinRole[];
+  disableIfUnauthorized?: boolean;
```
emersion: Now this field is unused so can be dropped

front/src/reducers/user/index.ts
```diff
@@ -39,12 +44,16 @@ export const userSlice = createSlice({
     logoutSuccess() {
       return userInitialState;
     },
+    setUserRoles(state, action: PayloadAction<{ userRoles: BuiltinRole[] | undefined }>) {
+      state.userRoles = action.payload.userRoles ?? [];
+    },
```
clarani: You can simplify your action like this
```suggestion
    setUserRoles(state, action: PayloadAction<BuiltinRole[] | undefined>) {
      state.userRoles = action.payload || [];
    },
```

front/src/reducers/user/userReducer.spec.ts
```diff
@@ -63,6 +64,17 @@ describe('userReducer', () => {
     expect(userState).toEqual(userInitialState);
   });
 
+  it('should handle setUserRoles', () => {
+    const store = createStore(userInitialState);
+    const action = setUserRoles({ userRoles: ['Stdcm'] });
+    store.dispatch(action);
```
clarani: You can directly dispatch
```suggestion
    store.dispatch(setUserRoles(['Stdcm']));
```

front/src/applications/stdcm/components/StdcmHeader.tsx
```diff
@@ -18,16 +21,18 @@ const StdcmHeader = ({ isDebugMode, onDebugModeToggle }: StdcmHeaderProps) => {
...
-          type="button"
-          aria-label="stdcm-debug"
-          className={cx({ 'debug-on': isDebugMode, 'debug-off': !isDebugMode })}
-          onClick={() => onDebugModeToggle(!isDebugMode)}
-        >
-          <Bug />
-        </button>
-      </div>
+      <RoleBasedContent requiredRoles={REQUIRED_USER_ROLES_FOR.FEATURES.ACCESS_STDCM_DEBUG}>
+        <div className="stdcm-v2-header_debug">
```
clarani: (It may comes from a rebase :) )
```suggestion
        <div className="stdcm-header_debug">
```

front/src/applications/stdcm/components/StdcmHeader.tsx
```diff
@@ -18,12 +22,14 @@ const StdcmHeader = ({ isDebugMode, onDebugModeToggle }: StdcmHeaderProps) => {
...
         </span>
       </div>
+
       <div className="stdcm-header_debug">
         <button
           type="button"
           aria-label="stdcm-debug"
           className={cx({ 'debug-on': isDebugMode, 'debug-off': !isDebugMode })}
           onClick={() => onDebugModeToggle(!isDebugMode)}
+          aria-disabled={!isSuperUser}
```
emersion: I think we can just use the `disabled` attribute here, since this is a `button` element.

front/src/utils/allowedUserRoles.ts
```diff
@@ -0,0 +1,22 @@
+import useUserRoleCheck from 'common/authorization/hooks/useUserRoleCheck';
+import { REQUIRED_USER_ROLES_FOR } from 'common/authorization/roleBaseAccessControl';
+
+export default function useAllowedUserRoles() {
```
emersion: Maybe this function should be defined in `common/authorization/hooks/useAllowedUserRoles.ts`, instead of utils?

front/src/common/AddNewCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+type AddNewCardProps = {
+  translationNamespaces?: string[] | string;
+  testId: string;
+  className: string;
+  modalComponent: ReactNode;
+  legendTranslationKey: string;
+};
+
+const AddNewCard = ({
```
emersion: This is a component specific to operational studies, maybe should be in a different directory? Maybe `src/applications/operationalStudies/components/`?

front/src/reducers/user/userSelectors.ts
```diff
@@ -7,8 +9,35 @@ const makeUserSelector = makeSubSelector<UserState>(getUser);
...
+export const getIsSuperUser = (state: RootState) => getUserRoles(state).includes('Superuser');
+
+const makeUserHasAllRequiredRolesSelector =
+  (requiredRoles: BuiltinRole[]) => (state: RootState) => {
+    const userRoles = getUserRoles(state);
+    return requiredRoles.every((role) => userRoles.includes(role));
+  };
+export const getIsOnlyStdcmProfile = (state: RootState) => {
+  const userRoles = getUserRoles(state);
+  const requiredRolesForStdcm = [
```
emersion: Could we use `REQUIRED_USER_ROLES_FOR` instead?

front/src/common/authorization/components/InitialRedirect.tsx
emersion: This has been implemented now!

emersion (approved):
Thanks a bunch!

flomonster (approved):
Approve editoast part

PR #9244, opened 2024-10-08
editoast: bump async-trait from 0.1.82 to 0.1.83 in /editoast

woshilapin:
⚠️ Dependabot created [a PR](https://github.com/OpenRailAssociation/osrd/pull/8980) but ended-up in a weird state, so I just manually recreated it.

Bumps [async-trait](https://github.com/dtolnay/async-trait) from 0.1.82 to 0.1.83.
- [Release notes](https://github.com/dtolnay/async-trait/releases)
- [Commits](https://github.com/dtolnay/async-trait/compare/0.1.82...0.1.83)

---
updated-dependencies:
- dependency-name: async-trait dependency-type: direct:production update-type: version-update:semver-patch ...

### DIFF
--- a/editoast/Cargo.lock
+++ b/editoast/Cargo.lock
@@ -410,9 +410,9 @@ checksum = "8b75356056920673b02621b35afd0f7dda9306d03c79a30f5c56c44cf256e3de"
 
 [[package]]
 name = "async-trait"
-version = "0.1.82"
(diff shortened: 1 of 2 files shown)


woshilapin:
Dependabot finally did seem to recover, I still don't understand how all of that works.

PR #9249, opened 2024-10-08
front: fix inconsistent translations and variable names in operationa…

### LINKED ISSUE #7270: Update the labels after the study creation FR and ENG
### What happened?

The field label in the form of study creation is not the same when the study is created 
image
image

The same label is displayed in both the form and the study details
In the study détails we should have : 

- **Service code** instead of **GEREMI CODE**
- **Business code** instead of **AFFAIR CODE**
- **Real ending** instead of **REAL END**

1. Go to an Project
2. Click on "Create a study"
3. Set the fields Name , Service code, Business code and Real ending
4. Click on create the study

- **Browser**: Chrome
- **OS**: W11
- **Env**: rec

1482e8c

Synar:
Close  #7270

Geremi code becomes service code in French and English translations
Affair code becomes business code in French and English translations
Any "ending date" become "end date" in variable names and English translations


louisgreiner:
Temporary fix until we do 
- https://github.com/osrd-project/osrd-confidential/issues/694

Thanks !

front/public/locales/fr/operationalStudies/project.json
```diff
@@ -71,6 +69,8 @@
   "studiesCount_one": "1 étude",
   "studiesCount_other": "{{count}} études",
   "studiesCount_zero": "Aucune étude",
+  "studyBusinessCode": "Code business",
```
emersion: :sadface:

Synar: Could you be more specific? :eyes:

emersion: 🇫🇷 https://www.academie-francaise.fr/business 🥐 https://www.cnrtl.fr/definition/business 🥖

Yohh: mmmh ok... every new day brings its lesson, but is it really an issue?
image

Yohh: anyway, we can replace with "exploitation"

emersion: I don't think it's appropriate to use informal wording here. Also the meaning of this definition doesn't quite fit the bill IMHO.

Synar: Hmmm, exploitation or affaire (which was used to refer to this code half the time before) would both work, but the question is also what our end user expect. 

I believe @louisgreiner had a meeting with them and he told me to use business, maybe he can comment on whether one of these options would also be fine?

louisgreiner: Since this component is actually not used AT ALL by the users, and will be sort of refacto [here](https://github.com/osrd-project/osrd-confidential/issues/694). 

They indeed told me that business or anything else is not relevant since not used. I then propose this bikeshed by using "Affaire" 😸

Synar: Hahaha ok, done

emersion (approved):
Nice

Yohh (approved):
LGTM

SharglutDev (approved):
Lgtm. Congratz for first PR :)

You just need to update the e2e tests with these new trads.

emersion:
Please squash these two commits together before merging :) 

(We're trying to keep intermediate commits fully buildable and testable.)

Synar:
Hmmm, the integration tests failed before even getting to playwright. I could not understand why, ran them again without changing anything, and now they pass. Assuming it was a failure of the test env to properly initialize and not a stochastic test failing, is this kind of failure of the test env expected?

Synar:
I am thoroughly confused by the stochastic behavior of the integration tests

PR #9250, opened 2024-10-08
osrdyne: hide sensitive information in logs

woshilapin:
I thought about using `derivative` to just customize the implementation of `Debug` with ignoring `amqp_uri`, but then, bringing a new dependency only for this was maybe too much.

### DIFF
--- a/osrdyne/src/config.rs
+++ b/osrdyne/src/config.rs
@@ -20,7 +20,7 @@ pub enum WorkerDriverConfig {
     ProcessComposeDriver(PCDriverOptions),
 }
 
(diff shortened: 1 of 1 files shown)


ghost (approved):

Khoyo (changes requested):
I don't think this will work after this. https://github.com/OpenRailAssociation/osrd/pull/9261

ghost (commented):
Indeed, my bad : I reviewed without thinking of the changes I've made on the other PR.

flomonster:
Can we close this PR?

PR #9258, opened 2024-10-08
front: minor upgrade for several dependencies

### LINKED ISSUE #9257: front: minor upgrade for several dependencies
upgrade i18n from 23.5.1 to 23.8.2

Yohh:
close #9257


emersion:
Hm, it seems like this goes from 23.15.1 to 23.15.2?

Could we group this with other bugfix releases maybe?

PR #9260, opened 2024-10-08
editoast: fix rolling stock deletion 

### LINKED ISSUE #9180: front: Can not delete a rolling stock
### What happened?

image

Should be deleted successfully 

1. Go to Rolling stock editor
2. Delete an unlocked rolling stock

Recette (SNCF)

Google Chrome

1482e8c

Synar:
Close #9180

Delete rolling_stock_usage.rs (which search train schedules by rolling stock ids, which are no longer present in the model, hence why the deletion process which exploited it failed) in favor of using schedules_from_rolling_stock.rs (which search train schedules by rolling stock names)


woshilapin (approved):
Good catch removing this redundant structures and function. Thank you. One small thing, can you edit the title of the PR so it's not about the name of the branch... but about the content of the PR (prefixing with `editoast` in this case).

editoast/src/views/rolling_stock.rs
Sh099078, line 1: It might be a good timing to add a test that checks that trying to delete a used rolling stock triggers an error ?

Sh099078, line 1: We are losing the train schedules related to the rolling stock in `RollingStockError::IsUsed`. Was it a valuable information ?

Sh099078 (approved):
Nice one ! I left two comments, otherwise it looks good to me :)

editoast/src/views/rolling_stock.rs
Synar: Good idea. The issue #8078 suggests that there is a possibility that it should instead simply trigger a warning to the user (I'll ask), but in all case it should trigger *something* that we should be able to test.

Sh099078: Agreed, maybe deleting used rolling stocks should be allowed. It would be interesting to have @leovalais opinion on the subject.

As for the test, you could simply check that calling the delete endpoint on a rolling stock that is used somewhere succeeds and that the rolling stock has been effectively deleted. I am not convinced it deserves its own test if the rolling stock usage check gets removed from `rolling_stock::delete`: it would follow the same logic as deleting a rolling stock that is not used anywhere. However the latter is not tested neither: maybe that is the test that should be written ?

Synar: I have added tests that check the deletion works properly when unused and fails when used. The behavior when used should probably be changed to require confirmation from the user, but I think this is outside the scope of this PR. So the tests simply assert that the current behavior is correctly implemented.

editoast/src/views/rolling_stock.rs
```diff
@@ -1314,4 +1329,78 @@ pub mod tests {
...
+        // WHEN
+        app.fetch(request).assert_status(StatusCode::CONFLICT);
+
+        // THEN
+        let rolling_stock_exists =
+            RollingStockModel::exists(&mut db_pool.get_ok(), fast_rolling_stock.id)
+                .await
+                .expect("Failed to check if rolling stock exists");
+
+        assert_eq!(rolling_stock_exists, true);
```
Sh099078: `assert!(rolling_stock_exists)` should work there.

editoast/src/views/rolling_stock.rs
```diff
@@ -1314,4 +1329,78 @@ pub mod tests {
...
+        simple_train_schedule_changeset(timetable.id)
+            .rolling_stock_name(fast_rolling_stock.name.clone())
+            .create(&mut db_pool.get_ok())
+            .await
+            .expect("Failed to create train schedule");
+
+        let request = app.delete(format!("/rolling_stock/{}", fast_rolling_stock.id).as_str());
+
+        // WHEN
+        app.fetch(request).assert_status(StatusCode::CONFLICT);
```
Sh099078: `assert_status(StatusCode::CONFLICT)` is one of the two main things that this test intends to check with the fact that the rolling stock still exists afterwards. If the test is split into GIVEN... WHEN... THEN... parts, maybe it should go in the THEN... part ? If it is in the WHEN... part, I understand it as if the only behaviour this test is about is the persistence of the rolling stock afterwards and that the http response code is just part of the setup.

editoast/src/views/rolling_stock.rs
```diff
@@ -1314,4 +1329,78 @@ pub mod tests {
...
+        // WHEN
+        app.fetch(request).assert_status(StatusCode::NO_CONTENT);
+
+        // THEN
+        let rolling_stock_exists =
+            RollingStockModel::exists(&mut db_pool.get_ok(), fast_rolling_stock.id)
+                .await
+                .expect("Failed to check if rolling stock exists");
+
+        assert_eq!(rolling_stock_exists, false);
```
Sh099078: `assert!(krolling_stock_exists)` ?

editoast/src/views/rolling_stock.rs
```diff
@@ -1314,4 +1329,78 @@ pub mod tests {
...
+        let db_pool = app.db_pool();
+
+        let rs_name = "fast_rolling_stock_name";
+        let fast_rolling_stock = create_fast_rolling_stock(&mut db_pool.get_ok(), rs_name).await;
+        assert!(!fast_rolling_stock.locked);
+
+        let request = app.delete(format!("/rolling_stock/{}", fast_rolling_stock.id).as_str());
+
+        // WHEN
+        app.fetch(request).assert_status(StatusCode::NO_CONTENT);
```
Sh099078: Same comment that I did on `delete_unlocked_used_rolling_stock_fails()`: shouldn't the `assert_status` be in the THEN... part of the test or the GIVEN... WHEN... THEN... comments be removed ?

Sh099078 (commented):
Nice work, I left a few comments on the new tests. Thank you for fixing my broken test and typos :sweat_smile:

editoast/src/views/rolling_stock.rs
Synar: What if I follow what is done in the test above
```
// WHEN
        let response: InternalError = app
            .fetch(request)
            .assert_status(StatusCode::CONFLICT)
            .json_into();

        // THEN
        assert_eq!(response.error_type, "editoast:rollingstocks:IsLocked");)
```
and thus check the status in the when and the actual error in the then? 
Or should I change the already written test above too to only check the status later?

editoast/src/views/rolling_stock.rs
Synar: Actually thinking about it, there is already a force flag in the method that serves to bypass this check, so probably the only the front behavior should change to be able to use this flag after user confirmation, and this test should be the correct one.

editoast/src/views/rolling_stock.rs
Sh099078: This looks fine to me ! Feel free to resolve that conversation with whichever option you think is best: I might have been a little too nitpicky on that one ^^

Synar: Ok so I opted to change all the tests in the file to consistently put the assert_status in the THEN, hopefully this is fine

editoast/src/views/rolling_stock.rs
Synar: Done

editoast/src/views/rolling_stock.rs
Synar: Done

editoast/src/views/rolling_stock.rs
Synar: Sure

editoast/src/views/rolling_stock.rs
Synar: Talked with Florian about it and it should be fine

emersion (approved):
LGTM for the frontend side (not tested)!

Math-R (approved):
lgtm =)

PR #9264, opened 2024-10-09
front: import test functions from vitest in every test file

### LINKED ISSUE #9263: front: import every missing test functions
in almost every test files, `it`, `describe` and `expect` functions are not imported

Yohh:
close #9263


emersion:
While at it can we remove the vitest `globals: true` option?

Yohh:
> While at it can we remove the vitest `globals: true` option?

done, this showed me that I missed some functions

kmer2016 (approved):
Thank you !

PR #7515, opened 2024-05-26
docker: switch to valkey

emersion:
Recently-ish redis has switched to another license which isn't open-source. An open-source fork named valkey has been established under the Linux Foundation umbrella:
https://www.linuxfoundation.org/press/linux-foundation-launches-open-source-valkey-community

See also this LWN article:
https://lwn.net/Articles/966631/

valkey is a drop-in replacement for redis. Switch to valkey to avoid depending on proprietary software.

(Note, I've only done some basic testing to make sure this works as expected, by playing with the UI a bit.)

PR #9274, opened 2024-10-09
Rename redis to valkey

### LINKED ISSUE #8249: Rename redis by valkey everwhere
### Description and goal

We've switched from redis to valkey, but we keep calling it redis. We should stop doing that

### Acceptance criteria

```
$ git grep -i redis
$
```

- [ ] Redis is replaced by valkey everywhere

EthanPERRUZZA:
Fixes #8249.

### DIFF
--- a/editoast/src/map/mod.rs
+++ b/editoast/src/map/mod.rs
@@ -11,28 +11,28 @@ pub use self::layer_cache::get_layer_cache_prefix;
 pub use self::layer_cache::get_view_cache_prefix;
 pub use self::layer_cache::Tile;
 use crate::error::Result;
-use crate::RedisConnection;
+use crate::ValkeyConnection;
 
 /// Invalidates layer cache for a specific infra and view if provided
 ///
 /// # Arguments
 ///
-/// * `redis_pool` - Pool to use to connect to the redis
+/// * `valkey` - Pool to use to connect to the valkey
 /// * `infra_id` - Infra on which the layer must be invalidated
 /// * `layer_name` - Layer to invalidate
 ///
 /// Returns the number of deleted keys
 async fn invalidate_full_layer_cache(
-    redis: &mut RedisConnection,
+    valkey: &mut ValkeyConnection,
     infra_id: i64,
     layer_name: &str,
 ) -> Result<u64> {
     let prefix: String = get_layer_cache_prefix(layer_name, infra_id);
-    let matching_keys: Vec<String> = redis.keys(format!("{prefix}.*")).await?;
+    let matching_keys: Vec<String> = valkey.keys(format!("{prefix}.*")).await?;
     if matching_keys.is_empty() {
         return Ok(0);
     }
-    let number_of_deleted_keys = redis.del(matching_keys).await?;
+    let number_of_deleted_keys = valkey.del(matching_keys).await?;
     Ok(number_of_deleted_keys)
 }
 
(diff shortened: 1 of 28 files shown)


emersion:
Nit: we're not using the "chore" prefixes from conventional commits. Instead we use a prefix to indicate the component, e.g. "editoast".

Other nit: "rename XXX by YYY" doesn't sound right to me, maybe "rename XXX to YYY" would be better?

EthanPERRUZZA:
> Nit: we're not using the "chore" prefixes from conventional commits. Instead we use a prefix to indicate the component, e.g. "editoast".

Should I put editoast: front: docker: or editoast, front, docker: or maybe all: or multiple: ?

emersion:
"editoast, front, docker:" is common. "all:" is something I'd do (and e.g. Go does) but I haven't seen it in OSRD yet.

EthanPERRUZZA:
> "editoast, front, docker:" is common. "all:" is something I'd do (and e.g. Go does) but I haven't seen it in OSRD yet.

I've look through the commit and the first one is the one used ;)

Khoyo (commented):
Nice! I'm not approving because we'll need to update the charts at the same time.

Nit: it really is rename to: https://english.stackexchange.com/questions/100296/which-preposition-to-use-with-rename

emersion:
Gentle ping about this:

> Other nit: "rename XXX by YYY" doesn't sound right to me, maybe "rename XXX to YYY" would be better?

editoast/src/map/mod.rs
```diff
@@ -11,47 +11,47 @@ pub use self::layer_cache::get_layer_cache_prefix;
...
 }
 
 /// Invalidates all map layers of a specific infra
 ///
 /// # Arguments
 ///
-/// * `redis_pool` - Pool to use to connect to the redis
-/// * `layers` - Layers to invalidate
+/// * `valkey` - Pool to use to connect to the valkey
+/// * `layers` - Layers to valkey
```
woshilapin: Shouldn't this stay `invalidate`? Why the change?

```suggestion
/// * `layers` - Layers to invalidate
```

editoast/src/views/timetable.rs
```diff
@@ -287,7 +287,7 @@ async fn conflicts(
     }
 
     let db_pool = app_state.db_pool_v2.clone();
-    let redis_client = app_state.redis.clone();
+    let valkey_client: std::sync::Arc<crate::ValkeyClient> = app_state.valkey.clone();
```
woshilapin: Why do you need the type to be explicit now? Can't you keep it like it was before?

woshilapin (approved):
Thanks for the hard work.

flomonster:
@Khoyo Why do you want to rename all occurrences of redis to valkey? We always use redis in production and from editoast's point of view, the service used is redis. The lib name is redis.

> [!NOTE]
> I'm okay with:
> - renaming redis to valkey in the osrd-compose since the image used is valkey
> - using valkey in our CI

emersion:
Would be nice to switch to valkey in production as well.

Khoyo:
> We always use redis in production and from editoast's point of view, the service used is redis. The lib name is redis.

As far as the open source side of the project is concerned, we use an opensource software named valkey and formerly known as redis. I believe from the point of view of editoast, the service used should be the open source one, not a closed source fork.

The valkey crate name is currently unofficially reserved by the redis-rs maintainers, and I believe we *should* switch to it if it becomes necessary.

> using valkey in our CI

I believe that was @emersion first PR :) https://github.com/OpenRailAssociation/osrd/pull/7515

---

Now, as far as production is concerned, we use neither redis nor valkey but Amazon ElastiCache.

Seeing the current move toward supporting valkey in favor of redis at aws (see the elasticache [supported engine versions](https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/SelectEngine.html), and [this blogpost](https://aws.amazon.com/blogs/database/amazon-elasticache-and-amazon-memorydb-announce-support-for-valkey/)), we should probably move there too.

EthanPERRUZZA:
> Gentle ping about this:
> 
> > Other nit: "rename XXX by YYY" doesn't sound right to me, maybe "rename XXX to YYY" would be better?

Sorry, I had completly forgotten, thanks for the ping ;)

flomonster (dismissed):
LGTM.

We should have the osrd-chart (+ deployments) PRs ready.

SharglutDev (approved):

Khoyo (approved):

PR #9276, opened 2024-10-09
front: fix waypoint removal in itinerary pathfinding module

### LINKED ISSUE #9259: Pathfinding: removing a waypoint removes the origin
### What happened?

When finding a path from two operational point through the rocket tool, when I add then remove a waypoint with the "add waypoints" tool, it seems to remove the origin point.

It might be linked to the fact they share the same name (not the same ch code though).

https://github.com/user-attachments/assets/64bb42af-6de0-45d9-94f1-6e87d61d9b39

I expect the origin to ne be removable from this tool

1. Open the train creation interface
2. Search a path
3. Open the "add waypoints" tool
4. Add a waypoint (with the same name as the origin)
5. Remove the waypoint you added
6. Witness

Local

Firefox

3423be7

Synar:
Close #9259

The waypoint removal did not check the ch code, only the uic

I am not sure if the ch code should also be checked in the case of a track : ('track' in step && step.track !== op.track)

### DIFF
--- a/front/src/modules/pathfinding/components/Itinerary/ModalSuggestedVias.tsx
+++ b/front/src/modules/pathfinding/components/Itinerary/ModalSuggestedVias.tsx
@@ -37,7 +37,8 @@ const ModalSuggestedVias = ({ suggestedVias }: ModalSuggestedViasProps) => {
   const removeViaFromPath = (op: SuggestedOP) => {
     const updatedPathSteps = compact(pathSteps).filter(
       (step) =>
-        ('uic' in step && step.uic !== op.uic) || ('track' in step && step.track !== op.track)
+        ('uic' in step && (step.uic !== op.uic || step.ch !== op.ch)) ||
+        ('track' in step && (step.track !== op.track || step.offset !== op.offsetOnTrack))
     );
     dispatch(
       updatePathSteps({


louisgreiner (commented):
It is working well for Waypoints, but for track offset, the behaviour seems to be bugged a bit

https://github.com/user-attachments/assets/9aa0aed0-7509-48d0-b32d-02c29891a8ca

I'm not sure it is the scope of the issue but it would be nice to fix this at the same time

emersion:
> I am not sure if the ch code should also be checked in the case of a track : ('track' in step && step.track !== op.track)

So… It's a bit of a mess. I think `matchPathStepAndOp()` is the closest to what you're looking for, except it also matches `kp` and has a special case for `step.id` which are both not desirable in the general case. There is also `pathStepMatchesOp()` which does mostly the same thing but also a bit differently (and compares the plaintext name in some cases for some reason)…

I'd really like to have a clean `matchPathStepAndOp()` function which just matches `opId`/`uic+ch`/`trigram+ch`/`track+offset`. And then leverage that clean function in the existing weird helpers that we have (keep the weird quirks alongside the call to the clean function, and later investigate why we need these weird quirks and try to find a better way).

emersion:
> for track offset, the behaviour seems to be bugged a bit

FWIW would likely need to match `offsetOnTrack` with `offset` for that case.

Synar:
> > for track offset, the behaviour seems to be bugged a bit
> 
> FWIW would likely need to match `offsetOnTrack` with `offset` for that case.

Indeed that's what I ended up doing

But I agree having a single clean function to compare points would be more maintainable

emersion (approved):
Fixes the original issue and then some, so it's already a great improvement :)

(There is _so much_ cleanup to be done around `SuggestedOP`…)

louisgreiner (approved):

SharglutDev (approved):
Lgtm and tested, thank you. Totally agreed that we should have a more unified way to check step/op correspondance

PR #9278, opened 2024-10-09
Fix Core being unauthorized to load an infra when authentication is enabled

leovalais:
When querying `GET /infra/42/railjson`, Core is unable to set the `x-remote-user` header usually set by the gateway to authenticate the user. When it's missing, editoast rejects the request automatically... and no infra can be loaded anymore.

The problems also appears with `/health` btw.

What's done here:

* Editoast endpoints that do not require the user to have any role can be accessed even if the `x-remote-user` header is missing. Editoast still have to be reached though, so we're either dealing with an authenticated user that went through the gateway, or we're already behind it.
* When Editoast detects that the `x-osrd-core` header, all verification checks are short-circuited.
* Core sends the `x-osrd-core` by default. This can be customized using a CLI parameter that was already there. I'm not sure why it's there since editoast doesn't support authentication tokens. If it's useful, I'll add another option instead. (I'll also rename the existing one if changing its behavior is ok.)
* Gateway: each proxy target can be provided a list of headers to remove. This is a bit redundant with the `forwarded_headers` option. The default configuration rejects `x-osrd-core` so that no one besides Core will be able to make editoast interpret the header ([not even him](https://youtu.be/fQGbXmkSArs?feature=shared&t=20)).

> [!NOTE]
> The `AuthorizerExt` extractor is now incorrectly named. I'll change that in a follow up PR (or just before merging this one if you prefer) in order not to conflict with everyone.

### DIFF
--- a/core/src/main/java/fr/sncf/osrd/api/APIClient.java
+++ b/core/src/main/java/fr/sncf/osrd/api/APIClient.java
@@ -9,11 +9,11 @@ public abstract class APIClient {
 
     final OkHttpClient httpClient;
     final HttpUrl baseUrl;
-    final String authorizationToken;
+    final String authenticationHeader;
 
-    APIClient(String baseUrl, String authorizationToken, OkHttpClient httpClient) {
+    APIClient(String baseUrl, String authenticationHeader, OkHttpClient httpClient) {
         this.baseUrl = HttpUrl.parse(baseUrl);
-        this.authorizationToken = authorizationToken;
+        this.authenticationHeader = authenticationHeader;
         this.httpClient = httpClient;
     }
 
@@ -23,7 +23,7 @@ public abstract class APIClient {
                 .encodedQuery(queryParameters)
                 .build();
         var builder = new Request.Builder().url(url);
-        if (authorizationToken != null) builder = builder.header("Authorization", authorizationToken);
+        if (authenticationHeader != null) builder = builder.header(authenticationHeader, "");
         return builder.build();
     }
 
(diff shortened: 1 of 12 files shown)


Khoyo (commented):
Tentative accept, I'll let another op give it a second look.

ghost (approved):
Checked gateway, lgtm. Note that must be accompanied with a modification of the chart to add it on `templates/gateway/configmap.yml` so the behaviour is consistent in production.

Khoyo (approved):

woshilapin (approved):
Not sure I completely understand the underlying issue and therefore, the solution. But at least, nothing worrying on `editoast`.

leovalais:
@woshilapin 

## The authenticated user workflow, where everything is fine, even before this PR

```mermaid
sequenceDiagram
	front->>gateway:/infra/1
	gateway->>editoast: /infra/1 x-remote-user:foo/bar
	editoast-->>gateway: OK
```

## Before this PR, where Core fails to load an infrastructure
```mermaid
sequenceDiagram
	participant gateway
	participant editoast
	participant core
	core->>editoast: /infra/1/railjson
	editoast-->>core: unauthorized: missing x-remote-user
```

And Core making up a fake user to circumvent the problem isn't a great solution imo.

## After this PR
```mermaid
sequenceDiagram
	participant gateway
	participant editoast
	participant core
	core->>editoast: /infra/1/railjson x-osrd-core:true
	editoast-->>core: OK
```

## The header is also systematically removed from incoming requests in the gateway

```mermaid
sequenceDiagram
	actor WannabeHacker
	participant gateway
	participant editoast
	participant core
	WannabeHacker->>gateway: /infra/1/railjson x-osrd-core:true
	gateway->>editoast: /infra/1/railjson x-remote-user:hacker/man
	editoast-->>gateway: nope
	gateway-->>WannabeHacker: get rekt
```

This behavior is not strictly necessary since the `x-remote-user` header takes precedence over `x-osrd-core` when editoast parses the headers. But this additional layer can't hurt so that if we mess up this part in editoast one day, still no one will be able to maliciously pretend they're Core.

> [!NOTE]
> This bug doesn't occur when the authorization is disabled (by default currently). That's why it went a bit under the radar.

This problem also occurred with the health check because `docker-compose` cannot provide the `x-remote-user` header when querying `GET /health`.

Issue #9156, opened 2024-10-02
front: stdcm: the condition where the "get the simulation" button is displayed is overly restrictive


eckter:
### What happened?

On the stdcm page, we can only send requests when the button is enabled. It's very often not enabled.

More specifically, changing parameters in the "debug" fields is not enough to consider that the inputs have changed.

Generally speaking, being entirely unable to send the same request more than once makes debugging a lot more tedious.

### What did you expect to happen?

It should be easier to enable this button. At the very least, changing the debug inputs should enable it. *Ideally*, it would be amazing if we had a debug button to make it always enabled (though that would be a new feature and not just a bugfix)

### How can we reproduce it (as minimally and precisely as possible)?

1. Go to the stdcm page
2. Send any request
3. Change the debug fields
4. There's no button to send a new request

### On which environments the bug occurs?

Local

### On which browser the bug occurs?

Firefox

### OSRD version (top right corner `Account` button > `Informations`)

efce7b7a4bb0342c8f3ba1d05eb64326c893c013

PR #9284, opened 2024-10-10
front: stdcm: various fixes

SarahBellaha:
This PR addresses several issues:
- The viewport of the map is now updated every time the `pathProperties` change.
- The launch button is always enabled in debug mode. ([#9156](https://github.com/OpenRailAssociation/osrd/issues/9156))
- Indices no longer disappear from the via list when there is a null `pathStep`.


Uriel-Sautron (approved):
Tested, thank you !!!

clarani (approved):
LGTM :white_check_mark:

PR #9289, opened 2024-10-10
front: unify PathStep and SuggestedOP matching

emersion:
_See individual commits._

Ref https://github.com/OpenRailAssociation/osrd/pull/9276#issuecomment-2404612290

Depends on https://github.com/OpenRailAssociation/osrd/pull/9276

### DIFF
--- a/front/src/modules/pathfinding/utils.ts
+++ b/front/src/modules/pathfinding/utils.ts
@@ -38,17 +38,13 @@ export const matchPathStepAndOp = (step: PathStep, op: SuggestedOP) => {
   if ('operational_point' in step) {
     return step.operational_point === op.opId;
   }
-  // We match the kp in case two OPs have the same uic+ch (can happen when the
-  // infra is imported)
   if ('uic' in step) {
-    return step.uic === op.uic && step.ch === op.ch && step.kp === op.kp;
+    return step.uic === op.uic && step.ch === op.ch;
   }
   if ('trigram' in step) {
-    return step.trigram === op.trigram && step.ch === op.ch && step.kp === op.kp;
+    return step.trigram === op.trigram && step.ch === op.ch;
   }
-  // TODO: we abuse the PathStep.id field here, the backend also sets it to an
-  // ID which has nothing to do with OPs
-  return step.id === op.opId;
+  return step.track === op.track && step.offset === op.offsetOnTrack;
 };
 
 export const getPathfindingQuery = ({
--- a/front/src/modules/pathfinding/utils.ts
+++ b/front/src/modules/pathfinding/utils.ts
@@ -34,7 +34,10 @@ export const formatSuggestedOperationalPoints = (
     coordinates: getPointCoordinates(geometry, pathLength, op.position),
   }));
 
-export const matchPathStepAndOp = (step: PathStep, op: SuggestedOP) => {
+export const matchPathStepAndOp = (
+  step: PathStep,
+  op: Pick<SuggestedOP, 'opId' | 'uic' | 'ch' | 'trigram' | 'track' | 'offsetOnTrack'>
+) => {
   if ('operational_point' in step) {
     return step.operational_point === op.opId;
   }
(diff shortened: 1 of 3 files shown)


front/src/modules/pathfinding/utils.ts
Synar, line 1: The difference in behavior between matchPathStepAndOp and pathStepMatchesOp does not seem obvious to me from their name.  Is there a way to clarify this?

front/src/modules/pathfinding/utils.ts
```diff
@@ -34,21 +34,20 @@ export const formatSuggestedOperationalPoints = (
     coordinates: getPointCoordinates(geometry, pathLength, op.position),
   }));
 
-export const matchPathStepAndOp = (step: PathStep, op: SuggestedOP) => {
+export const matchPathStepAndOp = (
+  step: PathStep,
+  op: Pick<SuggestedOP, 'opId' | 'uic' | 'ch' | 'trigram' | 'track' | 'offsetOnTrack'>
+) => {
   if ('operational_point' in step) {
     return step.operational_point === op.opId;
```
Synar, line 42: You added a todo comment for the other 2 functions when matching based on id, should it be added to matchPathStepAndOp too?

front/src/modules/timesStops/helpers/utils.ts
Synar, line 1: For pathStepMatchesOp withKp is an optional argument, but for matchPathStep(...) the same condition is split in 2 functions. Is there a reason for this difference?

Synar (commented):
LGTM overall, but I left a few naive comments

front/src/modules/pathfinding/utils.ts
```diff
@@ -155,15 +158,22 @@ export const upsertPathStepsInOPs = (ops: SuggestedOP[], pathSteps: PathStep[]):
...
   pathStep: PathStep,
-  op: Pick<SuggestedOP, 'uic' | 'ch' | 'kp' | 'name' | 'opId'>,
+  op: Pick<
+    SuggestedOP,
+    'opId' | 'uic' | 'ch' | 'trigram' | 'track' | 'offsetOnTrack' | 'name' | 'kp'
+  >,
   withKP = false
-) =>
-  ('uic' in pathStep &&
-    'ch' in pathStep &&
```
emersion, line 166: Note to self: `'ch' in pathStep` is always true, that's why I removed it.

front/src/modules/pathfinding/utils.ts
emersion: Yeah, it's not great at all, OTOH I don't really have better naming ideas, other than understand why the differences exist, whether it makes sense that they exist, and whether we could do it in a better way…

If I really had to find a better name, maybe `matchPathStepAndOpWithKPOrNameButOnlyIfUICAndWithIDWorkaround`?

front/src/modules/pathfinding/utils.ts
emersion: There are two different `opId`-based checks:

- `pathStep.id === op.opId`: here we abuse `PathItemLocation.id` (which identifies a waypoint, not an OP). Path item location IDs and OP IDs are two different namespaces.
- `step.operational_point === op.opId`: a `PathItemLocation` variant which describes the waypoint with an OP ID, instead of a UIC+ch, or trigram+ch, etc.

So a `PathItemLocation` could look like this:

```ts
{
    id: 'waypoint42',
    operational_point: 'op69'
}
```

Does this make sense?

front/src/modules/timesStops/helpers/utils.ts
emersion: I'd really like to just get rid of this argument is the main reason I think. I'd prefer for it to not spread to `matchPathStepAndOp()`, especially when it's only really needed for one specific use-case of the function.

Synar: Hmmm, what about always matching with kp? For the waypoint removal it should also work, shouldn't it? Are there any places where we actually want to match points with different kp?

front/src/modules/pathfinding/utils.ts
Synar: Haha, maybe we should ask a PO?

front/src/modules/pathfinding/utils.ts
Synar: Sure, thanks!

front/src/modules/timesStops/helpers/utils.ts
emersion: I try to not mix up refactorings with change of behavior when possible. Makes it more difficult to bisect or understand what's up when there is breakage. This PR is purely a refactoring.

I don't think KP is a particularly good way to match - `positionOnPath` would be better (ref https://github.com/OpenRailAssociation/osrd/pull/8555#discussion_r1735975387). Or even better, when building the `SuggestedOP` list we can look at the `PathStep`s and record once and for all the matching `PathStep.id`.

front/src/modules/pathfinding/utils.ts
emersion: Not sure… Maybe a combination of looking at git-log/git-blame, asking original authors, trying out paths with loops…

Anyways, sounds out-of-scope for a refactoring IMHO. Incremental steps!

front/src/modules/timesStops/helpers/utils.ts
Synar: Ok sure!

front/src/modules/pathfinding/utils.ts
Synar: Fine with me!

Synar (approved):

emersion:
Gentle ping @clarani

clarani (approved):

PR #9292, opened 2024-10-10
front: OSRD → NGE node fixes

### LINKED ISSUE #8842: PR without "BV" or "00" ch code breaks NGE
### What happened?

image

Related to [this PR](https://github.com/OpenRailAssociation/osrd/pull/8748), OSRD actually recognizes it, but it is bugged in macroscopic mode

Do not create a node for every other ch

1. Open macroscopic view
2. Create 2 nodes : DPY and MAS
3. Draw trainrun between them
4. Go back to microscopic
5. Go back to macroscopic
6. Notice bug

Local

Firefox

c62c9db

### LINKED ISSUE #8926: Import train schedules from open data creates invalid nodes in NGE
### What happened?

image

it seems that the stations met by the path of imported train schedules from open data breaks NGE

related or potential duplicate of https://github.com/OpenRailAssociation/osrd/issues/8842

_No response_

image

1. import train schedules from open data
2. switch to macro
3. notice

Local

Firefox

staging fbf8e3e

emersion:
_See individual commits._

Closes: https://github.com/OpenRailAssociation/osrd/issues/8842  
Closes: https://github.com/OpenRailAssociation/osrd/issues/8926

### DIFF
--- a/front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
@@ -297,19 +297,6 @@ const importTimetable = async (
     return node;
   };
 
-  const nodesByOpId = new Map<string, NodeDto>();
-  for (const op of searchResults) {
-    const node = createNode({
-      trigram: op.trigram + (op.ch ? `/${op.ch}` : ''),
-      fullName: op.name,
-      positionX: op.geographic.coordinates[0],
-      positionY: op.geographic.coordinates[1],
-    });
-    nodesByOpId.set(op.obj_id, node);
-  }
-
-  convertGeoCoords(nodes);
-
   const DTOLabels: LabelDto[] = [];
   // Create one NGE train run per OSRD train schedule
   let labelId = 0;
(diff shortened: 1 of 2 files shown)


front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
```diff
@@ -384,13 +368,25 @@ const importTimetable = async (
...
     .map((trainSchedule) => {
       // Figure out the node ID for each path item
       const pathNodeIds = trainSchedule.path.map((pathItem) => {
         const op = findOpFromPathItem(pathItem, searchResults);
         if (op) {
-          return nodesByOpId.get(op.obj_id)!.id;
+          let node = nodesByOpId.get(op.obj_id);
+          if (!node) {
+            node = createNode({
+              trigram: op.trigram + (op.ch ? `/${op.ch}` : ''),
```
SharglutDev, line 381: How do we difference ops with a CH with actual value `''` and others who will get also `''` but not for the same reasons ?

SharglutDev (approved):
Lgtm and tested (expect for open data import because of a bug)

front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
emersion: Yeah, this logic won't be able to make the difference. But last time I asked around, `ch == ''` turned out to be a myth - there are no such OPs in the infra in practice (France or otherwise).

(Note, this is old code which just happens to be moved around, it's not new code.)

SharglutDev: Wow the ghost ch code :D Ok fine by me :)

SharglutDev:
Tested also with imports :)

PR #9300, opened 2024-10-10
core: add short slip safety speed to buffer stops

### LINKED ISSUE #9283: core: safety-speed: add safety speed ranges before buffer stops
See https://github.com/OpenRailAssociation/osrd/pull/8993#discussion_r1793733560

The PR was already large enough and the feature freeze is getting close, this will be done in a different branch (hopefully very soon)

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/9283

### DIFF
--- a/core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
+++ b/core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
@@ -16,6 +16,15 @@ import fr.sncf.osrd.utils.units.kilometersPerHour
 import fr.sncf.osrd.utils.units.meters
 
 /**
+ * Simple internal class representing a stop with safety speed. Makes the function logic more
+ * straightforward.
+ */
+private data class SafetySpeedStop(
+    val offset: Offset<TravelledPath>,
+    val isShortSlip: Boolean,
+)
+
+/**
  * Compute safety speed ranges, areas where the train has a lower speed limit because of a scheduled
  * stop. For details, see https://osrd.fr/en/docs/reference/design-docs/timetable/#modifiable-fields
  * (or google "VISA (VItesse Sécuritaire d'Approche)" for resources in French)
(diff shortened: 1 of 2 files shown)


core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
```diff
@@ -55,6 +74,20 @@ fun makeSafetySpeedRanges(
...
+ */
+private fun makeEndOfPathStop(
+    infra: RawSignalingInfra,
+    routes: StaticIdxList<Route>,
+    signalOffsets: List<Offset<TravelledPath>>
+): SafetySpeedStop {
+    val lastRouteExit = infra.getRouteExit(routes.last())
+    val isBufferStop = infra.isBufferStop(lastRouteExit.value)
+    return SafetySpeedStop(signalOffsets.last(), isBufferStop)
+}
```
eckter: Question: do we always want to add safety speeds at the end of the last route, or just if it ends with a buffer stop?

I don't think this was explicitly mentioned anywhere, but I believe it makes sense. We only allocate ressources up to the end of the last route, which means that we always assume it ends with a blocking signal. 

I can change it if we only want that to happen on buffer stop. But I believe there would be an inconsistency somewhere if we do that.

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
```diff
@@ -15,6 +15,15 @@ import fr.sncf.osrd.utils.units.Speed
 import fr.sncf.osrd.utils.units.kilometersPerHour
 import fr.sncf.osrd.utils.units.meters
 
+/**
+ * Simple internal class representing a stop with safety speed. Makes the function logic more
+ * straightforward.
+ */
+private data class SafetySpeedStop(
```
bougue-pe, line 22: ❤️

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
eckter: Actually, that doesn't work: sometimes we don't want the train to stop at the end of its simulation. The last stop should generally already be "on stop signal" anyway.

I'll change that.

Khoyo (approved):

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
```diff
@@ -79,6 +113,9 @@ fun getSignalOffsets(
         }
         prevZonePathsLength += rawInfra.getZonePathLength(zonePath).distance
     }
+    // Add one "signal" at the end of the last route no matter what.
+    // There must be either a signal or a buffer stop, on which we may end safety speed ranges.
+    res.add(Offset(prevZonePathsLength - pathStartOffset.distance))
```
bougue-pe, line 117: Is it possible that there is a signal attached to the bufferStop (or the routeExit) and we are "hiding" it if choosing this new one?

eckter: In the current version, the default safety speed applied at the buffer stop will always end at the buffer stop.

But if there's a signal a few meters before, it would still be included in the `signalOffsets`. It would be used if there's a stop on closed signal before that (on top of the buffer stop safety range)

bougue-pe: As discussed, I resolve this (and wait for the tests to be fixed):
* there could be a wrong offset on the safety-speed range if we ever have explicit signal upstream a buffer-stop, but it is not the case for France
* there could be an issue if the train finishes between the final signal of the route and the final detector of the route, but it's an "undefined behavior" case
* As for the final route not being the "best one" for the simulation, the usual issues are here:
  * we could apply a safety-speed because of a buffer-stop that is after the end of the path and choosing another route we would have avoided it
  * we could miss the application of a buffer-stop's safety-speed just because it's on a very short route after the last one used on the path

bougue-pe (approved):
Thanks for the digging on that!

PR #9310, opened 2024-10-11
editoast: add healthcheck command

ghost:


woshilapin (approved):
Quick, thank you.

editoast/src/main.rs
```diff
@@ -266,9 +267,22 @@ async fn run() -> Result<(), Box<dyn Error + Send + Sync>> {
...
 
+async fn healthcheck_cmd(
+    db_pool: Arc<DbConnectionPoolV2>,
+    redis_config: RedisConfig,
+) -> Result<(), Box<dyn Error + Send + Sync>> {
+    let redis = RedisClient::new(redis_config).unwrap();
+    check_health(db_pool.into(), redis.into())
+        .await
+        .map_err(|e| CliError::new(1, format!("❌ healthcheck failed: {0}", e)))?;
+    println!("❎ Healthcheck passed");
```
emersion: Green cross is leaving a bitter-sweet taste in my mouth :P 

(Do we really want to log each and every healthcheck request?)

ghost: It will not be logged in the output, it'll be executed by k8s directly, we do not have this in the log iirc.

In fact, we will not have the log of every healthcheck HTTP request in the main log anymore if we use an exec instead of a request.

leovalais (approved):
I wasn't in your discussion about the issue, sorry if you already answered this question, why keeping the endpoint? Can't we just change the docker compose health check to use the CLI?

ghost:
> I wasn't in your discussion about the issue, sorry if you already answered this question, why keeping the endpoint? Can't we just change the docker compose health check to use the CLI?

Because we it's relevant for the readiness check and for the tileserver. All in all, it is a more accurate check going through the http server than spawning a process beside in the container..

PR #9314, opened 2024-10-11
front: display new warnings in stdcm

### LINKED ISSUE #9299: Display infra loading in STDCM

End users

If the infra isn't loaded STDCM can't work, but this info isn't present in the front end. The user should know if the infrastructure is loading and not be able to run a simulation if it is the case.

- [ ] Someting like this should be implemented, with the same dot size as when the infrastructure's loading in a scenario:

Image

- [ ] The "Obtenir la simulation" button should be disabled while the infra is loading

Without this STDCM breaks

clarani:
closes #9299

Display a warning if the infra is loading or if the user launches a request when some waypoints do not have a location

### DIFF
--- a/front/src/applications/stdcmV2/components/StdcmConfig.tsx
+++ b/front/src/applications/stdcmV2/components/StdcmConfig.tsx
@@ -2,14 +2,15 @@ import { useEffect, useState } from 'react';
 
 import { Button } from '@osrd-project/ui-core';
 import cx from 'classnames';
+import { compact } from 'lodash';
 import { useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
 
 import { useOsrdConfActions, useOsrdConfSelectors } from 'common/osrdContext';
+import useInfraStatus from 'modules/pathfinding/hooks/useInfraStatus';
 import { Map } from 'modules/trainschedule/components/ManageTrainSchedule';
 import type { StdcmConfSliceActions } from 'reducers/osrdconf/stdcmConf';
 import { useAppDispatch } from 'store';
-import { extractHHMM } from 'utils/date';
 
 import StdcmConsist from './StdcmConsist';
 import StdcmDestination from './StdcmDestination';
@@ -47,6 +48,7 @@ const StdcmConfig = ({
 }: StdcmConfigProps) => {
   const { t } = useTranslation('stdcm');
 
+  const { infra } = useInfraStatus();
   const dispatch = useAppDispatch();
   const {
     updateGridMarginAfter,
@@ -56,15 +58,16 @@ const StdcmConfig = ({
     updateDestinationArrivalType,
   } = useOsrdConfActions() as StdcmConfSliceActions;
 
-  const { getOrigin, getDestination, getProjectID, getScenarioID, getStudyID } =
+  const { getOrigin, getDestination, getPathSteps, getProjectID, getScenarioID, getStudyID } =
     useOsrdConfSelectors();
   const origin = useSelector(getOrigin);
+  const pathSteps = useSelector(getPathSteps);
   const destination = useSelector(getDestination);
   const projectID = useSelector(getProjectID);
   const studyID = useSelector(getStudyID);
   const scenarioID = useSelector(getScenarioID);
 
-  const pathfinding = useStaticPathfinding();
+  const pathfinding = useStaticPathfinding(infra);
 
   const [formErrors, setFormErrors] = useState<StdcmConfigErrors>();
 
@@ -77,8 +80,20 @@ const StdcmConfig = ({
   };
 
   const startSimulation = () => {
-    if (pathfinding?.status === 'success' && !formErrors) {
+    const isPathfindingFailed = !!pathfinding && pathfinding.status !== 'success';
+    const formErrorsStatus = checkStdcmConfigErrors(
+      isPathfindingFailed,
+      origin,
+      destination,
+      compact(pathSteps),
+      t
+    );
+    if (pathfinding?.status === 'success' && !formErrorsStatus) {
       launchStdcmRequest();
+    } else {
+      // The console error is only for debugging the user tests (temporary)
+      console.warn('The form is not valid:', { pathfinding, formErrorsStatus });
+      setFormErrors(formErrorsStatus);
     }
   };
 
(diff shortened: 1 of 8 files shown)


front/src/applications/stdcmV2/utils/checkStdcmConfigErrors.ts
```diff
@@ -16,6 +21,10 @@ const checkStdcmConfigErrors = (
     return { errorType: StdcmConfigErrorTypes.PATHFINDING_FAILED };
   }
 
+  if (pathSteps.some((step) => 'uic' in step && step.uic === -1)) {
```
SharglutDev: We should be not compact `pathSteps` when passing it to the function (and change the type for the parameter) to handle case like this : 

Capture d’écran 2024-10-11 à 16 49 47

```suggestion
  if (pathSteps.some((step) => step === null || 'uic' in step && step.uic === -1)) {
```

front/src/applications/stdcmV2/components/StdcmWarningBox.tsx
```diff
@@ -24,7 +30,14 @@ const StdcmWarningBox = ({
       <span>
         <Alert variant="fill" />
```
SharglutDev: Too small compared to the mockup

```suggestion
        <Alert variant="fill" size="lg" />
```

front/src/applications/stdcmV2/components/StdcmWarningBox.tsx
```diff
@@ -24,7 +30,14 @@ const StdcmWarningBox = ({
...
       </span>
-      <p className="mb-0 text-justify">{t(`stdcmErrors.${errorType}`)}</p>
+      <p
+        className={cx('mb-0', {
+          'text-center': SHORT_TEXT_ERRORS.includes(errorType),
+          'text-justify': !SHORT_TEXT_ERRORS.includes(errorType),
```
SharglutDev, line 36: We can discuss it later but it's not justified in the mockup, @thibautsailly doesn't like this style option.

mockup
Capture d’écran 2024-10-11 à 16 56 46

front/src/applications/stdcmV2/components/StdcmConfig.tsx
```diff
@@ -117,6 +124,14 @@ const StdcmConfig = ({
     }
   }, [isDebugMode]);
 
+  useEffect(() => {
+    if (!infra || infra.state === 'CACHED') {
```
SharglutDev, line 129: Don't we want to display the warning if we have no infra ?

front/src/applications/stdcmV2/components/StdcmConfig.tsx
```diff
@@ -77,8 +80,19 @@ const StdcmConfig = ({
...
+      isPathfindingFailed,
+      origin,
+      destination,
+      compact(pathSteps),
+      t
+    );
+    if (pathfinding?.status === 'success' && !formErrorsStatus) {
       launchStdcmRequest();
+    } else {
+      console.error('The form is not valid:', { pathfinding, formErrorsStatus });
```
SharglutDev: We should keep it a warning ?

```suggestion
      console.warn('The form is not valid:', { pathfinding, formErrorsStatus });
```

front/src/applications/stdcmV2/components/StdcmConfig.tsx
```diff
@@ -91,22 +105,15 @@ const StdcmConfig = ({
...
+    const formErrorsStatus = checkStdcmConfigErrors(
+      isPathfindingFailed,
+      origin,
+      destination,
+      [],
+      t
+    );
     setFormErrors(formErrorsStatus);
-  }, [origin, destination, pathfinding]);
+  }, [origin, pathSteps, destination, pathfinding]);
```
SharglutDev: Why put pathSteps here if we don't use it in the useEffect ?

front/src/applications/stdcmV2/components/StdcmWarningBox.tsx
clarani: Done

front/src/applications/stdcmV2/components/StdcmWarningBox.tsx
clarani: Acknowledged, I don't really want to modify the style right now, only the behavior for Monday

front/src/applications/stdcmV2/components/StdcmConfig.tsx
clarani: Yes, I want to keep it for the user tests. I can add a comment :+1:

front/src/applications/stdcmV2/components/StdcmConfig.tsx
clarani: Done

front/src/applications/stdcmV2/components/StdcmConfig.tsx
clarani: If we have no infra, it's because the first call to the backend is not finished. The first call will send the infra, even if it is not loaded.
This condition prevents the interface to blink on refresh:
[Capture vidéo du 2024-10-11 17-14-09.webm](https://github.com/user-attachments/assets/1c1b18f0-6fc9-411a-af66-fb7921675085)

front/src/applications/stdcmV2/utils/checkStdcmConfigErrors.ts
clarani: Done :+1:

front/src/applications/stdcmV2/hooks/useStaticPathfinding.ts
```diff
@@ -3,17 +3,19 @@ import { useEffect, useMemo, useState } from 'react';
...
+  type PathfindingResult,
+} from 'common/api/osrdEditoastApi';
 import { useOsrdConfSelectors } from 'common/osrdContext';
-import useInfraStatus from 'modules/pathfinding/hooks/useInfraStatus';
 import usePathProperties from 'modules/pathfinding/hooks/usePathProperties';
 import { getPathfindingQuery } from 'modules/pathfinding/utils';
 import { useStoreDataForRollingStockSelector } from 'modules/rollingStock/components/RollingStockSelector/useStoreDataForRollingStockSelector';
 
-const useStaticPathfinding = () => {
+const useStaticPathfinding = (infra?: InfraWithState) => {
```
Yohh, line 16: instead of passing an optional param, shouldn't we check `infra` before using `useStaticPathfinding`?

clarani: I can not condition the execution of a hook (i.e. sometimes execute it and sometimes not, otherwise we will have the error: "not the same number of hooks"), so we have to pass the infra even if it is undefined and the test must be done in the hook ;)

Yohh: ah yes ok

SharglutDev (approved):
Lgtm and tested

PR #9316, opened 2024-10-11
editoast: use spawn_blocking for long processing tasks

woshilapin:
Also add some telemetry to look at the proportion in time.


woshilapin:
Globally, it seems that the projection on `editoast` is relatively fast (call to `compute_batch_space_time_curves`, green arrow on the screenshot), especially compared to the request to `core` (call to `compute_batch_signal_updates`, red arrow on the screenshot).

Screenshot 2024-10-14 at 16-12-59 APM Traces Datadog

I'm closing the PR since it won't bring improvement.

PR #9317, opened 2024-10-11
editoast: use pooling for amqp

### LINKED ISSUE #9318: editoast: we reach the limit of channel per connexion on amqp client and crash subsequent requests
We should pool connections and keep the number of channel per connection as stated [here in the documentation](https://www.rabbitmq.com/docs/channels#resource-usage)

ghost:
Closes #9318

### DIFF
--- a/editoast/src/core/mq_client.rs
+++ b/editoast/src/core/mq_client.rs
@@ -1,23 +1,21 @@
+use deadpool_lapin::{Config, CreatePoolError, Pool, PoolError, Runtime};
 use editoast_derive::EditoastError;
 use futures_util::StreamExt;
 use itertools::Itertools;
 use lapin::{
     options::{BasicConsumeOptions, BasicPublishOptions},
     types::{ByteArray, FieldTable, ShortString},
-    BasicProperties, Connection, ConnectionProperties,
+    BasicProperties,
 };
 use serde::Serialize;
 use serde_json::to_vec;
-use std::{fmt::Debug, sync::Arc};
+use std::fmt::Debug;
 use thiserror::Error;
-use tokio::{
-    sync::RwLock,
-    time::{timeout, Duration},
-};
+use tokio::time::{timeout, Duration};
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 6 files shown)


flomonster (approved):
LGTM (not tested)

bloussou (approved):
:pray:

clarani (approved):
LGTM :white_check_mark:

editoast/src/core/mq_client.rs
```diff
@@ -51,6 +49,12 @@ pub enum Error {
     #[error("Connection does not exist")]
     #[editoast_error(status = "500")]
     ConnectionDoesNotExist,
+    #[error("Cannot create the pool")]
+    #[editoast_error(status = "500")]
+    CeatePoolLapin(CreatePoolError),
```
bougue-pe: There are other occurrences, please `ctrl+F`:
```suggestion
    CreatePoolLapin(CreatePoolError),
```

editoast/src/core/mq_client.rs
```diff
@@ -51,6 +49,12 @@ pub enum Error {
     #[error("Connection does not exist")]
     #[editoast_error(status = "500")]
     ConnectionDoesNotExist,
```
bougue-pe, line 51: nit: "Connection" is not so expressive here (maybe other errors above are subject to this).
How about having the same suffix? (nit: Not sure `Lapin` is the most expressive)
```suggestion
    ConnectionDoesNotExistLapin,
```

bougue-pe (commented):
Feel free to resolve without addressing if in rush 🙏

Khoyo (approved):
I don't believe this address the root issue (ie. probably channel leaks), but it might hide it enough for now

editoast/src/core/mq_client.rs
ghost: Will do in another PR, didn't change the name of this one while working on this PR

ghost:
> I don't believe this address the root issue (ie. probably channel leaks), but it might hide it enough for now

Added explicit closing of the channels.

PR #9327, opened 2024-10-14
editoast: return conflicts list when STDCM request fails

### LINKED ISSUE #8639: Editoast: explanation of why the simulation is not feasible
AC : 
- [ ] modify STDCM response format to take into account when STDCM finds no solution
- [ ] when no solution found, editoast has to run a simulation (margin 4.5min/100km) (without creating a circulation objet), then returns the "fake" train to the front-end

hamz2a:
closes [#8639](https://github.com/OpenRailAssociation/osrd/issues/8639)

### DIFF
--- a/editoast/src/core/stdcm.rs
+++ b/editoast/src/core/stdcm.rs
@@ -11,12 +11,14 @@ use serde::Deserialize;
 use serde::Serialize;
 use utoipa::ToSchema;
 
+use super::conflict_detection::Conflict;
 use super::conflict_detection::TrainRequirements;
 use super::pathfinding::PathfindingResultSuccess;
 use super::pathfinding::TrackRange;
 use super::simulation::PhysicsRollingStock;
 use super::simulation::SimulationResponse;
 use crate::core::{AsCoreRequest, Json};
+use crate::views::path::pathfinding::PathfindingResult;
 
 #[derive(Debug, Serialize)]
 pub struct STDCMRequest {
@@ -113,7 +115,7 @@ pub struct UndirectedTrackRange {
     pub end: u64,
 }
 
-#[derive(Serialize, Deserialize, Clone, Debug, ToSchema)]
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, ToSchema)]
 #[serde(tag = "status", rename_all = "snake_case")]
 // We accepted the difference of memory size taken by variants
 // Since there is only on success and others are error cases
@@ -125,6 +127,10 @@ pub enum STDCMResponse {
         departure_time: DateTime<Utc>,
     },
     PathNotFound,
+    Conflicts {
+        pathfinding_result: PathfindingResult,
+        conflicts: Vec<Conflict>,
+    },
     PreprocessingSimulationError {
         error: SimulationResponse,
     },
--- a/editoast/src/views/timetable/stdcm.rs
+++ b/editoast/src/views/timetable/stdcm.rs
@@ -25,16 +25,19 @@ use utoipa::IntoParams;
 use utoipa::ToSchema;
 
 use super::SelectionSettings;
+use crate::core::conflict_detection::ConflictDetectionRequest;
 use crate::core::conflict_detection::TrainRequirements;
+use crate::core::conflict_detection::WorkSchedulesRequest;
 use crate::core::pathfinding::InvalidPathItem;
 use crate::core::pathfinding::PathfindingInputError;
 use crate::core::simulation::PhysicsRollingStock;
 use crate::core::simulation::SimulationParameters;
 use crate::core::simulation::{RoutingRequirement, SimulationResponse, SpacingRequirement};
+use crate::core::stdcm::STDCMPathItem;
+use crate::core::stdcm::STDCMRequest;
 use crate::core::stdcm::STDCMResponse;
-use crate::core::stdcm::TemporarySpeedLimit as CoreTemporarySpeedLimit;
-use crate::core::stdcm::{STDCMPathItem, UndirectedTrackRange, WorkSchedule as CoreWorkSchedule};
-use crate::core::stdcm::{STDCMRequest, STDCMStepTimingData};
+use crate::core::stdcm::STDCMStepTimingData;
+use crate::core::stdcm::UndirectedTrackRange;
 use crate::core::AsCoreRequest;
 use crate::core::CoreClient;
 use crate::error::Result;
@@ -159,7 +162,15 @@ struct InfraIdQueryParam {
     infra: i64,
 }
 
-/// Compute a STDCM and return the simulation result
+/// This function computes a STDCM and returns the result.
+/// It first checks user authorization, then retrieves timetable, infrastructure,
+/// train schedules, and rolling stock data, and runs train simulations.
+/// The result contains the simulation output based on the train schedules
+/// and infrastructure provided.
+///
+/// If the simulation fails, the function uses a virtual train to detect conflicts
+/// with existing train schedules. It then returns both the conflict information
+/// and the pathfinding result from the virtual train's simulation.
 #[utoipa::path(
     post, path = "",
     tag = "stdcm",
@@ -204,8 +215,8 @@ async fn stdcm(
     let infra =
         Infra::retrieve_or_fail(conn, infra_id, || STDCMError::InfraNotFound { infra_id }).await?;
 
-    let (trains, _): (Vec<_>, _) =
-        TrainSchedule::retrieve_batch(conn, timetable_trains.train_ids).await?;
+    let (train_schedules, _): (Vec<_>, _) =
+        TrainSchedule::retrieve_batch(conn, timetable_trains.train_ids.clone()).await?;
 
     let rolling_stock =
         RollingStockModel::retrieve_or_fail(conn, stdcm_request.rolling_stock_id, || {
@@ -219,7 +230,7 @@ async fn stdcm(
         conn,
         valkey_client.clone(),
         core_client.clone(),
-        &trains,
+        &train_schedules,
         &infra,
         stdcm_request.electrical_profile_set_id,
     )
@@ -227,7 +238,12 @@ async fn stdcm(
 
     // 2. Compute the earliest start time, maximum running time and maximum departure delay
     // Simulation time without stop duration
-    let simulation_run_time_result = get_simulation_run_time(
+    let (
+        simulation_run_time,
+        virtual_train_schedule,
+        virtual_train_sim_result,
+        virtual_train_pathfinding_result,
+    ) = simulate_train_run(
         db_pool.clone(),
         valkey_client.clone(),
         core_client.clone(),
@@ -237,7 +253,7 @@ async fn stdcm(
         timetable_id,
     )
     .await?;
-    let simulation_run_time = match simulation_run_time_result {
+    let simulation_run_time = match simulation_run_time {
         SimulationTimeResult::SimulationTime { value } => value,
         SimulationTimeResult::Error { error } => {
             return Ok(Json(STDCMResponse::PreprocessingSimulationError {
@@ -245,6 +261,7 @@ async fn stdcm(
             }))
         }
     };
+
     let earliest_step_tolerance_window = get_earliest_step_tolerance_window(&stdcm_request);
     let maximum_departure_delay = get_maximum_departure_delay(
         &stdcm_request,
@@ -262,8 +279,8 @@ async fn stdcm(
 
     // 3. Get scheduled train requirements
     let trains_requirements = build_train_requirements(
-        trains,
-        simulations,
+        train_schedules.clone(),
+        simulations.clone(),
         earliest_departure_time,
         latest_simulation_end,
     );
@@ -285,10 +302,20 @@ async fn stdcm(
         None => vec![],
     };
 
-    // 6. Build STDCM request
-    let stdcm_response = STDCMRequest {
+    // 6. Retrieve work schedules
+    let work_schedules = match stdcm_request.work_schedule_group_id {
+        Some(work_schedule_group_id) => {
+            let selection_setting = SelectionSettings::new()
+                .filter(move || WorkSchedule::WORK_SCHEDULE_GROUP_ID.eq(work_schedule_group_id));
+            WorkSchedule::list(conn, selection_setting).await?
+        }
+        None => vec![],
+    };
+
+    // 7. Build STDCM request
+    let stdcm_request = STDCMRequest {
         infra: infra.id,
-        expected_version: infra.version,
+        expected_version: infra.version.clone(),
         rolling_stock_loading_gauge: rolling_stock.loading_gauge,
         rolling_stock_supported_signaling_systems: rolling_stock
             .supported_signaling_systems
@@ -296,7 +323,7 @@ async fn stdcm(
         comfort: stdcm_request.comfort,
         path_items,
         start_time: earliest_departure_time,
-        trains_requirements,
+        trains_requirements: trains_requirements.clone(),
         maximum_departure_delay,
         maximum_run_time,
         speed_limit_tag: stdcm_request.speed_limit_tags,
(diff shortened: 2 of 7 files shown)


editoast/src/views/timetable/stdcm.rs
```diff
@@ -445,17 +499,23 @@ fn get_earliest_step_tolerance_window(data: &STDCMRequestPayload) -> u64 {
...
     data: &STDCMRequestPayload,
     infra: &Infra,
     rolling_stock: &RollingStockModel,
     timetable_id: i64,
-) -> Result<SimulationTimeResult> {
+) -> Result<(
+    SimulationTimeResult,
+    TrainSchedule,
+    (SimulationResponse, PathfindingResult),
+)> {
```
woshilapin: It feels like it's missing a type or at least an alias here. A tuple of tuple is not clear, and naming the intermediate tuple with type alias would help to understand why they're grouped like this (or maybe we need a real `struct`?).

editoast/src/views/timetable/stdcm.rs
```diff
@@ -276,44 +294,80 @@ async fn stdcm(
...
+            &work_schedules,
+            earliest_departure_time,
+            maximum_run_time,
+        ),
         temporary_speed_limits,
+    };
+
+    let stdcm_response = stdcm_request.fetch(core_client.as_ref()).await?;
+
+    if let STDCMResponse::PathNotFound = stdcm_response {
```
woshilapin, line 346: This big `if` block could probably benefit to be extracted in a separate function.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -543,21 +608,17 @@ fn build_single_margin(margin: Option<MarginValue>) -> Margins {
 }
 
 /// Build the list of work schedules for the given time range
-async fn build_work_schedules(
-    conn: &mut DbConnection,
-    time: DateTime<Utc>,
+fn build_stdcm_work_schedules(
```
woshilapin: `build_stdcm_work_schedule` and `work_conflict_work_schedule` are very similar. Shouldn't they share some code? I don't understand why we do have 2 different functions that almost do the same thing... but not quite. Available to discuss it if you need.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -638,7 +733,205 @@ async fn parse_stdcm_steps(
...
+        let mut core = core_mocking_client();
+        core.stub("/v2/stdcm")
+            .method(reqwest::Method::POST)
+            .response(StatusCode::OK)
+            .json(json!({"status": "path_not_found"}))
+            .finish();
+        core.stub("/v2/conflict_detection")
+            .method(reqwest::Method::POST)
+            .response(StatusCode::OK)
+            .json(json!({"conflicts": []}))
```
woshilapin: Might be interesting to add a conflict, just to see if we forward it correctly.

woshilapin (approved):
Again some more debuggability information send back to the front, I love it 😍 

In this PR, there is some renaming interlaced with changes of functionality. It would have been nice to first do the renaming in an independent commit to ease the review of the important part (note that the renaming is nice, just not a change that need attentive reviewing).

editoast/src/views/timetable/stdcm.rs
hamz2a:
```rust
) -> Result<(
    SimulationTimeResult,
    TrainSchedule,
    SimulationResponse,
    PathfindingResult,
)> {
```
I've replaced the nested tuple with explicit types, what do you think ?

editoast/src/views/timetable/stdcm.rs
```diff
@@ -264,10 +272,20 @@ async fn stdcm(
         None => vec![],
     };
 
-    // 6. Build STDCM request
-    let stdcm_response = STDCMRequest {
+    // 6. Retrieve work schedules
+    let work_schedules = match stdcm_request.work_schedule_group_id {
+        Some(work_schedule_group_id) => {
+            let selection_setting: SelectionSettings<WorkSchedule> = SelectionSettings::new()
```
```suggestion
            let selection_setting = SelectionSettings::new()
```

I think rust can infer the type here.

editoast/src/views/timetable/stdcm.rs

editoast/src/views/timetable/stdcm.rs
```diff
@@ -264,10 +272,20 @@ async fn stdcm(
...
+                .filter(move || WorkSchedule::WORK_SCHEDULE_GROUP_ID.eq(work_schedule_group_id));
+            WorkSchedule::list(conn, selection_setting).await?
+        }
+        None => vec![],
```
leovalais, line 312: Just pointing out you don't need to explicitly `match`, you can also `work_schedule_group_id.map(...).unwrap_or_default()`. Feel free to resolve.

editoast/src/views/timetable/stdcm.rs
leovalais: Agreed about moving the response post-processing into another function.

editoast/src/views/timetable/stdcm.rs
leovalais: Very strong agree about turning this into a real `struct`. Not only does it names the values returned, it also allows to document somewhere what `fake_train_schedule` and friends actually are and why we compute them.

editoast/src/views/timetable/stdcm.rs
leovalais:
```suggestion
fn filter_work_schedules_time_window(
```

The WorkSchedule list is now provided in argument. Agreed about factorizing the filtering part.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -638,7 +733,205 @@ async fn parse_stdcm_steps(
...
+
+    fn core_mocking_client() -> MockingClient {
+        let mut core = MockingClient::new();
+        core.stub("/v2/pathfinding/blocks")
+            .method(reqwest::Method::POST)
+            .response(StatusCode::OK)
+            .json(
+                serde_json::to_value(PathfindingResult::Success(pathfinding_result_success()))
+                    .unwrap(),
+            )
```
```suggestion
            .json(PathfindingResult::Success(pathfinding_result_success()))
```

editoast/src/views/timetable/stdcm.rs
```diff
@@ -638,7 +733,205 @@ async fn parse_stdcm_steps(
...
+    use crate::core::simulation::SimulationResponse;
+    use crate::core::simulation::SpeedLimitProperties;
+    use crate::core::stdcm::STDCMResponse;
+    use crate::models::fixtures::create_fast_rolling_stock;
+    use crate::models::fixtures::create_small_infra;
+    use crate::models::fixtures::create_timetable;
+    use crate::views::test_app::TestAppBuilder;
+
+    fn pathfinding_result_success() -> PathfindingResultSuccess {
+        PathfindingResultSuccess {
```
```suggestion
        PathfindingResult::Success(PathfindingResultSuccess {
```

editoast/src/views/timetable/stdcm.rs
hamz2a: Done.

leovalais (changes requested):
Thanks for this PR! Left a few comments.

On a more general note, the `/stdcm` endpoint is starting to do a whole lot of things and this process isn't documented anywhere. The endpoint documentation is just

```rust
/// Compute a STDCM and return the simulation result
```

which isn't exact anymore following your changes.


I'd suggest:

* documenting the whole process in the handler's documentation with the "happy path" and what we're doing if the simulation fails, making sure to separate both
* continue the series of scoping comments `// n. Nth step` in your additions, as it's really helpful to figure out where we're at in the global process
* split the individual steps into documented functions (you don't have to do that for existing steps, but it would be a nice thing to do for your changes so that we can start enforcing a better modularity from now on) 

Wdyt?

editoast/src/views/timetable/stdcm.rs

editoast/src/views/timetable/stdcm.rs

editoast/src/views/timetable/stdcm.rs

editoast/src/views/timetable/stdcm.rs

tests/tests/test_stdcm.py
```diff
@@ -266,7 +266,7 @@ def test_max_running_time(small_scenario: Scenario, fast_rolling_stock: int):
     r = requests.post(url, json=payload)
     response = r.json()
     assert r.status_code == 200
-    assert response == {"status": "path_not_found"}
+    assert response["status"] == "conflicts"
```
shenriotpro: Should we check the list?

hamz2a: i added the list. could you please check?

shenriotpro: Thanks, I'm not sure about the details (times for instance), I'll let @eckter check those.

tests/tests/test_stdcm.py
```diff
@@ -266,7 +266,75 @@ def test_max_running_time(small_scenario: Scenario, fast_rolling_stock: int):
...
     assert r.status_code == 200
-    assert response == {"status": "path_not_found"}
+    assert response == {
+        "status": "conflicts",
+        "conflicts": [
+            {
+                "conflict_type": "Spacing",
+                "start_time": "2024-01-01T07:30:00Z",
+                "end_time": "2024-01-01T16:00:00Z",
+                "train_ids": [0],
```
eckter: It's weird that there is a train_id there. The timetable is supposed to be empty with just work schedules. 

Is it the train id of the "fake" new train?

tests/tests/test_stdcm.py
```diff
@@ -266,7 +266,75 @@ def test_max_running_time(small_scenario: Scenario, fast_rolling_stock: int):
...
     response = r.json()
     assert r.status_code == 200
-    assert response == {"status": "path_not_found"}
+    assert response == {
+        "status": "conflicts",
+        "conflicts": [
+            {
+                "conflict_type": "Spacing",
+                "start_time": "2024-01-01T07:30:00Z",
+                "end_time": "2024-01-01T16:00:00Z",
```
eckter, line 275: If the train starts at 7:30, the conflict start time should be later than this (it happens on the last track). And the end time should be earlier, the path isn't that long.

(same comment for requirement times)

tests/tests/test_stdcm.py
hamz2a: `0` is the id of `fake_train`

eckter: I don't think it should be included in the response then.

By the way, `0` may also be the id of a valid train, right? What happens if the new trains conflicts with a scheduled train with id=0?

hamz2a: As discussed, I removed the ID of the fake train from the list.

eckter: Nice, thanks.

What about the case where there's also a train with id=0 in the database?

Maybe I'll ping someone more familiar with editoast because I'm not 100% confident about how ids work (ping @woshilapin @leovalais). Train IDs are incremental integers starting at 0 included, right? Could we just replace the value here with a `-1` or would that break things? (it shouldn't be saved in db)

tests/tests/test_stdcm.py
eckter: Resolving this, this weird behavior comes from core. I'll double check what's the intended behavior and open a different issue if relevant.

tests/tests/test_stdcm.py
hamz2a: We use `GENERATED BY DEFAULT AS IDENTITY` in Postgres, which starts generating IDs from 1, so we don’t have an issue with id=0 being generated by the database.
For more details: https://www.postgresql.org/docs/current/ddl-identity-columns.html

editoast/src/views/timetable/stdcm.rs
```diff
@@ -157,7 +160,15 @@ struct InfraIdQueryParam {
...
 }
 
-/// Compute a STDCM and return the simulation result
+/// This function computes a STDCM and returns the result.
+/// It first checks user authorization, then retrieves timetable, infrastructure,
+/// train schedules, and rolling stock data, and runs train simulations.
+/// The result contains the simulation output based on the train schedules
+/// and infrastructure provided.
+///
+/// If the simulation fails, the function uses a fake train to detect conflicts
```
```suggestion
/// If the simulation fails, the function uses a direct train to detect conflicts
```

`fake` feels like it's a completely unrelated simulation. If I record correctly, it's only a simulation on the most basic pathfinding that we re-run, right?

editoast/src/views/timetable/stdcm.rs
woshilapin: A way to avoid `#[allow(clippy::too_many_arguments)]` in this kind of case is to create a `struct` that holds all of the arguments as fields, then `impl` the function `handle_path_not_found` on the `struct`.

Another possibility is to regroup some of these input arguments in some `struct` (for example, `fake_train_sim_result` and `fake_train_pathfinding_result` into a `FakeTrainResult`, or `earliest_departure_time`, `maximum_run_time` and `latest_simulation_end` into a `SimulationConstraint`?).

And maybe the solution is somewhere in-between. Typically, you could have a `PathHandler` that is initialized with `infra_id`, `infra.version`, `train_schedules`, `work_schedules` and `core_client` (the mostly static part), and a function `generate_fake_train` that would return a `fake_train_schedule`, another function `run_simulation` that would return `FakeTrainResult` (that holds both `fake_train_sim_result` and `fake_train_pathfinding_result`), then finally a `get_conflicts` that would return what you're finally looking for. What do you think? (cc @leovalais)

Some of these ideas might not entirely fit the existing codebase but I hope I get you some ideas about how to deal with this?

shenriotpro (approved):
LGTM for tests/

editoast/src/views/timetable/stdcm.rs
```diff
@@ -217,15 +228,20 @@ async fn stdcm(
...
     .await?;
 
     // 2. Compute the earliest start time, maximum running time and maximum departure delay
     // Simulation time without stop duration
-    let simulation_run_time_result = get_simulation_run_time(
+    let (
+        simulation_run_time,
+        fake_train_schedule,
+        fake_train_sim_result,
+        fake_train_pathfinding_result,
```
woshilapin: I would really rename `fake_` with something else. "Fake" kind of mean it doesn't exist (and kind of also means it's crap). Maybe `virtual` would be at least not have this pejorative connotation (but still means "doesn't really exist"). But if I record correctly, this train is basically a direct train, with the shortest pathfinding (not taking into consideration conflicts). So maybe `direct_train` would be good? Not sure so feel free to propose something else.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -464,17 +575,25 @@ fn get_earliest_step_tolerance_window(data: &STDCMRequestPayload) -> u64 {
...
     infra: &Infra,
     rolling_stock: &RollingStockModel,
     timetable_id: i64,
-) -> Result<SimulationTimeResult> {
+) -> Result<(
+    SimulationTimeResult,
+    TrainSchedule,
+    SimulationResponse,
+    PathfindingResult,
+)> {
```
woshilapin, line 601: That's a big tuple, I would rather make a type out of it. Grouping the input (`TrainSchedule`) with the results (`SimulationResponse` and `PathFindingResult`) in a `struct` feels like it would make sense.

As for `SimulationTimeResult`, since it's a by-product of `SimulationResponse`, I would rather let the caller deal with it. What do you think? (cc @leovalais)

editoast/src/views/timetable/stdcm.rs
```diff
@@ -561,37 +685,56 @@ fn build_single_margin(margin: Option<MarginValue>) -> Margins {
...
+    start_time: DateTime<Utc>,
+    maximum_run_time: u64,
+) -> Option<WorkSchedulesRequest> {
+    if work_schedules.is_empty() {
+        return None;
+    }
+
+    let work_schedule_requirements = work_schedules
+        .iter()
+        .map(|ws| (ws.id, filter_core_work_schedule(ws, start_time)))
```
woshilapin, line 735: Since it's actually 2 steps, I find it better to write it as 2 steps, just fewer parents nesting too.

```suggestion
        .map(|ws| filter_core_work_schedule(ws, start_time))
        .map(|filtered| (ws.id, filtered))
```

woshilapin (changes requested):
Sorry for this second pass of review, but I feel we lack some improvements of code architecture. Let's see together if we can make the code more simple and maintainable.

editoast/src/views/timetable/stdcm.rs
leovalais: Agreed. The ideal code presentation to me would be something in the lines of:

```rust
fn stdcm(input: StdcmInput) -> Result<StdcmOutput> {
	let res1: Step1Result = step_1(input)?;
	let res2: Step2Result = step_2(res1)?;
	if res2.success() {
		return Ok(res2.into());
	}
	let res3: FallbackResult = step_fallback(res2)?;
	Ok(res4.into())
}
```

This way each step would have its own structure with exactly what it needs and nothing else making reading easier and changes more explicit in the diffs. And it provides many definitions to document the process properly.

You are probably discussing this with @woshilapin right now, and this can of course be a topic for another PR. But I feel we're reaching an amount of code (and of features) justifying a refactoring to keeps things organized.

editoast/src/views/timetable/stdcm.rs
hamz2a: I can't do the two steps because `filter_core_work_schedule` returns a struct without `id`.

woshilapin: 🤦

leovalais (approved):
LGTM, the remaining comments will be addressed in a separate PR with a refactoring of the endpoint and its utilities.

editoast/src/views/timetable/stdcm.rs
hamz2a: See https://github.com/OpenRailAssociation/osrd/pull/9509

editoast/src/views/timetable/stdcm.rs
hamz2a: See https://github.com/OpenRailAssociation/osrd/pull/9509

editoast/src/views/timetable/stdcm.rs

editoast/src/views/timetable/stdcm.rs

editoast/src/views/timetable/stdcm.rs
hamz2a: See https://github.com/OpenRailAssociation/osrd/pull/9434

editoast/src/core/stdcm.rs
```diff
@@ -125,6 +127,10 @@ pub enum STDCMResponse {
         departure_time: DateTime<Utc>,
     },
     PathNotFound,
+    Conflicts {
+        pathfinding_result: PathfindingResult,
+        conflicts: Vec<Conflict>,
+    },
```
flomonster, line 133: I don't know if this will be done in the following refacto PR. Currently the struct used for the endpoint response is the same as the one used for core response. We should have two different struct now since editoast enrich the response with conflicts.

woshilapin (approved):
Let's merge this one and address all the unaddressed comments in the follow-up PR.

editoast/src/core/stdcm.rs
hamz2a: This might take a bit more time to do in this PR. Is it possible to handle it in the refactor PR https://github.com/OpenRailAssociation/osrd/pull/9434 instead? @flomonster

PR #9333, opened 2024-10-14
front: handle accents in STDCM vias autocomplete

leovalais:


front/src/applications/stdcmV2/components/StdcmOperationalPoint.tsx
```diff
@@ -22,6 +22,18 @@ function formatChCode(chCode: string) {
   return chCode === '' ? 'BV' : chCode;
 }
 
+// Normalizes a string by removing accents and converting to lowercase
+//
+// Accent removal is done using `String.normalize` which converts accented characters to their
+// decomposed form (e.g. é -> e + ´). We then remove all characters in the range of U+0300 to U+036f
+// which contains all unicode combining diacritical marks (https://en.wikipedia.org/wiki/Combining_Diacritical_Marks).
+function normalized(s: string): string {
```
SarahBellaha: Maybe we should have this function in `front/src/utils/strings.ts` ?

SarahBellaha (approved):
LGTM, Thanks ! 💯

SharglutDev (approved):
Lgtm and tested, thank you !

front/src/applications/stdcmV2/components/StdcmOperationalPoint.tsx
SharglutDev: Agreed. You could also use the syntax `/** */` for the comment so we can see it when hovering the function.

leovalais: Done.

PR #9342, opened 2024-10-15
front: fix relief on map

### LINKED ISSUE #9341: front: relief is always displayed
### What happened?

relief is always displayed by default, even though we set the relief coefficient to 0 by default.

Relief should not be displayed by default

1. Open the stdcm map

Recette (SNCF)

Firefox

.

clarani:
closes #9341

### DIFF
--- a/front/src/applications/editor/Map.tsx
+++ b/front/src/applications/editor/Map.tsx
@@ -309,7 +309,6 @@ const MapUnplugged = ({
               <Hillshade
                 mapStyle={mapStyle}
                 layerOrder={LAYER_GROUPS_ORDER[LAYERS.BACKGROUND.GROUP]}
-                display={terrain3DExaggeration > 0}
               />
             </>
           )}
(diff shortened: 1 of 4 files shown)


front/src/common/Map/Layers/HillShade.tsx
```diff
@@ -0,0 +1,41 @@
+import { Source, type LayerProps } from 'react-map-gl/maplibre';
+
+import OrderedLayer from 'common/Map/Layers/OrderedLayer';
+import { useSelector } from 'react-redux';
+import { getTerrain3DExaggeration } from 'reducers/map/selectors';
```
anisometropie: Hillshade is just one word, no need to rename that file

anisometropie (changes requested):

front/src/common/Map/Layers/HillShade.tsx
clarani: Done :white_check_mark:

Synar (approved):
LGTM!

Math-R (approved):
lgtm !

anisometropie (approved):

PR #9343, opened 2024-10-15
Small fixes and improvement for mac users on the docker pr-test file and script

EthanPERRUZZA:
To test it is necessary to use a very recent PR (as an env var has been changed with the switch to Valkey from Redis making the script unusable for not rebased PR's) 
It can be test with the 9343 PR and it works for me


scripts/pr-infra-compose.sh
```diff
@@ -38,6 +44,16 @@ if [ "$2" = "up" ] || [ "$2" = "up-and-load-backup" ]; then
             -f "docker/docker-compose.pr-tests.yml" \
             up -d postgres valkey
```
bougue-pe: As discussed, it's probably simpler to just do (no `while after that`):
```suggestion
            up --wait postgres valkey
```

scripts/pr-infra-compose.sh
bougue-pe, line 1: nit: Is it possible to rename the file(s) so that it matches the compose file, like for other script+compose?
I would go for `pr-compose.sh` and `docker-compose.pr.sh`, but `pr-tests` (or `pr-infra`) for both would be OK to me too.

bougue-pe (commented):
Otherwise LGTM, thanks for all those improvements that would be nice even on a "regular" compose for OSRD 👍

scripts/cleanup-db.sh
```diff
@@ -71,7 +71,11 @@ docker exec "$OSRD_POSTGRES" psql -p "$OSRD_POSTGRES_PORT" -f //tmp/init.sql > /
 
 # Clear Valkey Cache
 echo "Deleting valkey cache..."
-docker exec "$OSRD_VALKEY" valkey-cli -p "$OSRD_VALKEY_PORT" FLUSHALL > /dev/null 2>&1 || docker volume rm -f osrd_valkey_data > /dev/null
+if [ "$PR_TEST" -eq 1 ]; then
+  docker exec "$OSRD_VALKEY" valkey-cli -p "$OSRD_VALKEY_PORT" FLUSHALL > /dev/null 2>&1 || docker volume rm -f osrd-pr-tests_vakey_data > /dev/null
+else
+  docker exec "$OSRD_VALKEY" valkey-cli -p "$OSRD_VALKEY_PORT" FLUSHALL > /dev/null 2>&1 || docker volume rm -f osrd_vakey_data > /dev/null
```
bougue-pe: typo (same above if you keep both):
```suggestion
  docker exec "$OSRD_VALKEY" valkey-cli -p "$OSRD_VALKEY_PORT" FLUSHALL > /dev/null 2>&1 || docker volume rm -f osrd_valkey_data > /dev/null
```

scripts/cleanup-db.sh
```diff
@@ -71,7 +71,11 @@ docker exec "$OSRD_POSTGRES" psql -p "$OSRD_POSTGRES_PORT" -f //tmp/init.sql > /
 
 # Clear Valkey Cache
 echo "Deleting valkey cache..."
-docker exec "$OSRD_VALKEY" valkey-cli -p "$OSRD_VALKEY_PORT" FLUSHALL > /dev/null 2>&1 || docker volume rm -f osrd_valkey_data > /dev/null
+if [ "$PR_TEST" -eq 1 ]; then
+  docker exec "$OSRD_VALKEY" valkey-cli -p "$OSRD_VALKEY_PORT" FLUSHALL > /dev/null 2>&1 || docker volume rm -f osrd-pr-tests_vakey_data > /dev/null
```
bougue-pe: How about mutualizing more by introducing a var:
```suggestion
OSRD_VALKEY_VOLUME=osrd_valkey_data
if [ "$PR_TEST" -eq 1 ]; then
   OSRD_VALKEY_VOLUME=osrd-pr-tests_valkey_data
fi

docker exec "$OSRD_VALKEY" valkey-cli -p "$OSRD_VALKEY_PORT" FLUSHALL > /dev/null 2>&1 || docker volume rm -f "$OSRD_VALKEY_VOLUME" > /dev/null
```

(probably even possible to mutualize the compose project value for the names)

scripts/pr-infra-compose.sh
EthanPERRUZZA: I renamed it `pr-tests-compose.sh` to be consistent with the docker file ;)

bougue-pe: LGTM, just a reminder here to update the corresponding doc in osrd-website 💡 
(and remove the PR number for the "down" and "down-and-clean"?).

scripts/cleanup-db.sh
Khoyo: I'm wondering if it might be nicer to have cleanup DB only provide defaults for `OSRD_VALKEY`, etc, and have the pr-test script pass those instead of `PR_TEST=1`

bougue-pe: Both OK to me, better (but not mandatory for me) if it's consistent all the way.

EthanPERRUZZA: > I'm wondering if it might be nicer to have cleanup DB only provide defaults for `OSRD_VALKEY`, etc, and have the pr-test script pass those instead of `PR_TEST=1`

I would agree it would be better if we were to spawn another DB on another port but I don't see the benefit of doing so as their will always be only two value possible and only one use-case. However, it could makes sense to do so if we want the user to be able to manually / separately use the cleanup script on the PR tests instance. Is it the case ?

EthanPERRUZZA: I'll add the volume var suggestion but with the other vars above. (I think it will make the script easier to understand if everything is defined at the same place)

bougue-pe: > > I'm wondering if it might be nicer to have cleanup DB only provide defaults for `OSRD_VALKEY`, etc, and have the pr-test script pass those instead of `PR_TEST=1`
> 
> I would agree it would be better if we were to spawn another DB on another port but I don't see the benefit of doing so as their will always be only two value possible and only one use-case. However, it could makes sense to do so if we want the user to be able to manually / separately use the cleanup script on the PR tests instance. Is it the case ?

I would go for a clean/simple step on the current use, and we would do another step if another use-case appears.
So OK to just have a clean support of `pr-tests` + regular cleanup.

docker-compose.yml
```diff
@@ -49,9 +49,9 @@ services:
       - rabbitmq_data:/var/lib/rabbitmq
     healthcheck:
       test: ["CMD", "rabbitmqctl", "status"]
-      interval: 30s
+      interval: 10s
       timeout: 30s
-      retries: 3
+      retries: 9
```
bougue-pe: This conflicts with the changes I just merged: https://github.com/OpenRailAssociation/osrd/pull/9166/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3
(same for other compose files was done in my PR).
Changes introduced aimed at being consistent on all deps of OSRD's stack (so after a rebase I would say your change is not necessary anymore).

scripts/pr-infra-compose.sh
EthanPERRUZZA: https://github.com/OpenRailAssociation/osrd-website/pull/247

bougue-pe (approved):
Thanks for this 🎉

Khoyo (approved):

PR #9348, opened 2024-10-15
front: remove infra loader while in macro mode

### LINKED ISSUE #9271: Micro/Macro toggle button is available while infra is loading
### What happened?

While infra is loading, we can still switch in Macro mode (but not the way back)

Image

disable the button while infra is loading

1. open scenario with infra france
2. switch to macro

Recette (SNCF)

Firefox

staging 1482e8c

RomainValls:
closes #9271

### DIFF
--- a/front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
+++ b/front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
@@ -166,6 +166,7 @@ const ScenarioContent = ({
 
           <div className={collapsedTimetable ? 'col-12' : 'col-hdp-9 col-xl-8 col-lg-7 col-md-6'}>
             {!isInfraLoaded &&
+              !isMacro &&
               displayTrainScheduleManagement !== MANAGE_TRAIN_SCHEDULE_TYPES.add &&
               displayTrainScheduleManagement !== MANAGE_TRAIN_SCHEDULE_TYPES.edit && (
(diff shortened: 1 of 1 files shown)


emersion:
I'm not a fan of this solution because nothing in NGE requires a loaded infra. Couldn't we hide the loader while switched to NGE?

cc @louisgreiner

theocrsb (commented):
I think isInfraLoaded is true when `infra.state === TRANSIENT_ERROR`. If so, is this the expected behavior?

louisgreiner:
@emersion, I agree NGE does not need this infra, but it can trigger simulations. Are simulations independent of the infra ? Wouldn't it lead to potential errors if a user create new nodes and trains on it ?

If the behaviour of NGE is safe even without loaded infra, we could just disable the loading screen and allow the user to use the Macro mode (which is very cool actually, while the infra is loading)

emersion:
> but it can trigger simulations

It shouldn't trigger simulations. NGE edits fill up a queue of pending operations which is flushed when the user goes back to micro mode. Only then the simulation should be fired up.

louisgreiner:
Then we can keep macro mode enabled while infra is loading (and remove the loading screen in macro mode)

RomainValls:
Removed the loader while in macro mode, thanks 👍

emersion (approved):
Looks good! Can you squash your commits together into a single one?

PR #9349, opened 2024-10-15
Move most CLI commands into the client module

leovalais:
This is a first step towards having a proper `client` module. Next steps include:

- The remaining few commands (openapi, healthcheck, runserver, ...)
- Move the big `match` in the `run` function in the `client` module
- Move the `AppState` in the `views` module (and remove its `db_pool_v1` component while we're at it)
- Replace most `return Box::new(CliError::new(...` by `anyhow::bail!` since `anyhow`'s paradigm fits perfectly our error management on the `client` side

> [!NOTE]
> The diff is large but this PR just moves things around without modification (except for imports and visibility).


woshilapin (approved):
Good initiative and thank you so much for separating into 5 commits 🙏

Issue #8854, opened 2024-09-12
front: the card of the train in the list is not correctly updated when we update a train


Tguisnet:
### What happened?

When we update a train, the card of the train in the train list is not correctly updated : 
- the arrival time is not updated
- the start time is not updated
- the date is not updated
- If the train honored is schedule but is now late after the edit, he is not in orange
- If the train was late but not anymore after the edit, he is still in orange

### What did you expect to happen?

The update of the list of the train should be correct
- Information about the arrival time, the start time, and the date should be correctly updated
- information on the schedule (honored or not) should be correctly updated

### How can we reproduce it (as minimally and precisely as possible)?

1. Go to a scenario
2. Create a train
3. Edit this train to have different information and make him late (to do it, create an arrival time impossible to keep)
4. Check infos on the list of the trains

### On which environments the bug occurs?

Recette (SNCF)

### On which browser the bug occurs?

Google Chrome

### OSRD version (top right corner `Account` button > `Informations`)

bc8b913

emersion:
Can you still reproduce?

Synar:
Closing unless someone can reproduce again.

PR #9359, opened 2024-10-15
Fix train update timetable

Synar:
Fix at least part of #8854 

So what happened was that, in order : 
- the timetable got updated via putTrainScheduleById in useUpdateTrainSchedule.ts
- the cache was supposed to be invalidated automatically, according to its invalidatesTags in generatedEditoastApi.ts
- the simulation summary was requested via postTrainScheduleSimulationSummary in useLazyLoadTrains.ts (editoast actually receives and correctly responds to the request every time)
- the result of the fetched simulation summary was not actually used, but instead a cached version was used and displayed in the train card, hence the outdated data

There seemed to be some sort of race condition between the cache invalidation and the call to postTrainScheduleSimulationSummary. 

Indeed, adding a one second timer between the call to putTrainScheduleById and postTrainScheduleSimulationSummary solves the issue.

Explicitly invalidating the cache also seems to solve the issue. 

Both of these solutions seem hackish, I would rather find a way to specifically await the automatic cache invalidation, but I don't know how. Any suggestions or corrections would be very welcome.

### DIFF
--- a/front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
+++ b/front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
@@ -88,10 +88,10 @@ const useLazyLoadTrains = ({
         // do not happen during the same react cycle.
         // if we update a train, one is going to re-fetch first and the 2 are out of sync during a few cycles.
         // these cycles do not make sense to render.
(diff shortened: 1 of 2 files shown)


Synar:
PR closed as the branch was renamed

emersion:
Ref https://github.com/OpenRailAssociation/osrd/pull/9363

PR #9363, opened 2024-10-16
front: fix train update timetable

Synar:
Fix at least part of https://github.com/OpenRailAssociation/osrd/issues/8854

So what happened was that, in order :

- the timetable got updated via putTrainScheduleById in useUpdateTrainSchedule.ts
- the cache was supposed to be invalidated automatically, according to its invalidatesTags in generatedEditoastApi.ts
- the simulation summary was requested via postTrainScheduleSimulationSummary in useLazyLoadTrains.ts (editoast actually receives and correctly responds to the request every time)
- the result of the fetched simulation summary was not actually used, but instead a cached version was used and displayed in the train card, hence the outdated data

There seemed to be some sort of race condition between the cache invalidation and the call to postTrainScheduleSimulationSummary.

Indeed, adding a one second timer between the call to putTrainScheduleById and postTrainScheduleSimulationSummary solves the issue.

Explicitly invalidating the cache also seems to solve the issue.

Both of these solutions seem hackish, I would rather find a way to specifically await the automatic cache invalidation, but I don't know how. Any suggestions or corrections would be very welcome.

### DIFF
--- a/front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
+++ b/front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
@@ -88,10 +88,10 @@ const useLazyLoadTrains = ({
         // do not happen during the same react cycle.
         // if we update a train, one is going to re-fetch first and the 2 are out of sync during a few cycles.
         // these cycles do not make sense to render.
-        const outOfSync = [...trainSchedulesById.values()].some((trainShedule) => {
-          const summary = rawSummaries[trainShedule.id];
+        const outOfSync = [...trainSchedulesById.values()].some((trainSchedule) => {
+          const summary = rawSummaries[trainSchedule.id];
           if (summary?.status === 'success') {
-            return trainShedule.path.length !== summary.path_item_times_final.length;
+            return trainSchedule.path.length !== summary.path_item_times_final.length;
           }
           return false;
         });
--- a/front/src/modules/trainschedule/components/ManageTrainSchedule/hooks/useUpdateTrainSchedule.ts
+++ b/front/src/modules/trainschedule/components/ManageTrainSchedule/hooks/useUpdateTrainSchedule.ts
@@ -51,6 +51,11 @@ const useUpdateTrainSchedule = (
           trainScheduleForm: trainSchedule,
         }).unwrap();
 
+        osrdEditoastApi.util.invalidateTags(['train_schedule']);
+        // We explicitely invalidate the cache to avoid a race condition between the automatic cache invalidation
+        // by putTrainScheduleById and the call to postTrainScheduleSimulationSummary in useLazyLoadTrain.ts
+        // Awaiting a set amount of time here would also work : await new Promise((resolve) => setTimeout(resolve, 1000));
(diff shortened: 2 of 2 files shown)


anisometropie (commented):
Can’t find a way to reproduce, it’s the same with or without the fix

Synar:
I believe I can reproduce it pretty consistently by : - creating a train schedule in the scenario with no time stop inputs
- modifying it by adding a required arrival time in the list of time stop inputs and nothing else, or modifying this arrival time
- the train card always lags by one edit, while the summary sent back by editoast is always correct 

But perhaps there's a race condition that's computer dependent? That would be a pain

emersion (commented):
I can't reproduce either :/

<details>

<summary>Note to self: hack patch for debugging</summary>

```diff
diff --git a/front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts b/front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
index b4c9a4df6e72..376c2fb630ce 100644
--- a/front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
+++ b/front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
@@ -83,6 +83,7 @@ const useLazyLoadTrains = ({
             electrical_profile_set_id: electricalProfileSetId,
           },
         }).unwrap();
+        console.log('postTrainScheduleSimulationSummary', _trainToFetchIds, rawSummaries);
 
         // the two rtk-query calls postTrainSchedule & postTrainScheduleSimulationSummary
         // do not happen during the same react cycle.
@@ -96,6 +97,8 @@ const useLazyLoadTrains = ({
           return false;
         });
 
+        console.log(outOfSync);
+
         if (!outOfSync) {
           // launch the projection of the trains
           setTrainIdsToProject((prev) => new Set([...prev, ...packageToFetch]));
diff --git a/front/src/applications/operationalStudies/hooks/useScenarioData.ts b/front/src/applications/operationalStudies/hooks/useScenarioData.ts
index 0c6765792339..a07278cb2478 100644
--- a/front/src/applications/operationalStudies/hooks/useScenarioData.ts
+++ b/front/src/applications/operationalStudies/hooks/useScenarioData.ts
@@ -132,6 +132,7 @@ const useScenarioData = (
       });
 
       const sortedTrainSchedulesToUpsert = sortBy(trainSchedulesToUpsert, 'start_time');
+      console.log('upsertTrainSchedules', sortedTrainSchedulesToUpsert);
       setTrainIdsToFetch(sortedTrainSchedulesToUpsert.map((trainSchedule) => trainSchedule.id));
     },
     [trainSchedules]
```

</details>

Synar:
I tried to reproduce the bug again on the current dev and could not. I'll check later if I can manage to reproduce it on my old branch, to know if something fixed it or if it's just a race condition that requires an exceptionally poor and overworked computer ^^''

Synar:
Tried to go back to this, but there were enough changes that I would need to find an old dump of the db. If nobody can reproduce the bug now, I suppose it's good enough for now.

PR #9366, opened 2024-10-16
front: spaceTimeChart: display occupancy blocks

### LINKED ISSUE #9365: front: space time chart: display occupancy blocks
For the result of a simulation, the user must be able to view the occupancy blocks and their different types. The user must be able to choose which block to display.
BAL or BAPR signal status:

Green: lane free
Red: semaphore
Yellow: warning

Akctarus:
closes #9365


front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -42,7 +48,45 @@ const ManchetteWithSpaceTimeChartWrapper = ({
...
+              : block.aspect_label === '(A)' || block.aspect_label === 'A'
+                ? OCCUPANCY_BLOCKS_COLORS.WARNING
+                : block.aspect_label === '300(VL)' ||
+                    block.aspect_label === '270A' ||
+                    block.aspect_label === '220A' ||
+                    block.aspect_label === '160A' ||
+                    block.aspect_label === '080A' ||
+                    block.aspect_label === '000'
+                  ? OCCUPANCY_BLOCKS_COLORS.GREY
+                  : '',
```
emersion: I think this would be more readable with if/else blocks:

```ts
let color = '';
if (block.aspect_label === 'VL' || block.aspect_label === '300VL') {
    color = OCCUPANCY_BLOCKS_COLORS.FREE;
} else if (...) {
    ...
}
```

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -42,7 +48,45 @@ const ManchetteWithSpaceTimeChartWrapper = ({
   );
 
   const [showSettingsPanel, setShowSettingsPanel] = useState(false);
-  const [settings, setSettings] = useState({ showConflicts: false });
+  const [settings, setSettings] = useState({
+    showConflicts: false,
+    showBAL: false,
+    showBAPR: false,
+    showTVM: false,
```
emersion: As @flomonster noted, could we avoid hardcoding particular signalling systems?

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -42,7 +48,45 @@ const ManchetteWithSpaceTimeChartWrapper = ({
...
+    showSignalsStates: false,
+  });
+
+  const occupancyBlocks = projectPathTrainResult
+    .filter((train) => train.signal_updates)
```
emersion: I wonder, do we need this `filter()`? I think the map below should handle empty arrays just fine?

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -42,7 +48,45 @@ const ManchetteWithSpaceTimeChartWrapper = ({
...
+    .flatMap((train) =>
+      train.signal_updates.flatMap((block) => ({
```
emersion: Nit: probably a `map()` is enough here?

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -42,7 +48,45 @@ const ManchetteWithSpaceTimeChartWrapper = ({
...
+        timeStart: +new Date(train.departure_time) + block.time_start,
+        timeEnd: +new Date(train.departure_time) + block.time_end,
+        spaceStart: block.position_start,
+        spaceEnd: block.position_end,
+        aspect_label: block.aspect_label,
```
emersion: I don't remember this field being part of the occupancy block type on the osrd-ui side, am I missing something?

(Same for `departure_time`)

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
emersion: Or even better, a `switch` or an object, since we're only testing a single value `aspect_label` here.

SharglutDev: Should we also extract it in a function to clear some space in the component (as it's getting bigger) ?

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
flomonster: We can use `/sprites/signaling_systems`

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
flomonster: Note: In the feature it's specified that the backend should returns a new field `aspect` which is a enum value.
Both `aspect_label` and `color` could be removed :shrug:

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
Akctarus: fixed

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
Akctarus: done

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
Akctarus: No you're right, fixed

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
Akctarus: I got rid of it, fixed

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
Akctarus: I changed it, fixed

front/public/locales/en/simulation.json
```diff
@@ -66,6 +62,17 @@
...
   "stopStopTime": "Stop",
   "stopTime": "Arrival",
+  "timeSpaceChartSettings": {
+    "capacity": "Capacity",
+    "conflicts": "Conflicts",
+    "paths": "Paths",
+    "signals": {
+      "BAL": "BAL",
+      "BAPR": "BAPR",
+      "TVM": "TVM",
```
emersion: These are unused now

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -42,7 +48,22 @@ const ManchetteWithSpaceTimeChartWrapper = ({
...
+
+  const occupancyBlocks = projectPathTrainResult.flatMap((train) => {
+    const departureTime = new Date(train.departure_time).getTime();
+
+    return train.signal_updates.map((block) => ({
+      timeStart: departureTime + block.time_start,
+      timeEnd: departureTime + block.time_end,
+      spaceStart: block.position_start,
+      spaceEnd: block.position_end,
+      color: ASPECT_LABELS_COLORS[block.aspect_label as AspectLabel] || '',
```
emersion: Nit: I'd slightly prefer to assert that `ASPECT_LABELS_COLORS` contains a valid color with `!` instead of `|| ''` because it would be a frontend bug to miss one of these. With `!`, a missing aspect color would trigger an error instead of silently not drawing anything.

(This is optional, if you disagree feel free to ignore!)

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
emersion: Nice, much better!

emersion:
@flomonster's [comment](https://github.com/OpenRailAssociation/osrd/pull/9366#discussion_r1810666238) remains to be fixed: we're still relying on a field meant to be displayed to the user, not something meant to be stable. If we want to merge this quickly, I'd suggest opening a follow-up issue.

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
emersion: There's a single checkbox now so we don't need this anymore.

emersion (approved):
Feel free to resolve remaining comments and merge :)

PR #9367, opened 2024-10-17
front: stdcm: fix stop time default value

### LINKED ISSUE #9354: Stop time in STDCM is automatically set to zero and triggers warning
### What happened?

Image

In case of "relève conducteur", the default stop time should be 3mn so the warning isn't triggered.
In case of "arrêt de service" the default value can remain 0min.
If a stop time is entered, it sould be memorized if the user switches from "arrêt de service" to "relève conducteur" and _vice-versa_.

.

Recette (SNCF)

Google Chrome

e66e534

SarahBellaha:
Closes #9354

### DIFF
--- a/front/src/applications/stdcmV2/components/StdcmInputVia.tsx
+++ b/front/src/applications/stdcmV2/components/StdcmInputVia.tsx
@@ -1,51 +1,50 @@
-import { useState, useMemo, useEffect } from 'react';
+import { useMemo, useEffect, useState } from 'react';
 
 import { Input } from '@osrd-project/ui-core';
 import { debounce } from 'lodash';
 import { useTranslation } from 'react-i18next';
 
 import type { PathStep } from 'reducers/osrdconf/types';
-import { ISO8601Duration2sec } from 'utils/timeManipulation';
+import { ISO8601Duration2sec, secToMin } from 'utils/timeManipulation';
 
 import { StdcmStopTypes } from '../types';
 
-const StdcmInputVia = ({
-  stopType,
-  pathStep,
-  updatePathStepStopTime,
-}: {
-  stopType?: StdcmStopTypes;
-  pathStep: PathStep;
+type StdcmInputViaProps = {
+  stopType: StdcmStopTypes;
+  stopDuration: PathStep['stopFor'];
   updatePathStepStopTime: (stopTime: string) => void;
-}) => {
+};
+
+const StdcmInputVia = ({ stopType, stopDuration, updatePathStepStopTime }: StdcmInputViaProps) => {
   const { t } = useTranslation('stdcm');
 
-  const [pathStepStopTime, setPathStepStopTime] = useState(
-    pathStep.stopFor ? `${ISO8601Duration2sec(pathStep.stopFor) / 60}` : ''
-  );
+  const computedStopTime = useMemo(() => {
+    const duration = stopDuration ? `${secToMin(ISO8601Duration2sec(stopDuration))}` : '';
+    switch (stopType) {
+      case StdcmStopTypes.PASSAGE_TIME:
+        return '0';
+      case StdcmStopTypes.DRIVER_SWITCH:
+        return duration || '3';
+      default:
+        return duration || '0';
+    }
+  }, [stopDuration, stopType]);
 
-  const stopWarning = stopType === StdcmStopTypes.DRIVER_SWITCH && Number(pathStepStopTime) < 3;
+  const [pathStepStopTime, setPathStepStopTime] = useState(computedStopTime);
+
+  const stopWarning = stopType === StdcmStopTypes.DRIVER_SWITCH && Number(computedStopTime) < 3;
 
   const debounceUpdatePathStepStopTime = useMemo(
-    () => debounce((value) => updatePathStepStopTime(value), 500),
+    () => debounce((value) => updatePathStepStopTime(value), 300),
     []
   );
 
   useEffect(() => {
-    let newStopTime = pathStepStopTime;
-    const isPassageTime = stopType === StdcmStopTypes.PASSAGE_TIME || stopType === undefined;
-    if (isPassageTime && pathStepStopTime !== '0') {
-      newStopTime = '0';
-    }
-    if (newStopTime !== pathStepStopTime) {
-      setPathStepStopTime(newStopTime);
-      updatePathStepStopTime(newStopTime);
-    }
-  }, [pathStep.stopFor, stopType]);
+    setPathStepStopTime(computedStopTime);
+  }, [computedStopTime]);
 
   return (
-    stopType !== StdcmStopTypes.PASSAGE_TIME &&
-    stopType !== undefined && (
+    stopType !== StdcmStopTypes.PASSAGE_TIME && (
       <div className="stdcm-v2-via-stop-for stop-time">
         <Input
           id="stdcm-v2-via-stop-time"
--- a/front/src/applications/stdcmV2/components/StdcmVias.tsx
+++ b/front/src/applications/stdcmV2/components/StdcmVias.tsx
@@ -22,6 +22,14 @@ import StdcmStopType from './StdcmStopType';
 import { StdcmStopTypes } from '../types';
 import type { StdcmConfigCardProps } from '../types';
 
+const generateUniqueId = (idsList: string[]): string => {
+  let id;
+  do {
+    id = nextId();
+  } while (idsList.includes(id));
+  return id;
+};
+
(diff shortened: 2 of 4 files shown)


Akctarus (commented):
Tested, the bug is well fixed, however, when you set a stop time (to 10 minutes for example), and reload the page, the stop type is set to “passage stop”, so its stop time is supposed to be zero, but if you click on “Get simulation”, you end up with a passage stop that stops for 10 minutes. 

The stop type needs to be saved when refreshing the page. But I don't know whether this bug should be fixed on this PR, or on a new one.

axrolld:
I agreee that refreshing should not change the via type, it's something we've noiced during the user tests, i was about to open a bug for that. Idk if it should be done in this PR though, as you prefer

front/src/applications/stdcmV2/components/StdcmVias.tsx
```diff
@@ -62,17 +62,12 @@ const StdcmVias = ({ disabled = false, setCurrentSimulationInputs }: StdcmConfig
...
-      updatePathStepStopTime(defaultStopTime, index, pathStepId);
-
-      return updatedStopTypes;
-    });
+    setStopTypes((prevStopTypes) => ({
+      ...prevStopTypes,
+      [pathStepId]: newStopType,
+    }));
+    const defaultStopTime = newStopType === StdcmStopTypes.DRIVER_SWITCH ? '3' : '';
+    updatePathStepStopTime(defaultStopTime, index, pathStepId);
```
SharglutDev, line 77: Nice thank you :)

front/src/applications/stdcmV2/components/StdcmInputVia.tsx
```diff
@@ -9,39 +9,44 @@ import { ISO8601Duration2sec } from 'utils/timeManipulation';
 
 import { StdcmStopTypes } from '../types';
 
+type StdcmInputViaProps = {
+  stopType?: StdcmStopTypes;
+  pathStepStopFor: PathStep['stopFor'];
+  updatePathStepStopTime: (stopTime: string) => void;
+};
```
SharglutDev: ❤️

front/src/applications/stdcmV2/components/StdcmInputVia.tsx
```diff
@@ -9,39 +9,44 @@ import { ISO8601Duration2sec } from 'utils/timeManipulation';
...
+
 const StdcmInputVia = ({
   stopType,
-  pathStep,
+  pathStepStopFor,
```
SharglutDev: As we don't use `pathStepStopType` for the other prop

```suggestion
  stopDuration,
```

front/src/applications/stdcmV2/components/StdcmInputVia.tsx
```diff
@@ -9,39 +9,44 @@ import { ISO8601Duration2sec } from 'utils/timeManipulation';
...
-  pathStep: PathStep;
-  updatePathStepStopTime: (stopTime: string) => void;
-}) => {
+}: StdcmInputViaProps) => {
   const { t } = useTranslation('stdcm');
 
-  const [pathStepStopTime, setPathStepStopTime] = useState(
-    pathStep.stopFor ? `${ISO8601Duration2sec(pathStep.stopFor) / 60}` : ''
+  const computedStopTime = useMemo(
+    () => (pathStepStopFor ? `${ISO8601Duration2sec(pathStepStopFor) / 60}` : ''),
```
SharglutDev: Could you add an util for that ?

```suggestion
    () => (pathStepStopFor ? `${secToMin(ISO8601Duration2sec(pathStepStopFor))}` : ''),
```

front/src/applications/stdcmV2/components/StdcmInputVia.tsx
```diff
@@ -9,39 +9,44 @@ import { ISO8601Duration2sec } from 'utils/timeManipulation';
...
-    () => debounce((value) => updatePathStepStopTime(value), 500),
+    () => debounce((value) => updatePathStepStopTime(value), 300),
     []
   );
 
   useEffect(() => {
-    let newStopTime = pathStepStopTime;
-    const isPassageTime = stopType === StdcmStopTypes.PASSAGE_TIME || stopType === undefined;
-    if (isPassageTime && pathStepStopTime !== '0') {
+    let newStopTime = computedStopTime;
```
SharglutDev: I feel a bit confused now. Since we rely on `computedStopTime` which comes from `pathStep.stopFor` which comes form the store (so we do like a round trip between the form and the store), do we still need the state ? It seems to be used only for the condition `if (newStopTime !== pathStepStopTime)` but why do we need this ?  If we need it why not do the same for the `stopType` ?

SharglutDev (commented):
I think we can fix the bug mentioned by @Akctarus and @axrolld in another PR.

Could you create an issue @axrolld ? :)

front/src/applications/stdcmV2/components/StdcmInputVia.tsx
SarahBellaha: `pathStepStopTime` is used to keep track of user input, as the Input component doesn't handle this internally like other ui-core components do. Otherwise the pathStep will just be updated with the last caracter entered.
`stopType` is just here to determine the default value of the `computedStopTime`, we don't need to do the same for it as we don't need to update it.
I didn't want to spend a lot of time on analysing the component as it was just a small bug fix and not a refacto but i made some additional changes and it also resolved the issue mentionned above by @axrolld & @Akctarus :)

front/src/applications/stdcmV2/components/StdcmInputVia.tsx
```diff
@@ -1,51 +1,54 @@
...
-  );
+  const computedStopTime = useMemo(() => {
+    const duration = stopDuration ? `${secToMin(ISO8601Duration2sec(stopDuration))}` : '';
+    switch (stopType) {
+      case StdcmStopTypes.PASSAGE_TIME:
+        return '0';
+      case StdcmStopTypes.DRIVER_SWITCH:
+        return duration ?? '3';
+      default:
+        return duration ?? '0';
```
SharglutDev: With l.22, `duration` will never be null or undefined so the `??` will never go in the right option.

front/src/applications/stdcmV2/components/StdcmInputVia.tsx
```diff
@@ -1,51 +1,54 @@
...
+  }, [pathStepStopTime]);
+
+  useEffect(() => {
+    setPathStepStopTime(computedStopTime);
+  }, [computedStopTime]);
 
   return (
-    stopType !== StdcmStopTypes.PASSAGE_TIME &&
-    stopType !== undefined && (
+    stopType !== StdcmStopTypes.PASSAGE_TIME && (
```
SharglutDev, line 47: We need to keep the `stopType !== undefined` (or maybe `stopType &&` might be enough) otherwise when adding a new via we have this : 
image

front/src/applications/stdcmV2/components/StdcmInputVia.tsx
```diff
@@ -1,51 +1,54 @@
...
-    if (isPassageTime && pathStepStopTime !== '0') {
-      newStopTime = '0';
-    }
-    if (newStopTime !== pathStepStopTime) {
-      setPathStepStopTime(newStopTime);
-      updatePathStepStopTime(newStopTime);
-    }
-  }, [pathStep.stopFor, stopType]);
+    updatePathStepStopTime(pathStepStopTime);
+  }, [pathStepStopTime]);
```
SharglutDev: Do we really need this useEffect ?

Akctarus (approved):
Tested, thanks also for fixing the added bug :) apart @SharglutDev comments, LGTM

SharglutDev (commented):
Thank you for the changes, everything looks good except this weird behavior in the store : 

https://github.com/user-attachments/assets/aaf0e395-ed31-4a67-ba94-62fa1f1ddd48

SarahBellaha:
> Thank you for the changes, everything looks good except this weird behavior in the store :
> 
>  Enregistrement.de.l.ecran.2024-10-24.a.09.00.28.mov

This behaviour is due to the fact that there are two different ways of updating the pathStep informations. First we update its `stopType`, which is stored in a state, then we update the `stopTime` directly in the store, which then triggers an update of the `pathStep` in the store with the correct `stopType`.
We should change this method and update the pathStep directly in the store in both cases.
This will be done in another PR, as seen with @SharglutDev -> here's the [issue ticket](https://github.com/OpenRailAssociation/osrd/issues/9442).

SharglutDev (approved):
Lgtm and tested, good job !

PR #9373, opened 2024-10-17
editoast: add correlation id on rmq rpc call with dispatch on same channels

ghost:
Remove connection pooling using one connection with many channels.
Dispatching between calls with correlation_id.

Reverting the addition of a connection pooler on rabbitmq.

image

### DIFF
--- a/editoast/src/core/mq_client.rs
+++ b/editoast/src/core/mq_client.rs
@@ -1,25 +1,151 @@
-use deadpool_lapin::{Config, CreatePoolError, Pool, PoolError, Runtime};
+use deadpool::managed::{Manager, Metrics, Pool, RecycleError, RecycleResult};
 use editoast_derive::EditoastError;
 use futures_util::StreamExt;
 use itertools::Itertools;
 use lapin::{
+    message::Delivery,
     options::{BasicConsumeOptions, BasicPublishOptions},
     types::{ByteArray, FieldTable, ShortString},
-    BasicProperties,
+    BasicProperties, Channel, Connection, ConnectionProperties,
 };
 use serde::Serialize;
 use serde_json::to_vec;
-use std::fmt::Debug;
+use std::{collections::HashMap, fmt::Debug, sync::Arc};
 use thiserror::Error;
-use tokio::time::{timeout, Duration};
+use tokio::{
+    sync::{oneshot, RwLock},
+    task,
+    time::{timeout, Duration},
+};
+use uuid::Uuid;
 
 #[derive(Debug, Clone)]
 pub struct RabbitMQClient {
-    pub pool: Pool,
+    pool: Pool<ChannelManager>,
     exchange: String,
     timeout: u64,
-    hostname: String,
     single_worker: bool,
+}
+
+#[derive(Debug, Clone)]
+pub struct ChannelManager {
+    connection: Arc<RwLock<Option<Connection>>>,
+    hostname: String,
+}
+
+impl ChannelManager {
+    pub fn new(connection: Arc<RwLock<Option<Connection>>>, hostname: String) -> Self {
+        ChannelManager {
+            connection,
+            hostname,
+        }
+    }
+}
+
+pub enum ChannelManagerError {
+    Lapin,
+    ConnectionNotFound,
+    BadChannelState,
+}
+
+impl Manager for ChannelManager {
+    type Type = ChannelWorker;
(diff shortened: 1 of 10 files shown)


flomonster (approved):
LGTM :rocket:

clarani (approved):
LGTM

editoast/src/core/mq_client.rs
```diff
@@ -67,28 +197,102 @@ pub struct MQResponse {
...
+            )
+            .await;
+
+            match new_connection {
+                Ok(new_connection) => {
+                    *connection.write().await = Some(new_connection);
+                    tracing::info!("Reconnected to RabbitMQ");
+                }
+                Err(e) => {
+                    tracing::error!("Error while reconnecting to RabbitMQ: {:?}", e);
```
woshilapin, line 283: Should we kill `editoast` after a few retries? Or maybe check the connection to the RabbitMQ in the healthcheck? Because I don't think it makes sense to retry indefinitely.

editoast/src/core/mq_client.rs
```diff
@@ -1,27 +1,158 @@
...
+    pool: Pool<ChannelManager>,
     exchange: String,
     timeout: u64,
-    hostname: String,
     single_worker: bool,
 }
 
+#[derive(Debug, Clone)]
+pub struct ChannelManager {
+    connection: Arc<RwLock<Option<Connection>>>,
```
woshilapin, line 32: I'm not entirely sure that I understand why you need to wrap the `Connection` into a `Option` since in any case, the `Connection` can be in a `Error` state too. How about putting just a `Connection` instead of a `Option<Connection>`?

bloussou (approved):
Tested

editoast/src/core/mq_client.rs
```diff
@@ -67,28 +192,102 @@ pub struct MQResponse {
...
+        let conn = guard.as_ref();
+        let status = match conn {
+            None => return false,
+            Some(conn) => conn.status().state(),
+        };
+        match status {
+            lapin::ConnectionState::Initial => true,
+            lapin::ConnectionState::Connecting => true,
+            lapin::ConnectionState::Connected => true,
+            lapin::ConnectionState::Closing => true,
```
woshilapin, line 251: Looks weird to consider both `Initial`,  `Connecting` and `Closing` as `true`. It feels that either we consider, it will eventually be ok, and in this case, `Closing` should be `false`, or we want to consider the connection can now take some traffic, and in this case, `Initial` and `Connecting` should be false. What do you think?

editoast/src/core/mq_client.rs
```diff
@@ -156,15 +354,22 @@ impl RabbitMQClient {
...
+    ) -> Result<MQResponse, MqClientError>
     where
         T: Serialize,
     {
-        // Create a channel
-        let connection = self.pool.get().await.map_err(Error::DeadpoolLapin)?;
-        let channel = connection.create_channel().await.map_err(Error::Lapin)?;
+        let correlation_id = Uuid::new_v4().to_string();
+
+        // Get the next channel
```
woshilapin, line 363: Should we put from this line until the `drop(channel_work)` into a function (would avoid to manually drop it and it also makes sense as a block of computation? What do you think?

woshilapin (commented):
Some propositions for improvement, I can implement them, but first, do we agree on them (or some of them)?

editoast/src/core/mq_client.rs
ghost: Agree on this (also add to the healthcheck)

editoast/src/core/mq_client.rs
ghost: Agree on this, we should only consider the connected state.

editoast/src/core/mq_client.rs
ghost: It was a legacy from the old way of doing, we can drop it.

editoast/src/core/mq_client.rs
ghost: I'd rather keep the explicit drop to mark the moment we give it back to the pool (this is really important and I do not want this to be on a side effect : keeping it explicit will keep a trace that this it required), wdyt ?

editoast/src/core/mq_client.rs
woshilapin: Improved the healthcheck in #9393.

PR #9375, opened 2024-10-17
front: update timestop table

### LINKED ISSUE #9352: front: new output table
### Description and goal

Relates to https://github.com/osrd-project/osrd-confidential/issues/668

The current output table of the scenario page does not match the design system.
We should create a new one (see the mock-up below):

Image

The scope of this ticket is only to create the new output table (and not the buttons below). The table can not be modified (for the moment).

There are still 2 things to clarify with @thibautsailly: what should we display in the column "Reception sur signal fermé" and which title do we want for the new column that @RomainValls is going to add in his PR (https://github.com/OpenRailAssociation/osrd/pull/9275) (will it change the global display ?) ?

### Acceptance criteria

The new output table is in the scenario page

Synar:
Close #9352 

A few points (name of the CI column, width of each column, text boldness and color in the ch column) remain open in the issue, and will be patched in this PR as soon as they are clarified. All other aspects should be implemented.


Synar:
The last commit hardcode the line breaks and thus split the translations in 2, but I would rather revert this change

clarani (commented):
- Can you set the correct background color ? (behind the table)
- the 3 first columns are not in the right order (see the mockup)

image

front/src/modules/timesStops/ReadOnlyTime.tsx
```diff
@@ -17,7 +17,7 @@ const ReadOnlyTime = ({ rowData }: ReadOnlyTimeProps) => {
     daySinceDeparture !== undefined && dayDisplayed
       ? `${time}${NO_BREAK_SPACE}${t('dayCounter', { count: daySinceDeparture })}`
       : time;
-  return <div className="read-only-time">{fullString}</div>;
+  return <div className="read-only-time text-right w-full">{fullString}</div>;
```
clarani: Can you put all the style in the css ? :) (put `text-right` and `w-full` in `read-only-time`)

Synar: Sure, done ^^

Synar:
Done ^^

front/public/locales/en/timesStops.json
```diff
@@ -1,18 +1,19 @@
...
-  "diffMargins": "Margins diff.",
-  "name": "Name",
+  "departureTime": "requested departure time",
+  "diffMargins": "margins diff. (s)",
+  "name": "CI",
+  "ch": "ch",
   "noPathLoaded": "No path loaded",
-  "realMargin": "Real margin",
-  "receptionOnClosedSignal": "Reception on Closed Signal",
+  "realMargin": "real margin (s)",
```
theocrsb: `"realMargin": "real margin (s.)",`

front/public/locales/en/timesStops.json
```diff
@@ -1,18 +1,19 @@
...
+  "receptionOnClosedSignal": "reception on closed signal",
   "shortSlipDistance": "Short Slip Distance",
-  "stopTime": "Stopping Time (s)",
-  "theoreticalMargin": "Theoretical Margin",
+  "stopTime": "stopping time (s)",
```
theocrsb, line 18: `"stopTime": "stopping time (s.)",`

front/public/locales/en/timesStops.json
```diff
@@ -1,18 +1,19 @@
...
+  "theoreticalMargin": "theoretical margin",
   "theoreticalMarginPlaceholder": "% or min/100km",
-  "theoreticalMarginSeconds": "Theoretical Margin (s)",
+  "theoreticalMarginSeconds": "linear theoretical margin (s)",
```
theocrsb: `"theoreticalMarginSeconds": "linear theoretical margin (s.)",`

front/public/locales/fr/timesStops.json
theocrsb, line 1: Same as for the English file. There are dots after the `(s)`.

front/public/locales/fr/timesStops.json
```diff
@@ -1,18 +1,19 @@
...
   "shortSlipDistance": "Faible glissement",
-  "stopTime": "Temps d'arrêt (s)",
-  "theoreticalMargin": "Marge théorique",
+  "realMargin": "marge réelle (s)",
+  "receptionOnClosedSignal": "réception sur signal fermé",
+  "stopTime": "temps d'arrêt (s)",
+  "theoreticalMargin": "marge théorique",
   "theoreticalMarginPlaceholder": "% ou min/100km",
-  "theoreticalMarginSeconds": "Marge théorique (s)",
+  "theoreticalMarginSeconds": "marge théorique linéaire (s)",
```
theocrsb: "theoreticalMarginSeconds": "marge linéaire théorique (s.)",

theocrsb (changes requested):
Thank you for this PR. I've posted a few comments

front/src/modules/timesStops/styles/_timesStopsDatasheet.scss
```diff
@@ -1,31 +1,61 @@
 .time-stops-datasheet {
+  padding-bottom: 16px;
+  border-radius: 10px;
+  box-shadow: 0 4px 10px rgba(0, 0, 0, 0.1);
```
theocrsb: The shadow box on the mockup is a little different: 
box-shadow: 0px 1px 0px rgba(255, 255, 255, 1) inset, 0px 4px 7px -3px rgba(255, 171, 88, 0.17), 0px 2px 4px rgba(0, 0, 0, 0.22);

theocrsb:
The name of the output table should be added:
image

front/public/locales/en/timesStops.json
Synar: I disagree with this. The SI way to write units like seconds does not take a dot, I don't see why we should differ on this.

front/public/locales/fr/timesStops.json
Synar: Same as above.

Synar:
> The name of the output table should be added

I was not sure whether the name was included in the table and thus the desired changes, but I don't mind adding it

theocrsb:
The font size is 14 everywhere except the header, which has a font size of 12px

theocrsb (changes requested):
The font size is 14 everywhere except the header, which has a font size of 12px

front/public/locales/en/timesStops.json
theocrsb: in the mockup there is a dot : 
image

front/src/modules/timesStops/styles/_timesStopsDatasheet.scss
Synar: Good catch thanks! How did you get the exact box-shadow css? I couldn't find it on the mockup in sketch.

Synar:
> > The name of the output table should be added
> 
> I was not sure whether the name was included in the table and thus the desired changes, but I don't mind adding it

Thibaut is a bit busy right now so I don't want to bother him for that, but for now I have added a title in line with the information and style already on the page

front/public/locales/en/timesStops.json
Synar: This is true, but I believe we should follow the standard unit conventions

Synar:
Good catch on the font size thanks, I did not think to check it

front/public/locales/en/timesStops.json
theocrsb: Perhaps we can ask @thibautsailly for confirmation

front/src/modules/timesStops/styles/_timesStopsDatasheet.scss
theocrsb: right click and `Elevation` :) 
image

front/public/locales/fr/simulation.json
```diff
@@ -67,6 +67,7 @@
   "stopStopTime": "Arrêt",
   "stopTime": "Arrivée",
   "timetable": "Table horaire",
+  "timetableOutput": "Table horaire de sortie",
```
theocrsb, line 73: In the mockup there is:
"timetableOutput": "Table de sortie",

Synar: I know as I said above, but there was already a translation key to "table horaire", the title of the corresponding input section is "horaires et arrets", and I don't know if the title in the mockup was set taking the page into account and not just the component, which the title is not really a part of. Same reason why I put the same style as the title already set above.

There is also a button next to it in the mockup, but it's not implemented, and so I don't think it makes sense to try and replicate this part of the mockup now. I think setting properly all the elements and buttons outside the table and uniformizing the sections in the page should probably be another mockup and pass. This is also why I didn't add any title in the first place.

front/src/modules/timesStops/hooks/useTimeStopsColumns.tsx
theocrsb, line 1: The mockup's widths are not exactly equal to the values used

theocrsb (changes requested):

theocrsb:
I can't see the `CI` anymore when I try to modify a train

front/src/modules/timesStops/hooks/useTimeStopsColumns.tsx
Synar: The mockup width were supposed to be dynamic. Our columns are fixed width. See the issue for this discussion.

theocrsb: mb I didn't see

Synar:
> I can't see the `CI` anymore when I try to modify a train

Good catch thanks, fixed, mb

front/public/locales/en/timesStops.json
Synar: I did, but in the meantime I think we should defer to the BIPM specifications for units :
> Les symboles d’unités sont des entités mathématiques et non des abréviations. Par conséquent, ils ne doivent pas être suivis d’un point, sauf s’ils se trouvent placés à la fin d’une phrase

front/public/locales/fr/timesStops.json
```diff
@@ -1,18 +1,19 @@
...
+  "diffMargins": "diff. marges (s)",
+  "name": "CI",
+  "ch": "CH",
   "noPathLoaded": "Aucun chemin chargé",
-  "realMargin": "Marge réelle",
-  "receptionOnClosedSignal": "Réception sur signal fermé",
-  "shortSlipDistance": "Faible glissement",
-  "stopTime": "Temps d'arrêt (s)",
-  "theoreticalMargin": "Marge théorique",
+  "shortSlipDistance": "faible glissement",
```
clarani: We usually order our keys by alphebetical order (maybe we should add an eslint rule for that...)

front/src/modules/timesStops/styles/_timesStopsDatasheet.scss
```diff
@@ -1,31 +1,79 @@
...
+      font-weight: normal;
     }
   }
 
+  .padded-header > * {
+    padding-left: 8px;
+    padding-right: 16px;
+  }
+
+  .cell-index {
```
clarani: Why cell-index for the classname ? Why not .secondary-code-column ?

front/src/modules/timesStops/TimesStopsOutput.tsx
```diff
@@ -57,9 +57,10 @@ const TimesStopsOutput = ({
         return cx({
           'warning-schedule': arrivalScheduleNotRespected,
           'warning-margin': negativeDiffMargins,
+          'cell-index': columnId === 'ch',
         });
       }}
-      headerRowHeight={65}
+      headerRowHeight={50}
```
clarani: It's 40px on the mockup
```suggestion
      headerRowHeight={40}
```

front/src/modules/timesStops/hooks/useTimeStopsColumns.tsx
```diff
@@ -24,14 +24,14 @@ const timeColumn = (isOutputTable: boolean) =>
...
 
-const fixedWidth = (width: number) => ({ minWidth: width, maxWidth: width });
-
 export const useTimeStopsColumns = <T extends TimeStopsRow>(
   tableType: TableType,
   allWaypoints: T[] = []
 ) => {
   const { t } = useTranslation('timesStops');
 
+  const fixedWidth = (width: number) => ({ minWidth: width, maxWidth: width });
```
clarani: Why did you move this function inside the hook ? I think you can leave it outside, this way it won't be re-instantiated at each execution

front/src/modules/timesStops/hooks/useTimeStopsColumns.tsx
```diff
@@ -41,25 +41,36 @@ export const useTimeStopsColumns = <T extends TimeStopsRow>(
...
               }),
-              ...fixedWidth(120),
+              headerClassName: 'padded-header',
+              ...fixedWidth(90),
             },
             {
-              ...disabledTextColumn('diffMargins', t('diffMargins'), { alignRight: true }),
-              ...fixedWidth(120),
+              ...disabledTextColumn('diffMargins', t('diffMargins'), {
+                alignRight: true,
```
clarani: On the mockup, the margin columns are not aligned on the right but on the left. Same for the arrival column
```suggestion
```

front/src/modules/timesStops/hooks/useTimeStopsColumns.tsx
```diff
@@ -69,30 +80,36 @@ export const useTimeStopsColumns = <T extends TimeStopsRow>(
       {
         ...keyColumn('name', createTextColumn()),
         title: t('name'),
+        ...(isOutputTable && {
+          component: ({ rowData }) => (
+            <span
+              title={rowData.name}
+              className="ml-1 whitespace-nowrap overflow-hidden"
+              style={{ whiteSpace: 'nowrap' }}
```
clarani:
```suggestion
              className="ml-1 text-nowrap overflow-hidden"
```

front/src/modules/timesStops/TimesStopsOutput.tsx
Synar: The issue is that in some languages we can get header names that are too long to fit on 2 lines. On the other hand, we could also make the columns wider to ensure they always fit on two lines regardless of the language. The mockup was based on the idea that the column width would adapt to the headers rather than be fixed.

If we do want to instead go for columns dynamically tightly fitted to the content size, I believe we may need to change the library we use for the tables.

front/src/modules/timesStops/styles/_readOnlyTime.scss
```diff
@@ -1,4 +1,6 @@
 .read-only-time {
   // using px to conform with the padding in the react-datasheet-grid library class .dsg-input
   padding: 0 10px;
+  text-align: right;
```
clarani: I think that the read-only-time column are align on the left on the mockup.
```suggestion
```

front/src/modules/timesStops/hooks/useTimeStopsColumns.tsx
Synar: I agree, but then on the issue thibaut stated that "All columns but index and CI are right aligned.".

I asked for clarification on this (perhaps he was refering to the headers only? But they also seem to be left aligned on the mockup), but he hasn't had time to answer yet.

Synar: image

front/src/modules/timesStops/styles/_readOnlyTime.scss
Synar: Same as above. While we're waiting for an answer from thibaut I don't mind aligning everything on the left though.

front/src/modules/timesStops/hooks/useTimeStopsColumns.tsx
Synar: I don't remember, so sure.

front/src/modules/timesStops/hooks/useTimeStopsColumns.tsx
Synar: Done, thanks!

I am deeply confused about the difference between text-nowrap and whitespace-nowrap though ^^''

front/src/modules/timesStops/styles/_timesStopsDatasheet.scss
Synar: I thought of it as the CH code was acting like a secondary index, like the very first index column, hence their similar color and thus the name.

I am fine with changing it to secondary-code-column though ^^

front/public/locales/fr/timesStops.json
Synar: oh right, shortSlipDistance was added in another commit and I put them in the wrong order when resolving the conflict, I'll fix it ^^

front/public/locales/en/timesStops.json
Synar: Thibaut confirmed the points were not intentional

theocrsb:
Is the size of the `CI` column as expected? 
image

Synar:
> Is the size of the `CI` column as expected? 
> image
> 

I believe so, see thibaut comment on this in the issue. There should be a title tag to let you read the rest by hovering

theocrsb:
Thibaut said ` CI values have a maximum of 35 characters for now`
I think we don't have 35 characters.
https://github.com/OpenRailAssociation/osrd/issues/9352#issuecomment-2416941708

Synar:
Maximum, not minimum

> • **All columns but CI** have a fixed width based on (...)
> • CI values have a **maximum** of 35 characters for now. (...)
> • When the viewport is too small to display the whole CI value, a **title tag** should carry over. (...)

I have reduced the max width of the CI column btw, though this will be pushed later.
I discussed with thibaut and the title and units are fine as is, but we are going to make the durations right aligned (the times will stay left aligned, both for ease of comparison). 
The headers in English, as well as a couple of French ones, will use abbreviations with title tags with the full values in order to both make column narrowers and reduce the height of the header.
Finally we may change the font of the min/100km unit, or otherwise find a way to make this column narrower.

theocrsb (commented):
I don't know if is this PR but we have a weird display in input table : 
image
When we enter 0% we don't see anything displayed, but the array has kept a value.
image

And we see the two points
image

Synar:
The fact that 0% values are hidden was already a hard coded behavior (in 2 different places, so clearly intentional). I don't understand why this is the case. I will discuss this with baptiste as soon as he's back, as well as other questions about margins displaying

Which column is the : with no seconds in?

theocrsb:
> The fact that 0% values are hidden was already a hard coded behavior (in 2 different places, so clearly intentional). I don't understand why this is the case. I will discuss this with baptiste as soon as he's back, as well as other questions about margins displaying
> 
> Which column is the : with no seconds in?

Ok. In `Requested Arrival Time`

Synar:
I will remove the margins changes from this PR, to keep its scope more contained, and add them to another PR later (so until then the 0% will stay hidden, sorry).

I can not reproduce your bug about times being cut. Perhaps the column was simply too narrow? I can change the input table column width to make sure everything fits well.

front/src/modules/timesStops/styles/_timesStopsDatasheet.scss
```diff
@@ -1,31 +1,79 @@
...
+  .warning-schedule {
+    background: var(--warning30);
+  }
+
+  .warning-margin {
+    background: rgba(var(--warning30-rgb), 0.5);
+  }
+
+  .dsg-cell {
+    font-size: 14px;
```
theocrsb: This font-size is not applied everywhere
image

theocrsb (changes requested):
I don't think we want to change the style of the input panel. (which would solve the problem of unseen seconds).

front/src/modules/timesStops/TimesStopsOutput.tsx
Synar: We now use abbreviations (with title tags for the full header text) for headers that are too long. The header height was thus adjusted back to 40px.

front/src/modules/timesStops/hooks/useTimeStopsColumns.tsx
Synar: After discussion with thibaut, durations should be right-aligned and times should be left-aligned for ease of comparison. Text columns will also be left-aligned by default.

front/src/modules/timesStops/styles/_timesStopsDatasheet.scss
Synar: Fixed. thanks

front/src/modules/timesStops/TimesStopsOutput.tsx
clarani: Ok, thanks for the explanations :pray:

clarani: Ah hum it's still 50px

front/src/modules/timesStops/styles/_timesStopsDatasheet.scss
```diff
@@ -1,31 +1,79 @@
...
+  --dsg-header-active-text-color: var(--black100);
+
   .activeRow {
-    font-weight: bold;
-    background-color: var(--coolgray1);
-    --dsg-cell-background-color: var(--coolgray1);
-    --dsg-cell-disabled-background-color: var(--coolgray1);
-    .dsg-cell {
-      font-weight: bold;
+    font-weight: semibold;
```
clarani: With `semibold`, I don't see any change :/ I think you can use 600 for now, we will have to add new vars in the css
```suggestion
    font-weight: 600;
```

front/src/modules/timesStops/styles/_timesStopsDatasheet.scss
```diff
@@ -1,31 +1,79 @@
...
+    0px 1px 0px rgba(255, 255, 255, 1) inset,
+    0px 4px 7px -3px rgba(255, 171, 88, 0.17),
+    0px 2px 4px rgba(0, 0, 0, 0.22);
+  background-color: var(--ambiantB15);
+  --dsg-cell-background-color: var(--white100);
+  --dsg-cell-disabled-background-color: var(--white100);
+  --dsg-header-text-color: var(--grey50);
+  --dsg-header-active-text-color: var(--black100);
+
   .activeRow {
```
clarani: Do you think you can encapsulate the css concerning a row in the class dsg-row  ? Like this:
```
.dsg-row {
    &.oddRow {
      --dsg-cell-background-color: var(--ambiantB5);
      --dsg-cell-disabled-background-color: var(--ambiantB5);
    }

    &.activeRow {
      font-weight: 600;
      color: var(--black100);
      .dsg-cell-gutter,
      .secondary-code-column {
        font-weight: normal;
      }
    }
  }
```

front/src/modules/timesStops/styles/_timesStopsDatasheet.scss
```diff
@@ -1,31 +1,79 @@
...
+    color: var(--black100);
+    .dsg-cell-gutter,
+    .secondary-code-column {
+      font-weight: normal;
     }
   }
 
+  .padded-header > * {
+    padding-left: 8px;
+    padding-right: 16px;
```
clarani: You can use padding-inline:
```suggestion
    padding-inline: 8px 16px;
```

front/src/modules/timesStops/styles/_timesStopsDatasheet.scss
clarani: Can you use rem instead of px for the font-size ? :) 
```suggestion
    font-size: .875rem;
```

front/src/modules/timesStops/styles/_timesStopsDatasheet.scss
```diff
@@ -1,31 +1,79 @@
...
+  }
+
+  .dsg-cell {
+    font-size: 14px;
+    border-top: none !important;
+    border-bottom: none !important;
+  }
+
+  .dsg-cell.dsg-cell-header {
+    font-size: 12px;
```
clarani: Same here
```suggestion
    font-size: .75rem;
```

front/src/modules/timesStops/styles/_timesStopsDatasheet.scss
Synar: > Ah hum it's still 50px

The last batch of changes isn't pushed yet

> With semibold, I don't see any change

sure

front/src/modules/timesStops/styles/_timesStopsDatasheet.scss
Synar: Sure thanks!

front/src/modules/timesStops/styles/_timesStopsDatasheet.scss
Synar: Sure!

front/src/modules/timesStops/styles/_timesStopsDatasheet.scss
Synar: Done

front/src/modules/timesStops/styles/_timesStopsDatasheet.scss
Synar: Sure thanks

front/public/locales/en/timesStops.json
```diff
@@ -1,18 +1,23 @@
...
+  "arrivalTime": "requested arrival time",
+  "calculatedArrivalTime": "calculated arrival time",
+  "calculatedDepartureTime": "calculated depart. time",
+  "calculatedDepartureTimeFull": "calculated departure time",
   "dayCounter": "D+{{count}}",
-  "departureTime": "Requested departure Time",
-  "diffMargins": "Margins diff.",
-  "name": "Name",
+  "departureTime": "requested depart. time",
+  "departureTimeFull": "requested depart. time",
```
clarani:
```suggestion
  "departureTimeFull": "requested departure time",
```

clarani (commented):
Almost good:
- can you make the column CI take all the remaining space ? So the table fills the whole page ?
image

- can you make the input table fill the whole space too in its tab ?
image

- can you rebase and solve the conflicts ?

theocrsb (approved):
LGTM and tested! Thank you for this PR

front/public/locales/en/timesStops.json
Synar: Oh my bad!

Synar: All done ^^

clarani (approved):
LGTM :white_check_mark:

PR #9381, opened 2024-10-17
editoast: update temporary speed limit groups import endpoint

### LINKED ISSUE #9399: Update editoast LTV endpoint to take signals instead of track sections
Currently, the temporary speed limits endpoint expects to receive the list of track ranges associates with a given temporary speed limit. Since this information does not exist as is, modify the import endpoint to get the list of signals (execution / resume) associated with the speed limits and retrieve in Editoast the list of track ranges between the temporary speed limit signals.

Notes:

- Some speed limits might not have a valid end signal on all directions. Stop adding track sections if the distance since the execution signal becomes too great (meaning that there might be a missing end signal for the temporary speed limit on the path being explored).
- Some speed limits do not end with a resume signal but with another temporary speed limit execution signal.

Sh099078:
fixes https://github.com/OpenRailAssociation/osrd/issues/9399

- [x] Add an endpoint `/infra/{infra_id}/delimited_area` that returns the list of track sections inside an area delimited by entry and exit directed locations (i.e. a position on a track and a direction, similar to a light version of a sign with only the relevant positional information).
- [x] Stop the graph exploration at a specified distance from the entries if an end position is missing. It is considered that and end position is missing if none has been found after exploring the track sections on a specified maximum distance from the entries without having found any exit.
- [x]  Refactorize `temporary_speed_limit.rs` tests.

### DIFF
--- a/editoast/src/views/infra/delimited_area.rs
+++ b/editoast/src/views/infra/delimited_area.rs
@@ -1,0 +1,929 @@
+use crate::error::Result;
+use crate::infra_cache::{Graph, InfraCache};
+use crate::models::Infra;
+use crate::views::infra::{InfraApiError, InfraIdParam};
+use crate::views::{AuthenticationExt, AuthorizationError};
+use crate::AppState;
+use crate::Retrieve;
+use axum::extract::{Path, State};
+use axum::{Extension, Json};
+use editoast_authz::BuiltinRole;
+use editoast_derive::EditoastError;
+use editoast_schemas::{
+    infra::{Direction, DirectionalTrackRange, Endpoint, Sign, TrackEndpoint},
+    primitives::Identifier,
+};
+use itertools::{Either, Itertools};
+use serde::{Deserialize, Serialize};
+use std::{
+    cmp::Ordering,
+    collections::{HashMap, HashSet},
+    result::Result as StdResult,
+};
+use thiserror::Error;
+use utoipa::ToSchema;
+
+crate::routes! {
+    "/delimited_area" => delimited_area,
+}
+
+editoast_common::schemas! {
+    DelimitedAreaResponse,
+    DirectedLocation,
+    InputError,
+}
+
+// Maximum distance the graph can be explored from a speed limit execution signal
+// without finding any legitimate ending to the speed limit before it is considered
+// there is not valid limit on the portion of the graph that is being explored.
+// TODO Magic number for now. Make it configurable ?
+const MAXIMUM_DISTANCE: f64 = 5000.;
+
+#[derive(Deserialize, ToSchema)]
+struct DelimitedAreaForm {
+    #[schema(inline)]
+    entries: Vec<DirectedLocation>,
+    #[schema(inline)]
+    exits: Vec<DirectedLocation>,
+}
+
+#[derive(Deserialize, Serialize, ToSchema)]
+struct DelimitedAreaResponse {
+    track_ranges: Vec<DirectionalTrackRange>,
+}
+
+#[derive(Debug, Deserialize, Serialize, ToSchema)]
+struct DirectedLocation {
+    #[schema(inline)]
+    track: Identifier,
+    position: f64,
+    direction: Direction,
+}
+
+#[derive(Debug, Error, Serialize, Deserialize, EditoastError)]
+#[editoast_error(base_id = "delimited_area")]
+enum DelimitedAreaError {
+    #[error("Some locations were invalid")]
+    #[editoast_error(status = 400)]
+    InvalidLocations {
+        invalid_locations: Vec<(DirectedLocation, InputError)>,
+    },
+}
+
+#[derive(Debug, Error, Serialize, Deserialize, ToSchema)]
+enum InputError {
+    #[error("Track '{0}' does not exist")]
+    TrackDoesNotExist(String),
+    #[error("Invalid input position '{position}' on track '{track}' of length '{track_length}'")]
+    LocationOutOfBounds {
+        track: String,
+        position: f64,
+        track_length: f64,
+    },
+}
+
+#[derive(Debug, Error, Serialize, Deserialize)]
+enum TrackRangeConstructionError {
+    #[error("Track identifiers do not match")]
+    TrackIdentifierMissmatch,
+    #[error(
+        "The input directions or locations on track do not allow to build a valid track range"
+    )]
+    InvalidRelativeLocations,
+    #[error("The location is not on track")]
+    LocationNotOnTrack,
+}
+
+impl From<Sign> for DirectedLocation {
+    fn from(value: Sign) -> Self {
+        let Sign {
+            track,
+            position,
+            direction,
+            ..
+        } = value;
+        DirectedLocation {
+            track,
+            position,
+            direction,
+        }
+    }
+}
+
+#[utoipa::path(
+    get, path = "",
+    tag = "delimited_area",
+    params(InfraIdParam),
+    request_body = inline(DelimitedAreaResponse),
+    responses(
+        (status = 200, body = inline(DelimitedAreaResponse), description = "The track ranges between a list entries and exits." ),
+    )
+)]
+/// Computes all tracks between a set of `entries` locations and a set of `exits` locations
+///
+/// Returns any track between one of the `entries` and one of the `exits`, i.e. any track that can
+/// be reached from an entry before reaching an exit.
+/// To prevent a missing exit to cause the graph traversal to never stop exploring, the exploration
+/// stops when a maximum distance is reached and no exit has been found.
+async fn delimited_area(
+    Extension(auth): AuthenticationExt,
+    State(AppState {
+        infra_caches,
+        db_pool_v2: db_pool,
+        ..
+    }): State<AppState>,
+    Path(InfraIdParam { infra_id }): Path<InfraIdParam>,
+    Json(DelimitedAreaForm { entries, exits }): Json<DelimitedAreaForm>,
+) -> Result<Json<DelimitedAreaResponse>> {
+    // TODO in case of a missing exit, return an empty list of track ranges instead of returning all
+    // the track ranges explored until the stopping condition ?
+    let authorized = auth
+        .check_roles([BuiltinRole::InfraWrite].into())
+        .await
+        .map_err(AuthorizationError::AuthError)?;
+    if !authorized {
+        return Err(AuthorizationError::Unauthorized.into());
+    }
+
+    // Retrieve the infra
+
+    let conn = &mut db_pool.get().await?;
+    let infra =
+        Infra::retrieve_or_fail(conn, infra_id, || InfraApiError::NotFound { infra_id }).await?;
+    let infra_cache = InfraCache::get_or_load(conn, &infra_caches, &infra).await?;
+    let graph = Graph::load(&infra_cache);
+
+    // Validate user input
+
+    let (valid_entries, invalid_entries): (Vec<_>, Vec<_>) =
+        entries
+            .into_iter()
+            .partition_map(|entry| match validate_location(&entry, &infra_cache) {
+                Ok(_) => Either::Left(entry),
+                Err(e) => Either::Right((entry, e)),
+            });
+    let (valid_exits, invalid_exits): (Vec<_>, Vec<_>) =
+        exits
+            .into_iter()
+            .partition_map(|exit| match validate_location(&exit, &infra_cache) {
+                Ok(_) => Either::Left(exit),
+                Err(e) => Either::Right((exit, e)),
+            });
+
+    if !(invalid_exits.is_empty() && invalid_entries.is_empty()) {
+        let invalid_locations = invalid_entries
+            .into_iter()
+            .chain(invalid_exits.into_iter())
+            .collect::<Vec<_>>();
+        return Err(DelimitedAreaError::InvalidLocations { invalid_locations }.into());
+    }
+
+    // Retrieve the track ranges
+
+    Ok(Json(DelimitedAreaResponse {
+        track_ranges: track_ranges_from_locations(valid_entries, valid_exits, &graph, &infra_cache),
+    }))
+}
+
+/// Check whether a location is valid on a given infra cache, i.e. if it matches a track on the infra
+/// and if it is within bounds.
+fn validate_location(
+    location: &DirectedLocation,
+    infra_cache: &InfraCache,
+) -> StdResult<(), InputError> {
+    // Check if the location track exists on the infra
+
+    let track_length = infra_cache
+        .track_sections()
+        .get(&location.track.0)
+        .ok_or(InputError::TrackDoesNotExist(location.track.0.clone()))?
+        .unwrap_track_section()
+        .length;
+
+    // Check if the location is within bounds on the track
+
+    if location.position < 0. || track_length < location.position {
+        Err(InputError::LocationOutOfBounds {
+            track: location.track.0.clone(),
+            position: location.position,
+            track_length,
+        })
+    } else {
+        Ok(())
+    }
+}
+
+fn track_ranges_from_locations(
+    entries: Vec<DirectedLocation>,
+    exits: Vec<DirectedLocation>,
+    graph: &Graph,
+    infra_cache: &InfraCache,
+) -> Vec<DirectionalTrackRange> {
+    entries
+        .iter()
+        .flat_map(|entry| {
+            impacted_tracks(
+                entry,
+                exits.iter().collect::<Vec<&DirectedLocation>>(),
+                graph,
+                infra_cache,
+                MAXIMUM_DISTANCE,
+            )
+        })
+        .collect::<Vec<_>>()
+}
+
+fn impacted_tracks(
+    entry: &DirectedLocation,
+    exits: Vec<&DirectedLocation>,
+    graph: &Graph,
+    infra_cache: &InfraCache,
+    max_distance: f64,
+) -> Vec<DirectionalTrackRange> {
+    // Map track identifiers to their list of associated exits:
+    let exits = {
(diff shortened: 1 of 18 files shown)


flomonster (commented):
I've just seen the issue. I don't really like the idea of having several endpoints to create the same object. I'd really prefer this to be done on the import side and not in editoast.
When importing, do you have information on the "line" "track" and "pk" of the start and end of the temporary speed limit zone?

> [!NOTE]
> Using the InfraCache is really not ideal for an endpoint. It makes the endpoint unscalable because it's stateful. If you keep the endpoint this way, you'll need to modify the Gateway configuration in prod (to redirect to editoast-statefull).

Khoyo:
> I've just seen the issue. I don't really like the idea of having several endpoints to create the same object. 

wdym? there is only `/temporary_speed_limit_group`

> I'd really prefer this to be done on the import side and not in editoast. When importing, do you have information on the "line" "track" and "pk" of the start and end of the temporary speed limit zone?

We have signals and their location, so we could get that, yes. The import side does not have the full track graph with the various corrections (eg. we do not run the kd-tree), and can't run them easily since it lacks track geometry.

flomonster:
> > I've just seen the issue. I don't really like the idea of having several endpoints to create the same object.
> 
> wdym? there is only `/temporary_speed_limit_group`

I mean that we should keep the CRUD endpoints of an object quite simple and close from it's model. Usualy transformation are done by the user (using utility endpoint or by it's own).
Example: Train schedule / Workschedule / Infra / RS...

Since we don't have LRS in OSRD (for now) it's really difficult to import linear objects such as speed limits (all of them). However, all these objects are converted from a start point to an end point into a list of track ranges in the import script and not by editoast. This PR goes against this paradigm, hence my caution.

> > I'd really prefer this to be done on the import side and not in editoast. When importing, do you have information on the "line" "track" and "pk" of the start and end of the temporary speed limit zone?
> 
> We have signals and their location, so we could get that, yes. The import side does not have the full track graph with the various corrections (eg. we do not run the kd-tree), and can't run them easily since it lacks track geometry.

Maybe editoast can help with this, but with a more generic and reusable endpoint :shrug:

Sh099078:
Following discussions with @Khoyo and @flomonster  the temporary speed limits group endpoint has been split into two:
- A temporary speed limit groups creation endpoint (unchanged, accepts an LTV with a list of track sections).
- A `/infra/{infra_id}/delimited_area` endpoint that returns the list of track sections inside an area delimited by a list of entries and exits (positions on a track with a direction).

editoast/src/views/temporary_speed_limits.rs
```diff
@@ -186,33 +190,114 @@ mod tests {
...
+                    }
+            ))
+        }
+    }
+
+    struct RequestParameters {
+        group_name: String,
+        obj_id: String,
+        time_period: TimePeriod,
+        track_ranges: Value,
```
leovalais: We have a `DirectionalTrackRange` in `editoast_schemas` which can type this value I think.

editoast/src/views/temporary_speed_limits.rs
```diff
@@ -186,33 +190,114 @@ mod tests {
...
+            let RequestParameters {
+                group_name,
+                obj_id,
+                time_period:
+                    TimePeriod {
+                        start_date_time,
+                        end_date_time,
+                    },
+                track_ranges,
+            } = parameters;
```
leovalais: You can destructure in the argument list if you want.

editoast/src/views/infra/delimited_area.rs
```diff
@@ -0,0 +1,837 @@
...
+struct DelimitedAreaResponse {
+    #[schema(inline)]
+    track_ranges: Vec<DirectionalTrackRange>,
+}
+
+#[derive(Debug, Deserialize, ToSchema)]
+struct DirectedLocation {
+    pub track: Identifier,
+    pub position: f64,
+    pub direction: Direction,
```
leovalais:
```suggestion
    track: Identifier,
    position: f64,
    direction: Direction,
```

editoast/src/views/infra/delimited_area.rs
```diff
@@ -0,0 +1,837 @@
...
+    params(InfraIdParam),
+    request_body = inline(DelimitedAreaResponse),
+    responses(
+        (status = 200, body = inline(DelimitedAreaResponse), description = "The track ranges between a list entries and exits." ),
+    )
+)]
+async fn delimited_area(
+    Extension(auth): AuthenticationExt,
+    State(db_pool): State<DbConnectionPoolV2>,
+    State(app_state): State<AppState>,
```
leovalais:
```suggestion
    State(AppState { infra_caches, db_pool_v2: db_pool }): State<AppState>,
```

suggestion

editoast/src/views/infra/delimited_area.rs
```diff
@@ -0,0 +1,837 @@
...
+        } = value;
+        DirectedLocation {
+            track,
+            position,
+            direction,
+        }
+    }
+}
+
+#[utoipa::path(
```
leovalais, line 113: We could add a description of the endpoint with `///`

editoast/src/views/infra/delimited_area.rs
```diff
@@ -0,0 +1,837 @@
...
+    entries.iter().for_each(|entry| {
+        track_ranges.extend(impacted_tracks(
+            entry,
+            exits.iter().collect::<Vec<&DirectedLocation>>(),
+            graph,
+            infra_cache,
+            MAXIMUM_DISTANCE,
+        ))
+    });
+    track_ranges
```
leovalais: You can use `entries.iter().flatmap`

editoast/src/views/infra/delimited_area.rs
```diff
@@ -0,0 +1,837 @@
...
+    exits.into_iter().for_each(|exit| {
+        let track_id = &exit.track;
+        if let Some(track_exits) = tracks_to_exits.get_mut(track_id) {
+            track_exits.push(exit);
+        } else {
+            let track_exits = vec![exit];
+            tracks_to_exits.insert(track_id, track_exits);
+        };
+    });
+    let exits = tracks_to_exits;
```
leovalais: You can use scopes to make steps more evident:

```suggestion
	let exits = {
		let mut tracks_to_exits = HashMap::new();
		for exit in exits {
			tracks_to_exits.entry(track_id).or_default().push(exit);
		}
		tracks_to_exits
	};
```

editoast/src/views/infra/delimited_area.rs
leovalais: Also better move the `collect` out of the loop.

editoast/src/views/infra/delimited_area.rs
```diff
@@ -0,0 +1,837 @@
...
+                exit,
+                curr_track_endpoint,
+                infra_cache,
+                false,
+            )
+            .unwrap_or_else(|| {
+                panic!(
+                    "Cannot form a track from {:?} and {:?}",
+                    exit, curr_track_endpoint
+                )
```
leovalais: This should probably be a 4xx error since I think this case arise when users provide wrong input data.

editoast/src/views/infra/delimited_area.rs
```diff
@@ -0,0 +1,837 @@
...
+            });
+            related_track_ranges.push(track_range);
+        } else {
+            let track_range =
+                track_range_from_endpoint(curr_track_endpoint, remaining_distance, infra_cache)
+                    .unwrap_or_else(|| {
+                        panic!("Cannot form a track from {:?}", curr_track_endpoint)
```
leovalais: Likewise

editoast/src/views/infra/delimited_area.rs
```diff
@@ -0,0 +1,837 @@
...
+                |e_1, e_2| match (track_endpoint.endpoint, e_1.position < e_2.position) {
+                    (Endpoint::Begin, true) | (Endpoint::End, false) => Ordering::Less,
+                    (Endpoint::Begin, false) | (Endpoint::End, true) => Ordering::Greater,
+                },
+            )
+            .map(|exit| &**exit)
+            .next()
+    } else {
+        None
+    }
```
leovalais: You can use the `?` operator to avoid the `if`

editoast/src/views/infra/delimited_area.rs
```diff
@@ -0,0 +1,837 @@
...
+}
+
+/// Return the directional track range starting at `entry` finishing at `exit`, or `None`
+/// if no track range can be built from them.
+fn track_range_between_two_locations(
+    entry: &DirectedLocation,
+    exit: &DirectedLocation,
+) -> Option<DirectionalTrackRange> {
```
leovalais:
```suggestion
) -> Result<DirectionalTrackRange> {
```

editoast/src/views/infra/delimited_area.rs
```diff
@@ -0,0 +1,837 @@
...
+}
+
+/// Return the directional track range delimited by a location and a track endpoint, or `None` if no
+/// track range can be built from them.
+fn track_range_between_endpoint_and_location(
+    location: &DirectedLocation,
+    endpoint: &TrackEndpoint,
+    infra_cache: &InfraCache,
+    entry: bool,
+) -> Option<DirectionalTrackRange> {
```
leovalais:
```suggestion
) -> Result<DirectionalTrackRange> {
```

editoast/src/views/infra/delimited_area.rs
```diff
@@ -0,0 +1,837 @@
...
+struct DelimitedAreaForm {
+    #[schema(inline)]
+    entries: Vec<DirectedLocation>,
+    #[schema(inline)]
+    exits: Vec<DirectedLocation>,
+}
+
+#[derive(Deserialize, Serialize, ToSchema)]
+struct DelimitedAreaResponse {
+    #[schema(inline)]
```
leovalais:
```suggestion
```

It's already in the `schemas`, no need to inline it

leovalais (commented):
Great PR! Not much to say on the process apart from the points we discussed orally a while ago, but that's for another PR, if any.

I believe a few `panic!`s should be turned into proper errors, and there is probably a few ways to linearize to code using `Option<>`'s functional interface and the `?` operator.

editoast/src/views/temporary_speed_limits.rs
Sh099078: Thank you, I don't have that reflex yet

editoast/src/views/infra/delimited_area.rs
Sh099078: I added a description, does it seem clear to you ? Also, don't you think it would be better to change the behavior so that if the missing exit signal stop condition happens, we do not return any of the track ranges reached from the entry signal that is being explored ? I added a TODO on that matter

editoast/src/views/infra/delimited_area.rs
Sh099078: At this point, we have already retrieved the track range that corresponds to the entry signal given by the user. The track ranges being explored are retrieved from the loaded infra graph. If we cannot form a track range inside the main loop, it is either caused by a bug in the implementation or by invalid data in the loaded graph. In both cases, wouldn't a 5xx error be more appropriate ?

editoast/src/views/infra/delimited_area.rs
Sh099078: Same question as in https://github.com/OpenRailAssociation/osrd/pull/9381#discussion_r1837041927

editoast/src/views/infra/delimited_area.rs
leovalais: Ok I get it now. Panicking is indeed fine in this case. But then we're never checking that all tracks provided in input do exist. Maybe iterate over both input track arrays and check that every track is in the cache. This way 

1. we validate user input, making the API bit more ergonomic
2. we know it's fine to assert after this point

wdyt?

kmer2016 (approved):
LGTM for front side :)

editoast/src/views/infra/delimited_area.rs
Sh099078: I think it is great. I updated the API so that it returns a list of invalid user inputs and their associated error in case of failure and kept the asserts in the graph exploration. Some utility methods were updated so that all asserts happen at the same level: directly in the main `impacted_tracks` method instead of having some failures like trying to build a track with invalid values panic in the utility functions that build the track ranges (for example, if the infra cache did not contain a matching track for an identifier).

Overall we have now:
* A `DelimitedAreaError` that handles http responses in case of invalid user input.
* A `TrackRangeConstructionError` that is used to catch internal errors which have nothing to do with the API.
* panics in the main graph exploration loop when falling on a supposedly impossible situation (bug in the implementation or incorrect data in the cache).

What do you think ?

editoast/src/views/infra/delimited_area.rs
Sh099078: This signature got affected by the changes mentioned in https://github.com/OpenRailAssociation/osrd/pull/9381#discussion_r1848511107. I updated the signature to use `std::result::Result` instead of `editoast_error::Result` since these are internal errors that will cause panics and never return to the endpoint handler. What do you think ?

editoast/src/views/infra/delimited_area.rs
Sh099078: Same question as https://github.com/OpenRailAssociation/osrd/pull/9381#discussion_r1848518649

editoast/src/views/infra/delimited_area.rs
```diff
@@ -0,0 +1,934 @@
...
+    params(InfraIdParam),
+    request_body = inline(DelimitedAreaResponse),
+    responses(
+        (status = 200, body = inline(DelimitedAreaResponse), description = "The track ranges between a list entries and exits." ),
+    )
+)]
+/// Return any track between one of the `entries` and one of the `exits`, i.e. any track that can
+/// be reached from an entry before reaching an exit.
+/// To prevent a missing exit to cause the graph traversal to never stop exploring, the exploration
+/// stops when a maximum distance is reached and no exit has been found.
```
leovalais:
```suggestion
/// Computes all tracks between a set of `entries` locations and a set of `exits` locations
///
/// Returns any track between one of the `entries` and one of the `exits`, i.e. any track that can
/// be reached from an entry before reaching an exit.
/// To prevent a missing exit to cause the graph traversal to never stop exploring, the exploration
/// stops when a maximum distance is reached and no exit has been found.
```

Rust documentation follows this scheme:

```rust
/// One-line summary
/// 
/// Multi-line documenta-
/// -tion.
fn my_def() -> lolz
```

`utoipa` interprets this string to determine which text to assign to `description` or `summary`. If you look at the generated OpenApi you'll see the split point is a bit broken ^^

editoast/src/views/infra/delimited_area.rs
leovalais: I agree that ideally we should be conservative in the function and only return results which we know for sure are correct. In the case of LTVs though, I guess this has to be double-checked with a PO because less speed limits may be worse than too many 🤷

editoast/src/views/infra/delimited_area.rs
```diff
@@ -0,0 +1,934 @@
...
+        let mut tracks_to_exits: HashMap<&Identifier, Vec<&DirectedLocation>> = HashMap::new();
+        exits.into_iter().for_each(|exit| {
+            let track_id = &exit.track;
+            if let Some(track_exits) = tracks_to_exits.get_mut(track_id) {
+                track_exits.push(exit);
+            } else {
+                let track_exits = vec![exit];
+                tracks_to_exits.insert(track_id, track_exits);
+            };
+        });
```
leovalais: [`entry`](https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#method.entry) may be useful here:

```suggestion
		for exit in exits {
            tracks_to_exits.entry(&exit.track).or_default().push(exit);
        }
```

editoast/src/views/infra/delimited_area.rs
```diff
@@ -0,0 +1,934 @@
...
+                curr_track_endpoint,
+                infra_cache,
+                false,
+            )
+            .expect("Failed to build track range");
+            related_track_ranges.push(track_range);
+        } else {
+            let track_range =
+                track_range_from_endpoint(curr_track_endpoint, remaining_distance, infra_cache)
+                    .expect("Failed to build track range");
```
leovalais: I believe that `curr_track_endpoint` gets pulled from the infra cache three times here. Not that it matters performance wise (I don't think so at least 🤔), but you can factorize that part by `get`ting it once and make these functions accept a `track: &infra_cache::TrackSectionCache`. That would simplify the error management a bit. Wdyt?

leovalais (approved):
LGTM. Sorry about the time I took to review this. I hope that didn't block rest of the implementation fo the feature. The last batch of comments / responses:

editoast/src/views/infra/delimited_area.rs
leovalais: Looks to me like the best thing to do here. Since `EditoastError` will be deprecated "soon", let's avoid using it in places where we can do something else.

editoast/src/views/infra/delimited_area.rs
leovalais: That's great. I believe it's much more readable like this :)

editoast/src/views/infra/delimited_area.rs
Sh099078: Agreed, it simplifies a bit the error management. Updated in https://github.com/OpenRailAssociation/osrd/pull/9381/commits/78da1db5bafb4a0ed3c316acdfadc8421f7df0f3

editoast/src/views/infra/delimited_area.rs
Sh099078: I checked with @Arthur-Lefebvre, we'll keep it the way it currently is for now: building abnormally big LTVs might be easier to spot and fix later on than the opposite

PR #9382, opened 2024-10-18
osrdyne: refresh worker on version changes

### LINKED ISSUE #9369: Osrdyne should clean and update kubernetes ressources
Currently, we can have the deployment that end up in inconsistent states, mainly related to upgrading issue not updating the core since it doesn't .

When starting up osrdyne should :
- Cleanup the stale ScaledObjects if the deployment is not required
- Upgrade the deployments charts

In order to retrigger a core deployment at boot properly, we should:
- Select a random string in the Kubernetes Driver at boot
- Set this string in a label of the deployment
- When starting up, the deployment should be upgraded (triggering a reload of the containers) 

(This mechanism is inspired from the `rollme` system that is sometimes used in charts)

ghost:
Closes #9369

### DIFF
--- a/osrdyne/src/drivers/docker.rs
+++ b/osrdyne/src/drivers/docker.rs
@@ -9,13 +9,15 @@ use bollard::{
     Docker,
 };
 use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
 use uuid::Uuid;
 
 use crate::Key;
 
 use super::{
     worker_driver::{DriverError, WorkerDriver, WorkerMetadata},
-    LABEL_MANAGED_BY, LABEL_WORKER_ID, LABEL_WORKER_KEY, MANAGED_BY_VALUE,
+    LABEL_MANAGED_BY, LABEL_VERSION_IDENTIFIER, LABEL_WORKER_ID, LABEL_WORKER_KEY,
+    MANAGED_BY_VALUE,
 };
 
 #[derive(Serialize, Deserialize, Clone, Debug)]
@@ -39,6 +41,7 @@ pub struct DockerDriver {
     options: DockerDriverOptions,
     amqp_uri: String,
     worker_pool: String,
+    version_identifier: String,
 }
 
 impl DockerDriver {
@@ -47,11 +50,17 @@ impl DockerDriver {
         amqp_uri: String,
         worker_pool: String,
     ) -> DockerDriver {
+        let version_identifier = std::env::var("OSRD_GIT_DESCRIBE")
+            .unwrap_or_else(|_| format!("run-{}", Uuid::new_v4()));
+
+        let hashed = format!("{:x}", Sha256::digest(version_identifier))[..16].to_string();
+
         DockerDriver {
             client: Docker::connect_with_socket_defaults().expect("Failed to connect to Docker"),
             options,
             amqp_uri,
             worker_pool,
+            version_identifier: hashed,
         }
     }
 }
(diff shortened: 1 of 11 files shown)


flomonster (approved):
LGTM (not tested)

bloussou (approved):
~Tested~

ElysaSrc:
> Tested

Not tested. No the right one.

osrdyne/src/drivers/docker.rs
```diff
@@ -47,11 +50,17 @@ impl DockerDriver {
         amqp_uri: String,
         worker_pool: String,
     ) -> DockerDriver {
+        let version_identifier = std::env::var("OSRD_GIT_DESCRIBE")
+            .unwrap_or_else(|_| format!("run-{}", Uuid::new_v4()));
+
+        let hashed = format!("{:x}", Sha256::digest(version_identifier))[..16].to_string();
```
Khoyo, line 56: Why only the first 16 bytes?

osrdyne/src/drivers/kubernetes/keda.rs
```diff
@@ -44,6 +44,6 @@ pub enum MetricType {
 pub struct Trigger {
     pub type_: String,
     pub metadata: HashMap<String, String>,
-    pub use_cached_metrics: bool,
+    pub use_cached_metrics: Option<bool>,
```
woshilapin, line 47: Why do we need that to be optional now? Can't understand why, and especially, I don't see the link with the current PR.

osrdyne/src/drivers/docker.rs
```diff
@@ -47,11 +50,17 @@ impl DockerDriver {
         amqp_uri: String,
         worker_pool: String,
     ) -> DockerDriver {
+        let version_identifier = std::env::var("OSRD_GIT_DESCRIBE")
+            .unwrap_or_else(|_| format!("run-{}", Uuid::new_v4()));
```
Khoyo, line 54: This should work for now, and this PR is time critical, so feel free to ignore it.

I'm not a fan of relying on an env var for this. Maybe we should embed a generated uuid at build time?

osrdyne/src/drivers/docker.rs
ghost: Because there is a capacity limit on labels on k8s.

osrdyne/src/drivers/kubernetes/keda.rs
ghost: It is related to this PR because we deserialize objects that does not have the field and we fail badly (meaning : osrdyne crash).

osrdyne/src/drivers/docker.rs
ghost: Noted, we will change this in a future PR.

ghost:
Fully working on our deployments.

osrdyne/src/drivers/worker_driver.rs
```diff
@@ -56,6 +60,13 @@ pub trait WorkerDriver: Send {
         worker_key: Key,
     ) -> Pin<Box<dyn Future<Output = Result<(), DriverError>> + Send + '_>>;
 
+    /// This methoid should check if current deployments are up to date and refresh them if needed.
+    /// NOTE: This method should be called periodically to ensure that the worker deployments are up to date.
+    /// It should not modify the identifiers of the workers in anyway.
```
Khoyo: Why? That seems wrong

Khoyo (approved):

osrdyne/src/drivers/worker_driver.rs
ghost: That's mainly due to the fact that some part of a deployment are immutable in kubernetes and basically the update gets rejected if we touch those parts of the deployment. Yes, that's dirty. Maybe we can think of something better when revamping the driver layer in the future ? WDYT ?

osrdyne/src/drivers/worker_driver.rs
```diff
@@ -56,6 +60,13 @@ pub trait WorkerDriver: Send {
         worker_key: Key,
     ) -> Pin<Box<dyn Future<Output = Result<(), DriverError>> + Send + '_>>;
 
+    /// This methoid should check if current deployments are up to date and refresh them if needed.
```
SergeCroise:
```suggestion
    /// This method should check if current deployments are up to date and refresh them if needed.
```
see https://www.merriam-webster.com/dictionary/methoid

osrdyne/src/drivers/worker_driver.rs
SergeCroise:
```suggestion
    /// It should not modify the identifiers of the workers in any way.
```

PR #9383, opened 2024-10-18
stdcm: towed rolling stock parameter

Wadjetz:
part of https://github.com/osrd-project/osrd-confidential/issues/368

- [x] Add towed rolling stock param
- [x] Compute physical rolling stock
- [x] Unit Tests

### DIFF
--- a/editoast/editoast_schemas/src/rolling_stock/gamma.rs
+++ b/editoast/editoast_schemas/src/rolling_stock/gamma.rs
@@ -12,7 +12,7 @@ editoast_common::schemas! {
 #[derivative(Hash)]
 pub struct Gamma {
     #[serde(rename = "type")]
-    gamma_type: String,
+    pub gamma_type: String,
     #[derivative(Hash(hash_with = "editoast_common::hash_float::<3,_>"))]
-    value: f64,
+    pub value: f64,
 }
--- a/editoast/editoast_schemas/src/rolling_stock/rolling_resistance.rs
+++ b/editoast/editoast_schemas/src/rolling_stock/rolling_resistance.rs
@@ -13,11 +13,14 @@ editoast_common::schemas! {
 #[allow(non_snake_case)]
 pub struct RollingResistance {
     #[serde(rename = "type")]
-    rolling_resistance_type: String,
+    pub rolling_resistance_type: String,
+    /// Solid friction in kN
     #[derivative(Hash(hash_with = "editoast_common::hash_float::<5,_>"))]
-    A: f64,
+    pub A: f64,
+    /// Viscosity friction in kN/(km/h)
     #[derivative(Hash(hash_with = "editoast_common::hash_float::<5,_>"))]
-    B: f64,
+    pub B: f64,
+    /// Aerodynamic drag in kN/(km/h)²
     #[derivative(Hash(hash_with = "editoast_common::hash_float::<5,_>"))]
-    C: f64,
+    pub C: f64,
 }
--- a/editoast/src/core/simulation.rs
+++ b/editoast/src/core/simulation.rs
@@ -5,6 +5,7 @@ use editoast_schemas::rolling_stock::EffortCurves;
 use editoast_schemas::rolling_stock::Gamma;
 use editoast_schemas::rolling_stock::RollingResistance;
 use editoast_schemas::rolling_stock::RollingStock;
+use editoast_schemas::rolling_stock::TowedRollingStock;
 use editoast_schemas::train_schedule::Comfort;
 use editoast_schemas::train_schedule::Distribution;
 use editoast_schemas::train_schedule::MarginValue;
@@ -35,7 +36,7 @@ editoast_common::schemas! {
 
 #[derive(Debug, Serialize, Derivative)]
 #[derivative(Hash)]
-pub struct PhysicsRollingStock {
+pub struct PhysicsConsist {
     pub effort_curves: EffortCurves,
     pub base_power_class: Option<String>,
     /// Length of the rolling stock in mm
@@ -46,8 +47,10 @@ pub struct PhysicsRollingStock {
     // Time in ms
     pub startup_time: u64,
     #[derivative(Hash(hash_with = "editoast_common::hash_float::<5,_>"))]
+    /// In m/s²
     pub startup_acceleration: f64,
     #[derivative(Hash(hash_with = "editoast_common::hash_float::<5,_>"))]
+    /// In m/s²
     pub comfort_acceleration: f64,
     pub gamma: Gamma,
     #[derivative(Hash(hash_with = "editoast_common::hash_float::<5,_>"))]
@@ -66,29 +69,153 @@ pub struct PhysicsRollingStock {
     pub raise_pantograph_time: Option<u64>,
 }
 
-#[derive(Debug, Default)]
-pub struct SimulationParameters {
+#[derive(Debug, Clone)]
+pub struct PhysicsConsistParameters {
+    /// In kg
     pub total_mass: Option<f64>,
+    /// In m
     pub total_length: Option<f64>,
+    /// In m/s
     pub max_speed: Option<f64>,
+    pub towed_rolling_stock: Option<TowedRollingStock>,
+    pub traction_engine: RollingStock,
 }
 
-impl PhysicsRollingStock {
-    pub fn new(traction_engine: RollingStock, params: SimulationParameters) -> Self {
-        let traction_engine_length = traction_engine.length * 1000.0;
-        let length = params
+impl PhysicsConsistParameters {
+    pub fn with_traction_engine(traction_engine: RollingStock) -> Self {
+        PhysicsConsistParameters {
+            max_speed: None,
+            total_length: None,
+            total_mass: None,
+            towed_rolling_stock: None,
+            traction_engine,
+        }
+    }
+}
+
+impl PhysicsConsistParameters {
+    pub fn compute_length(&self) -> u64 {
+        let towed_rolling_stock_length = self
+            .towed_rolling_stock
+            .as_ref()
+            .map(|trs| trs.length)
+            .unwrap_or(0.0);
+
+        let length = self
             .total_length
-            .map(|tl| tl * 1000.0)
-            .unwrap_or(traction_engine_length)
-            .round() as u64;
+            .unwrap_or(self.traction_engine.length + towed_rolling_stock_length);
 
-        let traction_engine_mass = traction_engine.mass;
-        let mass = params.total_mass.unwrap_or(traction_engine_mass).round() as u64;
+        // Convert m to mm
+        let length = length * 1000.0;
 
-        let max_speed = f64::min(
-            traction_engine.max_speed,
-            params.max_speed.unwrap_or(traction_engine.max_speed),
-        );
+        length.round() as u64
+    }
+
+    pub fn compute_max_speed(&self) -> f64 {
+        self.max_speed
+            .map(|max_speed_parameter| {
+                f64::min(self.traction_engine.max_speed, max_speed_parameter)
+            })
+            .unwrap_or(self.traction_engine.max_speed)
+    }
+
+    pub fn compute_startup_acceleration(&self) -> f64 {
+        self.towed_rolling_stock
+            .as_ref()
+            .map(|towed_rolling_stock| {
+                f64::max(
+                    self.traction_engine.startup_acceleration,
+                    towed_rolling_stock.startup_acceleration,
+                )
+            })
+            .unwrap_or(self.traction_engine.startup_acceleration)
+    }
+
+    pub fn compute_comfort_acceleration(&self) -> f64 {
+        self.towed_rolling_stock
+            .as_ref()
+            .map(|towed_rolling_stock| {
+                f64::min(
+                    self.traction_engine.comfort_acceleration,
+                    towed_rolling_stock.comfort_acceleration,
+                )
+            })
+            .unwrap_or(self.traction_engine.comfort_acceleration)
+    }
+
+    pub fn compute_inertia_coefficient(&self) -> f64 {
+        if let (Some(towed_rolling_stock), Some(total_mass)) =
+            (self.towed_rolling_stock.as_ref(), self.total_mass)
+        {
+            let mass_carriage = total_mass - self.traction_engine.mass;
+            let traction_engine_inertia =
+                self.traction_engine.mass * self.traction_engine.inertia_coefficient;
+            let towed_inertia = mass_carriage * towed_rolling_stock.inertia_coefficient;
+            (traction_engine_inertia + towed_inertia) / total_mass
+        } else {
+            self.traction_engine.inertia_coefficient
+        }
+    }
+
+    pub fn compute_mass(&self) -> u64 {
+        let traction_engine_mass = self.traction_engine.mass;
+        let towed_rolling_stock_mass = self
+            .towed_rolling_stock
+            .as_ref()
+            .map(|trs| trs.mass)
+            .unwrap_or(0.0);
+        let mass = self
+            .total_mass
+            .unwrap_or(traction_engine_mass + towed_rolling_stock_mass);
+        mass.round() as u64
+    }
+
+    pub fn compute_rolling_resistance(&self) -> RollingResistance {
+        if let (Some(towed_rolling_stock), Some(total_mass)) =
+            (self.towed_rolling_stock.as_ref(), self.total_mass)
+        {
+            let traction_engine_rr = &self.traction_engine.rolling_resistance;
+            let towed_rs_rr = &towed_rolling_stock.rolling_resistance;
+            let traction_engine_mass = self.traction_engine.mass; // kg
+
+            let carriage_mass = total_mass - traction_engine_mass; // kg
+
+            let traction_engine_solid_friction_a = traction_engine_rr.A * 1000.0; // convert from kN to N
+            let traction_engine_viscosity_friction_b = traction_engine_rr.B * 1000.0 * 3.6; // convert from kN/(km/h) to N/(m/s)
+            let traction_engine_aerodynamic_drag_c = traction_engine_rr.C * 1000.0 * 3.6 * 3.6; // convert from kN/(km/h)² to N/(m/s)²
+
+            let towed_solid_friction_a = towed_rs_rr.A * 1e-2 * carriage_mass; // convert from daN/t to N
+            let towed_viscosity_friction_b = towed_rs_rr.B * 1e-2 * carriage_mass * 3.6; // convert from (daN/t)/(km/h) to N/(m/s)
+            let towed_aerodynamic_drag_c = towed_rs_rr.C * 1e-2 * carriage_mass * 3.6 * 3.6; // convert from (daN/t)/(km/h)² to N/(m/s)²
+
+            let solid_friction_a = traction_engine_solid_friction_a + towed_solid_friction_a;
+            let viscosity_friction_b =
+                traction_engine_viscosity_friction_b + towed_viscosity_friction_b;
+            let aerodynamic_drag_c = traction_engine_aerodynamic_drag_c + towed_aerodynamic_drag_c;
+
+            RollingResistance {
+                rolling_resistance_type: traction_engine_rr.rolling_resistance_type.clone(),
+                A: solid_friction_a,
+                B: viscosity_friction_b,
+                C: aerodynamic_drag_c,
+            }
+        } else {
+            self.traction_engine.rolling_resistance.clone()
+        }
+    }
+}
+
+impl From<PhysicsConsistParameters> for PhysicsConsist {
+    fn from(params: PhysicsConsistParameters) -> Self {
+        let length = params.compute_length();
+        let max_speed = params.compute_max_speed();
+        let startup_acceleration = params.compute_startup_acceleration();
+        let comfort_acceleration = params.compute_comfort_acceleration();
+        let inertia_coefficient = params.compute_inertia_coefficient();
+        let mass = params.compute_mass();
+        let rolling_resistance = params.compute_rolling_resistance();
+
+        let traction_engine = params.traction_engine;
 
         Self {
             effort_curves: traction_engine.effort_curves,
@@ -97,11 +224,11 @@ impl PhysicsRollingStock {
             mass,
             max_speed,
             startup_time: (traction_engine.startup_time * 1000.0).round() as u64,
-            startup_acceleration: traction_engine.startup_acceleration,
-            comfort_acceleration: traction_engine.comfort_acceleration,
+            startup_acceleration,
+            comfort_acceleration,
             gamma: traction_engine.gamma,
-            inertia_coefficient: traction_engine.inertia_coefficient,
-            rolling_resistance: traction_engine.rolling_resistance,
+            inertia_coefficient,
+            rolling_resistance,
             power_restrictions: traction_engine.power_restrictions.into_iter().collect(),
             electrical_power_startup_time: traction_engine
                 .electrical_power_startup_time
@@ -297,7 +424,7 @@ pub struct SimulationRequest {
     pub speed_limit_tag: Option<String>,
     pub power_restrictions: Vec<SimulationPowerRestrictionItem>,
     pub options: TrainScheduleOptions,
-    pub rolling_stock: PhysicsRollingStock,
+    pub rolling_stock: PhysicsConsist,
     pub electrical_profile_set_id: Option<i64>,
 }
 
@@ -346,5 +473,135 @@ impl AsCoreRequest<Json<SimulationResponse>> for SimulationRequest {
 
     fn infra_id(&self) -> Option<i64> {
         Some(self.infra)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use editoast_schemas::rolling_stock::RollingResistance;
+    use pretty_assertions::assert_eq;
+
+    use crate::models::fixtures::create_simple_rolling_stock;
+    use crate::models::fixtures::create_towed_rolling_stock;
+
+    use super::PhysicsConsistParameters;
+
+    fn create_physics_consist() -> PhysicsConsistParameters {
+        PhysicsConsistParameters {
+            total_length: Some(100.0),
+            total_mass: Some(50000.0),
+            max_speed: Some(22.0),
+            towed_rolling_stock: Some(create_towed_rolling_stock()),
+            traction_engine: create_simple_rolling_stock(),
+        }
+    }
+
+    #[test]
+    fn physics_consist_compute_length() {
+        let mut physics_consist = create_physics_consist();
+        physics_consist.total_length = Some(100.0); // m
+        physics_consist.traction_engine.length = 40.0; // m
+
+        // We always take total_length
+        assert_eq!(physics_consist.compute_length(), 100000);
+
+        physics_consist.total_length = None;
+        // When no total_length we take towed length + traction_engine length
+        assert_eq!(physics_consist.compute_length(), 70000);
+
+        physics_consist.total_length = None;
+        physics_consist.towed_rolling_stock = None;
+        // When no user specified length and towed rolling stock, we take traction_engine length
+        assert_eq!(physics_consist.compute_length(), 40000);
+    }
+
+    #[test]
+    fn physics_consist_compute_mass() {
+        let mut physics_consist = create_physics_consist();
+        physics_consist.total_mass = Some(50000.0); // kg
+        physics_consist.traction_engine.mass = 15000.0; // kg
+
+        // We always take total_mass
+        assert_eq!(physics_consist.compute_mass(), 50000);
+
+        physics_consist.total_mass = None;
+        // When no total_mass we take towed mass + traction_engine mass
+        assert_eq!(physics_consist.compute_mass(), 65000);
+
+        physics_consist.total_mass = None;
+        physics_consist.towed_rolling_stock = None;
+        // When no user specified mass and towed rolling stock, we take traction_engine mass
+        assert_eq!(physics_consist.compute_mass(), 15000);
+    }
+
+    #[test]
+    fn physics_consist_max_speed() {
+        let mut physics_consist = create_physics_consist();
+        physics_consist.max_speed = Some(20.0); // m/s
+        physics_consist.traction_engine.max_speed = 22.0; // m/s
+
+        // We take the smallest max speed
+        assert_eq!(physics_consist.compute_max_speed(), 20.0);
+
+        physics_consist.max_speed = Some(25.0); // m/s
+        physics_consist.traction_engine.max_speed = 24.0; // m/s
+
+        assert_eq!(physics_consist.compute_max_speed(), 24.0);
+
+        physics_consist.max_speed = None;
+        assert_eq!(physics_consist.compute_max_speed(), 24.0);
+    }
+
+    #[test]
+    fn physics_consist_compute_startup_acceleration() {
+        let mut physics_consist = create_physics_consist(); // 0.06
+
+        // We take the biggest
+        assert_eq!(physics_consist.compute_startup_acceleration(), 0.06);
+
+        physics_consist.towed_rolling_stock = None;
+        assert_eq!(physics_consist.compute_startup_acceleration(), 0.04);
+    }
+
+    #[test]
+    fn physics_consist_compute_comfort_acceleration() {
+        let mut physics_consist = create_physics_consist(); // 0.2
+
+        // We take the smallest
+        assert_eq!(physics_consist.compute_comfort_acceleration(), 0.1);
+
+        physics_consist.towed_rolling_stock = None;
+        assert_eq!(physics_consist.compute_comfort_acceleration(), 0.1);
+    }
+
+    #[test]
+    fn physics_consist_compute_inertia_coefficient() {
+        let mut physics_consist = create_physics_consist();
+
+        assert_eq!(physics_consist.compute_inertia_coefficient(), 1.065);
+
+        physics_consist.towed_rolling_stock = None;
+        assert_eq!(physics_consist.compute_inertia_coefficient(), 1.10,);
+    }
+
+    #[test]
+    fn physics_consist_compute_rolling_resistance() {
+        let mut physics_consist = create_physics_consist();
+
+        assert_eq!(
+            physics_consist.compute_rolling_resistance(),
+            RollingResistance {
+                rolling_resistance_type: "davis".to_string(),
+                A: 1350.0,
+                B: 48.6,
+                C: 7.387200000000001
+            }
+        );
+
+        physics_consist.towed_rolling_stock = None;
+        assert_eq!(
+            physics_consist.compute_rolling_resistance(),
+            physics_consist.traction_engine.rolling_resistance,
+        );
     }
 }
--- a/editoast/src/models/fixtures.rs
+++ b/editoast/src/models/fixtures.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::io::Cursor;
 use std::ops::DerefMut;
 
@@ -10,7 +11,13 @@ use editoast_schemas::infra::DirectionalTrackRange;
 use editoast_schemas::infra::InfraObject;
 use editoast_schemas::infra::RailJson;
 use editoast_schemas::primitives::OSRDObject;
+use editoast_schemas::rolling_stock::EffortCurves;
+use editoast_schemas::rolling_stock::Gamma;
+use editoast_schemas::rolling_stock::LoadingGaugeType;
+use editoast_schemas::rolling_stock::RollingResistance;
 use editoast_schemas::rolling_stock::RollingStock;
+use editoast_schemas::rolling_stock::RollingStockSupportedSignalingSystems;
+use editoast_schemas::rolling_stock::TowedRollingStock;
 use editoast_schemas::train_schedule::TrainScheduleBase;
 use postgis_diesel::types::LineString;
 use serde_json::Value;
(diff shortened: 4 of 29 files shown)


editoast/src/core/simulation.rs
```diff
@@ -68,40 +69,63 @@ pub struct PhysicsRollingStock {
...
     pub max_speed: Option<f64>,
 }
 
 impl PhysicsRollingStock {
-    pub fn new(traction_engine: RollingStock, params: SimulationParameters) -> Self {
+    pub fn new(
+        traction_engine: RollingStock,
+        towed_rolling_stock: Option<TowedRollingStock>,
+        params: SimulationParameters,
+    ) -> Self {
```
woshilapin: How about putting both `RollingStock` and `Option<TowedRollingStock>` directly in `SimulationParameters`? I don't see a compelling reason to separate them from `mass`, `length` and `max_speed`.

editoast/src/core/simulation.rs
```diff
@@ -68,40 +69,63 @@ pub struct PhysicsRollingStock {
 
 #[derive(Debug, Default)]
 pub struct SimulationParameters {
+    /// In kg
     pub total_mass: Option<f64>,
     pub total_length: Option<f64>,
     pub max_speed: Option<f64>,
```
woshilapin: Can you put a comment on the unit of `length` and `max_speed`.

editoast/src/core/simulation.rs
```diff
@@ -68,40 +69,63 @@ pub struct PhysicsRollingStock {
...
 }
 
 impl PhysicsRollingStock {
```
woshilapin: Maybe this need to be rename into `PhysicsConsist`? Please do that in a separate commit.

editoast/src/core/simulation.rs
```diff
@@ -68,40 +69,63 @@ pub struct PhysicsRollingStock {
...
+            .unwrap_or_else(|| traction_engine_length + towed_rolling_stock_length)
             .round() as u64;
 
         let traction_engine_mass = traction_engine.mass;
-        let mass = params.total_mass.unwrap_or(traction_engine_mass).round() as u64;
+        let towed_rolling_stock_mass = towed_rolling_stock.as_ref().map(|trs| trs.mass);
+        let mass = params
+            .total_mass
+            .unwrap_or_else(|| traction_engine_mass + towed_rolling_stock_mass.unwrap_or(0.0))
+            .round() as u64;
```
woshilapin:
```suggestion
        let towed_rolling_stock_mass = towed_rolling_stock.as_ref().map(|trs| trs.mass).unwrap_or(0.0);
        let mass = params
            .total_mass
            .unwrap_or_else(|| traction_engine_mass + towed_rolling_stock_mass)
            .round() as u64;
```

editoast/src/core/simulation.rs
```diff
@@ -113,6 +137,84 @@ impl PhysicsRollingStock {
     }
 }
 
+fn compute_rolling_resistance(
```
woshilapin: nitpick: can you order the `compute_*` functions in the same order than the order of fields in the `struct`, might just be more logical.

editoast/src/core/simulation.rs
```diff
@@ -113,6 +137,84 @@ impl PhysicsRollingStock {
...
+    traction_engine: &RollingStock,
+    towed_rolling_stock: &Option<TowedRollingStock>,
+    total_mass: Option<f64>,
+) -> RollingResistance {
+    if let (Some(towed_rolling_stock), Some(total_mass)) = (towed_rolling_stock, total_mass) {
```
woshilapin: Functions with `Option` arguments are weird: either you call the function if you have all parameters, or you don't call it if you don't have all the information. I see 2 different solutions.

Either you keep this `if` logic  in the caller. Another possibility is to send `SimulationParameter` as an input argument, if you follow the other comment I've made about putting `RollingStock` and `TowedRollingStock` inside the `SimulationParameter`. In this latter case, `compute_rolling_resistance` (and all other `compute_*` functions) can be functions of `impl SimulationParameters`. What do you think?

editoast/src/models/towed_rolling_stock.rs
```diff
@@ -21,7 +21,9 @@ pub struct TowedRollingStockModel {
     pub railjson_version: String,
     pub locked: bool,
 
+    /// In kg
     pub mass: f64,
+    /// In m
     pub length: f64,
     pub comfort_acceleration: f64,
     pub startup_acceleration: f64,
```
woshilapin, line 29: If possible, can you also add the unit for the other fields?

editoast/src/views/timetable/stdcm.rs
```diff
@@ -74,8 +75,14 @@ enum STDCMError {
     TimetableNotFound { timetable_id: i64 },
     #[error("Rolling stock {rolling_stock_id} does not exist")]
     RollingStockNotFound { rolling_stock_id: i64 },
+    #[error("Towed rolling stock {towed_rolling_stock_id} does not exist")]
+    TowedRollingStockNotFound { towed_rolling_stock_id: i64 },
     #[error("Path items are invalid")]
     InvalidPathItems { items: Vec<InvalidPathItem> },
+    #[error(
+        "Invalid entry: total mass and length are required when towed rolling stock is provided."
```
woshilapin: `mass` and ` length` are optional. So I believe this error should not exist.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -120,6 +127,7 @@ pub struct STDCMRequestPayload {
     total_length: Option<f64>,
     /// Maximum speed of the consist in km/h
     max_speed: Option<f64>,
+    towed_rolling_stock_id: Option<i64>,
```
woshilapin: How about moving that field next to `rolling_stock_id`?

editoast/src/views/timetable/stdcm.rs
```diff
@@ -130,6 +138,12 @@ impl STDCMRequestPayload {
             max_speed: self.max_speed,
         }
     }
+
+    pub fn is_consist_invalid(&self) -> bool {
+        self.towed_rolling_stock_id.is_some()
+            && self.total_mass.is_none()
+            && self.total_length.is_none()
```
woshilapin: total mass and total length are optional, even when a towed rolling stock is selected. I agree that in practice, user will often (always?) fill up these fields, but we don't want to enforce it.

editoast/src/core/simulation.rs
```diff
@@ -113,6 +137,84 @@ impl PhysicsRollingStock {
...
+        let traction_engine_rr = &traction_engine.rolling_resistance;
+        let towed_rs_rr = &towed_rolling_stock.rolling_resistance;
+        let traction_engine_mass = traction_engine.mass; // kg
+
+        let mass_carriage = total_mass - traction_engine_mass; // kg
+
+        let rav_a_te = traction_engine_rr.A * 1000.0; // N
+        let rav_b_te = traction_engine_rr.B * 1000.0 * 3.6; // N/(m/s)
+        let rav_c_te = traction_engine_rr.C * 1000.0 * 3.6 * 3.6; // N/(m/s)²
```
axrolld:
```suggestion
        let rav_a_te = traction_engine_rr.A * 1000.0; // convert from kN to N
        let rav_b_te = traction_engine_rr.B * 1000.0 * 3.6; // convert from kN/(km/h) to N/(m/s)
        let rav_c_te = traction_engine_rr.C * 1000.0 * 3.6 * 3.6; // convert from kN/(km/h)² to N/(m/s)²
```

I suggest specifying the convertion, not only the final unit.

editoast/src/core/simulation.rs
```diff
@@ -113,6 +137,84 @@ impl PhysicsRollingStock {
...
+
+        let rav_a_towed = towed_rs_rr.A * 1e-2 * mass_carriage; // N
+        let rav_b_towed = towed_rs_rr.B * 1e-2 * mass_carriage * 3.6; // N/(m/s)
+        let rav_c_towed = towed_rs_rr.C * 1e-2 * mass_carriage * 3.6 * 3.6; // N/(m/s)²
```
axrolld: Same here
```suggestion
        let rav_a_towed = towed_rs_rr.A * 1e-2 * mass_carriage; // convert from daN/t to  N
        let rav_b_towed = towed_rs_rr.B * 1e-2 * mass_carriage * 3.6; // convert from (daN/t)/(km/h) to N/(m/s)
        let rav_c_towed = towed_rs_rr.C * 1e-2 * mass_carriage * 3.6 * 3.6; // convert from (daN/t)/(km/h)² to N/(m/s)²
```

editoast/src/views/timetable/stdcm.rs
```diff
@@ -664,20 +700,35 @@ enum SimulationTimeResult {
...
     use pretty_assertions::assert_eq;
 
     use crate::views::rolling_stock::tests::fast_rolling_stock_form;
 
     use super::*;
 
-    fn create_test_rolling_stock() -> RollingStockModel {
+    pub fn create_towed_rolling_stock() -> TowedRollingStock {
+        TowedRollingStock {
+            name: "FAKE VOITURE CORAIL".to_string(),
```
axrolld: I wouldn't specify the name here

front/src/applications/stdcmV2/components/StdcmConsist.tsx
```diff
@@ -126,6 +177,7 @@ const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
           type="number"
           min={0}
           value={totalMass ?? ''}
+          required={!!selectedTowedRollingStock}
```
woshilapin: You can remove it, it should stay optional.

front/src/applications/stdcmV2/components/StdcmConsist.tsx
```diff
@@ -135,6 +187,7 @@ const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
           type="number"
           min={0}
           value={totalLength ?? ''}
+          required={!!selectedTowedRollingStock}
```
woshilapin: You can remove it, it should stay optional.

woshilapin (commented):
A first pass of review.

axrolld (changes requested):

editoast/src/core/simulation.rs
Wadjetz: Yes, I had trouble finding a good solution. The fields don’t require the same optional values, but I’ll come up with something based on your suggestion

hamz2a (approved):
LGTM, thanks!

editoast/editoast_schemas/src/rolling_stock/gamma.rs
```diff
@@ -16,3 +16,9 @@ pub struct Gamma {
     #[derivative(Hash(hash_with = "editoast_common::hash_float::<3,_>"))]
     value: f64,
 }
+
+impl Gamma {
+    pub fn new(gamma_type: String, value: f64) -> Self {
+        Self { gamma_type, value }
+    }
+}
```
leovalais:
```suggestion
```

It's a schema, it's fine having its fields `pub`.

editoast/editoast_schemas/src/rolling_stock/rolling_resistance.rs
```diff
@@ -13,11 +13,25 @@ editoast_common::schemas! {
...
+impl RollingResistance {
+    pub fn new(rolling_resistance_type: String, a: f64, b: f64, c: f64) -> Self {
+        Self {
+            rolling_resistance_type,
+            A: a,
+            B: b,
+            C: c,
+        }
+    }
 }
```
leovalais:
```suggestion
```

The function does nothing other than forwarding its parameters to the struct initialization form. The fields being `pub` already, we can use the structure initialization directly.

editoast/src/core/simulation.rs
```diff
@@ -66,29 +69,135 @@ pub struct PhysicsRollingStock {
...
+            let rav_a = rav_a_te + rav_a_towed;
+            let rav_b = rav_b_te + rav_b_towed;
+            let rav_c = rav_c_te + rav_c_towed;
+
+            RollingResistance::new(
+                traction_engine_rr.rolling_resistance_type.clone(),
+                rav_a,
+                rav_b,
+                rav_c,
+            )
```
leovalais: What does `rav` stands for? A more explicit prefix or some documentation on `RollingResistance` would help.

editoast/src/core/simulation.rs
```diff
@@ -66,29 +69,135 @@ pub struct PhysicsRollingStock {
...
+    pub fn compute_comfort_acceleration(&self, traction_engine: &RollingStock) -> f64 {
+        if let Some(towed_rolling_stock) = self.towed_rolling_stock.as_ref() {
+            f64::min(
+                traction_engine.comfort_acceleration,
+                towed_rolling_stock.comfort_acceleration,
+            )
+        } else {
+            traction_engine.comfort_acceleration
+        }
+    }
```
leovalais: The functional style is mostly used in this file, can these be rewritten in this style for consistency?

editoast/src/models/fixtures.rs
```diff
@@ -214,6 +221,46 @@ pub async fn create_rolling_stock_livery(
         .expect("Failed to create rolling stock livery")
 }
 
+pub fn create_towed_rolling_stock() -> TowedRollingStock {
+    TowedRollingStock {
+        name: "TOWED ROLLING STOCK".to_string(),
+        mass: 50000_f64,
+        length: 30_f64, // m
```
leovalais:
```suggestion
        mass: 50000.0,
        length: 30.0, // m
```

editoast/src/views/timetable/stdcm.rs
```diff
@@ -213,6 +221,23 @@ async fn stdcm(
...
+                TowedRollingStockModel::retrieve_or_fail(conn, towed_rolling_stock_id, || {
+                    STDCMError::TowedRollingStockNotFound {
+                        towed_rolling_stock_id,
+                    }
+                })
+                .await?;
+            Some(towed_rolling_stock)
+        } else {
+            None
+        };
```
leovalais, line 235: I believe `map` and [`transpose`](https://doc.rust-lang.org/std/option/enum.Option.html#method.transpose) can be used here.

front/public/locales/en/stdcm.json
```diff
@@ -7,7 +7,8 @@
     "length": "Length",
     "tonnage": "Tonnage",
     "maxSpeed": "Max speed",
-    "tractionEngine": "Traction engine"
+    "tractionEngine": "Traction engine",
+    "towedRollingStock": "Matériel remorqué"
```
leovalais:
```suggestion
    "towedRollingStock": "Towed rolling stock"
```

editoast/src/models/fixtures.rs
```diff
@@ -214,6 +221,46 @@ pub async fn create_rolling_stock_livery(
...
+        inertia_coefficient: 1.05,
+        rolling_resistance: RollingResistance::new("davis".to_string(), 1.0, 0.01, 0.0002),
+        gamma: Gamma::new("CONST".to_string(), 1.0),
+        railjson_version: "3.4".to_string(),
+    }
+}
+
+pub fn create_simple_rolling_stock() -> RollingStock {
+    RollingStock {
+        name: "ROLLING_STOCK_NAME".to_string(),
```
leovalais:
```suggestion
        name: "SIMPLE_ROLLING_STOCK".to_string(),
```

leovalais (changes requested):
Wow se really need `uom` or something similar 😨, thanks for documenting the units in most places, it helps a lot!

I opened an issue about that yesterday https://github.com/OpenRailAssociation/osrd/issues/9431 (ping @woshilapin). The `serde` compatibility might not be that much of an issue after all I believe :)

A few comments otherwise:

editoast/src/core/simulation.rs
woshilapin, line 1: It would be nice to have unit tests for all the `compute_*` functions in this file, especially testing some edge cases (presence or absence of `TowedRollingStock`, or `mass`, or `length`? What is the behavior with negative values?)

editoast/src/core/simulation.rs
```diff
@@ -67,52 +69,135 @@ pub struct PhysicsRollingStock {
...
-        let comfort_acceleration =
-            compute_comfort_acceleration(&traction_engine, &towed_rolling_stock);
-        let startup_acceleration =
-            compute_startup_acceleration(&traction_engine, &towed_rolling_stock);
-        let inertia_coefficient =
-            compute_inertia_coefficient(&traction_engine, &towed_rolling_stock, params.total_mass);
-        let rolling_resistance =
-            compute_rolling_resistance(&traction_engine, &towed_rolling_stock, params.total_mass);
+impl PhysicsRollingStock {
+    pub fn new(traction_engine: RollingStock, physics_consist: PhysicsConsist) -> Self {
```
woshilapin: Why separating consist from rolling stock? A consist contains a rolling stock and towed rolling stock. I really think `PhysicsRollingStock` should completely disappear in favor of `PhysicsConsist`.

editoast/src/views/timetable/stdcm.rs
Wadjetz: I tried, how you deal with an async call inside the map ?

leovalais: I missed the `await`... good point, resolving 👀

front/src/applications/stdcm/utils/formatStdcmConfV2.ts
```diff
@@ -29,6 +29,7 @@ type ValidStdcmConfig = {
   totalMass?: number;
   totalLength?: number;
   maxSpeed?: number;
+  towedRollingStockID?: number;
```
woshilapin: Same than in the back, maybe putting this field next to `rollingStockId` would make sense (semantically because their close, but also in the UI, they're close).

front/src/reducers/osrdconf/types.ts
```diff
@@ -56,6 +56,7 @@ export interface OsrdStdcmConfState extends OsrdConfState {
   totalMass?: number;
   totalLength?: number;
   maxSpeed?: number;
+  towedRollingStockID?: number;
```
woshilapin, line 53: Why is the rolling stock ID not present in this interface?

front/src/modules/towedRollingStock/hooks/useFilterTowedRollingStock.ts
```diff
@@ -0,0 +1,93 @@
...
+) {
+  if (filters === initialFilters) return towedRollingStockList;
+  return towedRollingStockList.filter(({ id, name }) => {
+    if (filters.id !== undefined) {
+      return id === filters.id;
+    }
+    if (!filters.text) {
+      return true;
+    }
+    return name.trim().toLowerCase().includes(filters.text.trim().toLowerCase());
```
woshilapin, line 31: The first `if` handles the `id` filter, and the second `if` handles the `not text` filter... then the final line handle the `text` filter. It's not really logical and also, should they be composed together? Because what if both `filter.id` and `filter.text` are defined, I believe only `filter.id` will be taken into account.

front/src/modules/towedRollingStock/hooks/useFilterTowedRollingStock.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const initialFilters: TowedRollingStockFilters = {
+  text: '',
+};
+
+function filterTowedRollingStocks(
+  towedRollingStockList: TowedRollingStock[],
+  filters: TowedRollingStockFilters
+) {
+  if (filters === initialFilters) return towedRollingStockList;
```
achrafmohye: I think here you should do the comparison between the two objects using lodash's equal ?

front/src/applications/stdcmV2/components/StdcmConsist.tsx
```diff
@@ -118,6 +144,31 @@ const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
...
+          label={t('consist.towedRollingStock')}
+          value={towedRsFilters.text.toUpperCase()}
+          onClick={() => {
+            if (selectedTowedRollingStock?.id !== undefined) {
+              searchRollingStockById(selectedTowedRollingStock.id);
+            }
+          }}
+          onChange={(e) => {
+            searchTowedRollingStock(e.target.value);
+            if (e.target.value.trim().length === 0) {
```
achrafmohye:
```suggestion
            if (!e.target.value.trim()) {
```
Small enhancement but not mandatory :)

front/src/reducers/osrdconf/types.ts
Wadjetz: This struct extend OsrdConfState that has the rolling stock ID

front/src/modules/towedRollingStock/hooks/useFilterTowedRollingStock.ts
Wadjetz: But if you have an ID (which is unique), there is no need to match the search query.

front/public/locales/fr/errors.json
```diff
@@ -186,6 +186,7 @@
       "InfraNotFound": "Infrastructure '{{infra_id}}' non trouvée",
       "InvalidPathItems": "Point(s) de passage {{items}} invalide(s)",
       "RollingStockNotFound": "Matériel roulant '{{rolling_stock_id}}' non trouvé",
+      "TowedRollingStockNotFound": "Le matériel roulant remorqué {towed_rolling_stock_id} non trouvé",
```
```suggestion
      "TowedRollingStockNotFound": "Matériel roulant remorqué {towed_rolling_stock_id} non trouvé",
```

front/src/common/api/generatedEditoastApi.ts
SharglutDev, line 1: Issue has this : 

Capture d’écran 2024-10-25 à 08 36 40

Where does the front provide this information ?

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
SharglutDev, line 1: The consist zone looks like this  
Capture d’écran 2024-10-25 à 08 43 43

But it doesn't match the mockup : 
Capture d’écran 2024-10-25 à 08 44 37

front/src/modules/towedRollingStock/hooks/useFilterTowedRollingStock.ts
```diff
@@ -0,0 +1,93 @@
...
+      return id === filters.id;
+    }
+    if (!filters.text) {
+      return true;
+    }
+    return name.trim().toLowerCase().includes(filters.text.trim().toLowerCase());
+  });
+}
+
+export default function useFilterTowedRollingStock() {
```
SharglutDev: For custom hooks, we use `const`

```suggestion
const useFilterTowedRollingStock = () => {
```

front/src/modules/towedRollingStock/hooks/useFilterTowedRollingStock.ts
```diff
@@ -0,0 +1,93 @@
...
+  };
+
+  const resetFilters = () => {
+    setFilters(initialFilters);
+  };
+
+  useEffect(() => {
+    const newFilteredTowedRollingStock = filterTowedRollingStocks(allTowedRollingStocks, filters);
+    setFilteredTowedRollingStockList(newFilteredTowedRollingStock);
+  }, [isSuccess]);
```
SharglutDev: Seems like this `useEffect` does the same thing  as the last one. It's probably not needed.

front/src/modules/towedRollingStock/hooks/useFilterTowedRollingStock.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  useEffect(() => {
+    if (isError && error) {
+      dispatch(setFailure(castErrorToFailure(error)));
+    }
+  }, [isError]);
+
+  useEffect(() => {
+    const newFilteredTowedRollingStock = filterTowedRollingStocks(allTowedRollingStocks, filters);
```
SharglutDev:
```suggestion
    const newFilteredTowedRollingStocks = filterTowedRollingStocks(allTowedRollingStocks, filters);
```

front/src/modules/towedRollingStock/hooks/useFilterTowedRollingStock.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  const searchTowedRollingStock = (value: string) => {
+    setFilters({ ...filters, id: undefined, text: value });
+    setSearchIsLoading(true);
+  };
+
+  const searchTowedRollingStockById = (id?: number) => {
+    setFilters({ ...filters, id });
+    setSearchIsLoading(true);
+  };
```
SharglutDev: How can the user chose to search by name or id ? You don't use the search by id for now

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
```diff
@@ -118,7 +144,32 @@ const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
...
       </div>
-      <div className="stdcm-consist__properties">
+      <div className="towed-rolling-stock">
+        <ComboBox
+          id="towedRollingStock"
+          label={t('consist.towedRollingStock')}
+          value={towedRsFilters.text.toUpperCase()}
+          onClick={() => {
+            if (selectedTowedRollingStock?.id !== undefined) {
+              searchRollingStockById(selectedTowedRollingStock.id);
```
SharglutDev: Why do we use the `searchRollingStockById` when clicking on the towed rolling stock combobox and not `searchTowedRollingStockById` ?

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -43,6 +44,12 @@ export const stdcmConfSlice = createSlice({
     ) {
       state.maxSpeed = action.payload;
     },
+    updateTowedRollingStockID(
```
SharglutDev, line 57: Could you add a unit test for that in `describe('Consist updates', () => {...`

front/src/modules/towedRollingStock/hooks/useFilterTowedRollingStock.ts
```diff
@@ -0,0 +1,93 @@
...
+  }, [filters, allTowedRollingStocks]);
+
+  return {
+    allTowedRollingStocks,
+    filteredTowedRollingStockList,
+    filters,
+    searchIsLoading,
+    resetFilters,
+    searchTowedRollingStock,
+    searchTowedRollingStockById,
```
SharglutDev: There is some unused functions returned here. Did you created them for a future use ?

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -43,6 +44,12 @@ export const stdcmConfSlice = createSlice({
...
+      state: Draft<OsrdStdcmConfState>,
+      action: PayloadAction<OsrdStdcmConfState['towedRollingStockID']>
+    ) {
+      state.towedRollingStockID = action.payload;
+    },
     updateStdcmConfigWithData(
```
SharglutDev, line 63: There is something missing (that we also missed in the length/mass/vmax PR) : we should also pass here mass/length/vmax/towedRsId. Because if we switch between multiple simulation results, these fields aren't updated.

SharglutDev (commented):
Thank you for your PR, not tested yet but here are some comments

editoast/src/models/fixtures.rs
```diff
@@ -214,6 +221,62 @@ pub async fn create_rolling_stock_livery(
         .expect("Failed to create rolling stock livery")
 }
 
+pub fn create_towed_rolling_stock() -> TowedRollingStock {
```
woshilapin, line 224: Maybe there is a possible factorization with a similar function [here](https://github.com/OpenRailAssociation/osrd/blob/a54923883baad47b9803c4dc97794176481e4139/editoast/src/views/rolling_stock/towed.rs#L363-L389).

editoast/src/core/simulation.rs
woshilapin: Thank you very much for this refactorization, it looks much better to me. Hope you agree 😄

front/src/modules/towedRollingStock/hooks/useFilterTowedRollingStock.ts
woshilapin: True, so id filtering is an early return condition, got it.

woshilapin (approved):

front/src/common/api/generatedEditoastApi.ts
Wadjetz: I think the gabarit is postponed.
Capture d’écran 2024-10-28 à 10 53 05

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
Wadjetz: Sorry, it’s a rebase from dev that went bad. i fix it

editoast/src/core/simulation.rs
Wadjetz: it rolling resistance in french, i added more explicit prefixes

front/src/modules/towedRollingStock/hooks/useFilterTowedRollingStock.ts
Wadjetz: It is used when the user clicks on the item (the towed rolling stock object with an ID); we perform the search and update the state.

front/src/modules/towedRollingStock/hooks/useFilterTowedRollingStock.ts
SharglutDev: How about `searchIsLoading` and `resetFilters` ?

For `searchIsLoading`, it doesn't seem useful and for `resetFilters`, when we click on the combobox clear button, we call `searchTowedRollingStock` and the value is changed to empty string anyway so we can remove it as well.

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -43,13 +48,40 @@ export const stdcmConfSlice = createSlice({
...
+      action: PayloadAction<OsrdStdcmConfState['towedRollingStockID']>
+    ) {
+      state.towedRollingStockID = action.payload;
+    },
+    updateLoadingGaugeType(
+      state: Draft<OsrdStdcmConfState>,
+      action: PayloadAction<OsrdStdcmConfState['loadingGaugeType']>
+    ) {
+      state.loadingGaugeType = action.payload;
+    },
```
SharglutDev: As the user can't select any loading gauge and we don't display it, we don't need to have this information in the store. We can just pass the default value (with a comment) when calling the stdcm endpoint.

front/src/reducers/osrdconf/stdcmConf/index.ts
SharglutDev: The issue is still unsolved. The `selectedSimulation` doesn't get the new consist datas. They need to be add in `useStdcmForm`. (to test if it works, launch 2 different simulations and switch between them, the consist inputs should be automatically updated)

front/src/reducers/osrdconf/stdcmConf/selectors.ts
```diff
@@ -12,6 +12,8 @@ const buildStdcmConfSelectors = () => {
     getTotalMass: makeOsrdConfSelector('totalMass'),
     getTotalLength: makeOsrdConfSelector('totalLength'),
     getMaxSpeed: makeOsrdConfSelector('maxSpeed'),
+    getTowedRollingStockID: makeOsrdConfSelector('towedRollingStockID'),
+    getLoadingGaugeType: makeOsrdConfSelector('loadingGaugeType'),
```
SharglutDev: Remove this selector (see comment in index.ts)

front/src/reducers/osrdconf/stdcmConf/stdcmConfReducers.spec.ts
```diff
@@ -161,6 +161,16 @@ describe('stdcmConfReducers', () => {
...
+    it('should handle towedRollingStockID', () => {
+      store.dispatch(stdcmConfSliceActions.updateTowedRollingStockID(11));
+      const state = store.getState()[stdcmConfSlice.name];
+      expect(state.towedRollingStockID).toEqual(11);
+    });
+    it('should handle loadingGaugeType', () => {
+      store.dispatch(stdcmConfSliceActions.updateLoadingGaugeType('G1'));
+      const state = store.getState()[stdcmConfSlice.name];
+      expect(state.loadingGaugeType).toEqual('G1');
+    });
```
SharglutDev: you can remove the test (see comment in index.ts)

front/src/reducers/osrdconf/types.ts
```diff
@@ -49,6 +50,8 @@ export interface OsrdStdcmConfState extends OsrdConfState {
...
+  loadingGaugeType?: LoadingGaugeType;
```
SharglutDev: Same here

front/src/modules/towedRollingStock/hooks/useFilterTowedRollingStock.ts
```diff
@@ -0,0 +1,92 @@
...
+
+const initialFilters: TowedRollingStockFilters = {
+  text: '',
+};
+
+function filterTowedRollingStocks(
+  towedRollingStockList: TowedRollingStock[],
+  filters: TowedRollingStockFilters
+) {
+  if (eq(filters, initialFilters)) return towedRollingStockList;
```
SharglutDev: We should rather use `isEqual` from lodash as it make a deep comparison

front/src/reducers/osrdconf/stdcmConf/index.ts
SharglutDev: Also when we change length/mass/vmax/towed, it doesn't display the simulation button again. This will probably be solved when you fix the above comment.

front/src/applications/stdcm/views/StdcmView.tsx
```diff
@@ -73,6 +73,11 @@ const StdcmView = () => {
       dispatch(
         updateStdcmConfigWithData({
           rollingStockID: consist?.tractionEngine?.id,
+          towedRollingStockID: consist?.towedRollingStock?.id,
+          loadingGaugeType: consist?.tractionEngine?.loading_gauge,
```
SharglutDev: remove this (see index.ts)

front/src/common/api/generatedEditoastApi.ts
SharglutDev: Seems you provide G1 for `loading_gauge_type` instead of GA. But I'm not even sure where you provide this value ?

I think it should be hard coded in `formatStdcmPayload` : `loading_gauge_type: 'GA'; // default value as the user can't select one`

front/src/applications/stdcm/utils/formatStdcmConf.ts
```diff
@@ -193,5 +199,6 @@ export const formatStdcmPayload = (
     time_gap_before: toMsOrUndefined(validConfig.gridMarginAfter),
     work_schedule_group_id: validConfig.workScheduleGroupId,
     electrical_profile_set_id: validConfig.electricalProfileSetId,
+    loading_gauge_type: 'G2', // default value as the user can't select one
```
SharglutDev:
```suggestion
    loading_gauge_type: 'GA', // default value as the user can't select one
```

front/src/applications/stdcm/utils/formatStdcmConf.ts
```diff
@@ -20,6 +21,7 @@ import createMargin from './createMargin';
 // TODO: DROP STDCM V1: remove formattedStartTime, startTime and latestStartTime
 type ValidStdcmConfig = {
   rollingStockId: number;
+  towedRollingStockID?: number;
```
SharglutDev, line 23: We can remove this from valid config

front/src/applications/stdcm/types.ts
```diff
@@ -75,13 +77,14 @@ export type StdcmSimulationInputs = {
   pathSteps: (PathStep | null)[];
   consist?: {
     tractionEngine?: RollingStockWithLiveries;
-    /** In ton */
+    towedRollingStock?: TowedRollingStock;
     totalMass?: number;
```
SharglutDev, line 81: You can keep the comment for unit, it was nice

front/src/applications/stdcm/hooks/useStdcmTowedRollingStock.ts
```diff
@@ -0,0 +1,27 @@
...
+      },
+      {
+        skip: !towedRollingStockId,
+      }
+    );
+
+  return {
+    towedRollingStockId,
+    towedRollingStock: towedRollingStockId ? towedRollingStock : undefined,
+  };
```
SharglutDev: We don't use `towedRollingStockId` anywhere

```suggestion
  return towedRollingStock: towedRollingStockId ? towedRollingStock : undefined;
```

SharglutDev (approved):
Lgtm and tested, great job !

leovalais (approved):
LGTM, thanks for the changes!

editoast/src/core/simulation.rs
```diff
@@ -66,29 +69,153 @@ pub struct PhysicsRollingStock {
...
+    }
+
+    pub fn compute_inertia_coefficient(&self) -> f64 {
+        if let (Some(towed_rolling_stock), Some(total_mass)) =
+            (self.towed_rolling_stock.as_ref(), self.total_mass)
+        {
+            let mass_carriage = total_mass - self.traction_engine.mass;
+            let traction_engine_inertia =
+                self.traction_engine.mass * self.traction_engine.inertia_coefficient;
+            let towed_inertia = mass_carriage * towed_rolling_stock.inertia_coefficient;
```
axrolld, line 153: Not a big deal but the naming here could be more homogenous renaming `mass_carriage` in `towed_mass`

editoast/src/core/simulation.rs
```diff
@@ -348,3 +475,133 @@ impl AsCoreRequest<Json<SimulationResponse>> for SimulationRequest {
...
+        let mut physics_consist = create_physics_consist();
+        physics_consist.total_mass = Some(50000.0); // kg
+        physics_consist.traction_engine.mass = 15000.0; // kg
+
+        // We always take total_mass
+        assert_eq!(physics_consist.compute_mass(), 50000);
+
+        physics_consist.total_mass = None;
+        // When no total_mass we take towed mass + traction_engine mass
+        assert_eq!(physics_consist.compute_mass(), 65000);
```
axrolld, line 529: The consist should not be heavier without the user input. If towed mass + traction engin mass = 65t, then the user won't be able to set 50t in the front end, because it's lighter.
We need to make sure this check will be done by front end.

editoast/src/models/fixtures.rs
```diff
@@ -214,6 +221,63 @@ pub async fn create_rolling_stock_livery(
...
+        rolling_resistance: RollingResistance {
+            rolling_resistance_type: "davis".to_string(),
+            A: 1.0,
+            B: 0.01,
+            C: 0.0002,
+        },
+        gamma: Gamma {
+            gamma_type: "CONST".to_string(),
+            value: 1.0,
+        },
```
axrolld, line 242: It would be a good practice to specify all the units here, but it can be done in another PR IMO

editoast/src/models/fixtures.rs
```diff
@@ -214,6 +221,63 @@ pub async fn create_rolling_stock_livery(
...
+        locked: false,
+        metadata: None,
+        power_restrictions: HashMap::new(),
+        railjson_version: "12".to_string(),
+        rolling_resistance: RollingResistance {
+            rolling_resistance_type: "davis".to_string(),
+            A: 1.0,
+            B: 0.01,
+            C: 0.0005,
```
axrolld, line 273: Same here for the units

axrolld (commented):
I only checked the physics part and the tests but LGTM !
A few suggestions but that do not request immediate changes imo (just make sure that the front end will ensure the minimum value of mass / length).

It would be great to comment all the units of the test values in another PR though.

axrolld (approved):

maelysLeratRosso:
Unresolved conversations should not block the merge. 
The suggested improvements are in this issue https://github.com/OpenRailAssociation/osrd/issues/9528 and https://github.com/OpenRailAssociation/osrd/issues/9529 to be done asap

PR #9390, opened 2024-10-18
tests: add "locked" property to mocked rolling-stock

bougue-pe:
Since https://github.com/OpenRailAssociation/osrd/pull/9024, this is required to be able to use:
```sh
./scripts/load-railjson-rolling-stock.sh tests/data/rolling_stocks/fast_rolling_stock.json
```


shenriotpro (approved):
Thanks

woshilapin (approved):

PR #9393, opened 2024-10-18
editoast: add 'core' connectivity to the healthcheck

woshilapin:
We should also check connectivity to the `core` for the healthcheck.


clarani (approved):
LGTM for the front part :white_check_mark:

flomonster (approved):
LGTM (tested)

PR #9398, opened 2024-10-21
front: add waypoint menu in manchette

### LINKED ISSUE #8629: Manchette : hide a waypoint by clicking on it
### Description and goal

In the manchette, we want to be able to hide a waypoint when clicking on it.

To work, this issue https://github.com/OpenRailAssociation/osrd-ui/issues/399 needs be done first.

### Acceptance criteria

- [x] When clicking on a waypoint, an options menu should be displayed
- [x] "Hide this waypoint" should be the only option visible for now
- [x] The menu's design should match the [sketch](https://www.sketch.com/s/a26d84be-5fb2-41e4-9e84-96d90f46a997/a/OeJV9yG) one 
- [x] Clicking on "Hide this waypoint" should close the menu and properly hide the waypoint
- [x] The space time chart should resize accordingly
- [x] The hidden waypoint should still be present in the waypoint list modal so it can't be displayed again
- [ ] When the menu is open, clicking on another waypoint or the same one just closes the menu and we need one more click to open another one
- [x] When the menu is open, the scroll in the manchette, the manchette toolbar and pan in space time chart are disabled
- [ ] If there was only two waypoints left, it shouldn't be possible to hide it (behavior needs to be checked with @thibautsailly)

SharglutDev:
close #8629 

Mockup : 
https://www.sketch.com/s/a26d84be-5fb2-41e4-9e84-96d90f46a997/a/OeJV9yG


front/src/common/OSRDMenu.tsx
```diff
@@ -9,10 +9,11 @@ export type OSRDMenuItem = {
 type OSRDMenuProps = {
   menuRef: React.RefObject<HTMLDivElement>;
   items: OSRDMenuItem[];
+  style?: React.CSSProperties;
```
emersion: If this is exposed to allow callers to position the menu, should we instead add a `position` property? That would deter callers from (ab)using `style` for things which should stay in CSS files.

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -7,10 +7,12 @@ import { ConflictLayer, PathLayer, SpaceTimeChart } from '@osrd-project/ui-space
 import type { Conflict } from '@osrd-project/ui-spacetimechart';
 
 import type { OperationalPoint, TrainSpaceTimeData } from 'applications/operationalStudies/types';
+import WaypointMenu from 'common/OSRDMenu';
```
emersion: Should this be named `OSRDMenu`? It's a bit confusing to rename on import when there are no naming conflicts.

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -64,7 +85,19 @@ const ManchetteWithSpaceTimeChartWrapper = ({
...
         onScroll={handleScroll}
       >
-        <Manchette {...manchetteProps} />
+        <Manchette {...manchettePropsWithWaypointMenu}>
+          {waypointMenuData.menuPosition && (
+            <WaypointMenu
+              menuRef={waypointMenuData.menuRef}
+              items={waypointMenuData.menuItems}
+              style={{
+                width: '305px',
```
emersion: Could we move this to a CSS file?

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -41,8 +45,25 @@ const ManchetteWithSpaceTimeChartWrapper = ({
...
+  );
+
+  const manchettePropsWithWaypointMenu = useMemo(
+    () => ({
+      ...manchetteProps,
+      operationalPoints: manchetteProps.operationalPoints.map((op) => ({
+        ...op,
+        onClick: (id: string, ref: HTMLDivElement | null) => {
+          waypointMenuData.handleWaypointClick(id, ref);
+        },
```
emersion: Nit: can be simplified to:

```ts
onClick: waypointMenuData.handleWaypointClick,
```

front/src/modules/simulationResult/components/SpaceTimeChart/useWaypointMenu.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  const menuItems: OSRDMenuItem[] = [
+    {
+      title: t('waypointMenu.hide'),
+      icon: <EyeClosed />,
+      onClick: () => {
+        setMenuPosition(undefined);
+        setActiveOperationalPointId(undefined);
+        if (filteredWaypoints && setFilteredWaypoints) {
+          setFilteredWaypoints(
```
emersion: This is racy, because another update to `filteredWaypoints` could be pending. Instead, `setFilteredWaypoints` can take a callback which mutates the current value:

```ts
setFilteredWaypoints((filteredWaypoints) => filteredWaypoints.filter(…));
```

As a bonus, this removes the need to pass `filteredWaypoints` as a prop.

front/src/modules/simulationResult/components/SpaceTimeChart/useWaypointMenu.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { EyeClosed } from '@osrd-project/ui-icons';
+import { useTranslation } from 'react-i18next';
+
+import type { OperationalPoint } from 'applications/operationalStudies/types';
+import type { OSRDMenuItem } from 'common/OSRDMenu';
+import useModalFocusTrap from 'utils/hooks/useModalFocusTrap';
+import useOutsideClick from 'utils/hooks/useOutsideClick';
+
+const SPACETIME_CHART_HEADER_HEIGHT = 40;
+const WAYPOINT_MENU_OFFSET = 2;
```
emersion: This could be a CSS margin instead.

front/src/modules/simulationResult/components/SpaceTimeChart/useWaypointMenu.tsx
```diff
@@ -0,0 +1,71 @@
...
+  useOutsideClick(menuRef, closeMenu);
+  useModalFocusTrap(menuRef, closeMenu);
+
+  const menuItems: OSRDMenuItem[] = [
+    {
+      title: t('waypointMenu.hide'),
+      icon: <EyeClosed />,
+      onClick: () => {
+        setMenuPosition(undefined);
+        setActiveOperationalPointId(undefined);
```
emersion: We could call `closeMenu` here instead.

emersion (commented):
Overall I'm not a huge fan of having the menu outside of ui-manchette. IMHO the positioning and styling of the menu belongs to osrd-ui.

I'm especially concerned about the positioning logic which is quite fragile: it hardcodes a lot of assumptions about the manchette, and desynchronizes when elements move around. Example bug: zoom in the manchette, open the menu, and then scroll. This is just one example, other interactions may break the menu alike.

out

emersion:
Actually, maybe this coule be used instead?

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_anchor_positioning/Using

front/src/common/OSRDMenu.tsx
SharglutDev: We also need to set the width, it's not always the same depending of who's calling the menu. I made the call to pass this props instead of pass `width` and `position`. You would prefer that ?

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
SharglutDev: This would require a div wrapper as the width it's not always the same when calling the menu. I'm not a huge fan of that. Or you prefer passing a `className` prop instead of `style` ? How would you do ?

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
SharglutDev: Oh good point, done

front/src/modules/simulationResult/components/SpaceTimeChart/useWaypointMenu.tsx
SharglutDev: Yep I know but as it requires to pass the `setFilteredWaypoints` not as a standard function but as a `SetStateAction`, it needed to be changed everywhere we pass this props (but only two spots actually) and I was lazy :upside_down_face: But you're right I should do that, done :)

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
emersion: Wouldn't `.manchette .osrd-menu` work? If not, I'd suggest adding a `className` prop.

front/src/modules/simulationResult/components/SpaceTimeChart/useWaypointMenu.tsx
SharglutDev: Done

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
SharglutDev: Oh yes, you're right, done

front/src/modules/simulationResult/components/SpaceTimeChart/useWaypointMenu.tsx
SharglutDev: Done

SharglutDev:
> Overall I'm not a huge fan of having the menu outside of ui-manchette. IMHO the positioning and styling of the menu belongs to osrd-ui.
> 
> I'm especially concerned about the positioning logic which is quite fragile: it hardcodes a lot of assumptions about the manchette, and desynchronizes when elements move around. Example bug: zoom in the manchette, open the menu, and then scroll. This is just one example, other interactions may break the menu alike.
> 

You're totally right, I will check for solutions

front/src/common/OSRDMenu.tsx
Synar: I think it would be cleaner to specify width and position, this way callers know what they can and should provide

SharglutDev: Forgot to update here but on the second fixup, I put the width in the css directly.

For the position part, I'll probably have to put the menu inside osrd-ui directly so we might not need this props anymore. I'll try that tomorrow.

Synar (commented):
LGTM beyond the positioning issue pointed out by emersion. But it is my understanding that https://github.com/OpenRailAssociation/osrd-ui/pull/668/files should solve this issue?

SharglutDev:
> LGTM beyond the positioning issue pointed out by emersion. But it is my understanding that https://github.com/OpenRailAssociation/osrd-ui/pull/668/files should solve this issue?

Yes exactly it's on wip :)

SharglutDev:
Nothing new, just rebased with dev to be not too far behind with conflicts :)

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -178,8 +182,29 @@ const ManchetteWithSpaceTimeChartWrapper = ({
     }));
   });
 
+  const waypointMenuData = useWaypointMenu(
+    waypointsPanelData?.filteredWaypoints,
+    waypointsPanelData?.setFilteredWaypoints
+  );
```
clarani: Nit: you can simply give the waypointsPanelData directly
```suggestion
  const waypointMenuData = useWaypointMenu(waypointsPanelData);
```

front/src/styles/scss/common/components/_manchetteWithSpaceTimeChart.scss
```diff
@@ -37,6 +38,15 @@
...
     overflow-y: auto;
     overflow-x: hidden;
+
+    &.no-scroll {
+      overflow-y: clip;
+    }
+
+    .osrd-menu {
+      width: 305px;
+      margin-top: -2px;
```
theocrsb: I think it's better not to use a negative margin, a transform would be better.

theocrsb (approved):
LGTM and tested ! I just posted a little comment

clarani (approved):
LGTM :white_check_mark:

front/src/styles/scss/common/components/_manchetteWithSpaceTimeChart.scss
SharglutDev: Done

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx

PR #9424, opened 2024-10-22
Add security file

flomonster:


emersion:
Note, seems like this is copied over from https://github.com/electron/electron/blob/main/SECURITY.md

emersion:
- Security policies typically indicate which versions are currently supported.
- Might be worth mentioning that security issues should not be reported through public GitHub issues.

Khoyo (approved):
Didn't know github had that feature, that's really nice

ghost (approved):

bloussou (approved):
✅

emersion (approved):

clarani (approved):

PR #9426, opened 2024-10-22
script: change call arguments for shutting down the docker

EthanPERRUZZA:
I hadn't see that PE wanted those changes on the script rework too https://github.com/OpenRailAssociation/osrd-website/pull/247

\ + fix help not showing on certain cases


bougue-pe (approved):
I thought it already worked with or without it 🤭 
It was not mandatory, yet, thanks for this !

woshilapin (approved):

PR #9428, opened 2024-10-22
front: create type StdcmPathStep

### LINKED ISSUE #9392: front: create a simple StdcmPathStep type
Add a new stdcmPathSteps props in the stdcm reducer and create a new StdcmPathStep type (which extends PathStep for the moment):
```
export type StdcmPathStep = PathStep & {
  isVia: boolean;
  tolerances?: { before: number; after: number };
} & ({ isVia: true; stopType: StdcmStopTypes } | { isVia: false; arrivalType: ArrivalTimeTypes });
```

This will simplify the stdcm interface.

Will be followed by https://github.com/OpenRailAssociation/osrd/issues/9462

clarani:
closes #9392 

- create a new type StdcmPathStep
- add a new props stdcmPathSteps to the stdcmStore
- create new actions: updateStdcmPathStep, addStdcmVia, removeStdcmVia
- create new selectors: getStdcmPathSteps, getStdcmOrigin, getStdcmDestination
- clean:
  - remove several actions: updateOriginArrival, updateDestinationArrival, updateOriginArrivalType, updateDestinationArrivalType, updateOriginTolerances, updateDestinationTolerances
  - clean some stdcm components

Will be followed by https://github.com/OpenRailAssociation/osrd/issues/9462


front/src/reducers/osrdconf/types.ts
```diff
@@ -81,3 +82,8 @@ export type PathStep = PathItemLocation & {
     trackNumber: number;
   };
 };
+
+export type StdcmPathStep = PathStep & {
+  isVia: boolean;
```
emersion: I think we can drop this field since it's defined below as well

front/src/applications/stdcm/utils/formatStdcmConf.ts
```diff
@@ -1,22 +1,39 @@
...
 
+const getStepLocation = (step: StdcmPathStep): PathItemLocation => {
+  if ('track' in step) {
+    return { track: step.track, offset: mToMm(step.offset) };
+  }
+  if ('operational_point' in step) {
+    return { operational_point: step.operational_point };
+  }
+  if ('trigram' in step) {
+    return { trigram: step.trigram, secondary_code: step.secondary_code };
```
emersion: Hmm. I think we actually never use `PathStep.secondary_code` and instead always use `PathStep.ch`?

See: https://github.com/OpenRailAssociation/osrd/pull/8633

front/src/applications/stdcm/utils/formatStdcmConf.ts
```diff
@@ -1,22 +1,39 @@
...
 import { setFailure } from 'reducers/main';
-import type { OsrdStdcmConfState, StandardAllowance } from 'reducers/osrdconf/types';
+import type { OsrdStdcmConfState, StandardAllowance, StdcmPathStep } from 'reducers/osrdconf/types';
 import { dateTimeFormatting } from 'utils/date';
 import { kmhToMs, mToMm, tToKg } from 'utils/physics';
 import { ISO8601Duration2sec, sec2ms } from 'utils/timeManipulation';
 
 import createMargin from './createMargin';
 
+const getStepLocation = (step: StdcmPathStep): PathItemLocation => {
```
emersion: This would also be useful for operational studies if this took a `PathStep` instead, but I suppose the longer term goal is to make these two completely separate?

clarani: Yes :/ but maybe I can do smth like this:
```suggestion
const getStepLocation = (step: PathStep | StdcmPathStep): PathItemLocation => {
```

front/src/applications/stdcm/utils/formatStdcmConf.ts
clarani: Thanks for your PR :star_struck:
I can use ch for now, and then when we drop ch, we could use secondary_code instead ?

front/src/applications/stdcm/utils/formatStdcmConf.ts
emersion: Ah yes, that sounds like a workable solution!

For reference, the duplicated code: https://github.com/OpenRailAssociation/osrd/blob/86ad6408b1b1a567b691882dc1c3d543b4df9112/front/src/modules/pathfinding/utils.ts#L68

front/src/applications/stdcm/utils/formatStdcmConf.ts
emersion: Sounds good!

front/src/reducers/osrdconf/types.ts
emersion: Seems like the field is still here, am I missing something?

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -1,13 +1,19 @@
...
-
-import { updateOriginPathStep, updateDestinationPathStep } from '../helpers';
+import { addElementAtIndex } from 'utils/array';
+import type { ArrayElement } from 'utils/types';
 
 export const stdcmConfInitialState: OsrdStdcmConfState = {
+  // TODO: remove all the default uic values
+  stdcmPathSteps: [
+    { id: nextId(), uic: -1, isVia: false, arrivalType: ArrivalTimeTypes.PRECISE_TIME },
+    { id: nextId(), uic: -1, isVia: false, arrivalType: ArrivalTimeTypes.ASAP },
```
emersion, line 15: I'm not a big fan of using `uic: -1` to represent unset steps, I think the previous design where `null` means unset was safer. I'm guessing the reason you switched to -1 is to be able to set `arrivalType` for these unset steps? I suppose the TODO is about getting rid of that -1?

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -88,10 +94,10 @@ const StdcmConfig = ({
   };
 
   const removeOriginArrivalTime = () => {
-    dispatch(updateOriginArrivalType(ArrivalTimeTypes.ASAP));
+    dispatch(updateStdcmPathStep({ ...origin, arrivalType: ArrivalTimeTypes.ASAP }));
```
emersion: Hm, this seems racy when two dispatches happen in quick succession… Right?

Could we maybe do the merging with the current state inside the action?

front/src/reducers/osrdconf/types.ts
clarani: Oops, sorry, it's removed now

front/src/reducers/osrdconf/stdcmConf/index.ts
clarani: Yes exactly. You can have a look at the next [issue](https://github.com/OpenRailAssociation/osrd/issues/9462). There will be a new props location (which can be undefined)

emersion: Oh, excellent!

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -94,10 +94,15 @@ export const stdcmConfSlice = createSlice({
...
-      action: PayloadAction<ArrayElement<OsrdStdcmConfState['stdcmPathSteps']>>
+      action: PayloadAction<{
+        id: string;
+        updates: Partial<ArrayElement<OsrdStdcmConfState['stdcmPathSteps']>>;
+      }>
     ) {
       const newPathSteps = state.stdcmPathSteps.map((pathStep) =>
-        pathStep.id === action.payload.id ? action.payload : pathStep
+        pathStep.id === action.payload.id
+          ? ({ ...pathStep, ...action.payload.updates } as StdcmPathStep)
```
emersion, line 104: Nit: since we're spreading `pathStep`, do we really need `as StdcmPathStep`?

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -94,10 +94,15 @@ export const stdcmConfSlice = createSlice({
     },
     updateStdcmPathStep(
       state: Draft<OsrdStdcmConfState>,
-      action: PayloadAction<ArrayElement<OsrdStdcmConfState['stdcmPathSteps']>>
+      action: PayloadAction<{
+        id: string;
+        updates: Partial<ArrayElement<OsrdStdcmConfState['stdcmPathSteps']>>;
```
emersion, line 99: If we wanted to, we could have a flatter data structure with something like:

```ts
action: PayloadAction<{ id: string } & Partial<…>>
```

(Making the `id` property required and everything else optional.)

But the status quo seems perfectly fine to me as well (feel free to ignore).

emersion (approved):
Feel free to resolve the remaining two comments at your discretion.

Looks good, thanks a lot for this cleanup!

front/src/reducers/osrdconf/stdcmConf/index.ts
clarani: Yes we need it, because of `isVia` : the props are not the same depending on its value, so TS is not very happy about it :/

front/src/reducers/osrdconf/stdcmConf/index.ts
clarani: Good idea, but I prefer to keep the id and the rest separated :)

PR #9429, opened 2024-10-22
front: add scenario train ids to query params and local storage

### LINKED ISSUE #9018: Add the selected train and the projection to the scenario URL
### Description and goal

It would be very nice if the train I have selected and the projection I'm using on my GET are not reset after a refresh. https://github.com/OpenRailAssociation/osrd/issues/9011 exist.

We should encode the selected train and projection in the URL doing so will allow sharing a specific view to others.

### Technical design

The URL could look like this:
```
https://..../operational-studies/projects/64/studies/312/scenarios/592?selected_train=4242&projection=8484
```

Both `selected_train` and `projection` are train schedule ids.

> [!NOTE]
> If this query parameter is not provided / invalid (deleted train). These parameters should be replaced by the first valid train.

> [!TIP]
> **Bonus:** The selected train/projection could also be saved in the local storage. So the next time a user opens its scenario, he has the same page as before. 
> - Query params have priority
> - These parameters must be associated with the scenario (each scenario has it's last selected train/projection)

### Acceptance criteria

- [x] When refreshing a scenario page, the previously selected train and train used for the projection are kept
- [x] When sharing the URL we should retrieve the same selected train/projection

Synar:
Close #9018

### DIFF
--- a/front/src/applications/operationalStudies/hooks/useScenarioData.ts
+++ b/front/src/applications/operationalStudies/hooks/useScenarioData.ts
@@ -11,7 +11,6 @@ import {
   type TrainScheduleResult,
 } from 'common/api/osrdEditoastApi';
 import { setFailure } from 'reducers/main';
-import { updateSelectedTrainId, updateTrainIdUsedForProjection } from 'reducers/simulationResults';
 import {
   getSelectedTrainId,
   getTrainIdUsedForProjection,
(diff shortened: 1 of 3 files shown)


flomonster (approved):
Tested and works like a charm! Thanks :pray :

Math-R (approved):
lgtm ! nice work

PR #9439, opened 2024-10-23
core: improve worker lifetime

### LINKED ISSUE #9388: core: health check should be KO if invalid infra or non-existing (404 from editoast)
### What happened?

Core stays "healthy" no matter the outcome of infra loading.
(no health endpoint exists)

This leads to some cases where a timeout from editoast locks core workers and the queue stays full

We could either add a liveness server or kill the process in case something is going unexpected when loading the infra.
What's expected is:
* loading OK : obvious case
* loading leads to an invalid infra : in this case, stay alive and reject any request on the same infra and same version
* requesting an infra that doesn't exist anymore : stay alive and reject any request on that infra
* any other case should lead to a retry of loading, then shutdown

See above

Dev (SNCF)

Firefox

dev

bougue-pe:
🔍 review the commits separately, ignore whitespace on first commit.

fixes #9388

Done:
* fix rabbit's connection closing
* stop worker when unable to process requests after infra load

Details:
The worker is able to process requests if infra load:
* ... went fine, happy path (just process request)
* ... failed with a perennial error (expected behavior is reject requests on the same version of the same infra)

If infra load failed with a temporary error (and after possible retries), better just stop and let orchestrator decide to retry or not.

No retry implemented so far (easier to test the behavior... and easier to code 😇).
Hand-tested on the cases I could think of and reproduce, more test is welcome.

### DIFF
--- a/core/src/main/java/fr/sncf/osrd/cli/WorkerCommand.kt
+++ b/core/src/main/java/fr/sncf/osrd/cli/WorkerCommand.kt
@@ -3,11 +3,6 @@ package fr.sncf.osrd.cli
 import com.beust.jcommander.Parameter
 import com.beust.jcommander.Parameters
 import com.rabbitmq.client.*
-import com.rabbitmq.client.AMQP
-import com.rabbitmq.client.Channel
-import com.rabbitmq.client.ConnectionFactory
-import com.rabbitmq.client.DeliverCallback
-import com.rabbitmq.client.Delivery
 import fr.sncf.osrd.api.*
 import fr.sncf.osrd.api.api_v2.conflicts.ConflictDetectionEndpointV2
 import fr.sncf.osrd.api.api_v2.path_properties.PathPropEndpoint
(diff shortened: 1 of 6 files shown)


core/src/main/java/fr/sncf/osrd/cli/WorkerCommand.kt
```diff
@@ -225,6 +246,10 @@ class WorkerCommand : CliCommand {
                             "ERROR, exception received"
                                 .toByteArray() // TODO: have a valid payload for uncaught exceptions
                         status = "core_error".encodeToByteArray()
+                        // Stop worker and let another worker spawn eventually
+                        if (t is OSRDError && t.osrdErrorType == ErrorType.InfraSoftLoadingError) {
+                            throw t
```
woshilapin, line 276: Why `throw` and not directly [`exitProcess`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.system/exit-process.html) directly since you know you want to kill the worker in this case? I'd say bubbling up an error is useful if you can't handle it right now, or if you expect it to be displayed to the end-user. But I believe in this case, you can handle it right now. What do you think?

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -61,15 +61,12 @@ class PathfindingBlocksEndpointV2(private val infraManager: InfraManager) : Take
...
-                pathfindingLogger.info("Pathfinding failed: ${error.message}")
-                val response = PathfindingFailed(error)
+        } catch (ex: Throwable) {
+            if (ex is OSRDError && ex.osrdErrorType.isCacheable) {
+                pathfindingLogger.info("Pathfinding failed: ${ex.message}")
+                val response = PathfindingFailed(ex)
                 RsJson(RsWithBody(pathfindingResponseAdapter.toJson(response)))
             }
-        } catch (ex: Throwable) {
             ExceptionHandler.handle(ex)
```
woshilapin: If I understand correctly this modification, you only merged `catch (error: OSRDError)` with `catch (ex: Throwable)` only to factorize a single line which is `ExceptionHandler.handle(error)`. I'd say it was arguably better before because types were more specific in the branches, and the benefit of factorizing only one line might not worth it. But it's possible I missed something from this diff too. What do you think?

woshilapin (approved):
Will definitely improve the situation, thank you.

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/standalone_sim/SimulationEndpoint.kt
```diff
@@ -72,14 +72,11 @@ class SimulationEndpoint(
...
-                return ExceptionHandler.handle(error)
-            } else {
-                val response = SimulationFailed(error)
+        } catch (ex: Throwable) {
+            if (ex is OSRDError && ex.osrdErrorType.isCacheable) {
+                val response = SimulationFailed(ex)
                 return RsJson(RsWithBody(simulationResponseAdapter.toJson(response)))
             }
-        } catch (ex: Throwable) {
             return ExceptionHandler.handle(ex)
```
woshilapin: See https://github.com/OpenRailAssociation/osrd/pull/9439#discussion_r1814427452.

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
bougue-pe: You're right on what was done.
I was first mislead by the cognitive complexity while working in order to re-throw on `InfraSoftLoadingError` so I changed it 🤷
It's OK for me to change back if someone else votes for it 🙂

core/src/main/java/fr/sncf/osrd/cli/WorkerCommand.kt
bougue-pe: I'll let others (@Khoyo ?) react to this also, but here is my take.

* If we always want to exitProcess, we may do it [right away](https://github.com/OpenRailAssociation/osrd/pull/9439/files#diff-ce4c689b90b896953c59277e744de96c9fb100774f54ac3e0720c0575135864dR154) instead of throwing SoftLoadingError (so even remove that line).
  * (minor) If this is a real use-case we want to maintain: for multi-infra cores (either worker or web API), not sure exitProcess is always what we want as we loose the worker for all infras.
* Opinionated: exitProcess (and early returns) in the middle of the code feels like a trap to maintain the code. Letting the worker recover anytime before the actual end of the `main()` feels better.

eckter (approved):
LGTM (not tested though)

Though I feel a little uncomfortable about the `isCacheable` flag in `ErrorType`. Apparently it was there but not set for any error, now we set it for `InfraSoftLoadingError`. But then most of the logic is made by checking for `InfraSoftLoadingError` specifically. 

Maybe we could just remove the flag? Or always use it instead of checking for this specific error.

bougue-pe:
@eckter 
> Though I feel a little uncomfortable about the `isCacheable` flag in `ErrorType`. Apparently it was there but not set for any error, now we set it for `InfraSoftLoadingError`. But then most of the logic is made by checking for `InfraSoftLoadingError` specifically.
> 
> Maybe we could just remove the flag? Or always use it instead of checking for this specific error.

I'm OK to remove the flag (I won't have the time to do it before 4/11/2024).
There's a bit of work to have things contiguous/ensure consistency (between handler, path_finding and simulation):
* I'd probably rename the flag to "unrecoverable"
* then also add a unique "filter_recoverable" method to be used in all 3 places

bougue-pe:
@eckter : introduced a 'simple' `isRecoverable` in https://github.com/OpenRailAssociation/osrd/pull/9439/commits/da81c3de6a603fab1f426ff750dcc309a4de9674

core/src/main/java/fr/sncf/osrd/cli/WorkerCommand.kt
Khoyo: > If this is a real use-case we want to maintain

Multi infra workers are only there to ease opening core inside a debugger, and I'm not sure (?) that people *really* use multiple infras (vs only using it to not specify the worker key)

@Erashin @eckter 

Re throwing vs exitProcess(), at least in the jvm world I believe you'll at least get a stack trace with eg. the `context` of `t `, so I'd tend to err on the throwing side.

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
Khoyo: There is behavior change here, in your change you call `ExceptionHandler.handle(ex)` unconditionally. I'm not sure this is what we want to do.

core/src/main/java/fr/sncf/osrd/cli/WorkerCommand.kt
Erashin: Yeah, it's only to ease opening core inside a debugger for me.

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
bougue-pe: Not sure how the `return try` handles "early returns"... So [I made it more explicit](https://github.com/OpenRailAssociation/osrd/pull/9439/commits/954a328b13a59d01c65b57848ec30f2190350bed).

Still, if you like better the previous implementation, I can revert (probably without the `return try`).

bougue-pe: OK, as discussed you had a good point, and it's corrected now.
Resolving this.

PR #9440, opened 2024-10-24
orsdyne: add telemetry support

### LINKED ISSUE #8081: Osrdyne: add telemetry
Should we add telemetry on `osrdyne`? Can be done in another PR.

_Originally posted by @woshilapin in https://github.com/OpenRailAssociation/osrd/pull/8048#discussion_r1677546159_

ghost:
Adding telemetry in osrdyne

closes [#8081](https://github.com/OpenRailAssociation/osrd/issues/8081)

### DIFF
--- a/osrdyne/src/config.rs
+++ b/osrdyne/src/config.rs
@@ -10,6 +10,8 @@ use figment::{
     providers::{Env, Format, Serialized, Yaml},
     Figment,
 };
+use tracing::info;
+use url::Url;
 
 #[derive(Serialize, Deserialize, Clone, Debug)]
 #[serde(tag = "type")]
@@ -32,6 +34,22 @@ pub struct OsrdyneConfig {
     pub max_length_bytes: Option<usize>,
     pub api_address: String,
     pub extra_lifetime: Option<Duration>,
+    pub opentelemetry: Option<OpentelemetryConfig>,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct OpentelemetryConfig {
+    pub service_name: Option<String>,
+    pub endpoint: Url,
+}
+
+impl Default for OpentelemetryConfig {
+    fn default() -> Self {
+        Self {
+            service_name: None,
+            endpoint: "http://jaeger:4317".parse().unwrap(),
+        }
+    }
 }
 
 impl Default for OsrdyneConfig {
@@ -47,13 +65,14 @@ impl Default for OsrdyneConfig {
             max_length_bytes: None,
             api_address: "0.0.0.0:4242".into(), // TODO: decide on the port
             extra_lifetime: None,
+            opentelemetry: None,
         }
     }
 }
 
 pub fn parse_config(file: Option<PathBuf>) -> Result<OsrdyneConfig, figment::Error> {
     let provider = if let Some(file) = file {
-        log::info!("Using configuration file: {}", file.display());
+        info!(file = %file.display(), "load configuration file");
         Yaml::file_exact(file)
     } else {
         Yaml::file("osrdyne.yml")
--- a/osrdyne/src/drivers/kubernetes.rs
+++ b/osrdyne/src/drivers/kubernetes.rs
@@ -20,14 +20,15 @@ use k8s_openapi::{
 };
 use keda::{MetricType, ScaledObject};
 use kube::{api::ObjectMeta, Client};
-use log::{debug, info};
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 use std::{
     collections::{BTreeMap, HashMap},
+    fmt::Debug,
     future::Future,
     pin::Pin,
 };
+use tracing::{debug, info, instrument};
 use uuid::Uuid;
 
 mod keda;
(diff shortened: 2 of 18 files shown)


flomonster (approved):
We should also add env vars to the host and pr-tests docker-compose files. 

Otherwise LGTM

osrdyne/src/config.rs
```diff
@@ -32,6 +34,8 @@ pub struct OsrdyneConfig {
     pub max_length_bytes: Option<usize>,
     pub api_address: String,
     pub extra_lifetime: Option<Duration>,
+    pub service_name: String,
+    pub opentelemetry_endpoint: Option<Url>,
```
woshilapin: Do we want/need the `service_name` if the opentelemetry endpoint is not defined 🤔 If not, then maybe this can be done like this?
```suggestion
    pub opentelemetry_config: Option<OpenTelemetryConfig>,
```

with
```rust
struct OpentelemetryConfig {
  service_name: String,
  endpoint: Url
}
```

osrdyne/src/drivers/kubernetes.rs
```diff
@@ -207,7 +220,7 @@ impl KubernetesDriver {
         };
 
         if let Some(version) = current_hpa_version {
-            info!("Updating HPA: {:?}", hpa);
+            debug!("Updating HPA: {:?}", hpa);
```
woshilapin: Now that we use `tracing`, we might as well use it to send structured event logs instead? Take a look at all the modified log events of this PR, most of them are also concerned.

```suggestion
            debug!(?hpa, "Updating HPA");
```

osrdyne/src/pool.rs
```diff
@@ -477,12 +478,15 @@ async fn worker_control_loop(
     sleep_interval: Duration,
 ) {
     loop {
+        let span = span!(Level::INFO, "loop_iteration");
+        let _enter = span.enter();
```
woshilapin, line 482: If you don't reuse `span` anywhere else, you should be able to use `.entered()` which produce a guard that owns the span instead of borrowing it.

```suggestion
        let _entered_guard = span!(Level::INFO, "loop_iteration").entered();
```

osrdyne/src/main.rs
```diff
@@ -200,3 +203,53 @@ async fn main() -> Result<(), anyhow::Error> {
...
+
+        let otlp_tracer = otlp_tracer_provider.tracer(config.service_name.clone());
+
+        let layer = tracing_opentelemetry::layer()
+            .with_tracer(otlp_tracer)
+            .boxed();
+
+        opentelemetry::global::set_text_map_propagator(
+            opentelemetry_sdk::propagation::TraceContextPropagator::new(),
+        );
```
woshilapin, line 249: I didn't see where it was used? I would expect to use this propagator whenever we do request to other services (in the case of `osrdyne`, that would means any request to RabbitMQ should propagate into the headers of messages).

woshilapin (commented):
Nice to see more OpenTelemetry everywhere.

You might also want to configure `axum` so it catches the trace ID in incoming requests. See [here](https://github.com/OpenRailAssociation/osrd/blob/59775313d99fe067e5e950da181d41c5bb7d54e6/editoast/src/main.rs#L388). You may also add `TraceLayer` to automatically create spans on all incoming requests (see https://github.com/OpenRailAssociation/osrd/blob/59775313d99fe067e5e950da181d41c5bb7d54e6/editoast/src/main.rs#L392)).

osrdyne/src/main.rs
ghost: I've put it so it was already there. We can removing, but that does not seems to be an issue to let it as is even if not used yet ?

woshilapin: It's indeed not an issue to set it up even if used later. I just thought that would be part of the current PR. Feel free to resolve if you plan to use it in future PR.

osrdyne/src/pool.rs
ghost: Not compiling because worker_control_loop is called within a task with async move.

woshilapin (approved):
Thanks for the improvements.

flomonster:
Not working in local. Osrdyne logs:
```
Error: missing field `service_name` for key "OPENTELEMETRY" in `OSRDYNE__` environment variable(s)
```

osrdyne/src/config.rs
```diff
@@ -32,6 +34,22 @@ pub struct OsrdyneConfig {
...
+#[derive(Debug, Deserialize, Serialize)]
+pub struct OpentelemetryConfig {
+    pub service_name: Option<String>,
+    pub endpoint: Url,
+}
+
+impl Default for OpentelemetryConfig {
+    fn default() -> Self {
+        Self {
+            service_name: None,
```
woshilapin, line 49: Why making it optional? `"osrdyne".into()` was a good default, don't you think?

bloussou (approved):
Not tested but looks ok :)

osrdyne/src/config.rs
ghost: Nope, was still messing when set from environment variables

PR #9443, opened 2024-10-24
Remove datadog support

flomonster:
We don't use datadog anymore. We can drop the support, which makes our lives easier and encourages to use opentelemetry, a free and opensource solution.


docker/docker-bake.hcl
```diff
@@ -79,11 +79,7 @@ target "editoast-test" {
   }
 }
 
-target "base-editoast" {
-  args = {
-    CARGO_FEATURES = "datadog"
-  }
-}
```
flomonster, line 86: Should I also remove `CARGO_FEATURE` in the editoast `Dockerfile` ? 
I don't know if we plan to add features for editoast :shrug:

woshilapin: I'd say that if it's easy to remove, let's remove it, the code will be less cluttered, and I don't see a short-term opportunity to use that functionality anyway.

woshilapin (approved):
Nice, everyone loves these PRs only removing code 😄 Actually, I believe you may also remove [this line](https://github.com/OpenRailAssociation/osrd/blob/4af662cecd849e126c5abce1c01c2e20fb5d746b/core/Dockerfile#L18).

Also, I did [this PR](https://github.com/OpenRailAssociation/osrd-chart/pull/36) to remove Datadog support in our charts.

ghost (approved):

eckter (approved):

PR #9447, opened 2024-10-24
front: fix combo box suggestions in stdcm

### LINKED ISSUE #9296: stdcm: CI input keeps track of what was previously input, but without showing it
### What happened?

If i start typing something the results are filtered, clocking outside removes everything (so far it's normal), but when clicking again to type something the field is empty but the list is still filtered:

https://github.com/user-attachments/assets/c7c4a25a-407e-4337-ba2f-caeb1d9b5173

Clicking outside the input should clear eveything completely and clicking back inside the field should not filter with what was previously input.

1. reproduce the video

Recette (SNCF)

Firefox, Google Chrome

96c960e

RomainValls:
closes #9296

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useMemo } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 
 import { Select, ComboBox } from '@osrd-project/ui-core';
 import { useTranslation } from 'react-i18next';
@@ -30,13 +30,13 @@ const StdcmOperationalPoint = ({
   disabled,
 }: StdcmOperationalPointProps) => {
   const { t } = useTranslation('stdcm');
+  const [suggestionsVisible, setSuggestionsVisible] = useState(false);
 
   const { searchTerm, chCodeFilter, sortedSearchResults, setSearchTerm, setChCodeFilter } =
     useSearchOperationalPoint({ initialSearchTerm: point?.name, initialChCodeFilter: point?.ch });
 
   const operationalPointsSuggestions = useMemo(
     () =>
-      // Temporary filter added to show a more restrictive list of suggestions inside the stdcm app.
       sortedSearchResults
         .filter(
           (op) =>
@@ -120,7 +120,13 @@ const StdcmOperationalPoint = ({
     setSearchTerm(e.target.value);
     if (e.target.value.trim().length === 0) {
       dispatchNewPoint(undefined);
+    } else {
+      setSuggestionsVisible(true);
     }
+  };
+
+  const handleBlur = () => {
+    setSuggestionsVisible(false);
   };
 
   useEffect(() => {
@@ -141,8 +147,9 @@ const StdcmOperationalPoint = ({
           label={t('trainPath.ci')}
           value={searchTerm}
           onChange={onInputChange}
+          onBlur={handleBlur}
           autoComplete="off"
-          suggestions={operationalPointsSuggestions}
+          suggestions={suggestionsVisible ? operationalPointsSuggestions : []}
           disabled={disabled}
           getSuggestionLabel={(option: Option) => option?.label}
           onSelectSuggestion={onSelectSuggestion}
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
@@ -35,6 +35,10 @@ const StdcmOperationalPoint = ({
   const { searchTerm, chCodeFilter, sortedSearchResults, setSearchTerm, setChCodeFilter } =
     useSearchOperationalPoint({ initialSearchTerm: point?.name, initialChCodeFilter: point?.ch });
 
+  const handleBlur = () => {
+    setSuggestionsVisible(false);
+  };
+
   const operationalPointsSuggestions = useMemo(
     () =>
       sortedSearchResults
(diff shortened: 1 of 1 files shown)


front/src/applications/stdcmV2/components/StdcmOperationalPoint.tsx
```diff
@@ -141,8 +147,9 @@ const StdcmOperationalPoint = ({
           label={t('trainPath.ci')}
           value={searchTerm}
           onChange={onInputChange}
+          onBlur={handleBlur}
           autoComplete="off"
-          suggestions={operationalPointsSuggestions}
+          suggestions={suggestionsVisible ? operationalPointsSuggestions : []}
```
Akctarus:
```suggestion
          suggestions={suggestionsVisible && operationalPointsSuggestions || []}
```
When we have this kind of line, what do you recommend @emersion ?

Akctarus (commented):
Tested, We still have a frame in which the old filter is still displayed when a new letter is typed, but it disappears immediately, seems fine to me

front/src/applications/stdcmV2/components/StdcmOperationalPoint.tsx
```diff
@@ -120,9 +120,15 @@ const StdcmOperationalPoint = ({
     setSearchTerm(e.target.value);
     if (e.target.value.trim().length === 0) {
       dispatchNewPoint(undefined);
+    } else {
+      setSuggestionsVisible(true);
     }
   };
 
+  const handleBlur = () => {
+    setSuggestionsVisible(false);
```
theocrsb: I think this should fix the little bug @Akctarus noticed: 

```suggestion
    setSuggestionsVisible(false);
     setSearchTerm('');
```

front/src/applications/stdcmV2/components/StdcmOperationalPoint.tsx
emersion: Personally I find the ternary more straightforward. I find it tricky to remember associativity rules for `&&`/`||` each time.

front/src/applications/stdcmV2/components/StdcmOperationalPoint.tsx
```diff
@@ -30,13 +30,13 @@ const StdcmOperationalPoint = ({
   disabled,
 }: StdcmOperationalPointProps) => {
   const { t } = useTranslation('stdcm');
+  const [suggestionsVisible, setSuggestionsVisible] = useState(false);
 
   const { searchTerm, chCodeFilter, sortedSearchResults, setSearchTerm, setChCodeFilter } =
     useSearchOperationalPoint({ initialSearchTerm: point?.name, initialChCodeFilter: point?.ch });
 
   const operationalPointsSuggestions = useMemo(
```
emersion: I think this `useMemo` is missing a dependency on `searchTerm`? Same deal with `sortedChOptions` below.

(With that fixed, maybe we don't need `suggestionsVisible` at all?)

front/src/applications/stdcmV2/components/StdcmOperationalPoint.tsx
RomainValls: I don't think it's possible. Clicking on a suggestion will select it before instantly deleting the selected value. 
Maybe using something else than onBlur would solve it, but I don't know what else to use.

front/src/applications/stdcmV2/components/StdcmOperationalPoint.tsx
RomainValls: Tried adding the missing dependencies on operationalPointsSuggestions and sortedChOptions, the behavior of the input remaining in memory is still here.

theocrsb (changes requested):
There's still the problem @Akctarus pointed out

RomainValls:
Added a dependency to operationalPointsSuggestions, previous input value isn't visible anymore while typing a new value. 
There's still a frame, after typing in the input, where the suggestions seem to be sorted again, but this is the behavior in dev.

front/src/applications/stdcmV2/components/StdcmOperationalPoint.tsx
theocrsb: I tried again and it works for me
```
  const handleBlur = () => {
    setSuggestionsVisible(false);
    setSearchTerm('');
  };
```

RomainValls: Clicking on a suggestion will delete it right after. With the onBlur, selecting a value makes the combo box lose the focus and triggers setSearchTerm(''). 
Maybe we should use another method. 
But the fix still to be working now.

front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
```diff
@@ -53,7 +57,7 @@ const StdcmOperationalPoint = ({
           if (!isDuplicate) acc.push(newObject);
           return acc;
         }, [] as Option[]),
-    [sortedSearchResults]
+    [sortedSearchResults, handleBlur]
```
emersion, line 60: `handleBlur` is not used inside this `useMemo`, and since a new function is created each time the component is re-rendered, this effectively always re-computes `operationalPointsSuggestions`, making `useMemo` useless.

emersion (commented):
When typing something, then clicking elsewhere, it seems like the first result is selected. From the issue, the expected result is that the input field is cleared.

front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
```diff
@@ -1,4 +1,4 @@
-import { useEffect, useMemo } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 
 import { Select, ComboBox } from '@osrd-project/ui-core';
```
achrafmohye, line 3: I think if we want to handle the case of parent's onChange passed to the ComboBox, this bug must be fixed in `osrd-ui `

achrafmohye: Here is a [draft](https://github.com/OpenRailAssociation/osrd-ui/pull/695) to fix this in` osrd-ui`

clarani:
Not relevant anymore, the bug was fixed by this PR: https://github.com/OpenRailAssociation/osrd/pull/10293

PR #9464, opened 2024-10-24
Use standard oidc crate

flomonster:
We don't longer need to handle epoch as string.

### DIFF
--- a/gateway/Cargo.toml
+++ b/gateway/Cargo.toml
@@ -52,10 +52,7 @@ phf = "0.11"
 # actix_auth
 actix-web-httpauth = "0.8"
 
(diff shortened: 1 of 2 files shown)


ghost (approved):

Khoyo (approved):

gateway/Cargo.toml
```diff
@@ -53,9 +53,7 @@ phf = "0.11"
 actix-web-httpauth = "0.8"
 
 # switch to https://github.com/ramosbugs/openidconnect-rs/pull/165 if it ever lands
```
emersion: We can probably drop this comment now?

PR #9465, opened 2024-10-24
editoast: add 'label' to towed rolling stock

woshilapin:
Follow up of #9202. We're missing a long name. I used `label`.

### DIFF
--- a/editoast/migrations/2024-10-24-143158_add-towed-rolling-stock-field-description/up.sql
+++ b/editoast/migrations/2024-10-24-143158_add-towed-rolling-stock-field-description/up.sql
@@ -1,0 +1,2 @@
+ALTER TABLE towed_rolling_stock
+ADD label VARCHAR(255) NOT NULL DEFAULT '';
--- a/editoast/editoast_models/src/tables.rs
+++ b/editoast/editoast_models/src/tables.rs
@@ -696,6 +696,8 @@ diesel::table! {
         rolling_resistance -> Jsonb,
         gamma -> Jsonb,
         version -> Int8,
+        #[max_length = 255]
+        label -> Varchar,
     }
 }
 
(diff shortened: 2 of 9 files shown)


editoast/migrations/2024-10-24-143158_add-towed-rolling-stock-field-description/up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE towed_rolling_stock
+ADD description VARCHAR(255) NOT NULL DEFAULT '';
```
leovalais: Is `description` meant to be a free field, editable if we had a towed rs editor? If so shouldn't it be a `TEXT` instead?

We can change that later if we need it ofc, feel free to disregard this comment :)

leovalais (approved):
LGTM

editoast/migrations/2024-10-24-143158_add-towed-rolling-stock-field-description/up.sql
woshilapin: I'm not entirely sure when we should use `VARCHAR` or `TEXT` and I feel you do know more. It's more a human readable label. For example, `name` might be something like "ABX126" and `description` might be `FREIGHT CAR FOR 126 TONS". I do not think the string will be very long. Do you feel `TEXT` is needed? I've read a bit on the subject and `VARCHAR` is probably ok. But thanks for getting that to my attention, at the very least, I learned something.

woshilapin: Note: your comment might be a sign that `label` would be a better name than `description`,  what do you think?

leovalais: Alright then maybe not `description` indeed. `label` seems fine, I'm also thinking of `long_name`, your call :)

woshilapin: Done. I've chosen `label`.

Issue #9351, opened 2024-10-15
Wrong translation in list of trains "selected trains" after importing trains


louisgreiner:
### What happened?

The language is set to english

![Image](https://github.com/user-attachments/assets/d6fca20b-3aaa-4d85-a2db-f5dc319fcdde)

EDIT: the bug is inconsistent, on the same infra after importing trains from Viriato file:

![Image](https://github.com/user-attachments/assets/794914e3-3587-4ba9-9e42-2561174b81c6)

And if I delete all + create manually a train, the bug appears again:

![Image](https://github.com/user-attachments/assets/63fa6b72-44f9-4307-9384-05cfdd6c36f7)

### What did you expect to happen?

Have an actual english translation here

### How can we reproduce it (as minimally and precisely as possible)?

1. open scenario and select a train

### On which environments the bug occurs?

Recette (SNCF)

### On which browser the bug occurs?

Firefox

### OSRD version (top right corner `Account` button > `Informations`)

staging e66e534

PR #9471, opened 2024-10-24
front: fix pluralization in scenario train selection

### LINKED ISSUE #9470: front: pluralization is wrong in scenario train selection
### What happened?

The train word is not pluralized based on the number of trains in scenario train selection

_No response_

Create a scenario with a single trainschedule. 
Look at the train selection panel.

Recette (SNCF)

Firefox

Unknown

Synar:
Close #9470


flomonster (approved):

theocrsb (approved):
LGTM and tested :)

front/public/locales/en/operationalStudies/scenario.json
```diff
@@ -82,8 +82,11 @@
     "validityFilter": "Trains validity"
   },
   "toggleTimetable": "Toggle the timetable",
-  "trainCount_other": "{{ count }} out of {{ totalCount }} trains selected",
-  "trainCount_zero": "{{ totalCount }} trains",
+  "trainCount_one": "1 out of $t(train, { 'count': {{totalCount}} }) selected",
```
emersion, line 85: Wouldn't this show "1 out of train selected"?

emersion: Ah, no, nevermind!

emersion, line 85: Nit: do we really need the `trainCount_one` special case? Wouldn't this be taken care of by `trainCount_other`?

Synar: This is what I tried first but then something... peculiar happened
[Screencast_20241025_161918.webm](https://github.com/user-attachments/assets/c41dbb06-e29c-4508-b034-1a606ba38379)

Right then and there I decided I would save myself a headache by just being a bit more exhaustive

emersion: Oh my.

One less headache sounds good to me!

emersion (approved):

flomonster:
Also close #9351

PR #9474, opened 2024-10-25
core: remove rolling stock length from safety speeds

eckter:
Safety speeds end when the train *head* reaches the end, and not its *tail* like all other speed limits.


Erashin (approved):

Erashin:
You have some unit tests failing btw (not so surprising).

bougue-pe (approved):
LGTM, especially if the test fix is to remove RS length here: https://github.com/OpenRailAssociation/osrd/blob/a54923883baad47b9803c4dc97794176481e4139/core/src/test/kotlin/fr/sncf/osrd/standalone_sim/StandaloneSimulationTest.kt#L351

bougue-pe (approved):

SharglutDev (approved):

Maymanaf (approved):

PR #9475, opened 2024-10-25
editoast: fix work schedule filter

eckter:
In the previous version, we only considered the range from (earliest start time) to (earliest start time + maximum running time), which ignores the departure time window.

For example, when we looked for a 1h long path at `t+-4h`, we only considered work schedules that overlapped `[t-4h, t-3h]`.

While I was there, I renamed a few functions. There were 3 "filter work schedules" functions that don't quite do that.

### DIFF
--- a/editoast/src/views/timetable/stdcm.rs
+++ b/editoast/src/views/timetable/stdcm.rs
@@ -1111,4 +1111,45 @@ mod tests {
             }
         );
     }
+
+    #[rstest]
+    // A day before the 'start_time' -> FILTERED OUT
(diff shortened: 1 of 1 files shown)


leovalais (approved):
LGTM. Ping @hamz2a, you may have some conflicts with your work.

hamz2a (approved):
LGTM, thanks!

woshilapin (approved):
Nice catch and thanks for the function renaming. Some tests might be good, I pushed a commit for that, feel free to edit, modify, improve.

PR #9478, opened 2024-10-25
editoast: remove always-on directive

woshilapin:
This modification was done in order to bring logs for any request that comes in thanks to the `tower_http::trace::TraceLayer` middleware. But the code has a sneaky behavior where it's actually impossible to turn it of anymore.

For example, `RUST_LOG=info,tower_http=none` will still show `tower_http` logs.

Let's remove it.


flomonster (approved):
Thanks, should we add an env var to the dockerfile to ensure logs in prod?

woshilapin:
> Thanks, should we add an env var to the dockerfile to ensure logs in prod?

Good point. Done in [37a05a8](https://github.com/OpenRailAssociation/osrd/pull/9478/commits/37a05a84a37ca05587dff25195c5fb4e9b4c43b8).

PR #9480, opened 2024-10-25
editoast: do not panic on incorrect pagination

woshilapin:
Once, I did put a `page` and `page_size` that went out of bound... and got a 500. I don't expect `editoast` to panic in such case, and we could reply with either an empty `Vec` and a 200, or more likely something like a 404 (I chose this one). Note that on the way, I also remove panic for `page==0` and `page_size==0` too and transformed them into a 400.

Note that I renamed and change the semantic of the existing `InvalidPageSize` into `PageSizeTooBig` (because it was actually not triggered when `page_size==0`) and then I created a **new** `InvalidPageSize` that actually only check for a strictly positive number.

Here is the result.
```
 ❯ xh 'localhost:8090/light_rolling_stock?page_size=100&page=100'
HTTP/1.1 404 Not Found
Access-Control-Allow-Origin: *
Content-Length: 214
Content-Type: application/json
Date: Fri, 25 Oct 2024 15:21:28 GMT
Vary: origin, access-control-request-method, access-control-request-headers

{
    "status": 404,
    "type": "editoast:pagination:OutOfBound",
    "context": {
        "page_size": 100,
        "total_count": 464,
        "page": 100
    },
    "message": "no more information after page 100 when page size is 100 (total number of elements is 464)"
}
```

### DIFF
--- a/editoast/openapi.yaml
+++ b/editoast/openapi.yaml
@@ -4240,7 +4240,10 @@ components:
       - $ref: '#/components/schemas/EditoastOperationErrorInvalidPatch'
       - $ref: '#/components/schemas/EditoastOperationErrorModifyId'
       - $ref: '#/components/schemas/EditoastOperationErrorObjectNotFound'
+      - $ref: '#/components/schemas/EditoastPaginationErrorInvalidPageNumber'
       - $ref: '#/components/schemas/EditoastPaginationErrorInvalidPageSize'
+      - $ref: '#/components/schemas/EditoastPaginationErrorPageOutOfBound'
+      - $ref: '#/components/schemas/EditoastPaginationErrorPageSizeTooBig'
       - $ref: '#/components/schemas/EditoastPathfindingErrorInfraNotFound'
       - $ref: '#/components/schemas/EditoastPathfindingViewErrorsEndingTrackLocationNotFound'
       - $ref: '#/components/schemas/EditoastPathfindingViewErrorsInvalidNumberOfPaths'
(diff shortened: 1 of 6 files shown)


leovalais (commented):
I'm a bit conflicted about this change. On one hand replacing `assert`s with proper error handling is a good thing, but in this case one of these `assert`s panicking is more an indicator of a bug rather than a user error I believe. The workflow for a paginated endpoint is:

1. Receive pagination parameters: `PaginationQueryParams`
2. Validate these, returning an error on failure (the former `InvalidPageSize`)
3. Process the data (using `SelectionSettings` or not)
4. Build the response, that we enhance by returning some info on the pagination, `PaginationStats`

My point being that once we call `Pagination::new`, the `page` and `page_size` should have been validated already since we started processing the request. And they are (`PaginationQueryParams::validate`). So there is no reason for the errors to be triggered, unless it's a bug on our end. As for `current_page_count`, this value is always provided algorithmically, so an inconsistent value is also an implementation bug. And in an internal bug situation I believe panicking and thus returning `Internal Server Error` is the most correct behavior 🤷

However, the `PageOutOfBound` error shouldn't have been an `assert`, thanks for fixing that! 😄 

I'm not sure if that's convincing at all though. Here's my proposition:

1. Turning the `InvalidPageNumber` into a 5xx to indicate the user their request was valid
2. Merge `PageSizeTooBig` (which name is misleading since it's also descriptive of the case `page_size == 0`, cf. `validate`) and `InvalidPageSize` (the new one)
3. Maybe comment in `new` that the first two errors are not triggered because of a user error, but because of a bug on our end.

Wdyt? 

---

> Note that I renamed and change the semantic of the existing InvalidPageSize into PageSizeTooBig (because it was actually not triggered when page_size==0)

I'd say it works as expected, with which request did you trigger this situation?

```
~
❯ xh ":8090/light_rolling_stock?page_size=0"
HTTP/1.1 400 Bad Request
Access-Control-Allow-Origin: *
Content-Length: 192
Content-Type: application/json
Date: Mon, 28 Oct 2024 13:38:53 GMT
Vary: origin, access-control-request-method, access-control-request-headers

{
    "status": 400,
    "type": "editoast:pagination:InvalidPageSize",
    "context": {
        "provided_page_size": 0,
        "max_page_size": 1000
    },
    "message": "Invalid page size (0), expected an integer 0 < page_size <= 1000"
}
```

flomonster:
Can we close this PR?

PR #9481, opened 2024-10-28
front: redesign map in result page

theocrsb:
closes part of https://github.com/osrd-project/osrd-confidential/issues/670

For the redesign of the buttons I'm going to do another PR because this part is optional and I don't have the mockup.


front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
```diff
@@ -11,14 +13,36 @@ interface RenderItineraryProps {
...
+      <OrderedLayer
+        id="geojsonPathBackgroundLine"
+        type="line"
+        paint={paintBackgroundLine}
+        beforeId="geojsonPathLine"
+        layerOrder={layerOrder}
+      />
+      <OrderedLayer id="geojsonPathLine" type="line" paint={paintLine} layerOrder={layerOrder} />
+      <Marker longitude={endCoordinate[0]} latitude={endCoordinate[1]} anchor="bottom">
+        <img src={stdcmDestination} alt="destination" />
```
Akctarus:
```suggestion
        <img src={DestinationIcon} alt="destination" />
```

front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
```diff
@@ -1,6 +1,8 @@
 import type { Feature, LineString } from 'geojson';
-import { Source } from 'react-map-gl/maplibre';
+import { Marker, Source } from 'react-map-gl/maplibre';
 
+import stdcmDestination from 'assets/pictures/stdcm/destination.svg';
+import OriginIcon from 'assets/pictures/stdcm/start.svg';
```
Akctarus: Since it is no longer only used in the stdcm map, perhaps you could rename the folder "mapMarkers" or something like that ?

Akctarus (commented):
Tested, LGTM, I think STDCM map needs to be updated too to match this redesign. Left 2 comments

front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
theocrsb: Done

front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
theocrsb: Done

Akctarus (approved):

front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
```diff
@@ -11,14 +13,36 @@ interface RenderItineraryProps {
...
-  const paint = {
-    'line-width': 3,
-    'line-color': '#82be00',
+  const paintBackgroundLine = {
+    'line-width': 4,
+    'line-color': '#EDF9FF',
   };
 
+  const paintLine = {
+    'line-width': 1,
```
SharglutDev, line 22: Capture d’écran 2024-10-31 à 12 27 19

```suggestion
    'line-width': 3,
```

front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
```diff
@@ -11,14 +13,36 @@ interface RenderItineraryProps {
...
+    'line-color': '#158DCF',
+  };
+
+  const startCoordinate = geojsonPath.geometry.coordinates[0];
+  const endCoordinate =
+    geojsonPath.geometry.coordinates[geojsonPath.geometry.coordinates.length - 1];
```
clarani: - use .at() instead of `[...]`
- you need to handle the case where the coodinates are undefined
```suggestion
  const startCoordinate = geojsonPath.geometry.coordinates.at(0);
  const endCoordinate =
    geojsonPath.geometry.coordinates.at(- 1);
```

clarani (commented):
This PR will also add the origin/destination markers to the warped map, is it what we want ?

SharglutDev:
> This PR will also add the origin/destination markers to the warped map, is it what we want ?

Should we ask a PO since @thibautsailly isn't here ?

front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
theocrsb: Done

front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
theocrsb: I have something else in the mockup:  
image

theocrsb:
> > This PR will also add the origin/destination markers to the warped map, is it what we want ?
> 
> Should we ask a PO since @thibautsailly isn't here ?

I asked to @Tguisnet

front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
maelysLeratRosso: Seems good to me to have them on the wrapped map as well.
Anyway, I think we will have to think about this wrapped map globally so we can adjust it then. Furthermore, since the map is backwards compared to the Y-axis of the time-space chart, the departure and arrival icons limit the confusion risk.

front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
SharglutDev: Oh right, I looked at the wrong map mockup

SharglutDev (approved):
Lgtm and tested

SharglutDev:
Actually the map mockup says this : 
Capture d’écran 2024-10-31 à 13 25 19
So I think we should also display the vias

clarani (approved):
LGTM :white_check_mark:

Tguisnet:
`This PR will also add the origin/destination markers to the warped map, is it what we want ?`

There are no real specifications for the warp map, so it's complicated to know what we want to have on it.
But the warp map is currently reversed (from the implementation of the new GET), so at least we could see how its works now with this markers :).

I think we could merge this PR, there are a lot of adjustements coming soon for the result page, the warp map probably won't follow all of the changes, so we'll have to specify what we want for the warp map, but later.

PR #9484, opened 2024-10-28
editoast: add trigger for search_scenario table population

### LINKED ISSUE #8612: front: cannot search the scenario
### What happened?

I try to filter the scenario with the name or the label and no scenario is displayed

I should see the correct scenario

1. Navigate to a study with several scenari
2. Try to filter the scenari with its name or label

Local

Firefox

1482e8c

EthanPERRUZZA:
fixes : https://github.com/OpenRailAssociation/osrd/issues/8612#issue-2494431637


editoast/src/views/search.rs
```diff
@@ -654,14 +654,18 @@ pub(super) struct SearchResultItemStudy {
...
     joins = "
         INNER JOIN scenario ON scenario.id = search_scenario.id
         INNER JOIN infra ON infra.id = scenario.infra_id",
-    column(name = "id", data_type = "integer"),
-    column(name = "name", data_type = "string"),
-    column(name = "description", data_type = "string"),
-    column(name = "tags", data_type = "string"),
-    column(name = "study_id", data_type = "integer")
+    column(name = "name", data_type = "TEXT", sql = "scenario.name"),
+    column(name = "description", data_type = "TEXT", sql = "scenario.description"),
```
leovalais:
```suggestion
    column(name = "name", data_type = "TEXT", sql = "scenario.name", textual_search),
    column(name = "description", data_type = "TEXT", sql = "scenario.description", textual_search),
```

Signals that the column should be indexed to allow [full text search](https://www.postgresql.org/docs/current/textsearch.html) and more, including accent processing. Otherwise you can't find a scenario named "mon scénario" using "scé".

flomonster (commented):
LGTM (with Léo's change)

leovalais (approved):
LGTM, to please the CI you'll need to format your commits and run `cargo fmt` (probably best to setup you editor to format on save ;)

PR #9486, opened 2024-10-28
core: stdcm: add rolling stock length to speed limits in output sim

### LINKED ISSUE #9458: stdcm algorithm seems to accelerate before the tail of the train leaves a speed limit
### What happened?

The train is accelerating even though the tail has not exited the speed section:
Image

It should behave the same way as in the operational studies : the train is supposed to be allowed to accelerate only after the tail leaves the speed section. Visually, the spped should increase after the small vertical line on the GEV.

1. Get any result that reaches speed limits in STDCM debugging mode

Recette (SNCF)

Google Chrome

f91b8ec

eckter:
fix https://github.com/OpenRailAssociation/osrd/issues/9458

todo: write tests, fix current tests

Note: this is very likely to introduce bugs and will require https://github.com/OpenRailAssociation/osrd/issues/9487 to be "correctly" implemented. I don't know if it's worth merging the "potentially bugged" version before then.


flomonster:
Can we close this PR?

eckter:
Oh. I forgot about this.

Yeah we can close this, but it's important not to forget (again). Related to https://github.com/OpenRailAssociation/osrd/pull/14755 and adjacent issues.

PR #9489, opened 2024-10-28
front: project trains only in debug mode in stdcm

### LINKED ISSUE #9402: STDCM: even without debug mode calls to path_projection is done in the stdcm page
### What happened?

Open STDCM page and create a train.
Open the network you will see requests to path projection even if debug mode is not opened.

No calls if no space time chart

See above

Recette (SNCF), Dev (SNCF)

Firefox

dev 698b8f4

clarani:
closes #9402


Math-R (approved):
lgtm =)

PR #9490, opened 2024-10-28
front: add lazy loader message on GET in stdcm

### LINKED ISSUE #8610: Missing lazy loader info in STDCM GET
### What happened?

This info is missing in STDCM GET
image

_No response_

.

Recette (SNCF)

Google Chrome

1482e8c

clarani:
closes #8610

image

### DIFF
--- a/front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
+++ b/front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
@@ -26,6 +26,7 @@ import type {
 
 import SettingsPanel from './SettingsPanel';
 import ManchetteMenuButton from '../SpaceTimeChart/ManchetteMenuButton';
+import ProjectionLoadingMessage from '../SpaceTimeChart/ProjectionLoadingMessage';
 import WaypointsPanel from '../SpaceTimeChart/WaypointsPanel';
 
 type ManchetteWithSpaceTimeChartProps = {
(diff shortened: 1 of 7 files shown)


front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -29,6 +30,8 @@ type ManchetteWithSpaceTimeChartProps = {
   waypointsPanelData?: WaypointsPanelData;
   conflicts?: Conflict[];
   workSchedules?: PostWorkSchedulesProjectPathApiResponse;
+  totalTrains: number;
+  allTrainsProjected: boolean;
```
SharglutDev: As they both work together and are only here for the projection loader message, should we put them in an object to have less props to pass across all components ?

SharglutDev (approved):
Lgtm and tested, good job ! Left just one suggestion, you can resolve yourself after treating it or not :)

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
clarani: Yes we can, but how would you name this object ?

SharglutDev: `ProjectionLoaderData` ? Like we do for other part of the simulation

PR #9507, opened 2024-10-29
front: bump osrd-ui to use the new manchette props

### LINKED ISSUE #9506: front: bump osrd-ui to 53

clarani:
closes #9506


emersion:
Note, these two commits should be squashed together because the app won't build in-between.

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -52,90 +52,101 @@ const ManchetteWithSpaceTimeChartWrapper = ({
...
+      name: waypoint.extensions?.identifier?.name,
+      secondaryCode: waypoint.extensions?.sncf?.ch,
+    }));
+  }, [waypointsPanelData, operationalPoints]);
 
   const { manchetteProps, spaceTimeChartProps, handleScroll } = useManchettesWithSpaceTimeChart(
-    waypointsPanelData?.filteredWaypoints ?? operationalPoints,
-    spaceTimeChartLayersData.filteredProjectPathTrainResult,
+    manchetteWaypoints,
+    cutProjectedTrains,
```
SharglutDev, line 149: Here we pass the projected trains with `signalUpdates` but the hook doesn't use them. Typescript doesn't seem to mind that there is an extra property in its wanted type but maybe we should remove it ?

front/src/applications/stdcm/utils/formatStdcmIntoSpaceTimeData.ts
```diff
@@ -1,24 +1,23 @@
-import type { ProjectPathTrainResult } from 'common/api/osrdEditoastApi';
+import type { TrainSpaceTimeData } from 'applications/operationalStudies/types';
```
SharglutDev, line 1: Is it ok that stdcm depends on a type from operational studies ? We also have to deal with `signalUpdates` even though we don't need them

SharglutDev (commented):
Thank you for the PR. Everything works great. Just left a few comments/suggestions

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
clarani: Yes, I think typescript only checks that the required props are given, it does not mind if extra props are given.
To remove the extra props, we will have to add a map. I think it's not very useful and we can stay like this, if it's good for you

front/src/applications/stdcm/utils/formatStdcmIntoSpaceTimeData.ts
clarani: No I think that the type TrainSpaceTimeData should be in module/SpaceTimeChart :/
As it was already there, I did not move it. Do you want me to move it now ? I can also do it in another PR.

Concerning the signalUpdates, I think we should always keep it in the TrainSpaceTimeData type, as we might use it every time we are displaying the trains on the SpaceTimeChart

clarani:
@emersion why do you think we should squash the 2 commits ? I can build after the 1st commit (without the 2nd one) in local

emersion:
Hm, right… The first commit bumps half of osrd-ui and the other bumps the rest. I guess that's fine.

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
SharglutDev: Fine by me

front/src/applications/stdcm/utils/formatStdcmIntoSpaceTimeData.ts
SharglutDev: Yep, it can be done in a different PR :)

SharglutDev (approved):

PR #9434, opened 2024-10-23
editoast: explanation of why a simulation is not feasible refactor editoast

hamz2a:
closes https://github.com/OpenRailAssociation/osrd/issues/9476
Refactor error handling and method organization
- Simplified how `STDCMResponse::PathNotFound` error is handled.
- Reorganized `STDCMRequestPayload` methods for better structure.
- Moved `simulation_run_time` to `SimulationResponse` for clarity.

PR #9419, opened 2024-10-22
Add logic to handle non-feasibility in stdcm simulations

achrafmohye:
Closes #9163

Can only be merged once this PR is merged: https://github.com/OpenRailAssociation/osrd/pull/9509

PR #9509, opened 2024-10-29
editoast: refactoring STDCM handler

### LINKED ISSUE #9476: editoast: refactoring STDCM handler
### Description and goal

Refactored STDCM handler and addressed comments in this [PR](https://github.com/OpenRailAssociation/osrd/pull/9327).

### Acceptance criteria

Fix all comments.

hamz2a:
closes https://github.com/OpenRailAssociation/osrd/issues/9476

### DIFF
--- a/editoast/openapi.yaml
+++ b/editoast/openapi.yaml
@@ -2630,14 +2630,6 @@ paths:
                       - success
                 - type: object
                   required:
-                  - status
-                  properties:
-                    status:
-                      type: string
-                      enum:
-                      - path_not_found
-                - type: object
-                  required:
                   - pathfinding_result
                   - conflicts
                   - status
@@ -4265,11 +4257,6 @@ components:
       - $ref: '#/components/schemas/EditoastRollingStockErrorKeyNotFound'
       - $ref: '#/components/schemas/EditoastRollingStockErrorLiveryMultipartError'
       - $ref: '#/components/schemas/EditoastRollingStockErrorNameAlreadyUsed'
-      - $ref: '#/components/schemas/EditoastSTDCMErrorInfraNotFound'
-      - $ref: '#/components/schemas/EditoastSTDCMErrorInvalidPathItems'
-      - $ref: '#/components/schemas/EditoastSTDCMErrorRollingStockNotFound'
-      - $ref: '#/components/schemas/EditoastSTDCMErrorTimetableNotFound'
-      - $ref: '#/components/schemas/EditoastSTDCMErrorTowedRollingStockNotFound'
       - $ref: '#/components/schemas/EditoastScenarioErrorInfraNotFound'
       - $ref: '#/components/schemas/EditoastScenarioErrorNotFound'
       - $ref: '#/components/schemas/EditoastScenarioErrorTimetableNotFound'
@@ -4277,6 +4264,11 @@ components:
       - $ref: '#/components/schemas/EditoastSearchApiErrorSearchEngineError'
       - $ref: '#/components/schemas/EditoastSpriteErrorsFileNotFound'
       - $ref: '#/components/schemas/EditoastSpriteErrorsUnknownSignalingSystem'
+      - $ref: '#/components/schemas/EditoastStdcmErrorInfraNotFound'
+      - $ref: '#/components/schemas/EditoastStdcmErrorInvalidPathItems'
+      - $ref: '#/components/schemas/EditoastStdcmErrorRollingStockNotFound'
+      - $ref: '#/components/schemas/EditoastStdcmErrorTimetableNotFound'
+      - $ref: '#/components/schemas/EditoastStdcmErrorTowedRollingStockNotFound'
       - $ref: '#/components/schemas/EditoastStudyErrorNotFound'
       - $ref: '#/components/schemas/EditoastStudyErrorStartDateAfterEndDate'
       - $ref: '#/components/schemas/EditoastTemporarySpeedLimitErrorNameAlreadyUsed'
@@ -5164,126 +5156,6 @@ components:
           type: string
           enum:
           - editoast:rollingstocks:NameAlreadyUsed
-    EditoastSTDCMErrorInfraNotFound:
-      type: object
-      required:
-      - type
-      - status
-      - message
-      properties:
-        context:
-          type: object
-          required:
-          - infra_id
-          properties:
-            infra_id:
-              type: integer
-        message:
-          type: string
-        status:
-          type: integer
-          enum:
-          - 400
-        type:
-          type: string
-          enum:
-          - editoast:stdcm_v2:InfraNotFound
-    EditoastSTDCMErrorInvalidPathItems:
-      type: object
-      required:
-      - type
-      - status
-      - message
-      properties:
-        context:
-          type: object
-          required:
-          - items
-          properties:
-            items:
-              type: array
-        message:
-          type: string
-        status:
-          type: integer
-          enum:
-          - 400
-        type:
-          type: string
-          enum:
-          - editoast:stdcm_v2:InvalidPathItems
-    EditoastSTDCMErrorRollingStockNotFound:
-      type: object
-      required:
-      - type
-      - status
-      - message
-      properties:
-        context:
-          type: object
-          required:
-          - rolling_stock_id
-          properties:
-            rolling_stock_id:
-              type: integer
-        message:
-          type: string
-        status:
-          type: integer
-          enum:
-          - 400
-        type:
-          type: string
-          enum:
-          - editoast:stdcm_v2:RollingStockNotFound
-    EditoastSTDCMErrorTimetableNotFound:
-      type: object
-      required:
-      - type
-      - status
-      - message
-      properties:
-        context:
-          type: object
-          required:
-          - timetable_id
-          properties:
-            timetable_id:
-              type: integer
-        message:
-          type: string
-        status:
-          type: integer
-          enum:
-          - 404
-        type:
-          type: string
-          enum:
-          - editoast:stdcm_v2:TimetableNotFound
-    EditoastSTDCMErrorTowedRollingStockNotFound:
-      type: object
-      required:
-      - type
-      - status
-      - message
-      properties:
-        context:
-          type: object
-          required:
-          - towed_rolling_stock_id
-          properties:
-            towed_rolling_stock_id:
-              type: integer
-        message:
-          type: string
-        status:
-          type: integer
-          enum:
-          - 400
-        type:
-          type: string
-          enum:
-          - editoast:stdcm_v2:TowedRollingStockNotFound
     EditoastScenarioErrorInfraNotFound:
       type: object
       required:
@@ -5447,6 +5319,126 @@ components:
           type: string
           enum:
           - editoast:sprites:UnknownSignalingSystem
+    EditoastStdcmErrorInfraNotFound:
+      type: object
+      required:
+      - type
+      - status
+      - message
+      properties:
+        context:
+          type: object
+          required:
+          - infra_id
+          properties:
+            infra_id:
+              type: integer
+        message:
+          type: string
+        status:
+          type: integer
+          enum:
+          - 400
+        type:
+          type: string
+          enum:
+          - editoast:stdcm_v2:InfraNotFound
+    EditoastStdcmErrorInvalidPathItems:
+      type: object
+      required:
+      - type
+      - status
+      - message
+      properties:
+        context:
+          type: object
+          required:
+          - items
+          properties:
+            items:
+              type: array
+        message:
+          type: string
+        status:
+          type: integer
+          enum:
+          - 400
+        type:
+          type: string
+          enum:
+          - editoast:stdcm_v2:InvalidPathItems
+    EditoastStdcmErrorRollingStockNotFound:
+      type: object
+      required:
+      - type
+      - status
+      - message
+      properties:
+        context:
+          type: object
+          required:
+          - rolling_stock_id
+          properties:
+            rolling_stock_id:
+              type: integer
+        message:
+          type: string
+        status:
+          type: integer
+          enum:
+          - 400
+        type:
+          type: string
+          enum:
+          - editoast:stdcm_v2:RollingStockNotFound
+    EditoastStdcmErrorTimetableNotFound:
+      type: object
+      required:
+      - type
+      - status
+      - message
+      properties:
+        context:
+          type: object
+          required:
+          - timetable_id
+          properties:
+            timetable_id:
+              type: integer
+        message:
+          type: string
+        status:
+          type: integer
+          enum:
+          - 404
+        type:
+          type: string
+          enum:
+          - editoast:stdcm_v2:TimetableNotFound
+    EditoastStdcmErrorTowedRollingStockNotFound:
+      type: object
+      required:
+      - type
+      - status
+      - message
+      properties:
+        context:
+          type: object
+          required:
+          - towed_rolling_stock_id
+          properties:
+            towed_rolling_stock_id:
+              type: integer
+        message:
+          type: string
+        status:
+          type: integer
+          enum:
+          - 400
+        type:
+          type: string
+          enum:
+          - editoast:stdcm_v2:TowedRollingStockNotFound
     EditoastStudyErrorNotFound:
       type: object
       required:
@@ -8226,6 +8218,108 @@ components:
             type: integer
             format: int64
             minimum: 0
+    Request:
+      type: object
+      description: An STDCM request
+      required:
+      - steps
+      - rolling_stock_id
+      - comfort
+      properties:
+        comfort:
+          $ref: '#/components/schemas/Comfort'
+        electrical_profile_set_id:
+          type: integer
+          format: int64
+          nullable: true
+        loading_gauge_type:
+          allOf:
+          - $ref: '#/components/schemas/LoadingGaugeType'
+          nullable: true
+        margin:
+          type: string
+          description: Can be a percentage `X%`, a time in minutes per 100 kilometer `Xmin/100km`
+          example:
+          - 5%
+          - 2min/100km
+          nullable: true
+        max_speed:
+          type: number
+          format: double
+          description: Maximum speed of the consist in km/h
+          nullable: true
+        maximum_departure_delay:
+          type: integer
+          format: int64
+          description: |-
+            By how long we can shift the departure time in milliseconds
+            Deprecated, first step data should be used instead
+          nullable: true
+          minimum: 0
+        maximum_run_time:
+          type: integer
+          format: int64
+          description: |-
+            Specifies how long the total run time can be in milliseconds
+            Deprecated, first step data should be used instead
+          nullable: true
+          minimum: 0
+        rolling_stock_id:
+          type: integer
+          format: int64
+        speed_limit_tags:
+          type: string
+          description: Train categories for speed limits
+          nullable: true
+        start_time:
+          type: string
+          format: date-time
+          description: Deprecated, first step arrival time should be used instead
+          nullable: true
+        steps:
+          type: array
+          items:
+            $ref: '#/components/schemas/PathfindingItem'
+        temporary_speed_limit_group_id:
+          type: integer
+          format: int64
+          nullable: true
+        time_gap_after:
+          type: integer
+          format: int64
+          description: |-
+            Margin after the train passage in milliseconds
+
+            Enforces that the path used by the train should be free and
+            available at least that many milliseconds after its passage.
+          minimum: 0
+        time_gap_before:
+          type: integer
+          format: int64
+          description: |-
+            Margin before the train passage in seconds
+
+            Enforces that the path used by the train should be free and
+            available at least that many milliseconds before its passage.
+          minimum: 0
+        total_length:
+          type: number
+          format: double
+          description: Total length of the consist in meters
+          nullable: true
+        total_mass:
+          type: number
+          format: double
+          description: Total mass of the consist in kg
+          nullable: true
+        towed_rolling_stock_id:
+          type: integer
+          format: int64
+          nullable: true
+        work_schedule_group_id:
+          type: integer
+          format: int64
+          nullable: true
     RjsPowerRestrictionRange:
       type: object
       description: A range along the train path where a power restriction is applied.
@@ -8734,108 +8828,6 @@ components:
             type: string
         zone:
           type: string
-    STDCMRequestPayload:
-      type: object
-      description: An STDCM request
-      required:
-      - steps
-      - rolling_stock_id
-      - comfort
-      properties:
-        comfort:
-          $ref: '#/components/schemas/Comfort'
-        electrical_profile_set_id:
-          type: integer
-          format: int64
-          nullable: true
-        loading_gauge_type:
-          allOf:
-          - $ref: '#/components/schemas/LoadingGaugeType'
-          nullable: true
-        margin:
-          type: string
-          description: Can be a percentage `X%`, a time in minutes per 100 kilometer `Xmin/100km`
-          example:
-          - 5%
-          - 2min/100km
-          nullable: true
-        max_speed:
-          type: number
-          format: double
-          description: Maximum speed of the consist in km/h
-          nullable: true
-        maximum_departure_delay:
-          type: integer
-          format: int64
-          description: |-
-            By how long we can shift the departure time in milliseconds
-            Deprecated, first step data should be used instead
-          nullable: true
-          minimum: 0
-        maximum_run_time:
-          type: integer
-          format: int64
-          description: |-
-            Specifies how long the total run time can be in milliseconds
-            Deprecated, first step data should be used instead
-          nullable: true
-          minimum: 0
-        rolling_stock_id:
-          type: integer
-          format: int64
-        speed_limit_tags:
-          type: string
-          description: Train categories for speed limits
-          nullable: true
-        start_time:
-          type: string
-          format: date-time
-          description: Deprecated, first step arrival time should be used instead
-          nullable: true
-        steps:
-          type: array
-          items:
-            $ref: '#/components/schemas/PathfindingItem'
-        temporary_speed_limit_group_id:
-          type: integer
-          format: int64
-          nullable: true
-        time_gap_after:
-          type: integer
-          format: int64
-          description: |-
-            Margin after the train passage in milliseconds
-
-            Enforces that the path used by the train should be free and
-            available at least that many milliseconds after its passage.
-          minimum: 0
-        time_gap_before:
-          type: integer
-          format: int64
-          description: |-
-            Margin before the train passage in seconds
-
-            Enforces that the path used by the train should be free and
-            available at least that many milliseconds before its passage.
-          minimum: 0
-        total_length:
-          type: number
-          format: double
-          description: Total length of the consist in meters
-          nullable: true
-        total_mass:
-          type: number
-          format: double
-          description: Total mass of the consist in kg
-          nullable: true
-        towed_rolling_stock_id:
-          type: integer
-          format: int64
-          nullable: true
-        work_schedule_group_id:
-          type: integer
-          format: int64
-          nullable: true
     Scenario:
       type: object
       required:
--- a/editoast/src/core/simulation.rs
+++ b/editoast/src/core/simulation.rs
@@ -61,10 +61,10 @@ pub struct PhysicsConsist {
     /// Mapping of power restriction code to power class
     #[serde(default)]
     pub power_restrictions: BTreeMap<String, String>,
-    /// The time the train takes before actually using electrical power (in miliseconds).
+    /// The time the train takes before actually using electrical power (in milliseconds).
     /// Is null if the train is not electric or the value not specified.
     pub electrical_power_startup_time: Option<u64>,
-    /// The time it takes to raise this train's pantograph in miliseconds.
+    /// The time it takes to raise this train's pantograph in milliseconds.
     /// Is null if the train is not electric or the value not specified.
     pub raise_pantograph_time: Option<u64>,
 }
@@ -473,6 +473,21 @@ impl AsCoreRequest<Json<SimulationResponse>> for SimulationRequest {
 
     fn infra_id(&self) -> Option<i64> {
         Some(self.infra)
+    }
+}
+
+impl SimulationResponse {
+    pub fn simulation_run_time(&self) -> Option<u64> {
+        if let SimulationResponse::Success { provisional, .. } = self {
+            Some(
+                *provisional
+                    .times
+                    .last()
+                    .expect("core error: empty simulation result"),
+            )
+        } else {
+            None
+        }
     }
 }
 
--- a/editoast/src/core/stdcm.rs
+++ b/editoast/src/core/stdcm.rs
@@ -11,14 +11,13 @@ use serde::Deserialize;
 use serde::Serialize;
 use utoipa::ToSchema;
 
-use super::conflict_detection::Conflict;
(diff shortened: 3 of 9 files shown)


front/src/applications/stdcm/hooks/useStdcm.ts
```diff
@@ -105,8 +105,8 @@ const useStdcm = ({
           setStdcmTrainResult(stdcmTrain);
           dispatch(updateSelectedTrainId(STDCM_TRAIN_ID));
         } else {
-          // When the back-end send back result with status 'path_not_found' we consider that the result is empty
-          setIsStdcmResultsEmpty(response.status === 'path_not_found');
+          // When the back-end send back result with status 'conflicts' we consider that the result is empty
+          setIsStdcmResultsEmpty(response.status === 'conflicts');
```
SharglutDev: This might conflict with @achrafmohye PR : https://github.com/OpenRailAssociation/osrd/pull/9419

hamz2a: Ok, I’ll wait for this PR to be merged before.

hamz2a: I rebased, so there's no more conflict.

eckter:
Before checking the code itself: do we want to merge it during the feature freeze? Are there bug fixes in there, or is it just moving code around?

hamz2a:
> Before checking the code itself: do we want to merge it during the feature freeze? Are there bug fixes in there, or is it just moving code around?

This is purely a refactoring with no functional changes.
Are there any risks you see that would block merging it during the freeze?

editoast/src/core/stdcm.rs
```diff
@@ -115,28 +114,24 @@ pub struct UndirectedTrackRange {
...
 }
 
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, ToSchema)]
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 #[serde(tag = "status", rename_all = "snake_case")]
 // We accepted the difference of memory size taken by variants
 // Since there is only on success and others are error cases
 #[allow(clippy::large_enum_variant)]
-pub enum STDCMResponse {
+pub enum STDCMCoreResponse {
```
woshilapin: Follow up of https://github.com/OpenRailAssociation/osrd/pull/9434#discussion_r1816331474. I still believe that both should be named `STDCMResponse`, one will be named `core::STDCMResponse` and the other one will be `views::timetable::stdcm::STDCMResponse`.

woshilapin (approved):
Thanks for the addressed comments from #9434.

eckter:
> This is purely a refactoring with no functional changes.
> Are there any risks you see that would block merging it during the freeze?

The same as any kind of change, bugs. I believe refactorings are even more "unstable" than new features as they affect what's already there and not just new things.

This specific part of the code that translates front inputs into core payloads is especially sensitive. It's poorly tested and it's difficult to track problems there, and we've had *issues* recently. At a time where we're working on stabilizing stdcm, it would be painful to introduce instability.

Let's talk about this on matrix, it's not about this PR specifically.

editoast/openapi.yaml
```diff
@@ -2628,14 +2628,6 @@ paths:
...
                       enum:
                       - success
-                - type: object
-                  required:
-                  - status
-                  properties:
-                    status:
-                      type: string
-                      enum:
-                      - path_not_found
```
leovalais, line 2638: Since this is a refactoring I'd argue that the public API shouldn't change. Is there a good reason this variant was removed?

editoast/src/core/simulation.rs
```diff
@@ -476,6 +476,17 @@ impl AsCoreRequest<Json<SimulationResponse>> for SimulationRequest {
     }
 }
 
+impl SimulationResponse {
+    pub fn simulation_run_time(self) -> Result<u64, Box<Self>> {
```
leovalais: Why the `Box<Self>` instead of just `Self`?

editoast/src/core/simulation.rs
```diff
@@ -476,6 +476,17 @@ impl AsCoreRequest<Json<SimulationResponse>> for SimulationRequest {
...
+        match self {
+            SimulationResponse::Success { provisional, .. } => {
+                Ok(*provisional.times.last().expect("empty simulation result"))
```
leovalais: Are we 100% sure that Core doesn't send an empty list? If so I'd suggest making the message a bit more explicit

```suggestion
                Ok(*provisional.times.last().expect("core error: empty simulation result"))
```

If not, probably best to return an `Option<u64>` and let the caller deal with the situation.

editoast/src/core/stdcm.rs
leovalais: String agree. I'd also suggest to rename this `StdcmResponse` while doing so to respect the case (I know it's an acronym, but look at `Cow<>` for instance).

editoast/src/core/stdcm.rs
```diff
@@ -115,28 +114,24 @@ pub struct UndirectedTrackRange {
...
     Success {
         simulation: SimulationResponse,
         path: PathfindingResultSuccess,
         departure_time: DateTime<Utc>,
     },
     PathNotFound,
-    Conflicts {
-        pathfinding_result: PathfindingResult,
-        conflicts: Vec<Conflict>,
-    },
```
leovalais, line 133: Why removing this?

editoast/src/models/work_schedules.rs
```diff
@@ -38,3 +45,45 @@ pub struct WorkSchedule {
...
+        latest_simulation_end: DateTime<Utc>,
+    ) -> Option<(i64, crate::core::stdcm::WorkSchedule)> {
+        let search_window_duration = (latest_simulation_end - start_time).num_milliseconds() as u64;
+
+        let ws = self.map_to_core_work_schedule(start_time);
+        if ws.end_time > 0 && ws.start_time < search_window_duration {
+            return Some((self.id, ws));
+        }
+        None
+    }
```
leovalais: I'd suggest merging these two functions into a single `fn as_core_work_schedule(&self, start_time, latest_simulation_end) -> Option<core::stdcm::WorkSchedule>`. A function which cannot be named better that `do_this_and_do_that` is often an indicator of something sketchy.

Since you're not taking ownership of the `WorkSchedule`, the caller can still access its `id` afterwards.

Wdyt?

editoast/src/models/work_schedules.rs
```diff
@@ -38,3 +45,45 @@ pub struct WorkSchedule {
...
+}
+
+fn elapsed_since_time_ms(time: &NaiveDateTime, start_time: &DateTime<Utc>) -> u64 {
```
leovalais: big nit, feel free to ignore:

```suggestion
fn elapsed_time_since_ms(time: &NaiveDateTime, since: &DateTime<Utc>) -> u64 {
```

Also can't we have `time: &Datetime<Utc>` for consistency?

editoast/src/views/timetable.rs
```diff
@@ -1,4 +1,6 @@
+pub mod path_not_found_handler;
 pub mod stdcm;
+pub mod stdcm_request_payload;
```
leovalais: Wdyt about moving all that into the `stcdm` module?

editoast/src/views/timetable/path_not_found_handler.rs
leovalais, line 1: I'm a bit conflicted about the name of the struct: `PathNotFoundHandler` strikes me like a handler for an `Err` case, where it's actually what we do when the simulation "successfully failed". I could be wired wrong though x)

Wdty about?

* Renaming `PathNotFoundHandler` into `SimulationFailureHandler`
* Renaming the module `failure_handler.rs`

@woshilapin

editoast/src/views/timetable/path_not_found_handler.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub infra_version: String,
+    pub train_schedules: Vec<TrainSchedule>,
+    pub simulations: Vec<SimulationResponse>,
+    pub work_schedules: Vec<WorkSchedule>,
+    pub virtual_train_schedule: TrainSchedule,
+    pub virtual_train_sim_result: SimulationResponse,
+    pub virtual_train_pathfinding_result: PathfindingResult,
+    pub earliest_departure_time: DateTime<Utc>,
+    pub latest_simulation_end: DateTime<Utc>,
+}
```
leovalais: This structure is important (it drives this whole PR after all). It should be documented properly:

1. In which case it's useful
2. What it does
3. What a "virtual train" is

Also please restrict the visibility to the most restrictive scope `pub(super)` or something else.

editoast/src/views/timetable/path_not_found_handler.rs
```diff
@@ -0,0 +1,112 @@
...
+            })
+            .collect();
+
+        Some(WorkSchedulesRequest {
+            start_time: self.earliest_departure_time,
+            work_schedule_requirements,
+        })
+    }
+
+    pub async fn handle(self) -> Result<STDCMResponse> {
```
leovalais:
```suggestion
    pub(super?) async fn compute_conflicts(self) -> Result<STDCMResponse> {
```

editoast/src/views/timetable/path_not_found_handler.rs
```diff
@@ -0,0 +1,112 @@
...
+            vec![self.virtual_train_schedule.clone()],
+        ]
+        .concat();
+
+        // Combine the original simulations with the virtual train's simulation results.
+        let simulations = [
+            self.simulations.clone(),
+            vec![self.virtual_train_sim_result.clone()],
+        ]
+        .concat();
```
leovalais:
```suggestion
        let train_schedules = [
            self.train_schedules,
            vec![self.virtual_train_schedule],
        ]
        .concat();

        // Combine the original simulations with the virtual train's simulation results.
        let simulations = [
            self.simulations,
            vec![self.virtual_train_sim_result],
        ]
        .concat();
```

leovalais: Also `self.train_schedules.push(...)`?

editoast/src/views/timetable/path_not_found_handler.rs
```diff
@@ -0,0 +1,112 @@
...
+            train_schedules,
+            simulations,
+            self.earliest_departure_time,
+            self.latest_simulation_end,
+        );
+
+        // Prepare the conflict detection request.
+        let conflict_detection_request = ConflictDetectionRequest {
+            infra: self.infra_id,
+            expected_version: self.infra_version.clone(),
```
leovalais:
```suggestion
            expected_version: self.infra_version,
```

editoast/src/views/timetable/path_not_found_handler.rs
leovalais: Also what about renaming changing "handler" to "fallback" as the term handler also has a meaning for `axum`, and we're in `views`?

editoast/src/views/timetable/stdcm.rs
```diff
@@ -64,15 +53,29 @@ crate::routes! {
...
-editoast_common::schemas! {
-    STDCMRequestPayload,
-    PathfindingItem,
-    StepTimingData,
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, ToSchema)]
+#[serde(tag = "status", rename_all = "snake_case")]
+// We accepted the difference of memory size taken by variants
+// Since there is only on success and others are error cases
+#[allow(clippy::large_enum_variant)]
+pub enum STDCMResponse {
```
leovalais: `pub(in `

editoast/src/views/timetable/stdcm.rs
```diff
@@ -220,107 +157,56 @@ async fn stdcm(
...
-        &stdcm_request,
-        &infra,
-        &rolling_stock,
-        timetable_id,
-    )
-    .await?;
-    let simulation_run_time = match simulation_run_time {
-        SimulationTimeResult::SimulationTime { value } => value,
-        SimulationTimeResult::Error { error } => {
+    let simulation_run_time = match virtual_train_sim_result.clone().simulation_run_time() {
```
```suggestion
    let simulation_run_time = match virtual_train_sim_result.simulation_run_time() {
```

The syntax `let Ok(..) = .. else ` covers this precise pattern ;)

editoast/src/core/simulation.rs
leovalais: Also
```suggestion
    pub fn simulation_run_time(&self) -> Result<u64, Box<Self>> {
```

editoast/src/views/timetable/stdcm.rs
```diff
@@ -539,86 +368,23 @@ fn is_resource_in_range(
...
-        .map(|data| data.arrival_time_tolerance_before + data.arrival_time_tolerance_after)
-        .next()
-        .unwrap_or(0)
-}
-
 /// Returns a `Result` containing:
-/// * `SimulationTimeResult` - The result of the simulation time calculation.
 /// * `TrainSchedule` - The generated train schedule based on the provided data.
 /// * `SimulationResponse` - Simulation response.
 /// * `PathfindingResult` - Pathfinding result.
```
leovalais: I'm not sure this documentation is useful as it doesn't bring anything we can't read in the return type.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -539,86 +368,23 @@ fn is_resource_in_range(
...
     infra: &Infra,
     rolling_stock: &RollingStockModel,
     timetable_id: i64,
-) -> Result<(
-    SimulationTimeResult,
-    TrainSchedule,
-    SimulationResponse,
-    PathfindingResult,
-)> {
+) -> Result<(TrainSchedule, SimulationResponse, PathfindingResult)> {
```
leovalais: Shouldn't we have a `VirtualTrainRun` structure aggregating the information of the virtual train? That would also reduce the number of fields of `PathNotFoundHandler`.

editoast/src/views/timetable/stdcm_request_payload.rs
leovalais, line 1: Likewise about visibility

editoast/src/views/timetable/stdcm_request_payload.rs
```diff
@@ -0,0 +1,285 @@
...
+    pub arrival_time: DateTime<Utc>,
+    /// The train may arrive up to this duration before the expected arrival time
+    pub arrival_time_tolerance_before: u64,
+    /// The train may arrive up to this duration after the expected arrival time
+    pub arrival_time_tolerance_after: u64,
+}
+
+/// An STDCM request
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, ToSchema)]
+pub struct STDCMRequestPayload {
```
leovalais: A bit more bike shed on the naming: we're in `views`, I'd say we don't need the `Payload` suffix to understand this struct refers to the content of the http request of the STDCM endpoint. 

Additionally, I'd rather have this module to be included in the `stdcm` module rather than `timetable` directly (see previous comment). In this case I'd argue that `Request` is a good enough name: both `stdcm/request.rs` and `stdcm::Request` are explanatory enough.

Just like for the `core` definitions and the `Core` prefix, `views::timetable::stcdm` is already providing a lot of context.

Wdyt?

editoast/src/views/timetable/stdcm_request_payload.rs
```diff
@@ -0,0 +1,285 @@
...
+            .iter()
+            .map(|step: &PathfindingItem| step.duration.unwrap_or_default())
+            .sum()
+    }
+
+    // Returns the maximum departure delay for the train.
+    pub fn get_maximum_departure_delay(&self, simulation_run_time: u64) -> u64 {
+        self.maximum_departure_delay
+            .unwrap_or(simulation_run_time + self.get_earliest_step_tolerance_window())
+    }
```
leovalais:
```suggestion
    }

```

leovalais (changes requested):
Thanks for all this work, it looks much more organized now! I left a bunch of comments which are mostly about:

* module organization
* visibility being more permissive than required (I didn't add a comment in all places)
* naming

It would be nice to have some module-level documentation in `stdcm.rs` describing the whole workflow still, even if it's much clearer now thanks to structure.

I'm not as worried about bugs as @eckter is: we already released a tested version and we're shooting demo videos very soon. I'd say that if a bug arises we're better off fixing it in the refactored code directly since this won't impact the PIP. I'm fine with waiting until the end of the week before merging in case a critical bug happens preventing us from shooting the videos, but there's no point in waiting after that point imo. Wdyt?

SharglutDev (approved):
Lgtm for front part (not tested, we were'nt handling this status anymore)

editoast/openapi.yaml
hamz2a: This variant is in Core's response, but I created a new struct for `StdcmResponse` in views, where this variant isn’t included, so we’re replacing it with `conflicts`.

editoast/src/core/simulation.rs
hamz2a: I updated the error message.
We need the result to forward it within this part of the code.
```rust
    let simulation_run_time = match virtual_train_sim_result.clone().simulation_run_time() {
        Ok(value) => value,
        Err(error) => {
            return Ok(Json(StdcmResponse::PreprocessingSimulationError {
                error: *error,
            }))
        }
    };
```

editoast/src/core/stdcm.rs
hamz2a: Core isn't returning conflicts for the STDCM request.

editoast/src/models/work_schedules.rs
hamz2a: Done.

editoast/src/models/work_schedules.rs
hamz2a: Done.

editoast/src/core/simulation.rs
hamz2a: I had this warning:
```bash
pub fn simulation_run_time(self) -> Result<u64, Self> {
    |                                           ^^^^^^^^^^^^^^^^^
    |
    = help: try reducing the size of `core::simulation::SimulationResponse`, for example by boxing large elements or replacing it with `Box<core::simulation::SimulationResponse>`
```

leovalais: Oh alright then. Maybe add a comment explaining the presence of the Box since this is not straightforward (to me at least).

editoast/src/core/simulation.rs
leovalais: > If not, probably best to return an Option<u64> and let the caller deal with the situation.

I meant to return an `Ok(Some(...))` instead, that was unclear, sorry

editoast/src/views/timetable.rs
hamz2a: Done.

editoast/src/views/timetable/path_not_found_handler.rs
hamz2a: I prefer to keep `handler`; I find this name clearer. Wdyt ?
Otherwise, I've changed the module and struct names.

editoast/src/views/timetable/path_not_found_handler.rs
hamz2a: Done.

editoast/src/views/timetable/path_not_found_handler.rs
hamz2a: To do `self.train_schedules.push(...)`, `self` needs to be mutable.
Therefore, I decided to clone certain parts of `self` instead of making the entire self mutable.

editoast/src/views/timetable/path_not_found_handler.rs
hamz2a: i got this error:
```bash
   |
93 |             expected_version: self.infra_version,
   |                               ------------------ value partially moved here
94 |             trains_requirements,
95 |             work_schedules: self.get_work_schedules_request(),
   |                             ^^^^ value borrowed here after partial move
   |
   = note: partial move occurs because `self.infra_version` has type `std::string::String`, which does not implement the `Copy` trait
 ```

File: editoast/src/views/timetable/stdcm.rs (hunk 15 above)

File: editoast/src/views/timetable/stdcm_request_payload.rs (hunk 18 above)
Comment (hamz2a) [reply editoast/src/views/timetable/stdcm_request_payload.rs]: Thanks for the great feedback! I agree with your points. I'll go ahead and remove the `Payload` suffix and move the module to `stdcm`. `Request` sounds perfect, and the context is clear.

Comment (hamz2a) [reply editoast/src/views/timetable/stdcm_request_payload.rs:71]: Thanks for the great feedback! I agree with your points. I'll go ahead and remove the `Payload` suffix and move the module to `stdcm`. `Request` sounds perfect, and the context is clear.

File: editoast/src/views/timetable/path_not_found_handler.rs (hunk 10 above)
Comment (hamz2a) [reply editoast/src/views/timetable/path_not_found_handler.rs]: Done.

File: editoast/src/views/timetable/stdcm.rs (hunk 17 above)

File: editoast/src/views/timetable/path_not_found_handler.rs
Comment (woshilapin) [reply editoast/src/views/timetable/path_not_found_handler.rs:1]: Agreed about it's more about simulation... but it's for the special case of path not found. Do we think we might have another handler for another simulation error? I'd say maybe, but if you think we can stick all in this one, I'm fine with `SimulationFailureHandler`.

As for "handler" vs "fallback", I'll stick with @hamz2a on this one for 2 reasons:
- I indeed find "handler" clearer
- `axum` does have the notion of "handler" but it's not used anywhere in the code (almost, try a `grep -i handler editoast/` to see what I mean). And indeed, in `axum`, `Handler` is a trait (but most likely, we never need it, not even need to import it), `handler` is a module (might use it in a `use`), and there is `#[debug_handler]` (which should never really stick in production code).

Comment (leovalais) [reply editoast/src/views/timetable/path_not_found_handler.rs:1]: Ok, let's keep `-Handler` then.

File: editoast/src/views/timetable/path_not_found_handler.rs (hunk 12 above)
Comment (leovalais) [reply editoast/src/views/timetable/path_not_found_handler.rs]: The function takes ownership of `self` and never returns it. I'd say it's fine to make it `mut` since it's already clear that the `SimulationFailureHandler` won't live after the call. Wdyt?

File: editoast/src/views/timetable/path_not_found_handler.rs (hunk 13 above)
Comment (leovalais) [reply editoast/src/views/timetable/path_not_found_handler.rs]: `self.get_work_schedules_request` doesn't move `self`, I guess you can swap the two fields in the struct initializer and avoid the clone

Comment (hamz2a) [reply editoast/src/views/timetable/path_not_found_handler.rs]: Thanks for the suggestion! Swapping the fields did the trick, and it worked perfectly without needing the clone. Great idea, and I appreciate the insight!

File: editoast/src/views/timetable/path_not_found_handler.rs (hunk 12 above)
Comment (hamz2a) [reply editoast/src/views/timetable/path_not_found_handler.rs]: I tested using `mut`, but it caused excessive cloning, so I tried destructuring instead. Let me know what you think.

File: editoast/src/views/timetable/stdcm.rs [hunk 20]
```diff
@@ -60,19 +51,37 @@ use crate::Retrieve;
...
+    },
+    PreprocessingSimulationError {
+        error: SimulationResponse,
+    },
 }
 
 #[derive(Debug, Error, EditoastError, Serialize)]
 #[editoast_error(base_id = "stdcm_v2")]
-enum STDCMError {
+pub enum STDCMError {
```
```suggestion
enum StdcmError {
```

editoast/src/core/simulation.rs
leovalais: Up https://github.com/OpenRailAssociation/osrd/pull/9509/files#r1829667767

editoast/src/views/timetable/stdcm/failure_handler.rs
```diff
@@ -0,0 +1,125 @@
...
+
+use super::build_train_requirements;
+use super::VirtualTrainRun;
+
+/// `SimulationFailureHandler` is used when a simulation failure occurs,
+/// particularly when a train's path cannot be found. It helps detect
+/// conflicts between the `virtual_train` and existing train schedules and simulations.
+/// `virtual_train` is a simulated train created to detect conflicts
+/// when a real train’s path cannot be found during the simulation.
+pub struct SimulationFailureHandler {
```
leovalais:
```suggestion
pub(super) struct SimulationFailureHandler {
```

editoast/src/views/timetable/stdcm/failure_handler.rs
```diff
@@ -0,0 +1,125 @@
...
+            infra_version,
+            earliest_departure_time,
+            latest_simulation_end,
+            ..
+        } = self;
+        let VirtualTrainRun {
+            train_schedule,
+            simulation,
+            pathfinding,
+        } = virtual_train_run;
```
leovalais:
```suggestion
        let Self {
            mut train_schedules,
            mut simulations,
            virtual_train_run:
                VirtualTrainRun {
                    train_schedule,
                    simulation,
                    pathfinding,
                },
            infra_id,
            infra_version,
            earliest_departure_time,
            latest_simulation_end,
            ..
        } = self;

```

You can destructure several levels at once.

editoast/src/views/timetable/path_not_found_handler.rs
leovalais: I'm not sure why this would cause any legitimate `clone()`, can you provide an example?

Destructuring is fine by me, but the workflow became weird with the `let work_schedules = self.get_work_schedules_request();` that has to occur before that.

I'd suggest either:

1. Try to use `mut self` and keep the call to `self.get_work_schedules_request` where it was before
2. Make `get_work_schedules_request` a function that takes all the `self` data as parameters and call it using the variables created by the destructuring.

editoast/src/views/timetable/path_not_found_handler.rs
leovalais: Cf. https://github.com/OpenRailAssociation/osrd/pull/9509#discussion_r1834761313

editoast/src/views/timetable/stdcm/failure_handler.rs
```diff
@@ -0,0 +1,125 @@
...
+        let work_schedule_requirements = self
+            .work_schedules
+            .iter()
+            .filter_map(|ws| {
+                if let Some(core_ws) = ws
+                    .as_core_work_schedule(self.earliest_departure_time, self.latest_simulation_end)
+                {
+                    return Some((ws.id, core_ws));
+                };
+                None
```
leovalais: `Option.map`?

editoast/src/views/timetable/stdcm/request.rs
leovalais, line 1: Up about visibility.

editoast/src/views/timetable/stdcm.rs
leovalais, line 1: Likewise about visibility in this file.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -329,142 +217,82 @@ async fn stdcm(
...
-        path_items,
+        path_items: stdcm_request.get_stdcm_path_items(conn, infra_id).await?,
         start_time: earliest_departure_time,
         trains_requirements: trains_requirements.clone(),
-        maximum_departure_delay,
-        maximum_run_time,
-        speed_limit_tag: stdcm_request.speed_limit_tags,
+        maximum_departure_delay: stdcm_request.get_maximum_departure_delay(simulation_run_time),
+        maximum_run_time: stdcm_request.get_maximum_run_time(simulation_run_time),
+        speed_limit_tag: stdcm_request.speed_limit_tags.clone(),
```
leovalais: Moving `train_requirements` and `speed_limit_tags` further down the field list may make the `clone()` unnecessary.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -539,157 +367,69 @@ fn is_resource_in_range(
...
-    };
-    Ok((
-        simulation_run_time,
-        train_schedule,
-        sim_result,
-        pathfinding_result,
-    ))
-}
+impl VirtualTrainRun {
+    async fn simulate_virtual_train_run(
```
```suggestion
    async fn simulate(
```

leovalais (commented):
Thanks for the changes!

editoast/src/views/timetable/stdcm.rs

editoast/src/views/timetable/stdcm/failure_handler.rs
hamz2a: Done.

editoast/src/views/timetable/stdcm/failure_handler.rs
hamz2a: Done.

editoast/src/views/timetable/stdcm.rs

editoast/src/views/timetable/stdcm/request.rs
hamz2a: Done.

editoast/src/views/timetable/stdcm.rs
hamz2a: Done.

editoast/src/views/timetable/stdcm/request.rs
hamz2a: Done.

editoast/src/views/timetable/stdcm/request.rs
```diff
@@ -0,0 +1,288 @@
...
+        self.steps
+            .iter()
+            .map(|step: &PathfindingItem| step.duration.unwrap_or_default())
+            .sum()
+    }
+
+    // Returns the maximum departure delay for the train.
+    pub(super) fn get_maximum_departure_delay(&self, simulation_run_time: u64) -> u64 {
+        self.maximum_departure_delay
+            .unwrap_or(simulation_run_time + self.get_earliest_step_tolerance_window())
```
eckter, line 155: That feels wrong, I'd expect this instead:
```suggestion
            .unwrap_or(self.get_earliest_step_tolerance_window())
```

---

@Erashin this function has just been moved here and you wrote the previous version, do you remember if there's a reason to add the simulation run time there?

editoast/src/views/timetable/stdcm/request.rs
```diff
@@ -0,0 +1,288 @@
...
+    }
+    // Maximum duration between train departure and arrival, including all stops
+    pub(super) fn get_maximum_run_time(&self, simulation_run_time: u64) -> u64 {
+        self.maximum_run_time
+            .unwrap_or(2 * simulation_run_time + self.get_total_stop_time())
```
eckter, line 161: Something is likely wrong there as well... We're supposed to return the simulation duration where the time spent between stops has been doubled. Here I believe stop duration is already included in `simulation_run_time`.

If a train leaves at 10:00, stops from 13:00 to 13:05 then arrives at 14:05. We spend 2h moving and 5 minutes stopped: we should return `2 * 2h + 5m = 4:05`. But in this version we may be returning `2 * 2h05 + 5m = 4:15`.

I realize that this isn't part of the PR and the code has just been moved, I'll test this on the side.

eckter (approved):
Much clearer, thanks :+1: 

I've noticed a couple weird things (that were already there in the previous version), I've commented them and I'll check further

editoast/src/views/timetable/stdcm/failure_handler.rs
hamz2a: Done.

leovalais (approved):
LGTM. Thanks for the changes!

editoast/src/views/timetable/stdcm/request.rs
Erashin: In the case where the departure is provided, that is indeed the case. However, if it's the end that is provided, it will always fail if the simulation run time is a lot higher than the tolerance window.. the best would be to differentiate both cases :
- if the departure is provided, max departure delay should be the tolerance window
- if it's the end, it should be as specified right now.

Issue #9422, opened 2024-10-22
front: rework the play bar in scenario page


axrolld:
# Description

Rework the display of the play bar, without adding any new feature.

The UI should follow [this mockup](https://www.sketch.com/s/2fb064cd-c483-46ac-bd30-72705fd63e75/p/9C99FBA0-0439-435A-98E4-5439F7838F33/canvas#Inspect).

![Image](https://github.com/user-attachments/assets/8b6c978d-7c7a-4fb9-8b06-f3bd714b8beb)

# Goal

This feature matters because it will end the harmonization for the scenario page.

# Acceptance criteria

- [ ] The play bar is now as the mockup specifies
- [ ] The time component should be editable by hand with clic + keyboard
- [ ] Two buttons should allow the user to play forward or backward (forward by default)
- [ ] The speed buttons should multiply the speed of the play (x1 as default)

# Optionnal

- [ ] A pin button is located to the right. When activated, the play bar remains sticky on top of the screen. If not, the bar remains on top of the GET.

theocrsb:
For now, we've hidden the bar with a `display none` in this [PR](https://github.com/OpenRailAssociation/osrd/pull/9512) and updated e2e tests in `op-timetable-page-model.ts`

Tguisnet:
@Tguisnet Discuss the priority of this with Thibaut

Tguisnet:
Not very importante for the moment (answer from thibaut)

PR #9512, opened 2024-10-29
front: hide simulation sticky bar

### LINKED ISSUE #9511: Hide the `simulation-sticky-bar` until it's redesigned
### Description and goal

For the moment, the bar has no real use and doesn't match the design. 
So we'll have to hide it for the time being.

theocrsb:
closes #9511


RomainValls (approved):
lgtm and tested

SharglutDev (approved):
Lgtm but plz add to the issue for the new chart controller that this should be removed

theocrsb:
> Lgtm but plz add to the issue for the new chart controller that this should be removed

https://github.com/OpenRailAssociation/osrd/issues/9422#issuecomment-2446096192

PR #9513, opened 2024-10-29
front: fix space time chart when first or last waypoint are hidden

SharglutDev:
Filter data from projected path (path and occupancy blocks) and conflicts when the first or last waypoint is hidden. Allow the space time chart to look as if it started/ended with the first/last displayed waypoint.

part of #8628 (see last comment of the issue)

before : 
Capture d’écran 2024-10-29 à 16 46 57

after : 
Capture d’écran 2024-10-29 à 16 46 24

To test : 
- Add at least 2 trains with only 1min difference for departure time
- With the waypoints panel, hide the first x waypoints and/or the last x waypoints

### DIFF
--- a/front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
+++ b/front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
@@ -1,4 +1,4 @@
-import { useRef, useState } from 'react';
+import { useMemo, useRef, useState } from 'react';
 
 import { KebabHorizontal } from '@osrd-project/ui-icons';
 import { Manchette } from '@osrd-project/ui-manchette';
@@ -11,12 +11,18 @@ import {
   OccupancyBlockLayer,
 } from '@osrd-project/ui-spacetimechart';
 import type { Conflict } from '@osrd-project/ui-spacetimechart';
+import { compact } from 'lodash';
 
 import type { OperationalPoint, TrainSpaceTimeData } from 'applications/operationalStudies/types';
 import upward from 'assets/pictures/workSchedules/ScheduledMaintenanceUp.svg';
 import type { PostWorkSchedulesProjectPathApiResponse } from 'common/api/osrdEditoastApi';
+import cutSpaceTimeRect from 'modules/simulationResult/components/SpaceTimeChart/helpers/utils';
 import { ASPECT_LABELS_COLORS } from 'modules/simulationResult/consts';
-import type { AspectLabel, WaypointsPanelData } from 'modules/simulationResult/types';
+import type {
+  AspectLabel,
+  LayerRangeData,
+  WaypointsPanelData,
+} from 'modules/simulationResult/types';
 
 import SettingsPanel from './SettingsPanel';
 import ManchetteMenuButton from '../SpaceTimeChart/ManchetteMenuButton';
@@ -45,9 +51,91 @@ const ManchetteWithSpaceTimeChartWrapper = ({
 
   const [waypointsPanelIsOpen, setWaypointsPanelIsOpen] = useState(false);
 
+  // Cut the space time chart curves if the first or last waypoints are hidden
+  const spaceTimeChartLayersData = useMemo(() => {
+    let filteredProjectPathTrainResult = projectPathTrainResult;
+    let filteredConflicts = conflicts;
+
+    if (!waypointsPanelData || waypointsPanelData.filteredWaypoints.length < 2)
+      return { filteredProjectPathTrainResult, filteredConflicts };
+
+    const { filteredWaypoints } = waypointsPanelData;
+    const firstPosition = filteredWaypoints.at(0)!.position;
+    const lastPosition = filteredWaypoints.at(-1)!.position;
+
+    if (firstPosition !== 0 || lastPosition !== operationalPoints.at(-1)!.position) {
+      filteredProjectPathTrainResult = projectPathTrainResult.map((train) => ({
+        ...train,
+        space_time_curves: train.space_time_curves.map(({ positions, times }) => {
+          const cutPositions: number[] = [];
+          const cutTimes: number[] = [];
+
+          for (let i = 1; i < positions.length; i += 1) {
+            const currentRange: LayerRangeData = {
+              spaceStart: positions[i - 1],
+              spaceEnd: positions[i],
+              timeStart: times[i - 1],
+              timeEnd: times[i],
+            };
+
+            const interpolatedRange = cutSpaceTimeRect(currentRange, firstPosition, lastPosition);
+
+            // TODO : remove reformatting the datas when https://github.com/OpenRailAssociation/osrd-ui/issues/694 is merged
+            if (!interpolatedRange) continue;
+
+            if (i === 1 || cutPositions.length === 0) {
+              cutPositions.push(interpolatedRange.spaceStart);
+              cutTimes.push(interpolatedRange.timeStart);
+            }
+            cutPositions.push(interpolatedRange.spaceEnd);
+            cutTimes.push(interpolatedRange.timeEnd);
+          }
+
+          return {
+            positions: cutPositions,
+            times: cutTimes,
+          };
+        }),
+        signal_updates: compact(
+          train.signal_updates.map((signal) => {
+            const updatedSignalRange = cutSpaceTimeRect(
+              {
+                spaceStart: signal.position_start,
+                spaceEnd: signal.position_end,
+                timeStart: signal.time_start,
+                timeEnd: signal.time_end,
+              },
+              firstPosition,
+              lastPosition
+            );
+
+            if (!updatedSignalRange) return null;
+
+            // TODO : remove reformatting the datas when https://github.com/OpenRailAssociation/osrd-ui/issues/694 is merged
+            return {
+              ...signal,
+              position_start: updatedSignalRange.spaceStart,
+              position_end: updatedSignalRange.spaceEnd,
+              time_start: updatedSignalRange.timeStart,
+              time_end: updatedSignalRange.timeEnd,
+            };
+          })
+        ),
+      }));
+
+      filteredConflicts = compact(
+        conflicts.map((conflict) => cutSpaceTimeRect(conflict, firstPosition, lastPosition))
+      );
+
+      return { filteredProjectPathTrainResult, filteredConflicts };
+    }
+
+    return { filteredProjectPathTrainResult, filteredConflicts };
+  }, [waypointsPanelData?.filteredWaypoints, projectPathTrainResult, conflicts]);
+
   const { manchetteProps, spaceTimeChartProps, handleScroll } = useManchettesWithSpaceTimeChart(
     waypointsPanelData?.filteredWaypoints ?? operationalPoints,
-    projectPathTrainResult,
+    spaceTimeChartLayersData.filteredProjectPathTrainResult,
     manchetteWithSpaceTimeChartRef,
     selectedTrainScheduleId
   );
@@ -58,17 +146,19 @@ const ManchetteWithSpaceTimeChartWrapper = ({
     showSignalsStates: false,
   });
 
-  const occupancyBlocks = projectPathTrainResult.flatMap((train) => {
-    const departureTime = new Date(train.departure_time).getTime();
+  const occupancyBlocks = spaceTimeChartLayersData.filteredProjectPathTrainResult.flatMap(
+    (train) => {
+      const departureTime = new Date(train.departure_time).getTime();
 
-    return train.signal_updates.map((block) => ({
-      timeStart: departureTime + block.time_start,
-      timeEnd: departureTime + block.time_end,
-      spaceStart: block.position_start,
-      spaceEnd: block.position_end,
-      color: ASPECT_LABELS_COLORS[block.aspect_label as AspectLabel],
-    }));
-  });
+      return train.signal_updates.map((block) => ({
+        timeStart: departureTime + block.time_start,
+        timeEnd: departureTime + block.time_end,
+        spaceStart: block.position_start,
+        spaceEnd: block.position_end,
+        color: ASPECT_LABELS_COLORS[block.aspect_label as AspectLabel],
+      }));
+    }
+  );
 
   return (
     <div className="manchette-space-time-chart-wrapper">
@@ -134,7 +224,9 @@ const ManchetteWithSpaceTimeChartWrapper = ({
                 imageUrl={upward}
               />
             )}
-            {settings.showConflicts && <ConflictLayer conflicts={conflicts} />}
+            {settings.showConflicts && (
+              <ConflictLayer conflicts={spaceTimeChartLayersData.filteredConflicts} />
+            )}
             {settings.showSignalsStates && (
               <OccupancyBlockLayer occupancyBlocks={occupancyBlocks} />
             )}
--- a/front/src/modules/simulationResult/components/SpaceTimeChart/helpers/__tests__/utils.spec.ts
+++ b/front/src/modules/simulationResult/components/SpaceTimeChart/helpers/__tests__/utils.spec.ts
@@ -1,0 +1,86 @@
+import { describe, it, expect } from 'vitest';
+
+import cutSpaceTimeRect from '../utils';
+
+describe('interpolateRange', () => {
+  it('should return null if the interpolated range ends before the cut space', () => {
+    const range = {
+      spaceStart: 3,
+      spaceEnd: 5,
+      timeStart: 100,
+      timeEnd: 200,
+    };
+    const interpolatedRange = cutSpaceTimeRect(range, 1, 3);
+    expect(interpolatedRange).toBeNull();
+  });
+
+  it('should return null if the interpolated range starts after the cut space', () => {
+    const range = {
+      spaceStart: 3,
+      spaceEnd: 5,
+      timeStart: 100,
+      timeEnd: 200,
+    };
+    const interpolatedRange = cutSpaceTimeRect(range, 5, 7);
+    expect(interpolatedRange).toBeNull();
+  });
+
+  it('should return the same range if its ranges are inside the cut space', () => {
+    const range = {
+      spaceStart: 3,
+      spaceEnd: 5,
+      timeStart: 100,
+      timeEnd: 200,
+    };
+    const interpolatedRange = cutSpaceTimeRect(range, 2, 7);
+    expect(interpolatedRange).toEqual(range);
+  });
+
+  it('should return the interpolated range when the start position is outside the cut space', () => {
+    const range = {
+      spaceStart: 3,
+      spaceEnd: 5,
+      timeStart: 100,
+      timeEnd: 200,
+    };
+    const interpolatedRange = cutSpaceTimeRect(range, 4, 5);
+    expect(interpolatedRange).toEqual({
+      spaceStart: 4,
+      spaceEnd: 5,
+      timeStart: 150,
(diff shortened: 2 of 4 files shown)


front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -45,9 +50,51 @@ const ManchetteWithSpaceTimeChartWrapper = ({
...
 
+  // Filter the space time chart curves if the first or last waypoints are hidden
+  const spaceTimeChartLayersData = useMemo(() => {
+    let filteredProjectPathTrainResult = projectPathTrainResult;
+    let filteredConflicts = conflicts;
+
+    if (!waypointsPanelData) return { filteredProjectPathTrainResult, filteredConflicts };
+
+    const { filteredWaypoints } = waypointsPanelData;
+    const firstPositon = filteredWaypoints.at(0)!.position;
```
emersion: Typo: "position"

front/src/modules/simulationResult/helpers/utils.ts
```diff
@@ -0,0 +1,113 @@
...
+      if (spaceStart < cutPositionStart && spaceEnd > cutPositionStart) {
+        const interpolationFactor = (cutPositionStart - spaceStart) / (spaceEnd - spaceStart);
+        return {
+          spaceStart: cutPositionStart,
+          spaceEnd,
+          timeStart: timeStart + (timeEnd - timeStart) * interpolationFactor,
+          timeEnd,
+        };
+      }
+      if (spaceStart < cutPositionEnd && spaceEnd > cutPositionEnd) {
```
emersion: What if both of these conditions are true? That is, what if a conflict extends past the first visible OP _and_ past the last visible OP?

(These cut-off functions are no fun - I'd really rather not have them at all…)

SharglutDev: I forgot to leave a comment in the code for that but I assumed that even if we leave only a very few waypoints visible, it would be unlikely to have only range of the layer left...

@Tguisnet do you think it could happen often ?

Uriel-Sautron (approved):
Tested, thank you

front/src/modules/simulationResult/helpers/utils.ts
```diff
@@ -0,0 +1,113 @@
...
+  cutPositionStart: number,
+  cutPositionEnd: number
+): ProjectPathTrainResult['space_time_curves'] =>
+  spaceTimeCurves.map((spaceTimeCurve) => {
+    let { positions, times } = { ...spaceTimeCurve };
+
+    // If the position where we want to cut the curve isn't already in the curve positions,
+    // We need to interpolate the corresponding time and it and the position the curve
+
+    if (positions.at(0)! <= cutPositionStart && !positions.includes(cutPositionStart)) {
```
clarani: I think `<` is enough
```suggestion
    if (positions.at(0)! < cutPositionStart && !positions.includes(cutPositionStart)) {
```

front/src/modules/simulationResult/helpers/utils.ts
```diff
@@ -0,0 +1,113 @@
...
+        { positions, speeds: [], times },
+        cutPositionStart,
+        'times'
+      );
+
+      positions = [cutPositionStart, ...positions];
+      times = [interpolatedTime, ...times];
+    }
+
+    if (positions.at(-1)! >= cutPositionEnd && !positions.includes(cutPositionEnd)) {
```
clarani:
```suggestion
    if (positions.at(-1)! > cutPositionEnd && !positions.includes(cutPositionEnd)) {
```

front/src/modules/simulationResult/helpers/utils.ts
```diff
@@ -0,0 +1,113 @@
...
+    const filteredIndex = positions
+      .map((position, index) =>
+        position >= cutPositionStart && position <= cutPositionEnd ? index : -1
+      )
+      .filter((index) => index !== -1);
+
+    return {
+      positions: filteredIndex.map((index) => positions[index]),
+      times: filteredIndex.map((index) => times[index]),
+    };
```
clarani: Do you think it's clearer like this ?
```suggestion
    const result = { positions: [] as number[], times: [] as number[] };
    for (let i = 0; i < positions.length; i++) {
      const position = positions[i];
      if (position >= cutPositionStart && position <= cutPositionEnd) {
        result.positions.push(position);
        result.times.push(times[i]);
      }
    }
    return result;
```

front/src/modules/simulationResult/helpers/utils.ts
```diff
@@ -0,0 +1,113 @@
...
+export const filterSpaceTimeCurvesByPosition = (
+  spaceTimeCurves: ProjectPathTrainResult['space_time_curves'],
+  cutPositionStart: number,
+  cutPositionEnd: number
+): ProjectPathTrainResult['space_time_curves'] =>
+  spaceTimeCurves.map((spaceTimeCurve) => {
+    let { positions, times } = { ...spaceTimeCurve };
+
+    // If the position where we want to cut the curve isn't already in the curve positions,
+    // We need to interpolate the corresponding time and it and the position the curve
```
clarani: What did you want to say here ? :)

front/src/modules/simulationResult/helpers/utils.ts
```diff
@@ -0,0 +1,113 @@
...
+          time_end: signal.time_end - (signal.time_end - signal.time_start) * interpolationFactor,
+        };
+      }
+      return signal;
+    })
+    .filter(
+      (signal) => signal.position_start >= cutPositionStart && signal.position_end <= cutPositionEnd
+    );
+
+export const filterSpaceTimeCurvesByPosition = (
```
clarani:
```suggestion
export const cutSpaceTimeCurvesByPosition = (
```

front/src/modules/simulationResult/helpers/utils.ts
```diff
@@ -0,0 +1,113 @@
...
+      )
+      .filter((index) => index !== -1);
+
+    return {
+      positions: filteredIndex.map((index) => positions[index]),
+      times: filteredIndex.map((index) => times[index]),
+    };
+  });
+
+export const filterConflictsByPosition = (
```
clarani: What do you think of this version ? In this function, we only go through the array once 
```
export const cutConflictsByPosition = (
  conflicts: Conflict[],
  cutPositionStart: number,
  cutPositionEnd: number
) => {
  const results: Conflict[] = [];

  for (const conflict of conflicts) {
    const { spaceStart, spaceEnd, timeStart, timeEnd } = conflict;
    if (spaceStart < cutPositionStart && spaceEnd > cutPositionEnd) {
      continue;
    }
    if (spaceStart < cutPositionStart) {
      const interpolationFactor = (cutPositionStart - spaceStart) / (spaceEnd - spaceStart);
      results.push({
        spaceStart: cutPositionStart,
        spaceEnd,
        timeStart: timeStart + (timeEnd - timeStart) * interpolationFactor,
        timeEnd,
      });
      continue;
    }
    if (spaceEnd > cutPositionEnd) {
      const interpolationFactor = (spaceEnd - cutPositionEnd) / (spaceEnd - spaceStart);
      results.push({
        spaceStart,
        spaceEnd: cutPositionEnd,
        timeStart,
        timeEnd: timeEnd - (timeEnd - timeStart) * interpolationFactor,
      });
      continue;
    }
    results.push(conflict);
  }

  return results;
};
```

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -45,9 +50,51 @@ const ManchetteWithSpaceTimeChartWrapper = ({
...
   const [waypointsPanelIsOpen, setWaypointsPanelIsOpen] = useState(false);
 
+  // Filter the space time chart curves if the first or last waypoints are hidden
+  const spaceTimeChartLayersData = useMemo(() => {
+    let filteredProjectPathTrainResult = projectPathTrainResult;
+    let filteredConflicts = conflicts;
+
+    if (!waypointsPanelData) return { filteredProjectPathTrainResult, filteredConflicts };
+
+    const { filteredWaypoints } = waypointsPanelData;
```
clarani, line 62: Maybe we can check here that filteredWaypoints has at least 2 elements (and remove the check line 65)
```suggestion
    const { filteredWaypoints } = waypointsPanelData;
    if (filteredWaypoints.length < 2) throw new Error('filteredWaypoints has less than 2 points');
```

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -45,9 +50,51 @@ const ManchetteWithSpaceTimeChartWrapper = ({
 
   const [waypointsPanelIsOpen, setWaypointsPanelIsOpen] = useState(false);
 
+  // Filter the space time chart curves if the first or last waypoints are hidden
```
clarani:
```suggestion
  // Cut the space time chart curves if the first or last waypoints are hidden
```

front/src/modules/simulationResult/helpers/utils.ts
SharglutDev: You're right, done.

front/src/modules/simulationResult/helpers/utils.ts
SharglutDev: Done

front/src/modules/simulationResult/helpers/utils.ts
SharglutDev: Only one loop instead of 3, great suggestion thank you :) Done

front/src/modules/simulationResult/helpers/utils.ts
SharglutDev: Oups missed some words : `    // We need to interpolate the corresponding time and add it with the position to the curve
`

front/src/modules/simulationResult/helpers/utils.ts
SharglutDev: Done

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
SharglutDev: We have a small gap when `filteredWaypoints` isn't updated. If we do that, when we enter a scenario, we will have 
Capture d’écran 2024-10-31 à 14 23 36

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx

front/src/modules/simulationResult/helpers/utils.ts
```diff
@@ -0,0 +1,113 @@
+import type { Conflict } from '@osrd-project/ui-spacetimechart';
+
+import type { ProjectPathTrainResult } from 'common/api/osrdEditoastApi';
+import { interpolateValue } from 'modules/trainschedule/components/DriverTrainSchedule/utils';
+
+export const filterSignalUpdatesByPosition = (
```
emersion: Could we maybe define a single interpolation function, and use it everywhere for all curves and layers?

Typing something up real quick:

```ts
type Line = {
	spaceStart: number;
	spaceEnd: number;
	timeStart: number;
	timeEnd: number;
};

const interpolateLine = (line: Line, minSpace: number, maxSpace: number): Line | null => {
	let { timeStart, timeEnd } = line;
	const { spaceStart, spaceEnd } = line;

	if (spaceStart < minSpace || spaceEnd > maxSpace) {
		return null;
	}

	if (spaceStart < minSpace) {
		const interpolationFactor = (minSpace - spaceStart) / (spaceEnd - spaceStart);
		timeStart += (timeEnd - timeStart) * interpolationFactor;
	}

	if (spaceEnd > maxSpace) {
		const interpolationFactor = (spaceEnd - maxSpace) / (spaceEnd - spaceStart);
		timeEnd -= (timeEnd - timeStart) * interpolationFactor;
	}

	return {
		...line,
		timeStart,
		timeEnd,
	};
};
```

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -45,9 +51,94 @@ const ManchetteWithSpaceTimeChartWrapper = ({
 
   const [waypointsPanelIsOpen, setWaypointsPanelIsOpen] = useState(false);
 
+  // Cut the space time chart curves if the first or last waypoints are hidden
+  const spaceTimeChartLayersData = useMemo(() => {
+    let filteredProjectPathTrainResult = projectPathTrainResult;
+    let filteredConflicts = conflicts;
+
+    if (!waypointsPanelData) return { filteredProjectPathTrainResult, filteredConflicts };
```
clarani: You can check here that there is more than 2 waypoints (remove the condition line 66)
```suggestion
    if (!waypointsPanelData || waypointsPanelData.filteredWaypoints.length <= 2) return { filteredProjectPathTrainResult, filteredConflicts };
```

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -45,9 +51,94 @@ const ManchetteWithSpaceTimeChartWrapper = ({
...
+
+    const { filteredWaypoints } = waypointsPanelData;
+    const firstPositon = filteredWaypoints.at(0)!.position;
```
clarani: Typo
```suggestion
    const firstPosition = filteredWaypoints.at(0)!.position;
```

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -45,9 +51,94 @@ const ManchetteWithSpaceTimeChartWrapper = ({
...
+              spaceEnd: signal.position_end,
+              timeStart: signal.time_start,
+              timeEnd: signal.time_end,
+            };
+            const { spaceStart, spaceEnd, timeStart, timeEnd } = formattedSignal;
+            const updatedSignalRange = interpolateRange(
+              { spaceStart, spaceEnd, timeStart, timeEnd },
+              firstPositon,
+              lastPosition
+            );
```
clarani: You can directly call the function interpolateRange
```suggestion
            const updatedSignalRange = interpolateRange(
              {
                spaceStart: signal.position_start,
                spaceEnd: signal.position_end,
                timeStart: signal.time_start,
                timeEnd: signal.time_end,
              },
              firstPositon,
              lastPosition
            );
```

front/src/modules/simulationResult/helpers/utils.ts
```diff
@@ -0,0 +1,34 @@
+import type { LayerRangeData } from '../types';
+
+const interpolateRange = (
```
clarani: I'm not sure interpolateRange is the best name for this function. Maybe we can call it cutSpaceTimeChartBlock ? Or something like that ?

Also, this function can be in the folder SpaceTimeChart (I think SpaceTimeChart should be a module)

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
emersion: Typo: "position"

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -45,9 +51,94 @@ const ManchetteWithSpaceTimeChartWrapper = ({
...
+
+    if (
+      filteredWaypoints.length >= 2 &&
+      (firstPositon !== 0 || lastPosition !== operationalPoints.at(-1)!.position)
+    ) {
+      filteredProjectPathTrainResult = projectPathTrainResult.map((train) => ({
+        ...train,
+        space_time_curves: train.space_time_curves.map(({ positions, times }) => {
+          const cutPositions = new Set<number>();
+          const cutTimes = new Set<number>();
```
emersion: Why use a `Set` here instead of an `Array`?

A `Set` will forbid multiple positions/times with the same value. If a curve is not strictly monotonic (e.g. if a curve goes down and then up, or is constant), then we'd drop some data points. Also I think there is an assumption that there are as many position values are there are time values.

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -45,9 +51,94 @@ const ManchetteWithSpaceTimeChartWrapper = ({
...
+          }
+
+          return {
+            positions: Array.from(cutPositions),
+            times: Array.from(cutTimes),
+          };
+        }),
+        signal_updates: compact(
+          train.signal_updates.map((signal) => {
+            const formattedSignal = {
```
emersion: It doesn't seem like `formattedSignal` is used for anything apart from extracting fields to be fed to `interpolateRange()`? I think we can simplify this a bit:

```ts
interpolateRange({
  spaceStart: signal.position_start,
  // …
}, firstPosition, lastPosition)
```

front/src/modules/simulationResult/helpers/utils.ts
emersion: Right, I'm pleading guilty here. I picked this name because originally this function only did some interpolation and nothing else, but now it does quite a bit more. New name sounds good to me (obligatory alternative name for the bikeshed: `cutSpaceTimeRect`).

emersion:
Sorry for the duplicate comments. It seems @clarani and I ended up reviewing in parallel without noticing.

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
SharglutDev: Done but with `< 2`

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
SharglutDev: You're right done

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx

front/src/modules/simulationResult/helpers/utils.ts
SharglutDev: Ok let's go for @emersion proposal. And moved the util in space time chart folder

clarani (approved):

PR #9514, opened 2024-10-29
core: stdcm: fix priority between stop and departure delay

### LINKED ISSUE #9508: core: stdcm sometimes fails to find a path for no apparent reason
### What happened?

Sometimes, the stdcm search ends almost instantly with no solution found. The response is the "conflict" by default, but there's actually no conflict. Meaning that there was an easy solution available. 

A "conflict" response without any conflict doesn't make any sense

I have payloads that reproduce but it's an infra + a schedule + timetables + stdcm request, it's a mess

Local, dev

Firefox

508ac321f2791dcbb8ca6dfdbee03b4a51261635

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/9508

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/TimeData.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/TimeData.kt
@@ -1,5 +1,7 @@
 package fr.sncf.osrd.stdcm.graph
 
+import com.google.common.primitives.Doubles.min
+import kotlin.Double.Companion.POSITIVE_INFINITY
 import kotlin.math.max
 
 /**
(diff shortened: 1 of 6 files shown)


core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/TimeData.kt
```diff
@@ -27,6 +29,7 @@ data class TimeData(
      * stop has been reached yet.
      */
     val maxDepartureDelayingWithoutConflict: Double,
+    val maxFirstDepartureDelaying: Double,
```
Erashin, line 39: Might be worth adding a comment for this value here, just in case people get confused.

Erashin (commented):
Unit tests looking good as well.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/TimeData.kt
eckter: I've added a comment there.

I'll probably rename variables and rewrite comments in this class, it's a bit messy. But I'd do it in a PR that has no functional change.

Erashin (approved):

PR #9516, opened 2024-10-29
front: fix conflicts card displayed time

### LINKED ISSUE #9454: front: the times in the conflict cards are incorrect
### What happened?

I created a train which leaves at 12h. I duplicated, the 2nd train also leaves at 12h.
I should see a conflict starting at 12h, but it starts at 10h:
Image

I should see a conflict starting at 12h.

1. Create a scenario
2. Create a train, which leaves at 12h
3. Create a second train without changing any inputs.
4. Display the simulation results, open the conflicts tab

Recette (SNCF)

Firefox

.

Akctarus:
closes #9454

### DIFF
--- a/front/src/utils/date.ts
+++ b/front/src/utils/date.ts
@@ -20,7 +20,7 @@ const userTimeZone = dayjs.tz.guess(); // Format : 'Europe/Paris'
  * @param dateTimeString date string in ISO format
  * @returns string "HH:MM:SS"
  */
-export function extractHHMMSS(dateTimeString?: string) {
+function extractHHMMSS(dateTimeString?: string) {
   if (!dateTimeString) {
     return '';
   }
--- a/front/src/modules/conflict/components/ConflictCard.tsx
+++ b/front/src/modules/conflict/components/ConflictCard.tsx
@@ -1,9 +1,9 @@
 import dayjs from 'dayjs';
 import { useTranslation } from 'react-i18next';
 
-import { extractHHMMSS } from 'utils/date';
-
 import type { ConflictWithTrainNames } from '../types';
+
+const formatToLocalTime = (dateString: string) => dayjs.utc(dateString).local().format('HH:mm:ss');
 
 const ConflictCard = ({
   conflict,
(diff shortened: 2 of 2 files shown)


front/src/utils/date.ts
```diff
@@ -20,7 +20,10 @@ const userTimeZone = dayjs.tz.guess(); // Format : 'Europe/Paris'
  * @param dateTimeString date string in ISO format
  * @returns string "HH:MM:SS"
  */
-export function extractHHMMSS(dateTimeString?: string) {
+export const isoStringToLocalDate = (dateString: string) =>
```
emersion: Hm, it's not super clear from the name that this function is just formatting the hours+minutes+seconds, and leaving out the day+month+year.

Also other call-sites may need slightly different formatting (e.g. leaving out the seconds), so maybe we should move this inside `ConflictCard` for now?

emersion:
A good follow-up task may to be get rid of `extractHHMMSS()` completely, because that function is just a footgun.

Akctarus:
> A good follow-up task may to be get rid of `extractHHMMSS()` completely, because that function is just a footgun.

Yes, I'd thought of doing it in the refactor ticket for the date.ts file, there's a lot to do there.

SharglutDev (approved):
Lgtm and tested. Agreed with @emersion, maybe the name of the function could be better

Uriel-Sautron (approved):
LGTM, tested

front/src/utils/date.ts
Akctarus: Fixed :)

front/src/utils/date.ts
```diff
@@ -20,7 +20,7 @@ const userTimeZone = dayjs.tz.guess(); // Format : 'Europe/Paris'
  * @param dateTimeString date string in ISO format
  * @returns string "HH:MM:SS"
  */
-export function extractHHMMSS(dateTimeString?: string) {
+function extractHHMMSS(dateTimeString?: string) {
```
emersion, line 23: Oh, so is this function unused now? If so, can we nuke it?

Akctarus: it's used in timestampToHHMMSS, so no need to export it anymore, but we need to keep it until we refactor the file

emersion: Ah right, makes sense! Good to stop exporting it anyways :)

emersion (approved):
LGTM!

PR #9517, opened 2024-10-30
stdcm: show consist params in simulation raport

### LINKED ISSUE #9460: stdcm : parameters of the convoy are not taken into account in the report and in the table
### What happened?

I created a convoy in STDMC and exported the simulation report and looked at the table, and the information of the convoy are not displayed (still uses the default ones : length, max speed, weight)

Example for the weight: 
- My convoy : Image

- Report : Image
 
- Table : Image

I expected to have the correct information

Create a convey and observe the bug

Recette (SNCF)

Google Chrome

7451939

Wadjetz:
fix #9460

Show user-defined consist value in the simulation results.

### DIFF
--- a/front/src/applications/stdcm/components/StdcmResults/StdcmResultsTable.tsx
+++ b/front/src/applications/stdcm/components/StdcmResults/StdcmResultsTable.tsx
@@ -6,12 +6,14 @@ import { useTranslation } from 'react-i18next';
 
 import type {
   StdcmResultsOperationalPointsList,
+  StdcmSimulationInputs,
   StdcmSuccessResponse,
 } from 'applications/stdcm/types';
 import { getStopDurationTime } from 'applications/stdcm/utils/formatSimulationReportSheet';
 
 type SimulationTableProps = {
   stdcmData: StdcmSuccessResponse;
+  consist: StdcmSimulationInputs['consist'];
   isSimulationRetained: boolean;
   operationalPointsList: StdcmResultsOperationalPointsList;
   onRetainSimulation: () => void;
@@ -19,6 +21,7 @@ type SimulationTableProps = {
 
 const StcdmResultsTable = ({
   stdcmData,
+  consist,
   onRetainSimulation,
   isSimulationRetained,
   operationalPointsList,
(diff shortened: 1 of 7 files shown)


front/src/applications/stdcm/hooks/useStdcmForm.ts
SharglutDev, line 1: You need to also pass totalMass, totalLength and maxSpeed to the useEffect dependency array. Otherwise, when we change a value the simulation button does'nt appear again.

front/src/applications/stdcm/components/StdcmResults/StdcmResultsTable.tsx
```diff
@@ -55,6 +58,9 @@ const StcdmResultsTable = ({
             const isPathStep =
               isFirstStep || isLastStep || (isRequestedPathStep && step.duration === 0);
             const isNotExtremity = !isFirstStep && !isLastStep;
+
+            const mass = stdcmSimulation.inputs.consist?.totalMass ?? stdcmData.rollingStock.mass / 1000;
```
SharglutDev: There is a linter error, you can `yarn prettier --write` to fix it

front/src/applications/stdcm/components/StdcmResults/StdcmResultsTable.tsx
```diff
@@ -6,19 +6,22 @@ import { useTranslation } from 'react-i18next';
...
 import type {
   StdcmResultsOperationalPointsList,
+  StdcmSimulation,
   StdcmSuccessResponse,
 } from 'applications/stdcm/types';
 import { getStopDurationTime } from 'applications/stdcm/utils/formatSimulationReportSheet';
 
 type SimulationTableProps = {
   stdcmData: StdcmSuccessResponse;
+  stdcmSimulation: StdcmSimulation;
```
SharglutDev: As we use only the mass here, we don't need to pass all this object in props

```suggestion
  selectedSimulationTotalMass?: number;
```

front/src/applications/stdcm/types.ts
```diff
@@ -28,6 +28,7 @@ export type StdcmSuccessResponse = Omit<
 
 export type SimulationReportSheetProps = {
   stdcmData: StdcmSuccessResponse;
+  stdcmSimulation: StdcmSimulation;
```
SharglutDev: Same here we could only the consist datas

```suggestion
  stdcmSimulationConsist: StdcmSimulation['inputs']['consist'];
```

SharglutDev (commented):
Thank you for the fix, left a few comment

front/src/applications/stdcm/components/StdcmResults/StdcmResultsTable.tsx
Wadjetz: in the future we will pass more parameters, but i can pass only the consist field

SharglutDev (approved):
Lgtm and tested

PR #9518, opened 2024-10-30
front: use Date in TrainScheduleWithDetails

emersion:
Instead of passing around formatted strings, pass a Date object. That way we don't need to undo the formatting in TimetableTrainCard later on, and we get improved type safety.

No behavior change intended.

### DIFF
--- a/front/src/modules/trainschedule/components/Timetable/TimetableTrainCard.tsx
+++ b/front/src/modules/trainschedule/components/Timetable/TimetableTrainCard.tsx
@@ -34,6 +34,9 @@ type TimetableTrainCardProps = {
   projectionPathIsUsed: boolean;
 };
 
+const formatFullDate = (d: Date) => dayjs(d).format('D/MM/YYYY HH:mm:ss');
+const formatDateHours = (d: Date) => dayjs(d).format('HH:mm');
+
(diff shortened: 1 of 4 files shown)


front/src/modules/trainschedule/components/Timetable/TimetableTrainCard.tsx
```diff
@@ -34,6 +34,9 @@ type TimetableTrainCardProps = {
   projectionPathIsUsed: boolean;
 };
 
+const formatFullDate = (d: Date) => dayjs(d).format('D/MM/YYYY HH:mm:ss');
+const formatDateHours = (d: Date) => dayjs(d).format('HH:mm:ss');
```
emersion: It would probably be better to use the `Intl` locale instead of dayjs for formatting here, because that would then use the locale's preferred date style instead hardcoding the French format for everybody. However that'd be a change of behavior so I'd prefer to keep it in a separate patch.

Akctarus (approved):
tested, LGTM, nice refactor :)

theocrsb (approved):
LGTM and tested!

clarani (approved):
LGTM :white_check_mark:

PR #9521, opened 2024-10-30
front: remove all remaining JavaScript files

emersion:
Use TypeScript everywhere.

Make some component properties optional to reflect current usage.


front/src/common/BootstrapSNCF/CardSNCF/CardSNCF.tsx
```diff
@@ -5,7 +5,7 @@ interface CardProps {
   link: string;
   img: string;
   title: string;
-  disabledLink: boolean;
+  disabledLink?: boolean;
```
theocrsb, line 8: Just one question: does this change have anything to do with the commit name?

theocrsb (commented):
LGTM, not tested

front/src/common/BootstrapSNCF/CardSNCF/CardSNCF.tsx
emersion: It's the "Make some component properties optional to reflect current usage." part :) 

The JavaScript files were constructing a `CardProps` object without a `disabledLink` property. This happens to work fine, but the type definition did not allow this. Previously the TypeScript compiler wouldn't complain because it doesn't check pure JavaScript files, but now that these are TypeScript, it errors out.

theocrsb: Thanks for the explanation !

Yohh (approved):

clarani (approved):
LGTM :white_check_mark:

PR #9527, opened 2024-10-31
tests: fix flaky test with hardcoded ID

eckter:
The actual ID would depend on the state of the DB, it gets incremented whenever a work schedule is added


shenriotpro (approved):
Thanks

PR #9553, opened 2024-11-04
Fix update pathfinding when changing rolling stock

### LINKED ISSUE #8570: Changing the rolling stock to make an invalid train valid does not exit the pathfinding debug interface
### What happened?

1. I start creating (WITHOUT SAVING) a train which is invalid due to constraints (PLY-DN with electric 25000V rolling stock) -> I enter the debug interface
2. I change the rolling stock to have a valid train (a thermic rolling stock) and go back to the itinerary tab -> I am still in the debug interface

Also seems to be the case the other way around : valid train (normal itinerary tab) -> invalid train (does not trigger the pathfinding debut)

I expected the itinerary tab to be updated without the debug interface when I use a correct rolling stock

as above

Dev (SNCF)

Google Chrome

1482e8c

Uriel-Sautron:
close #8570 

For tested: 
  
  - Create a PF that is invalid due to constraints ( ex: PLY -> MSC with RS 2500V) incompatibilities will be displayed at the bottom of the map.
  - Change RS for thermic, back to the Itinerary tab the incompatibilities should no longer appear.

### DIFF
--- a/front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
+++ b/front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
@@ -93,6 +93,8 @@ const ManageTrainSchedule = ({ trainIdToEdit }: ManageTrainScheduleProps) => {
       <RollingStockSelector
         rollingStockSelected={rollingStock}
         rollingStockComfort={rollingStockComfort}
+        pathProperties={pathProperties}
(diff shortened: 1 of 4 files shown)


Caracol3 (approved):
lgtm (tested)

Math-R (approved):
lgtm nice work !

PR #9554, opened 2024-11-04
front, scripts: add playwright container for e2e tests

### LINKED ISSUE #9505: front: add procedure to run e2e tests in docker to the readme
### Description and goal

Playwright does not support most linux distrib beyond ubuntu/debian. Integration/e2e tests can still be run locally running playwright inside docker. The instructions to do this should be added to the front readme.

Optionally, the instructions to run tests should be kept in their own readme, with a redirection from the main readme.

### Acceptance criteria

The procedure to run e2e tests in docker should be clearly explained in the readme

Synar:
Close #9505

Instead of simply expanding the readme as initially thought in the issue, 
a dockerfile is added for the playwright container,
and a script is used to launch it with optional ui flags for x11 and wayland.

The wayland flag is currently not working (hence the draft status), but the headless and x11 mode are working and ready for review. Interestingly, the x11 mode works on wayland through compatibility.

### DIFF
--- a/front/.dockerignore
+++ b/front/.dockerignore
@@ -1,3 +1,4 @@
 node_modules
 npm-debug.log
 Dockerfile
+playwright-report
--- a/front/README.md
+++ b/front/README.md
@@ -45,21 +45,27 @@ It requires:
 
 Now you can run the test with `cd front/ && yarn e2e-tests`.
 
-> [!CAUTION]
-> If you try to run `yarn start` instead of running it through docker, you'll notice it doesn't
-> work because the gateway can't access your local port from inside a container. 2 solutions:
+If you are using a Linux distribution not supported by playwright (which only supports Windows, macOS and Ubuntu/Debian),
+you can instead start the tests inside a docker container using `osrd/scripts/run-front-playwright-container.sh`.
+You may pass the same options and arguments to this script as you would to `yarn e2e-tests` or `yarn playwright test`.
+
+> [!CAUTION] If you try to run `yarn start` instead of running it through docker, you'll notice it
+> doesn't work because the gateway can't access your local port from inside a container. 2
+> solutions:
 >
 > - run all the components locally (you might keep Postgres and Valkey in containers)
 > - if on Linux, you can also launch all the containers on the host network: you can replace the
-> `docker compose <something>` above with `osrd/scripts/host-compose.sh <something>`
+>   `docker compose <something>` above with `osrd/scripts/host-compose.sh <something>`
 
 If the tests fail, you'll find a `front/test-results` folder that will contain videos of the fail
 test executions. They might be of help to understand what's going on. Note that the CI also exports
 these videos as artifacts.
 
-You may also want to explore the documentation of the test framework [Playwright](https://playwright.dev/).
-For example, try launching each test independently using `yarn playwright test --ui`, or debug a
-test with `yarn playwright test --debug`.
+You may also want to explore the documentation of the test framework
+[Playwright](https://playwright.dev/). For example, try launching each test independently using
+`yarn playwright test --ui`, debug a test with `yarn playwright test --debug`, or launch a specified
+test of a specified test file with a specified browser once with for example
+`yarn playwright test 011-op-times-and-stops-tab.spec.ts -g "should correctly set and display times and stops tables" --project=firefox  --retries=0`.
 
 ## Design rules
 
--- a/front/docker/Dockerfile.playwright
+++ b/front/docker/Dockerfile.playwright
@@ -1,0 +1,10 @@
+ARG PLAYWRIGHT_VERSION=latest
+FROM mcr.microsoft.com/playwright:$PLAYWRIGHT_VERSION
+
+COPY front/package.json front/yarn.lock /app/front/
+WORKDIR /app/front
+RUN yarn install --frozen-lockfile
+
+COPY front /app/front/
+COPY tests /app/tests
+RUN chmod a+w /app/front/
--- a/scripts/run-front-playwright-container.sh
+++ b/scripts/run-front-playwright-container.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+
+set -e
+
+# Open the base osrd folder, assuming the script is located in osrd/scripts
+cd "$(realpath "$(dirname "$0")"/..)"
+
+# Detect the playwright version installed in the front
+if ! command -v jq &> /dev/null; then
+  echo "Error: jq is not installed. Please install jq using your package manager to continue." >&2
+  exit 1
+fi
+cd front
+VERSION=$(yarn list --pattern playwright --json | jq -r '.data.trees[].name | split("@")[-1]' | sort -u)
+if [ "$(echo "$VERSION" | wc -l)" -ne 1 ]; then
+  echo "Error: Zero or multiple playwright versions found: $VERSION" >&2
+  exit 1
+fi
+cd ..
+
+# Loop through each argument passed to the scripts, and replace --ui with --ui=host=localhost
+args=()
+for arg in "$@"; do
+  if [ "$arg" = "--ui" ]; then
+    args+=("--ui-host=localhost")
+  else
+    args+=("$arg")
+  fi
+done
+
+docker build --build-arg PLAYWRIGHT_VERSION=v"$VERSION" -t osrd-playwright:latest -f front/docker/Dockerfile.playwright .
+
+# Create the bind mounted folders if they don't exist, to avoid them being created as root
+mkdir -p "$PWD/front/playwright-report"
+mkdir -p "$PWD/front/test-results"
+
+docker run -it --rm \
+    --ipc=host \
+    --network=host \
+    -v "$PWD/front/playwright-report:/app/front/playwright-report" \
+    -v "$PWD/front/test-results:/app/front/test-results" \
+    -u "$(stat -c %u:%g .)" \
+    osrd-playwright:latest yarn playwright test "${args[@]}"
--- a/flake.nix
+++ b/flake.nix
@@ -84,6 +84,7 @@
 
               # OSRD dev scripts
               osrd-dev-scripts
+              jq
 
             ]
             # Section added only on Linux systems


front/.dockerignore
```diff
@@ -1,3 +1,4 @@
 node_modules
 npm-debug.log
 Dockerfile
+playwright-report
```
Khoyo: Missing newline at EOF

scripts/run-front-playwright-container.sh
```diff
@@ -0,0 +1,47 @@
...
+        -v $XDG_RUNTIME_DIR/$WAYLAND_DISPLAY:/run/user/$(id -u)/$WAYLAND_DISPLAY \
+        -v "$PWD/front/playwright-reports:/app/front/playwright-reports" \
+        osrd-dev-playwright:latest bash
+
+else
+    docker run -it --rm \
+        --network="host" \
+        -v "$PWD/front/playwright-reports:/app/front/playwright-reports" \
+        osrd-dev-playwright:latest bash
+fi
```
Khoyo: Missing newline at EOF

scripts/run-front-playwright-container.sh
```diff
@@ -0,0 +1,47 @@
...
+    docker run -it --rm \
+        --network="host" \
+        -e DISPLAY=$DISPLAY \
+        -v /tmp/.X11-unix:/tmp/.X11-unix \
+        -v "$PWD/front/playwright-reports:/app/front/playwright-reports" \
+        osrd-dev-playwright:latest bash
+
+elif [[ "$1" == "--ui-wayland" ]]; then
+    docker run -it --rm \
+        --network="host" \
```
Khoyo: Unnessary quoting here

emersion:
Have you seen this? https://playwright.dev/docs/test-ui-mode#docker--github-codespaces

scripts/run-front-playwright-container.sh
```diff
@@ -0,0 +1,47 @@
...
+        -e WAYLAND_DISPLAY=$WAYLAND_DISPLAY \
+        -e XDG_RUNTIME_DIR=$XDG_RUNTIME_DIR \
+        -v $XDG_RUNTIME_DIR/$WAYLAND_DISPLAY:/run/user/$(id -u)/$WAYLAND_DISPLAY \
```
Khoyo: And missing quotes there

scripts/run-front-playwright-container.sh
```diff
@@ -0,0 +1,47 @@
...
+        -v "$PWD/front/playwright-reports:/app/front/playwright-reports" \
+        osrd-dev-playwright:latest bash
+
+else
+    docker run -it --rm \
+        --network="host" \
+        -v "$PWD/front/playwright-reports:/app/front/playwright-reports" \
+        osrd-dev-playwright:latest bash
```
Khoyo: Same comments here

scripts/run-front-playwright-container.sh
```diff
@@ -0,0 +1,47 @@
...
+    cleanup() {
+        echo "Revoking Docker's access to the X11 server..."
+        xhost -local:docker
+    }
+    # Set a trap to call the cleanup function on EXIT or when interrupted (e.g., with Ctrl+C)
+    trap cleanup EXIT
+
+    docker run -it --rm \
+        --network="host" \
+        -e DISPLAY=$DISPLAY \
```
Khoyo: And there

scripts/run-front-playwright-container.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/sh
+
+set -e
+
+if [[ $# -gt 1 || ( -n "$1" && "$1" != "--ui-x11" && "$1" != "--ui-wayland" ) ]]; then
+    echo "Usage is run-front-playwright, run-front-playwright --ui-x11, or run-front-playwright --ui-wayland"
```
Khoyo:
```suggestion
    echo "usage: $0 [--ui-x11|--ui-wayland]"
```

scripts/run-front-playwright-container.sh
```diff
@@ -0,0 +1,47 @@
...
+fi
+
+# Open the base osrd folder, assuming the script is located in osrd/scripts
+cd "$(dirname "$0")/.." 
+
+docker build -t osrd-dev-playwright:latest -f front/docker/Dockerfile.playwright .
+
+# Switch between the x11 or wayland ui mode or the headless mode
+if [[ "$1" == "--ui-x11" ]]; then
+    xhost +local:docker
```
Khoyo: I'm not sure why this is required, and I'm pretty sure it's wrong anyway.

Quoting from xhost manpage:

> The local family specifies all the local connections at once. However, the server interpreted address "si:localuser:username" can be used to specify a single local user. (See the Xsecurity(7) manual page for more details.)

So specifying docker here actually has no effects, all local users are now allowed access to the X server if I understand this correctly (@emersion ?) 

But anyway, isn't the user using the socket inside the container root anyway? I've never had to do anything like that when mounting the X11 socket inside a container.

scripts/run-front-playwright-container.sh
```diff
@@ -0,0 +1,47 @@
...
+    # Define a cleanup function to revoke access when the script exits
+    cleanup() {
+        echo "Revoking Docker's access to the X11 server..."
+        xhost -local:docker
```
Khoyo: If we do this, we need to make sure it wasn't setup like that before.

front/README.md
```diff
@@ -40,26 +40,35 @@ Launches end to end tests.
...
+  `docker compose up --no-build --detach valkey postgres gateway core editoast`
 - Running front with `docker compose up --build --detach front`
 
 Now you can run the test with `cd front/ && yarn e2e-tests`.
 
-> [!CAUTION]
-> If you try to run `yarn start` instead of running it through docker, you'll notice it doesn't
-> work because the gateway can't access your local port from inside a container. 2 solutions:
+If your os is not supported by playwright (which only supports windows, macos and ubuntu/debian),
+you can instead start the tests inside a docker container using
```
Khoyo: That is, if your OS isn't supported and is Linux.

Synar:
> Have you seen this? https://playwright.dev/docs/test-ui-mode#docker--github-codespaces

I had not, this should be much simpler

front/README.md
Synar: Sure, sure. But are there many other OS we care to support for dev purposes? ^^''

Khoyo: We don't really, so let's not pretend we do ;)

front/README.md
```diff
@@ -40,26 +40,35 @@ Launches end to end tests.
...
+../scripts/run-front-playwright-container.sh [--ui-x11] [--ui-wayland]. By default playwright will
```
Yohh:
```suggestion
`../scripts/run-front-playwright-container.sh [--ui-x11] [--ui-wayland]`. By default playwright will
```

front/README.md
```diff
@@ -40,26 +40,35 @@ Launches end to end tests.
...
+be started in headless mode, to use the --ui flag inside the container use the appropriate flag in
```
Yohh:
```suggestion
be started in headless mode, to use the `--ui` flag inside the container use the appropriate flag in
```

Yohh:
could we put `run-front-playwright-container.sh` in `front/scripts`?

Khoyo:
> could we put run-front-playwright-container.sh in front/scripts?

I mildly disagree, running the e2e tests shouldn't be a front thing.

Synar:
> Have you seen this? https://playwright.dev/docs/test-ui-mode#docker--github-codespaces

Tested, seems to work perfectly and should greatly simplify the script. Thanks a lot!

scripts/run-front-playwright-container.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/sh
+
+set -e
+
+if [[ $# -gt 1 || ( -n "$1" && "$1" != "--ui-x11" && "$1" != "--ui-wayland" ) ]]; then
```
Khoyo: `[[` is a bashism (ie. it's supported by bash, but not by all posix shells that could be at `/bin/sh`), either use it in the shebang or use `[`

scripts/run-front-playwright-container.sh
```diff
@@ -0,0 +1,47 @@
...
+    exit 1
+fi
+
+# Open the base osrd folder, assuming the script is located in osrd/scripts
+cd "$(dirname "$0")/.." 
+
+docker build -t osrd-dev-playwright:latest -f front/docker/Dockerfile.playwright .
+
+# Switch between the x11 or wayland ui mode or the headless mode
+if [[ "$1" == "--ui-x11" ]]; then
```
Khoyo: [[ is a bashism, either use it in the shebang or use [

scripts/run-front-playwright-container.sh
```diff
@@ -0,0 +1,47 @@
...
+    trap cleanup EXIT
+
+    docker run -it --rm \
+        --network="host" \
+        -e DISPLAY=$DISPLAY \
+        -v /tmp/.X11-unix:/tmp/.X11-unix \
+        -v "$PWD/front/playwright-reports:/app/front/playwright-reports" \
+        osrd-dev-playwright:latest bash
+
+elif [[ "$1" == "--ui-wayland" ]]; then
```
Khoyo: [[ is a bashism, either use it in the shebang or use [

scripts/run-front-playwright-container.sh
Synar: It didn't work for me without 

But anyway I will use emersion's solution which does not require this

front/README.md
Synar: Done, thanks!

front/README.md
Synar: Done!

front/README.md
Synar: Done, thanks!

scripts/run-front-playwright-container.sh
Synar: Deleted code. Thanks though!

scripts/run-front-playwright-container.sh
Synar: Done, thanks!

scripts/run-front-playwright-container.sh
Synar: Done, thanks!

front/.dockerignore
Synar: Done, thanks!

Synar:
Only problem remaining is that @ElysaSrc and @multun old hack to make sure created files are owned by the user and not root does not seem to be working here yet. Can you take a look @ElysaSrc ?

scripts/run-front-playwright-container.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+set -e
+
+# Open the base osrd folder, assuming the script is located in osrd/scripts
+cd "$(dirname "$0")/.."
```
Khoyo: For consistency, either the other scripts (eg. `scripts/host-compose.sh`) should do this or this one shouldn't

scripts/run-front-playwright-container.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+set -e
+
+# Open the base osrd folder, assuming the script is located in osrd/scripts
+cd "$(dirname "$0")/.." 
+
+docker build -t osrd-dev-playwright:latest -f front/docker/Dockerfile.playwright .
```
Khoyo: Just call it `osrd-playwright`

front/docker/Dockerfile.playwright
```diff
@@ -0,0 +1,19 @@
...
+WORKDIR /app
+COPY front front
+COPY tests tests
+WORKDIR /app/front
+RUN yarn install
+RUN yarn playwright install --with-deps
+
+# Start the app
+COPY front/docker/dev-entrypoint.sh /
+ENTRYPOINT ["/dev-entrypoint.sh"]
```
Khoyo: Don't use this entrypoint, it's only useful if you bindmount the whole `front/` folder.

scripts/run-front-playwright-container.sh
```diff
@@ -0,0 +1,13 @@
...
+
+docker run -it --rm \
+    --network=host \
```
Khoyo, line 39: Add `-u "$(stat -c %u:%g .)"` here

front/README.md
```diff
@@ -45,21 +45,28 @@ It requires:
 
 Now you can run the test with `cd front/ && yarn e2e-tests`.
 
-> [!CAUTION]
-> If you try to run `yarn start` instead of running it through docker, you'll notice it doesn't
-> work because the gateway can't access your local port from inside a container. 2 solutions:
+If you are using a linux distribution not supported by playwright (which only supports windows, macos and ubuntu/debian),
```
Khoyo:
```suggestion
If you are using a Linux distribution not supported by playwright (which only supports Windows, macOS and Ubuntu/Debian),
```

emersion: Actually the correct capitalization is "macOS" :P

scripts/run-front-playwright-container.sh
Synar: It is already done in sync-openapi.sh but not other scripts, so this needs to be settled before this pr is merged

front/README.md
Khoyo: Fix'd :p

front/docker/Dockerfile.playwright
```diff
@@ -0,0 +1,19 @@
...
+
+FROM mcr.microsoft.com/playwright
+
+COPY --from=exec_as_builder /exec-as /
+WORKDIR /app
+COPY front front
+COPY tests tests
+WORKDIR /app/front
+RUN yarn install
+RUN yarn playwright install --with-deps
```
Khoyo: This should probably not be dependent on the *whole* front, only its dependencies

front/docker/Dockerfile.playwright
```diff
@@ -0,0 +1,19 @@
+FROM alpine AS exec_as_builder
+
+RUN apk add build-base
+COPY front/docker/exec-as.c .
+RUN gcc -std=c99 -static -o /exec-as exec-as.c
+
+FROM mcr.microsoft.com/playwright
```
emersion: Hm, I wonder… We're pulling the Playwright Docker image, and then we're installing Playwright inside it: do we really need both?

Synar: I believe we can stop installing playwright with its deps, as long as we make sure to match the image we pull to the playwright version we have in the yarn.lock. This means bumping the dockerfile every time we bump our yarn playwright version, and thus more things to maintain.

I don't know which is preferable.

front/docker/Dockerfile.playwright
Synar: Done, thanks

front/README.md
Synar: Done, thanks

scripts/run-front-playwright-container.sh
Synar: Done, tahnks!

front/docker/Dockerfile.playwright
Synar: Done ^^

scripts/run-front-playwright-container.sh
Synar: Done

front/docker/Dockerfile.playwright
Synar: A middleground could be to keep the install with-deps but still try to match the versions.

The install with-deps is three times faster (on my computer) when the versions correctly match, so keeping it ensures that the script won't crash without sacrificing too much time.

ghost (approved):

scripts/run-front-playwright-container.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
```
Khoyo:
```suggestion
#!/usr/bin/env bash
```

We use arrays now, which are not a feature of (all) POSIX shells.

Khoyo (changes requested):

scripts/run-front-playwright-container.sh
```diff
@@ -0,0 +1,29 @@
...
+
+set -e
+
+# Open the base osrd folder, assuming the script is located in osrd/scripts
+cd "$(realpath "$(dirname "$0")"/..)"
```
Khoyo, line 6:
```suggestion
root_path="$(realpath "$(dirname "$0")"/..)"
cd "$root_path"
```

scripts/run-front-playwright-container.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Create the bind mounted folders if they don't exist, to avoid them being created as root
+mkdir -p "$PWD/front/playwright-report"
+mkdir -p "$PWD/front/test-results"
+
+docker run -it --rm \
+    --network=host \
+    -v "$PWD/front/playwright-report:/app/front/playwright-report" \
+    -v "$PWD/front/test-results:/app/front/test-results" \
+    -u "$(stat -c %u:%g .)" \
```
Khoyo, line 28: `-u "$(stat -c %u:%g "$root_path")"`

Khoyo (commented):
Nothing

front/docker/Dockerfile.playwright
emersion: We need to make sure to use the same Playwright version as `package.json`: we've had breakage in the past due to Playwright upgrades.

I'd like not to duplicate fixed dep versions in two places because it's so easy for them to go out of sync (especially with dependabot).

scripts/run-front-playwright-container.sh
Synar: Done thanks

Khoyo (approved):

front/docker/Dockerfile.playwright
Synar: In any cases we should end up using the correct version, the question is only if we try to match the image version to the yarn.lock to avoid installation or if we install it by hand.

If you want to match the versions, but don't want to repeat them, what would you suggest? @khoyo proposed adding a cli rule to enforce the versions matching. Would that be enough?

ghost: If you're willing to add a dependency to jq on your script (this package is available on all distributions and systems but it generally needs to be installed manually) you can do something like this:

```sh
VERSION=$(yarn list --pattern playwright --json | jq -r '.data.trees[].name | split("@")[-1]' | sort -u)
```

(The above method should be checked, I'm not sure it's the best way of extracting it).

Then you export this inside a variable and your variabilize the tag of the docker immage using a build arg:

```sh
docker build --build-arg PLAYWRIGHT_VERSION=$VERSION ....
```

with

```dockerfile
ARG PLAYWRIGHT_VERSION
FROM playwright:$PLAYWRIGHT_VERSION
```

Synar: Looks good thanks!

Yohh (approved):
once rebased, this PR totally fixes the known issues about e2e tests while using `host-compose.sh`, thank you

front/docker/Dockerfile.playwright
Maymanaf: The Playwright Docker image already includes the _Playwright browsers_ and all necessary _browser system dependencies_, so running `yarn playwright install --with-deps` is unnecessary, as it only installs _browser system dependencies_. 
We should always keep the Docker image updated to the latest version, as it primarily affects the browsers.

So the version of Playwright, specified in` package.json` doesn’t need to match the Docker image version. And the upgrade of this version can cause  breakages and failures. 
However,we can updated it to a stable and recent version, like v1.47.2.

Synar: > So the version of Playwright, specified in` package.json` doesn’t need to match the Docker image version. And the upgrade of this version can cause breakages and failures. However,we can updated it to a stable and recent version, like v1.47.2.

I believe you're wrong about this, [see](https://playwright.dev/docs/docker)

> It is recommended to always pin your Docker image to a specific version if possible. If the Playwright version in your Docker image does not match the version in your project/tests, Playwright will be unable to locate browser executables.

This corresponds to my experience. Before pinning the version, the script would fail if I didn't install --with-deps

Synar:
A bit outside the scope of this pr, but I noticed that the CI seems to spend a lot of time installing playwright deps with "npx playwright install --with-deps". Could we accelerate this using a microsoft/playwright based image, similar to what is done here?

Khoyo:
> A bit outside the scope of this pr, but I noticed that the CI seems to spend a lot of time installing playwright deps with "npx playwright install --with-deps". Could we accelerate this using a microsoft/playwright based image, similar to what is done here?

Potentially. A lot of the time is network transfer though, so it might not be much faster. Do you want to do it?

Synar:
Sure. We can even reuse Dockerfile.playwright wholesale right? I'll do this in another PR though

emersion (approved):

PR #9555, opened 2024-11-04
front: add vias in simulation results map and warped map

### LINKED ISSUE #9541: Add `vias` icon in scenario map

End users

part of https://github.com/osrd-project/osrd-confidential/issues/670

We have origin and destination icons. To be iso with the mockup, we need to add the vias

Akctarus:
closes #9541 

Vias should be displayed as on the STDCM map.

Note: Warped map don't handle vias on this PR. Itinerary markers are displayed with a gap on the wrap-around map, and also since the path is mostly vertical, displaying vias risks significantly reducing the visibility of the route.

### DIFF
--- a/front/src/applications/operationalStudies/views/SimulationResults.tsx
+++ b/front/src/applications/operationalStudies/views/SimulationResults.tsx
@@ -20,6 +20,7 @@ import type { ProjectionData } from 'modules/simulationResult/types';
 import TimesStopsOutput from 'modules/timesStops/TimesStopsOutput';
 import { updateViewport, type Viewport } from 'reducers/map';
 import { useAppDispatch } from 'store';
+import { getPointCoordinates } from 'utils/geometry';
 
 const SPEED_SPACE_CHART_HEIGHT = 521.5;
 const HANDLE_TAB_RESIZE_HEIGHT = 20;
@@ -64,6 +65,15 @@ const SimulationResults = ({
     pathProperties,
     infraId
   );
+
+  // Compute path items coordinates in order to place them on the map
+
+  const pathItemsCoordinates =
+    path &&
+    pathProperties &&
+    path.path_item_positions.map((positionOnPath) =>
+      getPointCoordinates(pathProperties.geometry, path.length, positionOnPath)
+    );
 
   const {
     operationalPoints: projectedOperationalPoints,
@@ -192,6 +202,7 @@ const SimulationResults = ({
                 }
               : undefined
           }
+          pathItemsCoordinates={pathItemsCoordinates}
           setMapCanvas={setMapCanvas}
         />
       </div>
--- a/front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
+++ b/front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
@@ -1,18 +1,23 @@
-import type { Feature, LineString } from 'geojson';
+import cx from 'classnames';
+import type { Feature, LineString, Position } from 'geojson';
 import { Marker, Source } from 'react-map-gl/maplibre';
 
 import destinationIcon from 'assets/pictures/mapMarkers/destination.svg';
+import viaIcon from 'assets/pictures/mapMarkers/intermediate-point.svg';
 import originIcon from 'assets/pictures/mapMarkers/start.svg';
 import OrderedLayer from 'common/Map/Layers/OrderedLayer';
 
 interface RenderItineraryProps {
   geojsonPath: Feature<LineString>;
+  pathItemsCoordinates?: Position[];
   layerOrder: number;
 }
 
-export default function RenderItinerary(props: RenderItineraryProps) {
-  const { geojsonPath, layerOrder } = props;
-
+export default function RenderItinerary({
+  geojsonPath,
+  pathItemsCoordinates,
+  layerOrder,
+}: RenderItineraryProps) {
   const paintBackgroundLine = {
     'line-width': 4,
     'line-color': '#EDF9FF',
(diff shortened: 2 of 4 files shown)


front/src/modules/simulationResult/components/SimulationResultsMap/SimulationResultsMap.tsx
```diff
@@ -92,6 +93,12 @@ const SimulationResultMap = ({
 
   const geojsonPath = useMemo(() => geometry && lineString(geometry.coordinates), [geometry]);
 
+  const { getVias } = useOsrdConfSelectors();
+  const vias = useSelector(getVias());
```
SharglutDev: You can't use the store to get the vias. It's not updated when we select a new train so you may not get the proper vias.

To get the vias coordinates, you can use the util `getPointCoordinates`. You already have the geometry, you can get the path length and the via's position by using the operational points we use in `SimulationResults`

SharglutDev (changes requested):

front/src/modules/simulationResult/components/SimulationResultsMap/SimulationResultsMap.tsx
Akctarus: fixed

front/src/applications/operationalStudies/views/SimulationResults.tsx
```diff
@@ -65,6 +66,13 @@ const SimulationResults = ({
     infraId
   );
 
+  const pathItemsCoordinates =
```
SharglutDev, line 71: Maybe a comment to explain why we do that

```suggestion
  // Compute path items coordinates in order to place them on the map
  const pathItemsCoordinates =
```

front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
```diff
@@ -23,15 +28,31 @@ export default function RenderItinerary(props: RenderItineraryProps) {
...
-  // We know we have a geometry so we have at least 2 coordinates
-  const [startLongitude, startLatitude] = geojsonPath.geometry.coordinates.at(0)!;
-  const [endLongitude, endLatitude] = geojsonPath.geometry.coordinates.at(-1)!;
+  if (!pathItemsCoordinates || pathItemsCoordinates.length < 2) {
+    return null;
+  }
+
+  const [originLongitude, originLatitude] = pathItemsCoordinates[0];
+  const [destinationLongitude, destinationLatitude] =
+    pathItemsCoordinates[pathItemsCoordinates.length - 1];
```
SharglutDev: We try to always use this syntax now. You can add the `!` because at this point we know there is at least 2 path items.

```suggestion
  const [originLongitude, originLatitude] = pathItemsCoordinates.at(0)!;
  const [destinationLongitude, destinationLatitude] =
    pathItemsCoordinates.at(-1)!;
```

front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
```diff
@@ -23,15 +28,31 @@ export default function RenderItinerary(props: RenderItineraryProps) {
...
 
   return (
     <Source type="geojson" data={geojsonPath}>
-      <Marker longitude={startLongitude} latitude={startLatitude} anchor="bottom">
+      <Marker longitude={originLongitude} latitude={originLatitude} anchor="bottom">
         <img src={originIcon} alt="origin" />
       </Marker>
+      {vias.map(([longitude, latitude], index) => (
+        <Marker key={`via-${index}`} longitude={longitude} latitude={latitude} anchor="bottom">
+          <img src={viaIcon} alt="via" />
```
SharglutDev:
```suggestion
          <img src={viaIcon} alt=`via ${index + 1}` />
```

front/src/applications/operationalStudies/views/SimulationResults.tsx
```diff
@@ -65,6 +66,13 @@ const SimulationResults = ({
...
+    path && path.length && path.path_item_positions
+      ? path.path_item_positions.map((positionOnPath) =>
+          getPointCoordinates(pathProperties!.geometry, path.length, positionOnPath)
+        )
+      : [];
+
```
SharglutDev: Not sure why but sometimes the `pathItemsCoordinates` received by `RenderItinerary` are undefined even though you set them to an empty array by default here. So better to allow this const to be undefined for consistency. 

You also need pathProperties to the condition in case it's still computing (and remove the `!` from `getPointCoordinates`) 

```suggestion
  const pathItemsCoordinates =
    path && pathProperties
      && path.path_item_positions.map((positionOnPath) =>
          getPointCoordinates(pathProperties.geometry, path.length, positionOnPath)
        );

```

SharglutDev (commented):
Tested and works great, left some comments for improving.

We still have one weird behavior though. If we have 2 vias very close and we are at a low zoom, the second one is hiding the first. But we have the same issue on the `ManageTrainSchedule` map. We should check that with @thibautsailly to fix it in another PR for both maps.

front/src/applications/operationalStudies/views/SimulationResults.tsx
Akctarus: done

front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
Akctarus: done

front/src/applications/operationalStudies/views/SimulationResults.tsx
Akctarus: done

front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
Akctarus: fixed

SharglutDev (approved):
Lgtm and tested, good job !

PR #9568, opened 2024-11-05
Fix: error message for non-recognized points in pathfinding

### LINKED ISSUE #9520: Behavior for non-recognized points in the path is not uniform
### What happened?

Depending on which point of the path is not recognized, the behavior is not the same (itinerary or not, error message or not)

### Message :  

If at least one point is not recognized (any point : origin, destination, waypoint), a red message should be shown to the user (the train will be invalid so it is important to show it as soon as possible) : 
- French : "Au moins un des points de passage n'a pas été reconnu"
- English : "At least one of the waypoints could not be recognized"
The red message should be here and look like this (with a different text)
Image

### Itinerary : 

If at least one point is not recognized (any point : origin, destination, waypoint), an itinerary ignoring these unrecognized points should be computed and displayed, except if only one point in the path is recognized. 
In both cases, all the recognized points (even if there is only one) should be shown on the map

Create a case where the origin is not recognized, and another case where a waypoint is not recognized, and observe the different behaviors.

Local

Google Chrome

na

achrafmohye:
Closes #9520

### DIFF
--- a/front/src/modules/pathfinding/components/Pathfinding/Pathfinding.tsx
+++ b/front/src/modules/pathfinding/components/Pathfinding/Pathfinding.tsx
@@ -10,6 +10,7 @@ import infraLogo from 'assets/pictures/components/tracks.svg';
 import { Spinner } from 'common/Loaders';
 import { useOsrdConfSelectors } from 'common/osrdContext';
 import { usePathfinding } from 'modules/pathfinding/hooks/usePathfinding';
+import { isPathStepInvalid } from 'modules/pathfinding/utils';
 import { useStoreDataForRollingStockSelector } from 'modules/rollingStock/components/RollingStockSelector/useStoreDataForRollingStockSelector';
 import { conditionalStringConcat, formatKmValue } from 'utils/strings';
 
@@ -23,7 +24,9 @@ type PathfindingProps = {
 const Pathfinding = ({ pathProperties, setPathProperties }: PathfindingProps) => {
   const { t } = useTranslation(['operationalStudies/manageTrainSchedule']);
 
-  const { getOrigin, getDestination } = useOsrdConfSelectors();
+  const { getOrigin, getDestination, getPathSteps } = useOsrdConfSelectors();
+  const pathSteps = useSelector(getPathSteps);
+  const hasInvalidPathStep = pathSteps.some((pathStep) => isPathStepInvalid(pathStep));
   const origin = useSelector(getOrigin, isEqual);
   const destination = useSelector(getDestination, isEqual);
   const { rollingStock } = useStoreDataForRollingStockSelector();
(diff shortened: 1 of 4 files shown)


front/src/modules/pathfinding/components/Pathfinding/Pathfinding.tsx
```diff
@@ -93,7 +97,9 @@ const Pathfinding = ({ pathProperties, setPathProperties }: PathfindingProps) =>
                 <Stop />
               </span>
               <span className="flex-grow-1">
-                {t('pathfindingError', { errorMessage: t(pathfindingState.error) })}
+                {hasInvalidPathStep
+                  ? t('InvalidTrainScheduleStep')
```
Caracol3, line 101: Maybe we should create this to be clearer:
French : "Au moins un des points de passage n'a pas été reconnu"
English : "At least one of the waypoints could not be recognized"

Caracol3:
We now lose the markers and the path on the map. I think we should keep this behavior.

Caracol3:
It works fine for the vias but when just the origin or the destination is invalid we still have the old message of success.
image

front/src/modules/pathfinding/components/Pathfinding/Pathfinding.tsx
achrafmohye: i'll add this :)

Caracol3:
When  going in itinerary, markers aren't displayed in the map when orign or destination is invalid.
If we had this in usePathding :
```
  useEffect(() => {
    if (origin?.isInvalid || destination?.isInvalid) {
      pathfindingDispatch({
        type: 'PATHFINDING_PARAM_CHANGED',
        params: { origin, destination, rollingStock },
      });
    }
  }, [origin, destination]);
  ```
  
  There is maybe a better way to do it but we have the markers in the map then.

**Review by Caracol3 — APPROVED** [review]
lgtm (tested)

File: front/src/modules/pathfinding/components/Pathfinding/Pathfinding.tsx [hunk 2]
```diff
@@ -23,16 +26,28 @@ type PathfindingProps = {
...
   } = usePathfinding(setPathProperties, pathProperties);
 
+  useEffect(() => {
+    if (origin?.isInvalid || destination?.isInvalid) {
+      pathfindingDispatch({
+        type: 'PATHFINDING_PARAM_CHANGED',
+        params: { origin, destination, rollingStock },
+      });
+    }
+  }, [origin, destination]);
```
clarani: Why did you add this useEffect ?

front/src/modules/pathfinding/components/Pathfinding/Pathfinding.tsx
```diff
@@ -23,13 +26,16 @@ type PathfindingProps = {
...
+  const { getOrigin, getDestination, getPathSteps } = useOsrdConfSelectors();
+  const pathSteps = useSelector(getPathSteps);
+  const hasInvalidPathStep = pathSteps.some((pathStep) => isPathStepInvalid(pathStep));
   const origin = useSelector(getOrigin, isEqual);
   const destination = useSelector(getDestination, isEqual);
   const { rollingStock } = useStoreDataForRollingStockSelector();
 
   const {
     pathfindingState,
+    pathfindingDispatch,
```
clarani: You can also remove this
```suggestion
```

achrafmohye: Done :)

clarani (approved):
LGTM

PR #9574, opened 2024-11-05
front: stdcm consist form validation

### LINKED ISSUE #9374: Handle invalid values in stdcm inputs
### Description and goal

Now that we have some inputs with number in stdcm, we need to handle some edge cases.

To be confirmed by @thibautsailly, but we probably want to handle this with the `statusWithMessage` prop in inputs and blocking the simulation button while in error like in the rest of the form. 

If we do that, it will need a refacto of the css of the stdcm consist part to keep matching the mockup.

### Acceptance criteria

- [ ] The following fileds with invalid or empty inputs display a warning message when focusing out. They are not displayed if no traction engine is selected.
    - tonnage (min = weight of traction engine + towed rolling stock, max = 10000t).
    🇫‍🇷 "Le tonnage total doit être compris entre ... et ...t". 
    🇬‍🇧 "The total weight must be between ... and ...t"
    - length (min = length of traction engine + towed rolling stock, max = 750m).
    🇫‍🇷 "La longueur totale doit être comprise entre ... et 750m"
    🇬‍🇧 "The total length must be between ... and 750m"
    - max speed (min = 30km/h, max = the value computed depending on traction engine, towed rs, and composition code).
    🇫‍🇷 "La vitesse max. doit être comprise entre 30 et ...km/h"
    🇬‍🇧 "The max speed must be between 30 and ...km/h"
- [ ] When an invalid value is entered in the consist card, the whole column gets larger to accommodate the display of the field error state. This pushes the operational points column to the right, and reduces the map's width ; all margins stay the same.
When the value becomes valid, all columns widths go back to their default values. All these interactions happen with a transition easeinout 300ms.

### LINKED ISSUE #9529: Add front check for mass and length consistency
To improve this PR https://github.com/OpenRailAssociation/osrd/pull/9383

Dependency : Needs this issue to be done : https://github.com/OpenRailAssociation/osrd/issues/9374

- Add a check in the front end to check that que input mass entered by the user is not smaller than towed mass + traction engin mass
- Add a check in the front end to check that que input length entered by the user is not smaller than towed length + traction engin length

Wadjetz:
fix #9529 
fix #9374

- [x] Add validation to consist form
- [x] Disable start simulation button with errors
- [x] Waiting for [PR](https://github.com/OpenRailAssociation/osrd-ui/pull/776)
- [x] Add tooltips status messages for inputs

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
@@ -5,9 +5,16 @@ import cx from 'classnames';
 import { useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
 
+import useStdcmTowedRollingStock from 'applications/stdcm/hooks/useStdcmTowedRollingStock';
 import { extractMarkersInfo } from 'applications/stdcm/utils';
+import {
+  validateMaxSpeed,
+  validateTotalLength,
+  validateTotalMass,
+} from 'applications/stdcm/utils/consistValidation';
 import { useOsrdConfActions, useOsrdConfSelectors } from 'common/osrdContext';
 import useInfraStatus from 'modules/pathfinding/hooks/useInfraStatus';
+import { useStoreDataForRollingStockSelector } from 'modules/rollingStock/components/RollingStockSelector/useStoreDataForRollingStockSelector';
 import { Map } from 'modules/trainschedule/components/ManageTrainSchedule';
 import { type StdcmConfSliceActions, resetMargins } from 'reducers/osrdconf/stdcmConf';
 import type { StdcmConfSelectors } from 'reducers/osrdconf/stdcmConf/selectors';
@@ -61,6 +68,9 @@ const StdcmConfig = ({
     getProjectID,
     getScenarioID,
     getStudyID,
+    getTotalMass,
+    getTotalLength,
+    getMaxSpeed,
   } = useOsrdConfSelectors() as StdcmConfSelectors;
   const origin = useSelector(getStdcmOrigin);
   const pathSteps = useSelector(getStdcmPathSteps);
@@ -69,10 +79,37 @@ const StdcmConfig = ({
   const studyID = useSelector(getStudyID);
   const scenarioID = useSelector(getScenarioID);
 
+  const totalMass = useSelector(getTotalMass);
+  const totalLength = useSelector(getTotalLength);
+  const maxSpeed = useSelector(getMaxSpeed);
+
   const pathfinding = useStaticPathfinding(infra);
   const formRef = useRef<HTMLDivElement>(null);
 
   const [formErrors, setFormErrors] = useState<StdcmConfigErrors>();
+
+  const { rollingStock } = useStoreDataForRollingStockSelector();
+  const towedRollingStock = useStdcmTowedRollingStock();
+
+  const consistErrors = useMemo(() => {
+    const totalMassError = validateTotalMass({
+      tractionEngineMass: rollingStock?.mass,
+      towedMass: towedRollingStock?.mass,
+      totalMass,
+    });
+
+    const totalLengthError = validateTotalLength({
+      tractionEngineLength: rollingStock?.length,
+      towedLength: towedRollingStock?.length,
+      totalLength,
+    });
+
+    return {
+      totalMass: totalMassError,
+      totalLength: totalLengthError,
+      maxSpeed: validateMaxSpeed(maxSpeed, rollingStock?.max_speed),
+    };
+  }, [rollingStock, towedRollingStock, totalMass, totalLength, maxSpeed]);
 
   const disabled = isPending || retainedSimulationIndex > -1;
 
@@ -146,7 +183,7 @@ const StdcmConfig = ({
           />
           <div className="stdcm-simulation-inputs">
             <div className="stdcm-consist-container">
-              <StdcmConsist disabled={disabled} />
+              <StdcmConsist consistErrors={consistErrors} disabled={disabled} />
             </div>
             <div className="stdcm__separator" />
             <div ref={formRef} className="stdcm-simulation-itinerary">
@@ -175,7 +212,10 @@ const StdcmConfig = ({
                   isDisabled={
                     disabled ||
                     !showBtnToLaunchSimulation ||
-                    formErrors?.errorType === StdcmConfigErrorTypes.INFRA_NOT_LOADED
+                    formErrors?.errorType === StdcmConfigErrorTypes.INFRA_NOT_LOADED ||
+                    !!consistErrors.totalMass ||
+                    !!consistErrors.totalLength ||
+                    !!consistErrors.maxSpeed
                   }
                 />
                 {formErrors && (
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
@@ -4,6 +4,11 @@ import { Input, ComboBox } from '@osrd-project/ui-core';
 import { useTranslation } from 'react-i18next';
 
 import useStdcmTowedRollingStock from 'applications/stdcm/hooks/useStdcmTowedRollingStock';
+import {
+  CONSIST_MAX_SPEED_MIN,
+  CONSIST_TOTAL_LENGTH_MAX,
+  CONSIST_TOTAL_MASS_MAX,
+} from 'applications/stdcm/utils/consistValidation';
 import type { LightRollingStockWithLiveries, TowedRollingStock } from 'common/api/osrdEditoastApi';
 import { useOsrdConfActions } from 'common/osrdContext';
 import SpeedLimitByTagSelector from 'common/SpeedLimitByTagSelector/SpeedLimitByTagSelector';
@@ -14,6 +19,7 @@ import useFilterRollingStock from 'modules/rollingStock/hooks/useFilterRollingSt
 import useFilterTowedRollingStock from 'modules/towedRollingStock/hooks/useFilterTowedRollingStock';
 import { type StdcmConfSliceActions } from 'reducers/osrdconf/stdcmConf';
 import { useAppDispatch } from 'store';
+import { kgToT, kmhToMs, msToKmh } from 'utils/physics';
 
 import StdcmCard from './StdcmCard';
 import useStdcmConsist from '../../hooks/useStdcmConsist';
@@ -33,7 +39,7 @@ const ConsistCardTitle = ({
   );
 };
 
-const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
+const StdcmConsist = ({ consistErrors = {}, disabled = false }: StdcmConfigCardProps) => {
   const { t } = useTranslation('stdcm');
   const { speedLimitByTag, speedLimitsByTags, dispatchUpdateSpeedLimitByTag } =
     useStoreDataForSpeedLimitByTagSelector({ isStdcm: true });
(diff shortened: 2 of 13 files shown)


front/src/utils/physics.ts
```diff
@@ -101,3 +101,11 @@ export function decimalToPercentage(value: number) {
 export function tToKg(value: number) {
   return value * 1000;
 }
+
+/**
+ * ex: converts 12000kg tp 12t
```
emersion: Typo: "to"

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
```diff
@@ -116,6 +123,28 @@ const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
...
+      totalLength
+    );
+
+    dispatch(
+      updateConsistErrors({
+        totalMass: totalMassError,
+        totalLength: totalLengthError,
+        maxSpeed: validateMaxSpeed(maxSpeed),
+      })
+    );
```
emersion: Do we really need to use the store for errors? Could we instead `useMemo()` and pass the errors object as a prop?

`StdcmConsist` is a direct child of `StdcmConfig`.

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
```diff
@@ -162,7 +191,13 @@ const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
           }}
         />
       </div>
-      <div className="stdcm-consist__properties">
+      <div
+        className={classNames({
+          'stdcm-consist__properties': !consistErrors.totalMass && !consistErrors.totalLength,
+          'stdcm-consist_with_errors__properties':
```
emersion: I'm not familiar with stdcm: what is this `__properties` suffix? Is it a pattern (and what is the reason for this pattern)? Or is it required by something?

front/src/applications/stdcm/utils/consistValidation.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  return undefined;
+};
+
+export const validateMaxSpeed = (maxSpeed?: number) => {
+  if (!maxSpeed) {
+    return undefined;
+  }
+
+  if (!maxSpeed || maxSpeed < 1) {
```
emersion: Nit: `!maxSpeed` is unnecessary because it's handled above already.

front/src/applications/stdcm/utils/consistValidation.ts
```diff
@@ -0,0 +1,66 @@
...
+export const validateTotalLength = (
+  tractionEngineLength?: number,
+  towedLength?: number,
+  totalLength?: number
+) => {
+  if (!totalLength) {
+    return undefined;
+  }
+
+  if (!totalLength || totalLength < 1) {
```
emersion: Ditto

emersion (commented):
Thanks for the patch! Here are a few comments.

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
Wadjetz: I think it's a CSS naming convention called BEM.

Wadjetz: yes, this new class do not follow the convention, i will rename it to "stdcm-consist__properties--with-errors"

clarani:
It would be nice to have the min length and weight of the rolling stock in the error text. What do you think ?

front/src/applications/stdcm/utils/consistValidation.ts
```diff
@@ -0,0 +1,66 @@
+import { kgToT } from 'utils/physics';
+
+export const validateTotalMass = (
+  tractionEngineMass?: number,
+  towedMass?: number,
```
SarahBellaha:
```suggestion
  towedMass: number = 0,
```

front/src/applications/stdcm/utils/consistValidation.ts
```diff
@@ -0,0 +1,66 @@
...
+  totalMass?: number
+) => {
+  if (!totalMass) {
```
SarahBellaha, line 16:
```suggestion
  if (!totalMass || !tractionEngineMass) {
```
I think we should also check that at the beginning, then you won't have to check it again or give it a default value :)

front/src/applications/stdcm/utils/consistValidation.ts
```diff
@@ -0,0 +1,66 @@
...
+    return undefined;
+  }
+
+  if (totalMass < 1) {
+    return 'consist.errors.totalMass.negative';
+  }
+
+  const tractionMassInTons = kgToT(tractionEngineMass || 0);
+  const consistMassInTons = kgToT((tractionEngineMass || 0) + (towedMass || 0));
```
SarahBellaha:
```suggestion
  const tractionMassInTons = kgToT(tractionEngineMass);
  const consistMassInTons = kgToT(tractionEngineMass + towedMass);
```

front/src/applications/stdcm/utils/consistValidation.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  if (towedMass && totalMass < consistMassInTons) {
+    return 'consist.errors.totalMass.lowerThanConsist';
+  }
+
+  return undefined;
+};
+
+export const validateTotalLength = (
```
SarahBellaha: We could apply the same changes as for "ValidateTotalMass"

front/src/applications/stdcm/utils/consistValidation.ts
Wadjetz: Shouldn’t this function still verify the value of the mass even if tractionEngineMass isn’t provided?

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
emersion: Sounds weird to use this convention here when the rest of OSRD doesn't use it. But it was there before your PR, so it's orthogonal.

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -71,10 +83,37 @@ const StdcmConfig = ({
...
+  const maxSpeed = useSelector(getMaxSpeed);
+
   const pathfinding = useStaticPathfinding(infra);
 
   const [formErrors, setFormErrors] = useState<StdcmConfigErrors>();
 
+  const { rollingStock } = useStoreDataForRollingStockSelector();
+  const towedRollingStock = useStdcmTowedRollingStock();
+
+  useEffect(() => {
```
emersion: This can be a `useMemo()` instead of a `useEffect()` + `useState()`.

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
```diff
@@ -1,9 +1,15 @@
 import { useEffect } from 'react';
 
 import { Input, ComboBox } from '@osrd-project/ui-core';
+import classNames from 'classnames';
```
emersion: Nit: everywhere else in the app, we import this as `cx`

front/src/applications/stdcm/utils/consistValidation.ts
```diff
@@ -0,0 +1,88 @@
...
+  if (totalMass < massLimit || totalMass >= CONSIST_TOTAL_MASS_MAX) {
+    return 'consist.errors.totalMass.range';
+  }
+
+  if (
+    totalMass < (towedMass ? consistMassInTons : tractionMassInTons) ||
+    totalMass >= CONSIST_TOTAL_MASS_MAX
+  ) {
+    return 'consist.errors.totalMass.range';
+  }
```
emersion: Sounds like this is a duplicate check? The `if` right above seems to check for the same thing.

front/src/applications/stdcm/utils/consistValidation.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  if (totalLength <= 0) {
+    return 'consist.errors.totalLength.negative';
+  }
+
+  const consistLength = tractionEngineLength + towedLength;
+
+  if (
+    totalLength < Math.max(tractionEngineLength, consistLength) ||
```
emersion: Can `tractionEngineLength` ever be lower than `consistLength`? I think these are all positive numbers right?

front/src/applications/stdcm/utils/consistValidation.ts
Wadjetz: It check the lower and higher bounds

emersion (commented):
Code LGTM, but after asking Thibaut it seems like it's undesirable to change the form layout when a field is invalid. A new mockup will be posted soon.

(For reference, the old mockup was: https://www.sketch.com/s/15e0909c-f17d-4200-a770-b9fc1154fa7e/a/Go2QJL7)

front/package.json
```diff
@@ -8,12 +8,12 @@
...
-    "@osrd-project/ui-manchette": "^0.0.57",
-    "@osrd-project/ui-manchette-with-spacetimechart": "^0.0.57",
-    "@osrd-project/ui-spacetimechart": "^0.0.57",
-    "@osrd-project/ui-speedspacechart": "^0.0.57",
+    "@osrd-project/ui-core": "^0.0.59",
+    "@osrd-project/ui-icons": "^0.0.59",
+    "@osrd-project/ui-manchette": "^0.0.59",
+    "@osrd-project/ui-manchette-with-spacetimechart": "^0.0.59",
+    "@osrd-project/ui-spacetimechart": "^0.0.59",
+    "@osrd-project/ui-speedspacechart": "^0.0.59",
```
emersion: Note, this version bump is done in https://github.com/OpenRailAssociation/osrd/pull/10233.

front/src/reducers/osrdconf/types.ts
```diff
@@ -47,6 +47,11 @@ export type OsrdStdcmConfState = OsrdConfState & {
   maxSpeed?: number;
   towedRollingStockID?: number;
   linkedTrains: LinkedTrains;
+  consistErrors: {
+    totalMass?: string;
+    totalLength?: string;
+    maxSpeed?: string;
+  };
```
emersion: I don't think we need these?

front/package.json
SharglutDev: Yep, he just needed the changes in v59 to continue this PR :)

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -39,6 +39,7 @@ export const stdcmConfInitialState: OsrdStdcmConfState = {
   totalLength: undefined,
   maxSpeed: undefined,
   towedRollingStockID: undefined,
+  consistErrors: {},
```
emersion: Ah, this can be dropped as well I believe.

front/src/styles/scss/applications/stdcm/_card.scss
```diff
@@ -89,7 +92,9 @@
     }
 
     .feed-back {
-      padding: 0;
+      --field-wrapper-padding-bottom: 9px;
```
emersion, line 95: Why do we need to override the osrd-ui styles here?

Wadjetz: It because we modified the input paddings in this form, and the tooltip offset is calculated based on the padding values.

front/package.json
SharglutDev: v59 has been merged. You can remove these modifications and rebase :)

front/src/styles/scss/applications/stdcm/_card.scss
emersion: Alright. I'm not in love with the way this is done, but I've dragged this PR for too long already, so let's leave it as-is.

emersion (approved):

Maymanaf (approved):

PR #9580, opened 2024-11-05
editoast: add CLI to import towed rolling stock

### LINKED ISSUE #8668: Implement the import of RollstocksJson into editoast
With the new format of RollstocksJson, `editoast` needs to evolve to import the new format.

- modify the CLI tool to be able to read the new RollstocksJson format
- populate correctly the database with all the new information
  - the new table for towed rolling stocks
  - the new flag "multiple unit" in the existing rolling stock table

woshilapin:
Closes #8668 

This new CLI brings a simple way to import Towed rolling stock in the DB from JSON files, without the need to start the `editoast` web service.

### DIFF
--- a/editoast/src/client/import_rolling_stock.rs
+++ b/editoast/src/client/import_rolling_stock.rs
@@ -1,14 +1,20 @@
+use std::error::Error;
+use std::fs::File;
+use std::io::BufReader;
 use std::path::PathBuf;
-use std::{error::Error, fs::File, io::BufReader, sync::Arc};
+use std::sync::Arc;
 
 use clap::Args;
 use colored::Colorize as _;
 use editoast_models::DbConnectionPoolV2;
 use editoast_schemas::rolling_stock::RollingStock;
+use editoast_schemas::rolling_stock::TowedRollingStock;
 use validator::ValidationErrorsKind;
 
 use crate::models::prelude::*;
-use crate::{models::RollingStockModel, CliError};
+use crate::models::towed_rolling_stock::TowedRollingStockModel;
+use crate::models::RollingStockModel;
+use crate::CliError;
 
 #[derive(Args, Debug)]
 #[command(about, long_about = "Import a rolling stock given a json file")]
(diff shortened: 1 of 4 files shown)


editoast/src/client/import_rolling_stock.rs
```diff
@@ -70,162 +76,251 @@ pub async fn import_rolling_stock(
     Ok(())
 }
 
+pub async fn import_towed_rolling_stock(
+    args: ImportRollingStockArgs,
+    db_pool: Arc<DbConnectionPoolV2>,
+) -> Result<(), Box<dyn Error + Send + Sync>> {
+    for towed_rolling_stock_path in args.rolling_stock_path {
+        let towed_rolling_stock_file = File::open(towed_rolling_stock_path)?;
+        let towed_rolling_stock_form: TowedRollingStock =
```
leovalais:
```suggestion
        let towed_rolling_stock: TowedRollingStock =
```

We conventionally use the `-form` suffix to represent what the frontend sends in the http api. Since we're using the schema directly here (rightfully so), the suffix may be misleading, wdyt?

editoast/src/client/import_rolling_stock.rs
```diff
@@ -70,162 +76,251 @@ pub async fn import_rolling_stock(
...
-        let RollingStockModel {
-            electrical_power_startup_time,
-            raise_pantograph_time,
-            ..
-        } = created_rs;
-        assert!(electrical_power_startup_time.is_some());
-        assert!(raise_pantograph_time.is_some());
+            // THEN
+            assert!(result.is_ok());
+            use crate::models::Retrieve;
```
leovalais: I think all `use ..::Retrieve` are unnecessary since the `prelude` is already imported at top-level and you `use super::*` in every test module.

leovalais (approved):
LGTM, thanks!

editoast/src/client/import_rolling_stock.rs
woshilapin: Done in [bfa4ce5](https://github.com/OpenRailAssociation/osrd/pull/9580/commits/bfa4ce50e951609623abde6fe570b979a579c0a9).

editoast/src/client/import_rolling_stock.rs
woshilapin: Done in [f582efe](https://github.com/OpenRailAssociation/osrd/pull/9580/commits/f582efe1d9ae87256116f815baf1b6b94455ff6c).

PR #9590, opened 2024-11-06
scripts: make all scripts that need it use realpath

Khoyo:


emersion:
There are two more I think:

```
scripts/cleanup-db.sh:80:echo "'diesel migration run --migration-dir $(dirname $0)/../editoast/migrations'  # 'docker compose up editoast' does it automatically"
scripts/load-backup.sh:47:$(dirname "$0")/cleanup-db.sh # Cleanup and init db (no migration)
```

Khoyo:
You're right, I'm reversing this PR to match the scripts popular vote

emersion:
The upside of this approach (grabbing the script's directory) is that scripts can be run from anywhere and still work fine. (It's still a bit fragile because `$0` may not be a path, e.g. because of `$PATH` lookup or passing a different `argv[0]` to `exec(3)`, but that's pretty unlikely to matter in practice.)

Khoyo:
> It's still a bit fragile because $0 may not be a path, e.g. because of $PATH lookup or passing a different argv[0] to exec(3), but that's pretty unlikely to matter in practice

Indeed which is why I naturally tend to err on the other side. But the most important part imo is to be consistent.

I'm a bit conflicted about the various compose scripts. Should they keep the docker-compose behavior of using the working directory (or any parent), or should they support being run from anywhere?

ghost (approved):

PR #9591, opened 2024-11-06
core: multithread core workers

### LINKED ISSUE #9572: core: when running as worker, we only use a single thread (per worker)
The logs mention 16(?) threads when starting up, but we only use one.

To test this, the simplest way to test it is to add a `Thread.sleep` when processing messages (with lots of messages in the queue). It's easy to tell that the sleep freezes the whole thing, it doesn't process other messages in the meantime.

Ideally, we should enable proper multithreading with ways to configure it. It would be much faster when running the app locally, while still having a single infra loaded in memory. 

Note: we *may* run into issues where several messages initiate an infra loading at the same time. `InfraManager` is supposed to be robust to this case though.

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/9572

Configurable with a `WORKER_THREADS=42` env variable, with a default set to the number of processors.

### DIFF
--- a/core/src/main/java/fr/sncf/osrd/cli/WorkerCommand.kt
+++ b/core/src/main/java/fr/sncf/osrd/cli/WorkerCommand.kt
@@ -3,6 +3,11 @@ package fr.sncf.osrd.cli
 import com.beust.jcommander.Parameter
 import com.beust.jcommander.Parameters
 import com.rabbitmq.client.*
+import com.rabbitmq.client.AMQP
+import com.rabbitmq.client.Channel
+import com.rabbitmq.client.ConnectionFactory
+import com.rabbitmq.client.DeliverCallback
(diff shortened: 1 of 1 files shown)


Khoyo:
Does this do parallel consume? Or does this launch a single task on the threadpool (and we would need to actually call consume multiple times on separate channels?)

eckter:
> Does this do parallel consume? Or does this launch a single task on the threadpool (and we would need to actually call consume multiple times on separate channels?)

The "consume" is only done on a single thread, but the processing itself is dispatched to the executor.

The alternative would be to create several channels and call several consumes, but I struggled to make it work and to have it actually work in parallel. (But I can try again later if the current solution bothers anyone)

Khoyo (approved):

Erashin (approved):

PR #9592, opened 2024-11-06
front: fix last waypoints hidden with warning

SharglutDev:
Add a padding-bottom to the waypoints panel body when the warning message is displayed, allowing the last waypoints from the manchette not to be hidden by it

fix last bug mentionned here https://github.com/OpenRailAssociation/osrd/issues/8628#issuecomment-2435259268

To display the warning message, there need to be less than 2 waypoints checked in the panel


kmer2016 (commented):
LGTM

kmer2016 (approved):

PR #9593, opened 2024-11-06
front: fix capitalized text in stdcm table results

### LINKED ISSUE #9525: front: typo in simulation result
### What happened?

Image
"De" and "Réf" should not be capitalized (see mock-up)

_No response_

1. go to STDCM
2. launch a simulation
3. go to the simulation result table

Recette (SNCF)

Firefox

86ad640

Akctarus:
closes #9525

### DIFF
--- a/front/src/styles/scss/applications/stdcm/_results.scss
+++ b/front/src/styles/scss/applications/stdcm/_results.scss
@@ -159,6 +159,9 @@
             padding-bottom: 0.313rem;
             vertical-align: middle;
           }
+          th:last-child {
+            text-transform: none;
+          }
           tbody tr:nth-child(odd) {
             background-color: rgb(239, 243, 245);
           }
--- a/front/public/locales/en/stdcm-simulation-report-sheet.json
+++ b/front/public/locales/en/stdcm-simulation-report-sheet.json
@@ -15,7 +15,7 @@
   "maxWeight": "max. weight",
   "operationalPoint": "OP",
   "passageStop": "passage",
-  "refEngine": "ref. engine",
+  "refEngine": "Ref. engine",
   "referenceEngine": "reference engine",
   "referencePath": "Reference path",
   "requestedRoute": "requested route",
--- a/front/public/locales/fr/stdcm-simulation-report-sheet.json
+++ b/front/public/locales/fr/stdcm-simulation-report-sheet.json
@@ -15,7 +15,7 @@
   "maxWeight": "tonnage max.",
   "operationalPoint": "jalon",
   "passageStop": "passage",
-  "refEngine": "engin de réf.",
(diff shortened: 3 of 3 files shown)


SharglutDev (approved):
Lgtm and tested

SarahBellaha (approved):
LGTM

front/src/applications/stdcm/components/StdcmResults/StdcmResultsTable.tsx
```diff
@@ -43,7 +44,7 @@ const StcdmResultsTable = ({
             <th>{t('passageStop')}</th>
             <th>{t('startStop')}</th>
             <th className="weight">{t('weight')}</th>
-            <th>{t('refEngine')}</th>
+            <th>{capitalizeFirstLetter(t('refEngine'))}</th>
```
emersion: Instead of post-processing the translated strings, could we update the strings in the translation files instead?

Akctarus: since it only used once, yes we can

front/src/styles/scss/applications/stdcm/_results.scss
```diff
@@ -159,6 +159,9 @@
             padding-bottom: 0.313rem;
             vertical-align: middle;
           }
+          th:last-child {
+            text-transform: none;
```
emersion, line 163: I'd prefer not to have a special case for the only header which happens to have multiple words: in the future, if one of the other translations get updated with a string which contains multiple words, `text-transform: capitalize` will not do what we want.

So maybe we can drop `text-transform: capitalize` and always just capitalize everything the way we want in the translation files?

Akctarus: We use these translations in another file and we need them to be in lowercase there :(

emersion: Maybe we can use `text-transform: lowercase` for the other place where the translations are used?

It's always easier to go from capitalized to full lowercase then the reverse.

Akctarus: This will be done in a future PR refactor of the simulation sheet

front/src/applications/stdcm/components/StdcmResults/StdcmResultsTable.tsx
Akctarus: fixed

front/src/styles/scss/applications/stdcm/_results.scss
emersion: Alright!

emersion (approved):

PR #9598, opened 2024-11-06
front: fix decimals error for study budget

### LINKED ISSUE #8607: Selecting float budget gives unreadable error
### What happened?

https://github.com/user-attachments/assets/0509e2b0-7616-4a73-89c8-2ed5c4640010

I expect the field to be highlighted in red, with a clear message stating that float numbers are forbidden

1. Go to operational studies page
2. Open project creation form
3. Enter the name and a floating point budget
4. Click on save, witness

Recette (SNCF)

Firefox

 1482e8c

Akctarus:
closes #8607

### DIFF
--- a/front/public/locales/en/operationalStudies/study.json
+++ b/front/public/locales/en/operationalStudies/study.json
@@ -46,7 +46,7 @@
   "confirmDeleteMessage_one": "Do you really want to delete {{count}} scenario?",
   "confirmDeleteMessage_other": "Do you really want to delete the {{count}} scenarios?",
   "studyBudget": "Budget",
-  "studyBudgetInvalid": "The budget for the study must not exceed 2147483647",
+  "studyBudgetInvalid": "The study budget must be a positive integer and must not exceed 2147483647 €",
   "studyBusinessCode": "Business code",
   "studyBusinessCodeInvalid": "The business code for the study must not exceed 128 characters",
   "studyCancel": "Cancel",
--- a/front/public/locales/fr/operationalStudies/study.json
+++ b/front/public/locales/fr/operationalStudies/study.json
@@ -45,7 +45,7 @@
     "byRecentDate": "Plus récents"
   },
   "studyBudget": "Budget",
-  "studyBudgetInvalid": "Le budget de l'étude ne doit pas dépasser 2147483647",
+  "studyBudgetInvalid": "Le budget de l'étude doit être un entier positif et ne doit pas dépasser 2147483647 €",
   "studyBusinessCode": "Code affaire",
   "studyBusinessCodeInvalid": "Le code affaire de l'étude ne doit pas dépasser 128 caractères",
   "studyCancel": "Annuler",
--- a/front/src/modules/study/utils.ts
+++ b/front/src/modules/study/utils.ts
@@ -35,5 +35,6 @@ export const checkStudyFields = (
   description: isInvalidString(SMALL_TEXT_AREA_MAX_LENGTH, study.description),
   business_code: isInvalidString(SMALL_INPUT_MAX_LENGTH, study.business_code),
   service_code: isInvalidString(SMALL_INPUT_MAX_LENGTH, study.service_code),
-  budget: (study.budget ?? 0) > 2147483647,
+  budget:
+    (study.budget ?? 0) > 2147483647 || (!Number.isInteger(study.budget) && study.budget !== null),
 });


front/src/modules/study/components/AddOrEditStudyModal.tsx
```diff
@@ -413,7 +413,7 @@ const AddOrEditStudyModal = ({ editionMode, study, scenarios }: AddOrEditStudyMo
           <div className="col-lg-4">
             <InputSNCF
               id="studyInputBudget"
-              type="number"
+              type="text"
```
emersion: Can we keep `number` still? This is useful for accessibility, on-screen keyboards, etc.

front/src/modules/study/components/AddOrEditStudyModal.tsx
```diff
@@ -432,12 +432,10 @@ const AddOrEditStudyModal = ({ editionMode, study, scenarios }: AddOrEditStudyMo
...
                   : ''
               }
-              onChange={(e) =>
-                handleStudyInputChange(
-                  'budget',
-                  e.target.value !== '' && +e.target.value >= 0 ? +e.target.value : null
-                )
-              }
+              onChange={(e) => {
+                const numericValue = e.target.value.replace(/\D/g, '');
```
emersion: Seems like this ignores any non-digit character? Might be surprising to the user that part of what they typed is just discarded?

front/src/modules/study/components/AddOrEditStudyModal.tsx
Akctarus: If we keep `number`, when we can still type letters and special characters, and if we save it, we have no budget.

emersion: But with `text` the user can also type letters and special characters? So we can keep `number`, and add some additional validation?

front/src/modules/study/components/AddOrEditStudyModal.tsx
Akctarus: fixed

front/src/modules/study/components/AddOrEditStudyModal.tsx
Akctarus: fixed

emersion:
This looks much better! I have one remaining question: we already have some validation logic in `checkStudyFields()`. I think it would make sense to have a single spot where all of the validation is performed, instead of having half of it in one place and the other somewhere else?

One approach would be to add an extra condition to `checkStudyFields()` and change the existing error message to "not an integer between 0 and $max". Another approach would be to return an error string (or translation key) from `checkStudyFields()` instead of a `boolean`.

front/public/locales/en/operationalStudies/study.json
```diff
@@ -47,6 +47,7 @@
   "confirmDeleteMessage_other": "Do you really want to delete the {{count}} scenarios?",
   "studyBudget": "Budget",
   "studyBudgetInvalid": "The budget for the study must not exceed 2147483647",
+  "studyBudgetNoDecimal": "The budget for the study must not be a decimal number",
```
emersion: This error message is a bit weird, because integers are decimal numbers. IOW, "42" is as good a decimal number as "-3.14".

Did you mean "fractional"? But then this doesn't convey that negative numbers are invalid.

I think "must be a positive integer" could do the trick?

Akctarus: Thank for your comments, fixed :)

emersion:
Seems like there is a duplicate PR? https://github.com/OpenRailAssociation/osrd/pull/9663 seems identical.

front/public/locales/en/operationalStudies/study.json
```diff
@@ -46,7 +46,7 @@
   "confirmDeleteMessage_one": "Do you really want to delete {{count}} scenario?",
   "confirmDeleteMessage_other": "Do you really want to delete the {{count}} scenarios?",
   "studyBudget": "Budget",
-  "studyBudgetInvalid": "The budget for the study must not exceed 2147483647",
+  "studyBudgetInvalid": "The study budget must be a positive integer, and must not exceed 2147483647 €.",
```
theocrsb:
```suggestion
  "studyBudgetInvalid": "The study budget must be a positive integer and must not exceed 2147483647 €",
```

front/public/locales/fr/operationalStudies/study.json
```diff
@@ -45,7 +45,7 @@
     "byRecentDate": "Plus récents"
   },
   "studyBudget": "Budget",
-  "studyBudgetInvalid": "Le budget de l'étude ne doit pas dépasser 2147483647",
+  "studyBudgetInvalid": "Le budget de l'étude doit être un entier positif, et ne doit pas dépasser 2147483647 €",
```
theocrsb:
```suggestion
  "studyBudgetInvalid": "Le budget de l'étude doit être un entier positif et ne doit pas dépasser 2147483647 €",
```

theocrsb (approved):
LGTM and tested :)

Akctarus:
> Seems like there is a duplicate PR? #9663 seems identical.

#9663 is for project budget, this one is for study budget :)

emersion (approved):
My bad! LGTM :)

PR #9603, opened 2024-11-06
docker: actually remove the port mapping in host mode

Khoyo:


emersion (approved):
LGTM

(For future reference, docs for this: https://docs.docker.com/reference/compose-file/merge/#reset-value)

ghost (approved):
Thanks ! 
Could you also add a commit to remove all `version: '3'` from the docker compose files since its deprecated in this PR ? (It's just, I'm thinking about doing a PR specifically for this for some time and it can fit this PR).

PR #9605, opened 2024-11-06
Isolate runserver CLI in client module and move AppState into views

leovalais:
Follow up of #9349

Moves in the `client` module the following commands:

- health check
- openapi generation
- runserver

Reaching a proper split (import wise) of the run server command required a bit of refactoring. We want `client` to depend of `views`, and not the other way around. So we cannot use `client` definitions anywhere in `views`.

What's done to achieve that:

- Introduce a `ServerConfig` structure that aggregates all server parameters.
- Move `AppState` into `views` (still re-exported at crate-level, for now at least)
- Add the config to the state.
- Get rid of most occurrences of `client` definitions usage outside of the module. `MapLayerConfig` has been removed from the state, `ValkeyConfig` moved in `mod valkey_utils`, `PostgresConfig` and `CoreConfig` duplicated in `mod views`.

What's been done simultaneously:

- Use the type `url::Url` instead of `String` where it makes sense 
- Remove unused CLI parameters
- Remove useless `Default` implementations in `client`
- Restrict `client` structs visibility where possible.

# Next steps

1. Move the big `match` in the `run` function in the `client` module
2. Expose `ROOT-URL` and `DYNAMIC_ASSET_PATH` in `clap` and remove their accessor functions from `client`
3. Move `get_app_version` in `editoast_common`
4. Move `valkey_utils` in `editoast_common` (wdyt?)
5. Remove `db_pool_v1` from the `AppState`
7. Replace most `return Box::new(CliError::new(...` by `anyhow::bail!` since `anyhow`'s paradigm fits perfectly our error management on the `client` side
8. Use the `AppState.config.disable_authorization` and remove `AppState.disable_authorization`
9. The `editoast_client` crate is ready to be split!

(Some of these points could be included in the current PR, but I figured it's big enough already...)

### DIFF
--- a/editoast/src/map/layers.rs
+++ b/editoast/src/map/layers.rs
@@ -1,7 +1,6 @@
 use std::collections::HashMap;
 
 use serde::Deserialize;
-use serde::Serialize;
 
 // select C.stuff from A inner join B C on C.id = C.id;
 //                       \___________________________/
@@ -10,7 +9,7 @@ use serde::Serialize;
 type JoinExpr = String;
 
 /// Layer view description
-#[derive(Clone, Serialize, Deserialize, PartialEq, Debug)]
+#[derive(Debug, Deserialize)]
 pub struct View {
     pub on_field: String,
     pub data_expr: String,
@@ -23,7 +22,7 @@ pub struct View {
 }
 
 /// Layer description
-#[derive(Clone, Serialize, Deserialize, PartialEq, Debug)]
+#[derive(Debug, Deserialize)]
 pub struct Layer {
     pub table_name: String,
     pub views: HashMap<String, View>,
@@ -33,14 +32,20 @@ pub struct Layer {
     pub attribution: Option<String>,
 }
 
-#[derive(Serialize, Deserialize, PartialEq, Debug)]
+#[derive(Debug, Deserialize)]
 pub struct MapLayers {
     pub layers: HashMap<String, Layer>,
 }
 
 impl MapLayers {
-    /// Parses file containing layers' description into MapLayers struct
-    pub fn parse() -> MapLayers {
-        serde_yaml::from_str(include_str!("../../map_layers.yml")).unwrap()
+    pub fn new(layers: HashMap<String, Layer>) -> Self {
+        Self { layers }
+    }
+}
+
+impl Default for MapLayers {
+    fn default() -> Self {
+        serde_yaml::from_str(include_str!("../../map_layers.yml"))
+            .expect("static data should be valid")
     }
 }
--- a/editoast/src/valkey_utils.rs
+++ b/editoast/src/valkey_utils.rs
@@ -17,8 +17,8 @@ use redis::ToRedisArgs;
 use serde::de::DeserializeOwned;
 use serde::Serialize;
 use tracing::trace;
+use url::Url;
 
-use crate::client::ValkeyConfig;
 use crate::error::Result;
 
 pub enum ValkeyConnection {
@@ -202,19 +202,26 @@ pub enum ValkeyClient {
     NoCache,
 }
 
+#[derive(Clone)]
+pub struct ValkeyConfig {
+    /// Disables caching. This should not be used in production.
+    pub no_cache: bool,
+    pub is_cluster_client: bool,
+    pub valkey_url: Url,
+}
+
 impl ValkeyClient {
     pub fn new(valkey_config: ValkeyConfig) -> Result<ValkeyClient> {
         if valkey_config.no_cache {
             return Ok(ValkeyClient::NoCache);
         }
-        let valkey_config_url = valkey_config.url()?;
         if valkey_config.is_cluster_client {
             return Ok(ValkeyClient::Cluster(
-                redis::cluster::ClusterClient::new(vec![valkey_config_url.as_str()]).unwrap(),
+                redis::cluster::ClusterClient::new(vec![valkey_config.valkey_url]).unwrap(),
             ));
         }
         Ok(ValkeyClient::Tokio(
-            redis::Client::open(valkey_config_url.as_str()).unwrap(),
+            redis::Client::open(valkey_config.valkey_url).unwrap(),
         ))
     }
 
--- a/editoast/src/valkey_utils.rs
+++ b/editoast/src/valkey_utils.rs
@@ -155,7 +155,7 @@ impl ValkeyConnection {
             Err(_) => {
                 return Err(RedisError::from((
                     ErrorKind::IoError,
-                    "An error occured serializing to json",
+                    "An error occurred serializing to json",
                 ))
                 .into())
             }
@@ -182,7 +182,7 @@ impl ValkeyConnection {
                     .map_err(|_| {
                         RedisError::from((
                             ErrorKind::IoError,
-                            "An error occured serializing to json",
+                            "An error occurred serializing to json",
                         ))
                         .into()
                     })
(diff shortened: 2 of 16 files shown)


leovalais:
To merge after https://github.com/OpenRailAssociation/osrd/pull/9580, it shouldn't conflict, but just to be sure.

Khoyo (commented):
I have to ask... Why is it the *client* module if it contains the runserver cli?

leovalais:
@Khoyo We only have one binary for now. And it likely won't change until we have a proper split in crates. In which case it will be trivial to move the `runserver` CLI implementation into another bin.

The biggest part of this PR is actually to remove the dependency of `client` in `views`, which is plainly wrong according to our split, whether we have 1, 2, or more binaries.

editoast/src/valkey_utils.rs
```diff
@@ -202,19 +202,26 @@ pub enum ValkeyClient {
...
-    pub fn new(valkey_config: ValkeyConfig) -> Result<ValkeyClient> {
+    pub fn new(valkey_config: &ValkeyConfig) -> Result<ValkeyClient> {
         if valkey_config.no_cache {
             return Ok(ValkeyClient::NoCache);
         }
-        let valkey_config_url = valkey_config.url()?;
         if valkey_config.is_cluster_client {
             return Ok(ValkeyClient::Cluster(
-                redis::cluster::ClusterClient::new(vec![valkey_config_url.as_str()]).unwrap(),
+                redis::cluster::ClusterClient::new(vec![valkey_config.valkey_url.as_str()])
```
woshilapin: If I understand well, `ClusterClient::new()` takes an iterator on anything that implements [`IntoConnectionInfo`](https://docs.rs/redis/latest/redis/trait.IntoConnectionInfo.html), which is the case for `Url` so I do not think you need the `.as_str()`, isn't it?

```suggestion
                redis::cluster::ClusterClient::new(vec![valkey_config.valkey_url])
```

editoast/src/valkey_utils.rs
```diff
@@ -202,19 +202,26 @@ pub enum ValkeyClient {
...
+                    .unwrap(),
             ));
         }
         Ok(ValkeyClient::Tokio(
-            redis::Client::open(valkey_config_url.as_str()).unwrap(),
+            redis::Client::open(valkey_config.valkey_url.as_str()).unwrap(),
```
woshilapin: See comment above.

```suggestion
            redis::Client::open(valkey_config.valkey_url).unwrap(),
```

editoast/src/valkey_utils.rs
```diff
@@ -202,19 +202,26 @@ pub enum ValkeyClient {
...
+pub struct ValkeyConfig {
+    /// Disables caching. This should not be used in production.
+    pub no_cache: bool,
+    pub is_cluster_client: bool,
+    pub valkey_url: Url,
+}
+
 impl ValkeyClient {
-    pub fn new(valkey_config: ValkeyConfig) -> Result<ValkeyClient> {
+    pub fn new(valkey_config: &ValkeyConfig) -> Result<ValkeyClient> {
```
woshilapin: Seems weird to not consume the configuration for creating the client. I believe the only purpose of the configuration is to create the client. Any reason you changed this?

editoast/src/map/layers.rs
```diff
@@ -33,14 +33,36 @@ pub struct Layer {
...
+
+/// Parses file containing layers' description
+fn parse() -> HashMap<String, Layer> {
+    #[derive(Deserialize)]
+    struct Layers {
+        layers: HashMap<String, Layer>,
+    }
+
+    let Layers { layers } = serde_yaml::from_str(include_str!("../../map_layers.yml"))
+        .expect("static data should be valid");
```
woshilapin: This doesn't work?

```suggestion
    let layers: HashMap<String, Layer> = serde_yaml::from_str(include_str!("../../map_layers.yml"))
        .expect("static data should be valid");
```

woshilapin (approved):
Looks fine, just a few minor comments.

editoast/src/map/layers.rs
leovalais: Sadly no:

```yaml
layers:
  track_sections:
    table_name: infra_layer_track_section
    id_field: id
    views:
      geo:
        on_field: geographic
  ...
```

I can remove the indirection in the YAML though, since `layer` is the only prop at top-level.

editoast/src/valkey_utils.rs
leovalais: I made it a reference because there was no reason to consume the config since none of its fields needed to be moved. I removed the `as_str` as suggested, which consumes the URL, so I removed the reference as well. Cf. [`3894923` (#9605)](https://github.com/OpenRailAssociation/osrd/pull/9605/commits/38949235d3746a6fae496d37ddfac81a98e89f34)

editoast/src/valkey_utils.rs
leovalais: Done in [`3894923` (#9605)](https://github.com/OpenRailAssociation/osrd/pull/9605/commits/38949235d3746a6fae496d37ddfac81a98e89f34)

editoast/src/valkey_utils.rs
leovalais: Done in [`3894923` (#9605)](https://github.com/OpenRailAssociation/osrd/pull/9605/commits/38949235d3746a6fae496d37ddfac81a98e89f34)

Khoyo (commented):
Re: client, could we call it *command*? I really don't see how it is a client

Otherwise, LGTM. (But I only did a cursory review)

leovalais:
Agreed on the naming. It never really was a true "client" since it just runs commands on its own (no command forge a request that is sent to an editoast server). `editoast_commands` or `editoast_cli` would be better crate names indeed. Let's keep that in mind when we'll create it.

editoast/src/map/layers.rs
woshilapin: ok, don't bother, I didn't have the entire context and I understand now.

editoast/src/valkey_utils.rs
woshilapin: Mmm, almost, there were 2 occurrences of this and you fixed the other one 😅

editoast/src/valkey_utils.rs
woshilapin: My remark was more on the side "yes, we technically don't need to consume it... but are we going to do with it after that anyway?". So it's a more logical reasoning than a technical one. Anyway, thanks for making the change 😄

editoast/src/valkey_utils.rs
leovalais: My bad... done in [`c85db53` (#9605)](https://github.com/OpenRailAssociation/osrd/pull/9605/commits/c85db53ca27f2ce0465b8620690abef45fe29771) (hopefully this time xD)

leovalais:
@woshilapin I rolled back the changes to the structure `MapLayers` since the `max_zoom` config is already available to each handler via the `AppState.config`. Diff in [`85a983d` (#9605)](https://github.com/OpenRailAssociation/osrd/pull/9605/commits/85a983d2c96dd354e7280d26d196ddd82bff9b64). Let me know if that's ok so that I cleanup the git history a bit before merging.

woshilapin:
Looks good, even for the rolled back on `MapLayers`. You'll need to rebase because of [this PR](https://github.com/OpenRailAssociation/osrd/pull/9658).

PR #9614, opened 2024-11-07
front: add track column to timestops table

### LINKED ISSUE #9526: Add a "Track" column to schedule tables

End users

Add a column to the tables (input and output) such that:

- [x] It is named "Track" in English and "Voie" in French
- [x] It is the third column in the table, right after the CH code of the operational points (or requested points)
- [x] For each row in a schedule table (input and output), it is filled with the name of the track the train is on when at this operational point
- [x] The width of the column is fixed (and thus, it takes space on the name of the operational points)

This feature matters because it:
- gives more information to the users about their schedule

The algorithm to get the name of the track can probably be scavenged from the "fiche train" component which is at the bottom of the scenario detail page today, and which displays a column with the data described above.

Synar:
Close #9526

### DIFF
--- a/front/src/applications/operationalStudies/hooks/useCachedTrackSections.ts
+++ b/front/src/applications/operationalStudies/hooks/useCachedTrackSections.ts
@@ -1,0 +1,41 @@
+import { useRef, useCallback } from 'react';
+
+import { uniq } from 'lodash';
+
+import { osrdEditoastApi } from 'common/api/osrdEditoastApi';
+import type { TrackSection } from 'common/api/osrdEditoastApi';
+
+export default function useCachedTrackSections(infraId: number) {
+  const trackIdsRef = useRef<Set<string>>(new Set());
+  const trackSectionsRef = useRef<Record<string, TrackSection>>({});
+  const [loadInfraObject, { isLoading }] =
+    osrdEditoastApi.endpoints.postInfraByInfraIdObjectsAndObjectType.useMutation();
+
+  const getTrackSectionsByIds = useCallback(
+    async (requestedTrackIds: string[]) => {
+      const uniqueNewIds = uniq(requestedTrackIds.filter((id) => !trackIdsRef.current.has(id)));
+      if (uniqueNewIds.length !== 0) {
+        try {
+          const fetchedSections = await loadInfraObject({
+            infraId,
+            objectType: 'TrackSection',
+            body: uniqueNewIds,
+          }).unwrap();
+
+          uniqueNewIds.forEach((id) => trackIdsRef.current.add(id));
+          fetchedSections.forEach((rawSection) => {
+            const trackSection = rawSection.railjson as TrackSection;
+            trackSectionsRef.current[trackSection.id] = trackSection;
+          });
+        } catch (error) {
+          console.error('Failed to fetch track sections:', error);
+        }
+      }
+
+      return trackSectionsRef.current;
+    },
+    [infraId]
+  );
+
+  return { getTrackSectionsByIds, isLoading };
+}
--- a/front/src/applications/operationalStudies/hooks/useScenarioContext.tsx
+++ b/front/src/applications/operationalStudies/hooks/useScenarioContext.tsx
@@ -1,0 +1,35 @@
+import { createContext, useContext, useMemo, type ReactNode } from 'react';
+
+import useCachedTrackSections from 'applications/operationalStudies/hooks/useCachedTrackSections';
+import type { TrackSection } from 'common/api/osrdEditoastApi';
+
+type ScenarioContextType = {
+  getTrackSectionsByIds: (requestedTrackIds: string[]) => Promise<Record<string, TrackSection>>;
+  infraId: number;
+  trackSectionsLoading: boolean;
+} | null;
+const ScenarioContext = createContext<ScenarioContextType>(null);
+
+type ScenarioContextProviderProps = { infraId: number; children: ReactNode };
+
+export const ScenarioContextProvider = ({ infraId, children }: ScenarioContextProviderProps) => {
+  const { getTrackSectionsByIds, isLoading: trackSectionsLoading } =
+    useCachedTrackSections(infraId);
+  const providedContext = useMemo(
+    () => ({
+      getTrackSectionsByIds,
+      infraId,
+      trackSectionsLoading,
+    }),
+    [getTrackSectionsByIds, infraId, trackSectionsLoading]
+  );
+  return <ScenarioContext.Provider value={providedContext}>{children}</ScenarioContext.Provider>;
+};
+
+export const useScenarioContext = () => {
+  const context = useContext(ScenarioContext);
+  if (!context) {
+    throw new Error('useScenarioContext must be used within a ScenarioContextProvider');
+  }
+  return context;
+};
--- a/front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
+++ b/front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
@@ -2,7 +2,6 @@ import { useEffect } from 'react';
 
 import { type Position } from '@turf/helpers';
 import { omit } from 'lodash';
-import { useSelector } from 'react-redux';
 
 import {
   osrdEditoastApi,
@@ -13,9 +12,9 @@ import {
   type TrainScheduleResult,
   type PathfindingResult,
 } from 'common/api/osrdEditoastApi';
-import { useOsrdConfActions, useOsrdConfSelectors } from 'common/osrdContext';
+import { useOsrdConfActions } from 'common/osrdContext';
 import {
-  formatSuggestedOperationalPointsWithTrackName,
+  formatSuggestedOperationalPoints,
   matchPathStepAndOp,
   upsertPathStepsInOPs,
 } from 'modules/pathfinding/utils';
@@ -29,6 +28,7 @@ import { castErrorToFailure } from 'utils/error';
 import { getPointCoordinates } from 'utils/geometry';
 
 import type { ManageTrainSchedulePathProperties } from '../types';
+import { useScenarioContext } from './useScenarioContext';
 
 type ItineraryForTrainUpdate = {
   pathSteps: (PathStep | null)[];
@@ -64,8 +64,6 @@ const useSetupItineraryForTrainUpdate = (
   setPathProperties: (pathProperties: ManageTrainSchedulePathProperties) => void,
   trainIdToEdit: number
 ) => {
-  const { getInfraID } = useOsrdConfSelectors();
-  const infraId = useSelector(getInfraID);
   const dispatch = useAppDispatch();
 
   const { updatePathSteps } = useOsrdConfActions();
@@ -77,6 +75,7 @@ const useSetupItineraryForTrainUpdate = (
     osrdEditoastApi.endpoints.postInfraByInfraIdPathfindingBlocks.useMutation();
   const [postPathProperties] =
     osrdEditoastApi.endpoints.postInfraByInfraIdPathProperties.useMutation();
+  const { infraId } = useScenarioContext();
 
   useEffect(() => {
     const computeItineraryForTrainUpdate = async (
@@ -124,14 +123,11 @@ const useSetupItineraryForTrainUpdate = (
       const stepsCoordinates = pathfindingResult.path_item_positions.map((position) =>
         getPointCoordinates(geometry, pathfindingResult.length, position)
       );
-      const suggestedOperationalPoints: SuggestedOP[] =
-        await formatSuggestedOperationalPointsWithTrackName(
-          operational_points,
-          geometry,
-          pathfindingResult.length,
-          infraId,
-          dispatch
-        );
+      const suggestedOperationalPoints: SuggestedOP[] = formatSuggestedOperationalPoints(
+        operational_points,
+        geometry,
+        pathfindingResult.length
+      );
 
       const computedpathSteps = computeBasePathSteps(trainSchedule);
       const updatedPathSteps: PathStep[] = updatePathStepsFromOperationalPoints(
--- a/front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
+++ b/front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
@@ -15,7 +15,7 @@ import {
 } from 'common/api/osrdEditoastApi';
 import { useOsrdConfActions, useOsrdConfSelectors } from 'common/osrdContext';
 import {
-  formatSuggestedOperationalPoints,
+  formatSuggestedOperationalPointsWithTrackName,
   matchPathStepAndOp,
   upsertPathStepsInOPs,
 } from 'modules/pathfinding/utils';
@@ -124,11 +124,14 @@ const useSetupItineraryForTrainUpdate = (
       const stepsCoordinates = pathfindingResult.path_item_positions.map((position) =>
         getPointCoordinates(geometry, pathfindingResult.length, position)
       );
-      const suggestedOperationalPoints: SuggestedOP[] = formatSuggestedOperationalPoints(
-        operational_points,
-        geometry,
-        pathfindingResult.length
-      );
+      const suggestedOperationalPoints: SuggestedOP[] =
+        await formatSuggestedOperationalPointsWithTrackName(
+          operational_points,
+          geometry,
+          pathfindingResult.length,
+          infraId,
+          dispatch
+        );
 
       const computedpathSteps = computeBasePathSteps(trainSchedule);
       const updatedPathSteps: PathStep[] = updatePathStepsFromOperationalPoints(
--- a/front/src/modules/pathfinding/utils.ts
+++ b/front/src/modules/pathfinding/utils.ts
@@ -1,19 +1,15 @@
-import { compact, uniq } from 'lodash';
+import { compact } from 'lodash';
 
-import type { TrackSectionEntity } from 'applications/editor/tools/trackEdition/types';
-import { osrdEditoastApi } from 'common/api/osrdEditoastApi';
 import type {
   GeoJsonLineString,
   PathProperties,
   PathfindingInput,
   PostInfraByInfraIdPathfindingBlocksApiArg,
   RollingStockWithLiveries,
-  TrackSection,
 } from 'common/api/osrdEditoastApi';
 import { getSupportedElectrification, isThermal } from 'modules/rollingStock/helpers/electric';
 import type { SuggestedOP } from 'modules/trainschedule/components/ManageTrainSchedule/types';
 import type { PathStep } from 'reducers/osrdconf/types';
-import type { AppDispatch } from 'store';
 import { addElementAtIndex } from 'utils/array';
 import { getPointCoordinates } from 'utils/geometry';
 
@@ -21,8 +17,8 @@ import getStepLocation from './helpers/getStepLocation';
 
 export const formatSuggestedOperationalPoints = (
   operationalPoints: NonNullable<Required<PathProperties['operational_points']>>,
-  geometry: GeoJsonLineString,
-  pathLength: number
+  geometry?: GeoJsonLineString,
+  pathLength?: number
 ): SuggestedOP[] =>
   operationalPoints.map((op) => ({
     opId: op.id,
@@ -37,54 +33,12 @@ export const formatSuggestedOperationalPoints = (
     offsetOnTrack: op.part.position,
     track: op.part.track,
     positionOnPath: op.position,
-    coordinates: getPointCoordinates(geometry, pathLength, op.position),
+    coordinates:
+      (geometry !== undefined &&
+        pathLength !== undefined &&
+        getPointCoordinates(geometry, pathLength, op.position)) ||
+      undefined,
   }));
-
-export const formatSuggestedOperationalPointsWithTrackName = async (
-  operationalPoints: NonNullable<Required<PathProperties['operational_points']>>,
-  geometry: GeoJsonLineString,
-  pathLength: number,
-  infraId: number,
-  dispatch: AppDispatch
-): Promise<SuggestedOP[]> => {
-  // Extract track IDs to fetch track names
-  const trackIds = uniq(operationalPoints.map((op) => op.part.track));
-  const trackSections = await dispatch(
-    osrdEditoastApi.endpoints.postInfraByInfraIdObjectsAndObjectType.initiate({
-      infraId,
-      objectType: 'TrackSection',
-      body: trackIds,
-    })
-  ).unwrap();
-
-  return operationalPoints.map((op) => {
-    const associatedTrackSection = trackSections.find(
-      (trackSection) => (trackSection.railjson as TrackSection).id === op.part.track
-    );
-
-    const trackName = associatedTrackSection
-      ? (associatedTrackSection.railjson as TrackSectionEntity['properties']).extensions?.sncf
-          ?.track_name
-      : null;
-
-    return {
-      opId: op.id,
-      name: op.extensions?.identifier?.name,
-      uic: op.extensions?.identifier?.uic,
-      ch: op.extensions?.sncf?.ch,
-      kp: op.part.extensions?.sncf?.kp,
-      chLongLabel: op.extensions?.sncf?.ch_long_label,
-      chShortLabel: op.extensions?.sncf?.ch_short_label,
-      ci: op.extensions?.sncf?.ci,
-      trigram: op.extensions?.sncf?.trigram,
-      offsetOnTrack: op.part.position,
-      track: op.part.track,
-      positionOnPath: op.position,
-      coordinates: getPointCoordinates(geometry, pathLength, op.position),
-      trackName: trackName || undefined,
-    };
-  });
-};
 
 export const matchPathStepAndOp = (
   step: PathStep,
--- a/front/src/modules/pathfinding/utils.ts
+++ b/front/src/modules/pathfinding/utils.ts
@@ -1,15 +1,19 @@
-import { compact } from 'lodash';
+import { compact, uniq } from 'lodash';
 
+import type { TrackSectionEntity } from 'applications/editor/tools/trackEdition/types';
+import { osrdEditoastApi } from 'common/api/osrdEditoastApi';
 import type {
   GeoJsonLineString,
   PathProperties,
   PathfindingInput,
   PostInfraByInfraIdPathfindingBlocksApiArg,
   RollingStockWithLiveries,
+  TrackSection,
 } from 'common/api/osrdEditoastApi';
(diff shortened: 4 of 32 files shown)


Synar:
(Update to the tests coming soon ^^)

Synar:
@Yohh je t'ajoute en review juste pour le mini changement de test

Akctarus (commented):
Tested, perhaps we could add a little padding-right for long track names to match the padding-left:

Capture d’écran 2024-11-08 à 11 25 24

Apart from that, LGTM!

Synar:
Thanks! I extended the width a bit.

Synar:
2 notes : - conflict with incoming https://github.com/OpenRailAssociation/osrd/pull/9642 to fix, probably by replacing the formatSuggestedOperationalPoints call by a formatSuggestedOperationalPointsWithTrackName call in use output table data
- we should probably cache the tracks information, possibly as a context in the scenario page

front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
Synar, line 1: The changes in this file are much more minor than they seem at first glance, we just added a `<ScenarioContextProvider infraId={infra.id} dispatch={dispatch}> `wrapper, but git doesn't catch this due to indentation changes.

Synar:
Todo :
 - [x] add tracksectionsLoading state to hook and context

front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
SharglutDev: Ty for clarifying this :)

front/src/applications/operationalStudies/hooks/useScenarioContext.tsx
```diff
@@ -0,0 +1,42 @@
...
+import useCachedTrackSections from 'applications/operationalStudies/hooks/useTrackSections';
+import type { InfraObjectWithGeometry } from 'common/api/osrdEditoastApi';
+import type { AppDispatch } from 'store';
+
+type ScenarioContextType = {
+  getTrackSectionsByIds: (requestedTrackIds: string[]) => Promise<InfraObjectWithGeometry[]>;
+  infraId: number;
+  trackSectionsLoading: boolean;
+} | null;
+export const ScenarioContext = createContext<ScenarioContextType>(null);
```
SharglutDev: I think we don't need to export this ?

front/src/applications/operationalStudies/hooks/useTrackSections.ts
```diff
@@ -0,0 +1,47 @@
+import { useState, useCallback } from 'react';
+
+import { osrdEditoastApi } from 'common/api/osrdEditoastApi';
+import type { InfraObjectWithGeometry } from 'common/api/osrdEditoastApi';
+import type { AppDispatch } from 'store';
+
+export default function useCachedTrackSections(infraId: number, dispatch: AppDispatch) {
```
SharglutDev: Since we are in a hook, can't we use `useInfraId` and `useAppDispatch` directly here so we don't have to pass them in param ?

front/src/applications/operationalStudies/hooks/useTrackSections.ts
```diff
@@ -0,0 +1,47 @@
...
+      const uniqueNewIds = newTrackIds.filter((id) => !trackIds.has(id));
+      if (uniqueNewIds.length === 0) return;
+
+      const fetchedSections = await dispatch(
+        osrdEditoastApi.endpoints.postInfraByInfraIdObjectsAndObjectType.initiate({
+          infraId,
+          objectType: 'TrackSection',
+          body: uniqueNewIds,
+        })
+      ).unwrap();
```
SharglutDev: Since we don't export this, maybe we could call this endpoint with `useMutation` in `useCachedTrackSections`'s body directly and call the function here ?

This way we don't need the dispatch. We can also provide an `isLoading` so you don't have to use the extra state.

```suggestion
   const [loadInfraObject, { isLoading }] = osrdEditoastApi.endpoints.postInfraByInfraIdObjectsAndObjectType.useMutation();
```

front/src/applications/operationalStudies/hooks/useTrackSections.ts
```diff
@@ -0,0 +1,47 @@
...
+    async (requestedTrackIds: string[]) => {
+      setLoading(true);
+      try {
+        await loadTracksSectionsByIds(requestedTrackIds);
+      } catch (error) {
+        console.error('Failed to fetch track sections:', error);
+      } finally {
+        setLoading(false);
+      }
+      return trackSections.filter((section) => requestedTrackIds.includes(section.obj_id));
```
SharglutDev: Are we sure `trackSections` will be already updated after the call of `loadTracksSectionsByIds` ?

Do we have to pass by a state for `trackSections` value ? Couldn't we just pass the content of `loadTracksSectionsByIds` in the `try` so we can access the result directly ?

front/src/applications/operationalStudies/hooks/useTrackSections.ts
```diff
@@ -0,0 +1,47 @@
...
+          infraId,
+          objectType: 'TrackSection',
+          body: uniqueNewIds,
+        })
+      ).unwrap();
+
+      setTrackIds((prev) => new Set([...prev, ...uniqueNewIds]));
+      setTrackSections((prev) => [...prev, ...fetchedSections]);
+    },
+    [trackIds, infraId, dispatch]
```
SharglutDev: Won't this be called twice since we `setTrackIds` at the end but we also put `trackIds` in the deps ?

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -181,7 +182,7 @@ const useSetupItineraryForTrainUpdate = (
       const stepsCoordinates = pathfindingResult.path_item_positions.map((position) =>
         getPointCoordinates(geometry, pathfindingResult.length, position)
       );
-      const suggestedOperationalPoints: SuggestedOP[] = formatSuggestedOperationalPoints(
+      const suggestedOperationalPoints: SuggestedOP[] = await formatSuggestedOperationalPoints(
```
SharglutDev: This one is not an async function

front/src/modules/pathfinding/utils.ts
```diff
@@ -36,6 +39,31 @@ export const formatSuggestedOperationalPoints = (
     coordinates: getPointCoordinates(geometry, pathLength, op.position),
   }));
 
+export const fetchTrackNamesForSuggestedOperationalPoints = async (
```
SharglutDev: Nit : for me, it implies that the function returns only the track names. Maybe we could try this (only suggestion)

```suggestion
export const addTrackNamesToSuggestedOperationalPoints = async (
```

front/src/modules/timesStops/TimesStopsInput.tsx
```diff
@@ -139,16 +147,23 @@ const TimesStopsInput = ({ allWaypoints, startTime, pathSteps }: TimesStopsInput
...
+          suggestedOPsWithTrackNames,
+          pathSteps || [],
+          t,
+          startTime,
+          TableType.Input
+        );
+        setRows(updateDaySinceDeparture(formatedRows, startTime, { keepFirstIndexArrival: true }));
+      }
+    };
+    fetchAndFormatRows();
```
SharglutDev: For readability :)

```suggestion
    };
    
    fetchAndFormatRows();
```

front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -37,19 +41,31 @@ function useOutputTableData(
...
+          getTrackSectionsByIds
+        );
+        setSuggestedOperationalPoints(formattedPointsWithTrackNames);
+      } else {
+        setSuggestedOperationalPoints([]);
+      }
+    };
+
+    fetchSuggestedOperationalPoints();
+  }, [pathProperties.operationalPoints, pathProperties.geometry, getTrackSectionsByIds]);
```
SharglutDev: Can't we keep the `useMemo` here and avoid the state ? Or the component won't rerender if we don't ?

SharglutDev (commented):
Thank you for the last changes ! Left some comments, they are totally debatable

front/src/applications/operationalStudies/hooks/useScenarioContext.tsx
Synar: Indeed thanks!

front/src/applications/operationalStudies/hooks/useTrackSections.ts
Synar: Agreed for appDispatch, but clara told me we want to get the infra id out of the store and into the context. And while this is just a preliminary step I would still prefer to avoid adding another reliance to the store for the infra id.

front/src/applications/operationalStudies/hooks/useTrackSections.ts
Synar: Perhaps usequery since we only fetch? I'm not sure I fully understand the implications of the choice of using dispatch, usemutation or usequery though, so I'll defer to your judgement 

(Nitpick : that loading will encompass less code though, but the sync code should be instantaneous so it probably doesn't matter at all.)

front/src/applications/operationalStudies/hooks/useTrackSections.ts
SharglutDev: Ok seems legit :)

front/src/applications/operationalStudies/hooks/useTrackSections.ts
Synar: I don't believe so since it's a usecallback not a useeffect, but it will indeed redefine the function every time.

front/src/applications/operationalStudies/hooks/useTrackSections.ts
SharglutDev: Like some other endpoints from the api, we fetch here but in reality, as we need to pass a body to the call, it's technically a post so we can only do a `useMutation` :)

If we want to use it with a query, we need to override the endpoint like we did for other of them in this file : `openapi-editoast-config.ts`

front/src/modules/timesStops/hooks/useOutputTableData.ts
Synar: Correct me if I'm wrong on this, but since the format operation is now async and useMemo doesn't natively support async operations, I believe we need to use useEffect to handle it. Then because useEffect does not return a value, I believe we need to use the state to track and store the result (in a way that is rerendered) once it completes.

I don't have a ton of experience with hooks yet so I may be overlooking something though

front/src/modules/timesStops/TimesStopsInput.tsx
Synar: Sure! ^^

front/src/modules/pathfinding/utils.ts
Synar: Sure! By calling it fetch I wanted to emphasize the api call, but the fact the function is async should be enough clues.

front/src/applications/operationalStudies/hooks/useTrackSections.ts
Synar: > Are we sure trackSections will be already updated after the call of loadTracksSectionsByIds ?

I assumed it would, but indeed it may not be.

> Do we have to pass by a state for trackSections value ? Couldn't we just pass the content of loadTracksSectionsByIds in the try so we can access the result directly ?

We can't just pass the value without also reading the state because we only fetch new ids, but you're right this needs to be reworked a bit.

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
Synar: Got mixed up, thanks!

front/src/applications/operationalStudies/hooks/useTrackSections.ts
Synar: Thanks for the explanation! I ended up going with useMutation.

Synar:
2 new fixup, the second one is the change from usestate to useref for the content of the cache, to avoid dependency loops and useless renders

front/src/applications/operationalStudies/hooks/useTrackSections.ts
Synar: Since the function was itself used as a dependency of useffects though, I opted to drop the states and go for useref instead. It seems to work properly when testing.

Synar:
Clara asked me to ping all maintainers about adding context to the scenario, so here : @kmer2016 @emersion @SharglutDev @clarani @Math-R . Sorry for the inconvenience, hope I didn't forget anyone ^^

front/src/applications/operationalStudies/hooks/useTrackSections.ts
SharglutDev, line 1: Feel like it's always better to have the file and hook name synchronized. Could we chose one or the other ? :)

front/src/applications/operationalStudies/hooks/useTrackSections.ts
```diff
@@ -0,0 +1,38 @@
...
+            objectType: 'TrackSection',
+            body: uniqueNewIds,
+          }).unwrap();
+        } catch (error) {
+          console.error('Failed to fetch track sections:', error);
+        }
+      }
+
+      uniqueNewIds.forEach((id) => trackIdsRef.current.add(id));
+      trackSectionsRef.current = [...trackSectionsRef.current, ...fetchedSections];
```
SharglutDev: Don't we want to do that in the `try` ?

front/src/modules/timesStops/hooks/useOutputTableData.ts
SharglutDev: Yep you're right

front/src/applications/operationalStudies/hooks/useTrackSections.ts
Synar: Done ^^

front/src/applications/operationalStudies/hooks/useTrackSections.ts
Synar: You're right! Thanks!

SharglutDev:
Lgtm and tested for me. Waiting for at least a 2nd maintainer approval for the context subject :) (@clarani)

front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
```diff
@@ -50,6 +52,8 @@ const ManageTrainSchedule = ({ trainIdToEdit }: ManageTrainScheduleProps) => {
   const { rollingStockId, rollingStockComfort, rollingStock } =
     useStoreDataForRollingStockSelector();
 
+  const { trackSectionsLoading } = useScenarioContext();
```
kmer2016: Since `trackSectionsLoading` is only used by `TimeStopInput`, and it already accesses the context for `getTrackSectionsByIds`, it would be cleaner and more maintainable to fetch it directly from the context, avoiding unnecessary prop drilling. What do you think"

kmer2016 (commented):
Nice job ! Just a small suggestion

emersion:
Overall the context approach sounds good to me :)

front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
Synar: Sure! I wanted to have timesstopsinput mirror timesstopoutput, but yes it probably makes sense to put it directly in timestopsinput here

front/src/modules/pathfinding/utils.ts
```diff
@@ -33,9 +36,38 @@ export const formatSuggestedOperationalPoints = (
...
+    );
+
+    const trackName = associatedTrackSection
+      ? (associatedTrackSection.railjson as TrackSectionEntity['properties']).extensions?.sncf
+          ?.track_name
+      : null;
+
+    return {
+      ...op,
+      trackName: trackName || undefined,
```
clarani: You can use undefined by default here and remove it l66
```suggestion
      : undefined;

    return {
      ...op,
      trackName: trackName,
```

front/src/modules/pathfinding/utils.ts
```diff
@@ -33,9 +36,38 @@ export const formatSuggestedOperationalPoints = (
...
+    coordinates:
+      (geometry !== undefined &&
+        pathLength !== undefined &&
+        getPointCoordinates(geometry, pathLength, op.position)) ||
+      undefined,
   }));
 
+export const addTrackNamesToSuggestedOperationalPoints = async (
+  operationalPoints: SuggestedOP[],
+  getTrackSectionsByIds: (requestedTrackIds: string[]) => Promise<InfraObjectWithGeometry[]>
```
clarani: Hum I don't think it's very useful to give the argument `getTrackSectionsById`, it would be better to call it outside `addTrackNamesToSuggestedOperationalPoints` and to give in argument only `operationalPoints` and `trackSections`.

front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -87,44 +95,60 @@ const useOutputTableData = (
...
+    const formatRows = async () => {
+      if (!operationalPoints || !startDatetime) {
+        setRows([]);
+        return;
       }
 
-      // compute arrival time
-      const matchingReportTrainIndex = simulatedTrain.positions.findIndex(
-        (position) => position === op.position
+      const formattedPoints = formatSuggestedOperationalPoints(operationalPoints);
```
clarani: Ah hum the type SuggestedOperationalPoint is very messy and I wish to avoid spreading it. Can you make the function addTrackNames more generic so it takes only operationalPoints which contain at least a trackId and returns the same list of operationalPoints with their trackNames ?

It would be smth like this:
```
export const addTrackNamesToOperationalPoints = async <T extends { track: string}>(
  operationalPoints: T[],
  trackSections: InfraObjectWithGeometry[]
): Promise<(T & { trackName?: string})[]> => {...}
```

front/src/modules/pathfinding/utils.ts
```diff
@@ -33,9 +36,38 @@ export const formatSuggestedOperationalPoints = (
...
+): Promise<SuggestedOP[]> => {
+  // Extract track IDs to fetch track names
+  const trackIds = uniq(operationalPoints.map((op) => op.track));
+  const trackSections = await getTrackSectionsByIds(trackIds);
+
+  return operationalPoints.map((op) => {
+    const associatedTrackSection = trackSections.find(
```
clarani: Maybe trackSections can be a Map instead of an array ? What do you think ?

front/src/applications/operationalStudies/hooks/useCachedTrackSections.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export default function useCachedTrackSections(infraId: number) {
+  const trackIdsRef = useRef<Set<string>>(new Set());
+  const trackSectionsRef = useRef<InfraObjectWithGeometry[]>([]);
+  const [loadInfraObject, { isLoading }] =
+    osrdEditoastApi.endpoints.postInfraByInfraIdObjectsAndObjectType.useMutation();
+
+  const getTrackSectionsByIds = useCallback(
+    async (requestedTrackIds: string[]) => {
+      let fetchedSections: InfraObjectWithGeometry[] = [];
```
clarani: You can use const directly l18
```suggestion
```

front/src/applications/operationalStudies/hooks/useCachedTrackSections.ts
```diff
@@ -0,0 +1,39 @@
+import { useRef, useCallback } from 'react';
+
+import { osrdEditoastApi } from 'common/api/osrdEditoastApi';
+import type { InfraObjectWithGeometry } from 'common/api/osrdEditoastApi';
+
+export default function useCachedTrackSections(infraId: number) {
+  const trackIdsRef = useRef<Set<string>>(new Set());
+  const trackSectionsRef = useRef<InfraObjectWithGeometry[]>([]);
```
clarani: Do you think it would be smart to store only the railjson, so the type would be directly `TrackSection[]` ?

front/src/applications/operationalStudies/hooks/useScenarioContext.tsx
```diff
@@ -0,0 +1,35 @@
...
+    }),
+    [getTrackSectionsByIds, infraId, trackSectionsLoading]
+  );
+  return <ScenarioContext.Provider value={providedContext}>{children}</ScenarioContext.Provider>;
+};
+
+export const useScenarioContext = () => {
+  const context = useContext(ScenarioContext);
+  if (!context) {
+    throw new Error('useScenarioContext must be used within a TrackSectionsProvider');
```
clarani:
```suggestion
    throw new Error('useScenarioContext must be used within a ScenarioContextProvider');
```

front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
clarani, line 1: I think you can even move the ScenarioContextProvider in the parent component, in Scenario.tsx

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
clarani, line 1: Nice :)

front/src/modules/simulationResult/hooks/useFormattedOperationalPoints.ts
```diff
@@ -16,28 +17,28 @@ import { formatOperationalPoints } from '../SimulationResultExport/utils';
...
-    if (train && simulatedTrain && pathProperties && infraId) {
+    if (train && simulatedTrain && pathProperties) {
       const fetchOperationalPoints = async () => {
         setLoading(true);
         const formattedOperationalPoints = await formatOperationalPoints(
           pathProperties.operationalPoints,
           simulatedTrain,
           train,
-          infraId
+          getTrackSectionsByIds
```
clarani: Would it be possible to get the `trackSections` outside the function `formatOperationalPoints` and to only give the trackSections as argument ?

front/src/modules/pathfinding/utils.ts
Synar: Sure!

front/src/modules/pathfinding/utils.ts
Synar: Honestly I just duplicated the old code for this, but sure!

front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
Synar: Ok!

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
Synar: Thanks!

front/src/modules/simulationResult/hooks/useFormattedOperationalPoints.ts
Synar: I think so!

front/src/applications/operationalStudies/hooks/useScenarioContext.tsx
Synar: Nice catch thanks!

front/src/modules/timesStops/hooks/useOutputTableData.ts
Synar: I thought about that, got lazy and decided to avoid messing with generics, but yeah you're right I will

front/src/modules/pathfinding/utils.ts
Synar: Nice catch! I don't understand why I wrote that in the first place ^^''

Synar: Actually let me simplify this further
`    const trackName = (
      associatedTrackSection?.railjson as TrackSectionEntity['properties'] | undefined
    )?.extensions?.sncf?.track_name;`

front/src/modules/pathfinding/utils.ts
Synar: It does mean that every caller will have to perform the 3 same steps instead of just calling the function though :
-> extract ids to fetch
-> fetch the ids
-> call this function (which is reduced to its last step)

Plus if we need to make it generic, we may as well delete the function and just inline everything

front/src/applications/operationalStudies/hooks/useCachedTrackSections.ts
Synar: Done!

front/src/applications/operationalStudies/hooks/useCachedTrackSections.ts
Synar: Good catch thanks! I used to use it outside the try, but there's no longer a reason to declare it before.

front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -104,44 +100,45 @@ const useOutputTableData = (
...
-      const formattedPointsWithTrackNames = await addTrackNamesToSuggestedOperationalPoints(
-        formattedPoints,
-        getTrackSectionsByIds
-      );
+      const trackIds = operationalPoints.map((op) => op.part.track);
+      const trackSections = await getTrackSectionsByIds(trackIds);
+      const formattedPointsWithTrackNames = operationalPoints.map((op) => ({
+        ...op,
+        trackName: trackSections[op.part.track]?.extensions?.sncf?.track_name,
+      }));
```
clarani: You don't need to do this map ;)
You can directly get the trackName of the op l114 and return it in the correct object
```suggestion
```

front/src/modules/pathfinding/utils.ts
clarani: Yes you're right. I find the code way cleaner and easier to understand, like this, thanks :star_struck:

front/src/modules/timesStops/hooks/useOutputTableData.ts
Synar: Oh right thanks! I was still thinking in terms of the previous function, but there's no longer a need for that ^^

Maymanaf (commented):
Nice PR! The changes look good for the minor update in the e2e test.

If possible, could you expand the scope to include a check for the track values? If adding it now isn't feasible, please create an issue and assign it to me for follow-up later.

Thanks! 😊

Synar:
> Nice PR! The changes look good for the minor update in the e2e test.
> 
> If possible, could you expand the scope to include a check for the track values? If adding it now isn't feasible, please create an issue and assign it to me for follow-up later.
> 
> Thanks! 😊

Done!

clarani (approved):
LGTM :white_check_mark:

Maymanaf (approved):
LGTM for e2e tests  ✅

PR #9622, opened 2024-11-07
Fix grant and revoke role to group using CLI

flomonster:


woshilapin (approved):
Looks good. Thank you.

leovalais (approved):
Indeed groups exist 🤦 Thanks for the fix!

PR #9640, opened 2024-11-07
front: stdcm: fix time interpolation in output table

### LINKED ISSUE #9626: stdcm - times in the pdf simulation sheet are not correct
### What happened?

I created a simulation (see pdf below to see the asked OPs).
The simulation works so I export the pdf file.
But in the file, one can see that the train goes from Iwuy to Caudry in 0 minutes (21:12 for all points in between). But in reality, when I compute the travel time between these 2 points, it is 14 minutes. 

[STDCM-1124-671-555.pdf](https://github.com/user-attachments/files/17660515/STDCM-1124-671-555.pdf)

I expected to have the real times

create a simulation with the input of the attached pdf file

Dev (SNCF)

Google Chrome

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/9626

The back-end returns sparse data for time per position and expects the front-end to interpolate, but we didn't. The results were significantly off.

This fixes both the table displayed in the result page and the one in the pdf file.


clarani (approved):

Akctarus (approved):
Tested, LGTM :)

PR #9642, opened 2024-11-07
front: refacto useOutputData

### LINKED ISSUE #8848: front: The first line of the output table is orange when there is a clicked via before
### What happened?

On the output table of a simulation, the first line is orange if the first real via is a clicked via on the map (meaning we are late, but it's not true)

I add a VIA just before the OP on a Paris-Marseille train
image

the first line of the output table is orange
image

we should see the via in the output table (there is an other issue for this)
and the first line should not be orange if we are not late

1. go to a scenario
2. create a train with a clicked via for the start of the schedule
3. check the output table after the simulation

Recette (SNCF)

Google Chrome

bc8b913

### LINKED ISSUE #8949: front: refacto useOutputTableData
### Description and goal

Refacto useOutputTableData

### Acceptance criteria

- [ ] useOutputTableData is clearer

clarani:
closes #8949 
will also close #8848


front/src/applications/operationalStudies/views/SimulationResults.tsx
Synar, line 1: > selectedTrainSchedule && pathProperties && selectedTrainSummary

Could some of these variables be in the dataIsLoading parameter instead?

front/src/modules/timesStops/helpers/__tests__/arrivalTime.spec.ts
Synar, line 1: Is the deletion of this test temporary, or is there a reason to delete it? (Same questions for each deleted file)

Synar (commented):
Very nice PR! useOutputTableData was a mess, it's a lot more straightforward now

Edit : nevermind, I think the line I chose for testing is just bugged, but it seems to work fine for other lines, and possibly that issue has nothing to do with the pr. I'm checking ^^

Edit edit : yep the bug (waypoints in the wrong order with the wrong informations for paris gare de lyon when going from gare de lyon to lyon) is there on dev also. Sorry!

front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -1,157 +1,126 @@
...
         };
-      }
-
-      return {
-        ...sugOpPoint,
-        calculatedArrival: secToHoursString(opPoint.time, { withSeconds: true }),
-      };
-    });
+      }),
+    [pathProperties.operationalPoints]
```
Synar: I think the usememo is lying a bit about its dependencies
```suggestion
    [pathProperties.operationalPoints, pathStepRows, simulatedTrain, selectedTrainSchedule]
```

front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -1,157 +1,126 @@
...
+        calculatedMargin,
+        diffMargins,
+        calculatedArrival: dateToHHMM(computedArrival, { withSeconds: true }),
+        calculatedDeparture,
+        positionOnPath: path.path_item_positions[index],
+      });
+    }
+
+    return result;
+  }, [selectedTrainSchedule, path]);
```
Synar: Same as below
```suggestion
  }, [selectedTrainSchedule, path, trainSummary.pathItemTimes]);
```

front/src/modules/timesStops/helpers/computeMargins.ts
```diff
@@ -30,47 +26,55 @@ function getTheoreticalMargin(selectedTrainSchedule: TrainScheduleResult, pathSt
...
+  }
+
+  const pathStepId = path[pathStepIndex].id;
   const theoreticalMargin = getTheoreticalMargin(selectedTrainSchedule, pathStepId);
 
-  // durations to go from current scheduled point to next scheduled point.
+  // find the previous pathStep where margin was defined
+  let prevIndex = 0;
+  // eslint-disable-next-line no-plusplus
+  for (let index = 1; index < pathStepIndex; index++) {
```
Synar: Why disable the eslint warning instead of doing index+=1?

```suggestion
  for (let index = 1; index < pathStepIndex; index += 1) {
```

front/src/applications/operationalStudies/views/SimulationResults.tsx
clarani: Done :)

front/src/modules/timesStops/hooks/useOutputTableData.ts
clarani: You're right, thanks !!

front/src/modules/timesStops/helpers/computeMargins.ts
clarani: You're right, thanks :)

front/src/modules/timesStops/helpers/__tests__/arrivalTime.spec.ts
clarani: I think this function is not used anymore, that is why I removed the test

Synar: Right, thanks! 

On this subject why did you remove the one minute "tolerance margin" for calculated arrival? Was it something the POs asked?

front/src/modules/timesStops/helpers/computeMargins.ts
```diff
@@ -30,47 +26,55 @@ function getTheoreticalMargin(selectedTrainSchedule: TrainScheduleResult, pathSt
...
+
+  // durations to go from the previous pathStep with margin to current one.
   // base = no margin
   // provisional = margins
   // final = margin + requested arrival times
-  const baseDuration = computeDuration(trainSimulation.base, opPoint, nextOpPoint);
-  const provisionalDuration = computeDuration(trainSimulation.provisional, opPoint, nextOpPoint);
-  const finalDuration = computeDuration(trainSimulation.final_output, opPoint, nextOpPoint);
+  const { base, provisional, final } = pathItemTimes;
+  const baseDuration = ms2sec(base[pathStepIndex + 1] - base[prevIndex]);
```
SharglutDev, line 64: Could you explain why we need the `+ 1` index ?

front/src/modules/timesStops/TimesStopsOutput.tsx
```diff
@@ -15,35 +15,37 @@ import { TableType, type TimeStopsRow } from './types';
...
     simulatedTrain,
+    trainSummary,
     pathProperties,
-    operationalPoints,
     selectedTrainSchedule,
     path
   );
-  if (dataIsLoading) {
+
+  if (dataIsLoading || !trainSummary || !pathProperties || !selectedTrainSchedule) {
```
SharglutDev: Shouldn't we put the loader before the call of `useOutputTableData` ? And this way maybe we could remove some early returns in it ?

front/src/modules/timesStops/helpers/__tests__/scheduleData.spec.ts
```diff
@@ -2,24 +2,34 @@ import { describe, it, expect } from 'vitest';
...
+    const startDatetime = new Date('2022-01-01T02:03:00Z');
+    const schedule = {
+      at: 'id325',
+      arrival: 'PT3600S',
+      stop_for: 'PT100S',
+      reception_signal: 'OPEN' as ReceptionSignal,
+    };
+
+    expect(formatSchedule(startDatetime, schedule)).toEqual({
+      calculatedDeparture: '02:04:40',
```
SharglutDev, line 28: Will this test work after the next timezone change in April ?

front/src/utils/date.ts
```diff
@@ -254,3 +254,8 @@ export const formatLocaleDate = (date: Date) => date.toLocaleString().substring(
 
 export const isEqualDate = (searchDate: Date, startDate: Date) =>
   formatLocaleDate(searchDate) === formatLocaleDate(startDate);
+
+export const dateToHHMM = (date: Date, { withSeconds } = { withSeconds: false }) => {
```
SharglutDev: As we already have functions `extractHHMM` and `extractHHMMSS` in `date.ts`. Maybe we could replace them with your new function which seems better ? Can be done in another PR but we should create the ticket not the forget it.

SharglutDev: Also the name is weird because sometimes, it will return a date in `HHMMSS` (like the use in `computeDayTimeFromStartTime`)

front/src/modules/timesStops/helpers/arrivalTime.ts
```diff
@@ -1,26 +1,58 @@
...
+  const arrivalDatetime = new Date(startDatetime.getTime() + duration * 1000);
+
+  const isAfterMidnight = arrivalDatetime.getDate() !== previousDatetime.getDate();
+
+  const timeExtraDay = {
+    time: dateToHHMM(arrivalDatetime, { withSeconds: true }),
+    daySinceDeparture:
+      Math.abs(arrivalDatetime.getTime() - startDatetime.getTime()) / (1000 * 60 * 60 * 24),
+    dayDisplayed: isAfterMidnight,
+  };
```
SharglutDev: We should add tests for this function because we lost most of the logic to calculate the difference between dates in days (was quite painful to find a "good" solution !) without using the old logic (holding 1 in a variable each time we get a `isAfterMidnight`).

`Math.abs()` won't return you an integer, you should use `Math.floor()` and the parenthesis weren't also properly placed to get an integer. 
Also, with this operation, you check if the difference between two dates last more than 1 day but how about if we compare these 2 dates : 22/11/24 23h55 and 23/11/24 02h59 ? We won't get the expected result even though we would want `1`.

Best solution I found is this (we can also use dayjs to avoid `1000 * 60...` :

```suggestion
  const startDate = new Date(
    startDatetime.getFullYear(),
    startDatetime.getMonth(),
    startDatetime.getDate()
  );
  const arrivalDate = new Date(
    arrivalDatetime.getFullYear(),
    arrivalDatetime.getMonth(),
    arrivalDatetime.getDate()
  );
  const timeExtraDay = {
    time: dateToHHMM(arrivalDatetime, { withSeconds: true }),
    daySinceDeparture: dayjs.duration(arrivalDate.getTime() - startDate.getTime()).asDays(),
    dayDisplayed: isAfterMidnight,
  };
```

front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -4,154 +4,124 @@ import { keyBy } from 'lodash';
...
-        sugOpIndex
+  const pathStepRows = useMemo(() => {
+    if (!path || !selectedTrainSchedule || !trainSummary?.pathItemTimes) return [];
+
+    const startDatetime = new Date(selectedTrainSchedule.start_time);
+
+    const result: (TimeStopsRow & { positionOnPath: number })[] = [];
+    let previousTime = new Date(selectedTrainSchedule.start_time);
+
+    for (let index = 0; index < selectedTrainSchedule.path.length; index += 1) {
```
SharglutDev: Since we loop on `selectedTrainSchedule.path`, couldn't be instead `return selectedTrainSchedule.path.map()` and put the logic inside ? 
This way, we could avoid : 
- initializing result
- define `pathStep` at line 40
- define `previousTime` : we could pass `pathStep[index - 1].arrival` (or departure if there is one) to `computeInputDatetimes` (except for index 0 where we pass `startTime`)
- we probably won't have to return `prev` in `computeInputDatetimes`

front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -4,154 +4,124 @@ import { keyBy } from 'lodash';
...
+      );
+
+      const { stopFor, shortSlipDistance, onStopSignal, calculatedDeparture } = formatSchedule(
+        computedArrival,
+        schedule
+      );
+      const { theoreticalMargin, theoreticalMarginSeconds, calculatedMargin, diffMargins } =
+        computeMargins(selectedTrainSchedule, index, trainSummary.pathItemTimes);
+
+      const inputs = computeInputDatetimes(startDatetime, previousTime, schedule, {
```
SharglutDev:
```suggestion
      const { arrival, departure, prev } = computeInputDatetimes(startDatetime, previousTime, schedule, {
```

front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -4,154 +4,124 @@ import { keyBy } from 'lodash';
...
+        isDeparture: index === 0,
+      });
+      const { arrival, departure } = inputs;
+      previousTime = inputs.prev;
```
SharglutDev: This property could be renamed `refDate` (just a suggestion) or something like that (`previousTime` is actually a `Date`)

```suggestion
      lastRefDate = inputs.refDate;
```

front/src/modules/timesStops/helpers/arrivalTime.ts
```diff
@@ -1,26 +1,58 @@
...
+      const departureInSeconds =
+        ISO8601Duration2sec(schedule.arrival) + ISO8601Duration2sec(schedule.stop_for);
+      const resultDeparture = computeDayTimeFromStartTime(startDatetime, departureInSeconds, prev);
+      departure = resultDeparture.timeExtraDay;
+      prev = resultDeparture.previousTime;
+    }
+  }
+
+  // if is departure and no arrival is defined, use the startDatetime
+  if (isDeparture && !arrival) {
```
SharglutDev: Can the departure have an `arrival` ?

front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -4,154 +4,124 @@ import { keyBy } from 'lodash';
...
+      });
+    }
+
+    return result;
+  }, [selectedTrainSchedule, path, trainSummary?.pathItemTimes]);
+
+  const rows: TimeStopsRow[] = useMemo(() => {
+    if (!pathProperties || !selectedTrainSchedule) return [];
+
+    const startDatetime = new Date(selectedTrainSchedule.start_time);
```
SharglutDev: We could define this only once at the top of the hook (won't to check for `undefined` if you moved the Loader before this hook call)

front/src/modules/timesStops/TimesStopsOutput.tsx
```diff
@@ -15,35 +15,37 @@ import { TableType, type TimeStopsRow } from './types';
 
 type TimesStopsOutputProps = {
   simulatedTrain: SimulationResponseSuccess;
-  pathProperties: PathPropertiesFormatted;
-  operationalPoints: OperationalPointWithTimeAndSpeed[];
-  selectedTrainSchedule: TrainScheduleResult;
+  trainSummary?: TrainScheduleWithDetails;
+  pathProperties?: PathPropertiesFormatted;
```
SharglutDev: It seems like we need only `operationalPoints` from `pathProperties`, we could pass only this as a prop

```suggestion
  operationalPoints: OperationalPoint[];
```

front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -4,154 +4,124 @@ import { keyBy } from 'lodash';
...
+
+    return pathProperties.operationalPoints.map((op) => {
+      const matchingPathStep = pathStepRows.find(
+        (pathStepRow) => op.position === pathStepRow.positionOnPath
```
SharglutDev, line 97: I'm afraid we might not get the proper matching for weird cases (import issues) where we have duplicates op at same position. If I have two ops at same position i.e. opA and opB at position 125, I put a stopFor on opB but the matching will be made with opA.

front/src/modules/timesStops/hooks/useOutputTableData.ts
```diff
@@ -4,154 +4,124 @@ import { keyBy } from 'lodash';
...
+import { interpolateValue } from 'modules/simulationResult/SimulationResultExport/utils';
+import type { TrainScheduleWithDetails } from 'modules/trainschedule/components/Timetable/types';
+import { dateToHHMM } from 'utils/date';
 
-import { checkAndFormatCalculatedArrival } from '../helpers/arrivalTime';
+import { computeInputDatetimes } from '../helpers/arrivalTime';
 import computeMargins from '../helpers/computeMargins';
-import { computeScheduleData, formatScheduleData } from '../helpers/scheduleData';
-import {
-  findNextScheduledOpPoint,
```
SharglutDev, line 21: We can remove this function from utils and its tests

front/src/modules/timesStops/types.ts
```diff
@@ -40,13 +40,4 @@ export enum TableType {
 
 export type PathStepOpPointCorrespondance = { correspondingOpPointIndex: number };
```
SharglutDev: Since we removed `findNextScheduledOpPoint`, we can remove this type

SharglutDev (commented):
Thank you for this great refacto, it's really easier to handle this hook now :)

Left some comments, feel free to contact me. to discuss some of them

SharglutDev (commented):
I also have a weird thing here

Capture d’écran 2024-11-22 à 16 19 27

The row shouldn't be orange for 1sec difference or do I miss something ?

front/src/modules/timesStops/TimesStopsOutput.tsx
clarani: No I can't, because we will have the issue "Not the same number of hooks at each re-render" :/

SharglutDev: Oh right forgot that

front/src/modules/timesStops/helpers/computeMargins.ts
clarani: I corrected the comment above this block of code :+1:

front/src/modules/timesStops/helpers/__tests__/scheduleData.spec.ts
clarani: Hum good question :/

front/src/utils/date.ts
clarani: Yes indeed. I would like to remove all the utils which manipulate strings, we should always manipulate Dates.
Here is an issue for this: https://github.com/OpenRailAssociation/osrd/issues/9835

clarani: You mean we may have either `HHMM` or `HHMMSS` ?

front/src/modules/timesStops/hooks/useOutputTableData.ts
clarani: Done !

front/src/modules/timesStops/hooks/useOutputTableData.ts
clarani: You're right, done !

front/src/modules/timesStops/hooks/useOutputTableData.ts
clarani: I moved the `startDatetime` at the top of the hook but as I can't move the loader before the hook, I added some tests to check that it's not `undefined`

front/src/modules/timesStops/TimesStopsOutput.tsx
clarani: Done !

front/src/modules/timesStops/hooks/useOutputTableData.ts
clarani: Done

front/src/modules/timesStops/helpers/arrivalTime.ts
clarani: Hum you're right, I changed the function a bit

front/src/modules/timesStops/helpers/arrivalTime.ts
clarani: You're right, thanks ! :pray:

front/src/modules/timesStops/hooks/useOutputTableData.ts
clarani: I replaced the for loop by a map, but I had to keep the previous time, because I don't have the arrival or departure in the `path[index-1]` since it's only `pathStep` and not `pathStepRow` :/

I think it's cleaner to keep the `previousTime`, if it's ok for you

front/src/modules/timesStops/hooks/useOutputTableData.ts
clarani: Yes, I agree, it's not very robust. Do you think we can try to solve this issue in an other PR ? This is the existing behaviour :/

front/src/modules/timesStops/helpers/__tests__/scheduleData.spec.ts
SharglutDev: The expected result is in UTC ? Couldn't we use dayjs to compute the expected result so it stays consistent even when changing hour ? Or just the Date api and extracting HHMMSS with our utils.

front/src/modules/timesStops/TimesStopsOutput.tsx
SharglutDev: Can't reproduce this issue if I switch the `if`. How did you do ?

front/src/utils/date.ts
SharglutDev: Exactly :)

front/src/modules/timesStops/hooks/useOutputTableData.ts
SharglutDev: Yes works for me thank you :)

front/src/modules/timesStops/helpers/arrivalTime.ts
```diff
@@ -1,26 +1,63 @@
...
+    arrivalInSeconds = 0;
+  } else if (schedule?.arrival) {
+    arrivalInSeconds = ISO8601Duration2sec(schedule.arrival); // duration from startTime
+  }
+
+  if (arrivalInSeconds !== undefined) {
+    theoriticalArrival = new Date(startDatetime.getTime() + arrivalInSeconds * 1000);
+    const result = computeDayTimeFromStartTime(startDatetime, arrivalInSeconds, refDate);
+    arrival = result.timeExtraDay;
+    refDate = result.previousTime;
```
SharglutDev:
```suggestion
const { timeExtraDay, previousTime } = computeDayTimeFromStartTime(
      startDatetime,
      arrivalInSeconds,
      refDate
    );
    arrival = timeExtraDay;
    refDate = previousTime;
```

front/src/modules/timesStops/hooks/useOutputTableData.ts
SharglutDev: Yes sure :)

front/src/modules/timesStops/helpers/__tests__/arrivalTime.spec.ts
clarani: Yes you're right, it was a mistake, I fixed it :)

front/src/utils/date.ts
clarani: I renamed the function and also change the default value of withSeconds (it's now true)

front/src/modules/timesStops/hooks/useOutputTableData.ts
clarani: Here is the issue: https://github.com/OpenRailAssociation/osrd/issues/9864

front/src/modules/timesStops/helpers/__tests__/scheduleData.spec.ts
SharglutDev: Updated the util by removing `arrival` which is not used in it

front/src/modules/timesStops/TimesStopsOutput.tsx
SharglutDev: There is still a warning if we switch the condition so better to it leave it like it is

front/src/utils/date.ts
clarani: Actually, I renamed the props `withoutSeconds` and set it to `false` by default

front/src/modules/timesStops/helpers/arrivalTime.ts
clarani: Done

SharglutDev (approved):
Lgtm and tested, great job !

front/src/modules/timesStops/helpers/__tests__/computeMargins.spec.ts
```diff
@@ -1,172 +1,50 @@
...
-      diffMargins: '0 s',
-    });
-  });
-});
-
-describe('findNextScheduledOpPoint', () => {
-  const operationalPoints = [
-    {
-      id: 'd9c92cb4-tsrn-x',
-      name: 'Gondor',
```
Synar, line 55: Gondor has fallen =/

front/src/modules/timesStops/helpers/arrivalTime.ts
```diff
@@ -1,26 +1,67 @@
...
+  return { timeExtraDay, previousTime: arrivalDatetime };
+};
+
+export const computeInputDatetimes = (
+  startDatetime: Date,
+  lastReferenceDate: Date,
+  schedule: ScheduleEntry | undefined,
+  { isDeparture }: { isDeparture: boolean }
+) => {
+  let theoriticalArrival: Date | undefined;
```
Synar: Typo : theoreticalArrival

Synar (approved):
Tested and LGTM, great work!

front/src/modules/timesStops/helpers/arrivalTime.ts
clarani: Done :white_check_mark:

front/tests/assets/operationStudies/timesAndStops/expectedOutputsCellsData.json
Maymanaf, line 1: You haven’t made any changes to related to pathfinding, right?
Then why have the margin values changed?

Maymanaf (approved):
LGTM for e2e tests

PR #9645, opened 2024-11-07
editoast: rename all test modules to follow rust idiom

Sh099078:
While the majority of editoast test modules are called "tests", some are mistakenly called "test". Rename all test modules to "tests" so that they follow rust idiom.


leovalais (approved):
LGTM, thanks!

PR #9646, opened 2024-11-07
docker: fix FromAsCasing warning in dockerfiles

Synar:
Fix the warning generated by docker when from and as casing do not match in dockerfiles, see https://docs.docker.com/reference/build-checks/from-as-casing/


Synar:
I will try and draft another pr soon to detect docker build --check warnings in the CI

emersion (approved):

Khoyo (approved):

PR #9647, opened 2024-11-08
core: docker: add profiles to single-worker compose

bougue-pe:
This will avoid dealing with unwanted services (especially at build)

### DIFF
--- a/core/README.md
+++ b/core/README.md
@@ -97,22 +97,28 @@ Using a specific script (just through `docker compose` CLI and a set of docker-c
 allows to run a single core worker for all infra on localhost network:
 
 ```sh
-./scripts/single-worker-compose.sh up -d
+# run all services:
+./scripts/single-worker-compose.sh --profile "*" up -d
 
 # or exclude 'core' service straight away:
-./scripts/single-worker-compose.sh up -d --scale core=0
+./scripts/single-worker-compose.sh --profile with-editoast up -d
+
+# one can also only exclude 'editoast' service
+./scripts/single-worker-compose.sh --profile with-core up -d
+
+# Please have a look at script or docker-compose file for advanced use.
 ```
 
 Then, it is easy to replace the desired component for debug purpose. \
 For core:
 
 ```sh
-./scripts/single-worker-compose.sh down core # if 'core' is running
+./scripts/single-worker-compose.sh  --profile "*" down core # if 'core' is running
 ./gradlew shadowJar && ALL_INFRA=true java -jar -ea -Xmx12g -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=dump.hprof build/libs/osrd-all.jar worker --editoast-url http://localhost:8090/
 ```
 
 Clean or restart the whole stack can be necessary sometimes and is also available
 through docker compose CLI (the following wipes the database too):
 ```sh
-./scripts/single-worker-compose.sh down -v
+./scripts/single-worker-compose.sh  --profile "*" down -v
 ```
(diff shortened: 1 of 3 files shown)


core/README.md
```diff
@@ -97,22 +97,23 @@ Using a specific script (just through `docker compose` CLI and a set of docker-c
 allows to run a single core worker for all infra on localhost network:
 
 ```sh
-./scripts/single-worker-compose.sh up -d
+# run all services
+./scripts/single-worker-compose.sh --profile "*" up -d
 
 # or exclude 'core' service straight away:
-./scripts/single-worker-compose.sh up -d --scale core=0
+./scripts/single-worker-compose.sh --profile with-editoast up -d
```
Erashin, line 104: Might be worth to add the `with-core` option as an example, since they are the only 2 services with this option, what do you think?

scripts/single-worker-compose.sh
```diff
@@ -6,6 +6,12 @@
 set -e
 
 # We override the default docker-compose file with "host" and "single-worker" versions in the docker folder
+
+# From docker/docker-compose.single-worker.yml, profiles are provided:
+# - no profile means all other service (except editoast and core)
```
Erashin:
```suggestion
# - no profile means all other services (except editoast and core)
```

Erashin (approved):

scripts/single-worker-compose.sh
bougue-pe: [Done](https://github.com/OpenRailAssociation/osrd/compare/64521221fc435ad22284e8073b3c469b7fe60320..5d67e80576cc800d26f8271ecb42d0ba4e1cf1e1#diff-dac8efaa7efba07d4bf82107ea44f6a4a7abce0d8eac0c78f23fcecee45469efR11)

core/README.md
bougue-pe: Indeed, more explicit readme pushed after your suggestion: https://github.com/OpenRailAssociation/osrd/compare/64521221fc435ad22284e8073b3c469b7fe60320..5d67e80576cc800d26f8271ecb42d0ba4e1cf1e1#diff-ea08f26e820e40e902af500eb960c815bbc5b5f23bacdc9e884cee42374024cb
Note that currently, `--profile with-editoast --profile with-core` == `--profile "*"`.

ghost:
It might better to hold off on using Docker Compose profiles for now. There is another work on those subjects in progress where we'll create multiple YAML files and a script that can combine them using stacked -f arguments. This approach better suits our needs since our setup involves mixing different configurations:

- Container networking (Network mode vs Host mode)
- Architecture (OSRDyne vs Single Worker)
- Frontend (Embedded vs Development mode)

While our default configuration will be 'Network mode + OSRDyne + Embedded frontend', using profiles for this setup presents several challenges:

- Some containers would conflict with each other
- Certain containers would only need to exist when specific configurations are combined (for example, the gateway container behaves differently depending on both network mode and frontend type)

@Khoyo and I are currently working on this. We might revisit using profiles after we complete our cleanup work, but for now, I suggest putting this PR on hold while we rework the scripts."

bougue-pe:
OK, fine to hold off on this as discussed with core's devs.
The need covered by this was to avoid build/manage core (`scale` is not so clean), which is a nice-to-have anyway: just sharing that this is a bonus in case it helps for some choices.

(I was thinking it would impact a small domain so it was OK, but it's way better to make it all consistent 👍)

Closing this, we can easily reopen something like this later anyway.

Side-note: not so fan of the way the `profiles` work for this use-case (but it was still an improvement IMO).

PR #9653, opened 2024-11-08
front: fix stdcm passage

### LINKED ISSUE #9655: front: stdcm stops at all vias, even if they are only "passage"
### What happened?

Try to create a stdcm with a via "PASSAGE" in debug mode. Check the GEV: the train stops at the via.

The train should not stop if the via is only a waypoint and not a stop.

1. Use debug mode
2. Create a stdcm with a via (select Passage for the stop type)
3. Check the GEV

Recette (SNCF)

Firefox

.

clarani:
closes #9655


emersion:
Ouch :/ would it really be that much work to fix in editoast instead?

clarani:
Right now, we just need a valid image so @bloussou can deploy on prod :/ I can ask on Element if someone can fix the bug in editoast

emersion:
Alternative fix for the first commit (edit: that commit has now been dropped): https://github.com/OpenRailAssociation/osrd/pull/9656

clarani:
I removed the last commit, related to the work schedules

Akctarus (changes requested):
Tested, but got this error message: `Last step of stdcm request needs to be a stop`

last step duration is equal to `undefined`, and the request fails because of this. We'd have to add a `duration = 0` to this step

clarani:
You're right, it's fixed

Akctarus (approved):
Tested, LGTM !

Math-R (approved):
Lgtm and tested !

PR #9654, opened 2024-11-08
editoast: lmr towed max speed

### LINKED ISSUE #9637: Towed rolling stocks with max speed at `0km/h` mess up the consist building
### What happened?

The consists are build with the formula:
`max_speed = min(max_speed_traction_engine, max_speed_towed)`

but when there's no data for the max speed towed the value is actually 0km/h, so the consist has a 0km/h value ⚙ 

We should not take into account max_speed of towed rolling stocks if they are at 0km/h

.

Recette (SNCF), Dev (SNCF)

Google Chrome

b74fd59

Wadjetz:
fix [#9637](https://github.com/OpenRailAssociation/osrd/issues/9637)

- [x] Add stdcm request validation
- [x] Handle max speed zero value in simulation computation (if it used in another place without request validation)
- [x] Add max_speed to towed rolling stock

The frontend validation done in another PR

### DIFF
--- a/editoast/src/core/simulation.rs
+++ b/editoast/src/core/simulation.rs
@@ -114,10 +114,17 @@ impl PhysicsConsistParameters {
     }
 
     pub fn compute_max_speed(&self) -> f64 {
-        self.max_speed
-            .map(|max_speed_parameter| {
-                f64::min(self.traction_engine.max_speed, max_speed_parameter)
-            })
+        let max_speeds = [
+            self.max_speed,
+            self.towed_rolling_stock
+                .as_ref()
+                .and_then(|towed| towed.max_speed),
+            Some(self.traction_engine.max_speed),
+        ];
+        max_speeds
+            .into_iter()
+            .flatten()
+            .reduce(f64::min)
             .unwrap_or(self.traction_engine.max_speed)
     }
 
@@ -555,6 +562,7 @@ mod tests {
 
     #[test]
     fn physics_consist_max_speed() {
+        // Towed max speed 35
         let mut physics_consist = create_physics_consist();
         physics_consist.max_speed = Some(20.0); // m/s
         physics_consist.traction_engine.max_speed = 22.0; // m/s
@@ -569,6 +577,12 @@ mod tests {
 
         physics_consist.max_speed = None;
         assert_eq!(physics_consist.compute_max_speed(), 24.0);
+
+        physics_consist.traction_engine.max_speed = 40.0; // m/s
+        assert_eq!(physics_consist.compute_max_speed(), 35.0);
+
+        physics_consist.towed_rolling_stock = None;
+        assert_eq!(physics_consist.compute_max_speed(), 40.0);
     }
 
     #[test]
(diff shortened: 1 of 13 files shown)


editoast/src/core/simulation.rs
```diff
@@ -112,11 +112,13 @@ impl PhysicsConsistParameters {
...
-            .map(|max_speed_parameter| {
+        match self.max_speed {
+            Some(0.0) => self.traction_engine.max_speed,
+            Some(max_speed_parameter) => {
                 f64::min(self.traction_engine.max_speed, max_speed_parameter)
-            })
-            .unwrap_or(self.traction_engine.max_speed)
+            }
+            None => self.traction_engine.max_speed,
+        }
```
woshilapin: You can express that with a filter.

```rust
self.max_speed
  .filter(|max_speed_parameter| *max_speed_parameter != 0.0)
  .map(|max_speed_parameter| f64::min(self.traction_engine.max_speed, max_speed_parameter))
  .unwrap_or(self.traction_engine.max_speed)
```

editoast/src/views/timetable/stdcm.rs
```diff
@@ -124,10 +125,13 @@ pub struct STDCMRequestPayload {
     #[schema(value_type = Option<String>, example = json!(["5%", "2min/100km"]))]
     margin: Option<MarginValue>,
     /// Total mass of the consist in kg
+    #[validate(range(min = 1.0))]
```
woshilapin: Where does this `1.0` value comes from. I believe the ticket only mentions excluding `0.0`. You should be able to do that with the following.

```suggestion
    #[validate(range(exclusive_min = 0.0))]
```

editoast/src/views/timetable/stdcm.rs
```diff
@@ -124,10 +125,13 @@ pub struct STDCMRequestPayload {
...
     total_mass: Option<f64>,
     /// Total length of the consist in meters
+    #[validate(range(min = 1.0))]
```
```suggestion
    #[validate(range(exclusive_min = 0.0))]
```

editoast/src/views/timetable/stdcm.rs
```diff
@@ -124,10 +125,13 @@ pub struct STDCMRequestPayload {
...
     total_length: Option<f64>,
     /// Maximum speed of the consist in km/h
+    #[validate(range(min = 1.0))]
```
```suggestion
    #[validate(range(exclusive_min = 0.0))]
```

woshilapin (commented):
The code is fine (only some minor comments), but I'm questioning the original ticket https://github.com/OpenRailAssociation/osrd/issues/9637 in which I made a comment. I propose to not merge the current PR until we do have an answer.

hamz2a (approved):
LGTM, thanks!

flomonster (approved):
LGTM. Can you fixup your commits before merge.

woshilapin (changes requested):
I don't think we should filter the zero, for the same reason I mentioned in https://github.com/OpenRailAssociation/osrd/issues/9637#issuecomment-2464324459. The fact that the field is optional should be enough.

woshilapin (approved):
Thank you.

leovalais:
Ping @Tristramg: changes in rolling stock speed calculations, probably best to rebase before typing quantities any further :)

SharglutDev (approved):
Lgtm for front (not tested)

PR #7049, opened 2024-03-28
editoast: reorganize all 'use' as unique items

woshilapin:
Can be configured with `rustfmt.toml` with the following options:

```
unstable_features = true
group_imports = "StdExternalCrate"
imports_granularity = "Item"
```

Then `cargo +nightly fmt`.

Note that we do not enforce that styling in CI for now since that would impose using `nightly` (necessary for unstable features) on developers.

PR #9656, opened 2024-11-08
editoast: use DateTime with timezone in work schedule API

### LINKED ISSUE #9652: Work schedule projection does not contain timezone
### What happened?

Dates returned in `work_schedule` projection does not contain time zone even if the timezone is specified in the database.

Have a `Z` at the end of the work schedule_date or the explicit timezone for the frontend

1. create a work schedule and open stdcm in debug mode

Recette (SNCF)

Firefox

v0.3.4

emersion:
The timezone was stored in the database, but was stripped at the HTTP API boundary.

This is a breaking change: the timezone is required when submitting a work schedule, and is always returned when projecting a work schedule.

Closes: https://github.com/OpenRailAssociation/osrd/issues/9652

### DIFF
--- a/editoast/src/models/work_schedules.rs
+++ b/editoast/src/models/work_schedules.rs
@@ -1,4 +1,5 @@
-use chrono::NaiveDateTime;
+use chrono::DateTime;
+use chrono::Utc;
 use editoast_derive::Model;
 use editoast_schemas::infra::TrackRange;
 use strum::FromRepr;
@@ -12,7 +13,7 @@ use utoipa::ToSchema;
 #[model(gen(ops = crd, batch_ops = c, list))]
 pub struct WorkScheduleGroup {
     pub id: i64,
-    pub creation_date: NaiveDateTime,
+    pub creation_date: DateTime<Utc>,
     pub name: String,
 }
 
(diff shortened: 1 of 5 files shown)


editoast/src/models/work_schedules.rs
```diff
@@ -1,4 +1,4 @@
-use chrono::NaiveDateTime;
+use chrono::{DateTime, Utc};
```
woshilapin: nit: we're trying to unmerge the `use` as much as possible (easier for the diff). We can't really automate it without it being a pain for developers though... (see https://github.com/OpenRailAssociation/osrd/pull/7049).

woshilapin (approved):
I sincerely like that change ❤️

emersion:
If this is deemed too intrusive for a bugfix, we could fix only the GET side of the API as a first step, and take more time to merge the POST side (and adapt the various users).

editoast/src/models/work_schedules.rs
emersion: Ah, I see. Split into two lines.

flomonster (approved):
LGTM

bloussou (approved):
Check with @Morgane-SJK no problem we are already using TZ everywhere in elt code

PR #9659, opened 2024-11-08
core: add a new updateMap() function to DistanceRangeMap

Khoyo:
The old one is now called updateMapIntersection(). The new one inserts the non overlapping ranges without modification.

### DIFF
--- a/core/kt-osrd-utils/src/main/kotlin/fr/sncf/osrd/utils/DistanceRangeMap.kt
+++ b/core/kt-osrd-utils/src/main/kotlin/fr/sncf/osrd/utils/DistanceRangeMap.kt
@@ -52,12 +52,25 @@ interface DistanceRangeMap<T> : Iterable<DistanceRangeMap.RangeMapEntry<T>> {
 
     /**
      * Updates the map with another one, using a merge function to fuse the values of intersecting
-     * ranges
+     * ranges. Doesn't keep any range from update where there is no intersection.
      */
-    fun <U> updateMap(update: DistanceRangeMap<U>, updateFunction: BiFunction<T, U, T>)
+    fun <U> updateMapIntersection(update: DistanceRangeMap<U>, updateFunction: BiFunction<T, U, T>)
+
+    /**
+     * Updates the map with another one, using a merge function to fuse the values of intersecting
+     * ranges. Calls default on the values of the ranges from update where there is no intersection.
+     */
+    fun updateMap(
+        update: DistanceRangeMap<T>,
+        updateFunction: (T, T) -> T,
+        default: (T) -> T = { it }
+    )
 
     /** Returns true if there is no entry at all */
     fun isEmpty(): Boolean
+
+    /** Clear the map */
+    fun clear()
 }
 
 fun <T> distanceRangeMapOf(
--- a/core/kt-osrd-utils/src/main/kotlin/fr/sncf/osrd/utils/DistanceRangeMapImpl.kt
+++ b/core/kt-osrd-utils/src/main/kotlin/fr/sncf/osrd/utils/DistanceRangeMapImpl.kt
@@ -190,12 +190,113 @@ data class DistanceRangeMapImpl<T>(
         return bounds.isEmpty()
     }
 
-    override fun <U> updateMap(update: DistanceRangeMap<U>, updateFunction: BiFunction<T, U, T>) {
+    override fun <U> updateMapIntersection(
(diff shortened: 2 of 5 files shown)


core/src/test/kotlin/fr/sncf/osrd/utils/DistanceRangeMapImplTest.kt
```diff
@@ -0,0 +1,214 @@
...
+        val update = DistanceRangeMapImpl<String>()
+        update.put(10.0.meters, 15.0.meters, "B")
+        map.updateMap(update) { old, new -> old + new }
+        assertEquals("A", map.get(2.5.meters))
+        assertEquals("B", map.get(12.5.meters))
+        assertNull(map.get(7.5.meters))
+    }
+
+    @Test
+    fun updateMap_partialOverlap() {
```
Erashin: Duplicate of test `updateMap()`.

core/kt-osrd-utils/src/main/kotlin/fr/sncf/osrd/utils/DistanceRangeMap.kt
```diff
@@ -52,12 +52,21 @@ interface DistanceRangeMap<T> : Iterable<DistanceRangeMap.RangeMapEntry<T>> {
...
+     * ranges. Doesn't keep any range from update where there is no intersection.
      */
-    fun <U> updateMap(update: DistanceRangeMap<U>, updateFunction: BiFunction<T, U, T>)
+    fun <U> updateMapIntersection(update: DistanceRangeMap<U>, updateFunction: BiFunction<T, U, T>)
+
+    /**
+     * Updates the map with another one, using a merge function to fuse the values of intersecting
+     * ranges. Keeps the values of the ranges from update where there is no intersection.
+     */
+    fun updateMap(update: DistanceRangeMap<T>, updateFunction: BiFunction<T, T, T>)
```
bougue-pe: This would feel more natural to me to have an `updateFunction`  that's always applied:
```suggestion
    fun updateMap(update: DistanceRangeMap<T>, updateFunction: BiFunction<T?, T?, T>)
```
The current way feels like a trap (documented, so it's still OK I guess) if the updateFunction is something like `a^2+b^2`.

core/src/test/kotlin/fr/sncf/osrd/utils/DistanceRangeMapImplTest.kt
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    @Test
+    fun updateMapIntersection() {
+        val map = DistanceRangeMapImpl<String>()
+        map.put(0.0.meters, 10.0.meters, "A")
+        val updateMap = DistanceRangeMapImpl<String>()
+        updateMap.put(5.0.meters, 15.0.meters, "B")
+        map.updateMapIntersection(updateMap) { old, new -> old + new }
+        assertEquals("AB", map.get(7.5.meters))
```
Erashin: Add the following to test that only the intersection is changed, and the values that are not intersecting in the update map are not added:
`assertEquals("A", map.get(2.5.meters))`
`assertNull(map.get(12.5.meters))`

core/src/test/kotlin/fr/sncf/osrd/utils/DistanceRangeMapImplTest.kt
bougue-pe, line 1: Should this be merged with https://github.com/OpenRailAssociation/osrd/blob/dev/core/kt-osrd-utils/src/test/kotlin/fr/sncf/osrd/utils/TestDistanceRangeMap.kt ?

Erashin (approved):
Good job, gets the work done :)

core/src/test/kotlin/fr/sncf/osrd/utils/DistanceRangeMapImplTest.kt
Khoyo: I completely missed that file (and intellij did too)

core/kt-osrd-utils/src/main/kotlin/fr/sncf/osrd/utils/DistanceRangeMap.kt
Khoyo: I took the middle ground where you can provide an additional mapping function for non overlapping ranges.

The usability of implementing `if (a == null) {a} else if (b == null) {b} else{ /*...*/}` is quite low for something I'm not even sure we will use

PR #9660, opened 2024-11-08
Use consists parameters to train simulation request

### LINKED ISSUE #9544: stdcm: towed rolling stock parameters aren't used for preprocessing simulations
### What happened?

When editoast receives an stdcm request, it first asks core for a simulation. This simulation is used to check that the train is physically capable of reaching its destination, as well as to have good parameters (max running time = 2 * fastest time).

This simulation uses the unedited rolling stock values.

The stdcm search then uses different values, which can result in mismatches, no solution found, and misleading errors.

The values of rolling stock sent to core for pre-processing should be the total consist and not the traction engine.

Check the values forwarded to core

Local, Dev (SNCF)

Firefox

a9261bc7d6a18fb28a9a73a214f7c901ca7e3988

Wadjetz:
fix [#9544](https://github.com/OpenRailAssociation/osrd/issues/9544)

### DIFF
--- a/editoast/src/core/simulation.rs
+++ b/editoast/src/core/simulation.rs
@@ -1,6 +1,9 @@
 use std::collections::BTreeMap;
 use std::collections::HashMap;
+use std::hash::Hash;
 
+use derivative::Derivative;
+use editoast_schemas::primitives::Identifier;
 use editoast_schemas::rolling_stock::EffortCurves;
 use editoast_schemas::rolling_stock::RollingResistance;
 use editoast_schemas::rolling_stock::RollingStock;
@@ -18,9 +21,6 @@ use super::pathfinding::TrackRange;
 use crate::core::{AsCoreRequest, Json};
 use crate::error::InternalError;
 use crate::views::path::pathfinding::PathfindingFailure;
-use derivative::Derivative;
-use editoast_schemas::primitives::Identifier;
-use std::hash::Hash;
 
 editoast_common::schemas! {
     CompleteReportTrain,
@@ -84,7 +84,7 @@ pub struct PhysicsConsistParameters {
 }
 
 impl PhysicsConsistParameters {
-    pub fn with_traction_engine(traction_engine: RollingStock) -> Self {
+    pub fn from_traction_engine(traction_engine: RollingStock) -> Self {
         PhysicsConsistParameters {
             max_speed: None,
             total_length: None,
--- a/editoast/src/views/path/pathfinding.rs
+++ b/editoast/src/views/path/pathfinding.rs
@@ -10,7 +10,9 @@ use axum::extract::State;
 use axum::Extension;
 use editoast_authz::BuiltinRole;
 use editoast_schemas::rolling_stock::LoadingGaugeType;
+use editoast_schemas::rolling_stock::RollingStock;
 use editoast_schemas::train_schedule::PathItemLocation;
+use itertools::Itertools;
 use ordered_float::OrderedFloat;
 use serde::Deserialize;
 use serde::Serialize;
@@ -336,24 +338,19 @@ pub async fn pathfinding_from_train(
     infra: &Infra,
     train_schedule: TrainSchedule,
 ) -> Result<PathfindingResult> {
-    let rolling_stocks =
+    let rolling_stock: Vec<RollingStock> =
         RollingStockModel::retrieve(conn, train_schedule.rolling_stock_name.clone())
             .await?
             .into_iter()
-            .map(|rs| (rs.name.clone(), rs))
+            .map_into()
             .collect();
 
-    Ok(pathfinding_from_train_batch(
-        conn,
-        valkey,
-        core,
-        infra,
-        &[train_schedule],
-        &rolling_stocks,
+    Ok(
+        pathfinding_from_train_batch(conn, valkey, core, infra, &[train_schedule], &rolling_stock)
+            .await?
+            .pop()
+            .unwrap(),
     )
-    .await?
-    .pop()
-    .unwrap())
 }
 
 /// Compute a path given a batch of trainschedule and an infrastructure.
@@ -363,7 +360,7 @@ pub async fn pathfinding_from_train_batch(
     core: Arc<CoreClient>,
     infra: &Infra,
     train_schedules: &[TrainSchedule],
-    rolling_stocks: &HashMap<String, RollingStockModel>,
+    rolling_stocks: &[RollingStock],
 ) -> Result<Vec<PathfindingResult>> {
     let mut results = vec![
         PathfindingResult::Failure(PathfindingFailure::PathfindingInputError(
@@ -371,12 +368,15 @@ pub async fn pathfinding_from_train_batch(
         ));
         train_schedules.len()
     ];
+
+    let rolling_stocks: HashMap<_, _> = rolling_stocks.iter().map(|rs| (&rs.name, rs)).collect();
+
     let mut to_compute = vec![];
     let mut to_compute_index = vec![];
     for (index, train_schedule) in train_schedules.iter().enumerate() {
         // Retrieve rolling stock
         let rolling_stock_name = &train_schedule.rolling_stock_name;
-        let Some(rolling_stock) = rolling_stocks.get(rolling_stock_name).cloned() else {
+        let Some(rolling_stock) = rolling_stocks.get(rolling_stock_name) else {
             let rolling_stock_name = rolling_stock_name.clone();
             results[index] = PathfindingResult::Failure(PathfindingFailure::PathfindingInputError(
                 PathfindingInputError::RollingStockNotFound { rolling_stock_name },
@@ -387,9 +387,14 @@ pub async fn pathfinding_from_train_batch(
         // Create the path input
         let path_input = PathfindingInput {
             rolling_stock_loading_gauge: rolling_stock.loading_gauge,
-            rolling_stock_is_thermal: rolling_stock.has_thermal_curves(),
-            rolling_stock_supported_electrifications: rolling_stock.supported_electrification(),
-            rolling_stock_supported_signaling_systems: rolling_stock.supported_signaling_systems.0,
+            rolling_stock_is_thermal: rolling_stock.effort_curves.has_thermal_curves(),
+            rolling_stock_supported_electrifications: rolling_stock
+                .effort_curves
+                .supported_electrification(),
+            rolling_stock_supported_signaling_systems: rolling_stock
+                .supported_signaling_systems
+                .0
+                .clone(),
             rolling_stock_maximum_speed: OrderedFloat(rolling_stock.max_speed),
             rolling_stock_length: OrderedFloat(rolling_stock.length),
             path_items: train_schedule
--- a/editoast/src/views/timetable/stdcm.rs
+++ b/editoast/src/views/timetable/stdcm.rs
@@ -45,7 +45,7 @@ use crate::models::train_schedule::TrainSchedule;
 use crate::models::Infra;
 use crate::models::RollingStockModel;
 use crate::views::path::pathfinding::PathfindingResult;
-use crate::views::train_schedule::train_simulation;
+use crate::views::train_schedule::consist_train_simulation_batch;
 use crate::views::train_schedule::train_simulation_batch;
 use crate::views::AuthenticationExt;
 use crate::views::AuthorizationError;
@@ -92,6 +92,8 @@ enum StdcmError {
     RollingStockNotFound { rolling_stock_id: i64 },
     #[error("Towed rolling stock {towed_rolling_stock_id} does not exist")]
     TowedRollingStockNotFound { towed_rolling_stock_id: i64 },
+    #[error("Train simulation fail")]
+    TrainSimulationFail,
     #[error("Path items are invalid")]
     InvalidPathItems { items: Vec<InvalidPathItem> },
 }
@@ -167,7 +169,19 @@ async fn stdcm(
                 rolling_stock_id: stdcm_request.rolling_stock_id,
             }
         })
-        .await?;
+        .await?
+        .into();
+
+    let physics_consist_parameters = PhysicsConsistParameters {
+        max_speed: stdcm_request.max_speed,
+        total_length: stdcm_request.total_length,
+        total_mass: stdcm_request.total_mass,
+        towed_rolling_stock: stdcm_request
+            .get_towed_rolling_stock(&mut conn)
+            .await?
+            .map(From::from),
+        traction_engine: rolling_stock,
+    };
 
     // 2. Compute the earliest start time and maximum departure delay
     let virtual_train_run = VirtualTrainRun::simulate(
@@ -176,7 +190,7 @@ async fn stdcm(
         core_client.clone(),
         &stdcm_request,
         &infra,
-        &rolling_stock,
+        &physics_consist_parameters,
         timetable_id,
     )
     .await?;
@@ -219,21 +233,12 @@ async fn stdcm(
     let stdcm_request = crate::core::stdcm::Request {
         infra: infra.id,
         expected_version: infra.version.clone(),
-        rolling_stock_loading_gauge: rolling_stock.loading_gauge,
-        rolling_stock_supported_signaling_systems: rolling_stock
+        rolling_stock_loading_gauge: physics_consist_parameters.traction_engine.loading_gauge,
+        rolling_stock_supported_signaling_systems: physics_consist_parameters
+            .traction_engine
             .supported_signaling_systems
             .clone(),
-        physics_consist: PhysicsConsistParameters {
-            max_speed: stdcm_request.max_speed,
-            total_length: stdcm_request.total_length,
-            total_mass: stdcm_request.total_mass,
-            towed_rolling_stock: stdcm_request
-                .get_towed_rolling_stock(&mut conn)
-                .await?
-                .map(From::from),
-            traction_engine: rolling_stock.into(),
-        }
-        .into(),
+        physics_consist: physics_consist_parameters.into(),
         temporary_speed_limits: stdcm_request
             .get_temporary_speed_limits(&mut conn, simulation_run_time)
             .await?,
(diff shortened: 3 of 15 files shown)


editoast/src/views/train_schedule.rs
```diff
@@ -358,66 +359,40 @@ async fn simulation(
...
 /// Note: The order of the returned simulations is the same as the order of the train schedules.
-pub async fn train_simulation_batch(
+pub async fn consist_train_simulation_batch(
     conn: &mut DbConnection,
     valkey_client: Arc<ValkeyClient>,
     core: Arc<CoreClient>,
     train_schedules: &[TrainSchedule],
     infra: &Infra,
     electrical_profile_set_id: Option<i64>,
+    physics_consist_parameters: PhysicsConsistParameters,
```
flomonster: This parameter must be set for each train schedule, not globally.

The `consist_train_simulation_batch` should be the function that actually do the simulation.

editoast/src/views/train_schedule.rs
```diff
@@ -428,6 +403,47 @@ pub async fn train_simulation_batch(
...
+
+async fn get_train_batch_simulation(
+    valkey_conn: &mut ValkeyConnection,
+    core: Arc<CoreClient>,
+    infra: &Infra,
+    train_schedules: &[TrainSchedule],
+    rolling_stocks: HashMap<String, RollingStockModel>,
+    pathfinding_results: Vec<PathfindingResult>,
+    electrical_profile_set_id: Option<i64>,
+    physics_consist_parameters: Option<PhysicsConsistParameters>,
```
flomonster:
```suggestion
    train_schedules: &[TrainSchedule],
    pathfinding_results: Vec<PathfindingResult>,
    electrical_profile_set_id: Option<i64>,
```

The `get_train_batch_simulation` should simply build the PhysicsConsistParameters from the rolling stock names then call the `consist_train_simulation_batch`

editoast/src/views/train_schedule.rs
```diff
@@ -524,13 +548,53 @@ pub async fn train_simulation_batch(
         .collect())
 }
 
+/// Compute in batch the simulation of a list of train schedule
+///
+/// Note: The order of the returned simulations is the same as the order of the train schedules.
+pub async fn train_simulation_batch(
```
flomonster: The `train_simulation_batch` should simply build the PhysicsConsistParameters from the rolling stock names then call the `consist_train_simulation_batch`

flomonster (commented):
You have three functions (`train_simulation_batch`, `get_train_batch_simulation` and `consist_train_simulation_batch`). Only two of them are required.

- `train_simulation_batch(&[TrainSchedule])`
    - Retrieve rolling stocks from `rolling_stock_name `field`
    - Convert this rolling stocks into `PhysicsConsistParameters`
    - Call `consist_train_simulation_batch`
- `consist_train_simulation_batch(&[TrainScheduleWithPhysicsConsist])`
    - Check valkey cache
    - Call the pathfinding
    - Cal the simulation 
    - Cache the response
    - Return the response

editoast/src/models/fixtures.rs
```diff
@@ -247,8 +247,10 @@ pub fn create_towed_rolling_stock() -> TowedRollingStock {
     }
 }
 
-pub fn create_simple_rolling_stock() -> RollingStock {
-    RollingStock {
+pub fn create_simple_rolling_stock() -> RollingStockModel {
+    RollingStockModel {
+        id: 1,
```
leovalais: Better to avoid hardcoding IDs as they can conflict with other entries. Besides since the function is called `create_simple_rolling_stock` we might as well `create` it for real and let Postgres assign it a unique ID.

editoast/src/core/simulation.rs
```diff
@@ -78,11 +78,16 @@ pub struct PhysicsConsistParameters {
     /// In m/s
     pub max_speed: Option<f64>,
     pub towed_rolling_stock: Option<TowedRollingStock>,
-    pub traction_engine: RollingStock,
+    pub traction_engine: RollingStockModel,
```
leovalais: Better avoid using the model here for two reasons:

1. Core shouldn't rely on how we store data in editoast, it probably should expect the RailJson schema instead.
2. The `models` module will depend on the `core` module as we'll store STDCM requests and responses. So using the model here will create a bidirectional dependency, resulting in more work when we'll get back to the split in crates.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -212,21 +229,12 @@ async fn stdcm(
     let stdcm_request = crate::core::stdcm::Request {
         infra: infra.id,
         expected_version: infra.version.clone(),
-        rolling_stock_loading_gauge: rolling_stock.loading_gauge,
+        rolling_stock_loading_gauge: rolling_stock.clone().loading_gauge,
         rolling_stock_supported_signaling_systems: rolling_stock
+            .clone()
             .supported_signaling_systems
             .clone(),
```
leovalais: Better avoid unnecessary cloning.

editoast/src/core/simulation.rs
```diff
@@ -78,11 +78,16 @@ pub struct PhysicsConsistParameters {
...
 }
 
 impl PhysicsConsistParameters {
-    pub fn with_traction_engine(traction_engine: RollingStock) -> Self {
+    pub fn set_traction_engine(mut self, traction_engine: RollingStockModel) -> Self {
```
leovalais: Naming is confusing here. We usually expect `set_xyz` methods to return `()` instead of `Self`.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -162,6 +163,17 @@ async fn stdcm(
...
 
+    let physics_consist_parameters = PhysicsConsistParameters {
+        max_speed: stdcm_request.max_speed,
+        total_length: stdcm_request.total_length,
+        total_mass: stdcm_request.total_mass,
+        towed_rolling_stock: stdcm_request
+            .get_towed_rolling_stock(conn)
+            .await?
+            .map(From::from),
+        traction_engine: rolling_stock.clone().into(),
```
leovalais: Can `PhysicsConsistParameter::traction_engine` can be used afterwards to avoid cloning the rolling stock?

editoast/src/views/train_schedule.rs
```diff
@@ -458,14 +480,15 @@ pub async fn train_simulation_batch(
...
+        let physics_consist_parameters = consists[&train_schedule.rolling_stock_name].clone();
+
         let simulation_request = build_simulation_request(
             infra,
             train_schedule,
             path_item_positions,
             path,
-            rolling_stock,
             electrical_profile_set_id,
+            physics_consist_parameters.clone(),
```
leovalais: Unnecessary clone

editoast/src/core/simulation.rs
flomonster: Why did you need to have a `RollingStockModel` here?

editoast/src/views/path/pathfinding.rs
```diff
@@ -363,7 +369,7 @@ pub async fn pathfinding_from_train_batch(
     core: Arc<CoreClient>,
     infra: &Infra,
     train_schedules: &[TrainSchedule],
-    rolling_stocks: &HashMap<String, RollingStockModel>,
+    consists: &HashMap<String, PhysicsConsistParameters>,
```
leovalais: Can this function take a `consists: &[PhysicsConsistParameters]`? That'd be more consistent with the rest of the signature and it would avoid unnecessary clones. A local hash map only containing references can be built internally.

hamz2a (approved):
LGTM, thanks!

editoast/src/views/path/pathfinding.rs
```diff
@@ -376,14 +382,16 @@ pub async fn pathfinding_from_train_batch(
     for (index, train_schedule) in train_schedules.iter().enumerate() {
         // Retrieve rolling stock
         let rolling_stock_name = &train_schedule.rolling_stock_name;
-        let Some(rolling_stock) = rolling_stocks.get(rolling_stock_name).cloned() else {
+        let Some(rolling_stock) = consists.get(rolling_stock_name).cloned() else {
```
leovalais:
```suggestion
        let Some(rolling_stock) = consists.get(rolling_stock_name) else {
```

editoast/src/views/train_schedule.rs
```diff
@@ -358,40 +358,38 @@ async fn simulation(
...
+}
+
+async fn retrieve_rolling_stocks_from_train_schedules(
+    conn: &mut DbConnection,
+    train_schedules: &[TrainSchedule],
+) -> Result<HashMap<String, RollingStockModel>> {
+    let rolling_stocks_ids = train_schedules
+        .iter()
+        .map::<String, _>(|t| t.rolling_stock_name.clone());
+    let (rolling_stocks, _): (Vec<_>, _) =
```
leovalais: Better use `retrieve_batch_or_fail` and deal with the error when ids are missing. If this situation can't happen, `retrieve_batch_unchecked` can be used with a comment explaining why it's safe to do so.

editoast/src/views/train_schedule.rs
```diff
@@ -358,40 +358,38 @@ async fn simulation(
...
+        RollingStockModel::retrieve_batch(conn, rolling_stocks_ids).await?;
+
+    let rolling_stocks: HashMap<_, _> = rolling_stocks
+        .into_iter()
+        .map(|rs| (rs.name.clone(), rs))
+        .collect();
```
leovalais: You can use `retrieve_batc_with_key_or_fail` (`_unchecked`?)

editoast/src/views/path/pathfinding.rs
flomonster: We don't need the consist parameters to compute pathfinding.

We can keep a map of `RollingStockModel`. Having a map of `RollingStock` makes more sense here.

editoast/src/views/train_schedule.rs
```diff
@@ -590,7 +613,7 @@ fn build_simulation_request(
         speed_limit_tag: train_schedule.speed_limit_tag.clone(),
         power_restrictions,
         options: train_schedule.options.clone(),
-        rolling_stock: PhysicsConsistParameters::with_traction_engine(rolling_stock.into()).into(),
+        rolling_stock: physics_consist_parameters.into(),
```
leovalais: Since the only use of this parameter is to `into()` it into something else, better use the type it's converted into in the function signature for readability;

leovalais (changes requested):
Cursory review. I'll come back to this when @flomonster's comments are addressed. Meanwhile, a few comments (mostly about excessive cloning 😄):

(The PR name and commit messages aren't really helpful to understand what's being fixed exactly, and it's not `front: ` commits 😅)

editoast/src/views/path/pathfinding.rs
woshilapin: In order to give more insight into why it might be a good idea to use a slice instead, I'd also argue it will help understand locally what the function is doing.

With the current code, as a reviewer, I have a `HashMap` with a key that is not really typed (a `String` can be a consist's name, or a consist's ID, or maybe something else). So I would need to dig the caller to know which is it. Note also that different callers could use different key (as long as it is a `String`) which would be a problem.

If you are able to change the function to take a slice and still need to construct a map locally, then I'll have the code under the eyes to understand it. And you won't have problems on the callers not being consist-ent (pun intended!).

editoast/src/views/path/pathfinding.rs
woshilapin:
```suggestion
        let Some(physics_consist_parameters) = consists.get(rolling_stock_name).cloned() else {
```

editoast/src/views/train_schedule.rs
```diff
@@ -405,26 +403,50 @@ pub async fn train_simulation_batch(
...
+
+    let consists: HashMap<String, PhysicsConsistParameters> = rolling_stocks
+        .iter()
+        .map(|(key, rs)| {
+            (
+                key.clone(),
+                PhysicsConsistParameters::from_traction_engine(rs.clone().into()),
+            )
+        })
+        .collect();
```
flomonster:
```suggestion
    let consists: HashMap<String, _> = rolling_stocks
        .into_iter()
        .map(|(key, rs)| (key, PhysicsConsistParameters::from_traction_engine(rs)))
        .collect();
```

editoast/src/views/timetable/stdcm.rs
```diff
@@ -381,6 +389,7 @@ impl VirtualTrainRun {
         stdcm_request: &Request,
         infra: &Infra,
         rolling_stock: &RollingStockModel,
+        consist_parameters: &PhysicsConsistParameters,
```
flomonster: Since a `PhysicsConsistParameters` contains a `RollingStock` we should only provide the consist.

flomonster (changes requested):
Thanks.

We need to make sure that the function parameters are as logical and relevant as possible (avoid hacks, as this suits our workflow).

editoast/src/models/fixtures.rs
Wadjetz: it for simple units tests, without database

leovalais: Okay, then in this case I suggest:

```rust
pub fn simple_rolling_stock() -> editoast_schemas::RollingStock
```

and to move that out of `models::fixtures` if no database nor model is involved

woshilapin (commented):
Just left some comments, needs a more thorough review still.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -192,12 +205,13 @@ async fn stdcm(
     let latest_simulation_end = stdcm_request.get_latest_simulation_end(simulation_run_time);
 
     // 3. Get scheduled train requirements
-    let simulations: Vec<_> = train_simulation_batch(
+    let simulations: Vec<_> = consist_train_simulation_batch(
```
flomonster: Using `consist_train_simulation_batch` doesn't work. We need to use the rolling stocks of timetable train schedules here, so `train_simulation_batch`

> [!CAUTION]
> Doing this should break our CI since STDCM is not working properly. The code as it is, won't be able to compute pathfinding for all train schedules that don't use the rolling stock provided by the STDCM query.
> So the STDCM computation will ignore almost all train schedules.

flomonster (changes requested):
The code looks good. We just have to fix the critical bug detected and heavily test this PR before merging.

editoast/src/views/train_schedule.rs
```diff
@@ -95,6 +96,9 @@ pub enum TrainScheduleError {
     #[error("{number} train schedule(s) could not be found")]
     #[editoast_error(status = 404)]
     BatchTrainScheduleNotFound { number: usize },
+    #[error("{number} rollings stock(s) could not be found")]
+    #[editoast_error(status = 404)]
+    BatchRollingStockNotFound { number: usize },
```
flomonster: We should not return an error in this case. The previous behavior was the right one.

If a RS is not found in the batch we should still compute all the simulations. Trains that don't have a valid RS are marked as invalid.

editoast/src/views/train_schedule.rs
```diff
@@ -401,29 +385,65 @@ pub async fn train_simulation_batch(
...
+        .iter()
+        .map::<String, _>(|t| t.rolling_stock_name.clone());
+
+    let rolling_stocks: Vec<RollingStockModel> =
+        RollingStockModel::retrieve_batch_or_fail(conn, rolling_stocks_ids, |missing| {
+            TrainScheduleError::BatchRollingStockNotFound {
+                number: missing.len(),
+            }
+        })
+        .await?;
```
flomonster: We should keep the behavior as it was.

editoast/src/views/path/pathfinding.rs
```diff
@@ -336,20 +337,23 @@ pub async fn pathfinding_from_train(
...
-            .collect();
+    let rolling_stock = RollingStockModel::retrieve_or_fail(
+        conn,
+        train_schedule.rolling_stock_name.clone(),
+        || PathfindingError::RollingStockNotFound {
+            rolling_stock_name: train_schedule.rolling_stock_name.clone(),
+        },
+    )
+    .await?
+    .into();
```
flomonster: Here it's working because we handle a single train. To avoid confusion and for consistency, I would keep the previous workflow.

flomonster (changes requested):

editoast/src/views/train_schedule.rs
flomonster: We **MUST NOT** return an error if a rolling stock is not found here. See the new comments, doing so introduced a huge regression.

editoast/src/views/path/pathfinding.rs
```diff
@@ -336,24 +337,19 @@ pub async fn pathfinding_from_train(
     infra: &Infra,
     train_schedule: TrainSchedule,
 ) -> Result<PathfindingResult> {
-    let rolling_stocks =
+    let rolling_stock: Vec<RollingStock> =
         RollingStockModel::retrieve(conn, train_schedule.rolling_stock_name.clone())
             .await?
             .into_iter()
-            .map(|rs| (rs.name.clone(), rs))
+            .map(|rs| rs.into())
```
leovalais:
```suggestion
            .map_into()
```

leovalais (approved):
LGTM, thanks for the changes!

editoast/src/views/train_schedule.rs
leovalais: Indeed. But then we have a documentation problem: we should at least add a comment above the instruction and document the function explicating its behavior. We can't just silently ignore a list of failures (albeit local ones) without "explaining why it's safe to do so".

@Wadjetz sorry about the misdirection, could you add these comments? You can also use `retrieve_batch_unchecked` to avoid computing the list of missing IDs. This will also simplify the `let` clause.

SharglutDev (approved):

flomonster (approved):
LGTM

front/src/assets/rollingStock/freightRollingStocks.ts
```diff
@@ -108,6 +108,7 @@ const FREIGHT_ROLLING_STOCKS = [
   'Y8000AP',
   'Y9000US',
   'electric_rolling_stock_test_e2e',
+  'FAST_ROLLING_STOCK_TEST_E2E',
```
flomonster, line 112: Can't we rename the fast rolling stock (used in the e2e test) to one of the whitelisted ?

Maymanaf (approved):

front/src/assets/rollingStock/freightRollingStocks.ts
flomonster: At least we can add a TODO comment to explain why we have this value and how we should get rid of this.

PR #9658, opened 2024-11-08
ci: add script and dockerfiles checks to github build action

Synar:
Close #9650

PR #9661, opened 2024-11-08
core: update start of free block in signal projection (space-time chart) when stopping on closed signal

### LINKED ISSUE #8814: Core: update free block start in space-time chart when stopping on closed-signal (match conflict detection)
Part of https://github.com/osrd-project/osrd-confidential/issues/568

Update the response for space-time chart when reception is done on closed signal to display a free block starting 20s before train restart (and not on signal sight as currently).

bougue-pe:
Fix #8814

This is done by amending signal-sightings with stop-ending (minus margin) when stop is on closed signal

Also:
* Adds corresponding integration test

A complete comment up-to-date is also worth reading: https://github.com/OpenRailAssociation/osrd/pull/9661#issuecomment-2491886171

🔍 review by commit is probably easier

### DIFF
--- a/editoast/src/core/simulation.rs
+++ b/editoast/src/core/simulation.rs
@@ -26,7 +26,7 @@ use std::hash::Hash;
 editoast_common::schemas! {
     CompleteReportTrain,
     RoutingRequirement,
-    SignalSighting,
+    SignalCriticalPosition,
     SpacingRequirement,
     RoutingZoneRequirement,
     ZoneUpdate,
@@ -306,14 +306,16 @@ pub struct ReportTrain {
 pub struct CompleteReportTrain {
     #[serde(flatten)]
     pub report_train: ReportTrain,
-    pub signal_sightings: Vec<SignalSighting>,
+    pub signal_critical_positions: Vec<SignalCriticalPosition>,
     pub zone_updates: Vec<ZoneUpdate>,
     pub spacing_requirements: Vec<SpacingRequirement>,
     pub routing_requirements: Vec<RoutingRequirement>,
 }
 
 #[derive(Debug, Clone, PartialEq, Hash, Serialize, Deserialize, ToSchema)]
-pub struct SignalSighting {
+/// First position (space and time) along the path where given signal must
+/// be free (sighting time or closed-signal stop ending)
+pub struct SignalCriticalPosition {
     pub signal: String,
     /// Time in ms
     pub time: u64,
--- a/tests/tests/test_timetable.py
+++ b/tests/tests/test_timetable.py
@@ -29,7 +29,7 @@ def test_conflicts(
     expected_conflict_types: set[str],
 ):
     requests.post(f"{EDITOAST_URL}infra/{small_infra.id}/load").raise_for_status()
-    train_schedule_payload = [
+    stopping_train_schedule_payload = [
         {
             "comfort": "STANDARD",
             "constraint_distribution": "STANDARD",
@@ -57,7 +57,11 @@ def test_conflicts(
             "train_name": "with_stop",
         }
     ]
-    requests.post(f"{EDITOAST_URL}/timetable/{timetable_id}/train_schedule", json=train_schedule_payload)
+    stopping_train_schedule_response = requests.post(
+        f"{EDITOAST_URL}/timetable/{timetable_id}/train_schedule", json=stopping_train_schedule_payload
+    )
+    stopping_train_schedule_response.raise_for_status()
+
     train_schedule_payload = [
         {
             "comfort": "STANDARD",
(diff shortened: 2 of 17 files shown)


emersion (approved):
LGTM for the frontend (not tested)

woshilapin (approved):
Nice improvement, and less dead code if I understand correctly. Also, you will need to rebase because [a CI PR](https://github.com/OpenRailAssociation/osrd/pull/9658) happened in between which explains why you see 50 steps (instead of 25) with half of them not running.

Khoyo:
(I don't see the dead code removal :p)

I believe we *need* signal sightings in order to only output green zones for signal that are effectively constraining the projected train. `path_signals` are about the path that's projected onto, and requirements are zone based, so we cannot only rely on those.

Here is an example:

image

The first green block in this image should not be there, as the first train never sees that signal

bougue-pe:
All rewritten, as discussed with @Khoyo:
Space-requirements start are not displayed in space-time chart, what's displayed for free block is the first moment where a signal must be free (usually signal sighting by the train, or the stop ending when on closed signal).
So:
* back to using "signal-sighting" (amended with closed stop positions)
* little refactors on the way (🔍 review by commit)
* ✅ hand-tested that @Khoyo's example is now OK
image
* ✅ hand-tested that it works as expected
image
* previous work is stored in branch https://github.com/OpenRailAssociation/osrd/tree/peb/core/use_spacing_reqs_for_free_block_in_space_time_chart

Khoyo (approved):
Nicely done 👍🏻

woshilapin (approved):
Is `signal_critical_position` better than `signal_sighting`? I'm sure I don't understand all of it, but intuitively, "signal sighting" means something in my mind (when you see the signal), whereas "signal critical position" feels like it's missing some context: **critical** for what (or for when)? Feel free to disregard if this is about a shared jargon term.

editoast/src/core/simulation.rs
```diff
@@ -306,14 +306,15 @@ pub struct ReportTrain {
...
-    pub signal_sightings: Vec<SignalSighting>,
+    pub signal_critical_positions: Vec<SignalCriticalPosition>,
     pub zone_updates: Vec<ZoneUpdate>,
     pub spacing_requirements: Vec<SpacingRequirement>,
     pub routing_requirements: Vec<RoutingRequirement>,
 }
 
 #[derive(Debug, Clone, PartialEq, Hash, Serialize, Deserialize, ToSchema)]
-pub struct SignalSighting {
+/// First position (space and time) along the path where given signal must be free
```
bougue-pe: @woshilapin (reply to https://github.com/OpenRailAssociation/osrd/pull/9661#pullrequestreview-2457946812): signal critical position's definition is exposed in openapi (even though this part is unread by front, only exchanged between core and editoast).
And it's indeed better than signal-sighting IMO because it's not always signal-sighting.

Would it be better with this change?
```suggestion
/// First position (space and time) along the path where given signal must be free (sighting time or closed-signal stop ending)
```

Note that it is already documented a bit [here](https://github.com/OpenRailAssociation/osrd/pull/9661/files#diff-042f91ba14bffb49566c677b16c9092f1a01538d07c608fb82b35897eda9ddb4R249-R252) by this PR

woshilapin: Yes, this improvement will help a little.

Note that I was not clear enough in my comment. I understood that the naming had to evolve and "signal sighting" was not correct enough. I was just trying to highlight that maybe "signal critical position" felt like an unfinished sentence: I would expect something after like "signal critical position before the train has to stop" (which is obviously too long for a name): "critical position" seems to refer to something that is not described in the name.

I don't really have a good solution to propose, so I'd be fine with the small improvement you propose above (feel free to resolve). Yet, I'll propose some terms here (in brainstorming mode), just in case it triggers something in you: "signal distance to stop", "maximum distance before signal", "ultimate position before stop".

bougue-pe: Pushed (directly amended) the suggestion: https://github.com/OpenRailAssociation/osrd/compare/ecce330889fd3885491eeb20b66a2b154a747394..461433f4d9aa532fc76cb4eb04dffa6d1b7d0ba5

Not really triggered by the others terms, we may dig along the fact that non-free signal would be a (broken) constraint from that position for this train-schedule. But I think that the current name is OK as we had an understanding already (between signaling-system devs at least).

So I resolve, but I am OK if anybody unresolves to bring up something they really think is clearer.

Khoyo: > "signal distance to stop", "maximum distance before signal", "ultimate position before stop"

I dislike all of those, sorry :/

It's really "spacetime point where the signal has to be green for no slowdown". It's not (only) about stops nor is it a distance, and the time component is more important. You can recover the spatial component from the time one if you have the enveloppe, you can't do it the other way - since it may be during a stop. (This is what I dislike most about "critical position", which come from a variable name in the routing ressource generation algorithm from before closed-signal stop support, and thus couldn't be a point during a stop)

bougue-pe: You have a point with the time vs. space, so `SignalCriticalTime` is better, then?
I'm OK to change this if you agree :)

woshilapin: Oh, to be honest, I dislike all of them too. I was just throwing more new words to see if it might trigger new ideas from the experts 😅 It didn't, and that's fine, especially since

> But I think that the current name is OK as we had an understanding already (between signaling-system devs at least).

"critical position" seems to be a shared jargon term between expert in signaling-system, let's keep it like this.

bougue-pe: As discussed aside: not worth changing too much the scope of this PR, let's keep it to `critical_position`, even if `critical_time` is a valid idea.

tests/tests/test_timetable.py
```diff
@@ -85,11 +88,45 @@ def test_conflicts(
...
     assert actual_conflicts == expected_conflict_types
 
+    # Check GET reservation block starts at the right time for the signal protecting switch.
+    # Train is received on closed (STOP/SHORT_SLIP_STOP) or OPEN signal.
+    # The free-block requirement must start at the same time as the spacing requirement of the switch's zone
+    # (signal sight for OPEN reception, or 20s before restart for STOP/SHORT_SLIP_STOP reception).
+    train_id = stopping_train_schedule_response.json()[0]["id"]
+    simu_response = requests.get(
+        f"{EDITOAST_URL}/train_schedule/{train_id}/simulation/?infra_id={small_infra.id}"
+    ).json()
```
shenriotpro: nit: you may want to check statuses

bougue-pe: Thanks, amended the 4th commit and pushed-force directly https://github.com/OpenRailAssociation/osrd/compare/461433f4d9aa532fc76cb4eb04dffa6d1b7d0ba5..34ba402dfbc926c3a485ff1b0d2a2431c1271daf

tests/tests/test_timetable.py
```diff
@@ -84,12 +88,51 @@ def test_conflicts(
...
-    requests.post(f"{EDITOAST_URL}/timetable/{timetable_id}/train_schedule", json=train_schedule_payload)
-    response = requests.get(f"{EDITOAST_URL}/timetable/{timetable_id}/conflicts/?infra_id={small_infra.id}")
-    assert response.status_code == 200
-    actual_conflicts = {conflict["conflict_type"] for conflict in response.json()}
+    requests.post(
+        f"{EDITOAST_URL}/timetable/{timetable_id}/train_schedule", json=train_schedule_payload
+    ).raise_for_status()
+
+    conflicts_response = requests.get(f"{EDITOAST_URL}/timetable/{timetable_id}/conflicts/?infra_id={small_infra.id}")
+    assert conflicts_response.status_code == 200
```
Khoyo: Nit, but you should use `conflicts_response.raise_for_status()` (ditto for the other status asserts)

bougue-pe: 🙂 hesitated with this (`== 200` vs. `200 =< ... < 400` + consitency)
One vote, so there you go (amended 4th commit again): https://github.com/OpenRailAssociation/osrd/compare/34ba402dfbc926c3a485ff1b0d2a2431c1271daf..08755cd77b29ec02968c6ad59a7b8678f891abce

shenriotpro (approved):
Thanks

PR #9665, opened 2024-11-08
core, editoast: rename rolling stock to physics consist

woshilapin:
Now that consists are supported instead of a single traction engine (named rolling stock in the entire application), it's reasonable to rename all along up to `core`.

It's a follow up of https://github.com/OpenRailAssociation/osrd/pull/9106.

❓ In `core`, I changed `PhysicsRollingStockModel` into `PhysicsConsistModel`. But there is also `RollingStockParser` which converts a `PhysicsConsistModel` into a `RollingStock`: I'm not sure how far I should push the renaming. Is `RollingStock` only about physics? Because `RollingStock` contains also information about identifier, or `supportedSignalingSystems` which not really be about the physics of the rolling stock?

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/api/api_v2/standalone_sim/RollingStockParser.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/api/api_v2/standalone_sim/RollingStockParser.kt
@@ -10,17 +10,17 @@ import fr.sncf.osrd.train.RollingStock.*
 
 /** Parse the rolling stock model into something the backend can work with */
 fun parseRawRollingStock(
-    rawRollingStock: PhysicsRollingStockModel,
+    rawPhysicsConsist: PhysicsConsistModel,
     loadingGaugeType: RJSLoadingGaugeType = RJSLoadingGaugeType.G1,
     rollingStockSupportedSignalingSystems: List<String> = listOf(),
 ): RollingStock {
     // Parse effort_curves
-    val rawModes = rawRollingStock.effortCurves.modes
+    val rawModes = rawPhysicsConsist.effortCurves.modes
 
-    if (!rawModes.containsKey(rawRollingStock.effortCurves.defaultMode))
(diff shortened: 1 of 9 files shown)


flomonster (approved):
LGTM

eckter (approved):
LGTM

> ❓ In core, I changed PhysicsRollingStockModel into PhysicsConsistModel. But there is also RollingStockParser which converts a PhysicsConsistModel into a RollingStock: I'm not sure how far I should push the renaming. Is RollingStock only about physics? Because RollingStock contains also information about identifier, or supportedSignalingSystems which not really be about the physics of the rolling stock?

I feel like a "rolling stock" is supposed to be "anything that can move on tracks". There's some semantics issues around identifiers and such, but I feel like we'd need to think more carefully about what we precisely intent there, we'd need to take some time to discuss all of this.

We should probably do it eventually, but maybe not quite yet

woshilapin:
@eckter I agree that, literally, `rolling stock` means what you said. But this term comes with the history of OSRD which actually associate `rolling stock` with: "something that has traction capability". Therefore, `rolling stock` today can be misunderstood, and I'd prefer to move to these terms instead:

1. traction engine (some part of `editoast` already adopted that),
2. towed rolling stock (since this feature is recent, I believe it's already relatively homogeneous everywhere),
3. consist

What do you think? And yes, we can discuss it.

Wadjetz (approved):

eckter:
@woshilapin 

Sorry I didn't really see the answer. I didn't mean to block this, it's fine as it is.

In the context of `core`, we don't really consider the distinction between "anything that moves on track" or "anything with traction", or even consists and such. We don't handle the database and we don't assemble anything together. So it doesn't really matter either way in this context, we can go with whatever is consistent with the rest of the codebase. 

The distinction between `RollingStock` and `PhysicsRollingStock` is mostly "legacy", I think it's related to circular dependencies between modules. The string identifier isn't really used as far as I know.

PR #9666, opened 2024-11-08
editoast: fix all units about rolling resistance

### LINKED ISSUE #9657: towed rolling stocks don't forward rolling resistance values in the right unit
### What happened?

When sending an stdcm request for a given rolling stock with no towed rolling stock, the rolling resistance have this value (in the payload editoast -> core):

```json
        "rolling_resistance": {
            "type": "davis",
            "A": 2200.0,
            "B": 64.8,
            "C": 5.832
        },
```

Same rolling stock but with a towed rolling stock:

```json
        "rolling_resistance": {
            "type": "davis",
            "A": 2200000.0,
            "B": 233280.0,
            "C": 75582.72
        },
```

Unsurprisingly, we always fail to find a solution. This happens for any value in the towed rolling stock data. 

Values with the same order of magnitude as before

Run any kind of request with a towed rolling stock (and weight and such otherwise it's ignored). Check the payloads received by core (can be done by setting `LOG_STDCM_REQUESTS=true`

Local, Recette (SNCF), Dev (SNCF)

Firefox

8461f3efed951b9b360c4039b9ae5d4adc60136c

woshilapin:
Fixes #9657 (or we hope it does at least).

The database is populated with SI units, which means Newton, kg, meter and seconds.

Another important piece of information is:
- Units of rolling resistance for traction engines (also named rolling stocks) are in N, N/(m/s) and N/(m/s)²
- Units of rolling resistance for towed rolling stocks are in (N/kg), (N/kg)/(m/s) and (N/kg)/(m/s)².

Therefore, we need 2 different `struct` for storing those, hence the introduction of `RollingResistancePerWeight`.

### DIFF
--- a/editoast/editoast_schemas/src/rolling_stock.rs
+++ b/editoast/editoast_schemas/src/rolling_stock.rs
@@ -10,6 +10,7 @@ pub use effort_curves::ModeEffortCurves;
 
 mod rolling_resistance;
 pub use rolling_resistance::RollingResistance;
+pub use rolling_resistance::RollingResistancePerWeight;
 
 mod energy_source;
 pub use energy_source::EnergySource;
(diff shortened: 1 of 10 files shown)


eckter (approved):
Tests seem to fail and I haven't tested yet, but the units match.

leovalais (approved):
LGTM

PR #9689, opened 2024-11-13
editoast: allow Model fields order to differ from table columns'

### LINKED ISSUE #9468: editoast: cannot declare Model fields in different order from the table's columns
### What happened?

```sql
CREATE TABLE toto(foo, bar, baz);
```

```rust
#[derive(Model)]
#[model(ops = crud)]
struct Toto {
    bar: (),
    baz: (),
    foo: (),
}
```

doesn't compile and the error is really confusing.

It should compile and work, or at the very least the compilation error should be informative.

Swap two fields of an existing model.

Local

Firefox

dev

leovalais:
fixes #9468


woshilapin:
@dependabot rebase

woshilapin (approved):
Thanks very much for that QoL improvement.

PR #9715, opened 2024-11-14
front: hide manchette op list edition button in stdcm debug

### LINKED ISSUE #9457: stdcm: button to edit the op list should not be displayed
### What happened?

This button does nothing in STDCM debug mode:
Image

I should be able to edit the OP just like the operationnal studies mode.

1. Go to STDCM
2. Activate the debugging mode
3. Get a result
4. Try to edit the manchette

Recette (SNCF)

Google Chrome

7451939

Math-R:
close #9457

### DIFF
--- a/front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
+++ b/front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
@@ -163,14 +163,18 @@ const ManchetteWithSpaceTimeChartWrapper = ({
   return (
     <div className="manchette-space-time-chart-wrapper">
       <div className="header">
-        <ManchetteMenuButton setWaypointsPanelIsOpen={setWaypointsPanelIsOpen} />
-        {waypointsPanelIsOpen && waypointsPanelData && (
-          <WaypointsPanel
-            waypointsPanelIsOpen={waypointsPanelIsOpen}
-            setWaypointsPanelIsOpen={setWaypointsPanelIsOpen}
-            waypoints={operationalPoints}
-            waypointsPanelData={waypointsPanelData}
-          />
+        {waypointsPanelData && (
+          <>
+            <ManchetteMenuButton setWaypointsPanelIsOpen={setWaypointsPanelIsOpen} />
+            {waypointsPanelIsOpen && (
+              <WaypointsPanel
+                waypointsPanelIsOpen={waypointsPanelIsOpen}
+                setWaypointsPanelIsOpen={setWaypointsPanelIsOpen}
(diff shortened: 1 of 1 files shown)


front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -163,14 +165,18 @@ const ManchetteWithSpaceTimeChartWrapper = ({
...
       <div className="header">
-        <ManchetteMenuButton setWaypointsPanelIsOpen={setWaypointsPanelIsOpen} />
-        {waypointsPanelIsOpen && waypointsPanelData && (
-          <WaypointsPanel
-            waypointsPanelIsOpen={waypointsPanelIsOpen}
-            setWaypointsPanelIsOpen={setWaypointsPanelIsOpen}
-            waypoints={operationalPoints}
-            waypointsPanelData={waypointsPanelData}
-          />
+        {allowWayPointsModification && (
```
emersion: Could we maybe only display this menu if `waypointsPanelData` is not undefined, instead of introducing a new boolean prop? It seems like it's only set for operational studies and is required to populate the menu.

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -163,14 +163,18 @@ const ManchetteWithSpaceTimeChartWrapper = ({
...
-          <WaypointsPanel
-            waypointsPanelIsOpen={waypointsPanelIsOpen}
-            setWaypointsPanelIsOpen={setWaypointsPanelIsOpen}
-            waypoints={operationalPoints}
-            waypointsPanelData={waypointsPanelData}
-          />
+        {waypointsPanelData && (
+          <>
+            <ManchetteMenuButton setWaypointsPanelIsOpen={setWaypointsPanelIsOpen} />
+            {waypointsPanelIsOpen && waypointsPanelData && (
```
emersion: Nit: `waypointsPanelData` is already checked to be non-undefined above, so we don't need to check it again in the nested condition.

emersion (approved):
Looks good apart from one nit, feel free to resolve it!

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
Math-R: my bad, brain still confused from the past three days ! thanks !

clarani (approved):
LGTM :white_check_mark:

PR #9725, opened 2024-11-15
front: handle stops in viriato imports

### LINKED ISSUE #9641: Viriato import doesn't handle stops
### What happened?

When importing a Viriato file, the stops seem to disappear:

file:
Image

generated train schedule:
Image

The stops at Embrun and Chorges for instance are missing

Bug with file Etoile de Veynes, ask file in mp

handle stops

1. load Etoile de Veynes
2. open schedule table
3. notice

Recette (SNCF)

Firefox

staging cd4cdfb

RomainValls:
closes #9641 

You can ask me to send you the test files by message

Now, when a step has an arrival and departure time, the calculated difference is used for the stop duration.

### DIFF
--- a/front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
+++ b/front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
@@ -170,28 +170,21 @@ const ImportTrainScheduleConfig = ({
     startDate: string
   ): Step[] =>
     ocpTTs
-      .map((ocpTT, index): Step | null => {
+      .map((ocpTT): Step | null => {
         const ocpRef = ocpTT.getAttribute('ocpRef');
         const times = ocpTT.getElementsByTagName('times')[0];
+        const isLastOcp = ocpTT === ocpTTs.at(-1);
+        const ocpType = ocpTT.getAttribute('ocpType');
         let departureTime = times?.getAttribute('departure') || '';
-        let arrivalTime = times?.getAttribute('arrival') || '';
-
-        const isLastOcpTT = index === ocpTTs.length - 1;
-
-        if (isLastOcpTT) {
-          arrivalTime = cleanTimeFormat(departureTime) || cleanTimeFormat(arrivalTime); // For the last sequence, arrival equals departure
-          departureTime = cleanTimeFormat(arrivalTime) || cleanTimeFormat(departureTime);
-        } else if (index !== 0) {
-          arrivalTime = times?.getAttribute('arrival') || times?.getAttribute('departure') || '';
-          arrivalTime = cleanTimeFormat(arrivalTime);
-          departureTime = times?.getAttribute('departure') || times?.getAttribute('arrival') || '';
-          departureTime = cleanTimeFormat(departureTime);
-        }
+        let arrivalTime = ocpType === 'pass' ? departureTime : times?.getAttribute('arrival') || '';
+        arrivalTime = cleanTimeFormat(arrivalTime);
+        departureTime = cleanTimeFormat(departureTime);
 
         if (!ocpRef) {
           console.error('ocpRef is null or undefined');
           return null;
         }
+
         const operationalPoint = cichDict[ocpRef];
 
         if (!operationalPoint) {
@@ -203,13 +196,30 @@ const ImportTrainScheduleConfig = ({
         const formattedArrivalTime = `${startDate} ${arrivalTime}`;
         const formattedDepartureTime = `${startDate} ${departureTime}`;
 
+        let stopFor: number | undefined;
+
+        const arrivalDate = new Date(`${startDate}T${arrivalTime}`);
+        const departureDate = new Date(`${startDate}T${departureTime}`);
+        if (ocpType === 'stop') {
+          if (arrivalTime && departureTime) {
+            stopFor = Math.round((departureDate.getTime() - arrivalDate.getTime()) / 1000);
+          } else {
+            stopFor = 0;
+          }
+        } else if (ocpType === 'pass') {
+          if (isLastOcp) {
+            stopFor = 0;
+          }
+        }
+
         return {
           id: nextId(),
           uic,
           chCode,
           name: ocpRef,
-          arrivalTime: cleanTimeFormat(formattedArrivalTime),
-          departureTime: cleanTimeFormat(formattedDepartureTime),
+          arrivalTime: formattedArrivalTime,
+          departureTime: formattedDepartureTime,
+          duration: stopFor,
         } as Step;
       })
       .filter((step): step is Step => step !== null);


RomainValls:
You might notice that, for some steps (in this test file, it happens a lot with the step "Veynes"), we have a step duplicate in the TimeStopsInput and the TimeStopsOutput. This causes the second duplicated value to be valued at day+1. 
I don't think it's related to the viriato import as it also happens when creating the train manually. 
I'll open a new [issue](https://github.com/OpenRailAssociation/osrd/issues/9727) about it.

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
```diff
@@ -170,28 +170,22 @@ const ImportTrainScheduleConfig = ({
...
-        if (isLastOcpTT) {
-          arrivalTime = cleanTimeFormat(departureTime) || cleanTimeFormat(arrivalTime); // For the last sequence, arrival equals departure
-          departureTime = cleanTimeFormat(arrivalTime) || cleanTimeFormat(departureTime);
-        } else if (index !== 0) {
-          arrivalTime = times?.getAttribute('arrival') || times?.getAttribute('departure') || '';
-          arrivalTime = cleanTimeFormat(arrivalTime);
-          departureTime = times?.getAttribute('departure') || times?.getAttribute('arrival') || '';
-          departureTime = cleanTimeFormat(departureTime);
-        }
+        arrivalTime = times?.getAttribute('arrival') || times?.getAttribute('departure') || '';
```
emersion: Seems like these variables are set, and then always overwritten with something else? Probably we can drop the first value they're set to?

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
```diff
@@ -203,13 +197,21 @@ const ImportTrainScheduleConfig = ({
...
         return {
           id: nextId(),
           uic,
           chCode,
           name: ocpRef,
-          arrivalTime: cleanTimeFormat(formattedArrivalTime),
-          departureTime: cleanTimeFormat(formattedDepartureTime),
+          arrivalTime: formattedArrivalTime,
+          departureTime: formattedDepartureTime,
+          duration: stopFor > 0 ? stopFor : undefined,
```
emersion: If the user has set the departure time equal to the arrival time, they've requested a zero stop duration. This is semantically different from an undefined stop duration: undefined stop duration means that the train goes through a waypoint without stopping, zero stop duration means that the train stops at a waypoint and immediately restarts.

To handle this, instead of checking whether `stopFor > 0`, we could probably define `stopFor` as a `number | undefined`?

```ts
let stopFor: number | undefined;
```

front/src/applications/operationalStudies/types.ts
```diff
@@ -18,6 +18,7 @@ import type { SuggestedOP } from 'modules/trainschedule/components/ManageTrainSc
 import type { ArrayElement } from 'utils/types';
 
 export interface Step {
+  id: string;
```
emersion: Is this used anywhere? If it's not used, we can probably just drop it?

Uriel-Sautron (approved):
Tested good work

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
```diff
@@ -170,28 +170,24 @@ const ImportTrainScheduleConfig = ({
     startDate: string
   ): Step[] =>
     ocpTTs
-      .map((ocpTT, index): Step | null => {
+      .map((ocpTT): Step | null => {
         const ocpRef = ocpTT.getAttribute('ocpRef');
         const times = ocpTT.getElementsByTagName('times')[0];
+        const isLastOcp = ocpTT === ocpTTs[ocpTTs.length - 1];
```
Uriel-Sautron:
```suggestion
        const isLastOcp = ocpTT === ocpTTs.at(- 1);
```

Uriel-Sautron: .at(-1) retrieves the last element of the array

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
```diff
@@ -203,13 +199,40 @@ const ImportTrainScheduleConfig = ({
...
+
+        const arrivalDate = new Date(`${startDate}T${arrivalTime}`);
+        const departureDate = new Date(`${startDate}T${departureTime}`);
+        if (ocpType === 'stop') {
+          if (arrivalTime && departureTime) {
+            if (arrivalTime === departureTime) {
+              stopFor = 0;
+            } else {
+              stopFor = Math.round((departureDate.getTime() - arrivalDate.getTime()) / 1000);
+            }
```
clarani: If arrivalTime and departureTime are the same, then stopFor will be 0 ;) You can remove the first test
```suggestion
            stopFor = Math.round((departureDate.getTime() - arrivalDate.getTime()) / 1000);
```

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
```diff
@@ -203,13 +199,40 @@ const ImportTrainScheduleConfig = ({
...
+          } else if (departureTime) {
+            stopFor = 0;
+          }
```
clarani: Do you think you can handle the case this way:
- if the stop has a departure and a arrival times, then compute the stop duration
- else set the stop duration to 0
```suggestion
          } else {
            stopFor = 0;
          }
```

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
```diff
@@ -203,13 +199,40 @@ const ImportTrainScheduleConfig = ({
...
+        } else if (ocpType === 'pass') {
+          if (arrivalTime && departureTime && arrivalTime !== departureTime) {
+            // Incorrect data, it shouldn't happen on a ocpType pass, skip step
+            return null;
+          }
+          if (isLastOcp && departureTime) {
```
clarani: Why do you need to check departureTime here ? :)

RomainValls: it is indeed useless :D thanks

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
```diff
@@ -203,13 +199,36 @@ const ImportTrainScheduleConfig = ({
...
         return {
           id: nextId(),
           uic,
           chCode,
           name: ocpRef,
-          arrivalTime: cleanTimeFormat(formattedArrivalTime),
-          departureTime: cleanTimeFormat(formattedDepartureTime),
+          arrivalTime: formattedArrivalTime,
+          departureTime: formattedDepartureTime,
+          duration: stopFor || undefined,
```
emersion: `|| undefined` breaks zero duration stops: `0` is transformed to `undefined`, which has a different meaning.

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
```diff
@@ -203,13 +199,36 @@ const ImportTrainScheduleConfig = ({
...
+          if (arrivalTime && departureTime) {
+            stopFor = Math.round((departureDate.getTime() - arrivalDate.getTime()) / 1000);
+          } else {
+            stopFor = 0;
+          }
+        } else if (ocpType === 'pass') {
+          if (arrivalTime && departureTime && arrivalTime !== departureTime) {
+            // Incorrect data, it shouldn't happen on a ocpType pass, skip step
+            return null;
+          }
```
emersion: Given the way `arrivalTime` is defined above, I don't think this can ever happen? That is, if `ocpType === 'pass'`, then `arrivalTime` is always equal to `departureTime`?

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
```diff
@@ -170,28 +170,24 @@ const ImportTrainScheduleConfig = ({
...
-          departureTime = cleanTimeFormat(arrivalTime) || cleanTimeFormat(departureTime);
-        } else if (index !== 0) {
-          arrivalTime = times?.getAttribute('arrival') || times?.getAttribute('departure') || '';
-          arrivalTime = cleanTimeFormat(arrivalTime);
-          departureTime = times?.getAttribute('departure') || times?.getAttribute('arrival') || '';
-          departureTime = cleanTimeFormat(departureTime);
-        }
+        let arrivalTime =
+          ocpType === 'pass'
+            ? times?.getAttribute('departure') || ''
```
emersion: Nit: this can be simplified to `departureTime`.

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
```diff
@@ -203,13 +199,36 @@ const ImportTrainScheduleConfig = ({
...
+          }
+        } else if (ocpType === 'pass') {
+          if (arrivalTime && departureTime && arrivalTime !== departureTime) {
+            // Incorrect data, it shouldn't happen on a ocpType pass, skip step
+            return null;
+          }
+          if (isLastOcp) {
+            stopFor = 0;
+          } else {
+            stopFor = undefined;
```
emersion: Optional nit: `undefined` is the initial value of `stopFor` so this isn't strictly needed (but I don't mind).

emersion (commented):
Apart from these LGTM!

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
clarani: I think `stopFor ?? undefined` should work :) (`??` checks only nullish value, so `null` and `undefined` whereas || checks falsy values (`null`, `undefined`, `0` and `''`)

emersion: But there's no need to convert anything here, the original value is already a `number | undefined`.

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
RomainValls: Indeed, we won't enter this condition, removing it

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
```diff
@@ -170,28 +170,22 @@ const ImportTrainScheduleConfig = ({
...
-          arrivalTime = cleanTimeFormat(departureTime) || cleanTimeFormat(arrivalTime); // For the last sequence, arrival equals departure
-          departureTime = cleanTimeFormat(arrivalTime) || cleanTimeFormat(departureTime);
-        } else if (index !== 0) {
-          arrivalTime = times?.getAttribute('arrival') || times?.getAttribute('departure') || '';
-          arrivalTime = cleanTimeFormat(arrivalTime);
-          departureTime = times?.getAttribute('departure') || times?.getAttribute('arrival') || '';
-          departureTime = cleanTimeFormat(departureTime);
-        }
+        let arrivalTime =
+          ocpType === 'pass' ? departureTime || '' : times?.getAttribute('arrival') || '';
```
emersion: Nit: `departureTime || ''` can be simplified to `departureTime`, because the line right above already has `|| ''`.

emersion (approved):
LGTM regardless (feel free to resolve comment)!

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
RomainValls: thanks 🙏

PR #9726, opened 2024-11-15
front, gateway: supporting launching front from outside the gateway reverse proxy

ghost:
For the developers working on macOS it can be a real pain to launch the front in a local fashion. Since we require the front to be served through the gateway, we must do gymnastics to serve the dev front.

This set of changes aims to solve this by providing a way to use the front from outside the stack, making it serve locally while consuming the services on the remote service (inside the Docker VM on macOS or windows, for instance).

In order to test this, start the stack in non-host mode, shutdown the front container and on your `.env.local` add:

```
VITE_PROXY_EDITOAST="http://localhost:4000/api"
VITE_PROXY_GATEWAY="http://localhost:4000"
```

Start the front locally, and it should work.

### DIFF
--- a/docker/gateway.dev.simple.toml
+++ b/docker/gateway.dev.simple.toml
@@ -5,6 +5,7 @@
 secret_key = "NOT+A+SECRET++NOT+A+SECRET++NOT+A+SECRET++NOT+A+SECRET++NOT+A+SECRET++NOT+A+SECRET++NOT+A+SECRET"
 
 listen_addr = "0.0.0.0"
+allowed_origins = ["http://localhost:3000"]
 
 [telemetry.tracing]
 enable = true
--- a/docker/gateway.dev.simple.toml
+++ b/docker/gateway.dev.simple.toml
@@ -5,7 +5,7 @@
 secret_key = "NOT+A+SECRET++NOT+A+SECRET++NOT+A+SECRET++NOT+A+SECRET++NOT+A+SECRET++NOT+A+SECRET++NOT+A+SECRET"
 
 listen_addr = "0.0.0.0"
-allowed_origins = ["http://localhost:3000"]
+allowed_origins = ["http://127.0.0.1:3000", "http://localhost:3000"]
 
 [telemetry.tracing]
 enable = true
--- a/gateway/src/main.rs
+++ b/gateway/src/main.rs
@@ -61,6 +61,9 @@ async fn main() -> std::io::Result<()> {
     // Enable telemetry
     config.telemetry.enable();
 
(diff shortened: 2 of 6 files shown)


docker/gateway.dev.simple.toml
```diff
@@ -5,6 +5,7 @@
 secret_key = "NOT+A+SECRET++NOT+A+SECRET++NOT+A+SECRET++NOT+A+SECRET++NOT+A+SECRET++NOT+A+SECRET++NOT+A+SECRET"
 
 listen_addr = "0.0.0.0"
+allowed_origins = ["http://localhost:3000"]
```
Khoyo: We should probably allow 127.0.0.1 too

gateway/src/main.rs
```diff
@@ -83,7 +97,7 @@ async fn main() -> std::io::Result<()> {
                     .configure(actix_auth::config)
                     .default_service(web::to(HttpResponse::NotFound)),
             );
-
+
```
Khoyo: 👀

Math-R (approved):
lgtm thanks for this one !

Khoyo (approved):

.github/workflows/build.yml
```diff
@@ -872,6 +872,7 @@ jobs:
       - name: Execute tests within container
         run: |
           docker run --name=front-test --net=host -v $PWD/output:/app/tests/unit \
+            -e PROXY_EDITOAST='/api' -e PROXY_GATEWAY='' \
```
emersion: Since these are the defaults, maybe we can just leave them unspecified?

front/src/common/api/osrdGatewayApi.ts
```diff
@@ -48,11 +48,11 @@ export type LoginRedirect = {
...
   | ({
-      type: 'success';
-    } & LoginSuccess)
+    type: 'success';
+  } & LoginSuccess)
   | ({
-      type: 'redirect';
-    } & LoginRedirect);
+    type: 'redirect';
+  } & LoginRedirect);
```
emersion: Seems like these changes are making CI unhappy.

emersion (approved):
LGTM for the front (not tested)

.github/workflows/build.yml
ghost: import-meta-env expects them to be defined also in the container here, test fails otherwise.

front/src/common/api/osrdGatewayApi.ts
ghost: Yes, reverting those changes.

PR #9729, opened 2024-11-15
front: fix translations for pathfinding error in itinerary

### LINKED ISSUE #9728: Wrong translations for pathfinding errors in itinerary
### What happened?

Image

_No response_

.

Recette (SNCF)

Firefox

.

RomainValls:
closes #9728

### DIFF
--- a/front/public/locales/fr/operationalStudies/manageTrainSchedule.json
+++ b/front/public/locales/fr/operationalStudies/manageTrainSchedule.json
@@ -9,6 +9,9 @@
   "cancelPathfinding": "Annuler",
   "chooseOnMap": "Choisir sur la carte",
   "composition": "Composition",
+  "coreErrors": {
+    "path_with_repeated_tracks": "Le chemin emprunte plusieurs fois la même voie"
+  },
   "deleteRoute": "Supprimer l'itinéraire",
   "deleteVias": "Supprimer les étapes",
   "destination": "Destination",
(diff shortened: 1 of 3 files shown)


emersion (commented):
The error message still isn't human-readable for me:

out

I think we need to set `message` to ``t(`pathfindingErrors.${…}`)`` instead.

front/public/locales/fr/operationalStudies/manageTrainSchedule.json
```diff
@@ -104,7 +104,8 @@
     "not_found_in_tracks": "Voie manquante",
     "not_enough_path_items": "Éléments manquants pour la recherche : Origine, Destination, Matériel roulant",
     "pathfinding_failure": "Aucun chemin trouvé",
-    "rolling_stock_not_found": "Matériel roulant non trouvé"
+    "rolling_stock_not_found": "Matériel roulant non trouvé",
+    "the path goes over the same track multiple times": "Le chemin emprunte plusieurs fois la même voie"
```
emersion: Ah. This is not an error code, it's a human-readable error message. So I don't think this is a stable string we can rely on.

For `pathfindingResult.error_type` we have an error code and we can find a suitable translation. For `pathfindingResult.core_error.message` specifically, we should probably just pass the message as-is without going through translations.

RomainValls: That's what I was wondering. But if the app is set in french, the user will have a mixed french-english error message ?

emersion: If we do want to provide an error message in French, we could use `pathfindingResult.core_error.type` maybe? But there are many such error types (see `core/osrd-reporting/src/main/java/fr/sncf/osrd/reporting/exceptions/ErrorType.java`) so not sure it's a rabbit hole we want to jump into.

emersion (approved):
LGTM!

PR #9733, opened 2024-11-15
front: call pathfinding only once

### LINKED ISSUE #6719: Pathfinding launched two times when adding a via
### What happened?

When adding a via via the modal or by clicking on map, the pathfinding is fired 2 times.

It should be fired only once.

1. Go to a scenario
2. Add a train and launch a pathfinding
3. Open the network tab and clean the history
4. Add a via 
5. You should see two pathfinding call in the network tab

- **Browser**: Firefox v123
- **OS**: macOS Sonoma
- **Env**: Dev

dev 1482e8c

anisometropie:
closes #6719

Ignore the vias changes made automatically from the pathfinding result which resulted in a new pathfinding request.

### DIFF
--- a/front/src/modules/pathfinding/hooks/usePathfinding.ts
+++ b/front/src/modules/pathfinding/hooks/usePathfinding.ts
@@ -28,6 +28,7 @@ import type { PathStep } from 'reducers/osrdconf/types';
 import { useAppDispatch } from 'store';
 import { isEmptyArray } from 'utils/array';
 import { castErrorToFailure } from 'utils/error';
+import { usePrevious } from 'utils/hooks/state';
 
 import useInfraStatus from './useInfraStatus';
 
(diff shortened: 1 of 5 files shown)


emersion:
Hm, won't this result in a pathfinding being started on stale data, and the latest via change will get ignored?

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -196,6 +196,9 @@ export const usePathfinding = (
   };
 
   useEffect(() => {
+    if (pathfindingState.running) {
+      return;
+    }
```
flomonster: It seems to fix the symptom but not the underlying issue.

anisometropie:
> Hm, won't this result in a pathfinding being started on stale data, and the latest via change will get ignored?

from the test I have run, haven’t seen a case where it doesn’t use the latest data

emersion:
What about this case:

- User sets a via
- Pathfinding starts, takes some time
- User changes their mind, sets another via
- Reply to the first pathfinding comes back

Latest user change is ignored here.

I'd argue correctness (user input not ignored) is more important than performance (1 pathfinding request instead of 2).

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -209,6 +226,12 @@ export const usePathfinding = (
   }, [vias]);
 
   useEffect(() => {
+    if (
+      (previousOrigin !== origin || previousDestination !== destination) &&
```
emersion, line 230: I don't think this is enough: an intermediary waypoint might've changed too, or rolling stock, or something else…

But, what is the root cause? Why does pathfinding run twice, which state change causes this?

anisometropie: intermediary waypoint don’t do aything in this useEffect, and rolling don’t create the bug.

Root cause is pathfinding updating vias when it succeeds. So we need to ignore exactly one call made and resume listening after, which is what `pathStepsCompletedWithPF` does

anisometropie: probably gonna change after the refacto https://github.com/OpenRailAssociation/osrd/pull/9924

PR #8861, opened 2024-09-13
front: redesigned top of scenario page

theocrsb:
this PR cannot be merged before this [PR](https://github.com/OpenRailAssociation/osrd/pull/8765)

closes https://github.com/osrd-project/osrd-confidential/issues/667

PR #9735, opened 2024-11-16
front: fix font weight in input timestop table

### LINKED ISSUE #9732: All entries of input table are bold
### What happened?

Everything is in bold in the input table

Image

_No response_

1. create any train schedule
2. open input table

Recette (SNCF)

Firefox

rec

Synar:
Close #9732

Bug introduced in https://github.com/OpenRailAssociation/osrd/pull/8861, looking at the pr it stands to reason that .active is supposed to be specialized to micro macro buttons. Inspecting the page, it does not seem that any other element affected by the pr was using the .active class.

### DIFF
--- a/front/src/styles/scss/applications/operationalStudies/_scenario.scss
+++ b/front/src/styles/scss/applications/operationalStudies/_scenario.scss
@@ -49,15 +49,15 @@
       margin-inline: 10px;
       background-color: var(--grey20);
     }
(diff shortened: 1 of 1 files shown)


SharglutDev (approved):
Lgtm and tested

theocrsb (approved):
LGTM and tested

PR #9739, opened 2024-11-16
front: fix row deletion in timestop input table

Synar:
Partial fix of #8959
-> IV. Multi selected fields modification is bugged
-> V. Delete the value of a stopping time field behaves weirdly

Prevents deletion of empty fields from creating a via

### DIFF
--- a/front/src/modules/timesStops/helpers/utils.ts
+++ b/front/src/modules/timesStops/helpers/utils.ts
@@ -196,7 +196,11 @@ export function updateRowTimesAndMargin(
       newRowData.stopFor = undefined;
     }
   }
-  if (!newRowData.stopFor && op.fromRowIndex !== allWaypointsLength - 1) {
+  if (
+    !newRowData.stopFor &&
+    newRowData.onStopSignal &&
+    op.fromRowIndex !== allWaypointsLength - 1
+  ) {
     newRowData.onStopSignal = false;
   }
   newRowData.isMarginValid = !(
@@ -211,6 +215,21 @@ export function updateRowTimesAndMargin(
     }
   }
   return newRowData;
+}
+
+/**
+ * This function is called before comparing rows to prevent a change from undefined to null (or the reverse)
+ * from being treated as an actual update of a row (otherwise changes would occur on deletion of an undefined field)
+ */
+export function normalizeNullablesInRow(row: TimesStopsInputRow): TimesStopsInputRow {
+  const normalizedRow = { ...row };
+  if (normalizedRow.stopFor === null) {
+    normalizedRow.stopFor = undefined;
+  }
+  if (normalizedRow.theoreticalMargin === null) {
+    normalizedRow.theoreticalMargin = undefined;
+  }
+  return normalizedRow;
 }
 
 /**
--- a/front/src/modules/timesStops/TimesStopsInput.tsx
+++ b/front/src/modules/timesStops/TimesStopsInput.tsx
@@ -15,6 +15,7 @@ import {
   durationSinceStartTime,
   formatSuggestedViasToRowVias,
   onStopSignalToReceptionSignal,
+  normalizeNullablesInRow,
   updateDaySinceDeparture,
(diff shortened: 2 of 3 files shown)


louisgreiner (approved):
Working perfectly, thanks !

emersion:
Hm, I must admit I'm not a huge fan of this workaround. Could we instead fix the function which checks whether two rows are equal, to consider that `null` is equivalent to `undefined`?

Synar:
> Hm, I must admit I'm not a huge fan of this workaround. Could we instead fix the function which checks whether two rows are equal, to consider that `null` is equivalent to `undefined`?

Right now we're just using lodash isEqual. I suppose we could call a normalizing function setting null to undefined (or both to false for stopSignal) on both rows before comparing. Would you prefer that?

emersion:
We could do something like:

```ts
const { stopFor, theoreticalMargin, ...rest } = row;
return isEqual(rest, …) && stopFor == … && theoreticalMargin == …;
```

Normalizing sounds fine to me as well.

emersion:
Another upside of normalizing is that we can stick to `===` so we don't need to think about the weird rules of `==`.

kmer2016 (approved):
LGTM !

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -213,6 +217,27 @@ export function updateRowTimesAndMargin(
   return newRowData;
 }
 
+/**
+ * This function prevent a change from undefined to null (or the reverse)
+ * from being treated as an actual update of a row (otherwise would occur on deletion of an undefined field)
+ */
+export function preventUpdateOnNullableToNullable(
```
emersion: Starting a stub discussion just so that we don't forget to bring this one to a conclusion before merging: https://github.com/OpenRailAssociation/osrd/pull/9739#issuecomment-2483556154

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -213,6 +217,21 @@ export function updateRowTimesAndMargin(
   return newRowData;
 }
 
+/**
+ * This function is called before comparing rows to prevent a change from undefined to null (or the reverse)
+ * from being treated as an actual update of a row (otherwise changes would occur on deletion of an undefined field)
+ */
+export function normalizeNullablesInRow(row: TimesStopsInputRow): TimesStopsInputRow {
+  const normalizedRow = _.clone(row);
```
emersion: Nit: I'd slightly prefer `= { ...row }` because it's standard JavaScript instead of relying on an external library.

front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -196,7 +196,11 @@ export function updateRowTimesAndMargin(
       newRowData.stopFor = undefined;
     }
   }
-  if (!newRowData.stopFor && op.fromRowIndex !== allWaypointsLength - 1) {
+  if (
+    !newRowData.stopFor &&
+    newRowData.onStopSignal &&
```
emersion, line 201: Nit: does this change really matter? If we reset the field to false when it's already false, it's a no-op so should be fine as well?

(In general, I prefer keeping the list of conditions in a `if` as small as possible.)

Synar: It does because we also set null/undefined to false, which means we need to add the row to the path to remember it. 

We could also put it in the format function, but here I think it makes more sense to directly avoid doing a modification rather than do it then drop it after the comparison

front/src/modules/timesStops/helpers/utils.ts
Synar: Sure, I agree!

front/src/modules/timesStops/helpers/utils.ts
emersion: Makes sense!

emersion (approved):
Thanks for the fix!

PR #9740, opened 2024-11-16
misc: add jetbrain folder .idea to gitignore

Synar:
Prevent files from jetbrain IDEs (datagrip, intellij, pycharm, webstorm, ...) from being tracked.

Jetbrain actually recommends part of the .idea folder to be shared via git and provides its own .gitignore for this purpose, but I'm not sure enough people use jetbrain in our project to warrant this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.backup
 __pycache__
 .vscode
+.idea
 
 # Direnv
 .envrc
--- a/core/.gitignore
+++ b/core/.gitignore
@@ -1,5 +1,4 @@
 .gradle
-.idea
 .kotlin
 build
 *.iml


ghost (approved):

ghost:
> Jetbrain actually recommends part of the .idea folder to be shared via git and provides its own .gitignore for this purpose, but I'm not sure enough people use jetbrain in our project to warrant this.

Let's not bother too much with this imho

Khoyo:
You should probable add it to the `.dockerignore` and remove it from `core/{.gitignore,.dockerignore}`

Synar:
> You should probable add it to the `.dockerignore` and remove it from `core/{.gitignore,.dockerignore}`

Removed it from core/.gitignore, but I don't believe there is such a thing as a general .dockerignore

Khoyo (approved):
That's actually right, oh well

PR #9741, opened 2024-11-16
ci: use microsoft playwright image in e2e tests

### LINKED ISSUE #9738: Accelerate build time by using prebuilt playwright image to run end-to-end tests in the CI
### Description and goal

When building a playwright docker image to run end-to-end tests in the CI, we currently use a bare linux image in which we install playwright and its dependencies. We could try to accelerate this build by directly using the microsoft/playwright docker image, which comes with its browser dependencies pre-installed.

### Acceptance criteria

- [x] The CI end to end tests use the microsoft/playwright docker image and no longer need to install playwright dependencies manually
- [x] The CI end to end tests build time is shorter than before

Synar:
Close #9738

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -188,8 +188,9 @@ jobs:
       - name: Build dummy test_data and static_assets images
         run: |
           echo -e "FROM scratch" > Dockerfile.empty
-          docker build -t test_data -f Dockerfile.empty .
-          docker build -t static_assets -f Dockerfile.empty .
+          docker build -t test_data - <Dockerfile.empty
+          docker build -t static_assets - <Dockerfile.empty
+          docker build -t front_tests:latest - <Dockerfile.empty
 
       - name: Find and check all Dockerfiles using docker build --check
         run: |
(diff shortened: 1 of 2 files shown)


Synar:
Before the changes, the playwright installation took 120s to 140s on 3 attempts.
After the changes, out of 3 attempts, all took 80s to 82s.
We will attempt a second speed up, by reusing the test-front image to multistage build the playwright image.

Synar:
Sadly the multistage approach does not appear to be any faster

The second "fixup" is thus likely to be dropped

.github/workflows/build.yml
```diff
@@ -1054,6 +1055,24 @@ jobs:
...
+          fi
+          echo "Detected Playwright version: $PLAYWRIGHT_VERSION"
+          echo "PLAYWRIGHT_VERSION=$PLAYWRIGHT_VERSION" >> $GITHUB_ENV
+
+      - name: Build Playwright container
+        run: |
+          docker build --build-arg PLAYWRIGHT_VERSION=v$PLAYWRIGHT_VERSION \
+            --build-arg FRONT_TESTS_IMAGE=${{ fromJSON(needs.build.outputs.stable_tags).front-tests }} \
+            -t osrd-playwright:latest \
+            -f front/docker/Dockerfile.playwright-ci .
```
Khoyo: If if the goal is to speed up the ci, we probably shouldn't transfer the whole repo into the context (since the dockerfile doesn't use it anyways) Something like `docker build -t osrd-playwright [etc] - <front/docker/Dockerfile.playwright-ci

front/docker/Dockerfile.playwright-ci
```diff
@@ -0,0 +1,9 @@
+ARG PLAYWRIGHT_VERSION=latest
```
Khoyo: Since this is a ci only file, I'd rather it lives in the /docker/ folder (to avoid pestering the front maintainers)

.github/workflows/build.yml
Synar: Done thanks!

front/docker/Dockerfile.playwright-ci
Synar: Sure

Khoyo (approved):

PR #9749, opened 2024-11-18
front: remove trigram from stdcmPathstep

### LINKED ISSUE #9745: STDCM - Input CI not coherent with the input
### What happened?

I open LMR and try to start a simulation, but the payload is not correct compared to my input. Here : 

Image

"DN" does not correspond to Perrigny but to another OR. 

Note : Does not seem to happen all the time, maybe a cache problem ?

I expected to have the correct payload

Not sure... Maybe do a first simulation, then another with another departure point, and see if the bug can be reproduced

Dev (SNCF)

Google Chrome

1bd519d

SarahBellaha:
Closes #9745

### DIFF
--- a/front/src/reducers/osrdconf/stdcmConf/index.ts
+++ b/front/src/reducers/osrdconf/stdcmConf/index.ts
@@ -163,7 +163,7 @@ export const stdcmConfSlice = createSlice({
       }>
     ) {
       const { linkedPathStep, trainName, pathStep, pathStepId } = action.payload;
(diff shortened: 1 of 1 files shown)


emersion (commented):
Should we remove `trigram` from `StdcmLinkedPathStep` as well?

emersion:
Ah, seems like we shouldn't, because it's used in the UI in `StdcmLinkedPathResults`.

emersion (approved):

clarani (approved):
LGTM, tested

PR #9756, opened 2024-11-18
front: use matchPathStepAndOp() in TimesStopsInput's clearPathStep()

emersion:
We already have a function which matches an OP and a path step.
Re-use it instead of hand-rolling our own logic.
    
In the UIC case, match by positionOnPath instead of name: this
should be more robust. In the track offset case, drop the opId
check because track offsets never have an OP ID.

### DIFF
--- a/front/src/modules/timesStops/TimesStopsInput.tsx
+++ b/front/src/modules/timesStops/TimesStopsInput.tsx
@@ -6,7 +6,7 @@ import type { Operation } from 'react-datasheet-grid/dist/types';
 import { useTranslation } from 'react-i18next';
 
 import { useOsrdConfActions } from 'common/osrdContext';
-import { isVia } from 'modules/pathfinding/utils';
+import { isVia, matchPathStepAndOp } from 'modules/pathfinding/utils';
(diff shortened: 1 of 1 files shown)


Synar:
Very nice! I haven't had the chance to test it yet, but the old way was indeed quite complex and flaky.

front/src/modules/timesStops/TimesStopsInput.tsx
SharglutDev, line 1: Could have been a nice clean up but not sure how this could work. We are comparing the row indexes (containing all ops from the path) and the pathstep indexes (only the waypoints or constraints ops) so their indexes won't match.

SharglutDev (commented):
Tested and it seems we can't remove any step now, see comment

front/src/modules/timesStops/TimesStopsInput.tsx
emersion: Ah, I suppose I just got lucky with the indexes somehow when testing. Will rework the patch.

emersion: Done, let's leave a larger refactoring for later.

SharglutDev (approved):
Lgtm and tested !

PR #9761, opened 2024-11-18
front: simplify OP matching in updatePathStepsFromOperationalPoints()

emersion:
We already have a `matchPathStepAndOp()` function to compare a `PathStep` and a `SuggestedOP`. There are two differences:

- The old code checks whether `PathStep.ch` is set before comparing it with `SuggestedOP.ch`, but that's incorrect. On the editoast side the `secondary_code` field is defined as `Option<String>`: `None` accepts OPs with any secondary code, `""` accepts only OPs without any secondary code. On the TypeScript side, `undefined` accepts OPs with any secondary code, and `''` accepts only OPs without any secondary code. However `'ch' in step` is true for `{ ch: undefined }`, which compares SuggestedOP.ch with undefined even if an undefined `PathStep.ch` is supposed to match any OP secondary code.
- The old code checks `PathStep.secondary_code`. However the `PathStep`s given to `updatePathStepsFromOperationalPoints()` come straight up from `computeBasePathSteps()`, which always leaves `secondary_code` undefined. As a result we never hit this case.

### DIFF
--- a/front/src/applications/operationalStudies/hooks/__tests__/useSetupItineraryForTrainUpdate.spec.ts
+++ b/front/src/applications/operationalStudies/hooks/__tests__/useSetupItineraryForTrainUpdate.spec.ts
@@ -226,20 +226,20 @@ describe('updatePathStepsFrom', () => {
         {
           id: 'whatev-0',
           trigram: 'GE',
-          secondary_code: 'BV',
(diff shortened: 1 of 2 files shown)


clarani (approved):
LGTM :white_check_mark:

PR #9763, opened 2024-11-18
front: leave ch code null when unspecified in XML import

emersion:
An empty string will specifically select OPs without any secondary
code. However that's very unlikely that's what the user wants: in
the French infrastructure, all OPs have a secondary code. Instead,
use undefined to match any secondary code.

While at it, drop the empty string default for ciCode: this is
always guaranteed to be a string (because split() always returns
an array with at least one element) so it's dead code.

To test: edit an XML file to remove a ch code (e.g. replace `SG/BV` with `SG`), then import it.

Also includes a cleanup commit.

### DIFF
--- a/front/src/applications/operationalStudies/types.ts
+++ b/front/src/applications/operationalStudies/types.ts
@@ -49,7 +49,7 @@ export type TrainScheduleImportConfig = {
 };
 
 export type CichDictValue = {
-  ciCode: number | string;
+  ciCode: number;
   chCode?: string;
 };
 
--- a/front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
+++ b/front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
@@ -157,10 +157,9 @@ const ImportTrainScheduleConfig = ({
       } as TrainScheduleImportConfig);
     }
   }
-  // EXTRACT-CI-CH-CODE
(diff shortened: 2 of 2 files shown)


RomainValls:
Does that mean we don't want the OP to appear in the pathSteps, but only in the "allWaypoints", in case the ch code isn't specified in the xml file ? 
If yes, tested, lgtm 👍

this is before removing the ch code for la freissinouse :
Capture d’écran 2024-11-19 à 14 46 32

this is after removing it : 
Capture d’écran 2024-11-19 à 14 51 05

emersion:
> Does that mean we don't want the OP to appear in the pathSteps, but only in the "allWaypoints", in case the ch code isn't specified in the xml file ?

This sounds like a separate, orthogonal bug: when `secondary_code` is left null in a `PathItemLocation`, the backend matches it with any OP's secondary code, but the times and stops table doesn't.

I believe this bug happens on the `dev` branch without this patch.

kmer2016:
> This sounds like a separate, orthogonal bug: when `secondary_code` is left null in a `PathItemLocation`, the backend matches it with any OP's secondary code, but the times and stops table doesn't.
> 
> I believe this bug happens on the `dev` branch without this patch.

You're absolutely right, this bug is indeed already present on the dev branch. It seems [this bug ticket](https://github.com/OpenRailAssociation/osrd/issues/9641) track it

kmer2016 (approved):

RomainValls (approved):

PR #9764, opened 2024-11-18
front: use matchPathStepAndOp() in upsertPathStepsInOPs()

emersion:
matchPathStepAndOp() is more general than special-casing UICs. It should make upsertPathStepsInOPs() work even if steps are defined in terms of trigrams or OP IDs.


clarani (approved):
LGTM :white_check_mark:

PR #9775, opened 2024-11-19
editoast: add field weight to op

### LINKED ISSUE #9746: editoast: update railjson to add field weight in OperationalPoints
# Description

Add a new Operational Point weight field in Railjson.
The purpose of this field is to mark which PRs are important. 
The heavier the weight, the more important the PR, and vice versa. 
This issue will be dealt with in 2 parts. 

# Implementation plan
 
Part 1: adding the field to the Railjson model (#9775)
Part 2: Adding DB weight data to dags.

younesschrifi:
Part of #9746

### DIFF
--- a/front/tests/assets/infra/infra.json
+++ b/front/tests/assets/infra/infra.json
@@ -1,5 +1,5 @@
 {
-  "version": "3.4.12",
+  "version": "3.4.13",
   "track_sections": [
     {
       "id": "TA1",
--- a/python/railjson_generator/railjson_generator/schema/infra/operational_point.py
+++ b/python/railjson_generator/railjson_generator/schema/infra/operational_point.py
@@ -9,13 +9,15 @@ class OperationalPoint:
     label: str
     trigram: str
     parts: List
+    weight: Optional[int]
     uic: int
 
-    def __init__(self, label: str, trigram: Optional[str] = None, uic: int = 0):
+    def __init__(self, label: str, trigram: Optional[str] = None, uic: int = 0, weight: Optional[int] = None):
         self.label = label
         self.trigram = trigram or label[:3].upper()
         self.parts = list()
         self.uic = uic
+        self.weight = weight
 
     def add_part(self, track, offset):
         op_part = OperationalPointPart(self, offset)
--- a/core/osrd-railjson/src/main/java/fr/sncf/osrd/railjson/schema/infra/RJSInfra.java
+++ b/core/osrd-railjson/src/main/java/fr/sncf/osrd/railjson/schema/infra/RJSInfra.java
@@ -18,7 +18,7 @@ public class RJSInfra {
     public static final JsonAdapter<RJSInfra> adapter =
             new Moshi.Builder().add(ID.Adapter.FACTORY).build().adapter(RJSInfra.class);
 
-    public static final transient String CURRENT_VERSION = "3.4.11";
+    public static final transient String CURRENT_VERSION = "3.4.13";
 
     /** The version of the infra format used */
     public String version;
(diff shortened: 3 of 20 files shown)


shenriotpro (approved):
LGTM for tests/

editoast/migrations/2024-11-19-143456_add_weight_to_operational_point/up.sql
```diff
@@ -0,0 +1,14 @@
+UPDATE infra_object_operational_point
+SET data = jsonb_set(
+    data,
+    '{weight}',
+    'null'::jsonb,
+    true
+);
```
flomonster: Since the field  has a default value it does not require a migration.

front/tests/assets/infra/infra.json
```diff
@@ -6757,7 +6757,8 @@
           "uic": 4,
           "name": "Mid_East_station"
         }
-      }
+      },
+      "weight": null
```
flomonster:
```suggestion
```

Same, it should not be required.

flomonster (changes requested):
Weirdly this file: `osrd-railjson/src/main/java/fr/sncf/osrd/railjson/schema/infra/RJSInfra.java` has a version set to `3.4.11`. Can you check? If it 's not use we can refacto this in another PR.

python/railjson_generator/railjson_generator/schema/infra/operational_point.py
```diff
@@ -9,13 +9,15 @@ class OperationalPoint:
...
     parts: List
+    weight: int
     uic: int
 
     def __init__(self, label: str, trigram: Optional[str] = None, uic: int = 0):
         self.label = label
         self.trigram = trigram or label[:3].upper()
         self.parts = list()
         self.uic = uic
+        self.weight = None
```
woshilapin: nitpick: Would it make sense to do that instead?

```suggestion
    def __init__(self, label: str, trigram: Optional[str] = None, uic: int = 0, weight: Optional[int] = None):
        self.label = label
        self.trigram = trigram or label[:3].upper()
        self.parts = list()
        self.uic = uic
        self.weight = weight
```

woshilapin (approved):
Nice job.

I believe all commits should be squashed together before merge since they all relate to the same atomic modification. Also, can you edit the title of the PR to not let the default branch name?

SharglutDev (approved):
Lgtm for front (not tested)

Erashin (approved):
Is the PR title fine? Not used to see this format.
Core lgtm otherwise.

Maymanaf (approved):
LGTM for e2e tests 
You just need to wait for this [PR](https://github.com/OpenRailAssociation/osrd/pull/9913) to be merged before

flomonster (approved):
LGTM

Khoyo:
> Is the PR title fine? Not used to see this format.

The default title for multi-commit pr is the name of the branch (instead of the commit title). It's not an ideal title but they don't end up in the git history anyway as we do fast-forward merges only

emersion:
They do end up in release notes though (and in general are not very readable).

younesschrifi:
Thank you @emersion for your comment. I'll do better next time :)

woshilapin (changes requested):
I'm sorry, I removed this PR from the merge queue.

Sure, most commits are separated which helps for review, but this doesn't make for a reasonable atomic change. So before merging, I believe these should be squashed. For example, changing the schemas in `osrd_schemas` without changing the corresponding schemas in `editoast` in the same commit is a mistake to me. That's the reason why we do have a monorepo and change can be done everywhere in one operation. What do you think?

And yes, sure, the title of the PR is not kept around in the git repository, but it's still kept in the Github where all the other PR titles respects some kind of standard formatting. And they end up in the release.

I could also argue that the description of the PR doesn't really help understand the full context (it links to an issue where there is nothing but a title).

Most of the contributions guidelines are [here](https://osrd.fr/en/docs/guides/contribute/contribute-code/share-changes/). I know we not always follow them to the line, but here, there are multiple problems and I believe some of them should be addressed (PR title, PR description, non-atomic commits).

woshilapin (approved):

PR #9777, opened 2024-11-19
display total length and running time in simulation tabs

### LINKED ISSUE #9678: lmr: front: display total length and running time in simulation tabs
# Description

Users requested to have a quick overview of the total length and duration for the different simulation.

# AC

- [ ] This UI has been implemented in the simulation tabs, with total length and trip duration under the date and time of the simulation

Image

kmer2016:
closes #9678

### DIFF
--- a/front/public/locales/en/translation.json
+++ b/front/public/locales/en/translation.json
@@ -50,6 +50,12 @@
         "title": "Failed to save modifications",
         "text": "Failed to publish your modifications"
       }
+    },
+    "units": {
+      "day": "d",
+      "hour": "h",
+      "km": "km",
+      "minute": "min"
     }
   },
   "Editor": {
--- a/front/public/locales/fr/translation.json
+++ b/front/public/locales/fr/translation.json
@@ -50,6 +50,12 @@
         "title": "Echec de l'enregistrement des modifications",
         "text": "Vos modifications n'ont pas été publiées"
       }
+    },
+    "units": {
+      "day": "j",
+      "hour": "h",
+      "km": "km",
+      "minute": "min"
     }
   },
   "Editor": {
--- a/front/src/applications/stdcm/components/StdcmResults/StdcmSimulationNavigator.tsx
+++ b/front/src/applications/stdcm/components/StdcmResults/StdcmSimulationNavigator.tsx
@@ -4,8 +4,9 @@ import { useTranslation } from 'react-i18next';
 
 import useHorizontalScroll from 'applications/stdcm/hooks/useHorizontalScroll';
 import type { StdcmSimulation } from 'applications/stdcm/types';
-import { hasConflicts } from 'applications/stdcm/utils/simulationOutputUtils';
-import { formatDateToString } from 'utils/date';
+import { hasConflicts, hasResults } from 'applications/stdcm/utils/simulationOutputUtils';
+import { formatDateToString, formatTimeDifference } from 'utils/date';
+import { mmToKm } from 'utils/physics';
 
 export const SIMULATION_ITEM_CLASSNAME = 'simulation-item';
 const ITEM_TO_SHOW_COUNT_ON_SCROLL = 3;
@@ -27,7 +28,7 @@ const StdcmSimulationNavigator = ({
   isCalculationFailed,
   onSelectSimulation,
 }: StdcmSimulationNavigatorProps) => {
-  const { t } = useTranslation('stdcm', { keyPrefix: 'simulation.results' });
+  const { t } = useTranslation();
 
   const { scrollableRef, showLeftBtn, showRightBtn, scrollLeft, scrollRight } = useHorizontalScroll(
     SIMULATION_ITEM_CLASSNAME,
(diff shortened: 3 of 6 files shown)


theocrsb:
Shouldn't the text be on one line?

mockup: 
image

app: 
image

kmer2016:
> Shouldn't the text be on one line?
> 
> mockup: image
> 
> app: image

Good catch! It seems this issue was introduced during the rebase !

theocrsb:
I think a small 10px margin is missing
mockup:
image

app: 
image

theocrsb (approved):
LGTM and tested

front/public/locales/en/translation.json
```diff
@@ -50,6 +50,11 @@
         "title": "Failed to save modifications",
         "text": "Failed to publish your modifications"
       }
+    },
+    "units": {
+      "day": "j",
```
SharglutDev: You reversed the en and fr trad :)

front/src/applications/stdcm/components/StdcmResults/StdcmSimulationNavigator.tsx
```diff
@@ -53,40 +54,64 @@ const StdcmSimulationNavigator = ({
...
-              <div className="simulation-name">
-                <div>
-                  {outputs && !hasConflicts(outputs)
-                    ? t('simulationName.withOutputs', { id })
-                    : t('simulationName.withoutOutputs')}
+          {simulationsList?.map(({ id, creationDate, outputs }, index) => {
+            let formatedTotalLength = '';
+            let formatedTripDuration = '';
+
+            if (hasResults(outputs) && outputs.results.simulation.final_output.times.length > 0) {
```
SharglutDev: If we have an output with result we are sure we will have at least 2 number in times no ?

front/src/applications/stdcm/components/StdcmResults/StdcmSimulationNavigator.tsx
```diff
@@ -53,40 +54,64 @@ const StdcmSimulationNavigator = ({
...
+              const { results } = outputs;
+              const lastPointTime = results.simulation.final_output.times.at(-1)!;
+              const departureTimeInTs = new Date(results.departure_time).getTime();
```
SharglutDev:
```suggestion
              const departureTimeInMs = new Date(results.departure_time).getTime();
```

front/public/locales/fr/translation.json
```diff
@@ -50,6 +50,11 @@
         "title": "Echec de l'enregistrement des modifications",
         "text": "Vos modifications n'ont pas été publiées"
       }
+    },
+    "units": {
+      "day": "d",
+      "hour": "h",
+      "minute": "min"
```
SharglutDev: Do we want the trad for these if they are the same between fr and en ? If yes, we should also add `km`

front/src/applications/stdcm/components/StdcmResults/StdcmSimulationNavigator.tsx
```diff
@@ -27,7 +28,7 @@ const StdcmSimulationNavigator = ({
   isCalculationFailed,
   onSelectSimulation,
 }: StdcmSimulationNavigatorProps) => {
-  const { t } = useTranslation('stdcm', { keyPrefix: 'simulation.results' });
+  const { t } = useTranslation();
```
SharglutDev, line 31: Why did you change this ?

front/src/styles/scss/applications/stdcm/_results.scss
```diff
@@ -75,13 +84,15 @@
...
+              margin-left: 8px;
               color: rgb(60, 202, 128);
-              padding-left: 8px;
             }
           }
 
           &.selected {
-            .creation-date {
+            .simulation-metadata {
               color: rgb(121, 118, 113);
```
SharglutDev: Not coming from your work but 

```suggestion
              color: var(--grey50);
```

front/src/styles/scss/applications/stdcm/_results.scss
```diff
@@ -97,7 +108,8 @@
           }
 
           &.retained {
-            .simulation-name {
+            .simulation-name,
+            .simulation-metadata {
               color: rgb(11, 114, 60);
```
SharglutDev: Same here

```suggestion
              color: var(--success60);
```

front/src/styles/scss/applications/stdcm/_results.scss
```diff
@@ -51,12 +51,11 @@
         overflow-x: hidden;
 
         .simulation-item {
-          width: 125px;
           display: inline-block;
-          margin: 9px 64px 0 0;
+          margin: 9px 39px 0 0;
```
SharglutDev: Not your fault, but there is no margin-top here

image

```suggestion
          margin-right: 39px;
```

front/src/styles/scss/applications/stdcm/_results.scss
```diff
@@ -97,7 +108,8 @@
           }
 
           &.retained {
-            .simulation-name {
+            .simulation-name,
```
SharglutDev, line 112: Still not your fault but when a simulation is retained, the margin right is not the same

image

```suggestion
            margin-right: 29px;
            
            .simulation-name,
```

front/src/styles/scss/applications/stdcm/_results.scss
SharglutDev, line 1: On screen with a resolution less than ~1390px, it goes like this : 

image

SharglutDev (commented):
Thank you for the PR, left a few comments

front/src/styles/scss/applications/stdcm/_results.scss
SharglutDev: Ok just had a call with @thibautsailly. Turns out that the box of the navigator header should always have a width of 160px. The underline was selected/retained should follow this width.

And the gap between each header should be 32px (mock up should have been updated).

Sorry for these requests but I think it's a good moment to update this since you are updating this component.

front/src/applications/stdcm/components/StdcmResults/StdcmSimulationNavigator.tsx
kmer2016: Good question, I wasn't sure, so I added the check to be safe.

front/src/applications/stdcm/components/StdcmResults/StdcmSimulationNavigator.tsx
kmer2016: I initially kept it this way because I was translating duration units directly in the component and they are in another namespace, but I later moved the translation inside the formatting function.

Now, I’m leaving it as is because I added 'Km' to the translation dictionary.

front/public/locales/fr/translation.json
```diff
@@ -50,6 +50,12 @@
         "title": "Echec de l'enregistrement des modifications",
         "text": "Vos modifications n'ont pas été publiées"
       }
+    },
+    "units": {
+      "day": "j",
+      "hour": "h",
+      "minute": "min",
+      "km": "km"
```
SharglutDev: For alphabetical order :) (same for en)

```suggestion
      "km": "km",
      "minute": "min"
```

front/src/styles/scss/applications/stdcm/_results.scss
SharglutDev: can't reproduce anymore

SharglutDev (approved):
Lgtm and tested, good job ! You can resolve the last comment when fixed and merge :)

PR #9778, opened 2024-11-19
Fix speed limit tags for stdcm

### LINKED ISSUE #9682: lmr: front: shorten the composition code list by only using the ones from the LMR infrastructure
# Description

Currently the composition code list in LMR is composed by all codes in OSRD, including all the ones from all the infrastructures. Only the ones associated to the infrastructure LMR is using are relevant in that case.

# AC

- [ ] The composition codes visible and selectable in LMR are the ones from the infrastructure LMR is based on

flomonster:
The list of speed limits was invalid for STDCM.

close https://github.com/OpenRailAssociation/osrd/issues/9682

### DIFF
--- a/front/src/applications/stdcm/consts.ts
+++ b/front/src/applications/stdcm/consts.ts
@@ -19,14 +19,6 @@ export const COMPOSITION_CODES = [
   'ME140',
   'MV160',
   'MVGV',
-  'Marchandise',
-  'Marchandise - MA100',
-  'Marchandise - MA80',
-  'Marchandise - MA90',
-  'Messagerie - ME100',
-  'Messagerie - ME120',
-  'Messagerie - ME140',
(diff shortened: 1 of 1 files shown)


SharglutDev:
There might an intersection with this issue ? https://github.com/OpenRailAssociation/osrd/issues/9682

Do you want to take care of it ?

maelysLeratRosso (approved):
LGTM

emersion (approved):

flomonster:
> There might an intersection with this issue ? #9682
> 
> Do you want to take care of it ?

After discussion with @bougue-pe, we don't need to check speed limit tags in the infra for STDCM. We only need a fixed list. For now the list is hardcoded, late it will be nice to handle it in the backend.

theocrsb (approved):

PR #9782, opened 2024-11-19
front: fix typing in i18n-api-errors script

emersion:
_See individual commits._

Prerequisite for switching to NPM.

### DIFF
--- a/front/scripts/i18n-api-errors.ts
+++ b/front/scripts/i18n-api-errors.ts
@@ -1,6 +1,7 @@
 import SwaggerParser from '@apidevtools/swagger-parser';
 import i18next from 'i18next';
 import { noop } from 'lodash';
+import type { OpenAPI } from 'openapi-types';
 
 // relative the project's root
 const openapi_path = '../editoast/openapi.yaml';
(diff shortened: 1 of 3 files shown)


front/scripts/i18n-api-errors.ts
```diff
@@ -65,19 +66,21 @@ async function checkI18N(error: any): Promise<string[]> {
...
-    const api = await SwaggerParser.validate(openapi_path);
+    const api: OpenAPI.Document = await SwaggerParser.validate(openapi_path);
+
+    if (!('openapi' in api)) throw new Error('Expected an OpenAPI schema');
 
     // Do some checks on the generic error
-    if (!api.components.schemas.EditoastError)
-      throw new Error(`"EditoastError" can't be found in "components > schemas"`);
-    if (!api.components.schemas.EditoastError.oneOf)
+    const EditoastError = api?.components?.schemas?.EditoastError;
```
SharglutDev: Would you consider to use a small case for this const ?

```suggestion
    const editoastError = api?.components?.schemas?.EditoastError;
```

front/scripts/i18n-api-errors.ts
```diff
@@ -65,19 +66,21 @@ async function checkI18N(error: any): Promise<string[]> {
...
+    if (!EditoastError) throw new Error(`"EditoastError" can't be found in "components > schemas"`);
+    if (!('oneOf' in EditoastError) || !EditoastError.oneOf)
```
SharglutDev: This is painful 😮‍💨

emersion: This is because `EditoastError` could be something else: OpenAPI allows schemas to be `{ "oneOf": … }` but also `true` or `{ "items": … }`.

front/scripts/i18n-api-errors.ts
emersion: Sure. This represents a schema type so I originally just kept the uppercase letter, but using a lowercase letter is fine by me as well.

Yohh (approved):

SharglutDev (approved):

PR #9784, opened 2024-11-19
front: switch oazapfts fork to osrd-project repository

emersion:
Fixes two issues:

- The commit of anisometropie/oazapfts that we were using was not in any branch. The master branch has been forced pushed to. This makes it tricky to browse the code actually used in OSRD, and the commit may be GC'ed since it's unreachable. The master branch's latest commit doesn't work with OSRD.
- The "prepare" script was running after installation, even though the fork has the built artifacts checked in. This is unnecessary and causes issues with NPM ("husky" wants a .git directory, it's there with yarn but NPM strips it). Fix for this in https://github.com/osrd-project/oazapfts/commit/3cdfac73c2615b923ba1cde9027ea096bc56ac54.

Additionally, a fork in a personal repository makes it difficult for others to push fixes.

(Alternatively, we could decide to just switch to upstream oazapfts, since we don't use the buggy type.)

### DIFF
--- a/front/package.json
+++ b/front/package.json
@@ -158,7 +158,7 @@
     "react-error-overlay": "6.0.9",
     "@nivo/line": "^0.80.0",
     "@nivo/core": "^0.80.0",
-    "oazapfts": "anisometropie/oazapfts#master"
+    "oazapfts": "osrd-project/oazapfts#master"
   },
   "scripts": {
     "compile": "tsc -b",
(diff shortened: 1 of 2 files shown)


anisometropie (approved):
Thanks, generate-types tested :heavy_check_mark:

Yohh (approved):

kmer2016 (approved):
Nice work !

PR #9785, opened 2024-11-19
editoast: work-schedules: add endpoints for easier edition and viewing

eckter:
Would fix 2/3 of https://github.com/osrd-project/osrd-confidential/issues/838, the rest is changing the etl for batched imports.

I've added a few more endpoints than just what was in the ticket, to have a complete workflow:

1. `POST /work_schedules/group`: create a new group (name can be empty and autofilled (the unique name thing is just painful))
1. `GET /work_schedules/group`: lists all the group ids
2. `PUT /work_schedules/group/42`: insets a work schedule into an existing group
3. `GET /work_schedules/group/42`: lists all the work schedules in the group
4. `DELETE /workçschedules/group/42`: deletes the group and its work schedules

What I **have not** added but could have been nice:

1. More details in `GET /work_schedules/group` (name, creation date, ...)
2. `GET /work_schedules` to list them all
3. Any kind of endpoint dealing with individual work schedules

### DIFF
--- a/editoast/src/views/work_schedules.rs
+++ b/editoast/src/views/work_schedules.rs
@@ -1,19 +1,4 @@
-use axum::extract::Json;
-use axum::extract::State;
-use axum::Extension;
-use chrono::DateTime;
-use chrono::Utc;
-use derivative::Derivative;
-use editoast_authz::BuiltinRole;
-use editoast_derive::EditoastError;
-use editoast_models::DbConnectionPoolV2;
-use serde::de::Error as SerdeError;
-use serde::Deserialize;
-use serde::Serialize;
-use std::result::Result as StdResult;
-use thiserror::Error;
-use utoipa::ToSchema;
-
+use super::pagination::PaginatedList;
 use crate::core::pathfinding::TrackRange as CoreTrackRange;
 use crate::error::InternalError;
 use crate::error::Result;
@@ -21,24 +6,62 @@ use crate::models::prelude::*;
 use crate::models::work_schedules::WorkSchedule;
 use crate::models::work_schedules::WorkScheduleGroup;
 use crate::models::work_schedules::WorkScheduleType;
+use crate::views::operational_studies::Ordering;
+use crate::views::pagination::PaginationQueryParam;
+use crate::views::pagination::PaginationStats;
 use crate::views::path::projection::Intersection;
 use crate::views::path::projection::PathProjection;
 use crate::views::AuthenticationExt;
 use crate::views::AuthorizationError;
-use crate::AppState;
-use editoast_schemas::infra::{Direction, TrackRange};
+use axum::extract::Json;
+use axum::extract::Path;
+use axum::extract::Query;
+use axum::extract::State;
+use axum::response::IntoResponse;
+use axum::Extension;
+use chrono::DateTime;
+use chrono::Utc;
+use derivative::Derivative;
+use editoast_authz::BuiltinRole;
+use editoast_derive::EditoastError;
+use editoast_models::DbConnectionPoolV2;
+use editoast_schemas::infra::Direction;
+use editoast_schemas::infra::TrackRange;
+use serde::de::Error as SerdeError;
+use serde::Deserialize;
+use serde::Serialize;
+use std::result::Result as StdResult;
+use thiserror::Error;
+use utoipa::IntoParams;
+use utoipa::ToSchema;
+use uuid::Uuid;
 
 crate::routes! {
     "/work_schedules" => {
         create,
         "/project_path" => project_path,
+        "/group" => {
+            create_group,
+            list_groups,
+            "/{id}" => {
+                delete_group,
+                get_group,
+                put_in_group,
+            },
+        },
     },
 }
 
 editoast_common::schemas! {
-    WorkScheduleCreateForm,
-    WorkScheduleCreateResponse,
+    WorkSchedule,
     WorkScheduleItemForm,
+    WorkScheduleType,
+}
+
+#[derive(IntoParams, Deserialize)]
+struct WorkScheduleGroupIdParam {
+    /// A work schedule group ID
+    id: i64,
 }
 
 #[derive(Debug, Error, EditoastError)]
@@ -47,6 +70,9 @@ enum WorkScheduleError {
     #[error("Name '{name}' already used")]
     #[editoast_error(status = 400)]
     NameAlreadyUsed { name: String },
+    #[error("Work schedule group '{id}' not found")]
+    #[editoast_error(status = 404)]
+    WorkScheduleGroupNotFound { id: i64 },
 }
 
 pub fn map_diesel_error(e: InternalError, name: impl AsRef<str>) -> InternalError {
@@ -136,49 +162,43 @@ struct WorkScheduleCreateResponse {
 #[utoipa::path(
     post, path = "",
     tag = "work_schedules",
-    request_body = WorkScheduleCreateForm,
+    request_body = inline(WorkScheduleCreateForm),
     responses(
-        (status = 201, body = WorkScheduleCreateResponse, description = "The id of the created work schedule group"),
+        (status = 201, body = inline(WorkScheduleCreateResponse), description = "The id of the created work schedule group"),
     )
 )]
 async fn create(
-    State(app_state): State<AppState>,
+    State(db_pool): State<DbConnectionPoolV2>,
     Extension(auth): AuthenticationExt,
     Json(WorkScheduleCreateForm {
         work_schedule_group_name,
         work_schedules,
     }): Json<WorkScheduleCreateForm>,
 ) -> Result<Json<WorkScheduleCreateResponse>> {
-    let authorized = auth
-        .check_roles([BuiltinRole::WorkScheduleWrite].into())
-        .await
-        .map_err(AuthorizationError::AuthError)?;
-    if !authorized {
-        return Err(AuthorizationError::Unauthorized.into());
-    }
+    // Create the group (using the method for the create group endpoint)
+    let work_schedule_group = create_group(
+        State(db_pool.clone()),
+        Extension(auth),
+        Json(WorkScheduleGroupCreateForm {
+            work_schedule_group_name: Some(work_schedule_group_name),
+        }),
+    )
+    .await?;
 
-    let db_pool = app_state.db_pool_v2.clone();
     let conn = &mut db_pool.get().await?;
-
-    // Create the work_schedule_group
-    let work_schedule_group = WorkScheduleGroup::changeset()
-        .name(work_schedule_group_name.clone())
-        .creation_date(Utc::now())
-        .create(conn)
-        .await;
-    let work_schedule_group =
-        work_schedule_group.map_err(|e| map_diesel_error(e, work_schedule_group_name))?;
 
     // Create work schedules
     let work_schedules_changesets = work_schedules
         .into_iter()
-        .map(|work_schedule| work_schedule.into_work_schedule_changeset(work_schedule_group.id))
+        .map(|work_schedule| {
+            work_schedule.into_work_schedule_changeset(work_schedule_group.work_schedule_group_id)
+        })
         .collect::<Vec<_>>();
     let _work_schedules: Vec<_> =
         WorkSchedule::create_batch(conn, work_schedules_changesets).await?;
 
     Ok(Json(WorkScheduleCreateResponse {
-        work_schedule_group_id: work_schedule_group.id,
+        work_schedule_group_id: work_schedule_group.work_schedule_group_id,
     }))
 }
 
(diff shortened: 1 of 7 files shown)


editoast/src/views/work_schedules.rs
```diff
@@ -26,19 +10,56 @@ use crate::views::path::projection::PathProjection;
 use crate::views::AuthenticationExt;
 use crate::views::AuthorizationError;
 use crate::AppState;
+use axum::extract::State;
+use axum::extract::{Json, Path};
```
woshilapin: nitpick: Can you unmerge the `use`, it's preferable for the diff (but [hard to enforce in the formatter without being a pain for developers](https://github.com/OpenRailAssociation/osrd/pull/7049))

front/public/locales/fr/errors.json
SharglutDev, line 1: Not trad for `en` ?

editoast/src/views/work_schedules.rs
```diff
@@ -275,6 +294,226 @@ async fn project_path(
...
+        .into_iter()
+        .map(|work_schedule| work_schedule.id);
+
+    // Delete them
+    use crate::models::DeleteBatch;
+    let conn = &mut db_pool.get().await?;
+    WorkSchedule::delete_batch(conn, work_schedule_ids).await?;
+
+    // Delete the group itself
+    WorkScheduleGroup::delete(&work_schedule_group, conn).await?;
```
woshilapin: Should we run all of this into a transaction (or at least the `delete_batch` and `delete`)? I wonder how much of a problem it is if remove all of the work schedule and miss the removal of the work schedule group? @eckter Any opinion? Feel free to resolve if you think that is not a problem.

editoast/src/views/work_schedules.rs
```diff
@@ -275,6 +294,226 @@ async fn project_path(
...
+        WorkScheduleError::WorkScheduleGroupNotFound { id: group_id }
+    })
+    .await?;
+
+    // Create work schedules
+    let work_schedules_changesets = work_schedules
+        .into_iter()
+        .map(|work_schedule| work_schedule.into_work_schedule_changeset(group_id))
+        .collect::<Vec<_>>();
+    let work_schedules = WorkSchedule::create_batch(conn, work_schedules_changesets).await?;
```
woshilapin: Same here, maybe we need a transaction, because I'm not sure what happens if the work schedule group is deleted in between?

woshilapin (approved):
Nice work, just some comments about if we should do more transactional operations or not.

editoast/src/views/work_schedules.rs
eckter: I didn't even know transactions were a thing in this context :smile: 

I don't think it would be an issue, but it would still generally *make sense* to run a transaction. I'll do the change.

editoast/src/views/work_schedules.rs
eckter: Fixed in f937a335fad70ebc83e5bc3a0992a73e553d2b67

It would have created "orphaned" work schedules that couldn't be used, queried, or deleted

editoast/src/views/work_schedules.rs
eckter: Fixed in f937a335fad70ebc83e5bc3a0992a73e553d2b67

Please let me know if I did the transaction wrong, I did the same thing as in other places

editoast/src/views/work_schedules.rs
eckter: Fixed in f937a335fad70ebc83e5bc3a0992a73e553d2b67

editoast/src/views/work_schedules.rs
```diff
@@ -275,6 +295,238 @@ async fn project_path(
...
+#[utoipa::path(
+    post, path = "",
+    tag = "work_schedules",
+    request_body = WorkScheduleGroupCreateForm,
+    responses(
+        (status = 200, body = WorkScheduleGroupCreateResponse, description = "The id of the created work schedule group"),
+    )
+)]
+async fn create_group(
+    State(app_state): State<AppState>,
```
```suggestion
    State(db_pool): State<DbConnectionPoolV2>,
```

I don't think you're using something else from the state, we have a shorthand for this case.

editoast/src/views/work_schedules.rs
```diff
@@ -275,6 +295,238 @@ async fn project_path(
...
+    params(WorkScheduleGroupIdParam),
+    responses(
+        (status = 204, description = "The work schedule group has been deleted"),
+        (status = 404, description = "The work schedule group does not exist"),
+    )
+)]
+async fn delete_group(
+    State(app_state): State<AppState>,
+    Extension(auth): AuthenticationExt,
+    raw_group_id: Path<WorkScheduleGroupIdParam>,
```
```suggestion
    Path(WorkScheduleGroupIdParam { id: group_id }): Path<WorkScheduleGroupIdParam>,
```

or just if you think that's too verbose:

```suggestion
    Path(group_id): Path<i64>,
```

editoast/src/views/work_schedules.rs
```diff
@@ -275,6 +295,238 @@ async fn project_path(
...
+    get, path = "",
+    tag = "work_schedules",
+    responses(
+        (status = 201, body = Vec<i64>, description = "The existing work schedule group ids"),
+    )
+)]
+async fn list_groups(
+    State(app_state): State<AppState>,
+    Extension(auth): AuthenticationExt,
+) -> Result<Json<Vec<i64>>> {
```
leovalais: Could this endpoint be paginated? Considering you're already using `models::List` that won't be a big change, you can use this as a template https://github.com/OpenRailAssociation/osrd/blob/d126f80d29dd7ab816eec1e4d0a9d8615a98be20/editoast/src/views/projects.rs#L189-L212.

I understand that this just returns a list of IDs so I'm not worried about the response size, but:

1. if we add more information to the response like you mentioned in the description, that might become necessary
2. for consistency with other listing endpoints
3. bounding requests size is a good practice (which we don't really follow 🙈, but at least we can try when it's an easy win...)

Wdyt?

editoast/src/views/work_schedules.rs
```diff
@@ -1,44 +1,66 @@
...
 
 editoast_common::schemas! {
+    WorkSchedule,
     WorkScheduleCreateForm,
     WorkScheduleCreateResponse,
+    WorkScheduleGroupCreateForm,
+    WorkScheduleGroupCreateResponse,
+    WorkScheduleInGroupCreateForm,
     WorkScheduleItemForm,
+    WorkScheduleType,
```
leovalais: I believe most of these schemas are only used once, so they can be inlined if you want.

leovalais (commented):
LGTM, a few non-blocking comments :)

editoast/src/views/work_schedules.rs
```diff
@@ -275,6 +295,238 @@ async fn project_path(
...
+            // List work schedules to be deleted
+            let selection_setting = SelectionSettings::new()
+                .filter(move || WorkSchedule::WORK_SCHEDULE_GROUP_ID.eq(group_id));
+            let work_schedule_ids = WorkSchedule::list(&mut conn.clone(), selection_setting)
+                .await?
+                .into_iter()
+                .map(|work_schedule| work_schedule.id);
+
+            // Delete them
+            use crate::models::DeleteBatch;
```
Sh099078: Shouldn't it be with the other imports at the top of the file even though it is only being used there ?

editoast/src/views/work_schedules.rs
```diff
@@ -275,6 +295,238 @@ async fn project_path(
...
+            let conn = &mut db_pool.get().await?;
+            WorkSchedule::delete_batch(conn, work_schedule_ids).await?;
```
Sh099078: They might be automatically deleted along with their group ? If so, that `delete_batch` of the work schedules can be skipped

editoast/src/views/work_schedules.rs
Sh099078: The other handlers could be updated in the same way

editoast/src/views/work_schedules.rs
```diff
@@ -510,4 +762,45 @@ pub mod tests {
...
+        ));
+        app.fetch(request).assert_status(StatusCode::OK);
+
+        // Get the content of the group
+        let request = app.get(&work_schedule_url);
+        let work_schedules = app
+            .fetch(request)
+            .assert_status(StatusCode::OK)
+            .json_into::<Vec<WorkSchedule>>();
+        assert_eq!(1, work_schedules.len());
```
Sh099078, line 787: Maybe check that the work schedule returned matches the work schedule sent in the put request and give one of its fields some random value (uuid in `obj_id` ?) to make sure it cannot be another pre-existing work schedule in the database ? I had issues in other tests where their behavior changed depending on the state of the local database when the tests assumed that a table was empty, that a value for a column with a unique constraint was not already taken etc.

Sh099078 (commented):
Great PR ! I added a few comments but mostly nitpicks

front/public/locales/en/errors.json
```diff
@@ -217,7 +217,8 @@
       "InvalidUrl": "Invalid url '{{url}}'"
     },
     "work_schedule": {
-      "NameAlreadyUsed": "A group of work schedules with '{{name}}' already exists"
+      "NameAlreadyUsed": "A group of work schedules with '{{name}}' already exists",
+      "WorkScheduleGroupNotFound": "No such work schedule group with id={{id}}"
```
SharglutDev: To keep it harmonized with other trads (same for fr)

```suggestion
      "WorkScheduleGroupNotFound": "No such work schedule group with id '{{id}}'"
```

SharglutDev (approved):
Lgtm for front part (not tested)

editoast/src/views/work_schedules.rs
eckter: I'm not sure (it's the first time I really deal with editoast tables).

From what I can see, the `work_schedule_group_id` is just an `i64` without any magic or macro to indicate that it's a foreign key, I don't think the database knows about this link. But that sounds like troubles, maybe there's something missing.

I don't know, @leovalais I'd need the input of someone more familiar

editoast/src/views/work_schedules.rs
eckter: Fixed in a1bd117b3c855cecd04c1dee1aba55d9435fcdd9

editoast/src/views/work_schedules.rs
eckter: Fixed in a1bd117b3c855cecd04c1dee1aba55d9435fcdd9

editoast/src/views/work_schedules.rs
eckter: Fixed in 64d4756f07e49aab97df92b70c1bfa4e38107ae6

editoast/src/views/work_schedules.rs
eckter: Right, I copied what was [there](https://github.com/OpenRailAssociation/osrd/blob/dev/editoast/src/views/train_schedule.rs#L256)

Fixed in a1bd117b3c855cecd04c1dee1aba55d9435fcdd9

editoast/src/views/work_schedules.rs
eckter: Fixed in 61426f6d6f5492a4585d822e904e34bccc38c9a9

editoast/src/views/work_schedules.rs
Sh099078: You can check constraints and foreign keys in the `up.sql` script that generates the table in `migrations` and get some insight about the tables in `tables.rs`. From what I see in the sql file, there is a foreign key with `ON DELETE CASCADE` between work schedules and their group, so only deleting the group should be fine

editoast/src/views/work_schedules.rs
eckter: I don't *really* think it's worth the troubles for the groups, there's not supposed to be a lot of them in the db (probably ranging from 0 to 2 most of the times).

That said, when getting the content of a group it would indeed make sense, I'll change it.

editoast/src/views/work_schedules.rs
Sh099078: Checking the `up.sql` script to get some info about the tables is what I find to be the most readable but if you do, make sure to grep that there isn't an `ALTER TABLE` later on in the migrations that impacts what you are looking for

editoast/src/views/work_schedules.rs
Sh099078: If I am not mistaken in this [comment](https://github.com/OpenRailAssociation/osrd/pull/9785#discussion_r1852372121) the delete_batch and the transaction might not be needed. What do you think @woshilapin ?

editoast/src/views/work_schedules.rs
eckter: Done in dbeed01b77f8549c185e4d08073f49b1c937b14e (just for the group content)

editoast/src/views/work_schedules.rs
leovalais: Thanks for spotting this! There's indeed no need to manually delete all work schedules, deleting the group should cascade.

The fact that it's a foreign key isn't really useful to `diesel` as a query builder, it only cares about which tables can be joined over which fields (cf. `editoast_models/src/tables.rs`). And since `Model` is just a light wrapper over `diesel`, that information isn't there as well.

Even if the foreign key information was kept by `diesel`, I'm not sure what would be the benefits of knowing that at compile time.

What I usually do to know which are the relations is to just open DBeaver and check out the schemas (there's also an UML diagram view which can be useful sometimes).

editoast/src/views/work_schedules.rs
```diff
@@ -275,6 +295,251 @@ async fn project_path(
...
+        .await
+        .map_err(AuthorizationError::AuthError)?;
+    if !authorized {
+        return Err(AuthorizationError::Unauthorized.into());
+    }
+
+    let ordering = ordering_params.ordering;
+    let settings = pagination_params
+        .validate(1000)?
+        .warn_page_size(100)
```
```suggestion
        .validate(100)
```

Since this is a new endpoint I'd suggest to enforce pagination directly. The `validate/warn` exists for some routes that used to not be paginated and for which the frontend expects all the data to be returned at once (like for the rolling stock selector for example).

leovalais (approved):
LGTM, thanks for the changes!

editoast/src/views/work_schedules.rs
eckter: Fixed in https://github.com/OpenRailAssociation/osrd/pull/9785/commits/fadd260d6bd8557a2c7764ed98357c71ca12787e

editoast/src/views/work_schedules.rs
woshilapin: Well if you only have to delete the `work_schedule_group` and the DB will do the rest, then yes, for sure it should be fine without a transaction.

That said, I see that you also have the `retrieve` in the transaction and that might still be useful to keep the `retrieve` then the `delete` in the same transaction. What if you retrieve some information, then something happen in between (modification of the work_schedule for example), then we delete it? Are we sure we want to continue the deletion in this case?

woshilapin: Looking again at the code, I'm actually wondering why we want to `retrieve` first? Is it only for reporting an error in case the identifier doesn't exist? Can't we report that directly from a failure of the `delete`? Same remark applies to `get_group` too.

eckter: Good point, the `delete` method wanted a group itself and not just an ID, but there seems to be other methods for that.

I haven't properly setup editoast on my laptop, I'll fix this tomorrow.

For `get_group` we'd return an empty list instead of a 404 though. We only use the group ID as a filter.

flomonster (approved):

editoast/src/views/work_schedules.rs
eckter: Fixed in https://github.com/OpenRailAssociation/osrd/pull/9785/commits/990a83ea339f97f3f497512bdaf8de4c2368ae91

woshilapin (approved):
Thanks for the improvements.

PR #9786, opened 2024-11-19
front: stdcm: update simulation progress button design

### LINKED ISSUE #9322: Keep the STDCM simulation results visible when a new simulation's running

End users

When running a STDCM simulation, all the previous simulations below are hidden by the loader (actually they are not even displayed anymore). I would be nice to keep them below the loader in order to navigate through the different simulations while the new one's running.

Image

It will enable users to check the previous simulations while the new one's running

_No response_

### LINKED ISSUE #9677: lmr: front: refacto `Calcul en cours` component
# Description

The actual `Calcul en cours` component caused a few interrogations during the users tests. Here's a small refactoring to make this phase clearer.

# Acceptance criteria
  
- [ ] The image and associated has been removed to make the component thiner heightwise
- [ ] The computing component now looks [like this](https://www.sketch.com/s/15e0909c-f17d-4200-a770-b9fc1154fa7e/p/CA8C9E9A-B795-4C87-9CF5-508DD0B3634A/canvas)

Image

- [ ] The `Calcul en cours...`message is animated
- [ ] https://github.com/OpenRailAssociation/osrd/issues/9322
- [ ] The contrast of the différent input fileds are harmonized (currently the numbers input are sharper than the others)

SharglutDev:
close #9677 
fix #9322 

To test popup behavior : 
- Animation is not on sketch yet and its duration will be validated with @thibautsailly on November 25th.
When launching a simulation, the launch button should fade out while the simulation progress popup should fade in from the bottom.
- The popup should be sticky to the bottom if it doesn't fit below the form and be fixed to the top when scrolling far bottom (if you have already a simulation in the results, you can toggle the op list to increase the height and test the behavior)


SharglutDev:
@Maymanaf, tagged you so you can check I don't break e2e tests

front/src/applications/stdcm/components/StdcmLoader.tsx
```diff
@@ -1,42 +1,57 @@
...
-          </div>
-          <img src={stdcmLoaderImg} alt={t('simulation.pendingSimulation')} />
-          <p className="stdcm-loader__img-signature">{t('loaderImageLegend')}</p>
+const StdcmLoader = ({ cancelStdcmRequest }: StdcmLoaderProps) => {
+  const { t } = useTranslation('stdcm');
+
+  useEffect(() => {
+    // Depending on the scroll change the position of the loader between fixed and sticky
+    const handleScroll = () => {
+      const element = document.querySelector('.stdcm-loader') as HTMLElement;
```
emersion: This is a bit fragile: React can re-render the element at any time, in which case our manual DOM changes here would be lost.

Would be nicer to add a state storing whether we're scrolled up or not, and the set a class on the loader element based on that state.

front/src/applications/stdcm/components/StdcmLoader.tsx
```diff
@@ -1,42 +1,57 @@
...
+      const scrollTop = window.scrollY;
+
+      if (scrollTop > window.innerHeight - LOADER_OFFSET) {
```
emersion: Do we need to have a constant for `LOADER_OFFSET`? Could we use a ref and find the height of the HTML element dynamically instead?

front/src/styles/scss/applications/stdcm/_loader.scss
```diff
@@ -1,61 +1,67 @@
...
+  bottom: 32px;
+  z-index: 2;
+  padding: 16px 16px 18px;
+  border-radius: 24px;
+  background-color: var(--white100);
+  width: 466px;
+  box-shadow:
+    0px 6px 21px -5px rgba(24, 68, 239, 0.26),
+    0px 16px 30px -5px rgba(0, 0, 0, 0.16),
+    0px 3px 5px -2px rgba(0, 0, 0, 0.1);
```
theocrsb:
```suggestion
    0 6px 21px -5px rgba(24, 68, 239, 0.26),
    0 16px 30px -5px rgba(0, 0, 0, 0.16),
    0 3px 5px -2px rgba(0, 0, 0, 0.1);
```

front/src/applications/stdcm/components/StdcmLoader.tsx
SharglutDev: You're right, done

front/src/applications/stdcm/components/StdcmLoader.tsx
SharglutDev: The threshold where we want to change is position depends of this offset. Also I made a mistake, it's actually not 32 but 32 * 2 (the offset bottom when it's sticky and the offset top when it's fixed). Do you see a patern I'm missing depending of the component's height ?

SharglutDev:
There is still an issue which needs to be investigated (probably related to your comment @emersion). When the form is long, the threshold for the position switch isn't good anymore because `scrollY` reaches `innerHeight - LOADER_OFFSET * 2` sooner

Maymanaf:
> @Maymanaf, tagged you so you can check I don't break e2e tests

Since the E2E tests for LMR run very quickly, I didn't include a check for the simulation progress to avoid flakiness, as the test might fail to detect it.

front/src/applications/stdcm/components/StdcmLoader.tsx
SharglutDev: I still kept a few `const` like this one. I'm not sure if having a bunch of ref to pass in props would be a better solution. What do you think ?

emersion: In general I dislike hardcoding style-related things in JS, when it really should just be in the CSS. The reason is that (1) it's fragile (depends on font size, padding, margin, text warping, general layout of the page, etc) and (2) it duplicates the information in two spots (CSS and JS).

emersion:
Random side note: really not a fan of these manual size computations. Would really prefer if this were just a simple pure CSS `position: sticky` element.

front/src/applications/stdcm/views/StdcmView.tsx
```diff
@@ -12,13 +12,14 @@ import { replaceElementAtIndex } from 'utils/array';
...
 import StdcmConfig from '../components/StdcmForm/StdcmConfig';
 import StdcmHeader from '../components/StdcmHeader';
-import StdcmLoader from '../components/StdcmLoader';
 import StdcmResults from '../components/StdcmResults';
 import StdcmStatusBanner from '../components/StdcmStatusBanner';
 import useStdcmEnvironment, { NO_CONFIG_FOUND_MSG } from '../hooks/useStdcmEnv';
 import useStdcmForm from '../hooks/useStdcmForm';
 import type { StdcmSimulation } from '../types';
 
+const LAUNCH_BUTTON_FADEOUT_DURATION = 500;
```
clarani: Can you add the unit here in comment (I think it should be ms ?)

front/src/styles/scss/applications/stdcm/_loader.scss
```diff
@@ -1,61 +1,102 @@
...
-  padding: 0 0 48px 338px;
-  background-color: #eff3f5;
-}
 .stdcm-loader {
-  padding: 48px 48px 19px 48px;
-  border-radius: 6px;
-  background-color: rgba(255, 255, 255, 1);
-  width: 546px;
+  z-index: 2;
+  padding: 16px 16px 18px;
```
clarani, line 3: I think we can do this
```suggestion
  padding: 16px 18px;
```

front/src/applications/stdcm/components/StdcmLoader.tsx
```diff
@@ -1,42 +1,104 @@
...
+    const handleScroll = () => {
+      if (!loaderRef.current || !launchButtonRef.current) return;
+
+      const { scrollY, innerHeight } = window;
+
+      const isLoaderFitting =
+        innerHeight - launchButtonRef.current.getBoundingClientRect().top >
+        LOADER_HEIGHT + LOADER_BOTTOM_OFFSET;
+
+      if (launchButtonRef.current && isLoaderFitting) {
```
clarani: You can handle only 3 cases:
```
  if the loader does not fit between the launch button and the bottom of the screen:
    setStatus('fixed-bottom');
    return;
  if the loader should stick to the top of the screen:
    setStatus('fixed-top');
  else:
    setStatus('absolute');
```

front/src/applications/stdcm/types.ts
```diff
@@ -200,3 +200,8 @@ export type StdcmLinkedPathResult = {
 };
 
 export type ExtremityPathStepType = 'origin' | 'destination';
+
+export type LoaderStatus = {
+  status: 'fixed' | 'sticky' | 'absolute';
```
clarani: Maybe we can use `fixed-top` and `fixed-bottom` instead of `fixed` and `sticky`
```suggestion
  status: 'fixed-top' | 'fixed-bottom' | 'absolute';
```

front/src/styles/scss/applications/stdcm/_loader.scss
```diff
@@ -1,61 +1,102 @@
...
+    100% {
       opacity: 1;
-      color: rgba(33, 100, 130, 1);
+    }
+  }
+
+  @keyframes slideUp {
+    0% {
+      // Le translate ne doit se passer que si la popup ne rentre pas dans l'espace restant
+      // La map ne doit jamais bouger quand le formulaire grandit ou retrécit ?
```
clarani:
```suggestion
```

front/src/applications/stdcm/views/StdcmView.tsx
SharglutDev: Actually I can remove this change (the `setTimeout` in `useEffect`), it used to be for something else but forgot to remove it. I also reremoved the `isPending` to activate the `fade-out` class of the button.

front/src/styles/scss/applications/stdcm/_loader.scss
SharglutDev: We can't. `padding: 16px 16px 18px;` stands for 16 top, 16 inline (left and right) and 18 bottom. This won't be the same behavior with `padding: 16px 18px;`

front/src/applications/stdcm/types.ts
SharglutDev: Done

front/src/styles/scss/applications/stdcm/_loader.scss
SharglutDev: Done

front/src/applications/stdcm/components/StdcmLoader.tsx
SharglutDev: Done

SharglutDev: I pushed another fixup to move the `LoadingStatus` state inside the `StdcmLoader` component. It seems to work the same and doesn't rerender all the form when we scroll.

front/src/styles/scss/applications/stdcm/_loader.scss
clarani: My bad !

clarani (approved):
LGTM :white_check_mark:

theocrsb (approved):
LGTM and tested

Maymanaf (approved):
LGTM for e2e tests

clarani (approved):
LGTM :white_check_mark:

PR #9789, opened 2024-11-20
Rename header to skip authorization

flomonster:
Core is not the only service that should bypass the authorization

osrd-chart PR: https://github.com/OpenRailAssociation/osrd-chart/pull/39


leovalais (approved):

Erashin (approved):

Khoyo (approved):

woshilapin (approved):
I realize, isn't this some kind of backdoor in `editoast` (nothing change since the apparition of the `core` header, I realize this PR doesn't make it worse than it already was). What the risk?

flomonster:
> I realize, isn't this some kind of backdoor in `editoast` (nothing change since the apparition of the `core` header, I realize this PR doesn't make it worse than it already was). What the risk?

It's a kind of backdoor but it cannot be exploit since the gateway reject requests with this header and editoast can't be queried from outside it's network.

Another way to do it is to use a token that have the necessary permissions and query the gateway instead.

PR #9795, opened 2024-11-20
front: fix stdcm speedlimitbytag reset

### LINKED ISSUE #9623: STDCM - Empty composition code in PDF in some cases
### What happened?

I make a simulation with a composition code, but I changed the destination after changing the composition code. Then the pdf has no composition code

https://github.com/user-attachments/assets/fc7f156c-a081-4521-91b5-1c19aaebfa6a

I expected to see my code in the pdf

As in the video (Before the video, I did another simulation and refreshed to have already a destination and origin written)

Dev (SNCF)

Google Chrome

Akctarus:
closes #9623


clarani (approved):

PR #9798, opened 2024-11-21
front: update btn map in scenario page and convert rem to px in css map file

theocrsb:

buttons list mockup: https://www.sketch.com/s/2fb064cd-c483-46ac-bd30-72705fd63e75/p/9559F6FC-69EE-4DB8-BE1A-77F60F10C97F/canvas#Inspect

button mockup: https://www.sketch.com/s/7b92c9be-3fa6-4af7-b8c6-66b63fec08fa/p/AB9D17BF-0EBF-4C30-A75D-0D6BB5C0BA68/canvas#Inspect


front/src/styles/scss/common/map/_map.scss
```diff
@@ -7,31 +7,66 @@
...
+  top: 0;
+  right: 0;
+  width: 72px;
+  height: 360px;
+  padding: 12px;
+}
+
+.osrd-btn-map {
+  color: var(--black100);
+  width: 48px;
```
anisometropie: where does 48px come from?
Screenshot_20241121_165841

anisometropie (changes requested):

front/src/styles/scss/common/map/_map.scss
theocrsb: I used the mockup `Graphic update for OSRD#2`
image

front/src/common/Map/Buttons/MapButtons.tsx
```diff
@@ -55,6 +56,7 @@ export default function MapButtons({
   bearing,
   editorProps,
   viewPort: viewportProps,
+  isNewButtons,
```
clarani: Set it to false by default, this way you can remove the props in the components which do not need the new buttons
```suggestion
  isNewButtons = false,
```

front/src/applications/editor/Editor.tsx
```diff
@@ -446,6 +446,7 @@ const Editor = () => {
                   activeTool: toolAndState.tool,
                 }}
                 viewPort={viewport}
+                isNewButtons={false}
```
clarani: You can remove this props if it's set by default to false
```suggestion
```

clarani (commented):
As you are re-writting these buttons, I think it would be better to do a bit of refacto :)
All these buttons are almost the same, the only things that change are the function onClick, the text in the span + its translation, the icon and the data-test-id.

You can create a generic component `MapButton` which will take only these 4-5 props, and in MapButtons, render this new component with the correct props if we are using the newButtons.

If you do this, you won't need to modify all the existing Buttons

front/src/common/Map/Buttons/MapButtons.tsx
theocrsb: Done

front/src/applications/editor/Editor.tsx
theocrsb: Done

theocrsb:
> As you are re-writting these buttons, I think it would be better to do a bit of refacto :) All these buttons are almost the same, the only things that change are the function onClick, the text in the span + its translation, the icon and the data-test-id.
> 
> You can create a generic component `MapButton` which will take only these 4-5 props, and in MapButtons, render this new component with the correct props if we are using the newButtons.
> 
> If you do this, you won't need to modify all the existing Buttons

You're right, I just proposed a refacto in my fixups :)

front/src/styles/scss/common/map/_map.scss
anisometropie: thanks. perhaps link it in the description so we know what you used as reference

theocrsb: Sure

anisometropie:
I think all the icons aren’t centered 
Screenshot_20241128_102413

theocrsb:
> I think all the icons aren’t centered Screenshot_20241128_102413

Thanks, it's done

front/src/styles/scss/common/map/_map.scss
```diff
@@ -7,31 +7,69 @@
...
+  position: absolute;
+  z-index: 2;
+  top: 0;
+  right: 0;
+  width: 72px;
+  height: 360px;
+  padding: 12px;
+}
+
+.osrd-btn-map {
```
clarani: Can you move `.osrd-btn-map` inside osrd-btn-map-container ?
Also, maybe we don't need the prefix `osrd` ? You can replace it by the prefix `new` if you need

clarani (commented):
There is a bug in the old buttons :/ Look at the compass button
image

front/src/styles/scss/common/map/_map.scss
theocrsb: Done

theocrsb:
> There is a bug in the old buttons :/ Look at the compass button image

Thanks, I just fixed

anisometropie (approved):

theocrsb:
It doesn't yet have “hover” and “active” behavior. 
I've just checked with @thibautsailly , the mockup are in progress.

front/src/common/Map/Buttons/MapButton.tsx
```diff
@@ -0,0 +1,42 @@
+import { type ReactNode } from 'react';
+
+import { useTranslation } from 'react-i18next';
+
+import Tipped from 'common/Tipped';
+
+type MapButtonProps = {
+  onClick: () => void;
+  isNewButtons: boolean;
```
clarani: nit: it's the prop of 1 button
```suggestion
  isNewButton: boolean;
```

front/src/styles/scss/common/map/_map.scss
```diff
@@ -18,6 +18,56 @@
...
+    span {
+      display: flex;
+      margin: auto;
+      width: 16px;
+      height: 16px;
+    }
+  }
+}
+
+.new-btn-map-resetviewport {
```
clarani: You can move this class in `new-btn-map-container`

clarani (approved):
LGTM :white_check_mark: Only left 2 small comments, you can close them once you have fixed them and merge without a new approval :+1: :)

front/src/common/Map/Buttons/MapButton.tsx
theocrsb: Done

front/src/styles/scss/common/map/_map.scss
theocrsb: Done

PR #9814, opened 2024-11-21
front: fix typing of TrainScheduleImportConfig.{from,to}

emersion:
These are defined as strings, but we were passing ImportStation objects. The type error was silenced with a "as" keyword.

We need to move the ImportStation type from StationCard to src/applications/operationalStudies/types.ts to use it in TrainScheduleImportConfig.

No functional change - only typing changes.


SharglutDev (approved):
Lgtm and tested

PR #9816, opened 2024-11-21
front: edit timetable trains in macro mode

RomainValls:
Part 2 of https://github.com/osrd-project/osrd-confidential/issues/840

We can now : 
- edit trains from the timetable in macro mode and see the update
- edit trains from NGE and see the updates in the timetable train card

### DIFF
--- a/front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
+++ b/front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, useCallback } from 'react';
+import { useState, useCallback, useEffect } from 'react';
 
 import { ChevronRight } from '@osrd-project/ui-icons';
 import cx from 'classnames';
@@ -24,7 +24,6 @@ import ScenarioLoaderMessage from 'modules/scenario/components/ScenarioLoaderMes
 import TimetableManageTrainSchedule from 'modules/trainschedule/components/ManageTrainSchedule/TimetableManageTrainSchedule';
 import Timetable from 'modules/trainschedule/components/Timetable/Timetable';
 import { useAppDispatch } from 'store';
-import { concatMap, mapBy } from 'utils/types';
 
 import ScenarioDescription from './ScenarioDescription';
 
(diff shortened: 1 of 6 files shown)


front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
```diff
@@ -92,33 +87,30 @@ const ScenarioContent = ({
   }, [scenario, isMacro]);
```
Uriel-Sautron: You can obtain the same result by adding `trainSchedules` in dependencies (use can fix typo un useEffect please `doImport` -> `dtoImport`),

Uriel-Sautron (approved):
Tested, good work

front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
```diff
@@ -64,60 +62,48 @@ const ScenarioContent = ({
     removeTrains,
   } = useScenarioData(scenario, timetable, infra);
 
+  const [editFromTimetable, setEditFromTimetable] = useState(false);
```
emersion: Hm, not a fan of using a state for this, sounds quite fragile… But since all updates trickle down via `useEffect()` chains it's tricky to find out why a timetable has been updated.

Maybe this could be somehow integrated in `useScenarioData()`. Will try to write up a proposal.

front/src/modules/trainschedule/components/ManageTrainSchedule/TimetableManageTrainSchedule.tsx
```diff
@@ -27,23 +28,25 @@ const TimetableManageTrainSchedule = ({
...
   setTrainIdToEdit,
+  dtoImport,
 }: TimetableManageTrainScheduleProps) => {
   const { t } = useTranslation('operationalStudies/manageTrainSchedule');
   const [isWorking, setIsWorking] = useState(false);
 
   const leaveManageTrainSchedule = () => {
     setDisplayTrainScheduleManagement(MANAGE_TRAIN_SCHEDULE_TYPES.none);
     setTrainIdToEdit(undefined);
+    dtoImport();
```
emersion: Since we already call `dtoImport()` in `useUpdateTrainSchedule()`, I think we don't need to call it a second time here?

front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
```diff
@@ -61,62 +60,41 @@ const ScenarioContent = ({
...
+  };
+
   const toggleMicroMacroButton = useCallback(
     (isMacroMode: boolean) => {
       setIsMacro(isMacroMode);
       if (!isMacroMode && collapsedTimetable) {
         setCollapsedTimetable(false);
       }
+      if (isMacroMode) {
+        dtoImport();
```
emersion: Since this is in a `useEffect()` with `collapsedTimetable` as a dependency, the NGE import is performed each time the sidebar is collapsed or expanded, even if nothing has changed. I think we should keep the `dtoImport()` call in a `useEffect()` which doesn't have this dependency.

emersion (approved):
LGTM, nice work!

PR #9817, opened 2024-11-21
front: simplify timestampToHHMMSS()

emersion:
Instead of manually extracting what we want from a larger string with string offsets, format the date.


Synar (approved):
Nice!

clarani (approved):
LGTM

PR #9824, opened 2024-11-22
core: add compatibility for G1 train gauge with FR3.3 track gauge

bougue-pe:
Based on:
* G1 (PPI, literally "pass-everywhere international gauge") is the minimal gauge (see https://en.wikipedia.org/wiki/Loading_gauge)
* the description from internal doc: "FR3.3 englobe le G1 pour des paramètres de circulations égaux."

Some doc is referenced in https://github.com/osrd-project/osrd-dags/issues/300


axrolld (approved):
A wise man once said : "G1 trains drive everywhere !"

Erashin (approved):

PR #9826, opened 2024-11-22
front: fix train edition setup

### LINKED ISSUE #9805: can't set or get the route of an invalid train schedule
### What happened?

The cache seems to override the route of a train schedule when we try to edit a train schedule (this train schedule has been imported and its route set, but the route displayed is from anothers, which i suspect is from my cache):
Image

This train should have this route:
Image

set right route

1. import train schedule in a timetable

Viriato files are from there : https://sncf.sharepoint.com/:u:/r/sites/OSRD644GrpO365/Documents%20partages/Donn%C3%A9es/3-Fichiers%20Viriato/Viriato%20fichiers%202908%202.zip?csf=1&web=1&e=gLgC4H

or edit any train schedule

Recette (SNCF), Dev (SNCF)

Firefox

dev dd9941f

clarani:
closes #9805

Since recently, when opening the view to edit a train, a pathfinding request is automatically launched. But it is launched before the store update, which creates the bug: the pathfinding is launched with the pathsteps stored in the store (and not the ones of the train we want to edit) and once the pathfinding request is finished, we update the store with the result of the request (and erase the path steps of the currently selected train).

To fix this, we can directly update the store before opening the ManageTrainSchedule view.


Caracol3 (approved):
lgtm (tested)

kmer2016 (approved):
LGTM !

PR #9827, opened 2024-11-22
Add unique constraint on group name

flomonster:
Without this constraint some commands does not work properly like `editoast group include`.

### DIFF
--- a/editoast/migrations/2024-11-22-151624_group-name-unique/down.sql
+++ b/editoast/migrations/2024-11-22-151624_group-name-unique/down.sql
@@ -1,0 +1,1 @@
+DROP CONSTRAINT group_name_unique;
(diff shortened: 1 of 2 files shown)


woshilapin (approved):

leovalais:
Looks fine to me, but do we really want groups to have a unique name? I hardly believe that a purely technical constraint like the CLI API is a good enough reason to ask users to never duplicate group names.

This also raises some UX questions: I (a lambda user) create a group with the name "My Group", which gets rejected. How do we display the error? Do we just show "A group named My Group already exists. Please choose another name."? (We can't show the group owner.) This is rather classic for user IDs and we can easily understand why, but for group names, as a user, I can hardly see the rationale behind that.

Wdyt?

PR #9829, opened 2024-11-22
front: add speed space chart e2e test

### LINKED ISSUE #8651: front: GEV e2e tests
# Description

Implement a first e2e test for the GEV component. We're only testing the basics for now, the test will be appended later once we're confident the recent features won't move too much.

# AC

- [ ] [This test case](https://github.com/osrd-project/osrd-tests/issues/70) has been implemented as an e2e test, using a snapshooting method.

Maymanaf:
Closes #8651
~~The visual assertion using the snapshot is only performed on Chrome due to issue #9909~~


front/playwright.config.ts
```diff
@@ -14,6 +14,7 @@ export default defineConfig({
   /* Maximum time one test can run for. */
   timeout: process.env.CI ? 90 * 1000 : 180 * 1000, // 90 seconds in CI, otherwise 180 seconds
   expect: {
+    toHaveScreenshot: { maxDiffPixels: 20 },
```
Yohh: why 20 pixels precisely?

front/tests/pages/op-simulation-results-page-model.ts
```diff
@@ -0,0 +1,57 @@
+import { expect, type Locator, type Page } from '@playwright/test';
+
+class OpSimulationResultPage {
+  readonly page: Page;
+
+  private readonly sscEllipsisButton: Locator;
```
Yohh: what is `ssc`?

front/tests/012-op-simulation-settings-tab.spec.ts
```diff
@@ -165,9 +171,11 @@ test.describe('Simulation Settings Tab Verification', () => {
     // Add the train schedule and verify output results
     await operationalStudiesPage.addTrainSchedule();
     await operationalStudiesPage.returnSimulationResult();
-    await opOutputTablePage.verifyTimeStopsDataSheetVisibility();
     await opTimetablePage.getTrainArrivalTime('11:53');
     await opTimetablePage.clickOnScenarioCollapseButton();
+    await opOutputTablePage.verifyTimeStopsDataSheetVisibility();
+    await simulationResultPage.checkAllSscCheckboxes();
+    await expect(opTimetablePage.speedSpaceChart).toHaveScreenshot('ElectricalProfile-SSC-ON.png');
```
Yohh: png can take a lot of storage, is it possible to use a lighter file extension?

Maymanaf: I tested with `webp`  and `jpeg`  but `toHaveScreenshot()` method only support `png`

front/tests/pages/op-simulation-results-page-model.ts
Maymanaf: SSC stands for Speed-Space Chart. 
To differentiate the locators, as we could add a Time Space Chart later on in the same file.

front/playwright.config.ts
Maymanaf: There’s a small pixel difference between running tests locally and in CI, usually between 1 and 18 pixels. I added a 2-pixel buffer to avoid flaky failures, so the tolerance is 20.

front/tests/pages/op-simulation-results-page-model.ts
Yohh: aaaah ok, I didn't see that

front/tests/pages/op-simulation-results-page-model.ts
```diff
@@ -0,0 +1,57 @@
...
+  private async toggleCheckboxIfUnchecked(index: number): Promise<void> {
+    const checkbox = this.sscCheckboxes.nth(index);
+    const isChecked = await checkbox.isChecked();
+
+    if (!isChecked) {
+      await this.sscCheckmarks.nth(index).click();
+      await expect(checkbox).toBeChecked();
+    } else {
+      await expect(checkbox).toBeChecked();
+    }
```
anisometropie:
```suggestion
    if (!isChecked) {
      await this.sscCheckmarks.nth(index).click();
    } 
    await expect(checkbox).toBeChecked();
```

anisometropie (changes requested):

anisometropie (approved):

anisometropie:
Review & tested, looking good

anisometropie:
btw, the fixup fa8fd20 should not go in the readme commit

Maymanaf:
> btw, the fixup [fa8fd20](https://github.com/OpenRailAssociation/osrd/commit/fa8fd2092c02990d30d7626a4cbc771fa1b46ea7) should not go in the readme commit

Yes my bad , I will fix that

Yohh (commented):
one thing that may be important:

when the tests are launched, snapshots are created with the name you give, but three times each with `-chromium-linux.png` or `-firefox-linux.png` added, so we have yet 16 png files, when I launch them on macos, playwright generates 16 new files with `-chromium-darwin.png` or `-firefox-darwin.png`

Maymanaf:
> one thing that may be important:
> 
> when the tests are launched, snapshots are created with the name you give, but three times each with `-chromium-linux.png` or `-firefox-linux.png` added, so we have yet 16 png files, when I launch them on macos, playwright generates 16 new files with `-chromium-darwin.png` or `-firefox-darwin.png`

Good catch! I think we can add a control to run the snapshot assertion only on Linux. The method toHaveScreenshot() generates images based on the operating system and browser, and we can't maintain multiple screenshots for every OS and browser combination
What do you think ?

Yohh:
that's the wisest solution I think

front/tests/pages/op-simulation-results-page-model.ts
```diff
@@ -0,0 +1,56 @@
+import { expect, type Locator, type Page } from '@playwright/test';
+
+class OpSimulationResultPage {
+  readonly page: Page;
+
+  private readonly sscEllipsisButton: Locator;
```
emersion: "ssc" is confusing. What is it referring to? An explicit name might be better. Or a different class.

front/tests/pages/common-page-model.ts
```diff
@@ -38,6 +41,10 @@ class CommonPage {
       await this.viteOverlay.evaluate((node) => node.setAttribute('style', 'display:none;'));
     }
   }
+
+  async closeToastNotification() {
+    if (this.toastSNCF) await this.closeToastButton.click();
```
emersion: `toastSNCF` will always be truthy. `getByTestId()` never returns null.

In other words, this won't work to check if there is a toast currently displayed. `count() > 0` might work better.

front/tests/pages/op-simulation-results-page-model.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  // Toggles the checkbox at a specific index if it is not already checked.
+  private async toggleCheckboxIfUnchecked(index: number): Promise<void> {
+    const checkbox = this.sscCheckboxes.nth(index);
+    const isChecked = await checkbox.isChecked();
+
+    if (!isChecked) {
+      await this.sscCheckmarks.nth(index).click();
+    }
+    await expect(checkbox).toBeChecked();
```
emersion: This can be replaced with `setChecked()`.

front/tests/pages/op-simulation-results-page-model.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  // Ensures all checkboxes in the settings panel are checked.
+  async checkAllSscCheckboxes(): Promise<void> {
+    await this.openSettingsPanel();
+
+    const checkboxCount = await this.sscCheckboxes.count();
+
+    for (let checkboxIndex = 0; checkboxIndex < checkboxCount; checkboxIndex += 1) {
+      await this.toggleCheckboxIfUnchecked(checkboxIndex);
+    }
```
emersion: `setChecked()` can probably be applied to `sscCheckboxes` directly.

front/playwright.config.ts
emersion: What are the differences? This should be exactly the same web browsers? Do you have both versions of the images so that we can compare?

front/tests/012-op-simulation-settings-tab.spec.ts
```diff
@@ -220,7 +258,20 @@ test.describe('Simulation Settings Tab Verification', () => {
     await operationalStudiesPage.returnSimulationResult();
     await opTimetablePage.getTrainArrivalTime('12:03');
     await opTimetablePage.clickOnScenarioCollapseButton();
+    // TODO: Remove the waitForTimeout and firefox from performOnSpecificOSAndBrowser when issue #9909 is fixed
+    await page.waitForTimeout(stabilityTimeout);
```
emersion: I believe this is a bad idea. This is a recipe for flaky tests.

The bug should be fixed before this test is merged.

front/tests/utils/index.ts
```diff
@@ -138,3 +138,39 @@ export const waitForInfraStateToBeCached = async (infraId: number): Promise<void
...
+/**
+ * Perform an action only on a specified OS and optionally skip a specific browser.
+ * @param action - The action to perform.
+ * @param options - Configuration options:
+ *                  - `currentBrowser`: The name of the current browser.
+ *                  - `os`: The target OS to run the action on (default: 'linux').
+ *                  - `browser`: The browser to skip (default: 'webkit').
+ *                  - `skipMessage`: Message to log if the action is skipped.
+ */
+export async function performOnSpecificOSAndBrowser(
```
emersion, line 151: This function is quite confusing. `browser` is a blacklist, while `os` is an whitelist.

front/tests/012-op-simulation-settings-tab.spec.ts-snapshots/CodeCompo-SSC-OFF-chromium-linux.png
emersion, line 1: Nit: this name is in French. Should be "CompositionCode" for English.

front/tests/012-op-simulation-settings-tab.spec.ts-snapshots/AllSettings-SSC-chromium-linux.png
emersion, line 1: Nit: the layout of these filenames is a bit mixed up. Would be better if it would start with the name of the component being tested, then any options, then the browser/os.

front/README.md
```diff
@@ -57,14 +57,19 @@ You may pass the same options and arguments to this script as you would to `yarn
...
 
-You may also want to explore the documentation of the test framework
-[Playwright](https://playwright.dev/). For example, try launching each test independently using
-`yarn playwright test --ui`, debug a test with `yarn playwright test --debug`, or launch a specified
-test of a specified test file with a specified browser once with for example
+If visual comparisons tests fail due to UI changes, new snapshots are required as the baseline.
+You can automatically update snapshots by running tests with the `--update-snapshots` flag:
+`yarn playwright test --update-snapshots`
+
+You may also want to explore [Playwright documentation](https://playwright.dev/docs/intro) for more insights.For example:
```
emersion: Nit: missing space after dot.

front/README.md
```diff
@@ -57,14 +57,19 @@ You may pass the same options and arguments to this script as you would to `yarn
 > - if on Linux, you can also launch all the containers on the host network: you can replace the
 >   `docker compose <something>` above with `osrd/scripts/host-compose.sh <something>`
 
-If the tests fail, you'll find a `front/test-results` folder that will contain videos of the fail
-test executions. They might be of help to understand what's going on. Note that the CI also exports
-these videos as artifacts.
+If the tests fail, a `front/test-results` folder will be created, containing videos and traces of the failed test executions.
+These files can help you understand what went wrong.
+Additionally, the CI system exports these videos and traces as artifacts.
+You can view the trace files using the [Playwright trace viewer](https://trace.playwright.dev/).
```
emersion: Nit: line wrapping is a bit off. In general we don't wrap the README by sentence, we wrap at 80 characters.

emersion (changes requested):
The language also affects the screenshots. The tests should be skipped if the current language doesn't match the screenshots.

front/tests/pages/op-simulation-results-page-model.ts
Yohh: this is for speed space chart, I don't like it either, as for "ellipsis button" that should be "settings button" or something like that probably.

`speedSpaceChartSettingsButton` ?

front/tests/012-op-simulation-settings-tab.spec.ts-snapshots/CodeCompo-SSC-OFF-chromium-linux.png
Yohh: `speedLimitTag` to be convenient

front/playwright.config.ts
Maymanaf: It's true that we have the same web browsers versions but we can have different OS enviremeent also the CI lack hardware acceleration which make graphic rendering a little bit differente 

Actual image from the CI: 
all-settings-GEV-actual
Expected image:
all-settings-GEV-expected
The difference between both images ( 18 pixels differences displayed in red ):
all-settings-GEV-diff

anisometropie: can’t you create the expected images from a similar environment ? so that the CI actually tests what we should get taking the environment into account

Maymanaf: We can generate the picture from the CI but we will have the same issue when we run the tests locally

Yohh (approved):
lgtm, thanks

front/README.md
```diff
@@ -257,11 +277,15 @@ You may also use `npm run lint-fix` to format/lint.
 When `package.json` changes (new packages or updated versions), developers may have issues running
 the app with Docker. New packages or versions might not be recognized by Docker.
 
+When `package.json` changes (new packages or updated versions), developers may have issues running
+the app with Docker. New packages or versions might not be recognized by Docker.
```
emersion: This paragraph is duplicated above.

front/tests/pages/op-simulation-results-page-model.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  // Ensures all checkboxes in the settings panel are checked.
+  async selectAllSpeedSpaceChartCheckboxes(): Promise<void> {
+    await this.openSettingsPanel();
+
+    const checkboxCount = await this.speedSpaceChartSettingsCheckboxes.count();
+
+    for (let checkboxIndex = 0; checkboxIndex < checkboxCount; checkboxIndex += 1) {
+      await this.speedSpaceChartCheckboxItems.nth(checkboxIndex).setChecked(true);
+    }
```
emersion: Continuing discussion from https://github.com/OpenRailAssociation/osrd/pull/9829#discussion_r1868067168

So, can this be replaced with `this.speedSpaceChartSettingsCheckboxes.setChecked(true)`? Or does this not work somehow?

front/tests/012-op-simulation-settings-tab.spec.ts
emersion: Thanks for fixing the root cause!

front/tests/pages/common-page-model.ts
```diff
@@ -38,6 +41,10 @@ class CommonPage {
       await this.viteOverlay.evaluate((node) => node.setAttribute('style', 'display:none;'));
     }
   }
+
+  async closeToastNotification() {
+    if (await this.toastSNCF.count()) await this.closeToastButton.first().click();
```
emersion: Do we really need the `first()` here? Shouldn't we close all notifications if there are multiple?

front/README.md
```diff
@@ -210,9 +228,11 @@ All common code (and shared components) supposed to be in `common/`.
 ### Javascript / Javascript-React
 
 - ESLint is used as linter and prettier as formatter. Both are configured as devDependencies to
-  enforce default eslint configuration eventually overidden by
+  enforce default eslint configuration eventually overridden by
   [airbnb rules](https://airbnb.io/javascript/) translation. A few rules (see eslintrc) has been
-  disabled and will be re-enabled in the near future):
+  disabled and will be re-enabled in the near future): [airbnb rules](https://airbnb.io/javascript/)
+  translation. A few rules (see eslintrc) has been disabled and will be re-enabled in the near
+  future):
```
emersion: Seems like this paragraph is corrupted?

front/README.md
```diff
@@ -38,35 +38,53 @@ Launches end to end tests.
...
-macOS and Ubuntu/Debian), you can instead start the tests inside a docker container using
-`osrd/scripts/run-front-playwright-container.sh`. You may pass the same options and arguments to
-this script as you would to `npm run e2e-tests` or `npx playwright test`.
+If you are using a Linux distribution not supported by Playwright (Playwright only supports Windows,
+macOS and Ubuntu/Debian), you can start the tests inside a Docker container using:
+`osrd/scripts/run-front-playwright-container.sh`. This script accepts the same options and arguments
+as `npm run e2e-tests` or `npx playwright test`. If you are using a Linux distribution not supported
+by Playwright (Playwright only supports Windows, macOS and Ubuntu/Debian), you can start the tests
+inside a Docker container using: `osrd/scripts/run-front-playwright-container.sh`. This script
+accepts the same options and arguments as `npm run e2e-tests` or `npx playwright test`.
```
emersion: This is duplicated as well

emersion (commented):
Looks a lot better, thanks for fixing the comments! Here are a few more.

front/tests/utils/index.ts
Maymanaf: Both the OS and the browser are now whitelisted

front/tests/pages/op-simulation-results-page-model.ts
Maymanaf: `setChecked()` method checks or unchecks an element can't be done for all the checkboxes

front/tests/pages/common-page-model.ts
Maymanaf: I updated the function to close all notifications if there are multiple ones

front/tests/pages/common-page-model.ts
```diff
@@ -38,6 +41,15 @@ class CommonPage {
       await this.viteOverlay.evaluate((node) => node.setAttribute('style', 'display:none;'));
     }
   }
+
+  async closeToastNotification(): Promise<void> {
+    const toastCount = await this.toastSNCF.count();
+
+    if (toastCount > 0) {
```
emersion: We don't need to check the length here. We can just call `.all()` always (which will return an empty array if there are no matching elements).

front/tests/pages/op-simulation-results-page-model.ts
emersion: Right. A cleaner way to do this would be `this.speedSpaceChartCheckboxItems.all((checkbox) => checkbox.setChecked(true))`.

front/tests/012-op-simulation-settings-tab.spec.ts
```diff
@@ -220,7 +257,21 @@ test.describe('Simulation Settings Tab Verification', () => {
...
```
emersion: Hm, now that the root cause has been fixed, this can be removed right?

Maymanaf: Oops, I forgot to remove them after the `autosquash `

front/tests/pages/common-page-model.ts
Maymanaf: I think it's okay to have this check to avoid unnecessary operations when the list is empty

emersion: There are no unnecessary operations performed when the list is empty. On the contrary, the current code performs two queries against the browser instead of a single one.

Maymanaf: Done

front/tests/012-op-simulation-settings-tab.spec.ts
Maymanaf: Removed

emersion (approved):
There is still one remaining comment for a small cleanup: https://github.com/OpenRailAssociation/osrd/pull/9829#discussion_r1884176835

If you're not willing to do it, you can resolve it.

Thanks!

front/tests/pages/op-simulation-results-page-model.ts
Maymanaf: Done

PR #9840, opened 2024-11-25
tests: add stdcm test generator for imported timetables

### LINKED ISSUE #8899: tests: improve fuzzer by including new features in the generated tests
There's a few things lacking with the fuzzer, new features that haven't been added to the generated scenarios:

- [ ] stdcm: we can request time for start or end points
- [ ] pathfinding locations can be defined as operational points for more realistic paths
- [ ] we can add scheduled points (though it's difficult to generate realistic times)

A little less clear:

We could tweak the script to be usable with an imported timetable. We wouldn't necessarily need to include the timetable data in the generated json.

eckter:
This is an utilitary script used to help reproduce bugs that only seem to appear on imported timetables, by "bruteforcing" realistic requests.

Fix https://github.com/OpenRailAssociation/osrd/issues/8899, in a way. It doesn't generate regression tests right out of the box.

### DIFF
--- a/tests/fuzzer/fuzzer_stdcm_single_timetable.py
+++ b/tests/fuzzer/fuzzer_stdcm_single_timetable.py
@@ -1,0 +1,202 @@
+import datetime
+import json
+import random
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Dict, Iterable, List, Optional, Union
+
+import requests
+
+from fuzzer.fuzzer import (
+    _get_random_rolling_stock,
+    _random_set_element,
+    _to_ms,
+    get_infra,
+)
+from tests.scenario import Scenario
+
+_TIMEOUT = 300
+
+_EDITOAST_URL = "http://127.0.0.1:8090/"
+_INFRA_NAME = "France"
+_TIMETABLE_ID = 0
+
+"""
+Generates random stdcm requests on a single infra + fixed timetable.
+Much lighter than `fuzzer.py`, but can't generate regression tests.
+Inputs that can't directly be set in the (non-debug) STDCM UI are fixed.
+
+This isn't automatically run anywhere, it's only used by hand for more convenient
+testing. May not always be up-to-date.
+
+Note: on an imported infra and timetables, tests take a while.
+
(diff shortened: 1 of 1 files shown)


shenriotpro:
Would it be hard to merge with the fuzzer?

tests/fuzzer/fuzzer_stdcm_single_timetable.py
```diff
@@ -0,0 +1,199 @@
...
+        date = self.start + datetime.timedelta(seconds=(random.randint(0, delta)))
+        return date.isoformat()
+
+
+def run(
+    editoast_url: str,
+    scenario: Scenario,
+    n_test: int = 1000,
+    log_folder: Optional[Path] = None,
+    seed: Optional[int] = None,
```
shenriotpro, line 61: nit: should we have a default seed for reproducibility? or make it mandatory

eckter: Reproducibility isn't a goal here, I can't really think of a use case where we'd need it. Other than debugging the script itself, but the optional parameter should be enough.

shenriotpro: I don't understand the workflow, how do you retrieve the request that failed? Is it from the exception message?

eckter:
> Would it be hard to merge with the fuzzer?

The use is a little different as it doesn't generate regression tests and requires a prefilled timetable.

I could cram all the functions in the same file and parametrize a single command, but I don't think it would help with maintainability.

Ultimately we should take some times to refactor the whole thing entirely, it has clearly become a mess. But I don't think it has a high priority just yet.

tests/fuzzer/fuzzer_stdcm_single_timetable.py
eckter: Yes, when we receive a 500 response, the content (detailed error) is put in an exception. It's then caught and logged in a json file, alongside the payload used to reproduce it.

shenriotpro: I see, thanks.

shenriotpro (approved):
Thanks

Erashin (approved):

PR #9843, opened 2024-11-25
front: fix duplicated op in times stops tables

### LINKED ISSUE #9727: Wrong duplicate waypoints bugs the time input and output
### What happened?

Some waypoints are identical and still appear twice in the Add Waypoints modal, while searching for an itinerary. 
When adding a value in the TimeStopsInput for one of the duplicates, the value gets added to both. 
It results in the processing of the second value as happening at Day+1. 

https://github.com/user-attachments/assets/489eb3c1-d31b-43f8-953f-36d63f7634b6

We should not have duplicates in the waypoints, or process them correctly in the timeStopsInput. 

Create a scenario
Search an itinerary with trigram VCE BRI
Try adding a waypoint with the modal, and see duplicates for Veynes
Go to Time and Stops tab and try modifying the values in the duplicate steps Veynes.

### Solutions for the fix, after workshop (16/02/24) : 

We noticed this was happening because the duplicated waypoint is located just between the end of the last track and the beginning of the new one.
We decided it was best to filter out one of the waypoint to avoid the issue mentioned in this ticket.
A solution to not lose informations about the fact that the waypoint is located between two tracks, is to display the two tracks in the track input of the input table. (Example : Track A - Waypoint - Track B = Track A-B in the input table).

We will need to know which tracks are connected, thanks to a new endpoint in the backend.
About the display in the graphs, we'll need to know which track offset to display, as it can be the one from track A or the one from track B (it won't change the overall calculations).

Local

Firefox

.

Akctarus:
closes #9727

### DIFF
--- a/front/src/modules/pathfinding/utils.ts
+++ b/front/src/modules/pathfinding/utils.ts
@@ -183,3 +183,29 @@ export const isStation = (chCode: string): boolean =>
   chCode === 'BV' || chCode === '00' || chCode === '';
 
 export const isPathStepInvalid = (step: PathStep | null): boolean => step?.isInvalid || false;
+
+/**
+ * Ensure that the array doesn't have any duplicated operational points by merging the duplicates.
+ * They usually can be differentiated by the track name.
+ * @param operationalPoints the list of operational points
+ * @param getDuplicatedOP a function that returns a key base on some op properties
+ * @param mergeOP a function checking that previous and next op differ by one property and merge this property into the previous op
+ * @returns the updated list with unique operational points
+ */
+export function getUniqueOperationalPoints<T>(
+  operationalPoints: T[],
+  extractKeyFromOp: (operationalPoint: T) => string,
+  mergeOP: (previousOperationalPoint: T, nextOperationalPoint: T) => void
+): T[] {
+  return Object.values(
+    operationalPoints.reduce<Record<string, T>>((uniqueOperationalPoints, operationalPoint) => {
+      const key = extractKeyFromOp(operationalPoint);
+      if (uniqueOperationalPoints[key]) {
+        mergeOP(uniqueOperationalPoints[key], operationalPoint);
+      } else {
+        uniqueOperationalPoints[key] = { ...operationalPoint };
+      }
+      return uniqueOperationalPoints;
+    }, {})
+  );
+}
--- a/front/src/modules/pathfinding/utils.ts
+++ b/front/src/modules/pathfinding/utils.ts
@@ -194,12 +194,12 @@ export const isPathStepInvalid = (step: PathStep | null): boolean => step?.isInv
  */
 export function getUniqueOperationalPoints<T>(
   operationalPoints: T[],
-  extractKeyFromOp: (operationalPoint: T) => string,
+  getDuplicatedOP: (operationalPoint: T) => string,
   mergeOP: (previousOperationalPoint: T, nextOperationalPoint: T) => void
 ): T[] {
   return Object.values(
     operationalPoints.reduce<Record<string, T>>((uniqueOperationalPoints, operationalPoint) => {
-      const key = extractKeyFromOp(operationalPoint);
+      const key = getDuplicatedOP(operationalPoint);
       if (uniqueOperationalPoints[key]) {
         mergeOP(uniqueOperationalPoints[key], operationalPoint);
       } else {
--- a/front/src/modules/timesStops/TimesStopsOutput.tsx
+++ b/front/src/modules/timesStops/TimesStopsOutput.tsx
@@ -45,7 +45,7 @@ const TimesStopsOutput = ({
 
     return getUniqueOperationalPoints(
       enrichedOperationalPoints,
-      (op) => `${op.name}-${op.ch}`,
+      (op) => `${op.opId}-${path?.path_item_positions}`,
       (previousOP, nextOP) => {
         if (previousOP.trackName !== nextOP.trackName) {
           previousOP.trackName = `${previousOP.trackName}/${nextOP.trackName}`;
--- a/front/src/modules/timesStops/TimesStopsOutput.tsx
+++ b/front/src/modules/timesStops/TimesStopsOutput.tsx
@@ -1,3 +1,5 @@
+import { useMemo } from 'react';
+
 import cx from 'classnames';
 
 import type {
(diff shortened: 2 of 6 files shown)


emersion:
Why do we get duplicate vias in the first place?

(I'm not a fan of using `name` or `ch` because these are not enough to recognize duplicate path steps: the name is human-readable and may be shared between different OPs, the ch may be unset or irrelevant.)

clarani (commented):
Nice job :)

I'm just thinking, do you think it would be possible to merge the op just after receiving the data from the backend ? This way, the op will be merged each time we need to display the waypoints.

You will have to use your function mergeDuplicatedOperationalPoints at 3 places (just after we call `postPathProperties`):
- in `preparePathPropertiesData` for the simulationResults
- in `usePathfiding` for the view ManageTrainSchedule
- in `useSetupItinerary` for the view ManageTrainSchedule, when we want to edit a train

front/src/modules/pathfinding/utils.ts
```diff
@@ -176,3 +176,21 @@ export const isStation = (chCode: string): boolean =>
...
 export const isPathStepInvalid = (step: PathStep | null): boolean => step?.isInvalid || false;
+
+export function mergeDuplicatedOperationalPoints<T>(
+  operationalPoints: T[],
+  getDuplicatedOP: (operationalPoint: T) => string,
+  mergeOP: (previousOperationalPoint: T, nextOperationalPoint: T) => void
+): T[] {
+  return Object.values(
+    operationalPoints.reduce<Record<string, T>>((mergedOperationalPoints, operationalPoint) => {
+      const key = getDuplicatedOP(operationalPoint);
```
SharglutDev, line 202: Not sure this is the right name for the function, it just buils a key for an op

front/src/modules/pathfinding/utils.ts
```diff
@@ -176,3 +176,21 @@ export const isStation = (chCode: string): boolean =>
...
 
 export const isPathStepInvalid = (step: PathStep | null): boolean => step?.isInvalid || false;
+
+export function mergeDuplicatedOperationalPoints<T>(
+  operationalPoints: T[],
+  getDuplicatedOP: (operationalPoint: T) => string,
+  mergeOP: (previousOperationalPoint: T, nextOperationalPoint: T) => void
+): T[] {
+  return Object.values(
+    operationalPoints.reduce<Record<string, T>>((mergedOperationalPoints, operationalPoint) => {
```
SharglutDev: Maybe it's more understandable

```suggestion
    operationalPoints.reduce<Record<string, T>>((uniqueOperationalPoints, operationalPoint) => {
```

front/src/modules/pathfinding/utils.ts
```diff
@@ -176,3 +176,21 @@ export const isStation = (chCode: string): boolean =>
   chCode === 'BV' || chCode === '00' || chCode === '';
 
 export const isPathStepInvalid = (step: PathStep | null): boolean => step?.isInvalid || false;
+
+export function mergeDuplicatedOperationalPoints<T>(
```
SharglutDev: This function could deserve some documentation. Here is a suggestion. Does that work for you @Akctarus @clarani ?
```suggestion
/**
 * Ensure that the array doesn't have any duplicated operational points by merging the duplicates.
 * They usually can be differentiated by the track name.
 * @param operationalPoints the list of operational points
 * @param getDuplicatedOP a function that returns a key base on some op properties
 * @param mergeOP a function checking that previous and next op differ by one property and merge this property into the previous op
 * @returns the updated list with unique operational points
 */
export function mergeDuplicatedOperationalPoints<T>(
```

front/src/modules/pathfinding/utils.ts
Akctarus: done

front/src/modules/pathfinding/utils.ts
Akctarus: done

front/src/modules/pathfinding/utils.ts
Akctarus: done

front/src/modules/timesStops/TimesStopsOutput.tsx
SharglutDev, line 1: I added a train PNO > LE and put a 30s stop to "Paris Nord La Souterraine 42/VRR" but it seems the ops are seen as the same in the output table : 

Capture d’écran 2024-12-19 à 14 13 05

front/src/modules/pathfinding/utils.ts
```diff
@@ -177,3 +177,29 @@ export const isStation = (chCode: string): boolean =>
...
+ * Ensure that the array doesn't have any duplicated operational points by merging the duplicates.
+ * They usually can be differentiated by the track name.
+ * @param operationalPoints the list of operational points
+ * @param getDuplicatedOP a function that returns a key base on some op properties
+ * @param mergeOP a function checking that previous and next op differ by one property and merge this property into the previous op
+ * @returns the updated list with unique operational points
+ */
+export function getUniqueOperationalPoints<T>(
+  operationalPoints: T[],
+  extractKeyFromOp: (operationalPoint: T) => string,
```
Uriel-Sautron:
```suggestion
  getDuplicatedOP: (operationalPoint: T) => string,
```
To comply with function documentation

SharglutDev: Actually I asked @Akctarus to change this param name but forgot to change it in the documentation proposal :)

front/src/modules/timesStops/TimesStopsOutput.tsx
```diff
@@ -36,9 +39,24 @@ const TimesStopsOutput = ({
     selectedTrainSchedule,
     path
   );
+
+  const mergedOperationalPoints = useMemo(() => {
+    if (!enrichedOperationalPoints) return [];
+
+    return getUniqueOperationalPoints(
+      enrichedOperationalPoints,
+      (op) => `${op.name}-${op.ch}`,
```
emersion: name + ch isn't a good key. The name is a human-readable string, so two different OPs can share the same name.

I believe OP ID + `positionOnPath` is what we want to use for deduplication.

front/src/modules/pathfinding/utils.ts
Akctarus: fixed

front/src/modules/timesStops/TimesStopsOutput.tsx
Akctarus: you're right, fixed

emersion (commented):
After discussion, we've concluded that we'd prefer to properly handle duplicate UIC/trigrams/secondary codes in a path rather than work around it.

Akctarus:
This PR can be closed until: https://github.com/osrd-project/osrd-confidential/issues/909 is finished

PR #9866, opened 2024-11-26
schemas, railjson_generator: comply to PEP 621

woshilapin:
This small change will help working with other tools than `poetry` which is [not yet](https://github.com/python-poetry/poetry/pull/9135) compatible.

With this simple change, you should be able to work with other tools like [uv](https://docs.astral.sh/uv/) with the following commands. Note that since we still depends on `poetry.lock` for versions, we can install with `poetry` the dependency, but through `uv` which means everything is done entirely in the virtual environment.

```sh
# setup a virtual environment with python from `.python-version` (either the system one, or install one)
$ uv venv
# install `poetry` dependencies from `poetry.lock` (note: `poetry`is never installed system-wide)
$ uv tool run poetry@1.8 install
# use `uv` like `poetry` to launch scripts
$ uv run ./some_script.py
```


shenriotpro (approved):
Thanks

flomonster (approved):
LGTM

PR #9867, opened 2024-11-26
front: fix margin and improve stopfor input in timesstops table

### LINKED ISSUE #9882: front :margin input bug in timestops table
### What happened?

If an invalid margin is entered in the timestopsinput table in the scenario page, the first correct margin that is entered to replace it is not taken into account (the margin is not saved and the corresponding op is not added to the path)

https://github.com/user-attachments/assets/e567d497-e025-4ac3-b444-aa478ac5ddf1

In the above video, the 2 margins above the one being inputted were also entered on top of an invalid value and thus not actually saved, hence they disappear once I actually enter a value that updates the path

Every valid value gets taken into account, even if inputed after an invalid value

-> Open the input table
-> Enter an invalid margin (e.g., "130" without a unit or "asd") on a row that is not in the path; the margin (correctly) turns red
-> Enter a valid margin instead; the margin returns to normal, but the row is not immediately added to the path, nor is the modification saved.

(See the video above)

Local

Firefox

N/A

Synar:
First commit : fix bug where setting an invalid margin, then a valid margin, would not register the valid margin in the path
Close #9882

Second commit : enable the user to input timestops with second unit, which were otherwise silently discarded upon saving the table

### DIFF
--- a/front/src/modules/timesStops/TimesStopsInput.tsx
+++ b/front/src/modules/timesStops/TimesStopsInput.tsx
@@ -103,26 +103,21 @@ const TimesStopsInput = ({ allWaypoints, startTime, pathSteps }: TimesStopsInput
       if (!updatedRows[operation.fromRowIndex].isMarginValid) {
         newRows[operation.fromRowIndex].isMarginValid = false;
         setRows(newRows);
-      } else if (
-        !rows[operation.fromRowIndex].isMarginValid &&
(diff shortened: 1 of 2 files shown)


clarani (commented):
Code LGTM :white_check_mark: 

I just need to test the 1st bug

clarani (approved):
LGTM :white_check_mark:

RomainValls (approved):
LGTM 👍

PR #9869, opened 2024-11-26
front: fix margin around scenario simulation results scrollbar

emersion:
Multiple issues here:

- The navbar height was incorrect. EDIT: fixed separately in https://github.com/OpenRailAssociation/osrd/pull/10089.
- Two elements had the same "scenario-sidemenu" CSS class.
- A CSS rule was overwriting Bootstrap's paddings for one column,
  but this messes up the layout because other columns weren't
  adjusted to take up more space. Use "no-gutters" instead, which
  is designed specifically for this use-case.
- Drop the min/max widths for the train list, because this creates
  a weird blank space for some screen sizes. Leave it up to the
  parent column to pick a reasonable width.

Before:

out-old

After:

out


theocrsb (commented):
Navbar size varies according to screen size. If we choose one or the other, a scrollbar appears. 
It might be better to set the navbar size to 40px (as seen with @thibautsailly).

image

emersion:
I've dropped the first commit which changes the navbar size. I'll fix it separately.

theocrsb (approved):
LGTM and tested

SharglutDev (approved):
Lgtm and tested

Maymanaf (approved):

PR #9870, opened 2024-11-26
front: display electrical profile set name in scenario

### LINKED ISSUE #8002: Name of the electrical profile set chosen for the scenario isn't displayed
### What happened?

As you can see on the screenshot, when you pick an electrical profile set or a scenario, only the ID is displayed and not the name. 

This happens only in V2 not V1

image

I would expect to see the name next to the infrastructure

1. Create a V2 scenario with an electrical profile set
2. In the train creation page, see that no name is displayed

- **Browser**: Firefox v115
- **OS**: Windows 11

803e0d1

clarani:
closes #8002 

After:
image

### DIFF
--- a/front/src/applications/operationalStudies/components/Scenario/ScenarioDescription.tsx
+++ b/front/src/applications/operationalStudies/components/Scenario/ScenarioDescription.tsx
@@ -1,9 +1,13 @@
-import { useEffect, useRef, useState } from 'react';
+import { useEffect, useMemo, useRef, useState } from 'react';
 
 import { Blocked, ChevronLeft, Pencil, X } from '@osrd-project/ui-icons';
 import { useTranslation } from 'react-i18next';
 
-import type { InfraWithState, ScenarioResponse } from 'common/api/osrdEditoastApi';
+import {
+  osrdEditoastApi,
+  type InfraWithState,
+  type ScenarioResponse,
+} from 'common/api/osrdEditoastApi';
 import { useModal } from 'common/BootstrapSNCF/ModalSNCF';
 import AddAndEditScenarioModal from 'modules/scenario/components/AddOrEditScenarioModal';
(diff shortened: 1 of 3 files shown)


SharglutDev:
Regarding your screenshot, it doesn't quite match the mockup : https://www.sketch.com/s/2fb064cd-c483-46ac-bd30-72705fd63e75/p/9935859D-9F0B-489C-9C2D-1E55C396D8DC/canvas

We should still see : "Profils électriques : FRANCE GAIA"

front/src/applications/operationalStudies/components/Scenario/ScenarioDescription.tsx
```diff
@@ -115,7 +128,7 @@ const ScenarioDescription = ({
       </div>
       <div className="scenario-details-electrical-profile-set">
         {scenario.electrical_profile_set_id
-          ? scenario.electrical_profile_set_id
+          ? electricalProfileSet?.name || ''
```
Akctarus: Is it possible that an electrical profile has no name? In which case, will we ever get the || ' ' case ?

Akctarus (approved):
Tested, LGTM, left a question in comment

clarani:
I modified my PR, here is what it looks like now:
image

front/src/applications/operationalStudies/components/Scenario/ScenarioDescription.tsx
clarani: Done :white_check_mark: (I display the name + the id, so even if there is no name, an identifier is displayed :+1:

Math-R (approved):
Looks good ! thanks !

front/src/applications/operationalStudies/components/Scenario/ScenarioDescription.tsx
Akctarus: Ty ! LGTM

PR #9871, opened 2024-11-26
front: fix map marker offset

### LINKED ISSUE #9633: Visual offset between the pointers and the track on the map
### What happened?

The pins are above the track with a not-negligeable distance
Image

They should stick to the tracks.

1. Get a train in operationnal studies
2. Scroll to the map on the result pages

Recette (SNCF)

Google Chrome

e9f3b96

emersion:
Set the `offset` prop to define the image "hotspot" (ie, pixel coordinate where the sign is planted on).

Closes: https://github.com/OpenRailAssociation/osrd/issues/9633

### DIFF
--- a/front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
+++ b/front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
@@ -28,6 +28,8 @@ export default function RenderItinerary({
     'line-color': '#158DCF',
   };
 
+  const markerOffset: [number, number] = [0, 8];
+
   if (!pathItemsCoordinates || pathItemsCoordinates.length < 2) {
     return null;
   }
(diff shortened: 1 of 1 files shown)


front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
```diff
@@ -38,18 +38,18 @@ export default function RenderItinerary({
 
   return (
     <Source type="geojson" data={geojsonPath}>
-      <Marker longitude={originLongitude} latitude={originLatitude} anchor="bottom">
+      <Marker longitude={originLongitude} latitude={originLatitude} anchor="bottom" offset={[0, 8]}>
```
emersion: Note, I've duplicated the offset 3 times because there's no real reason why these all need to match up (ie, they are the same "by chance"). If you prefer to store the offset in a `const` or something, that's fine by me as well.

Caracol3 (approved):
lgtm (tested)

front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
kmer2016: Good work! Since offset [0, 8]  is used to adjust visual alignment with the path, it might seem logical to extract it in a const like `const markerAlignmentOffset = [0, 8];`. However, if these offsets could diverge in the future (e.g., specific adjustments for start vs. intermediate points), keeping them separate as they are might actually make sense.

emersion: So, what is your preference?

kmer2016: store the offset in a const

emersion: Done. Had to explicitly make the const a `[number, number]` other TypeScript treats it as a `number[]`.

front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
SharglutDev, line 1: Should we pass these consts outside of the function ? (to avoid redefine them for when we zoom for example)

SharglutDev (approved):
Lgtm and tested. Feel free to solve the comment whether you fix it or not :)

front/src/modules/simulationResult/components/SimulationResultsMap/RenderItinerary.tsx
emersion: It shouldn't make any difference, and some other consts were there already.

Issue #9724, opened 2024-11-15
Log LMR requests


hamz2a:
### Description and goal

The goal of this ticket is to make it easier to debug LMR requests made by clients. It will allow a developer to "replay" an LMR request locally to analyze the cause of a failure.what is it for?

### Acceptance criteria

- [x] Create a PostgreSQL table with columns: `trace_id`, `input`, and `output`.
- [x] Store the input and output data of LMR requests in this table.
- [x] Add an environment variable to enable or disable logging (consider using Rust log as an alternative).
- [x] Create an endpoint accessible only to superusers to retrieve payloads. Allow fetching a payload by `trace_id` or a list of payloads.
- [x] Ensure data logging takes less than 0.1 seconds.

PR #9873, opened 2024-11-27
editoast: log stdcm requests

hamz2a:
Part of https://github.com/OpenRailAssociation/osrd/issues/9724

- Create a PostgreSQL table with columns: `trace_id`, `request`, `response`, `created` and `user_id`.
- Store the data of LMR requests in this table.


editoast/migrations/2024-11-21-145205_create_stdcm_logs/up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE stdcm_logs (
+    id int8 PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+    trace_id text UNIQUE NOT NULL,
```
leovalais:
```suggestion
    trace_id VARCHAR(X) UNIQUE NOT NULL,
```

where X is the maximum length of your trace_id

editoast/src/models/stdcm_log.rs
```diff
@@ -0,0 +1,23 @@
...
+use crate::core::stdcm::Request;
+use crate::core::stdcm::Response;
+
+editoast_common::schemas! {
+    StdcmLog,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, Model, ToSchema)]
+#[model(table = editoast_models::tables::stdcm_logs)]
+#[model(gen(ops = cru, list))]
```
```suggestion
#[model(gen(ops = c))]
```

They're only created now. We can always add more code gen when necessary.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -251,7 +258,34 @@ async fn stdcm(
...
+                            .trace_id();
+                        let _ = StdcmLog::changeset()
+                            .trace_id(trace_id.to_string())
+                            .request(stdcm_request)
+                            .response(stdcm_response_for_spawn)
+                            .create(conn)
+                            .await;
+                    }
+                    Err(_) => tracing::warn!("Database connection failed during log operation."),
+                }
```
leovalais: Wdyt about moving this part to a dedicated function?

editoast/src/views/timetable/stdcm.rs
```diff
@@ -251,7 +258,34 @@ async fn stdcm(
...
+        let _ = tokio::spawn(
+            async move {
+                match &mut db_pool.get().await {
+                    Ok(conn) => {
+                        let trace_id = tracing::Span::current()
+                            .context()
+                            .span()
+                            .span_context()
+                            .trace_id();
+                        let _ = StdcmLog::changeset()
```
```suggestion
                        StdcmLog::changeset()
```

this error shouldn't be ignored. We should log something if this fails.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -251,7 +258,34 @@ async fn stdcm(
...
+                            .trace_id(trace_id.to_string())
+                            .request(stdcm_request)
+                            .response(stdcm_response_for_spawn)
+                            .create(conn)
+                            .await;
+                    }
+                    Err(_) => tracing::warn!("Database connection failed during log operation."),
```
```suggestion
                    Err(_) => tracing::error!("Database connection failed during log operation."),
```

editoast/src/views/timetable/stdcm.rs
```diff
@@ -251,7 +258,34 @@ async fn stdcm(
 
     let stdcm_response = stdcm_request.fetch(core_client.as_ref()).await?;
 
-    // 6. Handle STDCM Core Response
+    // 6. Check if the current tracing level is debug or lower, and if so, log STDCM request and response
+    if tracing::Level::DEBUG <= tracing::level_filters::LevelFilter::current() {
```
leovalais: Is there any advantage of using Yoda conditions in rust since boolean assignment is syntactically invalid?

leovalais (commented):
Cursory review, I'll test later if this isn't merged beforehand.

I know the list of fields has been discussed in the refinement, but I can think of at least two other columns that could prove useful IMO:

* the date time of the request, allowing us to sort requests by recentness
* the user that performed the request

With these we'd be easily able to provide a feature like "replay *my* most recent request". Wdyt? (ping @ everyone)

editoast/src/views/timetable/stdcm.rs
hamz2a: I tried to put it into a function, but I couldn't get the context to retrieve the `trace_id`, so in the new function, I always have a `trace_id` of 0.

editoast/src/views/timetable/stdcm.rs
leovalais: Oh and it seems the condition is reversed if I'm not mistaken

editoast/src/views/timetable/stdcm.rs
leovalais: Oh you're right. But you still can make a new function and provide it the full changeset in parameter. Making it a function would be useful so you can manage errors with `?` in it, and just `tracing::error!` in the closure. Wdyt?

editoast/src/views/timetable/stdcm.rs
hamz2a: Done

editoast/migrations/2024-11-21-145205_create_stdcm_logs/up.sql
hamz2a: Done.

editoast/src/models/stdcm_log.rs
hamz2a: Done.

editoast/src/views/timetable/stdcm.rs
woshilapin: We actually discussed about the error handling in this specific scenario. Logging the requests and response is on a best-effort strategy (like logging) and I don't believe we want to impact in any way the behavior of the application. That's the reason why not bubbling up the error and not even considering it an `tracing::error!()` (I believe these should be used only when non recoverable errors that might affect the application, should be used). Still, we might want to know about these errors if they're too frequent, hence `tracing::warn!()`. What do you think?

editoast/src/views/timetable/stdcm.rs
woshilapin: See https://github.com/OpenRailAssociation/osrd/pull/9873#discussion_r1862289425 🤷

editoast/src/views/timetable/stdcm.rs
```diff
@@ -251,7 +258,34 @@ async fn stdcm(
...
+        let stdcm_response_for_spawn = stdcm_response.clone();
+        let _ = tokio::spawn(
+            async move {
```
woshilapin: Can you put a comment about why we don't want to bubble up errors in this function?

woshilapin (commented):
Were you able to test if the `trace_id` in the entrypoint of `editoast` is the same than the one you log? I'm wondering if it just generates a new one or not 🤷

editoast/src/views/timetable/stdcm.rs
leovalais: Well, I am not really convinced by the frontier between `error` and `warn`. We're talking about logs, which are meant for developers. To me, an error in this context would be resulting from a bug, while a warning would be meant to draw attention to a sketchy, or unstable, but ultimately legit situation. Take the warning we emit when we're receiving a request with a `page_size` above our limit (for some endpoints, rs for ex). We accept the request because it's legit, but still log something as a reminder that's a vulnerability. I'd hardly put an `INSERT INTO stdcm_logs` failure at the same level.

Besides, I'd argue that failing to acquire a connection, or failing to create the `StdcmLog` in the spawned task actually **are** unrecoverable errors as in “we have not implemented a way to recover from them“.

I'm not dying on that hill for sure, so feel free to resolve, it's just that `error!` felt more natural to me :)

editoast/src/views/timetable/stdcm.rs

editoast/src/views/timetable/stdcm.rs
leovalais: Follow up: my bad, the condition isn't reversed. Indeed `TRACE > DEBUG > INFO > ...`, which is a bit unintuitive imo, especially that the log constant values are in the reverse order 😅 More info: https://docs.rs/tracing-core/0.1.33/src/tracing_core/metadata.rs.html#871

editoast/src/views/timetable/stdcm.rs
```diff
@@ -283,6 +322,13 @@ async fn stdcm(
     }
 }
 
+async fn log_stdcm(conn: &mut DbConnection, stdcm_log_changest: Changeset<StdcmLog>) {
```
```suggestion
async fn log_stdcm(conn: &mut DbConnection, stdcm_log_changeset: Changeset<StdcmLog>) {
```

editoast/src/views/timetable/stdcm.rs
woshilapin, line 1: Shouldn't we also need a new endpoint to query this information from the DB? Or is it left for another PR (which would be nice to avoid a big PR).

hamz2a: Yes, I will do this in another PR.

editoast/src/models/stdcm_log.rs
```diff
@@ -0,0 +1,27 @@
...
+#[model(gen(ops = c))]
+pub struct StdcmLog {
+    pub id: i64,
+    pub trace_id: String,
+    #[model(json)]
+    pub request: Request,
+    #[model(json)]
+    pub response: Response,
+    created: DateTime<Utc>,
+    user_id: i64,
```
```suggestion
    pub created: DateTime<Utc>,
    pub user_id: i64,
```

for consistency (or remove the `pub` from the other fields if you prefer)

editoast/src/views/timetable/stdcm.rs
```diff
@@ -46,10 +51,14 @@ use crate::views::train_schedule::train_simulation_batch;
 use crate::views::AuthenticationExt;
 use crate::views::AuthorizationError;
 use crate::AppState;
+use crate::Model;
 use crate::Retrieve;
 use crate::RetrieveBatch;
 use crate::ValkeyClient;
 
+use super::Changeset;
+use super::Create;
```
```suggestion
use crate::models::prelude::*;
```

editoast/src/views/timetable/stdcm.rs
```diff
@@ -283,6 +324,13 @@ async fn stdcm(
     }
 }
 
+async fn log_stdcm(conn: &mut DbConnection, stdcm_log_changeset: Changeset<StdcmLog>) {
```
leovalais: I think there was a misunderstanding on what the function was for: I was thinking of something along these lines:

```suggestion
async fn log_stdcm(db_pool: Arc<DbConnectionPoolV2>, stdcm_log_changeset: Changeset<StdcmLog>) -> Result<()> {
```

Alternatively and even better, we should reuse the existing connection to the pool. Otherwise we might end up with the deadlocking issue we had some time ago while stressing out editoast. In this case, admittedly, a function might not be needed since there'll be only one source of error (`create`).

Ping @woshilapin for input on this.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -251,7 +260,39 @@ async fn stdcm(
...
+                        let trace_id = tracing::Span::current()
+                            .context()
+                            .span()
+                            .span_context()
+                            .trace_id();
+                        let stdcm_log_changeset = StdcmLog::changeset()
+                            .trace_id(trace_id.to_string())
+                            .request(stdcm_request)
+                            .response(stdcm_response_for_spawn)
+                            .user_id(user_id);
```
leovalais: these could be moved outside of the closure to better emphasize that we only try to create a log entry.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -251,7 +260,39 @@ async fn stdcm(
...
-    // 6. Handle STDCM Core Response
+    // 6. Check if the current tracing level is debug or lower, and if so, log STDCM request and response
+    if tracing::level_filters::LevelFilter::current() >= tracing::Level::DEBUG {
+        let user_id = auth.authorizer()?.user_id();
+        let stdcm_response_for_spawn = stdcm_response.clone();
+        let _ = tokio::spawn(
+            // We intentionally don't bubble up errors here because the spawned task
+            // runs asynchronously and is not expected to directly affect the caller.
+            // If errors were propagated, it could introduce unnecessary complexity
+            // and require additional error handling logic that is not needed in this context.
```
leovalais: Not sure I understand this part. We just don't await the creation of the log entry since we want the endpoint to return as soon as possible, and because failing to persist a log entry is not a very important error here.

Also referring the request issuer as the caller is a bit confusing as the caller of the `stdcm` function is actually `tower`.

leovalais (commented):
Thanks for the changes! A few more comments below:

editoast/src/models/stdcm_log.rs
hamz2a: Done.

editoast/src/views/timetable/stdcm.rs

editoast/src/views/timetable/stdcm.rs

editoast/src/views/timetable/stdcm.rs
hamz2a: As discussed, I updated the comment. Thanks again.

editoast/src/views/timetable/stdcm.rs
woshilapin: I usually refer to [this kind of definition](https://www.webdevsplanet.com/post/php-notice-warning-and-error-messages) about the difference between "warning" or "error". Note the

> when an error occurs execution of the script is halted.

which to me translates to not recoverable error, the application cannot continue to run properly.

That said, when searching a bit more, I also found [this](https://stackoverflow.com/a/2031209) which has "warning", "error" and "fatal" which might be closer to our case ("warn", "error" and "panic" ?). And then I understand how my definition doesn't necessary hold and... I kind of realize that since we can differentiate "non-recoverable but application can continue to run" and "non-recoverable and the application has to stop", I do agree with you 😄

editoast/src/views/timetable/stdcm.rs
woshilapin: Seems like I agree with `error` in the end! (see https://github.com/OpenRailAssociation/osrd/pull/9873#discussion_r1867444173)

editoast/src/views/timetable/stdcm.rs
woshilapin: Not sure that I understand what I'm expected for. If the idea is to put more in the function, less in the closure, yes, it does make sense.

editoast/src/views/timetable/stdcm.rs
leovalais: I really like this decision chart from your stack overflow post:

editoast/src/views/timetable/stdcm.rs
```diff
@@ -251,7 +255,38 @@ async fn stdcm(
 
     let stdcm_response = stdcm_request.fetch(core_client.as_ref()).await?;
 
-    // 6. Handle STDCM Core Response
+    // 6. Check if the current tracing level is debug or lower, and if so, log STDCM request and response
```
```suggestion
    // 6. Check if the current tracing level is debug or greater, and if so, log STDCM request and response
```

editoast/src/views/timetable/stdcm.rs
```diff
@@ -251,7 +255,38 @@ async fn stdcm(
...
+    if tracing::level_filters::LevelFilter::current() >= tracing::Level::DEBUG {
+        let user_id = auth.authorizer()?.user_id();
+        let stdcm_response_for_spawn = stdcm_response.clone();
```
woshilapin: You can inline this (no need for a new variable)... unless everything is moved inside the closure.

editoast/src/views/timetable/stdcm.rs
leovalais: It's more about using the new db connection vs opening a new one.

* If we open a new connection, we need the function to simplify the closure as there's two sources of error (`get` and `create`). But we may end up with the same deadlocking issues we encountered a while ago.
* If we clone the existing connection, we don't need the function since there's only one source of error anymore: `create`, so we can leave it in the closure.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -251,7 +255,38 @@ async fn stdcm(
 
     let stdcm_response = stdcm_request.fetch(core_client.as_ref()).await?;
 
-    // 6. Handle STDCM Core Response
+    // 6. Check if the current tracing level is debug or lower, and if so, log STDCM request and response
+    if tracing::level_filters::LevelFilter::current() >= tracing::Level::DEBUG {
+        let user_id = auth.authorizer()?.user_id();
```
woshilapin: Do we want to bubble up the error if `authorizer()` fails?

editoast/src/views/timetable/stdcm.rs
```diff
@@ -251,7 +255,38 @@ async fn stdcm(
 
     let stdcm_response = stdcm_request.fetch(core_client.as_ref()).await?;
 
-    // 6. Handle STDCM Core Response
+    // 6. Check if the current tracing level is debug or lower, and if so, log STDCM request and response
+    if tracing::level_filters::LevelFilter::current() >= tracing::Level::DEBUG {
```
woshilapin, line 264: What about putting everything that is inside the `if` inside the closure. I would expect minimal impact on the code, something like this

```rust
if LevelFilter::current >= Level::DEBUG {
  tokio::spawn(async move { /* ... */ });
}
```

hamz2a: We have to do the following `clone` outside of `spawn`, otherwise, we get a borrowing error.
```rust
let stdcm_response_for_spawn = stdcm_response.clone();
```

editoast/src/views/timetable/stdcm.rs
hamz2a: `user_id` is not null in the DB, and we need to know which user made the request. So, I think it's important to throw an error if we can't retrieve `user_id`.

editoast/src/views/timetable/stdcm.rs
hamz2a: As discussed, I deleted the function and fixed the borrowing issue with the connection.

editoast/src/views/timetable/stdcm.rs
leovalais: That's a good point actually. I'm not sure every stdcm request comes from an authenticated user. Aren't some run by scripts short-circuiting the gateway? If so, we may need to make the `user_id` nullable.

hamz2a: Is it possible to short-circuit in prod?
Are there any restrictions that prevent sending requests in production for editoast?
 @leovalais @woshilapin

leovalais: The gateway being the only exposed access point prevents that. But we may have unauthenticated, but legit, requests to editoast from behind it, on purpose.

In any case, I think the user_id should be nullable so that we can still log requests in development when authentication is disabled.

leovalais (changes requested):
Just tested. Works great when authenticated, but the log entry isn't created when authorization is disabled. Probably a `unique nulls` missing somewhere.

To keep the repo history clean, I suggest merging @woshilapin PR into yours instead of cherry-picking, and he should probably be Co-Author as well.

hamz2a:
> Just tested. Works great when authenticated, but the log entry isn't created when authorization is disabled. Probably a `unique nulls` missing somewhere.

If authentication is disabled, the superuser profile is granted to everyone:
```rust
if disable_authorization {
    return Ok(Authentication::Authenticated(Authorizer::new_superuser(
        PgAuthDriver::<BuiltinRole>::new(db_pool),
    )));
}
```
In the code, we assign `user_id = -1`:
```rust
pub fn new_superuser(storage_driver: S) -> Self {
    Self {
        user: UserInfo {
            identity: "superuser".to_string(),
            name: "Super User".to_string(),
        },
        user_id: -1,
        user_roles: HashSet::from([S::BuiltinRole::superuser()]),
        storage: storage_driver,
    }
}
```
This leads to an issue when trying to store the log, because it results in a foreign key error.
@leovalais

woshilapin (approved):
Thanks for integrating #9947 into your PR. I'll close #9947.

editoast/src/views/timetable/stdcm.rs
woshilapin: This has been resolved with #9947.

editoast/src/models/stdcm_log.rs
```diff
@@ -0,0 +1,55 @@
...
+            .trace_id();
+        let stdcm_log_changeset = StdcmLog::changeset()
+            .trace_id(trace_id.to_string())
+            .request(request)
+            .response(response.clone())
+            .user_id(user_id);
+        let _ = stdcm_log_changeset
+            .create(&mut conn)
+            .await
+            .map_err(|e| tracing::error!("Failed during log operation: {e}"));
```
```suggestion
        if let Err(e) = stdcm_log_changeset
            .create(&mut conn)
            .await {
      		tracing::error!("Failed during log operation: {e}"));
      	}
```

It's weird using `map` with a side effect and discarding the mapped value. You can use `inspect_err` and `ok` as well.

leovalais (approved):
LGTM. Thanks @hamz2a @woshilapin for the changes!

Issue #9023, opened 2024-09-26
stdcm: `Failed to compute a simulation that wouldn't cause conflicts` error


axrolld:
### What happened?

Tried to run a STDCM simulation in [this scenario](https://rec-osrd.reseau.sncf.fr/operational-studies/projects/19/studies/337/scenarios/644) with:
`{"comfort":"STANDARD","margin":"4.5min/100km","rolling_stock_id":55,"speed_limit_tags":"MA90","steps":[{"duration":0,"location":{"uic":87747006,"secondary_code":"BV"},"timing_data":{"arrival_time":"2024-09-26T12:45:00+02:00","arrival_time_tolerance_before":1800000,"arrival_time_tolerance_after":1800000}},{"duration":1800000,"location":{"uic":87741009,"secondary_code":"BV"}},{"duration":0,"location":{"uic":87746008,"secondary_code":"P2"}}],"time_gap_after":35000,"time_gap_before":35000}`

Lead to this error:
`"Failed to compute a simulation that wouldn't cause conflicts: mismatch between exploration and postprocessing (please open a bug report)"`

### What did you expect to happen?

I'm expecting to find a solution because i manage to find one in operational studies.

### How can we reproduce it (as minimally and precisely as possible)?

1. Reproduce the inputs from the description

### On which environments the bug occurs?

Recette (SNCF)

### On which browser the bug occurs?

Google Chrome

### OSRD version (top right corner `Account` button > `Informations`)

a5e9ca2

axrolld:
[train_schedules (7).json](https://github.com/user-attachments/files/17149697/train_schedules.7.json)

maelysLeratRosso:
Reproduced on recette (also dev) with the following inputs :  

Request : 
`/api/timetable/999/stdcm?infra=94`
Payload : 
```
[...]

PR #9874, opened 2024-11-27
core: stdcm: fix post-processing when arrival time is set

### LINKED ISSUE #9872: core: stdcm: array out of bounds in post-processing
### What happened?

```json
"error": {
        "status": 500,
        "type": "core:unknown_error",
        "context": {
            "message": "Index -1 out of bounds for length 1",
            "exception_type": "class java.lang.IndexOutOfBoundsException",
            "stack_trace": [
                "Preconditions.java:64",
                "Preconditions.java:70",
                "Preconditions.java:266",
                "Objects.java:361",
                "ArrayList.java:427",
                "STDCMPostProcessing.kt:184",
                "STDCMPostProcessing.kt:64",
                "STDCMPathfinding.kt:128",
                "STDCMPathfinding.kt:73",
                "STDCMEndpointV2.kt:99",
                "WorkerCommand.kt:249",
                "WorkerCommand.kt:303",
                "ThreadPoolExecutor.java:1136",
                "ThreadPoolExecutor.java:635",
                "Thread.java:840"
            ],
            "url": "/v2/stdcm"
        },
        "message": "An unknown exception was thrown"
    },
```

_No response_

I have a payload that reproduces it on a private timetable

Recette (SNCF)

Firefox

990c8fa9226a6783bf870bfd7716622fd686ce67

eckter:
We'd add too much delay to the departure time, resulting in inconsistencies. Includes a regression unit test. 

Fix https://github.com/OpenRailAssociation/osrd/issues/9872

Helps with https://github.com/OpenRailAssociation/osrd/issues/9023 as well, but it still happens, just much less often

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
@@ -169,6 +169,7 @@ class STDCMPostProcessing(private val graph: STDCMGraph) {
                 node,
                 nodes.subList(firstPlannedNodeIndex + 1, nodes.size),
                 mutableStopData,
+                timeData,
             )
         var actualStopAddedTime = min(maxAddedTime, timeDiff)
 
@@ -231,7 +232,9 @@ class STDCMPostProcessing(private val graph: STDCMGraph) {
         node: STDCMNode,
         nextNodes: List<STDCMNode>,
         mutableStopData: MutableList<StopTimeData>,
+        lastTimeData: TimeData,
     ): Double {
+        if (lastStopIndexBeforeNode == 0) return lastTimeData.maxFirstDepartureDelaying
         var maxTimeDiff = Double.POSITIVE_INFINITY
         var nextStopIndex = lastStopIndexBeforeNode
         if (node.stopDuration != null) nextStopIndex++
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
```diff
@@ -233,6 +233,7 @@ class STDCMPostProcessing(private val graph: STDCMGraph) {
         mutableStopData: MutableList<StopTimeData>,
     ): Double {
         var maxTimeDiff = Double.POSITIVE_INFINITY
+        if (lastStopIndexBeforeNode == 0) maxTimeDiff = node.timeData.maxFirstDepartureDelaying
```
Erashin: Hmm. Isn't it always true then? Shouldn't we just init `maxTimeDiff` to `node.timeData.maxFirstDepartureDelaying`?

eckter: It's true when the end time has been specified and there's no intermediate stop.

The alternative would be a ternary, but it's not a lot more readable. I don't mind either way.

Erashin: Nah let's keep it that way then.

Erashin (approved):

Erashin (approved):

PR #9876, opened 2024-11-27
editoast: consist compute gamma

### LINKED ISSUE #9806: stdcm: deceleration rate doesn't change when changing towed rolling stock
### What happened?

I simulated a train between PLF 00 and BGB P1 with a 22200G and different towed rolling stocks. The deceleration doesn't seem to change, though it's supposed to.

Here's the graph for a towed rolling stock RMI2N3M:
Image

And with MA, that's supposed to have a deceleration rate 3 times lower:
Image

The consist should have a deceleration rate that's the minimum between the one from the traction engine and the towed rolling stock.

1. try simulating a path with the towed rolling stocks i cited above
2. notice no difference in the deceleration curves

Recette (SNCF)

Google Chrome

6be0604

Wadjetz:
fix #9806
Add consist compute gamma

### DIFF
--- a/editoast/src/core/simulation.rs
+++ b/editoast/src/core/simulation.rs
@@ -212,6 +212,13 @@ impl PhysicsConsistParameters {
             self.traction_engine.rolling_resistance.clone()
         }
     }
+
+    pub fn compute_const_gamma(&self) -> f64 {
+        self.towed_rolling_stock
+            .as_ref()
+            .map(|towed| f64::min(towed.const_gamma, self.traction_engine.const_gamma))
+            .unwrap_or_else(|| self.traction_engine.const_gamma)
+    }
 }
 
 impl From<PhysicsConsistParameters> for PhysicsConsist {
@@ -223,6 +230,7 @@ impl From<PhysicsConsistParameters> for PhysicsConsist {
         let inertia_coefficient = params.compute_inertia_coefficient();
         let mass = params.compute_mass();
         let rolling_resistance = params.compute_rolling_resistance();
+        let const_gamma = params.compute_const_gamma();
 
         let traction_engine = params.traction_engine;
 
@@ -235,7 +243,7 @@ impl From<PhysicsConsistParameters> for PhysicsConsist {
             startup_time: (traction_engine.startup_time * 1000.0).round() as u64,
             startup_acceleration,
             comfort_acceleration,
-            const_gamma: traction_engine.const_gamma,
+            const_gamma,
             inertia_coefficient,
             rolling_resistance,
             power_restrictions: traction_engine.power_restrictions.into_iter().collect(),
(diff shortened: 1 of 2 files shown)


editoast/src/core/simulation.rs
```diff
@@ -203,6 +203,19 @@ impl PhysicsConsistParameters {
...
+
+    pub fn compute_gamma(&self) -> Gamma {
+        self.towed_rolling_stock
+            .as_ref()
+            .map(|towed| {
+                if towed.gamma.value < self.traction_engine.gamma.value {
+                    towed.gamma.clone()
+                } else {
+                    self.traction_engine.gamma.clone()
+                }
```
woshilapin: Once https://github.com/OpenRailAssociation/osrd/pull/9887 will be merged, you might be able to express that as `min`.

```suggestion
                std::cmp::min(towed.const_gamma, self.traction_engine.const_gamma)
```

axrolld: Should we wait for this to merge this PR ?

editoast/src/models/fixtures.rs
```diff
@@ -263,7 +263,7 @@ pub fn create_simple_rolling_stock() -> RollingStock {
         energy_sources: vec![],
         gamma: Gamma {
             gamma_type: "CONST".to_string(),
-            value: 1.0,
```
axrolld: what is this value ?

editoast/src/core/simulation.rs
```diff
@@ -621,4 +635,28 @@ mod tests {
...
+            physics_consist.compute_gamma(),
+            physics_consist.traction_engine.gamma,
+        );
+
+        physics_consist.towed_rolling_stock = None;
+        assert_eq!(
+            physics_consist.compute_gamma(),
+            physics_consist.traction_engine.gamma,
+        );
+    }
```
axrolld: I don't know any rust so i can't really review this 🙈

hamz2a (approved):
LGTM, thanks!

editoast/src/core/simulation.rs
```diff
@@ -623,4 +631,31 @@ mod tests {
...
+
+        assert_eq!(
+            physics_consist.compute_const_gamma(),
+            physics_consist.traction_engine.const_gamma,
+        );
+
+        physics_consist.towed_rolling_stock = None;
+        assert_eq!(
+            physics_consist.compute_const_gamma(),
+            physics_consist.traction_engine.const_gamma,
```
axrolld: Are you gonna catch the fact that removing the towed rolling stock isn't taken into account anymore if at the previous step the traction engin already has the lowest gamma ? I feel like the order you change and assert things matter here and it would be better to :
- start with a situation where the towed rolling stock gamma is the lowest
- remove the towed rs
- assert that the gamma of the consist has raised.

editoast/src/core/simulation.rs
```diff
@@ -623,4 +631,31 @@ mod tests {
             physics_consist.traction_engine.rolling_resistance,
         );
     }
+
+    #[test]
+    fn physics_consist_compute_gamma() {
+        let mut physics_consist = create_physics_consist();
+        physics_consist.traction_engine.const_gamma = 5.0;
```
axrolld: I know it's a test value but 5.0 is extremely high for a gamma, could you set it a 1.0 or something like this ?

axrolld (commented):
Just a few comments to make sure i got this right in the test. Otherwise the implementation looks good to me !

axrolld (approved):

woshilapin (approved):

PR #9877, opened 2024-11-27
front: fix flaky test 012 and 006

Maymanaf:
This PR addresses the flakiness in tests 006 and 012.
The Playwright runner is very fast, which sometimes prevents the results from being updated in time. To resolve this, I added timeouts

### DIFF
--- a/front/tests/012-op-simulation-settings-tab.spec.ts
+++ b/front/tests/012-op-simulation-settings-tab.spec.ts
@@ -30,9 +30,6 @@ test.describe('Simulation Settings Tab Verification', () => {
   // Set viewport to avoid scrolling issues and ensure elements are attached to the DOM
   test.use({ viewport: { width: 1920, height: 1080 } });
 
-  // TODO: Remove the skip when bug #9344 is fixed
-  test.skip(({ browserName }) => browserName === 'webkit', 'Skip this test on Safari');
-
   const expectedCellDataElectricalProfileON: StationData[] = readJsonFile(
     './tests/assets/operationStudies/simulationSettings/electricalProfiles/electricalProfileON.json'
   );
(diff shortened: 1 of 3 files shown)


Synar (commented):
Nice PR! There's a slight redundancy with https://github.com/OpenRailAssociation/osrd/pull/9614/commits/8248abec283ae79b5cbef44b455b8233b5904b82 which adds ```await page.waitForSelector(containerSelector, { state: 'visible' });``` to the top of the scroll container function

clarani (approved):
LGTM :white_check_mark:

Maymanaf:
> Nice PR! There's a slight redundancy with [8248abe](https://github.com/OpenRailAssociation/osrd/commit/8248abec283ae79b5cbef44b455b8233b5904b82) which adds `await page.waitForSelector(containerSelector, { state: 'visible' });` to the top of the scroll container function

I don't see any redundancy in this case, as I didn't modify the scroll function. Additionally, the use of waitForSelector shouldn't cause any issues since it doesn't rely on a fixed timeout.

Yohh (approved):
lgtm

Synar:
> > Nice PR! There's a slight redundancy with [8248abe](https://github.com/OpenRailAssociation/osrd/commit/8248abec283ae79b5cbef44b455b8233b5904b82) which adds `await page.waitForSelector(containerSelector, { state: 'visible' });` to the top of the scroll container function
> 
> I don't see any redundancy in this case, as I didn't modify the scroll function. Additionally, the use of waitForSelector shouldn't cause any issues since it doesn't rely on a fixed timeout.

Yes, but you are waiting for the table to be visible before calling the scroll function at multiple points, and my change makes the scroll function wait for the component it should scroll to be visible, so we solved the same bug twice. It shouldn't pose any issue though. If the table is visible once, it should be visible twice =p

PR #9878, opened 2024-11-27
front: remove rolling stock custom hook where unnecessary

theocrsb:
In some places, we use a custom hook making an api call when it's possible to use only the rolling stock id from the store


clarani (approved):
Nice job !! :white_check_mark:

Akctarus (approved):
tested, LGTM

PR #9879, opened 2024-11-27
front: make rs label shorter

### LINKED ISSUE #9358: stdcm: make rolling stock labels shorter

End users

In the list of rolling stocks here, the labels are very long, we can make them shorter by only displaying the reference of the rolling stock:

Image

From the payload the front gets, display the "name" and "series" if it exists, "reference" otherwise.
```
{
        "name": "JT42US",
        "detail": "CLASS 66",
        "reference": "JT42",
        "series": "CLASS 66"
}
```

This feature matters because it will make the rolling stock list clearer.

kmer2016:
close #9358

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
@@ -73,20 +73,8 @@ const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
   }, [towedRollingStock]);
 
   const getLabel = (rs: LightRollingStockWithLiveries) => {
-    let res = '';
-    const { metadata, name } = rs;
-
-    if (metadata) {
-      const series = metadata.series ?? metadata.reference;
-      const subseries =
-        metadata.series && metadata.series !== metadata.subseries
-          ? metadata.subseries
-          : metadata.detail;
-      if (series) res += series;
-      if (subseries) res += series ? ` (${subseries})` : subseries;
-    }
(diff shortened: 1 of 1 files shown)


axrolld (approved):
The logic's right and i saw what it looked like, it's great ! Thanks 😁

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
```diff
@@ -72,22 +72,8 @@ const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
...
-          ? metadata.subseries
-          : metadata.detail;
-      if (series) res += series;
-      if (subseries) res += series ? ` (${subseries})` : subseries;
-    }
-    if (name) res += ` -- ${name}`;
-    return res;
-  };
+  const getLabel = (rs: LightRollingStockWithLiveries) =>
+    `${rs.name} - ${rs.metadata?.series || rs.metadata?.reference || ''}`;
```
clarani: If we don't have any metadata, we will have the '-', is it on purpose ?

Akctarus (approved):
Tested, LGTM

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
kmer2016: Nope, I pushed the fix

clarani (approved):
LGTM :white_check_mark: Can you squash your commits ?

PR #9893, opened 2024-11-28
front: use the consist length in stdcm gev

### LINKED ISSUE #9459: stdcm : length of the convoy not taken into account in the speed-space chart
### What happened?

1. I create a convoy without specifying the length of the train : the chart looks like this (the speed limit lines are extended by the value of the length of the train, which is the default one : good
Image
2. I change the length of the convoy to be huge (10000m) : I expect the speed limit lines to be much larger, since the train is much longer, but they are notImage

I expected the line to depend on the length of the train

as above

Recette (SNCF)

Google Chrome

7451939

clarani:
closes #9459

Must wait for the fix in the backend.


clarani:
As I'm leaving the project, I'm going to close this PR. I'm not sure this bug is still relevant

PR #9894, opened 2024-11-28
Make stdcm path buttons clickable across the entire element

### LINKED ISSUE #9504: Make `Add an intermediate OP` button clickable everywhere

End users

Currently this button is only clickable on the text label. It would be nice if it was clickable everywhere, as some users tried several time to click outside the label to add an OP.
Image

Smoother UX

.

achrafmohye:
Closes #9504

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmDefaultCard.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmDefaultCard.tsx
@@ -16,12 +16,12 @@ const StdcmDefaultCard = ({
   disabled = false,
   className = 'add-via',
 }: StdcmCardProps) => (
-  <StdcmCard hasTip={hasTip} disabled={disabled} className={className}>
-    <button type="button" onClick={onClick}>
+  <button type="button" disabled={disabled} onClick={onClick}>
+    <StdcmCard hasTip={hasTip} disabled={disabled} className={className}>
       <span className="stdcm-default-card-icon">{Icon}</span>
       <span className="stdcm-default-card-button">{text}</span>
-    </button>
-  </StdcmCard>
+    </StdcmCard>
+  </button>
 );
 
 export default StdcmDefaultCard;
--- a/front/src/styles/scss/applications/stdcm/_home.scss
+++ b/front/src/styles/scss/applications/stdcm/_home.scss
@@ -107,6 +107,14 @@
 
         .stdcm-vias-list {
           margin-block: 16px;
+
+          button {
+            text-align: left;
+          }
+
+          button:focus {
+            outline: none;
+          }
         }
 
         .stdcm-origin-icon,
--- a/front/src/styles/scss/applications/stdcm/_card.scss
+++ b/front/src/styles/scss/applications/stdcm/_card.scss
@@ -197,10 +197,6 @@
       line-height: 24px;
       margin-left: 12px;
     }
-
-    button:focus {
-      outline: none;
-    }
   }
 
   &.disabled,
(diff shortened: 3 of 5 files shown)


front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -107,6 +107,14 @@
...
         .stdcm-vias-list {
           margin-block: 16px;
+
+          button {
+            text-align: left;
+          }
+
+          button:focus {
+            outline: none;
+          }
```
theocrsb, line 117: Why do we need to disable the outline?

achrafmohye: It was already there just moved the css property

theocrsb: The problem is that we don't have the information when navigating with the tab key. 
Maybe we can add an outline.
The only problem I can see is that it makes us look a bit weird : 
image

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -107,6 +107,14 @@
 
         .stdcm-vias-list {
           margin-block: 16px;
+
+          button {
+            text-align: left;
+          }
```
SharglutDev, line 113: Could you put that in `_card.scss` ? We will need this for the linked path buttons as well.

SharglutDev (changes requested):

front/src/styles/scss/applications/stdcm/_home.scss
achrafmohye: it might be a good idea maybe we have to check it's usage and how to handle the outline of the chevron

front/src/styles/scss/applications/stdcm/_home.scss
achrafmohye: Added it here `_linkedPath.scss`

front/src/styles/scss/applications/stdcm/_home.scss
SharglutDev: I think it doesn't have to block this PR as it was the behavior before. We should ask @thibautsailly how he wants to handle the focus in the form.

front/src/applications/stdcm/components/StdcmForm/StdcmDefaultCard.tsx
```diff
@@ -16,12 +16,12 @@ const StdcmDefaultCard = ({
   disabled = false,
   className = 'add-via',
 }: StdcmCardProps) => (
-  <StdcmCard hasTip={hasTip} disabled={disabled} className={className}>
-    <button type="button" onClick={onClick}>
+  <button type="button" onClick={onClick}>
```
SharglutDev: Otherwise the cards are still clickable when a simulation is retained

```suggestion
  <button type="button" disabled={disabled} onClick={onClick}>
```

SharglutDev (changes requested):

front/src/applications/stdcm/components/StdcmForm/StdcmDefaultCard.tsx
achrafmohye: Good point, thanks for the hint 👍

achrafmohye: Done

SharglutDev (approved):
Lgtm and tested

front/src/styles/scss/applications/stdcm/_home.scss
SharglutDev: @achrafmohye if you can talk with him this afternoon about that and maybe create an enhancement

theocrsb (approved):
LGTM and tested :)

Maymanaf (approved):
LGTM for e2e tests

PR #9901, opened 2024-11-29
front: simplify usePathfinding

### LINKED ISSUE #9900: front: simplify usePathfinding

clarani:
closes #9900 

see commits

### DIFF
--- a/front/src/modules/pathfinding/hooks/usePathfinding.ts
+++ b/front/src/modules/pathfinding/hooks/usePathfinding.ts
@@ -241,7 +241,7 @@ export const usePathfinding = (
         return step;
       });
 
-      dispatch(updatePathSteps({ pathSteps: updatedPathSteps }));
+      dispatch(updatePathSteps(updatedPathSteps));
       pathfindingDispatch({ type: 'PATHFINDING_FINISHED' });
       pathfindingDispatch({ type: 'PATHFINDING_PARAM_CHANGED' });
     } else {
@@ -338,7 +338,7 @@ export const usePathfinding = (
                 );
               }
               dispatch(
-                updatePathSteps({ pathSteps: updatedPathSteps, resetPowerRestrictions: true })
+                updatePathSteps(updatedPathSteps)
               );
 
               const allWaypoints = upsertPathStepsInOPs(
--- a/front/src/modules/pathfinding/hooks/usePathfinding.ts
+++ b/front/src/modules/pathfinding/hooks/usePathfinding.ts
@@ -7,7 +7,9 @@ import { useSelector } from 'react-redux';
 import { useScenarioContext } from 'applications/operationalStudies/hooks/useScenarioContext';
 import type { ManageTrainSchedulePathProperties } from 'applications/operationalStudies/types';
 import type {
+  IncompatibleConstraints,
   PathfindingInputError,
+  PathfindingResultSuccess,
   PostInfraByInfraIdPathPropertiesApiArg,
   PostInfraByInfraIdPathfindingBlocksApiArg,
   RollingStockWithLiveries,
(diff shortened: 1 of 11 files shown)


front/src/reducers/osrdconf/osrdConfCommon/index.ts
```diff
@@ -176,25 +176,38 @@ export function buildCommonConfReducers<S extends OsrdConfState>(): CommonConfRe
...
     },
     clearVias(state: Draft<S>) {
       state.pathSteps = [state.pathSteps[0], state.pathSteps[state.pathSteps.length - 1]];
+      state.powerRestriction = [];
+    },
+    // Use this action in the suggested ops list, not the via list
+    removeVia(state: Draft<S>, action: PayloadAction<string>) {
+      // Index takes count of the origin in the array
+      state.pathSteps = state.pathSteps.filter((step) => !step || step.id !== action.payload);
+      state.powerRestriction = [];
```
SharglutDev: Instead of having to do a `find` and then the `filter` in this action, could we do the `filter` before calling the action and here just do a `state.pathSteps = action.payload;` ?

SharglutDev (commented):
Thank you for the PR. Everything seems to be still working !

Some new actions are a bit strange but I guess it's the first step of your `pathfinding` refacto ? :)

front/src/reducers/osrdconf/osrdConfCommon/index.ts
clarani: Yes, done :white_check_mark:

SharglutDev (approved):
Lgtm :)

front/src/reducers/osrdconf/osrdConfCommon/index.ts
```diff
@@ -176,25 +177,40 @@ export function buildCommonConfReducers<S extends OsrdConfState>(): CommonConfRe
...
     },
-    updatePathSteps(
-      state: Draft<S>,
-      action: PayloadAction<{ pathSteps: S['pathSteps']; resetPowerRestrictions?: boolean }>
-    ) {
-      state.pathSteps = action.payload.pathSteps;
-      if (action.payload.resetPowerRestrictions) {
-        state.powerRestriction = [];
-      }
+    // update path steps without changing the itinerary (only add vias on the existing pathfinding,
```
anisometropie, line 180: shouldn’t we use replaceItinerary if we add a vias (and change the number of elements) ?
I think the more important distinction is the number of elements.

some use cases only add a flag to the pathSteps or update the coordinates, but we keep exactly the same start, vias, end. in that case the itinerary remains exactly the same (same number of elements, all the elements are the same) 

if we add a vias on the existing itinerary, the train uses exactly the same path, but it might stop somewhere it didn’t before.

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -252,152 +252,163 @@ export const usePathfinding = (
...
+          pathfindingDispatch({ type: 'PATHFINDING_FINISHED' });
+          return;
+        }
+
+        const incompatibleConstraintsCheck =
+          pathfindingResult.failed_status === 'pathfinding_not_found' &&
+          pathfindingResult.error_type === 'incompatible_constraints';
+
+        if (incompatibleConstraintsCheck) {
+          await populateStoreWithSuccessfulPathfinding(
```
anisometropie: it’s a bit puzzling to have this function named `withSuccessfulPathfinding` in a case where the status is actually failed.

anisometropie (changes requested):
Thanks ! That’s a really great improvement on the redux part. that’s really the direction we should take for our usage of redux, calling actions `replaceItinerary` and `reverseItinerary` emphasis the business rules we’re trying to implement instead of just`updateSomeElement` we have been doing so far in all our reducers

front/src/reducers/osrdconf/osrdConfCommon/index.ts
clarani: Actually, I created these new actions to simplify my next PR. The idea is that when we call this action `replaceItinerary`, we need to relaunch the pathfinding after that.
Adding a via from the map or adding a via from the suggested modal is not the same action: adding a via from the suggested modal will not modifiy the itinerary since the via is already on it (= no need to relaunch the pathfinding), whereas adding a via from the map may modify the itinerary (= need to relaunch the pathfinding).

Instead of creating a new action (which will be removed in my next PR), I thought I could reuse `replaceItinerary`.

anisometropie: Ok nice, that’ll be helpful for my fix on double calls on the pathfinding endpoint

anisometropie: Ok nice, that’ll be helpful for my fix on double calls on the pathfinding endpoint

anisometropie: can you comment both action to specify when they are used, and that one doesn’t require to recall pathfinding endpoint, and the other does ?

front/src/modules/pathfinding/hooks/usePathfinding.ts
clarani: You're right, I renamed it `populateStoreWithPathfinding` :+1:

anisometropie (approved):
tested

PR #9905, opened 2024-11-29
front: fix linked path button clickable if retained simulation

### LINKED ISSUE #9807: stdcm: I can still interact with linked path buttons after downloading the simulation sheet
### What happened?

I selected one of my stdcm simulation. All the form is disabled but this is still clickable:
Image

And the linked path anterior also is clickable.

It should be disabled like the rest of the form.

1. Get a STDCM simulation
2. Save it and download the simumation sheet
3. click on the buttons that are not disabled

Recette (SNCF)

Google Chrome

6be0604

SharglutDev:
Remove the unnecessary div wrapper for the card

fix #9807


theocrsb (approved):
LGTM and tested :)

Caracol3 (approved):
lgtm

Math-R (approved):

Math-R:
Lgtm ✅

PR #9906, opened 2024-11-29
core: remove the middle green signals in the STD

Khoyo:
Currently, VL (green) blocks only have a meaning before the train, and the meaning is "I require this signal to be open to keep to my schedule".
After/during the train, the meaning would be "I don't constrain this signal" which is actually what blank means. So let's keep those cases blank.

Case appears when diverging from the projected path, stopping on closed signal, then converging (stop in the same station, but different track)
Before:
image
After this PR:
image

Relates to 1 part of https://github.com/osrd-project/osrd-confidential/issues/886

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/signal_projection/SignalProjection.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/signal_projection/SignalProjection.kt
@@ -251,34 +251,20 @@ private fun signalUpdates(
             }
         }
 
-        for (i in 0 until events.size - 1) {
+        for (i in 0 until events.size) {
             val event = events[i]
-            val nextEvent = events[i + 1]
+            val timeEnd = if (i + 1 < events.size) events[i + 1].time else simulationEndTime
+
+            if (listOf("VL", "300VL").contains(event.newAspect)) {
+                continue
+            }
+
             signalUpdates.add(
                 SignalUpdate(
                     physicalSignalName!!,
                     signalingSystemName,
                     event.time,
-                    nextEvent.time,
-                    positionStart,
-                    positionEnd,
-                    color(event.newAspect),
-                    blinking(event.newAspect),
-                    event.newAspect,
-                )
-            )
-        }
-
-        // The last event only generates an update if the signal doesn't return to VL
-        if (events.last().newAspect != "VL" && events.last().newAspect != "300VL") {
-            val event = events.last()
-            val timeStart = event.time
-            signalUpdates.add(
-                SignalUpdate(
-                    physicalSignalName!!,
-                    signalingSystemName,
-                    timeStart,
-                    simulationEndTime,
+                    timeEnd,
                     positionStart,
                     positionEnd,
                     color(event.newAspect),
--- a/core/src/test/kotlin/fr/sncf/osrd/signal_projection/SignalProjectionTest.kt
+++ b/core/src/test/kotlin/fr/sncf/osrd/signal_projection/SignalProjectionTest.kt
@@ -1,0 +1,171 @@
+package fr.sncf.osrd.signal_projection
+
+import fr.sncf.osrd.api.SignalCriticalPosition
+import fr.sncf.osrd.api.ZoneUpdate
+import fr.sncf.osrd.api.project_signals.SignalUpdate
+import fr.sncf.osrd.api.project_signals.TrainSimulation
+import fr.sncf.osrd.path.interfaces.JsonTrainPath
+import fr.sncf.osrd.path.interfaces.JsonTrainPath.ObjectRange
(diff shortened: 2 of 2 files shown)


bougue-pe (dismissed):
(completed the main description)

It doesn't work from my test.
Maybe be should also shorten the first free block? 🤔

Khoyo:
> (completed the main description)
> 
> It doesn't work from my test. Maybe be should also shorten the first free block? 🤔

Nah, we should do it in the right file.

image


So... I also removed the old signal projection logic and adapted the v1 signal projection endpoint

core/src/main/java/fr/sncf/osrd/api/SignalProjectionEndpoint.kt
```diff
@@ -0,0 +1,168 @@
...
+            val chunkPath = makeChunkPath(infra.rawInfra, request.trainPath!!)
+
+            val routePath =
+                infra.rawInfra.convertRoutePath(request.trainPath!!.routePath.map { it.route })
+
+            val sigSystemManager = infra.signalingSimulator.sigModuleManager
+            val bal = sigSystemManager.findSignalingSystemOrThrow("BAL")
+            val bapr = sigSystemManager.findSignalingSystemOrThrow("BAPR")
+            val tvm300 = sigSystemManager.findSignalingSystemOrThrow("TVM300")
+            val tvm430 = sigSystemManager.findSignalingSystemOrThrow("TVM430")
```
bougue-pe: Those are unused vars

core/src/main/java/fr/sncf/osrd/api/SignalProjectionEndpoint.kt
bougue-pe, line 1: Shouldn't we put it in a kt dir?
Or should we just not care and add a `deprecated` comment right away?

core/src/main/kotlin/fr/sncf/osrd/signal_projection/SignalProjection.kt
```diff
@@ -277,6 +277,11 @@ private fun signalUpdates(
         for (i in 0 until events.size - 1) {
             val event = events[i]
             val nextEvent = events[i + 1]
```
bougue-pe: As there is no special filter for the last event anymore, the code can probably be mutualized.
Something along this (there is probably something more idiomatic to be done):
```suggestion
            val endTime = if (i < events.size) events[i + 1].time else simulationEndTime
```

bougue-pe (commented):
I'm not sure I understand what you did on the SignalProjectionEndpoint (v1): why can't we completely remove it?
If we keep it like this, we probably need a clear `todo` to accelerate future cleanup work.

core/src/main/java/fr/sncf/osrd/api/SignalProjectionEndpoint.kt
Khoyo: The kotlin directory is not necessary anymore since at least Kotlin 1.0. I believe we should probably just merge the kotlin and java dirs

Khoyo:
> I'm not sure I understand what you did on the SignalProjectionEndpoint (v1): why can't we completely remove it?

We're keeping the TSv1 endpoints for a few more months (I believe until some time in January). But there is not reason to keep the old logic if we can avoid it.

core/src/main/java/fr/sncf/osrd/api/SignalProjectionEndpoint.kt
Erashin: Hmm. Personal opinion: I'd rather not merge the directories, since the end goal is probably just to have the whole thing in kotlin, it makes a clear distinction between what we have to keep and what we have to migrate.
For that reason as well, as long as we separate them,  I'd rather have the kotlin files in the kotlin directory (especially since it already exists).

Erashin (approved):
This looks like it's working as intended.
Give me your opinion on the kotlin/java module question.
I'd love a unit test on this case as well.

core/src/main/java/fr/sncf/osrd/api/SignalProjectionEndpoint.kt
Khoyo: > I'd rather not merge the directories, since the end goal is probably just to have the whole thing in kotlin, it makes a clear distinction between what we have to keep and what we have to migrate.

I really don't get that. If it's about the kotlin/java migration, just `find -name '*.java'`?

If it's about the obsolete v1 APIs, this is an obsolete v1 api file :p

Erashin (commented):
Is this actually still up to date? Been one year.

Khoyo:
Well, I've just found it again while spring cleaning my git branches, the bug still isn't fixed, and I ~~finally~~ wrote the unit test, so yeah, now it is.

Erashin (approved):

PR #9910, opened 2024-12-02
fron: lmr consist prefill mass length max speed

### LINKED ISSUE #9712: lmr: front: pre-fill the total weight, length, and max speed of the consist when selecting traction engine and towed rs
# Description

> [!NOTE]
> Missing (compo code; max speed) map for now

Pre-fill the total weight, length, and max speed of the consist when selecting traction engine and towed rolling stock.

# AC

- [ ] Selecting the traction engin pre-fills the total weight, length, and max speed of the consist.
- [ ] Selecting the towed rolling stock updates the values 
     - [ ] Length = length of traction engine + length of towed rolling stock
     - [ ] Weight = weight of traction engine + weight of towed rolling stock
     - [ ] Max speed = minimum between speed of traction engine (if not 0) and speed of towed rolling stock (if not 0).
- [ ] Changing the traction engine or the towed rolling stock also updates the values.
- [ ] The value of the max speed must be updated when changing the composition code (requires a map of (compo code; max speed))
- [ ] The values manually set by the user must never be overwritten by the logics above.

Wadjetz:
part of #9712

Prefill the consist values based on the rolling stock and towed rolling stock selection.
Do not update the value if the user has manually changed it.

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
@@ -52,6 +52,7 @@ const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
     onTotalLengthChange,
     maxSpeed,
     onMaxSpeedChange,
+    prefillConsist,
   } = useStdcmConsist();
 
   const { filters, searchRollingStock, searchRollingStockById, filteredRollingStockList } =
@@ -91,6 +92,7 @@ const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
   };
 
   const onSelectSuggestion = (option?: LightRollingStockWithLiveries) => {
+    prefillConsist(option, towedRollingStock);
     dispatch(updateRollingStockID(option?.id));
   };
 
@@ -144,6 +146,7 @@ const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
           suggestions={filteredTowedRollingStockList}
           getSuggestionLabel={(suggestion: TowedRollingStock) => suggestion.name}
           onSelectSuggestion={(towed) => {
+            prefillConsist(rollingStock, towed);
             dispatch(updateTowedRollingStockID(towed?.id));
           }}
         />
--- a/front/src/applications/stdcm/hooks/useStdcmConsist.ts
+++ b/front/src/applications/stdcm/hooks/useStdcmConsist.ts
@@ -1,12 +1,21 @@
+import { useState } from 'react';
+
 import { useSelector } from 'react-redux';
 
+import type { LightRollingStockWithLiveries, TowedRollingStock } from 'common/api/osrdEditoastApi';
 import { useOsrdConfActions, useOsrdConfSelectors } from 'common/osrdContext';
 import { type StdcmConfSliceActions } from 'reducers/osrdconf/stdcmConf';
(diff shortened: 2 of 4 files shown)


front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
```diff
@@ -91,6 +92,7 @@ const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
   };
 
   const onSelectSuggestion = (option?: LightRollingStockWithLiveries) => {
+    prefillConsist(option);
```
SharglutDev: You need to pass the select towed rolling stock here (if there is one), otherwise the inputs are not updated correctly

front/src/applications/stdcm/hooks/useStdcmConsist.ts
```diff
@@ -15,28 +24,55 @@ const useStdcmConsist = () => {
...
     dispatch(updateMaxSpeed(totalMaxSpeed === 0 ? undefined : totalMaxSpeed));
   };
 
+  const prefillConsist = (
+    rollingStock?: LightRollingStockWithLiveries,
+    towed?: TowedRollingStock
+  ) => {
+    if (!rollingStock && !towed) {
+      return;
+    }
```
SharglutDev: If I clear the tower input and then I clear the rolling stock input, we pass here whereas we should clear all (unaffected by the user) inputs

SharglutDev (commented):
Thank you for the PR ! Almost good, left some comments

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
```diff
@@ -91,6 +92,7 @@ const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
   };
 
   const onSelectSuggestion = (option?: LightRollingStockWithLiveries) => {
```
SharglutDev, line 94: There is still an issue here. If I clear the towed in put and then clear the rs input, this function receives the stale towed data even though the towed id is at undefined and the inputs are refilled with these datas.

To fix this, you can do in `useStdcmTowedRollingStock` the same thing we did in `useStoreDataForRollingStockSelector`, changing `data` to `currentData` in the endpoint call. This will ensure to provide an `undefined` towed if the last call returned `undefined` instead of providing the stale data.

front/src/applications/stdcm/hooks/useStdcmConsist.ts
```diff
@@ -15,28 +24,53 @@ const useStdcmConsist = () => {
...
+    if (!totalLengthChanged) {
+      const consistLength = Math.floor((rollingStock?.length ?? 0) + (towed?.length ?? 0));
+      dispatch(updateTotalLength(consistLength > 0 ? consistLength : undefined));
+    }
+
+    if (!maxSpeedChanged) {
+      dispatch(
+        updateMaxSpeed(rollingStock?.max_speed ? Math.floor(rollingStock.max_speed) : undefined)
+      );
+    }
```
SharglutDev, line 63: We should leave a comment here to handle this case when the back bug for max speed in towed will be fixed (is that you handling it or another dev ?)

Wadjetz: Yes, the max speed for towed rolling stock is made in another PR

SharglutDev (approved):
Lgtm and tested, good job !

PR #9911, opened 2024-12-02
core: amend ertms braking simulator interface

Erashin:
We actually only want to compute the simulator from the mrsp. What's more, we need to know when to brake + the position of the svl if required. This required the following changes to the interface.

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/ertms/etcs/ETCSBrakingSimulator.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/ertms/etcs/ETCSBrakingSimulator.kt
@@ -11,28 +11,30 @@ interface ETCSBrakingSimulator {
     val rollingStock: RollingStock
     val timeStep: Double
 
-    /**
-     * Compute the ETCS braking envelope from the MRSP: for each decreasing speed transition,
-     * compute the corresponding ETCS braking curve.
-     */
-    fun getETCSBrakingEnvelopeFromMrsp(mrsp: Envelope)
-
-    /**
-     * Compute the ETCS braking envelope from target: compute the corresponding ETCS braking curve
-     * decreasing from rolling stock max speed and ending at target offset/speed-wise.
-     */
-    fun getETCSBrakingEnvelopeFromTarget(target: Target)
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/fr/sncf/osrd/ertms/etcs/ETCSBrakingSimulator.kt
```diff
@@ -11,28 +11,30 @@ interface ETCSBrakingSimulator {
...
-     * Compute the ETCS braking envelope from target: compute the corresponding ETCS braking curve
-     * decreasing from rolling stock max speed and ending at target offset/speed-wise.
-     */
-    fun getETCSBrakingEnvelopeFromTarget(target: Target)
+    /** Compute the ETCS braking envelope from the MRSP, for each LOA and EOA. */
+    fun getETCSBrakingEnvelope(
+        mrsp: Envelope,
+        limitsOfAuthority: Collection<LimitOfAuthority>,
+        endsOfAuthority: Collection<EndOfAuthority>
+    )
```
eckter:
```suggestion
    /** Compute the ETCS braking envelope from the MRSP, for each LOA and EOA. */
    fun addETCSBrakingParts(
        mrsp: Envelope,
        limitsOfAuthority: Collection<LimitOfAuthority>,
        endsOfAuthority: Collection<EndOfAuthority>
    ): Envelope
```

core/src/main/kotlin/fr/sncf/osrd/ertms/etcs/ETCSBrakingSimulator.kt
```diff
@@ -11,28 +11,30 @@ interface ETCSBrakingSimulator {
...
 
-enum class TargetType {
-    SLOWDOWN,
-    STOP
+data class EndOfAuthority(
+    val offsetEOA: Offset<Path>,
+    val offsetSVL: Offset<Path>?,
+) {
+    init {
+        if (offsetSVL != null) assert(offsetSVL > offsetEOA)
```
eckter:
```suggestion
        if (offsetSVL != null) assert(offsetSVL >= offsetEOA)
```

I assume there are some cases where they overlap, such as stopping right at a buffer stop

eckter (approved):

PR #9914, opened 2024-12-02
front: fix nge import when there is only one step

### LINKED ISSUE #9610: Macro mode - getSourceNode error
### What happened?

I get this error in the macro node after importing a specific Viriato file (I don't post it here, please ask @maelysLeratRosso the file to work on the bug) : 

Image

or also

Image

Linked to this [bug](https://github.com/OpenRailAssociation/osrd/issues/9594) probably

I expected to see the NGE board

1. Import the file
2. Go to macro mode
3. Observe the bug

Recette (SNCF)

Google Chrome

6d9ed21

RomainValls:
closes #9610 

When importing a train, it can happen that the train only has 1 step, which was resulting in the origin and destination being the same step. 
We check the length of the steps, if it's less than 2, we skip.

### DIFF
--- a/front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
@@ -242,7 +242,9 @@ const importTimetable = async (
       body: { ids: train_ids },
     })
   );
-  const trainSchedules = await trainSchedulesPromise.unwrap();
+  const trainSchedules = (await trainSchedulesPromise.unwrap()).filter(
+    (trainSchedule) => trainSchedule.path.length >= 2
+  );
 
   const searchPayload = buildOpQuery(infraId, trainSchedules);
   const searchResults = searchPayload ? await executeSearch(searchPayload, dispatch) : [];


front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
```diff
@@ -244,12 +244,16 @@ const importTimetable = async (
   );
   const trainSchedules = await trainSchedulesPromise.unwrap();
 
+  const filteredTrainSchedules = trainSchedules.filter(
```
emersion: Nit: I think I'd prefer to make `trainSchedules` mutable (with `let`) and overwrite it instead of using a separate `const`, otherwise I'm worried new code would forget about this stuff and use `trainSchedules` instead of `filteredTrainSchedules`.

emersion (approved):
LGTM!

front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
RomainValls: You're right, makes more sense keeping trainSchedules ! modified it 👍

Caracol3 (approved):
lgtm !

PR #9916, opened 2024-12-02
front: change imported train schedules number of step color

### LINKED ISSUE #9594: Viriato import with 0 OP matched
### What happened?

From a Viriato file import, if a train has no OCP matched with an OP (= we do not match every OP in our infra), the import is should not work.

See all "-1":
Image

I also break the Macro mode, but it is rather a consequence that will be solved by the above bug

Linked to this [bug](https://github.com/OpenRailAssociation/osrd/issues/9610)

When no stop is matched for a train, simply do not import it and console log it

Download the Viriato file ("MSC-AIX...") from this [folder](https://sncf.sharepoint.com/:f:/r/sites/OSRD644GrpO365/Documents%20partages/Donn%C3%A9es/3-Fichiers%20Viriato?csf=1&web=1&e=tsLWf3)

line 17 of this file, we see the OCP that is not match in our infra, used by the `trainPart id="tp_3_1975735"`

- import in operational studies
- notice 

Recette (SNCF)

Firefox

dev 95c1305

RomainValls:
closes https://github.com/osrd-project/osrd-confidential/issues/884
Will close #9594

### DIFF
--- a/front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleTrainDetail.tsx
+++ b/front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleTrainDetail.tsx
@@ -78,7 +78,7 @@ export default function ImportTrainScheduleTrainDetail({
         </span>
         <span
           className={`import-train-schedule-traindetail-stepnb ${
-            trainData.steps.length <= 2 ? 'bg-secondary' : 'bg-primary'
+            trainData.steps.length <= 2 ? 'import-invalid-stepnb' : 'bg-primary'
           }`}
         >
           {trainData.steps.length - 2}
--- a/front/src/modules/trainschedule/styles/ImportTrainSchedule/_importTrainScheduleTrainsList.scss
+++ b/front/src/modules/trainschedule/styles/ImportTrainSchedule/_importTrainScheduleTrainsList.scss
@@ -86,6 +86,9 @@
           height: 32px;
           text-align: center;
         }
+        .import-invalid-step-nb {
+          background-color: var(--error60);
+        }
       }
       .import-train-schedule-traindetail-with-hover {
         cursor: pointer;


front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleTrainDetail.tsx
```diff
@@ -78,7 +78,7 @@ export default function ImportTrainScheduleTrainDetail({
         </span>
         <span
           className={`import-train-schedule-traindetail-stepnb ${
-            trainData.steps.length <= 2 ? 'bg-secondary' : 'bg-primary'
+            trainData.steps.length <= 2 ? 'bg-danger' : 'bg-primary'
```
theocrsb: Maybe we should use a variable in our color variables? like `error`?

clarani: Ah yes you're right, I thought danger and error60 were the same but they are not. Can you use error60 (you can see it in our `styles/_variables.scss`

RomainValls: that's better, thanks !

theocrsb (approved):
LGTM (not tested)

louisgreiner (approved):
LGTM, tested, thanks

image

front/src/modules/trainschedule/styles/ImportTrainSchedule/_importTrainScheduleTrainsList.scss
```diff
@@ -86,6 +86,9 @@
           height: 32px;
           text-align: center;
         }
+        .import-invalid-stepnb {
```
clarani: ;)
```suggestion
        .import-invalid-steps-nb {
```

clarani (approved):
Left a small comment, but LGTM except that ! :rocket:

PR #9922, opened 2024-12-03
Ensure stdcm form and simulations clear on page refresh

### LINKED ISSUE #9754: LMR Refreshing the page completely clears the form and any simulations below.

achrafmohye:
Closes #9754

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
@@ -58,6 +58,7 @@ const StdcmConfig = ({
     updateGridMarginBefore,
     updateStdcmStandardAllowance,
     updateStdcmPathStep,
+    resetStdcmConfig,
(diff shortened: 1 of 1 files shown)


emersion:
Curious, I would've expected that to already be the case, seeing how `stdcmConfSlice` is in the blacklist in `persistConfig`.

Akctarus (approved):
Tested, lgtm !

front/src/reducers/osrdconf/stdcmConf/index.ts
SharglutDev, line 1: When I refresh the page, the fields for maxSpeed, length and mass are not empty

achrafmohye: i think after the merge of @Wadjetz pr, i'll push a commit that resets the consist form values

achrafmohye: Done !

SharglutDev (approved):
Lgtm and tested (don't forget to fixup your commits)

PR #9924, opened 2024-12-03
front: add ManageTrainSchedule context and remove Pathfinding's reducer 

### LINKED ISSUE #6719: Pathfinding launched two times when adding a via

### LINKED ISSUE #9902: front: editing a train with power restrictions resets its power restrictions
### What happened?

When clicking on 'edit' for a train which has power restrictions, the power restrictions are automatically reset.

The power restrictions should not be reset when clicking on editing

1. Open a scenario
2. Add a train with power restrictions (you can use a TGV)
3. See the results
4. Click on edit the train
5. See the warning displayed

Recette (SNCF)

Firefox

.

### LINKED ISSUE #9915: front: add Context for ManageTrainSchedule

### LINKED ISSUE #10038: No warning when removing origin or destination in the itinerary tab
### What happened?

Image

When removing the origin or destination, it should add a warning on the left of the rocket but it doesn't

We should wait for this [PR](https://github.com/OpenRailAssociation/osrd/pull/9924) to be merge before fixing this

Have the warning

1. Go to a scenario
2. Add a train and launch a pathfinding
3. Remove the origin or destination

Local

Firefox

.

clarani:
closes #9915 
closes #9902 (because the power restrictions were reset by the call to usePathfinding in the RollingStockSelector)
will also close #6719
and fix #10038


clarani:
Need to wait for https://github.com/OpenRailAssociation/osrd/pull/9901 first

front/src/modules/trainschedule/components/ManageTrainSchedule/Map.tsx
```diff
@@ -385,7 +385,7 @@ const Map = ({
               layerOrder={LAYER_GROUPS_ORDER[LAYERS.LINE_SEARCH.GROUP]}
               infraID={infraID}
             />
-            <RenderPopup pathProperties={pathProperties} />
+            {!showStdcmAssets && <RenderPopup pathProperties={pathProperties} />}
```
emersion, line 393: Hm, why this change? Is this fixing a bug?

clarani: This popup is displayed only if I click on the map. This feature is only available in the EEx, and inside the component, I use the ManageTrainScheduleContext.
If this component renders in Stdcm, then the app crashed because there is no ManageTrainScheduleContextProvider.

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -228,193 +199,193 @@ export const usePathfinding = (
...
+    }
 
-    const startPathFinding = async () => {
-      setPathProperties(undefined);
-      if (pathfindingState.running) {
+    try {
+      const pathfindingResult = await postPathfindingBlocks(pathfindingInput).unwrap();
+
+      if (pathfindingResult.status === 'success') {
+        await populateStoreWithPathfinding(steps as PathStep[], pathfindingResult);
```
emersion: Why do we need a cast here? Could we replace it with:

```ts
steps.map((step) => step!)
```

which should be safer in case the type of `steps` is changed to something else?

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -228,193 +199,193 @@ export const usePathfinding = (
...
 
-      if (!pathfindingInput) {
-        dispatch(
-          setFailure({
-            name: t('pathfinding'),
-            message: t('pathfindingMissingParamsSimple'),
-          })
+      if (incompatibleConstraintsCheck) {
+        await populateStoreWithPathfinding(
+          steps as PathStep[],
```
emersion: Ditto

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -228,193 +199,193 @@ export const usePathfinding = (
...
-          ...(correspondingOp && {
-            name: correspondingOp.name,
-            uic: correspondingOp.uic,
-            ch: correspondingOp.ch,
-            kp: correspondingOp.kp,
-            coordinates: correspondingOp.coordinates,
-          }),
-        };
-      });
+  const launchPathfinding = async (newPathSteps?: (PathStep | null)[]) => {
```
emersion: Do we need `newPathSteps` to be optional here? Doesn't seem like any caller sets it to `undefined`.

emersion: Should we `useCallback` here, to avoid always returning a new function inside the context even if all dependencies haven't changed?

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -228,193 +199,193 @@ export const usePathfinding = (
...
-
-    if (infra && infra.state === 'CACHED' && pathfindingState.mustBeLaunched) {
-      startPathFinding();
     }
-  }, [pathfindingState.mustBeLaunched, infra]);
+  };
 
-  useEffect(() => setIsPathfindingInitialized(true), []);
+  useEffect(() => {
+    launchPathfinding();
```
emersion: Previously we'd re-launch the pathfinding when the infra state switches from loading to cached. Now we don't do this anymore. Do we want to retain this behavior?

front/src/applications/operationalStudies/hooks/useManageTrainScheduleContext.tsx
```diff
@@ -2,17 +2,26 @@ import { createContext, useContext, useMemo, type ReactNode, useState, useEffect
...
 import type { ManageTrainSchedulePathProperties } from '../types';
 
+
 type ManageTrainScheduleContextType = {
   pathProperties?: ManageTrainSchedulePathProperties;
   setPathProperties: (pathProperties?: ManageTrainSchedulePathProperties) => void;
   voltageRanges: RangedValue[];
+  launchPathfinding: (pathSteps?: (PathStep | null)[]) => void;
+  pathfindingState: PathfindingState;
+  infraInfos: { infra?: InfraWithState; reloadCount: number };
```
emersion: Pedantic nit: "information" is uncountable, so shouldn't get pluralized.

emersion (commented):
Excellent work! Thanks so much for splitting into separate commits. Here are a few comments :)

front/src/modules/trainschedule/components/ManageTrainSchedule/Map.tsx
emersion: Aha, makes sense!

front/src/modules/pathfinding/hooks/usePathfinding.ts
clarani: Yes you're right, we should keep this behavior

front/src/modules/pathfinding/components/Itinerary/ModalSuggestedVias.tsx
```diff
@@ -110,7 +116,10 @@ const ModalSuggestedVias = ({ suggestedVias }: ModalSuggestedViasProps) => {
           <button
             className="btn btn-danger btn-sm btn-block mb-1"
             type="button"
-            onClick={() => dispatch(clearVias())}
+            onClick={() => {
+              const newPathSteps = [pathSteps.at(0) || null, pathSteps.at(-1) || null];
```
SharglutDev: This modal can't exist if we miss origin or destination (l.114 in `Itinerary.tsx`) so we should be able to do this
```suggestion
              const newPathSteps = [pathSteps.at(0)!, pathSteps.at(-1)!;
```

front/src/modules/pathfinding/components/Itinerary/ModalSuggestedVias.tsx
```diff
@@ -32,7 +35,10 @@ const ModalSuggestedVias = ({ suggestedVias }: ModalSuggestedViasProps) => {
     [destination]
   );
 
-  const removeViaFromPath = (op: SuggestedOP) => dispatch(removeVia(op));
+  const removeViaFromPath = (op: SuggestedOP) => {
+    const newPathSteps = pathSteps.filter((step) => !step || matchPathStepAndOp(step, op));
```
SharglutDev: I don't see how this could have ever work. Here we filter the steps which matches the op we want to remove so we will have only the removed op in path steps at the end.

Also, same as the other comment, here we know we won't have any null `pathSteps` so we should be able to remove the `!step` (if you did the suggested changed at the top of the file)

```suggestion
    const newPathSteps = pathSteps.filter((step) => !matchPathStepAndOp(step, op));
```

front/src/modules/pathfinding/components/Itinerary/ModalSuggestedVias.tsx
```diff
@@ -9,21 +9,24 @@ import ModalBodySNCF from 'common/BootstrapSNCF/ModalSNCF/ModalBodySNCF';
...
-const ModalSuggestedVias = ({ suggestedVias }: ModalSuggestedViasProps) => {
-  const { upsertViaFromSuggestedOP, clearVias, removeVia } = useOsrdConfActions();
-  const { getVias, getDestination } = useOsrdConfSelectors();
+const ModalSuggestedVias = ({ suggestedVias, launchPathfinding }: ModalSuggestedViasProps) => {
+  const { upsertViaFromSuggestedOP } = useOsrdConfActions();
+  const { getVias, getDestination, getPathSteps } = useOsrdConfSelectors();
   const dispatch = useAppDispatch();
   const vias = useSelector(getVias());
   const destination = useSelector(getDestination);
+  const pathSteps = useSelector(getPathSteps);
```
SharglutDev: We could add this here so we don't have to deal with `null` `PathStep` everywhere inside.
```suggestion
  const pathSteps = useSelector(getPathSteps) as PathStep[]; // this modal can't be open if we miss origin or destination
```

front/src/modules/pathfinding/components/Itinerary/ModalSuggestedVias.tsx
```diff
@@ -32,7 +35,10 @@ const ModalSuggestedVias = ({ suggestedVias }: ModalSuggestedViasProps) => {
     [destination]
   );
 
-  const removeViaFromPath = (op: SuggestedOP) => dispatch(removeVia(op));
+  const removeViaFromPath = (op: SuggestedOP) => {
+    const newPathSteps = pathSteps.filter((step) => !step || matchPathStepAndOp(step, op));
+    launchPathfinding(newPathSteps);
```
SharglutDev, line 40: To avoid relaunching the pathfinding if we remove a via which was already in the suggested op list, maybe we could pass to this component `pathProperties.suggestedOperationalPoints` (the pure list of ops on the path) to check if it was in it or not.

> [!NOTE]
> When the PR to "add a op via by map click" will be merged, maybe we will need some day to difference this kind of op from the others

front/src/modules/pathfinding/components/Itinerary/DisplayItinerary/Vias.tsx
```diff
@@ -87,7 +91,10 @@ const Vias = ({ zoomToFeaturePoint, shouldManageStopDuration }: ViasProps) => {
                         data-testid="delete-via-button"
                         className="btn btn-sm btn-only-icon btn-white ml-auto"
                         type="button"
-                        onClick={() => dispatch(deleteVia(index))}
+                        onClick={() => {
+                          const newPathSteps = removeElementAtIndex(pathSteps, index + 1);
+                          launchPathfinding(newPathSteps);
```
SharglutDev, line 96: Same suggestion as the one in `ModalSuggestedVias`

front/src/modules/pathfinding/components/Pathfinding/Pathfinding.tsx
```diff
@@ -98,13 +93,13 @@ const Pathfinding = ({ pathProperties, setPathProperties }: PathfindingProps) =>
...
-                  ? t('InvalidTrainScheduleStep')
-                  : t('pathfindingError', { errorMessage: t(pathfindingState.error) })}
+                {pathfindingState.error
+                  ? t('pathfindingError', { errorMessage: t(pathfindingState.error) })
+                  : t('InvalidTrainScheduleStep')}
               </span>
             </div>
           )}
-          {pathfindingState.missingParam && (
+          {pathfindingState.isMissingParam && (
```
SharglutDev, line 102: Strangely, if I remove the origin or destination by clicking on the cross from `Origin` or `Destination` components, this doesn't update this part of the component (probably just an update from the store doesn't rerender). Maybe `isMissingParam` should keep the actual missing param so this could update when removing one of them ?

I think this was broken before but this could be a good moment to fix it.

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -243,116 +86,131 @@ export const usePathfinding = (
...
+      if (infra?.state !== 'CACHED') {
+        return;
+      }
+
+      setIsRunning();
+      const pathfindingInput = getPathfindingQuery({
+        infraId,
+        rollingStock,
+        pathSteps: steps.filter(
+          (step) => step !== null && step.coordinates !== null && !step.isInvalid
```
SharglutDev: Why do we need to filter the steps without coordinates for a pathfinding ?

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -243,116 +86,131 @@ export const usePathfinding = (
...
+  };
+
+  const launchPathfinding = useCallback(
+    async (steps: (PathStep | null)[]) => {
+      dispatch(replaceItinerary(steps));
+      setPathProperties(undefined);
 
-    const startPathFinding = async () => {
-      if (pathfindingState.running) {
+      if (steps.some((step) => step === null)) {
```
SharglutDev, line 181: If you replace this line with this, typescript will understand that steps can't have any `null` value and you can avoid all null checks in the rest of the function and in `getPathfindingQuery`

```suggestion
      if (!steps.every((step) => step !== null)) {
```

front/src/modules/pathfinding/components/Itinerary/Itinerary.tsx
```diff
@@ -132,7 +129,12 @@ const Itinerary = ({
               className="col ml-1 my-1 text-white btn bg-info btn-sm"
               type="button"
               onClick={() =>
-                openModal(<ModalSuggestedVias suggestedVias={pathProperties.allWaypoints} />)
+                openModal(
+                  <ModalSuggestedVias
+                    suggestedVias={pathProperties.allWaypoints}
+                    launchPathfinding={launchPathfinding}
```
SharglutDev, line 134: Can't this component call `launchPathfinding` itself ?

front/src/modules/pathfinding/components/Pathfinding/TypeAndPath.tsx
```diff
@@ -171,8 +170,8 @@ const TypeAndPath = ({ setDisplayTypeAndPath }: TypeAndPathProps) => {
           id: nextId(),
         }));
 
-      dispatch(replaceItinerary(pathSteps));
       setDisplayTypeAndPath(false);
+      launchPathfinding(pathSteps);
```
SharglutDev, line 174: Seems a little weird that the function where we call this has the same name ()eslint doesn't complain strangely).

SharglutDev (commented):
Great refacto, thanks for that !

front/src/modules/pathfinding/components/Itinerary/ModalSuggestedVias.tsx
clarani: Done

front/src/modules/pathfinding/components/Itinerary/ModalSuggestedVias.tsx
clarani: You're right, done :+1:

front/src/modules/pathfinding/components/Itinerary/ModalSuggestedVias.tsx
clarani: Done !

front/src/modules/pathfinding/hooks/usePathfinding.ts
clarani: We should only filter the invalid steps, you're right :+1:

front/src/modules/pathfinding/components/Pathfinding/TypeAndPath.tsx
clarani: The trick here is that the function is `launchPathFinding` and not `launchPathfinding` ;)

I can change that but I think it's ok like this too

front/src/modules/pathfinding/components/Itinerary/Itinerary.tsx
clarani: No I can't because it's a modal, and it is not rendered within the ManageTrainSchedule ContextProvider :/

front/src/modules/pathfinding/hooks/usePathfinding.ts
clarani: Hum I still have the ts error, how did you do ?

front/src/modules/pathfinding/components/Pathfinding/Pathfinding.tsx
clarani: Can we do this in another PR ?

front/src/modules/pathfinding/components/Itinerary/ModalSuggestedVias.tsx
clarani: Hum no, unfortunately, we need to relaunch the pathfinding every time we delete a via, because maybe it modifies the path found by the algorithm.

Here is an example of workflow which will be broken:
- I create a path between A and C
- I add a via B on the map which changes the pathfinding
- the suggested OPs contain now B'
- I add a via on B'
- I remove B, the pathfinding is launched again and is still the same thanks to B'
- I remove B', but since it was added from the suggested OPs, the pathfinding is not relaunched
- however, if I refresh the page, the path will be different (initial path between A and C)

front/src/modules/pathfinding/components/Itinerary/DisplayItinerary/Vias.tsx
clarani: I think it does not work, because I can have an origin, a destination and several vias, and then remove the origin, the component will still be displayed :/

front/src/applications/operationalStudies/hooks/useManageTrainScheduleContext.tsx
```diff
@@ -0,0 +1,92 @@
...
+import type { RangedValue } from 'common/types';
+import getPathVoltages from 'modules/pathfinding/helpers/getPathVoltages';
+import usePathfinding from 'modules/pathfinding/hooks/usePathfinding';
+import type { PathfindingState } from 'modules/pathfinding/types';
+import { upsertPathStepsInOPs } from 'modules/pathfinding/utils';
+import type { PathStep } from 'reducers/osrdconf/types';
+
+import type { ManageTrainSchedulePathProperties } from '../types';
+
+type ManageTrainScheduleContextType = {
```
anisometropie, line 12: why do we carry that `manage` every where. can’t this just be TrainScheduleContextType ?

front/src/applications/operationalStudies/hooks/useManageTrainScheduleContext.tsx
```diff
@@ -0,0 +1,92 @@
...
+}: ManageTrainScheduleContextProviderProps) => {
+  const { getPathSteps } = useOsrdConfSelectors();
+  const pathSteps = useSelector(getPathSteps);
+
+  const [pathProperties, setPathProperties] = useState<ManageTrainSchedulePathProperties>();
+
+  const { launchPathfinding, pathfindingState, infraInfo } = usePathfinding(setPathProperties);
+
+  useEffect(() => {
+    if (pathProperties) {
```
anisometropie: the dependencies array is empty, so if this useEffect is called only once on the first render , why would pathProperties not be undefined?

front/src/applications/operationalStudies/hooks/useManageTrainScheduleContext.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+export const ManageTrainScheduleContextProvider = ({
+  children,
+}: ManageTrainScheduleContextProviderProps) => {
+  const { getPathSteps } = useOsrdConfSelectors();
+  const pathSteps = useSelector(getPathSteps);
+
+  const [pathProperties, setPathProperties] = useState<ManageTrainSchedulePathProperties>();
+
+  const { launchPathfinding, pathfindingState, infraInfo } = usePathfinding(setPathProperties);
```
anisometropie, line 30: I kinda fail to see what more that context does than just using `usePathfinding` directly.

front/src/modules/pathfinding/components/Itinerary/ModalSuggestedVias.tsx
SharglutDev: You're totally right !

front/src/modules/pathfinding/components/Pathfinding/Pathfinding.tsx
anisometropie: yes, keep this separate

SharglutDev: Sure I created an issue

front/src/modules/pathfinding/components/Pathfinding/TypeAndPath.tsx
anisometropie: This can be really confusing to see which one is called. I would call that `handleClick` since it’s an event handler

front/src/modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/setPointIti.ts
```diff
@@ -1,35 +1,37 @@
...
-      store.dispatch(updateDestination(pathStep));
+      newPathSteps = replaceElementAtIndex(pathSteps, -1, pathStep);
       break;
     default:
       if (pathProperties) {
-        store.dispatch(addVia({ newVia: pathStep, pathProperties }));
+        newPathSteps = insertViaFromMap(pathSteps, pathStep, pathProperties);
       } else {
-        store.dispatch(
-          replaceItinerary(addElementAtIndex(pathSteps, pathSteps.length - 1, pathStep))
```
anisometropie, line 30: how is this updated in the store now?

anisometropie (changes requested):

front/src/modules/pathfinding/hooks/usePathfinding.ts
SharglutDev: hmmm 

Capture d’écran 2024-12-12 à 17 31 05

front/src/modules/pathfinding/components/Itinerary/Itinerary.tsx
SharglutDev: Oh right ok

front/src/modules/pathfinding/components/Pathfinding/TypeAndPath.tsx
SharglutDev: Haha ok, took me a while to spot the difference. Maybe it's still better to rename it

front/src/modules/pathfinding/components/Itinerary/DisplayItinerary/Vias.tsx
SharglutDev: Right, my mistake. I thought this would remove it as well.

front/src/modules/pathfinding/hooks/usePathfinding.ts
SharglutDev: With your check l.181, we can be sure that all path steps aren't null so it should be ok to remove the `=== null` as well

front/src/applications/operationalStudies/hooks/useManageTrainScheduleContext.tsx
clarani: Hum I would rather keep the name of the view, even if I agree it's not the best naming ever. If you want to rename it, we can do it in another PR

front/src/modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/setPointIti.ts
clarani: The store is updated in launchPathfinding

front/src/applications/operationalStudies/hooks/useManageTrainScheduleContext.tsx
clarani: As we discussed it yesterday, I think we can resolve this conversation and talk about it next week ?

front/src/applications/operationalStudies/hooks/useManageTrainScheduleContext.tsx
clarani: I think I need to check with my other PR, but the dependencies should not be empty

front/src/modules/pathfinding/components/Pathfinding/TypeAndPath.tsx
clarani: Done

front/src/modules/pathfinding/hooks/usePathfinding.ts
SharglutDev: Ok so looks like I have the typescript 5.5+ on my setup (maybe because of vscode) which fixes this behavior but the app is still in 5.3.

We should keep this in mind when bumping typescript version

front/src/applications/operationalStudies/hooks/useManageTrainScheduleContext.tsx
anisometropie: it is empty line 51, so it’s only called once. and pathProperties has no reason to be defined here. so it looks like this useEffect never does anything

front/src/modules/pathfinding/hooks/usePathfinding.ts
SharglutDev: Same as next comment. Added issue here : https://github.com/OpenRailAssociation/osrd/issues/10081

front/src/modules/pathfinding/hooks/usePathfinding.ts
SharglutDev: issue is here : https://github.com/OpenRailAssociation/osrd/issues/10081

SharglutDev (approved):
Lgtm and tested, great job !

front/src/applications/operationalStudies/hooks/useManageTrainScheduleContext.tsx
clarani: Ok, I can remove it for now :+1:

anisometropie (approved):
thanks, good reworking

emersion (approved):

SharglutDev (approved):
Lgtm and tested again :)

Issue #9431, opened 2024-10-22
Type physical quantities in editoast


leovalais:
Tracking issue: #9435 

# Goals

- Have a more reliable way than comments to handle units.
- Explicit unit changes
- Type check quantity mismatch (e.g. `speed = length * time` shouldn't compile)
- Get rid of all manual conversions (millimeters to meters, etc)

# Constraints

- Should work with the current railjson format (which can change of course, but one migration at a time)
- Can work for integer and floating point arithmetic 

# Proposed solution: [`uom`](https://crates.io/crates/uom)

Benefits:

- Wide adoption
- Old, but still maintained
- Convenient API (a bit opinionated on the naming conventions)
- Supports all numeric types for quantities (even rational numbers!)
- Out of the box `serde` support

I experimented a bit with the `serde` interface of uom. We can't configure what is the unit the data we're (de)serializing: e.g. when serializing an `uom::si::u64::Length`, `uom` always serializes the value in `meters`. This is a bit annoying for us since we have weird units in our railjson format (`mm`, `km/s`, `minutes/100km` 🤯). Thankfully `serde` has a [feature](https://serde.rs/remote-derive.html) that will greatly limit the amount of boilerplate needed in most cases.

```rust
#[derive(Debug, Deserialize, Serialize)]
#[serde(remote = "uom::si::f64::Length")]
struct Millimeters(#[serde(getter = "uom::si::f64::Length::get::<millimeter>")] f64);

impl From<Millimeters> for uom::si::f64::Length {
    fn from(mm: Millimeters) -> Self {
        uom::si::f64::Length::new::<millimeter>(mm.0)
    }
}

#[derive(Debug, Deserialize, Serialize)]
pub struct UndirectedTrackRange {
    pub track_section: String,
    #[serde(with = "Millimeters")]
    pub begin: uom::si::f64::Length,
    #[serde(with = "Millimeters")]
    pub end: uom::si::f64::Length,
}
```

The approach of `uom` is great for our case since we only have to deal with units when we're importing, exporting or visualizing data. But when we use the quantity for computations, we only care about what the quantity is and not which unit we used to input it.

# Units in the RailJson format

> [!NOTE]
> This part is not in scope for this issue, but rather to initiate a discussion on how to manage quantities in the railjson format in general.

[...]

PR #9928, opened 2024-12-03
editoast: use uom to type units

Tristramg:
Refs [#9431](https://github.com/OpenRailAssociation/osrd/issues/9431)

This pull request isn’t ready, and was made to be a base for discussions. There are many values that still need to be typed

The models are hard to convert (I’m struggling with `derive(Model)`. But maybe it’s the right level of separation? In the database they are stored as floats, so maybe building the unit values in `impl From<RollingStockModel> for LightRollingStock` is actually the right place?

We get very long lines, such as `Acceleration::new::<meter_per_second_squared>(startup_acceleration)`. But I think it’s a good thing to have very explicit units.

In general I like that the units are now very explicit, and we no longer need comments to indicate what to do, and we can even remove them completly.

What do you think in general?

### DIFF
--- a/editoast/editoast_common/src/units.rs
+++ b/editoast/editoast_common/src/units.rs
@@ -167,3 +167,4 @@ define_unit!(kilogram_per_meter, AerodynamicDrag);
 define_unit!(per_meter, AerodynamicDragPerWeight);
 define_unit!(second, Time);
 define_unit!(millisecond, Time);
+define_unit!(basis_point, Ratio);
--- a/editoast/editoast_common/src/units.rs
+++ b/editoast/editoast_common/src/units.rs
@@ -43,6 +43,7 @@ pub type ViscosityFriction = uom::si::f64::MassRate;
 pub type ViscosityFrictionPerWeight = uom::si::f64::Frequency;
 pub type AerodynamicDrag = uom::si::f64::LinearMassDensity;
 pub type AerodynamicDragPerWeight = uom::si::f64::LinearNumberDensity;
+pub type Deceleration = uom::si::f64::Acceleration;
 
 macro_rules! quantity_to_path {
     (Length, $unit:ident) => {
--- a/editoast/editoast_common/src/units.rs
+++ b/editoast/editoast_common/src/units.rs
@@ -35,10 +35,14 @@
 //! ```
 
 /// Re-export the Quantities that are used in OSRD
-pub use uom::si::f64::{
-    Acceleration, Force, Frequency, Length, LinearMassDensity, LinearNumberDensity, Mass, MassRate,
-    Time, Velocity,
-};
+pub use uom::si::f64::{Acceleration, Length, Mass, Ratio, Time, Velocity};
+
+pub type SolidFriction = uom::si::f64::Force;
+pub type SolidFrictionPerWeight = uom::si::f64::Acceleration;
+pub type ViscosityFriction = uom::si::f64::MassRate;
+pub type ViscosityFrictionPerWeight = uom::si::f64::Frequency;
+pub type AerodynamicDrag = uom::si::f64::LinearMassDensity;
+pub type AerodynamicDragPerWeight = uom::si::f64::LinearNumberDensity;
 
 macro_rules! quantity_to_path {
     (Length, $unit:ident) => {
@@ -53,19 +57,19 @@ macro_rules! quantity_to_path {
     (Mass, $unit:ident) => {
         uom::si::mass::$unit
     };
-    (Force, $unit:ident) => {
+    (SolidFriction, $unit:ident) => {
         uom::si::force::$unit
     };
-    (MassRate, $unit:ident) => {
+    (ViscosityFriction, $unit:ident) => {
         uom::si::mass_rate::$unit
     };
-    (Frequency, $unit:ident) => {
+    (ViscosityFrictionPerWeight, $unit:ident) => {
         uom::si::frequency::$unit
     };
-    (LinearMassDensity, $unit:ident) => {
+    (AerodynamicDrag, $unit:ident) => {
         uom::si::linear_mass_density::$unit
     };
-    (LinearNumberDensity, $unit:ident) => {
+    (AerodynamicDragPerWeight, $unit:ident) => {
         uom::si::linear_number_density::$unit
     };
     (Time, $unit:ident) => {
@@ -79,8 +83,8 @@ macro_rules! quantity_to_path {
 macro_rules! define_unit {
     ($unit:ident, $quantity:ident) => {
         pub mod $unit {
+            use super::*;
             use serde::{Deserialize, Deserializer, Serialize, Serializer};
-            use uom::si::f64::*;
             type Unit = quantity_to_path!($quantity, $unit);
             pub type ReprType = f64;
 
@@ -155,10 +159,10 @@ define_unit!(millimeter, Length);
 define_unit!(meter_per_second, Velocity);
 define_unit!(meter_per_second_squared, Acceleration);
 define_unit!(kilogram, Mass);
-define_unit!(newton, Force);
-define_unit!(kilogram_per_second, MassRate);
-define_unit!(hertz, Frequency);
-define_unit!(kilogram_per_meter, LinearMassDensity);
-define_unit!(per_meter, LinearNumberDensity);
+define_unit!(newton, SolidFriction);
+define_unit!(kilogram_per_second, ViscosityFriction);
+define_unit!(hertz, ViscosityFrictionPerWeight);
+define_unit!(kilogram_per_meter, AerodynamicDrag);
+define_unit!(per_meter, AerodynamicDragPerWeight);
 define_unit!(second, Time);
 define_unit!(millisecond, Time);
--- a/editoast/editoast_common/src/units.rs
+++ b/editoast/editoast_common/src/units.rs
@@ -1,0 +1,164 @@
+//! Module to allow the use of serde with uom quantities
+//!
+//! The serde feature of uom doesn’t allow to specify in which unit the value will be serialized.
+//!
+//! Two helpers are provided for convenience:
+//! * `unit::new` (e.g. `meter::new(32)`) to build a new quantity from a f64 in the given unit
+//! * `unit::from` (e.g. `millimeter::from(length)`) to have the quantity as f64 in the given unit
+//!
+//! ## Usage
+//!
+//! ```ignore
+//! use editoast_model::units::*;
+//! #[derive(Debug, Serialize, Derivative)]
+//! struct Train {
+//!     // This means that serde with read and write the velocity in meters per second
+//!     #[serde(with="meter_per_second")]
+//!     max_speed: Velocity,
+//!     // When using optional values, we must add `default` and use ::option unit
+//!     // See https://stackoverflow.com/a/44303505
+//!     #[serde(default, with="meter::option")]
+//!     length: Option<Length>,
+//! }
+//!
+//! impl Train {
+//!     fn from_meter_per_seconds(mps: f64) -> Self {
+//!         Self {
+//!             max_speed: meter_per_second::new(mps),
+//!         }
+//!     }
+//!
+//!     fn print(&self) {
+//!         println!("The max speed is: {} km/h", kilometer_per_hour::from(self.max_speed));
+//!     }
+//! }
+//! ```
+
+/// Re-export the Quantities that are used in OSRD
+pub use uom::si::f64::{
+    Acceleration, Force, Frequency, Length, LinearMassDensity, LinearNumberDensity, Mass, MassRate,
+    Time, Velocity,
+};
+
+macro_rules! quantity_to_path {
+    (Length, $unit:ident) => {
+        uom::si::length::$unit
+    };
+    (Velocity, $unit:ident) => {
+        uom::si::velocity::$unit
+    };
+    (Acceleration, $unit:ident) => {
+        uom::si::acceleration::$unit
+    };
+    (Mass, $unit:ident) => {
+        uom::si::mass::$unit
+    };
+    (Force, $unit:ident) => {
+        uom::si::force::$unit
+    };
+    (MassRate, $unit:ident) => {
+        uom::si::mass_rate::$unit
+    };
+    (Frequency, $unit:ident) => {
+        uom::si::frequency::$unit
+    };
+    (LinearMassDensity, $unit:ident) => {
+        uom::si::linear_mass_density::$unit
+    };
+    (LinearNumberDensity, $unit:ident) => {
+        uom::si::linear_number_density::$unit
+    };
+    (Time, $unit:ident) => {
+        uom::si::time::$unit
+    };
+    (Ratio, $unit:ident) => {
+        uom::si::ratio::$unit
+    };
+}
+
+macro_rules! define_unit {
+    ($unit:ident, $quantity:ident) => {
+        pub mod $unit {
+            use serde::{Deserialize, Deserializer, Serialize, Serializer};
+            use uom::si::f64::*;
+            type Unit = quantity_to_path!($quantity, $unit);
+            pub type ReprType = f64;
+
+            pub fn serialize<S>(value: &$quantity, serializer: S) -> Result<S::Ok, S::Error>
+            where
+                S: Serializer,
+            {
+                value.get::<Unit>().serialize(serializer)
+            }
+
+            pub fn deserialize<'de, D>(deserializer: D) -> Result<$quantity, D::Error>
+            where
+                D: Deserializer<'de>,
+            {
+                let value = ReprType::deserialize(deserializer)?;
+                Ok($quantity::new::<Unit>(value))
+            }
+
+            pub fn new(value: ReprType) -> $quantity {
+                $quantity::new::<Unit>(value)
+            }
+
+            pub fn from(qty: $quantity) -> ReprType {
+                qty.get::<Unit>()
+            }
+
+            pub fn hash<H: std::hash::Hasher>(value: &$quantity, state: &mut H) {
+                crate::hash_float::<5, H>(&from(*value), state);
+            }
+
+            pub mod option {
+                use super::*;
+                pub type ReprType = Option<super::ReprType>;
+
+                pub fn serialize<S>(
+                    value: &Option<$quantity>,
+                    serializer: S,
+                ) -> Result<S::Ok, S::Error>
+                where
+                    S: Serializer,
+                {
+                    value.map(|value| value.get::<Unit>()).serialize(serializer)
+                }
+
+                pub fn deserialize<'de, D>(deserializer: D) -> Result<Option<$quantity>, D::Error>
+                where
+                    D: Deserializer<'de>,
+                {
+                    let value = Option::deserialize(deserializer)?;
+                    Ok(value.map(|value| $quantity::new::<Unit>(value)))
+                }
+
+                pub fn new(value: ReprType) -> Option<$quantity> {
+                    value.map(|v| $quantity::new::<Unit>(v))
+                }
+
+                pub fn from(qty: Option<$quantity>) -> ReprType {
+                    qty.map(|q| q.get::<Unit>())
+                }
+
+                pub fn hash<H: std::hash::Hasher>(value: &Option<$quantity>, state: &mut H) {
+                    super::hash(&value.unwrap_or_default(), state);
+                }
+            }
+        }
+    };
+}
+
+// Any new value here must also be added in editoast_derive/src/annotate_units.rs
+define_unit!(meter, Length);
+define_unit!(millimeter, Length);
+define_unit!(meter_per_second, Velocity);
+define_unit!(meter_per_second_squared, Acceleration);
+define_unit!(kilogram, Mass);
+define_unit!(newton, Force);
+define_unit!(kilogram_per_second, MassRate);
+define_unit!(hertz, Frequency);
+define_unit!(kilogram_per_meter, LinearMassDensity);
+define_unit!(per_meter, LinearNumberDensity);
+define_unit!(second, Time);
+define_unit!(millisecond, Time);
--- a/editoast/editoast_common/src/units.rs
+++ b/editoast/editoast_common/src/units.rs
@@ -35,15 +35,16 @@
 //! ```
 
 /// Re-export the Quantities that are used in OSRD
-pub use uom::si::f64::{Acceleration, Length, Mass, Ratio, Time, Velocity};
-
-pub type SolidFriction = uom::si::f64::Force;
-pub type SolidFrictionPerWeight = uom::si::f64::Acceleration;
-pub type ViscosityFriction = uom::si::f64::MassRate;
-pub type ViscosityFrictionPerWeight = uom::si::f64::Frequency;
-pub type AerodynamicDrag = uom::si::f64::LinearMassDensity;
-pub type AerodynamicDragPerWeight = uom::si::f64::LinearNumberDensity;
-pub type Deceleration = uom::si::f64::Acceleration;
+pub mod quantities {
+    pub use uom::si::f64::{Acceleration, Length, Mass, Ratio, Time, Velocity};
+    pub type SolidFriction = uom::si::f64::Force;
+    pub type SolidFrictionPerWeight = uom::si::f64::Acceleration;
+    pub type ViscosityFriction = uom::si::f64::MassRate;
+    pub type ViscosityFrictionPerWeight = uom::si::f64::Frequency;
+    pub type AerodynamicDrag = uom::si::f64::LinearMassDensity;
+    pub type AerodynamicDragPerWeight = uom::si::f64::LinearNumberDensity;
+    pub type Deceleration = uom::si::f64::Acceleration;
+}
 
 macro_rules! quantity_to_path {
     (Length, $unit:ident) => {
@@ -199,6 +200,7 @@ macro_rules! define_unit {
 }
 
 // Any new value here must also be added in editoast_derive/src/annotate_units.rs
+use quantities::*;
 define_unit!(meter, Length);
 define_unit!(millimeter, Length);
 define_unit!(meter_per_second, Velocity);
--- a/editoast/editoast_common/src/units.rs
+++ b/editoast/editoast_common/src/units.rs
@@ -150,6 +150,50 @@ macro_rules! define_unit {
                     super::hash(&value.unwrap_or_default(), state);
                 }
             }
+
+            pub mod u64 {
+                use super::*;
+
+                pub fn serialize<S>(value: &$quantity, serializer: S) -> Result<S::Ok, S::Error>
+                where
+                    S: Serializer,
+                {
+                    (value.get::<Unit>() as u64).serialize(serializer)
+                }
+
+                pub fn deserialize<'de, D>(deserializer: D) -> Result<$quantity, D::Error>
+                where
+                    D: Deserializer<'de>,
+                {
+                    super::deserialize(deserializer)
+                }
+
+                pub mod option {
+                    use super::*;
+                    pub type ReprType = Option<super::ReprType>;
+
+                    pub fn serialize<S>(
+                        value: &Option<$quantity>,
+                        serializer: S,
+                    ) -> Result<S::Ok, S::Error>
+                    where
+                        S: Serializer,
+                    {
+                        value
+                            .map(|value| value.get::<Unit>() as u64)
+                            .serialize(serializer)
+                    }
+
+                    pub fn deserialize<'de, D>(
+                        deserializer: D,
+                    ) -> Result<Option<$quantity>, D::Error>
+                    where
+                        D: Deserializer<'de>,
+                    {
+                        super::super::option::deserialize(deserializer)
+                    }
+                }
+            }
         }
     };
 }
--- a/editoast/editoast_derive/src/annotate_units.rs
+++ b/editoast/editoast_derive/src/annotate_units.rs
@@ -39,6 +39,7 @@ fn get_abbreviation(value: &str) -> Option<&'static str> {
         "kilogram_per_meter" => Some("Aerodynamic drag in kg·m⁻¹"),
         "kilogram_per_second" => Some("Viscosity friction in kg·s⁻¹"),
         "per_meter" => Some("Aerodynamic drag per kg in m⁻¹"),
+        "basis_point" => Some("Ratio 1:1"),
         _ => None,
     }
 }
--- a/editoast/editoast_derive/src/annotate_units.rs
+++ b/editoast/editoast_derive/src/annotate_units.rs
@@ -1,0 +1,75 @@
+use darling::Result;
+use proc_macro2::TokenStream;
+use quote::quote;
+use syn::{parse_quote, DeriveInput, LitStr};
+
+const OPTIONAL_SUFFIX: &str = "::option";
+struct UnitType {
+    unit: String,
+    optional: bool,
+}
+
+impl UnitType {
+    fn new(path: LitStr) -> Self {
+        let path = path.value();
+        let optional = path.ends_with(OPTIONAL_SUFFIX);
+        let unit = path
+            .strip_suffix(OPTIONAL_SUFFIX)
+            .unwrap_or(&path)
+            .rsplit("::")
+            .next()
+            .expect("String::split to return at least an empty string")
+            .to_owned();
+        Self { unit, optional }
+    }
+}
+
+fn get_abbreviation(value: &str) -> Option<&'static str> {
+    // Any new value here must also be added in editoast_common/src/units.rs
+    match value {
+        "second" => Some("Duration in s"),
+        "millisecond" => Some("Duration in ms"),
+        "meter" => Some("Length in m"),
+        "millimeter" => Some("Length in mm"),
+        "meter_per_second" => Some("Velocity in m·s⁻¹"),
+        "meter_per_second_squared" => Some("Acceleration in m·s⁻²"),
+        "kilogram" => Some("Mass in kg"),
+        "newton" => Some("Solid Friction in N"),
+        "hertz" => Some("Viscosity friction per weight in s⁻¹"),
+        "kilogram_per_meter" => Some("Aerodynamic drag in kg·m⁻¹"),
+        "kilogram_per_second" => Some("Viscosity friction in kg·s⁻¹"),
+        "per_meter" => Some("Aerodynamic drag per kg in m⁻¹"),
+        _ => None,
+    }
+}
+
+pub fn annotate_units(input: &mut DeriveInput) -> Result<TokenStream> {
+    // We look for fields that have #[sered(with="meter")] attributes
+    // and we push two new attributes to it to improve the OpenAPI
+    if let syn::Data::Struct(s) = &mut input.data {
+        for f in s.fields.iter_mut() {
+            for attr in f.attrs.clone() {
+                if attr.path().is_ident("serde") {
+                    let _ = attr.parse_nested_meta(|meta| {
+                        if meta.path.is_ident("with") {
+                            let value = meta.value()?;
+                            let s: LitStr = value.parse()?;
+                            let unit = UnitType::new(s);
+                            if let Some(abbreviation) = get_abbreviation(&unit.unit) {
+                                if unit.optional {
+                                    f.attrs
+                                        .push(parse_quote! {#[schema(value_type = Option<f64>)]});
+                                } else {
+                                    f.attrs.push(parse_quote! {#[schema(value_type = f64)]});
+                                }
+                                f.attrs.push(parse_quote! {#[doc = #abbreviation]});
+                            }
+                        }
+                        Ok(())
+                    });
+                }
+            }
+        }
+    }
+    Ok(quote! {#input})
+}
--- a/editoast/editoast_derive/src/lib.rs
+++ b/editoast/editoast_derive/src/lib.rs
@@ -1,5 +1,6 @@
 extern crate proc_macro;
 
+mod annotate_units;
 mod error;
 mod model;
 mod search;
@@ -269,3 +270,20 @@ mod test_utils {
 
 #[cfg(test)]
 use test_utils::assert_macro_expansion;
+/// Annotates fields of a structs with documentation and value_type for a better utoipa schema
+///
+/// It must be used on structs that use #[derive(ToSchema)]
+///
+/// On every field that has an attribute such as #[serde(with="millimeter")]
+/// It will add:
+/// * #[schema(value_type = f64)]
+/// * /// Length in mm
+#[proc_macro_attribute]
+pub fn annotate_units(_attr: TokenStream, input: TokenStream) -> TokenStream {
+    // We are using a macro attribute to modify in place the attributes of fields to annotate
+    // This requires to mutate the input
+    let mut input = parse_macro_input!(input as DeriveInput);
+    annotate_units::annotate_units(&mut input)
+        .unwrap_or_else(darling::Error::write_errors)
+        .into()
+}
--- a/editoast/editoast_derive/src/lib.rs
+++ b/editoast/editoast_derive/src/lib.rs
@@ -218,6 +218,7 @@ pub fn search_config_store(input: proc_macro::TokenStream) -> proc_macro::TokenS
 /// * `#[model(to_string)]`: calls `to_string()` before writing the field to the database and calls `String::from` after reading (diesel column type: String)
 /// * `#[model(to_enum)]`: is converted as `u8` before writing the field to the database and calls `FromRepr::from_repr` after reading (diesel column type: TinyInt)
 /// * `#[model(remote = "T")]`: calls `Into::<T>::into` before writing the field to the database and calls `T::from` after reading (diesel column type: T)
+/// * `#[model(uom_unit = "path::to::unit")]`: the value is the path to an unit defined in editoast_common, e.g. `"editoast_common::units::meter"`
 /// * `#[model(geo)]` **TODO**: TBD
 ///
 /// #### A note on identifiers
--- a/editoast/editoast_derive/src/model/parsing.rs
+++ b/editoast/editoast_derive/src/model/parsing.rs
@@ -165,6 +165,7 @@ impl ModelField {
             value.geo,
             value.to_string,
             to_enum,
+            value.uom_unit,
         )
         .map_err(|e| e.with_span(&ident))?;
         Ok(Self {
@@ -188,16 +189,18 @@ impl FieldTransformation {
         geo: bool,
         to_string: bool,
         to_enum: Option<syn::Type>,
+        uom_unit: Option<syn::Path>,
     ) -> darling::Result<Option<Self>> {
-        match (remote, json, geo, to_string, to_enum) {
-            (Some(ty), false, false, false, None) => Ok(Some(Self::Remote(ty))),
-            (None, true, false, false, None) => Ok(Some(Self::Json)),
-            (None, false, true, false, None) => Ok(Some(Self::Geo)),
-            (None, false, false, true, None) => Ok(Some(Self::ToString)),
-            (None, false, false, false, Some(ty)) => Ok(Some(Self::ToEnum(ty))),
-            (None, false, false, false, None) => Ok(None),
+        match (remote, json, geo, to_string, to_enum, uom_unit) {
+            (Some(ty), false, false, false, None, None) => Ok(Some(Self::Remote(ty))),
+            (None, true, false, false, None, None) => Ok(Some(Self::Json)),
+            (None, false, true, false, None, None) => Ok(Some(Self::Geo)),
+            (None, false, false, true, None, None) => Ok(Some(Self::ToString)),
+            (None, false, false, false, Some(ty), None) => Ok(Some(Self::ToEnum(ty))),
+            (None, false, false, false, None, Some(ty)) => Ok(Some(Self::UomUnit(ty))),
+            (None, false, false, false, None, None) => Ok(None),
             _ => Err(Error::custom(
-                "Model: remote, json, geo, to_string and to_enum attributes are mutually exclusive",
+                "Model: remote, json, geo, to_string, to_enum and uom_unit attributes are mutually exclusive",
             )),
         }
     }
--- a/editoast/editoast_schemas/src/rolling_stock.rs
+++ b/editoast/editoast_schemas/src/rolling_stock.rs
@@ -76,8 +76,8 @@ pub struct RollingStock {
     #[serde(with = "meter_per_second_squared")]
     pub const_gamma: Deceleration,
     pub etcs_brake_params: Option<EtcsBrakeParams>,
-    #[serde(with = "meter_per_second_squared")]
-    pub inertia_coefficient: Acceleration,
+    #[serde(with = "basis_point")]
+    pub inertia_coefficient: Ratio,
     #[serde(with = "kilogram")]
     pub mass: Mass,
     pub rolling_resistance: RollingResistance,
--- a/editoast/editoast_schemas/src/rolling_stock.rs
+++ b/editoast/editoast_schemas/src/rolling_stock.rs
@@ -36,7 +36,10 @@ pub use rolling_stock_livery::RollingStockLiveryMetadata;
 mod towed_rolling_stock;
 pub use towed_rolling_stock::TowedRollingStock;
 
-use editoast_common::units::*;
+use editoast_common::units;
+use editoast_common::units::quantities::{
+    Acceleration, Deceleration, Length, Mass, Ratio, Time, Velocity,
+};
 use serde::Deserialize;
 use serde::Serialize;
 use std::collections::HashMap;
@@ -61,24 +64,24 @@ pub struct RollingStock {
     pub locked: bool,
     pub effort_curves: EffortCurves,
     pub base_power_class: Option<String>,
-    #[serde(with = "meter")]
+    #[serde(with = "units::meter")]
     pub length: Length,
-    #[serde(with = "meter_per_second")]
+    #[serde(with = "units::meter_per_second")]
     pub max_speed: Velocity,
-    #[serde(with = "second")]
+    #[serde(with = "units::second")]
     pub startup_time: Time,
-    #[serde(with = "meter_per_second_squared")]
+    #[serde(with = "units::meter_per_second_squared")]
     pub startup_acceleration: Acceleration,
-    #[serde(with = "meter_per_second_squared")]
+    #[serde(with = "units::meter_per_second_squared")]
     pub comfort_acceleration: Acceleration,
     // The constant gamma braking coefficient used when NOT circulating
     // under ETCS/ERTMS signaling system
-    #[serde(with = "meter_per_second_squared")]
+    #[serde(with = "units::meter_per_second_squared")]
     pub const_gamma: Deceleration,
     pub etcs_brake_params: Option<EtcsBrakeParams>,
-    #[serde(with = "basis_point")]
+    #[serde(with = "units::basis_point")]
     pub inertia_coefficient: Ratio,
-    #[serde(with = "kilogram")]
+    #[serde(with = "units::kilogram")]
     pub mass: Mass,
     pub rolling_resistance: RollingResistance,
     pub loading_gauge: LoadingGaugeType,
@@ -89,11 +92,11 @@ pub struct RollingStock {
     pub energy_sources: Vec<EnergySource>,
     /// The time the train takes before actually using electrical power (in seconds).
     /// Is null if the train is not electric.
-    #[serde(default, with = "second::option")]
+    #[serde(default, with = "units::second::option")]
     pub electrical_power_startup_time: Option<Time>,
     /// The time it takes to raise this train's pantograph in seconds.
     /// Is null if the train is not electric.
-    #[serde(default, with = "second::option")]
+    #[serde(default, with = "units::second::option")]
     pub raise_pantograph_time: Option<Time>,
     pub supported_signaling_systems: RollingStockSupportedSignalingSystems,
     pub railjson_version: String,
--- a/editoast/editoast_schemas/src/rolling_stock.rs
+++ b/editoast/editoast_schemas/src/rolling_stock.rs
@@ -36,6 +36,7 @@ pub use rolling_stock_livery::RollingStockLiveryMetadata;
 mod towed_rolling_stock;
 pub use towed_rolling_stock::TowedRollingStock;
 
+use editoast_common::units::*;
 use serde::Deserialize;
 use serde::Serialize;
 use std::collections::HashMap;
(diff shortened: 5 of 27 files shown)


editoast/editoast_schemas/src/rolling_stock.rs
```diff
@@ -58,14 +59,14 @@ pub struct RollingStock {
     pub effort_curves: EffortCurves,
     pub base_power_class: Option<String>,
     /// In m
-    pub length: f64,
+    pub length: Length,
     /// In m/s
-    pub max_speed: f64,
-    pub startup_time: f64,
+    pub max_speed: Velocity,
+    pub startup_time: f64, // What unit ?
```
leovalais: [Time](https://docs.rs/uom/0.36.0/uom/si/time/type.Time.html) I guess?

leovalais (commented):
Looks great! 😃 

> We get very long lines, such as Acceleration::new::<meter_per_second_squared>(startup_acceleration). But I think it’s a good thing to have very explicit units.

I strongly agree. This makes reading the code much easier.

> The models are hard to convert (I’m struggling with derive(Model). But maybe it’s the right level of separation? In the database they are stored as floats, so maybe building the unit values in impl From<RollingStockModel> for LightRollingStock is actually the right place?

I'd rather have the `uom` quantities directly in the model. I believe you can use `#[model(remote = "f64")]` above the fields to instruct the macro to generate a few `.into()`s at the right places. This won't work if `uom::Quantity` doesn't implement `From<f64>` or if non-base units are stored. In which case we either need an adapter, or a new field transformation in `Model` like `#[model(remote = f64, unit = millimeter)]`. Wdyt about that? I can help with the derive macro, it's fairly easy to add such transformations.

---

I hope we can have this working as its makes everything so much clearer!

(Also we're in editoast around here, not in core, cf. PR title 😛)

Tristramg:
@leovalais thank you for the feedback. I tried to fiddle with macro by myself, but I have little experience here. See the commit https://github.com/OpenRailAssociation/osrd/pull/9928/commits/590a971fc1f43cd8127ca5b6dd25080f90da200b  for the change in the `Model` macro and [an example on one field](https://github.com/OpenRailAssociation/osrd/pull/9928/commits/38c7da73fcdad04dc1ed88a96dac4668c26b2e01)

If it’s ok for you, I’ll go brrr on all the fields of rolling stock.

I’m a bit worried about the interface with the http part: isn’t there a risk that serde will use the wrong unit?

editoast/editoast_schemas/src/rolling_stock.rs
```diff
@@ -54,14 +55,14 @@ pub struct RollingStock {
     pub effort_curves: EffortCurves,
     pub base_power_class: Option<String>,
     /// In m
-    pub length: f64,
+    pub length: Length,
     /// In m/s
-    pub max_speed: f64,
-    pub startup_time: f64,
+    pub max_speed: Velocity,
```
woshilapin, line 70: I would have expected this type to contain the unity, but it's not. If I understand, it would be possible to do, but that would be against the default generic which seems to be System International (meter, kilograms, seconds, etc.). Not sure what are your opinions about that? If I got it right, in any case, the type system will protect us against misuse of unities anyway...

editoast/editoast_derive/src/model/parsing.rs
```diff
@@ -188,16 +189,19 @@ impl FieldTransformation {
...
-            (None, false, false, true, None) => Ok(Some(Self::ToString)),
-            (None, false, false, false, Some(ty)) => Ok(Some(Self::ToEnum(ty))),
-            (None, false, false, false, None) => Ok(None),
+        match (remote, json, geo, to_string, to_enum, uom_unit) {
+            (Some(ty), false, false, false, None, None) => Ok(Some(Self::Remote(ty))),
+            (None, true, false, false, None, None) => Ok(Some(Self::Json)),
+            (None, false, true, false, None, None) => Ok(Some(Self::Geo)),
+            (None, false, false, true, None, None) => Ok(Some(Self::ToString)),
+            (None, false, false, false, Some(ty), None) => Ok(Some(Self::ToEnum(ty))),
+            //(None, false, false, false, None, Some(ty)) => Ok(Some(Self::Unit(ty))),
```
woshilapin: Can probably be removed before merge.

editoast/editoast_derive/src/model/parsing.rs
```diff
@@ -188,16 +189,19 @@ impl FieldTransformation {
...
+            (None, false, false, false, None, Some(ty)) => Ok(Some(Self::UomUnit(ty))),
+            (None, false, false, false, None, None) => Ok(None),
             _ => Err(Error::custom(
-                "Model: remote, json, geo, to_string and to_enum attributes are mutually exclusive",
+                "Model: remote, json, geo, to_string, uom_unit and to_enum attributes are mutually exclusive",
```
woshilapin: nit: Just to keep the same order everywhere.

```suggestion
                "Model: remote, json, geo, to_string, to_enum and uom_unit attributes are mutually exclusive",
```

editoast/src/core/simulation.rs
```diff
@@ -38,22 +39,21 @@ editoast_common::schemas! {
...
+    #[derivative(Hash(hash_with = "editoast_common::hash_acceleration::<5,_>"))]
+    pub startup_acceleration: Acceleration,
+    #[derivative(Hash(hash_with = "editoast_common::hash_acceleration::<5,_>"))]
+    pub comfort_acceleration: Acceleration,
     /// The constant gamma braking coefficient used when NOT circulating
     /// under ETCS/ERTMS signaling system in m/s^2
+    #[derivative(Hash(hash_with = "editoast_common::hash_float::<5,_>"))]
     pub const_gamma: f64,
     #[derivative(Hash(hash_with = "editoast_common::hash_float::<5,_>"))]
     pub inertia_coefficient: f64,
```
woshilapin: Shouldn't `intertia_coefficient` and `const_gamma` be of some units too? Not sure if they are ratio or actual units 🤷

editoast/editoast_schemas/src/rolling_stock.rs
woshilapin, line 1: I would also expect `RollingResistance` and `RollingResistancePerWeight` to be converted to `uom` too (I actually got some gray hairs because units were a mess in those).

editoast/src/core/simulation.rs
```diff
@@ -96,51 +96,41 @@ impl PhysicsConsistParameters {
...
             .map(|trs| trs.length)
-            .unwrap_or(0.0);
-
-        let length = self
-            .total_length
-            .unwrap_or(self.traction_engine.length + towed_rolling_stock_length);
-
-        // Convert m to mm
-        let length = length * 1000.0;
+            .unwrap_or_default();
```
woshilapin: When I see some `* 1000` disappear in this refactor, I'm thinking both "nice, we now **know**, thanks to the type system, that it's not the place to do it" 🚀 and "we're going to need a testing of the full OSRD stack once this PR is going to be merged to see if we communicate correctly with the `front` and `core`" 😨

editoast/src/core/simulation.rs
```diff
@@ -507,9 +501,9 @@ mod tests {
 
     fn create_physics_consist() -> PhysicsConsistParameters {
         PhysicsConsistParameters {
-            total_length: Some(100.0),
+            total_length: Some(Length::new::<meter>(100.0)),
```
woshilapin: Interesting that we can use `Velocity` as field types, without the need to specify the unit, but the `new` method **needs** to specify the unit. I'd say it might actually be a good design choice since the function is taking a literal (untyped considering the units)...

woshilapin (commented):
OK, in general, very interesting to discover how the integration of `uom` takes shape in a real project. After reading the entire PR, I now understand why fields are defined as `Length` and not `Length<meter>`: it doesn't really matter how it is stored. What is important is how you put it inside (and the `new` function **needs** the unit) and how you get it out (and the `get` function & friends **need** units).

That said, even if I now understand the technical reason, which seems to be a great design of the API, I'm still wondering how it will work in practice. As a developer, I'll probably still ask "what does `editoast` takes as inputs units?" then take a look at the endpoint, then the structs and look inside the struct, failing to find the information.

Introduction of `uom` seems to change the approach:
- before, we consider that a `length` is always in meter through the entire system (but not typed, hence we had bugs, but that's not the point here)
- with `uom`, we don't care how a `length` is stored, we care how it's read, so we might take millimeter on an endpoint, meter on another endpoint. Sure, in the end, everything is typed inside `editoast`, but not sure it's going to help our API consumer to get the correct information inside the system.

Not sure what to make of that information. Happy to discuss at some point.

editoast/editoast_schemas/src/rolling_stock.rs
Tristramg: There are many fields to be converted. I’m resolving for now until we decide to use uom and I’ll change all units

editoast/src/core/simulation.rs
Tristramg: You are pointing the thing that worries me. In the database, with the macro, we should be quite confident that everything is coherent.

However, with the serde extension, I _suppose_ that by default it will consider mksa units. Maybe we should write explicit serialization/deserialization functions to be sure. And yes, testing by hand, a lot

Tristramg:
https://github.com/OpenRailAssociation/osrd/actions/runs/12177764809/job/33967384585
Those checks on openapi will be helpful to me sure we didn’t miss anything in input/output (and check the documentation). Nice work!

leovalais (commented):
The macro changes look fine to me 😄 

I have not tested, but I'm pretty sure that the serialization to core and the deserialization from frontend requests will break. `uom` only uses base units in these cases, and that's not configurable on their end. We'll likely need adapters using `serde(remote`. These could be placed into `editoast_common`, wdyt? @Tristramg @woshilapin 

```rust
#[derive(Debug, Deserialize, Serialize)]
#[serde(remote = "uom::si::f64::Length")]
struct Millimeters(#[serde(getter = "uom::si::f64::Length::get::<millimeter>")] f64);

impl From<Millimeters> for uom::si::f64::Length {
    fn from(mm: Millimeters) -> Self {
        uom::si::f64::Length::new::<millimeter>(mm.0)
    }
}

#[derive(Debug, Deserialize, Serialize)]
pub struct UndirectedTrackRange {
    pub track_section: String,
    #[serde(with = "Millimeters")]
    pub begin: uom::si::f64::Length,
    #[serde(with = "Millimeters")]
    pub end: uom::si::f64::Length,
}
```

I guess this also answers @woshilapin question about where to get the unit information from editoast API.

Tristramg:
Thank you for the feedback.

Here are the main updates, the big stuff being in https://github.com/OpenRailAssociation/osrd/pull/9928/commits/0b1c9fc0966f9cf232e921ecffdc464fbc1dfe6d

I wanted to make a macro to generate all the structures, and it interfered with `serde(remote)` and `serde(getter)`. I’m not sure what was happing. So I used a mod to be able to deserialize our structures.

I also added converters in the mod to simplify a bit the usage in the code.

I’m not so happy with the way to handle `Option`, but it’s probably ok in the end.

leovalais:
Too bad `serde remote` doesn't work, but the `macro_rules` is not that complicated and it's straightforward to support new units.


As we discussed, we need to document the units in the OpenAPI. It'd be great to have some mechanism to avoid repeating the unit in a doc comment, and instruct `utoipa` to use a `float` when necessary. But it may require a fair amount of work to do that using metaprog. It's probably fine to repeat the unit in a comment, we'll just have to be extra careful when reviewing that the documented unit matches the one used in `serde(with =`.

Tristramg:
:partying_face: :tada: https://github.com/OpenRailAssociation/osrd/pull/9928/commits/d5e994006e31693f8a1afcb2601c6ce75d6c16cd :balloon:

leovalais:
So nice! I didn't know we could do that using proc-macros! (I guess Rust macro system isn't that bad after all... 😁) Thanks for keeping digging this, it's much better than manual comments!

Maybe a comment would be useful in `units.rs` above the `define_units!(...)` indicating that the new unit should also be added to the proc-macro hash map, and vice versa.

Or to avoid having units defined in two place, maybe `define_unit!` could define a `doc!` macro inside the module that expands to the doc literal, and call that macro in `#[doc = #unit::doc!()]`. Not too sure about this one, but it would make adding units easier. Wdyt?

Tristramg:
I was thinking calling from `editost_derive` the properties from `editoast_common`, but wouldn’t that defeat the separation of crates?

We might, maybe, also pull the information from the `uom` crate, but I got stock when trying to transform the string into the path

leovalais:
I'm not suggesting to "use" definitions from `common` in `derive`, just rather to expect in the macro expansion some definitions to be "there". What I'm suggesting would be something in the lines of:

```rust
// in common

define_unit!(Length, meter, "Length in m");
// expanding to
mod meter {
	pub fn serialize() { todo!() }
	// ...
	macro_rules! doc { () => { "Length in m" } }
	pub use doc!;
}

// and then, in derive
for
	for
		...
			f.attrs.push(quote! { #[doc = #unit::doc!()] });

```

I'm not sure what you mean by "also pull the information from the uom crate", does the crate already provide the units abbreviations?

Tristramg:
A new iteration, from my point of view, this is mergeable (I just need to squash commits, I tried to keep separated the specific changes.

In particular: in common https://github.com/OpenRailAssociation/osrd/pull/9928/commits/97fb3f03f9be18fbd674d93a9453557b245dcbaf it adds a hashing function and the type that is used to store the data. The goal is to maybe use a rational value on the long term? (plus it helps for some troubles I had with Options)

In derive, we now pass the unit module defined in `derive` instead of the unit from uom. This allows to use the `ReprType`defined in `common` making thing SOOOO much easier for Option https://github.com/OpenRailAssociation/osrd/pull/9928/commits/c95441abde9abb02a1ce6456a2831b9081fecbd5

What do you think? Do you think other things could be improved or are we good to go?

editoast/editoast_common/src/hash_rounded_float.rs
```diff
@@ -1,11 +1,79 @@
...
 
 /// Hash a float through a rounded integer value
 /// `hash_float<3,_>` means that the value is rounded to the nearest thousandth
 pub fn hash_float<const T: u8, H: Hasher>(value: &f64, state: &mut H) {
     ((value * 10i64.pow(T as u32) as f64).round() as i64).hash(state);
 }
 
+/// Hash a length through a rounded integer value
+/// `hash_length<3,_>` means that the value is rounded to the nearest thousandth
+pub fn hash_length<const T: u8, H: Hasher>(value: &Length, state: &mut H) {
```
leovalais: Do you still need these functions since `define_unit!` now generates them?

editoast/editoast_common/src/units.rs
```diff
@@ -0,0 +1,163 @@
...
+//! }
+//! ```
+
+/// Re-export the Quantities that are used in OSRD
+pub use uom::si::f64::{
+    Acceleration, Force, Frequency, Length, LinearMassDensity, LinearNumberDensity, Mass, MassRate,
+    Time, Velocity,
+};
+
+// Todo: maybe replace with https://crates.io/crates/casey
```
leovalais: Not sure we need a new crate in our (already big) dependency tree. The repetition looks good enough to me 🤷

You can use [`paste`](https://docs.rs/paste/1.0.15/paste/#case-conversion) which provides a similar feature: it's already used so we won't have to pull another package. (But it has been deprecated recently, maybe we should remove it as well 🤔)

Wdyt?

editoast/editoast_schemas/src/rolling_stock.rs
```diff
@@ -53,21 +54,22 @@ pub struct RollingStock {
...
+    #[serde(with = "second")]
+    pub startup_time: Time,
+    #[serde(with = "meter_per_second_squared")]
+    pub startup_acceleration: Acceleration,
+    #[serde(with = "meter_per_second_squared")]
+    pub comfort_acceleration: Acceleration,
     // The constant gamma braking coefficient used when NOT circulating
     // under ETCS/ERTMS signaling system in m/s^2
     pub const_gamma: f64,
```
leovalais: Shouldn't this be an `Acceleration` as well?

editoast/editoast_schemas/src/rolling_stock/rolling_resistance.rs
```diff
@@ -16,30 +18,37 @@ pub struct RollingResistance {
...
-    /// Viscosity friction in N/(m/s)
-    #[derivative(Hash(hash_with = "editoast_common::hash_float::<5,_>"))]
-    pub B: f64,
-    /// Aerodynamic drag in N/(m/s)²
-    #[derivative(Hash(hash_with = "editoast_common::hash_float::<5,_>"))]
-    pub C: f64,
+    #[derivative(Hash(hash_with = "newton::hash"))]
+    #[serde(with = "newton")]
+    pub A: Force,
+    /// Viscosity friction in N/(m/s) - N = kg⋅m⋅s−2 => viscosity friction kg/s
```
leovalais:
```suggestion
    /// Viscosity friction in N/(m/s) -- N = kg⋅m⋅s⁻² => viscosity friction kg/s
```

editoast/editoast_schemas/src/rolling_stock/rolling_resistance.rs
```diff
@@ -16,30 +18,37 @@ pub struct RollingResistance {
...
-    #[derivative(Hash(hash_with = "editoast_common::hash_float::<5,_>"))]
-    pub B: f64,
-    /// Aerodynamic drag in (N/kg)/(m/s)²
-    #[derivative(Hash(hash_with = "editoast_common::hash_float::<5,_>"))]
-    pub C: f64,
+    /// Solid friction in N/kg -- N = kg⋅m⋅s⁻²
+    #[derivative(Hash(hash_with = "meter_per_second_squared::hash"))]
+    #[serde(with = "meter_per_second_squared")]
+    pub A: Acceleration,
+    /// Viscosity friction in (N/kg)/(m/s) — N = kg⋅m⋅s−2 => 1/s
```
leovalais:
```suggestion
    /// Viscosity friction in (N/kg)/(m/s) -- N = kg⋅m⋅s⁻² => 1/s
```

editoast/src/core/simulation.rs
```diff
@@ -38,47 +39,55 @@ editoast_common::schemas! {
 pub struct PhysicsConsist {
     pub effort_curves: EffortCurves,
     pub base_power_class: Option<String>,
-    /// Length of the rolling stock in mm
```
leovalais, line 42: Yup. That line alone proves we needed `uom`... Thanks for adding the OpenAPI generation part, mistakes like this are way too easy to make.

editoast/src/core/simulation.rs
leovalais: Up

editoast/src/models/rolling_stock_model.rs
```diff
@@ -62,10 +77,14 @@ pub struct RollingStockModel {
     #[model(json)]
     pub energy_sources: Vec<EnergySource>,
     pub locked: bool,
-    #[schema(required)]
-    pub electrical_power_startup_time: Option<f64>,
-    #[schema(required)]
-    pub raise_pantograph_time: Option<f64>,
+    //#[schema(required)]
```
leovalais: Removing the `schema(required)` makes the field not required in the OpenAPI. But since we always serialize `None` to `null`, we can ensure the field's always there. Doing so is likely to change a `number | null` to `number | null | undefined` somewhere in the RTK file, maybe raising type errors on the TypeScript side.

editoast/src/models/rolling_stock_model.rs
```diff
@@ -96,8 +115,10 @@ impl RollingStockModelChangeset {
         match &self.effort_curves {
             Some(effort_curves) => validate_rolling_stock(
                 effort_curves,
-                self.electrical_power_startup_time.flatten(),
-                self.raise_pantograph_time.flatten(),
+                self.electrical_power_startup_time
+                    .flatten()
+                    .map(|t| second::new(t)),
+                self.raise_pantograph_time.flatten().map(|t| second::new(t)),
```
leovalais:
```suggestion
                    .map(second::new),
                self.raise_pantograph_time.flatten().map(second::new),
```

(maybe)

editoast/src/views/rolling_stock/towed.rs
```diff
@@ -53,14 +55,19 @@ struct TowedRollingStock {
...
+    #[serde(with = "meter_per_second_squared")]
+    comfort_acceleration: Acceleration,
+    #[serde(with = "meter_per_second_squared")]
+    startup_acceleration: Acceleration,
     inertia_coefficient: f64,
     rolling_resistance: RollingResistancePerWeight,
     const_gamma: f64,
-    max_speed: Option<f64>,
+    #[serde(default, with = "meter_per_second::option")]
+    max_speed: Option<Velocity>,
```
leovalais: `#[schema(required)]`

editoast/src/views/timetable/stdcm/request.rs
```diff
@@ -102,15 +104,16 @@ pub(super) struct Request {
     #[serde(default)]
     #[schema(value_type = Option<String>, example = json!(["5%", "2min/100km"]))]
     pub(super) margin: Option<MarginValue>,
-    /// Total mass of the consist in kg
-    #[validate(range(exclusive_min = 0.0))]
-    pub(super) total_mass: Option<f64>,
+    /// Total mass of the consist
+    //#[validate(range(exclusive_min = 0.0))] TODOUOM
```
leovalais: I suppose `Validate` isn't happy with the new types ^^

You could do the `remote = Self` trick to validate just after serializing: https://yossarian.net/til/post/remote-self-derivation-with-serde-for-custom-invariants/

leovalais (commented):
A few comments about details. I tried to make sure the units were consistent with the old definitions, but we'll have to test the application thoroughly to ensure no regression occurs... 😓

Thanks for setting all this up!

editoast/editoast_common/src/units.rs
Tristramg: Lets keep it simple and keep the repetition if it’s fine for you :)

leovalais (approved):
LGTM, thank you!

I'll fully test the application tomorrow to try to spot regressions. But I don't think I'm the most qualified to spot them efficiently. Maybe requesting a review from @OpenRailAssociation/osrd-physics and @OpenRailAssociation/osrd-core would be a good idea?

editoast/src/views/timetable/stdcm/request.rs
Tristramg: Validate somehow expected serializable structs and I didn’t find how to make him happy (I didn’t search much).

The `remote = "Self"` trick is funny, but it changes a bit the behavior: it would be a serialization error, and not explicitly in https://github.com/OpenRailAssociation/osrd/blob/dev/editoast/src/views/timetable/stdcm.rs#L144

Here is a suggestion: https://github.com/OpenRailAssociation/osrd/pull/9928/commits/89a22dfc7a4ea6810e0b2d3c64d5353cb69609df what do you think?

leovalais: Lgtm, resolving. It's fine to keep using `validate` types since we can't get rid of the crate just yet anyways :)

editoast/editoast_schemas/src/rolling_stock.rs
```diff
@@ -53,21 +54,24 @@ pub struct RollingStock {
...
+    pub comfort_acceleration: Acceleration,
     // The constant gamma braking coefficient used when NOT circulating
-    // under ETCS/ERTMS signaling system in m/s^2
-    pub const_gamma: f64,
-    pub inertia_coefficient: f64,
-    /// In kg
-    pub mass: f64,
+    // under ETCS/ERTMS signaling system
+    #[serde(with = "meter_per_second_squared")]
+    pub const_gamma: Acceleration,
```
axrolld: Don't know if that makes a difference but if this value is positive it's a deceleration, not an acceleration. Idk if you want this to be a generic type of unit then acceleration is fine, but if the model is more precise than that it could be good to differenciate acceleration and deceleration.

editoast/editoast_schemas/src/rolling_stock.rs
```diff
@@ -53,21 +54,24 @@ pub struct RollingStock {
...
+    #[serde(with = "meter_per_second_squared")]
+    pub inertia_coefficient: Acceleration,
```
axrolld: The inertia coefficient has no unit, it's a ratio. For example 1.05 -> 5% more mass due to inertia.

editoast/editoast_schemas/src/rolling_stock/rolling_resistance.rs
```diff
@@ -8,38 +9,46 @@ editoast_common::schemas! {
...
+    #[serde(with = "newton")]
+    pub A: Force,
+    /// Viscosity friction in N/(m/s) – N = kg⋅m⋅s⁻²
+    #[derivative(Hash(hash_with = "kilogram_per_second::hash"))]
+    #[serde(with = "kilogram_per_second")]
+    pub B: MassRate,
+    /// Aerodynamic drag in N/(m/s)² – N = kg⋅m⋅s⁻²
+    #[derivative(Hash(hash_with = "kilogram_per_meter::hash"))]
+    #[serde(with = "kilogram_per_meter")]
+    pub C: LinearMassDensity,
```
axrolld: I thought the `-` was a minus and got confused, how about this comment ? Feel free to make the change or not it's just a suggestion.
```suggestion
    /// Solid friction
    #[derivative(Hash(hash_with = "newton::hash"))]
    #[serde(with = "newton")]
    pub A: Force,
    /// Viscosity friction in N/(m/s) = kg/s
    #[derivative(Hash(hash_with = "kilogram_per_second::hash"))]
    #[serde(with = "kilogram_per_second")]
    pub B: MassRate,
    /// Aerodynamic drag in N/(m/s)² = kg/m
    #[derivative(Hash(hash_with = "kilogram_per_meter::hash"))]
    #[serde(with = "kilogram_per_meter")]
    pub C: LinearMassDensity,
```
Though the `MassRate` and `LinearMassDensity` are correct i don't think they help understanding what those parameters are. I don't know the purpose of that part of the code. If it's to summarize the unit under it's simplest form then i agree, if not we might be able to find more understandable units. But if i understand well it's the first situation, right ?

editoast/editoast_schemas/src/rolling_stock/rolling_resistance.rs
```diff
@@ -8,38 +9,46 @@ editoast_common::schemas! {
...
+    #[serde(with = "meter_per_second_squared")]
+    pub A: Acceleration,
+    /// Viscosity friction in (N/kg)/(m/s) – N = kg⋅m⋅s⁻²
+    #[derivative(Hash(hash_with = "hertz::hash"))]
+    #[serde(with = "hertz")]
+    pub B: Frequency,
+    /// Aerodynamic drag per kg in (N/kg)/(m/s)² – N = kg⋅m⋅s⁻²
+    #[derivative(Hash(hash_with = "per_meter::hash"))]
+    #[serde(with = "per_meter")]
+    pub C: LinearNumberDensity,
```
axrolld: Same here

editoast/editoast_schemas/src/rolling_stock/towed_rolling_stock.rs
```diff
@@ -1,22 +1,26 @@
...
+    #[serde(with = "kilogram")]
+    pub mass: Mass,
+    #[serde(with = "meter")]
+    pub length: Length,
+    #[serde(with = "meter_per_second_squared")]
+    pub comfort_acceleration: Acceleration,
+    #[serde(with = "meter_per_second_squared")]
+    pub startup_acceleration: Acceleration,
+    #[serde(with = "meter_per_second_squared")]
+    pub inertia_coefficient: Acceleration,
```
axrolld: Same as rolling stock, no unit here it's a ratio

editoast/editoast_schemas/src/rolling_stock/towed_rolling_stock.rs
```diff
@@ -1,22 +1,26 @@
...
     pub rolling_resistance: RollingResistancePerWeight,
     /// The constant gamma braking coefficient used when NOT circulating
-    /// under ETCS/ERTMS signaling system in m/s^2
-    pub const_gamma: f64,
-    pub max_speed: Option<f64>,
+    /// under ETCS/ERTMS signaling system
+    #[serde(with = "meter_per_second_squared")]
+    pub const_gamma: Acceleration,
+    #[serde(default, with = "kilometer_per_hour::option")]
+    pub max_speed: Option<Velocity>,
```
axrolld: Why is the rolling stock expecting a velocity in m/s and the towed rs expecting km/h ? Shouldn't we stick to m/s everywhere ?

axrolld: Maybe it's not the purpose of this PR though, just make sure the typing is correct here.

editoast/src/models/fixtures.rs
```diff
@@ -227,19 +228,19 @@ pub fn create_towed_rolling_stock() -> TowedRollingStock {
...
-        mass: 50000.0,             // kg
-        length: 30.0,              // m
-        comfort_acceleration: 0.2, // In m/s²
-        startup_acceleration: 0.06,
-        inertia_coefficient: 1.05,
+        mass: kilogram::new(50000.0),
+        length: meter::new(30.0),
+        comfort_acceleration: meter_per_second_squared::new(0.2),
+        startup_acceleration: meter_per_second_squared::new(0.06),
+        inertia_coefficient: meter_per_second_squared::new(1.05),
```
axrolld: no unit

editoast/src/models/fixtures.rs
```diff
@@ -227,19 +228,19 @@ pub fn create_towed_rolling_stock() -> TowedRollingStock {
...
         rolling_resistance: RollingResistancePerWeight {
             rolling_resistance_type: "davis".to_string(),
-            A: 1.0,    // In N
-            B: 0.01,   // In N/(m/s)
-            C: 0.0002, // In N/(m/s)²
+            A: meter_per_second_squared::new(1.0), // In N/kg
+            B: hertz::new(0.01),                   // In N/kg/(m/s)
+            C: per_meter::new(0.0002),             // In N/kg/(m/s)²
```
axrolld: According to the values i guess the units commented are wrong and were actually respectively daN/T, (daN/T)/(km/h), and (daN/T)/(km/h)².

If you want to keep m/s², Hz, and m^-1 you need to convert the units:
1.0 daN/T = 1e-2 N/kg = 1e-2 m/s²
0.01 (daN/T)/(km/h) = 0.01 * 1e-2 * 3.6 (N/kg)/(m/s) = 3.6e-4 Hz
0.0002 (daN/T)/(km/h)² = 0.0002 * 1e-2 * 3.6² (N/kg)/(m/s)² = 2.6e-5 m^-1

editoast/src/models/fixtures.rs
```diff
@@ -250,28 +251,28 @@ pub fn create_simple_rolling_stock() -> RollingStock {
         loading_gauge: LoadingGaugeType::G1,
         supported_signaling_systems: RollingStockSupportedSignalingSystems(vec![]),
         base_power_class: None,
-        comfort_acceleration: 0.1, // In m/s²
-        inertia_coefficient: 1.10,
-        startup_acceleration: 0.04, // In m/s²
-        startup_time: 1.0,
+        comfort_acceleration: meter_per_second_squared::new(0.1),
+        inertia_coefficient: meter_per_second_squared::new(1.10),
```
axrolld: no unit

editoast/src/models/rolling_stock_model.rs
```diff
@@ -43,16 +46,32 @@ pub struct RollingStockModel {
...
+    pub startup_acceleration: Acceleration,
+    #[serde(with = "meter_per_second_squared")]
+    #[model(uom_unit = "meter_per_second_squared")]
+    pub comfort_acceleration: Acceleration,
+    #[serde(with = "meter_per_second_squared")]
+    #[model(uom_unit = "meter_per_second_squared")]
+    pub const_gamma: Acceleration,
+    #[serde(with = "meter_per_second_squared")]
+    #[model(uom_unit = "meter_per_second_squared")]
+    pub inertia_coefficient: Acceleration,
```
axrolld: no unit

editoast/src/models/towed_rolling_stock.rs
```diff
@@ -21,18 +23,30 @@ pub struct TowedRollingStockModel {
...
+    #[serde(with = "kilogram")]
+    #[model(uom_unit = "kilogram")]
+    pub mass: Mass,
+    #[serde(with = "meter")]
+    #[model(uom_unit = "meter")]
+    pub length: Length,
+    #[schema(required)]
+    #[serde(default, with = "kilometer_per_hour::option")]
+    #[model(uom_unit = "kilometer_per_hour::option")]
+    pub max_speed: Option<Velocity>,
```
axrolld: same comment as above. It seems to be km/h here so the typing is correct, but maybe it would be goo to harmonize this in m/s in another PR

editoast/src/models/towed_rolling_stock.rs
```diff
@@ -21,18 +23,30 @@ pub struct TowedRollingStockModel {
...
+    #[model(uom_unit = "meter_per_second_squared")]
+    #[serde(with = "meter_per_second_squared")]
+    pub comfort_acceleration: Acceleration,
+    #[model(uom_unit = "meter_per_second_squared")]
+    #[serde(with = "meter_per_second_squared")]
+    pub startup_acceleration: Acceleration,
+    #[model(uom_unit = "meter_per_second_squared")]
+    #[serde(with = "meter_per_second_squared")]
+    pub inertia_coefficient: Acceleration,
```
axrolld: no unit

editoast/src/views/rolling_stock/form.rs
```diff
@@ -18,21 +19,30 @@ use crate::models::Changeset;
...
+    #[serde(with = "second")]
+    pub startup_time: Time,
+    #[serde(with = "meter_per_second_squared")]
+    pub startup_acceleration: Acceleration,
+    #[serde(with = "meter_per_second_squared")]
+    pub comfort_acceleration: Acceleration,
+    #[serde(with = "meter_per_second_squared")]
+    pub const_gamma: Acceleration,
+    #[serde(with = "meter_per_second_squared")]
+    pub inertia_coefficient: Acceleration,
```
axrolld: no unit

editoast/src/views/rolling_stock/light.rs
```diff
@@ -214,14 +216,22 @@ struct LightRollingStock {
...
+    #[serde(with = "second")]
+    startup_time: Time,
+    #[serde(with = "meter_per_second_squared")]
+    startup_acceleration: Acceleration,
+    #[serde(with = "meter_per_second_squared")]
+    comfort_acceleration: Acceleration,
+    #[serde(with = "meter_per_second_squared")]
+    const_gamma: Acceleration,
+    #[serde(with = "meter_per_second_squared")]
+    inertia_coefficient: Acceleration,
```
axrolld: no unit

editoast/src/views/rolling_stock/towed.rs
```diff
@@ -53,14 +55,22 @@ struct TowedRollingStock {
...
+    #[serde(with = "kilogram")]
+    mass: Mass,
+    #[serde(with = "meter")]
+    length: Length,
+    #[serde(with = "meter_per_second_squared")]
+    comfort_acceleration: Acceleration,
+    #[serde(with = "meter_per_second_squared")]
+    startup_acceleration: Acceleration,
+    #[serde(with = "meter_per_second_squared")]
+    inertia_coefficient: Acceleration,
```
axrolld: no unit

editoast/src/views/timetable/stdcm.rs
```diff
@@ -508,21 +509,21 @@ mod tests {
...
     fn simulation_with_towed_rolling_stock_parameters() {
         let mut rolling_stock = create_simple_rolling_stock();
-        rolling_stock.mass = 100000.0;
-        rolling_stock.length = 20.0;
-        rolling_stock.inertia_coefficient = 1.10; // m/s²
-        rolling_stock.comfort_acceleration = 0.1;
-        rolling_stock.startup_acceleration = 0.04; // m/s²
+        rolling_stock.mass = kilogram::new(100000.0);
+        rolling_stock.length = meter::new(20.0);
+        rolling_stock.inertia_coefficient = meter_per_second_squared::new(1.10);
```
axrolld: no unit

editoast/src/views/timetable/stdcm.rs
```diff
@@ -534,36 +535,39 @@ mod tests {
...
         let physics_consist: PhysicsConsist = simulation_parameters.into();
 
-        assert_eq!(physics_consist.mass, total_mass as u64);
+        assert_eq!(physics_consist.mass, total_mass);
 
-        assert_eq!(physics_consist.inertia_coefficient, 1.075_f64);
+        assert_eq!(
+            physics_consist.inertia_coefficient,
+            meter_per_second_squared::new(1.075)
+        );
```
axrolld: no unit

editoast/src/core/simulation.rs
```diff
@@ -38,47 +39,57 @@ editoast_common::schemas! {
...
-    pub inertia_coefficient: f64,
-    /// Mass of the rolling stock in kg
-    pub mass: u64,
+    /// under ETCS/ERTMS signaling system
+    #[derivative(Hash(hash_with = "meter_per_second_squared::hash"))]
+    #[serde(with = "meter_per_second_squared")]
+    pub const_gamma: Acceleration,
+    #[derivative(Hash(hash_with = "meter_per_second_squared::hash"))]
+    #[serde(with = "meter_per_second_squared")]
+    pub inertia_coefficient: Acceleration,
```
axrolld: no unit

editoast/src/core/simulation.rs
```diff
@@ -124,35 +129,33 @@ impl PhysicsConsistParameters {
...
-                )
+                self.traction_engine
+                    .comfort_acceleration
+                    .min(towed_rolling_stock.comfort_acceleration)
             })
             .unwrap_or(self.traction_engine.comfort_acceleration)
     }
 
-    pub fn compute_inertia_coefficient(&self) -> f64 {
+    pub fn compute_inertia_coefficient(&self) -> Acceleration {
```
axrolld: no unit

editoast/src/core/simulation.rs
```diff
@@ -533,96 +531,129 @@ mod tests {
...
+            physics_consist.compute_inertia_coefficient(),
+            meter_per_second_squared::new(1.065)
+        );
 
         physics_consist.towed_rolling_stock = None;
-        assert_eq!(physics_consist.compute_inertia_coefficient(), 1.10,);
+        assert_eq!(
+            physics_consist.compute_inertia_coefficient(),
+            meter_per_second_squared::new(1.10,)
+        );
```
axrolld: no unit

editoast/src/models/fixtures.rs
```diff
@@ -250,28 +251,28 @@ pub fn create_simple_rolling_stock() -> RollingStock {
...
         power_restrictions: HashMap::new(),
         railjson_version: "12".to_string(),
         rolling_resistance: RollingResistance {
             rolling_resistance_type: "davis".to_string(),
-            A: 1.0,    // In N
-            B: 0.01,   // In N/(m/s)
-            C: 0.0005, // In N/(m/s)²
+            A: newton::new(1.0),                // In N
+            B: kilogram_per_second::new(0.01),  // In N/(m/s)
+            C: kilogram_per_meter::new(0.0005), // In N/(m/s)²
```
axrolld: Same as above but with 0.0005 instead of 0.0002

editoast/src/views/timetable/stdcm.rs
```diff
@@ -508,21 +509,21 @@ mod tests {
...
+        rolling_stock.comfort_acceleration = meter_per_second_squared::new(0.1);
+        rolling_stock.startup_acceleration = meter_per_second_squared::new(0.04);
         rolling_stock.rolling_resistance = RollingResistance {
             rolling_resistance_type: "davis".to_string(),
-            A: 1.0,
-            B: 0.01,
-            C: 0.0005,
+            A: newton::new(1.0),
+            B: kilogram_per_second::new(0.01),
+            C: kilogram_per_meter::new(0.0005),
```
axrolld: Same convertion as one of my previous comments

editoast/src/views/timetable/stdcm.rs
```diff
@@ -534,36 +535,39 @@ mod tests {
...
         assert_eq!(
             physics_consist.rolling_resistance,
             RollingResistance {
                 rolling_resistance_type: "davis".to_string(),
-                A: 100001.0,
-                B: 1000.01,
-                C: 20.0005
+                A: newton::new(100001.0),
+                B: kilogram_per_second::new(1000.01),
+                C: kilogram_per_meter::new(20.0005),
```
axrolld: Here the values are already in N, N/(m/s), and N/(m/s)² already, so no convertion needs to be done 👍

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -2799,27 +2802,36 @@ export type RollingStockMetadata = {
...
 export type LightRollingStock = {
   base_power_class?: string | null;
+  /** Acceleration in m·s⁻² */
   comfort_acceleration: number;
+  /** Acceleration in m·s⁻² */
   const_gamma: number;
   effort_curves: LightEffortCurves;
   energy_sources: EnergySource[];
   id: number;
+  /** Acceleration in m·s⁻² */
```
axrolld: no unit

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -3020,55 +3034,75 @@ export type EffortCurves = {
...
   comfort_acceleration: number;
+  /** Acceleration in m·s⁻² */
   const_gamma: number;
   effort_curves: EffortCurves;
+  /** Duration in s */
   electrical_power_startup_time: number | null;
   energy_sources: EnergySource[];
   id: number;
+  /** Acceleration in m·s⁻² */
```
axrolld: no unit

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -3020,55 +3034,75 @@ export type EffortCurves = {
...
   comfort_acceleration: number;
+  /** Acceleration in m·s⁻² */
   const_gamma: number;
   effort_curves: EffortCurves;
-  /** The time the train takes before actually using electrical power (in seconds). Is null if the train is not electric. */
+  /** The time the train takes before actually using electrical power (in seconds). Is null if the train is not electric.
+    Duration in s */
   electrical_power_startup_time?: number | null;
   energy_sources?: EnergySource[];
+  /** Acceleration in m·s⁻² */
```
axrolld: no unit

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -3476,40 +3510,57 @@ export type TrainScheduleResult = TrainScheduleBase & {
...
   C: number;
   type: string;
 };
 export type TowedRollingStock = {
+  /** Acceleration in m·s⁻² */
   comfort_acceleration: number;
+  /** Acceleration in m·s⁻² */
   const_gamma: number;
   id: number;
+  /** Acceleration in m·s⁻² */
```
axrolld: no unit

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -3476,40 +3510,57 @@ export type TrainScheduleResult = TrainScheduleBase & {
...
   rolling_resistance: RollingResistancePerWeight;
+  /** Acceleration in m·s⁻² */
   startup_acceleration: number;
 };
 export type TowedRollingStockForm = {
+  /** Acceleration in m·s⁻² */
   comfort_acceleration: number;
+  /** Acceleration in m·s⁻² */
   const_gamma: number;
+  /** Acceleration in m·s⁻² */
```
axrolld: no unit

axrolld (changes requested):
Thanks for this huge piece of work ! It'll make the code much clearer and hopefully stop wasting time in converting units... It's always a pain.

I spotted a few convertion mistakes that need to be fixed, and the main issue in the inertia coefficient considered as an acceleration everywhere while it's just a ratio. I tried to spot all the occurrences but i might have missed some. Make sure this coefficient remains without unit everywhere.

editoast/editoast_schemas/src/rolling_stock/rolling_resistance.rs
Tristramg: The main goal is to guarantee unit homogeneity (it is now impossible to add an acceleration with a speed)

Documenting is a nice bonus. I made aliases, it should be better now

editoast/editoast_schemas/src/rolling_stock/towed_rolling_stock.rs
Tristramg: It’s still being investigated, probably the comments are wrong
https://github.com/OpenRailAssociation/osrd/blob/dev/editoast/src/models/towed_rolling_stock.rs#L28

editoast/editoast_schemas/src/rolling_stock.rs
Tristramg: So we has a comment thas was wrong: https://github.com/OpenRailAssociation/osrd/blob/dev/editoast/src/views/timetable/stdcm.rs#L513

I changed it to ratio in https://github.com/OpenRailAssociation/osrd/pull/9928/commits/65417f460cc2a3016f5b7828d023413d786c33a6

editoast/editoast_schemas/src/rolling_stock.rs
Tristramg: I aliased it as `Deceleration` in https://github.com/OpenRailAssociation/osrd/pull/9928/commits/e35922cadff84a1e59efb6b1b31057a756832e0c
It should make it clearer

Khoyo:
> * before, we consider that a `length` is always in meter through the entire system (but not typed, hence we had bugs, but that's not the point here)
> 
> * with `uom`, we don't care how a `length` is stored

I'm realizing now that we might not preserve integer units :/ I'm a bit worried, we went from meters/float everywhere to using millimeters/integers (aka. fixed point) in order to avoid some nasty bugs, are we going back to floats? Or is uom doing something *smart*?

Tristramg:
In this pull request, the representation is still f64 as it was before. Nowhere we switch from integers to floats.

The neat part of uom, is that we will be able to switch to rational values for the storage.
And when we will be ready to change the external part, it will be explicit, with correct generated comments

Khoyo:
You're right, I confused some of core API changes for TSv2 with editoast internal repr (I thought the switch was done already). Carry on 😄

Tristramg:
So, actually, we were serializing milliseconds as floats, not as seconds. I’m not sure it changes much, but let’s keep the existing behaviour.
I added the possibility to serialize as integers in https://github.com/OpenRailAssociation/osrd/pull/9928/commits/9fc176da1fc16229969e4cd2009d5cc7e50f5da3 (and found the unit error that was haunting me)

@leovalais

leovalais (approved):
LGTM, must have been fun debugging that 😓

As you suggested, there's probably some work on the generated units API (including traits, generics or what else). I'd prefer to keep that for a follow up PR though. This one has been opened for quite a while and rebasing this much code over and over isn't fun.

editoast/src/models/fixtures.rs
Tristramg: I added a comment for now because all our test are based on those values and putting more realistic values breaks a lot of tests

editoast/editoast_schemas/src/rolling_stock.rs
axrolld: Good catch, thanks ! I guess someone else misunderstood this earlier

editoast/src/models/fixtures.rs
axrolld: Fair

axrolld (approved):
LGTM, great job 🚀

editoast/editoast_common/src/units.rs
```diff
@@ -0,0 +1,214 @@
+//! Module to allow the use of serde with uom quantities
+//!
+//! The serde feature of uom doesn’t allow to specify in which unit the value will be serialized.
+//!
+//! Two helpers are provided for convenience:
+//! * `unit::new` (e.g. `meter::new(32)`) to build a new quantity form a f64 in the given unit
```
woshilapin:
```suggestion
//! * `unit::new` (e.g. `meter::new(32)`) to build a new quantity from a f64 in the given unit
```

editoast/editoast_derive/src/annotate_units.rs
```diff
@@ -0,0 +1,54 @@
...
+    if let syn::Data::Struct(s) = &mut input.data {
+        for f in s.fields.iter_mut() {
+            for attr in f.attrs.clone() {
+                if attr.path().is_ident("serde") {
+                    let _ = attr.parse_nested_meta(|meta| {
+                        if meta.path.is_ident("with") {
+                            let value = meta.value()?;
+                            let s: LitStr = value.parse()?;
+                            if let Some(abbreviation) = get_abbreviation(s.value()) {
+                                if s.value().ends_with("::option") {
```
woshilapin: I see a `.ends_with("::option")` here and `.replace("::option")` above. I wonder if we either should:

1. transform `s.value()` into its own type `struct UnitType { unit: String, optional: bool }` and we then could implement `get_abbreviation` on it without the need to do the `replace`, and the current `if` test without the need for testing `ends_with` (note: `unit` field could even be a `enum` instead of `String` with a `Display` implementation, replacing the need for `get_abbreviation`)
2. at the very least if you feel 1. is too much, how about `const OPTIONAL_SUFFIX: &str = "::option";` somewhere

editoast/editoast_derive/src/lib.rs
```diff
@@ -217,6 +218,7 @@ pub fn search_config_store(input: proc_macro::TokenStream) -> proc_macro::TokenS
 /// * `#[model(to_string)]`: calls `to_string()` before writing the field to the database and calls `String::from` after reading (diesel column type: String)
 /// * `#[model(to_enum)]`: is converted as `u8` before writing the field to the database and calls `FromRepr::from_repr` after reading (diesel column type: TinyInt)
 /// * `#[model(remote = "T")]`: calls `Into::<T>::into` before writing the field to the database and calls `T::from` after reading (diesel column type: T)
+/// * `#[model(uom_unit = "T")]`: the value is the path to an unit defined in editoast_common, e.g. `"editoast_common::units::meter"`
```
woshilapin: nitpick: T feels like you would expect a **type** but if I understand, we expect more a path here

```suggestion
/// * `#[model(uom_unit = "path::to::unit")]`: the value is the path to a unit defined in editoast_common, e.g. `"editoast_common::units::meter"`
```

editoast/editoast_schemas/src/rolling_stock.rs
```diff
@@ -60,22 +61,25 @@ pub struct RollingStock {
...
-    /// In m
-    pub length: f64,
-    /// In m/s
-    pub max_speed: f64,
-    pub startup_time: f64,
-    /// In m/s²
-    pub startup_acceleration: f64,
-    /// In m/s²
-    pub comfort_acceleration: f64,
+    #[serde(with = "meter")]
```
woshilapin: `"meter"` seems like magic... I wonder if we make more explicit that `meter` is actually a module's path, it would help reduce the magic effect? (that's a general problem of `*` imports, since a simple 'search' in the file for occurrence of `meter` won't return any result, and probably `rust-analyzer` won't help either to get to the definition, it's hard to understand what is going on).

```suggestion
    #[serde(with = "editoast_common::units::meter")]
```

I understand it will make all the macro annotation more noisy, but to me, it would be better. What do you think?

editoast/src/views/timetable/stdcm/request.rs
```diff
@@ -289,3 +291,35 @@ impl Request {
         Ok(Some(towed_rolling_stock))
     }
 }
+
+impl Validate for Request {
```
woshilapin: We're trying to get rid of `validator`, and trying more a [parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) strategy. So we are now trying to implement `Deserialize` with the validation inside. See [here](https://github.com/OpenRailAssociation/osrd/blob/4e1f4613eaf97202920801f67911e27a5dfece8e/editoast/editoast_schemas/src/rolling_stock.rs#L99-L127) (with the tiny line [here](https://github.com/OpenRailAssociation/osrd/blob/4e1f4613eaf97202920801f67911e27a5dfece8e/editoast/editoast_schemas/src/rolling_stock.rs#L57) too) for an example on how to do it.

woshilapin (approved):
Looking good. A few more comments.

editoast/editoast_derive/src/annotate_units.rs
Tristramg: Would https://github.com/OpenRailAssociation/osrd/pull/9928/commits/d9536a1a667f3d4c36b928046ecac7b401aec277 fit?

editoast/editoast_schemas/src/rolling_stock.rs
Tristramg: https://github.com/OpenRailAssociation/osrd/pull/9928/commits/791877dbaa81e61182489fbbb1403374c129bc2e here you go

editoast/src/views/timetable/stdcm/request.rs
Tristramg: https://github.com/OpenRailAssociation/osrd/pull/9928/commits/30394fb90c4c5fd49f29cdee66b760c9d6535ec6 done here

editoast/editoast_derive/src/annotate_units.rs
```diff
@@ -0,0 +1,82 @@
...
+        } else {
+            Self {
+                unit: paths
+                    .last()
+                    .expect("String::split to return at least an empty string")
+                    .to_string(),
+                optional: false,
+            }
+        }
+    }
```
woshilapin: A suggestion to use functions of `str` to make the code more concise (see [Rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=55a1f62250405a705c4ba2832d369372)) by using `strip_suffix` and `rsplit`.

editoast/editoast_derive/src/annotate_units.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+fn get_abbreviation(value: &UnitType) -> Option<&'static str> {
```
woshilapin: nitpick: `get_abbreviation` don't need the whole `UnitType` to do it's job, it doesn't even need to know about `UnitType` at all, it could just take a `String` as input to do it's job.

editoast/editoast_derive/src/annotate_units.rs
woshilapin: Yes, I've made some small improvements comments but I do prefer the general structure.

editoast/src/views/timetable/stdcm/request.rs
woshilapin: Thank you very much for this side quest ⚔️

woshilapin (approved):
Thank you for the improvements.

Issue #9707, opened 2024-11-14
lmr: front: catch missing fields and adapt warning message


axrolld:
# Description

Some fields can be left empty and the simulation button does nothing. We need to catch empty fields and warn the user.

# AC

- [ ] If one of the following fileds is not filled in, the warning box pops up above the button `Obtenir la simulation` when the user clicks it only (not before). The button `Obtenir la simulation` slides down, and if it gets out of the screen, there should be an automatic scroll to see the warning message entirely and the simulation button.
    - traction engine
    - consist total weight
    - consist total length
    - consist max speed
    - CI of origin/destination
    - departure or arrival time

![Image](https://github.com/user-attachments/assets/5d8912d8-9c32-4472-b7a1-633362335c43)

# Test e2e

**User scenario :** https://github.com/osrd-project/osrd-tests/issues/106
**End to end tests strategy:**
- Create a new `.spec.ts` test file
- Implement a new modal for web elements and methods
- Follow the user scenario to define the test steps

Seen with @SharglutDev, invalid fields will be handled in this ticket https://github.com/OpenRailAssociation/osrd/issues/11278.

thibautsailly:
Mockup for fields with errors: `Sketch > Last Minute Request > Request form > #9707 - Warning message display`

![Image](https://github.com/user-attachments/assets/ad1db5e8-20b8-4c98-99a6-67b9c48d77ad)

thibautsailly:
I uploaded the mockup in the wrong issue, following a mislabelling in Sketch. Sorry for the misunderstanding and for the potential wasted time, I'll double check before posting from now on!
Deleting the uploaded mockup to prevent more confusion.

axrolld:
There's this bug currently : it's possible to run a simulation with plenty of wrong values. I guess closing this issue will solve this, isn't it @Caracol3 ?
![Image](https://github.com/user-attachments/assets/8fb4d559-2968-4545-a008-6bf6fb16c6a9)

Caracol3:
It's not clearly stated in the issue, but I think it would be logical to prevent the calculation when there are invalid values, rather than just displaying an informative message.

axrolld:
Hm, you're right, the issue is actually just mentioning empty fields. What do you think about blocking invalid ones too? Should I open a new issue?

PR #9931, opened 2024-12-03
editoast, front: forbid zero-length paths in pathfinding endpoint

### LINKED ISSUE #9444: 0-length paths aren't explicitely forbidden but break most endpoints
### What happened?

Nothing in the API or UI forbids the user from creating a path with overlapping origin and destination, resulting in a 0-length path.

Surprisingly, the pathfinding endpoint gives a valid result. But any subsequent call (path_properties or simulation) result in a 500.

I don't think it's a significant issue. But if we want to handle that case properly, we could add an error case in the pathfinding endpoint. 

I *expected* it to fail right away at the pathfinding endpoint.

I think we *should* either fail with a 4xx at the pathfinding endpoint, or handle it all the way (but let's not do that).

Rocket tool, pick any OP, then ask for a path like `MES MES` on small_infra

Local

Firefox

59775313d99fe067e5e950da181d41c5bb7d54e6

Sh099078:
Fixes https://github.com/OpenRailAssociation/osrd/issues/9444
- ~~Editoast: immediately return a `422 Unprocessable Content` error at deserialization when the endpoint `pathfinding/blocks` is called with the same origin and destination.~~
- Editoast: `pathfinding/blocks`: fail in post-processing with a new `PathfindingInputError::ZeroLengthPath` variant when core returns zero-length paths.
- Front: add a new type of stdcm error type when trying to make an STDCM request with the same origin and destination, update the stdcm inputs validation method.

### DIFF
--- a/editoast/src/core/pathfinding.rs
+++ b/editoast/src/core/pathfinding.rs
@@ -133,6 +133,7 @@ pub enum PathfindingInputError {
     RollingStockNotFound {
         rolling_stock_name: String,
     },
+    ZeroLengthPath,
 }
 
 // Enum for not-found results and incompatible constraints
--- a/editoast/src/views/path/pathfinding.rs
+++ b/editoast/src/views/path/pathfinding.rs
@@ -86,7 +86,12 @@ pub enum PathfindingResult {
 impl From<PathfindingCoreResult> for PathfindingResult {
     fn from(core_result: PathfindingCoreResult) -> Self {
         match core_result {
-            PathfindingCoreResult::Success(success) => PathfindingResult::Success(success),
+            PathfindingCoreResult::Success(success) => match success.length {
+                0 => PathfindingResult::Failure(PathfindingFailure::PathfindingInputError(
+                    PathfindingInputError::ZeroLengthPath,
+                )),
+                _ => PathfindingResult::Success(success),
+            },
             PathfindingCoreResult::NotFoundInBlocks {
                 track_section_ranges,
                 length,
@@ -457,6 +462,53 @@ pub mod tests {
     use crate::views::test_app::TestAppBuilder;
 
     #[rstest]
+    async fn pathfinding_fails_when_core_responds_with_zero_length_path() {
+        let db_pool = DbConnectionPoolV2::for_tests();
+        let mut core = MockingClient::new();
+        core.stub("/v2/pathfinding/blocks")
+            .method(reqwest::Method::POST)
+            .response(StatusCode::OK)
+            .json(json!({
+                "blocks":[],
+                "routes": [],
+                "track_section_ranges": [],
+                "path_item_positions": [],
+                "length": 0,
+                "status": "success"
+            }))
+            .finish();
+        let app = TestAppBuilder::new()
+            .db_pool(db_pool.clone())
+            .core_client(core.into())
+            .build();
+        let small_infra = create_small_infra(&mut db_pool.get_ok()).await;
+
+        let request = app
+            .post(format!("/infra/{}/pathfinding/blocks", small_infra.id).as_str())
+            .json(&json!({
+                "path_items":[
+                {"trigram":"WS","secondary_code":"BV"},
+                {"trigram":"WS","secondary_code":"BV"}
+            ],
+                "rolling_stock_is_thermal":true,
+                "rolling_stock_loading_gauge":"G1",
+                "rolling_stock_supported_electrifications":[],
+                "rolling_stock_supported_signaling_systems":["BAL","BAPR"],
+                "rolling_stock_maximum_speed":22.00,
+                "rolling_stock_length":26.00
+            }));
+
+        let pathfinding_result: PathfindingResult =
+            app.fetch(request).assert_status(StatusCode::OK).json_into();
+        assert_eq!(
+            pathfinding_result,
+            PathfindingResult::Failure(PathfindingFailure::PathfindingInputError(
+                PathfindingInputError::ZeroLengthPath,
+            ))
+        );
+    }
+
+    #[rstest]
     async fn pathfinding_with_invalid_path_items_returns_invalid_path_items() {
         let app = TestAppBuilder::default_app();
         let db_pool = app.db_pool();
@@ -560,7 +612,7 @@ pub mod tests {
                 "routes": [],
                 "track_section_ranges": [],
                 "path_item_positions": [],
-                "length": 0,
+                "length": 1,
                 "status": "success"
             }))
             .finish();
@@ -593,7 +645,7 @@ pub mod tests {
                 blocks: vec![],
                 routes: vec![],
                 track_section_ranges: vec![],
-                length: 0,
+                length: 1,
                 path_item_positions: vec![]
             })
         );
--- a/front/public/locales/en/stdcm.json
+++ b/front/public/locales/en/stdcm.json
@@ -114,7 +114,8 @@
     "noResults": "No path found",
     "noScheduledPoint": "The calculation requires either the origin or destination schedule.",
     "pathfindingFailed": "No path have been found for these waypoints.",
-    "requestFailed": "Calculation error for last minute train path"
+    "requestFailed": "Calculation error for last minute train path",
+    "zeroLengthPath": "The path cannot have a null length."
   },
   "stdcmResults": "Results",
   "stdcmSimulationReport": "Path simulation report",
--- a/front/src/applications/stdcm/hooks/useStaticPathfinding.ts
+++ b/front/src/applications/stdcm/hooks/useStaticPathfinding.ts
@@ -58,6 +58,16 @@ const useStaticPathfinding = (infra?: InfraWithState) => {
         return;
       }
 
+      // Don't run the pathfinding if the origin and destination are the same:
+      const origin = pathSteps.at(0)!;
+      const destination = pathSteps.at(-1)!;
+      if (
+        origin.location!.uic === destination.location!.uic &&
+        origin.location!.secondary_code === destination.location!.secondary_code
+      ) {
+        return;
+      }
+
       const payload = getPathfindingQuery({
         infraId: infra.id,
         rollingStock,
--- a/front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
+++ b/front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
@@ -23,6 +23,13 @@ const checkStdcmConfigErrors = (
     throw new Error('Last step can not be a via');
   }
 
+  if (
+    origin.location!.uic === destination.location!.uic &&
+    origin.location!.secondary_code === destination.location!.secondary_code
+  ) {
+    return { errorType: StdcmConfigErrorTypes.ZERO_LENGTH_PATH };
+  }
+
   if (pathfindingStateError) {
     return { errorType: StdcmConfigErrorTypes.PATHFINDING_FAILED };
   }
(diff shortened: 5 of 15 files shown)


editoast/src/views/path/pathfinding.rs
```diff
@@ -73,6 +73,45 @@ struct PathfindingInput {
...
+        let Internal {
+            rolling_stock_loading_gauge,
+            rolling_stock_is_thermal,
+            rolling_stock_supported_electrifications,
+            rolling_stock_supported_signaling_systems,
+            path_items,
+            rolling_stock_maximum_speed,
+            rolling_stock_length,
+        } = Internal::deserialize(deserializer)?;
+        if path_items.first() == path_items.last() {
```
emersion: This is not enough to figure out if the path length will be zero. For instance, a UIC+ch could refer to the same OP as a trigram+ch.

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
```diff
@@ -25,6 +25,10 @@ const checkStdcmConfigErrors = (
     return { errorType: StdcmConfigErrorTypes.PATHFINDING_FAILED };
   }
 
+  if (origin.name === destination.name && origin.ch === destination.ch) {
```
emersion: Would be more reliable to check the `uic` here instead of `name`: `name` is a human-readable string which may not be unique.

editoast/src/views/path/pathfinding.rs
Sh099078: I am missing some knowledge about what a trigram and an UIC represent and how to correlate them. Would you mind to elaborate or hint at some documentation on them ?

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
Sh099078: Thanks for the speedy review !

Fixed in https://github.com/OpenRailAssociation/osrd/commit/9cbbbaa531c152b60aca995288f3e8d5be94fe7d

editoast/src/views/path/pathfinding.rs
emersion: Sure. There are multiple ways to refer to an operational point in a path: https://github.com/OpenRailAssociation/osrd/blob/9dedb376fde0341f4a6d6f94090dc1044c90d523/editoast/editoast_schemas/src/train_schedule/path_item.rs#L64

One way is via a _UIC code_. The UIC code is a worldwide-unique number which identifies a location. For instance, "87686006" is "Paris-Gare-de-Lyon". In France, the UIC is not enough to precisely identify an operational point: there are multiple operational points in "Paris-Gare-de-Lyon". So a _secondary code_ is added to pin down a specific one (often also called "ch" for "code chantier"). For instance, "BV" often refers to the place where the platforms are in a station ("bâtiment voyageur"). Side note, the UIC is made up of two parts, a country prefix ("87" for France) and the rest is called _CI_ ("code immuable", local French code, e.g. "686006").

Another way is via a _trigram_. It's a 3-letter code. It's essentially another way to write a CI, perhaps easier to remember for humans. For instance, "PLY" is the trigram for "Paris-Gare-de-Lyon".

Yet another way is via an _operational point ID_. It's just an opaque string identifier (e.g. "1e083f8b-c7db-48f4-bb12-ac3bb21d8963").

So, both of these refer to the same operational point:

```rust
OperationalPointDescription {
    trigram: "PLY",
    secondary_code: "BV",
}
OperationalPointUic {
    uic: 686006,
    secondary_code: "BV",
}
```

In other words, if these are given as input to the pathfinding algorithm, it'll freak out the same way it would if given two exact same path item locations.

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
emersion: Looks good! Unfortunately we still have to write these `'uic' in …` checks, hopefully this can go away once Clara's refactorings are merged.

editoast/src/views/path/pathfinding.rs
Sh099078: Thank you for the explanation ! I added a new error variant `ZeroLengthPath` that transforms `core` success responses with zero-length paths into errors instead of checking whether the origin and destination are the same location before executing the pathfinding.

editoast/src/views/path/pathfinding.rs
```diff
@@ -448,6 +453,53 @@ pub mod tests {
...
+            .core_client(core.into())
+            .build();
+        let small_infra = create_small_infra(&mut db_pool.get_ok()).await;
+
+        let request = app
+            .post(format!("/infra/{}/pathfinding/blocks", small_infra.id).as_str())
+            .json(&json!({
+                "path_items":[
+                {"trigram":"WS","secondary_code":"BV"},
+                {"trigram":"SWS","secondary_code":"BV"}
```
woshilapin: Hmm, as a reviewer of the code, I was expecting 2 identical path items, but I see 2 different trigrams and I'm left wondering why do we have a zero-length path. Isn't it possible to make the test even more obvious with the following?

```suggestion
                {"trigram":"WS","secondary_code":"BV"},
                {"trigram":"WS","secondary_code":"BV"}
```

front/public/locales/en/stdcm.json
```diff
@@ -91,7 +91,8 @@
     "noResults": "No path found",
     "noScheduledPoint": "The calculation requires either the origin or destination schedule.",
     "pathfindingFailed": "No path have been found for these waypoints.",
-    "requestFailed": "Calculation error for last minute train path"
+    "requestFailed": "Calculation error for last minute train path",
+    "sameOriginAndDestination": "The origin and destination cannot be the same."
```
woshilapin: nitpick: avoiding negation makes sometimes a clearer sentence?

```suggestion
    "sameOriginAndDestination": "The origin and destination must be different."
```

woshilapin (approved):
Nice fix. Thank you.

editoast/src/views/path/pathfinding.rs
emersion: I think part of the point here is to ensure that the endpoint correctly returns the error even if the trigrams don't match?

front/public/locales/en/stdcm.json
Sh099078: Fixed in https://github.com/OpenRailAssociation/osrd/pull/9931/commits/9719af2565978331c107cd86b909b6b3c5aab73b

editoast/src/views/path/pathfinding.rs
Sh099078: Fixed in https://github.com/OpenRailAssociation/osrd/pull/9931/commits/ed12a878549fa1b8e8199791aeb95defd7a9ba87

editoast/src/core/pathfinding.rs
```diff
@@ -133,6 +133,7 @@ pub enum PathfindingInputError {
     RollingStockNotFound {
         rolling_stock_name: String,
     },
+    ZeroLengthPath,
```
emersion, line 136: Each pathfinding error needs a translation in `front/public/locales/$lang/errors.json`.

emersion (commented):
Apart from this, looks good!

SharglutDev:
> [!NOTE]
> This will conflict with https://github.com/OpenRailAssociation/osrd/pull/9924

flomonster:
Should we authorize pathfinding that has a total length higher than `0m` but between two steps has a length of `0m`?

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
```diff
@@ -30,6 +30,17 @@ const checkStdcmConfigErrors = (
...
     origin.arrivalType === ArrivalTimeTypes.RESPECT_DESTINATION_SCHEDULE;
 
+  if (
+    'uic' in origin &&
+    'uic' in destination &&
+    'ch' in origin &&
+    'ch' in destination &&
+    origin.uic === destination.uic &&
+    origin.ch === destination.ch
+  ) {
```
SharglutDev: We don't use `ch` in the code (too much sncf) but `secondary_code` (editoast type).
Both `uic` and `secondary_code` are defined if `location` is defined in `StdcmPathStep` type.
As we checked l.13 that we don't have any path step without location, we can assert that we will have one at this point.
```suggestion
 if (
    origin.location!.uic === destination.location!.uic &&
    origin.location!.secondary_code === destination.location!.secondary_code
  ) {
```

Sh099078: Fixed in https://github.com/OpenRailAssociation/osrd/commit/df926d1310b72c1d79879d4843d06ee2bf606ecb.

I also put the origin / destination check before the pathfinding response check as a bit of a hotfix, otherwise it would return the standard error when pathfinding fails instead of the new error that is more specific. It might be useful to give to `checkStdcmConfigErrors` more than just a boolean in case of pathfinding failure ? If it also received the error type returned by Editoast, we could directly return the  `zero_length_path` error instead of doing the current check on the front-end side.

SharglutDev: Yep, you're totally right, it should be done [here](https://github.com/OpenRailAssociation/osrd/issues/9707) :)
But some errors/warning we want to display doesn't exist on editoast so it won't go away totally.

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
SharglutDev, line 1: We could go a little further and check in `useStaticPathfinding.ts` (probably going to conflict with 2 other PR atm) if origin and destination are the same to avoid launching the pathfinding. There is a check l.57 where we could do the same thing as here (maybe we can extract this check in a util like in stdcm/utils.index.ts)

Sh099078: Test added for same origin and destination in https://github.com/OpenRailAssociation/osrd/pull/9931/commits/30a4ecb19ddf9be2d69d59a9988eadb093a7632e

front/src/applications/stdcm/hooks/useStaticPathfinding.ts
SharglutDev, line 1: Now that the pathfinding isn't launched when "zeroLengthPath", we don't get the warning below the form anymore. Indeed, in `StdcmConfig.tsx`, in the `useEffect` l.59, we update the warning only if we have a pathfinding but we don't anymore, so we should remove the if here.

Then we need to change the `checkStdcmConfigErrors` params a bit : if we leave it as it is, when the pathfinding actually works, you will see for a very short time the "pathfindingFailed" warning. This is because the `useEffect` is first updated when `pathSteps` are changing (we get the "pathfindingFailed" warning) and then the pathfinding starts and the warning disappears for the pathfinding loader.

To avoid that, in the function params, instead of passing `pathfindingStateError: boolean`, we can : 
- pass `pathfindingStatus?: 'success' | 'failure'` (at the end of the params because it can be undefined)
- change the `if` l.33 for `if (pathfindingStatus && pathfindingStatus === 'failure')`
- update the 2 spots where the function is called

Probably it would be better to make these changes in a separate commit.

SharglutDev (commented):
Thank you for the changes, one more thing :)

front/src/applications/stdcm/hooks/useStaticPathfinding.ts
Sh099078: Done in https://github.com/OpenRailAssociation/osrd/pull/9931/commits/a7b27cece696cc837a2b1e3a6188ad7142aaaf14

SharglutDev (approved):
Lgtm and tested, great job !

PR #9932, opened 2024-12-03
front: fix constraint distribution state in simulation settings

### LINKED ISSUE #9811: Regularity margin parameter is set for all trains when you change it
### What happened?

When you change the regularity margin parameter (linear / Mareco), it changes for all trains.

So if we calculate another train, it changes this parameter for the calculus of the train without knowing it.

https://github.com/user-attachments/assets/f1d4baac-1a94-4314-8b9c-9903a6c0c147

The parameter should be set individually for each train.

1. Open a scenario with several trains
2. Modify a train and go to Simulation parameters
3. Change the regularity margin (linear / Mareco)
4. Validate the modifications
5. Modify another train
6. Go to simulation parameter and see that the regularity margin is the same

Recette (SNCF)

Firefox

6be0604

RomainValls:
close #9811

### DIFF
--- a/front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
+++ b/front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
@@ -40,8 +40,7 @@ const ManageTrainSchedule = ({ trainIdToEdit }: ManageTrainScheduleProps) => {
   const destination = useSelector(getDestination);
   const pathSteps = useSelector(getPathSteps);
   const constraintDistribution = useSelector(getConstraintDistribution);
-  const rawStartTime = useSelector(getStartTime);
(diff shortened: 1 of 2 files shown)


clarani (changes requested):
Hum it would be better to fix this directly in the action `selectTrainToEdit` ;) you can get the `constraint_distribution` of the trainSchedule there and directly set the state

front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
clarani, line 1: You don't need the change of this file anymore ;)

clarani (approved):
LGTM :white_check_mark:

Math-R (approved):
LGTM !

Yohh (approved):

PR #9933, opened 2024-12-03
front: use Date instead of a string in times and stops helpers

emersion:
Improves type safety.

### DIFF
--- a/front/src/modules/timesStops/helpers/utils.ts
+++ b/front/src/modules/timesStops/helpers/utils.ts
@@ -5,7 +5,7 @@ import { round, isEqual, isNil } from 'lodash';
 import { keyColumn, createTextColumn } from 'react-datasheet-grid';
 
 import type { ReceptionSignal } from 'common/api/osrdEditoastApi';
-import type { IsoDateTimeString, IsoDurationString, TimeString } from 'common/types';
+import type { IsoDurationString, TimeString } from 'common/types';
 import { matchPathStepAndOp } from 'modules/pathfinding/utils';
 import type { SuggestedOP } from 'modules/trainschedule/components/ManageTrainSchedule/types';
 import type { PathStep } from 'reducers/osrdconf/types';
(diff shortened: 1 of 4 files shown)


front/src/modules/timesStops/helpers/utils.ts
```diff
@@ -282,14 +282,14 @@ export function updateDaySinceDeparture(
...
-  startTime?: IsoDateTimeString,
+  startTime?: Date,
   stepTimeDays?: TimeExtraDays
 ): IsoDurationString | null {
   if (!startTime || !stepTimeDays?.time || stepTimeDays?.daySinceDeparture === undefined) {
     return null;
   }
   const start = dayjs(startTime);
-  const step = dayjs(`${startTime.split('T')[0]}T${stepTimeDays.time}`).add(
+  const step = dayjs(`${start.format('YYYY-MM-DD')}T${stepTimeDays.time}`).add(
```
emersion, line 292: Note: we should really get rid of this kind of string manipulation…

SharglutDev (approved):
Lgtm and tested

front/src/modules/timesStops/helpers/utils.ts
anisometropie: can’t completely avoid it, but what we should avoid is mixing very low level manipulation, link strings like this, with higher abstraction

emersion: We can avoid it, e.g. by using year/month/day/hour/minute setters instead of manipulating strings.

anisometropie (approved):
Alright, tested and looking good

front/src/modules/timesStops/helpers/utils.ts
anisometropie: I mean I’m not overly shocked to have string manipulation inside low level utils function. As long as we can hide those details from the rest of the code and are able to think about the higher level code/business code without all these low levels details cluttering the space.

But yeah for dates we have decents apis, we might as well use them here

emersion: Yeah, I don't mind doing string operations in helper functions when it makes sense. My main beef with the current code is that it's doing string operations to manipulate dates, when we already have date APIs which can do the same thing in a safer and more readable way.

Issue #8816, opened 2024-09-10
front: add type-safe APIs for durations and distances


emersion:
### Description and goal

We're planning to use `Date` instead of ad-hoc strings/numbers for timestamps for improved type safety. We don't have an equivalent dedicated type for durations and distances.

Upside of using a dedicated type is that mixups are prevented. Typically when doing math on duration/distance values, it can be easy to mix [millimeters and kilometers](https://github.com/OpenRailAssociation/osrd/pull/8234), or [millimeters and a number of elements](https://github.com/OpenRailAssociation/osrd/issues/8725). A dedicated type would make the TypeScript compiler complain about such mistakes.

### Possible solution 1: TypeScript subtypes

See https://timjohns.ca/typescripts-hidden-feature-subtypes.html

```ts
type Distance = A & { readonly __type: unique symbol };
```

Downside: math on distances doesn't work, it returns a bare number.

```ts
const a = 42 as Distance;
const b = 43 as Distance;
const d: Distance = a + b; // error
```

To fix, we'd need to define helper functions such as:

```ts
const addDistances = (a: Distance, b: Distance) => a + b as Distance;
```

But it's quite tempting for users to just litter casts everywhere and loose the type safety benefits.

### Possible solution 2: dedicated class

Introduce dedicated classes for durations and distances.

Similar to the upcoming temporal standard: https://tc39.es/proposal-temporal/docs/duration.html

```ts
class Duration {
  seconds: number; // XXX: should we use milliseconds instead?

  constructor(seconds: number) {
    this.seconds = seconds;
  }

  // Make comparison operators work on Duration objects (>, <, etc)
  valueOf() {
    return this.seconds;
  }

  toString() {
    return formatToISO(this.seconds);
  }

  add(other: Duration) {
    return new Duration(this.seconds + other.seconds);
  }
}
```

Downside: math operators don't work (+, -, etc), need to use the dedicated methods. But that's similar to `Date` here.

`valueOf` is possibly too tempting for users to perform math operations, so maybe should leave it out.

### Acceptance criteria

.

emersion:
The `Duration` constructor ought to be improved to make it clearer in which unit the number passed in is, and maybe make it possible to pass in numbers in other units (much like the temporal proposal linked above).

Ref https://github.com/OpenRailAssociation/osrd/pull/10230#discussion_r1922336375

emersion:
Still TODO in this issue: introduce a type for distances.

PR #9934, opened 2024-12-03
front: introduce Duration type

emersion:
See https://github.com/OpenRailAssociation/osrd/issues/8816

_See individual commits._

Follow-up work: drop duration-related helpers from `utils/timeManipulation`, use `Duration` everywhere instead of strings/numbers.


theocrsb (approved):
LGTM and tested :)

front/src/utils/duration.ts
```diff
@@ -0,0 +1,38 @@
...
+    return new Duration(a.getTime() - b.getTime());
+  }
+
+  // Return the number of milliseconds, so that comparison operators work as expected.
+  valueOf() {
+    return this.ms;
+  }
+
+  /** Format this duration as an ISO 8601 string. */
+  toString() {
```
clarani: Maybe toISOString would be a better name ?

clarani (approved):

front/src/utils/duration.ts
emersion: Indeed, and it would match `Date` better!

emersion: Done

PR #9938, opened 2024-12-04
Setup snapshot testing for editoast_derive using insta

### LINKED ISSUE #7467: Add snapshot testing to `editoast_derive`

Software developers

Add a snapshot test for every piece of code generated by `editoast_derive`:

* `ModelV2` (most important to test, as the generated code is non-trivial)
* `Search`
* `EditoastError` (we might keep the derive macro to generate the `ToSchema` impl once we'll be ready to add specific errors in the endpoints' openapi)

The crate https://insta.rs seems like a good choice in terms of popularity, developer experience and tooling.

1. Allows writing serious tests involving code generation.
2. Highlight changes made during macro evolutions (the snapshot diffs appear in the PR)
3. Helps with debugging as the faulty macro expansion is kept in a snapshot after the test fails

_No response_

leovalais:
closes #7467 

Setup snapshot testing library [`insta`](https://insta.rs) in order to unit-test our derive macros. This approach has several benefits:

- Tests are easy to write, so we can have many
- When the snapshot changes, insta shows a nice diff, making debugging our changes to the macro easier
- Facilitates the review: snapshots are committed, so their diff appear in the PR description

Since we may soon have some extensive changes coming to our macros with the new error system, having tests and ease the review process are essential.

A few drawbacks I can think of:

- The macro expansion need to be determinist: no more `HashMap`s are allowed
- We need to refrain as much as possible from using macros in our code gen, otherwise it hinders the readability of the snapshots
- We have to "accept" diffs manually (either by renaming the file or by using [`cargo insta`](https://insta.rs/docs/cli/), their official interactive snapshot management CLI tool)

> [!IMPORTANT]
> Review this PR commit by commit to see how the snapshots evolve when the macro output changes.


.github/workflows/build.yml
```diff
@@ -589,6 +589,7 @@ jobs:
           docker load --input ./osrd-editoast-test.tar
 
       - name: Execute tests within container
+        # snapshot testing library `insta` requires CI=true
```
woshilapin: Maybe this can be moved to line 599? (just above the line that actually needs it?)

editoast/editoast_derive/src/lib.rs
```diff
@@ -237,3 +237,37 @@ pub fn model(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+#[cfg(test)]
+#[allow(unused_macros)]
+macro_rules! assert_macro_expansion {
+    ($expansion:path, $derive_input:expr) => {
+        let input: syn::DeriveInput = $derive_input;
+        let source = crate::pretty_tokens(&<syn::DeriveInput as quote::ToTokens>::to_token_stream(&input));
+        let expansion = $expansion(&input).expect("macro should expand faultlessly");
+        let expected = crate::pretty_tokens(&expansion);
+
+        // HACK: sadly insta doesn't let us print multiline strings in the scnapshot description
```
woshilapin:
```suggestion
        // HACK: sadly insta doesn't let us print multiline strings in the snapshot description
```

editoast/editoast_derive/src/snapshots/editoast_derive__error__tests__construction.snap
```diff
@@ -0,0 +1,71 @@
...
+                axum::http::StatusCode::try_from(404)
+                    .expect("EditoastError: invalid status expression")
+            }
+            #[allow(unused)]
+            Self::BadRequest { message } => {
+                axum::http::StatusCode::try_from(400)
+                    .expect("EditoastError: invalid status expression")
+            }
+            #[allow(unused)]
+            Self::InternalError => axum::http::StatusCode::from_u16(500u16).unwrap(),
```
woshilapin, line 46: Unrelated to current PR: Interesting to see that sometimes, it's expanded to `StatusCode::try_from(400)` and sometimes to `StatusCode::from_u16(500u16)`. We might want to make it the same? Or do I miss a reason why it's different?

editoast/editoast_derive/src/snapshots/editoast_derive__model__tests__construction.snap
```diff
@@ -0,0 +1,1246 @@
...
+paste::paste! {
+    impl Document { pub const [< id_ : snake : upper >] : crate ::models::ModelField <
+    Document, i64, id > = crate ::models::ModelField::new(); pub const [< content_type :
+    snake : upper >] : crate ::models::ModelField < Document, String,
+    editoast_models::tables::osrd_infra_document::content_type > = crate
+    ::models::ModelField::new(); pub const [< data : snake : upper >] : crate
+    ::models::ModelField < Document, Vec < u8 >,
+    editoast_models::tables::osrd_infra_document::data > = crate
+    ::models::ModelField::new(); }
+}
```
woshilapin, line 38: Interesting, `prettyplease` completely gave up on this... Macros really gives a hard time to the tooling.

editoast/editoast_derive/Cargo.toml
woshilapin, line 1: Should we opt into [this](https://insta.rs/docs/quickstart/#optional-faster-runs)?

editoast/editoast_derive/src/lib.rs
```diff
@@ -237,3 +237,37 @@ pub fn model(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+    let file = syn::parse_file(tokens.to_string().as_str()).unwrap();
+    prettyplease::unparse(&file)
+}
+
+// Extra allows because of some false negative lints, not sure why.
+// Maybe because editoast_derive is a proc-macro crate?
+
+#[cfg(test)]
+#[allow(unused_macros)]
+macro_rules! assert_macro_expansion {
```
woshilapin: Why do we need a macro here? Isn't it a function that would take a `Fn` and a `DeriveInput` as input arguments?

editoast/editoast_derive/src/lib.rs
```diff
@@ -237,3 +237,37 @@ pub fn model(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
         .unwrap_or_else(darling::Error::write_errors)
         .into()
 }
+
+#[cfg(test)]
```
woshilapin, line 241: There is 3 elements exposed below only in case of tests. How about putting them all inside a `mod tests`?

editoast/editoast_derive/src/lib.rs
```diff
@@ -237,3 +237,37 @@ pub fn model(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+pub(crate) fn pretty_tokens(tokens: &proc_macro2::TokenStream) -> String {
```
woshilapin: Not sure that it matters to have `pub(crate)` instead of `pub` since it's protected by `#[cfg(test)]` anyway 🤷

woshilapin (approved):
> We have to "accept" diffs manually (either by renaming the file or by using [cargo insta](https://insta.rs/docs/cli/), their official interactive snapshot management CLI tool)

I actually don't see this as a drawback. Sure, it's more work for the developer, but we actually always do that even for "normal" unit tests: a expected output changed, we need to review it manually and validate that the new output is valid. So it's actually a good thing. Maybe you meant it's more cumbersome to do that operation of validating a different test output?

In general, this PR is going to bring much more to the experience of dealing with macros...

editoast/editoast_derive/Cargo.toml
flomonster: Sound a good idea to set this up.

flomonster:
Great idea! Thanks for the PR.

.github/workflows/build.yml
leovalais: Done in 3abe9aaa2

editoast/editoast_derive/src/snapshots/editoast_derive__error__tests__construction.snap
leovalais: Not sure why it's different. We could correct that in another PR but since this macro will be removed sooner than later, idk 🤷

editoast/editoast_derive/src/snapshots/editoast_derive__model__tests__construction.snap
leovalais: Yep, I had spotted this one as well... The case conversion could be done in the `Model` macro. I didn't want to add too much to this PR (the `chunked_for_libpq!` was really bad, that's why I fixed it here). Follow up PR if any?

editoast/editoast_derive/Cargo.toml
leovalais: I didn't want to set that up by default since `insta` is just used in `editoast_derive` for now and not many people run the tests of that crate, let alone work on it. I figured it was best to let people configure that locally if needed instead of imposing the cost of that compilation to everyone.

But alright I'll do that.

editoast/editoast_derive/src/lib.rs
leovalais: It's to keep the right failure source location. If we put the assert in a function, should it fail, the top stack frame would indicate that function, instead of the expected location in our failing test.

I don't have a strong preference for the macro, but I find it a bit cleaner, lmk what you prefer.

editoast/editoast_derive/src/lib.rs
leovalais: Done in 61411e63a

editoast/editoast_derive/src/lib.rs
leovalais: Done in 61411e63a. I left the `use` at top-level to be able to access the macro at crate-level: `crate::assert_macro_expansion!(...)` instead of `crate::test_utils::assert_macro_expansion!(...)`.

leovalais:
Thanks for the feedback! I've added `*.snap.new` to our `.gitignore` to avoid accidentally commit snapshots that still require some review (missing `cargo insta review`).

editoast/editoast_derive/src/lib.rs
leovalais: Update in 227047406

Apparently placing everything in a module prevents the `unused_macro` false positive... 😑

editoast/editoast_derive/src/lib.rs
leovalais: Actually, cf. 227047406

I have to `pub(crate) use assert_macro_expansion;` otherwise Rust complains about a `missing #[macro_export]`. I figured it would be more consistent to keep both definitions visible at `pub(crate)` (even if you're right, it's not different from `pub` in this context). I'll let you resolve if you agree, I don't mind changing it either :)

editoast/editoast_derive/src/lib.rs
woshilapin: Ah, I see, then maybe you need [`#[track_caller]`](https://doc.rust-lang.org/reference/attributes/codegen.html#the-track_caller-attribute) on your function?

editoast/editoast_derive/src/lib.rs
woshilapin: I should have added `nitpick` to my comment 😅 sure it's fine

Khoyo:
We should probably add some documentation about working with insta, either in editoast README or in a new one for editoast_derive

editoast/editoast_derive/src/lib.rs
leovalais: I didn't know about this one thanks! 😮

Seems like what I want :)

leovalais: It works for the assertion location, but `insta` does some magic to retrieve the name of the test, which basically resolves to getting the function name. From the expansion of `assert_snapshot!`:

```rust
            fn f(){}
            
            fn type_name_of_val<T>(_:T) ->  &'static str {
                std::any::type_name::<T>()
            }
            let mut name = type_name_of_val(f).strip_suffix("::f").unwrap_or("");
            while let Some(rest) = name.strip_suffix("::{{closure}}"){
                name = rest;
            }name
```

Placing the `assert` in a function results with all the snapshots being suffixed by `assert_macro_expansion` instead of the test name:

image

I guess it's better to keep the macro under these conditions 🤷

editoast/editoast_derive/Cargo.toml
leovalais: Done in 4882aa790

editoast/editoast_derive/src/lib.rs
woshilapin: I see, berk. I've looked a bit, and `assert_snapshot!()` does call `_assert_snapshot_base!()` (exposed but `#[doc(hidden)]`) which itself calls the function `insta::_macro_support::assert_snapshot` (the entire module `_macro_support` is `pub` but also `#[doc(hidden)]`). So it might be possible... but I don't have the time to dig that rabbit hole. I'm resolving.

woshilapin (approved):
Thanks for digging up about the "macro into function" proposition, even if it ended up not satisfying. And thanks for the additional documentation. Great improvement for `editoast_derive` in general! ❤️

Khoyo (approved):

PR #9939, opened 2024-12-04
front: fix lazy loading for rolling stock images

emersion:
We were unconditionally fetching the rolling stock image via getDocument(). Instead, side-step getDocument() entirely, and directly use the document URL as the image source. That way, the loading="lazy" attribute can be used and the image is only fetched if it's visible.

The deleted comment is wrong: authentication is required, but is performed via a cookie. Cookies are always sent for same-origin image requests.

### DIFF
--- a/front/src/modules/rollingStock/components/RollingStock2Img.tsx
+++ b/front/src/modules/rollingStock/components/RollingStock2Img.tsx
@@ -2,7 +2,7 @@ import React, { useEffect, useState } from 'react';
 
 import placeholderRollingStockElectric from 'assets/pictures/placeholder_rollingstock_elec.gif';
 import placeholderRollingStockThermal from 'assets/pictures/placeholder_rollingstock_thermal.gif';
-import { getDocument } from 'common/api/documentApi';
+import { getDocumentUrl } from 'common/api/documentApi';
 import type {
   LightRollingStockWithLiveries,
   RollingStockWithLiveries,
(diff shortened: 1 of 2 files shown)


theocrsb (approved):

front/src/modules/rollingStock/components/RollingStock2Img.tsx
```diff
@@ -13,14 +13,14 @@ type RollingStock2ImgProps = {
...
-  // as the image is stored in the database and can be fetched only through api (authentication needed),
-  // the direct url can not be given to the <img /> directly. Thus the image is fetched, and a new
-  // url is generated and stored in imageUrl (then used in <img />).
   const [imageUrl, setImageUrl] = useState('');
 
-  const getRollingStockImage = async () => {
+  useEffect(() => {
     const { liveries } = rollingStock;
-    if (!rollingStock || !Array.isArray(liveries)) return;
+    if (!rollingStock || !Array.isArray(liveries)) {
```
clarani:
```suggestion
    if (!Array.isArray(liveries)) {
```

clarani (approved):
LGTM :white_check_mark:

front/src/modules/rollingStock/components/RollingStock2Img.tsx
emersion: Done

PR #9941, opened 2024-12-04
core: etcs: use the braking curves in the simulation

### LINKED ISSUE #9705: core: plug ETCS module into core

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/9705

Doesn't use the new data yet, but there's clear entry points

### DIFF
--- a/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/EnvelopeSimContext.kt
+++ b/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/EnvelopeSimContext.kt
@@ -1,14 +1,32 @@
 package fr.sncf.osrd.envelope_sim
 
 import com.google.common.collect.RangeMap
+import fr.sncf.osrd.sim_infra.api.Path
+import fr.sncf.osrd.utils.DistanceRangeSet
+import fr.sncf.osrd.utils.units.Offset
 
-class EnvelopeSimContext(
+class EnvelopeSimContext
+@JvmOverloads
+constructor(
     @JvmField val rollingStock: PhysicsRollingStock,
     @JvmField val path: PhysicsPath,
     @JvmField val timeStep: Double,
     @JvmField
-    val tractiveEffortCurveMap: RangeMap<Double, Array<PhysicsRollingStock.TractiveEffortPoint>>
+    val tractiveEffortCurveMap: RangeMap<Double, Array<PhysicsRollingStock.TractiveEffortPoint>>,
+    /** If the train should follow ETCS rules, this contains some extra context */
+    val etcsContext: ETCSContext? = null,
 ) {
+
+    data class ETCSContext(
+        /** Braking curves are computing using ETCS rules when they end in these ranges. */
+        val brakingRanges: DistanceRangeSet,
+        /**
+         * List of switch and buffer stops offsets on the path, up to the first switch/buffer stop
+         * *after* the end of the path (or right at the end).
+         */
+        val dangerPointOffsets: List<Offset<Path>>,
+    )
+
     /** Creates a context suitable to run simulations on envelopes */
     init {
         checkNotNull(tractiveEffortCurveMap)
--- a/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/EnvelopeSimContext.kt
+++ b/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/EnvelopeSimContext.kt
@@ -18,10 +18,13 @@ constructor(
 ) {
 
     data class ETCSContext(
-        /** Braking curves are computing using ETCS rules when they end in these ranges. */
-        val brakingRanges: DistanceRangeSet,
         /**
-         * List of switch and buffer stops offsets on the path, up to the first switch/buffer stop
+         * Ranges where ETCS rules are applied. Braking curves are computed using ETCS rules if they
+         * *end* in these ranges.
+         */
+        val applicationRanges: DistanceRangeSet,
+        /**
+         * List of switch and buffer stop offsets on the path, up to the first switch/buffer stop
          * *after* the end of the path (or right at the end).
          */
         val dangerPointOffsets: List<Offset<Path>>,
--- a/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/pipelines/MaxSpeedEnvelope.kt
+++ b/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/pipelines/MaxSpeedEnvelope.kt
@@ -47,32 +47,44 @@ object MaxSpeedEnvelope {
      * Generate braking curves overlay everywhere the mrsp decrease (increase backwards) with a
      * discontinuity
      */
-    private fun addBrakingCurves(
+    private fun addSlowdownBrakingCurves(
         etcsSimulator: ETCSBrakingSimulator,
         context: EnvelopeSimContext,
         mrsp: Envelope
     ): Envelope {
         var envelope = mrsp
-        envelope = addETCSBrakingCurves(etcsSimulator, context, envelope)
+        envelope = addETCSSlowdownBrakingCurves(etcsSimulator, context, envelope)
+        envelope = addConstSlowdownBrakingCurves(context, envelope)
+        return envelope
+    }
+
+    /**
+     * Generate braking curves overlay everywhere the mrsp decreases (increases backwards) with a
+     * discontinuity using constant deceleration (outside ETCS ranges).
+     */
+    private fun addConstSlowdownBrakingCurves(
+        context: EnvelopeSimContext,
+        envelope: Envelope
+    ): Envelope {
         val builder = OverlayEnvelopeBuilder.backward(envelope)
         val cursor = EnvelopeCursor.backward(envelope)
         var lastPosition = envelope.endPos
         while (cursor.findPartTransition(MaxSpeedEnvelope::increase)) {
-            if (cursor.getPosition() > lastPosition) {
+            if (cursor.position > lastPosition) {
                 // The next braking curve already covers this point, this braking curve is hidden
                 cursor.nextPart()
                 continue
             }
             val partBuilder = EnvelopePartBuilder()
-            partBuilder.setAttr<EnvelopeProfile>(EnvelopeProfile.BRAKING)
+            partBuilder.setAttr(EnvelopeProfile.BRAKING)
             val overlayBuilder =
                 ConstrainedEnvelopePartBuilder(
                     partBuilder,
                     SpeedConstraint(0.0, EnvelopePartConstraintType.FLOOR),
                     EnvelopeConstraint(envelope, EnvelopePartConstraintType.CEILING)
                 )
-            val startSpeed = cursor.getSpeed()
-            val startPosition = cursor.getPosition()
+            val startSpeed = cursor.speed
+            val startPosition = cursor.position
             // TODO: link directionSign to cursor boolean reverse
             EnvelopeDeceleration.decelerate(
                 context,
@@ -89,12 +101,12 @@ object MaxSpeedEnvelope {
     }
 
     /** Add braking curves following ETCS rules in relevant places */
-    private fun addETCSBrakingCurves(
+    private fun addETCSSlowdownBrakingCurves(
         etcsSimulator: ETCSBrakingSimulator,
         context: EnvelopeSimContext,
         envelope: Envelope
     ): Envelope {
-        val etcsRanges = context.etcsContext?.brakingRanges ?: return envelope
+        val etcsRanges = context.etcsContext?.applicationRanges ?: return envelope
         val cursor = EnvelopeCursor.backward(envelope)
         val limitsOfAuthority = mutableListOf<LimitOfAuthority>()
         while (cursor.findPartTransition(MaxSpeedEnvelope::increase)) {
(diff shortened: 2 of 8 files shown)


core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/EnvelopeSimContext.kt
```diff
@@ -1,14 +1,30 @@
...
 
 class EnvelopeSimContext(
     @JvmField val rollingStock: PhysicsRollingStock,
     @JvmField val path: PhysicsPath,
     @JvmField val timeStep: Double,
     @JvmField
-    val tractiveEffortCurveMap: RangeMap<Double, Array<PhysicsRollingStock.TractiveEffortPoint>>
+    val tractiveEffortCurveMap: RangeMap<Double, Array<PhysicsRollingStock.TractiveEffortPoint>>,
+    /** If the train should follow ETCS rules, this contains some extra context */
+    val etcsContext: ETCSContext? = null,
```
Erashin, line 17: `class EnvelopeSimContext @JvmOverloads constructor(` generates two constructors, enabling you not to modify every java class to add a `null` default value if needed.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/EnvelopeSimContext.kt
```diff
@@ -1,14 +1,32 @@
...
 ) {
+
+    data class ETCSContext(
+        /** Braking curves are computing using ETCS rules when they end in these ranges. */
```
bougue-pe:
```suggestion
        /** Braking curves are computed using ETCS rules when they end in these ranges. */
```

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/EnvelopeSimContext.kt
```diff
@@ -1,14 +1,32 @@
...
     @JvmField val path: PhysicsPath,
     @JvmField val timeStep: Double,
     @JvmField
-    val tractiveEffortCurveMap: RangeMap<Double, Array<PhysicsRollingStock.TractiveEffortPoint>>
+    val tractiveEffortCurveMap: RangeMap<Double, Array<PhysicsRollingStock.TractiveEffortPoint>>,
+    /** If the train should follow ETCS rules, this contains some extra context */
+    val etcsContext: ETCSContext? = null,
 ) {
+
+    data class ETCSContext(
```
bougue-pe, line 20: If this struct is meant to focus only on braking, maybe more `ETCSBrakingContext`.

IMO, the name can stay, but I would then remove the "braking" vocabulary in comments to allow for a larger use/maintenance.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/EnvelopeSimContext.kt
```diff
@@ -1,14 +1,32 @@
...
+        /** Braking curves are computing using ETCS rules when they end in these ranges. */
+        val brakingRanges: DistanceRangeSet,
```
bougue-pe: Not sure: `applicationRanges` seems closer to the meaning, the "braking" intent (or general intent) should stay in class title.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/pipelines/MaxSpeedEnvelope.kt
```diff
@@ -29,10 +47,16 @@ object MaxSpeedEnvelope {
...
-        val builder = OverlayEnvelopeBuilder.backward(mrsp)
-        val cursor = EnvelopeCursor.backward(mrsp)
-        var lastPosition = mrsp.endPos
+    private fun addBrakingCurves(
+        etcsSimulator: ETCSBrakingSimulator,
+        context: EnvelopeSimContext,
+        mrsp: Envelope
+    ): Envelope {
+        var envelope = mrsp
+        envelope = addETCSBrakingCurves(etcsSimulator, context, envelope)
```
bougue-pe: To me, it would be more consistent and easy to read to have either:
```kt
addSlowdownCurves() {
  envelope = addETCSSlowdownCurves(etcsSimulator, context, envelope)
  envelope = addConstSlowdonCurves(etcsSimulator, context, envelope)
}
```
or a direct call to the 2 functions.

Ditto for `addStopCurves` being split in `addConstStopCurves` and `addETCSStopCurves`.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/pipelines/MaxSpeedEnvelope.kt
```diff
@@ -29,10 +47,16 @@ object MaxSpeedEnvelope {
...
+        val builder = OverlayEnvelopeBuilder.backward(envelope)
+        val cursor = EnvelopeCursor.backward(envelope)
+        var lastPosition = envelope.endPos
         while (cursor.findPartTransition(MaxSpeedEnvelope::increase)) {
```
bougue-pe, line 72: Should we mutualize this loop too? (my "method split" comment would be impacted).
I'm also wondering if there might be inconsistencies when not processing all backward.
Maybe add a `makeSimSlowdowns`?

core/src/main/kotlin/fr/sncf/osrd/standalone_sim/etcsContextBuilder.kt
```diff
@@ -0,0 +1,121 @@
...
+    chunkPath: ChunkPath,
+    routePath: StaticIdxList<Route>
+): List<Offset<Path>> {
+    val zonePaths = routePath.flatMap { infra.getRoutePath(it) }
+    var currentZonePathStartOffset = -getRoutePathStartOffset(infra, chunkPath, zonePaths).distance
+
+    val res = mutableSetOf<Offset<Path>>()
+    for (zonePath in zonePaths) {
+        val moveableElements = infra.getZonePathMovableElements(zonePath)
+        val moveableElementPositions = infra.getZonePathMovableElementsPositions(zonePath)
```
bougue-pe:
```suggestion
        val movableElements = infra.getZonePathMovableElements(zonePath)
        val movableElementPositions = infra.getZonePathMovableElementsPositions(zonePath)
```

core/src/main/kotlin/fr/sncf/osrd/standalone_sim/etcsContextBuilder.kt
```diff
@@ -0,0 +1,121 @@
...
+    val lastChunk = chunkPath.chunks.last()
+    var lastChunkEndOffset = chunkPath.beginOffset * -1.0
+    for (chunk in chunkPath.chunks) {
+        lastChunkEndOffset += infra.getTrackChunkLength(chunk.value).distance
+    }
+    val lastTrack = infra.getTrackFromChunk(lastChunk.value)
+    val endOfLastTrackPathOffset =
+        getEndOfLastTrackPathOffset(infra, lastTrack, lastChunk, lastChunkEndOffset)
+
+    // Iterate on the tracks until there either a switch or buffer stop
```
bougue-pe:
```suggestion
    // Iterate on the tracks until finding either a switch or a buffer stop
```

core/src/main/kotlin/fr/sncf/osrd/standalone_sim/etcsContextBuilder.kt
```diff
@@ -0,0 +1,121 @@
...
+    // As an offset on the undirected last track, location of the start of the directed last chunk
+    val lastDirectedChunkStartOffsetOnTrack =
+        if (lastChunk.direction == Direction.INCREASING) lastUndirectedChunkStartOffsetOnTrack
+        else lastUndirectedChunkStartOffsetOnTrack + lastChunkLength.distance
+
+    val distanceFromChunkEndToTrackEnd =
+        if (lastChunk.direction == Direction.INCREASING)
+            lastTrackLength.distance -
+                (lastDirectedChunkStartOffsetOnTrack.distance + lastChunkLength.distance)
+        else lastDirectedChunkStartOffsetOnTrack.distance - lastChunkLength.distance
```
bougue-pe: You may also save accessor calls in `DECREASING` case, but you can at least simplify:
```suggestion
    val distanceFromChunkEndToTrackEnd =
        if (lastChunk.direction == Direction.INCREASING)
            lastTrackLength.distance -
                (lastUndirectedChunkStartOffsetOnTrack.distance + lastChunkLength.distance)
        else lastUndirectedChunkStartOffsetOnTrack.distance
```

core/src/main/kotlin/fr/sncf/osrd/standalone_sim/etcsContextBuilder.kt
```diff
@@ -0,0 +1,121 @@
...
+        etcsRanges,
+        buildETCSDangerPoints(infra, chunkPath, routePath)
+    )
+}
+
+/**
+ * Builds the list of ETCS "danger points" (switches and buffer stops). Goes up to the first one at
+ * or after the end of the path. Does not return the point at offset 0 if any (wouldn't be useful).
+ * May return any number of point beyond the end of the path, specifically any point covered by the
+ * routes used by the path.
```
bougue-pe, line 36: What is coded is smarter (and easier) than what I imagined ❤️
But I don't get this phrase for 2 reasons:
* why would there be more than one point after the end?
* this doesn't use route (simplifying the logic a lot) so I wouldn't mention it, as it's confusing, especially when the (shorter) routePath doesn't cover the danger point.

bougue-pe (commented):
Looks good, no major comment, thank you! 👍

core/src/main/kotlin/fr/sncf/osrd/standalone_sim/etcsContextBuilder.kt
eckter: It does use routes, they're used to generate the zone paths. We return a value for any switch in the zone paths, even if they're after the end of the path. That can result in having several points after then end. (which should never be an issue in this context)

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/pipelines/MaxSpeedEnvelope.kt
eckter: This one is a little different, adding a deceleration curve ending at a discontinuity may erase other discontinuities. If we pre-compute the discontinuities we'd need to filter them and it's not necessarily that easy to do, the current version feels more robust.

Not processing it strictly backwards means that we may compute unnecessary deceleration curves that would be later "absorbed" by a different one. It's not ideal but not really a big deal either.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/EnvelopeSimContext.kt
eckter: I feel like this *could* become more general than just braking. I'm not 100% sure what we'll need for the signaling part.

What comments are you thinking about? The ones about this class and instances don't specifically mention braking (but I may be missing some).

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/pipelines/MaxSpeedEnvelope.kt
eckter: Fixed in https://github.com/OpenRailAssociation/osrd/pull/9941/commits/d7d2531580caf4341191020fcd174646bae2d0a8

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/pipelines/MaxSpeedEnvelope.kt
bougue-pe: Functionally: As we can now have different shape of braking curves, we probably need to process them all and make sure to really use the minimum of all curves.
Ex: I imagine a quite "flat" ETCS braking curve ending late in the path, and in the middle of it (so ending sooner in the path) a steep const curve, we probably need to apply all curves. (this is a side attention point)

As for the code mutualisation: it may make sense to process curves one by one (const, ETCS LoA, ETCS EoA, ETCS SvL) and then apply the min for each of them.
If this is a good idea, it may also be done later.

core/src/main/kotlin/fr/sncf/osrd/standalone_sim/etcsContextBuilder.kt
bougue-pe: Ah, thanks for your response and my bad, I was focused on what's done after the route path 🙈

We may filtrer to the first dangerPoint after the chunkPath's end? (and also avoid calls to findLastDangerPoint)
Not sure perf gain or code clarity is worth it, though.
(resolve this if you don't think so).

bougue-pe (approved):

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/EnvelopeSimContext.kt
```diff
@@ -1,14 +1,33 @@
...
+    /** If the train should follow ETCS rules, this contains some extra context */
+    val etcsContext: ETCSContext? = null,
 ) {
+
+    data class ETCSContext(
+        /** Ranges where ETCS rules are applied. Braking curves are computed using ETCS rules
+         * if they *end* in these ranges. */
+        val applicationRanges: DistanceRangeSet,
+        /**
+         * List of switch and buffer stops offsets on the path, up to the first switch/buffer stop
```
Erashin:
```suggestion
         * List of switch and buffer stop offsets on the path, up to the first switch/buffer stop
```

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/pipelines/MaxSpeedEnvelope.kt
```diff
@@ -29,26 +47,44 @@ object MaxSpeedEnvelope {
...
+        mrsp: Envelope
+    ): Envelope {
+        var envelope = mrsp
+        envelope = addETCSSlowdownBrakingCurves(etcsSimulator, context, envelope)
+        envelope = addConstSlowdownBrakingCurves(context, envelope)
+        return envelope
+    }
+
+    /**
+     * Generate braking curves overlay everywhere the mrsp decrease (increase backwards) with a
```
Erashin:
```suggestion
     * Generate braking curves overlay everywhere the mrsp decreases (increases backwards) with a
```

Erashin (approved):
Details below, good job.

PR #9947, opened 2024-12-04
editoast: extract all the STDCM log logic into a function

woshilapin:
Here is a suggestion to put everything inside a single function and extract all this weird debugging logic as far as possible of the main business logic that is in STDCM endpoint.

⚠️ Note that I targeted directly the branch of #9873


editoast/src/models/stdcm_log.rs
```diff
@@ -1,12 +1,17 @@
...
+use opentelemetry::trace::TraceContextExt;
 use serde::Deserialize;
 use serde::Serialize;
+use tracing_opentelemetry::OpenTelemetrySpanExt;
 use utoipa::ToSchema;
 
 use crate::core::stdcm::Request;
 use crate::core::stdcm::Response;
+use crate::models::prelude::*;
+use crate::views::Authentication;
```
leovalais, line 14: A `views` import in `models` :/

Given how things are going, the authentication and authorization are likely to stay in `views` and in a dedicated crate. The `models` "crate" won't probably know about that layer. Wdyt about taking an `Option<i64>` instead?

editoast/src/models/stdcm_log.rs
```diff
@@ -25,3 +30,37 @@ pub struct StdcmLog {
     pub created: DateTime<Utc>,
     pub user_id: Option<i64>,
 }
+
+impl StdcmLog {
+    // We just don't await the creation of the log entry since we want
+    // the endpoint to return as soon as possible, and because failing
+    // to persist a log entry is not a very important error here.
```
leovalais, line 37: Shouldn't this comment stay near the `spawn`? It mentions an "endpoint" and awaiting or not the return value is a decision for the caller.

leovalais (approved):
Great to keep the `stdcm` handler as clean as possible! There's just a few things I'd leave in that handler ⬇️

editoast/src/models/stdcm_log.rs
woshilapin: Good catch, so agreed. I really quickly prototyped this solution (maybe I should have opened a draft). This is one thing to improve for sure.

editoast/src/models/stdcm_log.rs
woshilapin: Good point.

hamz2a:
As discussed, I have integrated the changes into the [PR](https://github.com/OpenRailAssociation/osrd/pull/9873), thank you very much for the suggestions. @woshilapin @leovalais

woshilapin:
This work has been integrated into https://github.com/OpenRailAssociation/osrd/pull/9873, I'm closing it.

PR #9948, opened 2024-12-04
front: cancel fetch request when user cancels simulation in lmr

### LINKED ISSUE #9779: stdcm - results are printed for canceled requests
### What happened?

If I cancel a request, I still receive the results in the frontend which creates the following weird behaviors : 
- Minor : if I just cancel a request without launching another one, I receive a result which is confusing
- Major : VERY CONFUSING : If I cancel a request (n°1) and then start another one (n°2), I first receive the result of n°1 (but the inputs in the form are n°2) and then I receive n°2, which overwrites the first one (in the same simulation tab, the picture here shows what I have after the first result)

Image

I expected either 
- If possible, when a request is canceled, stopping the computation of the result
- If the previous point is not possible, ignoring the responses which correspond to canceled simulations, and not displaying them

Create a first simulation, cancel it, change the inputs, and create a 2nd simulation. One sees the results of the first simulation, then these results are overwritten with those of the second simulation.

Dev (SNCF)

Google Chrome

a4daa6d

RomainValls:
close #9779 

Also, had a minor bug where the button to launch the simulation disappeared after canceling a request, added a fix.

### DIFF
--- a/front/src/applications/stdcm/hooks/useStdcm.ts
+++ b/front/src/applications/stdcm/hooks/useStdcm.ts
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useRef, useState } from 'react';
 
 import { useTranslation } from 'react-i18next';
 import nextId from 'react-id-generator';
@@ -52,6 +52,7 @@ const useStdcm = ({
   const { getConf, getTimetableID } = useOsrdConfSelectors();
   const osrdconf = useSelector(getConf) as OsrdStdcmConfState;
   const timetableId = useSelector(getTimetableID);
+  const requestPromise = useRef<ReturnType<typeof postTimetableByIdStdcm>>();
 
   const stdcmResults = useStdcmResults(stdcmResponse, stdcmTrainResult, setPathProperties);
 
@@ -64,9 +65,6 @@ const useStdcm = ({
       },
       { skip: !osrdconf.rollingStockID }
     );
-
-  // https://developer.mozilla.org/en-US/docs/Web/API/AbortController
-  const controller = new AbortController();
 
   const { speedLimitByTag } = useStoreDataForSpeedLimitByTagSelector({ isStdcm: true });
 
@@ -85,7 +83,10 @@ const useStdcm = ({
     if (validConfig) {
       const payload = formatStdcmPayload(validConfig);
       try {
-        const response = await postTimetableByIdStdcm(payload).unwrap();
+        const promise = postTimetableByIdStdcm(payload);
+        requestPromise.current = promise;
+
+        const response = await promise.unwrap();
 
         if (
           response.status === 'success' &&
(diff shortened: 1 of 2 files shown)


front/src/applications/stdcm/hooks/useStdcm.ts
```diff
@@ -52,6 +52,8 @@ const useStdcm = ({
   const { getConf, getTimetableID } = useOsrdConfSelectors();
   const osrdconf = useSelector(getConf) as OsrdStdcmConfState;
   const timetableId = useSelector(getTimetableID);
+  const requestPromise = useRef<ReturnType<typeof postTimetableByIdStdcm>>();
+  const isRequestCancelled = useRef(false);
```
emersion: Instead of having a ref, could we `try`/`catch` and check if the error is an `AbortError`?

RomainValls: You're saying we should use AbortController, with signal, in a try catch, to see if we have an AbortError ? And if it's the case, return early ?

emersion: No need for `AbortController`, I think the error should already be an `AbortError` if we call `abort()` like already done in this PR? Or is the error a completely different kind?

RomainValls:
When canceling a simulation, the button to start another one isn't displayed anymore, unless we change the value of one of the inputs. I understand that we might not want to launch a simulation again with the same inputs if we decided to cancel it (@SharglutDev), but it does feel weird. 
If we keep the button hidden until a value changes, can we at least inform the user he has to change a value in the form ? 
@thibautsailly any idea ?

SharglutDev:
> When canceling a simulation, the button to start another one isn't displayed anymore, unless we change the value of one of the inputs. I understand that we might not want to launch a simulation again with the same inputs if we decided to cancel it (@SharglutDev), but it does feel weird. If we keep the button hidden until a value changes, can we at least inform the user he has to change a value in the form ? @thibautsailly any idea ?

If the user cancels the simulation in progress, it's probably because he forgot one parameter in the form or the simulation was taking too long because some inputs weren't good enough. So in any case, I don't think he will want to launch the same simulation again.

It's the same thing if the simulation succeeds, does he might want to relaunch the same simulation ? We decided he can't at the moment.

emersion (approved):
LGTM!

front/src/applications/stdcm/views/StdcmView.tsx
```diff
@@ -57,7 +57,6 @@ const StdcmView = () => {
   const handleSelectSimulation = (index: number) => {
     if (retainedSimulationIndex === -1) {
       setSelectedSimulationIndex(index);
-      setShowBtnToLaunchSimulation(false);
```
SharglutDev: Also here, I'm not sure what it has to do with this PR.

Moreover, when the user switches in a simulation, he won't want to relaunch the exact same one. That's why the launch button was hidden I suppose. It will be displayed again when the user changes a param. I think we should leave it that way.

front/src/applications/stdcm/hooks/useStdcm.ts
```diff
@@ -132,23 +133,24 @@ const useStdcm = ({
...
 
   const cancelStdcmRequest = () => {
-    // when http ready https://axios-http.com/docs/cancellation
-
-    controller.abort();
+    if (typeof requestPromise.current?.abort === 'function') {
+      requestPromise.current.abort();
+    }
+    requestPromise.current = undefined;
     setCurrentStdcmRequestStatus(STDCM_REQUEST_STATUS.canceled);
```
SharglutDev, line 154: Seems like we don't do anything with this status. Should we ask @thibautsailly if we want to display something when the user cancels the request ?

front/src/applications/stdcm/views/StdcmView.tsx
RomainValls: No problem, I put it back 😅

thibautsailly:
Sorry for being so late to answer, my bad 😬.
I can see 3 possible situations in which the computation could be cancelled by the user:
1. She realises that one of the parameter isn't right, and needs to fix it.
2. She changes her mind about this configuration being the right one to test.
3. She mis-clicks the button.

In situation 1, the button presence will be guaranteed by the form update.
Situation 2 leads to two other ones: A, she carries on and updates the form to test another configuration; B, she realises that actually this configuration is worth testing and needs to restart it right away. 2A is similar to 1, but in 2B we would need to display the button.
In situation 3, we would need to display the button.

In the nominal use case, the button isn't displayed because it is replaced by the confirmation message. Since we changed the "ongoing compute" message, we could also alter how the compute result is shown, and show the button in all cases, with a confirmation message if we detect the user is about to use the same configuration as a previous one. But it's too much work for now, so let's just display the button when the user cancels the simulation computation.

SharglutDev (approved):
Lgtm and tested, great job !

front/src/applications/stdcm/views/StdcmView.tsx
```diff
@@ -107,6 +108,12 @@ const StdcmView = () => {
     }
   }, [isDebugMode]);
 
+  useEffect(() => {
+    if (isCanceled) {
+      setShowBtnToLaunchSimulation(true);
+    }
+  }, [isCanceled]);
```
emersion, line 115: Nit: instead of this `useEffect()`, we could pass `showBtnToLaunchSimulation={showBtnToLaunchSimulation || isCanceled}` in the JSX below. This would make it easier to reason about the data flow.

SharglutDev: oh good idea

PR #9950, opened 2024-12-04
editoast: reduce logging level to trace for some verbose events

leovalais:
Proposition PR.

Are demoted to trace level:

* Query payloads sent to core
* PG connection instrumentation

### DIFF
--- a/editoast/editoast_models/src/db_connection_pool/tracing_instrumentation.rs
+++ b/editoast/editoast_models/src/db_connection_pool/tracing_instrumentation.rs
@@ -27,7 +27,7 @@ impl Instrumentation for TracingInstrumentation {
         match event {
             InstrumentationEvent::StartEstablishConnection { url, .. } => {
                 let url = Url::parse(url).unwrap();
-                let span = tracing::info_span!(
+                let span = tracing::trace_span!(
                     "connection",
                     // opentelemetry_semantic_conventions::attribute::DB_SYSTEM
                     "db.system" = "postgresql",
@@ -40,7 +40,7 @@ impl Instrumentation for TracingInstrumentation {
                 );
                 {
                     let _guard = span.enter();
-                    tracing::debug!("establishing a connection");
+                    tracing::trace!("establishing a connection");
                 }
                 self.connection_span = span;
             }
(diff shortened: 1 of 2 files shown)


woshilapin (commented):
Before reviewing or anything, have you consider using `RUST_LOG=editoast_models=none` before changing the log level? I often realize that these kinds of PR comes from a problematic situation when debugging something... but for a specific case, which might not apply in all situations. I consider that these logs are about `editoast_models`... and therefore, everything related to DB interactions have some interest, at least more than `trace` in my opinion.

And probably the same applies for the core client, if it ends up being moved inside its own crate (but it's already its own module so actually, the argument might still already be valid).

🤓 I'm not saying no to this PR, I'm just asking a bit more to be convinced it's the right move.

leovalais:
I'm not 100% sure either, but the current situation still feels quite wrong to me. (Hence the "proposition PR", I don't mind closing it.)

Actually, I cracked when I was trying to observe the DB interactions while testing the STDCM request logging feature. I couldn't spot the event of `StdcmLog::create` because it was drowned under the extensive logging of the DB connection and the core client. Having both `Model` functions and DB queries at the same log level feels sketchy to me.

Besides:

> at least more than trace in my opinion

Then I wonder what goes into the `trace` level?

My take is that we must not have too much trouble finding `debug` logs. Otherwise it's not a convenient "debug"ging tool. In other words, if `debug` logs are too verbose, or too numerous, it hinders the debug ability and thus defeats the purpose of the log. So I'd rather have anything that falls into these categories at the `trace` level, where we expect to be overflowed by logs.

I'm not saying I'm right or that we should do things this way, but I guess that's where I draw the line between `debug` and `trace`.

Wdyt?

flomonster (approved):
LGTM. I don't have a preference for the default log level.

woshilapin:
I've been incapable to find a reasonable definition to separate `DEBUG` and `TRACE`. In order to move on, and without a good definition, I propose to take the next best choice (or I think it is): let's take the best default (the one that will, in general, be the most reasonable). And even if I had to debug once and those logs were useful, I'd say that in most cases, they're not needed. So let's keep them as `TRACE`, maybe the future will teach us more anyway.

woshilapin (approved):

PR #9952, opened 2024-12-04
script: docker: remove docker images of the tested PR when cleaning

bougue-pe:
Saves disk resources (and forces re-pull, which may be useful for new version)

### DIFF
--- a/scripts/cleanup-db.sh
+++ b/scripts/cleanup-db.sh
@@ -23,7 +23,7 @@ OSRD_VALKEY="osrd-valkey"
 OSRD_VALKEY_VOLUME="osrd_valkey_data"
 OSRD_POSTGRES_PORT=5432
 OSRD_VALKEY_PORT=6379
-if [ "$PR_TEST" -eq 1 ]; then
+if [ "$PR_TEST" = 1 ]; then
   OSRD_POSTGRES="osrd-postgres-pr-tests"
   OSRD_EDITOAST="osrd-editoast-pr-tests"
   OSRD_VALKEY="osrd-valkey-pr-tests"
--- a/scripts/load-backup.sh
+++ b/scripts/load-backup.sh
@@ -41,7 +41,7 @@ fi
 # These variables are necessary to load the infra on the correct instance (the pr-infra or the dev one)
 OSRD_POSTGRES="osrd-postgres"
 OSRD_POSTGRES_PORT=5432
-if [ "$PR_TEST" -eq 1 ]; then
+if [ "$PR_TEST" = 1 ]; then
   OSRD_POSTGRES="osrd-postgres-pr-tests"
   OSRD_POSTGRES_PORT=5433
 fi
--- a/scripts/pr-tests-compose.sh
+++ b/scripts/pr-tests-compose.sh
@@ -74,10 +74,20 @@ elif [ "$1" = "down" ]; then
 
 else
 
-    # Shutdown and clean the docker instance
+    # Shutdown and clean the docker instance (remove "osrd" images too)
+    osrd_images=$(docker compose \
+        -p "osrd-pr-tests" \
+        -f "docker/docker-compose.pr-tests.yml" \
+        images --format json \
+        core editoast osrdyne gateway front \
+        | jq --raw-output '.[].ID')
+
     docker compose \
         -p "osrd-pr-tests" \
         -f "docker/docker-compose.pr-tests.yml" \
         down -v
+
+    # shellcheck disable=SC2086
+    docker rmi ${osrd_images}
 
 fi


scripts/cleanup-db.sh
```diff
@@ -23,7 +23,7 @@ OSRD_VALKEY="osrd-valkey"
 OSRD_VALKEY_VOLUME="osrd_valkey_data"
 OSRD_POSTGRES_PORT=5432
 OSRD_VALKEY_PORT=6379
-if [ "$PR_TEST" -eq 1 ]; then
+if [ "$PR_TEST" -eq "1" ]; then
```
Khoyo: This quoting seems spurious

scripts/load-backup.sh
```diff
@@ -41,7 +41,7 @@ fi
 # These variables are necessary to load the infra on the correct instance (the pr-infra or the dev one)
 OSRD_POSTGRES="osrd-postgres"
 OSRD_POSTGRES_PORT=5432
-if [ "$PR_TEST" -eq 1 ]; then
+if [ "$PR_TEST" -eq "1" ]; then
```
Khoyo: Ditto

scripts/pr-tests-compose.sh
```diff
@@ -74,10 +74,22 @@ elif [ "$1" = "down" ]; then
...
-    # Shutdown and clean the docker instance
+    # Shutdown and clean the docker instance (remove "osrd" images too)
+    osrd_images=$(docker compose \
+        -p "osrd-pr-tests" \
+        -f "docker/docker-compose.pr-tests.yml" \
+        images \
+        core editoast osrdyne gateway front \
+        | grep "pr-tests" \
+        | sed -e "s/[[:space:]]\+/ /g" \
+        | cut -f4 -d ' ')
```
woshilapin: Just to remove some of the magic of it...

```suggestion
        | cut --fields=4 --delimiter=' ')
```

and actually, most of the above code too `docker compose --file ... --profile ...`, `sed --expression` (you might also use `--regexp-extended` to avoid to escape `\+`)...

woshilapin (approved):

scripts/pr-tests-compose.sh
Khoyo: `awk '/pr-tests/ {print $4}'`

Not really magic, no grep, sed then cut required :p

Although I'd prefer having docker compose outputting some json and using jq

woshilapin: Well, it was a general comment on my general aversion 👼 for [short options in scripts](https://hole.tuziwo.info/don-t-be-lazy.html).

Khoyo: And it was a general comment about my love for awk :p

bougue-pe: Agreed for the long options.
Agreed for moving to `awk`.

Before changing this:
Agree that `jq` is even better, but I didn't want to ask for that dependency.
If you all think it's better, it sure won't bother me.

scripts/cleanup-db.sh
bougue-pe: The alternative correction (for a bad type comparison when empty) is:
```sh
if [[ "$PR_TEST" -eq 1 ]]; then
```
But it casts empty `$PR_TEST` to `0` which is less maintainable IMO.

Khoyo: huh? This would require bash - which the script currently doesn't, but in shell `1` and `"1"` have the same type and value, they are both the `1` *string*. Even if we then do an integer comparison for some reason (which we shouldn't anyway since it makes test spuriously error on non integer looking inputs). *The shell strongly/stringly typed nature strikes again*

If you want to fix it in this PR, I'd rather keep it POSIX and use `[ ! -z "$PR_TEST" ] && [ ! "$PR_TEST" = 0 ]` (ie. check for unset, empty or zero) - or just `[ "$PR_TEST"  = 1 ]` if we wish to keep the same behavior

edited: `-a` is not POSIX in test

scripts/pr-tests-compose.sh
Khoyo: I don't think jq will pose a problem for most people. If it does, we can always circle back to awk

scripts/cleanup-db.sh
bougue-pe: Indeed, I just moved to `[ "$PR_TEST"  = 1 ]` in both tests (rebased/amended)

scripts/pr-tests-compose.sh
bougue-pe: Move to `jq` done in the amended commit.

Didn't expand all the existing options for docker compose, which should still be OK.

scripts/load-backup.sh
bougue-pe: See other conversation (resolving this)

Khoyo (approved):

PR #9954, opened 2024-12-05
front: use Date instead of string in OsrdConfState.startTime

emersion:
_See individual commits._

Improves type safety.

### DIFF
--- a/front/src/reducers/osrdconf/operationalStudiesConf/index.ts
+++ b/front/src/reducers/osrdconf/operationalStudiesConf/index.ts
@@ -5,7 +5,6 @@ import type { TrainScheduleWithDetails } from 'modules/trainschedule/components/
 import computeBasePathStep from 'modules/trainschedule/helpers/computeBasePathStep';
 import { defaultCommonConf, buildCommonConfReducers } from 'reducers/osrdconf/osrdConfCommon';
 import type { OsrdConfState } from 'reducers/osrdconf/types';
-import { convertIsoUtcToLocalTime } from 'utils/date';
 import { msToKmh } from 'utils/physics';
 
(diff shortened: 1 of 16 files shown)


front/src/reducers/osrdconf/operationalStudiesConf/index.ts
```diff
@@ -34,7 +33,7 @@ export const operationalStudiesConfSlice = createSlice({
 
       state.rollingStockID = rollingStock?.id;
       state.pathSteps = computeBasePathSteps(action.payload);
-      state.startTime = convertIsoUtcToLocalTime(start_time);
+      state.startTime = new Date(start_time);
```
clarani: Instead of retransforming the start_time here, you can directly use startTime (which is already a Date ;) )

clarani (commented):
Nice cleaning !! Thanks !

Found a bug: 
- when editing a train, the date is not displayed

RomainValls (approved):
LGTM (except for the bug mentioned by @clarani), tested ✅

emersion:
Ooops, sounds like `formatLocalDateTime()` was using a bad format string. Fixed:

```diff
 export const formatLocalDateTime = (date: Date) =>
-  dayjs(date).local().format('YYYY-MM-DDTHH:MM:SS');
+  dayjs(date).local().format('YYYY-MM-DDTHH:mm:ss');
```

front/src/reducers/osrdconf/operationalStudiesConf/index.ts
emersion: I was worried about changing behavior in a refactoring, but it seems like this should be safe. Done!

clarani (approved):

PR #9956, opened 2024-12-05
editoast: fix projection endpoint

### LINKED ISSUE #9534: Path projection endpoint fails
### What happened?

Path projection fails with specific train schedules.

The path projection should be successful.

1. Create a scenario with this [trains_fails.json](https://github.com/user-attachments/files/17589075/trains_fails.json)
2. Observe the path projection call fail

Recette (SNCF)

Firefox

6d9ed21

flomonster:
close #9534

We found out that this code was broken since the `trains` variable does not have the same length as `trains_details` and `cached_projections`.

### DIFF
--- a/editoast/src/views/train_schedule.rs
+++ b/editoast/src/views/train_schedule.rs
@@ -796,6 +796,8 @@ async fn get_path(
 #[cfg(test)]
 mod tests {
     use axum::http::StatusCode;
+    use chrono::DateTime;
+    use chrono::Utc;
     use pretty_assertions::assert_eq;
     use rstest::rstest;
     use serde_json::json;
@@ -920,8 +922,7 @@ mod tests {
         )
     }
 
-    async fn app_infra_id_train_schedule_id_for_simulation_tests() -> (TestApp, i64, i64) {
-        let db_pool = DbConnectionPoolV2::for_tests();
+    fn mocked_core_pathfinding_sim_and_proj(train_id: i64) -> MockingClient {
         let mut core = MockingClient::new();
         core.stub("/v2/pathfinding/blocks")
             .method(reqwest::Method::POST)
@@ -975,10 +976,18 @@ mod tests {
                 }
             }))
             .finish();
-        let app = TestAppBuilder::new()
-            .db_pool(db_pool.clone())
-            .core_client(core.into())
-            .build();
+        core.stub("/v2/signal_projection")
+            .method(reqwest::Method::POST)
+            .response(StatusCode::OK)
+            .json(json!({
+                "signal_updates": {train_id.to_string(): [] },
+            }))
+            .finish();
+        core
+    }
+
+    async fn app_infra_id_train_schedule_id_for_simulation_tests() -> (TestApp, i64, i64) {
+        let db_pool = DbConnectionPoolV2::for_tests();
         let small_infra = create_small_infra(&mut db_pool.get_ok()).await;
         let rolling_stock =
             create_fast_rolling_stock(&mut db_pool.get_ok(), "simulation_rolling_stock").await;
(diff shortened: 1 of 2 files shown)


editoast/src/views/train_schedule/projection.rs
```diff
@@ -277,40 +274,40 @@ async fn project_path(
...
+    let trains_hash_values: HashMap<_, _> = trains_details
         .iter()
-        .map(|t| t.id)
+        .map(|t| t.train_id)
         .zip(trains_hash_values)
         .collect();
     let mut new_items = vec![];
-    for id in miss_cache.keys() {
-        let hash = &trains_hash_values[id];
+    for id in miss_cache.iter().map(|t| t.train_id) {
```
woshilapin: I'd argue that you might want to actually be explicit about which kind of identifier it is (I'm under the impression that it's actually because of that kind of things that the bug exists in the first place?).

```suggestion
    for train_id in miss_cache.iter().map(|t| t.train_id) {
```

editoast/src/views/train_schedule/projection.rs
```diff
@@ -277,40 +274,40 @@ async fn project_path(
...
-    let mut project_path_result = HashMap::new();
     let rolling_stock_length: HashMap<_, _> = rolling_stocks
         .into_iter()
         .map(|rs| (rs.name, rs.length))
         .collect();
 
     // 4.2 Build the projection response
-    for (id, cached) in hit_cache {
+    let mut project_path_result = HashMap::new();
+    for (cached, id) in hit_cache {
```
woshilapin: Same here, maybe we can be explicit about which kind of `id`?

editoast/src/views/train_schedule/projection.rs
```diff
@@ -277,40 +274,40 @@ async fn project_path(
     let signal_updates = signal_updates?;
 
     // 3. Store the projection in the cache (using pipeline)
-    let trains_hash_values: HashMap<_, _> = trains
+    let trains_hash_values: HashMap<_, _> = trains_details
         .iter()
-        .map(|t| t.id)
+        .map(|t| t.train_id)
```
woshilapin, line 279: Am I correct if I say that, should we use a `TrainId` type instead of a `i64` everywhere we need identifiers, we would not have had this bug? Is this something we want to investigate?

editoast/src/views/train_schedule/projection.rs
woshilapin, line 1: I'm surprised a bug fix doesn't change any test behavior. We probably should add a test which would fail under the old code, and success with the bug fix.

editoast/src/views/train_schedule/projection.rs
flomonster: It's not a type issue, so even with a `TrainId` type, we would have the bug.

editoast/src/views/train_schedule/projection.rs
flomonster: Fixed

editoast/src/views/train_schedule.rs
```diff
@@ -1024,4 +1038,143 @@ mod tests {
...
         app.fetch(request).assert_status(StatusCode::OK);
     }
+
+    #[derive(Deserialize)]
+    struct PartialProjectPathTrainResult {
+        departure_time: DateTime<Utc>,
+        // Ignore the rest of the payload
+    }
+
+    fn mocked_core_for_project_path(train_id: i64) -> MockingClient {
```
flomonster:
```suggestion
    fn mocked_core_pathfinding_sim_projection(train_id: i64) -> MockingClient {
```

editoast/src/views/train_schedule.rs
```diff
@@ -920,8 +922,7 @@ mod tests {
         )
     }
 
-    async fn app_infra_id_train_schedule_id_for_simulation_tests() -> (TestApp, i64, i64) {
-        let db_pool = DbConnectionPoolV2::for_tests();
+    fn mocked_core_pathfinding_and_sim() -> MockingClient {
```
flomonster: Remove this function and keep only one `MockedClient`.

woshilapin (approved):
Thanks for adding the test. The improvements look good.

PR #9960, opened 2024-12-05
front: fix simulation results width on wide screen

### LINKED ISSUE #9959: front: wrong width on wide screen
### What happened?

actually on screen resolution superior to 1280p the simulation results looks like this:

Image

the simulation results should stretch to the maximum width available:

Image

create a simulation with a screen resolution superior to 1280p

Recette (SNCF), Local, Dev (SNCF), Integration (SNCF)

Firefox, Google Chrome, Microsoft Edge, Safari, Opera

-

Yohh:
close #9959

### DIFF
--- a/front/src/styles/scss/applications/operationalStudies/_scenario.scss
+++ b/front/src/styles/scss/applications/operationalStudies/_scenario.scss
@@ -120,7 +120,9 @@
     height: var(--content-height);
     background-color: var(--black5);
     min-width: 392px;
-    max-width: 472px;
+    @media screen and (max-width: 1280px) {
+      max-width: 472px;
+    }
     padding: 14px 24px 0 24px;
     box-shadow: inset -1px 0 var(--black5);
 


Yohh:
NB: I m not sure that the max-width is really usefull at all, even in lower resolutions

SharglutDev:
> NB: I m not sure that the max-width is really usefull at all, even in lower resolutions

Probably. Doesn't it achieve the same result if you remove `max-width` totally ?

Yohh:
> > NB: I m not sure that the max-width is really usefull at all, even in lower resolutions
> 
> Probably. Doesn't it achieve the same result if you remove `max-width` totally ?

it releases the `col` classes, so the width fit responsively, that seems much better to me, I added the media query only because I thought it was a mockup criteria

in fact, combining `col` and min/max-width seems to lack consistence

SharglutDev:
Totally missread the issue. I thought we want to stretch the results, not the sidebar. Your solution seems fine to me.

We should probably get rid off the `col`s when we work more on the responsive design.

SharglutDev (approved):

Yohh:
on the contrary, `col`s are more responsive than using `px`, if I remove max-width and min-width totally the behavior seems fine:


https://github.com/user-attachments/assets/12639d16-5c1f-481b-9c1d-79ed4ea75926

SharglutDev:
> on the contrary, `col`s are more responsive than using `px`, if I remove max-width and min-width totally the behavior seems fine:
> Screencast.From.2024-12-05.14-23-05.mp4

Yes but as you mentioned, the mockup specifies that the default sidemenu width should be 432px

Capture d’écran 2024-12-05 à 15 02 54

Yohh:
> Yes but as you mentioned, the mockup specifies that the default sidemenu width should be 432px

ok go for it

Caracol3:
I would also tend to prefer the use of col in the application. I find it more responsive. Though, for pixel-perfect concerns, I understand that pixels are used, but the question might be whether Thibaut is familiar with using col. If not, maybe it’s up to us to adapt the design by converting pixels to columns and get it validated by Thibaut afterward.

Caracol3:
Lgtm (tested)

Caracol3 (approved):
lgtm (tested)

SharglutDev:
> I would also tend to prefer the use of col in the application. I find it more responsive. Though, for pixel-perfect concerns, I understand that pixels are used, but the question might be whether Thibaut is familiar with using col. If not, maybe it’s up to us to adapt the design by converting pixels to columns and get it validated by Thibaut afterward.

We totally should have a discussion with him next time. I added it in the discussions list for next front meeting.

PR #9961, opened 2024-12-05
editoast: skip serializing empty TrackSection extensions

Khoyo:
This causes osrd_schemas pydantic validator to complain when an extension isn't in use (for example, when trying to parse small infra tracksections).

### DIFF
--- a/editoast/editoast_schemas/src/infra/track_section_extensions.rs
+++ b/editoast/editoast_schemas/src/infra/track_section_extensions.rs
@@ -9,7 +9,9 @@ use crate::infra::TrackSectionSourceExtension;
 #[serde(deny_unknown_fields)]
 pub struct TrackSectionExtensions {
     #[schema(inline)]
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub sncf: Option<TrackSectionSncfExtension>,
     #[schema(inline)]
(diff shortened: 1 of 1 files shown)


woshilapin (approved):

leovalais:
I'm wondering why this isn't a fix on the python side? I'd be surprised if pedantic cannot be configured to accept this scheme. What would we do with pydantic if we hadn't control on the API?

Anyways, if there are good reasons why this isn't possible, I'd suggest to annotate all other infra objects extension for consistency. Wdyt?

Khoyo:
> I'm wondering why this isn't a fix on the python side? I'd be surprised if pedantic cannot be configured to accept this scheme. What would we do with pydantic if we hadn't control on the API?

I'm not enough of a pydantic wizard for that (especially with the dynamic extension registration that we have going on)

> Anyways, if there are good reasons why this isn't possible, I'd suggest to annotate all other infra objects extension for consistency. Wdyt?

Sure. Could we maybe pass this as a rubberband either way? (it's a blocker for dags work)

leovalais:
> Could we maybe pass this as a rubberband either way?

Sure np

flomonster (approved):

Khoyo:
Turns out I was just being dumb with pydantic. Superseeded by #9976

PR #9962, opened 2024-12-05
front: display path steps even if the simulation failed

### LINKED ISSUE #8418: Missing path item on map for invalid train schedule
### What happened?

When having an invalid path finding train schedule.
Path items are not visible on map even if they match a PR.

image

> [!NOTE]
> I tried with path items using `operational_point` variant. But it might be the case with other variant such as `track_offset`, `trigram`, `uic`...

I should see path items on the map.

1. Import [train_schedules.1.json](https://github.com/user-attachments/files/17246674/train_schedules.1.json)
2. Edit the "nope" train and go to the "Route" tab
3. We can see the list of path items but nothing on the map 

Local

Firefox

clarani:
closes #8418


emersion:
osrdToNge already does the exact same kind of logic. Would be nice to de-duplicate.

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -75,9 +78,66 @@ const useSetupItineraryForTrainUpdate = (
...
+          let op: SearchResultItemOperationalPoint | undefined;
+          if ('uic' in step) {
+            op = ops.find((o) => o.uic === step.uic);
+            name = `${op?.name} - ${step.uic}`;
+          } else if ('trigram' in step) {
+            op = ops.find((o) => o.trigram === step.trigram);
+            name = `${op?.name} - ${step.trigram}`;
+          } else {
+            op = ops.find((o) => o.obj_id === step.operational_point);
+            name = `${op?.name} - ${step.operational_point}`;
```
SharglutDev: Do we really want to add uic/trigram/opId after the name ?
Capture d’écran 2024-12-06 à 13 13 33
It's a bit messy on the map and when adding path step the regular way, we see only the name (or some track meta data information if it's a via added by map click)

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -75,9 +78,66 @@ const useSetupItineraryForTrainUpdate = (
...
+      // get operational points
+      let ops: SearchResultItemOperationalPoint[] = [];
+      const searchPayload = buildOpSearchQuery(infraId, [trainSchedule]);
+
+      if (searchPayload) {
+        ops = (await postSearch({
+          searchPayload,
+          pageSize: 1000,
+        }).unwrap()) as SearchResultItemOperationalPoint[];
+      }
```
SharglutDev: We could also do that. Maybe it's safer if someone later somehow changes the value of `ops` because it's a `let`.

```suggestion
      const searchPayload = buildOpSearchQuery(infraId, [trainSchedule]);

      const ops = searchPayload
        ? ((await postSearch({
            searchPayload,
            pageSize: 1000,
          }).unwrap()) as SearchResultItemOperationalPoint[])
        : [];
```

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -75,9 +78,66 @@ const useSetupItineraryForTrainUpdate = (
...
+        }
+
+        return {
+          ...computeBasePathStep(trainSchedule, index),
+          coordinates,
+          name,
+        };
+      });
+
+      // update le store
```
SharglutDev:
```suggestion
```

SharglutDev (commented):
Thank you for the fix, works great, just a couple comment :)

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
clarani: Done

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
clarani: Actually, I added the uic/trigram/opId because I thought it might be useful to debug the basic trainSchedules, but after spending 3 hours trying to debug, I think we don't need it :)

It's removed :+1:

SharglutDev (approved):
Lgtm and tested :)

PR #9967, opened 2024-12-05
front: stop putting path step IDs in SuggestedOP.opId

emersion:
The opId field was (ab)used for track offset waypoints: we were putting path step IDs in there. Instead, make opId optional and add a new field for the path step ID.

This paves the way for simplifying SuggestedOP/PathStep matching: we'll be able to use pathStepId instead of trying to match with a random subset of uic/trigram/secondary_code/kp/positionOnPath and friends.


Uriel-Sautron (approved):
Tested

clarani (approved):
LGTM :white_check_mark:

PR #9971, opened 2024-12-05
core: pathfinding: filter out paths with repeated tracks during exploration

eckter:
Prevents issues where the post-processing fails because of "repeated tracks in the path".

The result changes to "no path found" in most cases, but this time with some more details (such as constraints). It finds *some* new paths, but they make some extreme detours. 

It does come with a mild performance hit (roughly +10% computation time).

### DIFF
--- a/core/kt-osrd-utils/src/main/kotlin/fr/sncf/osrd/utils/AppendOnlySet.kt
+++ b/core/kt-osrd-utils/src/main/kotlin/fr/sncf/osrd/utils/AppendOnlySet.kt
@@ -1,0 +1,36 @@
+package fr.sncf.osrd.utils
+
+/**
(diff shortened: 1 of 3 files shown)


Khoyo (approved):
LGTM, although I'm not sure I'd call a 10% hit mild...

I also wonder if we could do something about it, eg. paying copies vs traversing a linked list or better tuning of the bloom filter

eckter:
> I also wonder if we could do something about it, eg. paying copies vs traversing a linked list or better tuning of the bloom filter

I already did and tried some optimizations, in the current version almost all of the extra computing time is spent actually creating the linked list itself. I don't know if that could be avoided.

~~I can *think* of some vague idea about using bitsets the size of the number of tracks in the infra, and marking them that way. But it would be quite tricky.~~ edit: wait no that sounds bad

PR #9973, opened 2024-12-05
front: fix lmr max speed prefill

### LINKED ISSUE #9972: front: max speed in the consist is in m/s
### What happened?

When selecting a rolling stock in the stdcm interface, the max speed is automatically set, but the value is incorrect because it is in m/s and not in km/h.
Then the simulation fails if I don't remove the max speed.

The max speed should be correct in km/h

1. open the stdcm page
2. select a rolling stock
3. see the max speed

Bonus:
5. select path steps and try to launch a request
6. the request fails

Recette (SNCF)

Firefox

.

Wadjetz:
fix #9972


SharglutDev (approved):
Lgtm and tested, thank you for the quick fix :)

Akctarus (approved):
Tested, LGTM !

PR #9977, opened 2024-12-06
front: set default tolerances in the store

### LINKED ISSUE #9803: The default tolerance value is not taken into account
### What happened?

The default tolerence value is not charged in the payload

Image

The default tolerence value -30/+30 should charged in the Payload

1. Go to LMR
2. Start a simulation without changing the tolerence
3. Observe the payload 

Recette (SNCF)

Firefox, Google Chrome

6be0604

clarani:
closes #9803


theocrsb (changes requested):
Not working for me :/ 
image

theocrsb:
> Not working for me :/ image

After cleaning the local storage, it works well for me

theocrsb (approved):
LGTM and tested !

Math-R (approved):
lgtm !

PR #9980, opened 2024-12-06
core: stdcm: optimize incremental conflict detection

### LINKED ISSUE #9801: core: stdcm: improve conflict detection performances
In stdcm on long requests, we spend roughly 70% of the computation time identifying conflicts. And it's not even running the signaling simulation and similar *complex* things, it's really just listing and comparing requirements. Specifically, `IncrementalConflictDetectorImpl.checkConflicts` and `IncrementalConflictDetectorImpl.analyzeConflicts` take respectively 16% + 42% of the global execution time.

We could certainly do better.

Maybe instead of having a `MutableList<SpacingZoneRequirement>` for each zone, we could use something like a `RangeSet`, with time ranges where it's free / used.

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/9801
Helps with https://github.com/OpenRailAssociation/osrd/issues/10034 without entirely fixing it

Possible review by commit (though it's possible to just ignore the previous version and read the new one on its own):

1. Split "static" and "incremental" conflict detection classes (the uses and possible optimizations differ quite a lot)
2. Remove incremental *routing* conflict detection (wasn't used and brought lots of complexity, we can add it back later)
3. Use range sets / tree maps for zone uses instead of a list of requirements

I haven't run a *proper benchmark* as it requires a pretty specific setup (thousands of trains in a realistic timetable), but I do have some metrics:

* In the profiler, this section of the code goes from ~60% of execution time ~down to 2-5%~
* On the slowest request I've logged (uncapped), execution time goes from 330s down to 190s. Still a timeout but not by much (180s threshold)

This PR also reduces code complexity by quite a lot, and *may* fix some of the "mismatch" crashes. 

---

**edit**: hold on, the new version *still* takes up a lot of time when it timeout (a lot more than 2-5% execution time). On one massive request, the time we spend in the conflict detection methods goes from 71% to 45% execution time: it's much better but still quite a lot. *edit2*: partially fixed by https://github.com/OpenRailAssociation/osrd/pull/10126.

### DIFF
--- a/core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
+++ b/core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
@@ -19,6 +19,8 @@ data class ConflictProperties(
     val maxDelayWithoutConflicts: Double,
     // If there are no conflicts, minimum begin time of the next requirement that could conflict
     val timeOfNextConflict: Double,
+    val hasConflict: Boolean,
+    val firstConflictTime: Double?,
 )
 
 fun incrementalConflictDetectorFromTrainReq(
@@ -34,11 +36,6 @@ fun incrementalConflictDetectorFromReq(
 }
 
 interface IncrementalConflictDetector {
-    fun checkConflicts(
-        spacingRequirements: List<SpacingRequirement>,
-        routingRequirements: List<RoutingRequirement>
-    ): List<Conflict>
-
     fun analyseConflicts(
         spacingRequirements: List<SpacingRequirement>,
         routingRequirements: List<RoutingRequirement>
@@ -115,7 +112,7 @@ class IncrementalConflictDetectorImpl(requirements: List<Requirements>) :
         }
     }
 
-    override fun checkConflicts(
+    private fun checkConflicts(
         spacingRequirements: List<SpacingRequirement>,
         routingRequirements: List<RoutingRequirement>
     ): List<Conflict> {
@@ -129,7 +126,7 @@ class IncrementalConflictDetectorImpl(requirements: List<Requirements>) :
         return res
     }
 
-    fun checkSpacingRequirement(req: SpacingRequirement): List<Conflict> {
+    private fun checkSpacingRequirement(req: SpacingRequirement): List<Conflict> {
         val requirements = spacingZoneRequirements[req.zone] ?: return listOf()
 
         val res = mutableListOf<Conflict>()
@@ -190,10 +187,17 @@ class IncrementalConflictDetectorImpl(requirements: List<Requirements>) :
         spacingRequirements: List<SpacingRequirement>,
         routingRequirements: List<RoutingRequirement>
     ): ConflictProperties {
+        val conflicts = checkConflicts(spacingRequirements, routingRequirements) // TODO remove this
         val minDelayWithoutConflicts =
             minDelayWithoutConflicts(spacingRequirements, routingRequirements)
         if (minDelayWithoutConflicts != 0.0) { // There are initial conflicts
-            return ConflictProperties(minDelayWithoutConflicts, 0.0, 0.0)
+            return ConflictProperties(
+                minDelayWithoutConflicts,
+                0.0,
+                0.0,
+                true,
+                conflicts.first().startTime,
+            )
         } else { // There are no initial conflicts
             var maxDelay = Double.POSITIVE_INFINITY
             var timeOfNextConflict = Double.POSITIVE_INFINITY
@@ -227,7 +231,13 @@ class IncrementalConflictDetectorImpl(requirements: List<Requirements>) :
                     }
                 }
             }
-            return ConflictProperties(minDelayWithoutConflicts, maxDelay, timeOfNextConflict)
+            return ConflictProperties(
+                minDelayWithoutConflicts,
+                maxDelay,
+                timeOfNextConflict,
+                false,
+                null,
+            )
         }
     }
 
--- a/core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
+++ b/core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
@@ -1,301 +1,154 @@
 package fr.sncf.osrd.conflicts
 
-import fr.sncf.osrd.api.ConflictDetectionEndpoint.ConflictDetectionResult.Conflict
-import fr.sncf.osrd.api.ConflictDetectionEndpoint.ConflictDetectionResult.Conflict.ConflictType
-import fr.sncf.osrd.api.ConflictDetectionEndpoint.ConflictDetectionResult.ConflictRequirement
-import fr.sncf.osrd.standalone_sim.result.ResultTrain.RoutingRequirement
+import com.google.common.collect.Range
+import com.google.common.collect.RangeSet
+import com.google.common.collect.TreeRangeSet
 import fr.sncf.osrd.standalone_sim.result.ResultTrain.SpacingRequirement
+import java.util.TreeMap
 import kotlin.math.max
 import kotlin.math.min
 
 const val DEFAULT_WORK_SCHEDULE_ID: Long = -1
 
-data class ConflictProperties(
-    // If there are conflicts, minimum delay that should be added to the train so that there are no
-    // conflicts anymore
+sealed interface IncrementalConflictResponse
+
+data class ConflictResponse(
+    // minimum delay that should be added to the train so that there are no conflicts anymore
     val minDelayWithoutConflicts: Double,
-    // If there are no conflicts, maximum delay that can be added to the train without creating any
-    // conflict
+    // Time at which the first conflict happened
+    val firstConflictTime: Double,
+) : IncrementalConflictResponse
+
+data class NoConflictResponse(
+    // maximum delay that can be added to the train without creating any conflict
     val maxDelayWithoutConflicts: Double,
-    // If there are no conflicts, minimum begin time of the next requirement that could conflict
+    // minimum begin time of the next requirement that could conflict
     val timeOfNextConflict: Double,
-    val hasConflict: Boolean,
-    val firstConflictTime: Double?,
-)
+) : IncrementalConflictResponse
 
 fun incrementalConflictDetectorFromTrainReq(
     requirements: List<TrainRequirements>
 ): IncrementalConflictDetector {
-    return IncrementalConflictDetectorImpl(convertTrainRequirements(requirements))
+    return IncrementalConflictDetector(convertTrainRequirements(requirements))
 }
 
 fun incrementalConflictDetectorFromReq(
     requirements: List<Requirements>
 ): IncrementalConflictDetector {
-    return IncrementalConflictDetectorImpl(requirements)
+    return IncrementalConflictDetector(requirements)
 }
 
-interface IncrementalConflictDetector {
-    fun analyseConflicts(
-        spacingRequirements: List<SpacingRequirement>,
-        routingRequirements: List<RoutingRequirement>
-    ): ConflictProperties
-}
-
-class IncrementalConflictDetectorImpl(requirements: List<Requirements>) :
-    IncrementalConflictDetector {
-    private val spacingZoneRequirements =
-        mutableMapOf<String, MutableList<SpacingZoneRequirement>>()
-    private val routingZoneRequirements =
-        mutableMapOf<String, MutableList<RoutingZoneRequirement>>()
+/**
+ * This class takes a list of requirements as input, and can only be used to compare them to a *new*
+ * set of requirements. The initial requirements cannot be modified. Conflicts between initial
+ * trains are not tested.
+ *
+ * In practice, this is used for STDCM, where the initial requirements represent the timetable
+ * trains, and new requirements come from the train we are trying to fit in the timetable.
+ */
+class IncrementalConflictDetector(requirements: List<Requirements>) {
+    // Zone name -> (end time -> Range(start time, end time)).
+    // The range is partially redundant, but it makes for easier and clearer processing.
+    private val spacingZoneUses: Map<String, TreeMap<Double, Range<Double>>>
 
     init {
-        generateSpacingRequirements(requirements)
-        generateRoutingRequirements(requirements)
+        spacingZoneUses = generateSpacingZoneUses(requirements)
     }
 
-    data class SpacingZoneRequirement(
-        val id: RequirementId,
-        override val beginTime: Double,
-        override val endTime: Double,
-    ) : ResourceRequirement
-
-    private fun generateSpacingRequirements(requirements: List<Requirements>) {
-        // organize requirements by zone
+    private fun generateSpacingZoneUses(
+        requirements: List<Requirements>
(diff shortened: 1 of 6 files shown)


core/src/test/kotlin/fr/sncf/osrd/conflicts/IncrementalConflictDetectorTests.kt
```diff
@@ -0,0 +1,63 @@
+package fr.sncf.osrd.conflicts
+
+import fr.sncf.osrd.standalone_sim.result.ResultTrain
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class IncrementalConflictDetectorTests {
+    data class SimpleRequirement(
+        val id: Int,
```
bougue-pe: Nit, a little bit clearer to me:
```suggestion
        val zone_id: Int,
```

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
```diff
@@ -0,0 +1,150 @@
...
+ * This class takes a list of requirements as input, and can only be used to compare them to a *new*
+ * set of requirements. The initial requirements cannot be modified. Conflicts between initial
+ * trains are not tested.
+ *
+ * In practice, this is used for STDCM, where the initial requirements represent the timetable
+ * trains, and new requirements come from the train we are trying to fit in the timetable.
+ */
+class IncrementalConflictDetector(requirements: List<Requirements>) {
+    // Zone name -> (end time -> Range(start time, end time))
+    private val spacingZoneUses: Map<String, TreeMap<Double, Range<Double>>>
```
bougue-pe: This is a bit confusing (redundant?), would it be valid/clearer to move to:
```suggestion
    // Zone name -> (start availability time -> end availability time)
    private val spacingZoneAvailability: Map<String, TreeMap<Double, Double>>
```

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
```diff
@@ -0,0 +1,150 @@
...
+        }
+        return rangeSets
+            .map { it.key to TreeMap(it.value.asRanges().associate { it.upperEndpoint() to it }) }
+            .toMap()
+    }
+
+    /**
+     * Check for any conflicts between the initial requirements and the ones given here as method
+     * input. Returns a polymorphic response with different extra data for either case (conflict /
+     * no conflict).
```
bougue-pe: nit (Checks/Returns consistency + any):
```suggestion
     * Checks for any conflict between the initial requirements and the ones given here as method
     * input. Returns a polymorphic response with different extra data for either case (conflict /
     * no conflict).
```

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
```diff
@@ -0,0 +1,150 @@
...
+                val firstEntryStartAfter = map.higherEntry(requirementStart) ?: continue
+                if (firstEntryStartAfter.value.lowerEndpoint() >= requirementEnd) continue
+
+                val extraDelay = firstEntryStartAfter.value.upperEndpoint() - requirementStart
+                minDelay += extraDelay
+                hasIncreasedDelay = true
+            }
+            // No new conflicts
+            if (!hasIncreasedDelay || minDelay.isInfinite()) return minDelay
+        }
```
bougue-pe, line 152: Can't we just remove the surrounding `while(true) {}`, remove `hasIncreasedDelay` and `return minDelay` all the time?

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
```diff
@@ -0,0 +1,150 @@
...
+    private fun earliestConflictTime(
+        spacingRequirements: List<SpacingRequirement>,
+    ): Double {
+        var res = Double.POSITIVE_INFINITY
+        for (spacingRequirement in spacingRequirements) {
+            val map = spacingZoneUses[spacingRequirement.zone] ?: continue
+            val entry = map.higherEntry(spacingRequirement.beginTime) ?: continue
+            if (entry.value.lowerEndpoint() > spacingRequirement.endTime) continue
+            val firstConflictTime = max(entry.value.lowerEndpoint(), spacingRequirement.beginTime)
+            res = min(res, firstConflictTime)
```
bougue-pe, line 119: If `spacingRequirements` are time-sorted (not sure about the best way to ensure that), we could skip half of the exploration on average.

Also if it can be sorted, there is probably some room to get some indications from the process of `minDelayWithoutConflicts`. This may bloat the code though, so not sure it's worth it.

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
```diff
@@ -0,0 +1,150 @@
...
+            )
+        } else { // There are no initial conflicts
+            var maxDelay = Double.POSITIVE_INFINITY
+            var timeOfNextConflict = Double.POSITIVE_INFINITY
+            for (spacingRequirement in spacingRequirements) {
+                val map = spacingZoneUses[spacingRequirement.zone] ?: continue
+                val entry = map.higherEntry(spacingRequirement.beginTime) ?: continue
+                val nextUse = entry.value.lowerEndpoint()
+                maxDelay = min(maxDelay, nextUse - spacingRequirement.endTime)
+                timeOfNextConflict = min(timeOfNextConflict, nextUse)
```
bougue-pe, line 97: The remark on sorting and early returning is probably true here also.

bougue-pe (commented):
Thanks for this works, found the new code quite easy to read indeed 🙏 
(some of my suggestions may tend to old code intricacy...)

Didn't review the old code and what might be missing from it.

As I understand it, we just don't check the routing requirements since STDCM v2, and this PR removes the check from v1. Which is fine by me (trusting you on a late non-incremental check for routing conflicts 🙈).

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
bougue-pe: Or maybe even type the sorted datetime bounds?
```kt
    // Zone name -> sorted (time availability change, becomesAvailable)
    private val spacingZoneAvailability: Map<String, List<Pair<Double, Boolean>>>
```
If we want to dig on the perf here, there may be some room for caching, too: would be useful when time windows are moving on the same path for different requests... The cache could be stable (same infra/STDCM scenario) but quite big (and it requires a shared cache between cores, so valkey would be the minimal option).

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
Erashin: No. The point of this method is that by adding this delay, we have 0 conflicts with the spacingZoneUses. Each time the delay is increased, we risk hitting a requirement further along on the same zone. With your proposal, we wouldn't be checking that with this newly computed delay, we aren't creating conflicts with other requirements further along.

Example: spacingZoneUses: (ZoneA, [(0.0, 100.0), (150.0, 200.0)])
              spacingRequirements: (ZoneA, (0.0, 75.0))
With your solution, we would determine the minDelay to add is 100.0: to avoid the first zone use on A, we add 100.0.
However, if we add this delay, the new spacing requirement would be ZoneA(100.0, 175.0), which now conflicts with the second zone use on A. The correct min delay is 200.0.

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
```diff
@@ -0,0 +1,150 @@
...
+            }
+            return NoConflictResponse(
+                maxDelay,
+                timeOfNextConflict,
+            )
+        }
+    }
+
+    /**
+     * Returns the earliest time at which there is a conflict (a ressource is used by the new train
```
Erashin:
```suggestion
     * Returns the earliest time at which there is a conflict (a resource is used by the new train
```

Erashin (commented):
Removing the `routingRequirements` + `higherEntry` really simplifies everything. Maybe we'll be able to do things more easily with this structure as well once/if we have to add the routing requirements back (it was really bloated with the previous way it was coded).
I'll let you deal with the other comments, but this looks pretty decent already.

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
bougue-pe: OK, I see your point, the handling of the "too narrow availabilities" would be wrong.

Moreover, my idea was wrong on the fact that between zones it could brake again, for example:
spacingZoneUses: (ZoneA, [(0, 100), (150, 200), (300, 400)]), (ZoneB, [(0, 160), (170, 300)])
spacingRequirements: (ZoneA, (0, 75)), (ZoneB, (75, 85))
From A, we would get 200 delay, then from B we would extend to 300, but we wouldn't go back to check A (and extend to 400).

bougue-pe: That being said, here is an idea that might change the worst case-complexity (although not sure at all on the average) and remove the `while(true)` red-flag.

image
* The idea would be to process "possible delay ranges" for each zone, then stack them to get the "soonest" common delay.
* On a zone:
  *  we would get the "really possible start ranges" by just shifting sooner the start of each use from the requirement's duration ➡️ We could already do it on the current algorithm
  * then we would shift all those ranges from the value of the start of the requirement (direction of the shift is probably wrong on the schema)

The complexity has to be considered on collecting the ranges (always considering all the uses, but just once), then also on stacking them for "collision".

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
eckter: > If we want to dig on the perf here, there may be some room for caching, too: would be useful when time windows are moving on the same path for different requests... The cache could be stable (same infra/STDCM scenario) but quite big (and it requires a shared cache between cores, so valkey would be the minimal option).

In theory yes, but because the requirements use string IDs, checking against the cache takes too long to be faster (I tried). We'd need to fix https://github.com/OpenRailAssociation/osrd/issues/10124 first.

I'll check the other part of the comment soon

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
eckter: I don't think it would be more efficient: zone uses can span very far in the future (as seen in [this other PR](https://github.com/OpenRailAssociation/osrd/pull/10126)). Processing the entirety of one zone is likely more computations that what we'd do in the current version over all zones.

And different zones in the same call are generally grouped together in close/overlapping time ranges, it's unlikely to iterate for a long time.

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
eckter: I don't think the "sorted bounds" would be better, we'd miss out on the `.higherEntry`.

Using availability instead of uses may indeed be clearer though. I'll try to make the change and see how it unfolds.

core/src/test/kotlin/fr/sncf/osrd/conflicts/IncrementalConflictDetectorTests.kt
eckter: Fixed in https://github.com/OpenRailAssociation/osrd/pull/9980/commits/d9a0f84e16ff9c84831a2165a9cf23f8c4676afb

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
eckter: There's no early return here, we still need to go through all zones in order to find the minimum.

We *could* sort trough the different input requirements for one given zone, but generally speaking there should only be one requirement per zone in most cases (during exploration).

(same for the other sorting comment)

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
eckter: > Using availability instead of uses may indeed be clearer though. I'll try to make the change and see how it unfolds.

I tried to make the change but it was too much troubles to make it work. ~~Sometimes we'd end up wanting to do what ends up being `inf - inf`, and~~ figuring out the relevant entry isn't always as straightforward.

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
bougue-pe: It looks like we are already iterating over all the zone uses at init when calling `generateSpacingZoneUses`.
So it's a bigger refactor, but an idea could be to regroup all of them right away.

But it may:
* impact the perf as the size of the range set would grow
* generate other difficulties
* complexify code

I'm not sure it's worth digging more now or in the future (we may open an "idea" issue?), so it's fine to resolve this conversation.

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
bougue-pe: OK, thanks for trying, maybe adding a little comment on redundancy of `end time` would be nice, I let you resolve the conversation (it's fine without comment too).

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
bougue-pe: Maybe a little comment on the `while(true)` loop can help read the code here.

bougue-pe (approved):
Improved perf and much clearer to me too, thanks for all this, really helps to dive in this part 🙏

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
eckter: Commented in this commit: https://github.com/OpenRailAssociation/osrd/pull/9980/commits/77d5703cca4e60f2af93b43afe9c9e553fc3fde5

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
eckter: Commented in this commit: https://github.com/OpenRailAssociation/osrd/pull/9980/commits/77d5703cca4e60f2af93b43afe9c9e553fc3fde5

At init we do iterate over the timetable requirements, but we only do it once per request, this is called much more often

Erashin (approved):
Gj!

PR #9983, opened 2024-12-07
front: add lmr help module

### LINKED ISSUE #9570: Implement help component in last minute request app
### Description and goal

What's this task about, and what is it for?

This task is about the development and the implementation of the "Aide" component, in last minute request app.
It aims to help the user with several information on how and what to set to launch a simulation.

### Acceptance criteria

A list of conditions which define when this task shall be considered complete

- [x] When the "Aide" button is clicked in the menu, a modal sheet must display on the right side of the screen
- [x] The modal sheet must contain a title "Aide"
- [x] The modal sheet must contain a short description
- [x] The modal sheet must contain several buttons to navigate between the different "aide" topics
- [x] When a topic is selected, a button to navigate back must be displayed in the top of the sheet
- [x] The Asu support information must be always displayed in the bottom of the sheet
- [x] The "Aide" sheet can be closed by clicking the " X " button in the top right corner of the sheet

Math-R:
close #9570 

To test it, navigate through LMR application and then click on the "Aide" button on the top right corner of the screen

[maquette](https://www.sketch.com/s/15e0909c-f17d-4200-a770-b9fc1154fa7e/p/BE7E4397-21E5-4496-B2CF-37D14E976F2A/canvas)

### DIFF
--- a/front/public/locales/en/stdcm-help-section.json
+++ b/front/public/locales/en/stdcm-help-section.json
@@ -1,0 +1,321 @@
+{
+  "help": "Help",
+  "backToIndex": "help index",
+  "asu": [
+    {
+      "title": "Application name",
+      "value": "OSRD"
+    },
+    {
+      "title": "Assistance",
+      "value": "Support applicatif ASU"
+    },
+    {
+      "title": "Phone",
+      "value": "70 70 70 (ou 09 72 72 12 70)"
+    },
+    {
+      "title": "Application Code (AVAYA)",
+      "value": "176"
+    }
+  ],
+  "asuLink": "ASU support link",
+  "sections": {
+    "application": {
+      "title": "The application",
+      "content": [
+        {
+          "type": "text",
+          "value": "Last Minute Request (LMR) is a tool that allows freight train operators to quickly simulate the feasibility of last-minute path creation requests."
+        },
+        {
+          "type": "text",
+          "value": "LMR allows you to:"
+        },
+        {
+          "type": "list",
+          "items": [
+            {
+              "value": "Carry out fully independent last-minute capacity search simulations"
+            },
+            {
+              "value": "Download an LMR simulation sheet"
+            }
+          ]
+        }
+      ],
+      "subSections": [
+        {
+          "title": "Improvements",
+          "content": [
+            {
+              "type": "text",
+              "value": "We have a great roadmap for making Last Minute Request more relevant and useful to our users. Your feedback is very important for us to learn from your use, and to prioritise these efforts correctly."
+            },
+            {
+              "type": "text",
+              "value": "Please don't hesitate to share your ideas and comments."
+            },
+            {
+              "type": "text",
+              "value": "<a href='mailto:supportclients.SI@sncf.fr' target=\"_blank\">Contact us</a>"
+            }
+          ]
+        },
+        {
+          "title": "Design",
+          "content": [
+            {
+              "type": "text",
+              "value": "Last Minute Request is part of a wider programme dedicated to facilitating rail operations planning: Open Source Railway Designer. The simulation engine, OSRD, is fed by a representation of the rail network, the various types of rolling stocks, train schedules, and works schedules. SNCF Réseau and other European infrastructure managers are participating in this effort, which is being led by a team of engineers, developers and managers who are passionate about trains."
+            }
+          ]
+        }
+      ]
+    },
+    "convoy": {
+      "title": "The consist",
+      "content": [
+        {
+          "type": "text",
+          "value": "Defining the consist allows you to obtain a simulation that is as close as possible to reality. The more closely the information you enter matches the actual composition of your consist, the closer the speed of the simulated train will be to operational."
+        },
+        {
+          "type": "text",
+          "value": "The tool does not yet allow you to change the composition en route. If you need to do this, enter the most restrictive consist in terms of weight, length and maximum speed."
+        }
+      ],
+      "subSections": [
+        {
+          "title": "Traction unit",
+          "content": [
+            {
+              "type": "text",
+              "value": "Enter the first few letters to find the locomotive(s) for your consist in the catalogue."
+            }
+          ]
+        },
+        {
+          "title": "Towed rolling stock",
+          "content": [
+            {
+              "type": "text",
+              "value": "In the same way as for the traction unit, enter the first few letters to find the equipment that will make up your consist."
+            }
+          ]
+        },
+        {
+          "title": "Weight, length, max. speed",
+          "content": [
+            {
+              "type": "text",
+              "value": "Indicate the total weight of your consist (in tons), the total length (in meters) and the maximum speed (in km/h)."
+            }
+          ]
+        },
+        {
+          "title": "Composition code",
+          "content": [
+            {
+              "type": "text",
+              "value": "Select your consist's composition code from the list."
+            }
+          ]
+        }
+      ]
+    },
+    "path": {
+      "title": "The route",
+      "content": [
+        {
+          "type": "text",
+          "value": "Indicate the origin, destination and any stopping points on your route (excluding turn-offs) to obtain a simulation."
+        },
+        {
+          "type": "text",
+          "value": "In the case of linked paths, you can also enter an previous or posterior path."
+        },
+        {
+          "type": "text",
+          "value": "The points you enter appear on the map as you enter them."
+        },
+        {
+          "type": "text",
+          "value": "If a warning message tells you that no path has been found for these staking points, check all the CIs and CHs on your route, as well as the compatibility of your consist with this route."
+        }
+      ],
+      "subSections": [
+        {
+          "title": "Previous path",
+          "content": [
+            {
+              "type": "text",
+              "value": "As an option, the previous path allows you to indicate a path linked upstream of your simulation. Simply search for its number and date of circulation, then click on \"Find\"."
+            },
+            {
+              "type": "text",
+              "value": "If several paths match your search, select the one you want from the list."
+            },
+            {
+              "type": "text",
+              "value": "Once you have selected your train path, the origin information is pre-filled with a 30-minute margin on the departure time. All fields can still be modified."
+            }
+          ]
+        },
+        {
+          "title": "Origin",
+          "content": [
+            {
+              "type": "text",
+              "value": "Enter the CI code using a text search on the name or trigram of your desired point of origin. A CH will appear in the CH selector. Select the correct CH code from the list."
+            },
+            {
+              "type": "text",
+              "value": "Enter the start date and time using the mouse or keyboard, then specify the tolerance before and after your start time (in minutes)."
+            },
+            {
+              "type": "text",
+              "value": "If you want to impose a timetable on the destination rather than the origin, replace ‘precise timetable’ with ‘respect destination time’, and enter the times in the destination block."
+            }
+          ]
+        },
+        {
+          "title": "Intermediate OP",
+          "content": [
+            {
+              "type": "text",
+              "value": "Intermediate OPs are optional and allow you to impose a via or stop at a specific point. Simply enter the CI and CH codes as for the origin, then select the type of staking from the three options: via, driver changeover, service stop."
+            },
+            {
+              "type": "text",
+              "value": "In the case of driver changeover or service stop, enter the minimum stop time required, in minutes. To find more solutions, the tool can extend these stops in the simulation."
+            },
+            {
+              "type": "text",
+              "value": "N.B.: Turnarounds are not yet taken into account in the app."
+            }
+          ]
+        },
+        {
+          "title": "Destination",
+          "content": [
+            {
+              "type": "text",
+              "value": "As for the origin, enter the CI and CH codes."
+            },
+            {
+              "type": "text",
+              "value": "If you wish to impose a timetable on the destination rather than the origin, replace ‘as soon as possible’ with ‘precise timetable’. The <code>Date</code>, <code>Time</code> and <code>Tolerance</code> fields then appear and can be filled in as for the origin."
+            },
+            {
+              "type": "text",
+              "value": "N.B.: You must enter either an origin time or a destination timetable, not both."
+            }
+          ]
+        },
+        {
+          "title": "Posterior path",
+          "content": [
+            {
+              "type": "text",
+              "value": "Optional, it works in the same way as the previous path but acts on the destination rather than the origin."
+            }
+          ]
+        }
+      ]
+    },
+    "result": {
+      "title": "The result of the calculation",
+      "content": [
+        {
+          "type": "text",
+          "value": "Depending on the capacity available for your train at the times requested, the tool will suggest a simulated path, or inform you of the reasons why this is not possible."
+        }
+      ],
+      "subSections": [
+        {
+          "title": "Path simulation",
+          "content": [
+            {
+              "type": "text",
+              "value": "If a path has been found, a table summarising the calculated timetables is displayed. Initially, only the requested milestones are shown, but you can extend this list by clicking on ‘Show all milestones’. The route is also shown on a map."
+            },
+            {
+              "type": "text",
+              "value": "If you are happy with the simulated path, click on ‘Keep this simulation’. A new button will appear allowing you to download a simulation form."
+            },
+            {
+              "type": "text",
+              "value": "If the simulated path does not suit you, go back to the form and adjust the parameters of your request to run a new simulation. This will then appear next to the previous one once the calculation has been completed."
+            }
+          ]
+        },
+        {
+          "title": "No capacity",
+          "content": [
+            {
+              "type": "text",
+              "value": "If no train path has been found, you will be informed of the various conflicts encountered with other train paths or work. Here again, you can go back to the form to adjust the parameters of your request and run a simulation again."
+            }
+          ]
+        }
+      ]
+    },
+    "glossary": {
+      "title": "Glossary",
+      "content": [
+        {
+          "type": "definitions",
+          "definitions": [
+            {
+              "term": "CI",
+              "definition": "Permanent Code"
+            },
+            {
+              "term": "CH",
+              "definition": "Worksite Code"
+            },
+            {
+              "term": "LMR",
+              "definition": "Last Minute Request"
+            },
+            {
+              "term": "OP",
+              "definition": "Operational Point"
+            }
+          ]
+        }
+      ]
+    },
+    "faq": {
+      "title": "Frequently asked questions",
+      "content": [
+        {
+          "type": "faq",
+          "questions": [
+            {
+              "question": "What is a simulation sheet?",
+              "answer": "A simulation sheet is a document presenting the simulation information as well as the timetable of the proposal made by LMR."
+            },
+            {
+              "question": "LMR is not offering me a train path, why?",
+              "answer": "LMR does not create capacity. The tool only proposes a path according to your needs and the residual capacity available at the time of your simulation."
+            },
+            {
+              "question": "When I run my simulation, is the capacity blocked?",
+              "answer": "An LMR simulation does not block capacity. The simulation only allows the user to know what capacity is available."
+            },
+            {
+              "question": "Is it possible to make requests for turnarounds in the app?",
+              "answer": "LMR does not allow requests for turnarounds. These developments will nevertheless be taken into account for future versions of the app."
+            },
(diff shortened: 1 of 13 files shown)


front/src/applications/stdcm/components/StdcmHelpModule/StdcmHelpModule.tsx
```diff
@@ -0,0 +1,81 @@
...
+        </div>
+      </footer>
+      {sections.map((section, index) => (
+        <HelpSection
+          section={section}
+          closeHelpSection={closeHelpSection}
+          isActive={section === activeSection}
+          key={index}
+        />
+      ))}
```
Uriel-Sautron:
```suggestion
     {activeSection && (
        <HelpSection section={activeSection} closeHelpSection={closeHelpSection} isActive />
      )}
```

front/src/applications/stdcm/views/StdcmView.tsx
```diff
@@ -206,6 +215,7 @@ const StdcmView = () => {
               )}
             </div>
           )}
+          <StdcmHelpModule showHelpModule={showHelpModule} toggleHelpModule={toggleHelpModule} />
```
Uriel-Sautron, line 218:
```suggestion
          {showHelpModule && (
            <StdcmHelpModule showHelpModule={showHelpModule} toggleHelpModule={toggleHelpModule} />
          )}
```

Uriel-Sautron: Prevents component rendering if not used.

front/src/applications/stdcm/components/StdcmHelpModule/StdcmHelpModule.tsx
Uriel-Sautron: Just a suggestion, it might avoid rendering unused sections.

Uriel-Sautron (approved):
Nice job !!! Tested, pleasant and smooth behavior.
I have left some comments.

front/src/applications/stdcm/components/StdcmHelpModule/StdcmHelpModule.tsx
Math-R: Module and section can't have smooth slide effect if not rendered before be cause thay will render being already active and be unmount so they will not slide out but just snap

front/src/applications/stdcm/views/StdcmView.tsx
Math-R: same as above

Uriel-Sautron (approved):
LGTM !!!

emersion:
In the mockup, there is a paragraph below the main "Help" heading:

out

Is the mockup outdated, or is the paragraph missing?

front/src/styles/scss/applications/stdcm/_helpModule.scss
```diff
@@ -0,0 +1,142 @@
...
+  top: 0;
+  right: -100%;
+  height: 100vh;
+  width: 520px;
+  background-color: var(--white100);
+  box-shadow:
+    0px 6px 21px -5px rgba(24, 68, 239, 0.26),
+    0px 16px 30px -5px rgba(0, 0, 0, 0.16),
+    0px 3px 5px -2px rgba(0, 0, 0, 0.1);
+  opacity: 1;
```
emersion: I think most of these `opacity: 1` rules can be dropped since it's the default value and never set to anything else.

front/src/styles/scss/applications/stdcm/_helpModule.scss
```diff
@@ -0,0 +1,142 @@
...
+  &__close {
+    position: absolute;
+    top: 24px;
+    right: 24px;
+    cursor: pointer;
+  }
+  &__title {
+    opacity: 1;
+    color: rgba(0, 0, 0, 1);
+    font-style: SemiBold;
```
emersion: "SemiBold" is not a valid value for [`font-style`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-style). Reading `_fonts.scss`, I believe the `font-weight` should be enough to select the right font.

front/src/applications/stdcm/views/StdcmView.tsx
```diff
@@ -65,6 +67,8 @@ const StdcmView = () => {
     dispatch(resetStdcmConfig());
   };
 
+  const toggleHelpModule = () => setShowHelpModule(!showHelpModule);
```
emersion: Nit: it doesn't really matter here, but in general it's unreliable to read the previous state like this when setting some new state. Indeed, the state updates are put in a queue so we might be working with outdated state here. An update function fixes these issues: `setShowHelpModule((show) => !show)`. See https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state

front/src/applications/stdcm/components/StdcmHelpModule/types.ts
```diff
@@ -0,0 +1,38 @@
+export type Section = {
+  id: string;
+  title: string;
+  content?: Content[];
+  subSections?: Omit<Section, 'subSection'>[];
```
emersion: Typo: `'subSections'` (plural)

front/src/applications/stdcm/components/StdcmHelpModule/SectionContentManager.tsx
```diff
@@ -0,0 +1,54 @@
...
+                  </li>
+                ))}
+            </ol>
+          );
+        case 'definitions':
+          return (
+            <table>
+              <tbody>
+                {content.definitions.map((item) => (
+                  <tr key={`${item.term}`}>
```
emersion: Nit: the string template is unnecessary here, can be simplified to `key={item.term}`.

front/src/styles/scss/applications/stdcm/_helpModule.scss
```diff
@@ -0,0 +1,142 @@
+.stdcm__help-module {
```
emersion, line 1: Random side note: I'm personally not a huge fan of this CSS naming convention.

front/src/styles/scss/applications/stdcm/_header.scss
```diff
@@ -42,37 +43,39 @@
...
-    margin-right: 48px;
+  .stdcm-header__debug {
     button:focus {
       outline: none;
     }
     button {
-      height: 40px;
-      width: 40px;
-      border-radius: 6px;
+      height: 2.5rem;
```
emersion: I think we should retain `px` values for these (applies to the whole file).

front/src/applications/stdcm/components/StdcmHelpModule/StdcmHelpModule.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const { t } = useTranslation('stdcm-help-section');
+
+  const [activeSection, setActiveSection] = useState<string | null>(null);
+
+  const closeHelpModule = () => {
+    setActiveSection(null);
+    toggleHelpModule();
+  };
+  const closeHelpSection = () => setActiveSection(null);
+  const support = t('asu', { returnObjects: true }) as Array<Record<string, string>>;
```
emersion: I think we can more precisely define the type here:

```ts
t('asu', { returnObjects: true }) as { title: string; value: string }[]
```

front/src/applications/stdcm/components/StdcmHelpModule/StdcmHelpModule.tsx
```diff
@@ -0,0 +1,84 @@
...
+      </div>
+      <footer>
+        {Array.isArray(support) &&
+          support.map((item, index) => (
+            <>
+              <div className="support-info">
+                <div className="support-info__title">{item.title}</div>
+                <div className="support-info__content">{item.value}</div>
+              </div>
+              {index !== support.length - 1 && <hr />}
```
emersion, line 57: Nit: I'd prefer to use CSS to display a separator between items, because it's a purely cosmetic element. `:not(:last_child)` can be used to skip the separator for the last element.

front/src/applications/stdcm/components/StdcmHelpModule/StdcmHelpModule.tsx
```diff
@@ -0,0 +1,84 @@
...
+    <div className={cx('stdcm__help-module', { active: showHelpModule })}>
+      <div className="stdcm__help-module__header">
+        <button type="button" className="stdcm__help-module__close" onClick={closeHelpModule}>
+          <X size="lg" iconColor="#B6B2AF" />
+        </button>
+      </div>
+      <div className="stdcm__help-module__content">
+        <h1 className="stdcm__help-module__title">{t('help')}</h1>
+        <div className="stdcm__help-module__chapters">
+          {sections &&
```
emersion: I don't think `sections` can ever be falsy?

front/src/applications/stdcm/components/StdcmHelpModule/StdcmHelpModule.tsx
```diff
@@ -0,0 +1,84 @@
...
+            sections.map((section, index) => (
+              <>
+                <button
+                  type="button"
+                  key={index}
```
emersion: Nit: doesn't really matter since the list items are static, but `section` would be slightly better to as key use here.

front/src/applications/stdcm/components/StdcmHelpModule/StdcmHelpModule.tsx
```diff
@@ -0,0 +1,84 @@
...
+                    <ChevronRight iconColor="#B6B2AF" />
+                  </div>
+                </button>
+                {index !== sections.length - 1 && <hr />}
+              </>
+            ))}
+        </div>
+      </div>
+      <footer>
+        {Array.isArray(support) &&
```
emersion, line 50: Can this ever be something else?

front/src/applications/stdcm/components/StdcmHelpModule/StdcmHelpModule.tsx
```diff
@@ -0,0 +1,84 @@
...
+            {t('asuLink')}
+          </a>
+        </div>
+      </footer>
+      {sections.map((section, index) => (
+        <HelpSection
+          section={section}
+          closeHelpSection={closeHelpSection}
+          isActive={section === activeSection}
+          key={index}
```
emersion: Ditto

front/src/applications/stdcm/components/StdcmHelpModule/StdcmHelpModule.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  const closeHelpModule = () => {
+    setActiveSection(null);
+    toggleHelpModule();
+  };
+  const closeHelpSection = () => setActiveSection(null);
+  const support = t('asu', { returnObjects: true }) as Array<Record<string, string>>;
+  const sections = Object.keys(t('sections', { returnObjects: true }));
+  return (
+    <div className={cx('stdcm__help-module', { active: showHelpModule })}>
```
emersion, line 27: I was wondering why we use a class here instead of omitting the element from the DOM when inactive. I take it it's to make CSS transitions work?

front/src/applications/stdcm/components/StdcmHelpModule/StdcmHelpModule.tsx
```diff
@@ -0,0 +1,84 @@
...
+      <div className="stdcm__help-module__header">
+        <button type="button" className="stdcm__help-module__close" onClick={closeHelpModule}>
+          <X size="lg" iconColor="#B6B2AF" />
```
emersion: Instead of hardcoding the color here, I think we can set the `color` property of the outer `<button>`. That way, we can use a CSS variable.

(Applies to other icons as well.)

front/src/applications/stdcm/components/StdcmHelpModule/StdcmHelpModule.tsx
```diff
@@ -0,0 +1,84 @@
...
+          {sections &&
+            sections.map((section, index) => (
+              <>
+                <button
+                  type="button"
+                  key={index}
+                  className="stdcm__help-module__title-button flex"
+                  onClick={() => setActiveSection(section)}
+                >
+                  <div className="">{t(`sections.${section}.title`)}</div>
```
emersion: Nit: empty `className` attribute can be removed.

front/src/applications/stdcm/components/StdcmHelpModule/SectionContentManager.tsx
```diff
@@ -0,0 +1,54 @@
...
+};
+
+const SectionContentManager = ({ content }: SectionContentManagerProps) => (
+  <div className="section__content">
+    {(() => {
+      switch (content.type) {
+        case 'list':
+          return (
+            <ul>
+              {content.items &&
```
emersion: I don't think `content.items` can be falsy?

front/src/applications/stdcm/components/StdcmHelpModule/SectionContentManager.tsx
```diff
@@ -0,0 +1,54 @@
...
+            </ul>
+          );
+
+        case 'text':
+          // eslint-disable-next-line react/no-danger
+          return <p dangerouslySetInnerHTML={{ __html: content.value }} />;
+        case 'faq':
+          return (
+            <ol>
+              {content.questions &&
```
emersion, line 26: I don't think `content.questions` can be falsy?

front/src/applications/stdcm/components/StdcmHelpModule/SectionContentManager.tsx
```diff
@@ -0,0 +1,54 @@
...
+                content.questions.map((question, index) => (
+                  <li key={index}>
+                    <h3>{`${index + 1}. ${question.question}`}</h3>
```
emersion: Nit: can be simplified to `<h3>{index + 1}. {question.question}</h3>`

front/src/applications/stdcm/components/StdcmHelpModule/SectionContentManager.tsx
```diff
@@ -0,0 +1,54 @@
...
+                    <p> {question.answer}</p>
```
emersion: Nit: extra space

front/public/locales/en/stdcm-help-section.json
emersion, line 1: I've been wondering whether it would make more sense to just use a Markdown or HTML file instead of re-inventing markup in JSON...

An intermediate solution would be to embed Markdown/HTML inside JSON, with one array string per line, and then use `t('foo', { joinArrays: '\n' })`.

front/src/applications/stdcm/components/StdcmHelpModule/types.ts
```diff
@@ -0,0 +1,38 @@
...
+};
+
+export type Content = ListContent | TextContent | FAQContent | GlossaryContent;
+
+export type ListContent = {
+  type: 'list';
+  items: Record<string, string>[];
```
emersion: It seems like the only accepted field is `value`, so this can be tightened to

```ts
items: { value: string }[];
```

But at that point we might as well just use strings without the wrapper?

```ts
items: string[];
```

emersion (commented):
Nice! Here are a few comments.

The result doesn't _exactly_ look like the mockup, e.g. some margins are a bit different. However that simplifies the code so I'm inclined to just ignore these differences.

Math-R:
> In the mockup, there is a paragraph below the main "Help" heading:
> 
> out
> 
> Is the mockup outdated, or is the paragraph missing?

The mockup is outdated, @axrolld asks me to not put this paragraph

front/src/styles/scss/applications/stdcm/_helpModule.scss
Math-R: Sorry for that i think i just copy paste from the sketch

front/src/styles/scss/applications/stdcm/_helpModule.scss
Math-R: Oh me neither, but i found some line where you did the same in _stdcm.scss it's why i used bem here, i could completely change it for more concise and easy naming so :D

front/src/applications/stdcm/components/StdcmHelpModule/StdcmHelpModule.tsx
Math-R: Yes it is, if there not present they will appear as already active and not slide-in out

front/src/applications/stdcm/components/StdcmHelpModule/StdcmHelpModule.tsx
Math-R: I put it because typescript was furious, if you find something else i can make a change but i didn't on my side

front/src/applications/stdcm/components/StdcmHelpModule/StdcmHelpModule.tsx
Math-R: I can buy this, but `hr` is also \<hr>: The Thematic Break (Horizontal Rule) element] =>  https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr ,   maybe we can discuss this later  and put a new rule to our contribution guide?

front/src/styles/scss/applications/stdcm/_helpModule.scss
Math-R: sure thanks

front/src/applications/stdcm/components/StdcmHelpModule/SectionContentManager.tsx
Math-R: not nit, I hate this one 😆 thanks !

front/public/locales/en/stdcm-help-section.json
Math-R: I think that could have been a solution too but we would have had to use one more dependency to render the markdown right ?

front/src/applications/stdcm/components/StdcmHelpModule/SectionContentManager.tsx
Math-R: i probably change something during the development but you're right it not useful anymore

front/public/locales/en/stdcm-help-section.json
emersion: We already depend on `react-markdown`.

front/src/applications/stdcm/components/StdcmHelpModule/SectionContentManager.tsx
Math-R: same : i probably change something during the development but you're right it not useful anymore

front/src/applications/stdcm/components/StdcmHelpModule/StdcmHelpModule.tsx
Math-R: yes true, thanks !

front/src/applications/stdcm/components/StdcmHelpModule/StdcmHelpModule.tsx
Math-R: yes thanks

front/src/applications/stdcm/views/StdcmView.tsx
Math-R: Interesting thanks !

front/src/applications/stdcm/components/StdcmHelpModule/types.ts
Math-R: agreed with the first point, but for the second i think i would just make the json less consistent and readable, because all the other type of item works with key: value

front/public/locales/en/stdcm-help-section.json
Math-R: Sure but in this case it's not useful to use it since it just for rendering that can be achieve with native html, plus markdown render html.

front/src/styles/scss/applications/stdcm/_helpModule.scss
SharglutDev: As it's at a lot of places in our .scss files, it can be changed later if we want to.

front/src/applications/stdcm/components/StdcmHelpModule/StdcmHelpModule.tsx
SharglutDev: App crashes if we remove this, it's a little strange...

SharglutDev (approved):
Lgtm and tested, good work !

PR #9984, opened 2024-12-09
front: lmr: reduce grid margins from 35s to 15s

### LINKED ISSUE #9946: front: stdcm: reduce grid margins from 35s to 15s
Grid margins are margins we add before and after each occupations, they're also referenced as `time_gap_before` and `time_gap_after` in the code.

The front-end sets this value, there's no default value in the back. It can only be configured with debug mode.

Apparently it's not *really* useful, and we're looking for ways to improve the positive response rate. We should *probably* keep a few seconds to avoid simulation inaccuracies.

The default value should be updated in the debug mode elements, and when sending a request without debug mode on.

Image

**EDIT**: apparently we still need *some* margin, 15s seems like a better value

SarahBellaha:
Closes #9946


clarani (approved):
LGTM

RomainValls (approved):
LGTM

Maymanaf (approved):
LGTM for e2e tests

PR #9985, opened 2024-12-09
editoast: change stdcm environment date type

### LINKED ISSUE #9974: editoast: search environment's dates have no timezone
### What happened?

In the frontend, when we get the search environment dates, they have no timezone but it's not normal.

The dates should always have their timezone.

Open a stdcm page with the network opened, check the environment search returned in the network

Recette (SNCF)

Firefox

.

SarahBellaha:
Closes #9974

### DIFF
--- a/editoast/src/models/stdcm_search_environment.rs
+++ b/editoast/src/models/stdcm_search_environment.rs
@@ -1,4 +1,5 @@
-use chrono::NaiveDateTime;
+use chrono::DateTime;
+use chrono::Utc;
 use diesel::ExpressionMethods;
 use diesel::QueryDsl;
 use diesel_async::RunQueryDsl;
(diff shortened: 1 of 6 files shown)


editoast/src/models/stdcm_search_environment.rs
```diff
@@ -127,11 +128,31 @@ pub mod tests {
...
             .temporary_speed_limit_group_id(Some(temporary_speed_limit_group.id))
             .timetable_id(timetable.id)
-            .search_window_begin(NaiveDate::from_ymd_opt(2024, 1, 1).unwrap().into())
-            .search_window_end(NaiveDate::from_ymd_opt(2024, 1, 15).unwrap().into());
+            .search_window_begin(
+                NaiveDate::from_ymd_opt(2024, 1, 1)
+                    .unwrap()
+                    .and_hms_opt(0, 0, 0)
+                    .unwrap()
+                    .and_utc(),
```
woshilapin: You can replace by this.

```suggestion
                 DateTime::parse_from_rfc3339("2024-01-01T00:00:00Z").unwrap().to_utc()),
```

woshilapin (approved):
Thank you very much, this really improves the entire OSRD stack!

editoast/src/models/stdcm_search_environment.rs
woshilapin: Oh, you found an even better API for that 😍 I’m resolving.

clarani (approved):
LGTM :white_check_mark:

Yohh (approved):

PR #9986, opened 2024-12-09
Fix accents support in `x-remote-user-name`

leovalais:


flomonster (approved):
LGTM

editoast/src/views/mod.rs
```diff
@@ -201,14 +202,12 @@ async fn authenticate(
         }
         return Ok(Authentication::Unauthenticated);
     };
-    let identity = identity
-        .to_str()
+    let identity = str::from_utf8(identity.as_bytes())
         .expect("unexpected non-ascii characters in x-remote-user-identity");
```
woshilapin, line 206:
```suggestion
        .expect("unexpected non-utf8 characters in x-remote-user-identity");
```

PR #9987, opened 2024-12-09
lmr: putting ma100 as default value in the speedlimittag selector

### LINKED ISSUE #9584: Make composition code mandatory with default `MA100` in LMR
# Description

Make composition code mandatory with default MA100 in LMR

# Goal

Save some time to our users

# AC

- The default value of this component is now `MA100`
- `- Aucun -` has been removed from the list
Image

Caracol3:
closes #9584 
Edit: 'No speedLimit tag' should only be displayed in debug mode on stdcm page


front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
```diff
@@ -25,6 +25,13 @@ export default function SpeedLimitByTagSelector({
   className = '',
 }: SpeedLimitByTagSelectorProps) {
   const { t } = useTranslation(['operationalStudies/manageTrainSchedule']);
+  const defaultTag = 'MA100';
```
Math-R: I don't think that `MA100` value il related to the component  but rather to the use we make of it.
Maybe you can put it in an external consts file as `DEFAULT_TAG` and import the value. It will be less `hardcoded`

Caracol3: Yes, good catch. I will change this 👌

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
```diff
@@ -1,11 +1,12 @@
-import { useMemo } from 'react';
+import { useEffect, useMemo } from 'react';
 
 import { Select } from '@osrd-project/ui-core';
 import cx from 'classnames';
 import { isEmpty } from 'lodash';
 import { useTranslation } from 'react-i18next';
 
 import { createStringSelectOptions } from 'utils/uiCoreHelpers';
+import { DEFAULT_COMPOSITION_CODE } from 'applications/stdcm/consts';
```
Akctarus: Got a little eslint warning : 
`applications/stdcm/consts` import should occur before import of `utils/uiCoreHelpers`

Akctarus (approved):
Tested, I left a comment but apart that, LGTM ! 
I approve the PR and let you make the quick fix

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
clarani, line 1: I think this component is also used in the scenario page right ? But I'm not sure we want to set the speed limit tag by default to MA100 in the scenario page.

I think the best way to do it would be:
- initialize the stdcm store with the MA100 speed limit tag
- in the stdcm page, if no speed limit tag exists, dispatch the MA100 speed limit tag

This way, you don't have to modify the component which is shared between the stdcm and scenario pages ;)

clarani (changes requested):

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
Caracol3: done

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
Caracol3: I removed the import so we don't have this warning anymore

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
Caracol3: done

front/src/common/SpeedLimitByTagSelector/SpeedLimitByTagSelector.tsx
```diff
@@ -48,7 +48,6 @@ export default function SpeedLimitByTagSelector({
           id="speed-limit-by-tag-selector"
           value={speedLimitByTag || ''}
           label={t('speedLimitByTagAbbrev')}
-          placeholder={t('noSpeedLimitByTag')}
```
clarani, line 51: Why do you remove this placeholder ?

Caracol3: It's a mistake, I removed it without realizing it also affected operational studies. I have put it back.

front/src/common/SpeedLimitByTagSelector/useStoreDataForSpeedLimitByTagSelector.ts
```diff
@@ -45,6 +45,12 @@ export const useStoreDataForSpeedLimitByTagSelector = ({ isStdcm } = { isStdcm:
     }
   }, [error]);
 
+  useEffect(() => {
+    if (isStdcm && !speedLimitByTag) {
+      dispatchUpdateSpeedLimitByTag(DEFAULT_COMPOSITION_CODE);
+    }
+  }, [isStdcm, speedLimitByTag, dispatchUpdateSpeedLimitByTag]);
```
clarani: I think speedLimitByTag should not be a dependence of the useEffect, what if I select the tag 'no speed limit' ? Then the useEffect will set it to MA100.

I'm not sure it's the intended behaviour
```suggestion
  }, [isStdcm, dispatchUpdateSpeedLimitByTag]);
```

Caracol3: Good catch ! I deleted it.

front/tests/pages/stdcm-page-model.ts
```diff
@@ -260,7 +260,7 @@ class STDCMPage {
       this.destinationChField,
     ];
     for (const field of emptyFields) await expect(field).toHaveValue('');
-    await expect(this.codeCompoField).toHaveValue('__PLACEHOLDER__');
+    await expect(this.codeCompoField).toHaveValue('MA100');
```
Yohh: the function is named `verifyAllFieldsEmpty()`, if this field has a default value it can't be empty, you should rename the function, maybe add a comment to be more explicit

Yohh (commented):
just to clean a little: can you rename the locator `codeCompoField` to `speedLimitTagField`

front/tests/pages/stdcm-page-model.ts
Caracol3: Indeed ! I will change that

Math-R (approved):

Yohh (approved):

clarani (changes requested):
I still have this bug, can you reproduce it too ?
[Capture vidéo du 2024-12-19 14-08-47.webm](https://github.com/user-attachments/assets/8be64986-dcd7-4151-ae68-9f851b40cc88)

clarani (approved):
LGTM :white_check_mark:

front/tests/pages/stdcm-page-model.ts
```diff
@@ -352,11 +353,11 @@ class STDCMPage {
       this.originChField,
       this.destinationChField,
     ];
-    const { arrivalDate, arrivalTime, tolerance, speedLimitTag } = DEFAULT_DETAILS;
+    const { arrivalDate, arrivalTime, tolerance } = DEFAULT_DETAILS;
     for (const field of emptyFields) await expect(field).toHaveValue('');
     await expect(this.originArrival).toHaveValue(ORIGIN_DETAILS.arrivalType.default);
     await expect(this.destinationArrival).toHaveValue(DESTINATION_DETAILS.arrivalType.default);
-    await expect(this.speedLimitTagField).toHaveValue(speedLimitTag);
+    await expect(this.speedLimitTagField).toHaveValue('MA100');
```
Maymanaf: you should update the value of `speedLimitTag` to `'MA100'`

front/tests/pages/stdcm-page-model.ts
```diff
@@ -413,10 +414,10 @@ class STDCMPage {
       await dropdownField.dispatchEvent('blur');
       await expect(dropdownField).toHaveValue(selectedValue);
 
-      const { expectedTonnage, expectedLength, expectedMaxSpeed } = expectedValues;
+      const { expectedTonnage, expectedLength } = expectedValues;
       await expect(this.tonnageField).toHaveValue(expectedTonnage);
       await expect(this.lengthField).toHaveValue(expectedLength);
-      await expect(this.maxSpeedField).toHaveValue(expectedMaxSpeed);
+      await expect(this.maxSpeedField).toHaveValue('100');
```
Maymanaf: Avoid magic numbers !
Add a condition in `calculateTowedPrefilledValues` to ensure that `speedLimitTagField` is not empty before proceeding with the calculation

front/tests/pages/stdcm-page-model.ts
```diff
@@ -448,7 +449,7 @@ class STDCMPage {
       {
         expectedTonnage: tractionEngineTonnage,
         expectedLength: tractionEngineLength,
-        expectedMaxSpeed: tractionEngineMaxSpeed,
+        expectedMaxSpeed: '100',
```
Maymanaf: magic number!

Maymanaf (approved):
LGTM

PR #9988, opened 2024-12-09
doc, github: suggest propagation and bump when interfaces changes

bougue-pe:
automatically add a PR comment to suggest check: part of https://github.com/osrd-project/osrd-dags/pull/596#issuecomment-2517776699

part of #11584

### DIFF
--- a/.github/workflows/interface_change.yml
+++ b/.github/workflows/interface_change.yml
@@ -1,0 +1,23 @@
+name: interface_change
+on:
+  pull_request:
+    paths:
+      - python/osrd_schemas
+      - editoast/openapi.yml
+jobs:
+  comment:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v7
+        with:
+          script: |
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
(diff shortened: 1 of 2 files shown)


Khoyo:
Could you open a real issue for this PR comment?

As for the PR, there are really two parts to osrd-schemas interface: the API (used directly by some projects but not by OSRD itself), and the RailJSON specification (used by the front editor forms generator). There are also no checks that the editoast schemas definitions are in sync with the osrd-schemas one.

So I believe we should:

- Specify how to bump the RJS version number (it's not semver today. Almost all changes are more or less breaking changes but did not provoke a major version bump)
- Add a check on `front/src/reducers/osrdconf/infra_schema.json` (which the CI guarantees will stay in sync with osrd-schemas) to ask to check for a potential RJS bump
- Add a reminder to check for osrd_schemas and editoast_schemas sync (automating it would be a must, but I realize that's a whole other endeavor).

I did not do a technical review, will do tommorrow

bougue-pe:
After discussion with @woshilapin an issue was created to regroup ideas: #11584

bougue-pe:
Superseded by #11581 (and the rest is tracked in #11584).

PR #9999, opened 2024-12-10
front: adjusting map buttons

theocrsb:

### DIFF
--- a/front/src/styles/scss/common/components/_tipped.scss
+++ b/front/src/styles/scss/common/components/_tipped.scss
@@ -4,6 +4,10 @@
   & > .tooltip {
     position: absolute;
 
+    .tooltip-inner {
+      font-family: 'IBM Plex Sans';
+    }
+
     // Layout
     &.bs-tooltip-left,
     &.bs-tooltip-right,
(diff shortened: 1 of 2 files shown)


Caracol3:
@theocrsb Just to confirm, on this PR there are only the tooltip changes? The infrastructure style modifications is postponed, right ?

emersion:
Hm, seems like the top margin isn't 16px (seems to be 12px)…

out

Also I'd suggest changing `new-btn-map` to not have a top margin at all, only a bottom margin. That way we can apply a fixed 16px margin to its container without having to account for the inner one.

emersion (commented):
Hm, seems like the top margin isn't 16px (seems to be 12px)…

out

Also I'd suggest changing `new-btn-map` to not have a top margin at all, only a bottom margin. That way we can apply a fixed 16px margin to its container without having to account for the inner one.

theocrsb:
> @theocrsb Just to confirm, on this PR there are only the tooltip changes? The infrastructure style modifications is postponed, right ?

yes, the font and spacing

theocrsb:
> Hm, seems like the top margin isn't 16px (seems to be 12px)…
> 
> out
> 
> Also I'd suggest changing `new-btn-map` to not have a top margin at all, only a bottom margin. That way we can apply a fixed 16px margin to its container without having to account for the inner one.

It's done :)

emersion (approved):
Looks good!

PR #10004, opened 2024-12-10
front: handle uncaught exceptions for e2e tests

### LINKED ISSUE #8883: front: log and make E2E tests fail on uncaught exceptions
### Description and goal

Right now when an E2E test fails we don't have lots of details. When it trips on an exception we instead get a timeout and no details.

See:

- https://playwright.dev/docs/api/class-testconfig#test-config-global-setup
- https://playwright.dev/docs/api/class-test#test-extend
- https://playwright.dev/docs/api/class-page/#page-event-page-error
- https://github.com/microsoft/playwright/issues/16355#issuecomment-2082933441
- https://github.com/microsoft/playwright/issues/27277#issuecomment-1734437655
- https://github.com/microsoft/playwright/issues/5546#issuecomment-783425772

### Acceptance criteria

.

Maymanaf:
Closes #8883

Only test 008 is excluded from this PR due to multiple exceptions causing it to fail. We're waiting for fixes from issues #11257 and #11256


emersion (dismissed):

Yohh (dismissed):

front/tests/pages/stdcm/via-section.ts
```diff
@@ -117,6 +120,7 @@ class ViaSection extends STDCMPage {
         await this.getViaType(viaNumber).selectOption(SERVICE_STOP);
         await expect(this.getViaStopTime(viaNumber)).toHaveValue(serviceStop.default);
         await this.getViaStopTime(viaNumber).fill(serviceStop.input);
+        await this.page.waitForTimeout(EXPLICIT_UI_STABILITY_TIMEOUT); // Wait for the animation to complete
```
emersion: Could we replace this with `scrollIntoViewIfNeeded()`, which waits for the element to be stable? Or listen for the `animationend` event?

Yohh (approved):

front/src/applications/operationalStudies/views/SimulationResults.tsx
```diff
@@ -159,7 +159,7 @@ const SimulationResults = ({
   }
 
   return (
-    <div className="simulation-results">
+    <div data-testid="simulation-results" className="simulation-results">
```
SharglutDev: Is this data-testid used ?

front/tests/008-timetable-item.spec.ts
SharglutDev, line 1: For the second commit "fix timetable test", would you mind to add a description in the commit with the reason why they needed a fix, it's not really clear for me

front/tests/014-stdcm-linked-train.spec.ts
```diff
@@ -101,19 +101,19 @@ test.describe('Verify stdcm simulation page', () => {
 
   /** *************** Test 2 **************** */
   test('Verify STDCM posterior linked train', async ({ page: _ }, testInfo) => {
+    await originSection.fillOriginDetailsLight(undefined, 'respectDestinationSchedule', true);
+    await linkedTrainSection.posteriorLinkedPathDetails();
+    await viaSection.fillAndVerifyViaDetails({
+      viaNumber: 1,
+      ciSearchText: 'mid_east',
+    });
```
SharglutDev, line 111: Same as the previous comment. A description in the 3rd commit of what was the bug and why the changes fixes it. Does moving this block of tests fixes something ? Maybe we should also leave a comment here in the code if it is so someone doesn't move it again later. 
I find it a bit not logic to test first the itinerary part and not the consist one. In most of the case, a user will first fill the consist form and then the itinerary one.

front/tests/002-project-management.spec.ts
SharglutDev, line 1: Also here an explanation fr the 4th commit of why it was broken with the previous method and not this one plz :)

front/src/applications/operationalStudies/utils.ts
```diff
@@ -79,26 +79,33 @@ export const transformBoundariesDataToPositionDataArray = <T extends 'gradient'
...
+    !Array.isArray(boundariesData.boundaries) ||
+    !Array.isArray(boundariesData.values)
+  ) {
+    return [];
+  }
+
+  const { boundaries, values } = boundariesData;
+
+  const formattedData: ElectricalRangesData<T>[] = boundaries.map((boundary, index) => ({
+    start: index === 0 ? 0 : mmToM(boundaries[index - 1]),
```
SharglutDev: Do these changes fixes the [bug](https://github.com/OpenRailAssociation/osrd/issues/11256) or just hide it ? If you found a explanation of the issue, could you update its description so its easier to fix ?

front/src/applications/operationalStudies/views/SimulationResults.tsx
Maymanaf: it's not but nothing wrong with having the data-testid 
should I remove it ?

front/src/applications/operationalStudies/utils.ts
Maymanaf: It doesn't fix the bug, but it does fix an uncaught exception caused by undefined values in `transformBoundariesDataToRangesDat`a

front/tests/008-timetable-item.spec.ts
Maymanaf: The commit handles some uncaught exceptions related to the timetable.
I will squash all the commits into the first one

front/tests/014-stdcm-linked-train.spec.ts
Maymanaf: The commit handles some uncaught exceptions related to stdcm
I will squash all the commits into the first one

SharglutDev (approved):
Lgtm and tested, thanks for the PR !

front/src/applications/operationalStudies/utils.ts
```diff
@@ -79,26 +79,33 @@ export const transformBoundariesDataToPositionDataArray = <T extends 'gradient'
...
+  boundariesData: ElectricalBoundariesData<T> | null | undefined,
   pathLength: number
 ): ElectricalRangesData<T>[] => {
   // TODO : remove electrical profiles
-  const formatedData = boundariesData.boundaries.map((boundary, index) => ({
-    start: index === 0 ? 0 : mmToM(boundariesData.boundaries[index - 1]),
+  if (
+    !boundariesData ||
+    !Array.isArray(boundariesData.boundaries) ||
+    !Array.isArray(boundariesData.values)
```
emersion: These two conditions can't be possible according to the type system.

front/src/applications/operationalStudies/utils.ts
```diff
@@ -79,26 +79,33 @@ export const transformBoundariesDataToPositionDataArray = <T extends 'gradient'
...
 >(
-  boundariesData: ElectricalBoundariesData<T>,
+  boundariesData: ElectricalBoundariesData<T> | null | undefined,
   pathLength: number
 ): ElectricalRangesData<T>[] => {
   // TODO : remove electrical profiles
-  const formatedData = boundariesData.boundaries.map((boundary, index) => ({
-    start: index === 0 ? 0 : mmToM(boundariesData.boundaries[index - 1]),
+  if (
+    !boundariesData ||
```
emersion: Would be cleaner to handle this in the caller: `transformBoundariesDataToRangesData(electrifications ?? [])`

emersion:
Sorry, had some more comments!

emersion (approved):
Thanks!

PR #10006, opened 2024-12-10
front: match PathStep and SuggestedOP with pathStepId

emersion:
Last episode: https://github.com/OpenRailAssociation/osrd/pull/9967

Matching via uic/trigram/ch/etc is unreliable because:

- Two OPs can have the same trigram/ch.
- The same OP could appear multiple times at different spots in a
  path.

Moreover, we had multiple variants of the matching with slight
differences. Some of the matching logic got adjusted to account
for specific bugs, while others were left alone.

Each path step can be identified with a unique ID stored in
PathStep.id. Use that instead.

(The refactoring has been split into small digestible commits to ease bisecting and make it more obvious why type and test adjustments are necessary.)

### DIFF
--- a/front/src/applications/stdcm/hooks/useStdcm.ts
+++ b/front/src/applications/stdcm/hooks/useStdcm.ts
@@ -99,7 +99,13 @@ const useStdcm = ({
       path: response.status === 'conflicts' ? response.pathfinding_result : response.path,
     } as StdcmResponse;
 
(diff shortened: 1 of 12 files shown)


SharglutDev:
Fyi : https://github.com/OpenRailAssociation/osrd/pull/10334#discussion_r1920020968

emersion:
Closing in favor of https://github.com/OpenRailAssociation/osrd/pull/17984

PR #10014, opened 2024-12-10
core: implement etcs braking simulator for EOAs

### LINKED ISSUE #9704: core: implement ETCS module interface to compute ETCS braking curves for EOAs

Erashin:
Restriction: in the end, only implement EOA part of the ETCS braking curves (without any behavior change for SVLs).

### DIFF
--- a/core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/EnvelopeSimPath.java
+++ b/core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/EnvelopeSimPath.java
@@ -106,6 +106,31 @@ public final class EnvelopeSimPath implements PhysicsPath {
         return (getCumGrade(end) - getCumGrade(begin)) / (end - begin);
     }
 
+    @Override
+    public double getMinGrade(double begin, double end) {
+        // TODO: Optimise method by adding in a cache.
+        int indexBegin = getIndexBeforePos(begin);
+        int indexEnd = getIndexBeforePos(end);
+        var lowestGradient = gradeValues[indexBegin];
+        for (int i = indexBegin; i < indexEnd; i++) {
+            var grad = gradeValues[i];
+            if (grad < lowestGradient) lowestGradient = grad;
+        }
+        return lowestGradient;
+    }
+
+    /** For a given position, return the index of the position just before in gradePositions */
+    public int getIndexBeforePos(double position) {
+        // TODO: Optimise method by using binary search.
+        if (position <= gradePositions[0]) return 0;
+        if (position >= gradePositions[gradePositions.length - 1]) return gradePositions.length - 1;
+        for (int i = 0; i < gradePositions.length; i++) {
+            var pos = gradePositions[i];
+            if (pos > position) return i - 1;
+        }
+        return gradePositions.length - 1;
+    }
+
     private RangeMap<Double, Electrification> getModeAndProfileMap(
             String powerClass, Range<Double> range, boolean ignoreElectricalProfiles) {
         if (ignoreElectricalProfiles) powerClass = null;
--- a/core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/PhysicsRollingStock.java
+++ b/core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/PhysicsRollingStock.java
@@ -1,5 +1,11 @@
 package fr.sncf.osrd.envelope_sim;
 
+import static fr.sncf.osrd.envelope_sim.TrainPhysicsIntegrator.GRAVITY_ACCELERATION;
+import static fr.sncf.osrd.envelope_sim.etcs.ConstantsKt.mRotatingMax;
+import static fr.sncf.osrd.envelope_sim.etcs.ConstantsKt.mRotatingMin;
+
+import fr.sncf.osrd.railjson.schema.rollingstock.RJSEtcsBrakeParams;
+
 public interface PhysicsRollingStock {
     /** The mass of the train, in kilograms */
     double getMass();
@@ -18,6 +24,8 @@ public interface PhysicsRollingStock {
 
     /** The first derivative of the resistance to movement at a given speed, in kg/s */
     double getRollingResistanceDeriv(double speed);
+
+    RJSEtcsBrakeParams getRJSEtcsBrakeParams();
 
     /** Get the effort the train can apply at a given speed, in newtons */
     static double getMaxEffort(double speed, TractiveEffortPoint[] tractiveEffortCurve) {
@@ -48,6 +56,16 @@ public interface PhysicsRollingStock {
         double coeff =
                 (previousPoint.maxEffort() - nextPoint.maxEffort()) / (previousPoint.speed() - nextPoint.speed());
         return previousPoint.maxEffort() + coeff * (Math.abs(speed) - previousPoint.speed());
+    }
+
+    /**
+     * The gradient acceleration of the rolling stock taking its rotating mass into account, in m/s².
+     * Grade is in m/km.
+     * mRotating (Max or Min) is in %, as seen in ERA braking curves simulation tool v5.1.
+     */
+    static double getGradientAcceleration(double grade) {
+        var mRotating = grade >= 0 ? mRotatingMax : mRotatingMin;
+        return -GRAVITY_ACCELERATION * grade / (1000.0 + 10.0 * mRotating);
     }
 
     /** The maximum constant deceleration, in m/s^2 */
--- a/core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/TrainPhysicsIntegrator.java
+++ b/core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/TrainPhysicsIntegrator.java
@@ -1,14 +1,21 @@
 package fr.sncf.osrd.envelope_sim;
 
+import static fr.sncf.osrd.envelope_sim.PhysicsRollingStock.getGradientAcceleration;
+import static fr.sncf.osrd.envelope_sim.PhysicsRollingStock.getMaxEffort;
+
 import com.google.common.collect.RangeMap;
+import fr.sncf.osrd.envelope_sim.etcs.BrakingType;
 
 /**
- * An utility class to help simulate the train, using numerical integration. It's used when
+ * A utility class to help simulate the train, using numerical integration. It's used when
  * simulating the train, and it is passed to speed controllers so they can take decisions about what
  * action to make. Once speed controllers took a decision, this same class is used to compute the
  * next position and speed of the train.
  */
 public final class TrainPhysicsIntegrator {
+    // Gravity acceleration, in m/s²
+    public static final double GRAVITY_ACCELERATION = 9.81;
+
     // A position delta lower than this value will be considered zero
     // Going back and forth with Distance and double (meters) often causes 1e-3 errors,
     // we need the tolerance to be higher than this
@@ -25,18 +32,21 @@ public final class TrainPhysicsIntegrator {
     private final double directionSign;
 
     private final RangeMap<Double, PhysicsRollingStock.TractiveEffortPoint[]> tractiveEffortCurveMap;
+    private final BrakingType brakingType;
 
     private TrainPhysicsIntegrator(
             PhysicsRollingStock rollingStock,
             PhysicsPath path,
             Action action,
             double directionSign,
-            RangeMap<Double, PhysicsRollingStock.TractiveEffortPoint[]> tractiveEffortCurveMap) {
+            RangeMap<Double, PhysicsRollingStock.TractiveEffortPoint[]> tractiveEffortCurveMap,
+            BrakingType brakingType) {
         this.rollingStock = rollingStock;
         this.path = path;
         this.action = action;
         this.directionSign = directionSign;
         this.tractiveEffortCurveMap = tractiveEffortCurveMap;
+        this.brakingType = brakingType;
     }
 
     /** Simulates train movement */
@@ -46,8 +56,19 @@ public final class TrainPhysicsIntegrator {
             double initialSpeed,
             Action action,
             double directionSign) {
+        return step(context, initialLocation, initialSpeed, action, directionSign, BrakingType.CONSTANT);
+    }
+
+    /** Simulates train movement */
+    public static IntegrationStep step(
+            EnvelopeSimContext context,
+            double initialLocation,
+            double initialSpeed,
+            Action action,
+            double directionSign,
+            BrakingType brakingType) {
         var integrator = new TrainPhysicsIntegrator(
-                context.rollingStock, context.path, action, directionSign, context.tractiveEffortCurveMap);
+                context.rollingStock, context.path, action, directionSign, context.tractiveEffortCurveMap, brakingType);
         return integrator.step(context.timeStep, initialLocation, initialSpeed, directionSign);
     }
 
@@ -65,16 +86,15 @@ public final class TrainPhysicsIntegrator {
     }
 
     private IntegrationStep step(double timeStep, double position, double speed) {
+        if (action == Action.BRAKE) return newtonStep(timeStep, speed, getDeceleration(speed, position), directionSign);
+
         double tractionForce = 0;
         var tractiveEffortCurve = tractiveEffortCurveMap.get(Math.min(Math.max(0, position), path.getLength()));
         assert tractiveEffortCurve != null;
-        double maxTractionForce = PhysicsRollingStock.getMaxEffort(speed, tractiveEffortCurve);
+        double maxTractionForce = getMaxEffort(speed, tractiveEffortCurve);
         double rollingResistance = rollingStock.getRollingResistance(speed);
-        double weightForce = getWeightForce(rollingStock, path, position);
-
-        if (action == Action.ACCELERATE) tractionForce = maxTractionForce;
-
-        boolean isBraking = (action == Action.BRAKE);
+        double averageGrade = getAverageGrade(rollingStock, path, position);
+        double weightForce = getWeightForce(rollingStock, averageGrade);
 
         if (action == Action.MAINTAIN) {
             tractionForce = rollingResistance - weightForce;
@@ -82,20 +102,53 @@ public final class TrainPhysicsIntegrator {
             else tractionForce = maxTractionForce;
         }
 
-        double acceleration = computeAcceleration(
-                rollingStock, rollingResistance, weightForce, speed, tractionForce, isBraking, directionSign);
+        if (action == Action.ACCELERATE) tractionForce = maxTractionForce;
+        double acceleration =
+                computeAcceleration(rollingStock, rollingResistance, weightForce, speed, tractionForce, directionSign);
         return newtonStep(timeStep, speed, acceleration, directionSign);
     }
 
-    /** Compute the weight force of a rolling stock at a given position on a given path */
-    public static double getWeightForce(PhysicsRollingStock rollingStock, PhysicsPath path, double headPosition) {
+    private double getDeceleration(double speed, double position) {
+        assert (action == Action.BRAKE);
+        if (brakingType == BrakingType.CONSTANT) return rollingStock.getDeceleration();
+
+        var grade = getMinGrade(rollingStock, path, position);
+        var gradientAcceleration = getGradientAcceleration(grade);
+        return switch (brakingType) {
+                // See Subset referenced in RJSEtcsBrakeParams: §3.13.6.2.1.3.
+            case ETCS_EBD -> -rollingStock.getRJSEtcsBrakeParams().getSafeBrakingAcceleration(speed)
+                    + gradientAcceleration;
+                // See Subset referenced in RJSEtcsBrakeParams: §3.13.6.3.1.3.
+            case ETCS_SBD -> -rollingStock.getRJSEtcsBrakeParams().getServiceBrakingAcceleration(speed)
+                    + gradientAcceleration;
+                // See Subset referenced in RJSEtcsBrakeParams: §3.13.6.4.3.
+            case ETCS_GUI -> -rollingStock.getRJSEtcsBrakeParams().getNormalServiceBrakingAcceleration(speed)
+                    + gradientAcceleration
+                    + rollingStock.getRJSEtcsBrakeParams().getGradientAccelerationCorrection(grade, speed);
+            default -> throw new UnsupportedOperationException("Braking type not supported: " + brakingType);
+        };
+    }
+
+    /** Compute the average grade of a rolling stock at a given position on a given path in m/km */
+    public static double getAverageGrade(PhysicsRollingStock rollingStock, PhysicsPath path, double headPosition) {
         var tailPosition = Math.min(Math.max(0, headPosition - rollingStock.getLength()), path.getLength());
         headPosition = Math.min(Math.max(0, headPosition), path.getLength());
-        var averageGrade = path.getAverageGrade(tailPosition, headPosition);
-        // get an angle from a meter per km elevation difference
+        return path.getAverageGrade(tailPosition, headPosition);
+    }
+
+    /** Compute the weight force of a rolling stock at a given position on a given path */
+    public static double getWeightForce(PhysicsRollingStock rollingStock, double grade) {
+        // get an angle from a m/km elevation difference
         // the curve's radius is taken into account in meanTrainGrade
-        var angle = Math.atan(averageGrade / 1000.0); // from m/km to m/m
-        return -rollingStock.getMass() * 9.81 * Math.sin(angle);
+        var angle = Math.atan(grade / 1000.0); // from m/km to m/m
+        return -rollingStock.getMass() * GRAVITY_ACCELERATION * Math.sin(angle);
+    }
+
+    /** Compute the min grade of a rolling stock at a given position on a given path in m/km */
+    public static double getMinGrade(PhysicsRollingStock rollingStock, PhysicsPath path, double headPosition) {
+        var tailPosition = Math.min(Math.max(0, headPosition - rollingStock.getLength()), path.getLength());
+        headPosition = Math.min(Math.max(0, headPosition), path.getLength());
+        return path.getMinGrade(tailPosition, headPosition);
     }
 
     /**
@@ -107,14 +160,9 @@ public final class TrainPhysicsIntegrator {
             double weightForce,
             double currentSpeed,
             double tractionForce,
-            boolean isBraking,
             double directionSign) {
 
         assert tractionForce >= 0.;
-
-        if (isBraking) {
-            return rollingStock.getDeceleration();
-        }
 
         if (currentSpeed == 0 && directionSign > 0) {
             // If we are stopped and if the forces are not enough to compensate the opposite force,
--- a/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/Constants.kt
+++ b/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/Constants.kt
@@ -1,0 +1,6 @@
+package fr.sncf.osrd.envelope_sim.etcs
+
+/** See Subset referenced in ETCSBrakingSimulator: table in Appendix A.3.1. */
+const val tDriver = 4.0 // s
+const val mRotatingMax = 15.0 // %
+const val mRotatingMin = 2.0 // %
--- a/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
+++ b/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
@@ -1,0 +1,263 @@
+package fr.sncf.osrd.envelope_sim.etcs
+
+import fr.sncf.osrd.envelope.Envelope
+import fr.sncf.osrd.envelope.OverlayEnvelopeBuilder
+import fr.sncf.osrd.envelope.part.ConstrainedEnvelopePartBuilder
+import fr.sncf.osrd.envelope.part.EnvelopePart
+import fr.sncf.osrd.envelope.part.EnvelopePartBuilder
+import fr.sncf.osrd.envelope.part.constraints.EnvelopeConstraint
+import fr.sncf.osrd.envelope.part.constraints.EnvelopePartConstraintType
+import fr.sncf.osrd.envelope.part.constraints.PositionConstraint
+import fr.sncf.osrd.envelope.part.constraints.SpeedConstraint
+import fr.sncf.osrd.envelope_sim.*
+import fr.sncf.osrd.envelope_sim.overlays.EnvelopeDeceleration
+import kotlin.math.max
+import kotlin.math.min
+
+enum class BrakingCurveType {
+    EBD, // Emergency Brake Deceleration
+    EBI, // Emergency Brake Intervention
+    SBD, // Service Brake Deceleration
+    SBI, // Service Brake Intervention
+    GUI, // Guidance
+    PS, // Permitted Speed
+    IND // Indication
+}
+
+enum class BrakingType {
+    CONSTANT,
+    ETCS_EBD,
+    ETCS_SBD,
+    ETCS_GUI
+}
+
+/** Compute braking curves at every end of authority. */
+fun addBrakingCurvesAtEOAs(
+    envelope: Envelope,
+    context: EnvelopeSimContext,
+    endsOfAuthority: Collection<EndOfAuthority>
+): Envelope {
+    val sortedEndsOfAuthority = endsOfAuthority.sortedBy { it.offsetEOA }
+    var beginPos = 0.0
+    val builder = OverlayEnvelopeBuilder.forward(envelope)
+    for (endOfAuthority in sortedEndsOfAuthority) {
+        val targetPosition = endOfAuthority.offsetEOA.distance.meters
+        val targetSpeed = 0.0
+        val overhead =
+            Envelope.make(
+                EnvelopePart.generateTimes(
+                    listOf(EnvelopeProfile.CONSTANT_SPEED),
+                    doubleArrayOf(beginPos, targetPosition),
+                    doubleArrayOf(envelope.maxSpeed, envelope.maxSpeed)
+                )
+            )
+        val guiCurve =
+            computeBrakingCurve(
+                context,
+                overhead,
+                beginPos,
+                targetPosition,
+                targetSpeed,
+                BrakingType.ETCS_GUI
+            )
+        val sbdCurve =
+            computeBrakingCurve(
+                context,
+                overhead,
+                beginPos,
+                targetPosition,
+                targetSpeed,
+                BrakingType.ETCS_SBD
+            )
+        val fullIndicationCurve =
+            computeIndicationBrakingCurveFromRef(
+                context,
+                sbdCurve,
+                BrakingCurveType.SBD,
+                guiCurve,
+                beginPos
+            )
+        val indicationCurve = keepBrakingCurveUnderOverlay(fullIndicationCurve, envelope, beginPos)
+        builder.addPart(indicationCurve)
+        beginPos = targetPosition
+    }
+    return builder.build()
+}
+
+/** Compute braking curve: used to compute EBD, SBD or GUI. */
+private fun computeBrakingCurve(
+    context: EnvelopeSimContext,
+    envelope: Envelope,
+    beginPos: Double,
+    targetPosition: Double,
+    targetSpeed: Double,
+    brakingType: BrakingType
+): EnvelopePart {
+    // If the stopPosition is below begin position, the input is invalid
+    // If the stopPosition is after the end of the path, the input is invalid except if it is an
+    // SVL, i.e. the target speed is 0 and the curve to compute is an EBD
+    if (
+        targetPosition <= beginPos ||
+            (targetPosition > context.path.length &&
+                targetSpeed == 0.0 &&
+                brakingType != BrakingType.ETCS_EBD)
+    )
+        throw RuntimeException(
+            String.format(
+                "Trying to compute ETCS braking curve from out of bounds ERTMS end/limit of authority: %s",
+                targetPosition
+            )
+        )
+    val partBuilder = EnvelopePartBuilder()
+    partBuilder.setAttr(EnvelopeProfile.BRAKING)
+    val overlayBuilder =
+        ConstrainedEnvelopePartBuilder(
+            partBuilder,
+            PositionConstraint(beginPos, targetPosition),
+            SpeedConstraint(0.0, EnvelopePartConstraintType.FLOOR),
+            EnvelopeConstraint(envelope, EnvelopePartConstraintType.CEILING)
+        )
+    EnvelopeDeceleration.decelerate(
+        context,
+        targetPosition,
+        targetSpeed,
+        overlayBuilder,
+        -1.0,
+        brakingType
+    )
+    val brakingCurve = partBuilder.build()
+    return brakingCurve
+}
+
+/**
+ * Compute Indication curve: EBI/SBD -> SBI -> PS -> IND. See Subset referenced in
+ * ETCSBrakingSimulator: figures 45 and 46.
+ */
+private fun computeIndicationBrakingCurveFromRef(
+    context: EnvelopeSimContext,
+    refBrakingCurve: EnvelopePart,
+    refBrakingCurveType: BrakingCurveType,
+    guiCurve: EnvelopePart,
+    beginPos: Double
+): EnvelopePart {
+    assert(refBrakingCurve.endPos > beginPos)
+    val rollingStock = context.rollingStock
+    val tBs =
+        when (refBrakingCurveType) {
+            BrakingCurveType.EBI -> rollingStock.rjsEtcsBrakeParams.tBs2
+            BrakingCurveType.SBD -> rollingStock.rjsEtcsBrakeParams.tBs1
+            else ->
+                throw IllegalArgumentException(
+                    "Expected EBI or SBD reference braking curve type, found: $refBrakingCurveType"
+                )
+        }
+
+    val reversedNewPos = ArrayList<Double>()
+    val reversedNewSpeeds = ArrayList<Double>()
+    for (i in refBrakingCurve.pointCount() - 1 downTo 0) {
+        val reversedNewPosIndex = refBrakingCurve.pointCount() - 1 - i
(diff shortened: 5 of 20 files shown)


core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/PhysicsRollingStock.java
```diff
@@ -50,6 +70,11 @@ static double getMaxEffort(double speed, TractiveEffortPoint[] tractiveEffortCur
         return previousPoint.maxEffort() + coeff * (Math.abs(speed) - previousPoint.speed());
     }
 
+    /** The gradient acceleration of the rolling stock taking its rotating mass into account, in m/s^2 */
+    static double getGradientAcceleration(double grade, double rotatingMassPercentage) {
+        return -9.81 * grade / (1000.0 + 10.0 * rotatingMassPercentage);
```
bougue-pe: To be explicit, can we make a const value for gravity acceleration instead of a magic number?

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.java
```diff
@@ -596,19 +596,19 @@ public EnvelopePart slice(
      * Returns a new EnvelopePart, where all positions are shifted by positionDelta. Resulting
      * positions are clipped to [minPosition; maxPosition].
      */
-    public EnvelopePart copyAndShift(double positionDelta, double minPosition, double maxPosition) {
+    public EnvelopePart copyAndShift(double positionDelta, double minPosition, double maxPosition, double speedDelta) {
```
eckter: Comment should probably be updated.

What happens if we get close to or below 0?

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/EnvelopeSimPath.java
```diff
@@ -106,6 +106,29 @@ public double getAverageGrade(double begin, double end) {
         return (getCumGrade(end) - getCumGrade(begin)) / (end - begin);
     }
 
+    @Override
+    public double getMinGrade(double begin, double end) {
```
eckter, line 110: We could add a comment in the body of the function (this one and the next) to indicate that it can easily be optimized if needed.

(though I entirely approve the simple and non-optimized implementation for a first version :+1: )

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/PhysicsRollingStock.java
eckter: What a weird operation. I'm sure it comes from the document somewhere with sound logic, but it feels surprising.

The effect of gravity depends on the rotating mass percentage ?
And that percentage depends on whether the slope is going forwards or backwards?

I trust you on translating the PDFs, but maybe that could use an extra comment somewhere

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/TrainPhysicsIntegrator.java
```diff
@@ -65,39 +85,66 @@ private IntegrationStep step(double timeStep, double initialLocation, double ini
...
-        var angle = Math.atan(averageGrade / 1000.0); // from m/km to m/m
+        var angle = Math.atan(grade / 1000.0); // from m/km to m/m
         return -rollingStock.getMass() * 9.81 * Math.sin(angle);
     }
 
+    /** Compute the min grade of a rolling stock at a given position on a given path in m/km */
+    public static double getMinGrade(PhysicsRollingStock rollingStock, PhysicsPath path, double headPosition) {
+        var tailPosition = Math.min(Math.max(0, headPosition - rollingStock.getLength()), path.getLength());
+        headPosition = Math.min(Math.max(0, headPosition), path.getLength());
+        return path.getMinGrade(tailPosition, headPosition);
```
eckter, line 151: Oh, if we call it at *every integration step* we may need the optimized version sooner than I thought. Though that can still wait until after the first functional POC.

eckter (commented):
Not done yet with the review, but 'll do the rest later on, I can comment what I have

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/PhysicsRollingStock.java
Erashin: You can check out page 104 of the subset, paragraph 3.13.4.3, the theory looks sound. I'll add in a comment though you're right ;)
To summarize, part of the mass is rolling and doesn't actually translate into linear motion, hence the factor added to reduce the total gradient acceleration.
image

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/TrainPhysicsIntegrator.java
```diff
@@ -65,39 +85,66 @@ private IntegrationStep step(double timeStep, double initialLocation, double ini
...
+        double acceleration =
+                computeAcceleration(rollingStock, rollingResistance, weightForce, speed, tractionForce, directionSign);
         return newtonStep(timeStep, speed, acceleration, directionSign);
     }
 
-    /** Compute the weight force of a rolling stock at a given position on a given path */
-    public static double getWeightForce(PhysicsRollingStock rollingStock, PhysicsPath path, double headPosition) {
+    private double getDeceleration(double speed, double position) {
+        assert (action == Action.BRAKE);
+        if (brakingType == BrakingType.CONSTANT) return rollingStock.getDeceleration();
```
bougue-pe, line 113: nit: a newline after early return would be clearer IMO

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/TrainPhysicsIntegrator.java
```diff
@@ -65,39 +85,66 @@ private IntegrationStep step(double timeStep, double initialLocation, double ini
...
+        var grade = getMinGrade(rollingStock, path, position);
+        var mRotating = grade >= 0 ? mRotatingMax : mRotatingMin;
```
bougue-pe: I would move that part in PhysicsRollingStock as this is closely tight to the formula.

core/envelope-sim/src/testFixtures/java/fr/sncf/osrd/envelope_sim/SimpleRollingStock.java
```diff
@@ -90,6 +90,47 @@ public double getRollingResistanceDeriv(double speed) {
         return B + 2 * C * speed;
     }
 
+    // TODO: implement these methods?
+    @Override
+    public double getTractionCutOff() {
+        return 0;
```
bougue-pe: (same for all) how about:
```suggestion
        throw new UnsupportedOperationException("To be implemented");
```

core/src/main/java/fr/sncf/osrd/train/RollingStock.java
```diff
@@ -260,7 +327,7 @@ public boolean isThermal() {
     }
 
     /** Return whether this rolling stock's has an electric mode */
-    public final boolean isElectric() {
+    public boolean isElectric() {
```
bougue-pe, line 266: Why that change?

core/src/main/java/fr/sncf/osrd/train/RollingStock.java
```diff
@@ -123,6 +125,73 @@ public double getRollingResistanceDeriv(double speed) {
         return B + 2 * C * speed;
     }
 
+    @Override
+    public double getTractionCutOff() {
+        assert etcsBrakeParams != null;
+        return etcsBrakeParams.tTractionCutOff;
+    }
```
bougue-pe: How about moving it all to etcsBrakeParams directly (checking for nullity once and for all, remove the simple getters) ?

core/src/main/java/fr/sncf/osrd/train/RollingStock.java
```diff
@@ -123,6 +125,73 @@ public double getRollingResistanceDeriv(double speed) {
         return B + 2 * C * speed;
     }
 
+    @Override
+    public double getTractionCutOff() {
```
bougue-pe: (if method is kept)
```suggestion
    public double getTTractionCutOff() {
```

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingSimulator.kt
```diff
@@ -38,18 +40,21 @@ data class EndOfAuthority(
...
-    override val rollingStock: PhysicsRollingStock,
-    override val timeStep: Double
-) : ETCSBrakingSimulator {
+class ETCSBrakingSimulatorImpl(override val context: EnvelopeSimContext) : ETCSBrakingSimulator {
+    override fun addSlowdownBrakingCurves(
+        envelope: Envelope,
+        limitsOfAuthority: Collection<LimitOfAuthority>
+    ): Envelope {
+        if (limitsOfAuthority.isEmpty()) return envelope
+        // TODO: implement braking at LOAs CORRECTLY
```
bougue-pe, line 54: We should probably rename that PR for now, then.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingSimulator.kt
```diff
@@ -38,18 +40,21 @@ data class EndOfAuthority(
...
-        mrsp: Envelope,
-        limitsOfAuthority: Collection<LimitOfAuthority>,
+    override fun addStopBrakingCurves(
+        envelope: Envelope,
         endsOfAuthority: Collection<EndOfAuthority>
     ): Envelope {
-        if (limitsOfAuthority.isEmpty() && endsOfAuthority.isEmpty()) return mrsp
-        TODO("Not yet implemented")
+        if (endsOfAuthority.isEmpty()) return envelope
+        return addBrakingCurvesAtEOAs(envelope, context, endsOfAuthority)
```
bougue-pe, line 64:
```suggestion
        // TODO compute curve at SvL and intersect
        return addBrakingCurvesAtEOAs(envelope, context, endsOfAuthority)
```

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.java
Erashin: Not gonna lie, this method isn't used in this pr anymore, I'll just remove the changes, there's no point.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/EnvelopeSimPath.java
Erashin: I'll add a TODO in at least in the meantime.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/TrainPhysicsIntegrator.java
```diff
@@ -65,39 +85,66 @@ private IntegrationStep step(double timeStep, double initialLocation, double ini
...
+        var gradientAcceleration = getGradientAcceleration(grade, mRotating);
+        return switch (brakingType) {
+            case ETCS_EBD -> -rollingStock.getSafeBrakingAcceleration(speed) + gradientAcceleration;
+            case ETCS_SBD -> -rollingStock.getServiceBrakingAcceleration(speed) + gradientAcceleration;
+            case ETCS_GUI -> -rollingStock.getNormalServiceBrakingAcceleration(speed)
+                    + gradientAcceleration
+                    + rollingStock.getGradientAccelerationCorrection(gradientAcceleration, speed);
```
eckter: The parameters for `rollingStock.getGradientAccelerationCorrection` are `double grade, double rotatingMassPercentage`

Is it a "grade" or an "acceleration"? What's the expected unit?

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/PhysicsRollingStock.java
```diff
@@ -19,6 +19,26 @@ public interface PhysicsRollingStock {
...
+    double getSafeBrakingAcceleration(double speed);
+
+    /** The service braking acceleration which can be applied at a given speed, in m/s^2 */
+    double getServiceBrakingAcceleration(double speed);
+
+    /** The normal service braking acceleration which can be applied at a given speed, in m/s^2 */
+    double getNormalServiceBrakingAcceleration(double speed);
+
+    /** The gradient acceleration correction applied for guidance computation in ETCS 2, in m/s^2 */
+    double getGradientAccelerationCorrection(double grade, double speed);
```
eckter: Related to the other comment: `grade` unit should be mentionned in the comment, assuming it's something like "meters per 100m" or something it's not entirely obvious

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -0,0 +1,213 @@
...
+}
+
+/** Compute braking curves at every end of authority. */
+fun addBrakingCurvesAtEOAs(
+    envelope: Envelope,
+    context: EnvelopeSimContext,
+    endsOfAuthority: Collection<EndOfAuthority>
+): Envelope {
+    val builder = OverlayEnvelopeBuilder.backward(envelope)
+    for (endOfAuthority in endsOfAuthority.reversed()) {
```
eckter: It's not sorted in any meaningful way

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -0,0 +1,213 @@
...
+        val targetPosition = endOfAuthority.offsetEOA.distance.meters
+        val targetSpeed = 0.0
+        val overhead =
+            Envelope.make(
+                EnvelopePart.generateTimes(
+                    listOf(EnvelopeProfile.CONSTANT_SPEED),
+                    doubleArrayOf(0.0, targetPosition),
+                    doubleArrayOf(envelope.maxSpeed, envelope.maxSpeed)
```
eckter, line 51: If we want the envelopes not to stop too early (so that we can shift them), we may want `context.rollingStock.maxSpeed` instead?

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -0,0 +1,213 @@
...
+                context,
+                overhead,
+                targetPosition,
+                targetSpeed,
+                BrakingType.ETCS_SBD
+            )
+        val fullIndicationCurve =
+            computeIndicationBrakingCurveFromRef(context, sbdCurve, BrakingCurveType.SBD, guiCurve)
+        val indicationCurve = keepBrakingCurveUnderOverlay(fullIndicationCurve, envelope)
+        builder.addPart(indicationCurve)
```
eckter, line 81: We *may* have cases where the parts overlap if the points are too close. I'm not sure it would be an issue, but I'd write a test case with points a few meters apart to be sure

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -0,0 +1,213 @@
...
+/** Compute braking curve: used to compute EBD, SBD or GUI. */
+private fun computeBrakingCurve(
+    context: EnvelopeSimContext,
+    envelope: Envelope,
+    targetPosition: Double,
+    targetSpeed: Double,
+    brakingType: BrakingType
+): EnvelopePart {
+    // If the stopPosition is below zero, the input is invalid
+    if (targetPosition <= 0.0)
```
eckter: Is it valid if it's higher than the path length? Considering the error message I'd expect both to be checked

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/TrainPhysicsIntegrator.java
Erashin: The expected unit for `rollingStock.getGradientAccelerationCorrection` is an acceleration, and the input should indeed be a grade, and not an acceleration. Well spotted!
image

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Nope. The only correct case is an SVL with a target position higher than the path length, associated to an EOA with a target position lower than the path length. I'll add that in :)
Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/537a2e40a390b8d8a7fb217d491bd4d4578fbae5

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/PhysicsRollingStock.java
Erashin: It's in meters per kilometers. Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/83348a7b07ee46f0dcf828cb0546b811b5f4576b

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/TrainPhysicsIntegrator.java
Erashin: Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/83348a7b07ee46f0dcf828cb0546b811b5f4576b

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: In this case, not really. For an EOA, when calculating the indication curve from the SBD curve, we only shift the positions, and apply no modifications to the speed. So `envelope.maxSpeed` is perfect for that.
In the other cases (LOA and SVL), where we compute an indication curve from the EBD curve, we indeed shift the speeds as well, by `deltaVbec(speed)`. When we'll implement it, we'll use `envelope.maxSpeed + maxDeltaVbecSpeed` instead. But that'll be done in the other PR (https://github.com/OpenRailAssociation/osrd/pull/10316).

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: We're actually gonna need to sort it to solve your pb below about overlaps! See https://github.com/OpenRailAssociation/osrd/pull/10014/commits/b8d98ccbcb2b162ffc6fcaba38770059320735fc

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/EnvelopeSimPath.java
```diff
@@ -106,6 +106,30 @@ public double getAverageGrade(double begin, double end) {
...
+        var lowestGradient = gradeValues[indexBegin];
+        for (int i = indexBegin; i < indexEnd; i++) {
+            var grad = gradeValues[i];
+            if (grad < lowestGradient) lowestGradient = grad;
+        }
+        return lowestGradient;
+    }
+
+    /** For a given position, return the index of the position just before in gradePositions */
+    public int getIndexBeforePos(double position) {
```
bougue-pe, line 123: We may add a TODO on binary search here too.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/PhysicsRollingStock.java
```diff
@@ -50,6 +73,14 @@ static double getMaxEffort(double speed, TractiveEffortPoint[] tractiveEffortCur
         return previousPoint.maxEffort() + coeff * (Math.abs(speed) - previousPoint.speed());
     }
 
+    /**
+     * The gradient acceleration of the rolling stock taking its rotating mass into account, in m/s^2.
+     * Grade is in meters per kilometers.
```
bougue-pe: nit: In this PR, there are `meters per kilometers`, `meter per km`, `m/km`.
I'd put the last one everywhere.

bougue-pe (commented):
ETCSBrakingCurves.kt is out of the scope for this review 😅

core/src/main/java/fr/sncf/osrd/train/RollingStock.java
Erashin: It was redundant, so i just removed it (add it back, you'll see what i mean).

core/src/main/java/fr/sncf/osrd/train/RollingStock.java
Erashin: Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/7eccec26e618803b194df654161dfc8dd0b130ed

core/src/main/java/fr/sncf/osrd/train/RollingStock.java
Erashin: Not keeping it, all good :)

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: You're correct. What we're gonna need is to sort the EOAs, and compute their braking curves along the path. For each EOA, their braking curve's begin position can't be located before the previous EOA (or 0 if this is the first EOA).
Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/b8d98ccbcb2b162ffc6fcaba38770059320735fc.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/PhysicsRollingStock.java
Erashin: Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/7eccec26e618803b194df654161dfc8dd0b130ed

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/TrainPhysicsIntegrator.java
Erashin: Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/7eccec26e618803b194df654161dfc8dd0b130ed

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/TrainPhysicsIntegrator.java
Erashin: Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/7eccec26e618803b194df654161dfc8dd0b130ed

core/envelope-sim/src/testFixtures/java/fr/sncf/osrd/envelope_sim/SimpleRollingStock.java
Erashin: Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/7eccec26e618803b194df654161dfc8dd0b130ed
However this will probably change when I start adding unit tests.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/EnvelopeSimPath.java
Erashin: Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/7eccec26e618803b194df654161dfc8dd0b130ed

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/PhysicsRollingStock.java
Erashin: Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/7eccec26e618803b194df654161dfc8dd0b130ed

core/osrd-railjson/src/main/java/fr/sncf/osrd/railjson/schema/rollingstock/RJSEtcsBrakeParams.java
```diff
@@ -9,39 +9,45 @@
  */
 public class RJSEtcsBrakeParams {
 
+    /** National Default Value: Available Adhesion */
+    private final double mNvavadh = 0.0;
+
+    /** National Default Value: Emergency Brake Confidence Level in % */
```
bougue-pe: As this is not expressed as percentage here:
```suggestion
    /** National Default Value: Emergency Brake Confidence Level ratio */
```

core/osrd-railjson/src/main/java/fr/sncf/osrd/railjson/schema/rollingstock/RJSEtcsBrakeParams.java
```diff
@@ -60,6 +66,38 @@ public class RJSEtcsBrakeParams {
...
+        var kWet = getRollingStockCorrectionFactorWet(speed);
+        return kDry * (kWet + mNvavadh * (1 - kWet)) * aBrakeEmergency;
+    }
+
+    private double getEmergencyBrakingDeceleration(double speed) {
+        return gammaEmergency.getValue(speed);
+    }
+
+    private double getRollingStockCorrectionFactorDry(double speed) {
+        return kDry.getValue(speed) * mNvebcl;
```
bougue-pe: I didn't find it in the documentation subset, is it possible to add a comment on that?
To my understanding, this getRollingStockCorrectionFactorDry is indeed a function of speed and confidence level, but I didn't see any precise formula, and I understood it more as "the manufacturer has to provide a correction factor function of the speed on which they are x% sure that the emergency braking will provide at least the power described" (§ 3.13.2.2.9.1.4 of the subset).
So to my understanding, this value would only be used to pick a curve if there were multiple ones available depending on the country.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -0,0 +1,221 @@
...
+    )
+    val brakingCurve = partBuilder.build()
+    return brakingCurve
+}
+
+/** Compute Indication curve: EBI/SBD -> SBI -> PS -> IND */
+private fun computeIndicationBrakingCurveFromRef(
+    context: EnvelopeSimContext,
+    refCurve: EnvelopePart,
+    refBrakingCurveType: BrakingCurveType,
```
bougue-pe: For consistency:
```suggestion
    refBrakingCurve: EnvelopePart,
    refBrakingCurveType: BrakingCurveType,
```
(no `Braking` at all also works for me)

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -0,0 +1,221 @@
...
+    guiCurve: EnvelopePart
+): EnvelopePart {
+    assert(
+        refBrakingCurveType == BrakingCurveType.EBI || refBrakingCurveType == BrakingCurveType.SBD
+    )
+    val rollingStock = context.rollingStock
+    val tBs =
+        if (refBrakingCurveType == BrakingCurveType.EBI) rollingStock.rjsEtcsBrakeParams.tBs2
+        else rollingStock.rjsEtcsBrakeParams.tBs1
```
bougue-pe: nit: how about
```suggestion
    val rollingStock = context.rollingStock
    val tBs = when (refBrakingCurveType) {
        BrakingCurveType.EBI -> rollingStock.rjsEtcsBrakeParams.tBs2
        BrakingCurveType.SBD -> rollingStock.rjsEtcsBrakeParams.tBs1
        else -> throw IllegalArgumentException("Wrong ref curve type") 
    }
```

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -0,0 +1,221 @@
...
+
+    val pos = refCurve.clonePositions()
+    val speeds = refCurve.cloneSpeeds()
```
bougue-pe: Looks like we can avoid the clone, no?

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -0,0 +1,221 @@
...
+fun getPermittedSpeedPosition(
+    sbiPosition: Double,
+    speed: Double,
+    guiPosition: Double = Double.POSITIVE_INFINITY
+): Double {
+    return min(getPreviousPosition(sbiPosition, speed, tDriver), guiPosition)
+}
+
+fun getIndicationPosition(psPosition: Double, speed: Double, tBs: Double): Double {
+    val tIndication = max((0.8 * tBs), 5.0) + tDriver
```
bougue-pe, line 253: Maybe a comment to refer to the doc would be fine to explain those numbers a bit: I found §3.13.9.3.6.2 , p127 of SUBSET-026-3 v360.pdf.
Better idea welcome!

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -0,0 +1,221 @@
...
+
+fun getPermittedSpeedPosition(
+    sbiPosition: Double,
+    speed: Double,
+    guiPosition: Double = Double.POSITIVE_INFINITY
+): Double {
+    return min(getPreviousPosition(sbiPosition, speed, tDriver), guiPosition)
+}
+
+fun getIndicationPosition(psPosition: Double, speed: Double, tBs: Double): Double {
```
bougue-pe: I'd name it more explicitly to "enforce" the link:
```suggestion
fun getIndicationPosition(permittedSpeedPosition: Double, speed: Double, tBs: Double): Double {
```

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -0,0 +1,221 @@
...
+
+    val pos = refCurve.clonePositions()
+    val speeds = refCurve.cloneSpeeds()
+    val reversedNewPos = ArrayList<Double>()
+    val reversedNewSpeeds = ArrayList<Double>()
+    for (i in refCurve.pointCount() - 1 downTo 0) {
+        val speed = speeds[i]
+        val sbiPosition = getSbiPosition(pos[i], speed, tBs)
+        val permittedSpeedPosition =
+            getPermittedSpeedPosition(sbiPosition, speed, guiCurve.interpolatePosition(speed))
```
bougue-pe: Trying to "self-document" a bit the code (and after searching a bit the subset, §3.13.9.3.1.4 p118):
I would separate a "permittedSpeed" that would only derive from warning/sbi, its couterpart being the guidance, and an "adjustedPermittedSpeed" being the minimum of those two.
The consistency of the calculation would be more explicit IMO (very small functions).

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -0,0 +1,221 @@
...
+        val indicationPosition = getIndicationPosition(permittedSpeedPosition, speed, tBs)
+        if (indicationPosition >= 0) {
+            reversedNewPos.add(indicationPosition)
+            reversedNewSpeeds.add(speed)
+        } else if (i != refCurve.pointCount() - 1 && reversedNewPos[i + 1] > 0) {
```
bougue-pe:
```suggestion
        } else if (i != refCurve.pointCount() - 1 && reversedNewPos[i + 1] > 0) {
        // interpolate to zero if reaching a negative position
```

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -0,0 +1,221 @@
...
+
+    val nbPoints = reversedNewPos.size
+    val newPosArray = DoubleArray(nbPoints)
+    val newSpeedsArray = DoubleArray(nbPoints)
+    for (i in newPosArray.indices) {
+        newPosArray[i] = reversedNewPos[nbPoints - 1 - i]
+        newSpeedsArray[i] = reversedNewSpeeds[nbPoints - 1 - i]
+    }
+    val brakingCurve =
+        EnvelopePart.generateTimes(listOf(EnvelopeProfile.BRAKING), newPosArray, newSpeedsArray)
```
bougue-pe, line 193: Not sure: could we save some processing and wait until the envelope is cut before generating times?

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -0,0 +1,221 @@
...
+
+/**
+ * Keep the part of the full braking curve which is located underneath the overlay and intersects
+ * with it.
+ */
+private fun keepBrakingCurveUnderOverlay(
+    fullBrakingCurve: EnvelopePart,
+    overlay: Envelope
+): EnvelopePart {
+    assert(fullBrakingCurve.maxSpeed >= overlay.minSpeed)
```
bougue-pe, line 207: nit: This is necessary but not sufficient, so I'm not sure I would keep that check.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/Constants.kt
```diff
@@ -0,0 +1,5 @@
+package fr.sncf.osrd.envelope_sim.etcs
+
+const val tDriver = 4.0 // s
+const val mRotatingMax = 15.0 // %
+const val mRotatingMin = 2.0 // %
```
bougue-pe, line 6: From the subset I'm not sure if we should use 0.15 or 15 in the formulas (same for the min) 🤷

bougue-pe (commented):
This is a great first step for the ERTMS 👣 
Nice job, just little questions remaining on the formulas, but it should be fine to track them (maybe in the ERA xls file?).
🎉

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: I need and `EnvelopePart` in the method that cuts this part out, so I don't think so no.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: I agree it's not sufficient, but it's the bare minimum for this method to work. This does guarantee the user doesn't do something dumb with this method. I'd rather keep it personally.

core/osrd-railjson/src/main/java/fr/sncf/osrd/railjson/schema/rollingstock/RJSEtcsBrakeParams.java
Erashin: I can add a comment on that for sure! I also checked its value and will add a comment on that as well. Might be as you said and it'll depend on the country/manufacturer, honestly not sure :)
Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/c1bc4feb0ae2e7792d1eb8b8c9171b225d546229

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Modified it to fit to the new code, where the left-most position limit is `beginPosition` instead of 0.
Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/c1bc4feb0ae2e7792d1eb8b8c9171b225d546229

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingSimulator.kt
Erashin: Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/c1bc4feb0ae2e7792d1eb8b8c9171b225d546229

core/osrd-railjson/src/main/java/fr/sncf/osrd/railjson/schema/rollingstock/RJSEtcsBrakeParams.java
Erashin: Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/c1bc4feb0ae2e7792d1eb8b8c9171b225d546229

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/c1bc4feb0ae2e7792d1eb8b8c9171b225d546229

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/c1bc4feb0ae2e7792d1eb8b8c9171b225d546229

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/c1bc4feb0ae2e7792d1eb8b8c9171b225d546229

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: I also put all the references to the subset for every formula/constant i could find in the rest of the pr :)
Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/c1bc4feb0ae2e7792d1eb8b8c9171b225d546229

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/c1bc4feb0ae2e7792d1eb8b8c9171b225d546229

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Fixed in https://github.com/OpenRailAssociation/osrd/pull/10014/commits/c1bc4feb0ae2e7792d1eb8b8c9171b225d546229

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/Constants.kt
Erashin: It's 15 and 2. Went and checked the xls to make sure of it :)

core/osrd-railjson/src/main/java/fr/sncf/osrd/railjson/schema/rollingstock/RJSEtcsBrakeParams.java
bougue-pe: As discussed, removed it completely, as the subset seems oriented to multiple curves functions of speed (those curves being computed to cover the confidence level required). And also no use could be found in ERA braking curves simulation tool v5.1.

eckter (approved):
Nice

bougue-pe (approved):
Noiiice !
Integration tests to be done in a separate PR #10294

PR #10018, opened 2024-12-11
front: remove high speed train from stdcm

axrolld:
Remove this high-speed train that sneaked among the freight locomotives.
It's just a kick fix i haven't checked that it actually works cause i can't deploy locally, but i assume this small fix should work. If you can try it quickly at the review i'd be great :)
Thanks !


clarani (approved):
LGTM :white_check_mark:

PR #10019, opened 2024-12-11
front: revert tolerances! in StdcmOrigin and StdcmDestination

clarani:
tolerances are stored in the store and in the local storage. For some users, this field is still undefined in their local storage. However, since https://github.com/OpenRailAssociation/osrd/pull/9977, we assume that tolerances can not be undefined, which is not true if the users have not reset their store (=clean their cookies) or manullaly filled it.

To prevent the app from crashing when a user without tolerances opens the stdcm page, I revert the changes in StdcmOrigin and StdcmDestination for now.

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmDestination.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmDestination.tsx
@@ -31,8 +31,8 @@ const StdcmDestination = ({ disabled = false }: StdcmConfigCardProps) => {
         ? extractDateAndTimefromISO(destination.arrival)
         : undefined,
       destinationToleranceValues: {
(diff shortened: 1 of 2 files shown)


Uriel-Sautron (approved):
LGTM, tested.

emersion (commented):
Shouldn't this be `DEFAULT_TOLERANCE` instead of 0? (That was the previous behavior)

clarani:
No because if we display DEFAULT_TOLERANCE_VALUE, the user thinks that the value is defined but it is not in the store and then not sent with the request :/

If the value in the store is undefined, we should not display something else

theocrsb (approved):
LGTM and tested

SharglutDev (approved):
Lgtm and tested, thank you !

PR #10020, opened 2024-12-11
front: display stdcm selected simulation's inputs

### LINKED ISSUE #9889: stdcm - inputs displayed don't change if several simulations
### What happened?

> [!IMPORTANT]
> This bug does not happen only on the time input. All the inputs from "Origine" seem to behave that way, and probably destination too. See [this comment](https://github.com/OpenRailAssociation/osrd/issues/9889#issuecomment-2523190165).

I have two simulations : 
- A first one which starts at 20h Image

- A second one which starts at 21h Image

But when I navigate between the two simulations, the inputs are the same for both (the inputs of the 2nd simulation) : if I go to the first simulation, the departure time is 21h instead of 20h Image

I expected the input to be adapted to the selected simulation

1. Create a 1st simulation which starts at 20h
2. Create a 2nd simulation which starts at 21h
3. Go back to see the first simulation and observe that the time is 21h and not 20h

Recette (SNCF)

Google Chrome

deac2f4

clarani:
closes #9889

### DIFF
--- a/front/src/applications/stdcm/types.ts
+++ b/front/src/applications/stdcm/types.ts
@@ -13,7 +13,7 @@ import type {
 } from 'common/api/osrdEditoastApi';
 import type { SpeedSpaceChartData } from 'modules/simulationResult/types';
 import type { SuggestedOP } from 'modules/trainschedule/components/ManageTrainSchedule/types';
-import type { PathStep } from 'reducers/osrdconf/types';
+import type { PathStep, StdcmPathStep } from 'reducers/osrdconf/types';
 import type { ValueOf } from 'utils/types';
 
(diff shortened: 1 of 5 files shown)


Akctarus (commented):
tested and works well, but there is an error on stdcmConfReducers.spec.ts l.104, I think you can change this line to something like this: 

```
stdcmPathSteps: [
          {
            ...paris,
            isVia: false,
            arrivalType: ArrivalTimeTypes.PRECISE_TIME,
          },
          {
            ...rennes,
            isVia: false,
            arrivalType: ArrivalTimeTypes.ASAP,
          },
        ],
```

Akctarus (approved):
LGTM !

Math-R (approved):
lgtm & tested in local env so can't reproduce exactly the same as @maelysLeratRosso but seems to works fine

PR #10021, opened 2024-12-11
python: bump osrd-schemas version

shenriotpro:


leovalais (approved):
LGTM

flomonster (approved):
Lgtm

Khoyo:
I think the latest change also require a bump in railjson generator, no?

flomonster:
> I think the latest change also require a bump in railjson generator, no?

You are right, actually I don't know if the railjson-generator is still used but I don't think so. Maybe we should drop this component?

Khoyo:
> You are right, actually I don't know if the railjson-generator is still used but I don't think so. Maybe we should drop this component?

It's how we generate tiny-infra, small-infra, etc.

(btw, I believe you also need to bump the railjson-generator lockfile, otherwise poetry might fail to bump the dependency if it's not the first install)

PR #10025, opened 2024-12-12
front: add towed rolling stock e2e test

### LINKED ISSUE #9743: front: LMR towed rolling stock e2e test
### Description and goal

The towed rolling stock is a new field in LMR Consist. It needs to be tested.

### Acceptance criteria

- [ ] The simulation result in LMR wil take into consideration the towed rolling stock

Maymanaf:
Closes #9743 

This PR:  
- Keeps only the required fake train for e2e tests in `freightRollingStocks.ts`.  
- Adjusts the method responsible for filling the consist fields to ensure that selecting a rolling stock or towed rolling stock correctly pre-fills the tonnage, length, and maximum speed fields.  
- Creates two simulations: one without outputs and another with outputs using a towed rolling stock, verifying the table outputs accordingly.  
- Replaces all `console` usage with the appropriate `logger` type.  
- Fixes typos.

### DIFF
--- a/front/tests/pages/stdcm-page-model.ts
+++ b/front/tests/pages/stdcm-page-model.ts
@@ -2,7 +2,6 @@ import { expect, type Locator, type Page } from '@playwright/test';
 
 import enTranslations from '../../public/locales/en/stdcm.json';
 import frTranslations from '../../public/locales/fr/stdcm.json';
-import { electricRollingStockName } from '../assets/project-const';
 import {
   CI_SUGGESTIONS,
   DEFAULT_DETAILS,
@@ -13,7 +12,8 @@ import {
   VIA_STOP_TIMES,
   VIA_STOP_TYPES,
 } from '../assets/stdcm-const';
-import { readJsonFile } from '../utils';
+import { logger } from '../test-logger';
+import { handleAndVerifyInput, readJsonFile } from '../utils';
 
 interface TableRow {
   index: number;
@@ -25,6 +25,16 @@ interface TableRow {
   weight: string | null;
   refEngine: string | null;
 }
+
+export interface ConsistFields {
+  tractionEngine: string;
+  towedRollingStock?: string;
+  tonnage?: string;
+  length?: string;
+  maxSpeed?: string;
+  speedLimitTag?: string;
+}
+const EXPECT_TO_PASS_TIMEOUT = 90_000; // Since toPass ignores custom expect timeouts, this timeout is set to account for all actions within the function.
 
 class STDCMPage {
   readonly page: Page;
@@ -43,11 +53,13 @@ class STDCMPage {
 
   readonly tractionEngineField: Locator;
 
+  readonly towedRollingStockField: Locator;
+
   readonly tonnageField: Locator;
 
   readonly lengthField: Locator;
 
-  readonly codeCompoField: Locator;
+  readonly speedLimitTagField: Locator;
 
   readonly maxSpeedField: Locator;
 
@@ -135,6 +147,10 @@ class STDCMPage {
 
   readonly viaResultMarker: Locator;
 
+  readonly simulationResultTable: Locator;
+
+  readonly simulationLengthAndDuration: Locator;
+
   readonly helpButton: Locator;
 
   constructor(page: Page) {
@@ -147,9 +163,10 @@ class STDCMPage {
     this.originCard = page.locator('.stdcm-card:has(.stdcm-origin-icon)');
     this.destinationCard = page.locator('.stdcm-card:has(.stdcm-destination-icon)');
     this.tractionEngineField = page.locator('#tractionEngine');
+    this.towedRollingStockField = page.locator('#towedRollingStock');
     this.tonnageField = page.locator('#tonnage');
     this.lengthField = page.locator('#length');
-    this.codeCompoField = page.locator('#speed-limit-by-tag-selector');
+    this.speedLimitTagField = page.locator('#speed-limit-by-tag-selector');
     this.maxSpeedField = page.locator('#maxSpeed');
     this.addViaButton = page.locator('.stdcm-vias-list button .stdcm-card__body.add-via');
     this.launchSimulationButton = page.getByTestId('launch-simulation-button');
@@ -193,7 +210,7 @@ class STDCMPage {
     this.dynamicOriginCi = this.originCard.locator('[id^="id"][id$="-ci"]');
     this.dynamicDestinationCi = this.destinationCard.locator('[id^="id"][id$="-ci"]');
     this.simulationStatus = page.getByTestId('simulation-status');
-    this.simulationList = page.locator('.simulation-list .simulation-name');
+    this.simulationList = page.locator('.stdcm-results .simulation-list');
     this.incrementButton = page.locator('.minute-button', { hasText: '+1mn' });
     this.allViasButton = page.getByTestId('all-vias-button');
     this.retainSimulationButton = page.getByTestId('retain-simulation-button');
@@ -206,6 +223,10 @@ class STDCMPage {
     this.originResultMarker = this.mapResultContainer.locator('img[alt="origin"]');
     this.destinationResultMarker = this.mapResultContainer.locator('img[alt="destination"]');
     this.viaResultMarker = this.mapResultContainer.locator('img[alt="via"]');
+    this.simulationResultTable = page.locator('.simulation-results table.table-results');
+    this.simulationLengthAndDuration = page.locator(
+      '.simulation-metadata .total-length-trip-duration'
+    );
   }
 
   // Dynamic selectors for via cards
@@ -213,11 +234,11 @@ class STDCMPage {
     return this.page.locator(`.stdcm-card:has(.stdcm-via-icons:has-text("${viaNumber}"))`);
   }
 
-  private getViaCh(viaNumber: number): Locator {
+  private getViaCH(viaNumber: number): Locator {
     return this.getViaCard(viaNumber).locator('[id^="id"][id$="-ch"]');
   }
 
-  private getViaCi(viaNumber: number): Locator {
+  private getViaCI(viaNumber: number): Locator {
     return this.getViaCard(viaNumber).locator('[id^="id"][id$="-ci"]');
   }
 
@@ -241,6 +262,16 @@ class STDCMPage {
   private async setHourLocator(hourValue: string) {
     const hourLocator = this.page.locator('.time-grid .hour', { hasText: hourValue });
     await hourLocator.click();
+  }
+
+  private getSimulationLengthAndDurationLocator(simulationNumber: number): Locator {
+    return this.simulationList
+      .locator('.simulation-metadata .total-length-trip-duration')
+      .nth(simulationNumber - 1);
+  }
+
+  private getSimulationNameLocator(simulationNumber: number): Locator {
+    return this.simulationList.locator('.simulation-name').nth(simulationNumber - 1);
   }
 
   private async verifySuggestions(expectedSuggestions: string[]) {
@@ -272,6 +303,7 @@ class STDCMPage {
   async verifyAllFieldsEmpty() {
     const emptyFields = [
       this.tractionEngineField,
+      this.towedRollingStockField,
       this.tonnageField,
       this.lengthField,
       this.maxSpeedField,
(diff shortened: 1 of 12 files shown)


front/tests/pages/stdcm-page-model.ts
```diff
@@ -413,24 +505,30 @@ class STDCMPage {
...
     }
   }
 
   // Launches the simulation and checks if simulation-related elements are visible
   async launchSimulation() {
+    await this.launchSimulationButton.waitFor();
     await expect(this.launchSimulationButton).toBeEnabled();
-    await this.launchSimulationButton.click();
-    await expect(this.simulationList).toBeVisible();
+    await this.page.waitForTimeout(500); // Small buffer
```
emersion: Can we avoid introducing more of these?

Maymanaf: I know using `waitForTimeout` is discouraged, but the application often has latency, especially in local tests. I updated the function to work in CI, but it's hard to confirm locally

emersion: If we need a `waitForTimeout()`, that means that we're missing a condition somewhere. In other words, instead of `waitForTimeout()`, we should wait for the app to update itself in another way (an element becomes visible, etc).

front/tests/pages/stdcm-page-model.ts
```diff
@@ -477,24 +573,20 @@ class STDCMPage {
...
+    await this.retainSimulationButton.click();
     await expect(this.downloadSimulationButton).toBeVisible();
     await expect(this.downloadSimulationButton).toBeEnabled();
     await expect(this.startNewQueryButton).toBeVisible();
   }
 
   async downloadSimulation(browserName: string) {
     // Wait for the download event
-    await this.page.waitForTimeout(500);
+    await this.page.waitForTimeout(1000);
```
Yohh: can't we use [expect.poll() and expect.toPass()](https://playwright.dev/docs/test-assertions#expectpoll) for those kind of situations?

I think we could avoid many `waitForTimeout()` this way.

front/tests/pages/stdcm-page-model.ts
```diff
@@ -383,20 +467,28 @@ class STDCMPage {
...
+    language?: string;
+  }) {
+    const translations = language === 'English' ? enTranslations : frTranslations;
     const warning = this.getViaWarning(viaNumber);
     // Helper function to fill common fields
     const fillVia = async (selectedSuggestion: Locator) => {
       await this.addViaButton.nth(viaNumber - 1).click();
       expect(await this.addViaButton.count()).toBe(viaNumber + 1);
-      await this.getViaCi(viaNumber).fill(viaSearch);
+      await this.getViaCi(viaNumber).fill(ciSearchText);
```
Yohh: nit: CI is an acronym, I think `getViaCI()` should be more explicit

front/tests/utils/api-setup.ts
```diff
@@ -260,3 +264,46 @@ export async function setStdcmEnvironment(stdcmEnvironment: StdcmSearchEnvironme
...
+    { page_size: 50 }
+  );
+  const towedRollingStock = towedRollingStocks.results.find(
+    (t: TowedRollingStock) => t.name === towedRollingStockName
+  );
+  return towedRollingStock as TowedRollingStock;
+};
+
+/**
+ * Creates a towed rolling stock using predefined data from the imported JSON file.
```
Yohh: nit: for consistence, make a choice between `Creates` and `Create`, you use the two kinds of spelling but basically we use `Create` in other comments

front/tests/utils/api-setup.ts
```diff
@@ -260,3 +264,46 @@ export async function setStdcmEnvironment(stdcmEnvironment: StdcmSearchEnvironme
...
+  towedRollingStockName: string
+): Promise<TowedRollingStock | undefined> => {
+  const towedRollingStocks: GetTowedRollingStockApiResponse = await getApiRequest(
+    '/api/towed_rolling_stock',
+    { page_size: 50 }
+  );
+  const towedRollingStock = towedRollingStocks.results.find(
+    (t: TowedRollingStock) => t.name === towedRollingStockName
+  );
+  return towedRollingStock as TowedRollingStock;
```
SharglutDev: You shouldn't need this

```suggestion
  return towedRollingStock;
```

front/tests/utils/api-setup.ts
```diff
@@ -260,3 +264,46 @@ export async function setStdcmEnvironment(stdcmEnvironment: StdcmSearchEnvironme
...
   );
 }
+
+/**
+ * Retrieve a towed rolling stock by name.
+ *
+ * @param towedRollingStockName - The name of the towed rolling stock to retrieve.
+ * @returns {Promise<TowedRollingStock >} - The matching towed rolling stock data .
+ */
+export const getTowedRollingStock = async (
```
SharglutDev: This would be more telling (and matches the syntax from editoast endpoints)
```suggestion
export const getTowedRollingStockByName = async (
```

front/tests/utils/api-setup.ts
```diff
@@ -260,3 +264,46 @@ export async function setStdcmEnvironment(stdcmEnvironment: StdcmSearchEnvironme
...
+  );
+  return towedRollingStock as TowedRollingStock;
+};
+
+/**
+ * Creates a towed rolling stock using predefined data from the imported JSON file.
+ *
+ * @returns {Promise<TowedRollingStock>} - The created towed rolling stock.
+ */
+export async function setTowedRollingStock(): Promise<TowedRollingStock> {
```
SharglutDev, line 292: This function is a little weird because we don't control the created towed rs.
As we always want to create the same one, should we just check if 'towedRollingStock.length > 0' ? Checking the name would be relevant if we had the possibility to create different ones.

front/tests/pages/stdcm-page-model.ts
```diff
@@ -294,15 +325,68 @@ class STDCMPage {
...
+        await dropdownField.dispatchEvent('blur');
+        await expect(dropdownField).toHaveValue(selectedValue);
+
+        // Verify the prefilled values
+        const { tonnage, length, maxSpeed } = prefilledValues;
+        await expect(this.tonnageField).toHaveValue(tonnage);
+        await expect(this.lengthField).toHaveValue(length);
+        await expect(this.maxSpeedField).toHaveValue(maxSpeed);
+      }
+    };
```
SharglutDev: If I understand correctly (didn't launch the test yet), this function take the values (length/mass/maxspeed) from a rs or towed rs and check that the input fields contain the values for these properties.

But below, if `fillAndVerifyConsistDetails` receives the prefilledValues from both a rs and towed rs, it first check that the files contain values from the rs and then from the towed rs. 
This shouldn't work for mass and length because they are (should be) added together and not just replaced.
For the max speed, the field keeps the lower value between the one from rs/towed rs/compo code when they are updated.

front/tests/pages/stdcm-page-model.ts
```diff
@@ -413,24 +505,29 @@ class STDCMPage {
...
     }
   }
 
   // Launches the simulation and checks if simulation-related elements are visible
   async launchSimulation() {
+    await this.launchSimulationButton.waitFor({ timeout: 30 * 1000 });
     await expect(this.launchSimulationButton).toBeEnabled();
-    await this.launchSimulationButton.click();
-    await expect(this.simulationList).toBeVisible();
+    await this.launchSimulationButton.click({ force: true });
```
SharglutDev: Why do we need to force ?

front/tests/pages/stdcm-page-model.ts
```diff
@@ -524,5 +616,36 @@ class STDCMPage {
...
+      ? `Simulation n°${simulationNumber}`
+      : translations.simulation.results.simulationName.withoutOutputs;
+
+    // Validate simulation name
+    const actualSimulationName =
+      await this.getSimulationNameLocator(simulationNumber).textContent();
+    expect(actualSimulationName).toEqual(expectedSimulationName);
+
+    // Determine expected length and duration
+    const expectedLengthAndDuration = isResultTableVisible ? simulationLengthAndDuration : '— ';
```
SharglutDev: How does it work if `simulationLengthAndDuration` isn't provided ? It's gonna be undefined and then you compare it with `actualLengthAndDuration` which can only be `string` or `null` so the test will fail.
If we know `simulationLengthAndDuration` will be defined if `isResultTableVisible` is true, maybe we should redesign the logic a bit.

front/tests/pages/stdcm-page-model.ts
```diff
@@ -230,6 +250,16 @@ class STDCMPage {
...
   }
 
+  private getSimulationLengthAndDurationLocator(simulationNumber: number): Locator {
+    return this.simulationList
+      .locator('.simulation-metadata .total-length-trip-duration')
+      .nth(simulationNumber - 1);
+  }
+
+  private getSimulationNameLocator(simulationNumber: number): Locator {
+    return this.simulationList.locator('.simulation-name').nth(simulationNumber - 1);
```
Yohh, line 274: couldn't it be `return this.simulationList.locator('.simulation-name').last();` ?

front/tests/pages/stdcm-page-model.ts
```diff
@@ -260,7 +291,7 @@ class STDCMPage {
       this.destinationChField,
     ];
     for (const field of emptyFields) await expect(field).toHaveValue('');
-    await expect(this.codeCompoField).toHaveValue('__PLACEHOLDER__');
+    await expect(this.speedLimitTagField).toHaveValue('__PLACEHOLDER__');
```
Yohh, line 316: you may have conflicts with [this pr](https://github.com/OpenRailAssociation/osrd/pull/9987)

front/tests/pages/stdcm-page-model.ts
```diff
@@ -294,15 +325,68 @@ class STDCMPage {
...
+      fields.tractionEngine
+    );
+
+    // Handle towed rolling stock field
+    await handleAndVerifyDropdown(
+      this.towedRollingStockField,
+      {
+        tonnage: prefilledTowedRollingStockTonnage || '',
+        length: prefilledTowedRollingStockLength || '',
+        maxSpeed: prefilledTowedRollingStockMaxSpeed || '',
```
Yohh: you could set a default `''` value

front/tests/pages/stdcm-page-model.ts
```diff
@@ -294,15 +325,68 @@ class STDCMPage {
...
-    await this.maxSpeedField.fill('180');
+  async fillAndVerifyConsistDetails(
+    fields: ConsistFields,
+    prefilledTractionEngineTonnage: string,
+    prefilledTractionEngineLength: string,
+    prefilledTractionEngineMaxSpeed: string,
+    prefilledTowedRollingStockTonnage?: string,
+    prefilledTowedRollingStockLength?: string,
+    prefilledTowedRollingStockMaxSpeed?: string
+  ): Promise<void> {
```
Yohh:
```suggestion
  async fillAndVerifyConsistDetails(
    fields: ConsistFields,
    prefilledTractionEngineTonnage: string,
    prefilledTractionEngineLength: string,
    prefilledTractionEngineMaxSpeed: string,
    prefilledTowedRollingStockTonnage?: string,
    prefilledTowedRollingStockLength?: string,
    prefilledTowedRollingStockMaxSpeed?: string
  ): Promise<void> {
  
    const { tractionEngine, towedRollingStock, tonnage, length, maxSpeed, speedLimitTag } = fields;

```

then you can remove all `fields.`

front/tests/pages/stdcm-page-model.ts
Maymanaf: If `simulationLengthAndDuration` isn't provided  the `expectedLengthAndDuration` will take `'— '` as `isResultTableVisible` can't be false if we don't have  `simulationLengthAndDuration`

front/tests/pages/stdcm-page-model.ts
Maymanaf: to avoid `intercepts pointer events`

front/tests/pages/stdcm-page-model.ts
Maymanaf: We can't use `last()` because this function is intended to test two simulations. It would be more useful when we enhance the test by adding functionality to navigate between simulations

front/tests/utils/api-setup.ts
Maymanaf: It's better to check by name because, if there are other tower rolling stocks besides the one designated for the E2E test, your logic might fail. This is especially important when running the tests locally with the dump

front/tests/pages/stdcm-page-model.ts
SharglutDev: Since your recent discover, will it be enough or should we change it for `dispatchEvent` ?

front/tests/pages/stdcm-page-model.ts
SharglutDev: That's what I thought but it's not documented anywhere so another dev manipulating this function might make a mistake, that's why I suggested to change the logic a bit. For example, we could check `simulationResultTable.isVisible` before calling `verifySimulationDetails` ? If the simulation fails, we won't any simulation number anyway : 

Capture d’écran 2024-12-19 à 18 32 46

I'm not sure to understandd why we make the first "without output" ?

front/tests/pages/stdcm-page-model.ts
Maymanaf: Done

front/tests/pages/stdcm-page-model.ts
Maymanaf: `click({ force: true })` is enough

front/tests/pages/stdcm-page-model.ts
Maymanaf: Changed the logic to go with #9712

front/tests/pages/stdcm-page-model.ts
Maymanaf: Done

front/tests/pages/stdcm-page-model.ts
Maymanaf: Changed the logic to go with #9712

front/tests/pages/stdcm-page-model.ts
Maymanaf: I use the simulation number to get the simulation name locator and simulationLengthAndDuration, if available.
I renamed `'—'` to n`oCapacityLengthAndDuration` to make it more explicit.
This is a combined test for cases "without output" and "towed rolling stock"

SharglutDev (approved):
Lgtm and tested, good job ! (still confused about my weird test fail for the help button in local)

Yohh (approved):
lgtm

PR #10030, opened 2024-12-12
editoast: better sql query to split a track section

### LINKED ISSUE #8947: Bug when splitting a track in infrastructure editor
### What happened?

I wanted to manually fix [this data error](https://github.com/osrd-project/osrd-dags/issues/426) by adding a switch to go to the `UNIQUE` track at the south east of Clisson.

First step was to split the track, so i did it:
image

image

But after the split the actual place where the track got split was very far away from where i asked:
image

I'm expecting the track to split where i asked it to be split.

1. Duplicate infra France
2. Go to the fork south-est of clisson
3. try to reproduce the steps above

Recette (SNCF), Integration (SNCF)

Google Chrome

v0.3.2 6ea7e56

sim51:
Fix #8947

New query is not using the `ST_LineSubstring` which gives bad result with geography data that has long segment. By computing the geo point and doing a split of the TS near this point, it's a lot better.

### DIFF
--- a/editoast/src/models/infra/sql/get_split_track_section_with_data.sql
+++ b/editoast/src/models/infra/sql/get_split_track_section_with_data.sql
@@ -1,27 +1,34 @@
-SELECT object_table.obj_id as obj_id,
-    object_table.data as railjson,
-    ST_AsGeoJSON(
-        ST_Transform(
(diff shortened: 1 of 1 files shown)


flomonster (approved):
LGTM

PR #10032, opened 2024-12-12
front: setting power restrictions don't remove pathSteps

### LINKED ISSUE #9361: Setting power restrictions totally overwrite the via points
### What happened?

I had an itinerary that i spent quite some time on:
Image

But when setting power restrictions, all my waypoints dissapeared and got replaced by the changing points of the power restrictions:
Image
Image

Some via points should be added where the power restriction changes but it should definitely not reset all the manually via points.

1. Create a path with manually set via points, or import this
[train_schedules (26).json](https://github.com/user-attachments/files/17393202/train_schedules.26.json)

2. Add power restrictions changes

Dev (SNCF)

Google Chrome

8a66782

clarani:
closes #9361 

When manipulating power restrictions, the pathSteps without any constraints (stop, arrival time or margin) were cleaned. This was the expected behaviour before, because we could not have pathSteps without constraints.

As the expected behaviour changed (we now authorize pathSteps without constraints), the pathSteps created from the power restrictions selector have a new flag and may be cleaned if they are not linked anymore to a power restriction and have no constraints. The other pathSteps will not be cleaned.

:warning: There is currently another bug which resets automatically the power restrictions when editing a train (#9902). It should be fixed in another PR (#9924).

To test:
- create a train between PE and SG (select a rolling stock with power restrictions)
- add some via on the path (don't add any stop, arrival times or margins)
- add some power restrictions: cut a range in 2 and select a value
- check in the itinerary tab that a new via has appeared (or 2) for the power restriction range, but the other vias should still be there
- remove the power restriction range
- check in the itinerary tab that the vias linked to the power restriction range have disappeared, but not the other ones.

### DIFF
--- a/front/src/reducers/osrdconf/operationalStudiesConf/utils.ts
+++ b/front/src/reducers/osrdconf/operationalStudiesConf/utils.ts
@@ -95,7 +95,7 @@ export const cleanPathSteps = (
   powerRestrictions: PowerRestriction[]
 ): PathStep[] =>
   pathSteps.reduce((acc, pathStep, index) => {
-    if (index === 0 || index === pathSteps.length - 1) {
+    if (index === 0 || index === pathSteps.length - 1 || !pathStep.isFromPowerRestriction) {
       acc.push(pathStep);
       return acc;
     }
--- a/front/src/reducers/osrdconf/types.ts
+++ b/front/src/reducers/osrdconf/types.ts
@@ -89,6 +89,14 @@ export type PathStep = PathItemLocation & {
     trackNumber: number;
   };
   isInvalid?: boolean;
+  /** Flag specifying whether the pathStep was created from the power restriction selector or not
+   *
+   * If true, the pathStep might be cleaned if its power restriction is removed (except if it has time, stop or margin constraints)
+   *
+   * This flag will only work if the user has not saved their change. Once the change is saved, the flag will be removed and the pathStep
+   * will become permanent.
+   */
+  isFromPowerRestriction?: boolean;
 };
 
 export type StdcmPathStep = {
--- a/front/src/modules/powerRestriction/helpers/createPathStep.ts
+++ b/front/src/modules/powerRestriction/helpers/createPathStep.ts
@@ -75,6 +75,7 @@ const createPathStep = (
     ...trackOffset,
     // TODO: we should return the offset in mm once it is stored in mm in the store
     offset: mmToM(trackOffset.offset),
+    isFromPowerRestriction: true,
   };
 };
 
@@ -124,6 +125,7 @@ export const createCutAtPathStep = (
     id: nextId(),
     positionOnPath: cutAtPosition,
     coordinates: coordinatesAtCut,
+    isFromPowerRestriction: true,
     ...trackOffset,
     // TODO: we should return the offset in mm once it is stored in mm in the store
     offset: mmToM(trackOffset.offset),


front/src/reducers/osrdconf/operationalStudiesConf/utils.ts
```diff
@@ -83,7 +83,7 @@ export const addPathStep = (pathSteps: PathStep[], newPathStep: PathStep): PathS
     const index = pathSteps.findIndex(
       (step) => step.positionOnPath && step.positionOnPath > newPathStep.positionOnPath!
     );
-    return addElementAtIndex(pathSteps, index, newPathStep);
+    return addElementAtIndex(pathSteps, index, { ...newPathStep, isFromPowerRestriction: true });
```
emersion: This flag will be set when the path step is created, but when reloading the page this won't work anymore right?

RomainValls (commented):
Found a little bug : 
Try cutting the range twice to add 2 power restrictions (ex: 0-50k ; 50k - 100k)
Cut the second restriction in two (like at 75k)
Check the pathsteps, everything is ok for now. 
Delete the newly created power restriction (the one from 50k to 75k) 
Delete the second power restriction (75k - 100k) 
We can still see the via at 75k

clarani:
You're right @RomainValls, thanks !
I forgot to add the flag when adding a new pathStep by "cutting a range", it's fixed now ;)

front/src/reducers/osrdconf/operationalStudiesConf/utils.ts
clarani: Yes, it won't work anymore, but it's the best way I found to avoid removing by mistake pathSteps which should not be removed :/

emersion: This will only result in false negatives, ie. after page reload path steps related to power restrictions won't be cleared. Since there is no data loss, that's fine.

front/src/applications/operationalStudies/hooks/useManageTrainScheduleContext.tsx
```diff
@@ -29,11 +34,27 @@ export const ManageTrainScheduleContextProvider = ({
...
+      const allWaypoints = upsertPathStepsInOPs(
+        pathProperties.suggestedOperationalPoints,
+        compact(pathSteps)
+      );
+      setPathProperties({
+        ...pathProperties,
+        allWaypoints,
+      });
+    }
+  }, [pathSteps]);
```
emersion: Why is this necessary now? What does this fix?

front/src/reducers/osrdconf/operationalStudiesConf/utils.ts
emersion: I'm a bit antsy to set this flag in a generic `addPathStep()` function… Could we maybe set it from the caller side?

front/src/reducers/osrdconf/operationalStudiesConf/utils.ts
```diff
@@ -95,7 +95,7 @@ export const cleanPathSteps = (
   powerRestrictions: PowerRestriction[]
 ): PathStep[] =>
   pathSteps.reduce((acc, pathStep, index) => {
-    if (index === 0 || index === pathSteps.length - 1) {
+    if (index === 0 || index === pathSteps.length - 1 || !pathStep.isFromPowerRestriction) {
```
emersion, line 98: Can we keep a TODO documenting the limitations of this approach?

front/src/applications/operationalStudies/hooks/useManageTrainScheduleContext.tsx
clarani: AllWaypoints is used to compute the inputs table.
Without this useEffect, if I add some vias from the power restrictions selector, then they are not added in allWaypoints and not displayed in the inputs table.

Note: I think it would be better to have a useMemo for allWaypoints, rather than to store them artificially in pathProperties

emersion: Agreed! 

(Side note, this `useEffect()` depends on `pathProperties.suggestedOperationalPoints`, but that's missing from the dependencies.)

emersion: > Note: I think it would be better to have a useMemo for allWaypoints, rather than to store them artificially in pathProperties

Done in https://github.com/OpenRailAssociation/osrd/pull/10169

front/src/reducers/osrdconf/operationalStudiesConf/utils.ts
clarani: Done

front/src/reducers/osrdconf/operationalStudiesConf/utils.ts
clarani: It's done :+1:

emersion (approved):

front/src/reducers/osrdconf/types.ts
```diff
@@ -89,6 +89,14 @@ export type PathStep = PathItemLocation & {
     trackNumber: number;
   };
   isInvalid?: boolean;
+  /** Flag specifying whether the pathStep was created from the power restriction selector or not
+   *
+   * If true, the pathStep might be cleaned if its power restriction is removed (except if it has time, stop or margin constraints)
+   *
+   * This flag will only work if the user has not saved its change. Once the change is saved, the flag will be removed and the pathStep
```
emersion: Nit: the "its" pronoun is not used for words referring to people, such as "the user". "[their](https://en.wikipedia.org/wiki/Singular_they)" can be used instead.

PR #10035, opened 2024-12-12
front: display simulation results only if a train is selected or get data is ready

### LINKED ISSUE #10031: My scenario tries to load some output table though i have no trains in it
### What happened?

I just created a scenario, so i have no trains in it, my the front end displays this:

Image

The scenrio should be completely empty.

I don't really know how to reproduce this unfortunately, it just happened. I'm guessing it's some kind of cache problem but i'm really not sure.

Edit : apperently opening any new scenario does that.

Recette (SNCF)

Google Chrome

9e75ce2

clarani:
closes #10031

### DIFF
--- a/front/src/applications/operationalStudies/views/SimulationResults.tsx
+++ b/front/src/applications/operationalStudies/views/SimulationResults.tsx
@@ -116,6 +116,10 @@ const SimulationResults = ({
     }
   }, [extViewport]);
 
+  if ((!selectedTrainSchedule || !trainSimulation) && !projectionData) {
+    return null;
+  }
+
   return (
     <div className="simulation-results">
       {/* SIMULATION : STICKY BAR */}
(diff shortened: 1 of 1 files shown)


front/src/applications/operationalStudies/views/SimulationResults.tsx
```diff
@@ -116,6 +116,10 @@ const SimulationResults = ({
     }
   }, [extViewport]);
 
+  if (!selectedTrainSchedule && !projectionData) {
```
SharglutDev: Should we be more restrictive ? Sometimes there is a bug where we see the simulation results but not the space time chart (and we need to change the projection in order to display it)
```suggestion
  if (!selectedTrainSchedule || !projectionData) {
```

SharglutDev (commented):
Works great, just one question :)

theocrsb (approved):
LGTM and tested, but I agree with @SharglutDev :)

front/src/applications/operationalStudies/views/SimulationResults.tsx
clarani: Ah I had in mind that as the SpaceTimeChart does not depend on the selected train, it should not disappear if the simulation of the selectedTrain is loading.

But I can hide all the results below the SpaceTimeChart if we have no selected train ?

front/src/applications/operationalStudies/views/SimulationResults.tsx
```diff
@@ -137,16 +141,16 @@ const SimulationResults = ({
...
       {/* SIMULATION : SPACE TIME CHART */}
-      <ResizableSection
-        height={manchetteWithSpaceTimeChartHeight}
-        setHeight={setManchetteWithSpaceTimeChartHeight}
-        minHeight={MANCHETTE_WITH_SPACE_TIME_CHART_DEFAULT_HEIGHT}
-      >
-        <div
-          className="simulation-warped-map d-flex flex-row align-items-stretch mb-2"
-          style={{ height: manchetteWithSpaceTimeChartHeight }}
+      {projectionData && projectionData.projectedTrains.length > 0 && (
```
SharglutDev: Hmm with this condition, when we switch projection, the space taken by the space time disappears resulting with an unpleasant animation where the rest of the simulation goes up

front/src/applications/operationalStudies/views/SimulationResults.tsx
SharglutDev: Why not, but we shouldn't do that for "no projection" (see other comment)

front/src/applications/operationalStudies/views/SimulationResults.tsx
clarani: You're right, I revert my changes :+1:

SharglutDev (commented):
If I delete the selected train (which is not the projected one), it seems there is no new one selected (or probably it goes on an invalid one) and we keep seeing the map/output table loader.


https://github.com/user-attachments/assets/2acba9f3-f186-46cd-8fb1-364e1f9ec057

clarani:
You're right, I think it's fixed like this, what do you think ?

SharglutDev (approved):
Thank you for the last change, everything looks great !

PR #10039, opened 2024-12-12
front: lmr prefill consist max speed from tag

### LINKED ISSUE #10016: Link the max speed of the consist to the composition code in STDCM
# Description

> [!IMPORTANT]
> Updated after [this comment](https://github.com/OpenRailAssociation/osrd/pull/10039#discussion_r1887193308)

Leftover work from https://github.com/OpenRailAssociation/osrd/issues/9712

Composition codes can affect the max speed of a train. For freight trains, here is the mapping : 

| Code  | Max speed (km/h) |
|-------|------------------|
| HLP   | 100                 |
| MA80  | 80               |
| MA90  | 90               |
| MA100 | 100              |
| ME100 | 100              |
| ME120 | 120              |
| ME140 | 140              |
| ME160 | 160              |
| MVGV  |  200                |

# AC

- [ ] The table above has been implemented and changing the composition code update the max speed if the chosen code is supposed to lower it. User's input are still not overwritten.

Wadjetz:
fix #10016

Prefill consist max speed from speed limit tag

### DIFF
--- a/front/src/applications/stdcm/consts.ts
+++ b/front/src/applications/stdcm/consts.ts
@@ -20,3 +20,15 @@ export const COMPOSITION_CODES = [
   'MV160',
   'MVGV',
 ];
+
+export const COMPOSITION_CODES_MAX_SPEEDS: Record<string, number | undefined> = {
+  MA80: 80,
+  MA90: 90,
+  MA100: 100,
+  ME100: 100,
+  ME120: 120,
+  ME140: 140,
+  ME160: 160,
+  HLP: 100,
+  MVGV: 200,
+};
--- a/front/src/applications/stdcm/hooks/useStdcmConsist.ts
+++ b/front/src/applications/stdcm/hooks/useStdcmConsist.ts
@@ -8,7 +8,9 @@ import { useOsrdConfActions, useOsrdConfSelectors } from 'common/osrdContext';
 import { type StdcmConfSliceActions } from 'reducers/osrdconf/stdcmConf';
 import type { StdcmConfSelectors } from 'reducers/osrdconf/stdcmConf/selectors';
 import { useAppDispatch } from 'store';
-import { kgToT, msToKmh } from 'utils/physics';
+import { kgToT, kmhToMs, msToKmh } from 'utils/physics';
+
+import maxSpeedFromSpeedLimitByTag from '../utils/maxSpeedFromSpeedLimitByTag';
 
 const useStdcmConsist = () => {
   const dispatch = useAppDispatch();
@@ -45,7 +47,8 @@ const useStdcmConsist = () => {
 
   const prefillConsist = (
     rollingStock?: LightRollingStockWithLiveries,
-    towed?: TowedRollingStock
+    towed?: TowedRollingStock,
+    maxSpeedTag?: string | null
   ) => {
     if (!totalMassChanged) {
       const consistMass = Math.floor(kgToT((rollingStock?.mass ?? 0) + (towed?.mass ?? 0)));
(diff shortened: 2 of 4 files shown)


theocrsb:
Shouldn't the speed be reset when -none- is clicked?

front/src/applications/stdcm/utils/maxSpeedFromSpeedLimitByTag.ts
```diff
@@ -0,0 +1,24 @@
...
+    MA80: 80,
+    MA90: 90,
+    MA100: 100,
+    ME100: 100,
+    ME120: 120,
+    ME140: 140,
+    ME160: 160,
+    HLP: undefined,
+    MVGV: undefined,
+  };
```
SharglutDev: We don't want to redefine this every time we call this function. We can move that in `stdcm/consts.ts`

front/src/applications/stdcm/hooks/useStdcmConsist.ts
```diff
@@ -58,7 +61,12 @@ const useStdcmConsist = () => {
     }
 
     if (!maxSpeedChanged) {
-      const consistMaxSpeed = min([rollingStock?.max_speed, towed?.max_speed]);
+      const maxSpeedFromTag = maxSpeedFromSpeedLimitByTag(maxSpeedTag);
```
SharglutDev, line 64: I think that's not the expected behavior, there is this mention in the issue : `if the chosen code is supposed to lower it.`. So for example, switching from MA80 to MA120 shouldn't affect the vmax.
Is it what you meant @axrolld ?

Wadjetz:
> Shouldn't the speed be reset when -none- is clicked?

@theocrsb It resets the value if neither a traction engine nor towed rolling stock is selected. If one of these values is selected, it uses the selected value.

front/src/applications/stdcm/consts.ts
```diff
@@ -20,3 +20,15 @@ export const COMPOSITION_CODES = [
...
+export const COMPOSITION_CODES_MAX_SPEEDS: Record<string, number | undefined> = {
+  MA80: 80,
+  MA90: 90,
+  MA100: 100,
+  ME100: 100,
+  ME120: 120,
+  ME140: 140,
+  ME160: 160,
+  HLP: undefined,
+  MVGV: undefined,
```
SharglutDev: Checked with @axrolld and @loic-hamelin 

```suggestion
export const COMPOSITION_CODES_MAX_SPEEDS = {
  MA80: 80,
  MA90: 90,
  MA100: 100,
  ME100: 100,
  ME120: 120,
  ME140: 140,
  ME160: 160,
  HLP: 100,
  MVGV: 200,
```

front/src/applications/stdcm/hooks/useStdcmConsist.ts
SharglutDev: Ok behavior seems perfect !

SharglutDev: Just wondering, on my computer the tower rollingstock endpoint returns an empty array. Same for you ?

Wadjetz: No, on my computer, the towed endpoint returns data.

SharglutDev (approved):
Lgtm and tested, good job ! Don't forget to fixup your last commit :)

maelysLeratRosso:
This is the mapping I observe on staging 

| Code | vMax |
| -------|-------|
| HLP | 100 |
| HLP | 100 |
| HLP | 100 |
| HLP | 100 |
| HLP | 100 |
| HLP | 100 |
| HLP | 100 |
| HLP | 100 |

PR #10078, opened 2024-12-13
front: deactivate marker check for e2e tests

Maymanaf:
Deactivate marker check for e2e tests until osm server is up again

### DIFF
--- a/front/tests/010-op-route-tab.spec.ts
+++ b/front/tests/010-op-route-tab.spec.ts
@@ -58,7 +58,7 @@ test.describe('Route Tab Verification', () => {
   );
 
   /** *************** Test 1 **************** */
-  test('Select a route for operational study', async ({ page, browserName }) => {
+  test('Select a route for operational study', async ({ page }) => {
     const operationalStudiesPage = new OperationalStudiesPage(page);
     const routePage = new RoutePage(page);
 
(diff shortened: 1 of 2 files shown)


front/tests/010-op-route-tab.spec.ts
```diff
@@ -92,50 +93,51 @@ test.describe('Route Tab Verification', () => {
...
     await routePage.clickOnDeleteOPButtons(OSRDLanguage);
     await routePage.verifyNoSelectedRoute(OSRDLanguage);
 
     // Perform pathfinding again and verify map markers in Chromium
-    await routePage.performPathfindingByTrigram('WS', 'SES', 'MWS');
-    if (browserName === 'chromium') {
-      await routePage.verifyMapMarkers(...expectedMapMarkersValues);
-    }
+    // TODO: Uncomment this part when osm server is up again
+    // await routePage.performPathfindingByTrigram('WS', 'SES', 'MWS');
```
emersion: This line can stay uncommented I think

emersion (approved):
Nit: instead of commenting, `if (false)` could be used to reduce the diff and avoid TypeScript errors due to unused variables. But I'm fine with commenting as well.

clarani (approved):
LGTM

Yohh (approved):

PR #10079, opened 2024-12-13
front: show map markers/itinerary when OSM map is unavailable

emersion:
When the OSM map background is unavailable, we can still display the map markers and itinerary.

Reverts part of b038d3eba049 ("front: display itinerary markers name when editing a train").


theocrsb (approved):
LGTM and tested ! Thank you because I'm going to need this PR ;)

clarani (approved):
LGTM

PR #10080, opened 2024-12-13
front: update top positon of compass in map buttons

theocrsb:

Check that the last button and the compass button are centered.

The 4px shift is that in the previous PR we used a 4px margin block and now simply an 8px margin bottom.


Akctarus (approved):
tested, LGTM !

SharglutDev (approved):
Lgtm and tested

PR #10084, opened 2024-12-16
front: update typescript-eslint

SharglutDev:
- Remove the ban-types related to React.FC in .eslintrc as ban-types is
not supported anymore and as we don't use RC in the appr anymore anyway.
- Remove some unneeded asserts
- Error in a catch now need to be handled, use the error in setFailure
  to display notifications to the user


front/src/applications/editor/components/EntitySumUp.tsx
```diff
@@ -61,6 +63,7 @@ async function getAdditionalEntities(
         try {
           return { [trackId]: await getEntity(infra, trackId, 'TrackSection', dispatch) };
         } catch (e) {
+          setFailure(castErrorToFailure(e));
```
anisometropie: is there a reason we were ignoring all these errors, and that what shouldn’t ignore them anymore ?

front/src/applications/editor/tools/switchEdition/utils.ts
```diff
@@ -29,7 +29,7 @@ export function isSwitchValid(entity: Partial<SwitchEntity>, type: SwitchType):
 /**
  * "Flat switch" management:
  */
-export const FLAT_SWITCH_PORTS_PREFIX = 'port::' as const;
```
anisometropie, line 32: any idea why this was here in the first place?

front/src/common/uploadFileModal.tsx
```diff
@@ -28,7 +30,12 @@ const UploadFileModal = ({ handleSubmit }: UploadFileModalProps) => {
       }
       return undefined;
     } catch (error) {
-      return t('errorMessages.errorInvalidXMLFormat').toString();
+      setFailure(
+        castErrorToFailure(error, {
+          message: t('errorMessages.errorInvalidXMLFormat'),
+        })
+      );
+      return undefined;
```
anisometropie: The function always returns `undefined` so the `if (xmlError) {` line 62 is never entered

anisometropie: I guess remove the 2 returns + the if l:62

anisometropie (changes requested):

Uriel-Sautron (approved):
LGTM.

front/src/applications/editor/tools/switchEdition/utils.ts
SharglutDev: Maybe this was planned to be used with an enum at some point or something like that

front/src/applications/editor/components/EntitySumUp.tsx
SharglutDev: the new version requires now that we handle `e` in a catch. Instead of removing this rule, or just adding a console.error(e), @emersion suggested that it couldn't hurt to display the message to the user as well.

front/src/common/uploadFileModal.tsx
SharglutDev: Done

front/.eslintrc
```diff
@@ -46,22 +46,6 @@
     "@typescript-eslint/no-unnecessary-type-assertion": "error",
     "@typescript-eslint/no-unsafe-return": "error",
     "@typescript-eslint/no-unsafe-call": "error",
-
-    "@typescript-eslint/ban-types": [
```
emersion, line 51: `ban-types` has been superseded by `no-restricted-types` it seems?

https://typescript-eslint.io/rules/no-restricted-types

front/src/common/uploadFileModal.tsx
```diff
@@ -41,21 +46,18 @@ const UploadFileModal = ({ handleSubmit }: UploadFileModalProps) => {
...
       } else if (
         fileToValidate.type === 'application/railml' ||
         fileToValidate.name.endsWith('.railml')
       ) {
         const fileContent = await fileToValidate.text();
-        const xmlError = parseXML(fileContent);
-        if (xmlError) {
-          return xmlError;
-        }
+        parseXML(fileContent);
```
emersion: Hm, so when `parseXML` fails, we return `undefined` indicating no error?

front/src/common/Map/Layers/blankStyle.ts
```diff
@@ -13,6 +15,7 @@ const isValidUrl = async (url: string) => {
       const response = await fetch(`${url}.${extension}`);
       return response.ok;
     } catch (error) {
+      setFailure(castErrorToFailure(error));
```
emersion: This is one of the cases where we don't want to bubble the error to the user, I think.

front/src/applications/editor/components/EntitySumUp.tsx
emersion: Right, but this code was maybe intentionally ignoring errors?

front/.eslintrc
SharglutDev: Yes exactly. I started to replace it but I noticed we don't use FC anymore in the app. 
Do you want to keep this rule in case someone was tempted to use one ?

front/src/common/uploadFileModal.tsx
SharglutDev: I kept it because it seems it can still throw an error if some conditions are met

front/.eslintrc
emersion: We added the rule back when we removed all FC uses to make sure we don't re-introduce them by mistake IIRC.

front/src/common/Map/Layers/blankStyle.ts
SharglutDev: Done

front/.eslintrc
SharglutDev: Put the rule back (will edit the commit description before merging)

front/src/applications/editor/components/EntitySumUp.tsx
SharglutDev: For some of them, there were  `TODO`s to remind if we wanted to display something or not. Not sure what we do then ? Displaying to the user, only dev or just eslint-disabled these ?

SharglutDev: Ok so after all, decided to keep the old behavior for the infra editor and not display anything to the user, just to devs to avoid the eslint warning

anisometropie (approved):

front/src/common/uploadFileModal.tsx
emersion: What I mean is that the original code worked fine, no?

With the new code, `validateFile()` no longer returns an error when it fails. There are cases where it returns `undefined` on XML parsing errors.

SharglutDev: Now if `parseXML` fails, instead of returning a string, we return a notification. We discussed with @RomainValls and we weren't sure where was the string displayed anyway ? Does it pop next to the input ? We couldn't simulate a failing import.
That's why I changed it but if you say we can properly see the error somewhere,I can revert this.

SharglutDev: Will probably wait for this [PR](https://github.com/OpenRailAssociation/osrd/pull/10139) to be merged because a lot of things are changing in these files

emersion: The string is supposed to be displayed under the file input. See `isValid`.

I don't think an XML format error should be treated differently from an unsupported file type error, a JSON error or an empty file error. Either all should display a string under the input (status quo), either all should show an error bubble.

SharglutDev: Rebased after merge of #10139, should be good now.

emersion (approved):

PR #10087, opened 2024-12-16
front: replace casts with assertions

emersion:
This is safer (we get a type error if a type is changed), and more readable.

Enable [@typescript-eslint/non-nullable-type-assertion-style](https://typescript-eslint.io/rules/non-nullable-type-assertion-style/) to make sure we don't introduce more of these.


SharglutDev (approved):
Lgtm and tested, thank you for this

PR #10089, opened 2024-12-16
front: stop hardcoding navbar height in scenario

emersion:
Drop `--navbar-height`. Replace it with full-height containers, and simplify the overall structure.

Fixes a blank margin below the conflict list on smaller screens.

This PR is better reviewed with the "Hide whitespace" option enabled in the GitHub diff view settings.

Depends on https://github.com/OpenRailAssociation/osrd/pull/9869

### DIFF
--- a/front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
+++ b/front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
@@ -101,107 +101,113 @@ const ScenarioContent = ({
   };
 
   return (
-    <main className="mastcontainer mastcontainer-no-mastnav">
-      <div className="scenario">
-        <div className="row no-gutters scenario-container">
-          <div
-            data-testid="scenario-sidemenu"
-            className={collapsedTimetable ? 'd-none' : 'col-hdp-3 col-xl-4 col-lg-5 col-md-6'}
-          >
-            <div className="scenario-sidemenu">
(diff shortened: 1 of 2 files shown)


theocrsb (approved):
LGTM and tested :)

front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
```diff
@@ -101,107 +101,113 @@ const ScenarioContent = ({
...
-                      upsertTrainSchedules={upsertTrainSchedules}
-                      trainIdToEdit={trainIdToEdit}
-                      setTrainIdToEdit={setTrainIdToEdit}
-                      infraState={infra.state}
-                      dtoImport={dtoImport}
-                    />
-                  )}
-                  <Timetable
+    <main className="mastcontainer mastcontainer-no-mastnav scenario">
+      <div className="row no-gutters scenario-container h-100">
```
SharglutDev:
```suggestion
      <div className="row no-gutters h-100">
```

SharglutDev (approved):
Nice ! Lgtm and tested. You can close the comment after fixing it

front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
emersion: Indeed, thanks!

PR #10091, opened 2024-12-16
front: remove useless requests when deleting projects, studies, scenari

### LINKED ISSUE #7752: Removing a project/study/scenario fires and endpoint for nothing
### What happened?

When deleting one of these 3 elements, as the related rtk query tag is invalidated, the query to refetch this element is fired automatically.

image

The fetch is called because when the tag is invalidated, even though we are redirected to another page, it happens instantly before the redirection.

We should redirect to the other page and close the modal regardless of the success of the delete endpoint or not (validated by the ux @thibautsailly). This way, the endpoint won't be fired.
This behavior happens in all delete functions in AddOrEditModals from project/study/scenario.

1. Go to a scenario
2. Open the network tab in devtools and clean it
3. Delete the scenario
4. See that there has been a getScenarioById fired with a 404

- **Browser**: Firefox v126
- **OS**: Ubuntu
- **Env**: Dev

dev 0f97f32

### LINKED ISSUE #10008: front: unwanted request is sent when deleting a project
### What happened?

When we delete a project, an error is displayed in the console:
Image

RTK tries to fetch the deleted project, but it does not have any sense anymore.

This bug blocks @Maymanaf.

No request to fetch the deleted project should be sent.

1. Create a project
2. Open it and open the console
3. Delete the project
4. See the error in the console

Recette (SNCF)

Firefox

.

clarani:
closes #10008 
closes #7752

### DIFF
--- a/front/src/common/api/osrdEditoastApi.ts
+++ b/front/src/common/api/osrdEditoastApi.ts
@@ -42,24 +42,92 @@ const osrdEditoastApi = generatedEditoastApi.enhanceEndpoints({
       // we don't want to invalidate the trainschedule tag here to prevent multiple calls
       invalidatesTags: ['timetable', 'scenarios'],
     },
-    // Invalidate the children count and last update timestamp
+
+    // Project handling
+    getProjects: {
+      providesTags: (result) => [
+        { type: 'projects', id: 'LIST' },
+        ...(result?.results || []).map((project) => ({
+          type: 'projects' as const,
+          id: project.id,
+        })),
+      ],
+    },
+    getProjectsByProjectId: {
+      providesTags: (_result, _error, args) => [{ type: 'projects', id: args.projectId }],
+    },
+    postProjects: {
+      invalidatesTags: [{ type: 'projects', id: 'LIST' }],
+    },
+    patchProjectsByProjectId: {
+      invalidatesTags: (_result, _error, args) => [{ type: 'projects', id: args.projectId }],
+    },
+    deleteProjectsByProjectId: {
+      invalidatesTags: [{ type: 'projects', id: 'LIST' }],
+    },
+
+    // Studies handling
+    getProjectsByProjectIdStudies: {
+      providesTags: (result) => [
+        { type: 'studies', id: 'LIST' },
+        ...(result?.results || []).map(({ id }) => ({
+          type: 'studies' as const,
+          id,
+        })),
+      ],
+    },
+    getProjectsByProjectIdStudiesAndStudyId: {
+      providesTags: (_result, _error, args) => [{ type: 'studies', id: args.studyId }],
+    },
     postProjectsByProjectIdStudies: {
-      invalidatesTags: ['studies', 'projects'],
+      invalidatesTags: (_result, _error, args) => [
+        { type: 'projects', id: args.projectId },
+        { type: 'studies', id: 'LIST' },
+      ],
     },
     patchProjectsByProjectIdStudiesAndStudyId: {
-      invalidatesTags: ['studies', 'projects'],
+      invalidatesTags: (_result, _error, args) => [
+        { type: 'projects', id: args.projectId },
+        { type: 'studies', id: args.studyId },
+      ],
     },
     deleteProjectsByProjectIdStudiesAndStudyId: {
-      invalidatesTags: ['studies', 'projects'],
+      invalidatesTags: (_result, _error, args) => [
+        { type: 'projects', id: args.projectId },
+        { type: 'studies', id: 'LIST' },
+      ],
+    },
+
+    // Scenari handling
+    getProjectsByProjectIdStudiesAndStudyIdScenarios: {
+      providesTags: (result) => [
+        { type: 'scenarios', id: 'LIST' },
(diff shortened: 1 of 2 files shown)


front/src/modules/project/components/AddOrEditProjectModal.tsx
```diff
@@ -216,12 +216,14 @@ export default function AddOrEditProjectModal({
       cleanLocalStorageByProject(project!.id, projectStudies, dispatch);
     }
 
+    // navigate before deleting because otherwise RTK fires a new request to fetch the deleted project
+    navigate(`/operational-studies/projects/`);
```
SharglutDev: Not sure how to do better, but here we navigate regardless of the success of the deletion. How would you do @emersion ?

emersion: Unfortunately, it doesn't seem like there is a very good answer to this particular problem: https://github.com/reduxjs/redux-toolkit/issues/1672

SharglutDev: This topic is interesting. We already discussed some month ago about using the syntax mentionned in the comments.
https://redux-toolkit.js.org/rtk-query/usage/automated-refetching#invalidating-cache-data

This would be a larger refacto but as this issue isn't critical, this could be a good moment to take the time to test this logic in `osrdEditoastApi.ts` for these endpoints which aren't affecting a large part of the app.

front/src/common/api/osrdEditoastApi.ts
```diff
@@ -42,24 +42,115 @@ const osrdEditoastApi = generatedEditoastApi.enhanceEndpoints({
...
+        { type: 'studies', id: 'LIST' },
+        { type: 'studies', id: args.studyId },
+        { type: 'scenarios', id: 'LIST' },
+        { type: 'scenarios', id: args.scenarioId },
+      ],
     },
     deleteProjectsByProjectIdStudiesAndStudyIdScenariosScenarioId: {
-      invalidatesTags: ['scenarios', 'studies', 'projects'],
+      invalidatesTags: (_result, _error, args) => [
+        { type: 'projects', id: 'LIST' },
```
emersion: Why does this invalidate `{ type: 'projects', id: 'LIST' }`? This request doesn't create/delete a project right?

front/src/common/api/osrdEditoastApi.ts
```diff
@@ -42,24 +42,115 @@ const osrdEditoastApi = generatedEditoastApi.enhanceEndpoints({
...
     },
-    // Invalidate the children count and last update timestamp
+
+    // Project handling
+    getProjects: {
+      providesTags: (result) => [
+        'projects',
+        { type: 'projects', id: 'LIST' },
+        ...(result?.results || []).map((project) => ({
+          type: 'projects' as const,
```
emersion, line 51: Sometimes we're using `as const`, sometimes not - do we need it?

front/src/common/api/osrdEditoastApi.ts
SharglutDev, line 1: Do we need all this invalidation each time ?

- If I delete a scenario (I'm in the scenario page, this will be redirect me to the scenarios list page), I expect to refetch only the scenarios list (`{ type: 'scenarios', id: 'LIST' },`) but not the project or studies 
- If I edit a scenario, I expect to refetch only the related scenario id : `{ type: 'scenarios', id: args.scenarioId },`
- If I add a new scenario, I think I expect nothing to happen as I am redirected to the scenario page

Same for project and studies.

SharglutDev:
Ping :)

SharglutDev (approved):
Lgtm and tested, nice !

One thing though, we use a custom hook called `useMultiSelection` that update the list of project/study/scenario regardless of what rtk-query does/return. It was necessary before because of our flaky use of the endpoints but now that you made these changes, we probably don't need this other item list state. This could be handled in another PR ?

clarani:
Yes sure, this can be handled in another PR :+1:

front/src/common/api/osrdEditoastApi.ts
clarani: As seen this morning, ts requires the `as const` in the `map`

emersion: Yeah, plus this is what's used in the official docs!

front/src/common/api/osrdEditoastApi.ts
```diff
@@ -42,24 +42,101 @@ const osrdEditoastApi = generatedEditoastApi.enhanceEndpoints({
       // we don't want to invalidate the trainschedule tag here to prevent multiple calls
       invalidatesTags: ['timetable', 'scenarios'],
     },
-    // Invalidate the children count and last update timestamp
+
+    // Project handling
+    getProjects: {
+      providesTags: (result) => [
+        'projects',
```
emersion: Do we need to keep the bare `projects` tag? I think the two tag types defined below should be enough?

It seems like this tag is not used: it's not invalidated by mutations below.

emersion (approved):
LGTM!

PR #10096, opened 2024-12-16
front: fix typo in classname

### LINKED ISSUE #9594: Viriato import with 0 OP matched

RomainValls:
close #9594 

I've let a typo in the previous feature fixing it.

### DIFF
--- a/front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleTrainDetail.tsx
+++ b/front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleTrainDetail.tsx
@@ -78,7 +78,7 @@ export default function ImportTrainScheduleTrainDetail({
         </span>
         <span
           className={`import-train-schedule-traindetail-stepnb ${
-            trainData.steps.length <= 2 ? 'import-invalid-stepnb' : 'bg-primary'
+            trainData.steps.length <= 2 ? 'import-invalid-step-nb' : 'bg-primary'
           }`}
         >
           {trainData.steps.length - 2}


theocrsb (approved):
This is a detail, but perhaps we could call the `import-invalid-stepnb` class to match the parent. 
But otherwise the PR works fine :)

emersion (approved):

RomainValls:
> This is a detail, but perhaps we could call the `import-invalid-stepnb` class to match the parent. But otherwise the PR works fine :)

I agree ! But I remember @clarani requested this change in the last pull request i sent. What do you think @clarani ?

clarani (approved):

clarani:
Yes I think it's easier to read like this, we should rather change the parent class :)

PR #10105, opened 2024-12-17
front: drop pathfinding error message bubble

emersion:
The "pathfindingError" translation needs an "errorMessage" property, so it's not displayed properly.
    
Since the error is already displayed to the user in a different way, we can just drop the bubble.

To reproduce:

- Select a train in small_infra
- Switch to another infra (e.g. France)
- Open the "Create new train" modal


clarani (approved):
LGTM, I think you can even remove this warning

emersion:
Updated to drop the error bubble.

PR #10108, opened 2024-12-17
front: upgrade turf to v7

emersion:
_See individual commits._

Similar patch for osrd-ui, for reference:
https://github.com/OpenRailAssociation/osrd-ui/pull/670


SharglutDev (approved):
Lgtm and tested

PR #10110, opened 2024-12-17
editoast: use 'const' syntax from 'tracing' for opentelemetry semantic convention

woshilapin:
We can actually use variables as keys in span metadata, as long as they're `const`. See [the documentation](https://docs.rs/tracing/latest/tracing/index.html#recording-fields).


emersion (approved):
Nice!

hamz2a (approved):
Thanks!

leovalais (approved):
Much better!

PR #10113, opened 2024-12-17
editoast: 'OtelAxumLayer' act as 'INFO' level

woshilapin:
By default, the 'OtelAxumLayer' only creates span at 'TRACE' level. That means that if you try `tracing::Span::current()` in a handler, you will get a `Span::none()` in response. This span is useless (it doesn't have a subscriber, no metadata): it can't be used with OpenTelemetry since all interactions between 'tracing' and 'opentelemetry' happens through `Span::with_subscriber()`.

### DIFF
--- a/editoast/Cargo.toml
+++ b/editoast/Cargo.toml
@@ -94,7 +94,9 @@ axum-extra = { version = "0.9.6", default-features = false, features = [
   "typed-header",
 ] }
 axum-test = { version = "16.4.1", default-features = false }
-axum-tracing-opentelemetry = { version = "0.24.1", default-features = false }
+axum-tracing-opentelemetry = { version = "0.24.1", default-features = false, features = [
+  "tracing_level_info",
+] }
 chrono.workspace = true
 clap = { version = "4.5.23", features = ["derive", "env"] }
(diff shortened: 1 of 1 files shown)


leovalais (approved):
LGTM.

---

As discussed yesterday, I wonder if we should really rely on trace IDs for our non logging-related features. I believe editoast should behave the same way regardless of the current `RUST_LOG` value. As you said, a middleware ensuring a trace ID is created could fix this issue.

woshilapin:
> As discussed yesterday, I wonder if we should really rely on trace IDs for our non logging-related features. I believe editoast should behave the same way regardless of the current RUST_LOG value. As you said, a middleware ensuring a trace ID is created could fix this issue.

I agree. But in the case of STDCM debug logging, it's actually a logging functionality so I find it fine to tie this functionality to the logging framework. Actually, the STDCM debug logging functionality is configured (activated or not) through `RUST_LOG`.

That said, I believe the current PR has not relation with the STDCM logging functionality, we should not create traces only at `TRACE` level for our endpoints (or at least, I'm waiting for someone to come explain to me why it would make sense).

PR #7511, opened 2024-05-24
Add utils to TestApp and tracing setup for unit tests

leovalais:

PR #10122, opened 2024-12-18
editoast: setup tracing with opentelemetry for `TestApp`

### LINKED ISSUE #10123: Editoast: setup tracing with opentelemetry for `TestApp`
### Description and goal

Refactor `init_tracing` to accept a custom `SpanExporter` for better flexibility, including support for testing scenarios

hamz2a:
closes https://github.com/OpenRailAssociation/osrd/issues/10123

I added the `OpenTelemetry` layer in `TestApp` to capture the `trace_id` set in the header, so we can test the endpoint we'll add in the [ticket](https://github.com/OpenRailAssociation/osrd/issues/9724).

- Refactored `init_tracing` to accept a custom `SpanExporter` for better flexibility, including support for testing scenarios.
- Updated `main.rs` to configure and initialize tracing.
- Modified setup tracing in `test_app.rs` to use opentelemetry layer.


editoast/src/main.rs
```diff
@@ -116,11 +117,21 @@ fn init_tracing(mode: EditoastMode, telemetry_config: &client::TelemetryConfig)
...
         }
     };
+
+    if for_test {
+        let subscriber = tracing_subscriber::registry()
+            .with(telemetry_layer)
+            .with(env_filter_layer)
+            .with(fmt_layer);
+        return Some(tracing::subscriber::set_default(subscriber));
+    }
```
woshilapin: Did you try to entirely remove that? [`set_default` set a subscriber only for the current scope](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/util/trait.SubscriberInitExt.html#method.set_default). But below, we use `.init()` which will globally set up the subscriber. Do we need a subscriber per-test? Maybe it's because `.init()` will panic if call multiple time in the same process and since, all tests are launch in parallel in the same process, it doesn't work?

In any case, I believe it should be possible to keep the code identical for production code and test code. I see 2 options.

1. Either we use `set_default` everywhere and we keep the guard in the `run()` function as well as in `TestApp` so the subscriber lives for the entire application (but as mentioned [here](https://docs.rs/tracing/0.1.21/tracing/dispatcher/index.html#setting-the-default-subscriber), it will not work well with `thread::spawn`, I wonder about `tokio::spawn`?). 
2. We use `try_init` instead of `init()`. It should work the same for production code, and for tests, only the first test will initialize the subscriber.

One last thing, I believe we don't need `for_test` variable, we can use `#[cfg(test)]` to detect if we're in tests or not.

editoast/Cargo.toml
```diff
@@ -94,7 +94,9 @@ axum-extra = { version = "0.9.6", default-features = false, features = [
   "typed-header",
 ] }
 axum-test = { version = "16.4.1", default-features = false }
-axum-tracing-opentelemetry = { version = "0.24.1", default-features = false }
+axum-tracing-opentelemetry = { version = "0.24.1", default-features = false, features = [
+  "tracing_level_info",
+] }
```
woshilapin: Already done in #10113.

woshilapin (commented):
Thank you for the PR. Let's see if we can improve it.

editoast/Cargo.toml
hamz2a: Should I wait for this PR to be merged?

editoast/src/main.rs
hamz2a: Yes, you're right. The issue comes from calling `.init()` multiple times in the same process, which panics when used in parallel tests.
Option 2 works great and solved the issue for me. Thanks a lot for the suggestion! It’s much cleaner and avoids the panic. Appreciate the help!

editoast/Cargo.toml
hamz2a: Done.

editoast/src/main.rs
```diff
@@ -116,11 +116,12 @@ fn init_tracing(mode: EditoastMode, telemetry_config: &client::TelemetryConfig)
...
         }
     };
-    tracing_subscriber::registry()
+
+    let _ = tracing_subscriber::registry()
         .with(telemetry_layer)
         .with(env_filter_layer)
         .with(fmt_layer)
-        .init();
+        .try_init();
```
woshilapin: We might want to ignore an error for tests, but I believe we want to crash the application at startup in case of an error in production. It would not be good to start our application in production with an incorrect subscriber. Maybe `init_tracing` can return an error?

editoast/src/main.rs
woshilapin: Reading a bit more the documentation, I realize that we might set up a global subscriber, then also have per-scope subscriber. In this case, [the global subscriber will only serve as a fall-back when a scoped one is not found](https://docs.rs/tracing/latest/tracing/dispatcher/fn.set_global_default.html).

I'm not entirely sure on the effect of having only one global subscriber for all the tests. Maybe we should have one scoped subscriber for each test? Therefore, maybe the subscriber itself should be returned by the function? the `run()` would just `.init()` (or `tracing::dispatcher::set_global_default(subscriber.into())`, it's equivalent). `TestApp` would use `set_default()`. What do you think? Maybe @leovalais can give some insights about why it was scoped in the first place with `DefaultGuard` (it was introduced in 35b9369ebc8d7fba1637c2c0f50f7d7f26db8536 of https://github.com/OpenRailAssociation/osrd/pull/7511)?

editoast/src/main.rs
hamz2a: Done.

editoast/Cargo.toml
```diff
@@ -146,6 +146,7 @@ opentelemetry-otlp = { version = "0.27.0", default-features = false, features =
   "trace",
 ] }
 opentelemetry-semantic-conventions.workspace = true
+opentelemetry-stdout = { version = "0.27.0" }
```
leovalais: This is a `dev-dependency`

editoast/src/views/test_app.rs
```diff
@@ -14,12 +14,14 @@ use tower_http::trace::TraceLayer;
...
 use crate::{
+    client::{TelemetryConfig, TelemetryKind},
     core::{mocking::MockingClient, CoreClient},
     generated_data::speed_limit_tags_config::SpeedLimitTagIds,
     infra_cache::InfraCache,
     map::MapLayers,
+    try_init_tracing,
     valkey_utils::ValkeyConfig,
-    AppState, ValkeyClient,
+    AppState, EditoastMode, ValkeyClient,
```
leovalais: Wrong dependency order. We want `client` and `bin editoast` to depend on `views`

editoast/src/views/test_app.rs
```diff
@@ -106,16 +108,12 @@ impl TestAppBuilder {
...
-            )
-            .with_span_events(tracing_subscriber::fmt::format::FmtSpan::CLOSE)
-            .finish();
-        let tracing_guard = tracing::subscriber::set_default(sub);
+        let exporter = opentelemetry_stdout::SpanExporter::default();
+        let telemety_config = TelemetryConfig {
+            telemetry_kind: TelemetryKind::Opentelemetry,
+            ..Default::default()
+        };
+        let _ = try_init_tracing(EditoastMode::Webservice, &telemety_config, exporter);
```
leovalais: Errors shouldn't be silently ignored even in tests. Maybe `expect` or we should at least log something.

editoast/src/views/test_app.rs
```diff
@@ -188,8 +185,6 @@ pub(crate) struct TestApp {
     server: TestServer,
     db_pool: Arc<DbConnectionPoolV2>,
     core_client: Arc<CoreClient>,
-    #[allow(unused)] // included here to extend its lifetime, not meant to be used in any way
-    tracing_guard: tracing::subscriber::DefaultGuard,
```
leovalais: The point of using the guard was to avoid setting up a global subscriber that would be shared between tests, breaking isolation. Is the new setup test-local?

leovalais (changes requested):
Thanks for this PR. A few comments about architecture and test isolation. Besides, with the context of what you're working on I suppose this is about collecting logs for unit testing, but it's not clear to me how that would work. Why does printing logs on stdout help? Do you plan to capture the stream? A bit more context in the PR description or in the related issue would help.

editoast/src/main.rs
leovalais: It was to avoid "leaking" tracing configuration between tests. If I'm not mistaken, the global subscriber is thread-local. Maybe there's another way to achieve that, but the `set_default` and keeping the guard in the `TestApp` was an easy way to keep the configuration test-local.

hamz2a:
> Thanks for this PR. A few comments about architecture and test isolation. Besides, with the context of what you're working on I suppose this is about collecting logs for unit testing, but it's not clear to me how that would work. Why does printing logs on stdout help? Do you plan to capture the stream? A bit more context in the PR description or in the related issue would help.

Apologies for that. I've added more details in the description, and I hope it helps clarify things.

leovalais:
> I added the OpenTelemetry layer in TestApp to capture the trace_id set in the header, so we can test the endpoint we'll add in the https://github.com/OpenRailAssociation/osrd/issues/9724.

Thanks for adding some context. So if I get it right, the point is not to necessarily print on stdout, but rather to setup OpenTelemetry so that the request header can be read. The fact that it prints on stdout is irrelevant, or am I missing smth?

I'm surely lacking OpenTelemetry and tracing knowledge, but I still don't understand what setting up a span exporter has to do with reading the HTTP request headers? Especially since the middleware `OtelAxumLayer::default()` is set in the `TestApp`.

hamz2a:
> Thanks for adding some context. So if I get it right, the point is not to necessarily print on stdout, but rather to setup OpenTelemetry so that the request header can be read. The fact that it prints on stdout is irrelevant, or am I missing smth?
> 
> I'm surely lacking OpenTelemetry and tracing knowledge, but I still don't understand what setting up a span exporter has to do with reading the HTTP request headers? Especially since the middleware `OtelAxumLayer::default()` is set in the `TestApp`.


You're right, printing to `stdout` is irrelevant. 
The goal is to have the same `init_tracing` setup for both editoast webservice and `TestApp`. That's why I kept the same configuration. @woshilapin, do you confirm?

editoast/src/views/test_app.rs

editoast/Cargo.toml
hamz2a: Done.

editoast/src/main.rs
woshilapin: Then I would change the way I proposed above, the function `init_tracing` should not be in charge of installing (globally or not) the subscriber, it should only return it. Then `.run()` will be able to set it up with `.init()`. And `TestApp` will be able to use `set_default()` and keep the `DefaultGuard`. I feel it would be a better design. What do you think?

woshilapin:
> I'm surely lacking OpenTelemetry and tracing knowledge, but I still don't understand what setting up a span exporter has to do with reading the HTTP request headers? Especially since the middleware `OtelAxumLayer::default()` is set in the `TestApp`.

@leovalais From what I understand, `OtelAxumLayer` emits span events, the `SpanExporter` is used in the subscriber to these events and will consume these span events to export them somewhere (to Datadog, to stdout, etc.). But you're right, we shouldn't need `opentelemetry_stdout` since we already have the `tracing::layer::fmt()` configured, which prints to the terminal.

But as said by @hamz2a, one of the main idea of this PR is to have a similar setup of the subscriber for tests and for production code. We had a hard time debugging what was not working and part of it was that it was working well with a `cargo run -- runserver` but not with `cargo test`, and some of the problems came from this difference between production code and tests.

editoast/src/main.rs
hamz2a: Changes made as per your suggestion. Please check the code. Thanks again! @woshilapin @leovalais

editoast/src/views/test_app.rs
hamz2a: Got it, but I need these structs for `init_tracing` in `test_app`:

```rust
let telemetry_config = TelemetryConfig {
    telemetry_kind: TelemetryKind::Opentelemetry,
    ..Default::default()
};
let sub = init_tracing(EditoastMode::Webservice, &telemetry_config, exporter);
let tracing_guard = tracing::subscriber::set_default(sub);
```
Any suggestions on how to better isolate this?

editoast/src/views/test_app.rs
```diff
@@ -106,15 +108,12 @@ impl TestAppBuilder {
...
-            .pretty()
-            .with_env_filter(
-                tracing_subscriber::EnvFilter::builder()
-                    .with_default_directive(tracing_subscriber::filter::LevelFilter::DEBUG.into())
-                    .from_env_lossy(),
-            )
-            .with_span_events(tracing_subscriber::fmt::format::FmtSpan::CLOSE)
-            .finish();
+        let exporter = opentelemetry_stdout::SpanExporter::default();
+        let telemety_config = TelemetryConfig {
```
woshilapin:
```suggestion
        let telemetry_config = TelemetryConfig {
```

editoast/src/views/test_app.rs
hamz2a: I've just reintegrated the code in the latest commit.

editoast/src/views/test_app.rs
```diff
@@ -106,15 +108,12 @@ impl TestAppBuilder {
...
-        let sub = tracing_subscriber::fmt()
-            .pretty()
-            .with_env_filter(
-                tracing_subscriber::EnvFilter::builder()
-                    .with_default_directive(tracing_subscriber::filter::LevelFilter::DEBUG.into())
-                    .from_env_lossy(),
-            )
-            .with_span_events(tracing_subscriber::fmt::format::FmtSpan::CLOSE)
-            .finish();
+        let exporter = opentelemetry_stdout::SpanExporter::default();
```
woshilapin: As mentioned in some previous comments, I don't think we need `opentelemetry_stdout` since we already have `tracing_subscriber::layer::fmt()` which will take care of printing to the terminal. I would instead suggest using the [`NoopSpanExporter`](https://docs.rs/opentelemetry_sdk/latest/opentelemetry_sdk/testing/trace/struct.NoopSpanExporter.html).

```suggestion
        let exporter = opentelemetry_sdk::testing::trace::NoopSpanExporter::default();
```

Note that you will need to add the `testing` feature. Please do not add it on the `opentelemetry_sdk` crate in the `[dependencies]`. Add a new line for `opentelemetry_sdk` in the `[dev-dependencies]` (we don't want the `testing` feature to leak in production code).

editoast/Cargo.toml
```diff
@@ -199,6 +199,7 @@ axum = { version = "0.7.9", default-features = false, features = [
 editoast_authz = { workspace = true, features = ["fixtures"] }
 editoast_models = { workspace = true, features = ["testing"] }
 editoast_osrdyne_client = { workspace = true, features = ["mock_client"] }
+opentelemetry_sdk = { version = "0.27.1", features = ["testing"] }
```
woshilapin: Since it's defined twice now, maybe we want to define it in the `workspace` dependencies, in order to define the version in only one place?

woshilapin (approved):
I believe it's in a much better state now 🚀. Thank you 🎉

editoast/Cargo.toml
hamz2a: Done.

editoast/Cargo.toml
```diff
@@ -199,6 +200,7 @@ axum = { version = "0.7.9", default-features = false, features = [
 editoast_authz = { workspace = true, features = ["fixtures"] }
 editoast_models = { workspace = true, features = ["testing"] }
 editoast_osrdyne_client = { workspace = true, features = ["mock_client"] }
+opentelemetry_sdk = { workspace = true, features = ["testing"] }
```
leovalais: Feature `testing` requires `async-std`, which we don't want since we use `tokio`. It's already in our dev-dependencies because `rstest` requires it, but we'll be able to lift that restriction when we'll have our `DbPoolV3`. In the meantime, could we remove this feature and implement a `NoopExporter` manually?

editoast/src/main.rs
```diff
@@ -66,12 +67,16 @@ pub use valkey_utils::ValkeyConnection;
...
 ///   output (and not extra information), it should be on stdout
 #[derive(Debug, PartialEq)]
-enum EditoastMode {
+pub enum EditoastMode {
     Webservice,
     Cli,
 }
 
-fn init_tracing(mode: EditoastMode, telemetry_config: &client::TelemetryConfig) {
+pub fn init_tracing<T: SpanExporter + 'static>(
```
leovalais:
```suggestion
pub fn crate_tracing_subscriber<T: SpanExporter + 'static>(
```

leovalais (changes requested):

editoast/src/views/test_app.rs
leovalais: You could move `init_tracing` in `editoast_common`, but `EditoastMode` and the `telemetry_config` do not make sense there. You'll need some kind of mapping.

editoast/src/main.rs
hamz2a: Done.

editoast/src/views/test_app.rs
hamz2a: I’ve updated the function signature for `create_tracing_subscriber` to the following:
```rust
pub fn create_tracing_subscriber<T: SpanExporter + 'static>(
    mode: Option<EditoastMode>,
    telemetry_config: Option<TelemetryConfig>,
    exporter: T,
) -> impl tracing::Subscriber
```
I believe it adds more flexibility. what do you think?

editoast/Cargo.toml
hamz2a: Done.

editoast/src/views/test_app.rs
leovalais: I'm not entirely convinced 😕

1. This does not solves the dependency problem: the function is still imported in `views` from the `crate` level.
2. We have two calls to that function and we know each time which mode and config we want. So there's no need for flexibility when 100% of our usage do not require that.
3. I'd argue that hiding a parameter behind a `None` degrades readability: now we have to look inside the function to find the actual values of the parameters, both of them not having an obvious default value.

Wdyt?

hamz2a: I’ve updated the code based on your feedback. Please verify if the changes align with your suggestions. @leovalais

editoast/src/views/test_app.rs
```diff
@@ -9,12 +9,17 @@ use axum_tracing_opentelemetry::middleware::OtelAxumLayer;
...
+use opentelemetry_sdk::export::trace::ExportResult;
+use opentelemetry_sdk::export::trace::SpanData;
+use opentelemetry_sdk::export::trace::SpanExporter;
 use serde::de::DeserializeOwned;
 use tower_http::trace::TraceLayer;
 use url::Url;
 
 use crate::{
     core::{mocking::MockingClient, CoreClient},
+    create_tracing_subscriber,
```
leovalais: This is the import that breaks the dependency order. `views` cannot `use` something from the `crate` or `client` level. I'd suggest moving `create_tracing_subscriber` to `editoast_common`.

leovalais (changes requested):
Almost there, thanks for the changes! There's still a dependency order issue, but apart from that, lgtm.

leovalais:
> si je bouge create_tracing_subscriber je dois bouger aussi
> ```
>    mode: EditoastMode,
>    telemetry_config: TelemetryConfig,
> ```

Yes, but `TelemetryConfig` is a `clap` definition and `EditoastMode` doesn't make sense outside of the CLI. I'd suggest making a new struct for configuration:


```rust
struct TracingConfig {
   stream: enum { Stderr, Stdout },
   telemetry: Option<struct {
		service_name: String,
		telemetry_endpoint: Url
	}>
}
```

Wdyt?

editoast/src/main.rs
```diff
@@ -66,61 +58,18 @@ pub use valkey_utils::ValkeyConnection;
 /// - we *expect* a webserver to output logging information, so since it's an expected
 ///   output (and not extra information), it should be on stdout
 #[derive(Debug, PartialEq)]
-enum EditoastMode {
+pub enum EditoastMode {
```
leovalais:
```suggestion
enum EditoastMode {
```

editoast/src/views/mod.rs
```diff
@@ -22,7 +22,7 @@ pub mod train_schedule;
 pub mod work_schedules;
 
 #[cfg(test)]
-mod test_app;
+pub mod test_app;
```
```suggestion
mod test_app;
```

editoast/src/views/test_app.rs
```diff
@@ -106,15 +121,7 @@ impl TestAppBuilder {
...
-        let sub = tracing_subscriber::fmt()
-            .pretty()
-            .with_env_filter(
-                tracing_subscriber::EnvFilter::builder()
-                    .with_default_directive(tracing_subscriber::filter::LevelFilter::DEBUG.into())
-                    .from_env_lossy(),
-            )
-            .with_span_events(tracing_subscriber::fmt::format::FmtSpan::CLOSE)
-            .finish();
+        let sub = create_tracing_subscriber(TracingConfig::default(), NoopSpanExporter);
```
```suggestion
        let sub = create_tracing_subscriber(TracingConfig {
        	...
        }, NoopSpanExporter);
```

Could the default value be built here? This way we gather all default test constants here instead of having them scattered across different crates. The `impl Default` and `derive(Derivative)` from `common::tracing` can be removed.

leovalais (approved):
LGTM, thanks for all the modifications!

editoast/src/main.rs
hamz2a: Done.

editoast/src/views/mod.rs
hamz2a: Done.

editoast/src/views/test_app.rs

editoast/src/views/test_app.rs

PR #10129, opened 2024-12-19
editoast: add tracing for ping operations

woshilapin:
It seems that our healtcheck needs a relatively long time (#10026) but we're missing information about what could take time. Let's first add a way to measure it and see what we can do with it later.


flomonster (approved):
LGTM

PR #10134, opened 2024-12-19
front: fix power restrictions reset when editing a train

### LINKED ISSUE #9902: front: editing a train with power restrictions resets its power restrictions

clarani:
closes #9902 

To test:
- create a train with power restrictions
- try to edit it
- its power restrictions are not reset

### DIFF
--- a/front/src/modules/pathfinding/hooks/usePathfinding.ts
+++ b/front/src/modules/pathfinding/hooks/usePathfinding.ts
@@ -51,6 +51,10 @@ const usePathfinding = (
   const [pathfindingState, setPathfindingState] =
     useState<PathfindingState>(initialPathfindingState);
 
+  // isInitialized is used to prevent the pathfinding to be launched multiple times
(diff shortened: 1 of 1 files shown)


Akctarus (approved):
Tested, LGTM, ty for this fix.
Also I have a question, not directly linked to this PR, do you know if this is the expected behavior when places where power restrictions are cut create vias?

clarani:
Thanks @Akctarus ! Yes, it is the expected behavior for now, because to add power restriction in the backend, we need to locate them with pathsteps ids ;)

kmer2016 (approved):
LGTM

PR #10137, opened 2024-12-19
editoast: improve tests for projection

woshilapin:
We're planning a rework of the projection algorithm to support a case that is not supported today: project multiple times the same track range (possibly different offset). This use case is brought to us by work schedule which might have this pattern.

In order to have a good refactoring without breaking anything, let's have even more tests.

> [!NOTE]
> Review by commit
> 1. first commit just make use of `rstest` cases to express tests in a more concise way
> 2. second commit add more tests
> 3. third commit, automate switching direction of inputs (multiply by 4 the numbers of tests)

> [!WARNING]
> The third commit significantly make the test code harder to read so I'd be happy to discuss it...

### DIFF
--- a/editoast/src/views/path/projection.rs
+++ b/editoast/src/views/path/projection.rs
@@ -288,6 +288,7 @@ pub enum TrackLocationFromPath {
 mod tests {
     use super::*;
     use rstest::rstest;
+    use std::iter::DoubleEndedIterator;
 
     #[test]
     #[should_panic]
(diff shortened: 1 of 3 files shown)


editoast/src/views/path/projection.rs
```diff
@@ -407,111 +407,44 @@ mod tests {
...
-        let boundaries = projection.get_intersections(&track_ranges);
-
-        let expected: Vec<Intersection> = vec![];
-        assert_eq!(boundaries, expected);
-    }
-
-    #[test]
-    fn get_boundaries_case_5() {
-        let path = vec![TrackRange::new("A", 0, 100, Direction::StartToStop)];
+    #[rstest]
```
hamz2a, line 410: Could you please add a comment in the test code to explain how it works, especially the business logic part? It would help a lot in understanding it. Thanks!

editoast/src/views/path/projection.rs
```diff
@@ -437,20 +437,68 @@ mod tests {
         #[case] path: &[&str],
         #[case] track_ranges: &[&str],
         #[case] expected_intersections: &[(u64, u64)],
+        #[values(false, true)] toggle_path: bool,
+        #[values(false, true)] toggle_track_ranges: bool,
     ) {
-        let path: Vec<TrackRange> = path.into_iter().map(|s| s.parse().unwrap()).collect();
```
hamz2a: Just a reminder from our call: some of the code here could be moved into a separate function for reuse.

hamz2a (commented):
Thank you for the PR. I’ve left two comments.

editoast/src/views/path/projection.rs
```diff
@@ -437,20 +437,68 @@ mod tests {
...
+                    track_range.direction = track_range.direction.toggle();
+                }
+                track_range
+            })
+            .collect();
+        let path = if toggle_path {
+            path.into_iter().rev().collect()
+        } else {
+            path
+        };
```
hamz2a: Just a reminder from our call: some of the code here could be moved into a separate function for reuse.

editoast/src/views/path/projection.rs
```diff
@@ -407,111 +407,98 @@ mod tests {
...
+    #[case::complex_path_two_intersections(
+        &["A+50-100", "B+200-0", "C+0-300", "D+250-0", "E+100-25"],
+        &["X+0-100", "B+0-200", "C+200-150", "E+30-100", "Z+0-100"],
+        &[(100, 350), (350, 420)]
+    )]
+    #[case::complex_path_three_intersections(
+        &["A+50-100", "B+200-0", "C+0-300", "D+250-0", "E+100-25"],
+        &["A+0-100", "B+0-200", "X+0-100", "C+200-150", "Z+0-100", "E+30-100"],
+        &[(50, 300), (400, 450), (550, 620)]
+    )]
```
hamz2a, line 478: Could you please add a comment in the test code to explain how it works, especially the business logic part? It would help a lot in understanding it.

editoast/src/views/path/projection.rs
woshilapin: I've done [5969536](https://github.com/OpenRailAssociation/osrd/pull/10137/commits/59695360472393d452051d0d5a01c09faf8f0483) and [ba541fa](https://github.com/OpenRailAssociation/osrd/pull/10137/commits/ba541faed7d0ee2d3968e4c27878fc6621a09d6a) to extract some functions and make the test function itself, shorter and more explicit about what it does. What do you think?

editoast/src/views/path/projection.rs
woshilapin: I've made a few comments in multiple places in [3750ea8](https://github.com/OpenRailAssociation/osrd/pull/10137/commits/3750ea87f7a87178d9504f2b3e0cd30b0a112cec), what do you think?

hamz2a: Looks good, thanks for the updates!

editoast/src/views/path/projection.rs
hamz2a: Looks good, thanks!

leovalais (approved):
LGTM. Much better!

PR #10139, opened 2024-12-19
front: improve import view

### LINKED ISSUE #8115: Weird behavior and error messages importing train schedule files
### What happened?

Here is a list of weird behaviors when importing train schedules from files:

1. We cannot import a file if its extension is not `.json` even if it's a valid json file
2. If importing a valid json file like `{}` nothing happens (no errors are reported)
3. If we import an almost-valid file like this one [almost_valid.json](https://github.com/user-attachments/files/16283842/almost_valid.json). We have the following error message "All running time calculations have failed.". This is confusing since no running time calculation is performed. The file is just invalid.

1. Being able to import json file with an invalid extension
2. Display an error like: "Malformed json file"
3. Change the error like: "Invalid train schedule file format"

1. Create a scenario
2. Try to import files like describe above

Local

Firefox

580140d

clarani:
close #8115 

You can ask me some files to test the PR

### DIFF
--- a/front/public/locales/fr/operationalStudies/importTrainSchedule.json
+++ b/front/public/locales/fr/operationalStudies/importTrainSchedule.json
@@ -26,12 +26,12 @@
   "selectStation": "Sélectionner cette station",
   "startTime": "DÉBUT",
   "status": {
-    "calculatingTrainScheduleCompleteAllSuccess_one": "Le train a été importé.",
-    "calculatingTrainScheduleCompleteAllSuccess": "Tous les trains ont été importés.",
     "invalidTrainSchedules_one": "La circulation à importer n'est pas au bon format.",
-    "invalidTrainSchedules": "Les circulations à importer ne sont pas au bon format."
+    "invalidTrainSchedules": "Les circulations à importer ne sont pas au bon format.",
+    "successfulImport_one": "Le train a été importé.",
+    "successfulImport": "Tous les trains ont été importés."
   },
   "success": "Opération réussie",
   "to": "Destination",
-  "trainsFound": "trains trouvé(s)"
+  "trainsFound": "train(s) trouvé(s)"
 }
--- a/front/public/locales/fr/operationalStudies/importTrainSchedule.json
+++ b/front/public/locales/fr/operationalStudies/importTrainSchedule.json
@@ -9,13 +9,11 @@
     "error": "Une erreur est survenue",
     "errorEmptyFile": "Fichier vide",
     "errorImport": "Impossible de convertir les données en TrainSchedule",
-    "errorInvalidJSONFormat": "Format JSON invalide",
-    "errorInvalidXMLFormat": "Format XML invalide",
     "errorNoDate": "Vous devez renseigner une date.",
     "errorNoFrom": "Vous devez renseigner une origine.",
     "errorNoTo": "Vous devez renseigner une destination.",
     "errorSameFromTo": "L'origine et la destination doivent être différentes.",
-    "errorUnsupportedFileType": "Type de fichier non supporté"
+    "errorInvalidFile": "Fichier invalide"
   },
   "failure": "Opération échouée",
   "from": "Origine",
--- a/front/public/locales/fr/operationalStudies/importTrainSchedule.json
+++ b/front/public/locales/fr/operationalStudies/importTrainSchedule.json
@@ -28,7 +28,8 @@
   "status": {
     "calculatingTrainScheduleCompleteAllSuccess_one": "Le train a été importé.",
     "calculatingTrainScheduleCompleteAllSuccess": "Tous les trains ont été importés.",
-    "calculatingTrainScheduleCompleteAllFailure": "Tous les calculs de marche ont échoué."
+    "invalidTrainSchedules_one": "La circulation à importer n'est pas au bon format.",
+    "invalidTrainSchedules": "Les circulations à importer ne sont pas au bon format."
   },
   "success": "Opération réussie",
   "to": "Destination",
--- a/front/src/common/uploadFileModal.tsx
+++ b/front/src/common/uploadFileModal.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useContext, useState } from 'react';
+import { useContext, useState } from 'react';
 
 import { Download } from '@osrd-project/ui-icons';
 import { isNil } from 'lodash';
@@ -16,50 +16,6 @@ const UploadFileModal = ({ handleSubmit }: UploadFileModalProps) => {
   const { t } = useTranslation(['operationalStudies/importTrainSchedule']);
   const { closeModal } = useContext(ModalContext);
   const [selectedFile, setSelectedFile] = useState<File | undefined>(undefined);
-  const [isValid, setIsValid] = useState<string | undefined>(undefined);
-
-  const parseXML = (xmlString: string) => {
-    try {
-      const parser = new DOMParser();
-      const xmlDoc = parser.parseFromString(xmlString, 'application/xml');
-      const parserError = xmlDoc.getElementsByTagName('parsererror');
-      if (parserError.length > 0) {
-        throw new Error('Invalid XML');
-      }
-      return undefined;
-    } catch (error) {
-      return t('errorMessages.errorInvalidXMLFormat').toString();
-    }
-  };
-  // TODO : create the translation keys
-  const validateFile = useCallback(
-    async (fileToValidate: File): Promise<string | undefined> => {
-      if (fileToValidate.size === 0) {
-        return t('errorMessages.errorEmptyFile').toString();
-      }
-      if (fileToValidate.type === 'application/json') {
-        try {
-          JSON.parse(await fileToValidate.text());
-        } catch (e) {
-          return t('errorMessages.errorInvalidJSONFormat').toString();
-        }
-      } else if (
-        fileToValidate.type === 'application/railml' ||
-        fileToValidate.name.endsWith('.railml')
-      ) {
-        const fileContent = await fileToValidate.text();
-        const xmlError = parseXML(fileContent);
-        if (xmlError) {
-          return xmlError;
-        }
-      } else {
-        return t('errorMessages.errorUnsupportedFileType').toString();
-      }
-
-      return undefined;
-    },
-    [t]
-  );
 
   return (
     <>
@@ -73,21 +29,15 @@ const UploadFileModal = ({ handleSubmit }: UploadFileModalProps) => {
           <input
             type="file"
             name="file"
-            accept=".json,.xml,.railml"
+            accept=".json,.txt,.xml,.railml"
             onChange={async (e) => {
               if (e.target.files && e.target.files.length > 0) {
-                const error = await validateFile(e.target.files[0]);
-                setIsValid(error);
-                if (isNil(error)) {
-                  setSelectedFile(e.target.files[0]);
-                }
+                setSelectedFile(e.target.files[0]);
               } else {
                 setSelectedFile(undefined);
-                setIsValid(undefined);
               }
             }}
           />
-          {!isNil(isValid) && <div className="text-danger">{isValid}</div>}
         </>
       </ModalBodySNCF>
       <ModalFooterSNCF>
@@ -105,7 +55,7 @@ const UploadFileModal = ({ handleSubmit }: UploadFileModalProps) => {
             <div className="col-6">
               <button
                 type="button"
-                disabled={isNil(selectedFile) || !isNil(isValid)}
+                disabled={isNil(selectedFile)}
                 className="btn btn-block btn-sm btn-primary"
                 onClick={() => {
                   if (selectedFile) handleSubmit(selectedFile);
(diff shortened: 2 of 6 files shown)


front/src/common/uploadFileModal.tsx
```diff
@@ -76,18 +32,12 @@ const UploadFileModal = ({ handleSubmit }: UploadFileModalProps) => {
             accept=".json,.xml,.railml"
```
SharglutDev: Shouldn't we add .txt ? This way I think we wouldn't have to select `all files` in the os window

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/handleParseFiles.ts
```diff
@@ -9,71 +9,40 @@ export const handleFileReadingError = (error: Error) => {
...
 ) => {
-  try {
-    const importedTrainSchedules: TrainScheduleBase[] = JSON.parse(fileContent);
-    if (importedTrainSchedules && importedTrainSchedules.length > 0) {
-      setTrainsJsonData(importedTrainSchedules);
-    }
-  } catch (error) {
-    handleJsonParsingError(error as Error, dispatch);
+  const importedTrainSchedules: TrainScheduleBase[] = JSON.parse(fileContent);
+  if (importedTrainSchedules && importedTrainSchedules.length > 0) {
```
SharglutDev: On the line above, we type `importedTrainSchedules: TrainScheduleBase[]` so we shouldn't have to check for that ?
```suggestion
  if (importedTrainSchedules.length > 0) {
```

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleTrainsList.tsx
```diff
@@ -104,7 +104,7 @@ const ImportTrainScheduleTrainsList = ({
       dispatch(
         setFailure({
           name: t('failure'),
-          message: t('status.calculatingTrainScheduleCompleteAllFailure', {
+          message: t('status.invalidTrainSchedules', {
             trainsList,
             count: trainsList.length,
```
SharglutDev: The trad doesn't seem to take count into account ?

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/handleParseFiles.ts
```diff
@@ -9,71 +9,40 @@ export const handleFileReadingError = (error: Error) => {
...
+    dispatch(
+      setFailure({
+        name: t('errorMessages.error'),
+        message: t('errorMessages.errorEmptyJSON'),
+      })
+    );
   }
 };
 
 export const processXmlFile = async (
```
SharglutDev, line 91: Don't think it's related to your changes because the import technically works but after editing the rs for one of the train in the list I have this : 

Capture d’écran 2024-12-20 à 11 42 40

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/handleParseFiles.ts
```diff
@@ -9,71 +9,40 @@ export const handleFileReadingError = (error: Error) => {
...
-    }
-
-    const importedTrainSchedules = await parseRailML(xmlDoc);
-    if (importedTrainSchedules && importedTrainSchedules.length > 0) {
-      updateTrainSchedules(importedTrainSchedules);
-    }
-  } catch (error) {
-    handleXmlParsingError(error as Error, dispatch);
+  if (parserError.length > 0) {
+    throw new Error('Invalid XML');
```
SharglutDev, line 101: If the file is invalid we throw the error in console but the user doesn't see any message/notification. Wouldn't we want to display something ?
In `uploadFileModal.tsx`, it used to have an `errorInvalidXMLFormat` message displayed in the catch (not sure how it was suppose to be displayed).

SharglutDev, line 101: If the json has a syntax error, we see "invalid XML" in console even though we sent a .json. Seems the wrong function is called.
Should we keep the file extensions condition like it was before ?

front/src/common/uploadFileModal.tsx
clarani: Yes sure !

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/handleParseFiles.ts
clarani: Yes I can add a check :)

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleTrainsList.tsx
clarani: Done :+1:

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/handleParseFiles.ts
clarani: The goal of this ticket is to handle all types of files (try to parse it into json and xml even though it's not a .json or a .xml), so I can't really keep the file extensions condition like before :/
I added an early return if the extension is .json, is it enough for you ?

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/handleParseFiles.ts
```diff
@@ -9,71 +9,94 @@ export const handleFileReadingError = (error: Error) => {
...
-  console.error('Error parsing JSON:', error);
-  dispatch(
-    setFailure({
-      name: t('errorMessages.error'),
-      message: t('errorMessages.errorInvalidJSONFormat'),
-    })
-  );
-};
+const validateTrainSchedules = (
+  importedTrainSchedules: Record<string, unknown>[]
```
SharglutDev: This should work to avoid many `any` in the function ?
```suggestion
  importedTrainSchedules: Partial<TrainScheduleBase>[]
```

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/handleParseFiles.ts
```diff
@@ -9,71 +9,94 @@ export const handleFileReadingError = (error: Error) => {
...
+  const isInvalidTrainSchedules = importedTrainSchedules.some((trainSchedule) => {
+    if (
+      ['constraint_distribution', 'path', 'rolling_stock_name', 'start_time', 'train_name'].some(
+        (key) => !(key in trainSchedule)
+      ) ||
+      !Array.isArray(trainSchedule.path)
+    ) {
+      return true;
+    }
+    const hasInvalidSteps = trainSchedule.path.some((step) => ['id'].some((key) => !(key in step)));
```
SharglutDev: Can't we do that instead ? (since we check only the `id` key)
```suggestion
    const hasInvalidSteps = trainSchedule.path.some((step) => !('id' in step));
```

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/handleParseFiles.ts
```diff
@@ -9,71 +9,94 @@ export const handleFileReadingError = (error: Error) => {
...
+): TrainScheduleBase[] | null => {
+  const isInvalidTrainSchedules = importedTrainSchedules.some((trainSchedule) => {
+    if (
+      ['constraint_distribution', 'path', 'rolling_stock_name', 'start_time', 'train_name'].some(
+        (key) => !(key in trainSchedule)
+      ) ||
+      !Array.isArray(trainSchedule.path)
+    ) {
+      return true;
+    }
```
SharglutDev: Might be safer (in case of misspell) to do that
```suggestion
    const compulsoryKeys: (keyof TrainScheduleBase)[] = [
      'constraint_distribution',
      'path',
      'rolling_stock_name',
      'start_time',
      'train_name',
    ];
    if (
      compulsoryKeys.some((key) => !(key in trainSchedule)) ||
      !Array.isArray(trainSchedule.path)
    ) {
      return true;
    }
```

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleTrainsList.tsx
```diff
@@ -94,7 +94,7 @@ const ImportTrainScheduleTrainsList = ({
       dispatch(
         setSuccess({
           title: t('success'),
-          text: t('status.calculatingTrainScheduleCompleteAllSuccess', {
+          text: t('status.successfulImport', {
             trainsList,
             count: trainsList.length,
```
SharglutDev: For json imports, I think `trainList` will be empty (same for l.109)
```suggestion
            count: trainsList.length || trainsJsonData.length,
```

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/handleParseFiles.ts
```diff
@@ -9,71 +9,94 @@ export const handleFileReadingError = (error: Error) => {
...
-  dispatch(
-    setFailure({
-      name: t('errorMessages.error'),
-      message: t('errorMessages.errorInvalidJSONFormat'),
-    })
-  );
-};
+const validateTrainSchedules = (
+  importedTrainSchedules: Record<string, unknown>[]
+): TrainScheduleBase[] | null => {
```
SharglutDev: Where do we return `null` in the function ?

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/handleParseFiles.ts
```diff
@@ -9,71 +9,94 @@ export const handleFileReadingError = (error: Error) => {
...
+      );
+    }
+    return isJsonFile;
+  }
+
+  // validate the trainSchedules
   try {
-    const importedTrainSchedules: TrainScheduleBase[] = JSON.parse(fileContent);
+    const importedTrainSchedules = validateTrainSchedules(rawContent);
     if (importedTrainSchedules && importedTrainSchedules.length > 0) {
```
SharglutDev: If I understood correctly, if we miss a compulsory key in `rawContent`, we will throw an error (in `validateTrainSchedules`) so we will fallback in `catch` here with `errorInvalidFile`. 

So same thing as before, checking `if (importedTrainSchedules &&...` isn't needed since `validateTrainSchedules` will always return an array (empty or not) if we are still in the `try`.

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/handleParseFiles.ts
SharglutDev: Still don't work if the json is in a .txt but it's kinda edgy so works for me :)

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/handleParseFiles.ts
clarani: Done :)

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/handleParseFiles.ts
clarani: Good catch, thanks !

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/handleParseFiles.ts
clarani: Done :+1:

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/handleParseFiles.ts
clarani: Done

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/handleParseFiles.ts
clarani: Done

front/public/locales/fr/operationalStudies/importTrainSchedule.json
```diff
@@ -28,9 +26,10 @@
...
-    "calculatingTrainScheduleCompleteAllSuccess_one": "Le train a été importé.",
-    "calculatingTrainScheduleCompleteAllSuccess": "Tous les trains ont été importés.",
-    "calculatingTrainScheduleCompleteAllFailure": "Tous les calculs de marche ont échoué."
+    "successfulImport_one": "Le train a été importé.",
+    "successfulImport": "Tous les trains ont été importés.",
+    "invalidTrainSchedules_one": "La circulation à importer n'est pas au bon format.",
+    "invalidTrainSchedules": "Les circulations à importer ne sont pas au bon format."
   },
   "success": "Opération réussie",
   "to": "Destination",
```
SharglutDev, line 35: The line below, if you could just do : `"trainsFound": "train(s) trouvé(s)"`

SharglutDev (approved):
Lgtm and tested, good job ! You can fix and resolve the last comments when done :)

front/src/modules/trainschedule/components/ManageTrainSchedule/helpers/handleParseFiles.ts
clarani: Here is the bug : https://github.com/OpenRailAssociation/osrd/issues/10219

front/public/locales/fr/operationalStudies/importTrainSchedule.json
clarani: It's done :white_check_mark:

PR #10140, opened 2024-12-19
front: display signals on map when searching signals

### LINKED ISSUE #6065: When searching for a signal the signal layer is not automatically activated
### What happened?

When you search a signal, the map will be focused on it but if the layer is not already activated, we can not see anything on the map  

When a signal is searched and found, the signals layer is automatically activated  

1. Go to the map 
2. Desactivate the signal layer
3. Search for a signal 

staging 

fb39845

clarani:
closes #6065


Caracol3 (approved):
lgtm (tested)

Math-R (approved):
lgtm !

PR #10141, opened 2024-12-19
front: change time tolerance from 1s to 2s

### LINKED ISSUE #8951: Schedule points marked as not honored even though they're within tolerance
### What happened?

image

The steps are marked as not honored even though they're within the tolerance our simulator can reach. 
Indeed the simulator only guarantees that the sum of delays is lower than 2s. Here, we fall within this error margin

Scheduled times: [0, 433000, 1033000, N/A]
Actual times:       [0, 432542, 1031964, 1581827]

Also this makes the train appear like this in the train list:
image

I expect schedule points which are less than 2s off to not be displayed as not honored (2s for each schedule is too lenient, but that's easier to implement).
I also expect the train to appear normal in the train list.

1. Create a scenario on small infra
2. Upload [this train_schedule](https://github.com/user-attachments/files/17075525/train_schedules.7.json)
3. Witness the scheduled points injustly marked wrong

Dev (SNCF)

Firefox

b448226

axrolld:
fixes #8951


clarani (dismissed):
LGTM :white_check_mark:

clarani (approved):

Yohh (approved):

PR #10147, opened 2024-12-20
Add ERTMS-ETCS brake params and signaling to rolling-stock model

### LINKED ISSUE #9706: core: add mock infra and rolling-stock to be used for ETCS testing/demo
Test part of the model: add core tests using new infra and RS, but probably only using "classic" braking curve (test to be updated when ETCS is handled)

Also:
* ~editoast, front: check that infra editor allows ERTMS signal creation/edition~

### LINKED ISSUE #9708: editoast, front, core: add ETCS parameters to rolling stock model
What's new for front users :
* possibility to declare a train ETCS_LEVEL2 (but not usable as it requires a ETCS brake params that are not available in front)
* possibility to declare a signal as ETCS_LEVEL2
* it can be "seen" if loading the new etcs_infra and etcs_rolling_stock data available in the repo (but the simulation breaks currently as it's not implemented)
* breaks nothing else as no ERTMS/ETCS data is available

To be added to RS : https://github.com/OpenRailAssociation/osrd/pull/1737/files#diff-ae258262711f96ff8379d40f67afd4a6bdd1b6dd8a8a5ba54500711834b070ed (et values in https://github.com/OpenRailAssociation/osrd/pull/1737/files#diff-d1333ce74e6f24539524cd9fa092979d3d0d008ca7fe30eba9c53a0c469c9212)

See also (just close on the way?) https://github.com/osrd-project/osrd-confidential/issues/607

bougue-pe:
Note: not managing towed rolling-tock so far (may not have a max_speed, not sure how to merge brake values).

Also:
* Add ETCS_LEVEL2 signaling for infra and rolling-stock (complete duplicate of TVM430 for now).
* Compute ETCS_LEVEL2 ranges of the path
* Fix some small bits (🔍 review by commit)
* Tests:
  * add tests ETCS_LEVEL2 rolling-stock (derived from fast_rolling_stock).
  * add tests ETCS infra (including electrical_profile). Derived from small_infra, just switching all the signaling from BAL to
ETCS_LEVEL2.

✅ Hand-tested only for now:
* keeps ETCS params when editing RS
* when removing the `TODO()` on core ETCS-brake part processing, it "works" as expected, with the `TODO()` it fails in core

fix #9708 and #9706

## TODO after this PR:
* add integration tests (circumvent ETCS intervals for now with a TODO) > For this PI : https://github.com/OpenRailAssociation/osrd/pull/10294
* prioritize signaling systems for blocks of pathfinding: ETCS>TVM430>TVM300>BAL>BAPR
* use the signaling system output from pathfinding everywhere (remove recoverBlockPath - only deprecated now)
* merge small_infra BAL and ERTMS (and TVM?)


editoast/editoast_schemas/src/rolling_stock.rs
```diff
@@ -87,3 +93,78 @@ pub struct RollingStock {
     #[serde(default)]
     pub metadata: Option<RollingStockMetadata>,
 }
+
+impl<'de> Deserialize<'de> for RollingStock {
```
woshilapin, line 99: To make the implementation simpler, you might be able to use [this trick](https://stackoverflow.com/a/70293354) which allow to not have to maintain `Inner` with `RollingStock`.

```rust
#[derive(Deserialize)]
#[serde(remote = "self")]
struct RollingStock {
  /* ... */
}

impl<'de> Deserialize<'de> for RollingStock {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>,
    {
        let rolling_stock = RollingStock::deserialize(deserializer)?;

        if rolling_stock
            .supported_signaling_systems
            .0
            .contains(&"ETCS_LEVEL2".to_string())
            && rolling_stock.etcs_brake_params.is_none()
        {
            return Err(serde::de::Error::custom(
                "invalid rolling-stock, supporting ETCS_LEVEL2 signaling system requires providing ETCS brake parameters.",
            ));
        }
        Ok(rolling_stock)
    }
}
```

Basically, `#[serde(remote = "self")]` doesn't really implement the `Deserialize` trait, but create a function of the object `RollingStock` ([this `remote` is actually here because of problems around the orphan rule](https://serde.rs/remote-derive.html)).

editoast/editoast_schemas/src/rolling_stock.rs
```diff
@@ -87,3 +93,78 @@ pub struct RollingStock {
...
+            pub metadata: Option<RollingStockMetadata>,
+        }
+
+        let inner = Inner::deserialize(deserializer)?;
+
+        if inner
+            .supported_signaling_systems
+            .0
+            .contains(&"ETCS_LEVEL2".to_string())
+            && inner.etcs_brake_params.is_none()
```
woshilapin: This validation makes sense, but I realize we already have some validation of coherence of the input data, in [a different place](https://github.com/OpenRailAssociation/osrd/blob/7582bddb071ae51efa1c5b13039ea3c59dd1307b/editoast/src/models/rolling_stock_model.rs#L93-L140). At the very least, I would expect all of these validations to happen at the same place.

I can understand why we did it somewhere else: in a 3-tier architecture, deserialization should only check if it's a valid considering the protocol (JSON in this case), not if the data inside is coherent relatively to the business logic. Then the model which will be inserted into DB has to be checked logically too, so conversion from `RollingStock` to `RollingStockModel` will do this verification.

That said, I'm also a fervent defender [Parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/), basically, validate early and don't think about it anymore (and hopefully, don't forget to call the `.validate_imported_rolling_stock()` function. cc @leovalais 

In any case, I believe we should move everything to one place or the other in the end.

editoast/editoast_schemas/src/rolling_stock/etcs_brake_params.rs
```diff
@@ -0,0 +1,115 @@
...
+    pub k_n_neg: SpeedIntervalValueCurve,
+    /// T_traction_cut_off: time delay in s from the traction cut-off command to the moment the acceleration due to traction is zero
+    #[derivative(Hash(hash_with = "editoast_common::hash_float::<5,_>"))]
+    pub t_traction_cut_off: f64,
+    /// T_bs1: time service break in s used for SBI1 computation
+    #[derivative(Hash(hash_with = "editoast_common::hash_float::<5,_>"))]
+    pub t_bs1: f64,
+    /// T_bs2: time service break in s used for SBI2 computation
+    #[derivative(Hash(hash_with = "editoast_common::hash_float::<5,_>"))]
+    pub t_bs2: f64,
```
woshilapin, line 48: Is there no unit for these? (thanks for documenting relatively extensively the other fields with units 🙏)

editoast/editoast_schemas/src/rolling_stock/etcs_brake_params.rs
```diff
@@ -0,0 +1,115 @@
...
+    pub t_be: f64,
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize, ToSchema, Derivative)]
+#[derivative(Hash)]
+#[serde(deny_unknown_fields)]
+pub struct SpeedIntervalValueCurve {
+    #[derivative(Hash(hash_with = "editoast_common::hash_float_slice::<3,_>"))]
+    #[schema(example = json!([8.333333, 19.444444]))]
+    /// Speed in m/s (sorted)
```
woshilapin: nitpick: sorted in which order?

```suggestion
    /// Speed in m/s (sorted from smallest to biggest)
```
> I prefer `ascending`

editoast/editoast_schemas/src/rolling_stock/etcs_brake_params.rs
```diff
@@ -0,0 +1,115 @@
...
+    /// External bounds are implicit to [0, rolling_stock.max_speed]
+    boundaries: Vec<f64>,
+    #[derivative(Hash(hash_with = "editoast_common::hash_float_slice::<3,_>"))]
+    #[schema(min_items = 1, example = json!([0.5, 0.6, 0.5]))]
+    /// Interval values (unit to be made explicit at use)
+    /// There must be one more value than boundaries
+    values: Vec<f64>,
+}
+
+impl<'de> Deserialize<'de> for SpeedIntervalValueCurve {
```
woshilapin, line 70: The trick about `#[serde(remove = "self")]` can be used here too.

editoast/src/core/simulation.rs
```diff
@@ -219,6 +221,11 @@ impl PhysicsConsistParameters {
             .map(|towed| f64::min(towed.const_gamma, self.traction_engine.const_gamma))
             .unwrap_or_else(|| self.traction_engine.const_gamma)
     }
+
+    pub fn compute_etcs_brake_params(&self) -> Option<EtcsBrakeParams> {
+        // TODO: handle towed rolling-stock when applying ERTMS to that case
```
woshilapin, line 226: Do we have a follow-up about that? I feel we should probably do that in the same PR because intermediate feature in this case doesn't make sense, isn't it?

woshilapin (commented):
I've only looked at the `editoast` part. Looks good. I'm only worry about the "towed rolling stock" not being handled yet, which might means an incorrect feature (not an incomplete one). What's your take on that?

editoast/editoast_schemas/src/rolling_stock/etcs_brake_params.rs
bougue-pe: There are, in the middle of the comment `in s`.
I can add parenthesis for emphasis if you prefer.

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -2905,6 +2931,7 @@ export type RollingStock = {
   effort_curves: EffortCurves;
   electrical_power_startup_time: number | null;
   energy_sources: EnergySource[];
+  etcs_brake_params: EtcsBrakeParams | null;
```
emersion, line 3054: Hm, why is this field not optional (`?`) while it's optional in `RollingStockMetadata`?

editoast/src/core/simulation.rs
bougue-pe: IMO, this is too tricky to be done in this PR.
1. We don't know what data we will get for ETCS towed RS, especially: will we force providing a max_speed? Should we provide a dedicated etcs_max_speed? (the idea is to prevent propagating illegitimate braking gamma for towed to high speeds, up to the traction engine's max_speed). If I were to choose now, I'd go force max_speed when etcs.
2. We didn't read/search about the merging of etcs_brake_params in documentation (and it's heavy). The conservative way is getting the minimum for all case, but it's a bit more code to merge curves, and it may occupy "useless" extra capacity if the merging is smarter than that.

Possible "middle ground": We may just assert that no etcs value is provided on towed RS to explicitly reject such unhandled case. Would that be OK for you?

core/kt-fast-collections-generator/src/main/kotlin/collections/ArrayList.kt
```diff
@@ -258,11 +261,22 @@ private fun CollectionItemType.generateArrayList(context: GeneratorContext, curr
...
+                val res = Mutable${simpleName}ArrayList${paramsUse}(4)
+                res.add(a)
+                res.add(b)
+                res.add(c)
+                res.add(d)
+                return res
+            }
+
+            /** GENERATED CODE */
+            fun ${paramsDecl} mutable${simpleName}ArrayListOf(a: $type, b: $type, c: $type, d: $type, e: $type): Mutable${simpleName}ArrayList${paramsUse} {
```
eckter, line 273: At some point we could just add a `vararg` version
(though it's fine as it is)

core/kt-osrd-sncf-signaling/src/main/kotlin/fr/sncf/osrd/signaling/etcs_level2/ETCS_LEVEL2.kt
```diff
@@ -0,0 +1,58 @@
...
+import fr.sncf.osrd.signaling.SignalingSystemDriver
+import fr.sncf.osrd.signaling.SignalingTrainState
+import fr.sncf.osrd.sim_infra.api.SigParametersSchema
+import fr.sncf.osrd.sim_infra.api.SigSettingsSchema
+import fr.sncf.osrd.sim_infra.api.SigState
+import fr.sncf.osrd.sim_infra.api.SigStateSchema
+import fr.sncf.osrd.utils.units.Speed
+import fr.sncf.osrd.utils.units.kilometersPerHour
+
+object ETCS_LEVEL2 : SignalingSystemDriver {
```
eckter:
```suggestion

/** PLACEHOLDER CODE copied from TVM signaling */
object ETCS_LEVEL2 : SignalingSystemDriver {
```

It's probably best to mention it somewhere in the code base

core/osrd-railjson/src/main/java/fr/sncf/osrd/railjson/schema/rollingstock/RJSEtcsBrakeParams.java
```diff
@@ -0,0 +1,81 @@
+package fr.sncf.osrd.railjson.schema.rollingstock;
+
+import com.squareup.moshi.Json;
+
+public class RJSEtcsBrakeParams {
+
+    // A_brake_emergency: the emergency deceleration curve (values in m/s²)
```
eckter: We could explain where to look for these names (`A_brake_emergency` and such).

Ideally we should directly link the PDF somewhere, even if it's not publicly available (I think?)

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/standalone_sim/SimulationEndpoint.kt
```diff
@@ -51,13 +52,20 @@ class SimulationEndpoint(
             val chunkPath = makeChunkPath(infra.rawInfra, request.path.trackSectionRanges)
             val routePath = convertRoutePath(infra.rawInfra, request.path.routes)
             val pathProps = makePathProperties(infra.rawInfra, chunkPath, routePath.toList())
+            val blockPath = mutableStaticIdxArrayListOf<Block>()
+            for (blockName in request.path.blocks) {
+                val blockId = infra.blockInfra.getBlockFromName(blockName)
+                assert(blockId != null) { "An input block from path is missing in infrastructure" }
+                blockPath.add(blockId!!)
```
eckter: *Ideally* we'd create an error similar to `UnknownRoute`, and we could write it like this:

```suggestion
                val blockId = infra.blockInfra.getBlockFromName(blockName)
                    ?: throw OSRDError(ErrorType.UnknownBlock)
                blockPath.add(blockId)
```

core/src/main/kotlin/fr/sncf/osrd/standalone_sim/etcsContextBuilder.kt
```diff
@@ -10,22 +12,42 @@ import fr.sncf.osrd.utils.distanceRangeSetOf
...
+    routePath: StaticIdxList<Route>,
+    blockPath: StaticIdxList<Block>
 ): EnvelopeSimContext.ETCSContext? {
-    val isRollingStockETCSCompatible = false // TODO
-    if (!isRollingStockETCSCompatible) return null
+    val blockInfra = infra.blockInfra
+    val etcsRanges = distanceRangeSetOf()
+    val etcsLevel2 =
+        infra.signalingSimulator.sigModuleManager.findSignalingSystemOrThrow(ETCS_LEVEL2.id)
+    var offsetSinceStart = Offset<Path>(0.meters)
```
eckter: I'm afraid it's not as simple as that, the path may start in the middle of a block. We'd need to identify where the path starts exactly, based on `chunkPath.beginOffset`. That includes matching the blocks with the chunks, it's a little tedious.

-> Also: take the minimum with the length of the path at the end.

eckter (commented):
Nice job!

For the test infra, we may not need something as complex as a clone of small_infra. Maybe an infra with a few tracks, a switch and a few slopes would be enough and easier to maintain. (But it's fine as it is if you prefer it this way)
-> Adding a non-ERTMS section would be nice.

editoast/editoast_schemas/src/rolling_stock/etcs_brake_params.rs
woshilapin: Oops, no don't bother. Let's say it was a Friday 😴

core/osrd-railjson/src/main/java/fr/sncf/osrd/railjson/schema/rollingstock/RJSEtcsBrakeParams.java
```diff
@@ -0,0 +1,81 @@
+package fr.sncf.osrd.railjson.schema.rollingstock;
+
+import com.squareup.moshi.Json;
+
+public class RJSEtcsBrakeParams {
```
eckter, line 10: nit: maybe we could add asserts for things like signs? When I read "deceleration (m/s²)", I never know if I should expect something positive or negative

-> maybe report (part of) checks from editoast to core

editoast/editoast_schemas/src/rolling_stock.rs
Wadjetz: You can also use this library [https://crates.io/crates/validator](https://crates.io/crates/validator) it is already used in the project

editoast/src/models/rolling_stock_model.rs
```diff
@@ -49,6 +50,9 @@ pub struct RollingStockModel {
     pub startup_acceleration: f64,
     pub comfort_acceleration: f64,
     pub const_gamma: f64,
+    #[model(json)]
+    #[schema(required)]
```
Wadjetz, line 53: This marks the value as required, but the type is an option. Could this be causing the inconsistency in the OpenAPI ?

front/src/common/api/generatedEditoastApi.ts
bougue-pe: As pointed by @Wadjetz [here](https://github.com/OpenRailAssociation/osrd/pull/10147#discussion_r1900660548) this is because it's marked required, which seems to be towed by the `derive(Model)`.
I tried different approaches (like `#[model(remote = "Option<ETCSBrakeParams>"]`) to circumvent it but didn't manage.

Unless @leovalais has a better idea, I think it will stay as-is (at least the practice didn't change).

My opinion is the better would be to get rid of the required everywhere, but if it's not possible, changing all to required for consistency might be an idea, but I would have a small vote to let it like that.

editoast/src/models/rolling_stock_model.rs
bougue-pe: Resolving it to regroup conversation [here](https://github.com/OpenRailAssociation/osrd/pull/10147#discussion_r1893810953), but thank you for pointing this!

editoast/editoast_schemas/src/rolling_stock.rs
bougue-pe: As discussed with @woshilapin and @leovalais : the target is to remove validator crate.
Then mostly do things in deserialize, with a free function to share code (in rolling_stock_form for example).

front/src/common/api/generatedEditoastApi.ts
leovalais: If I'm not misunderstanding the problem here.

We put `#[schema(required)]` over an `Option<T>` to indicate that the field needs to be there in the schema, whether it's part of a request or a response. In our case, `RollingStockModel` and `LightRollingStock` are only used as responses (their `-Form` counterpart are used for requests). Since unless told otherwise `serde` serializes optional fields to `null` instead of skipping them, the `etcs_brake_params` cannot be `undefined`.

**Imo it is `LightRollingStock` that's missing a bunch of `#[schema(required)]`.**

As for the `Model` macro, all `#[model(json)]` does to a `field: T` is to read/write it using `diesel` as being a `diesel_json::Json<T>`. So the type used to talk to the DB here is a `diesel_json::Json<EtcsBrakeParams>`. This means:

- `None` becomes serialized as `'null'::jsonb`
- `Some(etcs)` becomes `'etcs'::jsonb`

So keeping the `etc_brake_params JSONB NOT NULL` seems right to me.

The macro doesn't support `etcs_brake_params JSONB NULLABLE` => `Option<diesel_json::Json<EtcsBrakeParams>>`, which I agree would be cleaner, but in this context `NULL` is semantically equivalent to `'null'::jsonb`. If it's necessary I can implement that though, lmk.

We also don't support telling apart `NULL` and `null::jsonb`. That can still be done with `#[model(remote = "Option<diesel_json::Json<Option<T>>>")]` and a manually implemented conversion.

Wdyt?

editoast/editoast_schemas/src/rolling_stock/etcs_brake_params.rs
bougue-pe: Done in https://github.com/OpenRailAssociation/osrd/pull/10147/commits/983dc60fa77a8739dea21dc67bb93a2a03f10265

core/kt-osrd-sncf-signaling/src/main/kotlin/fr/sncf/osrd/signaling/etcs_level2/ETCS_LEVEL2.kt
bougue-pe: Done in https://github.com/OpenRailAssociation/osrd/pull/10147/commits/983dc60fa77a8739dea21dc67bb93a2a03f10265

core/osrd-railjson/src/main/java/fr/sncf/osrd/railjson/schema/rollingstock/RJSEtcsBrakeParams.java
bougue-pe: Done in https://github.com/OpenRailAssociation/osrd/pull/10147/commits/983dc60fa77a8739dea21dc67bb93a2a03f10265

core/osrd-railjson/src/main/java/fr/sncf/osrd/railjson/schema/rollingstock/RJSEtcsBrakeParams.java
bougue-pe: So far, I didn't implement the checks done in editoast (>= 0), just commented more in https://github.com/OpenRailAssociation/osrd/pull/10147/commits/983dc60fa77a8739dea21dc67bb93a2a03f10265

Not sure that I will prioritize those checks, so feel free to reply if you think they are needed.

core/kt-fast-collections-generator/src/main/kotlin/collections/ArrayList.kt
bougue-pe: Agreed, but I don't want to dive into this now, so I will let this as-is.

editoast/editoast_schemas/src/rolling_stock.rs
bougue-pe: Thanks, done in https://github.com/OpenRailAssociation/osrd/pull/10147/commits/39bda275b5f8eecc547adbe5c5d22a54ba30dda9 (requires custom impl of `serialize()`too).

editoast/editoast_schemas/src/rolling_stock/etcs_brake_params.rs
bougue-pe: Done in https://github.com/OpenRailAssociation/osrd/pull/10147/commits/39bda275b5f8eecc547adbe5c5d22a54ba30dda9

editoast/src/core/simulation.rs
bougue-pe: Just added an assert in https://github.com/OpenRailAssociation/osrd/pull/10147/commits/e5d9736ffbb15b9a6d7b036a8025276cb4ca2b3a, let me know if it's not enough for you.

editoast/editoast_schemas/src/rolling_stock.rs
bougue-pe: As mentioned, the checks are staying in `deserialize()`, and in https://github.com/OpenRailAssociation/osrd/pull/10147/commits/380264e0e27408be3ce5718284f529a38d8b0dc4 the work to remove `validator` crate was only bootstrapped: I will provide a separate branch with part of the work I tried, in order to have a separate work on this.

@leovalais and @woshilapin: Please let me know at least if the state of this PR is OK, I will keep you updated when the new branch is pushed.
EDIT: drafted PR in https://github.com/OpenRailAssociation/osrd/pull/10273

front/src/common/api/generatedEditoastApi.ts
bougue-pe: OK, indeed as it's only used for output and serde serializes optional fields anyway, added the `required` in LightRollingStock in https://github.com/OpenRailAssociation/osrd/pull/10147/commits/d1ab72ff25f9c30b5918b2bebac68e6e32f8aab0, so this should be OK now.

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/standalone_sim/SimulationEndpoint.kt
bougue-pe: Done in https://github.com/OpenRailAssociation/osrd/pull/10147/commits/ce97208b8579677e096b53dad101cada8b79a540

core/src/main/kotlin/fr/sncf/osrd/standalone_sim/etcsContextBuilder.kt
bougue-pe: As discussed a bit aside, a dedicated issue was created https://github.com/OpenRailAssociation/osrd/pull/10147/commits/9d87edcb8a5f25fe34b2998bc3299369452e9259.
In this PR, I first made `TravelledPath` available (and commented) in https://github.com/OpenRailAssociation/osrd/pull/10147/commits/82425fda81b47dab1cad5a588e82d974faa8884b, then did the main work related to this in https://github.com/OpenRailAssociation/osrd/pull/10147/commits/9d87edcb8a5f25fe34b2998bc3299369452e9259.

editoast/editoast_schemas/src/rolling_stock.rs
woshilapin: Sounds good for the direction, will look more closely to the other PR once the current PR is merged and the other one rebased.

woshilapin (approved):
Thank you for the improvements. The `editoast` part looks good!

eckter (approved):
LGTM

Great job there! :+1:

Erashin (approved):
Gj! We'll need to report the todo things in our meta issue.

front/src/common/api/generatedEditoastApi.ts
emersion: Ah, since it's only used as output then it's indeed better to keep it that way. Thanks for the detailed explanation!

emersion (approved):
LGTM for the frontend part!

achrafmohye (approved):
the front LGTM 👍

Khoyo (approved):

PR #10151, opened 2024-12-20
front: handle train drag

Math-R:


front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
SharglutDev, line 1: Gonna start to check the code but after testing, I have some weird behaviors : 
- When dragging a train to the left, the viewport follows the start of dragged train but it doesn't move if I drag the train to the right. Is it the expected behavior ?

https://github.com/user-attachments/assets/99bcb149-60ed-4a0c-ad8e-0787166fe619


- If I pan to the right and then drag a train (which doesn't fit in viewport anymore), the viewport doesn't follow the drag. It seems it takes count only of the initial viewport position.

https://github.com/user-attachments/assets/6df3223e-d2e3-4ace-b3ea-2a1c203b47d7

- After dragging a train, a lot of endpoints are fired. I think we should just see the `putTrainScheduleById`, `summaries` (only for the dragged train) and the `getTimetableByIdConflicts` ? It probably comes from the update which invalidates the `train_schedule` tag resulting to refetching all endpoints with this tag. This can maybe be handled by customizing the endpoint like in this [PR](https://github.com/OpenRailAssociation/osrd/pull/10091). If you don't want to deal with that now, could you create an issue to handle that so we don't forget it ? :)

https://github.com/user-attachments/assets/331e3137-19fa-4b81-9bd3-a3bbfc3e9e23

front/src/modules/trainschedule/components/Timetable/Timetable.tsx
```diff
@@ -47,6 +48,7 @@ const Timetable = ({
   dtoImport,
 }: TimetableProps) => {
   const { t } = useTranslation(['operationalStudies/scenario', 'common/itemTypes']);
+  const selectedTrainId = useSelector(getSelectedTrainId);
```
SharglutDev: nit : would be nice to group the `useSelector` together :)

front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
```diff
@@ -217,7 +215,7 @@ const ScenarioContent = ({
           )}
           <div className="scenario-results">
             {isMacro ? (
-              <div className="h-100 p-1">
+              <div className={cx(collapsedTimetable ? 'macro-container' : 'h-100 p-1')}>
```
SharglutDev: A remnant from a rebase ? `macro-container` class doesn't seem to exist.

front/src/applications/operationalStudies/views/SimulationResults.tsx
```diff
@@ -106,6 +116,30 @@ const SimulationResults = ({
 
   const conflictZones = useProjectedConflicts(infraId, conflicts, projectionData?.path);
 
+  const selectedTrainSummary = useMemo(
+    () => trainScheduleSummaries?.find((train) => train.id === selectedTrainSchedule?.id),
+    [trainScheduleSummaries, selectedTrainSchedule]
+  );
+
+  const handleTrainDrag = async (
+    draggedTrain: TrainSpaceTimeData,
```
SharglutDev: Seems we use only the dragged train id here. Maybe we could use only that in this function ?

```suggestion
    draggedTrainId: number,
```

front/src/applications/operationalStudies/views/SimulationResults.tsx
```diff
@@ -106,6 +116,30 @@ const SimulationResults = ({
...
+    newDepartureTime: Date,
+    { stopPanning }: { stopPanning: boolean }
+  ) => {
+    if (stopPanning) {
+      // update in the database
+      dispatch(updateSelectedTrainId(draggedTrain.id));
+      updateTrainDepartureTime(draggedTrain.id, newDepartureTime);
+    } else {
```
SharglutDev, line 134: When we are dragging, the dragged train becomes blue meaning it is now the selected one but in the timetable, we update the selected train only when `stopPanning`. Maybe we should also use a state here to "pre" update the selected train in the timetable before the dispatch ?

front/src/applications/operationalStudies/views/SimulationResults.tsx
```diff
@@ -36,17 +38,19 @@ type SimulationResultsProps = {
   collapsedTimetable: boolean;
   infraId?: number;
   projectionData?: ProjectionData;
-  selectedTrainSummary?: TrainScheduleWithDetails;
+  trainScheduleSummaries?: TrainScheduleWithDetails[];
```
SharglutDev, line 41: Why do we need change this since we use only the selectedTrainSummary after all ?

front/src/applications/operationalStudies/hooks/useScenarioData.ts
```diff
@@ -87,13 +84,18 @@ const useScenarioData = (
   useAutoUpdateProjection(infra, timetable.train_ids, trainScheduleSummaries);
 
   useEffect(() => {
-    if (!rawTrainSchedules) {
-      setTrainSchedules(undefined);
-    } else {
+    const fetchTrains = async () => {
```
SharglutDev, line 87: Since we do the same thing as before but with a `lazyQuery`. Is it also because of the `invalidateTags` to avoid some unnecessary refetching ?

front/src/applications/operationalStudies/hooks/useScenarioData.ts
```diff
@@ -160,29 +155,104 @@ const useScenarioData = (
...
+        return newProjectedTrainsById;
+      });
+
+      setTrainSchedules((prev) => {
+        const newTrainSchedulesById = {
+          ...keyBy(prev, 'id'),
+          ...keyBy([trainScheduleResult], 'id'),
+        };
+        const newTrainSchedules = sortBy(Object.values(newTrainSchedulesById), 'start_time');
+        return newTrainSchedules;
```
SharglutDev: To harmonize with other part l.123
```suggestion
        return sortBy(Object.values(newTrainSchedulesById), 'start_time');
```

front/src/applications/operationalStudies/views/SimulationResults.tsx
```diff
@@ -77,8 +81,14 @@ const SimulationResults = ({
     pathProperties
   );
 
-  // Compute path items coordinates in order to place them on the map
+  const [projectPathTrainResult, setProjectPathTrainResult] = useState<TrainSpaceTimeData[]>([]);
+  useEffect(() => {
```
SharglutDev: Improves readability :)

```suggestion
  const [projectPathTrainResult, setProjectPathTrainResult] = useState<TrainSpaceTimeData[]>([]);
  
  useEffect(() => {
```

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -11,6 +11,10 @@ import {
   OccupancyBlockLayer,
 } from '@osrd-project/ui-spacetimechart';
 import type { Conflict } from '@osrd-project/ui-spacetimechart';
+import type {
+  SpaceTimeChartProps,
+  HoveredItem,
+} from '@osrd-project/ui-spacetimechart/dist/lib/types';
```
SharglutDev, line 22: Is it ok to import these from `dist` ? Should we rather make them "exportable" in osrd-ui ?

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -44,6 +51,12 @@ type ManchetteWithSpaceTimeChartProps = {
     totalTrains: number;
     allTrainsProjected: boolean;
   };
+  handleTrainDrag?: (
+    draggedTrain: TrainSpaceTimeData,
+    newDepartureTime: Date,
+    { stopPanning }: { stopPanning: boolean }
+  ) => Promise<void>;
+  selectTrainPath?: (trainId: number) => void;
```
SharglutDev: Seems like we don't use this prop

front/src/applications/operationalStudies/views/SimulationResults.tsx
```diff
@@ -106,6 +116,30 @@ const SimulationResults = ({
 
   const conflictZones = useProjectedConflicts(infraId, conflicts, projectionData?.path);
 
+  const selectedTrainSummary = useMemo(
+    () => trainScheduleSummaries?.find((train) => train.id === selectedTrainSchedule?.id),
+    [trainScheduleSummaries, selectedTrainSchedule]
+  );
+
+  const handleTrainDrag = async (
```
SharglutDev, line 125: Should move this function in `ManchetteWithSpaceTimeChartWrapper` ? Since there is the `handleClick` on a train inside which now requires `dispatch` in the component. I think both these functions should be at the same place.

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -279,6 +363,9 @@ const ManchetteWithSpaceTimeChartWrapper = ({
             }
             timeOrigin={Math.min(...projectPathTrainResult.map((p) => +p.departureTime))}
             {...spaceTimeChartProps}
+            onPan={onPanOverloaded}
+            onClick={handleClick}
+            onHoveredChildUpdate={handleHoveredChildUpdate}
```
SharglutDev, line 393: We should change the cursor when hovering a train so we know it's interactive. And probably when dragging, we would want to see the "grabbing" cursor ?
Maybe @thibautsailly also wants to highlight the train differently when just hovering it ?

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -185,6 +212,49 @@ const ManchetteWithSpaceTimeChartWrapper = ({
...
+  const onPanOverloaded: SpaceTimeChartProps['onPan'] = async (payload) => {
+    const { isPanning } = payload;
+
+    if (!handleTrainDrag) {
+      // if no handleTrainDrag, we pan normally
+      spaceTimeChartProps.onPan(payload);
+      return;
+    }
+
+    // if dragging
```
SharglutDev, line 233: Maybe not useful 
```suggestion
```

front/src/modules/timesStops/ReadOnlyTime.tsx
```diff
@@ -9,10 +9,11 @@ type ReadOnlyTimeProps = CellProps<TimeExtraDays | undefined, string>;
 
 const ReadOnlyTime = ({ rowData }: ReadOnlyTimeProps) => {
   const { time, daySinceDeparture, dayDisplayed } = rowData || {};
+  const { t } = useTranslation('timesStops');
```
SharglutDev, line 12: Is it to fix a bug ? Maybe it should go in a separate commit :)

SharglutDev (commented):
Thank you for the PR, left some comments.
> [!NOTE]
> Before merging, we should be carefull of some of your commits with spelling mistakes  if you keep them :)

front/src/modules/timesStops/ReadOnlyTime.tsx
Math-R: Actually yes is to fix a bug fired by the trainupdate after drag, it seems like it was there for some times But we had never matched the case where it was fired.
It's not even core it's just a line change because of a miss linting of code, don't think its relevant to put it in a specific commit

emersion: I'd recommend splitting this into a separate commit, because:

- It's not part of the train drag feature, it's an unrelated bug which happens by chance to be triggered by a change in a completely different spot.
- It reduces cognitive load for reviewers: they first see the bugfix in a commit, then the changes related to the new feature, instead of mixed up changes.
- If we need to revert your commit because it causes a regression, we don't want to revert this bugfix.

See: https://www.bitsnbites.eu/git-history-work-log-vs-recipe/

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
Math-R: The behaviour who recalculate view port are handle by the component itself and also the hook useManchetteWithSpaceTimeChart, it can be discuss but treated in another osrd-ui PR

front/src/modules/timesStops/ReadOnlyTime.tsx
Math-R: makes sense

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
SharglutDev: Ok after some digging, turns out there is already a mockup for that : https://www.sketch.com/s/bcbac4ff-7f74-4f1a-9e10-08d115c47ed8/p/0A77AD5F-6060-4AB9-9DED-DAD4279597A1/canvas.

The cursor isn't supposed to change but the style (color and width) of the curve changes when hovered or selected.

front/src/applications/operationalStudies/hooks/useScenarioData.ts
Math-R: i don't understand your question, seems like it misses some words

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
Math-R: We probably could make them "exportable" but for my part yes it's ok to do this like this since there is no other solution right now, it could be treated as an improvement for both projects in further development

Math-R:
> Thank you for the PR, left some comments.
> 
> Note
> 
> Before merging, we should be carefull of some of your commits with spelling mistakes if you keep them :)

Sure, there will be some squash fixup and reword during the rebase

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
Math-R: I can't revert the cursor change but for the spacetimechart render style, it has to be done in osrd-ui

front/src/applications/operationalStudies/hooks/useScenarioData.ts
Math-R: sure 👍🏻

front/src/applications/operationalStudies/views/SimulationResults.tsx
Math-R: @clarani can you help me on this one ? i don't have the answer on my side

front/src/applications/operationalStudies/views/SimulationResults.tsx
Math-R: I made the change

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
Math-R: The function is an overload of an existing one gave by manchetteWithSpaceTimaChart hook and i think it's more friendly to keep some commentaries for further development task

front/src/applications/operationalStudies/views/SimulationResults.tsx
Math-R: made a change to update with the handle click lonely if the dragged train is not actually the one selected, and to update the selected one as soon as the drag occurs

front/src/applications/operationalStudies/views/SimulationResults.tsx
Math-R: Not sure about this one, maybe the dispatch use for selected train should be pass as a props so that the manchetteWithSpaceTimechartWrapper don't have to know how to modify store ?

front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
Math-R: removed =)

front/src/applications/operationalStudies/hooks/useScenarioData.ts
SharglutDev: To reformulate : why did we change the `useQuery` for a `lazyQuery` here ? Seems like we are just fetching the train schedules when mounting the component (as a `useQuery` does) and sorting the result (like we also did before).

front/src/applications/operationalStudies/views/SimulationResults.tsx
SharglutDev: Agreed. I think that's what we try to do in the rest of the app. But rather than passing the dispatch in props, we should maybe pass the whole `handleClick`function to harmonize with the `handleTrainDrag` (meaning both handler should be in the same component either in `SimulationResult` if we keep the dispatch there or in `MancheWithSpaceTime` if we allow to call a dispatch over there)

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
SharglutDev, line 1: Not sure if it's perceptible on the video, but it seems like there like a small "lag" when I start dragging. The drag movement is a bit hatched at the beginning and then it becomes fluid again. Don't know where it comes from.

https://github.com/user-attachments/assets/dad1d653-3ce0-47fc-927c-8b967b4a9667

Math-R: it's seems that it is due to the modification to highlight the train in the timetable during the path click, it's all the app who seems to render and makes a small lag, so it adresses the whole store gestion in this part of the app

Math-R: it's seems that it is due to the modification to highlight the train in the timetable during the path click, it's all the app who seems to render and makes a small lag, so it adresses the whole store gestion in this part of the app

front/src/applications/operationalStudies/hooks/useScenarioData.ts
Math-R: We should ask @clarani about what motivate this change

front/src/applications/operationalStudies/views/SimulationResults.tsx
Math-R: In that case if it's requires refactoring stuff, could it be in a further PR to ?

SharglutDev: Actually the `handleDrag` and `handleClick` both come from your PR so it would be better to fix it now I think. You just went for two different options for these two handlers and we should correct that :)

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
SharglutDev: Yep that's what I thought because I don't remember having this "lag" before the change to directly update the selectedTrainId.
One way to fix this would be like an "optimistic" update. We keep dispatching only when drag is over but we could just change the trains background style when the drag starts (but seems like a pain to do that without context since both components are far from each another)

front/src/applications/operationalStudies/views/SimulationResults.tsx
SharglutDev: After discussions, it helps for performance

front/src/applications/operationalStudies/hooks/useScenarioData.ts
SharglutDev: Performance change as well

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
SharglutDev: After some discussions with @maelysLeratRosso, she thinks the synchro change between the timetable and the space time chart for the selected train is necessary so we can keep it like that. If this "lag" is too much trouble, we will find a solution later

front/src/applications/operationalStudies/views/SimulationResults.tsx
Math-R: So to handle this i choose to keep the handleClick inside the Manchette[..]Wrapper, so it can make the difference about wich element is clicked but i pass the dispatch as a new prop

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
Math-R: Added a new function to determine the path color

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
SharglutDev: After discussions with @maelysLeratRosso. These behaviors should be discussed in refinement and will be addressed in a new issue. As this issue doesn't specify any behavior for the feature, it shouldn't block it

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
SharglutDev: Will be done in a different PR after some changes in osrd-ui

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
SharglutDev: The width and color changes for the selected train and default will be done in an osrd-ui PR

SharglutDev (approved):
Lgtm and tested, good job

PR #10154, opened 2024-12-20
front: fix train selection in the anterior or posterior train list

### LINKED ISSUE #10133: STDCM: Can't select train for linked path
### What happened?

Image

The train is selected and the designited fields are filled 

1. Go to STDCM page
2. Search for a train in the "Anterior" or "Posterior" train list
Image
3. Click on the train from the search results

Recette (SNCF)

Firefox, Google Chrome

6580cfe

theocrsb:
closes #10133

### DIFF
--- a/front/src/applications/stdcm/types.ts
+++ b/front/src/applications/stdcm/types.ts
@@ -183,7 +183,7 @@ export type StdcmLinkedTrainExtremity = {
   ch: string;
   date: string;
   geographic: GeoJsonPoint;
-  isoArrivalTime: string;
+  arrivalDate: Date;
   name: string;
   obj_id: string;
   time: string;
--- a/front/src/applications/stdcm/utils/computeOpSchedules.ts
+++ b/front/src/applications/stdcm/utils/computeOpSchedules.ts
@@ -29,12 +29,12 @@ const computeOpSchedules = (startTime: Date, msFromStartTime: number) => {
     origin: {
       date: originDate,
       time: originTime,
-      isoArrivalTime: subtractDurationFromDate(startTime, 1800).toISOString(),
+      arrivalDate: subtractDurationFromDate(startTime, 1800),
     },
     destination: {
       date: destinationDate,
       time: destinationTime,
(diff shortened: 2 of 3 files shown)


clarani (changes requested):
Actually, the issue is not in this function, it's because we should not receive a string if we're expecting a date.

I let you search a bit, the problem comes from the action `updateLinkedTrainExtremity`, where we update the pathStep :wink:

front/src/applications/stdcm/utils/computeOpSchedules.ts
```diff
@@ -29,12 +29,12 @@ const computeOpSchedules = (startTime: Date, msFromStartTime: number) => {
     origin: {
       date: originDate,
       time: originTime,
-      isoArrivalTime: subtractDurationFromDate(startTime, 1800).toISOString(),
+      isoArrivalTime: new Date(subtractDurationFromDate(startTime, 1800)),
```
clarani: `subtractDurationFromDate` already returns a date, so you just need to remove the toISOString()
```suggestion
      isoArrivalTime: subtractDurationFromDate(startTime, 1800),
```

front/src/applications/stdcm/types.ts
```diff
@@ -183,7 +183,7 @@ export type StdcmLinkedTrainExtremity = {
   ch: string;
   date: string;
   geographic: GeoJsonPoint;
-  isoArrivalTime: string;
+  isoArrivalTime: Date;
```
clarani: Can you rename the field here ? Maybe arrivalDate ?

theocrsb: Done

front/src/applications/stdcm/utils/computeOpSchedules.ts
theocrsb: Done

clarani (approved):
LGTM :white_check_mark:

Akctarus (approved):
tested, LGTM :)

PR #10158, opened 2024-12-20
front: drop StdcmResultsOperationalPoint.departureTime

emersion:
This is unused, and always set to the same constant value for all operational points.


Yohh (approved):

SharglutDev (approved):
Lgtm and tested

PR #10160, opened 2024-12-20
editoast: bump osm4routing to avoid two compilations of protobuf

Tristramg:
This allows a small compilation time improvement, as there is no longer two protobuf versions to compile

The total compilation time from scratch goes down from 186 seconds to 174. It’s a oneshot benchmark, so not very reliable, but there aren’t much downsides to this bump

Here is a screenshot of the situation before obtained by calling `cargo build --timings`
image

### DIFF
--- a/editoast/Cargo.lock
+++ b/editoast/Cargo.lock
@@ -2836,7 +2836,7 @@ dependencies = [
  "log",
  "num-traits",
  "pointy",
- "protobuf 3.7.1",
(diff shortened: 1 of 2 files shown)


leovalais (approved):
LGTM, less things to compile! ☢️ ⚡

Jean is vacationing btw.

woshilapin (approved):
🚀

PR #10168, opened 2024-12-23
front: fix combobox behavior

### LINKED ISSUE #8932: STDCM: The mouse scroll for the combo box field isn't working
### What happened?

https://github.com/user-attachments/assets/2a7e1e7f-e1f7-42d1-8d05-85a458883ed1

I can scroll using both the mouse and keyboard

Try scrolling using the mouse **( not mouse wheel)**

Recette (SNCF)

Google Chrome

fbf8e3e

Uriel-Sautron:
close #8932

### DIFF
--- a/front/src/applications/stdcm/views/StdcmView.tsx
+++ b/front/src/applications/stdcm/views/StdcmView.tsx
@@ -181,7 +181,8 @@ const StdcmView = () => {
   if (error && error.message !== NO_CONFIG_FOUND_MSG) throw error;
 
   return (
-    <div role="button" tabIndex={0} className="stdcm" onClick={() => setShowStatusBanner(false)}>
+    // eslint-disable-next-line jsx-a11y/interactive-supports-focus
+    <div role="button" className="stdcm" onClick={() => setShowStatusBanner(false)}>
       <StdcmHeader
         isDebugMode={isDebugMode}
         onDebugModeToggle={setIsDebugMode}


front/src/applications/stdcm/views/StdcmView.tsx
```diff
@@ -181,7 +181,8 @@ const StdcmView = () => {
   if (error && error.message !== NO_CONFIG_FOUND_MSG) throw error;
 
   return (
-    <div role="button" tabIndex={0} className="stdcm" onClick={() => setShowStatusBanner(false)}>
```
emersion, line 184: Removing this has accessibility consequences: it's no longer possible to activate the button via the keyboard.

emersion: Hm, but it's not really a button since it wraps the whole STDCM view… I think we should drop `role="button"` as well?

Uriel-Sautron: Interacting with this div just removes the simulation banner. Do we need to activate it on the keyboard?

emersion: The container doesn't act like a button, so I don't think `role="button"` is appropriate here.

There are many ways to "interact" with a form, using the mouse to do so is just one of them. For instance, the Tab key focuses the next form element, or the "Next" key on an on-screen keyboard (e.g. on an Android tablet) has the same effect.

Maybe `onFocus` is what we want here instead of `onClick`?

clarani:
Not relevant anymore, since the bug was solved by https://github.com/OpenRailAssociation/osrd/pull/10293

PR #10169, opened 2024-12-23
front: move allWaypoints to ManageTrainScheduleContextType

emersion:
This value should be updated every time pathSteps or suggestedOperationalPoints change. Instead of manually updating it when touching these two values, memoize it.

### DIFF
--- a/front/src/applications/operationalStudies/hooks/useManageTrainScheduleContext.tsx
+++ b/front/src/applications/operationalStudies/hooks/useManageTrainScheduleContext.tsx
@@ -23,7 +23,7 @@ type ManageTrainScheduleContextType = {
   pathfindingState: PathfindingState;
   infraInfo: { infra?: InfraWithState; reloadCount: number };
   /** Operational points along the path (including origin and destination) and vias added by clicking on map */
-  allWaypoints?: SuggestedOP[];
+  pathStepsAndSuggestedOPs?: SuggestedOP[];
 } | null;
 
 const ManageTrainScheduleContext = createContext<ManageTrainScheduleContextType>(null);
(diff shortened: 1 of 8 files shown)


front/src/applications/operationalStudies/hooks/useManageTrainScheduleContext.tsx
```diff
@@ -16,6 +22,8 @@ type ManageTrainScheduleContextType = {
   launchPathfinding: (pathSteps: (PathStep | null)[]) => void;
   pathfindingState: PathfindingState;
   infraInfo: { infra?: InfraWithState; reloadCount: number };
+  /** Operational points along the path (including origin and destination) and vias added by clicking on map */
+  allWaypoints?: SuggestedOP[];
```
emersion: The original comment is bogus: the list also contains all suggested OPs. The name `allWaypoints` is also confusing for the same reason.

Maybe we should rename this to something else while at it? For instance, `waypointsAndSuggestedOPs`?

clarani: I used waypoints because it's all the special points on the path (pathSteps or operational points). If you want to rename it, `pathStepsAndSuggestedOPs` would be better I think :)

clarani (approved):
LGTM :white_check_mark:

theocrsb (approved):

front/src/applications/operationalStudies/hooks/useManageTrainScheduleContext.tsx
emersion: Added a commit to rename the field.

clarani (approved):
LGTM :white_check_mark:

PR #10170, opened 2024-12-23
front: drop TimesStopsRow.isWaypoint

emersion:
Now we have an unambiguous way to figure out whether a row is a waypoint explicitly selected by the user: if it has a path step ID, then it's a waypoint.

Check the path step ID instead of having to keep a separate field synced up.


Uriel-Sautron (approved):
LGTM, tested !!!

kmer2016 (approved):
LGTM !

PR #10192, opened 2024-12-24
front: refacto stdcm op schedule

### LINKED ISSUE #10190: stdcm: refacto stdcm op schedule

clarani:
closes #10190

See commits

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmOpSchedule.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmOpSchedule.tsx
@@ -24,10 +24,6 @@ type StdcmOpScheduleProps = {
     arrivalTimeHours: number;
     arrivalTimeMinutes: number;
   };
-  opToleranceValues: {
-    arrivalToleranceBefore: number;
-    arrivalToleranceAfter: number;
-  };
   opId: string;
   isOrigin?: boolean;
 };
@@ -43,7 +39,6 @@ const StdcmOpSchedule = ({
   pathStep,
   opTimingData,
   opScheduleTimeType,
-  opToleranceValues,
   opId,
   isOrigin = false,
 }: StdcmOpScheduleProps) => {
@@ -54,26 +49,29 @@ const StdcmOpSchedule = ({
   const { getSearchDatetimeWindow } = useOsrdConfSelectors();
   const searchDatetimeWindow = useSelector(getSearchDatetimeWindow);
 
-  const { arrivalDate, arrivalTime, arrivalTimeHours, arrivalTimeMinutes, arrivalToleranceValues } =
-    useMemo(() => {
-      const isArrivalDateValid =
-        opTimingData?.arrivalDate &&
-        isArrivalDateInSearchTimeWindow(opTimingData.date, searchDatetimeWindow);
+  const { arrivalDate, arrivalTime, arrivalTimeHours, arrivalTimeMinutes } = useMemo(() => {
+    const isArrivalDateValid =
+      opTimingData?.arrivalDate &&
+      isArrivalDateInSearchTimeWindow(opTimingData.date, searchDatetimeWindow);
 
-      return {
-        arrivalDate:
-          opTimingData && isArrivalDateValid
-            ? opTimingData.date
-            : defaultDate(searchDatetimeWindow?.begin),
-        arrivalTime: opTimingData?.arrivalTime,
-        arrivalTimeHours: opTimingData?.arrivalTimeHours,
-        arrivalTimeMinutes: opTimingData?.arrivalTimeMinutes,
-        arrivalToleranceValues: {
-          minusTolerance: opToleranceValues.arrivalToleranceBefore,
-          plusTolerance: opToleranceValues.arrivalToleranceAfter,
-        },
-      };
-    }, [opTimingData, opToleranceValues, searchDatetimeWindow]);
+    return {
+      arrivalDate:
+        opTimingData && isArrivalDateValid
+          ? opTimingData.date
+          : defaultDate(searchDatetimeWindow?.begin),
+      arrivalTime: opTimingData?.arrivalTime,
+      arrivalTimeHours: opTimingData?.arrivalTimeHours,
+      arrivalTimeMinutes: opTimingData?.arrivalTimeMinutes,
+    };
+  }, [opTimingData, searchDatetimeWindow]);
+
+  const tolerances = useMemo(
+    () => ({
+      minusTolerance: pathStep.tolerances.before,
+      plusTolerance: pathStep.tolerances.after,
+    }),
+    [pathStep.tolerances]
+  );
 
   const selectableSlot = useMemo(
     () =>
@@ -181,7 +179,7 @@ const StdcmOpSchedule = ({
             <TolerancePicker
               id={`stdcm-tolerance-${opId}`}
               label={t('trainPath.tolerance')}
-              toleranceValues={arrivalToleranceValues}
+              toleranceValues={tolerances}
               onChange={() => {}}
               onToleranceChange={({ minusTolerance, plusTolerance }) => {
                 dispatch(
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmOpSchedule.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmOpSchedule.tsx
@@ -4,7 +4,10 @@ import { DatePicker, Select, TimePicker, TolerancePicker } from '@osrd-project/u
 import { useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
 
-import { useOsrdConfSelectors } from 'common/osrdContext';
+import { useOsrdConfActions, useOsrdConfSelectors } from 'common/osrdContext';
+import type { StdcmConfSliceActions } from 'reducers/osrdconf/stdcmConf';
+import type { StdcmPathStep } from 'reducers/osrdconf/types';
+import { useAppDispatch } from 'store';
 import { formatDateString, isArrivalDateInSearchTimeWindow } from 'utils/date';
 import { createStringSelectOptions } from 'utils/uiCoreHelpers';
 
@@ -12,15 +15,7 @@ import type { ArrivalTimeTypes, ScheduleConstraint } from '../../types';
 
 type StdcmOpScheduleProps = {
   disabled: boolean;
-  onArrivalChange: ({ date, hours, minutes }: ScheduleConstraint) => void;
-  onArrivalTypeChange: (arrivalType: ArrivalTimeTypes) => void;
-  onArrivalToleranceChange: ({
-    toleranceBefore,
-    toleranceAfter,
-  }: {
-    toleranceBefore: number;
-    toleranceAfter: number;
-  }) => void;
+  pathStep: Extract<StdcmPathStep, { isVia: false }>;
   opScheduleTimeType: ArrivalTimeTypes;
   opTimingData?: {
     date: Date;
@@ -45,9 +40,7 @@ const defaultDate = (date?: Date) => {
 
 const StdcmOpSchedule = ({
   disabled,
-  onArrivalChange,
-  onArrivalTypeChange,
-  onArrivalToleranceChange,
+  pathStep,
   opTimingData,
   opScheduleTimeType,
   opToleranceValues,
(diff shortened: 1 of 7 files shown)


Uriel-Sautron (commented):
Good work. I have nothing to say about the code. I'm still testing the application's behavior.
Thanks for the refacto.

front/src/applications/stdcm/components/StdcmForm/StdcmOpSchedule.tsx
```diff
@@ -1,83 +1,55 @@
...
+    };
+  }, [pathStep.arrival]);
+
+  const tolerances = useMemo(
+    () => ({
+      minusTolerance: pathStep.tolerances?.before || 0,
+      plusTolerance: pathStep.tolerances?.after || 0,
+    }),
+    [pathStep.tolerances]
+  );
```
SharglutDev: Seems like `tolerances` is always defined (we can also update `StdcmPathStep` type) and we can't set it to undefined.

```suggestion
  const tolerances = useMemo(
    () => ({
      minusTolerance: pathStep.tolerances.before,
      plusTolerance: pathStep.tolerances.after,
    }),
    [pathStep.tolerances]
  );
```

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -150,11 +151,34 @@ export const stdcmConfSlice = createSlice({
...
-      if (action.payload.workScheduleGroupId) {
-        state.workScheduleGroupId = action.payload.workScheduleGroupId;
-      }
-      if (action.payload.temporarySpeedLimitGroupId) {
-        state.temporarySpeedLimitGroupId = action.payload.temporarySpeedLimitGroupId;
+      state.workScheduleGroupId = action.payload.workScheduleGroupId;
+      state.temporarySpeedLimitGroupId = action.payload.temporarySpeedLimitGroupId;
+
+      // check that the arrival dates are in the search time window
+      const { searchDatetimeWindow } = action.payload;
```
SharglutDev: You can define this at the beginning of the function to reuse it l.153

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -150,11 +151,34 @@ export const stdcmConfSlice = createSlice({
...
+      if (searchDatetimeWindow) {
```
SharglutDev: This always defined

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -150,11 +151,34 @@ export const stdcmConfSlice = createSlice({
...
+        const origin = state.stdcmPathSteps.at(0)! as Extract<StdcmPathStep, { isVia: false }>;
+        const destination = state.stdcmPathSteps.at(-1)! as Extract<
+          StdcmPathStep,
+          { isVia: false }
+        >;
```
SharglutDev: If you cast, I don't think you need the `!`

```suggestion
        const origin = state.stdcmPathSteps.at(0) as Extract<StdcmPathStep, { isVia: false }>;
        const destination = state.stdcmPathSteps.at(-1) as Extract<
          StdcmPathStep,
          { isVia: false }
        >;
```

front/src/applications/stdcm/components/StdcmForm/StdcmOpSchedule.tsx
```diff
@@ -101,26 +73,27 @@ const StdcmOpSchedule = ({
...
-    ) {
-      onArrivalChange({
-        date: defaultDate(searchDatetimeWindow?.begin),
-        hours: arrivalTimeHours || 0,
-        minutes: arrivalTimeMinutes || 0,
-      });
-    }
-  }, [searchDatetimeWindow, opScheduleTimeType]);
+  const onArrivalChange = ({ date, hours, minutes }: ScheduleConstraint) => {
+    const newDate = new Date(date);
```
SharglutDev, line 79: `date` is already a `Date`

front/src/applications/stdcm/components/StdcmForm/StdcmOpSchedule.tsx
```diff
@@ -161,7 +134,11 @@ const StdcmOpSchedule = ({
             hours={arrivalTimeHours}
             minutes={arrivalTimeMinutes}
             onTimeChange={({ hours, minutes }) => {
-              onArrivalChange({ date: arrivalDate, hours, minutes });
+              onArrivalChange({
+                date: pathStep.arrival || searchDatetimeWindow!.begin,
```
SharglutDev, line 140: I think you anticipated it but with your changes, `pathStep.arrival` can't be undefined anymore

front/tests/utils/setup-utils.ts
```diff
@@ -177,8 +177,8 @@ export async function createDataForTests(): Promise<void> {
 
     const stdcmEnvironment = {
       infra_id: smallInfra.id,
-      search_window_begin: '2024-10-17T00:00:00Z',
-      search_window_end: '2024-10-18T23:59:59Z',
+      search_window_begin: '2024-10-17T00:00:01+02:00',
+      search_window_end: '2024-10-18T23:59:59+02:00',
```
SharglutDev: I'm not sure this is gonna work :/

SharglutDev (commented):
Really great refacto !

front/src/reducers/osrdconf/stdcmConf/index.ts
clarani: Done

front/src/reducers/osrdconf/stdcmConf/index.ts
clarani: Done

front/src/applications/stdcm/components/StdcmForm/StdcmOpSchedule.tsx
clarani: Actually here we need to create a new object to allow React to detect a change. Otherwise, the useEffect/useMemo won't be triggered

front/src/applications/stdcm/components/StdcmForm/StdcmOpSchedule.tsx
clarani: Yes, but in the store, we don't always have a searchDatetimeWindow, and the arrival can be undefined... We will be able to fix this once the data is in the context rather than in the store

front/tests/utils/setup-utils.ts
clarani: Hum I modified it, is it better like this ?

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -134,28 +135,46 @@ export const stdcmConfSlice = createSlice({
...
+        infraID: number;
+        timetableID: number;
+        searchDatetimeWindow: {
+          begin: Date;
+          end: Date;
+        };
+        electricalProfileSetId?: number;
+        temporarySpeedLimitGroupId?: number;
+        workScheduleGroupId?: number;
+      }>
```
SharglutDev: I'm sad since a long time (start with `PathProperties` that we have trouble to customize our types for this kind of mix of NonNullable properties in an object. I recently found a nice video to build a "complex "type helping for that which I customized for our needs. With some documentation, it would look like this : 
```
/**
 * Transform the provided keys of an object to be non-nullable.
 *
 * -? removes the optional modifier
 *
 * NonNullable removes the nullable modifier
 * @param Type The object type we want to transform
 * @param K The keys we want to make non-nullable
 * @example NonNullableObject<{ a?: string | null; b?: string; c?: string }, 'a' | 'c'> = { a: string; b?: string; c: string }
 */
type NonNullableObject<Type, K extends keyof Type> = {
  [Property in keyof Type as Property extends K ? Property : never]-?: NonNullable<Type[Property]>;
} & {
  [Property in keyof Type as Property extends K ? never : Property]: Type[Property];
};

/**
 * Shortcut type to pick some properties of an object and make some of them non-nullable.
 * @example PickAndNonNullableFields<{ a?: string | null; b?: string; c?: string }, 'a' | 'c', 'a'> = { a: string; c?: string }
 */
type PickAndNonNullableFields<
  Type,
  PickKeys extends keyof Type,
  TransformKeys extends keyof Type,
> = NonNullableObject<Pick<Type, PickKeys>, Extract<TransformKeys, PickKeys>>;
```
What do you think ? We could define them in `src/utils/types.ts`. And in the future, we could refacto some `PathProperties` types and maybe others.

For here, it would look like that
```suggestion
      action: PayloadAction<
        PickAndNonNullableFields<
          OsrdStdcmConfState,
          | 'infraID'
          | 'timetableID'
          | 'electricalProfileSetId'
          | 'workScheduleGroupId'
          | 'temporarySpeedLimitGroupId'
          | 'searchDatetimeWindow',
          'infraID' | 'timetableID' | 'searchDatetimeWindow'
        >
      >
```

front/src/applications/stdcm/components/StdcmForm/StdcmOpSchedule.tsx
SharglutDev: Ok make sense. Thank you for the comment :)

front/src/applications/stdcm/components/StdcmForm/StdcmOpSchedule.tsx
SharglutDev: Hmm if we aren't sure to always have searchDatetimeWindow, how can we be sure here that the fallback will work ? What is the difference ?

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -134,28 +135,46 @@ export const stdcmConfSlice = createSlice({
...
+      ) {
+        newOrigin = { ...origin, arrival: searchDatetimeWindow.begin };
       }
-      if (action.payload.temporarySpeedLimitGroupId) {
-        state.temporarySpeedLimitGroupId = action.payload.temporarySpeedLimitGroupId;
+      if (
+        !destination.arrival ||
+        !isArrivalDateInSearchTimeWindow(destination.arrival, searchDatetimeWindow)
+      ) {
+        newDestination = { ...destination, arrival: searchDatetimeWindow.begin };
```
SharglutDev: Will it work with `arrival: searchDatetimeWindow.begin` ? This means the train will be able to leave before the time window. Is it acceptable ?

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -134,28 +135,46 @@ export const stdcmConfSlice = createSlice({
...
-          | 'searchDatetimeWindow'
-        >
-      >
+      action: PayloadAction<{
+        infraID: number;
+        timetableID: number;
+        searchDatetimeWindow: {
+          begin: Date;
+          end: Date;
+        };
```
SharglutDev: Hmm I spoke too fast for `searchDateTimeWindow`. It seems the change breaks in `ScenarioExplorer` because `scenarioDateTimeWindow` can be undefined and `trainsDetails` can be `undefined` or `length === 0`. 
But is there any utility to test stdcm on an empty scenario ? Should we filter them from the explorer ?

front/tests/utils/setup-utils.ts
SharglutDev: Let's see when the CI works again :)

front/src/reducers/osrdconf/stdcmConf/index.ts
clarani: You're right, it's much better :)

SharglutDev (approved):
Lgtm and tested, good job !

front/tests/utils/setup-utils.ts
```diff
@@ -177,8 +177,8 @@ export async function createDataForTests(): Promise<void> {
 
     const stdcmEnvironment = {
       infra_id: smallInfra.id,
-      search_window_begin: '2024-10-17T00:00:00Z',
-      search_window_end: '2024-10-18T23:59:59Z',
+      search_window_begin: new Date('2024-10-17T00:00:01+02:00').toISOString(),
```
Maymanaf: Why not directly use the calculated date?
` 'search_window_begin": "2024-10-16T22:00:01Z'`
and 
`  'search_window_end": "2024-10-18T21:59:59Z'`

Maymanaf (approved):
Good work ! 
LGTM for e2e-tests

PR #10193, opened 2024-12-24
front: add stdcm simulation sheet e2e test

### LINKED ISSUE #9587: front: LMR simulation sheet e2e test
# Description

Test the simulation sheet under its pdf form to make sure the display's correct and no information is missing.

# AC

- [ ] TBD, i don't know if we're gonna use snapshooting or not
- [ ] The sure thing is this test needs to be added at the end of the LMR test because it requires a simulation result

Maymanaf:
Closes #9587


Wadjetz (approved):

front/package.json
```diff
@@ -69,6 +69,7 @@
     "maplibre-gl": "^4.0.0",
     "openapi-typescript-codegen": "^0.29.0",
     "party-js": "^2.2.0",
+    "pdf-parse": "^1.1.1",
```
emersion: This should not be a runtime dep, since it's used only in tests. It should be a dev dep.

front/tests/utils/simulationSheet.ts
```diff
@@ -0,0 +1,194 @@
+import fs from 'fs';
+import path from 'path';
+
+import { expect } from '@playwright/test';
+
+export interface Simulation {
```
Yohh: if exported, maybe should if be in a types.ts file

nit: don´t really care about  it, but do you need it to be an interface? we usually use types

front/tests/utils/simulationSheet.ts
```diff
@@ -0,0 +1,194 @@
...
+        referenceEngine?: string | null;
+        stopType?: string | null;
+      };
+    };
+    disclaimer: string;
+  };
+}
+
+/**
+ * Finds the first PDF file in a directory.
```
Yohh:
```suggestion
 * Find the first PDF file in a directory.
```

front/tests/utils/simulationSheet.ts
```diff
@@ -0,0 +1,194 @@
...
+  try {
+    const pdfFile = fs.readdirSync(directory).find((file) => file.endsWith('.pdf'));
+    return pdfFile ? path.resolve(directory, pdfFile) : null;
+  } catch {
+    return null;
+  }
+}
+
+/**
+ * Verifies the PDF content against the expected simulation.
```
Yohh:
```suggestion
 * Verify the PDF content against the expected simulation.
```

Yohh (approved):

front/tests/utils/simulationSheet.ts
```diff
@@ -0,0 +1,194 @@
...
+ * @returns The absolute path to the PDF file, or `null` if no PDF file is found.
+ */
+export function findFirstPdf(directory: string): string | null {
+  try {
+    const pdfFile = fs.readdirSync(directory).find((file) => file.endsWith('.pdf'));
+    return pdfFile ? path.resolve(directory, pdfFile) : null;
+  } catch {
+    return null;
+  }
+}
```
kmer2016: Currently, the `findFirstPdf` function returns null when no PDF is found, but the error handling lacks details. Consider adding error logs or making exceptions more descriptive to help with debugging.

```typescript
export function findFirstPdf(directory: string): string | null {
  try {
    const pdfFile = fs.readdirSync(directory).find((file) => file.endsWith('.pdf'));
    if (!pdfFile) {
      console.error(`No PDF files found in directory: ${directory}`);
      return null;
    }
    return path.resolve(directory, pdfFile);
  } catch (error) {
    console.error(`Error reading directory: ${directory}`, error);
    return null;
  }
}
```

front/tests/013-stdcm-simulation-sheet.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  test.beforeAll('Fetch infrastructure', async () => {
+    infra = await getInfra();
+  });
+
+  test.beforeEach('Navigate to the STDCM page', async ({ page }) => {
+    // Retrieve OSRD language and navigate to STDCM page
+    const homePage = new HomePage(page);
+    await homePage.goToHomePage();
+    OSRDLanguage = await homePage.getOSRDLanguage();
```
kmer2016, line 50: The `OSRDLanguage` is fetched in every test case, but it seems like this value doesn't change often. Fetching it once in `beforeAll` could improve performance and reduce redundancy. What do you think ?

kmer2016 (commented):
good job ! Small suggestion

front/tests/013-stdcm-simulation-sheet.spec.ts
Maymanaf: I'm working on that in this [PR](https://github.com/OpenRailAssociation/osrd/pull/10423)

front/tests/utils/simulationSheet.ts
Maymanaf: Done

kmer2016 (approved):

Issue #6025, opened 2023-12-06
front: remove unused store variables


nicolaswurtz:
### Description
Some store variable are unused, they should be removed or moved to another reducer.

- [ ] harmonize the sub-reducers (remove the types in `PowerRestrictionsReducer`...)
- [ ] move all the types in the types file (especially `OperationalStudiesConfState`)
- [ ] export the specific selectors or actions of each part of the store to avoid casting `useOsrdConfActions() as StdcmConfSliceActions` or `useOsrdConfSelectors() as StdcmConfSelectors`

Variables to remove:
- [x] osrdsimulation
  - [x] redirectToGraph
  - [x] mustRedraw
  - [x] signalBase
  
Variables to move to another reducer:
- osrdConfCommon -> operationalStudiesConf
  - [x] labels
  - [x] allowances
  - [x] trainCount
  - [x] trainDelta
  - [x] trainStep
  - [x] usingElectricalProfiles
  - [x] powerRestrictionRanges
  - [x] initialSpeed
- osrdConfCommon -> stdcmConf
  - [x] gridMarginBefore 
  - [x] gridMarginAfter
  - [x] originUpperBoundDate
  - [x] originUpperBoundTime
  - [x] originLinkedBounds
  - [x] destinationDate
  - [x] destinationTime

### AC & Implementation plan
- [ ] the store is smaller and cleaner

clarani:
I also add some new tasks related to this refacto:
- harmonize the sub-reducers (remove the types in `PowerRestrictionsReducer`...)
- move all the types in the types file (especially `OperationalStudiesConfState`)
- export the specific selectors or actions of each part of the store to avoid casting `useOsrdConfActions() as StdcmConfSliceActions` or `useOsrdConfSelectors() as StdcmConfSelectors`

PR #10194, opened 2024-12-24
front: create train settings reducer

clarani:
Move some actions from the commonConf to operationalStudiesConf since it is only used in the operational studies application

To merge after https://github.com/OpenRailAssociation/osrd/pull/10202

### DIFF
--- a/front/src/applications/stdcm/utils/formatStdcmConf.ts
+++ b/front/src/applications/stdcm/utils/formatStdcmConf.ts
@@ -2,11 +2,7 @@ import type { TFunction } from 'i18next';
 import { compact } from 'lodash';
 import type { Dispatch } from 'redux';
 
-import type {
-  PathfindingItem,
-  PostTimetableByIdStdcmApiArg,
-  TrainScheduleBase,
-} from 'common/api/osrdEditoastApi';
+import type { PathfindingItem, PostTimetableByIdStdcmApiArg } from 'common/api/osrdEditoastApi';
 import getStepLocation from 'modules/pathfinding/helpers/getStepLocation';
 import type { InfraState } from 'reducers/infra';
 import { setFailure } from 'reducers/main';
@@ -23,7 +19,6 @@ type ValidStdcmConfig = {
   towedRollingStockID?: number;
   timetableId: number;
   infraId: number;
-  rollingStockComfort: TrainScheduleBase['comfort'];
   path: PathfindingItem[];
   speedLimitByTag?: string;
   totalMass?: number;
@@ -46,7 +41,6 @@ export const checkStdcmConf = (
     stdcmPathSteps: pathSteps,
     timetableID,
     speedLimitByTag,
-    rollingStockComfort,
     infraID,
     rollingStockID,
     towedRollingStockID,
@@ -191,7 +185,6 @@ export const checkStdcmConf = (
     infraId: infraID!,
     rollingStockId: rollingStockID!,
     timetableId: timetableID!,
-    rollingStockComfort,
     path,
     speedLimitByTag,
     totalMass,
@@ -216,7 +209,7 @@ export const formatStdcmPayload = (
   infra: validConfig.infraId,
   id: validConfig.timetableId,
   body: {
-    comfort: validConfig.rollingStockComfort || 'STANDARD',
+    comfort: 'STANDARD',
     margin: createMargin(validConfig.margin),
     rolling_stock_id: validConfig.rollingStockId,
     towed_rolling_stock_id: validConfig.towedRollingStockID,
--- a/front/src/reducers/osrdconf/operationalStudiesConf/index.ts
+++ b/front/src/reducers/osrdconf/operationalStudiesConf/index.ts
@@ -1,5 +1,6 @@
 import { createSlice, type Draft, type PayloadAction } from '@reduxjs/toolkit';
 
+import type { Comfort, Distribution } from 'common/api/osrdEditoastApi';
 import type { SuggestedOP } from 'modules/trainschedule/components/ManageTrainSchedule/types';
 import type { TrainScheduleWithDetails } from 'modules/trainschedule/components/Timetable/types';
 import computeBasePathStep from 'modules/trainschedule/helpers/computeBasePathStep';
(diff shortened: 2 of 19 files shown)


achrafmohye (approved):
Seems good to me thanks for the refacto :)

front/src/reducers/osrdconf/operationalStudiesConf/index.ts
```diff
@@ -8,16 +9,43 @@ import type { OsrdConfState } from 'reducers/osrdconf/types';
 import { msToKmh } from 'utils/physics';
 
 import { builPowerRestrictionReducer } from './powerRestrictionReducer';
+import buildTrainSettingsReducer from './trainSettingsReducer';
 import { upsertPathStep } from '../helpers';
 
-export type OperationalStudiesConfState = OsrdConfState;
+export type OperationalStudiesConfState = OsrdConfState & {
```
SharglutDev, line 15: We could move this type in `osrdconf/types.ts`

front/src/reducers/osrdconf/types.ts
```diff
@@ -36,12 +27,9 @@ export type OsrdConfState = InfraState & {
   rollingStockID?: number;
   speedLimitByTag?: string;
   powerRestriction: PowerRestriction[];
```
SharglutDev, line 29: Should we also move this one in `OperationalStudiesConfState` ?

front/src/reducers/osrdconf/types.ts
```diff
@@ -36,12 +27,9 @@ export type OsrdConfState = InfraState & {
...
-  initialSpeed?: number;
   gridMarginBefore?: number;
   gridMarginAfter?: number;
```
SharglutDev, line 31: These should be in `OsrdStdcmConfState`

front/src/reducers/osrdconf/types.ts
```diff
@@ -36,12 +27,9 @@ export type OsrdConfState = InfraState & {
...
   pathSteps: (PathStep | null)[];
```
SharglutDev, line 32: Should be in `OperationalStudiesConfState`

front/src/reducers/osrdconf/operationalStudiesConf/trainSettingsReducer.ts
```diff
@@ -0,0 +1,66 @@
+import type { PayloadAction } from '@reduxjs/toolkit';
+import type { Draft } from 'immer';
+
+import type { OperationalStudiesConfState } from '.';
+
+const buildTrainSettingsReducer = () => ({
```
SharglutDev, line 6: Should we type it to harmonize with `powerRestrictionReducer` ?

front/src/reducers/osrdconf/operationalStudiesConf/selectors.ts
```diff
@@ -1,20 +1,52 @@
...
+    getTrainCount: makeOsrdConfSelector('trainCount'),
+    getTrainDelta: makeOsrdConfSelector('trainDelta'),
+    getTrainStep: makeOsrdConfSelector('trainStep'),
   };
 };
 
 const selectors = buildOperationalStudiesConfSelectors();
 
-export const { getOperationalStudiesConf } = selectors;
+export const {
```
SharglutDev, line 37: We have `useOsrdConfSelectors` where we can get all these selectors. Do we really need to export them here ? Or do you plan to remove the context ?
(same question for actions in `index.ts`)

front/src/applications/stdcm/utils/formatStdcmConf.ts
```diff
@@ -216,7 +209,7 @@ export const formatStdcmPayload = (
   infra: validConfig.infraId,
   id: validConfig.timetableId,
   body: {
-    comfort: validConfig.rollingStockComfort || 'STANDARD',
+    comfort: 'STANDARD',
```
SharglutDev, line 212: If we never want to chose something else than this comfort for stdcm, should the back handle this prop ?

SharglutDev (commented):
Thank you for this clean up. Left some comments (you probably already thought of most of them for another PR 🙂)

front/src/reducers/osrdconf/types.ts
SharglutDev: Oups you took care of that in https://github.com/OpenRailAssociation/osrd/pull/10203 :)

front/src/reducers/osrdconf/operationalStudiesConf/index.ts
clarani: Yes I agree, but let's do this in another PR ?

front/src/reducers/osrdconf/types.ts
clarani: Yes, it's done in this PR : https://github.com/OpenRailAssociation/osrd/pull/10205 :)

front/src/reducers/osrdconf/types.ts
clarani: Same, it will be done in this PR: https://github.com/OpenRailAssociation/osrd/pull/10205

front/src/applications/stdcm/utils/formatStdcmConf.ts
clarani: You're right, we don't have this need now, but since it does not cost much to keep it here, I think we can left the code like this for now :) 
But if you want, we can discuss it with the editoast maintainers

front/src/reducers/osrdconf/operationalStudiesConf/selectors.ts
clarani: The context was created to get the field from the correct part of the store (for instance, if I want to get the selected rollingStock id, we have only 1 selector which can get the data from 2 parts of the store: stdcm or operational studies (it knows in which part of the application it is thanks to the context). Thus, we can store 2 different rolling stocks, for each part of the application).

This does not make sense for selectors which exist only in a part of the application. Also, if I keep using the context, I would have to add cast everywhere like this: `useOsrdConfSelectors() as StdcmConfSelectors`, which is quite fragile :/

front/src/reducers/osrdconf/operationalStudiesConf/trainSettingsReducer.ts
clarani: Yes I agree :+1: But I think we should rather clean the powerRestrictionReducer. I think the type does not bring anything here

clarani:
@SharglutDev  I added a comment about your remarks in this [issue](https://github.com/OpenRailAssociation/osrd/issues/6025), if you want to have a look :wink:

front/src/reducers/osrdconf/operationalStudiesConf/trainSettingsReducer.ts
SharglutDev: Totally agreed. Will you take care of that in #10205 ?

front/src/reducers/osrdconf/operationalStudiesConf/selectors.ts
SharglutDev: Ok I understand

front/src/applications/stdcm/utils/formatStdcmConf.ts
SharglutDev: Yes, it's not hard for us to maintain that :)
Was asking just to know the good practice for that, would be curious to see what they think yes :) (the edge between "it's dead code but it's too much pain to remove it in case we use it in the future" or "we remove it as we don't need it now and don't plane to use it in the future atm")

SharglutDev (approved):
Lgtm and tested, you can close the last remaining comments after reading them :)

front/src/reducers/osrdconf/operationalStudiesConf/trainSettingsReducer.ts
clarani: Nope, but it's reported here https://github.com/OpenRailAssociation/osrd/issues/6025 :)

PR #10195, opened 2024-12-24
front: drop ViaStopDurationSelector

clarani:
Not used anymore since the end of TS V2 migration and the new STDCM view page


achrafmohye (approved):
LGTM and tested 👍

kmer2016 (approved):
LGTM !

Issue #8111, opened 2024-07-18
Implement support for single PK field in ModelV2


leovalais:
### Description and goal

We want the following code to work:

```rust
#[derive(Debug, Clone, ModelV2)]
#[derive(table = XYZ)]
struct MyModel {
     id: i64
}
```

It currently fails because the macro generate the following changeset definition:

```rust
#[derive(Debug, Default, Queryable, AsChangeset, Insertable)]
#[diesel(table_name = XYZ)]
struct MyModelChangeset {}
```

But an empty struct is not supported by diesel's `AsChangeset` and `Insertable`.

Sadly, the current `trait Model` definition **requires** every model to have a changeset type implementing `AsChangeset` (and should also require `Insertable` tbh).

# 🔧 Suggested fix

0. Complete and merge https://github.com/OpenRailAssociation/osrd/pull/7898
1. Relax the bound `AsChangeset` of `trait Model`. We can do this since the queries are generated by the macro and we don't rely on generic type parameters. So this bound is not strictly necessary.
2. Generate and empty changeset struct that doesn't `derive(AsChangeset, Insertable)`.
3. Generate a different implementation for create and update operations if the model that's being expanded contains no non-PK field.
  * `create` and `create_batch` could use `.default_values()` expression method
  * updates don't need to be implemented as PK changes is out of ModelV2's scope
4. If `derive(ModelV2)` is applied to a struct that only contains a PK, the macro can change its `ImplPlan` to avoid generating the `Update*` impl blocks. If those were asked explicitly, an error should be raised.

### Acceptance criteria

```rust
#[derive(Debug, Clone, ModelV2)]
#[derive(table = XYZ)]
struct MyModel {
     id: i64
}
```

compiles and works as expected

PR #10196, opened 2024-12-24
editoast: remove InternalError usage in Delete and Create Model API

leovalais:
* Creates a custom error type wrapping `diesel::result::Error` in `editoast_models`
* Adapt the API of deletion and creation traits (the rest will come in another PR(s))
* Propagate the API change in `views` (mostly)
    * new error variants are created to accommodate some (valid) typing constraints, bringing us closer to our target error management structure
    * simplified a few `map_diesel_error` here and there
* The traits will be moved into `editoast_models::model` later

> [!TIP]
> Review commit by commit

### DIFF
--- a/editoast/Cargo.toml
+++ b/editoast/Cargo.toml
@@ -70,6 +70,7 @@ rangemap = "1.5.1"
 opentelemetry = { version = "0.27.1", default-features = false, features = [
   "trace",
 ] }
+regex = "1.11"
 reqwest = { version = "0.11.27", features = ["json"] }
 rstest = { version = "0.19.0", default-features = false }
 serde = { version = "1.0.217", features = ["derive"] }
@@ -170,7 +171,7 @@ redis = { version = "0.28", default-features = false, features = [
   "tokio-comp",
   "tokio-native-tls-comp",
 ] }
-regex = "1.11.1"
+regex.workspace = true
 reqwest.workspace = true
 serde = { workspace = true, features = ["derive"] }
 serde_json.workspace = true
--- a/editoast/editoast_models/src/model.rs
+++ b/editoast/editoast_models/src/model.rs
@@ -1,0 +1,83 @@
+use std::sync::LazyLock;
+
+use diesel::result::{DatabaseErrorInformation, DatabaseErrorKind};
+use regex::Regex;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("unique constraint violation \"{constraint}\" on column \"{column}\" with value \"{value}\"")]
+    UniqueViolation {
+        constraint: String,
+        column: String,
+        value: String,
+    },
+    #[error("check constraint violation of \"{constraint}\"")]
+    CheckViolation { constraint: String },
+    #[error(transparent)]
+    DatabaseError(diesel::result::Error),
+}
+
+fn try_parse_unique_violation(e: &(dyn DatabaseErrorInformation + Send + Sync)) -> Option<Error> {
+    static RE: LazyLock<Regex> =
+        LazyLock::new(|| Regex::new(r#"duplicate key value violates unique constraint"#).unwrap());
+    if RE.is_match(e.message()) {
+        static RE: LazyLock<Regex> = LazyLock::new(|| {
+            Regex::new(r#"Key \(([^)]+)\)=\(([^)]+)\) already exists\."#).unwrap()
+        });
+        RE.captures(e.details().expect("PostgreSQL should provide details"))
+            .map(|captures| Error::UniqueViolation {
+                constraint: e
+                    .constraint_name()
+                    .expect("PostgreSQL should provide the constraint name")
+                    .to_owned(),
+                column: captures.get(1).unwrap().as_str().to_owned(),
+                value: captures.get(2).unwrap().as_str().to_owned(),
+            })
+    } else {
+        None
+    }
+}
+
+fn try_parse_check_violation(e: &(dyn DatabaseErrorInformation + Send + Sync)) -> Option<Error> {
+    static RE: LazyLock<Regex> = LazyLock::new(|| {
+        Regex::new(r#"new row for relation .* violates check constraint"#).unwrap()
+    });
+    if RE.is_match(e.message()) {
+        Some(Error::CheckViolation {
+            constraint: e
+                .constraint_name()
+                .expect("PostgreSQL should provide the constraint name")
+                .to_owned(),
+        })
+    } else {
+        None
+    }
+}
+
+impl From<diesel::result::Error> for Error {
+    fn from(e: diesel::result::Error) -> Self {
+        match &e {
+            diesel::result::Error::DatabaseError(DatabaseErrorKind::UniqueViolation, inner) => {
+                try_parse_unique_violation(inner.as_ref()).unwrap_or_else(move || {
+                    // falling back to the generic error — since it's still semantically correct, logging the error is enough
+                    tracing::error!(
+                        error = %e,
+                        "failed to parse PostgreSQL details message"
+                    );
+                    Self::DatabaseError(e)
+                })
+            }
+            diesel::result::Error::DatabaseError(DatabaseErrorKind::CheckViolation, inner) => {
+                try_parse_check_violation(inner.as_ref()).unwrap_or_else(|| {
+                    // falling back to the generic error — since it's still semantically correct, logging the error is enough
+                    tracing::error!(
+                        error = %e,
+                        "failed to parse PostgreSQL details message"
+                    );
+                    Self::DatabaseError(e)
+                })
+            }
+            _ => Self::DatabaseError(e),
+        }
+    }
+}
--- a/editoast/src/error.rs
+++ b/editoast/src/error.rs
@@ -265,6 +265,19 @@ impl EditoastError for editoast_schemas::errors::GeometryError {
     }
 }
 
+inventory::submit! {
+    ErrorDefinition::new("editoast:model:ModelError", "", "ModelError", 500u16, r#"{}"#)
+}
+impl EditoastError for editoast_models::model::Error {
+    fn get_status(&self) -> StatusCode {
+        StatusCode::INTERNAL_SERVER_ERROR
+    }
+
+    fn get_type(&self) -> &str {
+        "editoast:ModelError"
+    }
+}
+
 // error definition : uses by the macro EditoastError to generate
 // the list of error and share it with the openAPI generator
 #[derive(Debug)]
--- a/editoast/src/models/prelude/delete.rs
+++ b/editoast/src/models/prelude/delete.rs
@@ -40,24 +40,26 @@ pub trait Delete: Model {
 ///
 /// You can implement this type manually but its recommended to use the `Model`
 /// derive macro instead.
-#[async_trait::async_trait]
-pub trait DeleteStatic<K>: Sized
+pub trait DeleteStatic<K>: Model
 where
-    for<'async_trait> K: Send + 'async_trait,
+    K: Send,
 {
     /// Deletes the row #`id` from the database
-    async fn delete_static(conn: &mut DbConnection, id: K) -> Result<bool>;
+    async fn delete_static(
+        conn: &mut DbConnection,
+        id: K,
+    ) -> std::result::Result<bool, Self::Error>;
 
     /// Just like [DeleteStatic::delete_static] but returns `Err(fail())` if the row didn't exist
-    async fn delete_static_or_fail<E, F>(conn: &mut DbConnection, id: K, fail: F) -> Result<()>
+    async fn delete_static_or_fail<E, F>(conn: &mut DbConnection, id: K, fail: F) -> Result<(), E>
     where
-        E: EditoastError,
-        F: FnOnce() -> E + Send + 'async_trait,
+        E: From<Self::Error>,
+        F: FnOnce() -> E + Send,
     {
         match Self::delete_static(conn, id).await {
             Ok(true) => Ok(()),
-            Ok(false) => Err(fail().into()),
-            Err(e) => Err(e),
+            Ok(false) => Err(fail()),
+            Err(e) => Err(E::from(e)),
         }
     }
 }
--- a/editoast/src/models/prelude/delete.rs
+++ b/editoast/src/models/prelude/delete.rs
@@ -1,11 +1,12 @@
-use editoast_models::DbConnection;
+use std::result::Result;
+
+use editoast_models::{model, DbConnection};
 
 use crate::error::EditoastError;
-use crate::error::Result;
 
 use super::Model;
 
-/// Describes how a [Model](super::Model) can be deleted from the database
+/// Describes how a [Model] can be deleted from the database
 ///
 /// You can implement this type manually but its recommended to use the `Model`
 /// derive macro instead.
@@ -16,11 +17,7 @@ pub trait Delete: Model {
     async fn delete(&self, conn: &mut DbConnection) -> std::result::Result<bool, Self::Error>;
 
     /// Just like [Delete::delete] but returns `Err(fail())` if the row didn't exist
-    async fn delete_or_fail<E, F>(
-        &self,
-        conn: &mut DbConnection,
-        fail: F,
-    ) -> std::result::Result<(), E>
+    async fn delete_or_fail<E, F>(&self, conn: &mut DbConnection, fail: F) -> Result<(), E>
     where
         E: From<Self::Error>,
         F: FnOnce() -> E + Send,
@@ -33,7 +30,7 @@ pub trait Delete: Model {
     }
 }
 
-/// Describes how a [Model](super::Model) can be deleted from the database
+/// Describes how a [Model] can be deleted from the database
 ///
 /// This trait is similar to [Delete] but it doesn't take a reference to the model
 /// instance. This is useful for models that don't have to be retrieved before deletion.
@@ -64,36 +61,39 @@ where
     }
 }
 
-/// Describes how a [Model](super::Model) can be deleted from the database given a batch of keys
+/// Describes how a [Model] can be deleted from the database given a batch of keys
 ///
 /// You can implement this type manually but its recommended to use the `Model`
 /// derive macro instead.
-#[async_trait::async_trait]
-pub trait DeleteBatch<K>: Sized
+pub trait DeleteBatch<K>: Model
 where
-    for<'async_trait> K: Send + 'async_trait,
+    K: Send,
 {
     /// Deletes a batch of rows from the database given an iterator of keys
     ///
     /// Returns the number of rows deleted.
-    async fn delete_batch<I: IntoIterator<Item = K> + Send + 'async_trait>(
+    async fn delete_batch<I: IntoIterator<Item = K> + Send>(
         conn: &mut DbConnection,
         ids: I,
-    ) -> Result<usize>;
+    ) -> Result<usize, Self::Error>;
 
     /// Just like [DeleteBatch::delete_batch] but returns `Err(fail(missing))` where `missing`
     /// is the number of rows that were not deleted
-    async fn delete_batch_or_fail<I, E, F>(conn: &mut DbConnection, ids: I, fail: F) -> Result<()>
+    async fn delete_batch_or_fail<I, E, F>(
+        conn: &mut DbConnection,
+        ids: I,
+        fail: F,
+    ) -> Result<(), E>
     where
-        I: Send + IntoIterator<Item = K> + 'async_trait,
-        E: EditoastError,
-        F: FnOnce(usize) -> E + Send + 'async_trait,
+        I: Send + IntoIterator<Item = K>,
+        E: From<Self::Error>,
+        F: FnOnce(usize) -> E + Send,
     {
         let ids = ids.into_iter().collect::<Vec<_>>();
         let expected_count = ids.len();
         let count = Self::delete_batch(conn, ids).await?;
         if count != expected_count {
-            Err(fail(expected_count - count).into())
+            Err(fail(expected_count - count))
         } else {
             Ok(())
         }
--- a/editoast/src/models/prelude/delete.rs
+++ b/editoast/src/models/prelude/delete.rs
@@ -3,27 +3,32 @@ use editoast_models::DbConnection;
 use crate::error::EditoastError;
 use crate::error::Result;
 
+use super::Model;
+
 /// Describes how a [Model](super::Model) can be deleted from the database
 ///
 /// You can implement this type manually but its recommended to use the `Model`
 /// derive macro instead.
-#[async_trait::async_trait]
-pub trait Delete: Sized {
+pub trait Delete: Model {
     /// Deletes the row corresponding to this model instance
     ///
     /// Returns `true` if the row was deleted, `false` if it didn't exist
-    async fn delete(&self, conn: &mut DbConnection) -> Result<bool>;
+    async fn delete(&self, conn: &mut DbConnection) -> std::result::Result<bool, Self::Error>;
 
     /// Just like [Delete::delete] but returns `Err(fail())` if the row didn't exist
-    async fn delete_or_fail<E, F>(&self, conn: &mut DbConnection, fail: F) -> Result<()>
+    async fn delete_or_fail<E, F>(
+        &self,
+        conn: &mut DbConnection,
+        fail: F,
+    ) -> std::result::Result<(), E>
     where
-        E: EditoastError,
-        F: FnOnce() -> E + Send + 'async_trait,
+        E: From<Self::Error>,
+        F: FnOnce() -> E + Send,
     {
         match self.delete(conn).await {
             Ok(true) => Ok(()),
-            Ok(false) => Err(fail().into()),
-            Err(e) => Err(e),
+            Ok(false) => Err(fail()),
+            Err(e) => Err(E::from(e)),
         }
     }
 }
--- a/editoast/src/models/prelude/mod.rs
+++ b/editoast/src/models/prelude/mod.rs
@@ -31,6 +31,7 @@ pub trait Model: std::fmt::Debug + Clone + Sized + Send {
     type Row: Into<Self> + Send;
     type Changeset: Default + From<Self> + Send;
     type Table: diesel::Table + Send;
+    type Error: std::error::Error + From<editoast_models::model::Error> + Send;
 
     /// Returns an empty changeset for this model
     fn changeset() -> Self::Changeset {
@@ -56,15 +57,19 @@ pub trait Model: std::fmt::Debug + Clone + Sized + Send {
 
 /// A type alias for the [Model::Row] associated type
 ///
-/// Helps silent compiler errors about type amibiguity.
+/// Helps silent compiler errors about type ambiguity.
 #[allow(unused)]
 pub type Row<M> = <M as Model>::Row;
 
 /// A type alias for the [Model::Changeset] associated type
 ///
-/// Helps silent compiler errors about type amibiguity.
-#[allow(unused)]
+/// Helps silent compiler errors about type ambiguity.
 pub type Changeset<M> = <M as Model>::Changeset;
+
+/// A type alias for the [Model::Error] associated type
+///
+/// Helps silent compiler errors about type ambiguity.
+pub type ModelError<M> = <M as Model>::Error;
 
 /// A struct persisting the column and type information of each model field
 ///
--- a/editoast/src/models/timetable.rs
+++ b/editoast/src/models/timetable.rs
@@ -24,6 +24,19 @@ pub struct Timetable {
     pub id: i64,
 }
 
+impl crate::models::Model for Timetable {
+    type Row = Self;
+    type Changeset = Option<i64>;
+    type Table = editoast_models::tables::timetable::table;
+    type Error = editoast_models::model::Error;
+}
+
+impl From<Timetable> for Option<i64> {
+    fn from(timetable: Timetable) -> Self {
+        Some(timetable.id)
+    }
+}
+
 impl Timetable {
     #[tracing::instrument(name = "model:create<Timetable>", skip_all, err)]
     pub async fn create(conn: &mut DbConnection) -> Result<Self> {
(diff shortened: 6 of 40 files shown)


editoast/editoast_models/src/model/error.rs
woshilapin, line 1: How about defining this directly in `src/model.rs`? To me, defining the error is also about locality, meaning, where it will be used. I'm not a fan of defining error in its own module for that reason. On top of that:
- `src/model.rs` is empty so it won't clutter much (we can think later if it's start to not be the case)
- you will avoid a `EditoastModelErrorError` in OpenAPI I believe and transform it into `EditoastModelError` 😃

editoast/src/error.rs
```diff
@@ -265,6 +265,19 @@ impl EditoastError for editoast_schemas::errors::GeometryError {
...
+inventory::submit! {
+    ErrorDefinition::new("editoast:model:ModelError", "Error", "ModelError", 500u16, r#"{}"#)
+}
+impl EditoastError for editoast_models::model::Error {
+    fn get_status(&self) -> StatusCode {
+        StatusCode::INTERNAL_SERVER_ERROR
+    }
+
+    fn get_type(&self) -> &str {
+        "editoast:ModelError"
```
woshilapin, line 277: Do we need the `Error` suffix since this trait is already all about errors anyway? I know we have others like this so for consistency, we could say that's fine... but if I understand well, most of the others may disappear, and therefore, it might be right to challenge that again, isn't it?

editoast/src/models/prelude/delete.rs
```diff
@@ -12,18 +12,25 @@ pub trait Delete: Sized {
...
     /// Returns `true` if the row was deleted, `false` if it didn't exist
-    async fn delete(&self, conn: &mut DbConnection) -> Result<bool>;
+    async fn delete(
+        &self,
+        conn: &mut DbConnection,
+    ) -> std::result::Result<bool, editoast_models::model::Error>;
 
     /// Just like [Delete::delete] but returns `Err(fail())` if the row didn't exist
-    async fn delete_or_fail<E, F>(&self, conn: &mut DbConnection, fail: F) -> Result<()>
+    async fn delete_or_fail<E, F>(
```
woshilapin: Any reason why `delete_or_fail()` is generic over the error `E` but not `delete()` above?

editoast/src/models/timetable.rs
```diff
@@ -62,12 +62,14 @@ impl Timetable {
 impl DeleteStatic<i64> for Timetable {
     #[allow(clippy::blocks_in_conditions)] // TODO: Remove this once using clippy 0.1.80
     #[tracing::instrument(name = "model:delete_static<Timetable>", skip_all, ret, err)]
-    async fn delete_static(conn: &mut DbConnection, id: i64) -> Result<bool> {
-        diesel::delete(dsl::timetable.filter(dsl::id.eq(id)))
+    async fn delete_static(
+        conn: &mut DbConnection,
+        id: i64,
+    ) -> std::result::Result<bool, editoast_models::model::Error> {
```
woshilapin: Every place you change `Result<something>` to `std::result::Result<something, editoast_models::model::Error>`, you can do this instead `Result<something, editoast_models::model::Error>` because our `crate::error::Result` is generic over the error with a default value (so you can just override it).

editoast/src/models/timetable.rs
```diff
@@ -62,12 +62,14 @@ impl Timetable {
 impl DeleteStatic<i64> for Timetable {
     #[allow(clippy::blocks_in_conditions)] // TODO: Remove this once using clippy 0.1.80
```
woshilapin: Does this TODO still holds or can it be removed? (sorry, not part of the PR... but close enough of the diff 😝)

editoast/editoast_models/src/model/error.rs
```diff
@@ -1,11 +1,58 @@
...
+        constraint: String,
+    },
     #[error(transparent)]
     DatabaseError(#[from] crate::DatabaseError),
 }
 
 impl From<diesel::result::Error> for Error {
     fn from(e: diesel::result::Error) -> Self {
-        Self::DatabaseError(e.into())
+        match &e {
```
woshilapin: nit: not sure about my proposition: since all the workflow path ends up being `return                     Self::DatabaseError(e.into())` in case we can't do better, how about expressing this function with a bunch of early return possibilities, and always fallback to the default?

```rust
if let diesel::result::Error::DatabaseError(DatabaseErrorKind::UniqueViolation, inner) = &e {
  /* ... define RE ... */
  if let Some(captures) = RE.captures(something) {
    return Self::UniqueViolation { /* ... */ };
  } else {
    tracing::error!("...");
  }
}
/* same for CheckViolation */
return Self::DatabaseError(e.into());
```

editoast/src/views/temporary_speed_limits.rs
```diff
@@ -111,18 +111,32 @@ enum TemporarySpeedLimitError {
...
+    Database(model::Error),
 }
 
-fn map_diesel_error(e: InternalError, name: impl AsRef<str>) -> InternalError {
-    if e.message.contains(
-        r#"duplicate key value violates unique constraint "temporary_speed_limit_group_name_key""#,
-    ) {
-        TemporarySpeedLimitError::NameAlreadyUsed {
-            name: name.as_ref().to_string(),
+impl From<model::Error> for TemporarySpeedLimitError {
```
woshilapin: Will we have to write all `From` manually from now on? Or is it just a temporary situation until we finish the migration to the new error model (I feel I'm missing a part of the big picture here). If it's the former, then we might to discuss it because I'm not sure we would want that everywhere, what do you think?

editoast/src/views/temporary_speed_limits.rs
```diff
@@ -158,7 +172,7 @@ async fn create_temporary_speed_limit_group(
         .creation_date(Utc::now().naive_utc())
         .create(conn)
         .await
-        .map_err(|e| map_diesel_error(e, speed_limit_group_name))?;
+        .map_err(|e| TemporarySpeedLimitError::from(e).with_name(speed_limit_group_name))?;
```
woshilapin: Maybe implementing `From` is not adapted here. We might need a real `new()` constructor? What do you think? Or else, we might think about `UniqueViolation` containing the value of the conflicting key?

editoast/src/views/rolling_stock.rs
```diff
@@ -189,6 +192,35 @@ pub fn map_diesel_error(e: InternalError, name: impl AsRef<str>) -> InternalErro
...
+            {
+                Self::BasePowerClassEmpty
+            }
+            e => Self::Database(e),
+        }
+    }
+}
+
+impl RollingStockError {
+    fn with_name(self, name: String) -> Self {
```
woshilapin: You might be able to avoid this weird function/pattern if `model::Error::UniqueViolation` is able to retain the key that created the conflict.

editoast/editoast_models/src/model/error.rs
```diff
@@ -0,0 +1,58 @@
+use std::sync::LazyLock;
+
+use diesel::result::DatabaseErrorKind;
+use regex::Regex;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
woshilapin: Do you think we would need a `CreateError` and a `DeleteError`? From what I understand, they semantically can produce different set of error variants... but `diesel` seems to have made the choice to only have one `enum` for them all (even if different functions may only return a subset of it).

woshilapin (commented):
Nice job. Not approving yet, let's see where the comments goes first.

editoast/editoast_models/src/model/error.rs
leovalais: d5e8f29a2

Note: this is the only instruction that defines what's being included in the OpenAPI. The module path is only used by `derive(EditoastError)` to fill in `fn get_type`.

```rust
inventory::submit! {
    ErrorDefinition::new("editoast:model:ModelError", "Error", "ModelError", 500u16, r#"{}"#)
}
```

editoast/src/error.rs
leovalais: This defines the value of the `error_type` field of the response, so as you said consistency and all.

I've `impl EditoastError or model::Error` just as a hack to avoid changing all the error propagation of each view using Create and Delete traits. But this implementation is bound to disappear as we convert the last view errors to `trait ViewError`. 

I don't mind the `Error` suffix but I admit it's redundant as we'll be able to tell it's an error via the status code of the response. Would it be ok to challenge that when the first `ViewError` lands?

Also I removed the `-ErrorError`: 6852ad732

editoast/src/models/prelude/delete.rs
leovalais: The point of `delete` is to delete the row if it's there (hence the `Ok(bool)`). The "fail" or `delete_or_fail` represents the case where we expect the row to be there. In case it's not, it's a caller error that needs to be raised, not a `model::Error`:

```rust
        match self.delete(conn).await {
            Ok(true) => Ok(()),
            Ok(false) => Err(fail()), // fail(): E
            Err(e) => Err(E::from(e)), // E: From<model::Error>
        }
``` 

But PostgreSQL errors can still happen, so we need `E` to "accept" `model::Error` as well.

editoast/src/models/timetable.rs
leovalais: 2d7acaab2

editoast/src/models/timetable.rs
leovalais: 2d7acaab2

editoast/src/views/temporary_speed_limits.rs
leovalais: Only when we can't write:

```rust
#[error(transparent)]
#[editoast_error(status = 500)]
Database(#[from] model::Error),
```

because we want the `from` operation to perform some decisions about which variant to return, like here. When it's just a forwarding, `thiserror` generates this implementation.

editoast/editoast_models/src/model/error.rs
leovalais: Imperative programming, my favorite style... 🙃

Done in 1ed77d0b6, not sold on the proposition, but I may be biased tbh, lmk :)

editoast/src/views/temporary_speed_limits.rs
leovalais: Thanks for making me look again the information provided by the error. I believe we can parse the conflicting key and add it to the error.

There may be a way to interface the error with `trait Identifiable<K>` and `trait PreferredIdentifier<K>`. I'll look into it.

But before that can we resolve https://github.com/OpenRailAssociation/osrd/pull/10196#discussion_r1905511645 (conflicts...)

```rust
pub trait DatabaseErrorInformation {
    /// The primary human-readable error message. Typically one line.
    fn message(&self) -> &str;

    /// An optional secondary error message providing more details about the
    /// problem, if it was provided by the database. Might span multiple lines.
    fn details(&self) -> Option<&str>;

    /// An optional suggestion of what to do about the problem, if one was
    /// provided by the database.
    fn hint(&self) -> Option<&str>;

    /// The name of the table the error was associated with, if the error was
    /// associated with a specific table and the backend supports retrieving
    /// that information.
    ///
    /// Currently this method will return `None` for all backends other than
    /// PostgreSQL.
    fn table_name(&self) -> Option<&str>;

    /// The name of the column the error was associated with, if the error was
    /// associated with a specific column and the backend supports retrieving
    /// that information.
    ///
    /// Currently this method will return `None` for all backends other than
    /// PostgreSQL.
    fn column_name(&self) -> Option<&str>;

    /// The constraint that was violated if this error is a constraint violation
    /// and the backend supports retrieving that information.
    ///
    /// Currently this method will return `None` for all backends other than
    /// PostgreSQL.
    fn constraint_name(&self) -> Option<&str>;

    /// An optional integer indicating an error cursor position as an index into
    /// the original statement string.
    fn statement_position(&self) -> Option<i32>;
}
```

editoast/src/views/rolling_stock.rs
leovalais: https://github.com/OpenRailAssociation/osrd/pull/10196#discussion_r1905522292

editoast/editoast_models/src/model/error.rs
leovalais: https://github.com/OpenRailAssociation/osrd/pull/10196#discussion_r1905522292

editoast/editoast_models/src/model/error.rs
woshilapin: OK, so the [commit](https://github.com/OpenRailAssociation/osrd/commit/d5e8f29a253aedb5da03ca2806634f14cba45247) does solve the complicate naming in OpenAPI. But yet, do we need a dedicated Rust module for the error?

editoast/src/error.rs
woshilapin: I missed the fact that this implementation too will also disappear. So more than fine enough for consistency.

editoast/src/models/timetable.rs
woshilapin: Haha, well, you changed this only place, but there was a lot of places in your PR where you can do that.

editoast/src/models/timetable.rs
woshilapin: Thank you ❤️

editoast/editoast_models/src/model/error.rs
woshilapin: Oh, I was not entirely sure either. I just saw in the previous implementation, multiple possible outcomes of the function that were all doing the `Self::DatabaseError(e.into());`: they are all about "if I can't do anything, let's fallback to that".

My proposition is more about presenting the function with one `return` for each semantic outcome (`UniqueViolation`, `'CheckViolation` and fallback). But I could live with any versions of the implementation, so feel free to keep or drop the commit. I'm resolving anyway.

editoast/editoast_models/src/model/error.rs
woshilapin: I'm not sure that I understand how is this related to https://github.com/OpenRailAssociation/osrd/pull/10196#discussion_r1905522292.

editoast/editoast_models/src/model/error.rs
leovalais: * https://github.com/OpenRailAssociation/osrd/pull/10196/commits/6852ad73256e650ac8253f36ee3f9ff90d55f0ca fixes the OpenApi
* https://github.com/OpenRailAssociation/osrd/commit/d5e8f29a253aedb5da03ca2806634f14cba45247 moves the error to `model.rs`

I should have linked both, sorry 😅

editoast/src/models/timetable.rs
leovalais: Oh I thought your comment was about this one specifically since the traits are implemented manually for timetables. I'll change the occurrences in the trait functions but for the quoted code in the macros I'd prefer to keep the fully qualified path:

1. macro hygiene
2. it lifts ambiguity since "jump to definition" isn't available in snapshots or quoted code
3. `crate::error::Result` will disappear after the `ViewError` migration, so that's one less thing we'll have to do in the future

Wdyt?

editoast/editoast_models/src/model/error.rs
leovalais: It doesn't, sorry 😅 

I asked myself the same question. I don't think there will be much difference between each trait's error type. `Retrieve` may have less error cases, but I don't think that'd be worth dealing with 5+ error types for the `Model` API alone. I could be proved wrong after having migrated the rest of the traits, but at least for `Create` and `Delete`, I believe that holds. Wait and see, or do we split them now and merge them if that's deemed unnecessary later?

editoast/editoast_models/src/model/error.rs
leovalais: I've dropped the commit, but added a comment in a15b78c33 hopefully making things more readable.

editoast/src/views/temporary_speed_limits.rs
leovalais: Here's a proposition: 6257f64dd

I like it better this way, but I wish we didn't have to bother knowing table columns and constraints using models in views. Table definitions are "implementation details" from the views point of view. But that's out of scope of this PR (maybe ModelV3? :p)

editoast/editoast_models/src/model/error.rs
leovalais: Should be better in 6257f64dd

editoast/src/models/timetable.rs
woshilapin: I disagree about point 3 (yes it will disappear, but then you wouldn't need to change anything because if `crate::error::Result` doesn't exist anymore, prelude of Rust will fall back on `std::result::Result`).

BUT, 1. macro hygiene is a good reason by itself to be explicit to a maximum. I didn't have that in mind.

editoast/editoast_models/src/model.rs
```diff
@@ -0,0 +1,97 @@
...
+fn try_parse_unique_violation(
+    e: &Box<dyn DatabaseErrorInformation + Send + Sync>,
+) -> Option<Error> {
+    static RE: LazyLock<Regex> =
+        LazyLock::new(|| Regex::new(r#"duplicate key value violates unique constraint"#).unwrap());
+    if RE.is_match(e.message()) {
+        static RE: LazyLock<Regex> = LazyLock::new(|| {
+            Regex::new(r#"Key \(([^)]+)\)=\(([^)]+)\) already exists\."#).unwrap()
+        });
+        if let Some(captures) = RE.captures(e.details().expect("PostgreSQL should provide details"))
```
woshilapin: No need for panic, especially since [this might actually return nothing](https://docs.rs/diesel/latest/diesel/result/trait.DatabaseErrorInformation.html#tymethod.details).

```suggestion
        if let Some(captures) = RE.captures(e.details().unwrap_or("Database did not provided details"))
```

editoast/editoast_models/src/model.rs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            Some(Error::UniqueViolation {
+                constraint: e
+                    .constraint_name()
+                    .expect("PostgreSQL should provide the constraint name")
```
woshilapin, line 31: Why panic in this function when `Option<Error>` is returned anyway. We're trying to handle an error path, wouldn't it be better to not panic while already trying to handle a failure (panic might actually hide the primary reason of the failure) and instead:
1. log why we can't handle this failure
2. let the default behavior kick back (returning `None` will actually end-up returning `Self::DatabaseError(e.into())`

> [!WARNING]
> There is multiple panic point in this function, I'm talking about all of them

editoast/src/views/temporary_speed_limits.rs
woshilapin: I also do like it much more, but agree with you. That said, I believe it's only because `UniqueViolation` has not the right API. We're exposing errors to the users of `UniqueViolation` (in this case, `views::*`) and we already simplified the DB technical error by naming it `UniqueViolation` or `CheckViolation`... therefore, why keeping technical details like `constraint` or `column`. So here is the suggestion:

```rust
#[derive(Debug)]
pub enum UniqueKind {
    Id,
    Name,
}

#[derive(Debug)]
pub enum CheckKind {
    BasePowerClassEmpty,
}

pub enum Error {
    #[error("unique constraint violation of \"{kind}\" with value \"{value}\"")]
    UniqueViolation {
        kind: UniqueKind,
        value: String,
    },
    #[error("check constraint violation of \"{kind}\"")]
    CheckViolation { kind: CheckKind },
   /* ... */
}
```

or even a step further with


```rust
#[derive(Debug, thiserror::Error)]
pub enum UniqueViolation {
    #[error("unique constraint violation on the identifier \"{.0}\"")]
    Id(i64),
    #[error("unique constraint violation on the identifier \"{.0}\"")]
    Name(String),
}
#[derive(Debug, thiserror::Error)]
pub enum CheckViolation {
    #[error("base power class is empty")]
    BasePowerClassEmpty,
}
pub enum Error {
    #[error(transparent)]
    UniqueViolation(UniqueViolation),
    #[error(transparent)]
    CheckViolation(CheckViolation),
   /* ... */
}
```


What do you think?

editoast/editoast_models/src/model/error.rs
woshilapin: OK, let's keep it simple. It will also help bring other developers in it (if the API is too dense right from the beginning, it will be a friction to adoption).

editoast/src/views/temporary_speed_limits.rs
woshilapin: Yes thinking more about this, I believe this is part of being explicit in some cases where we want/need to be. So fine indeed.

editoast/editoast_models/src/model.rs
leovalais: According to what I understand from the error interface of `diesel` and my experiments with pg, the information is *systematically there* for *some* error kinds, but since they're all unified in a single interface, `Option`s are necessary.

From that, having matched on the correct error variant, not having the expected data as part of the error is actually a bug on our end (in the "error computing algorithm" so to speak). I'd argue we *want* things to `panic` early as it helps pinpoint the error easily. (Instead of having `None` coerced to another error now structurally valid, but invalid conceptually, that's being propagated and transformed. Debugging that won't be fun, especially since a plain `panic` basically tells us "precondition not met at file.rs:line".)

However I do agree on the inconsistency between the leniency of the `error!("failed to match regexp")` and the panicking. But now that I'm thinking about this, I'd be more inclined towards also panicking if the regexps do not match. 


Wdyt?

editoast/src/views/temporary_speed_limits.rs
leovalais: You're right about the interface. However these enums are specific to the RollingStock model so we need a way to provide the macro this information. I wanted to avoid making too much changes to the macro, but here's my proposition if that's ok:

```rust
#[derive(Model)]
#[model(error = Error)]
struct RollingStock {..}

// your error type
pub enum Error {
    #[error(transparent)]
    UniqueViolation(UniqueViolation),
    #[error(transparent)]
    CheckViolation(CheckViolation),
   /* ... */
}
```

We need to add the error as an associated type to `trait Model` as well:

```rust
pub trait Model: std::fmt::Debug + Clone + Sized + Send {
    type Row: Into<Self> + Send;
    type Changeset: Default + From<Self> + Send;
    type Table: diesel::Table + Send;

	// new: we keep the generic model::Error as the default to avoid
	// having too much to deal with in the generated code, but we add
	// a `.map_err(Model::Error::from)` after each query to allow custom errors
	type Error: From<model::Error> = model::Error
	..
}
```

then:

```rust
// rolling_stock_model.rs

impl From<model::Error> for Error {
	fn from(...) -> Self {
		// the conversion of the low-level error
		// model::Error::UniqueViolation { constraint, column, value } into
		// a structured error designed for the RollingStockModel
	}
}
```

So basically:

```
PG error --> diesel error --> model::Error --> custom Model error
```

editoast/editoast_models/src/model.rs
woshilapin: It feels like a prioritization problem. On one hand, we prioritize the consumer experience by always returning something, giving some information about a problem of the database, even in cases where we fail to characterize this database error more precisely. On the other hand, we prioritize the developer experience but highlighting bugs as soon as possible, leaving the consumer with a function call that never really finish. Do you feel that is a correct statement of the situation? If yes, I do not have an opinion on which one to prioritize (my opinion being solely based on eradicating panics everywhere 😈).

leovalais: Yep, seems like a good summary. Note that the function does finish in case of a panic and the endpoint returns a 5xx. Maybe a third opinion on the matter would help (odd numbers, @flomonster :eyes:)? I'd be fine with both ways.

editoast/src/views/temporary_speed_limits.rs
woshilapin: As discussed, let's go for the macro 😄

editoast/editoast_models/src/model.rs
flomonster: I don't see how the consumer experience is enhanced with your proposal @woshilapin. In both cases, the user's request will fail without any possibility of recovery. That's why I prioritize the “panic”.

Note that I'm not sure how panic behaves in editoast.

- Do we send back a 500?
- Is it logged (and therefore recoverable in datadog with an alert)?
- Is there any configuration work required to log panic events?

woshilapin: Panic does return a 500 (I believe `axum` catches it).

In this case, I'm really not sure we should return a 5xx. We're handling a `UniqueViolation` or a `CheckViolation` which means, it's a client error (said differently, the client should be able to fix its request with the information contained in the error: "oh, I see, the object already exists"). To me, that's a sign of a 4xx error. `panic` (so 500) will return the wrong information to the client telling him he can't do anything about it.

Here, we already know it's a `UniqueViolation` (we checked for the regexp). And this function is only trying to populate correctly the `UniqueViolation`... but we already know we're in a state where the client tried to insert a duplicate object so we should tell it with a 4xx. That's why I say it's a more consumer driven prioritization.

That said, the documentation from Diesel also states that `.constraint_name()` should always return `Some()` in a Postgresql environment. So `panic` might make sense since it might be a sign of a regression somewhere. More of a developer driven prioritization.

That's why I had this belief of "it might depend on what we want to prioritize?".

leovalais: > Here, we already know it's a UniqueViolation (we checked for the regexp).

We actually know that even before checking the regexp in `try_parse_unique_violation`:

```rust
impl From<diesel::result::Error> for Error {
    fn from(e: diesel::result::Error) -> Self {
        match &e {
            diesel::result::Error::DatabaseError(DatabaseErrorKind::UniqueViolation, inner) => {
                try_parse_unique_violation(inner).unwrap_or_else(|| {
```

Diesel detected that it's a unique violation already. If the regexp can't match, we in one of these cases:

1. our regexp is wrong
2. postgres is inconsistent with its error messages

.1. may happen, 2. is much more unlikely, but in any case it's one of our (we developers) assumptions that's wrong. Knowing from diesel that it's a unique violation, we literally "expect" the regexp to match.

I do agree the helpfulness of the error originating from a panic though. But maybe it's a feature? Do we really want database details to leak at the API level? Intuitively no, but maybe for convenience?

---

To remove the panics I'd suggest introducing a new variant `model::Error::UnexpectedDatabaseError(Box<dyn Error>)` to tell apart "I forward this error" and "I should have been able to parse this error, but failed". I'll go with that and drop the commit if we decide `panic`s are fine.

woshilapin: > I do agree the helpfulness of the error originating from a panic though. But maybe it's a feature? Do we really want database details to leak at the API level? Intuitively no, but maybe for convenience?

Agreed with that. Just find it a shame to not return some 4xx to the client, helping him to fix the problem. Let's keep the panic, it's the less work, we've probably spend too much time on it (although, it's always fun to challenge and try to dig a bit in order to build a common understanding of the situation). This situation will (should) probably never happen in practice.

I'll resolve.

editoast/Cargo.toml
```diff
@@ -65,6 +65,7 @@ postgres-openssl = "0.5.0"
 pretty_assertions = "1.4.1"
 rand = "0.8.5"
 rangemap = "1.5.1"
+regex = "*"
```
woshilapin: Ooops.
```suggestion
regex = "1.11"
```

editoast/src/models/prelude/mod.rs
```diff
@@ -63,9 +64,13 @@ pub type Row<M> = <M as Model>::Row;
 /// A type alias for the [Model::Changeset] associated type
 ///
 /// Helps silent compiler errors about type amibiguity.
-#[allow(unused)]
 pub type Changeset<M> = <M as Model>::Changeset;
 
+/// A type alias for the [Model::Error] associated type
+///
+/// Helps silent compiler errors about type amibiguity.
```
woshilapin:
```suggestion
/// Helps silent compiler errors about type ambiguity.
```

Hmm, and by the way, I believe this is copy/paste from above so it might need to be corrected above too.

editoast/src/models/timetable.rs
```diff
@@ -24,6 +24,19 @@ pub struct Timetable {
     pub id: i64,
 }
 
+impl crate::models::Model for Timetable {
+    type Row = Self;
+    type Changeset = Option<i64>;
+    type Table = editoast_models::tables::timetable::table;
+    type Error = editoast_models::model::Error;
+}
```
woshilapin, line 32: I do not understand why this piece of code appears in the diff? Didn't we already have an implementation of `Model`? Can't we use `derive(Model)`? (if no, do you think we need a comment to explain why we can't use `derive`?)

woshilapin (approved):
I think I like this `Model::Error` associated type, seems logical. I've put a few more comments related to the recent changes, but nothing big. Thank you.

leovalais:
Thanks for the review! Good to know it's the right track, it's still a work in progress though (I think one commit got nuked in the process I just noticed... 👀). I'll try to finish that on monday (if I'm not trapped on permissions again, that is 😨)

editoast/src/models/timetable.rs
leovalais: We can't because `Timetable` only has a primary key and the macro doesn't support that. It *almost* can though: https://github.com/OpenRailAssociation/osrd/issues/8111

woshilapin: But `Model` was not implemented on `Timetable` before this PR?

leovalais: `trait Model` wasn't. It wasn't used much except for listing, which is not supported for timetables. But now that we require `Delete: Model` to get `Self::Error`, we need this implementation.

leovalais: Note: the traits are implemented manually because the macro wants to generate a changeset type, which is undefined for objects with just an auto incremented ID.

SharglutDev (approved):
Lgtm for front (not tested)

editoast/src/models/timetable.rs
woshilapin: ok, ok, so this new `impl` was imposed by the change of architecture, so it really is a new `impl` explaining why we see it in the diff. I'm resolving.

editoast/Cargo.toml
leovalais: 600bd2ca8

editoast/src/models/prelude/mod.rs
leovalais: 600bd2ca8

editoast/Cargo.toml
leovalais: [`e999b7e` (#10196)](https://github.com/OpenRailAssociation/osrd/pull/10196/commits/e999b7eccc7291ee9bb20fe03e11c403421dfb6c)

leovalais:
@woshilapin I've added the `Error` associated type of `trait Model` and used it in `RollingStock`, `TemporarySpeedLimit` and `WorkSchedule`.

I also removed some `#[async_trait]`s that are not useful for these definitions anymore :-)

woshilapin (approved):

PR #10197, opened 2024-12-26
editoast: add configurable authorization to `TestAppBuilder`

hamz2a:
Update `TestAppBuilder` to support configurable authorization settings

- Modified `TestAppBuilder` to allow enabling/disabling authorization via the `enable_authorization` field.
- Added a test to verify that project creation fails with a FORBIDDEN status when authorization is enabled.

### DIFF
--- a/editoast/src/views/projects.rs
+++ b/editoast/src/views/projects.rs
@@ -359,15 +359,21 @@ async fn patch(
 #[cfg(test)]
 pub mod tests {
 
+    use std::collections::HashSet;
+
     use axum::http::StatusCode;
+    use editoast_authz::authorizer::UserInfo;
     use pretty_assertions::assert_eq;
     use rstest::rstest;
     use serde_json::json;
 
     use super::*;
+    use crate::core::mocking::MockingClient;
+    use crate::core::CoreClient;
     use crate::models::fixtures::create_project;
     use crate::models::prelude::*;
     use crate::views::test_app::TestAppBuilder;
+    use crate::views::test_app::TestRequestExt;
 
     #[rstest]
     async fn project_post() {
@@ -392,6 +398,32 @@ pub mod tests {
             .expect("Project not found");
 
         assert_eq!(project.name, project_name);
+    }
+
+    #[rstest]
+    async fn project_post_should_fail_when_authorization_is_enabled() {
+        let pool = DbConnectionPoolV2::for_tests();
+        let user = UserInfo {
+            identity: "user_identity".to_string(),
+            name: "user_name".to_string(),
+        };
+        let app = TestAppBuilder::new()
+            .db_pool(pool)
+            .core_client(CoreClient::Mocked(MockingClient::default()))
+            .enable_authorization(true)
+            .user(user.clone())
+            .roles(HashSet::from([BuiltinRole::OpsRead]))
+            .build();
+
+        let request = app.post("/projects").by_user(user).json(&json!({
+            "name": "test_project_failed",
+            "description": "",
+            "objectives": "",
+            "funders": "",
+        }));
+
+        // OpsWrite is required to complete this request successfully.
+        app.fetch(request).assert_status(StatusCode::FORBIDDEN);
     }
 
     #[rstest]
--- a/editoast/src/views/test_app.rs
+++ b/editoast/src/views/test_app.rs
@@ -2,17 +2,22 @@
 //! test actix server, database connection pool, and different mocking
 //! components.
 
+use std::collections::HashSet;
 use std::sync::Arc;
 
 use axum::Router;
 use axum_tracing_opentelemetry::middleware::OtelAxumLayer;
 use dashmap::DashMap;
+use editoast_authz::authorizer::StorageDriver;
+use editoast_authz::authorizer::UserInfo;
+use editoast_authz::BuiltinRole;
 use editoast_common::tracing::create_tracing_subscriber;
 use editoast_common::tracing::Stream;
 use editoast_common::tracing::Telemetry;
 use editoast_common::tracing::TracingConfig;
 use editoast_models::DbConnectionPoolV2;
 use editoast_osrdyne_client::OsrdyneClient;
+use futures::executor::block_on;
 use futures::future::BoxFuture;
 use opentelemetry_sdk::export::trace::ExportResult;
 use opentelemetry_sdk::export::trace::SpanData;
@@ -26,6 +31,7 @@ use crate::{
     generated_data::speed_limit_tags_config::SpeedLimitTagIds,
     infra_cache::InfraCache,
     map::MapLayers,
+    models::auth::PgAuthDriver,
     valkey_utils::ValkeyConfig,
     AppState, ValkeyClient,
 };
@@ -56,6 +62,9 @@ pub(crate) struct TestAppBuilder {
     db_pool: Option<DbConnectionPoolV2>,
     core_client: Option<CoreClient>,
     osrdyne_client: Option<OsrdyneClient>,
+    enable_authorization: bool,
+    user: Option<UserInfo>,
+    roles: HashSet<BuiltinRole>,
 }
 
 impl TestAppBuilder {
@@ -64,6 +73,9 @@ impl TestAppBuilder {
             db_pool: None,
             core_client: None,
             osrdyne_client: None,
+            enable_authorization: false,
+            user: None,
+            roles: HashSet::new(),
         }
     }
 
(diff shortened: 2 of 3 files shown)


editoast/src/views/projects.rs
```diff
@@ -394,6 +396,27 @@ pub mod tests {
...
+        let project_name = "test_project_failed";
+
+        let request = app.post("/projects").json(&json!({
+            "name": project_name,
+            "description": "",
+            "objectives": "",
+            "funders": "",
+        }));
+
+        app.fetch(request).assert_status(StatusCode::FORBIDDEN);
```
leovalais, line 426: Should be `401 Unauthorized` as the headers are missing.

editoast/src/views/test_app.rs
```diff
@@ -39,6 +39,7 @@ pub(crate) struct TestAppBuilder {
     db_pool: Option<DbConnectionPoolV2>,
     core_client: Option<CoreClient>,
     osrdyne_client: Option<OsrdyneClient>,
+    disable_authorization: bool,
```
```suggestion
    enable_authorization: bool,
```

I'd reverse the flag as we don't want to bother with authorization for most tests. But for those where we do, I think this reads more easily.

editoast/src/views/test_app.rs
```diff
@@ -152,7 +159,7 @@ impl TestAppBuilder {
             infra_caches,
             map_layers: Arc::new(MapLayers::default()),
             speed_limit_tag_ids,
-            disable_authorization: true,
+            disable_authorization: self.disable_authorization,
```
leovalais: This field should have been removed from the `AppState` as it's redundant with the value in `config`. Could we remove it in this PR, it's only used in a few places? 🙏

leovalais (commented):
This PR revealed a bug. When no authentication header is sent to editoast, we respond with a 403 instead of a 401. Its fix belongs to another PR, but I don't believe we should be testing a faulty behavior. Here's my proposition:

1. Add a function `TestApp::def_user(&self, UserInfo, impl IntoIterator<Item = BuiltinRole>)` that persists a user configuration for a test.
2. Add a convenience function `TestAppBuilder::user(self, UserInfo, impl IntoIterator<Item = BuiltinRole>)`
3. Add a function like `app.post("...").from_user(UserInfo { ... }).json(...` to add the headers to the request.
4. Change the test behavior to "make a request with a real authenticated user which doesn't have the required role", where responding 403 is a valid response.

Wdyt?

editoast/src/views/test_app.rs

editoast/src/views/test_app.rs

editoast/src/views/projects.rs
```diff
@@ -394,6 +400,33 @@ pub mod tests {
         assert_eq!(project.name, project_name);
     }
 
+    #[rstest]
+    async fn project_post_should_fail_when_authorization_is_enabled() {
+        let pool = DbConnectionPoolV2::for_tests();
+        let app = TestAppBuilder::new()
+            .db_pool(pool)
+            .core_client(CoreClient::Mocked(MockingClient::default()))
+            .enable_authorization(true)
```
leovalais, line 413:
```suggestion
            .enable_authorization(true)
            .user(&user, &[BuiltinRole::OpsRead])
```

a `fn user(self: TestAppBuilder, &userInfo, impl IntoIterator<Item = BuiltinRole>)` calling `def_user` after building would be convenient.

editoast/src/views/test_app.rs
```diff
@@ -213,18 +224,45 @@ impl TestApp {
...
+
     pub fn patch(&self, path: &str) -> TestRequest {
         self.server.patch(&trim_path(path))
     }
+
     pub fn delete(&self, path: &str) -> TestRequest {
         self.server.delete(&trim_path(path))
     }
+
+    pub async fn def_user(&self, user: &UserInfo, roles: HashSet<BuiltinRole>) {
```
```suggestion
    pub fn def_user(&self, user: &UserInfo, roles: impl IntoIterator<Item = BuiltinRole>) {
```

Let's try to make that sync with `futures::executor::block_on`. Saves us an .await in tests.

editoast/src/views/test_app.rs
```diff
@@ -213,18 +224,45 @@ impl TestApp {
...
+            .await
+            .expect("User should be created successfully");
+        driver
+            .ensure_subject_roles(uid, roles)
+            .await
+            .expect("Roles should be updated successfully");
+    }
+}
+
+pub trait WithUser {
```
```suggestion
pub trait TestRequestExt {
```

editoast/src/views/test_app.rs
```diff
@@ -213,18 +224,45 @@ impl TestApp {
...
+    fn with_user(self, user: UserInfo) -> Self;
```
```suggestion
    fn from_user(self, user: UserInfo) -> Self;
```

A request comes from a user.

leovalais (commented):
Thanks for the modifications! It looks much better, a few comments about API readability and convenience.

editoast/src/views/test_app.rs

editoast/src/views/test_app.rs
hamz2a: I used this name, because I got the following error :
```bash
warning: methods called from_* usually take no self
   --> src/views/test_app.rs:258:18
    |
258 |     fn from_user(self, user: UserInfo) -> Self;
    |                  ^^^^
    |
    = help: consider choosing a less ambiguous name
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention
    = note: #[warn(clippy::wrong_self_convention)] on by default
```

editoast/src/views/test_app.rs

editoast/src/views/projects.rs
hamz2a: I made some changes to the API. Could you take a look and let me know what you think?

editoast/src/views/test_app.rs
leovalais: 😑😑😑😑😑😑😑😑😑😑😑😑

Okay, so what about `by_user`, that should make clippy happy...

leovalais (approved):
LGTM, thanks!

editoast/src/views/test_app.rs
```diff
@@ -56,6 +62,9 @@ pub(crate) struct TestAppBuilder {
     db_pool: Option<DbConnectionPoolV2>,
     core_client: Option<CoreClient>,
     osrdyne_client: Option<OsrdyneClient>,
+    enable_authorization: bool,
```
woshilapin, line 65: I find it a bit weird that sometimes it's called `disable_authorization` (in `ServerConfig`), and sometimes `enable_authorization`. Is there a good reason? In general, I would prefer the non-negative version (hence `enable_authorization`, `disable` is negating the authorization). For example, if `disable_authorization` is `false`, then it's almost like a double negation ("don't disable the authorization" instead of "enable the authorization").

leovalais: Agreed about the double negation. I'd prefer to have `enable_` everywhere as well, except in the command line where authorization will probably be enabled by default eventually, in which case `editoast runserver --disable-authorization` reads more easily than `editoast runserver --enable-authorization=false`. WdyT?

woshilapin: In any case, I'd be more attached to have `disable_authorization` everywhere (or `enable_authorization` everywhere) than a mix of the two. And double negation is less intuitive, but I agree that in this case, since the default behavior is with authorization, maybe keeping `disable_authorization` everywhere is fine? (unless you have an idea to have `enable_authorization` everywhere, and provide a `--disable-authorization` on the CLI ?)

leovalais: Yep. My point was to have `enable_authorization` everywhere _except_ in the CLI. Sorry if that was unclear. (I'm fine with having `--enable-authorization` as well though, I'm not that attached to the CLI exception.)

hamz2a: I see there’s a discussion about using `disable_authorization` vs `enable_authorization`. For this PR, should I stick to one convention, or would you prefer to handle this differently?

And if we decide to change it, should we do that in this PR or in a future one to keep this PR focused?

Thanks! @woshilapin @leovalais

leovalais: Okay on second thought I do agree with @woshilapin about consistency even in the CLI. `enable-authorization` would be my preference. I'm fine with keeping things as they are for now in order to let you continue working on the stdcm log endpoints. It's not a large change and it can be done rather quickly anyways. Maybe after the other PR is merged?

woshilapin: From what I understand, @leovalais (and me) tends to prefer `enable_authorization` everywhere, and make the CLI accept the option `--disable-authorization` (which would then fill up a field `enable_authorization` with `false`). Not sure how easy is the last bit about CLI 🤔 If it's not, fine also for `--enable-authorization false` in the CLI. Is that a good summary for you @hamz2a?

hamz2a: Sure, I’ll make the change in a separate PR to keep the commit history clean. Thank you!

hamz2a: https://github.com/OpenRailAssociation/osrd/pull/10509

PR #10200, opened 2024-12-27
core, editoast, python: stop train at end of block instead of OP

louisgreiner:
Closes https://github.com/osrd-project/osrd-confidential/issues/892
Closes https://github.com/osrd-project/osrd-confidential/issues/830

Analysis of the solution, compared to the default behavior can be seen [here](https://github.com/osrd-project/osrd-dags/pull/655)

## Description

Note : the following screenshots are for review visualization only, nothing will be output in the front-end at the end

The trains were previously stopping exactly on the Operational Point Part position, which is often in the middle of the train platform. The position of the train is considered at the head of the train. This is an issue, because in the real life, a train (meaning its head) stops at the end of the platform (at a specific stop panel position), allowing the rest of the train not to overflow of the passed tracks, avoiding unnecessary track occupation. Also, this doesn't take account of the train direction.

The train stops are now shifted the least possible to the next signal position, with a minimal value that would put the tail of the train on the operational point.

This implementation makes the pathfinding algorithm move the stops of these trains closer to the end of the current block, to try to avoid that the tail of the train is on another track, and also never further than the next block-delimiting signal.
⚠️ the first and last stops are not moved ⚠️

- The feature is ✅ hard-enabled for STDCM (for the simulated train), and there is no deactivation possible.
- The feature is ✖️ hard-disabled in front for Operational Studies (trains created from this module cannot enable the feature)
  - however the feature is 🔲 activable in back (disabled by default), so imported trains that have the feature enabled (such as BASIC timetables) can be looked at in Operational Studies, or used in STDCM search environment.
- The feature may (☑️ will) be enabled for all the trains simulated in the STDCM search environment depending on what was parametrized when importing data.

#### With a `TGV2N2`, 400m train
image

#### With a `83500`, 109m train
image

#### Also working on STDCM, with a 200m train
image

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlockRequest.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlockRequest.kt
@@ -18,6 +18,7 @@ class PathfindingBlockRequest(
     @Json(name = "rolling_stock_maximum_speed") val rollingStockMaximumSpeed: Double,
     @Json(name = "rolling_stock_length") val rollingStockLength: Double,
     @Json(name = "speed_limit_tag") val speedLimitTag: String? = null,
+    @Json(name = "stops_at_end_of_block") val stopsAtEndOfBlock: Boolean? = null,
     val timeout: Double?,
     val infra: String,
     @Json(name = "expected_version") val expectedVersion: Int?,
--- a/core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
@@ -28,6 +28,7 @@ import fr.sncf.osrd.reporting.exceptions.ErrorType
 import fr.sncf.osrd.reporting.exceptions.OSRDError
 import fr.sncf.osrd.sim_infra.api.*
 import fr.sncf.osrd.utils.*
+import fr.sncf.osrd.utils.units.Distance.Companion.min
 import fr.sncf.osrd.utils.units.Length
 import fr.sncf.osrd.utils.units.Offset
 import fr.sncf.osrd.utils.units.meters
@@ -100,10 +101,32 @@ class PathfindingBlocksEndpoint(private val infraManager: InfraProvider) : Take 
 fun runPathfinding(infra: FullInfra, request: PathfindingBlockRequest): PathfindingBlockResponse {
     // Parse the waypoints
     val waypoints = ArrayList<Collection<EdgeLocation<BlockId, Block>>>()
-    for (step in request.pathItems) {
+    val destinationTrack = request.pathItems.last()
+    val destinationBlock = findWaypointBlocks(infra, destinationTrack)
+    request.pathItems.forEachIndexed { stepIndex, step ->
         val allStarts = HashSet<EdgeLocation<BlockId, Block>>()
         for (direction in Direction.entries) {
-            for (waypoint in step) allStarts.addAll(findWaypointBlocks(infra, waypoint, direction))
+            for (waypoint in step) {
+                val waypointBlocks = findDirectedWaypointBlocks(infra, waypoint, direction)
+                if (
+                    request.stopsAtEndOfBlock == true &&
+                        stepIndex != 0 &&
+                        stepIndex != request.pathItems.size - 1
+                ) {
+                    allStarts.addAll(
+                        waypointBlocks.map {
+                            findStopPositionAtEndOfBlockConsideringRollingStock(
+                                it,
+                                destinationBlock,
+                                request.rollingStockLength,
+                                infra,
+                            )
+                        }
+                    )
+                } else {
+                    allStarts.addAll(waypointBlocks)
+                }
+            }
         }
         waypoints.add(allStarts)
     }
@@ -278,7 +301,7 @@ private fun processPathfindingResponse(
  * @param waypoint corresponding waypoint.
  * @return corresponding edge location, containing a block id and its offset from the waypoint.
  */
-fun findWaypointBlocks(
+fun findDirectedWaypointBlocks(
     infra: FullInfra,
     waypoint: TrackLocation,
     direction: Direction,
@@ -305,6 +328,19 @@ fun findWaypointBlocks(
         res.add(EdgeLocation(block, offset))
     }
     return res
+}
+
+fun findWaypointBlocks(
+    infra: FullInfra,
+    waypoints: Collection<TrackLocation>,
+): Set<EdgeLocation<BlockId, Block>> {
+    val waypointBlocks = HashSet<EdgeLocation<BlockId, Block>>()
+    for (waypoint in waypoints) {
+        for (direction in Direction.entries) {
+            waypointBlocks.addAll(findDirectedWaypointBlocks(infra, waypoint, direction))
+        }
+    }
+    return waypointBlocks
 }
 
 private fun getTrackSectionChunkOnWaypoint(
@@ -398,4 +434,37 @@ private fun makeHeuristicsForPathfindingEdges(
         }
     }
     return remainingDistanceEstimators
+}
+
+/**
+ * Given a waypoint block (initial position of a train stop), returns the new stop position of a
+ * train: end of current block, with the constraints of staying in the same block and keeping the
+ * tail of the train on the initial position (waypointBlock). Note that nodes of the infra are not
+ * considered.
+ */
+fun findStopPositionAtEndOfBlockConsideringRollingStock(
+    waypointBlock: EdgeLocation<BlockId, Block>,
+    destinationBlock: Set<EdgeLocation<BlockId, Block>>,
+    rollingStockLength: Double,
+    infra: FullInfra,
+): EdgeLocation<BlockId, Block> {
+    // To ensure that the tail of the rolling stock is not further than the initial operational
+    // point (waypointBlock) position
+    val maxTailOffset = waypointBlock.offset + rollingStockLength.meters
+
+    // Can't go further than the end of the block, which is delimited by a block-delimiting signal
+    val maxHeadOffset = infra.blockInfra.getBlockLength(waypointBlock.edge)
+
+    val newWaypointOffset = Offset.min(maxTailOffset, maxHeadOffset)
+
+    val destinationOffset =
+        destinationBlock
+            .filter { it.edge == waypointBlock.edge && waypointBlock.offset <= it.offset }
+            .minOfOrNull { it.offset }
+
+    // The waypoint offset does not change in the case where the stop is after the destination
+    if (destinationOffset != null && destinationOffset <= newWaypointOffset)
+        return EdgeLocation(waypointBlock.edge, waypointBlock.offset)
+
+    return EdgeLocation(waypointBlock.edge, newWaypointOffset)
 }
--- a/core/src/main/kotlin/fr/sncf/osrd/api/stdcm/STDCMEndpoint.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/api/stdcm/STDCMEndpoint.kt
@@ -4,6 +4,7 @@ import com.google.common.collect.ImmutableRangeMap
 import com.google.common.collect.Range
 import com.google.common.collect.TreeRangeSet
 import fr.sncf.osrd.api.*
+import fr.sncf.osrd.api.pathfinding.findStopPositionAtEndOfBlockConsideringRollingStock
 import fr.sncf.osrd.api.pathfinding.findWaypointBlocks
 import fr.sncf.osrd.api.pathfinding.hasDuplicateTracks
 import fr.sncf.osrd.api.pathfinding.runPathfindingBlockPostProcessing
@@ -21,15 +22,12 @@ import fr.sncf.osrd.envelope_sim.allowances.AllowanceValue.Percentage
 import fr.sncf.osrd.envelope_sim.allowances.AllowanceValue.TimePerDistance
 import fr.sncf.osrd.envelope_sim_infra.computeMRSP
 import fr.sncf.osrd.pathfinding.Pathfinding
-import fr.sncf.osrd.pathfinding.Pathfinding.EdgeLocation
 import fr.sncf.osrd.railjson.schema.common.graph.EdgeDirection
 import fr.sncf.osrd.railjson.schema.rollingstock.Comfort
 import fr.sncf.osrd.railjson.schema.schedule.RJSTrainStop
 import fr.sncf.osrd.reporting.exceptions.ErrorType
 import fr.sncf.osrd.reporting.exceptions.OSRDError
 import fr.sncf.osrd.signaling.etcs_level2.ETCS_LEVEL2
-import fr.sncf.osrd.sim_infra.api.Block
-import fr.sncf.osrd.sim_infra.api.BlockId
 import fr.sncf.osrd.sim_infra.api.DirTrackChunkId
 import fr.sncf.osrd.sim_infra.api.SpeedLimitProperty
 import fr.sncf.osrd.sim_infra.api.TrackSectionId
@@ -114,7 +112,7 @@ class STDCMEndpoint(
                         it != ETCS_LEVEL2.id
                     },
                 )
-            val steps = parseSteps(infra, request.pathItems, request.startTime)
+            val steps = parseSteps(infra, request.pathItems, request.startTime, rollingStock.length)
             val requirements = getRequirements(request, infra, timetableCacheManager)
             val allowedTrackSections = parseTrackSectionIds(infra, request.allowedTrackSections)
 
@@ -306,6 +304,7 @@ fun parseSteps(
     infra: FullInfra,
     pathItems: List<STDCMPathItem>,
     startTime: ZonedDateTime,
+    rollingStockLength: Double,
 ): List<STDCMStep> {
     if (pathItems.last().stopDuration == null) {
         throw OSRDError(ErrorType.MissingLastSTDCMStop)
@@ -324,20 +323,34 @@ fun parseSteps(
     // that it's not a stop.
     pathItems.first().stopDuration = null
 
-    return pathItems.map {
-        STDCMStep(
-            findWaypointBlocks(infra, it.locations),
-            it.stopDuration?.seconds,
-            it.stopDuration != null,
-            if (it.stepTimingData != null)
-                PlannedTimingData(
-                    TimeDelta(between(startTime, it.stepTimingData.arrivalTime).toMillis()),
-                    it.stepTimingData.arrivalTimeToleranceBefore,
-                    it.stepTimingData.arrivalTimeToleranceAfter,
-                )
-            else null,
-        )
-    }
+    return pathItems
+        .mapIndexed { index, it ->
+            STDCMStep(
+                if (index != 0 && index != pathItems.size - 1) {
+                    val destinationBlock = findWaypointBlocks(infra, pathItems.last().locations)
+                    findWaypointBlocks(infra, it.locations).map { waypointBlock ->
+                        findStopPositionAtEndOfBlockConsideringRollingStock(
+                            waypointBlock,
+                            destinationBlock,
+                            rollingStockLength,
+                            infra,
+                        )
+                    }
+                } else {
+                    findWaypointBlocks(infra, it.locations)
+                },
+                it.stopDuration?.seconds,
+                it.stopDuration != null,
+                if (it.stepTimingData != null)
+                    PlannedTimingData(
+                        TimeDelta(between(startTime, it.stepTimingData.arrivalTime).toMillis()),
+                        it.stepTimingData.arrivalTimeToleranceBefore,
+                        it.stepTimingData.arrivalTimeToleranceAfter,
+                    )
+                else null,
+            )
+        }
+        .toList()
 }
 
 /**
@@ -414,19 +427,6 @@ private fun parseSimulationScheduleItems(
             SimulationScheduleItem(Offset(it.position.meters), null, duration, it.receptionSignal)
         }
     )
-}
-
-private fun findWaypointBlocks(
-    infra: FullInfra,
-    waypoints: Collection<TrackLocation>,
-): Set<EdgeLocation<BlockId, Block>> {
-    val waypointBlocks = HashSet<EdgeLocation<BlockId, Block>>()
-    for (waypoint in waypoints) {
-        for (direction in Direction.entries) {
-            waypointBlocks.addAll(findWaypointBlocks(infra, waypoint, direction))
-        }
-    }
-    return waypointBlocks
 }
 
 fun parseTrackSectionIds(infra: FullInfra, trackSectionName: Set<String>?): Set<TrackSectionId>? {
--- a/core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingTest.kt
+++ b/core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingTest.kt
@@ -20,6 +20,7 @@ import kotlin.test.assertEquals
 import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.api.Assertions.assertThatThrownBy
 import org.assertj.core.api.AssertionsForClassTypes
+import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.TestInstance
 
@@ -27,6 +28,7 @@ fun getPathfindingBlockRequest(
     rs: RollingStock,
     pathItems: List<Collection<TrackLocation>>,
     infra: String = "unused_name",
+    stopsAtEndOfBlock: Boolean? = false,
 ): PathfindingBlockRequest {
     return PathfindingBlockRequest(
         rs.loadingGaugeType,
@@ -36,6 +38,7 @@ fun getPathfindingBlockRequest(
         rs.maxSpeed,
         rs.length,
         null,
+        stopsAtEndOfBlock,
         null,
         infra,
         1,
@@ -60,6 +63,7 @@ fun checkPathfindingSuccess(
             .plus(expectedIntermediatePathItemPosition)
             .plusElement(Offset(pathSuccess.length.distance))
     assertEquals(expectedPathItemsPos, pathSuccess.pathItemPositions)
+    assertEquals(pathSuccess.pathItemPositions.sorted(), pathSuccess.pathItemPositions)
 
     if (expectedBlocks != null) {
         assertEquals(
@@ -600,5 +604,320 @@ class PathfindingTest : ApiTest() {
         val response = PathfindingBlocksEndpoint(infraManager).act(RqFake(requestBody))
         val parsed = pathfindingResponseAdapter.fromJson(response.body())!!
         return parsed
+    }
+}
+
+class PathfindingV2StopsAtEndOfBlock : ApiTest() {
+    fun callPathfindingEndpoint(
+        rs: RollingStock,
+        pathItems: List<Collection<TrackLocation>>,
+        infra: String,
+        stopsAtEndOfBlock: Boolean,
+    ): PathfindingBlockResponse {
+        val requestBody =
+            pathfindingRequestAdapter.toJson(
+                getPathfindingBlockRequest(rs, pathItems, infra, stopsAtEndOfBlock)
+            )
+        val response = PathfindingBlocksEndpoint(infraManager).act(RqFake(requestBody))
+        val parsed = pathfindingResponseAdapter.fromJson(response.body())!!
+        return parsed
+    }
+
+    private lateinit var waypoints: List<List<TrackLocation>>
+    private val firstIntermediateStopDistance = 12050.meters
+    private val secondIntermediateStopDistance = 26500.meters
+    private val reversedFirstIntermediateStopDistance = 19400.meters
+    private val reversedSecondIntermediateStopDistance = 33850.meters
+
+    @BeforeEach
+    override fun setUp() {
+        super.setUp()
+        // West_Station
+        val startWaypoint = TrackLocation("TA1", Offset(500.meters))
+        // Mid_West_Station
+        val firstIntermediateWaypoint = TrackLocation("TC1", Offset(550.meters))
+        // Mid_East_Station
+        val secondIntermediateWaypoint = TrackLocation("TD0", Offset(14000.meters))
+        // South_East_Station
+        val endWaypoint = TrackLocation("TH1", Offset(4400.meters))
+        waypoints =
+            listOf(
+                listOf(startWaypoint),
+                listOf(firstIntermediateWaypoint),
+                listOf(secondIntermediateWaypoint),
+                listOf(endWaypoint),
+            )
+    }
+
+    @Test
+    fun nonStopsAtEndOfBlockTest() {
+        // Stops are not moved - nothing to do here
+
+        val parsed =
+            callPathfindingEndpoint(
+                TestTrains.REALISTIC_FAST_TRAIN,
+                waypoints,
+                "small_infra/infra.json",
+                false,
+            )
+        checkPathfindingSuccess(
+            parsed,
+            45900.meters,
+            expectedIntermediatePathItemPosition =
+                listOf(
+                    Offset(firstIntermediateStopDistance),
+                    Offset(secondIntermediateStopDistance),
+                ),
+        )
+    }
+
+    @Test
+    fun reversedNonStopsAtEndOfBlockTest() {
+        // Stops are not moved - nothing to do here
+
+        val parsed =
+            callPathfindingEndpoint(
+                TestTrains.REALISTIC_FAST_TRAIN,
+                waypoints.reversed(),
+                "small_infra/infra.json",
+                false,
+            )
+        checkPathfindingSuccess(
+            parsed,
+            45900.meters,
+            expectedIntermediatePathItemPosition =
+                listOf(
+                    Offset(reversedFirstIntermediateStopDistance),
+                    Offset(reversedSecondIntermediateStopDistance),
+                ),
+        )
+    }
+
+    @Test
+    fun shortTrainTest() {
+        // Only intermediate stops are moved to the end of their block
+        val recalculatedFirstIntermediateStopDistance =
+            firstIntermediateStopDistance + TestTrains.VERY_SHORT_FAST_TRAIN.length.meters
+        val recalculatedSecondIntermediateStopDistance =
+            secondIntermediateStopDistance + TestTrains.VERY_SHORT_FAST_TRAIN.length.meters
+
+        val parsed =
+            callPathfindingEndpoint(
+                TestTrains.VERY_SHORT_FAST_TRAIN,
+                waypoints,
+                "small_infra/infra.json",
+                true,
+            )
+        checkPathfindingSuccess(
+            parsed,
+            45900.meters,
+            expectedIntermediatePathItemPosition =
+                listOf(
+                    Offset(recalculatedFirstIntermediateStopDistance),
+                    Offset(recalculatedSecondIntermediateStopDistance),
+                ),
+        )
+    }
+
+    @Test
+    fun reversedShortTrainTest() {
+        // Only intermediate stops are moved to the end of their block
+        val recalculatedFirstIntermediateStopDistance =
+            reversedFirstIntermediateStopDistance + TestTrains.VERY_SHORT_FAST_TRAIN.length.meters
+        val recalculatedSecondIntermediateStopDistance =
+            reversedSecondIntermediateStopDistance + TestTrains.VERY_SHORT_FAST_TRAIN.length.meters
+
+        val parsed =
+            callPathfindingEndpoint(
+                TestTrains.VERY_SHORT_FAST_TRAIN,
+                waypoints.reversed(),
+                "small_infra/infra.json",
+                true,
+            )
+        checkPathfindingSuccess(
+            parsed,
+            45900.meters,
+            expectedIntermediatePathItemPosition =
+                listOf(
+                    Offset(recalculatedFirstIntermediateStopDistance),
+                    Offset(recalculatedSecondIntermediateStopDistance),
+                ),
+        )
+    }
+
+    @Test
+    fun longTrainTest() {
+        // The first intermediate stop is moved to the next block-delimiting signal, but the
+        // distance available is less than the length of the train, so it is moved to the end of the
+        // block (270 meters after the first intermediate stop)
+        val recalculatedFirstIntermediateStopDistance = firstIntermediateStopDistance + 270.meters
+        // The second intermediate stop is also moved to the end of the block (at the "DD0_9"
+        // detector's position, 37.5 meters after the second intermediate stop)
+        val recalculatedSecondIntermediateStopDistance =
+            secondIntermediateStopDistance + 37.5.meters
+
+        val parsed =
+            callPathfindingEndpoint(
+                TestTrains.VERY_LONG_FAST_TRAIN,
+                waypoints,
+                "small_infra/infra.json",
+                true,
+            )
+        checkPathfindingSuccess(
+            parsed,
+            45900.meters,
+            expectedIntermediatePathItemPosition =
+                listOf(
+                    Offset(recalculatedFirstIntermediateStopDistance),
+                    Offset(recalculatedSecondIntermediateStopDistance),
+                ),
+        )
+    }
+
+    @Test
+    fun reversedLongTrainTest() {
+        // The first intermediate stop is moved to the next block-delimiting signal, but the
+        // distance available is less than the length of the train, so it is moved to the end of the
+        // block, the 9th block of the path (1500 meters after the first intermediate stop)
+        val recalculatedFirstIntermediateStopDistance =
+            reversedFirstIntermediateStopDistance + 1500.meters
+        // The second intermediate stop is also moved to the end of the block, the 13th block of the
+        // path (370 meters after the second intermediate stop)
+        val recalculatedSecondIntermediateStopDistance =
+            reversedSecondIntermediateStopDistance + 370.meters
+
+        val parsed =
+            callPathfindingEndpoint(
+                TestTrains.VERY_LONG_FAST_TRAIN,
+                waypoints.reversed(),
+                "small_infra/infra.json",
+                true,
+            )
+        checkPathfindingSuccess(
+            parsed,
+            45900.meters,
+            expectedIntermediatePathItemPosition =
+                listOf(
+                    Offset(recalculatedFirstIntermediateStopDistance),
+                    Offset(recalculatedSecondIntermediateStopDistance),
+                ),
+        )
+    }
+
+    @Test
+    fun penultimateCorrectlyMovedJustBeforeDestination() {
+        var waypoints: List<List<TrackLocation>>
+        val startWaypoint = TrackLocation("TC1", Offset(500.meters))
+        val intermediateWaypoint = TrackLocation("TD0", Offset(12550.meters))
+        val endWaypoint = TrackLocation("TD0", Offset(13000.meters))
+        waypoints = listOf(listOf(startWaypoint), listOf(intermediateWaypoint), listOf(endWaypoint))
+
+        val intermediateStopDistance = 13450.meters
+
+        val parsed =
+            callPathfindingEndpoint(
+                TestTrains.REALISTIC_FAST_TRAIN,
+                waypoints,
+                "small_infra/infra.json",
+                true,
+            )
+        checkPathfindingSuccess(
+            parsed,
+            13500.meters,
+            expectedIntermediatePathItemPosition = listOf(Offset(intermediateStopDistance)),
+        )
+    }
+
+    @Test
+    fun penultimateStopNotMovedExactlyToDestination() {
+        var waypoints: List<List<TrackLocation>>
+        val startWaypoint = TrackLocation("TC1", Offset(500.meters))
+        val intermediateWaypoint = TrackLocation("TD0", Offset(12600.meters))
+        val endWaypoint = TrackLocation("TD0", Offset(13000.meters))
+        waypoints = listOf(listOf(startWaypoint), listOf(intermediateWaypoint), listOf(endWaypoint))
+
+        // The intermediate stop is moved to the next block-delimiting signal, but the
+        // distance available is more than the length of the train, so it would be moved by the
+        // length of the train (400m) but this is also exactly the destination so the intermediate
+        // stop remains the same
+        val intermediateStopDistance = 13100.meters
+
+        val parsed =
+            callPathfindingEndpoint(
+                TestTrains.REALISTIC_FAST_TRAIN,
+                waypoints,
+                "small_infra/infra.json",
+                true,
+            )
+        checkPathfindingSuccess(
+            parsed,
+            13500.meters,
+            expectedIntermediatePathItemPosition = listOf(Offset(intermediateStopDistance)),
+        )
+    }
+
+    @Test
+    fun penultimateStopTooCloseToDestination() {
+        var waypoints: List<List<TrackLocation>>
+        val startWaypoint = TrackLocation("TC1", Offset(500.meters))
+        val intermediateWaypoint = TrackLocation("TD0", Offset(12990.meters))
+        val endWaypoint = TrackLocation("TD0", Offset(13000.meters))
+        waypoints = listOf(listOf(startWaypoint), listOf(intermediateWaypoint), listOf(endWaypoint))
+
+        // The intermediate stop is moved to the next block-delimiting signal, but the
+        // distance available is more than the length of the train, so it would be moved by the
+        // length of the train, but that is after the destination so the intermediate stop remains
+        // the same
+        val intermediateStopDistance = 13490.meters
+
+        val parsed =
+            callPathfindingEndpoint(
+                TestTrains.REALISTIC_FAST_TRAIN,
+                waypoints,
+                "small_infra/infra.json",
+                true,
+            )
+        checkPathfindingSuccess(
+            parsed,
+            13500.meters,
+            expectedIntermediatePathItemPosition = listOf(Offset(intermediateStopDistance)),
+        )
+    }
+
+    @Test
+    fun twoPenultimatesStopTooCloseToDestination() {
+        var waypoints: List<List<TrackLocation>>
+        val startWaypoint = TrackLocation("TC1", Offset(500.meters))
+        val intermediateWaypoint = TrackLocation("TD0", Offset(12950.meters))
+        val secondIntermediateWaypoint = TrackLocation("TD0", Offset(12990.meters))
+        val endWaypoint = TrackLocation("TD0", Offset(13000.meters))
+        waypoints =
+            listOf(
+                listOf(startWaypoint),
+                listOf(intermediateWaypoint),
+                listOf(secondIntermediateWaypoint),
+                listOf(endWaypoint),
+            )
+
+        // The intermediate stops are moved to the next block-delimiting signal, but the
+        // distance available is more than the length of the train, so it would be moved by the
+        // length of the train, but that is after the destination so the intermediate stops remains
+        // the same
+        val intermediateStopDistance = 13450.meters
+        val secondIntermediateStopDistance = 13490.meters
+
+        val parsed =
+            callPathfindingEndpoint(
+                TestTrains.REALISTIC_FAST_TRAIN,
+                waypoints,
+                "small_infra/infra.json",
+                true,
+            )
+        checkPathfindingSuccess(
+            parsed,
+            13500.meters,
+            expectedIntermediatePathItemPosition =
+                listOf(Offset(intermediateStopDistance), Offset(secondIntermediateStopDistance)),
+        )
     }
 }
--- a/editoast/openapi.yaml
+++ b/editoast/openapi.yaml
@@ -10953,6 +10953,13 @@ components:
           - string
           - 'null'
           description: Speed limit tag, used to estimate the travel time
+        stops_at_end_of_block:
+          type:
+          - boolean
+          - 'null'
+          description: |-
+            Stop the train at the next block-delimiting signal,
+            staying in the same block and keeping the tail on the initial position
     PathfindingItem:
       type: object
       required:
@@ -14209,6 +14216,8 @@ components:
     TrainScheduleOptions:
       type: object
       properties:
+        stops_at_end_of_block:
+          type: boolean
         use_electrical_profiles:
           type: boolean
         use_speed_limits_for_simulation:
--- a/editoast/src/views/path/pathfinding.rs
+++ b/editoast/src/views/path/pathfinding.rs
@@ -65,6 +65,9 @@ pub(in crate::views) struct PathfindingInput {
     rolling_stock_length: OrderedFloat<f64>,
     /// Speed limit tag, used to estimate the travel time
     speed_limit_tag: Option<String>,
+    /// Stop the train at the next block-delimiting signal,
+    /// staying in the same block and keeping the tail on the initial position
+    stops_at_end_of_block: Option<bool>,
 }
 
 impl PathfindingInput {
@@ -389,6 +392,7 @@ fn build_pathfinding_request(
         rolling_stock_maximum_speed: pathfinding_input.rolling_stock_maximum_speed,
         rolling_stock_length: pathfinding_input.rolling_stock_length,
         speed_limit_tag: pathfinding_input.speed_limit_tag.clone(),
+        stops_at_end_of_block: pathfinding_input.stops_at_end_of_block,
     })
 }
 
@@ -479,6 +483,7 @@ pub async fn pathfinding_from_train_batch<T: TrainScheduleLike>(
                 .map(|item| item.location.clone())
                 .collect(),
             speed_limit_tag: train_schedule.speed_limit_tag().cloned(),
+            stops_at_end_of_block: Some(train_schedule.options().stops_at_end_of_block()),
         };
         to_compute.push(path_input);
         to_compute_index.push(index);
--- a/front/src/common/api/generatedEditoastApi.ts
+++ b/front/src/common/api/generatedEditoastApi.ts
@@ -3586,6 +3586,9 @@ export type PathfindingInput = {
   rolling_stock_supported_signaling_systems: string[];
   /** Speed limit tag, used to estimate the travel time */
   speed_limit_tag?: string | null;
+  /** Stop the train at the next block-delimiting signal,
+    staying in the same block and keeping the tail on the initial position */
+  stops_at_end_of_block?: boolean | null;
 };
 export type RoutePath = {
   switches_directions: {
@@ -3933,6 +3936,7 @@ export type Margins = {
   values: string[];
 };
 export type TrainScheduleOptions = {
+  stops_at_end_of_block?: boolean;
   use_electrical_profiles?: boolean;
   use_speed_limits_for_simulation?: boolean;
 };
--- a/front/tests/assets/stdcm/linked-train/anterior-linked-train-table.json
+++ b/front/tests/assets/stdcm/linked-train/anterior-linked-train-table.json
@@ -15,7 +15,7 @@
     "operationalPoint": "North_station",
     "code": "BV",
     "track": "V1bis",
-    "endStop": "13:26",
+    "endStop": "13:25",
     "passageStop": "4 min",
     "startStop": "13:30",
     "weight": "=",
--- a/python/osrd_schemas/osrd_schemas/train_schedule.py
+++ b/python/osrd_schemas/osrd_schemas/train_schedule.py
@@ -201,6 +201,8 @@ class TrainScheduleOptions(BaseModel):
     - `use_electrical_profiles` : use the electrical profiles for the standalone simulation
     - `use_speed_limits_for_simulation` : use the speed limits coming from the infrastructure for
        the standalone simulation
+    - `stops_at_end_of_block` : stop the train at the next block-delimiting signal instead of on the waypoint
+
     """
 
     use_electrical_profiles: bool = Field(
@@ -210,6 +212,11 @@ class TrainScheduleOptions(BaseModel):
     use_speed_limits_for_simulation: bool = Field(
         default=True,
         description="If false, the speed limits of the infrastructure are ignored in the standalone simulation",
+    )
+    stops_at_end_of_block: bool = Field(
+        default=False,
+        description="If true, stop positions will be shifted toward the next block-delimiting signal, \
+        staying in the same block and keeping the tail on the initial position",
     )
 
 
--- a/core/src/main/java/fr/sncf/osrd/cli/BenchSTDCM.kt
+++ b/core/src/main/java/fr/sncf/osrd/cli/BenchSTDCM.kt
@@ -156,7 +156,8 @@ class BenchSTDCM : CliCommand {
                                 it != ETCS_LEVEL2.id
                             },
                         )
-                    val steps = parseSteps(infra, request.pathItems, request.startTime)
+                    val steps =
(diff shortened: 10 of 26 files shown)


editoast/editoast_schemas/src/train_schedule/train_schedule_options.rs
```diff
@@ -14,8 +14,22 @@ pub struct TrainScheduleOptions {
     #[derivative(Default(value = "true"))]
     #[serde(default = "default_use_electrical_profiles")]
     use_electrical_profiles: bool,
+
+    #[derivative(Default(value = "true"))]
```
louisgreiner, line 18:
```suggestion
    #[derivative(Default(value = "false"))]
```

We keep this flag to false for review purpose, but before merging, we need to set it to `false`

editoast/editoast_schemas/src/train_schedule/train_schedule_options.rs
```diff
@@ -14,8 +14,22 @@ pub struct TrainScheduleOptions {
...
+    stop_at_next_signal: bool,
 }
 
 fn default_use_electrical_profiles() -> bool {
     true
 }
+
+fn default_stop_at_next_signal() -> bool {
+    true
+}
```
louisgreiner, line 29:
```suggestion
```

We keep this flag to `true` for review purpose, but before merging, we need to set it to `false`

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -149,6 +149,7 @@ const useSetupItineraryForTrainUpdate = (trainIdToEdit: number) => {
           rolling_stock_supported_signaling_systems: rollingStock.supported_signaling_systems,
           rolling_stock_maximum_speed: rollingStock.max_speed,
           rolling_stock_length: rollingStock.length,
+          stop_at_next_signal: true, // TODO: change to false and set to true elsewhere if needed
```
louisgreiner, line 152:
```suggestion
          stop_at_next_signal: false,
```

We keep this flag to `true` for review purpose, but before merging, we need to set it to `false`

front/src/modules/pathfinding/utils.ts
```diff
@@ -79,6 +79,7 @@ export const getPathfindingQuery = ({
         rolling_stock_supported_signaling_systems: rollingStock.supported_signaling_systems,
         rolling_stock_maximum_speed: rollingStock.max_speed,
         rolling_stock_length: rollingStock.length,
+        stop_at_next_signal: true, // TODO: change to false and set to true elsewhere if needed
```
louisgreiner, line 82:
```suggestion
        stop_at_next_signal: false,
```

We keep this flag to `true` for review purpose, but before merging, we need to set it to `false`

editoast/editoast_schemas/src/train_schedule/train_schedule_options.rs
```diff
@@ -14,8 +14,22 @@ pub struct TrainScheduleOptions {
     #[derivative(Default(value = "true"))]
     #[serde(default = "default_use_electrical_profiles")]
     use_electrical_profiles: bool,
+
+    #[derivative(Default(value = "true"))]
+    #[serde(default = "default_stop_at_next_signal")]
```
louisgreiner:
```suggestion
    #[derivative(Default(value = "false"))]
```

editoast/editoast_schemas/src/train_schedule/train_schedule_options.rs
louisgreiner:
```suggestion
```

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
louisgreiner:
```suggestion
          stop_at_next_signal: false,
```

front/src/modules/pathfinding/utils.ts
louisgreiner:
```suggestion
        stop_at_next_signal: false,
```

louisgreiner (commented):
We keep this flag to `true` for review purpose, but before merging, we need to set it to `false`

louisgreiner:
Questions : 
- Currently it moves all the waypoints of the train schedules instead of only the stops ; is it a real problem ?

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -369,3 +384,37 @@ private fun getBlockOffset(
...
+    rollingStockLength: Double
+): Offset<Block> {
+    val signalsPositions = infra.blockInfra.getSignalsPositions(blockId)
+    val blockLength = infra.blockInfra.getBlockLength(blockId).distance
+    val nextSignalPosition = signalsPositions.firstOrNull { it.distance >= blockOffset.distance }
+
+    // some blocks are < 1m long (even 0m), we can't get further in the block
+    val maxHeadOffset =
+        if (blockOffset.distance < 1.meters) {
+            blockOffset.distance
```
eckter: That `if` feels wrong, if the train starts at the beginning of the block, then we don't use the next signal offset?

eckter (commented):
That's the right approach, good job :+1:

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
louisgreiner: You are right indeed, it sucks 😄 

Also, I couldn't manage otherwise to deal with that later in the code, if the train stops **exactly** on one of a border of the block (either `start` or `end` if block is `[start ; end]`), I then don't know how it will react.
That's why I introduced a magic value (1m) to make sure this will not happen.

Do you have any better solution ? Or insight

eckter: I don't think we should really handle it in any specific way.

When the point is exactly at the end of a block, there's not much we can do anyway to move it further. It's supposed to be near a signal or buffer stop anyway. 

When it's at the beginning of a block, we can handle it normally by moving it forward using the same rules as usual.

louisgreiner:
Question : is the type of signalisation (here `TVM` instead of `BAL`) could affect the behaviour of the code ? 

image

Because here, the waypoint is moved AFTER the signal, which is not correct (but I only see that behaviour on `TVM` signals)

Could it be also related to the current bug that breaks the projection of intermediary waypoints on the map ?

Edit: looks like I was right, but difficult to be sure (the position here is now correct for the end pathstep)

image

editoast/editoast_schemas/src/train_schedule/train_schedule_options.rs
woshilapin: Hmm, I'm curious, but why not keeping `true` as a default value? Seems (naively maybe) like a good default, isn't it?

woshilapin (approved):
Looks good. I'm just wondering about making it a default option?

editoast/src/views/path/pathfinding.rs
```diff
@@ -472,7 +476,8 @@ pub mod tests {
                 "rolling_stock_supported_electrifications":[],
                 "rolling_stock_supported_signaling_systems":["BAL","BAPR"],
                 "rolling_stock_maximum_speed":22.00,
-                "rolling_stock_length":26.00
+                "rolling_stock_length":26.00,
+                "stop_at_next_signal":false
```
flomonster: We can omit this and use the default value.

editoast/editoast_schemas/src/train_schedule/train_schedule_options.rs
flomonster: The current behavior is `false`. If we set the new default behavior to `true` we need to add a migration to this PR. I do prefer to keep the behavior as `false`.

flomonster (changes requested):

Caracol3 (approved):
LGTM

front/src/modules/pathfinding/utils.ts
```diff
@@ -84,6 +84,7 @@ export const getPathfindingQuery = ({
...
```
clarani: Hmm I think it should be false by default ?

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
clarani: Same here

front/src/modules/pathfinding/utils.ts
louisgreiner: Indeed, it wrote above a comment about this : I kept it `true` to make it easier to test, but I will update these values at `false` as soon as the tests are done from the reviewers' side

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -369,3 +385,41 @@ private fun getBlockOffset(
...
+        getNextSignalOffset(waypointBlock.edge, waypointBlock.offset, infra, rollingStockLength)
+    return PathfindingEdgeLocationId(waypointBlock.edge, nextSignalOffset)
+}
+
+private fun getNextSignalOffset(
+    blockId: BlockId,
+    blockOffset: Offset<Block>,
+    infra: FullInfra,
+    rollingStockLength: Double
+): Offset<Block> {
```
Khoyo: This needs some documentation, it's quite unclear why a function called `getNextSignalOffset()` cares about the `rollingStockLength`

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -369,3 +385,41 @@ private fun getBlockOffset(
         String.format("getBlockOffset: Track chunk %s not in block %s", trackChunkId, blockId)
     )
 }
+
+/**
+ * Returns the block of the signal next to a waypoint block.
```
Khoyo: This doesn't return a Block

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -369,3 +385,41 @@ private fun getBlockOffset(
...
+    blockOffset: Offset<Block>,
+    infra: FullInfra,
+    rollingStockLength: Double
+): Offset<Block> {
+    val signalsPositions = infra.blockInfra.getSignalsPositions(blockId)
+    val blockLength = infra.blockInfra.getBlockLength(blockId).distance
+    val nextSignalPosition = signalsPositions.firstOrNull { it.distance >= blockOffset.distance }
+
+    val minTailOffset = blockOffset.distance + rollingStockLength.meters
+    val maxHeadOffset = if (blockOffset.distance < 0.01.meters) {
```
Khoyo: We should probably use `10.millimeters` instead of a floating point approximation. (It should also be a constant instead of a magic number)

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -369,3 +385,41 @@ private fun getBlockOffset(
...
+        blockOffset.distance
+    } else {
+        (nextSignalPosition?.distance ?: blockLength) - 0.01.meters
```
Khoyo: Ditto

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
Erashin: It is unclear whether this is still necessary or not in the overall PR. I'm down for a check/quick explanation :)

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
louisgreiner: You're totally right ; I moved and renamed some of this stuff, and also re-wrote the documentation attached. Is that ok that way ?

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
louisgreiner: I rephrased, in 
```kt
 * @return stop position on the signal next to the waypoint block.
```
But I also could have had been more concise : 

```kt
 * @return the stop position
```

What do you think ?

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
louisgreiner: It has been removed, since a heavenly fix came meanwhile

editoast/editoast_schemas/src/train_schedule/train_schedule_options.rs
```diff
@@ -14,8 +14,17 @@ pub struct TrainScheduleOptions {
     #[derivative(Default(value = "true"))]
     #[serde(default = "default_use_electrical_profiles")]
     use_electrical_profiles: bool,
+
+    #[derivative(Default(value = "false"))]
```
flomonster: 1. Default `bool` is already `false`. Check the doc [here](https://doc.rust-lang.org/std/primitive.bool.html#impl-Default-for-bool).
2. We should indicate to serde that the value can be omitted. 

```suggestion
    #[serde(default)]
```

flomonster (changes requested):

editoast/src/core/pathfinding.rs
```diff
@@ -42,6 +42,8 @@ pub struct PathfindingRequest {
     pub rolling_stock_maximum_speed: f64,
     /// Rolling stock length in meters:
     pub rolling_stock_length: f64,
+    /// If the train should stop on the next signal instead of on the operational point
```
bougue-pe: nit: This could match the comment from editoast's API (which may be rephrased as mentioned in another comment)

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -369,3 +385,38 @@ private fun getBlockOffset(
...
 }
+
+/**
+ * Returns the stop position of a train on the signal next to a waypoint block, considering its
+ * rolling stock length.
+ *
+ * @param waypointBlock waypoint block.
+ * @param rollingStockLength length of the rolling stock.
+ * @param infra full infra.
+ * @return stop position on the signal next to the waypoint block.
```
bougue-pe: nit: Those can be removed as, to my knowledge, it's not a target to document all inputs/outputs in core, and those don't add information to the existing code or comment above.

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -369,3 +385,38 @@ private fun getBlockOffset(
         String.format("getBlockOffset: Track chunk %s not in block %s", trackChunkId, blockId)
     )
 }
+
+/**
+ * Returns the stop position of a train on the signal next to a waypoint block, considering its
+ * rolling stock length.
```
bougue-pe: As:
* waypoint block is not so obvious name
* it does more than just get the position of the next signal
```suggestion
 * Given a waypoint block (initial position of a train stop), returns the new stop position of a train: next signal, with the constraint of staying in the same block and keeping the tail of the train on the initial position.
 * Note that nodes of the infra are not considered.
```

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -369,3 +385,38 @@ private fun getBlockOffset(
...
+ * @param infra full infra.
+ * @return stop position on the signal next to the waypoint block.
+ */
+fun findStopPositionOnNextSignalConsideringRollingStock(
+    waypointBlock: PathfindingEdgeLocationId<Block>,
+    rollingStockLength: Double,
+    infra: FullInfra
+): PathfindingEdgeLocationId<Block> {
+    // To ensure that the tail of the rolling stock is not further back than the operational point
+    val minTailOffset = waypointBlock.offset.distance + rollingStockLength.meters
```
bougue-pe: It's more of a max limit, no? (applying the min to the 2 max values at the end)
```suggestion
    val maxTailOffset = waypointBlock.offset.distance + rollingStockLength.meters
```

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -369,3 +385,38 @@ private fun getBlockOffset(
...
+ * @param rollingStockLength length of the rolling stock.
+ * @param infra full infra.
+ * @return stop position on the signal next to the waypoint block.
+ */
+fun findStopPositionOnNextSignalConsideringRollingStock(
+    waypointBlock: PathfindingEdgeLocationId<Block>,
+    rollingStockLength: Double,
+    infra: FullInfra
+): PathfindingEdgeLocationId<Block> {
+    // To ensure that the tail of the rolling stock is not further back than the operational point
```
bougue-pe: nit:
```suggestion
    // To ensure that the tail of the rolling stock is not further back than the operational point (waypointBlock)
```

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -369,3 +385,38 @@ private fun getBlockOffset(
...
+    val minTailOffset = waypointBlock.offset.distance + rollingStockLength.meters
+
+    // If there is no signal after the block, we consider the end of the block as the stop position
```
bougue-pe:
```suggestion
    // If there is no signal in the block after the operational point (waypointBlock), we consider the end of the block as the new stop position
```

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingV2Test.kt
```diff
@@ -239,4 +244,165 @@ class PathfindingV2Test : ApiTest() {
...
+    @Test
+    fun stopAtNextSignalTest() {
+        val waypointStart = TrackLocation("TA1", Offset(500.meters))
+        val waypointIntermediate = TrackLocation("TC1", Offset(550.meters))
+        val waypointEnd = TrackLocation("TD0", Offset(14000.meters))
+        val waypointsStart = listOf(waypointStart)
+        val waypointsIntermediate = listOf(waypointIntermediate)
+        val waypointsEnd = listOf(waypointEnd)
+        val waypoints = listOf(waypointsStart, waypointsIntermediate, waypointsEnd)
+        val reverseWaypoints = listOf(waypointsEnd, waypointsIntermediate, waypointsStart)
```
bougue-pe: I'd use `waypoints.reversed()` directly, instead.
(then maybe remove the 3 intermediate lists `waypointsStart`, etc. is a valid option)

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingV2Test.kt
```diff
@@ -239,4 +244,165 @@ class PathfindingV2Test : ApiTest() {
...
+                    rollingStockSupportedSignalingSystems = listOf("BAL"),
+                    rollingStockMaximumSpeed = 320.0,
+                    rollingStockLength = 200.0,
+                    timeout = null,
+                    infra = "small_infra/infra.json",
+                    expectedVersion = "1",
+                    pathItems = waypoints,
+                    stopAtNextSignal = false,
+                )
+            )
```
bougue-pe: nit: This could be mutualized in a free fn.

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingV2Test.kt
```diff
@@ -239,4 +244,165 @@ class PathfindingV2Test : ApiTest() {
...
+                .act(RqFake("POST", "/v2/pathfinding/blocks", reversedLongTrainRequestBody))
+        val reversedLongTrainResponse = TakesUtils.readBodyResponse(reversedLongTrainRawResponse)
+        val reversedLongTrainParsed =
+            (pathfindingResponseAdapter.fromJson(reversedLongTrainResponse)
+                as? PathfindingBlockSuccess)!!
+
+        assertEquals(3, nonStopAtNextSignalParsed.pathItemPositions.size)
+        assertEquals(0.meters, nonStopAtNextSignalParsed.pathItemPositions[0].distance)
+        assertEquals(12050.meters, nonStopAtNextSignalParsed.pathItemPositions[1].distance)
+        assertEquals(26500.meters, nonStopAtNextSignalParsed.pathItemPositions[2].distance)
```
bougue-pe: nit: I would use this one also for the reversed case: it tests/explains the control-case.

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingV2Test.kt
```diff
@@ -239,4 +244,165 @@ class PathfindingV2Test : ApiTest() {
...
+
+        assertEquals(3, reversedShortTrainParsed.pathItemPositions.size)
+        assertEquals(0.meters, reversedShortTrainParsed.pathItemPositions[0].distance)
+        assertEquals(14460.meters, reversedShortTrainParsed.pathItemPositions[1].distance)
+        assertEquals(26510.meters, reversedShortTrainParsed.pathItemPositions[2].distance)
+
+        assertEquals(3, longTrainParsed.pathItemPositions.size)
+        assertEquals(0.meters, longTrainParsed.pathItemPositions[0].distance)
+        assertEquals(12250.meters, longTrainParsed.pathItemPositions[1].distance)
+        assertEquals(26517.5.meters, longTrainParsed.pathItemPositions[2].distance)
```
bougue-pe: * Is this the only case tested on the signal?
* Should we also test the block-end?

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingV2Test.kt
```diff
@@ -239,4 +244,165 @@ class PathfindingV2Test : ApiTest() {
...
+                as? PathfindingBlockSuccess)!!
+
+        assertEquals(3, nonStopAtNextSignalParsed.pathItemPositions.size)
+        assertEquals(0.meters, nonStopAtNextSignalParsed.pathItemPositions[0].distance)
+        assertEquals(12050.meters, nonStopAtNextSignalParsed.pathItemPositions[1].distance)
+        assertEquals(26500.meters, nonStopAtNextSignalParsed.pathItemPositions[2].distance)
+
+        assertEquals(3, shortTrainParsed.pathItemPositions.size)
+        assertEquals(0.meters, shortTrainParsed.pathItemPositions[0].distance)
+        assertEquals(12060.meters, shortTrainParsed.pathItemPositions[1].distance)
```
bougue-pe: nit for explicitness:
How about defining before (ditto for other values):
```kt
val shortTrainLength = 10.meters
val initialIntermediateStop = 12050.meters
```
then use `initialIntermediateStop + shortTrainLength`

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingV2Test.kt
```diff
@@ -239,4 +244,165 @@ class PathfindingV2Test : ApiTest() {
...
+            )
+
+        val nonStopAtNextSignalRawResponse =
+            PathfindingBlocksEndpointV2(infraManager)
+                .act(RqFake("POST", "/v2/pathfinding/blocks", nonStopAtNextSignalRequestBody))
+        val nonStopAtNextSignalResponse =
+            TakesUtils.readBodyResponse(nonStopAtNextSignalRawResponse)
+        val nonStopAtNextSignalParsed =
+            (pathfindingResponseAdapter.fromJson(nonStopAtNextSignalResponse)
+                as? PathfindingBlockSuccess)!!
```
bougue-pe: nit: Mutualizing this would be nice too.

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingV2Test.kt
```diff
@@ -239,4 +244,165 @@ class PathfindingV2Test : ApiTest() {
...
+        val reversedLongTrainParsed =
+            (pathfindingResponseAdapter.fromJson(reversedLongTrainResponse)
+                as? PathfindingBlockSuccess)!!
+
+        assertEquals(3, nonStopAtNextSignalParsed.pathItemPositions.size)
+        assertEquals(0.meters, nonStopAtNextSignalParsed.pathItemPositions[0].distance)
+        assertEquals(12050.meters, nonStopAtNextSignalParsed.pathItemPositions[1].distance)
+        assertEquals(26500.meters, nonStopAtNextSignalParsed.pathItemPositions[2].distance)
+
+        assertEquals(3, shortTrainParsed.pathItemPositions.size)
```
bougue-pe: nit: Also, I would rather have each case separated (one test fn for each, or at least group body+request+assert)

editoast/src/views/path/pathfinding.rs
```diff
@@ -75,6 +75,8 @@ struct PathfindingInput {
     /// Rolling stock length
     #[schema(value_type = f64)]
     rolling_stock_length: OrderedFloat<f64>,
+    /// Stops the train at next signal instead of on path item
```
bougue-pe: As it does more than just this (see suggestion of code comment in core), it would be nice to elaborate more in the doc of the API.

python/osrd_schemas/osrd_schemas/train_schedule.py
```diff
@@ -170,12 +170,19 @@ class PowerRestrictionRanges(RootModel):
...
+    - `stop_at_next_signal` : for dealing with stopped trains that overflow on switches during imports
+    """
 
     ignore_electrical_profiles: bool = Field(
         default=False,
         description="If true, the electrical profiles are ignored in the standalone simulation",
     )
+    stop_at_next_signal: bool = Field(
+        default=False,
+        description="If true, the train will stop at the next signal instead of at the operational point",
```
bougue-pe: Same here, it would be nice to have the same description than editoast's API (and more explicit on what's done).

bougue-pe (commented):
🎉 Nice job (especially with the side-bug messing around)

Just comments on the form, the content is good to me.
Feel free to reject the nitpicks, they're not mandatory IMO and sometimes just a matter of taste.

editoast/editoast_schemas/src/train_schedule/train_schedule_options.rs
louisgreiner: thanks for the explanation :-)

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
louisgreiner: removed then, I thought we add documentation to public functions

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
louisgreiner: Totally right !! 💯

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingV2Test.kt
louisgreiner: Right, added

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingV2Test.kt
louisgreiner: Using small infra, it's hard to test more hazardous cases :/

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingV2Test.kt
louisgreiner: Done, is it conform to what you expected ?

bougue-pe: Yes, thank you very much!

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingV2Test.kt
bougue-pe: As discussed:
* this does test the signal case indeed.
* we should add a test on the "end-of-block" case, and the easiest is to use a very long (600m) train on the reverse case and check that it stops after 500m (buffer-stop of TA1)

bougue-pe (approved):
Thanks!

front/tests/assets/trainSchedule/train_schedules.json
```diff
@@ -92,7 +92,7 @@
     "constraint_distribution": "MARECO",
     "speed_limit_tag": "HLP",
     "power_restrictions": [],
-    "options": { "use_electrical_profiles": false }
+    "options": { "use_electrical_profiles": true, "stop_at_next_signal": false }
```
bougue-pe: From previous value, it looks like a bad copy-paste:
```suggestion
    "options": { "use_electrical_profiles": false, "stop_at_next_signal": false }
```

front/tests/assets/trainSchedule/train_schedules.json
```diff
@@ -264,7 +264,7 @@
     "constraint_distribution": "MARECO",
     "speed_limit_tag": null,
     "power_restrictions": [],
-    "options": { "use_electrical_profiles": false }
+    "options": { "use_electrical_profiles": true, "stop_at_next_signal": false }
```
bougue-pe:
```suggestion
    "options": { "use_electrical_profiles": false, "stop_at_next_signal": false }
```

front/tests/assets/trainSchedule/train_schedules.json
```diff
@@ -493,7 +493,7 @@
     "constraint_distribution": "STANDARD",
     "speed_limit_tag": "HLP",
     "power_restrictions": [],
-    "options": { "use_electrical_profiles": false }
+    "options": { "use_electrical_profiles": true, "stop_at_next_signal": false }
```
bougue-pe:
```suggestion
    "options": { "use_electrical_profiles": false, "stop_at_next_signal": false }
```

front/tests/assets/trainSchedule/train_schedules.json
```diff
@@ -672,6 +672,6 @@
     "constraint_distribution": "MARECO",
     "speed_limit_tag": "MA100",
     "power_restrictions": [],
-    "options": { "use_electrical_profiles": false }
+    "options": { "use_electrical_profiles": true, "stop_at_next_signal": false }
```
bougue-pe:
```suggestion
    "options": { "use_electrical_profiles": false, "stop_at_next_signal": false }
```

front/tests/assets/trainSchedule/train_schedules.json
louisgreiner: oops, thanks for the spot 🎯

louisgreiner:
## Feature not ready for STDCM simulation

Last commit `core: add feature flag to disable feature for stdcm` is meant to disable the feature for the simulated trains of STDCM. 

The issue is that, after having moved the waypoints, we can't match the new track positions with actual operational points. Then, the output tables can't match the stops information of the newly moved waypoints with actual operational points.

Example what we should get versus what we get with the feature:
Capture d'écran 2025-02-07 110951
Capture d'écran 2025-02-07 111552

However, it is still not an ideal trade-off, because with the feature disabled for STDCM only, if we want to debug the STDCM timetables, we still face this issue on the Operational Studies module.

Example, the stop information has disapeared from the output table, because, and we see it, the waypoint has moved:
Capture d'écran 2025-02-07 165731
Capture d'écran 2025-02-07 165744
Capture d'écran 2025-02-07 165851

## Feature still interesting for filling the STDCM timetable

The feature remains enabled for simulating the trains of the timetable STDCM stands on, thus potentially gaining still a good part of the capacity the feature should provide. But as mentionned above, it breaks the debug functionnality for STDCM timetable for the stops.

With the [current state of code](https://github.com/OpenRailAssociation/osrd/blob/dev/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt#L104-L106), it is difficult not to break anything to make the new positions of waypoints match with actual waypoints.

flomonster (approved):
LGTM for the editoast part, we should add a migration to drop this field (up.sql should be empty).

Synar (approved):
E2e-tests : LGTM!

louisgreiner:
Thanks to @Synar [PR](https://github.com/OpenRailAssociation/osrd/pull/12474) that fixes the issue on the operational studies output table (the moved stops were not displayed in the output table, when debugging an STDCM timetable), this PR will soon come to its end 🤞 

We need to wait that the above PR is merged, then final review can be done. I also fixed some tests to make sure that only the intermediate stops are moved close to the following signal (and not the start nor the last stop).

louisgreiner:
Dear reviewers, this PR is finally working 🎉 Because the PR is old, it might be necessary to review it once again.

Quick recap:
- this PR enables to move the stops of a pathfinding closer to the signal next to the requested stop
  - ⚠️ **the first and last stops are not moved** ⚠️
- the feature is enabled for STDCM (for all the trains simulated in the stdcm search environment AND for the simulated train)
- the feature is disabled for Operational Studies (trains created from this module cannot enable the feature, however imported trains that have the feature enabled (such as BASIC timetables) can be looked at in this module)
- [osrd-data associated MR](https://gitlab-repo-res.apps.eul.sncf.fr/dsir/groupedxs-dsir/04735/osrd-data/-/merge_requests/8)

To test this PR:
- either import this timetable in the Operational Studies module (you can first look at the stops on the map that they are indeed moved, train `"870004"` that stops at `"Gragnague"`, [timetable_chenille.json](https://github.com/user-attachments/files/21599681/timetable_chenille.json)), use this scenario as STDCM search environment (debug mode) and simulate trains in STDCM
- either locally import a BASIC timetable and run STDCM on it (also take a look at this timetable in OP module)

Questions: 
- should I do something about the [API changes automatic comment](https://github.com/OpenRailAssociation/osrd/pull/10200#issuecomment-3153925166)?
- ETCS simulations are missing in this PR, should it be part of it?

Khoyo:
> should I do something about the [API changes automatic comment](https://github.com/OpenRailAssociation/osrd/pull/10200#issuecomment-3153925166)?

You already did by opening the `osrd-data` MR. AFAIK, it's the only consumer project of editoast's API that we know of ;)

> ETCS simulations are missing in this PR, should it be part of it?

For now, ETCS is not supported in STDCM - but I don't see what ETCS (with fixed blocks) would have to do with this?

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
```diff
@@ -486,3 +506,36 @@ private fun makeHeuristicsForPathfindingEdges(
...
+    infra: FullInfra
+): PathfindingEdgeLocationId<Block> {
+    // To ensure that the tail of the rolling stock is not further than the initial operational
+    // point (waypointBlock) position
+    val maxTailOffset = waypointBlock.offset.distance + rollingStockLength.meters
+
+    // If there is no signal in the block after the operational point (waypointBlock), we consider
+    // the end of the block as the new stop position
+    val nextSignalPosition =
+        findNextSignalPositionFromBlockOffset(infra, waypointBlock.edge, waypointBlock.offset)
```
eckter: I don't think we need to bother with that. Only block-delimiting signals are relevant here (no point in stopping at a BAPR announcer signal). And block delimiting signals, well, they delimit blocks, can't move past one without changing block.

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
```diff
@@ -486,3 +506,36 @@ private fun makeHeuristicsForPathfindingEdges(
...
+    val maxHeadOffset =
+        nextSignalPosition?.distance ?: infra.blockInfra.getBlockLength(waypointBlock.edge).distance
+
+    return PathfindingEdgeLocationId(waypointBlock.edge, Offset(min(maxTailOffset, maxHeadOffset)))
```
eckter: We can keep offset types as `Offset<Block>` without too much troubles. You'd need the specific min method here (`Offset.min`), and remove a few `.distance` here and there. (note: `Offset + Distance` is OK and returns an `Offset`)

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlockRequest.kt
```diff
@@ -17,6 +17,7 @@ class PathfindingBlockRequest(
     val rollingStockSupportedSignalingSystems: List<String>,
     @Json(name = "rolling_stock_maximum_speed") val rollingStockMaximumSpeed: Double,
     @Json(name = "rolling_stock_length") val rollingStockLength: Double,
+    @Json(name = "stops_at_next_signal") val stopsAtNextSignal: Boolean,
```
eckter: I think I would have preferred if the flag were optional (with default to false), like how it's optional in the editoast API. The behavior with flag enable not that intuitive, while the behavior with no flag is what you'd normally expect. And the usecase is fairly niche. 

Ideally someone could start using the endpoint without having to wonder what that actually does and which value to chose.

eckter (approved):
Some small comments, but for the most part, LGTM

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
louisgreiner: Oh yes, I did not thought of this case. Should we then rename the feature (flag and etc.) to precise that we specificly stop at block-delimiting signals only? Like "stops_at_end_of_block" or something similar?

louisgreiner: Ok, done

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlockRequest.kt
louisgreiner: I agree, done

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
louisgreiner: Nice, done

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
```diff
@@ -104,10 +103,29 @@ class PathfindingBlocksEndpoint(private val infraManager: InfraProvider) : Take
...
+    request.pathItems.forEachIndexed { stepIndex, step ->
         val allStarts = HashSet<PathfindingEdgeLocationId<Block>>()
         for (direction in Direction.entries) {
-            for (waypoint in step) allStarts.addAll(findWaypointBlocks(infra, waypoint, direction))
+            for (waypoint in step) {
+                val waypointBlocks = findWaypointBlocks(infra, waypoint, direction)
+                if (
+                    request.stopsAtEndOfBlock == true &&
+                        stepIndex != 0 &&
+                        stepIndex != request.pathItems.size - 1
```
bougue-pe, line 114: We could end with a penultimate stop being pushed after the ultimate one, no?

What we could do about it:
* Add a test on a penultimate stop being before the ultimate one and in the same block, ultimate stop being before the end of the block.
* Cap the offset to the initial length of the TravelledPath (if we have it) while finding new stop position, or after when filling `waypoints` (to be applied for STDCM too in that case). It may require passing the positions of the possible last steps to compare if on the same edge?
* I don't think it's necessary to prevent the move before first stop, but if it's easy and clear we can add it.

front/tests/assets/stdcm/linked-train/anterior-linked-train-table.json
```diff
@@ -13,7 +13,7 @@
     "index": 2,
     "operationalPoint": "North_station",
     "code": "BV",
-    "endStop": "13:18",
+    "endStop": "13:17",
```
bougue-pe: Just curious: why those 1 min changes?

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingTest.kt
```diff
@@ -37,6 +39,7 @@ fun getPathfindingBlockRequest(
         rs.maxSpeed,
         rs.length,
         null,
+        false,
```
bougue-pe: nit: you may add a new param defaulting to `false`

also nit: you may use this method in your tests (there is a `VERY_SHORT_FAST_TRAIN`, `REALISTIC_FAST_TRAIN` and `VERY_LONG_FAST_TRAIN`)


Please feel very free to dismiss this as the fatigue on this much delayed PR is perfectly understandable.

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingTest.kt
```diff
@@ -620,3 +623,255 @@ class PathfindingTest : ApiTest() {
...
+            rollingStockLength = rollingStockLength,
+            timeout = null,
+            infra = "small_infra/infra.json",
+            expectedVersion = 1,
+            pathItems = pathItems,
+            stopsAtEndOfBlock = stopsAtEndOfBlock,
+        )
+    }
+
+    private fun pathfindingRequest(
```
bougue-pe: It looks like it can now be replaced by `checkPathfindingSuccess()`
The expected length will have to be provided, which is easy to have (may be manually compared to the case of `stopsAtEndOfBlock = false` if not already done for the last stop).

python/osrd_schemas/osrd_schemas/train_schedule.py
```diff
@@ -211,6 +213,11 @@ class TrainScheduleOptions(BaseModel):
         default=True,
         description="If false, the speed limits of the infrastructure are ignored in the standalone simulation",
     )
+    stops_at_end_of_block: bool = Field(
```
bougue-pe, line 216: Should we change (bump from 0.8.19 to 0.8.20) the version of the Python package in [`osrd_schemas::pyproject.toml`](https://github.com/OpenRailAssociation/osrd/tree/dev/python/osrd_schemas/pyproject.toml#L3) ?
I guess it depends mostly on how we want to handle it in osrd-data.

bougue-pe (commented):
At least testing the case of penultimate "overtaking" the ultimate stop seems important.

I can give a hand around this, NP.

python/osrd_schemas/osrd_schemas/train_schedule.py
louisgreiner: Done

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingTest.kt
louisgreiner: Done

front/tests/assets/stdcm/linked-train/anterior-linked-train-table.json
louisgreiner: It may be because the previous stop is now closer to this stop? I don't really know

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingTest.kt
louisgreiner: Done, the long train tests now fail. It outputs something interesting: 
```sh
expected: <[0m, 112050m, 26537.500m, 45900m]> but was: <[0m, 12320m, 26537.500m, 45900m]>
```

The first stop is way further than the next ones (because now the rs length is 100 000m)

It must be fixed

python/osrd_schemas/osrd_schemas/train_schedule.py
```diff
@@ -201,6 +201,8 @@ class TrainScheduleOptions(BaseModel):
     - `use_electrical_profiles` : use the electrical profiles for the standalone simulation
     - `use_speed_limits_for_simulation` : use the speed limits coming from the infrastructure for
        the standalone simulation
+    - `stops_at_end_of_block` : stops the train at the next block-delimiting signal instead of on the waypoint
```
shenriotpro: nit: use imperative form (ditto for all occurrences)

python/osrd_schemas/osrd_schemas/train_schedule.py
```diff
@@ -211,6 +213,11 @@ class TrainScheduleOptions(BaseModel):
         default=True,
         description="If false, the speed limits of the infrastructure are ignored in the standalone simulation",
     )
+    stops_at_end_of_block: bool = Field(
+        default=False,
+        description="If true, the train will stop at a new stop position: near next block-delimiting signal, \
```
shenriotpro: nit: not sure new is the right word, it won't be new for long..

shenriotpro (approved):

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
ppitou: @bougue-pe I handled this case and added tests if you could review it again.

python/osrd_schemas/osrd_schemas/train_schedule.py
Khoyo: "If true, stop positions will be shifted toward the next block-delimiting signal" ?

bougue-pe: > "If true, stop positions will be shifted toward the next block-delimiting signal" ?

👍 that (and keeping the details of the second line below) would be nice IMO.

python/osrd_schemas/osrd_schemas/train_schedule.py
bougue-pe: It (expectedly) got lost in rebases.

front/tests/assets/stdcm/linked-train/anterior-linked-train-table.json
bougue-pe: It's now gone, resolving it.

It may have been that the application of safe approach speed around signal matches better the train's braking to a stop 🤷

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingTest.kt
bougue-pe: Oh, maybe add a check on the sort of offsets would be nice on the way, then?

I let you see if you want to handle it or not, resolving without it is fine by me.

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingTest.kt
```diff
@@ -623,3 +626,265 @@ class PathfindingTest : ApiTest() {
...
+                ),
+        )
+    }
+
+    @Test
+    fun longTrainTest() {
+        // The first intermediate stop is moved to the next block-delimiting signal, but the
+        // distance available is less than the length of the train, so it is moved to the end of the
+        // block
+        // (270 meters after the first intermediate stop)
```
bougue-pe: nit: you got bitten by spotless (ditto on all the comments of the file) 😿
And my other suggestions might be affected too, of course 😓
```suggestion
        // block (270 meters after the first intermediate stop)
```

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingTest.kt
```diff
@@ -623,3 +626,265 @@ class PathfindingTest : ApiTest() {
...
+        val startWaypoint = TrackLocation("TC1", Offset(500.meters))
+        val intermediateWaypoint = TrackLocation("TD0", Offset(12600.meters))
+        val endWaypoint = TrackLocation("TD0", Offset(13000.meters))
+        waypoints = listOf(listOf(startWaypoint), listOf(intermediateWaypoint), listOf(endWaypoint))
+
+        // The intermediate stop is moved to the next block-delimiting signal, but the
+        // distance available is less than the length of the train, so it is moved to the end of the
+        // block,
+        // but the end of the block is also the destination so the intermediate stop remains the
+        // same
```
bougue-pe: Not testing versus the end of the block here, just the destination:
```suggestion
        // distance available is more than the length of the train, so it would be moved by the length of the train (400m)
        // but this is also exactly the destination so the intermediate stop remains the same
```

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingTest.kt
```diff
@@ -623,3 +626,265 @@ class PathfindingTest : ApiTest() {
...
+            expectedIntermediatePathItemPosition =
+                listOf(
+                    Offset(recalculatedFirstIntermediateStopDistance),
+                    Offset(recalculatedSecondIntermediateStopDistance),
+                ),
+        )
+    }
+
+    @Test
+    fun penultimateStopInsideDestinationBlock() {
```
bougue-pe: To me this test is currently a corner case (exact limit) of the next test, and the naming may be misleading.
```suggestion
    fun penultimateStopNotMovedExactlyToDestination() {
```

bougue-pe: This may be worth testing that it's correctly moved in the same block, with intermediateWaypoint at 12550 (previous signal is at 12500): i.e. not hitting destination.

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingTest.kt
```diff
@@ -623,3 +626,265 @@ class PathfindingTest : ApiTest() {
...
+        var waypoints: List<List<TrackLocation>>
+        val startWaypoint = TrackLocation("TC1", Offset(500.meters))
+        val intermediateWaypoint = TrackLocation("TD0", Offset(12990.meters))
+        val endWaypoint = TrackLocation("TD0", Offset(13000.meters))
+        waypoints = listOf(listOf(startWaypoint), listOf(intermediateWaypoint), listOf(endWaypoint))
+
+        // The intermediate stop is moved to the next block-delimiting signal, but the
+        // distance available is less than the length of the train, so it is moved to the end of the
+        // block,
+        // but the end of the block after the destination so the intermediate stop remains the same
```
bougue-pe:
```suggestion
        // distance available is more than the length of the train, so it would be moved by the length of the train,
        // but that is after the destination so the intermediate stop remains the same
```

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingTest.kt
```diff
@@ -623,3 +626,265 @@ class PathfindingTest : ApiTest() {
...
+            )
+        checkPathfindingSuccess(
+            parsed,
+            13500.meters,
+            expectedIntermediatePathItemPosition = listOf(Offset(intermediateStopDistance)),
+        )
+    }
+
+    @Test
+    fun penultimateStopCloseToDestination() {
```
bougue-pe:
```suggestion
    fun penultimateStopTooCloseToDestination() {
```

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
```diff
@@ -101,10 +102,33 @@ class PathfindingBlocksEndpoint(private val infraManager: InfraProvider) : Take
...
+                ) {
+                    val destinationTrack = request.pathItems.last().last()
+                    val destinationBlock =
+                        findWaypointBlocks(infra, destinationTrack, direction).last()
```
bougue-pe: There's a problem here: the pathItems (destination also) can be OP (with multiple OP-parts, on different tracks) and each can be in different blocks (one for each signaling system protecting the range).
So we're only considering the last OP-part of the destination, and the last signaling-system, which may not be the ones actually used as destination.

It looks like:
* we should get all the blocks+offsets of the destination (function without `direction` param)
* then only in the `findStopPositionAtEndOfBlockConsideringRollingStock`, filter on the actual block of the waypoint, keeping only the minimal offset if multiple parts are in the block.

core/src/main/kotlin/fr/sncf/osrd/api/stdcm/STDCMEndpoint.kt
```diff
@@ -325,20 +327,34 @@ fun parseSteps(
...
-                    it.stepTimingData.arrivalTimeToleranceAfter,
-                )
-            else null,
-        )
-    }
+    return pathItems
+        .mapIndexed { index, it ->
+            STDCMStep(
+                if (index != 0 && index != pathItems.size - 1) {
+                    val destinationBlock = findWaypointBlocks(infra, it.locations).last()
```
bougue-pe: Kind of the same problem here: we're considering (only) the last block-location of the current item (not the destination).

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingTest.kt
```diff
@@ -623,3 +626,265 @@ class PathfindingTest : ApiTest() {
...
+                "small_infra/infra.json",
+                true,
+            )
+        checkPathfindingSuccess(
+            parsed,
+            13500.meters,
+            expectedIntermediatePathItemPosition = listOf(Offset(intermediateStopDistance)),
+        )
+    }
+}
```
bougue-pe, line 923: After the review, should we test also:
* 2 penultimate stops being too close from destination
* destination and penultimate stops having different parts and different signaling systems

?

IMO: "No, it's not worth the hastle" is a valid answer (especially for the second one), but if you think of an easy way to manage it...

bougue-pe (commented):
The last 3 comments are especially important.

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingTest.kt
```diff
@@ -623,3 +626,318 @@ class PathfindingTest : ApiTest() {
...
+            listOf(
+                listOf(startWaypoint),
+                listOf(intermediateWaypoint),
+                listOf(secondIntermediateWaypoint),
+                listOf(endWaypoint),
+            )
+
+        // The intermediate stop is moved to the next block-delimiting signal, but the
+        // distance available is more than the length of the train, so it would be moved by the
+        // length of the train, but that is after the destination so the intermediate stop remains the same
```
bougue-pe:
```suggestion
        // The intermediate stops are moved to the next block-delimiting signal, but the
        // distance available is more than the length of the train, so it would be moved by the
        // length of the train, but that is after the destination so the intermediate stops remain the same
```

core/src/main/kotlin/fr/sncf/osrd/api/stdcm/STDCMEndpoint.kt
```diff
@@ -417,7 +433,7 @@ private fun parseSimulationScheduleItems(
     )
 }
 
-private fun findWaypointBlocks(
+fun findWaypointBlocks(
```
bougue-pe: nit: better grouping it next to the other one in `PathfindingBlocksEndpoint.kt` (and if a rename is necessary, I'd go for "find`Directed`WaypointBlocks" for the other function).

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
```diff
@@ -101,10 +104,32 @@ class PathfindingBlocksEndpoint(private val infraManager: InfraProvider) : Take
...
-            for (waypoint in step) allStarts.addAll(findWaypointBlocks(infra, waypoint, direction))
+            for (waypoint in step) {
+                val waypointBlocks = findWaypointBlocks(infra, waypoint, direction)
+                if (
+                    request.stopsAtEndOfBlock == true &&
+                        stepIndex != 0 &&
+                        stepIndex != request.pathItems.size - 1
+                ) {
+                    val destinationTrack = request.pathItems.last()
+                    val destinationBlock = findWaypointBlocks(infra, destinationTrack)
```
bougue-pe: nit: this can be moved out of the `forEachIndexed` loop and computed only once.

bougue-pe (commented):
✅ LGTM: The code logic matches the business logic as I understand it, thanks for the work!

I suspect tests are breaking now with the correction :-/
I can probably peer on fixing tests if needed.

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingTest.kt
bougue-pe: ✅ OK to resolve IMO (I let it open just in case it acts as a reminder for you)

✔️ Done for the first suggestion of two intermediate stops too close to destination.
✖️ Not done for the multiple parts and multiple signaling systems one (much heavier probably)

python/osrd_schemas/osrd_schemas/train_schedule.py
ppitou: Done

python/osrd_schemas/osrd_schemas/train_schedule.py
ppitou: Done

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
ppitou: Done

core/src/main/kotlin/fr/sncf/osrd/api/stdcm/STDCMEndpoint.kt
ppitou: Done

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingTest.kt
ppitou: Done

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
ppitou: Thank you ! Done

core/src/main/kotlin/fr/sncf/osrd/api/stdcm/STDCMEndpoint.kt
ppitou: Done

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingTest.kt
ppitou: Thank you ! :dragon:

bougue-pe (approved):
LGTM, thanks very much for the work!

There are 2 comments hanging (conveniently 😬 hidden by github's UX):
* https://github.com/OpenRailAssociation/osrd/pull/10200#discussion_r2569079043
* https://github.com/OpenRailAssociation/osrd/pull/10200#discussion_r2272878629

Also gl for the squash and the fix of e2e tests 💪

python/osrd_schemas/osrd_schemas/train_schedule.py
ppitou: Done :P

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingTest.kt
ppitou: Done

bougue-pe:
It looks like we are only missing a front review (+squash) before merge, no?

[What is the strategy for client repository on this](https://github.com/OpenRailAssociation/osrd/pull/10200#issuecomment-3153925166)? Merging this one in OSRD, and later activate it in client project (data import), I guess?

Khoyo:
Indeed, since this is a non-breaking change, we can merge it now and the downstream repos can use it at their leisure. For osrd-data, that'll be any point after this hits the deployed dev env.

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -3933,6 +3936,7 @@ export type Margins = {
   values: string[];
 };
 export type TrainScheduleOptions = {
+  stops_at_end_of_block?: boolean;
```
SharglutDev, line 3939: Maybe we could also handle the case in `selectTrainToEdit` function (l.82) in `front/src/reducers/osrdconf/operationalStudiesConf/index.ts`

SharglutDev (approved):
Lgtm (not tested), good job ! You can close the comment after fixing it.

editoast/openapi.yaml
```diff
@@ -14209,6 +14216,8 @@ components:
     TrainScheduleOptions:
       type: object
       properties:
+        stops_at_end_of_block:
```
bougue-pe, line 14219: If I understand correctly, this is optional for pathfinding (and TrainSchedule when storing the information).
And this is now automatic for STDCM, right?

🙏 Can you please edit the main (top) comment of the PR to express that (or correct me if I'm mistaken)?
Not sure if we can document it clearly in STDCM's API, also.

ppitou: Is it clearer for the top comment ?

front/src/common/api/generatedEditoastApi.ts
ppitou: Done

editoast/openapi.yaml
bougue-pe: It is already clearer, yes, thanks.

---

For the trains simulated in the stdcm search environment, it depends on what was picked when importing data (like BASIC imports), no? (in practice we are gonna use it, but it's not tied to OSRD, just how we use it)

And I also plan to mention that:
* the feature can't be deactivated (front or back) for STDCM's path output.
* the feature can be switched on/off for pathfinding and train-schedule (operational studies) in the backend, but it can't be activated in the frontend.

If you don't correct me, I'm OK to change a bit more myself.

ppitou: Yes it's good for me !

bougue-pe: Thanks, done.

Khoyo (approved):
Ship it

PR #10203, opened 2024-12-30
front: move grid margins props from commonConf to stdcmConf reducer

clarani:
:white_check_mark: To merge after https://github.com/OpenRailAssociation/osrd/pull/10202

### DIFF
--- a/front/src/reducers/osrdconf/stdcmConf/index.ts
+++ b/front/src/reducers/osrdconf/stdcmConf/index.ts
@@ -31,9 +31,9 @@ export const stdcmConfInitialState: OsrdStdcmConfState = {
     },
   ],
   margins: {
-    standardAllowance: undefined,
-    gridMarginBefore: undefined,
-    gridMarginAfter: undefined,
+    standardAllowance: { type: 'time_per_distance', value: 4.5 },
+    gridMarginBefore: 15,
+    gridMarginAfter: 15,
   },
   totalMass: undefined,
   totalLength: undefined,
@@ -106,6 +106,13 @@ export const stdcmConfSlice = createSlice({
       state.maxSpeed = action.payload.maxSpeed;
       state.stdcmPathSteps = action.payload.stdcmPathSteps;
       state.speedLimitByTag = action.payload.speedLimitByTag;
+    },
+    resetMargins(state: Draft<OsrdStdcmConfState>) {
+      state.margins = {
+        standardAllowance: { type: 'time_per_distance', value: 4.5 },
+        gridMarginBefore: 15,
+        gridMarginAfter: 15,
+      };
     },
     updateStandardAllowance(
       state: Draft<OsrdStdcmConfState>,
@@ -223,8 +230,12 @@ export const stdcmConfSlice = createSlice({
 
 export const stdcmConfSliceActions = stdcmConfSlice.actions;
 
-export const { updateGridMarginAfter, updateGridMarginBefore, updateStandardAllowance } =
-  stdcmConfSliceActions;
+export const {
+  resetMargins,
+  updateGridMarginAfter,
+  updateGridMarginBefore,
+  updateStandardAllowance,
+} = stdcmConfSliceActions;
 
 export type StdcmConfSlice = typeof stdcmConfSlice;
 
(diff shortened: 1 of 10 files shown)


front/src/reducers/osrdconf/types.ts
SharglutDev, line 1: Checked where we use `StandardAllowance`, seems like `createAllowanceValue` function isn't used anymore (and maybae some other `consts` / `types` related to it. Could you remove it a separate commit ? :)

SharglutDev, line 1: Also `AllowanceValue` can accept `value_type: 'time';` but we don't use it anywhere, we could maybe remove it as well (and it the related `consts` / `types`)

front/src/reducers/osrdconf/stdcmConf/stdcmConfReducers.spec.ts
```diff
@@ -71,20 +71,38 @@ describe('stdcmConfReducers', () => {
     expect(state).toEqual(stdcmConfInitialState);
   });
 
-  it('should handle updateStdcmStandardAllowance', () => {
-    const initialTimeStandardAllowance = testDataBuilder.buildTimeStandardAllowance(10);
-    const store = createStore({
-      standardStdcmAllowance: initialTimeStandardAllowance,
-    });
+  it('should handle margins update', () => {
```
SharglutDev: To fix your failing test

```suggestion
  describe('should handle margins update', () => {
```

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -30,7 +30,11 @@ export const stdcmConfInitialState: OsrdStdcmConfState = {
       tolerances: { before: DEFAULT_TOLERANCE, after: DEFAULT_TOLERANCE },
     },
   ],
-  standardStdcmAllowance: undefined,
+  margins: {
```
SharglutDev, line 33: In `StdcmConfig.tsx` l.120, we give some initial values for margins if we are not in debug mode, which is the default behavior. Maybe we could remove this useEffect and directly initialize the state with these values ?

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -207,6 +223,9 @@ export const stdcmConfSlice = createSlice({
 
 export const stdcmConfSliceActions = stdcmConfSlice.actions;
 
+export const { updateGridMarginAfter, updateGridMarginBefore, updateStandardAllowance } =
```
SharglutDev: Same question as your other store PR, is it useful to export these (same for selectors) as we can already access them in `osrdContext` ? (I think it was the point of the context)

front/src/reducers/osrdconf/stdcmConf/stdcmConfReducers.spec.ts
clarani: Done !

front/src/reducers/osrdconf/stdcmConf/index.ts
clarani: I added the new default value :+1: 
I kept the useEffect because if I switch from debugMode to classicMode, I need to reset these values (except if we want to handle the debug mode when creating the payload for stdcm ?)

I also added a new action, I find it cleaner like this

clarani:
@SharglutDev  I fixed some of your comments related to cleaning AllowanceValue type in this [PR](https://github.com/OpenRailAssociation/osrd/pull/10222), since it was a bit out of scope :+1:

front/src/reducers/osrdconf/types.ts
clarani: Done here https://github.com/OpenRailAssociation/osrd/pull/10222 :)

clarani: Done here https://github.com/OpenRailAssociation/osrd/pull/10222

front/src/reducers/osrdconf/stdcmConf/index.ts
SharglutDev: Perfect !

SharglutDev (approved):
Lgtm and tested

PR #10206, opened 2024-12-30
front: convert stopFor in seconds

clarani:
It may break the localStorage :/


clarani:
Duplicate with https://github.com/OpenRailAssociation/osrd/pull/9934, not relevant anymore

PR #10212, opened 2024-12-31
editoast: endpoints to retrieve stdcm payloads

hamz2a:
Part of https://github.com/OpenRailAssociation/osrd/issues/9724 

> [!WARNING]
> We need to wait for this [PR](https://github.com/OpenRailAssociation/osrd/pull/10197) to be merged before merging that one.

### DIFF
--- a/editoast/openapi.yaml
+++ b/editoast/openapi.yaml
@@ -2590,6 +2590,69 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/StdcmSearchEnvironment'
+  /stdcm_log:
+    get:
+      tags:
+      - stdcm_log
+      parameters:
+      - name: trace_id
+        in: query
+        required: false
+        schema:
+          type: string
+          nullable: true
+      - name: id
+        in: query
+        required: false
+        schema:
+          type: integer
+          format: int64
+          nullable: true
+      responses:
+        '200':
+          description: The STDCM log
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/StdcmLog'
+  /stdcm_logs:
+    get:
+      tags:
+      - stdcm_log
+      parameters:
+      - name: page
+        in: query
+        required: false
+        schema:
+          type: integer
+          format: int64
+          default: 1
+          minimum: 1
+      - name: page_size
+        in: query
+        required: false
+        schema:
+          type: integer
+          format: int64
+          default: 25
+          nullable: true
+          minimum: 1
+      responses:
+        '200':
+          description: The list of STDCM Logs
+          content:
+            application/json:
+              schema:
+                allOf:
+                - $ref: '#/components/schemas/PaginationStats'
+                - type: object
+                  required:
+                  - results
+                  properties:
+                    results:
+                      type: array
+                      items:
+                        $ref: '#/components/schemas/StdcmLogListItem'
   /temporary_speed_limit_group:
     post:
       tags:
@@ -4748,6 +4811,9 @@ components:
       - $ref: '#/components/schemas/EditoastStdcmErrorTimetableNotFound'
       - $ref: '#/components/schemas/EditoastStdcmErrorTowedRollingStockNotFound'
       - $ref: '#/components/schemas/EditoastStdcmErrorTrainSimulationFail'
+      - $ref: '#/components/schemas/EditoastStdcmLogErrorMissingIdAndTraceId'
+      - $ref: '#/components/schemas/EditoastStdcmLogErrorNotFound'
+      - $ref: '#/components/schemas/EditoastStdcmLogErrorTraceIdNotFound'
       - $ref: '#/components/schemas/EditoastStudyErrorNotFound'
       - $ref: '#/components/schemas/EditoastStudyErrorStartDateAfterEndDate'
       - $ref: '#/components/schemas/EditoastTemporarySpeedLimitErrorNameAlreadyUsed'
@@ -5986,6 +6052,73 @@ components:
           type: string
           enum:
           - editoast:stdcm_v2:TrainSimulationFail
+    EditoastStdcmLogErrorMissingIdAndTraceId:
+      type: object
+      required:
+      - type
+      - status
+      - message
+      properties:
+        context:
+          type: object
+        message:
+          type: string
+        status:
+          type: integer
+          enum:
+          - 400
+        type:
+          type: string
+          enum:
+          - editoast:stdcm_log:MissingIdAndTraceId
+    EditoastStdcmLogErrorNotFound:
+      type: object
+      required:
+      - type
+      - status
+      - message
+      properties:
+        context:
+          type: object
+          required:
+          - id
+          properties:
+            id:
+              type: integer
+        message:
+          type: string
+        status:
+          type: integer
+          enum:
+          - 404
+        type:
+          type: string
+          enum:
+          - editoast:stdcm_log:NotFound
+    EditoastStdcmLogErrorTraceIdNotFound:
+      type: object
+      required:
+      - type
+      - status
+      - message
+      properties:
+        context:
+          type: object
+          required:
+          - trace_id
+          properties:
+            trace_id:
+              type: string
+        message:
+          type: string
+        status:
+          type: integer
+          enum:
+          - 404
+        type:
+          type: string
+          enum:
+          - editoast:stdcm_log:TraceIdNotFound
     EditoastStudyErrorNotFound:
       type: object
       required:
@@ -8317,7 +8450,26 @@ components:
           $ref: '#/components/schemas/PathItemLocation'
         timing_data:
           allOf:
-          - $ref: '#/components/schemas/StepTimingData'
+          - type: object
+            required:
+            - arrival_time
+            - arrival_time_tolerance_before
+            - arrival_time_tolerance_after
+            properties:
+              arrival_time:
+                type: string
+                format: date-time
+                description: Time at which the train should arrive at the location
+              arrival_time_tolerance_after:
+                type: integer
+                format: int64
+                description: The train may arrive up to this duration after the expected arrival time
+                minimum: 0
+              arrival_time_tolerance_before:
+                type: integer
+                format: int64
+                description: The train may arrive up to this duration before the expected arrival time
+                minimum: 0
           nullable: true
     PathfindingNotFound:
       oneOf:
@@ -8964,153 +9116,6 @@ components:
             type: integer
             format: int64
             minimum: 0
-    Request:
-      type: object
-      description: An STDCM request
-      required:
-      - steps
-      - rolling_stock_id
-      - comfort
-      properties:
-        comfort:
-          $ref: '#/components/schemas/Comfort'
-        electrical_profile_set_id:
-          type: integer
-          format: int64
-          nullable: true
-        loading_gauge_type:
-          allOf:
-          - $ref: '#/components/schemas/LoadingGaugeType'
-          nullable: true
-        margin:
-          type: string
-          description: Can be a percentage `X%`, a time in minutes per 100 kilometer `Xmin/100km`
-          example:
-          - 5%
-          - 2min/100km
-          nullable: true
-        max_speed:
-          type: number
-          format: double
-          description: |-
-            Maximum speed of the consist in km/h
-            Velocity in m·s⁻¹
-          nullable: true
-        maximum_departure_delay:
-          type: integer
-          format: int64
-          description: |-
-            By how long we can shift the departure time in milliseconds
-            Deprecated, first step data should be used instead
-          nullable: true
-          minimum: 0
-        maximum_run_time:
-          type: integer
-          format: int64
-          description: |-
-            Specifies how long the total run time can be in milliseconds
-            Deprecated, first step data should be used instead
-          nullable: true
-          minimum: 0
-        rolling_stock_id:
-          type: integer
-          format: int64
-        speed_limit_tags:
-          type: string
-          description: Train categories for speed limits
-          nullable: true
-        start_time:
-          type: string
-          format: date-time
-          description: Deprecated, first step arrival time should be used instead
-          nullable: true
-        steps:
-          type: array
-          items:
-            $ref: '#/components/schemas/PathfindingItem'
-        temporary_speed_limit_group_id:
-          type: integer
-          format: int64
-          nullable: true
-        time_gap_after:
-          type: integer
-          format: int64
-          description: |-
-            Margin after the train passage in milliseconds
-
-            Enforces that the path used by the train should be free and
-            available at least that many milliseconds after its passage.
-          minimum: 0
-        time_gap_before:
-          type: integer
-          format: int64
-          description: |-
-            Margin before the train passage in seconds
-
-            Enforces that the path used by the train should be free and
-            available at least that many milliseconds before its passage.
-          minimum: 0
-        total_length:
-          type: number
-          format: double
-          description: |-
-            Total length of the consist in meters
-            Length in m
-          nullable: true
-        total_mass:
-          type: number
-          format: double
-          description: |-
-            Total mass of the consist
-            Mass in kg
-          nullable: true
-        towed_rolling_stock_id:
-          type: integer
-          format: int64
-          nullable: true
-        work_schedule_group_id:
-          type: integer
-          format: int64
-          nullable: true
-    Response:
-      oneOf:
-      - type: object
-        required:
-        - simulation
-        - path
-        - departure_time
-        - status
-        properties:
-          departure_time:
-            type: string
-            format: date-time
-          path:
-            $ref: '#/components/schemas/PathfindingResultSuccess'
-          simulation:
-            $ref: '#/components/schemas/SimulationResponse'
-          status:
-            type: string
-            enum:
-            - success
-      - type: object
-        required:
-        - status
-        properties:
-          status:
-            type: string
-            enum:
-            - path_not_found
-      - type: object
-        required:
-        - error
-        - status
-        properties:
-          error:
-            $ref: '#/components/schemas/SimulationResponse'
-          status:
-            type: string
-            enum:
-            - preprocessing_simulation_error
     RjsPowerRestrictionRange:
       type: object
       description: A range along the train path where a power restriction is applied.
@@ -10797,7 +10802,6 @@ components:
       type: object
       required:
       - id
-      - trace_id
       - request
       - response
       - created
@@ -10809,15 +10813,276 @@ components:
           type: integer
           format: int64
         request:
-          $ref: '#/components/schemas/Request'
+          $ref: '#/components/schemas/StdcmRequest'
         response:
-          $ref: '#/components/schemas/Response'
+          $ref: '#/components/schemas/StdcmResponse'
         trace_id:
           type: string
+          nullable: true
         user_id:
           type: integer
           format: int64
           nullable: true
+    StdcmLogListItem:
+      type: object
+      required:
+      - id
+      properties:
+        id:
+          type: integer
+          format: int64
+        trace_id:
+          type: string
+          nullable: true
+    StdcmRequest:
+      type: object
+      required:
+      - infra
+      - expected_version
+      - path_items
+      - rolling_stock_loading_gauge
+      - rolling_stock_supported_signaling_systems
+      - comfort
+      - physics_consist
+      - trains_requirements
+      - start_time
+      - maximum_departure_delay
+      - maximum_run_time
+      - time_gap_before
+      - time_gap_after
+      - work_schedules
+      - temporary_speed_limits
+      properties:
+        comfort:
+          $ref: '#/components/schemas/Comfort'
+        expected_version:
+          type: string
+          description: Infrastructure expected version
+        infra:
+          type: integer
+          format: int64
+          description: Infrastructure id
+        margin:
+          allOf:
+          - oneOf:
+            - type: object
+              required:
+              - Percentage
+              properties:
+                Percentage:
+                  type: number
+                  format: double
+            - type: object
+              required:
+              - MinPer100Km
+              properties:
+                MinPer100Km:
+                  type: number
+                  format: double
+          nullable: true
+        maximum_departure_delay:
+          type: integer
+          format: int64
+          description: Maximum departure delay in milliseconds.
+          minimum: 0
+        maximum_run_time:
+          type: integer
+          format: int64
+          description: Maximum run time of the simulation in milliseconds
+          minimum: 0
+        path_items:
+          type: array
+          items:
+            $ref: '#/components/schemas/PathItem'
+          description: List of waypoints. Each waypoint is a list of track offset.
+        physics_consist:
+          type: object
+          required:
+          - effort_curves
+          - length
+          - max_speed
+          - startup_time
+          - startup_acceleration
+          - comfort_acceleration
+          - const_gamma
+          - inertia_coefficient
+          - mass
+          - rolling_resistance
+          properties:
+            base_power_class:
+              type: string
+              nullable: true
+            comfort_acceleration:
+              type: number
+              format: double
+            const_gamma:
+              type: number
+              format: double
+              description: |-
+                The constant gamma braking coefficient used when NOT circulating
+                under ETCS/ERTMS signaling system
+            effort_curves:
+              $ref: '#/components/schemas/EffortCurves'
+            electrical_power_startup_time:
+              type: integer
+              format: int64
+              description: |-
+                The time the train takes before actually using electrical power.
+                Is null if the train is not electric or the value not specified.
+              nullable: true
+              minimum: 0
+            etcs_brake_params:
+              allOf:
+              - $ref: '#/components/schemas/EtcsBrakeParams'
+              nullable: true
+            inertia_coefficient:
+              type: number
+              format: double
+            length:
+              type: integer
+              format: int64
+              description: Length of the rolling stock
+              minimum: 0
+            mass:
+              type: integer
+              format: int64
+              description: Mass of the rolling stock
+              minimum: 0
+            max_speed:
+              type: number
+              format: double
+              description: Maximum speed of the rolling stock
+            power_restrictions:
+              type: object
+              description: Mapping of power restriction code to power class
+              additionalProperties:
+                type: string
+            raise_pantograph_time:
+              type: integer
+              format: int64
+              description: |-
+                The time it takes to raise this train's pantograph.
+                Is null if the train is not electric or the value not specified.
+              nullable: true
+              minimum: 0
+            rolling_resistance:
+              $ref: '#/components/schemas/RollingResistance'
+            startup_acceleration:
+              type: number
+              format: double
+            startup_time:
+              type: integer
+              format: int64
+              minimum: 0
+        rolling_stock_loading_gauge:
+          $ref: '#/components/schemas/LoadingGaugeType'
+        rolling_stock_supported_signaling_systems:
+          $ref: '#/components/schemas/RollingStockSupportedSignalingSystems'
+        speed_limit_tag:
+          type: string
+          nullable: true
+        start_time:
+          type: string
+          format: date-time
+        temporary_speed_limits:
+          type: array
+          items:
+            type: object
+            description: Lighter description of a work schedule with only the relevant information for core
+            required:
+            - speed_limit
+            - track_ranges
+            properties:
+              speed_limit:
+                type: number
+                format: double
+                description: Speed limitation in m/s
+              track_ranges:
+                type: array
+                items:
+                  $ref: '#/components/schemas/TrackRange'
+                description: Track ranges on which the speed limitation applies
+          description: List of applicable temporary speed limits between the train departure and arrival
+        time_gap_after:
+          type: integer
+          format: int64
+          description: Gap between the created train and following trains in milliseconds
+          minimum: 0
+        time_gap_before:
+          type: integer
+          format: int64
+          description: Gap between the created train and previous trains in milliseconds
+          minimum: 0
+        time_step:
+          type: integer
+          format: int64
+          description: Numerical integration time step in milliseconds. Use default value if not specified.
+          nullable: true
+          minimum: 0
+        trains_requirements:
+          type: object
+          additionalProperties:
+            type: object
+            required:
+            - start_time
+            - spacing_requirements
+            - routing_requirements
+            properties:
+              routing_requirements:
+                type: array
+                items:
+                  $ref: '#/components/schemas/RoutingRequirement'
+              spacing_requirements:
+                type: array
+                items:
+                  $ref: '#/components/schemas/SpacingRequirement'
+              start_time:
+                type: string
+                format: date-time
+        work_schedules:
+          type: array
+          items:
+            $ref: '#/components/schemas/WorkSchedule'
+          description: List of planned work schedules
+    StdcmResponse:
+      oneOf:
+      - type: object
+        required:
+        - simulation
+        - path
+        - departure_time
+        - status
+        properties:
+          departure_time:
+            type: string
+            format: date-time
+          path:
+            $ref: '#/components/schemas/PathfindingResultSuccess'
+          simulation:
+            $ref: '#/components/schemas/SimulationResponse'
+          status:
+            type: string
+            enum:
+            - success
+      - type: object
+        required:
+        - status
+        properties:
+          status:
+            type: string
+            enum:
+            - path_not_found
+      - type: object
+        required:
+        - error
+        - status
+        properties:
+          error:
+            $ref: '#/components/schemas/SimulationResponse'
+          status:
+            type: string
+            enum:
+            - preprocessing_simulation_error
     StdcmSearchEnvironment:
       type: object
       required:
@@ -10883,27 +11148,6 @@ components:
           type: integer
           format: int64
           nullable: true
-    StepTimingData:
-      type: object
-      required:
-      - arrival_time
-      - arrival_time_tolerance_before
-      - arrival_time_tolerance_after
-      properties:
-        arrival_time:
-          type: string
-          format: date-time
-          description: Time at which the train should arrive at the location
-        arrival_time_tolerance_after:
-          type: integer
-          format: int64
-          description: The train may arrive up to this duration after the expected arrival time
-          minimum: 0
-        arrival_time_tolerance_before:
-          type: integer
-          format: int64
-          description: The train may arrive up to this duration before the expected arrival time
-          minimum: 0
     Study:
       type: object
       required:
--- a/editoast/src/client/runserver.rs
+++ b/editoast/src/client/runserver.rs
@@ -43,6 +43,10 @@ pub struct RunserverArgs {
     // only recieve 401 responses.
     #[clap(long, env = "EDITOAST_DISABLE_AUTHORIZATION", default_value_t = true)]
     disable_authorization: bool,
+    /// If this option is set, logging for the STDCM will be enabled.
+    /// When enabled, relevant logs will be captured to aid in debugging and monitoring.
+    #[clap(long, env = "ENABLE_STDCM_LOGGING", default_value_t = true)]
+    enable_stdcm_logging: bool,
     #[clap(long, env = "OSRDYNE_API_URL", default_value_t = Url::parse("http://127.0.0.1:4242/").unwrap())]
     osrdyne_api_url: Url,
     /// The timeout to use when performing the healthcheck, in milliseconds
@@ -64,6 +68,7 @@ pub async fn runserver(
                 core_client_channels_size,
             },
         disable_authorization,
+        enable_stdcm_logging,
         osrdyne_api_url,
         health_check_timeout_ms,
     }: RunserverArgs,
@@ -76,6 +81,7 @@ pub async fn runserver(
         health_check_timeout: Duration::milliseconds(health_check_timeout_ms as i64),
         map_layers_max_zoom: map_layers_config.max_zoom as u8,
         disable_authorization,
+        enable_stdcm_logging,
         postgres_config: postgres.into(),
         osrdyne_config: views::OsrdyneConfig {
             mq_url,
--- a/editoast/src/core/stdcm.rs
+++ b/editoast/src/core/stdcm.rs
@@ -20,10 +20,12 @@ use crate::core::AsCoreRequest;
 use crate::core::Json;
 
 editoast_common::schemas! {
+    Request,
     Response,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+#[schema(as = StdcmRequest)]
 pub struct Request {
     /// Infrastructure id
     pub infra: i64,
@@ -42,9 +44,11 @@ pub struct Request {
     /// The comfort of the train
     pub comfort: Comfort,
     pub speed_limit_tag: Option<String>,
+    #[schema(inline)]
     pub physics_consist: PhysicsConsist,
 
     // STDCM search parameters
+    #[schema(inline)]
     pub trains_requirements: HashMap<i64, TrainRequirements>,
     /// Numerical integration time step in milliseconds. Use default value if not specified.
     pub time_step: Option<u64>,
@@ -58,10 +62,12 @@ pub struct Request {
     /// Gap between the created train and following trains in milliseconds
     pub time_gap_after: u64,
     /// Margin to apply to the whole train
+    #[schema(inline)]
     pub margin: Option<MarginValue>,
     /// List of planned work schedules
     pub work_schedules: Vec<WorkSchedule>,
     /// List of applicable temporary speed limits between the train departure and arrival
+    #[schema(inline)]
     pub temporary_speed_limits: Vec<TemporarySpeedLimit>,
 }
 
@@ -98,7 +104,7 @@ pub struct WorkSchedule {
 }
 
 /// Lighter description of a work schedule with only the relevant information for core
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
 pub struct TemporarySpeedLimit {
     /// Speed limitation in m/s
     pub speed_limit: f64,
@@ -123,6 +129,7 @@ pub struct UndirectedTrackRange {
 // We accepted the difference of memory size taken by variants
 // Since there is only on success and others are error cases
 #[allow(clippy::large_enum_variant)]
+#[schema(as = StdcmResponse)]
 pub enum Response {
     Success {
         simulation: SimulationResponse,
--- a/editoast/src/models/stdcm_log.rs
+++ b/editoast/src/models/stdcm_log.rs
@@ -2,10 +2,8 @@ use chrono::DateTime;
 use chrono::Utc;
 use editoast_derive::Model;
 use editoast_models::DbConnection;
-use opentelemetry::trace::TraceContextExt;
 use serde::Deserialize;
 use serde::Serialize;
-use tracing_opentelemetry::OpenTelemetrySpanExt;
 use utoipa::ToSchema;
 
 use crate::core::stdcm::Request;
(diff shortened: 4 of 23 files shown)


front/src/common/api/generatedEditoastApi.ts
```diff
@@ -3411,19 +3456,52 @@ export type SimulationResponse =
...
       status: 'simulation_failed';
     };
-export type StepTimingData = {
-  /** Time at which the train should arrive at the location */
-  arrival_time: string;
-  /** The train may arrive up to this duration after the expected arrival time */
-  arrival_time_tolerance_after: number;
-  /** The train may arrive up to this duration before the expected arrival time */
-  arrival_time_tolerance_before: number;
+export type Response =
```
SharglutDev: Looks like the same type as `PostTimetableByIdStdcmApiResponse`, do we need another one ? Or we could reuse this in it with : `export type PostTimetableByIdStdcmApiResponse = Response`.

Also `Request` and `Response` are too vague imo, we could do `StdcmRequest` and `StdcmResponse` ?

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -3276,30 +3294,57 @@ export type StdcmSearchEnvironmentCreateForm = {
...
-  end_time: string;
-  /** List of requirements causing the conflict */
-  requirements: ConflictRequirement[];
-  /** Datetime of the start of the conflict */
-  start_time: string;
-  /** List of train ids involved in the conflict */
-  train_ids: number[];
-  /** List of work schedule ids involved in the conflict */
-  work_schedule_ids: number[];
+export type Request = {
```
SharglutDev: Same type as `PostTimetableByIdStdcmApiArg['body']`, we could reuse it inside ?

front/src/common/api/generatedEditoastApi.ts
hamz2a: We have structures that are quite similar, but some are specific to Core. This is the case for the two I use for the logs.
I have renamed them.

front/src/common/api/generatedEditoastApi.ts
hamz2a: Please look at the previous comment.

editoast/src/core/stdcm.rs
```diff
@@ -20,11 +20,13 @@ use crate::core::AsCoreRequest;
...
-    Response,
+    StdcmResponse,
+    StdcmRequest,
+    TemporarySpeedLimit,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct Request {
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct StdcmRequest {
```
leovalais: The renaming should only occur in the OpenAPI.

editoast/src/core/stdcm.rs
```diff
@@ -98,7 +100,7 @@ pub struct WorkSchedule {
 }
 
 /// Lighter description of a work schedule with only the relevant information for core
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
 pub struct TemporarySpeedLimit {
```
leovalais, line 108: Could be inlined instead.

editoast/src/models/stdcm_log.rs
```diff
@@ -18,37 +22,48 @@ editoast_common::schemas! {
 
 #[derive(Clone, Debug, Serialize, Deserialize, Model, ToSchema)]
 #[model(table = editoast_models::tables::stdcm_logs)]
-#[model(gen(ops = c))]
+#[model(gen(ops = cd, list))]
 pub struct StdcmLog {
     pub id: i64,
     pub trace_id: String,
```
```suggestion
	#[model(identifier)]
    pub trace_id: String,
```

Allows you to use Model traits by trace id.

editoast/src/models/stdcm_log.rs
```diff
@@ -18,37 +22,48 @@ editoast_common::schemas! {
...
+        use editoast_models::tables::stdcm_logs::dsl;
+
+        dsl::stdcm_logs
+            .filter(dsl::trace_id.eq(trace_id))
+            .first::<StdcmLogRow>(conn.write().await.deref_mut())
+            .await
+            .map(Into::into)
+            .optional()
+            .map_err(Into::into)
+    }
```
leovalais: `Retrieve` should be used instead, cf. previous comments.

editoast/src/views/stdcm_logs.rs
```diff
@@ -0,0 +1,206 @@
...
+        "/{trace_id}" => {
+            stdcm_log_by_trace_id,
+        },
+    },
+}
+
+#[derive(Debug, Error, EditoastError)]
+#[editoast_error(base_id = "stdcm_log")]
+pub enum StdcmLogError {
+    #[error("StdcmLog '{trace_id}', could not be found")]
```
```suggestion
    #[error("STDCM log entry '{trace_id}' could not be found")]
```

editoast/src/views/stdcm_logs.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    let stdcm_logs = StdcmLog::list(conn, SelectionSettings::new()).await?;
+
+    Ok(Json(stdcm_logs))
+}
+
+#[derive(IntoParams)]
+#[allow(unused)]
+pub struct StdcmLogTraceIdParam {
+    trace_id: String,
```
leovalais: There's no notion of `trace` in the API.

```suggestion
    id: String,
```

editoast/src/views/stdcm_logs.rs
```diff
@@ -0,0 +1,206 @@
...
+    tag = "stdcm_log",
+    responses(
+        (status = 200, body = StdcmLog, description = "The stdcm log"),
+        (status = 404, body = InternalError, description = "The stdcm log was not found"),
+    )
+)]
+async fn stdcm_log_by_trace_id(
+    State(db_pool): State<DbConnectionPoolV2>,
+    Extension(auth): AuthenticationExt,
+    Path(trace_id): Path<String>,
```
```suggestion
    Path(StdcmLogTraceIdParam { id: trace_id }): Path<StdcmLogTraceIdParam>,
```

editoast/src/views/test_app.rs
```diff
@@ -293,13 +297,13 @@ impl TestApp {
 }
 
 pub trait TestRequestExt {
-    fn with_user(self, user: UserInfo) -> Self;
+    fn with_user(self, user: &UserInfo) -> Self;
```
woshilapin: nitpick: Why did you change this signature? The intent of the function is to consume the information in order to inject it somewhere (`with_user` means you'll take that `UserInfo` to make something about it, not just read it). I would rather let the caller of this function `.clone()` if it needs the `UserInfo` instance multiple times.

editoast/src/views/stdcm_logs.rs
```diff
@@ -0,0 +1,206 @@
...
+    trace_id: String,
+}
+
+#[utoipa::path(
+    get, path = "",
+    params(StdcmLogTraceIdParam),
+    tag = "stdcm_log",
+    responses(
+        (status = 200, body = StdcmLog, description = "The stdcm log"),
+        (status = 404, body = InternalError, description = "The stdcm log was not found"),
```
```suggestion
```

editoast/src/views/stdcm_logs.rs
```diff
@@ -0,0 +1,206 @@
...
+    tag = "stdcm_log",
+    responses(
+        (status = 200, body = Vec<StdcmLog>, description = "The list of Stdcm Logs"),
+    )
+)]
+async fn list_stdcm_logs(
+    State(db_pool): State<DbConnectionPoolV2>,
+    Extension(auth): AuthenticationExt,
+) -> Result<Json<Vec<StdcmLog>>> {
+    let authorized = auth
```
leovalais: Should be paginated.

front/public/locales/en/errors.json
```diff
@@ -186,6 +186,9 @@
     "stdcm": {
       "InfraNotFound": "Infrastructure '{{infra_id}}' does not exist"
     },
+    "stdcm_log": {
+      "NotFound": ""
```
woshilapin: Is there a reason no message is provided here (same in `fr/errors.json`).

editoast/src/views/timetable/stdcm.rs
```diff
@@ -474,7 +490,7 @@ fn build_single_margin(margin: Option<MarginValue>) -> Margins {
 }
 
 #[cfg(test)]
-mod tests {
+pub mod tests {
```
woshilapin: How about moving all the needed functions into `fixtures.rs` instead of making `pub` this module and the functions (the pattern seems really weird).

leovalais: I'd prefer to duplicate the test data instead of sharing them in unrelated contexts. I think locality is important especially for test readability. In case of a schema change we'll rely on the help of the compiler anyways (provided we change the `json!` by a struct initialization).

front/public/locales/en/errors.json
leovalais: An error message would be nice

front/public/locales/fr/errors.json
```diff
@@ -188,6 +188,9 @@
     "stdcm": {
       "InfraNotFound": "Infrastructure {{infra_id}} non trouvée"
     },
+    "stdcm_log": {
+      "NotFound": ""
```
leovalais: An error message would be nice

woshilapin (changes requested):
Nice work, especially on the tests! I left a few comments.

leovalais (changes requested):

front/src/common/api/generatedEditoastApi.ts
leovalais: You can remove the `inline` in the stdcm endpoint to `$ref` it.

editoast/src/models/stdcm_log.rs
woshilapin: Doesn't retrieve forces you consider cases where no response or more than one response is returned? `.first()` does get you only one result or none.

editoast/src/views/stdcm_logs.rs
woshilapin: Maybe the struct can be renamed too?

editoast/src/models/stdcm_log.rs
leovalais: The `Retrieve` implementation that's generated includes the `first()`. And you can choose how to handle a missing row. If you expect multiple rows you need to use `RetrieveBatch`.

```rust
pub trait Retrieve<K>: Sized
where
    for<'async_trait> K: Send + 'async_trait,
{
    /// Retrieves the row #`id` and deserializes it as a model instance
    async fn retrieve(conn: &mut DbConnection, id: K) -> Result<Option<Self>>;

    /// Just like [Retrieve::retrieve] but returns `Err(fail())` if the row was not found
    async fn retrieve_or_fail<E, F>(
        conn: &'async_trait mut DbConnection,
        id: K,
        fail: F,
    ) -> Result<Self>
    where
        E: EditoastError,
        F: FnOnce() -> E + Send + 'async_trait,
    {
        match Self::retrieve(conn, id).await {
            Ok(Some(obj)) => Ok(obj),
            Ok(None) => Err(fail().into()),
            Err(e) => Err(e),
        }
    }
}
```

editoast/src/views/stdcm_logs.rs
leovalais: * Only a few endpoints list their errors, so I'd still suggest to remove it, at least for consistency's sake.
* Besides, it's not the only error the endpoint can return (authentication, forbidden access, database-related error, serde, ...). I'd argue that no list is more correct than an incomplete one.
* This part will be reworked (soon hopefully) once the new error system is embarked (which design doc I updated recently, input much appreciated 👀😬)

I don't really mind if the line stays, but it doesn't bring much information imo and `$ref`s a schema that will disappear so :shrug:

editoast/src/models/stdcm_log.rs
leovalais: We should checked that the `trace_id` column is actually indexed in DB

front/public/locales/en/errors.json
hamz2a: The API will be used directly without going through the frontend, so no message is needed.

hamz2a: The API will be used directly without going through the frontend, so no message is needed.

front/public/locales/fr/errors.json
hamz2a: The API will be used directly without going through the frontend, so no message is needed.

editoast/src/views/stdcm_logs.rs
hamz2a: Done.

editoast/src/views/test_app.rs

editoast/src/models/stdcm_log.rs
hamz2a: Done.

editoast/src/models/stdcm_log.rs
hamz2a: The `stdcm_logs` table has `id` and `trace_id`. In this case, I want to search by `trace_id` instead of `id`, but I don't know how to use `retrieve` for this. Can you give me an example, please?

editoast/src/views/stdcm_logs.rs
hamz2a: Done.

editoast/src/views/stdcm_logs.rs
hamz2a: Maybe I misunderstood the comment, but the purpose of the endpoint is to search by `trace_id`, not by the stdcm_log `id`.

editoast/src/views/stdcm_logs.rs
hamz2a: I got this error:
```bash
error[E0277]: the trait bound `fn(axum::extract::State<editoast_models::DbConnectionPoolV2>, axum::Extension<views::Authentication>, axum::extract::Path<views::stdcm_logs::StdcmLogTraceIdParam>) -> impl futures_util::Future<Output = std::result::Result<axum::Json<models::stdcm_log::StdcmLog>, error::InternalError>> {views::stdcm_logs::stdcm_log_by_trace_id}: axum::handler::Handler<_, views::AppState>` is not satisfied
   --> src/views/openapi.rs:202:92
    |
202 |           $crate::views::openapi::make_method_router::<$crate::routes!(@utoipa_type $route), _, _>(
    |                                                                                              ^ the trait `axum::handler::Handler<_, views::AppState>` is not implemented for fn item `fn(State<DbConnectionPoolV2>, ..., ...) -> ... {stdcm_log_by_trace_id}`
    |
   ::: src/views/stdcm_logs.rs:19:1
    |
19  | / crate::routes! {
20  | |     "/stdcm_logs" => {
21  | |         list_stdcm_logs,
22  | |         "/{trace_id}" => {
...   |
25  | |     },
26  | | }
    | |_- in this macro invocation
    |
    = note: the full name for the type has been written to '/Users/home/Desktop/osrd/editoast/target/debug/deps/editoast-5034878a3500a78c.long-type-13579499270931352073.txt'
    = note: consider using `--verbose` to print the full type name to the console
    = note: Consider using `#[axum::debug_handler]` to improve the error message
    = help: the following other types implement trait `axum::handler::Handler<T, S>`:
              `axum::handler::Layered<L, H, T, S>` implements `axum::handler::Handler<T, S>`
              `axum::routing::MethodRouter<S>` implements `axum::handler::Handler<(), S>`
              `axum_extra::handler::IntoHandler<H, T, S>` implements `axum::handler::Handler<T, S>`
              `axum_extra::handler::Or<L, R, Lt, Rt, S>` implements `axum::handler::Handler<(M, Lt, Rt), S>`
```

editoast/src/views/stdcm_logs.rs
woshilapin: Mmmh, I believe you're right indeed. So maybe it makes sense to expose the notion of "trace" in the API. If I recall correctly, the idea is to detect a problem through our observability stack (OpenTelemetry), and therefore find a trace ID, then ask `editoast` for more information about it (STDCM logs). Do you agree @leovalais? We might be able to resolve that comment and leave the code as is?

editoast/src/views/stdcm_logs.rs
woshilapin: You may need to `#[derive(serde::Deserialize)]` on `StdcmLogTraceIdParam` 🤷 (I'm looking at a similar case in `src/views/train_schedule.rs` with `TrainScheduleIdParam`). You can try, as suggested, to add (temporary, only for debugging the problem), the `#[axum::debug_handler]` above the endpoint's function, the compiler might give you a better hint on what the problem is.

hamz2a: It was missing `#[derive(serde::Deserialize)]`, thank you!

hamz2a: Done.

editoast/src/views/stdcm_logs.rs
hamz2a: Done.

editoast/src/core/stdcm.rs
hamz2a: Done.

editoast/src/views/timetable/stdcm.rs
hamz2a: I created `views::fixtures` to place these functions, not in `models::fixtures`, because they depend on `axum::http::StatusCode` and `views::path::pathfinding::PathfindingResult`.

editoast/src/core/stdcm.rs
hamz2a: Done.

editoast/src/core/stdcm.rs
hamz2a: I used `#[schema(inline)]`.

editoast/src/views/timetable/stdcm.rs
hamz2a: I created `views::fixtures` to place these functions, not in `models::fixtures`, because they depend on `axum::http::StatusCode` and `views::path::pathfinding::PathfindingResult`.

front/public/locales/en/errors.json
woshilapin: Right, but yet, I'd rather add a message even if never used by the front. My idea is to not cause a later developer that come see this piece of code and ask question around why, and lose some time about this weirdness, what do you think?

woshilapin (approved):
Thank you for all the improvements, I've resolved most of my comments (only left one opened, feel free to resolve if you fix, if you disagree, either way).

front/public/locales/en/errors.json
hamz2a: Done.

hamz2a: I added the message.

front/public/locales/fr/errors.json
hamz2a: I added the message.

editoast/openapi.yaml
```diff
@@ -2590,6 +2590,61 @@ paths:
...
+        required: false
+        schema:
+          type: integer
+          format: int64
+          default: 25
+          nullable: true
+          minimum: 1
+      responses:
+        '200':
+          description: The list of Stdcm Logs
```
leovalais:
```suggestion
          description: The list of STDCM logs
```

editoast/openapi.yaml
```diff
@@ -2590,6 +2590,61 @@ paths:
...
+      - stdcm_log
+      parameters:
+      - name: trace_id
+        in: path
+        required: true
+        schema:
+          type: string
+      responses:
+        '200':
+          description: The stdcm log
```
leovalais:
```suggestion
          description: The STDCM log
```

editoast/src/models/stdcm_log.rs
leovalais: `StcdmLog::retrieve(conn, String::from("trace id"))` should work, lmk if it doesnt

editoast/src/views/stdcm_logs.rs
```diff
@@ -0,0 +1,223 @@
...
+        .map_err(AuthorizationError::AuthError)?;
+    if !authorized {
+        return Err(AuthorizationError::Forbidden.into());
+    }
+
+    let conn = &mut db_pool.get().await?;
+
+    let settings = pagination_params
+        .validate(1000)?
+        .warn_page_size(100)
```
```suggestion
        .validate(10)?
```

editoast/src/views/stdcm_logs.rs
leovalais: You're right. These are just IDs when looking just at the API, but since this is the goal workflow let's keep `trace_id`.

editoast/src/core/stdcm.rs
```diff
@@ -19,11 +19,7 @@ use super::simulation::SimulationResponse;
 use crate::core::AsCoreRequest;
 use crate::core::Json;
 
-editoast_common::schemas! {
-    Response,
-}
-
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
```
leovalais, line 27: I think @SharglutDev wants it to be named `StdcmRequest` and referenced in the OpenApi. This can be done without renaming the strut using utoipa.

editoast/src/views/stdcm_logs.rs
```diff
@@ -0,0 +1,223 @@
...
+        list_stdcm_logs,
+        "/{trace_id}" => {
+            stdcm_log_by_trace_id,
+        },
+    },
+}
+
+#[derive(Debug, Error, EditoastError)]
+#[editoast_error(base_id = "stdcm_log")]
+pub enum StdcmLogError {
```
```suggestion
enum StdcmLogError {
```

editoast/src/views/stdcm_logs.rs
```diff
@@ -0,0 +1,223 @@
...
+    let (stdcm_logs, stats) = StdcmLog::list_paginated(conn, settings).await?;
+
+    Ok(Json(StdcmLogListResponse {
+        results: stdcm_logs,
+        stats,
+    }))
+}
+
+#[derive(IntoParams, Deserialize)]
+pub struct StdcmLogTraceIdParam {
```
```suggestion
struct StdcmLogTraceIdParam {
```

editoast/src/core/stdcm.rs
leovalais: https://github.com/OpenRailAssociation/osrd/pull/10212#discussion_r1907607694

editoast/src/views/timetable/stdcm.rs
leovalais: I'm not at all against having views fixtures. But since we did not have any discussion involving @woshilapin on this matter, can you at least argue about why you went this way?

editoast/openapi.yaml
hamz2a: Done.

editoast/openapi.yaml
hamz2a: Done.

editoast/src/views/stdcm_logs.rs
hamz2a: Done.

editoast/src/views/stdcm_logs.rs
hamz2a: Done.

editoast/src/views/stdcm_logs.rs
hamz2a: Done.

editoast/src/core/stdcm.rs
woshilapin: Something like `#[schema(as = StdcmRequest)]` I believe.

hamz2a: `StdcmRequest` isn’t used on the front; the main struct is `StdcmLog`. @SharglutDev, would it be okay to inline `StdcmRequest` in `StdcmLog`?

editoast/src/views/timetable/stdcm.rs
hamz2a: We have functions used in two places, so I created a file to centralize them. Normally, such functions go in `models::fixtures`, but since they depend on `axum::http::StatusCode` and `views::path::pathfinding::PathfindingResult`, I placed them in `views::fixtures` to avoid introducing dependencies between models, views, and axum. what do you think? @leovalais @woshilapin

editoast/src/models/stdcm_log.rs
hamz2a: it works, thank you!

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -3438,6 +3449,131 @@ export type SimulationResponse =
...
+    margin?:
+      | (
+          | {
+              Percentage: number;
+            }
+          | {
+              MinPer100Km: number;
+            }
+        )
+      | null;
```
SharglutDev: Is it normal that this margin is more specific than the margin in `PostTimetableByIdStdcmApiArg` ?

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -3438,6 +3449,131 @@ export type SimulationResponse =
...
+    /** Maximum departure delay in milliseconds. */
+    maximum_departure_delay: number;
+    /** Maximum run time of the simulation in milliseconds */
+    maximum_run_time: number;
+    /** List of waypoints. Each waypoint is a list of track offset. */
+    path_items: PathItem[];
```
SharglutDev: Why not use `PathfindingItem` which is more "stdcm" specific since it also has the duration and timing data ?

editoast/src/core/stdcm.rs
SharglutDev: It isn't used in the front we can manage the type safety by using `PostTimetableByIdStdcmApiResponse` and `PostTimetableByIdStdcmApiArg`. But I just thought it's a shame that we duplicate these 2 types in `postTimetableByIdStdcm` args and response and in this new `StdcmLog` when we could only have one source of truth and one type to maintain if the endpoint changes in the future.

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -3438,6 +3449,131 @@ export type SimulationResponse =
       core_error: InternalError;
       status: 'simulation_failed';
     };
+export type StdcmLog = {
```
SharglutDev, line 3581: If I understand correctly, this type differs from the one we use in `postTimetableByIdStdcm` because these are the datas editoast sends to core and they might differ from the one we use in the front.
If that's the case, would it be possible to add a comment/doc on this type to explain that because most of the front devs don't know this and might think it's strange.

front/src/common/api/generatedEditoastApi.ts
SharglutDev: It contains information about the tolerance, the time arrival, the stop duration. Don't the log should give us these informations ?

editoast/src/views/timetable/stdcm.rs
woshilapin: I'm fine with your proposition @hamz2a

front/src/common/api/generatedEditoastApi.ts
hamz2a: `StdcmLog` represents the data sent to and received from the `core` system. This format may differ from the one used in the frontend (like in `postTimetableByIdStdcm`), as it is specifically designed for communication with the `core` system. It may include extra fields, different formatting, or structures that aren't needed in the frontend.

Unfortunately, I can't add this comment to the frontend file.

front/src/common/api/generatedEditoastApi.ts
hamz2a: https://github.com/OpenRailAssociation/osrd/pull/10212#discussion_r1910152922

front/src/common/api/generatedEditoastApi.ts
hamz2a: https://github.com/OpenRailAssociation/osrd/pull/10212#discussion_r1910152922

editoast/src/core/stdcm.rs
hamz2a: https://github.com/OpenRailAssociation/osrd/pull/10212#discussion_r1910152922

editoast/src/views/stdcm_logs.rs
```diff
@@ -0,0 +1,229 @@
...
+        let mut core = core_mocking_client();
+        core.stub("/v2/stdcm")
+            .method(reqwest::Method::POST)
+            .response(StatusCode::OK)
+            .json(json!({
+                "status": "success",
+                "simulation": serde_json::to_value(simulation_response()).unwrap(),
+                "path": serde_json::to_value(pathfinding_result_success()).unwrap(),
+                "departure_time": "2024-01-02T00:00:00Z"
+            }))
```
leovalais: Can the struct representing the response be used instead (avoids `to_value` + type system)?

editoast/src/views/fixtures.rs
```diff
@@ -0,0 +1,100 @@
...
+          }
+        },
+        { "duration": 0, "location": { "trigram": "MWS", "secondary_code": "BV" } }
+      ],
+      "time_gap_after": 35000,
+      "time_gap_before": 35000
+    })
+}
+
+pub fn core_mocking_client() -> MockingClient {
```
leovalais: The name isn't specific enough

editoast/src/views/fixtures.rs
```diff
@@ -0,0 +1,100 @@
...
+    let mut core = MockingClient::new();
+    core.stub("/v2/pathfinding/blocks")
+        .method(reqwest::Method::POST)
+        .response(StatusCode::OK)
+        .json(pathfinding_result_success())
+        .finish();
+    core.stub("/v2/standalone_simulation")
+        .method(reqwest::Method::POST)
+        .response(StatusCode::OK)
+        .json(serde_json::to_value(simulation_response()).unwrap())
```
leovalais:
```suggestion
        .json(simulation_response())
```

should work?

hamz2a: Done.

editoast/src/views/fixtures.rs
hamz2a: `init_core_mocking_client` ?

editoast/src/views/timetable/stdcm.rs
leovalais: I not really satisfied with view fixtures (*here, not in general*) for two reasons (mainly):

1. Locality is important. For example, when the functions of `views::fixtures` were in `views::stdcm`, I believe things were much easier to read and explicit. Take `fn core_mocking_client() -> MockingClient`, by not being `pub`, we know it's only used in the tests module. So we immediately understand that it returns a configured mocking client for STDCM. Now that it's in a shared context (`fixtures`), it becomes named incorrectly (not specific enough), and because of it's `pub` visibility, we don't know to which extent it's used without searching for all it's references. Additionally, since they're not localized to a single `mod tests`, we don't know which fixtures are related: here `pathfinding_...` and `simulation_response` are tied together to form a basis for a STDCM call. When they were just local `fn`s, it was easy to figure that out, but know they all live freely in a shared context, that's not so obvious. I don't think we should have fixtures that are not self-contained. 
3. Before the test refactoring a year ago, we had an `src/fixtures.rs` file centralizing all shared or atomic test data. Theoretically it works, but in practice, apart from infras and rolling stocks, it was mostly a pain to use. Sharing fixtures was difficult, we often needed to change a value here and there to write a new test, but since these were shared with others, we often ended up duplicating the whole fixture. We had quite a lot of fixtures that were almost identical, but not strictly. Many of them were therefore used in only one or two tests, often in the same `mod tests`, defeating the purpose of fixtures and hindering readability.

Imo fixtures are useful in the following cases:

1. We know the data is general enough and represent a constant well-defined object (`small_infra`, `fast_rolling_stock`, ...)
2. The data is already stored in a different file (`src/tests/*.json`), so since we loose locality, we might as well make it available to every test.

I don't intent to block this PR at all, but I'm not convinced this is the right move yet, sorry.

Wdyt?

editoast/src/views/fixtures.rs
leovalais: https://github.com/OpenRailAssociation/osrd/pull/10212#discussion_r1910244205

front/src/common/api/generatedEditoastApi.ts
leovalais: Okay, by looking at the OpenAPI, it is wrong. Currently:

* `Request` represents the STDCM **view request**
* `Response` represents the STDCM **core response**
* `StdcmLog` is linked to both of them, so it's wrong

Here's my proposition. **In the OpenAPI only**:

1. inline the view request
2. put the core STDCM responses and requests in `components/schemas` as `StdcmRequest` and `StdcmResponse`
3. `$ref` them in `StdcmLog`

Is everyone ok with that? (plz :+1: and click the notification bell 🔔)

editoast/src/views/stdcm_logs.rs
hamz2a: I tried this:
```rust
            .json(crate::core::stdcm::Response::Success {
                simulation: simulation_response(),
                path: pathfinding_result_success(),
                departure_time: DateTime::from_str("2024-01-02T00:00:00Z")
                    .expect("Failed to parse datetime"),
            })
```
i got this error:
```bash
mocked response body should deserialize faultlessly: InternalError { status: 500, error_type: "editoast:coreclient:CoreResponseFormatError", context: {"msg": String("missing field `status` at line 1 column 91")}, message: "Cannot parse Core response: missing field `status` at line 1 column 91" }
```

what you think ? @leovalais

editoast/src/views/timetable/stdcm.rs
hamz2a: I see your point. I wanted to avoid duplicating the code, but I understand it can get tricky later. So, I’m fine with redefining `core_mocking_client`, `pathfinding_result_success`, and `simulation_response`.
What you think @woshilapin ?

front/src/common/api/generatedEditoastApi.ts
hamz2a: https://github.com/OpenRailAssociation/osrd/blob/hai/editoast-endpoint-to-retrieve-stdcm-payloads/editoast/src/models/stdcm_log.rs#L20
`StdcmLog` is linked only to `crate::core::stdcm::Request` and `crate::core::stdcm::Response`

leovalais: Not in the openapi

hamz2a: If I understand correctly, `ToSchema` in this [code](https://github.com/OpenRailAssociation/osrd/blob/hai/editoast-endpoint-to-retrieve-stdcm-payloads/editoast/src/models/stdcm_log.rs#L17) doesn't work. The `openapi.yml` is a generated file.

editoast/src/views/stdcm_logs.rs
leovalais: That's weird, it should be identical 🤔 

```rust
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, ToSchema)]
// We accepted the difference of memory size taken by variants
// Since there is only on success and others are error cases
#[allow(clippy::large_enum_variant)]
pub enum Response {
    Success {
        simulation: SimulationResponse,
        path: PathfindingResultSuccess,
    },
    PathNotFound,
    PreprocessingSimulationError {
        error: SimulationResponse,
    },
}
```

Is the response serialized correctly in the stubbed response?

editoast/src/views/timetable/stdcm.rs
woshilapin: @leovalais Thanks for the detail response and I see your point and do agree to a degree. I see 2 solutions (and I'd be fine with both):
- keeping the concept of fixtures, but moving them in a less general module (so, instead of `views`, moving them one level deeper in `views::stdcm` and not `pub` (they should be accessible in sub-modules if you keep them in `views::stdcm` and not `views::stdcm::tests`).
- just keep it simple and copy/paste the functions, we don't get to factorize code but I sometimes see that less as a problem for tests (for reasons @leovalais mentioned, locality and simplicity of code instead of abstractions).

editoast/src/views/stdcm_logs.rs
hamz2a: it works now 😂

front/src/common/api/generatedEditoastApi.ts
leovalais: `utoipa` doesn't care which type is used in the rust struct, it only reads the type name to generate the $ref, and the schema `Request` in the OpenAPI describes the view request, not core's one.

editoast/src/views/timetable/stdcm.rs
leovalais: About the first solution: the initial problem was that `pub` was added to the module because of the module tree (we currently have `crate::views::stdcm_logs` and `crate::views::timetable::stdcm`). I would have been fine with the addition of a `pub(in crate::views::stdcm)` if we had that module. I don't see `stdcm_logs` fitting as a submodule of `timetable` though. 🤷

Maybe I can see another solution: we could extend the `CoreClient` to provide an `stdcm_on_small_infra(&mut self, rs_id: i64) -> StdcmFixture`. `StdcmFixture` would gather all information related to the STDCM call and the stubbed core API calls. Thus making the fixture “atomic“. But that's quite a lot of work for just two usages. And we loose locality.

hamz2a: As we discussed, I duplicated the code and fixed the schemas.

front/src/common/api/generatedEditoastApi.ts
hamz2a: https://github.com/OpenRailAssociation/osrd/pull/10212#discussion_r1913306144

editoast/src/views/fixtures.rs
hamz2a: https://github.com/OpenRailAssociation/osrd/pull/10212#discussion_r1913306144

editoast/src/views/timetable/stdcm/request.rs
leovalais, line 1: I don't think that's necessary to change the visibility here? 🤔

leovalais (commented):
LGTM, thank you for all the changes!

editoast/src/views/timetable/stdcm/request.rs
hamz2a: I got this error:
```bash
error[E0603]: struct `PathfindingItem` is private
   --> src/views/stdcm_logs.rs:154:50
    |
154 |     use crate::views::timetable::stdcm::request::PathfindingItem;
    |                                                  ^^^^^^^^^^^^^^^ private struct
    |
note: the struct `PathfindingItem` is defined here
   --> src/views/timetable/stdcm/request.rs:35:1
    |
35  | pub(super) struct PathfindingItem {
```

leovalais: Ha, the tests... Ok nevermind. Maybe we should revise the module tree for stdcm features. We'll see later

SharglutDev (approved):
Lgtm

leovalais:
While trying to test the new endpoints I cam across a few bugs 🥲

With authorization disabled: 
image

With authorization enabled, the first request works, but its trace id is `00000000....`.
The second request fails:
Capture d’écran 2025-01-14 à 09 59 09

Steps to reproduce:

1. (I loaded a new DB dump)
2. Authorization enabled
3. Open the frontend
4. Give mock.identity the role Superuser
5. LMR request
6. The log appear on DB with trace-id 00000000000...
7. Another LMR request
8. Contraint violation

editoast/src/views/stdcm_logs.rs
```diff
@@ -0,0 +1,460 @@
...
+        enable_telemetry: bool,
+    }
+
+    struct TestContextOutput {
+        app: TestApp,
+        user: UserInfo,
+        trace_id: String,
+    }
+
+    async fn prepare_test_context(config: TestContextConfig) -> TestContextOutput {
```
woshilapin: Not a huge fan of this "function does it all", especially the not so obvious way it creates `App` and returns it. An idea would just be to separate things a bit by having a `init_app` then a `prepare_test_context` (and maybe name it better, something like `execute_stdcm_request`?).

```rust
struct TestContextOutput {
  user: UserInfo,
  trace_id: String,
}
fn init_app(config: &TestContextConfig) -> TestApp {
  /* ... */
}
async fn execute_stdcm_request(app: &TestApp, config: &TestContextConfig) -> TestContextOutput {
  /* ... */
}
#[rstest]
async fn list_stdcm_logs_return_success() {
  let config = TestContextConfig {
    disable_superuser: false,
    enable_authorization: true,
    enable_stdcm_log: true,
    enable_telemetry: true,
  };
  let app = init_app(&config);
  let TestContextOutput {
    user,
    trace_id,
  } = execute_stdcm_request(&app, &config).await;
  /* ... */
}
```

woshilapin (approved):
Really nice and exhaustive suite of tests. Thank you.

editoast/src/views/stdcm_logs.rs
hamz2a: Done.

editoast/src/client/runserver.rs
```diff
@@ -43,6 +43,10 @@ pub struct RunserverArgs {
     // only recieve 401 responses.
     #[clap(long, env = "EDITOAST_DISABLE_AUTHORIZATION", default_value_t = true)]
     disable_authorization: bool,
+    /// If this option is set, logging for the STDCM will be enabled.
+    /// When enabled, relevant logs will be captured to aid in debugging and monitoring.
+    #[clap(long, env = "ENABLE_STDCM_LOG", default_value_t = true)]
+    enable_stdcm_log: bool,
```
leovalais: nit: would

```suggestion
    enable_stdcm_logging: bool,
```

or

```suggestion
    log_stdcm: bool,
```

be more correct?

editoast/src/views/stdcm_logs.rs
```diff
@@ -0,0 +1,445 @@
...
+        .check_roles([BuiltinRole::Superuser].into())
+        .await
+        .map_err(AuthorizationError::AuthError)?;
+    if !authorized {
+        return Err(AuthorizationError::Forbidden.into());
+    }
+
+    let conn = &mut db_pool.get().await?;
+
+    let stdcm_log = StdcmLog::retrieve_or_fail(conn, Some(trace_id.clone()), || {
```
```suggestion
    let stdcm_log = StdcmLog::retrieve_or_fail(conn, Some(trace_id), || {
```

leovalais (approved):
Thanks a lot for all the changes! I'll test it a bit to make sure I didn't miss anything but things look fine to me :)

editoast/src/views/stdcm_logs.rs
hamz2a: We need to call `clone()`, as `trace_id` is used in the next line.

editoast/src/client/runserver.rs
hamz2a: Done.

editoast/src/views/stdcm_logs.rs
leovalais: Oh yeah the error I missed it, my bad

leovalais:
Sorry found a new problem:

```
❯ xh :8090/stdcm_logs x-remote-user-identity:'mock.identity/Example User'
HTTP/1.1 400 Bad Request
Access-Control-Allow-Origin: *
Content-Length: 190
Content-Type: application/json
Date: Thu, 16 Jan 2025 11:31:29 GMT
Vary: origin, access-control-request-method, access-control-request-headers

{
    "status": 400,
    "type": "editoast:pagination:InvalidPageSize",
    "context": {
        "max_page_size": 10,
        "provided_page_size": 25
    },
    "message": "Invalid page size (25), expected an integer 0 < page_size <= 10"
}
```

hamz2a:
> Sorry found a new problem:
> 
> ```
> ❯ xh :8090/stdcm_logs x-remote-user-identity:'mock.identity/Example User'
> HTTP/1.1 400 Bad Request
> Access-Control-Allow-Origin: *
> Content-Length: 190
> Content-Type: application/json
> Date: Thu, 16 Jan 2025 11:31:29 GMT
> Vary: origin, access-control-request-method, access-control-request-headers
> 
> {
>     "status": 400,
>     "type": "editoast:pagination:InvalidPageSize",
>     "context": {
>         "max_page_size": 10,
>         "provided_page_size": 25
>     },
>     "message": "Invalid page size (25), expected an integer 0 < page_size <= 10"
> }
> ```

The error occurs because the default value for [pagination is 25](https://github.com/OpenRailAssociation/osrd/blob/hai/editoast-endpoint-to-retrieve-stdcm-payloads/editoast/src/views/pagination.rs#L19), but for this endpoint, we've set it to [10 as specified in the pagination validation](https://github.com/OpenRailAssociation/osrd/blob/hai/editoast-endpoint-to-retrieve-stdcm-payloads/editoast/src/views/stdcm_logs.rs#L73).

hamz2a:
As we discussed, we'll keep the `page_size` at 25, but the endpoint is less verbose, it now returns results like: `[{'id': 1, 'trace_id': 'AZERT'}, {'id': 2, 'trace_id': null}]`.
I've also updated the endpoint to allow access to the logs by`id`.

leovalais (approved):
Thanks and sorry about all the back and forth! I can see a few things to improve (especially the `/stdcm_log` vs `/stdcm_logS`, these mistakes are hard to find 😨), but that's for another PR if any. Let's merge this :)

PR #10215, opened 2025-01-01
editoast: extract osm_to_railjson in separate binary

Tristramg:
This is more to open a discussion:

As compile time are unbearable, we could split the editoast binary into smaller ones. As we already started separating crates, this could be quite simple.

In this pull request, osm-to-railjson in separated.

It now must be called as `cargo run -p osm_to_railjson`

The improvement are not spectacular:

```
Now:
- 541 units
- total build time: 175.5s
- editoast bin: 109s
- minor change build time: 25s
- binary size (debug): 312Mb

Before:
- 561 units
- total build time: 182.4s
- editoast bin: 109s
- minor change build time: 26secs
- binary size (debug): 314Mb
```

I’m not sure if it would make sens to extract the other commands that are more close to the editoast core?

### DIFF
--- a/editoast/Cargo.lock
+++ b/editoast/Cargo.lock
@@ -1324,7 +1324,6 @@ dependencies = [
  "opentelemetry-semantic-conventions",
  "opentelemetry_sdk",
  "ordered-float",
- "osm_to_railjson",
  "paste",
  "pathfinding",
  "postgis_diesel",
@@ -3066,6 +3065,7 @@ dependencies = [
 name = "osm_to_railjson"
 version = "0.1.0"
 dependencies = [
+ "clap",
  "editoast_schemas",
  "geo-types",
  "geos",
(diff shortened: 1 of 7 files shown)


woshilapin (approved):
We might need an exhaustive search for `osm_to_railjson` in our repos to see if it's used somewhere (documentation, CI scripts, deployment, charts, etc.). But in general, I'd say it'd be nice that the binaries are separated, it might indeed help a bit the compilation times (especially not wasting time compiling dependencies that are not used.

Castavo:
Isn't splitting the binary for `osm-to-railjson` especially beneficial to `osm-to-railjson` ? This binary alone must be much faster to compile now, isn't it ?

Tristramg:
Il did a grep on my `sncf` folder with different projects for osm-to-railjson and osm_to_railjson and didn’t found anything. Maybe it’s used somewhere else I’m not aware of? I don’t think as it’s quite a nice use

@Castavo indeed :) but the much more people working on the general editoast than osm-to-railjson that hasn’t changed in over a year

leovalais (approved):
Thanks!

PR #10218, opened 2025-01-02
front: rename and refacto StdcmInputVia

clarani:


front/src/applications/stdcm/components/StdcmForm/StopDurationInput.tsx
```diff
@@ -0,0 +1,79 @@
...
+          id: pathStep.id,
+          updates: { stopFor: newStopDuration },
+        })
+      );
+    }
+  }, [debouncedStopDuration]);
+
+  if (pathStep.stopType === StdcmStopTypes.PASSAGE_TIME) {
+    return null;
+  }
```
Math-R: I think it would be better to don't let the component decide to render or not but keep it in the parent to prevent unnecessary mounting

front/src/applications/stdcm/components/StdcmForm/StopDurationInput.tsx
```diff
@@ -0,0 +1,79 @@
...
+    <div className="stop-time">
+      <Input
+        id="stdcm-via-stop-time"
+        type="text"
+        label={t('trainPath.stopFor')}
+        onChange={(e) => {
+          // TODO: Find a better way to prevent user from entering decimal values
+          const value = e.target.value.replace(/[\D.,]/g, '');
+          setStopDuration(value);
+        }}
```
Math-R: Maybe it would be better to don't automatically replace value with a regex and warn the user that the value can't be used.
Because if I put `3.5`, it compute `35` and if i don't pay attention, i will launch the simulation using a `35` minutes stop

Math-R: So you can make a handleChange function, that will turn a flag and display a warning if the value don't respect the expexted value

Math-R: Another solution would be to round up or down the value to keep minutes only, but this need to be discuss

clarani: This portion of code is not related to my PR, can we discuss this later ? :/
We won't forget it since there is already a TODO comment above the problematic lines

front/src/applications/stdcm/components/StdcmForm/StopDurationInput.tsx
clarani: Done !

front/src/applications/stdcm/components/StdcmForm/StopDurationInput.tsx
thibautsailly: Following Mathieu's invitation, here is my user experience comment. This behaviour is problematic as the change is difficult to spot (a little larger letter spacing and a dot) and could affect results.
That said, we haven't witnessed users entering decimal values during our tests.
As there would be no way for us to know if users get this problem or not, I'm inviting POs to prioritize the issue as high, especially since we currently don't have the possibility to cancel computations.

clarani: The value is now rounded. It's a quick fix, this will need to be discussed again with the POs

SharglutDev: > especially since we currently don't have the possibility to cancel computations.

We can since a few weeks :)

Math-R (approved):
lgtm

PR #10222, opened 2025-01-02
front: clean allowanceValue type and related code

clarani:
to merge after https://github.com/OpenRailAssociation/osrd/pull/10203

Remove some dead code and simplify code in StdcmAllowance

### DIFF
--- a/front/src/modules/stdcmAllowances/components/StdcmAllowances.tsx
+++ b/front/src/modules/stdcmAllowances/components/StdcmAllowances.tsx
@@ -1,42 +1,44 @@
 import { useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
 
-import type { AllowanceValue } from 'applications/stdcm/types';
+import { MarginType } from 'applications/stdcm/types';
 import InputGroupSNCF from 'common/BootstrapSNCF/InputGroupSNCF';
 import type { InputGroupSNCFValue } from 'common/BootstrapSNCF/InputGroupSNCF';
 import InputSNCF from 'common/BootstrapSNCF/InputSNCF';
-import { ALLOWANCE_UNITS_KEYS } from 'modules/stdcmAllowances/allowancesConsts';
 import {
   updateGridMarginAfter,
   updateGridMarginBefore,
   updateStandardAllowance,
 } from 'reducers/osrdconf/stdcmConf';
 import { getMargins } from 'reducers/osrdconf/stdcmConf/selectors';
-import type { StandardAllowance } from 'reducers/osrdconf/types';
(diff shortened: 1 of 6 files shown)


front/src/reducers/osrdconf/types.ts
SharglutDev, line 1: We also need to update in `stdcmConf/index.ts`

front/src/modules/stdcmAllowances/components/StdcmAllowances.tsx
```diff
@@ -92,10 +94,10 @@ const StdcmAllowances = ({ disabled = false }: { disabled?: boolean }) => {
         <label htmlFor="standardAllowanceTypeSelect">{t('allowances:standardAllowance')}</label>
         <InputGroupSNCF
           id="standardAllowanceTypeSelect"
-          options={standardAllowanceTypes}
+          options={STANDARD_MARGIN_UNITS}
           onChange={onchangeType}
           currentValue={{
-            unit: standardAllowance?.type || 'percentage',
+            unit: standardAllowance?.type || MarginType.PERCENTAGE,
```
SharglutDev, line 100: Can the fallback ever happen since the store initialize an unit ?

clarani: I think we need to keep it, because we can set the whole standardAllowance to `undefined` :/

Can we keep this behvaviour for now ?  I think we should check with the POs whether setting `standardAllowance` to `undefined` is something we want to allow before making any further changes

SharglutDev: Sure :)

front/src/modules/stdcmAllowances/components/StdcmAllowances.tsx
```diff
@@ -1,42 +1,44 @@
...
-      label: ALLOWANCE_UNITS_KEYS.time_per_distance,
-    },
-  ];
 
-  const onchangeType = <U extends string>(newTypeValue: InputGroupSNCFValue<U>) => {
-    const processedType: StandardAllowance = {
-      type: newTypeValue.unit as AllowanceValue['value_type'],
-      value: newTypeValue.value === undefined ? undefined : Math.abs(newTypeValue.value),
-    };
+  const onchangeType = (newTypeValue: InputGroupSNCFValue<MarginType>) => {
```
SharglutDev: nit :)
```suggestion
  const onChangeType = (newTypeValue: InputGroupSNCFValue<MarginType>) => {
```

SharglutDev (approved):
Lgtm 🧹

PR #10224, opened 2025-01-03
front: display rs name instead of rs reference in StdcmReportSheet

### LINKED ISSUE #10152: LMR : The name of the rolling stock in the simulation report is not correct
### What happened?

When you launch a simulation with a rolling stock, the name that appears in the simulation report is not correct.

Example, if I launch a simulation wiith a 27000UM, I have this in the simulation report :
Image

Maybe it's the "reference" of the rolling stock that appears here. It should be the "name" (this statement has to be challenged).

The information "US" or "UM" must be displayed.

Launch a LMR simulation with the 27000UM and inspect the simulation report.

Recette (SNCF)

Firefox

6580cfe

clarani:
closes #10152


achrafmohye (approved):
LGMT 👍

SharglutDev (approved):
Lgtm and tested

PR #10225, opened 2025-01-03
github: group osrd-ui dependabot upgrades

emersion:
osrd-ui has an unusual versioning scheme: it uses 0.0.x and all releases may contain breaking changes.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -36,6 +36,9 @@ updates:
       nivo:
         patterns:
           - "@nivo/*"
+      osrd-ui:
+        patterns:
+          - "@osrd-project/ui-*"
       rjsf:
         patterns:
           - "@rjsf/*"


SharglutDev:
This still allows dependabot to bump osrd-ui ? I'm afraid some "ghost" changes which doesn't break the CI - like one in v0.0.58 - might be forgotten. Or should we still allow it and be more rigorous for these kind of changes (like always creating a separated issue from the bump) ?

emersion:
dependabot will create a separate PR with all osrd-ui upgrades. Then it's up to the reviewer to check the changelog and make sure the breaking changes are all accounted for. I don't think there's anything really osrd-ui-specific here, any dependency upgrade can pass CI and still break stuff :P

SharglutDev (approved):

SharglutDev:
It's kind of an edge case anyway as most of the time we bump `osrd-ui` ourself just after a new release

flomonster (approved):
LGTM

Khoyo (approved):
Reviewed the syntax, no opinion on this

PR #10227, opened 2025-01-03
Filter TowedRollingStocks to include only allowed values

### LINKED ISSUE #10120: stdcm: front: filter towed rolling stocks to keep freight ones
# Description

Only the following towed rolling stocks should be kept because the others aren't relevant for freight trains:

```
AFAMOD
AFAMOD2
MA
MAL100
MALOUR
MAVIDE
ME100
ME120
ME140
ME200SER
MEPA
TRAINLOC
```

# AC

- [ ] The towed rolling stocks in stdcm are filtered using the list above.
- [ ] Ideally, all the towed rolling stocks are kept when using the debugging mode (optionnal)

achrafmohye:
Closes #10120

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
@@ -4,6 +4,7 @@ import { Input, ComboBox } from '@osrd-project/ui-core';
 import { useTranslation } from 'react-i18next';
 
 import useStdcmTowedRollingStock from 'applications/stdcm/hooks/useStdcmTowedRollingStock';
+import type { ConsistErrors } from 'applications/stdcm/types';
 import {
   CONSIST_MAX_SPEED_MIN,
   CONSIST_TOTAL_LENGTH_MAX,
@@ -23,7 +24,6 @@ import { kgToT, kmhToMs, msToKmh } from 'utils/physics';
 
 import StdcmCard from './StdcmCard';
 import useStdcmConsist from '../../hooks/useStdcmConsist';
-import type { StdcmConfigCardProps } from '../../types';
 
 const ConsistCardTitle = ({
   rollingStock,
@@ -39,7 +39,13 @@ const ConsistCardTitle = ({
   );
 };
 
-const StdcmConsist = ({ consistErrors = {}, disabled = false }: StdcmConfigCardProps) => {
+export type StdcmConsistProps = {
+  isDebugMode: boolean;
+  disabled?: boolean;
+  consistErrors?: ConsistErrors;
+};
+
+const StdcmConsist = ({ isDebugMode, consistErrors = {}, disabled = false }: StdcmConsistProps) => {
   const { t } = useTranslation('stdcm');
   const { speedLimitByTag, speedLimitsByTags, dispatchUpdateSpeedLimitByTag } =
     useStoreDataForSpeedLimitByTagSelector({ isStdcm: true });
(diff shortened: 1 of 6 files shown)


front/src/applications/stdcm/consts.ts
SharglutDev, line 1: For consistency, it would be nice to have this list at the same place as `FREIGHT_ROLLING_STOCKS` in `assets/rollingStock`

front/src/modules/towedRollingStock/hooks/useFilterTowedRollingStock.ts
```diff
@@ -40,14 +41,22 @@ const useFilterTowedRollingStock = () => {
...
     pageSize: 50,
   });
 
+  const allTowedRollingStocks = useMemo(
+    () =>
+      fetchedTowedRollingStocks.filter((stock) =>
+        ALLOWED_TOWED_ROLLING_STOCKS.includes(stock.name)
+      ),
+    [fetchedTowedRollingStocks]
+  );
```
SharglutDev: It's only an optional AC but in `StdcmConfig`, we have access to `isDebugMode`. 
We could pass it to `StdcmConsist` and then to this hook so we can still have access to the full list of towed rs when we are in debug mode.

Uriel-Sautron (commented):
Good work, tested.
Agree with @SharglutDev comments

front/src/modules/towedRollingStock/hooks/useFilterTowedRollingStock.ts
achrafmohye: Done

front/src/applications/stdcm/types.ts
```diff
@@ -148,6 +148,11 @@ export type StdcmConfigCardProps = {
   disabled?: boolean;
 };
 
+export type StdcmConsistProps = {
+  isDebugMode: boolean;
+  disabled?: boolean;
+};
+
```
SharglutDev: For props, we usually keep the type just above the related component (except some rare exceptions in stdcm)

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
SharglutDev, line 1: In debug mode, if I select a towed rs from the debug list (`CORAIL` for example), and then I switch back to "none debug mode", I still have CORAIL selected. We should do something about that.
What do you think @Arthur-Lefebvre @Melissa266 @thibautsailly ?

Arthur-Lefebvre: You mean that it results with a selected RS that couldn't be selected otherwise ?

SharglutDev: Exactly. Shouldn't happen in real life because users won't have access to debug mode, but it could be nice to handle this case.

Arthur-Lefebvre: Since the debug mode is available only for restricted users, it seems not to be a great problem, but if it is simple to handle I thiink it's better.

Allowing these RS if we stay in debug mode would be great though.

SharglutDev: After some discussions with @thibautsailly, he proposed to clear the input for this kind of use case. The combobx shouldn't display a value which isn't in the provided list. The osrd-ui bug has been reported [here](https://github.com/OpenRailAssociation/osrd-ui/issues/802).

SharglutDev (approved):
Lgtm and tested

Maymanaf (approved):
LGTM for e2e-tests

PR #10230, opened 2025-01-03
front: drop formatDurationAsISO8601()

emersion:
See https://github.com/OpenRailAssociation/osrd/issues/8816

_See individual commits._

Depends on https://github.com/OpenRailAssociation/osrd/pull/9934

### DIFF
--- a/front/src/modules/timesStops/TimesStopsInput.tsx
+++ b/front/src/modules/timesStops/TimesStopsInput.tsx
@@ -12,6 +12,8 @@ import type { SuggestedOP } from 'modules/trainschedule/components/ManageTrainSc
 import type { OperationalStudiesConfSliceActions } from 'reducers/osrdconf/operationalStudiesConf';
 import type { PathStep } from 'reducers/osrdconf/types';
 import { useAppDispatch } from 'store';
+import { Duration } from 'utils/duration';
+import { sec2ms } from 'utils/timeManipulation';
 
 import {
   durationSinceStartTime,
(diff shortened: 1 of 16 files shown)


kmer2016 (approved):
LGTM !

front/src/modules/timesStops/TimesStopsInput.tsx
```diff
@@ -119,11 +121,12 @@ const TimesStopsInput = ({
           (row, index) =>
             !isEqual(normalizeNullablesInRow(row), normalizeNullablesInRow(rows[index]))
         )
-        .map(({ shortSlipDistance, onStopSignal, arrival, departure, ...row }) => ({
+        .map(({ shortSlipDistance, onStopSignal, arrival, departure, stopFor, ...row }) => ({
           ...row,
           arrival: durationSinceStartTime(startTime, arrival),
           departure: durationSinceStartTime(startTime, departure),
           receptionSignal: onStopSignalToReceptionSignal(onStopSignal, shortSlipDistance),
+          stopFor: stopFor ? new Duration(sec2ms(Number(stopFor))) : null,
```
anisometropie, line 129: We could have a static `Duration.fromSeconds`
this seems like an operation we would need to do often

front/src/reducers/index.ts
```diff
@@ -82,10 +83,25 @@ const operationalStudiesDateTransform = createTransform(
...
       if (!pathStep) return null;
-      return { ...pathStep, arrival: pathStep.arrival ? Duration.parse(pathStep.arrival) : null };
+
+      let stopFor: Duration | null = null;
+      if (pathStep.stopFor) {
+        if (pathStep.stopFor.startsWith('P')) {
+          stopFor = Duration.parse(pathStep.stopFor);
+        } else {
+          // Compatibility with older store format
+          stopFor = new Duration(sec2ms(Number(stopFor)));
```
anisometropie, line 96: if you go with the `fromSeconds` idea, here too

front/src/modules/timesStops/TimesStopsInput.tsx
emersion: Yeah, I'd prefer to do this as a second step, because I get a feeling we'll have a lot less of these once everything is converted to `Duration`.

Perhaps by making the `Duration` constructor take an object e.g. `{ seconds?: number; milliseconds?: number }`, because currently it's not super from the caller side that the number passed to the constructor is a number of milliseconds.

anisometropie (approved):
Neat

front/src/modules/timesStops/TimesStopsInput.tsx
anisometropie: Fine by me !

PR #10235, opened 2025-01-06
front: display correct ch code after linked train selection

### LINKED ISSUE #10033: stdcm: selecting a linked train doesn't set the ch code
### What happened?

Selecting a linked path only sets the CI code, and not the CH code:

Here I select a train which arrives at Lyon-Guillotière EH, but in stdcm it sets the origin to Lyon-Guillotière 27

https://github.com/user-attachments/assets/73edc7e1-4402-437d-acc0-f15db620c56f

I expect the CH in stdcm to be set to the same CH of the linked train.

1. Pick a train you know the CI/CH code of the destination in the timetable used for stdcm
2. In the stdcm form, pick it
3. Witness the destination being set to the wrong CH

Recette (SNCF)

Firefox

9e75ce2

SarahBellaha:
Closes #10033


Uriel-Sautron (approved):
LGTM, tested !!!

kmer2016 (approved):
LGTM !

PR #10236, opened 2025-01-06
front: add default CH code when not provided during open data import

### LINKED ISSUE #10159: Open data import picks random CH codes instead of BV/00
### What happened?

I imported trains between Nantes and Cholet using the open data import, but all the waypoints (including origin and destination) are with a random CH and not the BV/00.

I expect the import to pick either the right CH if we have the data, or BV/00 as a default

1. Import trains between Nantes and Cholet
2. Notice that the origin, destination, and stops are not at BV/00

Other

Google Chrome

94d3a84

SarahBellaha:
Closes #10159

### DIFF
--- a/front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleTrainsList.tsx
+++ b/front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleTrainsList.tsx
@@ -1,18 +1,20 @@
-import { useMemo } from 'react';
+import { useCallback, useMemo } from 'react';
 
 import { Rocket } from '@osrd-project/ui-icons';
 import type { TFunction } from 'i18next';
 import { keyBy } from 'lodash';
 import { useTranslation } from 'react-i18next';
 
-import type { ImportedTrainSchedule } from 'applications/operationalStudies/types';
+import type { ImportedTrainSchedule, Step } from 'applications/operationalStudies/types';
 import {
   osrdEditoastApi,
   type LightRollingStockWithLiveries,
+  type SearchResultItemOperationalPoint,
   type TrainScheduleBase,
   type TrainScheduleResult,
 } from 'common/api/osrdEditoastApi';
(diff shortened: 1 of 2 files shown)


front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleTrainsList.tsx
```diff
@@ -72,21 +80,50 @@ const ImportTrainScheduleTrainsList = ({
...
+      const stepsQuery = ['or', ...steps.map((step) => ['=', ['uic'], Number(step.uic)])];
+      const mainChCodeConstraint = [
+        'or',
+        ...MAIN_OP_CH_CODES.map((chCode) => ['=', ['ch'], chCode]),
+      ];
+
+      try {
+        const payloadSteps = {
+          object: 'operationalpoint',
+          query: ['and', stepsQuery, mainChCodeConstraint],
```
emersion, line 94: I think we also need to filter by infra ID here?

emersion (commented):
Hm, I'm not a fan of looking at the infrastructure during the import step. Right now the import is completely disconnected from the infrastructure (much like etl-basic) and that's a very good property IMHO.

emersion:
FWIW, when the CH is unspecified in a path step, the backend is supposed to automatically pick BV/00.

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleTrainsList.tsx
SarahBellaha: As we retrieve information from the OP with its UIC code, which is normally unique regardless of the infra (thanks to the "country code"), i don't think it's necessary here ?

emersion: A test infra created by a random person might have a BV CH which doesn't really exist in the real infra the user is interested in.

PR #10238, opened 2025-01-06
ci: bump python version

ghost:


Khoyo (approved):

emersion (approved):

emersion (commented):
Hm, actually, seems like there are 2 more of these in this file?

ghost:
> Hm, actually, seems like there are 2 more of these in this file?

Yes, currently trying things randomly

.github/workflows/build.yml
```diff
@@ -230,6 +230,11 @@ jobs:
       - name: Generate railjson
         run: |
           mkdir /tmp/generated_infras
+          pwd
+          file tests/infra-scripts/circle_infra.py
+          stat tests/infra-scripts/circle_infra.py
+          ls -lah tests/infra-scripts/circle_infra.py
+          poetry -C python/railjson_generator run python -c 'from pathlib import Path; print(Path.cwd())'
```
emersion: We should probably remove these?

.github/workflows/build.yml
```diff
@@ -215,12 +215,12 @@ jobs:
...
       - uses: actions/checkout@v4
       - name: Install poetry
-        run: pipx install poetry
+        run: pipx install 'poetry<2.0'
 
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.9"
+          python-version: "3.11"
```
emersion, line 223: If we bump the python version, can we do it everywhere in this file? Alternatively, we could choose not to bump.

.github/workflows/build.yml
```diff
@@ -389,11 +389,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Install poetry
-        run: pipx install poetry
+        run: pipx install 'poetry<2.0'
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.9"
+          python-version: "3.11"
```
woshilapin, line 396: When updating this, it would be nice to also update [this](https://github.com/OpenRailAssociation/osrd/blob/dev/python/railjson_generator/.python-version). That said, I see that [`actions/setup-python` supports reading `.python-version`](https://github.com/marketplace/actions/setup-python#basic-usage), maybe we can leave the version in one place only?

PR #10241, opened 2025-01-06
front: disable get the simulation button when infra is not loaded

### LINKED ISSUE #9452: STDCM:  "Get the simulation" button is active when infra is not loaded
### What happened?

The second part of this ticket is not done :  [#9299](https://github.com/OpenRailAssociation/osrd/issues/9299)

Image

The button should be disabled

1. Go to STDCM

Recette (SNCF)

Firefox, Google Chrome

7451939

theocrsb:
closes #9452


SharglutDev (approved):
Lgtm and tested

PR #10242, opened 2025-01-06
front: fix missing await statements for asynchronous calls

Maymanaf:


front/.eslintrc
```diff
@@ -154,6 +154,12 @@
       "rules": {
         "@typescript-eslint/no-explicit-any": ["error", { "fixToUnknown": true }]
       }
+    },
+    {
+      "files": ["tests/**/*.ts", "tests/**/*.tsx"],
+      "rules": {
+        "@typescript-eslint/no-floating-promises": "error"
```
emersion, line 161: Note, ideally this rule would be enabled for all files (ref https://github.com/OpenRailAssociation/osrd/pull/8772).

Maymanaf: If it's planned soon, I can close this PR, as I noticed there are many errors to address if this rule is applied to all files. Alternatively, we could activate it partially

Yohh: @emersion is it ok for you if we merge it? it can be set for all files in another PR since you have it in your checklist

emersion: Sure

Yohh (approved):

SharglutDev (approved):

PR #10264, opened 2025-01-07
front: display duration time only for requested pathsteps

### LINKED ISSUE #9981: Ops considered as stops when they aren't in stdcm result table
### What happened?

Image
Rennes BD is not among the vias in the form but he still gets a duration as if it was a stop.

This is because in `getStopDurationBetweenTwoPositions` function, we check only if an op position on path is the same as one from the simulation results but not if it's actually a path step or not.

Only the steps in the form should have special columns filled.

1. Go to stdcm
2. Launch a simulation for a Brest > Rennes > Strasbourg and add a stop duration to Rennes.
3. Check the result table.

Local

Firefox

.

### LINKED ISSUE #10012: LMR : Application date in the simulation report sheet appears in french when the document is in english
### What happened?

When you download the report sheet in english, the application date is written in french.

Date must be in english

Reproduce the steps in the video.

https://github.com/user-attachments/assets/7e2af9ef-c99d-47f3-9d04-12061959a790

Recette (SNCF)

Firefox

94d3a84

SarahBellaha:
Closes #9981

### DIFF
--- a/front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
+++ b/front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
@@ -64,6 +64,7 @@ const StcdmResults = ({
     return getOperationalPointsWithTimes(
       outputs.pathProperties?.suggestedOperationalPoints || [],
       outputs.results.simulation,
+      outputs.results.simulationPathSteps,
       outputs.results.departure_time
     );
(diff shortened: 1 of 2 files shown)


Caracol3 (approved):
lgtm !

SharglutDev (approved):
Lgtm and tested :)

PR #10266, opened 2025-01-07
front:  fix adding intermediate points hides the result table and result map

### LINKED ISSUE #10204: stdcm: adding intermediate points hides the result table and result map
### What happened?

https://github.com/user-attachments/assets/f52deb72-560e-4d5b-86df-c1f0534e6221

When adding intermediate points in a new request, the result table of previous requests as well as the accompanying map are hidden

The table and the map should stay visible with any number of intermediate points

1. go to stdcm
2. get a first simulation
3. look at the result table
4. start a new simulation with more intermdiate points
5. try to look at the result table or map

Recette (SNCF)

Firefox

dev bf0ab67

kmer2016:
close #10204

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmVias.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmVias.tsx
@@ -77,9 +77,9 @@ const StdcmVias = ({ disabled = false }: StdcmConfigCardProps) => {
 
     const cancelListener = () => cancelAnimationFrame(requestId);
 
-    newElement.addEventListener('animationend', cancelListener);
+    newElement.parentElement!.addEventListener('animationend', cancelListener);
     return () => {
-      newElement.removeEventListener('animationend', cancelListener);
+      newElement.parentElement!.removeEventListener('animationend', cancelListener);
       cancelListener();
     };
   }, [newIntermediateOpIndex]);


SharglutDev (approved):
Lgtm and tested, good job !

front/src/applications/stdcm/components/StdcmForm/StdcmVias.tsx
```diff
@@ -77,10 +77,9 @@ const StdcmVias = ({ disabled = false }: StdcmConfigCardProps) => {
 
     const cancelListener = () => cancelAnimationFrame(requestId);
 
-    newElement.addEventListener('animationend', cancelListener);
+    newElement.parentElement!.addEventListener('animationend', cancelListener);
     return () => {
-      newElement.removeEventListener('animationend', cancelListener);
-      cancelListener();
```
emersion, line 83: Hm, I think we should still cancel the requested animation frame here?

SharglutDev: Oh right, missed that 🫤

kmer2016: Oh, mee too, thank you ! :pray:

emersion (approved):

Issue #10144, opened 2024-12-19
stdcm: map shakes when adjusting time


Castavo:
### What happened?

When clicking on the time adjusting buttons of the form (among other inputs), the map "shakes" I guess it's reloading.

https://github.com/user-attachments/assets/d656b357-44d2-4e48-aa33-9f9fa9b9b719

### What did you expect to happen?

I expect the map to either stay stable or to not reload at all when editing the time component

### How can we reproduce it (as minimally and precisely as possible)?

1. Start a STDCM request
2. Fill in the origin
3. Fill in a waypoint / the destination (in order to have two pins placed on the map)
4. Now adjust the time component as shown on video
5. Witness

### On which environments the bug occurs?

Recette (SNCF)

### On which browser the bug occurs?

Firefox

### OSRD version (top right corner `Account` button > `Informations`)

6580cfe

theocrsb:
I can't reproduce the bug. Is it still there?

Castavo:
This might be obscured by #10240

sim51:
The map is recenter automatically when a user changes : 
- traction engine 
- time
- date 
- tolerance
- A CH code
- a CI
-  ...

Question : does the map should be automatically recenter on the "path" when the form's modifications don't impact it ?

sim51:
I respond to my question : yes we need to recenter it (except for the time, date & tolerance).
When an input is changed, we call the API to find a path between the origin/destination, and the found path can be completely different than the previous one, and so the BBOX can changed. 

I can reproduce  it with : 

* engine : 26000US - BB 26000
* Origin: valence 
* Destination: Miramas
* via : orange

Bug explanation : 
- when we change a field, a call is made to find a path (see `useStaticPathfinding`).  
- during this process the pathfinding is set to `undefined` before to be set it with the API return
- For each new instance of the pathfinding (even undefined), the map is redraw and the BBOX computed. And if the path is `undefined`, the BBOX is computed with the `markersInfo`

So when we change the time field, we have one map render with a bbox for the marker, and a second later with pathfinding bbox.  The blink effect comes from those 2 renders which can be different.

thibautsailly:
This blinking doesn't scream "quality". Ideally, there should be a map state of "computing new path", and a transition between the two displays rather than a sudden move when the new path becomes available.
As a first step we can prevent it from happening when time related data is edited, and improve upon it later.

SharglutDev:
[...]

PR #10268, opened 2025-01-07
front: enhance viewport computation function

### LINKED ISSUE #10240: stdcm: map isn't fitted to the waypoints asked
### What happened?

When selecting waypoints (including start and stop), the map window isn't fitted to include the waypoints asked.

https://github.com/user-attachments/assets/04b47f3c-6552-4a57-a3f3-a3e35fad9290

The map window is expected to include all waypoints when they're found of the map (seen with @axrolld)

1. Go to the stdcm page
2. Fill start
3. Fill stop so it's outside of the map window
4. Witness that the map doesn't move

Recette (SNCF)

Firefox

bf0ab67

sim51:
Adding optional parameters on function `computeBBoxViewport` to take care of the map height/width

Related to #10240 & #10144

NOTE: I also increased the padding to be bigger than the button on the map (it avoid to be hidden by them).

### DIFF
--- a/front/src/common/Map/WarpedMap/core/helpers.ts
+++ b/front/src/common/Map/WarpedMap/core/helpers.ts
@@ -187,15 +187,23 @@ export function simplifyFeature(feature: MapGeoJSONFeature): Feature {
   };
 }
 
-export function computeBBoxViewport(boundingBox: BBox | Position, initialViewport: Viewport) {
+export function computeBBoxViewport(
+  boundingBox: BBox | Position,
+  initialViewport: Viewport,
+  opts?: { width?: number; height?: number; padding?: number }
+): Viewport {
   const [minLng, minLat, maxLng, maxLat] = boundingBox;
-  const viewportTemp = new WebMercatorViewport({ ...initialViewport, width: 600, height: 400 });
+  const viewportTemp = new WebMercatorViewport({
+    ...initialViewport,
+    width: opts?.width || 600,
+    height: opts?.height || 400,
+  });
   const { longitude, latitude, zoom } = viewportTemp.fitBounds(
     [
       [minLng, minLat],
       [maxLng, maxLat],
     ],
-    { padding: 40 }
+    { padding: opts?.padding ?? 40 }
   );
 
   return {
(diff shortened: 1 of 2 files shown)


theocrsb (approved):
LGTM and tested

front/src/common/Map/WarpedMap/core/helpers.ts
```diff
@@ -187,15 +187,23 @@ export function simplifyFeature(feature: MapGeoJSONFeature): Feature {
...
+    width: opts?.width || 600,
+    height: opts?.height || 400,
+  });
   const { longitude, latitude, zoom } = viewportTemp.fitBounds(
     [
       [minLng, minLat],
       [maxLng, maxLat],
     ],
-    { padding: 40 }
+    { padding: opts?.padding || 40 }
```
SharglutDev: Maybe this is safer if someone passes `padding: 0` (not sure why) in options ?
```suggestion
    { padding: opts?.padding ?? 40 }
```

front/src/modules/trainschedule/components/ManageTrainSchedule/Map.tsx
```diff
@@ -90,17 +90,24 @@ const Map = ({
...
+    if (isReadOnly && pathGeometry) {
+      const mapContainer = mapRef.current?.getContainer();
+      return computeBBoxViewport(bbox(pathGeometry), viewport, {
+        width: mapContainer?.clientWidth,
+        height: mapContainer?.clientHeight,
+        padding: 60,
+      });
+    }
+    return viewport;
+  }, [isReadOnly, pathGeometry, viewport, mapRef]);
```
SarahBellaha, line 110:
```suggestion
      return computeBBoxViewport(bbox(pathGeometry), viewport, {
        width: mapContainer?.clientWidth,
        height: mapContainer?.clientHeight,
        padding: 60,
      });
    }
    return viewport;
  }, [isReadOnly, pathGeometry, viewport, mapContainer]);
```

front/src/modules/trainschedule/components/ManageTrainSchedule/Map.tsx
```diff
@@ -90,17 +90,24 @@ const Map = ({
   const infraID = useInfraID();
   const terrain3DExaggeration = useSelector(getTerrain3DExaggeration);
   const { viewport, mapSearchMarker, mapStyle, showOSM, layersSettings } = useSelector(getMap);
+  const mapRef = useRef<MapRef | null>(null);
```
SarahBellaha, line 93:
```suggestion
  const mapRef = useRef<MapRef | null>(null);
  const mapContainer = useMemo(() => mapRef.current?.getContainer(), [mapRef.current]);
```
Perhaps we could declare mapContainer once here and use it in the dependency arrays of useEffect(l.216) and useMemo(l.100) that use mapRef?

SarahBellaha (approved):
LGTM :)

SharglutDev (approved):
Lgtm and tested !

PR #10271, opened 2025-01-07
lmr: add mandatory fields and error messages to the form

### LINKED ISSUE #9707: lmr: front: catch missing fields and adapt warning message

Caracol3:
close #9707

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
@@ -33,7 +33,7 @@ const ConsistCardTitle = ({
   );
 };
 
-const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
+const StdcmConsist = ({ disabled = false, showErrors }: StdcmConfigCardProps) => {
   const { t } = useTranslation('stdcm');
   const { speedLimitByTag, speedLimitsByTags, dispatchUpdateSpeedLimitByTag } =
     useStoreDataForSpeedLimitByTagSelector({ isStdcm: true });
@@ -64,6 +64,12 @@ const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
     searchTowedRollingStockById,
     filters: towedRsFilters,
   } = useFilterTowedRollingStock();
+
+  const isTractionEngineEmpty = !filters.text.trim();
+  const isTowedRollingStockEmpty = !towedRsFilters.text.trim();
+  const isTotalMassEmpty = totalMass === undefined || totalMass === null;
+  const isTotalLengthEmpty = totalLength === undefined || totalLength === null;
+  const isMaxSpeedEmpty = maxSpeed === undefined || maxSpeed === null;
 
   useEffect(() => {
     if (towedRollingStock) {
@@ -128,6 +134,11 @@ const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
           suggestions={filteredRollingStockList}
           getSuggestionLabel={(suggestion: LightRollingStockWithLiveries) => getLabel(suggestion)}
           onSelectSuggestion={onSelectSuggestion}
+          statusWithMessage={
+            showErrors && isTractionEngineEmpty
+              ? { tooltip: 'left', status: 'error', message: t('form.missingValue') }
+              : undefined
+          }
         />
       </div>
       <div className="towed-rolling-stock">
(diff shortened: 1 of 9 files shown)


anisometropie (commented):
Screenshot_20250107_175234

Several issues:
- message is not positionned correctly, it should be below the input
- message gets on top of other inputs and we can’t see or interact with them. 
- towed field is required but doesn’t allow me to write anything

What prototype did you use?

We should be able to see the main information of the page, and see at first glance what fields failed validation

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
```diff
@@ -65,6 +65,12 @@ const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
     filters: towedRsFilters,
   } = useFilterTowedRollingStock();
 
+  const isTractionEngineEmpty = !filters.text.trim();
+  const isTowedRollingStockEmpty = !towedRsFilters.text.trim();
+  const isTotalMassEmpty = totalMass === undefined || totalMass === null;
+  const isTotalLengthEmpty = totalLength === undefined || totalLength === null;
+  const isMaxSpeedEmpty = maxSpeed === undefined || maxSpeed === null;
```
anisometropie, line 72: Can’t we use the same function for all number fields?

anisometropie (changes requested):

SharglutDev:
>     * message is not positionned correctly, it should be below the input
> 
>     * message gets on top of other inputs and we can’t see or interact with them.

This is because of overriding rules, we use in stdcm, we should get rid of these in another PR, it breaks all the osrd-ui styles... (the tooltip top position here)
There is one issue in osrd-ui for the `left` position which will be fixed later.
For now, I think we can fix these issues by changing the osrd-ui styles here directly with a TODO that we will remove later.
Fixing the top value of the tooltip will also fix your second point.

>     * towed field is required but doesn’t allow me to write anything

This field shouldn't be required, it will be changed

RomainValls (commented):
Do we need to have the "missing value" field on the towed rolling stock ? As it isn't necessary to run a simulation ?

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
Caracol3: Yes, you're right, it would be clearer. I will change this

Caracol3:
Indeed, towed rolling stock isn't required. I will remove it. thanks

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
SharglutDev: We could use `isNil()` from lodash ?

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
```diff
@@ -65,6 +65,12 @@ const StdcmConsist = ({ disabled = false }: StdcmConfigCardProps) => {
     filters: towedRsFilters,
   } = useFilterTowedRollingStock();
 
+  const isTractionEngineEmpty = !filters.text.trim();
+  const isTowedRollingStockEmpty = !towedRsFilters.text.trim();
+  const isTotalMassEmpty = totalMass === undefined || totalMass === null;
+  const isTotalLengthEmpty = totalLength === undefined || totalLength === null;
```
SharglutDev, line 71: Instead of that, I think you could use the same logic than @Wadjetz in his consist PR (in StdcmConfig). We could have one object having boolean properties for each required fields changing when these fields changes.

We should also `setShowError(false)` as soon as the field changes.

anisometropie:
> * message gets on top of other inputs and we can’t see or interact with them.

Discussed with @thibautsailly, it has to be annoying on purpose to force its resolution first

PR #10272, opened 2025-01-07
front: center text on stdcm error

### LINKED ISSUE #10135: lmr: Information message under error message is not centered
### What happened?

When there is a technical problem, I get this 

Image

It should look like this (ignore the blue arrow, it is for another [bug](https://github.com/OpenRailAssociation/osrd/issues/10295))
Image

So : 
- The text below should be centered

Create a very long train to get the error and observe

Recette (SNCF)

Google Chrome

6580cfe

sim51:
fix #10135 for the text alignment

image

### DIFF
--- a/front/src/styles/scss/applications/stdcm/_home.scss
+++ b/front/src/styles/scss/applications/stdcm/_home.scss
@@ -288,10 +288,11 @@
       }
 
       .error-message {
-        padding-top: 22px;
-        max-width: 546px;
+        padding: 22px 8px 0 8px;
+        width: 466px;
         color: rgba(49, 46, 43, 1);
         line-height: 24px;
+        text-align: center;
       }
     }
   }


front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -289,9 +289,10 @@
 
       .error-message {
         padding-top: 22px;
-        max-width: 546px;
+        width: 466px;
```
SharglutDev, line 292: In the mockup the error message block width is 450px. It has probably be designed like that to avoid the text overlapping on the white borders of the banner (if it fits exactly in his blocks)

Capture d’écran 2025-01-07 à 16 57 21

SharglutDev (approved):
Lgtm and tested, you can close the comment after fixing it :)

PR #10273, opened 2025-01-07
[superseded, on-hold before closing] editoast: remove 'validator' dependency

bougue-pe:
⚠️ Superseded for one part by https://github.com/OpenRailAssociation/osrd/pull/12904 and for the other part by https://github.com/OpenRailAssociation/osrd/issues/13333

---

After https://github.com/OpenRailAssociation/osrd/pull/10147#discussion_r1905598946

TODO:
- [ ] fix tests
- [ ] use Changeset<RollingStock> and #[serde(remote = "Self")] if possible
- [ ] maybe change error type returned by the free function
- [ ] remove any 'validator' use

### DIFF
--- a/editoast/src/client/import_rolling_stock.rs
+++ b/editoast/src/client/import_rolling_stock.rs
@@ -9,12 +9,10 @@ use colored::Colorize as _;
 use editoast_models::DbConnectionPoolV2;
 use editoast_schemas::rolling_stock::RollingStock;
 use editoast_schemas::rolling_stock::TowedRollingStock;
-use validator::ValidationErrorsKind;
 
 use crate::models::prelude::*;
 use crate::models::towed_rolling_stock::TowedRollingStockModel;
 use crate::models::RollingStockModel;
-use crate::CliError;
 
 #[derive(Args, Debug)]
 #[command(about, long_about = "Import a rolling stock given a json file")]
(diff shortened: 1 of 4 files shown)


editoast/src/models/rolling_stock_model.rs
```diff
@@ -93,50 +93,98 @@ impl RollingStockModel {
...
     }
 }
 
-pub fn validate_rolling_stock(
+pub fn validate_rolling_stock<'de, D>(
     effort_curves: &EffortCurves,
     electrical_power_startup_time: Option<f64>,
     raise_pantograph_time: Option<f64>,
-) -> std::result::Result<(), ValidationError> {
+) -> Result<(), D::Error>
```
woshilapin, line 154: I saw that it's part of your TODO list. The errors are going to change, but in the meantime, we might want to stick to the less ideal, but coherent way of doing: using `crate::error::Result` which uses `InternalError`.

woshilapin (commented):
So far, it looks good, most of the comments I was about to do are part of the TODO list, so waiting for the follow-up 😄

editoast/src/client/import_rolling_stock.rs
```diff
@@ -32,46 +30,24 @@ pub async fn import_rolling_stock(
         let rolling_stock: RollingStock =
             serde_json::from_reader(BufReader::new(rolling_stock_file))?;
         let rolling_stock: Changeset<RollingStockModel> = rolling_stock.into();
```
leovalais, line 32: In the spirit of "parse don't validate", I'd argue that the validation should occur at the deserialization of *the schema* and not changeset (unlike what's currently being done on dev). There is an `impl From<editoast_schema::RollingStock> for RollingStockChangeset` existing in `rolling_stock_model.rs`. It might also be easier to pull the `remote = "Self"` trick on the schema as it's struct isn't generated.

Wdyt?

editoast/src/models/rolling_stock_model.rs
leovalais: I think that this function being called by `Deserialize::deserialize` prevents from using `InternalError`

flomonster (commented):
Our supported_signaling_system has a weird behavior. We should deduce if a RS supports ETCS from the  `etcs_brake_params` field (and not provide explicitly).

> Note of @bougue-pe after discussion:
> The main idea is to keep a "free" `Vec<String>` but make it private, and check that `ETCS_LEVEL2` is not part of it (deserialization).
> Then make an "core" accessor for it that populates the variant if RS has `etcs_brake_params`.
> One question "stands": we should probably have a different "accessor" for front (stick to the internal model when serializing) and adapt front to also show `ETCS_LEVEL2` if `etcs_brake_params` exists. ~We could also use the same accessor for front, and then be permissive on input (allow `ETCS_LEVEL2` if `etcs_brake_params` is provided, but remove it before save).~
>
> Tracked in https://github.com/OpenRailAssociation/osrd/issues/13333

editoast/src/client/import_rolling_stock.rs
bougue-pe: After discussion with @leovalais and @flomonster, the options are not perfect, but the first one will be tested:
1. Use `editoast_schema::RollingStock` to replace `RollingStockForm` and remove the PATCH use, just keep a PUT (requesting all object). Then the deserializer can be used to centralize checks (like what's done in `editoast_schemas::train_schedule::TrainSchedule`).
2. ~Add extra constraints in DB: would centralize, but not a practice we want to have, especially when write perfs matter (not the case here, but being consistent is better)~
3. ~Have a free function to be called everywhere introduces the risk to miss the call (and classic duplicate issues). This is pretty much the case currently. Forcing a PG update to deserialize, and then rollback if deserialize checks break leads to identical duplicate issues, while also hiding the cause of the deserialization.~

bougue-pe:
Superseded for one part by https://github.com/OpenRailAssociation/osrd/pull/12904 and for the other part by https://github.com/OpenRailAssociation/osrd/issues/13333: closing it now.

PR #10274, opened 2025-01-07
front: fix scroll in stdcm result table

### LINKED ISSUE #9625: Scrolling in stdcm result scrolls through all waypoints even if not on the waypoints
### What happened?

I create a long train (DN-MSC) and try to scroll down the page. I have to scroll through all the waypoints.

https://github.com/user-attachments/assets/7198eab6-cc0c-4166-ac5c-bb09a750639c

I would expected to scroll in the page without scrolling through all the waypoints (except if the mouse is on the waypoints)

as in the video

Dev (SNCF)

Google Chrome

RomainValls:
close #9625 

Decided to put a max height of 80vh to prevent the results from increasing page size too much. 
Can be changed for lesser / higher value if you think it's better to.

### DIFF
--- a/front/src/styles/scss/applications/stdcm/_results.scss
+++ b/front/src/styles/scss/applications/stdcm/_results.scss
@@ -150,11 +150,12 @@
       .table-container {
         width: 804px;
         margin-right: 32px;
-
+        height: calc(100vh - 180px);
+        overflow-y: auto;
+        scrollbar-width: none;
         .table-results {
           border-radius: 6px;
           background-color: rgba(0, 0, 0, 0.05);
-
           thead {
             position: sticky;
             top: 0;


Yohh (commented):
seems ok but just a suggestion:

to fit on screens, you could set a `height` on `.stdcm__body`:
```css
    height: calc(100vh - 120px);
```
(100vh - stdcm navbar) 

then remove your `max-height` and set a height on `.simulation-results`:
```css
    height: calc(100vh - 180px);
```
(100vh - y-paddings - simulation-list height) 

this way the should always fit on every screen heights

can you try it?
@kmer2016 do you agree?

kmer2016:
> seems ok but just a suggestion:
> 
> to fit on screens, you could set a `height` on `.stdcm__body`:
> 
> ```css
>     height: calc(100vh - 120px);
> ```
> 
> (100vh - stdcm navbar)
> 
> then remove your `max-height` and set a height on `.simulation-results`:
> 
> ```css
>     height: calc(100vh - 180px);
> ```
> 
> (100vh - y-paddings - simulation-list height)
> 
> this way the should always fit on every screen heights
> 
> can you try it? @kmer2016 do you agree?

I agree with the idea of making the height dynamic to improve responsiveness. However, after testing this approach, it slightly breaks the current layout.
Based on your suggestion, I propose the following
```css
.table-container {
    height: calc(100vh - 120px);
    overflow-y: auto;
}
```

RomainValls:
Thank you for the suggestions @Yohh and @kmer2016 🙏

kmer2016 (approved):
LGTM !

RomainValls:
The calc on the stdcm__body is important for everything to be well displayed, but elements inside it (like the map) should be reworked so that they fit correctly. 

@kmer2016 I modified 120px to 180px so that we can see the simulation banner above the waypoints.

Yohh (approved):

PR #10275, opened 2025-01-07
editoast: add weight field to operational points path_properties

Akctarus:


editoast/src/core/path_properties.rs
```diff
@@ -78,6 +78,8 @@ pub struct OperationalPointOnPath {
     extensions: OperationalPointExtensions,
     /// Distance from the beginning of the path in mm
     position: u64,
+    /// Importance of the operational point
+    weight: Option<u8>
```
leovalais:
```suggestion
    /// Importance of the operational point
    #[schema(required)]
    weight: Option<u8>
```

will mark the field as required (we serialize `None` to `null` by default), so the field will never be `undefined` in the frontend

leovalais (approved):
LGTM, thanks!

Erashin (approved):
Core lgtm :)

editoast/src/core/path_properties.rs
younesschrifi: Done 👍🏽

front/src/applications/operationalStudies/views/SimulationResults.tsx
```diff
@@ -120,6 +120,8 @@ const SimulationResults = ({
     return null;
   }
 
+  console.log(projectedOperationalPoints, trainSimulation);
```
SharglutDev:
```suggestion
```

SharglutDev (approved):
Lgtm except the `console.log` :)

front/src/applications/operationalStudies/views/SimulationResults.tsx
Akctarus: done !

SharglutDev (approved):

Khoyo:
You need to `./gradlew spotlessApply` in `core/`

PR #10277, opened 2025-01-07
fix invalid op time stdcm

### LINKED ISSUE #10267: simulation loader pops even if time range is wrong
### What happened?

If I start a simulation with a selected time not in the stdcmEnv range, the warning toast pops but the simulation pops as if the simulation started but its not.

- [ ] Simulation loader shouldn't pop
- [ ] We could display a warning message in the form

1. Go to stdcm
2. Try to launch a simulation with a time not in the env time range

Local

Firefox

.

SharglutDev:
fix #10267


clarani (approved):
LGTM :white_check_mark:

kmer2016 (approved):
LGTM ! Nice catch

Issue #6019, opened 2023-12-06
front: refacto map component across the entire application


nicolaswurtz:
Today, map instances are duplicated across all application, where only one should be necessary (a lot of code is duplicated).

Desired state is a single shared component for all maps in applications, except for infrastructure editor that would have it's own specific map component.

PR #10280, opened 2025-01-08
front: Initialize the new version of the Map.

### LINKED ISSUE #9776: stdcm: front: make the result map great again
# Description

The result map in STDCM got frozen in terms of interactions in order to screenshot the result to send it to the simulation sheet. It would be great to bring back the interactions with the map in order to at least zoom in/out to check out the path.

# AC

- [ ] The user can interact with the result map in terms of zoom (+/-) and displacement.
- [ ] Other map settings do not need to be implemented

kmer2016:
closes #9776
This commit initializes [the refacto](https://github.com/OpenRailAssociation/osrd/issues/6019) of the Map component. The Map is now decoupled from the store. Specifically:
- The zoom button callbacks no longer depend on the store.
- Other buttons have been temporarily hidden.

This change allows multiple Map instances to be rendered on the same page without conflicts.

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
@@ -15,7 +15,7 @@ import {
 import { useOsrdConfActions, useOsrdConfSelectors } from 'common/osrdContext';
 import useInfraStatus from 'modules/pathfinding/hooks/useInfraStatus';
 import { useStoreDataForRollingStockSelector } from 'modules/rollingStock/components/RollingStockSelector/useStoreDataForRollingStockSelector';
-import { Map } from 'modules/trainschedule/components/ManageTrainSchedule';
+import NewMap from 'modules/trainschedule/components/ManageTrainSchedule/NewMap';
 import { type StdcmConfSliceActions, resetMargins } from 'reducers/osrdconf/stdcmConf';
 import type { StdcmConfSelectors } from 'reducers/osrdconf/stdcmConf/selectors';
 import { useAppDispatch } from 'store';
@@ -246,7 +246,8 @@ const StdcmConfig = ({
         </div>
 
         <div className="osrd-config-item-container osrd-config-item-container-map stdcm-map">
-          <Map
+          <NewMap
+            id="stdcm-map-config"
             hideAttribution
             hideItinerary
             preventPointSelection
--- a/front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
+++ b/front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
@@ -13,7 +13,7 @@ import {
 } from 'applications/stdcm/utils/formatSimulationReportSheet';
 import { hasConflicts, hasResults } from 'applications/stdcm/utils/simulationOutputUtils';
 import { type TrackRange } from 'common/api/osrdEditoastApi';
-import { Map } from 'modules/trainschedule/components/ManageTrainSchedule';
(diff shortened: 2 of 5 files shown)


front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
```diff
@@ -164,15 +164,17 @@ const StcdmResults = ({
...
+          <NewMap
+            id="stdcm-map-result"
             hideAttribution
             showStdcmAssets
             isFeasible={!hasConflictResults}
             pathGeometry={outputs?.pathProperties?.geometry}
             simulationPathSteps={markersInfo}
+            withSearchButton={false}
+            withToggleLayersButton={false}
+            withMapKeyButton={false}
```
clarani: It would be better to set these props to false by default, then you can simply remove them
```suggestion
```

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -199,13 +199,17 @@ const StdcmConfig = ({
...
+            id="stdcm-map-config"
             hideAttribution
             hideItinerary
             preventPointSelection
             pathGeometry={pathfinding?.geometry}
             showStdcmAssets
             simulationPathSteps={markersInfo}
+            withSearchButton={false}
+            withToggleLayersButton={false}
+            withMapKeyButton={false}
```
clarani:
```suggestion
```

front/tests/pages/stdcm-page-model.ts
```diff
@@ -162,7 +162,7 @@ class STDCMPage {
     this.notificationHeader = page.locator('#notification');
     this.debugButton = page.getByTestId('stdcm-debug-button');
     this.helpButton = page.getByTestId('stdcm-help-button');
-    this.mapContainer = page.locator('#map-container');
+    this.mapContainer = page.locator('stdcm-map-config');
```
Maymanaf: `page.locator('#stdcm-map-config');`

Maymanaf (changes requested):

Yohh:
just to be sure, is the name `NewMap` going to change when the refacto will be done or is it it's real and definitive name?

front/src/common/Map/Buttons/MapButtons.tsx
```diff
@@ -34,6 +34,10 @@ type MapButtonsProps = {
   map?: MapRef;
   resetPitchBearing: () => void;
   closeFeatureInfoClickPopup?: () => void;
+  zoomIn?: () => void;
+  zoomOut?: () => void;
+  withSearchButton?: boolean;
+  withToggleLayersButton?: boolean;
```
Yohh, line 40: `with` is usually used for HOC components I guess, it can be a little bit confusing, should'nt we use `is` instead?

kmer2016:
> just to be sure, is the name `NewMap` going to change when the refacto will be done or is it it's real and definitive name?

Oh no, the name will also change once the refactor is complete

front/src/common/Map/Buttons/MapButtons.tsx
kmer2016: I just followed the naming convention used by other props that were already present like `withMapKeyButton` or `withInfraButton`. Would you be okay if we address this in the next step of the refactor?

clarani (approved):
LGTM :white_check_mark:

RomainValls (approved):
LGTM and tested 👍

Maymanaf (approved):
LGTM for e2e-tests

PR #10282, opened 2025-01-08
lmr: add messages during pathfinding and when its done

### LINKED ISSUE #9451: `Obtenir la simulation` button doesn't send a request when there's no pathfinding
### What happened?
Part of https://github.com/osrd-project/osrd-confidential/issues/815

Clicking the `Obtenir la simulation` button does nothing, and actually doesn't even send a request. The console shows this:

Image

> [!NOTE]
> Edit: actually core was down so the pathfinding wasn't completed. So the button could not work.
> Edit n°2: same thing happens when no rolling stock is selected, there can't be a pathfinding
> Edit n°3: same when origin, destination, or waypoint is missing

- [ ] This message should be displayed when the pathfinding is running:

Image

- [ ] When the path is found, this message should appear for 2 seconds then disappearing with a fading

Image

Inputs are in the screenshot.

Recette (SNCF)

Google Chrome

7451939

Caracol3:
close #9451

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useMemo, useRef, useState } from 'react';
+import { useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react';
 
 import { Button } from '@osrd-project/ui-core';
 import cx from 'classnames';
@@ -90,8 +90,12 @@ const StdcmConfig = ({
   const totalLength = useSelector(getTotalLength);
   const maxSpeed = useSelector(getMaxSpeed);
 
-  const pathfinding = useStaticPathfinding(infra);
+  const [showMessage, setShowMessage] = useState(false);
+
+  const { pathfinding, isPathFindingLoading } = useStaticPathfinding(infra);
+
   const formRef = useRef<HTMLDivElement>(null);
+  const pathfindingBannerRef = useRef<HTMLDivElement>(null);
 
   const [formErrors, setFormErrors] = useState<StdcmConfigErrors>();
 
@@ -145,6 +149,13 @@ const StdcmConfig = ({
     );
   };
 
+  const getStatusMessage = () => {
+    if (isPathFindingLoading) {
+      return t('pathfindingStatus.calculating');
+    }
+    return t('pathfindingStatus.success');
+  };
+
   useEffect(() => {
     if (pathfinding) {
       const formErrorsStatus = checkStdcmConfigErrors(
@@ -176,6 +187,31 @@ const StdcmConfig = ({
       dispatch(restoreStdcmConfig(state));
     }
   }, []);
+
+  useEffect(() => {
+    if (isPathFindingLoading) {
+      setShowMessage(true);
+    }
+
+    if (pathfinding?.status === 'failure') {
+      setShowMessage(false);
+    }
+  }, [isPathFindingLoading, pathfinding?.status]);
+
+  useLayoutEffect(() => {
+    const handleAnimationEnd = () => {
+      setShowMessage(false);
+    };
+
+    if (!showMessage || formErrors) {
+      return undefined;
+    }
+    pathfindingBannerRef.current!.addEventListener('animationend', handleAnimationEnd);
+
+    return () => {
+      pathfindingBannerRef.current?.removeEventListener('animationend', handleAnimationEnd);
+    };
+  }, [showMessage, formErrors]);
 
   return (
     <div className="stdcm__body">
@@ -238,6 +274,20 @@ const StdcmConfig = ({
                 )}
               </div>
 
+              {!formErrors && showMessage && (
+                <div className="simulation-status-banner">
+                  <div className="banner-content">
+                    <div
+                      ref={pathfindingBannerRef}
+                      className={cx('pathfinding-status', {
+                        'pathfinding-status-success': pathfinding?.status === 'success',
+                      })}
+                    >
+                      {getStatusMessage()}
+                    </div>
+                  </div>
+                </div>
+              )}
               {isPending && (
                 <StdcmLoader
                   cancelStdcmRequest={cancelStdcmRequest}
--- a/front/src/applications/stdcm/hooks/useStaticPathfinding.ts
+++ b/front/src/applications/stdcm/hooks/useStaticPathfinding.ts
@@ -32,7 +32,7 @@ const useStaticPathfinding = (infra?: InfraWithState) => {
 
   const [pathfinding, setPathfinding] = useState<PathfindingResult>();
 
-  const [postPathfindingBlocks] =
+  const [postPathfindingBlocks, { isFetching }] =
     osrdEditoastApi.endpoints.postInfraByInfraIdPathfindingBlocks.useLazyQuery();
 
   // When pathSteps changed
@@ -84,7 +84,7 @@ const useStaticPathfinding = (infra?: InfraWithState) => {
     [pathfinding, pathProperties]
   );
 
-  return result;
+  return { pathfinding: result, isPathFindingLoading: isFetching };
 };
 
 export default useStaticPathfinding;
--- a/front/src/styles/scss/applications/stdcm/_home.scss
+++ b/front/src/styles/scss/applications/stdcm/_home.scss
@@ -258,6 +258,9 @@
     .banner-content {
       padding: 0 0 16px 378px;
       background: linear-gradient(180deg, rgba(239, 243, 245) 40px, rgba(233, 239, 242) 40px);
+      &:has(.pathfinding-status) {
+        padding: 0;
+      }
       .status {
         width: 466px;
         display: flex;
@@ -293,6 +296,45 @@
         color: rgba(49, 46, 43, 1);
         line-height: 24px;
         text-align: center;
+      }
+
+      .pathfinding-status {
+        left: -8px;
+        position: absolute;
+        bottom: -95px;
+        width: calc(100% + 16px);
+        display: flex;
+        justify-content: center;
+        line-height: 24px;
+        border-radius: 8px;
+        color: var(--info60);
+        background-color: var(--info5);
+        padding-block: 22px 26px;
+        box-shadow:
+          0 0 0 4px rgba(255, 255, 255, 1) inset,
+          0 0 0 5px rgb(112, 193, 229) inset,
+          0 3px 7px -3px rgba(28, 28, 217, 0.4);
+        &-success {
+          color: var(--success60);
+          background-color: var(--success5);
+          opacity: 0;
+          box-shadow:
+            0 0 0 4px rgba(255, 255, 255, 1) inset,
+            0 0 0 5px rgb(60, 202, 128) inset,
+            0 3px 7px -3px rgba(11, 114, 60, 0.4);
+          animation: pathfinding-success 2.7s;
+        }
+      }
+      @keyframes pathfinding-success {
+        0% {
+          opacity: 1;
+        }
+        74% {
+          opacity: 1;
+        }
+        100% {
+          opacity: 0;
+        }
       }
     }
   }
--- a/front/public/locales/en/stdcm.json
+++ b/front/public/locales/en/stdcm.json
@@ -50,6 +50,10 @@
   },
   "noCorrespondingResults": "No results matching your search.",
   "notificationTitle": "Phase 1: from D-7 to D-1 5pm, on the Perrigny-Miramas axis.",
+  "pathfindingStatus": {
+    "calculating": "Validating path...",
+    "success": "The requested path is valid."
+  },
   "pleaseWait": "Please wait…",
   "simulation": {
     "calculatingSimulation": "Calculation in progress...",
--- a/front/public/locales/fr/stdcm.json
+++ b/front/public/locales/fr/stdcm.json
@@ -50,6 +50,10 @@
   },
   "noCorrespondingResults": "Aucun résultat ne correspond à votre recherche.",
   "notificationTitle": "Phase 1 : de J-7 à J-1 17h, sur l’axe Perrigny—Miramas.",
+  "pathfindingStatus": {
+    "calculating": "Validation de l'itinéraire en cours...",
+    "success": "L'itinéraire demandé est valide."
+  },
   "pleaseWait": "Veuillez patientez…",
   "simulation": {
     "calculatingSimulation": "Calcul en cours...",
--- a/front/src/styles/scss/_variables.scss
+++ b/front/src/styles/scss/_variables.scss
@@ -67,6 +67,7 @@ $colors: (
   'primary80': #1f0f96,
   'primary90': #180f47,
   'selection20': #fff2b3,
+  'success5': #e6f7ee,
   'success30': #3cca80,
   'success60': #0b723c,
   'warning5': #fdf5e1,


theocrsb (commented):
image

I have these two messages at the same time

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -110,6 +110,7 @@ const StdcmConfig = ({
       maxSpeed: validateMaxSpeed(maxSpeed, rollingStock?.max_speed),
     };
   }, [rollingStock, towedRollingStock, totalMass, totalLength, maxSpeed]);
+  const [validationMessage, setValidationMessage] = useState<string | null>(null);
```
theocrsb: I think we can just use undefined
```suggestion
  const [validationMessage, setValidationMessage] = useState<string>();
```

Caracol3:
> image
> 
> I have these two messages at the same time

This is weird, i never had this. Did you restart the front ?

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
Caracol3: Good catch ! Done

Math-R (dismissed):
Lgtm tested & live reviewed with @Caracol3

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -173,6 +174,25 @@ const StdcmConfig = ({
     }
   }, []);
 
+  useEffect(() => {
+    if (origin.location && destination.location & rollingStock?.id) {
```
Math-R: & => &&

theocrsb:
> > image
> > I have these two messages at the same time
> 
> This is weird, i never had this. Did you restart the front ?

Yes, You can try these inputs 

image

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
Caracol3: done

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -87,6 +87,7 @@ const StdcmConfig = ({
   const formRef = useRef<HTMLDivElement>(null);
 
   const [formErrors, setFormErrors] = useState<StdcmConfigErrors>();
+  const [ciChanged, setCiChanged] = useState(false);
```
theocrsb: I think it's better to use a variable name with an `is` because it's a boolean

Caracol3: Good catch !
done

theocrsb (approved):
LGTM and tested :)

Math-R (approved):
lgtm

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -267,7 +271,7 @@
         justify-content: center;
         color: rgb(11, 114, 60);
         padding-top: 23px;
-        font-size: 1.5rem;
+        font-size: 24px;
```
SharglutDev: font-size is the only style who should remain in rem :) (same in `.validation-message`)

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -294,6 +298,35 @@
...
+        display: flex;
+        font-weight: 400;
+        letter-spacing: 0px;
+        text-align: center;
+        justify-content: center;
+        line-height: 32px;
+        height: 80px;
+        border-radius: 8px;
+        color: rgb(33, 100, 130);
+        background-color: rgb(230, 247, 255);
```
SharglutDev: We try to directly use the design system theme from `_variables.scss`
```suggestion
        background-color: var(--info5);
```

You can also use `info60` for the text color, add `success5` (in hex unit) for the success background and use `success60` for the success text

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -294,6 +298,35 @@
...
         text-align: center;
       }
+
+      .validation-message {
+        width: 450px;
+        display: flex;
+        font-weight: 400;
+        letter-spacing: 0px;
+        text-align: center;
+        justify-content: center;
```
SharglutDev: Are these two necessary ? Maybe `justify-content` is enough (didn't test)

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -258,6 +258,10 @@
     .banner-content {
       padding: 0 0 16px 378px;
       background: linear-gradient(180deg, rgba(239, 243, 245) 40px, rgba(233, 239, 242) 40px);
+      &:has(.validation-message) {
+        padding: 0 !important;
```
SharglutDev: If you need to override something, could you leave a comment here to explain why ?

front/src/styles/scss/applications/stdcm/_home.scss
SharglutDev, line 1: Capture d’écran 2025-01-10 à 11 33 55

Did you add the fade-out effect ? I don't see it

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -173,6 +175,30 @@ const StdcmConfig = ({
     }
   }, []);
 
+  useEffect(() => {
+    setIsCiChanged(true);
+  }, [origin.location, destination.location]);
```
SharglutDev: What about if a via has changed ?

front/src/styles/scss/applications/stdcm/_home.scss
Caracol3: done

front/src/styles/scss/applications/stdcm/_home.scss
Caracol3: done

front/src/styles/scss/applications/stdcm/_home.scss
Caracol3: indeed, justify-content alone was enough

front/src/styles/scss/applications/stdcm/_home.scss
Caracol3: Good catch. I put it in the beginning of my work but there is no use for the ``` ìmportant ``` now. I removed it.
We still need the padding though.

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -110,6 +111,7 @@ const StdcmConfig = ({
       maxSpeed: validateMaxSpeed(maxSpeed, rollingStock?.max_speed),
     };
   }, [rollingStock, towedRollingStock, totalMass, totalLength, maxSpeed]);
+  const [validationMessage, setValidationMessage] = useState<string>();
```
SharglutDev: Instead of adding a new state, do you think we could use a logic similar of the one in this [PR](https://github.com/OpenRailAssociation/osrd/pull/10292/files) ? When a CI changes (origin, via or destination), the pathfinding is relaunched so we can use its loading state to display the info message.

front/src/styles/scss/applications/stdcm/_home.scss
Caracol3: done

front/src/styles/scss/applications/stdcm/_home.scss
Caracol3: done

front/src/styles/scss/applications/stdcm/_home.scss
Caracol3: We have it now

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
Caracol3: The logic is now different, we have the message even when a via is added or removed

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
Caracol3: done

front/src/applications/stdcm/hooks/useStaticPathfinding.ts
```diff
@@ -84,7 +84,7 @@ const useStaticPathfinding = (infra?: InfraWithState) => {
     [pathfinding, pathProperties]
   );
 
-  return result;
+  return { pathfinding: result, isFetching };
```
SharglutDev: Better to do that because when we use it on other component it's hard to follow what is actually fetching.
```suggestion
  return { pathfinding: result, isPathfindingLoading: isFetching };
```

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -294,6 +298,39 @@
         line-height: 24px;
         text-align: center;
       }
+
+      .validation-message {
+        width: 450px;
```
SharglutDev:
```suggestion
        width: 466px;
```

image

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -294,6 +298,39 @@
...
+        display: flex;
+        font-weight: 400;
+        letter-spacing: 0px;
+        justify-content: center;
+        line-height: 32px;
```
SharglutDev:
```suggestion
        line-height: 24px;
```

image

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -294,6 +298,39 @@
...
+        height: 80px;
```
SharglutDev:
```suggestion
        height: 72px;
```

image

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -294,6 +298,39 @@
...
+        border-radius: 8px;
+        color: var(--info60);
+        background-color: var(--info5);
+        margin-top: 16px;
+        padding-top: 23px;
+        font-size: 1rem;
```
SharglutDev: Since it's 1rem after all, we probably don't need to set this

```suggestion
```

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -294,6 +298,39 @@
         line-height: 24px;
         text-align: center;
       }
+
+      .validation-message {
+        width: 450px;
+        display: flex;
+        font-weight: 400;
+        letter-spacing: 0px;
```
SharglutDev: Since we don't justify the text, I don't think we need to specify this

```suggestion
```

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -294,6 +298,39 @@
...
+            0 0 0 4px rgba(255, 255, 255, 1) inset,
+            0 0 0 5px rgb(60, 202, 128) inset,
+            0 3px 7px -3px rgba(11, 114, 60, 0.4);
+        }
+      }
+
+      .fade-out {
+        opacity: 0;
+        transition: opacity 0.3s ease-out;
+      }
```
SharglutDev: The animation doesn't seem to work. It's because when `isFadingOut` becomes true, the banner is removed from the DOM and the fade-out class does nothing.

front/src/styles/scss/applications/stdcm/_home.scss
Caracol3: done

front/src/styles/scss/applications/stdcm/_home.scss
Caracol3: good catch, i removed it

front/src/styles/scss/applications/stdcm/_home.scss
Caracol3: done

front/src/styles/scss/applications/stdcm/_home.scss
Caracol3: done

front/src/styles/scss/applications/stdcm/_home.scss
Caracol3: done

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -294,6 +297,52 @@
...
+        color: var(--info60);
+        background-color: var(--info5);
+        margin-top: 16px;
+        padding-top: 23px;
+        box-shadow:
+          0 0 0 4px rgba(255, 255, 255, 1) inset,
+          0 0 0 5px rgb(112, 193, 229) inset,
+          0 3px 7px -3px rgba(28, 28, 217, 0.4);
+        &-success {
+          color: var(--success60);
```
SharglutDev, line 318: We should add this to prevent for the blinking effect after the fade out
```suggestion
          opacity: 0;
          color: var(--success60);
```

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -173,6 +185,20 @@ const StdcmConfig = ({
...
+    }
+    document.addEventListener('animationend', () => {
+      setShowMessage(false);
+    });
+    return () => {
+      document.removeEventListener('animationend', () => {
+        setShowMessage(false);
+      });
+    };
+  }, [isPathFindingLoading, pathfinding?.status]);
```
SharglutDev:
```suggestion
  }, [isPathFindingLoading]);
```

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -138,16 +141,25 @@ const StdcmConfig = ({
...
+    }
+    return t('pathfindingStatus.success');
+  };
+
+  const formErrorsStatus = useMemo(() => {
+    if (pathfinding && origin.location && destination.location) {
+      return checkStdcmConfigErrors(pathfinding.status !== 'success', pathSteps, t);
+    }
+    return undefined;
+  }, [pathfinding, pathSteps, t]);
```
SharglutDev: Not sure why you change the old useEffect for this ?

SharglutDev (commented):
Added some comments we discussed (+1 bonus) not to forget them

front/src/applications/stdcm/hooks/useStaticPathfinding.ts
Caracol3: done

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -173,6 +185,27 @@ const StdcmConfig = ({
...
+  useEffect(() => {
+    if (isPathFindingLoading) {
+      setShowMessage(true);
+    }
+
+    const handleAnimationEnd = () => {
+      setShowMessage(false);
+    };
+
+    const element = document.querySelector('.pathFinding-status');
```
SharglutDev: It's better to use a `ref` when possible in React. You can declare it next to the other `useRef` of the component.
```suggestion
    const pathfindingBannerRef = useRef<HTMLDivElement>(null);
```

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -173,6 +185,27 @@ const StdcmConfig = ({
...
+    if (element) {
+      element.addEventListener('animationend', handleAnimationEnd);
+    }
+
+    return () => {
+      if (element) {
```
SharglutDev: Don't think you need to put the `if` here. With the ref, you should be able to do 
```suggestion
      pathfindingBannerRef.current?.removedEvent...
```

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
Caracol3: It was when we didn’t yet know why there was an infinite loop. The error was pointing to this useEffect. I thought at the time that a useMemo could be a solution to fix it. But I can put it back since the issue was actually elsewhere.

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
Caracol3: Indeed, with the ref, its useless now. I removed it

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
Caracol3: Good catch ! 
Done !

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -173,6 +184,22 @@ const StdcmConfig = ({
     }
   }, []);
 
+  useEffect(() => {
+    if (isPathFindingLoading) {
+      setShowMessage(true);
+    }
+
```
SharglutDev, line 200: We need to add this case, otherwise the hidden banner is never removed from the DOM if the pathfinding fails.
```suggestion
    if (pathfinding?.status === 'failure') {
      setShowMessage(false);
    }
```
Also add `pathfinding?.status` in the deps array

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -173,6 +184,22 @@ const StdcmConfig = ({
...
+    const handleAnimationEnd = () => {
+      setShowMessage(false);
+    };
+
+    pathfindingBannerRef.current?.addEventListener('animationend', handleAnimationEnd);
+
+    return () => {
+      pathfindingBannerRef.current?.removeEventListener('animationend', handleAnimationEnd);
```
SharglutDev: Sorry to change again but it seems `showBanner` never goes back to false (and the hidden banner is never remove from the dom) if we launch 2 pathfinding in a row : launch a PNO>LEW, clean LEW and then launch a PNO>ARR and watch the element not being removed in the inspector)
```suggestion
    document.addEventListener('animationend', handleAnimationEnd);

    return () => {
      document.removeEventListener('animationend', handleAnimationEnd);
```

emersion: `document` will catch animations from all elements on the page, so I don't think it's a good idea. For instance if a completely unrelated element happens to animate too then we'd mix things up here.

SharglutDev: Yeah that's why we changed it before, but somehow the `showMessage` state is not reset to true (only the first time) if we keep it with the ref on the banner. Do you have any idea about that ?

emersion: Have you tried `useLayoutEffect` and putting the ref on the `.pathFinding-status`?

SharglutDev: Yep, didn't work

SharglutDev: Fixed

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -294,6 +297,53 @@
...
+        }
+        74% {
+          opacity: 1;
+        }
+        100% {
+          opacity: 0;
+        }
+      }
+
+      .fade-out {
```
emersion: This class is unused.

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -294,6 +297,53 @@
...
+        margin-top: 16px;
+        padding-top: 23px;
+        box-shadow:
+          0 0 0 4px rgba(255, 255, 255, 1) inset,
+          0 0 0 5px rgb(112, 193, 229) inset,
+          0 3px 7px -3px rgba(28, 28, 217, 0.4);
+        &-success {
+          color: var(--success60);
+          background-color: var(--success5);
+          opacity: 0;
```
emersion, line 320: I think this `opacity` can be dropped.

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -294,6 +297,53 @@
         line-height: 24px;
         text-align: center;
       }
+
+      .pathFinding-status {
```
emersion: Nit: we use kebab-case instead of camelCase for CSS classes.

Hm, I suppose the class is used elsewhere so we can't change it?

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -294,6 +297,53 @@
...
+        left: -8px;
+        position: absolute;
+        bottom: -95px;
+        width: 466px;
+        display: flex;
+        font-weight: 400;
```
emersion: This is the default font-weight, so shouldn't be necessary.

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -294,6 +297,53 @@
...
+        justify-content: center;
+        line-height: 24px;
+        height: 72px;
+        border-radius: 8px;
+        color: var(--info60);
+        background-color: var(--info5);
+        margin-top: 16px;
```
emersion: Seems like removing this `margin-top` doesn't change the result.

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -294,6 +297,53 @@
         line-height: 24px;
         text-align: center;
       }
+
+      .pathFinding-status {
+        left: -8px;
+        position: absolute;
+        bottom: -95px;
+        width: 466px;
```
emersion: Would be slightly nicer to use `calc(100% + 16px)`: would make it more obvious why there is a `left: -8px`.

emersion (approved):
Just some nits, feel free to resolve!

front/src/styles/scss/applications/stdcm/_home.scss
SharglutDev: done

front/src/styles/scss/applications/stdcm/_home.scss
emersion: It cannot: when the animation ends the state setter doesn't cause an immediate re-render to remove the element from the DOM, but we want to keep a zero opacity still.

front/src/styles/scss/applications/stdcm/_home.scss
SharglutDev: done

front/src/styles/scss/applications/stdcm/_home.scss
SharglutDev: done

front/src/styles/scss/applications/stdcm/_home.scss
SharglutDev: done

front/src/styles/scss/applications/stdcm/_home.scss
SharglutDev: done

front/src/styles/scss/applications/stdcm/_home.scss
SharglutDev: we need it

PR #10284, opened 2025-01-08
front: fix "rolling stock" spelling in English translations

emersion:
"Rollingstock" is not an English word.


SharglutDev (approved):
In `rollingstock.json`, there are some use of `rollingstock` as well. Should we fix them ?

Caracol3 (approved):
lgtm

emersion:
> In rollingstock.json, there are some use of rollingstock as well. Should we fix them ?

Good catch, fixed!

PR #10285, opened 2025-01-08
front: fix nge update after importing or deleting multiple trains

### LINKED ISSUE #10028: Macro mode : multiple deletion or importation from the train list does not update the graph
### What happened?

1. I have several trains and I am in macro mode 
Image

2. I delete my trains at once in the train list
Image

3. My trains are still displayed in NGE even if they were deleted
Image

The same thing happens when importing several trains from Viriato for example : I see my trains in the list but not in NGE

I expected the trains to be deleted

as above

Recette (SNCF)

Google Chrome

9e75ce2

RomainValls:
close #10028


Uriel-Sautron (approved):
Thank you for this fixup, LGTM tested.

emersion (approved):
LGTM as well!

Issue #9533, opened 2024-10-31
STDCM: Starting a new query does not reset all the fields in the consist card


Maymanaf:
### What happened?

![Image](https://github.com/user-attachments/assets/d3b34009-bab8-4135-a58c-62d2b1ad58be)

### What did you expect to happen?

The fields should be emptied

### How can we reproduce it (as minimally and precisely as possible)?

1. Go to stdcm, launch a request
2. Select the simulation
3. Click on reset the form

### On which environments the bug occurs?

Recette (SNCF)

### On which browser the bug occurs?

Firefox

### OSRD version (top right corner `Account` button > `Informations`)

6d9ed21

axrolld:
Sould be fixed by https://github.com/OpenRailAssociation/osrd/issues/9695

celine-sncf:
Seems fixed for the "Convoi" card but still happening for the "Hour" field

theocrsb:
I can't reproduce the bug

sim51:
I reproduce it for the "hour field". 
In fact this field is also not changed when you switch from one simulation to an other.

I found a way to fix the "clean query" issue : in the component `StdcmOpSchedule` we pass to `TimePicker` the attribut `value` (as well as `hours` & `minutes` which are the good data).  By removing it, it works. 

But it doesn't fix my issue on simulation switch.

theocrsb:
> I reproduce it for the "hour field". In fact this field is also not changed when you switch from one simulation to an other.
> 
> I found a way to fix the "clean query" issue : in the component `StdcmOpSchedule` we pass to `TimePicker` the attribut `value` (as well as `hours` & `minutes` which are the good data). By removing it, it works.
> 
> But it doesn't fix my issue on simulation switch.

The bug when switch from one simulation to an other should be [fix by this PR](https://github.com/OpenRailAssociation/osrd/pull/10192)

sim51:
[PR](https://github.com/OpenRailAssociation/osrd/pull/10286) that fix the clear the hour field

EthanPERRUZZA:
PR were merged, bug is fixed

PR #10286, opened 2025-01-08
front: stdcm - new query should clear the hour field

sim51:
Related to #9533

TimePicker component should not have a `value` property, it only needs `hours`, `minutes` & `seconds`.
Giving a `value` property to the component lead to some unwanted behaviors.

See https://github.com/OpenRailAssociation/osrd-ui/issues/807 for more information


SharglutDev:
This component is gonna change very soon [here](https://github.com/OpenRailAssociation/osrd/pull/10192). But you're probably right for the value thing :)

SharglutDev (approved):
Lgtm and tested

PR #10287, opened 2025-01-08
front: fix some issues in stdcm consist errors tooltips

SharglutDev:
- Tooltip position in consist form should be always left
- The launch button shouldn't be disabled if a tooltip is displayed

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
@@ -218,10 +218,7 @@ const StdcmConfig = ({
                   isDisabled={
                     disabled ||
                     !showBtnToLaunchSimulation ||
-                    formErrors?.errorType === StdcmConfigErrorTypes.INFRA_NOT_LOADED ||
-                    !!consistErrors.totalMass ||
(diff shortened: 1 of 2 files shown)


front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -218,10 +218,7 @@ const StdcmConfig = ({
                   isDisabled={
                     disabled ||
                     !showBtnToLaunchSimulation ||
-                    formErrors?.errorType === StdcmConfigErrorTypes.INFRA_NOT_LOADED ||
-                    !!consistErrors.totalMass ||
-                    !!consistErrors.totalLength ||
-                    !!consistErrors.maxSpeed
```
emersion, line 224: With these removed, we can now launch the simulation with invalid parameters.

SharglutDev: I understand but Thibaut doesn't want the launch button to be blocked by that. We will soon have a validation on the back-end and one with the warning box in the front-end. As the tooltip is pretty much visible, we can hope that the user will fix it before sending the simulation :)

emersion (approved):

PR #10288, opened 2025-01-08
front: avoid unecessary path finding call in stdcm

### LINKED ISSUE #10144: stdcm: map shakes when adjusting time

sim51:
fix #10144

Adding a state for `pathStepsLocations` and only changes its value when there is a deep diff

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
@@ -147,7 +147,7 @@ const StdcmConfig = ({
       );
       setFormErrors(formErrorsStatus);
     }
-  }, [pathfinding]);
+  }, [pathfinding, pathSteps, t]);
 
   useEffect(() => {
     if (!isDebugMode) {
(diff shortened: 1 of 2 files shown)


emersion (commented):
Why are the `pathSteps` updated when touching the calendar in the first place?

Shouldn't we avoid this update instead? Otherwise we'd need to carry this workaround each time `pathStep` is in a `useEffect()`/`useMemo()` dependency.

sim51:
Temporal inputs are stored in the corresponding step. Steps defines the whole path with its constraints, which makes sense.

Ex: 
Step 1 : 2025-01-05 at 10h22 Nantes
Step 2: Ancenis with a 5min Stop
Step 3: Angers

In our case we need a derivative data because we just want to find a path (not a schedule).

emersion (approved):
Eh, sorry, I thought we had separate arrays for `path` and `schedule`, but everything is in a single `StdcmPathStep` type.

I don't see a better way to fix this bug!

SharglutDev (approved):
Lgtm and tested

PR #10292, opened 2025-01-09
front: enhance pathfinding loading state management

SarahBellaha:
Completes [this PR](https://github.com/OpenRailAssociation/osrd/pull/10288)

This PR will add a loader while the pathfinding is still computing, and the map's viewport will be updated only once it is finished, avoiding the 'shaking' effect.

### DIFF
--- a/front/src/applications/stdcm/hooks/useStdcmPathfinding.ts
+++ b/front/src/applications/stdcm/hooks/useStdcmPathfinding.ts
@@ -3,11 +3,7 @@ import { useEffect, useMemo, useState } from 'react';
 import { compact, isEqual } from 'lodash';
 import { useSelector } from 'react-redux';
 
-import {
-  osrdEditoastApi,
-  type InfraWithState,
-  type PathfindingResult,
-} from 'common/api/osrdEditoastApi';
+import { osrdEditoastApi, type InfraWithState } from 'common/api/osrdEditoastApi';
 import { useOsrdConfSelectors } from 'common/osrdContext';
 import usePathProperties from 'modules/pathfinding/hooks/usePathProperties';
 import { getPathfindingQuery } from 'modules/pathfinding/utils';
@@ -24,16 +20,11 @@ function pathStepsToLocations(
   return compact(pathSteps.map((s) => s.location));
 }
 
-const useStaticPathfinding = (infra?: InfraWithState) => {
+const useStdcmPathfinding = (infra?: InfraWithState) => {
   const { getStdcmPathSteps } = useOsrdConfSelectors() as StdcmConfSelectors;
   const pathSteps = useSelector(getStdcmPathSteps);
   const [pathStepsLocations, setPathStepsLocations] = useState(pathStepsToLocations(pathSteps));
   const { rollingStock } = useStoreDataForRollingStockSelector();
-
-  const [pathfinding, setPathfinding] = useState<PathfindingResult>();
-
-  const [postPathfindingBlocks, { isFetching }] =
-    osrdEditoastApi.endpoints.postInfraByInfraIdPathfindingBlocks.useLazyQuery();
 
   // When pathSteps changed
   // => update the pathStepsLocations (if needed by doing a deep comparison).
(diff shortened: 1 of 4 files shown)


front/src/applications/stdcm/hooks/useStaticPathfinding.ts
```diff
@@ -21,6 +21,7 @@ const useStaticPathfinding = (infra?: InfraWithState) => {
   const { rollingStock } = useStoreDataForRollingStockSelector();
 
   const [pathfinding, setPathfinding] = useState<PathfindingResult>();
+  const [isPathfindingLoading, setIsPathfindingLoading] = useState(false);
```
SharglutDev: Instead of this state maybe we could use the `isLoading` from the `postInfraByInfraIdPathfindingBlocks` endpoint  provided by rtk-query ?

front/src/applications/stdcm/hooks/useStaticPathfinding.ts
```diff
@@ -20,56 +16,45 @@ const useStaticPathfinding = (infra?: InfraWithState) => {
...
-      const pathfindingResult = await postPathfindingBlocks(payload).unwrap();
-
-      setPathfinding(pathfindingResult);
-    };
-
-    launchPathfinding();
+  const pathfindingPayload = useMemo(() => {
+    if (infra?.state !== 'CACHED' || !rollingStock || pathStepsLocations.length < 2) {
+      return undefined;
+    }
```
SharglutDev: The other checks are done in `getPathfindingQuery`. And to match `getPathfindingQuery` return, maybe it' better to return also `null` if infra state is not cached.
```suggestion
    if (infra?.state !== 'CACHED') {
      return null;
    }
```

front/src/applications/stdcm/hooks/useStaticPathfinding.ts
SharglutDev, line 1: As it's not "static" anymore (no `lazyQuery`), maybe we could rename this file `useStdcmPathfinding`

front/src/applications/stdcm/hooks/useStaticPathfinding.ts
```diff
@@ -20,56 +16,45 @@ const useStaticPathfinding = (infra?: InfraWithState) => {
...
-        ? { status: pathfinding.status, geometry: pathProperties?.geometry ?? undefined }
-        : null,
-    [pathfinding, pathProperties]
-  );
-
-  return result;
+  const { pathfinding, isFetching } =
+    osrdEditoastApi.endpoints.postInfraByInfraIdPathfindingBlocks.useQuery(pathfindingPayload!, {
+      skip: !pathfindingPayload,
+      refetchOnMountOrArgChange: true,
```
SharglutDev: Don't think we need this one
```suggestion
```

Caracol3 (approved):
LGTM and tested !

SharglutDev (commented):
Tested and everything works great, good job ! Just a few comments.

Also before merging, I would rather show the loader animation on the map to @thibautsailly to be sure it's ok for him since he also requested to @Caracol3 the blue loader below the form.

front/src/modules/trainschedule/components/ManageTrainSchedule/Map.tsx
SharglutDev, line 1: Don't think we still need these changes on this map since it is used only for manage train schedule

front/src/modules/trainschedule/components/ManageTrainSchedule/NewMap.tsx
SharglutDev, line 1: Seems like now when the pathfinding is fired, the `computeBbox` happen on the previous path, not the current one (or not compute at all if it's the first pathfinding launched)

[Capture vidéo du 15-01-2025 14:08:24.webm](https://github.com/user-attachments/assets/5eca0120-407f-47b5-a5b6-3fc8bee08614)

front/src/modules/trainschedule/components/ManageTrainSchedule/Map.tsx
```diff
@@ -12,6 +12,7 @@ import { useParams } from 'react-router-dom';
 import captureMap from 'applications/operationalStudies/helpers/captureMap';
 import type { ManageTrainSchedulePathProperties } from 'applications/operationalStudies/types';
 import type { PathProperties } from 'common/api/osrdEditoastApi';
+import { LoaderFill } from 'common/Loaders';
```
SharglutDev:
```suggestion
```

SharglutDev (commented):
Lgtm and tested, one last comment. Just waiting for you to fix conflicts to test one more time and we should be good to go :)

SharglutDev:
This PR can be rebase with last changes from this [one](https://github.com/OpenRailAssociation/osrd/pull/10282) :)

front/src/applications/stdcm/hooks/useStdcmPathfinding.ts
SharglutDev, line 1: When a pathfinding is loading, it seems the loading banner is always green and not blue

SharglutDev, line 1: For a path like Brest > Narbonne > Perpignan (but should work for any path with a via), if I remove the origin, `pathProperties` endpoint is fired again (but it shouldn't, it's not the case right now) and bbox on map is updated with a path containing the via as if it was the origin and the destination.
When checking the network tab in dev tools, we see only `pathProperties` endpoint launching but in its payload, we can see that the `track_section_ranges` doesn't have the same length. So rtk query is probably relaunching `pathfinding blocks` as well in the dark.

SharglutDev (commented):
After the rebase, I have 2 behaviors issues. Both are probably related and coming from the switch from `useLazyQuery` to `useQuery`.

PR #10294, opened 2025-01-09
tests: add ETCS braking curve integration tests for stops only

### LINKED ISSUE #10298: tests: try to merge small_infra BAL and ERTMS
and maybe add TVM?

bougue-pe:
* Added the final stop to get a final braking curve (more realistic) on small_infra test.
* Mutualize small_infra and etcs_infra generation
* Test stop braking curves (rough shape, target position)

🔍 review by commit is advised (especially the infra-generation part)

Fixes https://github.com/OpenRailAssociation/osrd/issues/10298
Fixes part of https://github.com/OpenRailAssociation/osrd/issues/10495


tests/conftest.py
```diff
@@ -37,6 +37,13 @@ def small_infra() -> Iterator[Infra]:
     requests.delete(EDITOAST_URL + f"infra/{infra_id}/")
 
 
+@pytest.fixture(scope="session")
+def etcs_infra() -> Iterator[Infra]:
```
shenriotpro, line 41: how small is it? :p

tests/conftest.py
```diff
@@ -115,9 +132,9 @@ class TestRollingStock:
 TestRollingStock.__test__ = False
 
 
-def create_fast_rolling_stocks(test_rolling_stocks: Optional[List[TestRollingStock]] = None):
+def create_rolling_stock(rolling_stock_json_path: Path, test_rolling_stocks: Optional[List[TestRollingStock]] = None):
```
shenriotpro: nit: can we add a return type?

tests/conftest.py
```diff
@@ -187,6 +213,7 @@ def west_to_south_east_simulation(
                     {"offset": 837034, "track": "TA2", "id": "a"},
                     {"offset": 4386000, "track": "TH1", "id": "b"},
                 ],
+                "schedule": [{"at": "b", "stop_for": "PT0S"}],
```
shenriotpro, line 219: Is there a description of the new infra/rs somewhere?

tests/conftest.py
```diff
@@ -103,6 +116,10 @@ def get_rolling_stock(editoast_url: str, rolling_stock_name: str) -> int:
 
 FAST_ROLLING_STOCK_JSON_PATH = Path(__file__).parents[1] / "editoast" / "src" / "tests" / "example_rolling_stock_1.json"
 
+ETCS_ROLLING_STOCK_JSON_PATH = (
+    Path(__file__).parents[1] / "tests" / "data" / "rolling_stocks" / "etcs_level2_rolling_stock.json"
```
shenriotpro, line 121: it looks like that file is missing, perhaps true for the infra as well?

tests/tests/test_train_schedule.py
```diff
@@ -65,3 +66,27 @@ def test_editoast_delete(west_to_south_east_simulations: Sequence[Any]):
...
+    # TODO: x-fail test: those 2 brake starts MUST move once ETCS braking is plugged
+    speed_section1_const_brake_start_offset = 35718795
+    assert get_current_or_next_speed_at(simulation_final_output, speed_section1_const_brake_start_offset) == 80
+    assert get_current_or_next_speed_at(simulation_final_output, speed_section1_const_brake_start_offset + 1) < 80
+    final_stop_const_brake_start_offset = 43975031
+    assert get_current_or_next_speed_at(simulation_final_output, final_stop_const_brake_start_offset) == 39.444
+    assert get_current_or_next_speed_at(simulation_final_output, final_stop_const_brake_start_offset + 1) < 39.444
+
+
+def get_current_or_next_speed_at(simulation_final_output: Dict[str, Any], position: int) -> int:
```
shenriotpro: nit: if we don't want to export it, let's call it _get_current_or_next_speed_at

tests/conftest.py
shenriotpro: Sorry, I was only looking at 1 commit.

leovalais (approved):

tests/tests/test_train_schedule.py
```diff
@@ -65,3 +66,27 @@ def test_editoast_delete(west_to_south_east_simulations: Sequence[Any]):
...
+    simulation_final_output = response.json()["final_output"]
+
+    assert len(simulation_final_output["positions"]) == len(simulation_final_output["speeds"])
+
+    # TODO: x-fail test: those 2 brake starts MUST move once ETCS braking is plugged
+    speed_section1_const_brake_start_offset = 35718795
+    assert get_current_or_next_speed_at(simulation_final_output, speed_section1_const_brake_start_offset) == 80
+    assert get_current_or_next_speed_at(simulation_final_output, speed_section1_const_brake_start_offset + 1) < 80
+    final_stop_const_brake_start_offset = 43975031
+    assert get_current_or_next_speed_at(simulation_final_output, final_stop_const_brake_start_offset) == 39.444
```
woshilapin: nitpick: Can we (should we) rely on float comparisons for tests?

woshilapin (approved):

tests/conftest.py
bougue-pe: Well, it's the same size than the small_infra... So not so small, to be realistic.
As seen in the recent commits, this ETCS infra is directly derived from small_infra (see railjson_generator scripts).

tests/conftest.py
bougue-pe: I added a little comment for the rolling-stock in https://github.com/OpenRailAssociation/osrd/pull/10294/commits/9dc61c0da6a7e001d8d9f8a7b9bd9c6e47429151
And the infra is described in the railjson-generator script.

tests/tests/test_train_schedule.py
bougue-pe: The tests were completely refactored, but yes, some tolerance was added, thanks for pointing this.

tests/tests/test_train_schedule.py
```diff
@@ -65,3 +67,232 @@ def test_editoast_delete(west_to_south_east_simulations: Sequence[Any]):
...
+
+
+def test_etcs_schedule_result_never_reach_mrsp(etcs_scenario: Scenario, etcs_rolling_stock: int):
+    rolling_stock_response = requests.get(EDITOAST_URL + f"light_rolling_stock/{etcs_rolling_stock}")
+    etcs_rolling_stock_name = rolling_stock_response.json()["name"]
+    ts_response = requests.post(
+        f"{EDITOAST_URL}timetable/{etcs_scenario.timetable}/train_schedule/",
+        json=[
+            {
+                "train_name": "nearby EoAs + brake uphill/downhill grade + dodge LoA break",
```
Erashin: Honestly, if we don't have a single LoA here, i'd rather just not mention them. Or just say `+ no LoAs`.

Edit: ok you want to put the focus on the fact that this simulation has a curve that barely dodges the LoA near the end, hence if at some point it actually hits it due to code changes, it should fail. I feel like there should be a clearer test description then, maybe directly above the test method? Same for the following test as well.

tests/tests/test_train_schedule.py
```diff
@@ -65,3 +67,232 @@ def test_editoast_delete(west_to_south_east_simulations: Sequence[Any]):
...
+
+
+def test_etcs_schedule_result_brake_from_mrsp(etcs_scenario: Scenario, etcs_rolling_stock: int):
+    rolling_stock_response = requests.get(EDITOAST_URL + f"light_rolling_stock/{etcs_rolling_stock}")
+    etcs_rolling_stock_name = rolling_stock_response.json()["name"]
+    ts_response = requests.post(
+        f"{EDITOAST_URL}timetable/{etcs_scenario.timetable}/train_schedule/",
+        json=[
+            {
+                "train_name": "brake from MRSP: max_speed + just after LoA",
```
Erashin: Mentionning LoAs here is good since there are some, but nowhere in the test is it said that we don't actually check them, people might think you've forgotten a case. Since you'll add that check later on (probably in another additional test), it might be less confusing to make it clearer, either in the test description or the train name.

tests/tests/test_train_schedule.py
```diff
@@ -65,3 +67,232 @@ def test_editoast_delete(west_to_south_east_simulations: Sequence[Any]):
...
+    simulation_final_output = simu_response.json()["final_output"]
+
+    assert len(simulation_final_output["positions"]) == len(simulation_final_output["speeds"])
+
+    # To debug this test: please add a breakpoint then use front to display speed-space chart
+    # (activate Context for Slopes and Speed limits).
+
+    # Check that the curves does respect Ends of Authority (EoA = stops), and
+    #   that there is an acceleration then deceleration in between (never reach the MRSP given the acceleration curves).
+    # This check is especially interesting on the first 2 stops that are so close that their braking curves are
```
Erashin: This does not apply to this test case (13km difference between the first 2 stops).

tests/tests/test_train_schedule.py
```diff
@@ -65,3 +67,232 @@ def test_editoast_delete(west_to_south_east_simulations: Sequence[Any]):
...
+        for pos_index in range(start_pos_index, end_pos_index):
+            current_speed = simulation_final_output["speeds"][pos_index]
+            if accelerating:
+                if prev_speed > current_speed:
+                    accelerating = False
+            else:
+                assert prev_speed >= current_speed
+            prev_speed = current_speed
+
+    # Check that the braking curves for the 2 middle stops start at the expected offset
```
Erashin: You've named the stop offsets first second and third, might be better to use that to specify which stops you're referring to -> `for the first 2 stops` or `first and second stops`.

Erashin (approved):
Ticks all the EoA test cases. This will probably be affected by incoming SvL code changes, kinda depends on how the infra is built.
Thanks for that!

shenriotpro (approved):

tests/tests/test_train_schedule.py
bougue-pe: Indeed, changed the name of the tests and the trains + added a little comment on LoAs in https://github.com/OpenRailAssociation/osrd/pull/10294/commits/22ab1abb3fa44ab5e8d5befe7920bf48de1f506a

tests/tests/test_train_schedule.py
bougue-pe: Right, leftover of a copy-paste, fixed in https://github.com/OpenRailAssociation/osrd/pull/10294/commits/22ab1abb3fa44ab5e8d5befe7920bf48de1f506a

tests/tests/test_train_schedule.py
bougue-pe: Done in https://github.com/OpenRailAssociation/osrd/pull/10294/commits/22ab1abb3fa44ab5e8d5befe7920bf48de1f506a

tests/tests/test_train_schedule.py
bougue-pe: Tried to clarify this in https://github.com/OpenRailAssociation/osrd/pull/10294/commits/22ab1abb3fa44ab5e8d5befe7920bf48de1f506a

PR #10302, opened 2025-01-09
Change name for interne deployment

### LINKED ISSUE #10281: Personnaliser les déploiements internes avec les nouveaux noms
https://github.com/osrd-project/osrd-confidential/issues/842

Uriel-Sautron:
close #10281 
For internal deployment we need to display the LMR and Horizon logos.

For test:
  - create a folder `overrides` at the front root
  - add in this folder the images you find [here ](https://github.com/osrd-project/osrd-confidential/issues/842#issuecomment-2454430710) and `overrides.json` 
  
```json
// overrides.json
{
   "names": {
     "stdcm": "Last Minute Request",
     "digital_twin": "Horizon"
   },
 "icons": {
    "stdcm": {
      "light": "Logotype_LastMinuteRequest_light",
      "dark": "Logotype_LastMinuteRequest_dark"
    },
    "digital_twin": {
      "light": "Logotype_Horizon_light",
      "dark": "Logotype_Horizon_dark"
    }
  }
} 
```


front/src/utils/hooks/useLogo.ts
```diff
@@ -0,0 +1,58 @@
...
+const useLogo = () => {
+  const [digitalTwin, setDigitalTwin] = useState<{ logo: string; digitalTwinLogo: string }>({
+    logo: defaultLogo,
+    digitalTwinLogo: defaultOsrdLogo,
+  });
+  const [stdcm, setStdcm] = useState<{ stdcmLogo?: string; stdcmPngLogo: string }>({
+    stdcmLogo: undefined,
+    stdcmPngLogo: defaultLogo,
+  });
+  const [isIntenalProd, setIsIntenalProd] = useState(false);
```
ghost: Maybe we should not name this internalProd, but "customizedDeployment", it seems more accurate.

front/src/applications/stdcm/components/StdcmResults/SimulationReportSheet.tsx
```diff
@@ -15,6 +16,7 @@ import type { SimulationReportSheetProps } from '../../types';
 import { getStopDurationTime } from '../../utils/formatSimulationReportSheet';
 
 const getSecondaryCode = ({ location }: StdcmPathStep) => location!.secondary_code;
+const lmrLogoPath = '../../overrides/Logotype_LastMinuteRequest_light@2x.png';
```
ghost: We should not reference any override directly in terms of path, they are dynamically defined by the JSON.

ghost (changes requested):
Thanks for the PR !

In my understanding of the front (which might get a bit rusty over time) I think we should reference directly only the files that are present in the code base (the default logo) and switch if the override file exists on the server and defines the fields on the json.

front/src/applications/stdcm/components/StdcmHeader.tsx
```diff
@@ -20,10 +28,11 @@ const StdcmHeader = ({
 }: StdcmHeaderProps) => {
   const { t } = useTranslation(['stdcm', 'translation']);
   const isSuperUser = useSelector(getIsSuperUser);
+  const { stdcm } = useLogo();
 
   return (
     <div className="stdcm-header d-flex">
-      <span className="stdcm-header__title pl-5">ST DCM</span>
+      <LogoSTDCM logo={stdcm.stdcmLogo} />
```
Yohh: ~~it seems that `stdcm-header__title` isn't used anymore, you can remove `__title` from `__header.scss` I guess~~

nevermind, it's still used

front/src/utils/hooks/useLogo.ts
Uriel-Sautron: Done

front/src/applications/stdcm/components/StdcmResults/SimulationReportSheet.tsx
Uriel-Sautron: Done

front/src/utils/hooks/useLogo.ts
```diff
@@ -0,0 +1,61 @@
...
+          if (new Date().getMonth() === MONTH_VALUES.DECEMBER) {
+            setDigitalTwin({ logo: xmasLogo, digitalTwinLogo: xmasOsrdLogo });
+          }
+        } else {
+          const overridesData = await response.json();
+          const { icons } = overridesData;
+
+          const lmrLogoPath = `../../overrides/${icons.stdcm.light}.svg`;
+          const lmrPngLogoPath = `../../overrides/${icons.stdcm.light}@2x.png`;
+          const horizonFullLogoPath = `../../overrides/${icons.digital_twin.dark}.svg`;
```
ghost: Any reasons to use ../../overrides vs /overrides ?

ghost (approved):
Overall LGTM, thanks !

front/src/utils/hooks/useLogo.ts
Uriel-Sautron: You're right it's better, thank's.

Uriel-Sautron: Done.

front/src/common/BootstrapSNCF/NavBarSNCF.tsx
```diff
@@ -24,17 +25,25 @@ type Props = {
   logo?: string;
```
SharglutDev, line 25: Since like we don't use this prop anymore with your changes

```suggestion
```

front/src/utils/hooks/useLogo.ts
```diff
@@ -0,0 +1,61 @@
...
+            setDigitalTwin({ logo: xmasLogo, digitalTwinLogo: xmasOsrdLogo });
+          }
+        } else {
+          const overridesData = await response.json();
+          const { icons } = overridesData;
+
+          const lmrLogoPath = `/overrides/${icons.stdcm.light}.svg`;
+          const lmrPngLogoPath = `/overrides/${icons.stdcm.light}@2x.png`;
+          const horizonFullLogoPath = `/overrides/${icons.digital_twin.dark}.svg`;
+          const horizonLogoPath = `/overrides/${icons.digital_twin.dark}_logo.svg`;
```
SharglutDev: What's the difference between these 2 ?

front/src/utils/hooks/useLogo.ts
```diff
@@ -0,0 +1,61 @@
...
+          setCustomizedDeployment(false);
+          if (new Date().getMonth() === MONTH_VALUES.JUNE) {
+            setDigitalTwin({ logo: proudLogo, digitalTwinLogo: proudOsrdLogo });
+          }
+          if (new Date().getMonth() === MONTH_VALUES.DECEMBER) {
+            setDigitalTwin({ logo: xmasLogo, digitalTwinLogo: xmasOsrdLogo });
+          }
+        } else {
+          const overridesData = await response.json();
+          const { icons } = overridesData;
```
SharglutDev: It shouldn't happen but maybe we should handle the case where `icons` is undefined

front/src/utils/hooks/useLogo.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const MONTH_VALUES = {
+  JUNE: 5,
+  DECEMBER: 11,
+};
+
+const useLogo = () => {
+  const [digitalTwin, setDigitalTwin] = useState<{ logo: string; digitalTwinLogo: string }>({
+    logo: defaultLogo,
+    digitalTwinLogo: defaultOsrdLogo,
```
SharglutDev: We are already in `digitalTwin` property, maybe we could rename it like that ?
```suggestion
    logoWithName: defaultOsrdLogo,
```

front/src/applications/stdcm/components/StdcmResults/SimulationReportStyleSheet.ts
```diff
@@ -42,6 +42,12 @@ const styles = {
       height: '24',
       marginLeft: '20',
     },
+    lmrLogo: {
+      width: 'auto',
+      height: 'auto',
+      marginTop: 'auto',
+      marginRight: 'auto',
+    },
```
SharglutDev: Are these necessary ?

front/src/utils/hooks/useLogo.ts
Uriel-Sautron: I come back to you after discussion with @thibautsailly

front/src/utils/hooks/useLogo.ts
Uriel-Sautron: horizonFullLogoPath is logo with name

front/src/utils/hooks/useLogo.ts
Uriel-Sautron: Done

front/src/applications/stdcm/components/StdcmResults/SimulationReportStyleSheet.ts
Uriel-Sautron: I removed the height and marginTop properties, but width and marginRight are necessary

front/src/utils/hooks/useLogo.ts
```diff
@@ -0,0 +1,61 @@
...
+const useLogo = () => {
+  const [digitalTwin, setDigitalTwin] = useState<{ logo: string; logoWithName: string }>({
+    logo: defaultLogo,
+    logoWithName: defaultOsrdLogo,
+  });
+  const [stdcm, setStdcm] = useState<{ stdcmLogo?: string; stdcmPngLogo?: string }>({
+    stdcmLogo: undefined,
+    stdcmPngLogo: undefined,
+  });
+  const [customizedDeployment, setCustomizedDeployment] = useState(false);
```
clarani: You don't need these 3 states values, one which contains everything is enough and avoid extra re-renderings

front/src/utils/hooks/useLogo.ts
```diff
@@ -0,0 +1,61 @@
...
+          const lmrPngLogoPath = `/overrides/${icons.stdcm.light}@2x.png`;
+          const horizonFullLogoPath = `/overrides/${icons.digital_twin.dark}.svg`;
+          const horizonLogoPath = `/overrides/${icons.digital_twin.dark}_logo.svg`;
+
+          setCustomizedDeployment(true);
+          setStdcm({ stdcmLogo: lmrLogoPath, stdcmPngLogo: lmrPngLogoPath });
+          setDigitalTwin({ logo: horizonLogoPath, logoWithName: horizonFullLogoPath });
+        }
+      } catch {
+        setCustomizedDeployment(false);
```
clarani: At this step, if the await response.json() failed, customizedDeployment is still false, so you don't need to update it again.
Maybe you can add a console.error indicating that we failed to load the logos ?
```suggestion
```

front/src/utils/hooks/useLogo.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+const useLogo = () => {
+  const [digitalTwin, setDigitalTwin] = useState<{ logo: string; logoWithName: string }>({
+    logo: defaultLogo,
+    logoWithName: defaultOsrdLogo,
+  });
+  const [stdcm, setStdcm] = useState<{ stdcmLogo?: string; stdcmPngLogo?: string }>({
+    stdcmLogo: undefined,
+    stdcmPngLogo: undefined,
```
clarani: Maybe stdcmSimulationSheetLogo would be a better name ?
```suggestion
    stdcmSimulationSheetLogo: undefined,
```

front/src/utils/hooks/useLogo.ts
```diff
@@ -0,0 +1,61 @@
...
+          setDigitalTwin({ logo: horizonLogoPath, logoWithName: horizonFullLogoPath });
+        }
+      } catch {
+        setCustomizedDeployment(false);
+      }
+    };
+    fetchInternalProd();
+  }, []);
+
+  return { digitalTwin, stdcm, customizedDeployment };
```
clarani: It would be better to return a plain object like this:
```suggestion
  return { digitalTwinLogo, digitalTwinLogoWithName, stdcmLogo, stdcmSimulationSheetLogo, customizedDeployment };
```

front/src/common/BootstrapSNCF/NavBarSNCF.tsx
```diff
@@ -24,17 +25,25 @@ type Props = {
...
 
-const LegacyNavBarSNCF = ({ appName, logo = getLogo() }: Props) => {
+const LegacyNavBarSNCF = ({ appName, logo }: Props) => {
   const { openModal } = useModal();
+  const { digitalTwin, customizedDeployment } = useLogo();
   const safeWord = useSelector(getUserSafeWord);
   const { t } = useTranslation('home/navbar');
   const { logout, username } = useAuth();
 
+  const headerLogo = logo ?? digitalTwin?.logo;
```
clarani: `digitalTwin` is always defined since you initialized in `useLogo`. 
- you can remove the `?` 
- you can remove the test line 41 since logo will always be defined
- I think this const is not needed, you can directly use the image like this:
  ``
```suggestion
  const headerLogo = logo ?? digitalTwin.logo;
```

front/src/common/BootstrapSNCF/NavBarSNCF.tsx
```diff
@@ -24,17 +25,25 @@ type Props = {
...
-      <div className="mastheader-logo flex-grow-0">
+      <div
+        className={cx(
+          customizedDeployment && logo ? `mastheader-logo__horizon` : `mastheader-logo`,
+          `flex-grow-0`
+        )}
+      >
         <Link to="/">
-          <img src={logo} data-testid="osrd-logo" alt="OSRD Logo" />
+          <img src={headerLogo} data-testid="osrd-logo" alt="OSRD Logo" />
```
clarani: You need to change the data-test-id and the alt

front/src/utils/hooks/useLogo.ts
Uriel-Sautron: Done

front/src/common/BootstrapSNCF/NavBarSNCF.tsx
Uriel-Sautron: Done

front/src/common/BootstrapSNCF/NavBarSNCF.tsx
Uriel-Sautron: Done

front/src/utils/hooks/useLogo.ts
Uriel-Sautron: Done

front/src/utils/hooks/useLogo.ts
Uriel-Sautron: Done

front/src/utils/hooks/useLogo.ts
Uriel-Sautron: Done

ghost:
Tested on internal deployments, works properly. LGTM.

front/src/utils/hooks/useDeploymentSettings.ts
```diff
@@ -0,0 +1,90 @@
...
+          setCustomizedDeploymentSetting((prev) => ({
+            ...prev,
+            digitalTwinName: names.digital_twin,
+            digitalTwinLogo: horizonLogoPath,
+            digitalTwinLogoWithName: horizonLogoWithNamePath,
+            stdcmName: names.stdcm,
+            stdcmLogo: lmrLogoPath,
+            stdcmSimulationSheetLogo: lmrPngLogoPath,
+            isCustomizedDeployment: true,
+          }));
```
clarani: Here you completely re-write the whole state, so you don't need to use `prev`
```suggestion
          setCustomizedDeploymentSetting{
            digitalTwinName: names.digital_twin,
            digitalTwinLogo: horizonLogoPath,
            digitalTwinLogoWithName: horizonLogoWithNamePath,
            stdcmName: names.stdcm,
            stdcmLogo: lmrLogoPath,
            stdcmSimulationSheetLogo: lmrPngLogoPath,
            isCustomizedDeployment: true,
          });
```

front/src/utils/hooks/useDeploymentSettings.ts
```diff
@@ -0,0 +1,90 @@
...
+              digitalTwinLogoWithName: proudOsrdLogo,
+            }));
+          }
+          if (new Date().getMonth() === MONTH_VALUES.DECEMBER) {
+            setCustomizedDeploymentSetting((prev) => ({
+              ...prev,
+              digitalTwinLogo: xmasLogo,
+              digitalTwinLogoWithName: xmasOsrdLogo,
+            }));
+          }
```
clarani: It would be better to have only 1 set at the end of this block of code to avoid extra re-renders
```suggestion
          let digitalTwinLogo = defaultLogo;
          let digitalTwinLogoWithName = defaultOsrdLogo;
          const currentMonth = new Date().getMonth();
          
          if (currentMonth === MONTH_VALUES.JUNE) {
            digitalTwinLogo = proudLogo;
            digitalTwinLogoWithName = proudOsrdLogo;
          }
          else if (currentMonth === MONTH_VALUES.DECEMBER) {
            digitalTwinLogo = xmasLogo;
            digitalTwinLogoWithName = xmasOsrdLogo;
          }
          
          setCustomizedDeploymentSetting((prev) => ({
            ...prev,
            isCustomizedDeployment: false,
            digitalTwinLogo,
            digitalTwinLogoWithName
          }));
```

front/src/utils/hooks/useDeploymentSettings.ts
```diff
@@ -0,0 +1,90 @@
...
+          }));
+        }
+      } catch (error) {
+        console.error('Error fetching overrides.json', error);
+      }
+    };
+    fetchInternalProd();
+  }, []);
+
+  return { ...customizedDeploymentSetting };
```
clarani: No need to destructure it
```suggestion
  return customizedDeploymentSetting;
```

front/src/utils/hooks/useDeploymentSettings.ts
Uriel-Sautron: Done

front/src/utils/hooks/useDeploymentSettings.ts
Uriel-Sautron: Done

front/src/utils/hooks/useDeploymentSettings.ts
Uriel-Sautron: Done

clarani (approved):
LGTM :white_check_mark:

flomonster (commented):
There are several occurrences of `lmr` which is not (yet) used for the open source part. Ideally, we should only use `stdcm` for consistency.

Maymanaf (approved):

Issue #10000, opened 2024-12-10
Clean up and enhance simulation report sheet


achrafmohye:
Implements the Clean up and enhance simulation report sheet [feature](https://github.com/osrd-project/osrd-confidential/issues/848)

axrolld:
I found this tiny tiny bug but i think it would be nice to fix it before the release. Those labels should not start with a capital letter according to the mockup:
![Image](https://github.com/user-attachments/assets/3f3d5a10-7029-41cd-8ece-4f4e492623b7)

axrolld:
LGTM thanks for the fix !

PR #10308, opened 2025-01-09
Fix table headers to start with lowercase letters for consistency

achrafmohye:
This is a quick fix for this [comment](https://github.com/OpenRailAssociation/osrd/issues/10000#issuecomment-2580259798) in #10000


front/public/locales/en/stdcm.json
```diff
@@ -4,7 +4,7 @@
   "cancelRequest": "Cancel request",
   "consist": {
     "consist": "Consist",
-    "length": "Length",
+    "length": "length",
```
clarani: Hum this translation is also used in the consist card, I don't think it's a good idea to modify it here.
It would be better to add a new translation

image

achrafmohye: Good call, i'm on it 👍

achrafmohye: Done

clarani (approved):
LGTM :white_check_mark:

PR #10312, opened 2025-01-09
core: fix max speed in concatenated envelopes

eckter:
When there was a single envelope, we'd actually return the maximum value over the whole envelope, ignoring ranges.

This caused a nasty bug in stdcm where the "max speed" would be higher in post-processing than during the graph traversal, resulting in conflict mismatches (when using TVM)

### DIFF
--- a/core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/EnvelopeConcat.java
+++ b/core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/EnvelopeConcat.java
@@ -1,6 +1,8 @@
 package fr.sncf.osrd.envelope;
 
 import static fr.sncf.osrd.envelope_utils.DoubleUtils.clamp;
+import static java.lang.Math.max;
+import static java.lang.Math.min;
 
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/EnvelopeConcat.java
```diff
@@ -158,9 +160,11 @@ public double maxSpeedInRange(double beginPos, double endPos) {
         assert firstEnvelopeIndex != -1 : "Trying to interpolate time outside of the envelope";
         var firstEnvelope = envelopes.get(firstEnvelopeIndex);
         var beginSpeed = firstEnvelope.envelope.maxSpeedInRange(
```
Erashin: `firstEnvelopeSpeed` and `lastEnvelopeSpeed` (or even `firstEnvelopeMaxSpeed`, but it gets lengthy)? Anything to make it clearer that these are the max speeds in the given range on the first and last envelopes.

Erashin (approved):
LGTM.

PR #10313, opened 2025-01-09
front: fix stdcm consist unwanted warning

### LINKED ISSUE #10306: Picking some rolling stocks triggers unwanted warning
### What happened?

Picking MAL100 as a towed rolling stocks triggers this:
Image

There seems to be a tolerance problem here because the value is correct.

I'm expecting no warning if the value is within the acceptable range.

1. Select MAL100 as a towed rolling stock

Recette (SNCF)

Google Chrome

fdfd8ac

SharglutDev:
Round to bottom the conversion to ton of the sum of the rolling stock and towed rolling stock mass to match the behavior of the tonnage input.

fix #10306

### DIFF
--- a/front/src/applications/stdcm/utils/consistValidation.ts
+++ b/front/src/applications/stdcm/utils/consistValidation.ts
@@ -22,7 +22,7 @@ export const validateTotalMass = ({
   }
 
   const tractionMassInTons = kgToT(tractionEngineMass);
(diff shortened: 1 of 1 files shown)


Wadjetz (approved):

kmer2016 (approved):
LGTM !

PR #10316, opened 2025-01-10
core: add ETCS LOA logic to ETCS braking simulator

### LINKED ISSUE #10330: core: ETCS: implement LOA braking curves logic
Follow up to #9704: in the end, only EOAs were implemented in that issue.
Implement ETCS braking curves for LOAs.

Restriction: simplification: TODOs were added:
- maxVbecDeltaSpeed is not calculated and is 50km/h. Hence, we calculate the EBD curve until `envelope.maxSpeed + 50 / 3.6`. This is less important until we actually get a bug where the actual maxVbecDeltaSpeed > 50km/h: unlikely.
- Finding EBD foot is done by just shifting the envelope: approximation on the gradients.

Erashin:
LOA logic (see Subset given in code): compute indication curve from EBD.
To do so, couple of technicalities:
- EBD foot is not (targetPos, targetSpeed). It is a point after targetPos, (targetPos + delta, targetSpeed) so that the EBD passes by the point (targetPos, targetSpeed + dvEbi).
- The indication curve reaches the target speed before the LOA: create maintain target speed curve until target position.
- We build the LOA braking curves along the path, so that, same as EOA, we handle overlaps of the braking curves. However, we must handle intersections differently, since here, the LOAs don't all have the same target speed. To do so, build the LOA braking curve into the envelope at each step, so that it is used as a ceiling for the following LOAs.
- The indication curve positions can all be negative, if the LOA is close to the path start: handle the case so we still add a maintain speed from the start of the path until the LOA (second screenshot).

image

image

### DIFF
--- a/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/Constants.kt
+++ b/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/Constants.kt
@@ -1,6 +1,69 @@
 package fr.sncf.osrd.envelope_sim.etcs
 
-/** See Subset referenced in ETCSBrakingSimulator: table in Appendix A.3.1. */
+/**
+ * Formulas are found in `SUBSET-026-3v400.pdf` from the file at
+ * https://www.era.europa.eu/system/files/2023-09/index004_-_SUBSET-026_v400.zip and in
+ * `SUBSET-041_v400.pdf` from the file at
+ * https://www.era.europa.eu/system/files/2023-09/index014_-_SUBSET-041_v400.pdf
+ */
+
+/**
+ * National Default Value: permission to inhibit the compensation of the speed measurement accuracy.
+ * See Subset 026: table in Appendix A.3.2.
+ */
+const val qNvinhsmicperm = false
+
+/**
+ * Estimated acceleration during tBerem, worst case scenario (aEst2 is between 0 and 0.4), expressed
+ * in m/s². See Subset 026: §3.13.9.3.2.9.
+ */
+const val aEst2 = 0.4
+
+/** See Subset 026: table in Appendix A.3.1. */
+const val dvEbiMin = 7.5 / 3.6 // m/s
+const val dvEbiMax = 15.0 / 3.6 // m/s
+const val vEbiMin = 110.0 / 3.6 // m/s
+const val vEbiMax = 210.0 / 3.6 // m/s
+const val tWarning = 2.0 // s
 const val tDriver = 4.0 // s
 const val mRotatingMax = 15.0 // %
 const val mRotatingMin = 2.0 // %
+
+/** See Subset 041: §5.3.1.2. */
+const val vUraMinLimit = 30 / 3.6 // m/s
+const val vUraMaxLimit = 500 / 3.6 // m/s
+const val vUraMin = 2 / 3.6 // m/s
+const val vUraMax = 12 / 3.6 // m/s
+
+/** See Subset 041: §5.3.1.2. */
+fun vUra(speed: Double): Double {
+    return interpolateLinearSpeed(speed, vUraMinLimit, vUraMaxLimit, vUraMin, vUraMax)
+}
+
+/** See Subset 026: §3.13.9.3.2.10. */
+fun vDelta0(speed: Double): Double {
+    return if (!qNvinhsmicperm) vUra(speed) else 0.0
+}
+
+/** See Subset 026: §3.13.9.2.3. */
+fun dvEbi(speed: Double): Double {
+    return interpolateLinearSpeed(speed, vEbiMin, vEbiMax, dvEbiMin, dvEbiMax)
+}
+
+/**
+ * The linear curve is the following: below minSpeedLimit = minSpeed, above maxSpeedLimit =
+ * maxSpeed, in between is a linear curve. This method takes a speed input and converts it
+ * accordingly.
+ */
+private fun interpolateLinearSpeed(
+    speed: Double,
+    minSpeedLimit: Double,
+    maxSpeedLimit: Double,
+    minSpeed: Double,
+    maxSpeed: Double
+): Double {
+    return if (speed <= minSpeedLimit) minSpeed
+    else if (speed < maxSpeedLimit)
+        (maxSpeed - minSpeed) / (maxSpeedLimit - minSpeedLimit) * (speed - minSpeedLimit) + minSpeed
+    else maxSpeed
+}
--- a/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
+++ b/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
@@ -14,6 +14,10 @@ import fr.sncf.osrd.envelope_sim.overlays.EnvelopeDeceleration
 import kotlin.math.max
 import kotlin.math.min
 
+/**
+ * Formulas are found in `SUBSET-026-3v400.pdf` from the file at
+ * https://www.era.europa.eu/system/files/2023-09/index004_-_SUBSET-026_v400.zip
+ */
 enum class BrakingCurveType {
     EBD, // Emergency Brake Deceleration
     EBI, // Emergency Brake Intervention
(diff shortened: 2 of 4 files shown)


core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/Constants.kt
```diff
@@ -1,6 +1,44 @@
...
 const val mRotatingMax = 15.0 // %
 const val mRotatingMin = 2.0 // %
+
+/** See https://www.era.europa.eu/system/files/2023-09/index014_-_SUBSET-041_v400.pdf: §5.3.1.2. */
+fun vUra(speed: Double): Double {
+    return if (speed <= 30 / 3.6) 2 / 3.6
+    // vUra(30km/h) = 2km/h & vUra(500km/h) = 12km/h with a linear interpolation in between
+    // this gives the following equation : y = (x + 64) / 47, still in km/h
+    else ((speed + 64) / 47) / 3.6
+}
```
eckter: What are the input/output units?

I'm expecting m/s for both, so id the formula is in km/h we should probably convert `speed`

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/Constants.kt
```diff
@@ -1,6 +1,44 @@
...
+
+/** See Subset referenced in ETCSBrakingSimulator: §3.13.9.3.2.10. */
```
eckter: We keep referencing the pdf everywhere, there's already a link in the code somewhere but we should probably repeat it in every relevant file. I can be confusing to start with this file and see § everywhere, while having to look through the module to find what it means.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -60,28 +69,102 @@ fun addBrakingCurvesAtEOAs(
                 targetSpeed,
                 BrakingType.ETCS_GUI
             )
-        val sbdCurve =
+
+        val fullIndicationCurve =
+            computeIndicationBrakingCurveFromRef(context, sbdCurve, BrakingCurveType.SBD, guiCurve)
+        // Indication curve for EOAs will always have at least 2 points with positive positions.
+        // This method cannot
+        // return a null.
```
eckter:
```suggestion
        // Indication curve for EOAs will always have at least 2 points with positive positions.
        // This method cannot return a null.
```

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -93,9 +176,14 @@ private fun computeBrakingCurve(
...
-    // If the stopPosition is below begin position, the input is invalid
+    assert(
+        brakingType == BrakingType.ETCS_EBD ||
+            brakingType == BrakingType.ETCS_SBD ||
+            brakingType == BrakingType.ETCS_GUI
+    )
+    // If the stopPosition is below begin position, the input is invalid.
     // If the stopPosition is after the end of the path, the input is invalid except if it is an
-    // SVL, i.e. the target speed is 0 and the curve to compute is an EBD
+    // SVL, i.e. the target speed is 0 and the curve to compute is an EBD.
```
eckter, line 226: If the point can be after the path end *only if the speed is 0*, then I think the condition is backwards.

It could be clearer if it's split into several parts.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -125,10 +213,65 @@ private fun computeBrakingCurve(
...
-    val brakingCurve = partBuilder.build()
+    var brakingCurve = partBuilder.build()
+
+    if (brakingType == BrakingType.ETCS_EBD && targetSpeed != 0.0) {
+        // TODO: by doing this, there is an approximation on the gradient used. TBD at a later date.
+        // When target is an LOA, EBD reaches target position at target speed + dVEbi: shift
+        // envelope to make it so. See Subset referenced in ETCSBrakingSimulator: §3.13.8.3.1,
+        // figure 40.
+        val dvEbi = dvEbi(targetSpeed)
+        val intersection = brakingCurve.interpolatePosition(targetSpeed + dvEbi)
```
eckter: Are we certain that `targetSpeed + dvEbi <= brakingCurve.maxSpeed`?

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -125,10 +213,65 @@ private fun computeBrakingCurve(
...
+        val deltaPos = deltaPosAndDeltaSpeed.component1()
+        val deltaSpeed = deltaPosAndDeltaSpeed.component2()
+        val newPos = ebdCurve.getPointPos(i) - deltaPos
+        val newSpeed = speed - deltaSpeed
+        reversedNewPos.add(newPos)
+        reversedNewSpeeds.add(newSpeed)
+    }
+
+    val nbPoints = reversedNewPos.size
+    val newPosArray = DoubleArray(nbPoints)
```
eckter:
```suggestion
    val newPosArray = reversedNewPos.toDoubleArray()
```

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -216,16 +342,76 @@ private fun keepBrakingCurveUnderOverlay(
...
+/**
+ * Compute the position and speed offsets between EBD and EBI curves, for a given speed. See Subset
+ * referenced in ETCSBrakingSimulator: 3.13.9.3.2.
+ */
+private fun computeBecDeltaPosAndSpeed(
+    context: EnvelopeSimContext,
+    ebd: EnvelopePart,
+    speed: Double,
+    targetSpeed: Double
+): Pair<Double, Double> {
```
eckter: Could return a simple data class

eckter (commented):
Nice job! I only have some minor comments.


I haven't double-checked the formulas as much as I'd have liked to, it's quite difficult to get familiar enough with the reference doc.


Do you think we could add some `asserts` here and there? It would be valuable in this context.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Technically yeah. `dvEbiMax = 15 km/h`, and we compute the EBD until `envelope.MaxSpeed + 50 km/h`. so even in the worst case scenario, we'll still be abovec `targetSpeed + dvEbi` by at least 35 km/h :)

EDIT: ok nvm. We might have a bug. Basically, EBD curve HAS to stop at 0.0 if it reaches it because otherwise we'll try to call tractive effort curve map for a position which is located before the path -> not gonna work.
Solution: approximation: we don't shift if we don't intersect with dvEbi. This will result in an approx of ebd curves, but only at the start of the path. The impact should not be seen, as we should accelerate and hit the mrsp rather than the approximate etcs braking curves we'll have computed.
Fixed in https://github.com/OpenRailAssociation/osrd/pull/10316/commits/f95b0b10bae839ded2e5f38cb9ba7b408aec2ad9

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/Constants.kt
Erashin: I just noticed this is the same as how we compute dvEbi. I'm just gonna merge both methods, way easier, and it should be easier to understand.
Fixed in https://github.com/OpenRailAssociation/osrd/pull/10316/commits/f95b0b10bae839ded2e5f38cb9ba7b408aec2ad9

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Yeah the condition is incorrect. It should throw if :
before path or (after path and notSvl), i.e. before path or (after path and !(target speed = 0 and EBD)).
Fixed in https://github.com/OpenRailAssociation/osrd/pull/10316/commits/f95b0b10bae839ded2e5f38cb9ba7b408aec2ad9

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: The formatter isn't the greatest at managing comments sometimes 😅
Fixed in https://github.com/OpenRailAssociation/osrd/pull/10316/commits/f95b0b10bae839ded2e5f38cb9ba7b408aec2ad9

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: They aren't reversed anymore as well, so I'll rename the value and simplify the code.
Fixed in https://github.com/OpenRailAssociation/osrd/pull/10316/commits/f95b0b10bae839ded2e5f38cb9ba7b408aec2ad9

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/Constants.kt
Erashin: Fixed in https://github.com/OpenRailAssociation/osrd/pull/10316/commits/f95b0b10bae839ded2e5f38cb9ba7b408aec2ad9

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Fixed in https://github.com/OpenRailAssociation/osrd/pull/10316/commits/f95b0b10bae839ded2e5f38cb9ba7b408aec2ad9

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -38,69 +42,185 @@ fun addBrakingCurvesAtEOAs(
...
+    val beginPos = envelope.beginPos
+    var envelopeWithLoaBrakingCurves = envelope
+    var builder = OverlayEnvelopeBuilder.forward(envelopeWithLoaBrakingCurves)
+    for (limitOfAuthority in sortedLimitsOfAuthority) {
+        val targetPosition = limitOfAuthority.offset.distance.meters
+        val targetSpeed = limitOfAuthority.speed
+        val maxSpeedEnvelope = envelopeWithLoaBrakingCurves.maxSpeed
+
+        // Add maxBecDeltaSpeed to EBD curve overlay so it reaches sufficiently high speeds to
+        // guarantee that the corresponding EBI curve intersects with envelope max speed.
```
bougue-pe: nit (keep only what you want, it's fine as-is if you prefer):
```suggestion
        // Add maxBecDeltaSpeed to EBD curve overlay so it reaches sufficiently high speed to
        // guarantee that, after speed translation, the corresponding EBI curve does intersect with envelope max speed.
```

bougue-pe (approved):
Did not review LoA precisely (will do even if merged, improving late is better than nothing on the matter).

It's OK to merge IMO as long as we can hand-test this, and as long as it doesn't break "normal" braking. So please do if I don't have the time to review LoA before the deadline 🙏

eckter (approved):

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -38,69 +42,191 @@ fun addBrakingCurvesAtEOAs(
...
             )
+        val sbdCurve =
+            computeBrakingCurve(
+                context,
+                overhead,
+                targetPosition,
+                targetSpeed,
+                BrakingType.ETCS_SBD
+            )
+        assert(
```
bougue-pe, line 67: nit: It should be easier to debug if the assertions are separated (ditto for all)

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -38,69 +42,191 @@ fun addBrakingCurvesAtEOAs(
...
+        val maxBecDeltaSpeed = maxBecDeltaSpeed()
+        val maxSpeedEbd = maxSpeedEnvelope + maxBecDeltaSpeed
+        val overhead =
+            Envelope.make(
+                EnvelopePart.generateTimes(
+                    listOf(EnvelopeProfile.CONSTANT_SPEED),
+                    doubleArrayOf(0.0, targetPosition),
+                    doubleArrayOf(maxSpeedEbd, maxSpeedEbd)
+                )
+            )
```
bougue-pe, line 140: nit: this (or part of) may stay out of the loop (the `overhead` would be as long as the path if so).

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -38,69 +42,191 @@ fun addBrakingCurvesAtEOAs(
...
+        if (indicationCurve != null) {
+            assert(
+                indicationCurve.beginSpeed <= maxSpeedEnvelope &&
+                    indicationCurve.endSpeed == targetSpeed &&
+                    indicationCurve.endPos <= targetPosition
+            )
+            builder.addPart(indicationCurve)
+        }
+
+        val endOfIndicationCurve = indicationCurve?.endPos ?: beginPos
```
bougue-pe, line 191: If true, not sure about how to solve this, but: Can't we get a "no intersection" case when there is a super-short raise of the MRSP?
Example of MRSP(m, km/h) being: (0, 30), (100, 30), (100, 300), (101, 300), (101, 100), (2000, 100).

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -38,69 +42,191 @@ fun addBrakingCurvesAtEOAs(
...
+                fullIndicationCurve,
+                envelopeWithLoaBrakingCurves,
                 beginPos
             )
-        val indicationCurve = keepBrakingCurveUnderOverlay(fullIndicationCurve, envelope, beginPos)
-        builder.addPart(indicationCurve)
-        beginPos = targetPosition
+
+        if (indicationCurve != null) {
+            assert(
```
bougue-pe, line 183: Can't we also:
```kt
assert(indicationCurve.beginPos >= 0)
```

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -38,69 +42,191 @@ fun addBrakingCurvesAtEOAs(
...
+    assert(
+        brakingType == BrakingType.ETCS_EBD ||
+            brakingType == BrakingType.ETCS_SBD ||
+            brakingType == BrakingType.ETCS_GUI
+    )
     // If the stopPosition is after the end of the path, the input is invalid except if it is an
-    // SVL, i.e. the target speed is 0 and the curve to compute is an EBD
+    // SVL, i.e. the target speed is 0 and the curve to compute is an EBD.
     if (
-        targetPosition <= beginPos ||
```
bougue-pe, line 100: Can't we keep that assert?

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -125,22 +251,73 @@ private fun computeBrakingCurve(
...
     )
-    val brakingCurve = partBuilder.build()
+    var brakingCurve = partBuilder.build()
+
+    if (brakingType == BrakingType.ETCS_EBD && targetSpeed != 0.0) {
+        // TODO: by doing this, there is an approximation on the gradient used. TBD at a later date.
+        // When target is an LOA, EBD reaches target position at target speed + dVEbi: shift
+        // envelope to make it so. See Subset 026: §3.13.8.3.1, figure 40.
+        val dvEbi = dvEbi(targetSpeed)
+        val vTargetPosition = targetSpeed + dvEbi
```
bougue-pe, line 261: Shouldn't this be named "vTarget`Speed`"?

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -125,22 +251,73 @@ private fun computeBrakingCurve(
         -1.0,
         brakingType
     )
-    val brakingCurve = partBuilder.build()
+    var brakingCurve = partBuilder.build()
+
+    if (brakingType == BrakingType.ETCS_EBD && targetSpeed != 0.0) {
+        // TODO: by doing this, there is an approximation on the gradient used. TBD at a later date.
```
bougue-pe, line 257: As discussed, we can probably begin steps from `(targetPos, targetSpeed+dvEbi)` both backward (like usual) and forward until `targetSpeed` is intersected.
This would use the correct gradient, just applying these part backward and part forward, which seems OK to me.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -216,35 +372,102 @@ private fun keepBrakingCurveUnderOverlay(
...
+    // The remaining time during which the traction effort is not present. See Subset:
+    // §3.13.9.3.2.6.
+    val tBerem = max(rollingStock.rjsEtcsBrakeParams.tBe - tTraction, 0.0)
+    // Speed correction due to the braking system not being active yet. See Subset: §3.13.9.3.2.10.
+    val vDelta2 = aEst2 * tBerem
+
+    // Compute dBec and vBec. See Subset: §3.13.9.3.2.10.
+    val maxV = max(speed + vDelta0 + vDelta1, targetSpeed)
+    val dBec =
+        max(speed + vDelta0 + vDelta1 / 2, targetSpeed) * tTraction + (maxV + vDelta1 / 2) * tBerem
```
bougue-pe, line 437: From the formula, it should be:
max(speed + vDelta0 + vDelta1 / 2, targetSpeed) * tTraction + (maxV + vDelta`2` / 2) * tBerem

bougue-pe (commented):
Nice, and the simplifications you did here and there are very welcome 🙏 
We may dig a bit on mutualizing the code... Let's see how it goes with SvL!

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: The reason this assert was deleted was to remove the `beginPos` from the method, because this was its only use (it was used before for a position constraint, but we should always go until 0.0). I can put it somewhere else though if you want.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Oh wait no i got bamboozled, this just means velocity at target position. But well, this also means this isn't clear enough.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: I summarised this issue here: https://github.com/OpenRailAssociation/osrd/issues/10430

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Good catch.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Going to group them by type at least yeah? Otherwise there's gonna be so much.
All the fixes were fixed in this pr: https://github.com/OpenRailAssociation/osrd/pull/10440

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Technically, `maxSpeedEnvelope` could diminish during the loop, so you'd gain some computation time, but that's in best case scenarios and very specific case, so let's do that.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: I think this case is already handled by the ConstrainedEnvelopeBuilder, since we don't have that pb for const gamma decelerations.

Erashin: Ok I see the pb. In this case, we actually need to add the maintain curve before hand to the indication curve, THEN call `keepBrakingCurveOverlay`. This will work all the time. Added in issue: https://github.com/OpenRailAssociation/osrd/issues/10443
image

PR #10317, opened 2025-01-10
front: scroll into warning box

### LINKED ISSUE #10311: front: stdcm - warning for "no path found" is not visible on a 1920*1080 screen
### What happened?

When  I create a new simulation on STDCM with "a no path can be found" error, nothing happened on my screen. 
If I scroll down, yes  I see the warning message

Image

The red line is the bottom of my screen

As discussed with @thibautsailly the solution is to make scroll down to let appears the warning.

Data used for the simulation : 

* rolling stock : 6500G - 6500
* origin: Nantes (NS)
* destination: Angers (ASL)

Dev (SNCF)

Firefox

62932d4d465181056752d104b8075e6c3cc07fc8

kmer2016:
closes #10311

### DIFF
--- a/front/src/applications/stdcm/components/StdcmWarningBox.tsx
+++ b/front/src/applications/stdcm/components/StdcmWarningBox.tsx
@@ -1,3 +1,5 @@
+import { useEffect, useRef } from 'react';
+
 import { Button } from '@osrd-project/ui-core';
 import { Alert } from '@osrd-project/ui-icons';
 import cx from 'classnames';
@@ -25,8 +27,16 @@ const StdcmWarningBox = ({
   removeDestinationArrivalTime,
 }: StdcmWarningBoxProps) => {
   const { t } = useTranslation('stdcm');
+  const ref = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (errorType && errorType !== StdcmConfigErrorTypes.MISSING_LOCATION && ref.current) {
+      ref.current.scrollIntoView({ behavior: 'smooth' });
+    }
+  }, [errorType]);
+
   return (
-    <div data-testid="warning-box" className="warning-box">
+    <div ref={ref} data-testid="warning-box" className="warning-box">
       <span>
         <Alert variant="fill" size="lg" />
       </span>


front/src/applications/stdcm/components/StdcmWarningBox.tsx
```diff
@@ -25,8 +27,16 @@ const StdcmWarningBox = ({
   removeDestinationArrivalTime,
 }: StdcmWarningBoxProps) => {
   const { t } = useTranslation('stdcm');
+  const ref = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (ref.current) {
+      ref.current.scrollIntoView({ behavior: 'smooth' });
+    }
+  });
```
SharglutDev: Shouldn't we add this ?

```suggestion
  useEffect(() => {
    if (ref.current) {
      ref.current.scrollIntoView({ behavior: 'smooth' });
    }
  }, []);
```

kmer2016: Not really but we need to add `errorType` and `errorDetails` in dependencies

front/src/applications/stdcm/components/StdcmWarningBox.tsx
```diff
@@ -25,8 +27,16 @@ const StdcmWarningBox = ({
   removeDestinationArrivalTime,
 }: StdcmWarningBoxProps) => {
   const { t } = useTranslation('stdcm');
+  const ref = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (ref.current) {
+      ref.current.scrollIntoView({ behavior: 'smooth' });
+    }
+  }, [errorType, errorDetails]);
```
SharglutDev: With these dependencies, I also have a scroll when to warning box disappears. Maybe we should add a condition.

front/src/applications/stdcm/components/StdcmWarningBox.tsx
SharglutDev, line 1: One side effect of this scroll is when I want to change an op, when I clear the input to type a new one, the scroll fires and I think it's kind of annoying to lose the eye focus of the input. Maybe we should disable the pop of the warning box when we clear an input ?


https://github.com/user-attachments/assets/6d702d5d-bffe-4e27-8898-6968a08f30c5

front/src/applications/stdcm/components/StdcmWarningBox.tsx
```diff
@@ -25,8 +27,16 @@ const StdcmWarningBox = ({
   removeDestinationArrivalTime,
 }: StdcmWarningBoxProps) => {
   const { t } = useTranslation('stdcm');
+  const ref = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (errorType && ref.current) {
+      ref.current.scrollIntoView({ behavior: 'smooth' });
+    }
+  }, [errorType, errorDetails]);
```
SharglutDev: I think `errorType` should be enough. `errorDetails` is only defined when `errorType` goes to `bothPointAreScheduled` so the scroll will happen anyway.
```suggestion
  }, [errorType]);
```

front/src/applications/stdcm/components/StdcmWarningBox.tsx
kmer2016: Your suggestion makes sense, but I'd like to get @thibautsailly 's feedback on this.

front/src/applications/stdcm/components/StdcmWarningBox.tsx
kmer2016: I wasn’t fully sure how `errorType` and `errorDetails` interact. I thought there might be cases where the same `errorType` happens with different `errorDetails` between renders.

Right now, based on how the form works, it’s true that `errorDetails` cannot change without `errorType` going to a different state first. However, I included it to handle any future changes or edge cases where this might happen.

If we’re confident this won’t be needed, I can simplify the dependency array and rely only on `errorType`. Let me know what you think.

SharglutDev: I find it weird to add a dependency in the deps array that isn't actually a dependency (not use in the useEffect).
It doesn't seem to be a use case to have the same `errorType` with different `errorDetails`. Even if others errors come in the future with an `errorDetails`, this shouldn't affect the `useEffect` since the `errorType` will always change.
Is it ok for you ?

front/src/applications/stdcm/components/StdcmWarningBox.tsx
thibautsailly: Yes, disabling the scroll behaviour is necessary in this case.

SharglutDev (approved):
Lgtm and tesed, good job !

PR #10320, opened 2025-01-10
Fix git describe version injected to the frontend

flomonster:
- Stop using meta-env-import and use the builtin vite feature instead.
- Set the value to `development` when running `docker compose ...`

Example when pulling the image of this PR.

> [!WARNING]
> I did not test with the front end embedded with the gateway.

image


emersion (approved):
Nice, LGTM!

front/docker/Dockerfile.nginx
```diff
@@ -23,18 +28,10 @@ COPY --from=test_data . /tests/data
 
 FROM nginx:alpine
 
-RUN apk add npm && \
-    npm install -g @import-meta-env/cli && \
-    rm /etc/nginx/conf.d/default.conf
```
emersion, line 28: Note: previously we were deleting this file, and now we're keeping it. If it doesn't break anything, this is completely fine by me - just wondering if it matters at all. (Feel free to resolve.)

flomonster: Fixed

ghost (approved):
Thanks ! LGTM

PR #10322, opened 2025-01-10
front: group IGN layers

clarani:


clarani:
Hm not relevant anymore, it seems that the IGN layers are already grouped on dev :thinking:

PR #10323, opened 2025-01-10
front: fix ch suggestions update

### LINKED ISSUE #10033: stdcm: selecting a linked train doesn't set the ch code

SarahBellaha:
Closes #10033

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
@@ -145,6 +145,10 @@ const StdcmOperationalPoint = ({ location, pathStepId, disabled }: StdcmOperatio
   useEffect(() => {
     if (location) {
       setSearchTerm(location.name);
+      // Clear the list of CH suggestions if the location has changed to avoid showing outated suggestions
+      if (!chSuggestions.some((suggestion) => suggestion.label === location.secondary_code)) {
+        setChSuggestions([]);
+      }
     } else {
       setSearchTerm('');
       setChSuggestions([]);
--- a/front/src/reducers/osrdconf/stdcmConf/index.ts
+++ b/front/src/reducers/osrdconf/stdcmConf/index.ts
@@ -229,10 +229,10 @@ export const stdcmConfSlice = createSlice({
       }>
     ) {
       const { linkedTrainExtremity, trainName, pathStep, pathStepId } = action.payload;
-      const { name, ch, uic, geographic, arrivalDate, date, time } = pathStep;
+      const { name, ch, uic, geographic, arrivalDate, date, time, trigram } = pathStep;
 
       const newPathStep = {
-        location: { name, coordinates: geographic.coordinates, uic, secondary_code: ch },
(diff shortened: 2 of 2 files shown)


RomainValls (approved):
LGTM, tested 👍

front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
```diff
@@ -132,12 +136,22 @@ const StdcmOperationalPoint = ({ location, pathStepId, disabled }: StdcmOperatio
...
       setSearchTerm('');
       setChSuggestions([]);
     }
   }, [location]);
 
+  useEffect(() => {
+    if (location && searchResults.length > 0) {
+      updateChSuggestions(location.name);
+    }
+  }, [searchResults]);
```
clarani: Actually this useEffect introduces a bug...
[Capture vidéo du 2025-01-15 15-47-57.webm](https://github.com/user-attachments/assets/64c5e194-da3b-459e-b94c-7227fa3bc602)

Hum when we update the location from the linkedTrains, we should fetch the ch of the selected ci and update the suggestions, but only at this moment and not each time the searchResults change.

What you can do: 
- in StdcmPathStep location, add a new field `availableSecondaryCodes`

- in StdcmOperationalPoint, when we change the ci, then we update availableSecondaryCodes
- chSuggestions is built from availableSecondaryCodes
- when we select an op in linkedTrain, we also update the available secondaryCodes (either we already have the secondaryCodes available, or we fetch them with `searchOperationalPoints`, see below)
- [optional] in the hook useSearchOperationalPoint, extract from the function searchOperationalPoints the set state and change it so it returns its results and takes in argument the query

If you have another solution, we can discuss it ! I'm not sure mine is the simplest...

clarani (changes requested):

front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
SarahBellaha: I've changed the condition in this `useEffect` a bit so that we don't have the bug anymore.
I like your idea, i've created an [issue](https://github.com/OpenRailAssociation/osrd/issues/10401) to do this in another PR :)

clarani (approved):
:white_check_mark: LGTM because it fixes a major bug: if a user now selects a linked train, then the ci and ch are correctly set.

However, the management of the ch suggestions is still broken. It will be solved by this [issue](https://github.com/OpenRailAssociation/osrd/issues/10401)

PR #10324, opened 2025-01-10
front: fix stdcm add via scroll

### LINKED ISSUE #9713: lmr: front: rework the addition of intermediate OPs
# Description

Adding an intermediaite OP isn't clear enough for the users, we'd like to make this interaction clearer.

# AC

- [ ] There should be an automatic scroll if the card exits the view.
- [ ] The CI field must be focus after clicking "Add an intermediaite OP"
- [ ] The cards appear with an animation (mockup un comment below)

SharglutDev:
- Focus the CI input of the new added via
- Reset the state giving the via index to scroll to after the action to enable the animation even when the user adds 2 vias in a row at the same index

close #9713 (see last comment in the issue)

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmVias.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmVias.tsx
@@ -75,12 +75,19 @@ const StdcmVias = ({ disabled = false }: StdcmConfigCardProps) => {
 
     requestId = requestAnimationFrame(scrollWithAnimation);
 
-    const cancelListener = () => cancelAnimationFrame(requestId);
+    const newElementCIInput: HTMLInputElement | null = newElement.querySelector('.ci-input input');
 
-    newElement.parentElement!.addEventListener('animationend', cancelListener);
+    if (newElementCIInput) newElementCIInput.focus({ preventScroll: true });
+
+    const handleAnimationEnd = () => {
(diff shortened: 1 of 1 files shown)


front/src/applications/stdcm/components/StdcmForm/StdcmVias.tsx
```diff
@@ -75,12 +75,19 @@ const StdcmVias = ({ disabled = false }: StdcmConfigCardProps) => {
 
     requestId = requestAnimationFrame(scrollWithAnimation);
 
-    const cancelListener = () => cancelAnimationFrame(requestId);
+    const newElementCIInput: HTMLInputElement | null = newElement.querySelector('.ci-input input');
+
+    if (newElementCIInput) newElementCIInput.focus({ preventScroll: true });
+
+    const cancelListener = () => {
```
emersion: Nit: since this function now does more than just cancelling the listener, can we rename it to e.g. `handleAnimationEnd`?

emersion (approved):
LGTM regardless, feel free to resolve!

front/src/applications/stdcm/components/StdcmForm/StdcmVias.tsx
SharglutDev: Done

theocrsb (approved):
LGTM and tested

Issue #6146, opened 2023-12-18
Standardize OP code (CI,CH,trigram)


flomonster:
# Description

Currently operational points contains two extension:
- **identifier**: Contains the uic code (international)
- **sncf**: Contains ci, ch field and trigram

There should be a universal way to specify CI/CH. We could make this field part of the core railjson.

celine-sncf:
Il y a un lien avec Selene & le format international PLC/SLC

celine-sncf:
@nicolaswurtz est intéressé

celine-sncf:
Le service d'exposition du lien PLC <> PR existe déjà et peut être consommé par OSRD (et sera bientôt enrichi sur les PR étrangers). 
Les PLC ont tous des coordonnées x,y (pour la France cette coordonnée est calculée par PM-SIG donc pas très utile), ce qui permettrait d'importer les PR étrangers et de les positionner correctement
Format des identifiants PLC : code pays + code lieu 
Pour plus d'infos contacter Marion Postec

celine-sncf:
Catégories de PR : PR horairisés, PR travaux.. filtrer selon l'usage qui en est fait (c'est dans Gaïa)

loic-hamelin:
PLC = CI
SLC = CH

J'ai le fichier des PLC français. Voir avec Sélène également (référentiel des lieux)
PLC sur le RINF ?

louisgreiner:
For Germany, in OSM we could find something similar to trigrams: the `[railway:ref](https://wiki.openstreetmap.org/wiki/Key:railway:ref?uselang=fr)` key
https://www.openstreetmap.org/node/205364328

louisgreiner:
There should be enough data to cover Germany, but conflicts can happen on big infrastructures (such as France+Germany), i.e., `FF` is Frankfurt in Germany and Frauenfeld in Swiss

https://overpass-turbo.eu/s/1L4u
![image](https://github.com/OpenRailAssociation/osrd/assets/57864277/577ce74b-e875-4810-9074-7e596f267663)

Thanks @Tristramg for the work

emersion:
I wrote a small proposal for a RailJSON update below.

[...]

PR #10325, opened 2025-01-10
railjson: add is_station field for operational points

emersion:
A generic flag indicating whether an OP is a station will allow us to drop SNCF-specific logic.

This is part of #6146

Follow-up work:

- osrd-data
- Wire up search form
- Grep for "BV"/"00", switch to `is_station`


woshilapin (approved):
Nice, amazing how many pieces of code are touched.

core/src/test/kotlin/fr/sncf/osrd/sim_infra_adapter/PathPropertiesTests.kt
```diff
@@ -134,7 +134,8 @@ class PathPropertiesTests {
                         RJSOperationalPointSncfExtension(0, "BV", "B", "0", "TRI"),
                         null,
                     ),
-                    null
+                    null,
+                    false
```
bougue-pe, line 138: As the CH is "BV" above:
```suggestion
                    true
```

core/osrd-railjson/src/main/java/fr/sncf/osrd/railjson/schema/infra/RJSOperationalPoint.java
```diff
@@ -17,15 +18,21 @@ public class RJSOperationalPoint implements Identified {
     @Nullable
     public String weight;
 
+    @Nullable
```
bougue-pe, line 21: Do you have to make it nullable? (for backward compatibility, or missing default?)
As all the later use converts null to false, it seems better to disambiguate as soon as possible and make it non-nullable here.

emersion: Yeah, the intent is to continue supporting JSON files without the field. I haven't found a simple way to default this to false with moshi.

Khoyo (commented):
"stations" include freight, whereas "BV" is passenger specific. Was that considered?

emersion:
> "stations" include freight, whereas "BV" is passenger specific. Was that considered?

Does "00" include freight as well? Is there a different secondary code for freight?

From a UX PoV, do we want to include or exclude freight stations?

(My understanding is: yes, no, yes, respectively.)

core/osrd-railjson/src/main/java/fr/sncf/osrd/railjson/schema/infra/RJSOperationalPoint.java
bougue-pe: Not sure what's best, but after testing (remove `is_station` value from West_station in small_infra and see how core `PathPropEndpointTest` behave), it's possible to:
* make the value `private`
* define 
  ```java
    public Boolean getIsStation() {
        if (isStation == null) return false;
        return isStation;
    }
  ```
* remove the `@Nullable` here and in constructor, which is just an annotation for linters (but I probably would keep the annotation here and maybe replace constructor by a factory method with non-null `isStation`)
* get rid of the fallbacks added when using the value

Synar (approved):
Lgtm for the e2e tests.

Was this pr abandoned? It seemed like a nice generalization. (No pressure if you just didn't have time of course haha.)

emersion:
@Khoyo hasn't been addressed yet.

- If we go for a "station" field, I think we should include freight stations too. We don't have a good way to recognize these at SNCF, but it's fine to keep this best-effort and only populate the "station" field for a subset of stations (BV/00).
- If we want to specifically exclude freight stations from the special treatments (prioritization in UIs, default choice when user hasn't specified anything more specific), then we should rename the field to "passenger station".

Data question: do freight stations share the same trigram as passenger stations at SNCF?

shenriotpro (approved):
LGTM for tests..

flomonster:
@emersion Do we want to keep this PR open or with the new defined format we want to drop it?

emersion:
Seeing how many conflicts this contains, we'd probably better start from scratch again, and perhaps perform some other planned OP changes at the same time to reduce pain. This PR can still be used as a reference. Things missing:

- Name of the new field should be `is_passenger_station`.
- The migration SQL script should check for `BV`/`00` and set the new field accordingly.

PR #10331, opened 2025-01-13
front, editoast: add set -x option to the bash scripts

Sh099078:
It is useful when not knowing well the commands executed inside a script and their expected output to be able to see immediately which command failed and with which arguments it was called.

### DIFF
--- a/docker/gateway-entrypoint.sh
+++ b/docker/gateway-entrypoint.sh
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+set -x
+
 >/tmp/front_config jq -rn '{
     OSRD_GIT_DESCRIBE: env.OSRD_GIT_DESCRIBE,
 } | @json | "      globalThis.import_meta_env = \(.);"'
--- a/editoast/assets/signal_sprites/generate-atlas.sh
+++ b/editoast/assets/signal_sprites/generate-atlas.sh
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+set -x
+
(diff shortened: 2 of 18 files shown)


leovalais:
Personally, I consider -x more as a debugging tool rather than a way to help people understand a script. I think it's often too verbose and not readable enough for a non-debug usage, and kind of obfuscates the output of the script. I prefer plain `echo "I'm doing this task"` at every step.

I don't mind it at all though, feel free to disregard if there's consensus :)$

(Also you can write `set -ex`, not sure how standard this is though :eyes:)

woshilapin (commented):
I'm kind of thinking a bit like @leovalais, activating debug mode by default feels weird. That said, I also understand it might kind of a pain to add them manually each time you want to debug... We could depend on a variable environment to activate it, so by default it's off?

```
[ "$DEBUG" == 'true' ] && set -x
```

But like @leovalais, I won't block either way.

emersion:
> We could depend on a variable environment to activate it, so by default it's off?

Users can already easily enable it on a per-run basis, with `sh -x script/foo.sh`.

Sh099078:
> > We could depend on a variable environment to activate it, so by default it's off?
> 
> Users can already easily enable it on a per-run basis, with `sh -x script/foo.sh`.

Skill issue on my end, I didn't think about that. It is clearly better than imposing it to everyone or making it depend on an environment variable. I am closing the issue if that's fine with you @leovalais @woshilapin.

PR #10332, opened 2025-01-13
editoast: add consist params validation

### LINKED ISSUE #9543: editoast: towed rolling stocks should really have a minimum value
### What happened?

On the STDCM page, I tried a simulation with a train weighting 1t. That input is absurd, but the output was even more absurd: simulation failed, not enough traction.

Turns out, the RK4 integration doesn't really work out properly when the traction force is absurdly high compared to the weight. The first (2s) step goes up to 1500km/h. The second iteration starts at that speed and decelerates even faster. The average acceleration over 4 iterations end up negative => "not enough traction" to even go above 0km/h at the first step. 

Anything from these options:

1. A valid simulation where the train can start
2. Input is considered invalid in the front-end
3. A clear crash with a valid error message (even that would be better than "not enough traction")

Send an STDCM request with a 1t train. 

Local, Dev (SNCF)

Firefox

a9261bc7d6a18fb28a9a73a214f7c901ca7e3988

Wadjetz:
fix #9543

Adding backend validation for the consists values to prevent calculation errors.

### DIFF
--- a/editoast/src/views/timetable/stdcm.rs
+++ b/editoast/src/views/timetable/stdcm.rs
@@ -31,7 +31,6 @@ use tracing::Instrument;
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 use utoipa::IntoParams;
 use utoipa::ToSchema;
-use validator::Validate;
 
 use crate::core::conflict_detection::Conflict;
 use crate::core::conflict_detection::TrainRequirements;
@@ -100,6 +99,10 @@ enum StdcmError {
     TrainSimulationFail,
     #[error("Path items are invalid")]
     InvalidPathItems { items: Vec<InvalidPathItem> },
+    #[error("Invalid consist mass: {message}")]
+    InvalidConsistMass { message: String },
+    #[error("Invalid consist length: {message}")]
+    InvalidConsistLength { message: String },
 }
 
 #[derive(Debug, Default, Clone, Serialize, Deserialize, IntoParams, ToSchema)]
@@ -156,7 +159,6 @@ async fn stdcm(
 
     let trace_id = Some(trace_id).filter(|trace_id| *trace_id != TraceId::INVALID);
 
-    stdcm_request.validate()?;
     let mut conn = db_pool.get().await?;
 
     let timetable_id = id;
@@ -178,14 +180,18 @@ async fn stdcm(
         .await?
         .into();
 
+    let towed_rolling_stock = stdcm_request
+        .get_towed_rolling_stock(&mut conn)
+        .await?
+        .map(From::from);
+
+    stdcm_request.validate_consist(&rolling_stock, &towed_rolling_stock)?;
+
     let physics_consist_parameters = PhysicsConsistParameters {
         max_speed: stdcm_request.max_speed,
         total_length: stdcm_request.total_length,
         total_mass: stdcm_request.total_mass,
-        towed_rolling_stock: stdcm_request
-            .get_towed_rolling_stock(&mut conn)
-            .await?
-            .map(From::from),
+        towed_rolling_stock,
         traction_engine: rolling_stock,
     };
 
@@ -524,6 +530,10 @@ mod tests {
     use rstest::rstest;
     use serde_json::json;
     use std::str::FromStr;
+    use uom::si::length::meter;
+    use uom::si::length::Length;
+    use uom::si::mass::kilogram;
+    use uom::si::quantities::Mass;
     use uuid::Uuid;
 
     use crate::core::conflict_detection::Conflict;
@@ -535,6 +545,7 @@ mod tests {
     use crate::core::simulation::PhysicsConsist;
     use crate::core::simulation::ReportTrain;
     use crate::core::simulation::SpeedLimitProperties;
+    use crate::error::InternalError;
     use crate::models::fixtures::create_fast_rolling_stock;
     use crate::models::fixtures::create_simple_rolling_stock;
     use crate::models::fixtures::create_small_infra;
@@ -551,7 +562,12 @@ mod tests {
 
     use super::*;
 
-    fn get_stdcm_payload(rolling_stock_id: i64, work_schedule_group_id: Option<i64>) -> Request {
+    fn get_stdcm_payload(
+        rolling_stock_id: i64,
+        work_schedule_group_id: Option<i64>,
+        total_mass: Option<f64>,
+        total_length: Option<f64>,
+    ) -> Request {
         Request {
             start_time: Some(
                 DateTime::from_str("2024-01-01T10:00:00Z").expect("Failed to parse datetime"),
@@ -601,8 +617,8 @@ mod tests {
             time_gap_before: 35000,
             time_gap_after: 35000,
             margin: Some(MarginValue::MinPer100Km(4.5)),
-            total_mass: None,
-            total_length: None,
+            total_mass: total_mass.map(Mass::new::<kilogram>),
+            total_length: total_length.map(Length::new::<meter>),
             max_speed: None,
             loading_gauge_type: None,
         }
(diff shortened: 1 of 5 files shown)


editoast/src/views/timetable/stdcm/request.rs
```diff
@@ -103,17 +104,62 @@ pub(super) struct Request {
...
+    range_parse(5.0, 750.0, de)
+}
+
+pub fn max_speed_parse<'de, D>(de: D) -> Result<Option<f64>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let value: Option<f64> = Deserialize::deserialize(de)?;
+    if let Some(max_speed) = value {
+        if max_speed < 8.33 {
```
bougue-pe: As explained in the main comment, let's just keep it to a strictly positive value like before (no crash expected from a too low value).

editoast/src/views/timetable/stdcm/request.rs
```diff
@@ -103,17 +104,62 @@ pub(super) struct Request {
...
+    D: Deserializer<'de>,
+{
+    range_parse(20000.0, 10000000.0, de)
+}
+
+pub fn total_length_parse<'de, D>(de: D) -> Result<Option<f64>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    range_parse(5.0, 750.0, de)
```
bougue-pe: AFAIK, there is no technical issue with the length of the consist.
So for now I think it's fine to check that it's positive, and (if easy) that the consist length is >= length of the train engine.

editoast/src/views/timetable/stdcm/request.rs
```diff
@@ -103,17 +104,62 @@ pub(super) struct Request {
...
+        }
+    }
+    Ok(value)
+}
+
+pub fn total_mass_parse<'de, D>(de: D) -> Result<Option<f64>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    range_parse(20000.0, 10000000.0, de)
```
bougue-pe: This is the most important one, and I see 4 options:
1. Just staying strictly positive won't protect from weird cases described in the issue.
2. Find a better min value (I don't know if 20t is a worldwide realistic value)
3. Probably the best option now: At least enforce >= the mass of the train engine.
4. Probably the idealistic option: Find a coefficient to apply to `max (max braking force, max tractive force)` and ensure that the mass is > to the resulting value. Doing the same for the  `min (max braking force, max tractive force)` and ensure that the mass is < to the resulting value would be nice to ensure that the train is able to brake/accelerate. Maybe try and fail could help find a value, but it's probably too much work for the benefit.

I let anyone react on it, but going with the 3rd option is good enough IMO, if no one else reacts.

bougue-pe (commented):
After discussing in the office, the main goal of the issue is to protect from unclear crash.
The front part already does the main job, so we should just protect the backend from unclear crashes, without narrowing use when it's technically OK (at a reasonable cost).

👍 Thanks for getting rid of a new part of the 'validator' use 🎉

editoast/src/views/timetable/stdcm/request.rs
Wadjetz: The 20t comes from a value close to the lowest rolling stock we have.
We could also apply the same logic that we have on the frontend (point 2), but I’m not sure it’s a good idea to make a database call within the serde parser.

leovalais: I agree about option 3 being the lowest hanging fruit. But I'm a bit worried about separation of concerns here. If you ask me it's not up to editoast to perform such validation. Unless I misunderstood the issue here is Core algorithms failing unhelpfully for values outside of a certain domain. Relying on editoast to check for bounds Core provides (and thus has to maintain) seems very brittle to me. I can't think of an example where externalizing validation is a good design.

Imo the best way would be (maybe not the easiest if I understand what @bougue-pe said yesterday):

1. Core validates its inputs and returns an error for out of bounds values
2. Editoast just forwards the error without interpreting it. It's probably fine to do so as this validation only concerns editoast API users (the frontend doesn't send the request for oob values).

Wdyt?

woshilapin: I agree, `editoast` may protect itself from crashing, `core` should do the same for itself, and `front` may provide more user-friendly guidance about "reasonable values" (and if the user wants to send weird values: garbage in, garbage out, it's fine).

The only place where I could see `editoast` doing something is about the positive floating numbers. Sadly, Rust doesn't have types for non-negative floating numbers, so `Option<f64>` is not enough coercive. So we could keep a validation step of that (with a `deserialize_with`), or we could do what we've already done with `PositiveDuration`: create another type that carries this semantic.

bougue-pe: To help prioritize: Does the front already validate that consist mass/length >= rolling stock?

Agreed on roles, but it may not be as easy to do in core:
* Checking length/mass >= rolling stock in core means... sending those values just for check 😞
* The real "safety check" on values is a power-to-weight ratio (traction and brake), and it means finding this max ratio (maybe min also). Note that length is actually not an issue AFAIK.
* Introducing checks in core may be a bit of work as it's unusual (which is bad).

bougue-pe: After discussion:
* at parsing: just check that it's positive (like before)
* as soon as the data are retrieved (before sending core request): check the consistency of mass & length: consist >= (towed-RS + RS)
* side-note: not worth the cost to add some dedicated check in core for now :-/

editoast/src/views/timetable/stdcm.rs
```diff
@@ -100,6 +99,8 @@ enum StdcmError {
     TrainSimulationFail,
     #[error("Path items are invalid")]
     InvalidPathItems { items: Vec<InvalidPathItem> },
+    #[error("Request validation failed: {message}")]
+    RequestValidationFail { message: String },
```
bougue-pe: How about a more direct (if you accept, please report everywhere: translation, etc.):
```suggestion
    #[error("Invalid request: {message}")]
    InvalidRequest { message: String },
```

editoast/src/views/timetable/stdcm/request.rs
```diff
@@ -297,6 +303,72 @@ impl Request {
...
+        self.validate_consist_length(traction_engine, towed_rolling_stock)?;
+        Ok(())
+    }
+
+    fn validate_consist_mass(
+        &self,
+        traction_engine: &RollingStock,
+        towed_rolling_stock: &Option<TowedRollingStock>,
+    ) -> Result<()> {
+        let max = Mass::new::<kilogram>(10000000.0);
```
bougue-pe: AFAIU, we won't add a hard max limit in the backend, so this should disappear, no?

editoast/src/views/timetable/stdcm/request.rs
```diff
@@ -297,6 +303,72 @@ impl Request {
...
+
+        Ok(())
+    }
+
+    fn validate_consist_length(
+        &self,
+        traction_engine: &RollingStock,
+        towed_rolling_stock: &Option<TowedRollingStock>,
+    ) -> Result<()> {
+        let max = Length::new::<meter>(750.0);
```
bougue-pe: ditto

editoast/src/views/timetable/stdcm/request.rs
```diff
@@ -297,6 +303,72 @@ impl Request {
...
+                .as_ref()
+                .map(|t| t.mass)
+                .unwrap_or_default();
+        if let Some(request_total_mass) = self.total_mass {
+            if request_total_mass < consist_mass || request_total_mass > max {
+                return Err(StdcmError::RequestValidationFail {
+                    message: format!(
+                        "The total weight must be between {} and {}",
+                        &consist_mass.into_format_args(kilogram, DisplayStyle::Description),
+                        &max.into_format_args(kilogram, DisplayStyle::Description),
```
bougue-pe: To be explicit about the check and where the value comes from:
```suggestion
                        "The total mass must be greater than the sum of the rolling stock masses ({})",
                        &consist_mass.into_format_args(kilogram, DisplayStyle::Description),
```

editoast/src/views/timetable/stdcm/request.rs
```diff
@@ -297,6 +303,72 @@ impl Request {
...
+                .as_ref()
+                .map(|t| t.length)
+                .unwrap_or_default();
+        let consist_length = consist_length.floor::<meter>();
+
+        if let Some(request_total_length) = self.total_length {
+            if request_total_length < consist_length || request_total_length > max {
+                return Err(StdcmError::RequestValidationFail {
+                    message: format!(
+                        "The total length must be between {} and {}",
```
bougue-pe: ditto

editoast/src/views/timetable/stdcm.rs
```diff
@@ -958,6 +972,85 @@ mod tests {
...
+            .core_client(core.into())
+            .build();
+        let small_infra = create_small_infra(&mut db_pool.get_ok()).await;
+        let timetable = create_timetable(&mut db_pool.get_ok()).await;
+        let rolling_stock =
+            create_fast_rolling_stock(&mut db_pool.get_ok(), &Uuid::new_v4().to_string()).await;
+
+        let request = app
+            .post(format!("/timetable/{}/stdcm?infra={}", timetable.id, small_infra.id).as_str())
+            .json(&get_stdcm_payload(rolling_stock.id, None, Some(1.0), None));
```
bougue-pe: nit, for explicitness and realism of the test:
```suggestion
        let total_mass = Some(80_000.0)
        let request = app
            .post(format!("/timetable/{}/stdcm?infra={}", timetable.id, small_infra.id).as_str())
            .json(&get_stdcm_payload(rolling_stock.id, None, total_mass, None));
```

editoast/src/views/timetable/stdcm.rs
```diff
@@ -958,6 +972,85 @@ mod tests {
...
+            .core_client(core.into())
+            .build();
+        let small_infra = create_small_infra(&mut db_pool.get_ok()).await;
+        let timetable = create_timetable(&mut db_pool.get_ok()).await;
+        let rolling_stock =
+            create_fast_rolling_stock(&mut db_pool.get_ok(), &Uuid::new_v4().to_string()).await;
+
+        let request = app
+            .post(format!("/timetable/{}/stdcm?infra={}", timetable.id, small_infra.id).as_str())
+            .json(&get_stdcm_payload(rolling_stock.id, None, None, Some(1.0)));
```
bougue-pe: ditto
```suggestion
            .json(&get_stdcm_payload(rolling_stock.id, None, None, total_length));
```

editoast/src/views/timetable/stdcm.rs
```diff
@@ -958,6 +972,85 @@ mod tests {
         }
     }
 
+    #[rstest]
```
bougue-pe, line 922: Can you please add tests on the happy path (length OK and mass OK)?

editoast/src/views/timetable/stdcm/request.rs
bougue-pe: Resolving this, as the leftover is tracked in other comments.

And we won't dig more on this now unless it triggers alarms in someone else's head, but I'm having second thought on the validation of total length.
I'm not sure it make more sense to have a total that's longer than shorter (compared to the sum of tractive-engine and towed rolling-stock)...

front/public/locales/en/errors.json
```diff
@@ -208,7 +208,8 @@
       "RollingStockNotFound": "Rolling stock '{{rolling_stock_id}}' does not exist",
       "TowedRollingStockNotFound": "Towed rolling stock {towed_rolling_stock_id} does not exist",
       "TrainSimulationFail": "Train simulation failed",
-      "TimetableNotFound": "Timetable '{{timetable_id}}' does not exist"
+      "TimetableNotFound": "Timetable '{{timetable_id}}' does not exist",
+      "InvalidRequest": "Request validation failed"
```
bougue-pe: I think we should be consistent and more direct here too:
```suggestion
      "InvalidRequest": "Invalid request"
```

front/public/locales/fr/errors.json
```diff
@@ -210,7 +210,8 @@
       "RollingStockNotFound": "Matériel roulant '{{rolling_stock_id}}' non trouvé",
       "TowedRollingStockNotFound": "Matériel roulant remorqué {towed_rolling_stock_id} non trouvé",
       "TrainSimulationFail": "Échec de la simulation du train",
-      "TimetableNotFound": "Grille horaire '{{timetable_id}}' non trouvée"
+      "TimetableNotFound": "Grille horaire '{{timetable_id}}' non trouvée",
+      "InvalidRequest": "Échec de la validation de la requête"
```
bougue-pe: ditto:
```suggestion
      "InvalidRequest": "Requête invalide"
```

bougue-pe (approved):
LGTM.
Feel free to resolve those 2 comments if applied.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -100,6 +99,8 @@ enum StdcmError {
     TrainSimulationFail,
     #[error("Path items are invalid")]
     InvalidPathItems { items: Vec<InvalidPathItem> },
+    #[error("Invalid request: {message}")]
+    InvalidRequest { message: String },
```
leovalais: This is a bit vague imo, maybe `InvalidConsistLength` and `InvalidConsistMass` with numeric fields would be more descriptive?

woshilapin (approved):

SharglutDev (approved):
Lgtm for front (not tested)

SharglutDev (approved):

bougue-pe (approved):

bougue-pe (commented):
LGTM, just please squash before merge 🙏

editoast/src/views/timetable/stdcm.rs
```diff
@@ -100,6 +99,10 @@ enum StdcmError {
     TrainSimulationFail,
     #[error("Path items are invalid")]
     InvalidPathItems { items: Vec<InvalidPathItem> },
+    #[error("Invalid consist mass: {message}")]
+    InvalidConsistMass { message: String },
+    #[error("Invalid consist length: {message}")]
+    InvalidConsistLength { message: String },
```
leovalais, line 105: It would have been great to have numerical values instead of a generic message string :/

Because of that, 1) the full error message is split in several places and 2) in tests you have to `assert_eq` on strings instead of numerical values which are more stable.

PR #10333, opened 2025-01-13
core: add cli command to re-run payloads

eckter:
It's just a debug tool, it's a little more practical than using curl/postman for every run with the "api" command


Erashin (approved):
Nice!!!

PR #10334, opened 2025-01-13
front: fix displayed op on manchette

Akctarus:
Set the weight of the main PRs (origin, arrival and vias) to 100. 
For vias added by hand from the map, they were previously set to “null”, now they're set to 100, so they're displayed on the manchette. 

## Tests : 
- add a waypoint using the functionality, see it displayed on the basic cuff and GEV
- add a waypoint by clicking on the map, and see it displayed on the headline and GEV.
- check PR weights console.log(operationalPoints.weight) and verify that passage PRs have a weight of 100

### DIFF
--- a/front/src/applications/operationalStudies/helpers/upsertMapWaypointsInOperationalPoints.ts
+++ b/front/src/applications/operationalStudies/helpers/upsertMapWaypointsInOperationalPoints.ts
@@ -5,6 +5,8 @@ import type { PathfindingResultSuccess, TrainScheduleResult } from 'common/api/o
 
 import type { OperationalPoint } from '../types';
 
+const HIGHEST_PRIORITY_WEIGHT = 100;
+
 /**
  * Check if the train path used waypoints added by map click and add them to the operational points
  */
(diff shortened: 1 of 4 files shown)


front/src/modules/simulationResult/components/SpeedSpaceChart/SpeedSpaceChartContainer.tsx
```diff
@@ -20,7 +20,7 @@ import { formatData } from './helpers';
 export type SpeedSpaceChartContainerProps = {
   trainSimulation: SimulationResponseSuccess;
   selectedTrainPowerRestrictions?: LayerData<PowerRestrictionValues>[];
-  pathProperties: PathPropertiesFormatted;
+  pathProperties?: PathPropertiesFormatted;
```
SharglutDev: Don't think it makes sense to have a speed space chart without path properties

Akctarus: You're right, I fixed it

front/src/applications/operationalStudies/helpers/upsertMapWaypointsInOperationalPoints.ts
```diff
@@ -35,7 +35,7 @@ export const upsertMapWaypointsInOperationalPoints = (
         },
         part: { track: step.track, position: step.offset },
         position: positionOnPath,
-        weight: null,
+        weight: 100,
```
Caracol3: To avoid magic numbers and because it's also used in useGetProjectedTrainOperationalsPoints maybe we should use a const like 'HIGHEST_PRIORITY_WEIGHT' here

Caracol3 (commented):
Works great ! Good job ! 
I just left one comment

front/src/applications/operationalStudies/helpers/upsertMapWaypointsInOperationalPoints.ts
Akctarus: Fixed !

front/src/applications/operationalStudies/helpers/upsertMapWaypointsInOperationalPoints.ts
```diff
@@ -15,38 +15,53 @@ export const upsertMapWaypointsInOperationalPoints = (
   t: TFunction
 ): OperationalPoint[] => {
   let waypointCounter = 1;
+  const HIGHEST_PRIORITY_WEIGHT = 100;
```
SharglutDev: Should be outside of the function

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
```diff
@@ -76,14 +75,8 @@ const useGetProjectedTrainOperationalPoints = (
...
           >;
-        } else {
-          // If the manchette hasn't been saved, we want to display by default only
-          // the waypoints with CH BV/00/'' and the ones added by map click
-          operationalPointsWithUniqueIds = operationalPointsWithUniqueIds.filter((op) =>
-            op.extensions?.sncf ? isStation(op.extensions.sncf.ch) : true
-          );
         }
-        setFilteredOperationalPoints(operationalPointsWithUniqueIds);
+        setFilteredOperationalPoints(operationalPointsWithAllWaypoints);
```
SharglutDev: I don't think we should change this. Otherwise, we won't have any waypoint checked in the waypoint panel.
We should ask @Tguisnet if on linear mode of the manchette, we now want to view all points or just some of them if a weight superior than a specific number etc

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
SharglutDev, line 1: Not sure where it comes from, but there is a white screen if we try to switch to linear mode in the manchette

front/src/applications/operationalStudies/helpers/upsertMapWaypointsInOperationalPoints.ts
Akctarus: fixed

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
Akctarus: Fixed ! It came from pushing the OP to the end of the list rather than adding it to the index when given the weight of 100

front/src/applications/operationalStudies/helpers/upsertMapWaypointsInOperationalPoints.ts
```diff
@@ -18,35 +20,56 @@ export const upsertMapWaypointsInOperationalPoints = (
...
-      } else {
-        operationalPointsWithAllWaypoints.splice(indexToInsert, 0, formattedStep);
+      if ('uic' in step) {
+        const matchedIndex = operationalPointsWithAllWaypoints.findIndex(
+          (op) =>
+            'uic' in step &&
+            'secondary_code' in step &&
+            step.uic === op.extensions?.identifier?.uic &&
+            step.secondary_code === op.extensions?.sncf?.ch
+        );
```
SharglutDev, line 30: > [!NOTE]
> @emersion we don't access to a `SuggestedOP` here so we can't use `matchPathStep...` but this spot will be another one to watch in your refacto

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
```diff
@@ -76,12 +75,6 @@ const useGetProjectedTrainOperationalPoints = (
           operationalPointsWithUniqueIds = JSON.parse(stringifiedSavedWaypoints) as NonNullable<
             PathProperties['operational_points']
           >;
-        } else {
-          // If the manchette hasn't been saved, we want to display by default only
-          // the waypoints with CH BV/00/'' and the ones added by map click
-          operationalPointsWithUniqueIds = operationalPointsWithUniqueIds.filter((op) =>
-            op.extensions?.sncf ? isStation(op.extensions.sncf.ch) : true
-          );
```
SharglutDev: Ok so sorry but after all, @Tguisnet would prefer (and I agree) not to change the behavior and keep this by default.
On the other hand, we should be vigilant if the path step which aren't BV/00 are still well displayed in the manchette.

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
SharglutDev: Ok so if you created an issue for that, you can close it with a comment :)

Caracol3 (approved):
LGTM

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
Akctarus: It is fixed, ty for your comment !

SharglutDev (approved):
Lgtm and tested, good job !

PR #10340, opened 2025-01-13
schemas: provide a default value for ETCS brake parameters

woshilapin:
Without this default value, it's a breaking change and consumers of this repository will have to update their use of `RollingStock` class.


leovalais (approved):

bougue-pe (approved):
🙏 Thanks!

PR #10346, opened 2025-01-13
front: fix update coordinates when change ch code

### LINKED ISSUE #10310: front: stdcm: origin and destination markers are both at BV and on the wrong track
### What happened?

Tryied to look for a STDCM between BGB BV and BGB P1 but the map is showing both marker at the BV:
Image

And the result was this:
Image

I'm expecting the markers to be on the right PR and on the right track once the result is found.

Reproduce description

Recette (SNCF)

Google Chrome

fdfd8ac

kmer2016:
closes #10310

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
@@ -16,7 +16,8 @@ type StdcmOperationalPointProps = {
   disabled?: boolean;
 };
 
-type Option = StdcmPathStep['location'] & { label: string };
+type CIOption = StdcmPathStep['location'] & { label: string };
+type CHOption = { label: string; id: string; coordinates: [number, number] };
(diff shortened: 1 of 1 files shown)


front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
```diff
@@ -16,7 +16,8 @@ type StdcmOperationalPointProps = {
   disabled?: boolean;
 };
 
-type Option = StdcmPathStep['location'] & { label: string };
+type CIOption = StdcmPathStep['location'] & { label: string };
+type CHOption = { label: string; id: string; coordinates: [number, number] };
```
SarahBellaha, line 20:
```suggestion
type CHOption = { label: string; id: string; coordinates: GeoJsonPointValue };
```
Can't we use this type ?

Math-R (approved):
lgtm =)

PR #10347, opened 2025-01-13
editoast: filter train schedules for LMR

younesschrifi:
Part of this [issue](https://github.com/osrd-project/osrd-confidential/issues/847)

### DIFF
--- a/editoast/src/models/timetable.rs
+++ b/editoast/src/models/timetable.rs
@@ -5,9 +5,12 @@ use diesel::sql_query;
 use diesel::sql_types::Array;
 use diesel::sql_types::BigInt;
 use diesel_async::RunQueryDsl;
+use futures_util::stream::TryStreamExt;
 use std::ops::DerefMut;
 
 use crate::error::Result;
+use crate::models::prelude::*;
+use crate::models::train_schedule::TrainSchedule;
 use crate::models::Identifiable;
 use crate::models::{DeleteStatic, Retrieve};
 use crate::Exists;
@@ -55,6 +58,29 @@ impl Timetable {
             .load(conn.write().await.deref_mut())
             .await
             .map_err(Into::into)
+    }
+
+    pub async fn schedules_before_date(
+        self,
+        conn: &mut DbConnection,
+        time: DateTime<Utc>,
+    ) -> Result<Vec<TrainSchedule>> {
+        use diesel::prelude::*;
+        use diesel_async::RunQueryDsl;
+        use editoast_models::tables::train_schedule::dsl;
+
+        let train_schedules = dsl::train_schedule
+            .filter(dsl::start_time.le(time))
+            .filter(dsl::timetable_id.eq(self.id))
+            .load_stream::<Row<TrainSchedule>>(conn.write().await.deref_mut())
+            .await?
+            .map_ok(|ts| ts.into())
+            .try_collect::<Vec<TrainSchedule>>()
+            .await;
+        match train_schedules {
+            Ok(train_schedules) => Ok(train_schedules),
+            Err(err) => Err(err.into()),
+        }
     }
 }
 
--- a/editoast/src/models/train_schedule.rs
+++ b/editoast/src/models/train_schedule.rs
@@ -10,7 +10,7 @@ use editoast_schemas::train_schedule::ScheduleItem;
 use editoast_schemas::train_schedule::TrainScheduleBase;
 use editoast_schemas::train_schedule::TrainScheduleOptions;
 
-use super::Model as _;
+use crate::models::prelude::*;
 
 #[derive(Debug, Default, Clone, Model)]
 #[model(table = editoast_models::tables::train_schedule)]
--- a/editoast/src/views/timetable/stdcm.rs
+++ b/editoast/src/views/timetable/stdcm.rs
@@ -40,7 +40,7 @@ use crate::core::CoreClient;
 use crate::error::Result;
 use crate::models::prelude::*;
 use crate::models::stdcm_log::StdcmLog;
-use crate::models::timetable::TimetableWithTrains;
+use crate::models::timetable::Timetable;
 use crate::models::train_schedule::TrainSchedule;
 use crate::models::Infra;
 use crate::models::RollingStockModel;
@@ -149,19 +149,12 @@ async fn stdcm(
     let timetable_id = id;
     let infra_id = query.infra;
 
-    // 1. Retrieve Timetable / Infra / Trains / Simulation / Rolling Stock
-    let timetable_trains = TimetableWithTrains::retrieve_or_fail(&mut conn, timetable_id, || {
-        StdcmError::TimetableNotFound { timetable_id }
-    })
-    .await?;
+    // 1.  Infra / Timetable / Trains / Simulation / Rolling Stock
 
     let infra = Infra::retrieve_or_fail(&mut conn, infra_id, || StdcmError::InfraNotFound {
         infra_id,
     })
     .await?;
-
-    let (train_schedules, _): (Vec<_>, _) =
-        TrainSchedule::retrieve_batch(&mut conn, timetable_trains.train_ids.clone()).await?;
 
     let rolling_stock =
         RollingStockModel::retrieve_or_fail(&mut conn, stdcm_request.rolling_stock_id, || {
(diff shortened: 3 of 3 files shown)


editoast/src/models/train_schedule.rs
```diff
@@ -9,8 +11,14 @@ use editoast_schemas::train_schedule::PowerRestrictionItem;
...
 use editoast_schemas::train_schedule::TrainScheduleBase;
 use editoast_schemas::train_schedule::TrainScheduleOptions;
+use diesel::prelude::*;
+use diesel_async::RunQueryDsl;
+use futures_util::stream::TryStreamExt;
+use std::ops::DerefMut;
 
 use super::Model as _;
+use crate::diesel::ExpressionMethods;
+use crate::Row;
```
leovalais:
```suggestion
use crate::models::prelude::*; // contains Row, Model, and all traits
```

editoast/src/models/train_schedule.rs
```diff
@@ -74,3 +82,24 @@ impl From<TrainScheduleBase> for TrainScheduleChangeset {
             .options(options)
     }
 }
+
+pub async fn filter_train_by_start_time_and_timetable(
+    conn: &mut DbConnection,
+    time: DateTime<Utc>,
+    timetable_id: i64,
+) -> Result<Vec<TrainSchedule>> {
+    use editoast_models::tables::train_schedule::dsl;
```
leovalais: Could `diesel::prelude::*`, `RunQueryDsl` etc. be scoped as well?

editoast/src/views/timetable/stdcm.rs
```diff
@@ -191,6 +186,29 @@ async fn stdcm(
     let earliest_departure_time = stdcm_request.get_earliest_departure_time(simulation_run_time);
     let latest_simulation_end = stdcm_request.get_latest_simulation_end(simulation_run_time);
 
+    let timetable = Timetable::retrieve_or_fail(&mut conn, timetable_id, || {
+        StdcmError::TimetableNotFound { timetable_id }
+    })
+    .await?;
+    // Filter train
+    let mut train_schedules = filter_train_by_start_time_and_timetable(&mut conn, latest_simulation_end, timetable.id).await?;
```
leovalais: Since you already retrieved the timetable, could this be an associated method?

```suggestion
    let mut train_schedules = timetable.schedules_in_time_window(&mut conn, earliest_departure_time, latest_simulation_end).await?;
```

editoast/src/views/timetable/stdcm.rs
```diff
@@ -191,6 +186,29 @@ async fn stdcm(
...
+
+    train_schedules.retain(|train_schedule| {
+        if train_schedule.start_time >= earliest_departure_time {
+            true
```
leovalais: Any reason why this isn't checked in the SQL request?

editoast/src/views/timetable/stdcm.rs
```diff
@@ -191,6 +185,31 @@ async fn stdcm(
...
+        } else {
+            if let Some(last_item) = train_schedule.schedule.last() {
+                if let Some(last_path_item) = train_schedule.path.last() {
+                    last_item.at == last_path_item.id
+                } else {
+                    false
+                }
+            } else {
+                false
+            }
```
leovalais: Not sure what this does, isn't `last_item.at == last_path_item.id` always true for valid schedules?

editoast/src/views/timetable/stdcm.rs
```diff
@@ -148,19 +149,12 @@ async fn stdcm(
     let infra_id = query.infra;
 
     // 1. Retrieve Timetable / Infra / Trains / Simulation / Rolling Stock
```
leovalais: The comments scoping each step should be updated

editoast/src/views/timetable/stdcm.rs
```diff
@@ -191,6 +185,31 @@ async fn stdcm(
     let earliest_departure_time = stdcm_request.get_earliest_departure_time(simulation_run_time);
     let latest_simulation_end = stdcm_request.get_latest_simulation_end(simulation_run_time);
 
+    let timetable = Timetable::retrieve_or_fail(&mut conn, timetable_id, || {
+        StdcmError::TimetableNotFound { timetable_id }
+    })
+    .await?;
+    // Filter train
+    let mut train_schedules =
```
leovalais: Since the timetable is already fetched, wdyt about an associated method?

```suggestion
    let mut train_schedules = timetable.schedules_in_time_window(earliest_departure_time, latest_simulation_end).await?;
```

leovalais (commented):
Thanks! A few questions:

editoast/src/views/timetable/stdcm.rs
younesschrifi: Good idea :)

editoast/src/views/timetable/stdcm.rs
younesschrifi: No, maybe `last_item` should be renamed  `last_scheduled_item`  for more clarity. 
This `last_path_item.id` necessarily corresponds to the train arrival, whereas the `last_item.at` only corresponds to the train arrival if the arrival is timed.

editoast/src/views/timetable/stdcm.rs
younesschrifi: We designed this way with @flomonster. But I think it's a good idea to include this filter in the SQL request !

editoast/src/views/timetable/stdcm.rs
leovalais: Okay so unless I'm mistaken this basically filters out trains where no arrival time is scheduled. But if they're valid, why exclude them?

editoast/src/models/timetable.rs
```diff
@@ -56,6 +59,29 @@ impl Timetable {
             .await
             .map_err(Into::into)
     }
+
+    pub async fn schedules_in_time_window(
```
leovalais:
```suggestion
    pub async fn schedules_before_date(
```

since we're going this way

editoast/src/models/timetable.rs
```diff
@@ -5,12 +5,15 @@ use diesel::sql_query;
...
 use diesel_async::RunQueryDsl;
+use futures_util::stream::TryStreamExt;
 use std::ops::DerefMut;
 
 use crate::error::Result;
+use crate::models::train_schedule::TrainSchedule;
 use crate::models::Identifiable;
 use crate::models::{DeleteStatic, Retrieve};
 use crate::Exists;
+use crate::Row;
```
leovalais: Likewise about the `prelude`, helps cleaning the imports :)

editoast/src/views/timetable/stdcm.rs
```diff
@@ -191,6 +184,43 @@ async fn stdcm(
...
+    .await?;
+
+    // Filtering trains is done in 2 stages:
+    // 1) We filter all trains whose departure dates are before the latest_simulation_end via sql
+
+    let mut train_schedules = timetable
+        .schedules_in_time_window(&mut conn, latest_simulation_end)
+        .await?;
+
+    // 2) We then filter in rust according to the lmr train departure date.  Since TrainSchedule input is light, this rust filtering saves time.
```
leovalais: I think it would be more helpful to explain why we're doing this manipulation with `schedule` and `path`, and why the departure date isn't part of the SQL query. It'd be great if you could reproduce the schema you shown me on call in ascii-art, but at least explaining it would help better understand this part. (If you ask me again in six months I'm not I'd be able to tell why again...)

I'd also be cautious about "Since TrainSchedule input is light, this rust filtering saves time". We don't know that, we just assume it's good enough. I'd argue it's actually the opposite: rewriting the following filtering in SQL could be an optimization to bench for large grids, saving us transport cost.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -191,6 +184,43 @@ async fn stdcm(
...
+                    } else {
+                        true
+                    }
+                } else {
+                    true
+                }
+            } else {
+                false
+            }
+        }
```
leovalais: Could this be rewritten in a more monadic style to avoid this much nesting? 😇 

```rust
 } else {
                            true
                        }
                    } else {
                        true
                    }
                } else {
                    true
                }
            } else {
                false
            }
        }
```

image

leovalais (commented):
Thanks for the changes and the explications on call!

flomonster (approved):
LGTM

leovalais (approved):
Thanks a lot for the explanation, it really helps understand what's going on :)

PR #10370, opened 2025-01-13
front: decrease the electrification layer order priority to avoid masking path in map

### LINKED ISSUE #6546: Electrification layer covers the path found
### What happened?

When the electrification layer is activated I can not see the path found on the map because the layer is above

Image

I expected the path being displayed abov the electrification layer 

1. Add a train
2. Activate the electrification layer
3. Search a path

- **Browser**: Chrome
- **OS**: W11
- **Env**: Staging internal

1482e8c

Synar:
Close #6546

The electrification layer could also be drawn even earlier in LAYER_GROUPS.TOPOGRAPHY, or smaller layer groups could be created to separate tracks, electrification and path.


Caracol3 (approved):
LGTM

Math-R (approved):
Lgtm

Issue #8575, opened 2024-08-23
front: application is slow when opening a big timetable


bloussou:
### What happened?

Rolling stock images are sometimes not here:
![image](https://github.com/user-attachments/assets/000ebc17-3bad-43b1-9034-028785b091b8)

moving the mouse over different part of the page is very slow...

### What did you expect to happen?

Backend is pretty fast, is there a way to immprove the speed here ?

### How can we reproduce it (as minimally and precisely as possible)?

Go here: https://rec-osrd.reseau.sncf.fr/operational-studies/projects/4/studies/198/scenarios/362 (1000 different trains timetable)

1. move your mouse over the list of trains
2. click on different trains
3. when clicking on a train check the map and tables

### On which environments the bug occurs?

Recette (SNCF)

### On which browser the bug occurs?

Firefox

### OSRD version (top right corner `Account` button > `Informations`)

1482e8c

bloussou:
Probably here : https://github.com/OpenRailAssociation/osrd/blob/dev/front/src/utils/date.ts#L88
```
dayjs.tz.guess(), // Format : 'Europe/Paris'
```
<img width="1168" alt="image" src="https://github.com/user-attachments/assets/33d201a9-6ff3-4ef1-987f-e505015b0111">

Same for conflict cards
<img width="1168" alt="image" src="https://github.com/user-attachments/assets/1a1afa7f-9569-4da7-81c3-d4c041731a06">

Maymanaf:
The issue remains unresolved

Castavo:
This actually improved now that we display the list of trains before the infra being loaded.

The general usability of the train list for a large number of train has actually been converted to a feature : https://github.com/osrd-project/osrd-confidential/issues/1187

PR #10371, opened 2025-01-13
timetable performance

### LINKED ISSUE #10377: Large timetable loading performance
Reduce loading and projection time for large time table

1st step for #8575

anisometropie:
closes #10377

Performance improvement with a 1000 trains timetable on a short path. Every single train was being re-rendered on every single frame.

improvements: 
- virtualizing list of trains to only have elements that are actually displayed in the DOM.
- using useCallback on two props to prevent of re-renders on `TimetableTrainCard`
Screenshot_20250113_210147

the 2 `useCallback`s on their own prevent any re-render of `TimetableTrainCard` that are already painted.

adding `VList` on top of the 2 `useCallback`s makes an anonymous wrapper of `TimetableTrainCard` re-render (internal stuff of the library (z prop)) but it’s still an improvement from using useCallback alone.

### Timing

|  case |  timer |   |
|---|---|---|
|  base |  4:27 | https://github.com/user-attachments/assets/23e89906-72c4-43d3-81dc-3336af39c6f6  |
|  VList only | 1:13  |  https://github.com/user-attachments/assets/36366f06-c0ce-46c4-aaca-b61f1ea29984 | 
|  useCallback only |  1:25 | https://github.com/user-attachments/assets/238d8dc5-a106-4904-a7b3-a0f23671c068 | 
|  Both |  1:00 | https://github.com/user-attachments/assets/bbb3e9d6-fa5a-4816-b6d3-432d2a064cd6 | 

a bit of improvement for #8575, but still slow on a click

### DIFF
--- a/front/src/modules/trainschedule/components/Timetable/Timetable.tsx
+++ b/front/src/modules/trainschedule/components/Timetable/Timetable.tsx
@@ -151,7 +151,7 @@ const Timetable = ({
         />
         <Virtualizer overscan={15}>
           {displayedTrainSchedules.map((train: TrainScheduleWithDetails, index) => (
-            <div key={`timetable-train-card-${train.id}-${train.trainName}`}>
+            <div key={`timetable-train-card-${train.id}`}>
               {showDepartureDates[index] && (
                 <div className="scenario-timetable-departure-date">
                   {currentDepartureDates[index]}
--- a/front/src/modules/trainschedule/components/Timetable/Timetable.tsx
+++ b/front/src/modules/trainschedule/components/Timetable/Timetable.tsx
@@ -1,4 +1,4 @@
-import { useMemo, useState } from 'react';
+import { useMemo, useState, useCallback } from 'react';
 
 import cx from 'classnames';
 import dayjs from 'dayjs';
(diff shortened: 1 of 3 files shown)


bloussou:
Nice, will you change the way trains are added in the list (currently 10 after 10) in a following PR ? The backend is ready to accept first train list loading and then the remaining parts and this with parallelized batches.

RomainValls (approved):
I'm not familiar with the virtualizer solution, but performances and code LGTM

front/src/modules/trainschedule/components/Timetable/Timetable.tsx
```diff
@@ -104,6 +108,8 @@ const Timetable = ({
     });
   }, [currentDepartureDates]);
 
+  const VirtualizerRef = useRef<VirtualizerHandle>(null);
```
emersion: Nit:

- I don't think we need this ref?
- This is a regular variable, so should start with a lowercase letter

front/src/modules/trainschedule/components/Timetable/Timetable.tsx
```diff
@@ -142,30 +148,32 @@ const Timetable = ({
...
-              projectionPathIsUsed={
-                infraState === 'CACHED' && trainIdUsedForProjection === train.id
-              }
-              dtoImport={dtoImport}
-            />
-          </div>
-        ))}
+        <Virtualizer ref={VirtualizerRef} overscan={15}>
+          {displayedTrainSchedules.map((train: TrainScheduleWithDetails, index) => (
+            <div key={`timetable-train-card-${train.id}-${train.trainName}`}>
```
emersion: Unrelated to your PR, but I don't think it's a good idea to bake in the name of the train in the key. When renaming the train, React won't be able to match the previous element and the new one. The train ID should be enough (and the prefix is unnecessary).

Feel free to resolve/ignore (or fix in a separate commit if you feel like it!).

emersion (approved):
Nice and simple!

Testing with a timetable with 4k trains, the next bottleneck seems to be the conflict list. Real world data seems to contain a lot of these (I get ~15k) and it slows down the whole page significantly.

emersion:
Other nit for the commit message title: "performance" is uncountable in this context.

front/src/modules/trainschedule/components/Timetable/Timetable.tsx
anisometropie: yes, not needed. I removed it.

front/src/modules/trainschedule/components/Timetable/Timetable.tsx
anisometropie: done!

PR #10381, opened 2025-01-14
front: do not persist STDCM configuration

emersion:
@thibautsailly has confirmed that restoring the STDCM form values wasn't desirable when opening a new STDCM window.

Fixes crashes when reading an old local storage with data which doesn't conform to the new types.

### DIFF
--- a/front/src/reducers/osrdconf/stdcmConf/index.ts
+++ b/front/src/reducers/osrdconf/stdcmConf/index.ts
@@ -62,6 +62,12 @@ export const stdcmConfSlice = createSlice({
       state.speedLimitByTag = stdcmConfInitialState.speedLimitByTag;
       state.linkedTrains = stdcmConfInitialState.linkedTrains;
     },
+    restoreStdcmConfig(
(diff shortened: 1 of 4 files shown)


RomainValls (approved):
LGTM

Wadjetz (approved):

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -62,6 +62,12 @@ export const stdcmConfSlice = createSlice({
       state.speedLimitByTag = stdcmConfInitialState.speedLimitByTag;
       state.linkedTrains = stdcmConfInitialState.linkedTrains;
     },
+    restoreStdcmConfig(
+      _state: Draft<OsrdStdcmConfState>,
+      action: PayloadAction<OsrdStdcmConfState>
+    ) {
+      return action.payload;
```
clarani, line 69: I didn't know we could simply return like this ! Nice :)

clarani (approved):
LGTM, not tested

PR #10382, opened 2025-01-14
editoast, front : fix step update erasing parts of the study in studies page 

### LINKED ISSUE #6455: Changing study status erases starting/expected ending/real ending dates
### What happened?

Starting from this view, with dates working properly:
image

If I change the status here:
image

It erases starting/expected ending/real ending dates:
image

The dates should not change, except the `REAL END` date that sould follow today's date when `Status` is set to `Ended`

1. Create a study with dates here:
image

2. Change the status of the study

- **Browser**: Chrome
- **OS**: W11
- **Env**: satging internal

f062f1f

Synar:
Close #6455

I believe the issue came from the fact editoast does not properly handles a partial studyForm, as the attributes are nullable and editoast does not seem to distinguish between an undefined attribute and a null one. The study changeset is indeed option<option<attribute>>, but the studyForm is simply option<attribute>. I am unsure if we can handle this better, considering the limitations of open api and type conversions between js and rust.

In the meantime, we now send the full study everytime.

Also fix a couple inconsistent translations in study dates

Edit : patches now properly handled in editoast thanks to serde with double option, we only send partial study patches again


front/public/locales/en/operationalStudies/study.json
```diff
@@ -4,10 +4,10 @@
   "createScenario": "Create a scenario",
   "dates": {
     "creation": "Creation",
-    "estimatedend": "Expected ending",
```
RomainValls, line 7: estimatedEnd for readability ? 😄 
or maybe it should be expectedEnd ?

woshilapin (approved):
Great improvement for `editoast` form's. We should actually bring that to other forms now that we know 😄.

I actually tested a `PUT` on `/studies` with `{"description":"something"}`, then with `{"study_type":"something"}` (checking that `description` has not been erased in the process) and it works well. Then finally with `{"description":null}` which works also as expected.

hamz2a (approved):
LGTM, thanks!

leovalais:
> We should actually bring that to other forms now that we know 😄

That's actually quite a lot of them 😨 

I came to really dislike the `flat_*` functions of the changeset. I'd rather have something like `MaybeUnset<T>` which could also be used by API forms (much like what [SeaORM](https://www.sea-ql.org/SeaORM/) does with it's `ActiveRecords` :eyes:). Anyway it has to be refined and it's for another PR, thanks for the fix!

front/public/locales/en/operationalStudies/study.json
Synar: Good catch, lets go with expectedEnd ^^

Yohh:
NIT: if I "missclick" on `ended`, then rollback to `in progress` the `REAL END` keeps a date

Capture d’écran 2025-01-20 à 15 54 11

Synar:
> NIT: if I "missclick" on `ended`, then rollback to `in progress` the `REAL END` keeps a date

We could definitely set the real end date to null if the status is set to anything but ended, but I wanted to keep to the letter of the issue, which stated "The dates should not change, except the REAL END date that sould follow today's date when Status is set to Ended". I don't have a strong opinion on the subject though

clarani (approved):

I'm ok with the current behaviour, but I don't have a strong opinon on this too

Maymanaf (approved):

Issue #7136, opened 2024-04-08
core: stdcm: improve the way we check engineering allowance possibility


Erashin:
We currently run simulations with maximum speedup / slowdown, but this will be incompatible with the next version of running time simulations. Eliminating this simulation-reliant check for engineering allowance possibility would also make the whole process faster.

The idea is to find the range's highest possible acceleration taking the slope into account, and use it to compute a speed/time graph between t=0 and t=time we exit the range + time we want to add with the allowance, which decelerates, potentially levels out, and then accelerates at that given rate. Then use the graph to determine the distance the train would have travelled in this state, if it's above the range's distance, the check fails.

Warning: if the slope is very steep at a specific point of the range, this could skew the approximation. Consider using an average of the slope's steepness over the rolling stock's length.

eckter:
I'll comment this issue as it's the same goal, but I have a new plan for *how* we can improve this.

Context: when testing if an engineering allowance is possible, we're only interested in the most "compact" allowance: one braking phase followed by one accelerating phase. 

The accelerating phase needs to be fully simulated: the traction force and weight force change at every point, sometimes the train can't even maintain its current speed.

But the braking phase is based on a "const gamma": it's not realistically simulated, we just use a constant deceleration value (in m/s²). That means *we don't need to fully compute this section*, we can know from a very cheap computation how long it takes to brake (both time and distance). 

---

We can then use the following algorithm (very rough draft):

We compute the accelerating section, starting from the end. Normal simulation process.

At some point (every block transition?), we test if we've slowed down enough. We do this by using the const deceleration and comparing it to the last edge(s). We see if the total added time is enough.

We then have the "optimal" margin section. We can then test if there's no conflict. 


---


(todo: add some details)

PR #10386, opened 2025-01-15
core: stdcm: increase max length to test for engineering allowances

### LINKED ISSUE #10385: core: stdcm: we're (still) too optimistic on engineering allowances
### What happened?

```
[09:28:42,502] [ERROR]             [Allowance] Couldn't reach target time for allowance section.
[09:28:42,502] [ERROR]             [Allowance] Closest time = 54.482671, target time = 77.69749198843365 +- 0.003634815373274719

[...]


java.lang.RuntimeException: Failed to compute a simulation that wouldn't cause conflicts: mismatch between exploration and postprocessing (please open a bug report)
	at fr.sncf.osrd.stdcm.graph.PostProcessingSimulationKt.handlePostProcessingConflict-WS7ElmQ(PostProcessingSimulation.kt:472)
	at fr.sncf.osrd.stdcm.graph.PostProcessingSimulationKt.buildFinalEnvelope(PostProcessingSimulation.kt:104)
	at fr.sncf.osrd.stdcm.graph.PostProcessingSimulationKt.handlePostProcessingConflict-WS7ElmQ(PostProcessingSimulation.kt:457)
	at fr.sncf.osrd.stdcm.graph.PostProcessingSimulationKt.buildFinalEnvelope(PostProcessingSimulation.kt:104)
	at fr.sncf.osrd.stdcm.graph.PostProcessingSimulationKt.buildFinalEnvelope$default(PostProcessingSimulation.kt:51)
	at fr.sncf.osrd.stdcm.graph.STDCMPostProcessing.makeResult(STDCMPostProcessing.kt:83)
	at fr.sncf.osrd.stdcm.graph.STDCMPathfinding.findPath(STDCMPathfinding.kt:134)
	at fr.sncf.osrd.stdcm.graph.STDCMPathfindingKt.findPath(STDCMPathfinding.kt:77)
	at fr.sncf.osrd.api.api_v2.stdcm.STDCMEndpointV2.run(STDCMEndpointV2.kt:110)
	at fr.sncf.osrd.cli.ReproduceRequest.run(ReproduceRequest.java:48)
	at fr.sncf.osrd.App.main(App.java:44)
```

_No response_

Fuzzer, I have payloads

Local

Firefox

cf771fa328350158227e4fab8cda5d24bb3c4884

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/10385 (at least the few cases I've seen recently, I'll re-check if it still happens afterwards)

So apparently even 20km isn't always enough to add any amount of delay, because sometimes we barely have the traction to maintain the current speed. Ideally we should remove that check altogether, but it would still cause performance issues (repeatedly computing margins over potentially thousands of kilometers would certainly cause timeouts). We should look into https://github.com/OpenRailAssociation/osrd/issues/7136 for a proper solution. 

EDIT: this version is better but it's still not enough. I've found a case with a **300km long** allowance section that was still impossible. We may need to rework some stuff. 

---

So there's two extra fixes:

1. Small bugfix on how we compute the "slowest envelope": we'd stop early if the max speed envelope ended with a speedup (we'd "intersect" right away)
2. In post-processing, prefill the fixed time point list with the end of every engineering allowance. This avoids some unnecessary and constraining extra fixed points at the wrong places.

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/EngineeringAllowanceManager.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/EngineeringAllowanceManager.kt
@@ -43,7 +43,7 @@ class EngineeringAllowanceManager(private val graph: STDCMGraph) {
         if (affectedEdges.isEmpty()) return false // No space to try the allowance
 
         val length = affectedEdges.map { it.length.distance }.sumDistances()
-        if (length > 20_000.meters) {
+        if (length > 50_000.meters) {
             // If the allowance area is large enough to reasonably stop and accelerate again, we
             // just accept the solution. This avoids computation on very large paths
             // (which can be quite time expensive)
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/EngineeringAllowanceManager.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/EngineeringAllowanceManager.kt
@@ -7,6 +7,7 @@ import fr.sncf.osrd.envelope.part.EnvelopePart
 import fr.sncf.osrd.envelope.part.EnvelopePartBuilder
 import fr.sncf.osrd.envelope.part.constraints.EnvelopeConstraint
 import fr.sncf.osrd.envelope.part.constraints.EnvelopePartConstraintType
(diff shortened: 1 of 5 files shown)


Erashin (approved):
Nice!

PR #10387, opened 2025-01-15
front: fix selectableSlot in useLinkedTrainSearch

SarahBellaha:
For this stdcm environment search window  :
```
{
    "id": 55,
    "infra_id": 2,
    "timetable_id": 1,
    "search_window_begin": "2024-09-16T22:00:00Z",
    "search_window_end": "2024-09-18T21:59:59Z"
}
```
we should only have the 17th and the 18th displayed in the `DatePicker`.

But we have the wrong dates for the linked train search:
image


Math-R (approved):
lgtm & tested

PR #10402, opened 2025-01-16
core: add ETCS SVL logic to ETCS braking simulator

### LINKED ISSUE #10404: core: ETCS: implement SVL braking curves logic
Implement ETCS braking curves for SVLs. With release speed (compulsory).

Care bug SvL after path end: tractive effort curves won't have values for ebd calculation -> don't do it, issue created here: https://github.com/OpenRailAssociation/osrd/issues/11036

- 40 km/h release-speed fixed value only for SvL for now? -> YES
- Only consider one intersection? -> NO

Erashin:
Fixes #10404.
Add SVL logic to ETCS braking simulator.
How: compute svl and eoa indications, compute minimum curve, and compare it against overlay.
Limit: approximate braking curves' calculation when SvL is after the end of the path, due to the missing tractive effort and slopes post path.

### DIFF
--- a/core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/EnvelopeSimPath.java
+++ b/core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/EnvelopeSimPath.java
@@ -111,6 +111,10 @@ public final class EnvelopeSimPath implements PhysicsPath {
         // TODO: Optimise method by adding in a cache.
         int indexBegin = getIndexBeforePos(begin);
         int indexEnd = getIndexBeforePos(end);
+        // TODO: Remove if we extend path properties until last SvL > path.length.
+        if (indexBegin == indexEnd && indexBegin == gradePositions.length - 1)
+            // Take last grade value in this case
+            return gradeValues[gradeValues.length - 1];
         var lowestGradient = gradeValues[indexBegin];
         for (int i = indexBegin; i < indexEnd; i++) {
             var grad = gradeValues[i];
--- a/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/Constants.kt
+++ b/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/Constants.kt
@@ -11,43 +11,49 @@ package fr.sncf.osrd.envelope_sim.etcs
  * National Default Value: permission to inhibit the compensation of the speed measurement accuracy.
  * See Subset 026: table in Appendix A.3.2.
  */
-const val qNvinhsmicperm = false
+const val Q_NVINHSMICPERM = false
+
+/**
+ * National Default Value: permission to follow release speed at 40km/h near the EoA. See Subset
+ * 026: table in Appendix A.3.2.
+ */
+const val NATIONAL_RELEASE_SPEED = 40.0 / 3.6 // m/s
 
 /**
  * Estimated acceleration during tBerem, worst case scenario (aEst2 is between 0 and 0.4), expressed
  * in m/s². See Subset 026: §3.13.9.3.2.9.
  */
-const val aEst2 = 0.4
+const val A_EST_2 = 0.4
 
 /** See Subset 026: table in Appendix A.3.1. */
-const val dvEbiMin = 7.5 / 3.6 // m/s
-const val dvEbiMax = 15.0 / 3.6 // m/s
-const val vEbiMin = 110.0 / 3.6 // m/s
-const val vEbiMax = 210.0 / 3.6 // m/s
-const val tWarning = 2.0 // s
-const val tDriver = 4.0 // s
-const val mRotatingMax = 15.0 // %
-const val mRotatingMin = 2.0 // %
+const val DV_EBI_MIN = 7.5 / 3.6 // m/s
+const val DV_EBI_MAX = 15.0 / 3.6 // m/s
+const val V_EBI_MIN = 110.0 / 3.6 // m/s
+const val V_EBI_MAX = 210.0 / 3.6 // m/s
+const val T_WARNING = 2.0 // s
+const val T_DRIVER = 4.0 // s
+const val M_ROTATING_MIN = 2.0 // %
+const val M_ROTATING_MAX = 15.0 // %
 
 /** See Subset 041: §5.3.1.2. */
-const val vUraMinLimit = 30 / 3.6 // m/s
-const val vUraMaxLimit = 500 / 3.6 // m/s
-const val vUraMin = 2 / 3.6 // m/s
-const val vUraMax = 12 / 3.6 // m/s
+const val V_URA_MIN_LIMIT = 30 / 3.6 // m/s
+const val V_URA_MAX_LIMIT = 500 / 3.6 // m/s
+const val V_URA_MIN = 2 / 3.6 // m/s
+const val V_URA_MAX = 12 / 3.6 // m/s
 
 /** See Subset 041: §5.3.1.2. */
 fun vUra(speed: Double): Double {
-    return interpolateLinearSpeed(speed, vUraMinLimit, vUraMaxLimit, vUraMin, vUraMax)
+    return interpolateLinearSpeed(speed, V_URA_MIN_LIMIT, V_URA_MAX_LIMIT, V_URA_MIN, V_URA_MAX)
 }
 
 /** See Subset 026: §3.13.9.3.2.10. */
 fun vDelta0(speed: Double): Double {
-    return if (!qNvinhsmicperm) vUra(speed) else 0.0
+    return if (!Q_NVINHSMICPERM) vUra(speed) else 0.0
 }
 
 /** See Subset 026: §3.13.9.2.3. */
 fun dvEbi(speed: Double): Double {
-    return interpolateLinearSpeed(speed, vEbiMin, vEbiMax, dvEbiMin, dvEbiMax)
+    return interpolateLinearSpeed(speed, V_EBI_MIN, V_EBI_MAX, DV_EBI_MIN, DV_EBI_MAX)
 }
 
 /**
--- a/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
+++ b/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
@@ -47,56 +47,40 @@ fun addBrakingCurvesAtEOAs(
 ): Envelope {
     val sortedEndsOfAuthority = endsOfAuthority.sortedBy { it.offsetEOA }
     var beginPos = envelope.beginPos
+    val maxSpeedEnvelope = envelope.maxSpeed
     val builder = OverlayEnvelopeBuilder.forward(envelope)
     for (endOfAuthority in sortedEndsOfAuthority) {
         val targetPosition = endOfAuthority.offsetEOA.distance.meters
         assert(targetPosition > 0.0)
         val targetSpeed = 0.0
-        val maxSpeedEnvelope = envelope.maxSpeed
-        val overhead =
-            Envelope.make(
-                EnvelopePart.generateTimes(
-                    listOf(EnvelopeProfile.CONSTANT_SPEED),
-                    doubleArrayOf(0.0, targetPosition),
-                    doubleArrayOf(maxSpeedEnvelope, maxSpeedEnvelope)
-                )
-            )
-        val sbdCurve =
-            computeBrakingCurve(
-                context,
-                overhead,
-                targetPosition,
-                targetSpeed,
-                BrakingType.ETCS_SBD
-            )
-        assert(sbdCurve.beginPos >= 0 && sbdCurve.endPos == targetPosition)
-        assert(sbdCurve.endSpeed == targetSpeed)
-        val guiCurve =
-            computeBrakingCurve(
-                context,
-                overhead,
-                targetPosition,
-                targetSpeed,
-                BrakingType.ETCS_GUI
-            )
-        assert(guiCurve.beginPos >= 0.0 && guiCurve.endPos == targetPosition)
-        assert((guiCurve.beginSpeed == maxSpeedEnvelope || guiCurve.beginPos == 0.0))
-        assert(guiCurve.endSpeed == targetSpeed)
+        val fullIndicationCurveEoa =
+            computeSbdFullIndicationCurve(context, targetPosition, maxSpeedEnvelope)
+        var fullIndicationCurveStop = Envelope.make(fullIndicationCurveEoa)
 
-        val fullIndicationCurve =
-            computeIndicationBrakingCurveFromRef(context, sbdCurve, BrakingCurveType.SBD, guiCurve)
-        assert(fullIndicationCurve.endPos == targetPosition)
-        assert(fullIndicationCurve.endSpeed == targetSpeed)
+        if (endOfAuthority.offsetSVL != null) {
+            val targetPositionSvl = endOfAuthority.offsetSVL.distance.meters
+            val fullIndicationCurveSvl =
+                computeEbdFullIndicationCurve(
+                    context,
+                    targetPositionSvl,
+                    targetSpeed,
+                    maxSpeedEnvelope
+                )
+            fullIndicationCurveStop =
+                computeMinSvlEoaIndCurve(
+                    fullIndicationCurveEoa,
+                    fullIndicationCurveSvl,
+                    targetPosition
+                )
+        }
 
         val indicationCurve =
-            keepBrakingCurveUnderOverlay(Envelope.make(fullIndicationCurve), envelope, beginPos)
-                ?: continue
+            keepBrakingCurveUnderOverlay(fullIndicationCurveStop, envelope, beginPos) ?: continue
         assert(indicationCurve.beginPos >= beginPos && indicationCurve.endPos == targetPosition)
         assert(
             indicationCurve.beginSpeed <= maxSpeedEnvelope &&
                 indicationCurve.endSpeed == targetSpeed
         )
-
         builder.addPart(indicationCurve)
 
         // We build EOAs along the path. We need to handle overlaps with the next EOA. To do so, we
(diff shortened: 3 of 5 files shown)


core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -47,57 +47,61 @@ fun addBrakingCurvesAtEOAs(
 ): Envelope {
     val sortedEndsOfAuthority = endsOfAuthority.sortedBy { it.offsetEOA }
     var beginPos = envelope.beginPos
-    val builder = OverlayEnvelopeBuilder.forward(envelope)
+    val maxSpeedEnvelope = envelope.maxSpeed
+    var envelopeWithEoaBrakingCurves = envelope
```
bougue-pe: To distinguish EoA, SvL and stop (min of EoA and SvL):
```suggestion
    var envelopeWithStopBrakingCurves = envelope
```

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -203,6 +168,123 @@ fun addBrakingCurvesAtLOAs(
...
+    // TODO: uncomment once ebd is not shifted anymore for an LoA
+    // assert((ebdCurve.beginSpeed == maxSpeedEbd || ebdCurve.beginPos == 0.0))
+
+    val guiCurve =
+        computeBrakingCurve(context, overhead, targetPosition, targetSpeed, BrakingType.ETCS_GUI)
+    assert(guiCurve.beginPos >= 0.0 && guiCurve.endPos == targetPosition)
+    assert((guiCurve.beginSpeed == maxSpeedEbd || guiCurve.beginPos == 0.0))
+
+    val ebiCurve = computeEbiBrakingCurveFromEbd(context, ebdCurve, targetSpeed)
+    assert(ebiCurve.endSpeed == targetSpeed)
```
bougue-pe, line 306: How about a `computeBrakingCurve(context, overhead, targetPosition, targetSpeed, BrakingType.ETCS_EBI)` ?
As it doesn't need anything else than the EBD, we could "hide" it.

At least stick `ebiCurve` right after `ebdCurve` here would be nice/enough?

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/Constants.kt
```diff
@@ -13,6 +13,12 @@ package fr.sncf.osrd.envelope_sim.etcs
  */
 const val qNvinhsmicperm = false
 
+/**
+ * National Default Value: permission to follow release speed at 40km/h near the EoA. See Subset
+ * 026: table in Appendix A.3.2.
+ */
+const val releaseSpeed = 40.0 / 3.6 // m/s
```
bougue-pe: After reading the use, I wonder if these should be SCREAMING_SNAKE_CASE to advertize for constness (in a separate commit at least, maybe separate PR).

bougue-pe: Separate matter (as the release speed is probably actually the min of a lot of values, we may avoid name-squating it):
```suggestion
const val nationalReleaseSpeed = 40.0 / 3.6 // m/s
```

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -203,6 +168,123 @@ fun addBrakingCurvesAtLOAs(
...
+    fullIndicationCurveSvl: EnvelopePart,
+    targetPosition: Double,
+    overlay: Envelope,
+    beginPos: Double
+): EnvelopePart? {
+    // If no part of the indication curve is before the target position, ignore it.
+    if (fullIndicationCurveSvl.beginPos >= targetPosition) return null
+    val fullIndCurveSvlWithMaintain: Envelope
+    val releaseSpeedPosition =
+        if (fullIndicationCurveSvl.maxSpeed >= releaseSpeed)
```
bougue-pe: Can't we make `releaseSpeed` a param, and probably rename it `targetSpeed` ?
I guess you see where this is going for the LoA a bit later 😉

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -353,12 +441,25 @@ private fun keepBrakingCurveUnderOverlay(
...
+        fullBrakingCurve.minSpeed >
+            Envelope.make(
+                    *overlay.slice(
+                        max(fullBrakingCurve.beginPos, beginPos),
+                        min(fullBrakingCurve.endPos, overlay.endPos)
+                    )
+                )
+                .maxSpeed
+    ) {
+        // The full braking curve is above the overlay envelope: nothing to do here.
```
bougue-pe, line 487: Should we check for more, like "are they intersecting?": `fullBrakingCurve.minSpeed < overlaySlice.minSpeed && fullBrakingCurve.maxSpeed >= overlaySlice.maxSpeed `

Maybe we should just try/catch for the absence of intersection which is probably legitimate?

This questions will be less important after applying min instead of looking for just one intersection, but there might still be some cases between LoA and stop?

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -383,11 +484,23 @@ private fun keepBrakingCurveUnderOverlay(
...
+    var lastIndex = nbPoints - 1
+    while (
+        speeds[lastIndex] >
+            overlay
+                .get(overlay.findRightDir(positions[lastIndex], -1.0))
+                .interpolateSpeed(positions[lastIndex])
+    ) {
+        lastIndex--
+        if (lastIndex == 0) return null
+    }
```
bougue-pe: How about moving it to a dedicated method `findLastPointRespectingOverlay`? (maybe even inside OverlayBuilder?)
To help document this weird code a bit more.

tests/tests/test_train_schedule.py
```diff
@@ -301,6 +301,205 @@ def test_etcs_schedule_result_stop_brake_from_mrsp(etcs_scenario: Scenario, etcs
...
+    # Check a bending point for the first stop's braking curve (where the Guidance curve's influence stops).
+    offset_bending_guidance_point = 37_279_507
+    speed_at_bending_guidance_point = _get_current_or_next_speed_at(
+        simulation_final_output, offset_bending_guidance_point
+    )
+    _assert_equal_speeds(speed_at_bending_guidance_point, kph2ms(223.112_006_6))
+    # Check that the release part (where the speed stays at 40km/h) starts and ends at the expected offsets.
+    offset_start_release_speed = 40_827_882
+    speed_at_start_release_speed = _get_current_or_next_speed_at(simulation_final_output, offset_start_release_speed)
+    _assert_equal_speeds(speed_at_start_release_speed, kph2ms(40))
```
bougue-pe: Not sure: should we also make a const with the release speed? (ditto below)
```suggestion
    _assert_equal_speeds(speed_at_start_release_speed, RELEASE_SPEED_40)
```

tests/tests/test_train_schedule.py
```diff
@@ -301,6 +301,205 @@ def test_etcs_schedule_result_stop_brake_from_mrsp(etcs_scenario: Scenario, etcs
...
+        _get_current_or_next_speed_at(simulation_final_output, offset_start_first_brake + 1) < speed_before_first_brake
+    )
+    # Check the first bending point for the first stop's braking curve (where the Guidance curve's influence stops).
+    offset_bending_guidance_point = 37_240_601
+    speed_at_bending_guidance_point = _get_current_or_next_speed_at(
+        simulation_final_output, offset_bending_guidance_point
+    )
+    _assert_equal_speeds(speed_at_bending_guidance_point, kph2ms(219.751_941_6))
+    # Check the second bending point for the first stop's braking curve, where the indication curve followed switches
+    # from the SvL indication curve to the EoA indication curve.
```
bougue-pe: reminder: as checked, curves are not crossing on this case (because of GUIs)

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: This has disappeared, but it would have been a good point.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/Constants.kt
Erashin: Probably separate commit, but yeah, literally just thought about it at some point and forgot all about it. Good catch.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: The release speed is now encapsulated in a svl specific method, so I think this won't be relevant anymore. Feel free to tell me if not :)

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: I think your condition just does not work. For ex, a stop/speed limit close to the path would creat an sbd/ebd curve going 'til path start, with a max speed of maybe 10km/h, which would be below overlaySlice's max speed. Especially since the overlay is not continuous, it's tricky to check that there is an intersection going only on speed data, and not taking into account the position constraints.
The check I'm doing here guarantees a logical braking curve, by which i mean a part of it at least is under the overlay and should be added.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Actually, `overlay` is an `Envelope`, so we can't actually move it to `OverlayBuilder`. I'll add it to the non ending list of methods in this file though :D

tests/tests/test_train_schedule.py
Erashin: Yep this will change in the last commit.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Yeah I'm not quite sure I understand this.
If what you mean is that we go for `computeBrakingCurve(..., BrakingcurveType.EBI)` we'd need to add a ref curve in the arguments of the method to compute the ebi curve, and the only thing the code would do would be to call `computeEbiBrakingCurveFromEbd` -> not good imo, unnecessary.
If what you mean is that we just skip EBD curve calculation, and only do one call, which computes EBD into EBI directly, then I don't want to do it, since it would go against the future feature where we potentially wanna add all these different curves in the front end.
As to your proposition of putting it straight after ebd, nah, it wouldn't be coherent with the order of the method calculations in the other methods. What's more, i quite like regrouping closeby the curves that come directly from the rolling stock model (i.e. EBD SBD and GUI).

bougue-pe: Right, you understood correctly and have good points, resolving it.
The last point is a matter of taste I guess, and it's fine as-is (I like more to have an uninterrupted flow whenever possible).

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
bougue-pe: True, it doesn't work (I can think of other examples, like when the curves are intersecting twice...).

The current check prevents obvious never-intersecting cases, which is nice to have.

But I still think there can be other less-obvious legitimate cases, and we would throw, no? Not here probably, I'll try to keep it in mind.
After discussion, that should never happen, so I'm resolving it, thanks 🙏

Erashin (commented):
@bougue-pe : fixed in https://github.com/OpenRailAssociation/osrd/pull/10402/commits/e3dbe473581a9ad394a2ccf6c10ebc2f68a1af17 and https://github.com/OpenRailAssociation/osrd/pull/10402/commits/1e7ce6971e01a98ee1e6dbfaca7dde4bf7f054d6.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -47,56 +47,40 @@ fun addBrakingCurvesAtEOAs(
...
-            )
-        assert(guiCurve.beginPos >= 0.0 && guiCurve.endPos == targetPosition)
-        assert((guiCurve.beginSpeed == maxSpeedEnvelope || guiCurve.beginPos == 0.0))
-        assert(guiCurve.endSpeed == targetSpeed)
-
-        val fullIndicationCurve =
-            computeIndicationBrakingCurveFromRef(context, sbdCurve, BrakingCurveType.SBD, guiCurve)
-        assert(fullIndicationCurve.endPos == targetPosition)
-        assert(fullIndicationCurve.endSpeed == targetSpeed)
+            fullIndicationCurve =
```
bougue-pe: nit:
```suggestion
            fullIndicationCurveStop =
```

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -110,6 +90,95 @@ fun addBrakingCurvesAtEOAs(
     return builder.build()
 }
 
+private fun computeMinSvlEoaIndCurve(
```
bougue-pe, line 93: Not mandatory at all: If we wanted to generify this, we could provide a shortened SvL curve (cut below release speed) and we would have a generic min.
Adding the constant speed to join second curve when first curve stops would be OK I think.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -110,6 +90,95 @@ fun addBrakingCurvesAtEOAs(
...
+            val otherSpeed = slicedIndicationCurveSvl.interpolateSpeed(newPos)
+            indicationPositions[i] = newPos
+            // TODO: unneeded for now: interpolate to not approximate position at intersection.
+            indicationSpeeds[i] = min(otherSpeed, newSpeed)
+        } else {
+            indicationPositions[i] = otherCurve.endPos
+            indicationSpeeds[i] = otherCurve.endSpeed
+            // Clean up the last unneeded points in the arrays before exiting the loop.
+            indicationPositions = indicationPositions.dropLast(pointCount - 1 - i).toDoubleArray()
+            indicationSpeeds = indicationSpeeds.dropLast(pointCount - 1 - i).toDoubleArray()
```
bougue-pe, line 159: reminder: With this fix and the test update, this last commit must be squashed to the major commit too.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
bougue-pe: With the case change and all, I agree and resolve.
We'll change it eventually when the need arises.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -106,6 +90,95 @@ fun addBrakingCurvesAtEOAs(
...
+    var refCurve = slicedIndicationCurveEoa
+    var otherCurve = slicedIndicationCurveSvl
+    if (
+        slicedIndicationCurveEoa.interpolateSpeed(startIntersectingRange) >
+            slicedIndicationCurveSvl.interpolateSpeed(startIntersectingRange)
+    ) {
+        refCurve = slicedIndicationCurveSvl
+        otherCurve = slicedIndicationCurveEoa
+    }
+    val pointCount = refCurve.pointCount()
```
bougue-pe, line 140: Idea (feel free to resolve by yourself once decided what to do):
```suggestion
    // TODO: uneeded for now: interate over points of both curves to improve precision
    val pointCount = refCurve.pointCount()
```

tests/tests/test_train_schedule.py
```diff
@@ -480,7 +480,7 @@ def test_etcs_schedule_result_stop_with_eoa_and_svl_at_different_locations(
             prev_speed = current_speed
 
     # Check that the braking curve starts and ends at the expected offsets.
-    offset_start_first_brake = 33_261_530
+    offset_start_first_brake = 33_361_530
```
bougue-pe, line 484: In case you get an idea to "enforce" that: this is the same value as the previous test (driven by the SvL curve).
Maybe: just comment, mutualize in a `START_SVL_BRAKE_ON_PH1_SWITCH` value, nothing?

bougue-pe (approved):
👏 Nice, thank you very much for the work !

Only nits and not-mandatory comments (or even just ideas for the future): no problem resolving them by yourself once decided/replied 👍

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Extracted the minimum part and made it a static method in `EnvelopePart`. I'm just calling it here in this method :)

Edit: I actually got tired of keeping the TODOs. So I'll just do the fully correct method in `EnvelopePart` in another PR:
https://github.com/OpenRailAssociation/osrd/pull/10991

tests/tests/test_train_schedule.py
Erashin: It isn't. Since it's the EoA, i.e. the stop, that is 100m before in this case, they aren't the same values. However i can mutualize them, and put `START_SVL_BRAKE_ON_PH1_SWITCH - 100_000` here.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: See comment above: another PR incoming managing this issue directly :)
https://github.com/OpenRailAssociation/osrd/pull/10991

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Soooooo... yeah the commits conflicted with each others, as well as the screamin snake case commit... So i took the easy way out and just kept 2 commits: SVL logic with screamin snake case commit in the first, and then the hack commit.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/EnvelopeSimPath.java
```diff
@@ -111,6 +111,10 @@ public double getMinGrade(double begin, double end) {
         // TODO: Optimise method by adding in a cache.
         int indexBegin = getIndexBeforePos(begin);
         int indexEnd = getIndexBeforePos(end);
+        // TODO: Remove when we extend path properties until last SvL > path.length.
```
eckter: Do we plan on doing this? It would probably be best, but we assume quite often that we simulate the train up to the end of the path properties

eckter (approved):
Good job on the tests

I only skimmed through the core changes, I trust PE's review on this

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope_sim/EnvelopeSimPath.java
Erashin: I didn't check the code. Basically we'll open an issue, then look at what this modification would entail. If it messes too many things up, we might just not do it.

eckter: I'd just reword it and change the "when" into an "if" then

bougue-pe: OK for the if, and for the issue, opened: https://github.com/OpenRailAssociation/osrd/issues/11036

To be decided in the issue I guess: we should at least be pessimistic about the grade if we don't check what's next after the path.

tests/tests/test_train_schedule.py
```diff
@@ -301,6 +304,213 @@ def test_etcs_schedule_result_stop_brake_from_mrsp(etcs_scenario: Scenario, etcs
...
+
+    # This case hits an LoA curve (slowdown of the MRSP), but it's not the point to test it here.
+
+    # Check that the curves does respect the EoA + SvL (EoA = stops), and that there is an
+    # acceleration then deceleration in between (maintain speed when reach the MRSP).
+    first_stop_offset = SVL_STOP_OFFSET
+    final_stop_offset = 45_060_000
+    stop_offsets = [
+        0,
+        first_stop_offset,
```
bougue-pe: As this `SVL_STOP_OFFSET` is not so "test generic", I would remove it and just explicitly use the named value:
```suggestion
    svl_ph1_offset = 41_138_000
    final_stop_offset = 45_060_000
    stop_offsets = [
        0,
        svl_ph1_offset,
```

tests/tests/test_train_schedule.py
```diff
@@ -301,6 +304,213 @@ def test_etcs_schedule_result_stop_brake_from_mrsp(etcs_scenario: Scenario, etcs
...
+    # To debug this test: please add a breakpoint then use front to display speed-space chart
+    # (activate Context for Slopes and Speed limits).
+
+    # This case hits an LoA curve (slowdown of the MRSP), but it's not the point to test it here.
+
+    # Check that the curves does respect the EoA + SvL (EoA = stops), and that there is an
+    # acceleration then deceleration in between (maintain speed when reach the MRSP).
+    # Here, the EoA is at 41038m, and the SvL is at the switch at 41138m.
+    distance_to_svl = 100_000
+    first_stop_offset = SVL_STOP_OFFSET - distance_to_svl
```
bougue-pe: Following the same idea:
```suggestion
    # Check that the curves does respect the EoA + SvL (EoA = stops), and that there is an
    # acceleration then deceleration in between (maintain speed when reach the MRSP).
    # Here, stop (EoA) is 100m before the PH1 switch (SvL).
    svl_ph1_offset = 41_138_000
    first_stop_offset = svl_ph1_offset - 100_000
```

tests/tests/test_train_schedule.py
bougue-pe: Oh, sorry on the mistake on the values being identical. 🙏 
Then I think we can't mutualize (resolving this and opening a separate comment).

tests/tests/test_train_schedule.py
```diff
@@ -301,6 +304,213 @@ def test_etcs_schedule_result_stop_brake_from_mrsp(etcs_scenario: Scenario, etcs
...
+            current_speed = simulation_final_output["speeds"][pos_index]
+            if accelerating:
+                if prev_speed > current_speed:
+                    accelerating = False
+            else:
+                assert prev_speed >= current_speed
+            prev_speed = current_speed
+
+    # Check that the braking curve starts and ends at the expected offsets.
+    offset_start_first_brake = START_SVL_BRAKE_ON_PH1_SWITCH - distance_to_svl
```
bougue-pe: As mentioned, it looks to me like we shouldn't mutualize it, as it's only "lucky" that both EoA and SvL upper curve parts are under GUI influence and therefore separated by the same value as the EoA-SvL distance.
"Hard-documentation" though variable names seems too strong (and hides a lot of complexity about how the curves are computed) IMO.

Also, the values are not enough "test-generic" IMO (same as the `SVL_STOP_OFFSET`).

I wouldn't detail what's happening here at all (or just a comment, if you will).
```suggestion
    offset_start_first_brake = 33_361_530
```

🦜 Ditto for the "same location" test above.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -429,7 +580,12 @@ private fun computeBecParams(
             rollingStock.getRollingResistance(speed),
             weightForce,
             speed,
-            PhysicsRollingStock.getMaxEffort(speed, context.tractiveEffortCurveMap.get(position)),
+            PhysicsRollingStock.getMaxEffort(
+                speed,
+                // TODO: have a tractive effort curve map which extends until the last SvL instead
+                // of the end of the path.
+                context.tractiveEffortCurveMap.get(min(position, context.path.length))
```
bougue-pe, line 587: This is also part of the "end-of-the-path" hack (tracked it into https://github.com/OpenRailAssociation/osrd/issues/11036).
Yet, after thinking of it, we shouldn't have any of the 2 hacks in a separate commit, as the commit without the hacks will break: only one commit for the whole PR would be my pick on this 1️⃣.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
bougue-pe: OK, resolving this as it's fine as-is and the conversation in #10991 will keep track of those ideas.
(same for iteration over points of both curves)

PR #10405, opened 2025-01-16
front: fix project image display in project page

### LINKED ISSUE #10400: Project image not displayed on firefox
### What happened?

Using **firefox** the image project does not display when opening the project.

We should see the image (like chromium)

1. Use firefox
2. Create a project with an image
3. Open it and observe that we don't see the image 🐛 

Local

Firefox

9ddc06c0bd92b791c8a64356da59b2cbabbf8da3

RomainValls:
close #10400


emersion (approved):
Hm, I can't reproduce on my machine… The image displays correctly in both cases.

That said, I can't see why we were using absolute positioning here.

PR #10413, opened 2025-01-16
front: fix stdcm consist warning

### LINKED ISSUE #10409: lmr : warning for length and weight value excludes upper bound
### What happened?

Currently, in LMR we have : 

Image

(same for length with 750m)

We expect to have no warning when the value is equal to the upper bound

as above

Recette (SNCF)

Google Chrome

9151b80

SharglutDev:
fix #10409


bloussou (approved):
Thanks ✅

clarani (approved):

theocrsb (approved):
LGTM and tested :)

PR #10418, opened 2025-01-16
front: fix stdcm consist length and mass minimum values

### LINKED ISSUE #10306: Picking some rolling stocks triggers unwanted warning

SharglutDev:
The sum of the rolling stock and towed rolling stock mass was previously rounded down (instead of up) resulting to some invalid values send to the form. If a total mass was 80.7t, the user was allowed to send 80t.

fix #10306


theocrsb (approved):
LGTM

emersion (approved):

Akctarus (approved):
Tested, LGTM !

kmer2016 (approved):
LGTM !

Maymanaf (approved):

PR #10423, opened 2025-01-16
front: reduce e2e test execution time

### LINKED ISSUE #10321: front: reduce or remove unnecessary timeouts in e2e tests
### Description and goal

- Eliminate redundant timeouts
- Store the utilized timeouts in a constants file
- Shorten test timeout durations 
- Increase the number of workers (to be evaluated)
- Restrict the maximum number of failed tests (to be evaluated)
- Reduce the retry count
- use env variable for osrd language

### Acceptance criteria

- [x] All tests pass with reduced execution time
- [x] No flaky test when the pipeline is executed 10 times

Maymanaf:
Closes #10321


front/playwright.config.ts
```diff
@@ -12,31 +12,33 @@ export default defineConfig({
...
   /* Run tests in files in parallel */
   fullyParallel: true,
   /*
    * Limit parallelism in CI based on CPU capacity,
    * running 50% of the available workers when in CI.
-   * Otherwise, run tests with a single worker.
+   * Otherwise, run tests with 2 workers.
    */
-  workers: process.env.CI ? '50%' : 1,
+  workers: process.env.CI ? '50%' : 2,
```
Yohh:
```suggestion
  workers:'50%',
```

I think the most important is for the CI, a developper can set a specific number of workers him/herself if needed

front/playwright.config.ts
```diff
@@ -12,31 +12,33 @@ export default defineConfig({
...
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
-  /* Retry up to 2 times on CI, and 1 time otherwise */
-  retries: process.env.CI ? 2 : 1,
+  /* Retry 1 time if a test fail */
```
Yohh:
```suggestion
```

this comment and many others in this file should be removed as the key explained seems explicit enough, do you agree?

front/tests/002-project-management.spec.ts
```diff
@@ -46,15 +48,7 @@ test.describe('Validate the Operational Study Project workflow', () => {
...
   /** *************** Test 2 **************** */
-  test('Update an existing project', async ({ page }) => {
-    // Create a project
-    project = await createProject(generateUniqueName(projectData.name));
-
-    // Navigate to the Operational Studies projects page
-    await page.goto('/operational-studies/projects');
-
-    const homePage = new HomePage(page);
-    const projectPage = new ProjectPage(page);
```
Yohh: so tests 1, 2 and 3 can't be parallelized anymore?

front/tests/008-train-schedule.spec.ts
```diff
@@ -5,19 +5,20 @@ import {
...
-import HomePage from './pages/home-page-model';
+import test from './logging-fixture';
 import OperationalStudiesTimetablePage from './pages/op-timetable-page-model';
-import test from './test-logger';
 import { waitForInfraStateToBeCached } from './utils';
 import { getInfra, getProject, getScenario, getStudy } from './utils/api-setup';
 
 test.describe('Verify train schedule elements and filters', () => {
   test.slow();
+  test.use({ viewport: { width: 1920, height: 1080 } });
```
Yohh, line 15: why is it needed?

front/tests/assets/timeout-const.ts
```diff
@@ -0,0 +1,4 @@
+export const LOAD_PAGE_TIMEOUT = 30_000;
+export const SIMULATION_RESULT_TIMEOUT = 30_000;
+export const STDCM_SIMULATION_TIMEOUT = 30_000;
```
Yohh, line 3: could be one `DEFAULT_TIMEOUT`, or do you think it will change soon?

front/tests/utils/index.ts
```diff
@@ -119,17 +107,18 @@ export function formatDateToDayMonthYear(dateString: string, OSRDLanguage: strin
...
  * Displays the total time taken for the state to reach 'CACHED'.
  *
  * @param infraId - The ID of the infrastructure to retrieve and check.
- * @throws {Error} - Throws an error if the state does not become 'CACHED' within 5 minutes.
+ * @throws {Error} - Throws an error if the state does not become 'CACHED' within 60 seconds.
  * @returns {Promise<void>} - Resolves when the state is 'CACHED'.
  */
 export const waitForInfraStateToBeCached = async (infraId: number): Promise<void> => {
-  const maxRetries = 18; // Total attempts (3 minutes / 10 seconds)
+  const maxRetries = 6; // Total attempts (60 seconds / 10 seconds)
```
Yohh, line 105: why 6 precisely?

Yohh, line 121: why 6 precisely?

front/tests/003-study-management.spec.ts
```diff
@@ -72,14 +69,14 @@ test.describe('Validate the Study creation workflow', () => {
 
   /** *************** Test 2 **************** */
   test('Update an existing study', async ({ page }) => {
-    const studyPage = new StudyPage(page);
+    const PROJECT_LANGUAGE = process.env.PROJECT_LANGUAGE || '';
```
Yohh: what if `PROJECT_LANGUAGE = ''` ?

Yohh (commented):
great, thanks!!

Yohh (commented):
great, thanks!!

front/tests/002-project-management.spec.ts
Yohh: sorry, didn't see the `beforeach`

front/tests/011-op-times-and-stops-tab.spec.ts
```diff
@@ -113,15 +112,9 @@ test.describe('Times and Stops Tab Verification', () => {
...
-      new OperationalStudiesInputTablePage(page),
-      new OperationalStudiesOutputTablePage(page),
-      new OperationalStudiesPage(page),
-      new RoutePage(page),
-    ];
-
     // Set translations based on selected language
-    const translations = OSRDLanguage === 'English' ? enTranslations : frTranslations;
+    const PROJECT_LANGUAGE = process.env.PROJECT_LANGUAGE || '';
+    const translations = PROJECT_LANGUAGE === 'English' ? enTranslations : frTranslations;
```
Uriel-Sautron: This code is used twice in the file and could be included in the beforeEach or beforeAll

front/tests/008-train-schedule.spec.ts
Maymanaf: To avoid flakiness, as the train timetable may not display the list correctly on smaller viewport

front/tests/assets/timeout-const.ts
Maymanaf: It can be used in the future to lower the timeout if we achieve better performance for STDCM as an example

front/tests/utils/index.ts
Maymanaf: The total timeout is 60 seconds, so if we retry every 10 seconds, there will be 6 retries

front/playwright.config.ts
Maymanaf: Removed all the comments in `playwright.config.ts`

front/tests/003-study-management.spec.ts
Maymanaf: My bad, I missed that. Now that I’ve created a utility function, it should throw an error

Yohh (approved):

Uriel-Sautron (approved):
LGTM, tested

front/tests/global-setup.ts
```diff
@@ -22,5 +23,11 @@ setup('setup', async () => {
   await deleteRollingStocks(ROLLING_STOCK_NAMES);
 
   await createDataForTests();
+
+  logger.info('Retrieving project language ...');
+  const homePage = new HomePage(page);
+  await homePage.goToHomePage();
+  process.env.PROJECT_LANGUAGE = await homePage.getOSRDLanguage();
```
clarani, line 29: Hum is it a good practice to put things in the global environment like this ? :thinking:

front/tests/utils/index.ts
```diff
@@ -188,3 +189,26 @@ export async function performOnSpecificOSAndBrowser(
...
+ * @param {string} projectLanguage - The language of the project
+ * @param {{ en: T; fr: T }} translations - An object containing translations for English and French.
+ * @returns {T} - The translation object corresponding to the specified project language.
+ * @throws {Error} - Throws an error if the project language is unsupported.
+ */
+
+export function getTranslations<T = unknown>(
+  projectLanguage: string,
+  translations: { en: T; fr: T }
+): T {
```
clarani: - why not get the language from the process.env here ?
- you don't need to precise unknown
```suggestions
export function getTranslations<T>(translations: { en: T; fr: T }): T {
  const projectLanguage = process.env.PROJECT_LANGUAGE;
```

front/tests/002-project-management.spec.ts
```diff
@@ -1,23 +1,25 @@
...
 
-  /** *************** Test 1 **************** */
-  test('Create a new project', async ({ page }) => {
-    const projectPage = new ProjectPage(page);
-
-    // Navigate to the Operational Studies projects page
+  test.beforeEach('Navigate to the projects page', async ({ page }) => {
+    projectPage = new ProjectPage(page);
+    // Create a project
+    project = await createProject(generateUniqueName(projectData.name));
```
clarani: You don't delete the project created for the first test.

Maybe it would be better to add a afterEach which checks if the project created in the beforeEach still exists and deletes it if it exists.
It would be better, like this if we add a new test in this file, we won't forget to manually delete it at the end of the test

front/tests/global-setup.ts
clarani: I read a bit the documentation, it seems to be OK. If you use the process.env, then you should use it everywhere directly (you don't need to instantiate the const PROJECT_LANGUAGE in each test)

front/tests/utils/index.ts
Maymanaf: Done

front/tests/002-project-management.spec.ts
Maymanaf: The project is deleted at the end of each test. Since the names vary each time, some checks may be unnecessary

front/tests/global-setup.ts
Maymanaf: It's not recommended if the `process.env` value is dynamic or can change during the tests, but in our case, it's fine. An alternative solution is to use it in the `fixture`. I tried that but encountered some issues, which will take some time to resolve.
Maybe I should handle this in a separate issue, what do you think?

front/tests/006-stdcm.spec.ts
```diff
@@ -52,44 +49,40 @@ test.describe('Verify stdcm simulation page', () => {
 
   test.beforeEach('Navigate to the STDCM page', async ({ page }) => {
     // Retrieve OSRD language and navigate to STDCM page
```
clarani:
```suggestion
```

front/tests/013-stdcm-simulation-sheet.spec.ts
```diff
@@ -45,21 +40,18 @@ test.describe('Verify stdcm simulation page', () => {
 
   test.beforeEach('Navigate to the STDCM page', async ({ page }) => {
     // Retrieve OSRD language and navigate to STDCM page
```
clarani:
```suggestion
```

front/tests/pages/stdcm-page-model.ts
clarani, line 1: You need to use `getTranslations` in `fillAndVerifyViaDetails` too

front/tests/pages/stdcm-page-model.ts
```diff
@@ -508,7 +500,7 @@ class STDCMPage {
   }
 
   // Fill and verify destination details based on selected language
```
clarani:
```suggestion
  // Fill and verify destination details
```

front/tests/global-setup.ts
clarani: Yes it's fine for me

front/tests/002-project-management.spec.ts
clarani: I got your point, but I still think it's quite fragile.

Example in the first test `Create a new project`:
- a project 1 is created in the useBefore hook
- during the test, a project 2 is created manually
- at the end of the test, the project 2 is deleted
- the project 1 is not deleted

clarani (approved):
LGTM :white_check_mark:

PR #10345, opened 2025-01-13
core: aggregate errors that happen too often during infra loading

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/8773

After:

```
[15:46:49,574] [INFO]          [ValidateInfra] parsing json
[15:46:52,963] [INFO]          [ValidateInfra] parsing RailJSON
[15:46:57,358] [WARN]      [RawInfraRJSParser] Electrification conflict on track-range TRACK_ID[53m, 1168m]: 25000V != 3000V
[15:46:57,358] [WARN]      [RawInfraRJSParser] Electrification conflict on track-range TRACK_ID[23m, 487m]: 25000V != 3000V
[15:46:57,358] [WARN]      [RawInfraRJSParser] Electrification conflict on track-range TRACK_ID[487m, 974m]: 25000V != 3000V
[15:46:57,371] [WARN]      [RawInfraRJSParser] ... and 90 other similar errors
[15:47:10,282] [INFO]          [ValidateInfra] loading signals
[15:47:10,502] [INFO]          [ValidateInfra] building blocks
[15:47:12,318] [DEBUG]          [BlockBuilder] no signal at non buffer stop detector.DETECTOR_ID:DECREASING
[15:47:12,318] [DEBUG]          [BlockBuilder] no signal at non buffer stop detector.DETECTOR_ID:INCREASING
[15:47:12,318] [DEBUG]          [BlockBuilder] no signal at non buffer stop detector.DETECTOR_ID:INCREASING
[15:47:12,328] [DEBUG]          [BlockBuilder] ... and 1256 other similar errors
[15:47:12,536] [DEBUG] [SignalingSimulatorImpl] error in block BLOCK_ID: too_few_signals
[15:47:12,536] [DEBUG] [SignalingSimulatorImpl] error in block BLOCK_ID: too_few_signals
[15:47:12,536] [DEBUG] [SignalingSimulatorImpl] error in block BLOCK_ID: too_few_signals
[15:47:12,537] [DEBUG] [SignalingSimulatorImpl] ... and 955 other similar errors
[15:47:12,537] [INFO]          [ValidateInfra] done
```

Before: same thing but over 2k2 lines

PR #10426, opened 2025-01-16
core: add warning on route ending on non route delimiting signal in block generation

### LINKED ISSUE #4606: core: the block generator does not ensure routes are bounded by appropriate signals
### What happened?

In the real world, only some signals can be used to end / start routes. These signal cannot be passed when closed due to interlocking (`Nf` signals in french signaling).

Currently, we let block delimiting signals start and end routes, regardless of whether they can actually signal for the end of a route. 

We should check whether signals at the end / start of routes are actually belong there, and raise a warning if they don't.

In practice, the `is_route_delimiter` property is missing from the current implementation.

1. Create a route which ends with a non-route delimiting signal
2. No warning whatsoever

- **Browser**: Firefox v109
- **OS**: Nixos v22.11
- **Env**: Local

2421cbdb7d8085f6a541b0b2ffc053f51e1a9988

Synar:
Close #4606

We only check that route ends on Nf signals (or more generally route delimiting signal), not whether they start on it

### DIFF
--- a/core/kt-osrd-signaling/src/main/kotlin/fr/sncf/osrd/signaling/impl/BlockBuilder.kt
+++ b/core/kt-osrd-signaling/src/main/kotlin/fr/sncf/osrd/signaling/impl/BlockBuilder.kt
@@ -20,6 +20,7 @@ internal fun internalBuildBlocks(
     val signalDelimiters = findSignalDelimiters(rawSignalingInfra, loadedSignalInfra)
     val detectorEntrySignals = makeDetectorEntrySignals(loadedSignalInfra, signalDelimiters)
     val missingSignalLogAggregator = LogAggregator({ logger.debug(it) })
+    val nonRouteDelimitingSignalLogAggregator = LogAggregator({ logger.debug(it) })
     val result =
         blockInfraBuilder(loadedSignalInfra, rawSignalingInfra) {
             // Step 2) iterate on zone paths along the route path.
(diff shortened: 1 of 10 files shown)


Khoyo (approved):

core/kt-osrd-signaling/src/main/kotlin/fr/sncf/osrd/signaling/impl/BlockBuilder.kt
```diff
@@ -334,3 +347,26 @@ private fun BlockInfraBuilder.updatePartialBlocks(
...
+    for (associatedSignal in endSignals.values()) {
+        val logicalSignalId = associatedSignal.signal
+        val signalingSystem = loadedSignalInfra.getSignalingSystem(logicalSignalId)
+        val sigSettings = loadedSignalInfra.getSettings(logicalSignalId)
+        val routeEndsWithRouteEndingSignal =
+            sigModuleManager.isRouteDelimiter(signalingSystem, sigSettings)
+        if (!routeEndsWithRouteEndingSignal) {
+            logger.debug {
+                "Route ${rawSignalingInfra.getRouteName(route)} ends with non-route delimiting signal on signaling system ${signalingSystem}"
+            }
```
eckter: How often does it happen when loading France? I expect a lot?

I've just merge a branch that aggregate some of those warnings to avoid flooding the datadog: https://github.com/OpenRailAssociation/osrd/pull/10345. Do you think it could be possible to use it?

Synar: Sure, good idea, I will do so!

Btw, is this debug log not aggregated because it doesn't occur on prod?

>  logger.debug { 
>    "unterminated block at end of route ${rawSignalingInfra.getRouteName(route)}" 
>  }

Synar: Hmm, the error simply does not trigger when loading France. I am unsure whether the check is not broad enough or if the infra France is simply clean on this regard.

eckter: It may be normal actually. Routes are generated during the import, and they may be generated in a way that avoids this error entirely.

(Which makes my comment a little irrelevant, sorry about that)

eckter (approved):

PR #10429, opened 2025-01-17
core: stdcm: use the right limit when shifting departure time

eckter:


Erashin (approved):
Ah yeah, that's better.

Issue #10337, opened 2025-01-13
front: !debug mode only! stdcm path step are not updated when modifying date input


louisgreiner:
### What happened?

/!\ it only happens in debug mode

The STDCM path step input are not updated when i change the input in the form:
https://github.com/user-attachments/assets/c8f8ecbe-503f-4da7-9966-7ce3aba7b5cb

I thus can't re-simulate after a simulation if I don't trigger another stdcm path step modification

Also, I looks with redux the state, and i don't see any update stdcm path step entry : my guess is that the debug mode overloads the traces that update the state

However, the input of the request sent is correct :/

### What did you expect to happen?

input correctly updated

### How can we reproduce it (as minimally and precisely as possible)?

1. see the video
2. or just select a date and observe with redux

### On which environments the bug occurs?

Recette (SNCF)

### On which browser the bug occurs?

Firefox

### OSRD version (top right corner `Account` button > `Informations`)

-

kmer2016:
I believe this issue has already been resolved. I can no longer reproduce it, and the Redux store updates correctly.

clarani:
Validated ✅

PR #10432, opened 2025-01-17
front: display only trains around stdcm result

### LINKED ISSUE #9131: stdcm: only display trains that surround the result in the GET

Us, for debugging and demoing

When STDCM finds a result, it is displayed in the GET with all the trains from the time table. The lazy loading takes about 2-3min in big timetables.
However, viewing the trains that are far away from the result, time-speaking, is not really usefull.

The goal here is to filter the trains that are displayed to view only the ones surounding the result (something like -1h before the departure and +1h after the departure, it needs to be tested).

This feature matters because it will make the loading of all the trains in the GET much faster, by displaying less trains.

This needs to be tested, and maybe adjusted (try +/- 1h first, and maybe end up with +/-2h for example).

clarani:
closes #9131

### DIFF
--- a/front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
+++ b/front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
@@ -17,8 +17,8 @@ type UseLazyLoadTrainsProps = {
   infraId?: number;
   trainIdsToFetch?: number[];
   trainSchedules?: TrainScheduleResult[];
-  setTrainIdsToFetch: Dispatch<SetStateAction<number[] | undefined>>;
-  setTrainIdsToProject: Dispatch<SetStateAction<Set<number>>>;
+  setTrainIdsToFetch?: Dispatch<SetStateAction<number[] | undefined>>;
+  setTrainIdsToProject?: Dispatch<SetStateAction<Set<number>>>;
 };
 
 /**
@@ -79,9 +79,6 @@ const useLazyLoadTrains = ({
         });
 
         if (!outOfSync) {
-          // launch the projection of the trains
-          setTrainIdsToProject((prev) => new Set([...prev, ...packageToFetch]));
-
           // format the summaries to display them in the timetable
           const newFormattedSummaries = formatTrainScheduleSummaries(
             packageToFetch,
@@ -89,6 +86,9 @@ const useLazyLoadTrains = ({
             trainSchedulesById,
             rollingStocks!
           );
+
+          // launch the projection of the trains if needed
+          setTrainIdsToProject?.((prev) => new Set([...prev, ...packageToFetch]));
 
           // as formattedSummaries is a dictionary, we replace the previous values with the new ones
           setTrainScheduleSummariesById((prev) => concatMap(prev, newFormattedSummaries));
--- a/front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
+++ b/front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
@@ -3,13 +3,8 @@ import { useEffect, useState, type Dispatch, type SetStateAction, useMemo } from
 
 import { useSelector } from 'react-redux';
 
-import {
-  osrdEditoastApi,
-  type PathfindingResultSuccess,
-  type TrainScheduleResult,
-} from 'common/api/osrdEditoastApi';
+import { osrdEditoastApi, type TrainScheduleResult } from 'common/api/osrdEditoastApi';
 import { useOsrdConfSelectors } from 'common/osrdContext';
-import useLazyProjectTrains from 'modules/simulationResult/components/SpaceTimeChart/useLazyProjectTrains';
 import type { TrainScheduleWithDetails } from 'modules/trainschedule/components/Timetable/types';
 import { getBatchPackage } from 'utils/batch';
 import { concatMap, mapBy } from 'utils/types';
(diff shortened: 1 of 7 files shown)


bloussou:
changed

louisgreiner:
Nice ! Actually could also fix this https://github.com/OpenRailAssociation/osrd/issues/10337

front/src/applications/operationalStudies/hooks/useScenarioData.ts
```diff
@@ -40,21 +42,31 @@ const useScenarioData = (
...
-    setTrainIdsToFetch,
+    trainIdsToProject,
     path: projectionPath?.path,
     trainSchedules,
+    moreTrainsToCome: !allTrainsLoaded,
+    setTrainIdsToProject,
   });
 
+  useEffect(() => {
+    if (trainSchedules && projectionPath?.path) {
```
emersion: Previously, there was a `allTrainsLoaded` check here as well. Has it been removed on purpose?

front/src/applications/stdcm/components/StdcmResults/StdcmDebugResults.tsx
```diff
@@ -53,32 +53,36 @@ const StdcmDebugResults = ({
...
-          setHeight={setManchetteWithSpaceTimeChartHeight}
-          minHeight={MANCHETTE_WITH_SPACE_TIME_CHART_DEFAULT_HEIGHT}
-        >
-          <div
-            className="osrd-simulation-container mb-2"
-            style={{
-              height: manchetteWithSpaceTimeChartHeight,
-            }}
+      {projectedData &&
+        projectedData?.spaceTimeData.length > 0 &&
```
emersion, line 58: Follow-up: osrd-ui shouldn't crash when `spaceTimeData` is empty. Could we open an issue to track this bug?

front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
```diff
@@ -91,12 +88,15 @@ const useLazyLoadTrains = ({
...
+          // launch the projection of the trains if needed
+          setTrainIdsToProject?.((prev) => new Set([...prev, ...packageToFetch]));
+
           // as formattedSummaries is a dictionary, we replace the previous values with the new ones
           setTrainScheduleSummariesById((prev) => concatMap(prev, newFormattedSummaries));
         }
       }
 
-      setTrainIdsToFetch([]);
+      setTrainIdsToFetch?.([]);
```
emersion: Instead of making `setTrainIdsToFetch` optional, could we remove it and make it the caller's responsibility to reset the list of train IDs to fetch if they need to? They can do it with a `useEffect()` on `allTrainsLoaded` I think?

front/src/applications/stdcm/hooks/useProjectedTrainsForStdcm.ts
```diff
@@ -53,14 +102,35 @@ const useProjectedTrainsForStdcm = (stdcmResponse?: StdcmSuccessResponse) => {
   });
 
   useEffect(() => {
-    const newSpaceTimeData = Array.from(projectedTimetableTrainsById.values());
+    if (stdcmResponse) {
+      const relevantTrainScheduleIds = keepTrainsRunningDuringStdcm(
+        stdcmResponse,
+        trainScheduleSummariesById
+      );
```
emersion, line 105: Could we maybe extract this as a `useMemo()`? That way we don't have to duplicate it in both `useEffect()` calls.

emersion: Hm, sounds like that wouldn't work well since we omit dependencies in `useEffect()`…

front/src/applications/stdcm/hooks/useProjectedTrainsForStdcm.ts
```diff
@@ -2,14 +2,62 @@ import { useEffect, useMemo, useState } from 'react';
...
+import { addDurationToDate, subtractDurationFromDate } from 'utils/date';
 
 import formatStdcmTrainIntoSpaceTimeData from '../utils/formatStdcmIntoSpaceTimeData';
 
+/**
+ * Project only the trains which leave 1 hour max before the departure of the stdcm or
+ * arrive 1 hour max after the arrival of the stdcm
+ */
+const keepTrainsRunningDuringStdcm = (
+  stdcmResult: StdcmSuccessResponse | undefined,
```
emersion: All callers check that `stdcmResult` isn't undefined, so I think we can drop `| undefined` here?

emersion:
The first and last commits are reviewed-by me. The middle commit "front: display only trains around the stdcm train on the GET in debug mode" is not.

front/src/applications/stdcm/components/StdcmResults/StdcmDebugResults.tsx
clarani: Here is the [bug](https://github.com/OpenRailAssociation/osrd/issues/10496)

front/src/applications/operationalStudies/hooks/useScenarioData.ts
clarani: Hum it was a mistake I think, though I'm not sure what the expected behavior is if we change the projection when the trains are not all loaded...
I keep the previous behavior

front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
clarani: Done !

front/src/applications/stdcm/components/StdcmResults/StdcmDebugResults.tsx
emersion: Thanks!

front/src/applications/operationalStudies/hooks/useScenarioData.ts
emersion: WFM!

emersion (approved):
LGTM! :)

PR #10434, opened 2025-01-17
front: add version parameter to translations load path

### LINKED ISSUE #9965: Traductions not loading
### What happened?

About everywhere in the front, sometimes, traductions will not load, even when clearing every cache imaginable.
It seems to be linked to new features being released / new traductions entries being added.

Here is a recent example:
https://github.com/user-attachments/assets/d994c1c0-6fbf-4469-ae83-7e3fd7057661

I expect the traduction entries to always be available.

> [!NOTE]
> The translation files loaded by the browser should have a custom URL (containing a hash) so that when it changes everything is reloaded correctly.

1. Empty every local storage you've got
2. Reload the page
3. Go to a scenario page with an electrical profile set chosen, witness
4. Go to the output time table, witness

Recette (SNCF)

Firefox

94d3a84

emersion:
That way, every time we release a new OSRD version, we force browsers to refresh translation files. This is similar to the hash inserted in JS/CSS filenames by vite.

Docs: https://github.com/i18next/i18next-xhr-backend?tab=readme-ov-file#backend-options

Closes: https://github.com/OpenRailAssociation/osrd/issues/9965

### DIFF
--- a/front/src/i18n.ts
+++ b/front/src/i18n.ts
@@ -6,6 +6,8 @@ import { initReactI18next } from 'react-i18next';
 // Official languages codes to use from IANA
 // https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry
 
+const version = encodeURIComponent(import.meta.env.VITE_OSRD_GIT_DESCRIBE);
+
 i18n
   .use(Backend)
   .use(initReactI18next)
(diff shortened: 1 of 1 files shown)


kmer2016 (approved):
Good job !

RomainValls (approved):
LGTM

PR #10435, opened 2025-01-17
front: migrate to i18next-http-backend

emersion:
i18next-xhr-backend uses the old XMLHttpRequest API, so it has been deprecated and replaced with a fetch()-based implementation: https://github.com/i18next/i18next-xhr-backend?tab=readme-ov-file#deprecated


SharglutDev (approved):

Akctarus (approved):
tested, LGTM !

PR #10438, opened 2025-01-17
front: fix opstudies itinary reversal dysfunctioning when times and margin where set

### LINKED ISSUE #10036: Reversing the itinerary messes up the schedule points
### What happened?

Starting from a train that has specific schedule points, if i reverse the itinerary the timings and margins are kept, but they become inconsistent because they are in the wrong order : 

Image

Image

Image

3 options here:

- brutal one: reversing the itinerary changes the itinerary so all the input table should be reset
- 🔆nicer one🔆: keep the stop times but not the timings, so that stops remain stops
- 🔆🔆🔆nicer one premium shiny pack🔆🔆🔆: also keep the margins but adapt them so that they are correct : a margin applies on a zone, so reverting the itinerary should keep the margin on this zone. In the table this means placing the 5% at the beginning of the reversed zone.

Reproduce the description

Recette (SNCF)

Google Chrome

9e75ce2

Synar:
Close #10036 (🔆🔆🔆nicer one premium shiny pack🔆🔆🔆 version)

### DIFF
--- a/front/src/modules/pathfinding/components/Itinerary/Itinerary.tsx
+++ b/front/src/modules/pathfinding/components/Itinerary/Itinerary.tsx
@@ -14,6 +14,7 @@ import { useOsrdConfSelectors } from 'common/osrdContext';
 import Tipped from 'common/Tipped';
 import Pathfinding from 'modules/pathfinding/components/Pathfinding/Pathfinding';
 import TypeAndPath from 'modules/pathfinding/components/Pathfinding/TypeAndPath';
+import reversePathSteps from 'modules/pathfinding/helpers/reversePathSteps';
 import { setWarning } from 'reducers/main';
 import { updateViewport } from 'reducers/map';
 import { getMap } from 'reducers/map/selectors';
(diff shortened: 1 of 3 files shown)


front/src/modules/pathfinding/components/Itinerary/Itinerary.tsx
```diff
@@ -73,7 +73,30 @@ const Itinerary = () => {
 
   const inverseOD = () => {
     notifyRestrictionResetWarning();
-    const newPathSteps = [...pathSteps].reverse();
+
+    // Reverse start and end of margins, in prevision of reversing the list of path steps
+    const newMargins: (string | undefined)[] = [];
+    let prevMargin: string | undefined;
+    pathSteps.forEach((pathStep, index) => {
```
SharglutDev: I think we can `compact` here, the button isn't available if there is no origin/destination (the only ones who can be `null`). This way, you can avoid the other checks for that in the function.
```suggestion
    compact.pathSteps.forEach((pathStep, index) => {
```

Synar: Sure, done!

SharglutDev (approved):
Lgtm and tested, thank you for the shiny pack version 🔆🔆🔆

achrafmohye (approved):
LGTM and tested 👍

emersion:
I think we want to move the margin of each path step to the next path step before reversing. I think this version would be easier to follow:

```ts
const newPathSteps = pathSteps.map((step, index) => ({
  ...step,
  arrival: null,
  theoreticalMargin: pathSteps[index - 1]?.theoreticalMargin,
})).reverse();
```

Or am I missing something?

Would be nice to have tests for this.

emersion (commented):
Dear GitHub, I've looked at this PR, now you can hide it from my review queue.

Synar:
> I think we want to move the margin of each path step to the next path step before reversing. I think this version would be easier to follow:
> 
> ```ts
> const newPathSteps = pathSteps.map((step, index) => ({
>   ...step,
>   arrival: null,
>   theoreticalMargin: pathSteps[index - 1]?.theoreticalMargin,
> })).reverse();
> ```
> 
> Or am I missing something?
> 
> Would be nice to have tests for this.

The issue is that a lot of intermediary path steps don't have a theoretical margins, and we should skip them when exchanging margins. So you should only set margins that were already set (plus start/end), the value to replace it with should be the last margin that was actually set, not simply the previous margin.

To give an example, [0%, undef, 3%, undef, undef, undef, 1%, undef] means we have 0% margin for 2 steps traveled, then 3% for 4, then 1% for 1. 

So it should give first [undef, undef, 0%, undef, undef, undef, 3%, 1%] before reversing, then after [1%, 3%, undef, undef, undef, 0%, undef, undef] which is 1% for 1 step, 3% for 4, and  0% for 2. Same as the initial distrib, but reversed.

Your version would give [undef, 0%, undef, 3%, undef, undef, undef, 1%] then [1%, undef, undef, undef, 3%, undef, 0%, undef] so 1% for 4 steps, 3% for 2, and 0% for 1. Different from the initial distrib.

I will write a test ^^

Synar:
@emersion I've added a couple tests ^^

PR #10440, opened 2025-01-20
core: clean up etcs braking simulator

Erashin:
Rest of the clean ups commented by @bougue-pe in https://github.com/OpenRailAssociation/osrd/pull/10316


bougue-pe (approved):
Thanks !

eckter (approved):

PR #10442, opened 2025-01-20
front: upgrade node to v23, nivo to v0.88

emersion:
A newer node version [is required][1] to make require() work with
modules. This is necessary for a nivo upgrade, otherwise CI fails
with:

     FAIL  src/modules/rollingStock/components/RollingStockEditor/__tests__/RollingStockEditorForm.spec.ts [ src/modules/rollingStock/components/RollingStockEditor/__tests__/RollingStockEditorForm.spec.ts ]
    Error: require() of ES Module /app/node_modules/d3-interpolate/src/index.js from /app/node_modules/@nivo/core/dist/nivo-core.cjs.js not supported.
    Instead change the require of index.js in /app/node_modules/@nivo/core/dist/nivo-core.cjs.js to a dynamic import() which is available in all CommonJS modules.
     ❯ Object.<anonymous> node_modules/@nivo/core/dist/nivo-core.cjs.js:1:164

[1]: https://github.com/plouc/nivo/issues/2310#issuecomment-2495916321

See also: https://github.com/OpenRailAssociation/osrd/pull/10256

### DIFF
--- a/front/docker/Dockerfile.devel
+++ b/front/docker/Dockerfile.devel
@@ -5,7 +5,7 @@ COPY docker/exec-as.c .
 RUN gcc -std=c99 -static -o /exec-as exec-as.c
 
 
-FROM node:20
+FROM node:23
 
 # Install dependencies
 RUN apt-get update -yqq && \
--- a/front/docker/Dockerfile.nginx
+++ b/front/docker/Dockerfile.nginx
@@ -1,6 +1,6 @@
 ### BUILD STAGE
 
-FROM node:20-bookworm AS build
+FROM node:23-bookworm AS build
 
 WORKDIR /app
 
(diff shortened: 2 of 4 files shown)


Yohh (commented):
you should modify `flake.nix`too:
```nix
 fixedNode = pkgs.nodejs_23;
```

emersion:
> you should modify flake.nix too

Sorry, I can't do that. Updating `flake.nix` isn't enough AFAIU, the lock file also needs to be updated together, and I don't have/want the nix tool installed locally.

This upgrade is just about CI and Docker, and e.g. I continue to use my distro's node package: we don't need everybody to use the same version. Feel free to submit a separate PR to upgrade nix-related stuff.

SharglutDev (approved):
Lgtm, thank you !

PR #10445, opened 2025-01-20
front: lmr fix intermediate op id collision

### LINKED ISSUE #10375: front: multiple intermediate OP are changed together
### What happened?

When I modify the second intermediate OP, the first intermediate OP is also modified with the same values at the same time.

Each intermediate OP should be independently changed and isolated from one another.

1. Add a first intermediate OP and fill in some values
2. Add a second intermediate OP
3. Fill in some values in the second, and observe that the first one changed

See also the following video.

https://github.com/user-attachments/assets/44e6547a-7658-498d-9f24-a3b480f036b1

Other

Firefox

fdfd8ac

Wadjetz:
fix #10375

Use UUID instead of react-id-generator. The latter is a sequential ID generator with a global counter that can reset when the page is refreshed and data is restored from localStorage or the window object, potentially resulting in duplicated IDs.

- [ ] Fix tests


SharglutDev (approved):
Lgtm and tested (you probably need to rebase)

woshilapin (approved):
Looks fine although, I'm pretty sure I don't understand it in the details 😄

front/src/utils/react-id.ts
```diff
@@ -0,0 +1,10 @@
+import { v4 as uuidV4 } from 'uuid';
+
+/**
+ * Generate a unique id for react components
+ * @returns a unique id
+ */
+export default function nextId() {
+  // return `id-${uuidV4()}`;
+  return uuidV4();
```
emersion: I think we can use `uuidV4()` directly in places where it makes sense, rather than introduce a wrapper function. (Note, the name "react-id" is slightly confusing, since these IDs aren't used in anything React-related.)

emersion (approved):
Very nice, thanks!

Maymanaf (approved):
LGTM for e2e tests

PR #9105, opened 2024-09-30
front: nge saving node's positions

sim51:
Related to https://github.com/osrd-project/osrd-confidential/issues/597

What is done : 
- [X] Editoast API
- [X] Sync between OSRD/NGE
- [X] Using a layout algo for node's position
- [X] On load, deduplicate nodes
- [x] Creating new nodes in NGE
- [x] testing import from viriato, basic & opendata
- [x] check the trainrun modification

Close https://github.com/OpenRailAssociation/osrd/issues/9031

PR #10448, opened 2025-01-20
front: fix macro trainruns frequencies

### LINKED ISSUE #10389: nge: frequency feature is broken
### What happened?

Image

It seems we've lost the frequency feature : only 60' is remaining, and even if we set freq from OSRD it is not working

_No response_

1. try change freq in NGE

Recette (SNCF)

Firefox

-

### LINKED ISSUE #10552: nge: stop_for calculation is inverted
### What happened?

When a stop is set, here from minute 26 to 28:

Image

The `stop_for` value is wrong for the first intermediary node (the second being null is fine):

Image

28-26=2, not -2

1. create train schedule and set arrival and departure time from a waypoint

Recette (SNCF)

Firefox

-

louisgreiner:
After https://github.com/OpenRailAssociation/osrd/pull/9105, we've lost the feature the [macro trainrun frequencies](https://github.com/OpenRailAssociation/osrd/pull/8917)

This way of handling macro train frequencies will be revamp during https://github.com/osrd-project/osrd-confidential/issues/779

Closes https://github.com/OpenRailAssociation/osrd/issues/10389
Closes https://github.com/OpenRailAssociation/osrd/issues/10552

### DIFF
--- a/front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
@@ -6,123 +6,28 @@ import buildOpSearchQuery from 'modules/operationalPoint/helpers/buildOpSearchQu
 import type { AppDispatch } from 'store';
 import { Duration, addDurationToDate } from 'utils/duration';
 
+import {
+  TRAINRUN_CATEGORY_HALTEZEITEN,
+  NODE_LABEL_GROUP,
+  DEFAULT_TRAINRUN_FREQUENCIES,
+  DEFAULT_TRAINRUN_CATEGORY,
+  DEFAULT_TRAINRUN_FREQUENCY,
+  DEFAULT_TRAINRUN_TIME_CATEGORY,
+  TRAINRUN_LABEL_GROUP,
+  DEFAULT_TIME_LOCK,
+  DEFAULT_DTO,
+} from './consts';
 import MacroEditorState, { type NodeIndexed } from './MacroEditorState';
-import { deleteMacroNodeByDbId, getSavedMacroNodes } from './utils';
+import { deleteMacroNodeByDbId, getSavedMacroNodes, trainrunFrequencyFromLabel } from './utils';
 import {
   type PortDto,
   type TimeLockDto,
   type TrainrunSectionDto,
-  type TrainrunCategory,
-  type TrainrunTimeCategory,
   type TrainrunFrequency,
   type NetzgrafikDto,
-  type LabelGroupDto,
   PortAlignment,
   type LabelDto,
 } from '../NGE/types';
-
-const TRAINRUN_CATEGORY_HALTEZEITEN = {
-  HaltezeitIPV: { haltezeit: 0, no_halt: false },
-  HaltezeitA: { haltezeit: 0, no_halt: false },
-  HaltezeitB: { haltezeit: 0, no_halt: false },
-  HaltezeitC: { haltezeit: 0, no_halt: false },
-  HaltezeitD: { haltezeit: 0, no_halt: false },
-  HaltezeitUncategorized: { haltezeit: 0, no_halt: false },
-};
-
-const TRAINRUN_LABEL_GROUP: LabelGroupDto = {
-  id: 1,
-  name: 'Default',
-  labelRef: 'Trainrun',
-};
-const NODE_LABEL_GROUP: LabelGroupDto = {
-  id: 2,
-  name: 'Node',
-  labelRef: 'Node',
-};
-
-const DEFAULT_TRAINRUN_CATEGORY: TrainrunCategory = {
-  id: 1, // In NGE, Trainrun.DEFAULT_TRAINRUN_CATEGORY
-  order: 0,
-  name: 'Default',
-  shortName: '', // TODO: find a better way to hide this in the graph
-  fachCategory: 'HaltezeitUncategorized',
-  colorRef: 'EC',
-  minimalTurnaroundTime: 0,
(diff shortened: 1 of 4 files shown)


louisgreiner:
Depends on https://github.com/OpenRailAssociation/osrd/pull/10486

front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
```diff
@@ -303,16 +303,39 @@ export const loadAndIndexNge = async (
...
+      if (trainSchedule.labels) {
+        trainSchedule.labels.forEach((label) => {
+          if (!label.match(/^frequency::(?!30$|60$|120$)\d+$/)) {
+            const frequency = parseInt(label.split('::')[1], 10);
+            const trainrunFrequencyFound = DEFAULT_TRAINRUN_FREQUENCIES.find(
+              (freq) => freq.frequency === frequency
+            );
+            if (
+              trainrunFrequencyFound &&
+              (!trainrunFrequency || trainrunFrequencyFound.frequency < trainrunFrequency.frequency)
```
SharglutDev: Probably a lack of nge knowledge but what is the reason to keep the last lower frequency (and not any frequency) ?

front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
```diff
@@ -303,16 +303,39 @@ export const loadAndIndexNge = async (
...
-      labelIds: (trainSchedule.labels || []).map((l) =>
-        labels.findIndex((e) => e.label === l && e.labelGroupId === TRAINRUN_LABEL_GROUP.id)
-      ),
-    }));
+    .map((trainSchedule) => {
+      // NGE trainrun frequency is stored as OSRD labels ('frequency::30' or 'frequency::120')
+      let trainrunFrequency: TrainrunFrequency | undefined;
+      if (trainSchedule.labels) {
+        trainSchedule.labels.forEach((label) => {
+          if (!label.match(/^frequency::(?!30$|60$|120$)\d+$/)) {
```
SharglutDev: nit : since it is used two times, we could maybe extract `/^frequency::(?!30$|60$|120$)\d+$/` in a `const`.

front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
```diff
@@ -303,16 +303,39 @@ export const loadAndIndexNge = async (
...
+        });
+      }
+
+      return {
+        id: trainSchedule.id,
+        name: trainSchedule.train_name,
+        categoryId: DEFAULT_TRAINRUN_CATEGORY.id,
+        frequencyId: trainrunFrequency?.id || DEFAULT_TRAINRUN_FREQUENCY.id,
+        trainrunTimeCategoryId: DEFAULT_TRAINRUN_TIME_CATEGORY.id,
+        labelIds: (trainSchedule.labels || [])
```
SharglutDev: Maybe it could help to know why we want to store other frequencies than 30/60/120 (and why we don't want to keep the other regular labels)

SharglutDev (commented):
Works great just some questions :)

front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
louisgreiner: As a trainrun can have only 1 frequency, we intend to keep the most restrictive one

emersion:
It seems like this feature got lost in https://github.com/OpenRailAssociation/osrd/pull/9105.

I think this PR still creates `frequency::` labels in NGE (without using them in any trainrun)?

louisgreiner:
`frequency::` labels should be shared in NGE if they are not either `frequency::30`, or `frequency::60` or `frequency::120`.

The former feature allowed, for instance, `frequency::10` to be shared in NGE, even tho we don't match this frequency with an actual NGE frequency

@emersion

emersion:
Yeah, what I meant was that I think `frequency::30` and friends show up as a regular label in NGE with this PR.

emersion:
I'm not a huge fan of the RegExp-based label parsing. (BTW, it's also used in `ngeToOsrd`.)

Maybe this can all be consolidated in a single function, used everywhere instead of the RegExp and string checks?

```ts
const frequencyIdFromLabel = (label: string) => {
  if (!label.startsWith('frequency::')) return null;
  const n = parseInt(label.split('::', 2)[1], 10);
  const frequency = DEFAULT_TRAINRUN_FREQUENCIES.find((freq) => freq.frequency === n);
  return frequency?.id ?? null;
};
```

front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
emersion: This `if` matches labels such as `asdf::60`.

louisgreiner: this regex and the use of it was not great indeed, it has been changed

louisgreiner: I decided not to (miss-)use this regex anymore

louisgreiner:
Because I found this bug on the way, i'm also fixing it in a separate commit : https://github.com/OpenRailAssociation/osrd/issues/10552

SharglutDev (approved):
Lgtm and tested, good job :)

Uriel-Sautron (approved):
LGTM, tested.

Uriel-Sautron (approved):
LGTM, tested.

Issue #10401, opened 2025-01-16
Refacto: Improve CH code suggestions management.


SarahBellaha:
## Description
To improve the efficiency and accuracy of updating operational points and their suggestions, we need to refactor the current implementation. The following changes are proposed:

- Add `availableSecondaryCodes` in `StdcmPathStep` location.
- Update `availableSecondaryCodes` in `StdcmOperationalPoint` when changing the ci.
- Build `chSuggestions` from `availableSecondaryCodes`.
- Update `availableSecondaryCodes` when selecting an op in linked train:
  - This can be done either by using already available secondary codes or by fetching them with `searchOperationalPoints`.
- (Optional, if needed above): Refactor `useSearchOperationalPoint` hook:
  - Extract the state-setting logic from the `searchOperationalPoints` function.
  - Modify the function so that it returns its results and takes the query as an argument.

## Acceptance Criteria:

- [ ] Add `availableSecondaryCodes` field to `StdcmPathStep` location.
- [ ] Update `availableSecondaryCodes` in `StdcmOperationalPoint` when ci changes.
- [ ] Ensure `chSuggestions` are built from `availableSecondaryCodes`.
- [ ] Update `availableSecondaryCodes` when an op is selected in linked trains.
- [ ] (Optional) Refactor `useSearchOperationalPoint` hook to return results and accept a query argument.

PR #10451, opened 2025-01-20
front: fix stdcm empty ch list in new query

### LINKED ISSUE #10410: stdcm: when creating a request from an accepted simulation, waypoints have no CH Code
### What happened?

When clicking the aforementioned button, a new tab is opened (and at first I have a loader, don't mind it), and then the CH Code drop down menu is empty (I can open it but nothing's in it).

https://github.com/user-attachments/assets/8770c203-8098-4e2a-b93c-005808acb9e5

I expect this menu to always be filled when a CI code is selected.

1. Go to the stdcm page
2. Make a successful simulation
3. Accept it
4. Click on the "make another request from this one" button
5. Witness that the CH code drop down menu is empty

Recette (SNCF)

Firefox

9151b80

SharglutDev:
fix #10410

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
@@ -3,6 +3,7 @@ import { useEffect, useMemo, useState } from 'react';
 import { Select, ComboBox } from '@osrd-project/ui-core';
 import { useTranslation } from 'react-i18next';
 
+import type { SearchResultItemOperationalPoint } from 'common/api/osrdEditoastApi';
 import useSearchOperationalPoint from 'common/Map/Search/useSearchOperationalPoint';
 import { useOsrdConfActions } from 'common/osrdContext';
 import type { StdcmConfSliceActions } from 'reducers/osrdconf/stdcmConf';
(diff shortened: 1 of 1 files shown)


front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
```diff
@@ -23,6 +24,20 @@ function formatChCode(chCode: string) {
   return chCode === '' ? 'BV' : chCode;
 }
 
+const extractChCodes = (searchResults: SearchResultItemOperationalPoint[], selectedCI: CIOption) =>
+  searchResults
+    .filter((pr) => pr.name === selectedCI.name)
```
Akctarus:
```suggestion
    .filter((op) => op.name === selectedCI.name)
```
I think we should use "op" instead of "pr" that is a French acronym

Akctarus (approved):
tested, left a comment but does not block approval

front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
SharglutDev: Good catch !

clarani:
Related to https://github.com/OpenRailAssociation/osrd/issues/10401

kmer2016 (approved):
lgtm

PR #10471, opened 2025-01-21
Handle both invalid trigrams and uics as invalid op

### LINKED ISSUE #10116: UIC are not handled when not recognized in the infrastructure
### What happened?

Importing a train from a Viriato file uses UIC. If the code is not recognized, it should be handled by the functionnality "handled unrecognized waypoints" but it is not. We can see two problems : 

- infinite loading
- "undefined" written instead of the UIC code
Image

The train to import is the following :
[train_schedules (18).json](https://github.com/user-attachments/files/18168819/train_schedules.18.json)

I expect the same behavior as for unrecognized trigrams (error message + UIC code written in the list, but in red)

Import the train above

Dev (SNCF)

Google Chrome

fe3b339

achrafmohye:
Closes #10116

### DIFF
--- a/front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
+++ b/front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
@@ -10,6 +10,7 @@ import {
   type TrainScheduleResult,
   type PathfindingResult,
   type SearchResultItemOperationalPoint,
+  type PathfindingInputError,
 } from 'common/api/osrdEditoastApi';
 import { useOsrdConfActions } from 'common/osrdContext';
 import buildOpSearchQuery from 'modules/operationalPoint/helpers/buildOpSearchQuery';
@@ -74,8 +75,13 @@ const useSetupItineraryForTrainUpdate = (trainIdToEdit: number) => {
   const [postSearch] = osrdEditoastApi.endpoints.postSearch.useMutation();
 
   useEffect(() => {
-    const fetchPathStepsCoordinates = async (trainSchedule: TrainScheduleResult) => {
-      // get track sections
+    const fetchPathStepsCoordinates = async (
+      trainSchedule: TrainScheduleResult,
+      invalidPathItems: Extract<
+        PathfindingInputError,
+        { error_type: 'invalid_path_items' }
+      >['items'] = []
+    ) => {
       const trackSectionIds: string[] = [];
       trainSchedule.path.forEach((step) => {
         if ('track' in step) {
@@ -97,6 +103,8 @@ const useSetupItineraryForTrainUpdate = (trainIdToEdit: number) => {
       const pathStepsWithCoordinates = trainSchedule.path.map((step, index) => {
         let coordinates: Position | undefined;
         let name: string | undefined;
+        let uic: number | undefined;
+        let trigram: string | undefined;
 
         if ('track' in step) {
           const track = tracks[step.track];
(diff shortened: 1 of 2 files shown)


front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -74,7 +75,24 @@ const useSetupItineraryForTrainUpdate = (trainIdToEdit: number) => {
...
+    ) => {
+      const invalidIdentifiers = invalidPathItems
+        .map((item) => {
+          if ('trigram' in item.path_item) {
+            return item.path_item.trigram;
+          }
+          if ('uic' in item.path_item) {
+            return item.path_item.uic;
+          }
+          return null;
```
emersion: This doesn't indicate an invalid path step. A path step with an OP ID, or a track offset, is valid.

emersion: Oh, I misread the code. Let me correct myself: a OP ID or track offset may be invalid as well. Additionally, an UIC/trigram isn't enough to uniquely identify a path step. I'd suggest using `item.id` to identify invalid path steps, instead of trigram/UIC.

achrafmohye: what property for item identification you propose or use a combination of properties for each case ?

achrafmohye: @emersion there's a suggestion in an additional commit :)

emersion: What I mean is something like:

```ts
const isInvalid = invalidPathItems.find((item) => item.id === step.pathStepId)
```

and dropping `invalidIdentifiers` altogether.

achrafmohye: There's no direct identifier for the items hence my question, there is only an index and `PathItemLocation` which

Akctarus (approved):
Tested, ty for this fix

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
emersion: Seems like your comment got cut off?

Do you mean that `item.id` is always unset? Or do you mean that `step.pathStepId` is always unset? (Or something else?)

The former would surprise me. The latter less so, and should be fixed by https://github.com/OpenRailAssociation/osrd/pull/10006 - `matchPathStepAndOp()` can be used to avoid the issue.

emersion (commented):
Placeholder comment, just so that GitHub gets this PR off my TODO list :P 

See the discussion above.

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
achrafmohye: handled using the index of the `pathStep` exists in the `invalidPathItems`

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
```diff
@@ -107,19 +114,21 @@ const useSetupItineraryForTrainUpdate = (trainIdToEdit: number) => {
...
+            uic = step.uic;
           } else if ('trigram' in step) {
             op = ops.find((o) => o.trigram === step.trigram);
-          } else {
+          } else if ('operational_point' in step) {
             op = ops.find((o) => o.obj_id === step.operational_point);
           }
           coordinates = op?.geographic.coordinates;
-          name = `${op?.name}`;
+          name = `${op?.name || uic?.toString()}`;
```
emersion: Proposal: if it's a trigram, should we use it as a fallback?

```ts
`${op?.name || uic?.toString() || trigram}`
```

emersion (commented):
The new logic looks good!

It seems like the `usePathfinding()` hook has a similar bug: `handleInvalidPathItems()` there checks trigrams only. There is a TODO:

```ts
// TODO: we currently only handle invalid pathSteps with trigram. We will have to do it for trackOffset, opId and uic too.
```

Are you feeling up for adding a new commit to fix that bug and do something similar to the logic in this PR, or do you prefer to leave that for later?

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
emersion: Sweet!

front/src/applications/operationalStudies/hooks/useSetupItineraryForTrainUpdate.ts
achrafmohye: Done :)

emersion (approved):
Thanks, looks good!

(For next time: ideally the `usePathFinding` change should be in its own separate commit :P )

PR #10472, opened 2025-01-21
Fix unrecognized op and preserve invalid state

### LINKED ISSUE #10396: Unrecognized operational points error not displayed anymore
### What happened?

When importing a train with an unrecognized waypoint, I don't see the corresponding error 

Here is what happened 

1. I import  [this train](https://github.com/user-attachments/files/18427386/train_schedules.27.json)
, for which the 2nd waypoint is not recognized. If I enter the edition mode, select a rolling stock, and then go to the itinerary tab, I can see my error (red message + toto written in red). 

Image

2. I close and reopen the edition mode. Then if I go to the itinerary tab, I don't have the error (red message + toto in red) as expected. 

Image

Seems like the errors comes at the second opening of the train ?

Hint : Window reload + clear cache seems to reset the problem.

I expected to have always the unrecognized waypoint error

as above

Dev (SNCF)

Google Chrome

675d279

achrafmohye:
Closes #10396

### DIFF
--- a/front/src/modules/pathfinding/hooks/usePathfinding.ts
+++ b/front/src/modules/pathfinding/hooks/usePathfinding.ts
@@ -266,13 +266,13 @@ const usePathfinding = (
     if (isInitialized && infra?.state === 'CACHED') {
       launchPathfinding(pathSteps);
     }
-  }, [infra?.state]);
+  }, [infra?.state, isInitialized]);
 
   useEffect(() => {
     if (isInitialized) {
       launchPathfinding(pathSteps);
     }
-  }, [rollingStock]);
+  }, [rollingStock, isInitialized]);
 
   useEffect(() => {
     setIsInitialized(true);


RomainValls:
I tried to reproduce the bug, by importing the trainSchedule. When i modify this train, i can see in itinerary that the unrecognized waypoint is marked in red (correct behavior). When I modify this train, and go back to its itinerary tab, I only see an "undefined" with no red and no error message.

RomainValls (approved):
Tested, it works, thanks for the fix 👍

clarani:
Hum actually, this issue is more complex. It comes from the fact that the pathfinding is launched in setupItinerary, which does not handle the invalid items.
Adding this dependency in the useEffects will break the power restrictions, because the pathfinding will be launched twice: once in the hook useSetupItinerary and once in the usePathfinding hook (which will erase the power restrictions).

The best way to solve this bug would be to refacto the code so the pathfinding is launched only once when editing a train

achrafmohye:
@clarani do you manage to reporduce the bug ?

clarani:
The original bug must have been solved by https://github.com/OpenRailAssociation/osrd/pull/10696, it can't be reproduced now

Issue #9702, opened 2024-11-14
[meta] ETCS: Compute ETCS braking curves


Erashin:
### Description and goal

Compute ETCS braking curves. See https://github.com/osrd-project/osrd-confidential/issues/609 for more info.

### Acceptance criteria

- [x] #9703
- [x] #9704
- [x] #9705
- [x] #9708
- [x] #10330
- [x] #9706
- [x] #10495
- [x] #10404
- [x] #10991
- [x] #10443
- [x] #10430
- [x] #10446
- [x] #10296
- [ ] #10297 -> moved to pi 15
- [x] #10298 

To refine during next pi:
- no visa for etcs -> consider it an EOA with SVL at same target position
- conflict detection
  La SvL corresponde au point à protéger (si il n'y a) ou sinon au début d'un canton occupé. L'EOA est coincidente avec le SvL pour le cas d'un canton occupé ou à une distance fixe (200 m?) dans le cas d'un point à proteger
- add etcs curves to front: warning, ps, sbi1&2, gui, ebi, ebd, sbd (maybe improve reticule to clearly see breaking points between 2 consecutive curves)
- import etcs data in infra and import etcs rs
- check: calculer l'erreur odometrique = Q_LOCACC (précision dans la position de la balise - donnée d'infrastructure)+ 5m +5% de la distance parcourue de la dernière balise de localisation. Vérifier cette formule !! See paragraprah 3.6.4.1 of subset 026
- (bonus: modify rolling-stock etcs brake params in front)

Erashin:
All of the tasks have been finished except #11036, which was moved to #11443.

PR #10473, opened 2025-01-21
core: return all etcs curves in response

Erashin:
We'll be returning all the curves to the front end in the near future. This is just a preview draft where all of the intermediate curves are computed. See https://github.com/OpenRailAssociation/osrd/issues/9702


Erashin:
Done in https://github.com/OpenRailAssociation/osrd/pull/11864#event-18117156488

PR #10474, opened 2025-01-21
front: revert part of #10371

Khoyo:
Revert part of #10371 since it seems to make e2e tests flaky. We should investigate and follow up after.


emersion:
What is the motivation for this change? Ideally the commit message or GitHub PR description should explain.

Khoyo:
Indeed, this was opened fast to at least see if it has the desired effect :p

anisometropie (approved):

clarani (approved):
Code LGTM

PR #10475, opened 2025-01-21
front: fix flaky e2e test 008

### LINKED ISSUE #10377: Large timetable loading performance

Maymanaf:


emersion:
The issue might be `this.timetableTrains.count()`: when the list is virtualized, off-screen train HTML elements don't even exist so might not get counted? Hm, but that should only result in less trains being tested, not more tests failures?

Maymanaf:
> The issue might be `this.timetableTrains.count()`: when the list is virtualized, off-screen train HTML elements don't even exist so might not get counted? Hm, but that should only result in less trains being tested, not more tests failures?

The count works fine, the issue is with selecting the train card when it's not loaded in the DOM

clarani:
The commit has been reverted [here](https://github.com/OpenRailAssociation/osrd/commit/c1facc5e5329fd9fe85abb2e1085e01dcf2ae191), maybe we should re-add it ? @anisometropie

emersion:
Indeed, both commits from https://github.com/OpenRailAssociation/osrd/pull/10474 need to be re-applied to check whether the issue is indeed fixed. (Since it was flaky, need to test multiple times to ensure it doesn't succeed by chance.)

front/tests/pages/op-timetable-page-model.ts
```diff
@@ -94,7 +94,7 @@ class OperationalStudiesTimetablePage extends CommonPage {
 
   // Verify that simulation results are displayed
   async verifySimulationResultsVisibility(): Promise<void> {
-    await this.page.waitForLoadState('networkidle');
+    await this.page.waitForLoadState('networkidle', { timeout: 60_000 });
```
emersion: 60s is a very long time, do we really need it? I don't think https://github.com/OpenRailAssociation/osrd/pull/10371 increases the simulation time.

front/tests/pages/op-timetable-page-model.ts
```diff
@@ -200,12 +200,13 @@ class OperationalStudiesTimetablePage extends CommonPage {
...
     for (let currentTrainIndex = 0; currentTrainIndex < trainCount; currentTrainIndex += 1) {
-      await this.page.waitForLoadState('networkidle');
-      await this.waitForSimulationResults();
-      const trainButton = OperationalStudiesTimetablePage.getTrainButton(
-        this.timetableTrains.nth(currentTrainIndex)
-      );
+      const trainElementHandle = this.timetableTrains.nth(currentTrainIndex);
+
+      await trainElementHandle.scrollIntoViewIfNeeded();
+      await trainElementHandle.waitFor();
```
emersion: Does `waitFor()` really do anything here? I'd assume `scrollIntoViewIfNeeded()` to be enough to make the element visible.

front/tests/pages/op-timetable-page-model.ts
```diff
@@ -200,12 +200,13 @@ class OperationalStudiesTimetablePage extends CommonPage {
     const trainCount = await this.timetableTrains.count();
 
     for (let currentTrainIndex = 0; currentTrainIndex < trainCount; currentTrainIndex += 1) {
-      await this.page.waitForLoadState('networkidle');
-      await this.waitForSimulationResults();
-      const trainButton = OperationalStudiesTimetablePage.getTrainButton(
-        this.timetableTrains.nth(currentTrainIndex)
-      );
+      const trainElementHandle = this.timetableTrains.nth(currentTrainIndex);
```
emersion: Kind of orthogonal to this PR, but it's nicer to use `for (const trainElement of await this.timetableTrains.all())` instead of a counter.

anisometropie:
pushed back in my commits for #10377, E2e are running ok

Maymanaf:
Fixed in this [PR ](https://github.com/OpenRailAssociation/osrd/pull/10423)

PR #10476, opened 2025-01-21
front: fix margins reset after micro macro switch

### LINKED ISSUE #10428: nge erases margins (and maybe other attributes)
### What happened?

I created a train in the micro mode with a 5% margin on the first OP.

Then I update this train in macro mode (change its name for example)

Then my train has a 0% margin instead of 5%

I expected to keep my margin

Maybe while solving this issue, one could check the other attributes and check that the macro mode does not erase other attributes of the train schedule?

as above

Recette (SNCF)

Google Chrome

c9066aa

RomainValls:
close #10428 

Margins were reset as they contained path items references.
I haven't noticed any bug since putting them back, but I'm not sure I tried every edge cases.

### DIFF
--- a/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
@@ -146,8 +146,7 @@ const createPathItemFromNode = async (
   node: NodeDto,
   index: number,
   infraId: number,
-  dispatch: AppDispatch,
-  sectionToPathItemId: Map<number, string>
+  dispatch: AppDispatch
 ) => {
   const [trigram, secondaryCode] = node.betriebspunktName.split('/');
   let finalSecondaryCode: string | undefined;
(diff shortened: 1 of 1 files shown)


Uriel-Sautron (approved):
LGTM tested.

emersion (changes requested):
Doesn't work for me with this [exported train](https://github.com/user-attachments/files/18520160/train_schedules.6.json). What I did:

- Create a train from OSRD with 4 steps, add different margins at each step
- Switch to NGE, edit train name

The backend replies with a 422:

```
Failed to deserialize the JSON body into the target type: Invalid boundary, path waypoint 'id171' not found at line 1 column 606
```

emersion:
In general I think it's tricky to make this work in all cases. Maybe with a mapping of train run section IDs → path item IDs, and checking whether the prev/next path item IDs haven't changed for each margin span?

RomainValls:
Thanks for the tip. I tried implementing it, i just need to add the correct typing to fix the warnings. Wdyt about this ? @emersion

PR #10478, opened 2025-01-21
front: remove webkit from e2e tests

Maymanaf:
Remove WebKit from e2e tests as it slows down the test execution


Yohh (approved):

SharglutDev (approved):
Lgtm and tested !

PR #10483, opened 2025-01-22
editoast: fix cropped work schedule times

### LINKED ISSUE #9793: Work schedules start and end time seem cropped by the stdcm research window
### What happened?

Work schedules are announced between 1 and 5am here:
Image

While they actually span from 11pm to ~6am according to the official database.

I'm guessing what's happening here is that the timings are cropped by the stdcm research time window because the departure time is 2am +/- 1h, so the time window starts at 1am:
Image

Image
_sorry about the french text, didn't take time to change that_

I'm expecting the entirety of the work schedule to be announced.

It's kinda hard to reproduce cause we need the official database to compare what's displayed and the reality. I suppose it's easier to test by creating work schedules manually.

Other

Google Chrome

v0.3.4 cd4cdfb

hamz2a:
Fixes https://github.com/OpenRailAssociation/osrd/issues/9793

### DIFF
--- a/editoast/src/views/timetable/stdcm/failure_handler.rs
+++ b/editoast/src/views/timetable/stdcm/failure_handler.rs
@@ -51,12 +51,13 @@ impl SimulationFailureHandler {
             latest_simulation_end,
             ..
         } = self;
+        let start_time = work_schedules
+            .first()
+            .map(|ws| ws.start_date_time)
+            .unwrap_or(earliest_departure_time);
         let virtual_train_id = train_schedule.id;
-        let work_schedules = WorkSchedulesRequest::new(
-            work_schedules,
-            earliest_departure_time,
-            latest_simulation_end,
-        );
+        let work_schedules =
+            WorkSchedulesRequest::new(work_schedules, start_time, latest_simulation_end);
 
         // Combine the original train schedules with the virtual train schedule.
         train_schedules.push(train_schedule);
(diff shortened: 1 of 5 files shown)


editoast/src/views/timetable/stdcm/failure_handler.rs
```diff
@@ -51,12 +51,13 @@ impl SimulationFailureHandler {
             latest_simulation_end,
             ..
         } = self;
+        let start_time = work_schedules
+            .first()
```
woshilapin, line 55: Does this work (taking the first), because you assume that the work schedules are ordered by `start_date_time`? The way I see it is that there is a functionality here that rely on something done in a completely different place of the code. So not so much locality for the behavior, which means we might break that anytime in the future if we decide to remove the `.order_by()` or change it for another field.

That said, I'm not sure that I have a proposition for this? At the very least, we could document the API `get_work_schedules()` which is expected to return a sorted list of work schedules? This way, any future modification would need to take that into account.

What do you think? Any idea?

woshilapin (approved):
I'm approving since I'm not sure there is an easy solution to the small remark I made. I won't block the merge if we can't find an easy solution (I believe comment would be an ok solution).

editoast/src/views/timetable/stdcm/failure_handler.rs
hamz2a: I used the existing `order_by` API because I wanted to avoid redoing the sorting in Rust. But I agree, there’s a risk if the order changes in the future. Documenting the expected sorting behavior for `get_work_schedules()` sounds like a good idea to make this clear.

hamz2a: I’ve added a comment to document the expected sorting behavior for `get_work_schedules()`. Let me know what you think! @woshilapin

shenriotpro (approved):
LGTM for tests/

PR #10487, opened 2025-01-22
Change map tiles server and use pmtiles instead of mbtiles

Tristramg:
Use pmtiles provided by https://tuiles.codeureusesenliberte.fr/

There should be no visible change.

This only concerns the base layer, underneath the infrastructure handled by OSRD.

To make sure everything works fine, try the different styles (including terrain). Make sure that we see everything, look in the console that all the data comes either from osrd, either from https://osm.nbg1.your-objectstorage.com/ (the url of object storage by hetzner).

Make sure that the map loading feels at least as fast as before.

image

### DIFF
--- a/front/src/common/Map/Layers/Hillshade.tsx
+++ b/front/src/common/Map/Layers/Hillshade.tsx
@@ -1,6 +1,7 @@
 import { Source, type LayerProps } from 'react-map-gl/maplibre';
 import { useSelector } from 'react-redux';
 
+import { TERRAIN_URL } from 'common/Map/const';
 import OrderedLayer from 'common/Map/Layers/OrderedLayer';
 import { getTerrain3DExaggeration } from 'reducers/map/selectors';
 
@@ -29,7 +30,7 @@ const Hillshade = ({ mapStyle, layerOrder }: HillshadeProps) => {
       id="hillshade"
       type="raster-dem"
       encoding="terrarium"
-      url="https://osm.osrd.fr/data/terrain.json"
+      url={TERRAIN_URL}
       tileSize={256}
       maxzoom={12}
     >
--- a/front/src/common/Map/const.ts
+++ b/front/src/common/Map/const.ts
@@ -4,7 +4,9 @@ import config from 'config/config';
 export const MAP_URL = `${config.proxy_editoast}/layers`;
 export const SPRITES_URL = `${config.proxy_editoast}/sprites`;
 export const FONTS_URL = `${config.proxy_editoast}/fonts`;
-export const OSM_URL = 'https://osm.osrd.fr/data/v3.json';
+export const OSM_URL = 'pmtiles://https://osm.nbg1.your-objectstorage.com/planet.pmtiles';
+export const TERRAIN_URL =
+  'pmtiles://https://osm.nbg1.your-objectstorage.com/terrain-europe.pmtiles';
 
 export const MAP_MODES = {
   display: 'display',
--- a/front/package-lock.json
+++ b/front/package-lock.json
@@ -72,6 +72,7 @@
         "maplibre-gl": "^4.0.0",
         "openapi-typescript-codegen": "^0.29.0",
         "party-js": "^2.2.0",
+        "pmtiles": "^4.2.1",
         "prop-types": "^15.8.1",
         "rc-slider": "^11.1.8",
         "react": "^18.2.0",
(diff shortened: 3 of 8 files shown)


flomonster:
I ran some tests on "Gare Montparnasse".

2025-03-06-102342_2555x1264_scrot
2025-03-06-102406_2560x1266_scrot

- Left side is using your tiles, right side is using SNCF Réseau tiles.

With your version, we can still see `© SNCF Réseau` in the bottom right of the map.

front/src/common/Map/const.ts
```diff
@@ -4,7 +4,9 @@ import config from 'config/config';
 export const MAP_URL = `${config.proxy_editoast}/layers`;
 export const SPRITES_URL = `${config.proxy_editoast}/sprites`;
 export const FONTS_URL = `${config.proxy_editoast}/fonts`;
-export const OSM_URL = 'https://osm.osrd.fr/data/v3.json';
```
Synar, line 7: front/src/common/Map/Layers/Background.tsx containing platform backgounds still uses https://osm.osrd.fr/data/v3.json, should it be updated to use OSM_URL too?

If so, should we drop osm.osrd.fr? (I assume this last point is a @devops concern)

Synar: (In fact the Background layer component and the Platform layer component should probably be a single component with one source and two layers given their similarities, so I assume they should indeed keep using the same url)

front/src/common/Map/Layers/Hillshade.tsx
```diff
@@ -29,7 +30,7 @@ const Hillshade = ({ mapStyle, layerOrder }: HillshadeProps) => {
       id="hillshade"
       type="raster-dem"
       encoding="terrarium"
-      url="https://osm.osrd.fr/data/terrain.json"
+      url={TERRAIN_URL}
       tileSize={256}
       maxzoom={12}
```
Synar, line 35: Can't test, but we may need an 
'attribution="..."' in the source component props to respect the open street map attribution clause of https://tuiles.codeureusesenliberte.fr/

Edit: tested and adding that attribute appears to work ^^

Synar (commented):
Can't test right now, but overall lgtm ^^

Synar:
By the way, could you prepend your branch names with your trigram/identifier? Something like "tmg/test_pmtiles"

flomonster:
> By the way, could you prepend your branch names with your trigram/identifier? Something like "tmg/test_pmtiles"

No need to rename the branch and create a new PR for that. There's no requirement for branch names. Several people in the project (including me) use an identifier to prefix the branch name so that they can filter easily.

Synar:
> > By the way, could you prepend your branch names with your trigram/identifier? Something like "tmg/test_pmtiles"
> 
> No need to rename the branch and create a new PR for that. There's no requirement for branch names. Several people in the project (including me) use an identifier to prefix the branch name so that they can filter easily.

Oh I did not mean for this branch specifically of course, but more in general, it would be nicer to be able to quickly filter (I won't name and blame who fed me misinformation but I swear I was told it was a requirement though ''^^)

Tristramg:
Thank you for the branch names, in the future I’ll try to remember to follow the convention

front/src/common/Map/Layers/Hillshade.tsx
Tristramg: For terrain specifically, there is no need for attribution (the data is base on Copernicus : https://www.copernicus.eu/en/access-data/copyright-and-licences)

But I added it for OpenStreetMap data where the attribution was indeed missing

front/src/common/Map/const.ts
Tristramg: As far as I understood, this seems to be a bit a workaround of the limitations of the React wrapper around maplibre.
The background layer just sets the default background color and has no data, so it’s a bit weird to set a source.

The background is always present, the platforms not. So we shouldn’t merge them.

I added this commit to simplify a bit the backgound layer: https://github.com/OpenRailAssociation/osrd/pull/10487/commits/e6708c9a4bca10c773ba302e59d3efb87bf75350

Synar: Thanks! I assumed from its previous source it was setting platforms background, the new code is much cleaner ^^

front/src/common/Map/Layers/Hillshade.tsx
Synar: Thanks! (I thought they were both osm data.)

Synar (approved):
Code LGTM ^^

emersion (approved):

flomonster (approved):
LGTM

Issue #10412, opened 2025-01-16
STDCM: simulation failed for a specific path


Maymanaf:
### What happened?

```
{
    "status": 500,
    "type": "core:unknown_error",
    "context": {
        "message": "Failed to compute a simulation that wouldn't cause conflicts: mismatch between exploration and postprocessing (please open a bug report)",
        "stack_trace": [
            "PostProcessingSimulation.kt:472",
            "PostProcessingSimulation.kt:104",
            "PostProcessingSimulation.kt:457",
            "PostProcessingSimulation.kt:104",
            "PostProcessingSimulation.kt:51",
            "STDCMPostProcessing.kt:81",
            "STDCMPathfinding.kt:134",
            "STDCMPathfinding.kt:77",
            "STDCMEndpointV2.kt:110",
            "STDCMEndpointV2.kt:79",
            "WorkerCommand.kt:249",
            "WorkerCommand.kt:303",
            "ContextPropagatingRunnable.java:37",
            "ThreadPoolExecutor.java:1144",
            "ThreadPoolExecutor.java:642",
            "Thread.java:1583"
        ],
        "exception_type": "class java.lang.RuntimeException",
        "url": "/v2/stdcm"
    },
    "message": "An unknown exception was thrown"
}
```
![Image](https://github.com/user-attachments/assets/96b06aed-6b40-4663-9189-b99da99b5359)

### What did you expect to happen?

No error displayed 

### How can we reproduce it (as minimally and precisely as possible)?

Launch a simulation with this Payload: 
```json
{"comfort":"STANDARD","margin":"4.5min/100km","rolling_stock_id":6,"towed_rolling_stock_id":8,"speed_limit_tags":"MA100","total_mass":89000,"max_speed":27.77777777777778,"total_length":17,"steps":[{"duration":0,"location":{"trigram":"VCE","secondary_code":"BV"}},{"duration":900000,"location":{"trigram":"LPT","secondary_code":"BV"}},{"duration":0,"location":{"trigram":"MAS","secondary_code":"BV"},"timing_data":{"arrival_time":"2025-01-17T16:32:00.000Z","arrival_time_tolerance_before":4500000,"arrival_time_tolerance_after":6300000}}],"time_gap_after":15000,"time_gap_before":15000,"work_schedule_group_id":264,"loading_gauge_type":"GA"}
```

Also happened with this payload:
```json
[...]

PR #10488, opened 2025-01-22
core: stdcm: add fixed time points at start+end of engineering allowances

eckter:
Fix what I believe to be the last cases of https://github.com/OpenRailAssociation/osrd/issues/10412

Though while investigating I found this, when printing the engineering allowances on the final path: 

```
...
margin from 234,337m to 444,837m
margin from 234,337m to 446,622m
margin from 116,065m to 449,152m
margin from 234,337m to 451,259m
...
```

Meaning: a margin went further back than the previous margin. It doesn't seem to cause any significant issue yet, but it feels very wrong and worth looking into.


Erashin (approved):
Distance bug is probably caused by a length computation bug, which was never really apparent until now. As discussed, it's fine to merge this rn, and you'll look into it later on.

PR #10447, opened 2025-01-20
front: LMR: fix length, mass and speed reset

Akctarus:
closes #10106 and fix #10270

should respect the expectations of this: https://github.com/osrd-project/osrd-confidential/issues/903

PR #10490, opened 2025-01-22
front: lmr reset prefilled values changed by the user

### LINKED ISSUE #10270: front: changing a convoy value prevents it from being autofilled for all future rolling stock change
### What happened?

I set a value for convoy max speed. I then changed rolling stock, and the value remained. There wasn't any way to know the actual expected rolling stock max speed, or to reset it in any way. The only solution was refreshing the page and losing everything.

I expected all values to be updated when changing rolling stock.

At least, I thought that deleting the value entirely would reset it.

Set a max speed value and change rolling stock. 

Local, Other

Firefox

ef6e28ed5cf068bd874ed7bef9e266b5a8e2e78f

Wadjetz:
fix [#10270](https://github.com/OpenRailAssociation/osrd/issues/10270)
When the user modifies or enters a value, it disables the prefill for mass, length, and max speed. This PR re-enables the prefill if the user removes the value.

### DIFF
--- a/front/src/applications/stdcm/hooks/useStdcmConsist.ts
+++ b/front/src/applications/stdcm/hooks/useStdcmConsist.ts
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 
 import { min } from 'lodash';
 import { useSelector } from 'react-redux';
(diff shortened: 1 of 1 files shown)


SharglutDev:
I think this will conflict with this PR : https://github.com/OpenRailAssociation/osrd/pull/10447 @Akctarus

Wadjetz:
Close this PR it already fixed by https://github.com/OpenRailAssociation/osrd/pull/10447

PR #10497, opened 2025-01-23
Fix incorrect input saving when switching simulations during a running calculation

### LINKED ISSUE #10480: Wrong inputs saved for the simulation
### What happened?

https://github.com/user-attachments/assets/dbfb038d-d319-4fc4-b92a-25eebc5a2481

If I launch a simulation (simulation 4) and during the loading I switch to another existing one (simulation 2) and the simulation load finishes while I'm on this other simulation (n2), he inputs saved for the n4 will be the one from n2.

The saved inputs should be the ones from the simulation I launched, regardless of the simulation I'm visitng.

See description

Local

Firefox

.

kmer2016:
closes #10480 
When a different simulation is selected while a calculation is still running, the new simulation’s inputs overwrite the inputs of the ongoing calculation. This causes a mismatch between the inputs used for the calculation and the results.

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
@@ -5,11 +5,17 @@ import cx from 'classnames';
 import { useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
 
+import useStdcmForm from 'applications/stdcm/hooks/useStdcmForm';
 import { extractMarkersInfo } from 'applications/stdcm/utils';
 import DefaultBaseMap from 'common/Map/DefaultBaseMap';
 import { useOsrdConfSelectors } from 'common/osrdContext';
 import useInfraStatus from 'modules/pathfinding/hooks/useInfraStatus';
-import { resetMargins, restoreStdcmConfig, updateStdcmPathStep } from 'reducers/osrdconf/stdcmConf';
+import {
+  resetMargins,
+  restoreStdcmConfig,
+  updateStdcmPathStep,
+  addStdcmSimulation,
+} from 'reducers/osrdconf/stdcmConf';
 import {
   getStdcmDestination,
   getStdcmOrigin,
@@ -44,19 +50,23 @@ declare global {
 type StdcmConfigProps = {
   isDebugMode: boolean;
   isPending: boolean;
-  launchStdcmRequest: () => Promise<void>;
   retainedSimulationIndex?: number;
   showBtnToLaunchSimulation: boolean;
+  skipPathfindingStatusMessage: boolean;
+  launchStdcmRequest: () => Promise<void>;
   cancelStdcmRequest: () => void;
+  setSkipPathfindingStatusMessage: (value: boolean) => void;
 };
 
 const StdcmConfig = ({
   isDebugMode,
   isPending,
-  launchStdcmRequest,
   retainedSimulationIndex,
   showBtnToLaunchSimulation,
+  skipPathfindingStatusMessage,
+  setSkipPathfindingStatusMessage,
   cancelStdcmRequest,
+  launchStdcmRequest,
 }: StdcmConfigProps) => {
   const { t } = useTranslation('stdcm');
   const launchButtonRef = useRef<HTMLDivElement>(null);
@@ -81,13 +91,16 @@ const StdcmConfig = ({
 
   const [formErrors, setFormErrors] = useState<StdcmConfigErrors>();
 
+  const currentSimulationInputs = useStdcmForm();
+
   const disabled = isPending || retainedSimulationIndex !== undefined;
 
   const markersInfo = useMemo(() => extractMarkersInfo(pathSteps), [pathSteps]);
 
-  const startSimulation = () => {
+  const startSimulation = async () => {
     const formErrorsStatus = checkStdcmConfigErrors(pathSteps, t, pathfinding?.status);
     if (pathfinding?.status === 'success' && !formErrorsStatus) {
+      dispatch(addStdcmSimulation(currentSimulationInputs));
       launchStdcmRequest();
     } else {
       // The console error is only for debugging the user tests (temporary)
@@ -101,11 +114,14 @@ const StdcmConfig = ({
       updateStdcmPathStep({ id: origin.id, updates: { arrivalType: ArrivalTimeTypes.ASAP } })
     );
   };
+
   const removeDestinationArrivalTime = () => {
     dispatch(
       updateStdcmPathStep({ id: destination.id, updates: { arrivalType: ArrivalTimeTypes.ASAP } })
     );
   };
+
+  const onItineraryChange = () => setSkipPathfindingStatusMessage(false);
 
   const getStatusMessage = () => {
     if (isPathFindingLoading) {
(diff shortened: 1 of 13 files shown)


front/src/applications/stdcm/views/StdcmView.tsx
```diff
@@ -19,13 +19,16 @@ import StdcmResults from '../components/StdcmResults';
...
+import type { StdcmSimulation, StdcmSimulationInputs } from '../types';
 
 const StdcmView = () => {
   // TODO : refacto. state useStdcm. Maybe we can merge some state together in order to reduce the number of refresh
   const currentSimulationInputs = useStdcmForm();
   const stdcmConf = useSelector(getStdcmConf);
   const [simulationsList, setSimulationsList] = useState<StdcmSimulation[]>([]);
+  const [inProgressSimulationInputs, setInProgressSimulationInputs] = useState<
+    StdcmSimulationInputs | undefined
+  >(undefined);
```
SharglutDev:
```suggestion
  const [inProgressSimulationInputs, setInProgressSimulationInputs] = useState<
    StdcmSimulationInputs
  >();
```

front/src/applications/stdcm/views/StdcmView.tsx
SharglutDev, line 1: If I switch between simulations, it seems like the speed limit tag isn't reset

SharglutDev, line 1: If I switch between a simulation with a short form and a long one, the pathfinding loader/success message his hidden between the simulation navigator (probably because it's in position absolute and it needs to stay that way)
https://github.com/user-attachments/assets/af10f021-eced-4bf1-89c6-b15973bf62d7

Maybe you should talk with @thibautsailly for the expected behavior : 
- Remove the banners if we switch between simulation
- Display it somewhere else
- other option

SharglutDev, line 1: If I switch between a simulation with a short form and a long one, the viewport changes. I think it should stay as it is when we switch (you can double check with @thibautsailly). Also maybe we should remove the animation for the via in that case, it's a little weird.
https://github.com/user-attachments/assets/0fbadb77-d587-40b3-bd5b-58f828ffeae5

SharglutDev (commented):
Thank you for the fix, it works great !

But now that it works, it reveals some strange behaviors that we should correct I think.

RomainValls (approved):
Code LGTM, tested and it works well. ✅

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -107,7 +107,10 @@
         }
 
         .stdcm-vias-bundle {
-          animation: bouncin-in 0.75s cubic-bezier(0.567, -0.475, 0, 1);
+          
+          &.animated {
```
SharglutDev, line 110: With this behavior, we have the animation after switching simulation and clearing a via input

https://github.com/user-attachments/assets/7e84627b-49ee-4602-b15e-74a79efe74de

front/src/applications/stdcm/views/StdcmView.tsx
SharglutDev: The last part of the comment is fixed but the first part 

> I think it should stay as it is when we switch

is still there. What did @thibautsailly said about that ?

SharglutDev: Seems like this bug is still there ?

kmer2016: confirmed by @thibautsailly

kmer2016: I can't reproduce the bug
[Capture vidéo du 10-02-2025 18:13:25.webm](https://github.com/user-attachments/assets/2c2d6a8a-b367-4768-ab86-24e1e6afbc03)

SharglutDev: Me neither :)

front/src/reducers/osrdconf/types.ts
SharglutDev, line 1: There is still a wrong behavior.
In the video, I start a simulation for Paris > Mâcon > Lyon. While simulation in progress, I switch on another existing simulation with Brest > ... > Strasbourg. When the loader finishes, I'm switched on the last simulation but the inputs aren't Paris > ... > Mâcon but the ones from Brest.

https://github.com/user-attachments/assets/d52b9570-c5b8-4fc0-9666-57068af57f5c

front/src/applications/stdcm/views/StdcmView.tsx
```diff
@@ -177,6 +188,20 @@ const StdcmView = () => {
     }
   }, [isRejected]);
 
+  /*
+   * After the new content is rendered, this effect adjusts the scroll to compensate for any shift in the stdcmResult section.
+   * It compares the stdcmResult section position before the change (stored in previousResultSectionOffsetRef)
+   * with its current position after rendering. The calculated difference is used to perform a window.scrollBy,
+   *  ensuring that the section remains at the same relative position in the viewport, thus avoiding any visual jump.
+   */
+  useEffect(() => {
```
SharglutDev: I think that for DOM behaviors like that, it's better to use a [useLayoutEffect](https://fr.react.dev/reference/react/useLayoutEffect)

front/src/reducers/osrdconf/stdcmConf/selectors.ts
```diff
@@ -32,6 +32,10 @@ const buildStdcmConfSelectors = () => {
     }
   );
 
+  const getStdcmCompletedSimulations = createSelector([getStdcmSimulations], (simulations) =>
+    simulations.filter((simulation) => simulation.outputs !== undefined)
```
SharglutDev: Wouldn't it work like that ? 

```suggestion
    simulations.filter((simulation) => simulation.outputs)
```

front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -232,10 +246,15 @@ export const stdcmConfSlice = createSlice({
       );
       state.stdcmPathSteps = newPathSteps;
     },
-    // temporary solution to add a new simulation, only 1 action should remain after the refactoring
-    addNewStdcmResult(state: Draft<OsrdStdcmConfState>, action: PayloadAction<StdcmSimulation>) {
-      state.simulations.push(action.payload);
-      state.selectedSimulationIndex = state.simulations.length - 1;
+    createStdcmSimulation(
```
SharglutDev: Maybe it's more relevant ?
```suggestion
    addStdcmSimulation(
```

front/src/reducers/osrdconf/stdcmConf/stdcmConfReducers.spec.ts
```diff
@@ -201,58 +222,74 @@ describe('stdcmConfReducers', () => {
...
       creationDate: new Date(),
-      inputs: {
-        pathSteps: stdcmPathSteps,
-        linkedTrains: { anteriorTrain: undefined, posteriorTrain: undefined },
-      },
+      inputs: simulationInputs,
     };
 
-    it('should handle adding new simulations', () => {
+    it('should create new simulation', () => {
```
SharglutDev: If you made the other change
```suggestion
    it('should add a new simulation', () => {
```

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -126,6 +137,7 @@ const StdcmConfig = ({
     const formErrorsStatus = checkStdcmConfigErrors(pathSteps, t, pathfinding?.status);
     if (pathfinding?.status === 'success' && !formErrorsStatus) {
       launchStdcmRequest();
+      dispatch(createStdcmSimulation(currentSimulationInputs));
```
SharglutDev: Is there any (bad)chance where we could have a sync issue if inputs were updated before/after the same simulation outputs ?

SharglutDev (approved):
Lgtm and tested, great job !

PR #10499, opened 2025-01-23
front: fix path step not updated in debug mode

### LINKED ISSUE #10337: front: !debug mode only! stdcm path step are not updated when modifying date input

kmer2016:
closes #10337

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
@@ -248,13 +248,13 @@ const StdcmConfig = ({
                 <Button
                   data-testid="launch-simulation-button"
                   className={cx({
-                    'fade-out': !showBtnToLaunchSimulation,
+                    'fade-out': !isDebugMode && !showBtnToLaunchSimulation,
                   })}
                   label={t('simulation.getSimulation')}
                   onClick={startSimulation}
                   isDisabled={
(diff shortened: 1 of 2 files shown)


emersion:
Hm, that sounds a bit weird… So `currentSimulationInputs` is equal to the previous one but something changed?

Are we missing something from the check? Should we compare other values?

kmer2016:
> Hm, that sounds a bit weird… So `currentSimulationInputs` is equal to the previous one but something changed?
> 
> Are we missing something from the check? Should we compare other values?

Yes, it may seem strange, but it's intended for debug mode. We want users to be able to rerun the simulation with the same inputs to simplify troubleshooting. This doesn’t affect non-debug mode, which checks inputs to avoid unnecessary reruns.

Math-R (approved):
lgtm

PR #10502, opened 2025-01-23
ci, dev: remove frontend nginx version, dev stack with bundled front …

ghost:
The goal of this PR is to simplify the stack by:
- Removing the nginx front image
- Using the gateway front image by default
- Removing various duplications between dockerfiles
- Adding a script to manage all the docker compose stuff

Due to a limitation in Docker Compose not supporting targeting custom targets inside other Dockerfile in additional contexts, I've moved a part of the building system for the front on the Gateway Dockerfile. Not ideal, but does the trick.

NOTE : Before merging we should modify the protrected branch requirement


flomonster (approved):
LGTM

Maybe we should delete the `scripts/host-compose.sh` file?

docker/docker-compose.front.yml
```diff
@@ -0,0 +1,30 @@
+services:
+
```
Khoyo: spurious white line

osrd-compose
```diff
@@ -0,0 +1,107 @@
+#!/bin/bash
```
Khoyo:
```suggestion
#!/usr/bin/env bash
```

osrd-compose
```diff
@@ -0,0 +1,107 @@
...
+
+STATE_FILE=".osrd-compose-state"
+BASE_COMPOSE="docker-compose.yml"
+OVERRIDE_DIR="./docker"
```
Khoyo: I think most of our scripts try to find the correct base dir based on $0 instead of working from cwd

osrd-compose
```diff
@@ -0,0 +1,107 @@
...
+
+This script simplifies docker compose operations by managing override files.
+
+Usage: ./osrd-compose [FLAGS] [DOCKER_COMPOSE_COMMANDS]
+
+Flags (optional, can be combined):
+  front    Add front-end development container (docker/docker-compose.front.yml)
+  sw       Enable single worker mode (docker/docker-compose.single-worker.yml)
+  host     Use host networking mode (docker/docker-compose.host.yml)
+  default  Reset to base configuration only and clear saved state
```
Khoyo, line 20: Shouldn't those be options?

osrd-compose
```diff
@@ -0,0 +1,107 @@
...
+
+# Help
+show_help() {
+    cat << EOF
+OSRD Docker Compose Wrapper
+==========================
+
+This script simplifies docker compose operations by managing override files.
+
+Usage: ./osrd-compose [FLAGS] [DOCKER_COMPOSE_COMMANDS]
```
Khoyo: nit: `./osrd-compose` should be `$0` if we go toward the "not assuming cwd is correct" route

osrd-compose
```diff
@@ -0,0 +1,107 @@
...
+BASE_COMPOSE="docker-compose.yml"
+OVERRIDE_DIR="./docker"
+
+# Help
+show_help() {
+    cat << EOF
+OSRD Docker Compose Wrapper
+==========================
+
+This script simplifies docker compose operations by managing override files.
```
Khoyo: nit: to save a bit of screen size with long usage messages, drop the title, and move the description below the usage line

osrd-compose
```diff
@@ -0,0 +1,107 @@
...
+flags=()
+
+# Parse command line arguments for flags first
+while [[ $# -gt 0 ]]; do
+    case $1 in
+        default)
+            # Remove state file and use only base compose
+            rm -f "$STATE_FILE"
+            shift
+            break
```
Khoyo: This should probably set flags to empty/error if they are any (even still left?)

osrd-compose
```diff
@@ -0,0 +1,107 @@
...
+cmd="docker compose"
+for file in "${compose_files[@]}"; do
+    cmd+=" -f $file"
+done
+
+# Add remaining arguments
+cmd+=" $*"
+
+# Execute the command
+echo "Executing: $cmd"
```
Khoyo:
```suggestion
echo "$cmd"
```
nit: similar to bash behavior with eg. `!!`

osrd-compose
```diff
@@ -0,0 +1,111 @@
...
+            # Remove state file and use only base compose
+            rm -f "$STATE_FILE"
+            DEFAULT_FLAG=true
+            shift
+            ;;
+        front|sw|host)
+            if [ -n "$DEFAULT_FLAG" ]; then
+                echo "Error: Cannot use flags with 'default' flag"
+                exit 1
+            fi
```
Khoyo: You should check the flags after the loop, this won't catch `front default`

osrd-compose
```diff
@@ -0,0 +1,115 @@
...
+State Management:
+- Previous flags are saved to .osrd-compose-state
+- Without flags, last known state is loaded
+- Without flags and no state file, only base compose is used
+- 'default' flag clears saved state
+EOF
+}
+
+# Display help if requested or no args given
+if [ "$1" = "script-help" ] || [ $# -eq 0 ]; then
```
Khoyo: nit: we should support `-h` and `--help`, because when I don't remember how to use it that's definitely what I'm gonna try first

ElysaSrc: It will conflict with no flag + checking the help of `docker compose`

ghost:
Various updates : the order of the overrides matters more than I thought, so I switched the way I pars env flags.

Khoyo (approved):

osrd-compose
bougue-pe, line 1: Why not naming it `osrd-compose.sh`?

osrd-compose
```diff
@@ -0,0 +1,127 @@
...
+show_help() {
+    cat << EOF
+
+Usage: $0 [FLAGS] [DOCKER_COMPOSE_COMMANDS]
+
+This script simplifies docker compose operations by managing override files.
+
+Flags (optional, can be combined):
+  front    Add front-end development container (docker/docker-compose.front.yml)
+  sw       Enable single worker mode (docker/docker-compose.single-worker.yml)
```
bougue-pe, line 18: Then we can remove `scripts/single-worker-compose.sh` too 🙂

bougue-pe (approved):
Thanks! ❤️

Only tested and reviewed the script, and it suits my (core debug) need at least as well as before:
`./osrd-compose host up -d --build  osrdyne osrd-images jaeger gateway postgres valkey rabbitmq && cd editoast && diesel migration run` (then launching aside `editoast` no-cache & single-worker + `core` single-worker)
Improving a "without-back" dev-ex (no-cache, single-worker, exclude service instead of include) is another topic.

osrd-compose
ghost: For the sake of brevity

osrd-compose
ghost: Will do, thanks for noticing

osrd-compose
bougue-pe: OK (for the sake of consistency I'd stick to `.sh`, but either one is fine).

Oh, forgot about this remark: should we document it in the main readme, then? (I'm OK to do this, once we agree on it, if you prefer)

ghost: I'd rather have it in the documentation than the readme since the instructions in the readme are still properly working for someone that is pulling the repository to start the project locally and we should cover the topic more than just "type this command".

Maybe we should add a page for "starting the development environment" with various scenarios depending on the components you'd like to work on ? Wdyt ?

docker/docker-compose.front.yml
```diff
@@ -0,0 +1,29 @@
+services:
```
emersion, line 1: Could we use [includes](https://docs.docker.com/compose/how-tos/multiple-compose-files/include/) for specialized docker-compose files to make it so users don't need to specify multiple `-f` flags?

ghost: In my understanding, this would not allow conditional inclusions ?

We use multiple -f flags to enable/disable features of the dev env and every dev requires different flags depending of their workflow.

emersion: Do we really need conditionals? Couldn't:

- `host` and `front` include the base config
- `host-front` include both of these

?

ghost: I can try, I had a lot of issues with overriding mounted files (on the gateway container) etc... I'm not completely sure it would work as expected, I'll give it a try today.

emersion:
I personally don't really like the custom script. IMHO open-source projects should use standard tooling, not force developers to run a magic script. (Also the script autosave functionality isn't tied to OSRD.)

I suppose I can just ignore it.

> Due to a limitation in Docker Compose not supporting targeting custom targets inside other Dockerfile in additional contexts, I've moved a part of the building system for the front on the Gateway Dockerfile. Not ideal, but does the trick.

Hm, what is the issue here? Why can't we use the latest stage (with everything built) as the `front` context for `gateway`?

(If we really can't do that, we could probably still use a separate Dockerfile in `front/`.)

docker/docker-compose.front.yml
emersion: I could also give it a try.

ghost:
> > Due to a limitation in Docker Compose not supporting targeting custom targets inside other Dockerfile in additional contexts, I've moved a part of the building system for the front on the Gateway Dockerfile. Not ideal, but does the trick.
> 
> 
> 
> Hm, what is the issue here? Why can't we use the latest stage (with everything built) as the `front` context for `gateway`?
> 
> 
> 
> (If we really can't do that, we could probably still use a separate Dockerfile in `front/`.)

Either way we must use additional contexts:
- The gateway needs the built front, or
- The front needs the built gateway

No matter how we integrate it, it does require additional contexts that can reference a specific stage of a docker file, a feature allowed by bakefiles but not compose file.

Since in production the front doesn't exist as a service but is bundled in the gateway, I've put the front building in the gateway dockerfile.

emersion:
Sorry, I still don't understand why we need to reference a specific stage in `additional_contexts`. Let's say `front/docker/Dockerfile` builds the frontend (runs these `npm` commands). Can't `gateway` use that Dockerfile (specifically, its last stage) and copy the built files over?

I'd prefer to keep the NPM commands inside `front/`.

ghost:
> Sorry, I still don't understand why we need to reference a specific stage in `additional_contexts`. Let's say `front/docker/Dockerfile` builds the frontend (runs these `npm` commands). Can't `gateway` use that Dockerfile (specifically, its last stage) and copy the built files over?

From my understanding you can give it a path but not the built result of another dockerfile. If you find a way to achieve this without a bakefile we can do so, but I've tried without any success.


> I'd prefer to keep the NPM commands inside `front/`.

I'd prefer too.

.github/workflows/build.yml
```diff
@@ -105,12 +105,6 @@ jobs:
...
 
-      - name: Upload front-build artifact
-        uses: actions/upload-artifact@v4
-        if: steps.bake-metadata.outputs.output_method == 'artifact' && contains(matrix.targets, 'front-build')
-        with:
-          name: front-build
-          path: osrd-front-build.tar
       - name: Upload core-build artifact
         uses: actions/upload-artifact@v4
         if: steps.bake-metadata.outputs.output_method == 'artifact' && contains(matrix.targets, 'core-build')
```
bougue-pe, line 110: 2 possible leftovers:
* There is still an `Upload front-nginx artifact` [step](https://github.com/OpenRailAssociation/osrd/blob/9314f35d52eab075d62e0c2a49f78d8e12f4935b/.github/workflows/build.yml#L162-L167), is it on purpose?

* Also there is still a `front-nginx` [target](https://github.com/OpenRailAssociation/osrd/blob/9314f35d52eab075d62e0c2a49f78d8e12f4935b/.github/scripts/bake-metadata.py#L42) in bake-metadata.py (and a few comments referencing nginx there): should it be removed?

osrd-compose
bougue-pe: I remembered some doc already existed on part of my use-case, so I just moved it to the new script in https://github.com/OpenRailAssociation/osrd/pull/10530.
To me, we can resolve this with the open PR (please do if OK to you)

emersion:
> From my understanding you can give it a path but not the built result of another dockerfile. If you find a way to achieve this without a bakefile we can do so, but I've tried without any success.

Eh, you're correct. It seems like this is https://github.com/compose-spec/compose-spec/issues/336.

.github/workflows/build.yml
ghost: Nice catch, gonna remove those.

docker/docker-compose.front.yml
ghost: So... bad news... I've just had the same kind of issues with the overriding where the file in the container was not the right one in all the cases. I might be missing something, but I believe this change can be tried later, wdyt ?

emersion:
Is host+front mode really useful? If one runs in host mode _and_ wants to hack on the front, they'll probably just run it outside docker, no?

ghost:
> Is host+front mode really useful? If one runs in host mode _and_ wants to hack on the front, they'll probably just run it outside docker, no?

I feel that making this assumption is risky : we do not know each configuration the user is going to have relative to Docker and I believe we should keep it as unopinionated as possible on those subjects.

Also, the same issue exists with Single Worker and Front, there is no guarantee that someone working on Core could not require modifications on the front. I feel we should not remove an easy option for the developers, starting the stack outside docker can be cumbersome.

.github/workflows/build.yml
ghost: Done in latest commit.

clarani:
I think we need to update the README ?

EDIT (by @bougue-pe ):
-> done in https://github.com/OpenRailAssociation/osrd/pull/10530

gateway/Dockerfile
```diff
@@ -48,10 +48,43 @@ COPY . .
...
+COPY --from=front_src package.json package-lock.json /app/
+RUN npm install --frozen-lockfile
+
+# Fill build time values
+ARG OSRD_GIT_DESCRIBE
+ENV VITE_OSRD_GIT_DESCRIBE=${OSRD_GIT_DESCRIBE}
+
+# Generate the licenses file and build
+COPY --from=front_src . /app
+RUN npm run generate-licenses && npm run build
```
clarani, line 77: Building the gateway crashed on my laptop with the error `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`.

I had to add this line here to solve the issue

```suggestion
ENV NODE_OPTIONS="--max_old_space_size=4096"
RUN npm run generate-licenses && npm run build
```

clarani (approved):
LGTM :white_check_mark:

RomainValls (approved):
Tested on Mac, Docker desktop, works great 💪

PR #10504, opened 2025-01-23
OpenFGA client crate

leovalais:
> [!NOTE]
> This is a lengthy PR which also requires some amount of OpenFGA knowledge to be understood. It's probably better to peer review it.

* Meant to remain a new sub-crate which will be included in `editoast_authz` (in which we'll map the authorization model, define our objects, their conversion, relations, etc.)
* The `fga` binary is necessary to run unit tests for now: https://github.com/openfga/cli
* So is a running OpenFGA instance

TODO:

- [ ] Implement all the operations we'll need
	- [x] Operations needed for the roles reimplementation
	- [x] `POST /stores/{}/list-users`
	- [ ] `POST /stores/{}/stream-list-objects`
	- [ ] `POST /stores/{}/expand`
	- [ ] `POST /stores/{}/read`
	- [x] `POST /stores/{}/batch-check`
	- [ ] Other operations currently not needed, but since there's so few of them, we may want to implement them for completeness?
- [x] Keep an `authorization_model_id` in the client to forward it to requests—as advised by OpenFGA
- [x] Improve tests utils & debugging tools
	- [x] Client `check` assertions (make them public + async variant?)
	- [x] Parse OpenFGA errors instead of just forwarding them generically
	- [x] Expose `compile_model`?
- [x] Documentation (for real this time :eyes:)
	- [x] `mod client`
	- [x] `mod model`
	- [x] `crate` (ongoing)
- [x] **INSTRUMENT EVERYTHING**
- [x] Setup OpenFGA in `docker compose`
- [x] Add check for `User.id ≠ '*'`
- [ ] Add early verification of `User.id` and `Object.id`? Not really required for us.
- [x] Implement `list-users` as this will change the `trait User` API
- [x] Macros
	- [x] `fga!`
	- [x] `relations!`
	- [x] `derive(User)`
	- [x] `derive(Object)`

### About macros

This crates introduces a few macro_rules, which are only used for tests currently.

* `relations!` allows defining typed relations concisely to allow quick comparisons with the OpenFGA schema. I think we should keep this one, that'll help the review process and reduce straightforward boilerplate.
* `fga_type!` declares the newtype structs representing the `types` of OpenFGA and generates a few implementations. I think this one should remain a `#[cfg(test)]` definition. However I think derive macros `derive(fga::User, fga::Object)` would help us and cost nothing as they are trivial to implement.
* `fga!` allows manipulating and instantiating types and relations with *the same syntax* as OpenFGA, but in a type-safe way. This is particularly useful for the conciseness of unit tests. I think it should be exported.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -39,6 +39,32 @@ services:
       interval: 5s
       retries: 20
 
+  openfga-migrate:
+    depends_on:
+      postgres: { condition: service_healthy }
+    image: openfga/openfga:latest
+    container_name: osrd-openfga-migrate
+    command: migrate
+    environment:
+      OPENFGA_DATASTORE_ENGINE: "postgres"
+      OPENFGA_DATASTORE_URI: "postgres://osrd:password@postgres:5432/osrd?search_path=openfga"
+
+  openfga:
+    depends_on:
+      openfga-migrate:
+        condition: service_completed_successfully
+    image: openfga/openfga:latest
+    container_name: osrd-openfga
+    environment:
+      OPENFGA_DATASTORE_ENGINE: "postgres"
+      OPENFGA_DATASTORE_URI: "postgres://osrd:password@postgres:5432/osrd?search_path=openfga"
+      OPENFGA_HTTP_ADDR: "0.0.0.0:8091"
+      OPENFGA_PLAYGROUND_PORT: 8092
+    command: run
+    ports:
+      - "8091:8091" # http
+      - "8092:8092" # playground
+
   rabbitmq:
     image: rabbitmq:4-management
     container_name: osrd-rabbitmq
--- a/docker/docker-compose.pr-tests.yml
+++ b/docker/docker-compose.pr-tests.yml
@@ -119,10 +119,10 @@ services:
       valkey: { condition: service_healthy }
       rabbitmq: { condition: service_healthy }
     restart: unless-stopped
-    ports: [ "8091:8091" ]
+    ports: [ "8089:8089" ]
     environment:
-      EDITOAST_PORT: 8091
-      ROOT_URL: "http://localhost:8091"
+      EDITOAST_PORT: 8089
+      ROOT_URL: "http://osrd-editoast-pr-tests:8089"
       VALKEY_URL: "redis://valkey:6380"
       DATABASE_URL: "postgres://osrd:password@postgres:5433/osrd"
       TELEMETRY_KIND: "opentelemetry"
@@ -134,7 +134,7 @@ services:
       - -c
       - "diesel migration run && exec editoast runserver"
     healthcheck:
-      test: [ "CMD", "curl", "-f", "http://localhost:8091/health" ]
+      test: [ "CMD", "curl", "-f", "http://localhost:8089/health" ]
       start_period: 4s
       interval: 5s
       retries: 6
--- a/docker/docker-compose.pr-tests.yml
+++ b/docker/docker-compose.pr-tests.yml
@@ -54,6 +54,31 @@ services:
     networks:
       - pr-tests
 
+  openfga-migrate:
+    image: openfga/openfga:latest
+    container_name: osrd-openfga-migrate-pr-tests
+    command: migrate
+    environment:
+      OPENFGA_DATASTORE_ENGINE: "postgres"
+      OPENFGA_DATASTORE_URI: "postgres://osrd:password@postgres:5433/osrd?search_path=openfga"
+    networks:
+      - pr-tests
+
+  openfga:
+    image: openfga/openfga:latest
+    container_name: osrd-openfga-pr-tests
+    environment:
+      OPENFGA_DATASTORE_ENGINE: "postgres"
+      OPENFGA_DATASTORE_URI: "postgres://osrd:password@postgres:5433/osrd?search_path=openfga"
+      OPENFGA_HTTP_ADDR: "0.0.0.0:8191"
+      OPENFGA_PLAYGROUND_PORT: 8192
+    command: run
+    ports:
+      - "8191:8191" # http
+      - "8192:8192" # playground
+    networks:
+      - pr-tests
+
   rabbitmq:
     image: rabbitmq:4-management
     container_name: osrd-rabbitmq-pr-tests
--- a/editoast/Dockerfile
+++ b/editoast/Dockerfile
@@ -40,6 +40,16 @@ RUN rustup component add llvm-tools && \
     rustup component add rustfmt && \
     rustup component add clippy && \
     cargo install grcov
+
+# install `fga` binary used in `crate fga` unit tests
+RUN mkdir install \
+    && cd install \
+    && wget https://github.com/openfga/cli/releases/download/v0.6.4/fga_0.6.4_linux_amd64.tar.gz \
+    && tar xf fga_0.6.4_linux_amd64.tar.gz \
+    && mv fga /usr/local/bin \
+    && cd .. \
+    && rm -rf install
+
 COPY --from=planner /editoast/recipe.json recipe.json
 COPY --from=planner /editoast/editoast_derive/ editoast_derive
 COPY --from=test_data . /tests/data
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -517,31 +517,6 @@ jobs:
     name: Check editoast tests
     needs:
       - build
-
-    services:
-      postgres:
-        image: postgis/postgis
-        env:
-          POSTGRES_USER: postgres
-          POSTGRES_PASSWORD: password
-        ports:
-          - 5432:5432
-        # needed because the postgres container does not provide a healthcheck
-        options: >-
-          --health-cmd pg_isready
-          --health-interval 10s
-          --health-timeout 5s
-          --health-retries 5
-      valkey:
-        image: valkey/valkey
-        ports:
-          - 6379:6379
-        options: >-
-          --health-cmd "valkey-cli ping"
-          --health-interval 5s
-          --health-timeout 5s
-          --health-retries 5
-
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -558,13 +533,21 @@ jobs:
         run: |
           docker load --input ./osrd-editoast-test.tar
 
+      - name: Startup the test infrastructure
+        id: start_editoast_tests_infra
+        run: |
+          set -e
+
+          services='valkey openfga'
+          composes='-f docker-compose.yml'
+          docker compose $composes pull --policy missing $services
+          docker compose $composes up --no-build -d $services
+        env:
+          DOCKER_BUILDKIT: 1
+          COMPOSE_DOCKER_CLI_BUILD: 1
+
       - name: Execute tests within container
         run: |
-          docker run --rm --net=host \
-            -v $PWD/docker/init_db.sql:/init_db.sql \
-            postgis/postgis:latest \
-            psql postgresql://postgres:password@localhost:5432 -f /init_db.sql
-
           # snapshot testing library `insta` requires CI=true
           docker run --name=editoast-test --net=host -v $PWD/output:/output \
             -e DATABASE_URL="postgres://osrd:password@localhost:5432/osrd" \
--- a/docker/docker-compose.host.yml
+++ b/docker/docker-compose.host.yml
@@ -9,6 +9,18 @@ services:
     ports: !reset []
     network_mode: host
 
+  openfga-migrate:
+    ports: !reset []
+    network_mode: host
+    environment:
+      OPENFGA_DATASTORE_URI: "postgres://osrd:password@localhost:5432/osrd?search_path=openfga"
+
+  openfga:
+    ports: !reset []
+    network_mode: host
+    environment:
+      OPENFGA_DATASTORE_URI: "postgres://osrd:password@localhost:5432/osrd?search_path=openfga"
+
   editoast:
     ports: !reset []
     network_mode: host
--- a/docker/gateway.pr-tests.simple.toml
+++ b/docker/gateway.pr-tests.simple.toml
@@ -20,7 +20,7 @@ redirect_404_to_index = true
 [[targets]]
 tracing_name = "editoast"
 prefix = "/api"
-upstream = "http://osrd-editoast-pr-tests:8091"
+upstream = "http://editoast:8089"
 require_auth = true
 
 [[targets]]
--- a/docker/init_db.sql
+++ b/docker/init_db.sql
@@ -6,6 +6,9 @@ create database template_osrd WITH IS_TEMPLATE true;
 -- Connect to template_osrd
 \c template_osrd
 
+-- Create openfga schema
+create schema openfga;
+
 -- Enable required extensions
 create extension postgis;
 create extension pg_trgm;
@@ -13,6 +16,8 @@ create extension unaccent;
 
 -- Grant privileges to user
 grant all privileges on schema public to osrd;
+grant all privileges on schema openfga to osrd;
+
 
 -- Create osrd database using template_osrd
 create database osrd TEMPLATE template_osrd;
--- a/editoast/Cargo.lock
+++ b/editoast/Cargo.lock
@@ -865,7 +865,7 @@ version = "3.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fde0e0ec90c9dfb3b4b1a0891a7dcd0e2bffde2f7efed5fe7c9bb00e5bfb915e"
 dependencies = [
- "windows-sys 0.48.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1150,6 +1150,26 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 1.0.109",
+]
+
+[[package]]
+name = "derive_more"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "093242cf7570c207c83073cf82f79706fe7b8317e98620a47d5be7c3d8497678"
+dependencies = [
+ "derive_more-impl",
+]
+
+[[package]]
+name = "derive_more-impl"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bda628edc44c4bb645fbe0f758797143e4e07926f7ebf4e9bdfbd3d2ce621df3"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.98",
 ]
 
 [[package]]
@@ -1620,6 +1640,36 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1e6853b52649d4ac5c0bd02320cddc5ba956bdb407c4b75a2c6b75bf51500f8c"
 dependencies = [
  "simd-adler32",
+]
+
+[[package]]
+name = "fga"
+version = "0.1.0"
+dependencies = [
+ "derive_more",
+ "fga_derive",
+ "futures 0.3.31",
+ "futures-util",
+ "itertools",
+ "reqwest",
+ "serde",
+ "serde_json",
+ "stdext",
+ "tempfile",
+ "thiserror 2.0.11",
+ "tokio",
+ "tracing",
+ "tracing-subscriber",
+ "url",
+]
+
+[[package]]
+name = "fga_derive"
+version = "0.1.0"
+dependencies = [
+ "darling",
+ "quote",
+ "syn 2.0.98",
 ]
 
 [[package]]
@@ -2161,7 +2211,7 @@ dependencies = [
  "httpdate",
  "itoa",
  "pin-project-lite",
- "socket2 0.4.10",
+ "socket2 0.5.8",
  "tokio",
  "tower-service",
  "tracing",
@@ -4436,6 +4486,12 @@ name = "static_assertions"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
+
+[[package]]
+name = "stdext"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4af28eeb7c18ac2dbdb255d40bee63f203120e1db6b0024b177746ebec7049c1"
 
 [[package]]
 name = "stringprep"
--- a/editoast/Cargo.toml
+++ b/editoast/Cargo.toml
@@ -13,6 +13,8 @@ members = [
   "editoast_osrdyne_client",
   "editoast_schemas",
   "editoast_search",
+  "fga",
+  "fga_derive",
   "osm_to_railjson",
 ]
 
@@ -27,6 +29,7 @@ license = "LGPL-3.0"
 [workspace.dependencies]
 chrono = { version = "0.4.39", default-features = false, features = ["serde"] }
 clap = { version = "4.5", features = ["derive", "env"] }
+darling = "0.20"
 derivative = "2.2.0"
 diesel = { version = "2.2", default-features = false, features = [
   "32-column-tables",
@@ -66,9 +69,12 @@ paste = "1.0.15"
 postgis_diesel = { version = "2.4.1", features = ["serde"] }
 postgres-openssl = "0.5.1"
 pretty_assertions = "1.4.1"
+proc-macro2 = "1.0"
+quote = "1.0"
 rand = "0.9.0"
 rangemap = "1.5.1"
 regex = "1.11"
+syn = "2.0"
 # 0.12.0 to 0.12.4 have weird timeout issues https://github.com/seanmonstar/reqwest/issues/2283
 # This bug was introduced between 0.12.0 and 0.12.3.
 reqwest = { version = "0.11.27", features = ["json"] }
--- a/editoast/editoast_derive/Cargo.toml
+++ b/editoast/editoast_derive/Cargo.toml
@@ -7,12 +7,12 @@ edition.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-darling = "0.20"
+darling.workspace = true
 paste.workspace = true
-proc-macro2 = "1.0"
-quote = "1.0"
+proc-macro2.workspace = true
+quote.workspace = true
 serde_json.workspace = true
-syn = "2.0"
+syn.workspace = true
 
 [lib]
 proc-macro = true
--- a/editoast/fga/Cargo.toml
+++ b/editoast/fga/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+name = "fga"
+license.workspace = true
+version.workspace = true
+edition.workspace = true
+
+[dependencies]
+fga_derive = { path = "../fga_derive" }
+futures.workspace = true
+futures-util.workspace = true
+itertools.workspace = true
+reqwest.workspace = true
+serde.workspace = true
+serde_json.workspace = true
+tempfile.workspace = true
+thiserror.workspace = true
+tokio.workspace = true
+tracing.workspace = true
+url.workspace = true
+
+[dev-dependencies]
+derive_more = { version = "2.0.1", features = ["from"] }
+stdext = "0.3.3"
+tracing-subscriber.workspace = true
--- a/editoast/fga/src/client.rs
+++ b/editoast/fga/src/client.rs
@@ -1,0 +1,1005 @@
+mod authorization_models;
+mod queries;
+mod stores;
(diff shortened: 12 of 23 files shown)


fga/Cargo.lock
woshilapin, line 1: I know it's nothing to do with `editoast` (at least, supposed to be generic enough)... but there still is some opinionated restrictions in which OpenFGA is to be used (no heterogeneous types for objects in a relation, hence an associated type instead of a generic). I'd say this is still a crate tied to the current project. How about naming it `osrd_fga` or `editoast_fga` still? Also, it would help identify this folder as a crate (although, we might want to move all of our crates in a sub-folder name crates at some point since they start to be a lot of them).

By the way, shouldn't this be moved under the `editoast` folder (like `osm-to-railjson` is, in some way, not completely `editoast` but yet in the same workspace)?

fga/src/client.rs
```diff
@@ -0,0 +1,849 @@
...
+use crate::model::ParsingError;
+use crate::model::QueryObjects;
+use crate::model::{AsUser, Check, Object, Relation, Tuple, User};
+
+const OPENFGA_WRITES_MAX_TUPLES: usize = 100;
+
+#[derive(Debug, Clone)]
+pub struct Client {
+    store: Store,
+    authorization_model_id: Option<String>,
```
woshilapin: As mentioned in peer-review, this one could be a `String` and not an `Option<String>` since it basically have the same kind of constraint than the `store` above (doesn't exist during the constructor execution, but always exist after that).

fga/src/client.rs
```diff
@@ -0,0 +1,849 @@
...
+    ) -> Result<Client, InitializationError> {
+        Self::try_create_store_inner(store_name, settings, false).await
+    }
+
+    #[cfg(test)]
+    #[tracing::instrument(err)]
+    pub async fn try_create_store(
+        store_name: String,
+        settings: ConnectionSettings,
+        reset: bool,
```
woshilapin: As suggested, maybe this could be a conditional field of `ConnectionSettings` instead? You should be able to merge both `try_create_store` and `try_create_store_inner` into a single function that takes only 2 parameters.

fga/src/client.rs
```diff
@@ -0,0 +1,849 @@
...
+        err
+    }
+}
+
+// Public API of the client
+// -------------------------
+
+impl Client {
+    #[tracing::instrument(err)]
+    pub async fn try_init(
```
woshilapin: Only naming, but we basically have 2 types of constructor, one that return a client initialized with the latest store, the other one that creates a new store. First of all, I'd say both functions should contain the word "store"? And secondly, maybe using `new` instead of `create` would feel more like an constructor in Rust? So I'd suggest:
- `try_latest_store`
- `try_new_store`

What do you think?

fga/src/client.rs
```diff
@@ -0,0 +1,849 @@
...
+use tuples::RawTuple;
+
+use std::future::{self, Future};
+
+use futures::{stream, TryStreamExt as _};
+use itertools::Itertools as _;
+
+use crate::model::ParsingError;
+use crate::model::QueryObjects;
+use crate::model::{AsUser, Check, Object, Relation, Tuple, User};
```
woshilapin: Can be exploded in lines...

fga/src/client.rs
```diff
@@ -0,0 +1,849 @@
...
+/// the ownership of the [Client] (cloned internally, but that's a cheap operation).
+///
+/// The [ContinuationUnfolder::stream] closure is expected to return a `Vec<T>` of items at each call,
+/// but those will be flatten into a stream of `T` as expected, and not a stream of `Vec<T>`.
+///
+/// This unfolder can also be provided with a context that will be passed to the closure at each call.
+/// The closure is free to modify it but **must** provide it anew in its result using [UnfoldNextState].
+struct ContinuationUnfolder<C> {
+    client: Client,
+    ctx: C,
```
woshilapin: I'm not sure that I understand why you need the `ctx` in our usecase.

fga/src/client.rs
```diff
@@ -0,0 +1,849 @@
...
+/// the [UnfoldNextState::continuation] is empty. For lifetimes reasons, the closure is provided with
+/// the ownership of the [Client] (cloned internally, but that's a cheap operation).
+///
+/// The [ContinuationUnfolder::stream] closure is expected to return a `Vec<T>` of items at each call,
+/// but those will be flatten into a stream of `T` as expected, and not a stream of `Vec<T>`.
+///
+/// This unfolder can also be provided with a context that will be passed to the closure at each call.
+/// The closure is free to modify it but **must** provide it anew in its result using [UnfoldNextState].
+struct ContinuationUnfolder<C> {
+    client: Client,
```
woshilapin: I believe we can also get rid of capturing the `client`. I just tried to modify the `fn stores` that way and it seems to work.

```rust
pub fn stores(&self) -> impl stream::TryStream<Ok = Store, Error = RequestFailure> + '_ { // NOTE: the additional `'_` here
    ContinuationUnfolder::new(self.clone(), ()).stream(
        move |UnfoldArgs { // NOTE: the additional `move` here
                  client: _client, // NOTE: that I don't use the `client` in the closure anymore
                  continuation,
                  ctx: _ctx,
              }| async move {
            let (stores, continuation) = self // NOTE: the use of `self` here instead of `client`
                .get_stores(None, continuation.as_ref().map(String::as_str))
                .await?;
            Ok((
                stores,
                UnfoldNextState {
                    ctx: (),
                    continuation,
                },
            ))
        },
    )
}
```

fga/src/client.rs
```diff
@@ -0,0 +1,849 @@
...
+/// The [ContinuationUnfolder::stream] function takes a closure that will be called repeatedly until
+/// the [UnfoldNextState::continuation] is empty. For lifetimes reasons, the closure is provided with
+/// the ownership of the [Client] (cloned internally, but that's a cheap operation).
+///
+/// The [ContinuationUnfolder::stream] closure is expected to return a `Vec<T>` of items at each call,
+/// but those will be flatten into a stream of `T` as expected, and not a stream of `Vec<T>`.
+///
+/// This unfolder can also be provided with a context that will be passed to the closure at each call.
+/// The closure is free to modify it but **must** provide it anew in its result using [UnfoldNextState].
+struct ContinuationUnfolder<C> {
```
woshilapin: If we are able to remove `client` and `context` (like my comments below mention), we actually might be able to simplify all of that structure for a state inside the `try_unfold` that would only be `Continuation`. We can try that together tomorrow in peer if you'd like.

fga/src/client.rs
```diff
@@ -0,0 +1,849 @@
...
+            "set client authorization model ID"
+        );
+        Ok(())
+    }
+
+    /// Pushes a new authorization model into OpenFGA
+    ///
+    /// /!\ WARNING /!\ The `authorization_model_id` serialized in the [Client] won't change.
+    /// If you wan't this model to be used for following client usage, you need to call
+    /// [Client::actualize_authorization_model] afterwards.
```
woshilapin: Why can't we take a `&mut` and actually update the `authorization_model_id`? I guess it's an API choice, but since the client does contain a reference to one model, I would say that if you push a new model, you want it. If decide to do that, we may also change the name of the function to `update_authorization_model` which might better convey the fact that we change the client's reference too. Or there might be a case for both API 🤷

fga/src/client.rs
```diff
@@ -0,0 +1,849 @@
...
+        self.writes.push(RawTuple::from(tuple));
+        self
+    }
+
+    /// Concurrently sends write requests to OpenFGA in 100-tuple chunks
+    ///
+    /// /!\ WARNING /!\ No transactional state is set up, so should any request fail,
+    /// the tuples written by other successful requests will remain in OpenFGA.
+    /// This function also returns at the first failing request, so OpenFGA may still
+    /// write some tuples **after** this function exits.
```
woshilapin: I'm actually wondering: what if we would panic each time we are over this 100 tuples limit? Let's say we don't support non-transactional batches. That would be a nice property and let the client's user to deal to split up the batch into logicals ones that can be transactional. It really is an open question.

fga/src/client.rs
```diff
@@ -0,0 +1,849 @@
...
+        )
+        .unwrap()
+    }
+}
+
+/// Convenience trait to query OpenFGA from [crate::model] query types directly
+///
+/// For example:
+///
+/// ```ignore
```
woshilapin: We should be able to remove the `ignore` and make something that compiles here. If you're worry of all the setup for making this work that would be too much noise for what you're trying to document, you can actually use `/// #` instead of `///`, the code will be use for compilation, but not output in the documentation.

woshilapin (commented):
Only a first pass on `src/client.rs`. I'll take a look at the rest later.

fga/Cargo.lock
leovalais: The goal is to move the crate in editoast's workspace of course. I just worked on a different ws for dev comfort only :)

For the name, first, it's not `openfga` because I didn't want to coin that name since an official crate that we may want to use could (will?) appear. The crate has nothing editoast-specific in it. We name a function that does X and called by a procedure P `do_X` or `perform_X`, not `do_X_for_P` or `on_behalf_of_P_perform_X`. The context does not usally appear in the name when separation of concerns is respected. The only relation (badum tss) `crate fga` has with editoast is being a dependency. Let's say in the future editoast gets split in two services, the `crate editoast_fga` shouldn't need to be renamed imo. The same applies to `osm_to_railjson` btw, it's not `editoast_osn_to_railjson`.

For the same reason I'm not a huge fan of `osrd_fga`. If another project wants to use the crate, they'll have in their `Cargo.toml` something like:

```toml
fga = { git = "http://github.com/osrd.git/editoast/fga" } # I don't remember the URL exactly :eyes:
```

That's enough context to me. In my head, `osrd_fga` would be the crate that exposes the authorization model of OSRD. Take `axum_extras` or `diesel_async`, the `_` means "subsystem", not "dependency" or "relation".

As for being opinionated, I don't see the issue, even if we were to publish it. A crate can be opinionated in my book. It's up to users to decide whether they like it or not imo. I think not using the name `openfga` and masquerading as an official crate is enough 🤷

I'm not hell bent on keeping this name though. But I like `fga` it as it's short, refers to the service it talks to, and it's not used in crates.io yet, so it won't conflict with another crate should we need it.

fga/src/client.rs
leovalais: Currently, both invocations use `()`. But I added it in case we want to provide a page size (we always use OpenFGA's default currently, but there's a query param for that). Could we keep it while the public `Client` API isn't dry yet? I'll remove it if it's still unused when all OpenFGA operations have been implemented. It's much easier to do this way 😅

fga/src/client.rs
leovalais: image

fga/src/client.rs
leovalais: > Or there might be a case for both API

Currently there isn't, so `update_authorization_model(&mut self)` is indeed much better. We may want to add the other behavior back when we'll tackle the migration problem, but that's for later.

fga/src/client.rs
leovalais: `try_with_store` would be more accurate as we'll always want the `osrd-editoast` store. There will likely be more recent stores than this one created by unit tests.

fga/src/client.rs
leovalais: fbf7b50b1

fga/Cargo.lock
woshilapin: Well, you orally told me all of that and I didn't forgot... and did agree. My only point was about the "opinionated part". So basically, this part of your comment is interesting to me (the rest is nice for history of the PR though).

> As for being opinionated, I don't see the issue, even if we were to publish it. A crate can be opinionated in my book. It's up to users to decide whether they like it or not imo. I think not using the name openfga and masquerading as an official crate is enough 🤷

... and I believe I do agree, thanks for stating that publishing an opinionated piece is fine. That said, we might want to emphase this choice in the documentation of the `Relation` trait.

fga/src/client.rs
leovalais: Hmmm but that's deferring the burden of managing transactions/rollbacks to the caller. If rollbacks are possible and sound, I'd say that a high-level `Client` should take care of that detail for the us... but then again "explicit is better than implicit". Here's a proposition:

```rust
impl PreparedWrites {
	async fn execute(self) -> Result<(), BatchWritesError>;
}

struct BatchWritesError {
	failures: Vec<RequestFailure>,
	rollback: Rollback
}

/// Operations to do in order to rollback a failed tuple Writes operation
struct Rollback {
	writes: Vec<RawTuple>,
	deletes: Vec<RawTuple>,
}

impl Client {
	// here BatchWritesError.failures is the list of failures of failed rollback requests
	// BatchWritesError.rollback is a different `Rollback` instance containing only the
	// tuples that failed to be rolled back. This way we can try to `rollback` again.
	async fn rollback(&self, rollback: Rollback) -> Result<(), BatchWritesError>;
}

impl PreparedWrites {
	async fn execute_or_rollback(self) -> Result<(), BatchWritesError>;
}
```

Wdyt?

fga/src/client.rs
leovalais: 35913e894

fga/src/client.rs
leovalais: 91ec646ef

fga/src/client.rs
leovalais: 9c71d2845

fga/src/client.rs
leovalais: a767ba296

fga/src/client.rs
```diff
@@ -0,0 +1,841 @@
...
+            .await
+            .expect("Failed to initialize client")
+        };
+    }
+
+    #[tokio::test]
+    async fn test_try_init() {
+        setup_tracing();
+        let client = Client::try_with_store(
+            "lol".to_owned(),
```
leovalais: this store only exists on my machine, to change...

fga/src/client.rs
leovalais: While making the change I remembered why it's an `Option`: `try_new_store` returns a client that is configured for a store that doesn't have any authorization model yet. Returning a `Client` with `authorization_model_id = ""` is an invalid state (even though OpenFGA may accept an empty string as "not provided", I haven't checked).

Do we leave things as they are or do we require an authorization model to be provided as well in `try_new_store`?

fga/src/client.rs
leovalais: 7e47f29ac

fga/src/client.rs
woshilapin: wrong commit?

leovalais: indeed.. https://github.com/OpenRailAssociation/osrd/pull/10504/commits/937f800e08f7418279267885065bf3aa742a6254

fga/src/client.rs
leovalais: removed the test: if `try_with_store` has a regression, other tests will fail

141120c68

fga/src/client.rs
woshilapin: Nice. You may need to update the comment above `ContinuationUnfolder` too.

fga/src/client.rs
woshilapin: Do we need to keep `try_new_store`? It's not clear to me that we both need `try_with_store` and `try_new_store` and it feel we could only keep the former (to be honest, both signatures are the same and it's not clear why the differences). What do you think?

fga/Cargo.toml
woshilapin, line 1: We're missing a `.gitignore` that ignore the `target` folder at the root of `fga`.

fga/src/client.rs
woshilapin: ok, what do you think about [that](https://github.com/OpenRailAssociation/osrd/pull/10784) to simplify a bit all of this code for continuation handling?

fga/src/model.rs
```diff
@@ -0,0 +1,176 @@
...
+pub trait User: fmt::Debug + Sized {
+    const NAMESPACE: &'static str;
+
+    fn id(&self) -> &str;
+
+    fn fga_ident(&self) -> String {
+        format!("{}:{}", Self::NAMESPACE, self.id())
+    }
+
+    fn tbpa() -> Tbpa<Self> {
```
woshilapin: Maybe need to be renamed `all`? Or `wildcard`?

fga/Cargo.toml
woshilapin, line 1: We're missing `openfga` in the `docker-compose.yml`.

woshilapin, line 1: We're missing an additionnal step in the CI to launch the tests? (or do we wait for this crate to be merged into `editoast`'s workspace?

fga/src/model.rs
```diff
@@ -0,0 +1,176 @@
...
+            ident: ident.to_string(),
+            expected_type: Self::NAMESPACE,
+        })?;
+        if ns != Self::NAMESPACE {
+            return Err(ParsingError {
+                ident: ident.to_string(),
+                expected_type: Self::NAMESPACE,
+            });
+        }
+        Ok(Self::from(id.to_string()))
```
woshilapin:
```suggestion
        let prefix: String = format!("{}:", Self::NAMESPACE);
        if let Some(id) = ident.strip_prefix(&prefix) {
            Ok(Self::from(id.to_string()))
        } else {
            Err(ParsingError {
                ident: ident.to_string(),
                expected_type: Self::NAMESPACE,
            })
        }
```

fga/src/client.rs
leovalais: The `Client` is in an inconsistent state if it's not attached to a particular store. If we want to create a new store, we need a client, but to have a client, we need a store... We could solve this bootstrapping issue by exposing the `post_stores` low-level API, but then we'd trade an internal state management issue into an API issue by "leaking" implementation constraints to the user. Not a huge fan tbh.

fga/src/client.rs
leovalais: I kinda regret writing documentation now... cdfd7821e

fga/Cargo.toml
leovalais: `fga` will be moved into editoast's workspace. So it will benefit from its `.gitignore`, `.cargo`, etc. Do you want me to still add one in the meantime?

fga/src/doctest_setup.rs
```diff
@@ -0,0 +1,37 @@
...
+    Document {
+        reader: Person,
+        writer: Person,
+
+        can_read: Person,
+        can_write: Person
+    }
+}
+
+use fga::model::Relation as _;
```
woshilapin: Can be move at the top.

fga/src/model.rs
```diff
@@ -0,0 +1,241 @@
...
+}
+
+pub trait Object: fmt::Debug + From<String> {
+    const NAMESPACE: &'static str;
+
+    fn id(&self) -> &str;
+
+    fn fga_ident(&self) -> String {
+        format!("{}:{}", Self::NAMESPACE, self.id())
+    }
```
woshilapin: This piece of code is the same between `User` and `Object`. How about making an `trait Id` that both `Object` and `User` would be a subset of?

fga/src/model.rs
```diff
@@ -0,0 +1,241 @@
...
+            panic!(
+                "Refusing to generate an identifier for a type-bound public access\n\
+                without using the `fga::model::Tpba` type.\n\
+                This is a security measure to avoid having user identifiers coerced to\n\
+                public accesses maliciously."
+            );
+        }
+        format!("{}:{}", Self::User::NAMESPACE, id)
+    }
+}
```
woshilapin: I don't understand why we need this `trait`. It's exactly the same than `User` (except `User` has a bit more). It feels that everywhere a `AsUser` would be expected, a `User` bound would be enough.

docker-compose.yml
```diff
@@ -38,6 +38,32 @@ services:
...
+
+  openfga:
+    depends_on:
+      openfga-migrate:
+        condition: service_completed_successfully
+    image: openfga/openfga:latest
+    container_name: osrd-openfga
+    environment:
+      OPENFGA_DATASTORE_ENGINE: "postgres"
+      OPENFGA_DATASTORE_URI: "postgres://osrd:password@postgres:5432/osrd"
```
flomonster: We should use another database.

```suggestion
      OPENFGA_DATASTORE_URI: "postgres://osrd:password@postgres:5432/openfga"
```

docker-compose.yml
```diff
@@ -38,6 +38,32 @@ services:
...
+  openfga-migrate:
+    depends_on:
+      postgres:
+        condition: service_healthy
+    image: openfga/openfga:latest
+    container_name: osrd-openfga-migrate
+    command: migrate
+    environment:
+      OPENFGA_DATASTORE_ENGINE: "postgres"
+      OPENFGA_DATASTORE_URI: "postgres://osrd:password@postgres:5432/osrd"
```
flomonster: We don't need a service for that we can migrate before running the server.

```suggestion
```

docker-compose.yml
```diff
@@ -38,6 +38,32 @@ services:
...
+        condition: service_completed_successfully
+    image: openfga/openfga:latest
+    container_name: osrd-openfga
+    environment:
+      OPENFGA_DATASTORE_ENGINE: "postgres"
+      OPENFGA_DATASTORE_URI: "postgres://osrd:password@postgres:5432/osrd"
+    command: run --http-addr=:8091 --playground-port=8092
+    ports:
+      - "8091:8091"
+      - "8092:8092" # openfga playground /!\ DISABLE IN PROD
```
flomonster: Docker compose are not used in production
```suggestion
      - "8092:8092"
```

docker-compose.yml
```diff
@@ -38,6 +38,32 @@ services:
...
+      OPENFGA_DATASTORE_ENGINE: "postgres"
+      OPENFGA_DATASTORE_URI: "postgres://osrd:password@postgres:5432/osrd"
+
+  openfga:
+    depends_on:
+      openfga-migrate:
+        condition: service_completed_successfully
+    image: openfga/openfga:latest
+    container_name: osrd-openfga
+    environment:
```
flomonster, line 58: Instead of providing it through the command line, we can use env var.

```suggestion
    environment:
      OPENFGA_HTTP_ADDR=:8091
      OPENFGA_PLAYGROUND_ENABLED=true
```

docker-compose.yml
```diff
@@ -38,6 +38,32 @@ services:
...
+      OPENFGA_DATASTORE_ENGINE: "postgres"
+      OPENFGA_DATASTORE_URI: "postgres://osrd:password@postgres:5432/osrd"
+    command: run --http-addr=:8091 --playground-port=8092
```
flomonster:
```suggestion
    entrypoint: /openfga migrate && /openfga run
```

docker-compose.yml
flomonster, line 1: We should also handle other overriding docker-compose files.
See `ls docker/docker-compose*`

flomonster (commented):
Small review of the docker-compose

docker-compose.yml
leovalais: In the documentation, they seem to do it in two steps both in the docker compose and in the helm charts. Not sure why though. https://openfga.dev/docs/getting-started/setup-openfga/docker

docker-compose.yml
leovalais: That was just a reminder, also indicating the existence of the playground ^^

docker-compose.yml
leovalais: I tried to use these but they didn't seem to work, maybe I fumbled somewhere 🤷

docker-compose.yml
leovalais: I avoided these on purpose to avoid copy pasting config that could need changes (looks like I was right to do so 🙃)

fga/src/client.rs
leovalais: Let's keep the Option, as discussed.

fga/src/client.rs
leovalais: d8401ac52

fga/src/client.rs
leovalais: It's already gone, ciao

fga/src/client.rs
leovalais: The continuationunfolder is gone as well, ciao

fga/src/client.rs
leovalais: Wait and see, panic in a follow up PR.

fga/src/client.rs
leovalais: merged

fga/src/model.rs
leovalais: wildcard, as discussed

fga/src/model.rs
leovalais: done, by yourself

leovalais: trait Type

docker-compose.yml
flomonster: I've seen that too. But it seems to me that it's just duplicating configs (and therefore a source of errors).

flomonster: We actually can't do this since there is no `/bin/sh` in the docker image...

docker-compose.yml
flomonster: Not possible

fga/src/doctest_setup.rs
leovalais: [7d6841c](https://github.com/OpenRailAssociation/osrd/pull/10504/commits/7d6841c261622196c47dd8385ff0a2c26b3c61e9)

fga/src/model.rs
leovalais: 4279493c3

fga/src/model.rs
leovalais: e5416a8db

fga/src/model.rs
leovalais: Recap of what we discussed on call.

Take `query_objects` for instance. You may want to query the documents which can be read by `person:jean` (**one** subject), or by the members of my friend group `group:friends#member` (**several** subjects), or even publicly `person:*` (**many** subjects). Therefore we want the `query_user` functions to accept these three forms.

The trait `User` is too restrictive for that as it only represents the first form `type:id`. That's why the trait `AsUser<User: User>` exists. `AsUser<User = Person>` is implemented by `UserSet<R: Relation> where R::User = Person` for the second form and by `Wildcard<Person>` for the third.

sim51:
I tested the docker integration, and so far it's good (in `dev-front` and `host` mode).

As discussed with @flomonster , if you already have a database with some data and you start the stack, you will have some SQL errors from the container `openfga-migrate`. 
To solve it, you need to apply this SQL script :
```
create schema openfga;
grant all privileges on schema public to osrd;
```

On a fresh install of this stack (I tested it), the issue is not present: those commands are directly executed by the script `docker/init_db.sql`  which is ran at db creation.

woshilapin (approved):
Thanks for all the work on this PR. There are a bunch of unchecked items in the description, do we need to put them in an issue?

leovalais:
> There are a bunch of unchecked items in the description, do we need to put them in an issue?

Probably better, but we'll need those to be done eventually to have the full permission system we envision to work.

docker/docker-compose.pr-tests.yml
```diff
@@ -94,10 +119,9 @@ services:
       valkey: { condition: service_healthy }
       rabbitmq: { condition: service_healthy }
     restart: unless-stopped
-    ports: [ "8091:8091" ]
     environment:
-      EDITOAST_PORT: 8091
-      ROOT_URL: "http://localhost:8091"
+      EDITOAST_PORT: 8090
+      ROOT_URL: "http://osrd-editoast-pr-tests:8090"
```
Khoyo: pr-tests ports cannot overlap the main stack ports

Khoyo (changes requested):

docker/docker-compose.pr-tests.yml
flomonster: I don't expose ports anymore so it works since it's not the same network.

Khoyo: For consistency, could we expose nothing but the gateway then?

(*Ideally*, the port shuffling and exposition change would be done in a separate commit, but it's not that important)

flomonster: I moved editoast port from 8090 to 8089 so it can be exposed again.

flomonster: I also moved the diff in another commit.

editoast/Dockerfile
```diff
@@ -40,6 +40,16 @@ RUN rustup component add llvm-tools && \
     rustup component add rustfmt && \
     rustup component add clippy && \
     cargo install grcov
+
+# install `fga` binary used in `crate fga` unit tests
+RUN mkdir install \
+    && cd install \
+    && wget https://github.com/openfga/cli/releases/download/v0.6.4/fga_0.6.4_linux_amd64.tar.gz \
```
Khoyo, line 47: Low priority, but we should add a github check for updates on that

Khoyo (approved):
LGTM

editoast/Dockerfile
Khoyo: #10953

PR #10506, opened 2025-01-23
core: stdcm: account for previous margins when checking for new ones

### LINKED ISSUE #10505: core: stdcm: weird behavior on engineering allowance ranges
### What happened?

See https://github.com/OpenRailAssociation/osrd/pull/10488

We sometimes have consecutive / overlapping engineering allowances, where the start of an allowance range goes further back than the previous allowance, which isn't supposed to be possible. 

_No response_

Log the ranges

Local

Firefox

c2228f02b41e476ff1c6e0767e9ae5dbd554654a

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/10505


Erashin (approved):
LGTM.

PR #10513, opened 2025-01-23
editoast: hash train path to sort train schedules

younesschrifi:


editoast/src/views/train_schedule.rs
```diff
@@ -385,11 +386,17 @@ pub async fn train_simulation_batch(
     infra: &Infra,
     electrical_profile_set_id: Option<i64>,
 ) -> Result<Vec<(SimulationResponse, PathfindingResult)>> {
-    // Compute path
 
-    let train_batches = train_schedules.chunks(TRAIN_SIZE_BATCH);
+    // Sort train_schedules and create batches
```
flomonster: We could add some documentation on why we do this.

flomonster (approved):

editoast/src/views/train_schedule.rs
```diff
@@ -385,9 +386,15 @@ pub async fn train_simulation_batch(
...
-    let train_batches = train_schedules.chunks(TRAIN_SIZE_BATCH);
+    // Sort train_schedules by the hash of their path_item to group schedules with identical or similar paths together.
+    // This increases the likelihood of identical pathfinding and simulation results, optimizing cache usage and retrieval.
+    let mut sorted_train_schedules = train_schedules.to_vec();
+    sorted_train_schedules.sort_by(|a, b| {
+        let hash_a = path_input_hash(&a.path);
+        let hash_b = path_input_hash(&b.path);
+        hash_a.cmp(&hash_b)
+    });
+    let train_batches = sorted_train_schedules.chunks(TRAIN_SIZE_BATCH);
```
woshilapin, line 397: This comparison closure is possibly used multiple times on each item. This means each item is going to be hashed multiple times. Would it make sense to change this for something where each item is hashed only once?

woshilapin (approved):

editoast/src/views/train_schedule.rs
flomonster: I'm not sure I understand your proposal and what we can expect to gain.

woshilapin: I don't know how heavy is it to hash a list of path items, but doing it multiple time for the same train schedule seems like a waste of time. One way to not do it multiple times is to do something like this.

```rust
let sorted_train_schedules: BTreeMap<u64, &TrainSchedule> = train_schedules
  .iter()
  .map(|train_schedule| (path_input_path(&train_schedule.path), train_schedule))
  .collect();
```

With this, each train schedule is hashed only once (but you don't have a `Vec` in the end). Feel free to disregard if you believe the waste of computing time is negligeable.

flomonster:
> [!NOTE]
> This PR does not work since the path item id is part of the hash.

PR #10515, opened 2025-01-23
front: get correct coordinates of uic pathsteps

### LINKED ISSUE #8901: Waypoint added is not displayed correctly on the map
### What happened?

https://github.com/user-attachments/assets/823484c0-b51d-4a2b-8adc-3a648f1ee06a

When creating a path with vias as operational points, they're not displayed with the right position on the path

I expect the displayed position of the via to be the same as the operational point

1. Create a scenario on small_infra
2. Ask for a path going through 3 operational points (I used WS MES NES)
3. Witness that the via is not displayed where the operational point is

Dev (SNCF)

Firefox

8f3a2f4

clarani:
closes #8901 

See commits

The markers were not correctly placed because the calculations were wrong: we could not use a simple rule of three to find the correct point in the path geometry.

### DIFF
--- a/front/src/modules/pathfinding/helpers/findTrackSectionOffset.ts
+++ b/front/src/modules/pathfinding/helpers/findTrackSectionOffset.ts
@@ -1,0 +1,29 @@
+import type { TrackRange } from 'common/api/osrdEditoastApi';
+
+/**
+ * Return the track offset corresponding to the given position on path, composed of the track section id
+ * and the offset on this track section in mm
+ */
+const findTrackSectionOffset = (
+  positionOnPath: number, // in mm
+  tracksLengthCumulativeSums: number[], // in mm
+  trackRangesOnPath: TrackRange[]
+) => {
+  const index = tracksLengthCumulativeSums.findIndex(
+    (cumulativeSum) => positionOnPath <= cumulativeSum
+  );
+  const trackRange = trackRangesOnPath[index];
+  if (!trackRange) return null;
+
+  // compute offset
+  const inferiorSum = index > 0 ? tracksLengthCumulativeSums[index - 1] : 0;
+  const offsetOnTrackRange = positionOnPath - inferiorSum;
+  const offsetOnTrackSection =
+    trackRange.direction === 'START_TO_STOP'
+      ? trackRange.begin + offsetOnTrackRange
+      : trackRange.end - offsetOnTrackRange;
+
+  return { track: trackRange.track_section, offset: offsetOnTrackSection };
+};
+
+export default findTrackSectionOffset;
--- a/front/src/modules/pathfinding/helpers/findTrackSectionOffset.ts
+++ b/front/src/modules/pathfinding/helpers/findTrackSectionOffset.ts
@@ -13,7 +13,10 @@ const findTrackSectionOffset = (
     (cumulativeSum) => positionOnPath <= cumulativeSum
   );
   const trackRange = trackRangesOnPath[index];
-  if (!trackRange) return null;
+
+  if (!trackRange) {
+    throw new Error('No track range found for the given position on path');
+  }
 
   // compute offset
   const inferiorSum = index > 0 ? tracksLengthCumulativeSums[index - 1] : 0;
--- a/front/src/modules/pathfinding/helpers/getPointOnPathCoordinates.ts
+++ b/front/src/modules/pathfinding/helpers/getPointOnPathCoordinates.ts
@@ -22,9 +22,9 @@ const getPointOnPathCoordinates = (
     trackRanges
   );
 
-  const track = tracks[trackOffset!.track];
+  const track = tracks[trackOffset.track];
 
-  return getPointOnTrackCoordinates(track.geo, mToMm(track.length), trackOffset!.offset);
+  return getPointOnTrackCoordinates(track.geo, mToMm(track.length), trackOffset.offset);
 };
 
 export default getPointOnPathCoordinates;
--- a/front/src/modules/pathfinding/helpers/getPointOnPathCoordinates.ts
+++ b/front/src/modules/pathfinding/helpers/getPointOnPathCoordinates.ts
@@ -1,0 +1,30 @@
+import type { Position } from 'geojson';
+
+import type { TrackRange, TrackSection } from 'common/api/osrdEditoastApi';
+import { getPointOnTrackCoordinates } from 'utils/geometry';
+import { mToMm } from 'utils/physics';
+
+import findTrackSectionOffset from './findTrackSectionOffset';
+
+/**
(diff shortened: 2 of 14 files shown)


front/src/modules/pathfinding/helpers/getTrackLengthCumulativeSums.ts
```diff
@@ -0,0 +1,22 @@
+import type { TrackRange } from 'common/api/osrdEditoastApi';
+
+/**
+ * Given a list a trackRanges, return the array of cumulative sums of the lengths of the track ranges.
```
SharglutDev:
```suggestion
 * Given a list of track ranges, return the array of cumulative sums of the lengths of the track ranges.
```

front/src/modules/pathfinding/helpers/findTrackSectionOffset.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Return the track offset corresponding to the given position on path, composed of the track section id
+ * and the offset on this track section in mm
+ */
+const findTrackSectionOffset = (
+  positionOnPath: number, // in mm
+  tracksLengthCumulativeSums: number[], // in mm
+  trackRangesOnPath: TrackRange[]
+) => {
+  const index = findTrackSectionIndex(positionOnPath, tracksLengthCumulativeSums);
```
SharglutDev: Wouldn't it work to just do that ?
```suggestion
  const index = tracksLengthCumulativeSums.findIndex(
    (cumulativeSum) => positionOnPath <= cumulativeSum
  );
```

front/src/modules/pathfinding/helpers/__tests__/findTrackSectionOffset.ts
SharglutDev, line 1: we should keep this file extension `.spec.ts` otherwise the test won't run I think.

SharglutDev, line 1: Would also be nice to have a test like this 

```
  it('should correctly find the track offset if the infSum is 0', () => {
    const trackRangesLengthCumulativeSums = [1000, 2000, 3000, 4000, 5000];
    const trackRanges = [
      { track_section: 'track_0', begin: 500, end: 1500, direction: 'START_TO_STOP' },
      { track_section: 'track_1' },
      { track_section: 'track_2', begin: 60, end: 1060, direction: 'STOP_TO_START' },
      { track_section: 'track_3' },
      { track_section: 'track_4' },
    ] as TrackRange[];

    const offsetOnPath = 900;
    const result = findTrackSectionOffset(
      offsetOnPath,
      trackRangesLengthCumulativeSums,
      trackRanges
    );

    expect(result).toEqual({ track: 'track_0', offset: 1400 });
  });
```

front/src/modules/pathfinding/helpers/getPointOnPathCoordinates.ts
```diff
@@ -0,0 +1,28 @@
...
+  tracksLengthCumulativeSums: number[],
+  positionOnPath: number
+) => {
+  const trackOffset = findTrackSectionOffset(
+    positionOnPath,
+    tracksLengthCumulativeSums,
+    trackRanges
+  );
+
+  const track = tracks[trackOffset!.track];
```
SharglutDev: Is it safe to use `!` ? Should we handle the case if `trackOffset` is null ?

front/src/modules/pathfinding/helpers/getPointOnPathCoordinates.ts
SharglutDev, line 1: nit: maybe it would be better to have commit n8 before commit n3 ? But you might have some conflicts and would need to rename all the other commits, so not compulsory :)

SharglutDev (commented):
Work perfectly ! Just some small comments

front/src/modules/pathfinding/helpers/findTrackSectionOffset.ts
clarani: Yes you're right, it's better like this :)

front/src/modules/pathfinding/helpers/__tests__/findTrackSectionOffset.ts
clarani: Good catch thanks !

clarani: Nice suggestion, it's added :)

front/src/modules/pathfinding/helpers/getPointOnPathCoordinates.ts
clarani: I've modified the function `findTrackSectionOffset`: it will always return a trackOffset now, and throw if it does not find any.
I removed the ! :+1:

front/src/modules/pathfinding/helpers/getPointOnPathCoordinates.ts
clarani: I can do it once you validated all the other changes :+1:

front/src/modules/pathfinding/helpers/getTrackLengthCumulativeSums.ts
SharglutDev: Still missing the change from `a` to `of` :)

SharglutDev:
Lgtm and tested ! You can fix the remaining comments and conflicts and I'll give it another testing before approve when you're done :)

front/src/modules/pathfinding/helpers/getTrackLengthCumulativeSums.ts
clarani: It's done :)

front/src/modules/pathfinding/helpers/getPointOnPathCoordinates.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Compute the coordinates of a point on a path from its path offset
+ * and the list of tracks (track sections and ranges) composing the path.
+ */
+const getPointOnPathCoordinates = (
+  tracks: Record<string, TrackSection>,
+  trackRanges: TrackRange[],
+  tracksLengthCumulativeSums: number[],
+  positionOnPath: number
+) => {
```
Synar: Nitpick : could you add the explicit Point return type (as it is done in getPointOnPathCoordinates)?

Synar (approved):
Left a minor nitpick comment, but nice changes, LGTM!

front/src/modules/pathfinding/helpers/getPointOnPathCoordinates.ts
clarani: It's done :white_check_mark:

SharglutDev (approved):
Lgtm and tested, great job !

PR #10516, opened 2025-01-23
front: fix api and network errors not being caught and displayed in infra editor

### LINKED ISSUE #10514: front: infrastructure deletion fails silently with a success message
### What happened?

When an infrastructure fails to be deleted, a success message is improperly displayed instead of an error message

https://github.com/user-attachments/assets/caf830a0-b570-4448-b6ec-73a064c00312

_No response_

1. Create an undeletable infrastructure
2. Go to infra editor
3. Try to delete it

Local

Firefox

dev

Synar:
Close #10514


SharglutDev (approved):
Lgtm and tested, good job :)

PR #10518, opened 2025-01-23
front: delete infra at the end of e2e-tests

### LINKED ISSUE #10517: front e2e-tests: small_infra_test_e2e is never deleted
### What happened?

Running the e2e-tests creates an infra called small_infra_test_e2e, but does not delete it, making future tests on the same environment more brittle

A function exists for this in e2e-tests, but it is never called, and thus constitutes dead code

_No response_

1. Run the e2e-tests
2. List the infrastucture and notice a new infra called small_infra_test_e2e

Local

Firefox

dev

Synar:
Close #10517

Feel free to refuse the PR if there is a very good reason not to delete the infra

### DIFF
--- a/front/tests/global-teardown.ts
+++ b/front/tests/global-teardown.ts
@@ -45,6 +45,10 @@ teardown('teardown', async ({ browser }) => {
       force: true,
     });
 
+    // Delete infra, unless it is currently used as a foreign key by the stdcm env in the database
+    if (savedEnvironment && savedEnvironment.infra_id !== Number(process.env.TEST_INFRA_ID))
+      await deleteInfra(infrastructureName);
+
     logger.info('Test data teardown completed successfully.');
   } catch (error) {
     logger.error('Error during test data teardown:', error);
--- a/front/tests/global-teardown.ts
+++ b/front/tests/global-teardown.ts
@@ -4,11 +4,12 @@ import { test as teardown } from '@playwright/test';
 
 import ROLLING_STOCK_NAMES, {
   globalProjectName,
+  infrastructureName,
   trainScheduleProjectName,
 } from './assets/project-const';
 import { logger } from './logging-fixture';
 import { setStdcmEnvironment } from './utils/api-setup';
-import { deleteProject, deleteRollingStocks } from './utils/teardown-utils';
+import { deleteInfra, deleteProject, deleteRollingStocks } from './utils/teardown-utils';
 
 teardown('teardown', async ({ browser }) => {
   try {
@@ -38,6 +39,11 @@ teardown('teardown', async ({ browser }) => {
     } else {
       logger.warn('No STDCM environment to restore.');
     }
+
+    fs.rmSync('./tests/test-saved-environment/savedStdcmEnvironment.json', {
+      recursive: true,
+      force: true,
+    });
 
     logger.info('Test data teardown completed successfully.');
   } catch (error) {
--- a/front/tests/utils/setup-utils.ts
+++ b/front/tests/utils/setup-utils.ts
@@ -1,3 +1,5 @@
+import { promises } from 'fs';
+
 import type {
   Infra,
   PostInfraRailjsonApiResponse,
@@ -9,7 +11,13 @@ import type {
   StudyCreateForm,
 } from 'common/api/osrdEditoastApi';
 
-import { getApiRequest, getInfra, postApiRequest, setStdcmEnvironment } from './api-setup';
+import {
+  getApiRequest,
+  getInfra,
+  getStdcmEnvironment,
+  postApiRequest,
+  setStdcmEnvironment,
+} from './api-setup';
 import readJsonFile from './file-utils';
 import createScenario from './scenario';
 import { sendTrainSchedules } from './trainSchedule';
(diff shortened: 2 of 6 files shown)


Yohh (commented):
seems ok (and obvious) for me, I didn't even notice this disappeared until now, I wait for @Maymanaf to know why this is missing

Maymanaf:
The infrastructure is used for STDCM tests, so it cannot be deleted, and the teardown will always fail

Synar:
> The infrastructure is used for STDCM tests, so it cannot be deleted, and the teardown will always fail

Thanks! I had renamed all the infras in my db, so the setup and teardown were working fine for me.

I did notice I had an undeletable infra because it was used in an stdcmsearchcontext. Do you know when this is used, and whether we could define and destroy it each time?

Right now the tests are dependent on a specific state of the database, which make them more brittle. I would rather minimize such dependence if possible.

Maymanaf:
> > The infrastructure is used for STDCM tests, so it cannot be deleted, and the teardown will always fail
> 
> Thanks! I had renamed all the infras in my db, so the setup and teardown were working fine for me.
> 
> I did notice I had an undeletable infra because it was used in an stdcmsearchcontext. Do you know when this is used, and whether we could define and destroy it each time?
> 
> Right now the tests are dependent on a specific state of the database, which make them more brittle. I would rather minimize such dependence if possible.

The issue lies in the backend implementation. Currently, there is no dedicated endpoint available for deleting the `stdcm_search_environment`. The API only provides `POST` and `GET` methods for this resource, without a `DELETE` endpoint, it becomes impossible to directly remove the infra from stdcm

Synar:
Thanks, I'll look into it

front/tests/utils/setup-utils.ts
```diff
@@ -139,6 +146,42 @@ export async function createStudy(projectId: number, studyName = globalStudyName
   return study;
 }
 
+/**
+ * Loads and saves the stdcm environment to ensure it is not erased by the e2e-tests environment
```
Maymanaf:
```suggestion
 * Load and save the stdcm environment to ensure it is not erased by the e2e-tests environment
```

front/tests/utils/setup-utils.ts
```diff
@@ -139,6 +146,42 @@ export async function createStudy(projectId: number, studyName = globalStudyName
...
+        // The only way this can occur normally is if the tests were interrupted before teardown.
+        stdcmEnvironment = readJsonFile(savedStdcmEnvFilePath);
+      } catch (error: unknown) {
+        if (!(error instanceof Error && 'code' in error && error.code === 'ENOENT')) {
+          throw error; // Rethrow errors other than file not found
+        }
+      }
+    }
+    // Finally we put the current stdcm env in an env variable in prevision of the teardown,
+    // where we will delete the save file and put the stdcm env back into the database
```
Maymanaf:
```suggestion
    // where we will delete the saved file and put the stdcm env back into the database
```

front/tests/utils/setup-utils.ts
```diff
@@ -139,6 +146,42 @@ export async function createStudy(projectId: number, studyName = globalStudyName
...
+    process.env.STDCM_ENVIRONMENT = JSON.stringify(stdcmEnvironment);
+  } catch (error) {
+    console.error('Error handling saved STDCM environment: ', error);
```
Maymanaf:
```suggestion
    logger.error('Error handling saved STDCM environment: ', error);
```

Maymanaf (approved):
Great job! LGTM✅ 
Lefs some nit comments you can resovle when it's done

front/tests/global-teardown.ts
```diff
@@ -39,6 +40,15 @@ teardown('teardown', async ({ browser }) => {
       logger.warn('No STDCM environment to restore.');
     }
 
+    fs.rmSync('./tests/test-saved-environment/savedStdcmEnvironment.json', {
+      recursive: true,
+      force: true,
+    });
```
emersion, line 46: Hm, do we really need to write a file on disk? Can't we use a global variable?

Writing to `front/tests` is not great, e.g. in my container it's read-only.

emersion:
Maybe we could mock HTTP responses for the STDCM env if we're running locally instead? (ie, not in CI)

See https://playwright.dev/docs/mock#mock-api-requests

front/tests/global-teardown.ts
Synar: The file writing is an additional security, it is also written to process.env.

The reason I added this was to prevent the env from being lost if the tests are interrupted before finishing the teardown.

Synar: So if the tests/saved env folder/volume is not mounted, the tests should still proceed normally (as they do in the ci, where we're not mounting either), and you won't lose your previous env as long as you do not interrupt the tests.

emersion: Won't playwright execute the global teardown when the tests are interrupted?

Synar:
This could work, but mocking in the e2e tests seemed quite controversial last time it was discussed. Plus introducing significant differences in behavior between local and ci tests seems risky.

The best solution would probably be to have the e2e tests deal with a fully isolated database while running locally, but I'm unsure how we could do that.

front/tests/global-teardown.ts
Synar: Nope. There may be a way to change that, but all I found was a sad stack overflow question sitting with a couple upvotes and no answers.

Maymanaf:
I’m currently facing some setup issues, so I haven’t been able to test it yet. However, we could try using:
```
process.on('SIGINT', () => {
    /* Perform the teardown here */
    process.exit();
});

```
Or, to handle any possible interruption:
```
process.on('exit', () => {
    /* Perform the teardown here */
    process.exit();
});
```
I’m not sure if this is a recommended approach, but it could be a potential solution to force the teardown

ghost (approved):

front/tests/utils/setup-utils.ts
Synar: Done ^^

front/tests/utils/setup-utils.ts
Synar: Thanks, done!

front/tests/utils/setup-utils.ts
Synar: Done ^^

front/tests/utils/setup-utils.ts
Synar: I also added a commit to use logger instead of console in a sibling file

Synar:
> I’m currently facing some setup issues, so I haven’t been able to test it yet. However, we could try using:
> 
> ```
> process.on('SIGINT', () => {
>     /* Perform the teardown here */
>     process.exit();
> });
> ```
> 
> Or, to handle any possible interruption:
> 
> ```
> process.on('exit', () => {
>     /* Perform the teardown here */
>     process.exit();
> });
> ```
> 
> I’m not sure if this is a recommended approach, but it could be a potential solution to force the teardown

I like this idea (although I'm not sure it's recommended either), but I couldn't get it to work. Have you had any success?

front/tests/global-teardown.ts
emersion: Still not a fan of writing stuff to disk, will leave it up to others to approve.

Synar: Worth noting that writing files to the disk is something the e2e tests already do, for example to check the simulation reports. So this would simply be another file created during the tests and deleted at teardown.

emersion: Tests write to a "results" directory. Here we're talking about persisting state across test runs.

Synar: Not across tests runs in general, as the file gets deleted at teardown. So it would only propagate until a test does not get interrupted.

On the other hand, the current behavior on dev is that test data is propagated forever across test runs in the db, even without test interruption (and if the tests do get interrupted we permanently lose preexisting db data).

Synar: Eventually it would be better to actually avoid replacing that data at all in the db, but that would require a much bigger change.

Synar: Of course if maymanaf solution of managing to run teardown even on singint works that would be better too ^^

But otherwise I think this would be better than the current behavior, even if not perfect

kmer2016 (approved):
I find the proposed solution acceptable given the current constraints. While writing to disk may pose challenges in certain local environments, it provides necessary protection against test interruptions

Synar:
Merging for now, once a better solution is found we can remove this work around ^^

PR #10526, opened 2025-01-24
editoast: ensure spans are created in tests regardless of log level

hamz2a:
### Problem:
When running tests with `RUST_LOG=warn` or `RUST_LOG=error`, the `axum-tracing-opentelemetry` [crate doesn't create spans if the log level is below `info`](https://github.com/OpenRailAssociation/osrd/pull/10113). This causes tests of `views::stdcm_logs` to fail because there is no `trace_id` (span) generated.

### Solution:
To fix this, we modified the `create_tracing_subscriber` function to ignore the `RUST_LOG` environment variable during tests. This ensures that spans (`trace_id`) are always created, even when the log level is set to `warn` or `error`, allowing the tests to run correctly.

### DIFF
--- a/editoast/editoast_common/src/tracing.rs
+++ b/editoast/editoast_common/src/tracing.rs
@@ -32,12 +32,19 @@ pub struct TracingConfig {
 pub fn create_tracing_subscriber<T: SpanExporter + 'static>(
     tracing_config: TracingConfig,
     log_level: tracing_subscriber::filter::LevelFilter,
+    ignore_rust_log: bool,
     exporter: T,
 ) -> impl tracing::Subscriber {
-    let env_filter_layer = tracing_subscriber::EnvFilter::builder()
+    let env_filter_builder = tracing_subscriber::EnvFilter::builder()
         // Set the default log level to 'info'
-        .with_default_directive(log_level.into())
-        .from_env_lossy();
+        .with_default_directive(log_level.into());
+
+    let env_filter_layer = if ignore_rust_log {
+        env_filter_builder.parse_lossy("")
+    } else {
+        env_filter_builder.from_env_lossy()
+    };
+
     let fmt_layer = tracing_subscriber::fmt::layer()
         .pretty()
         .with_file(true)
--- a/editoast/src/client/runserver.rs
+++ b/editoast/src/client/runserver.rs
@@ -46,7 +46,7 @@ pub struct RunserverArgs {
     /// If this option is set, logging for the STDCM will be enabled.
     /// When enabled, relevant logs will be captured to aid in debugging and monitoring.
     #[clap(long, env = "ENABLE_STDCM_LOGGING", default_value_t = true)]
-    enable_stdcm_logging: bool,
+    pub(crate) enable_stdcm_logging: bool,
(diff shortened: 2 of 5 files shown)


editoast/src/client/runserver.rs
```diff
@@ -46,7 +46,7 @@ pub struct RunserverArgs {
     /// If this option is set, logging for the STDCM will be enabled.
     /// When enabled, relevant logs will be captured to aid in debugging and monitoring.
     #[clap(long, env = "ENABLE_STDCM_LOGGING", default_value_t = true)]
-    enable_stdcm_logging: bool,
+    pub(crate) enable_stdcm_logging: bool,
```
bougue-pe, line 49: Why do you need this `pub`?

woshilapin:
To give more precisions, we evaluated another solution with forcing the directive `otel::tracing=info` when configuring the tracer. It's not ideal because doing so is not deactivable afterwards (it takes higher precedence than `RUST_LOG`).

The rational for choosing this solution was: this is not an application problem, this is a "test" problem since we do not control enough the test environment (configuring `RUST_LOG` may break the tests). That the reason why we chose to give the test, the power to not take `RUST_LOG` into account.

editoast/editoast_common/src/tracing.rs
```diff
@@ -32,12 +32,19 @@ pub struct TracingConfig {
...
 ) -> impl tracing::Subscriber {
-    let env_filter_layer = tracing_subscriber::EnvFilter::builder()
+    let env_filter_builder = tracing_subscriber::EnvFilter::builder()
         // Set the default log level to 'info'
-        .with_default_directive(log_level.into())
-        .from_env_lossy();
+        .with_default_directive(log_level.into());
+
+    let env_filter_layer = if ignore_rust_log {
+        env_filter_builder.parse_lossy("")
```
bougue-pe, line 43: After discussing with @woshilapin, Opening a non-mandatory conversation after comparing `RUST_LOG=warn,otel::tracing=info cargo test` on `dev` with `RUST_LOG=warn cargo test` on this branch: both are working as for tests.
The "current branch case" still logs a lot more on affected tests.

So here are a few ideas (non-blocking, can be ignored because "not worth the time"):
1. just add `otel::tracing=info` instead of ignoring RUST_LOG for tests?
2. maybe really always force `otel::tracing=info` (except explicit, which is probably a pain to parse)
3. use INFO log level for effected tests (instead of current DEBUG)

bougue-pe (commented):
✅ Thanks for digging on this, after test it works for me (not approving as I'm not confident on how it interfaces with the feature, didn't dig on this).

editoast/src/client/runserver.rs
woshilapin: Probably a left over from previous modifications when we experimented different solutions, we don't need that anymore I believe.

editoast/editoast_common/src/tracing.rs
woshilapin: One solution we explored was to always have `otel::tracing=info` enabled. But this makes it not possible to deactivate through configuration. The only place where we want it activated is for tests where we want to control our test environment. This is something we didn't explore with @hamz2a and I made #10531 as an alternative proposition.

bougue-pe (approved):

hamz2a:
https://github.com/OpenRailAssociation/osrd/pull/10531

Issue #10452, opened 2025-01-20
STDCM queries take too much time


flomonster:
### What happened?

See the traces of an stdcm query:

![Image](https://github.com/user-attachments/assets/5b04b854-8066-4c4c-80fb-c20e1e5363ca)

Private datalog link to [the trace](https://claps.datadoghq.eu/apm/traces?query=service%3Aosrd-editoast%20%40deployment.environment.name%3Aosd-prod%20stdcm&agg_m=count&agg_m_source=base&agg_t=count&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&fromUser=false&graphType=waterfall&historicalData=false&messageDisplay=inline&panel_tab=flamegraph&query_translation_version=v0&shouldShowLegend=true&sort=desc&spanID=1280736463641084366&spanType=all&spanViewType=metadata&storage=hot&timeHint=1737391622862&trace=286e95605c7341863da96998498b7f4510254662739427355806&traceID=286e95605c7341863da96998498b7f45&view=spans&start=1737391780016&end=1737392680016&paused=false).

- [ ] Add `tracing::instrument` for critical functions.
- [ ] Compress the data to valkey
- [ ] Optimize the psql train schedule retrieve query
- [ ] Optimize the `PathItemCache::load` function (unnecessary queries when we hit cache 100%)
    - Skip query if no items
    - Use HashSet instead to avoid duplicates 
- [ ] Valkey queries seem to be executed sequentially. We must fix it so each batch can run in parallel.
    - `Tokio(ConnectionManager)` can handle only one physical connection.
    - We could use [deadpool-redis](https://crates.io/crates/deadpool-redis)

### What did you expect to happen?

Retrieving the simulation from valkey should be way more efficient.

- Check that the filter is working properly
- Check the size of the payload (is it transfer time)

Storing the stdcm log takes too much time:
- We are supposed to spawn a worker to insert in DB the input and output of the query.

### How can we reproduce it (as minimally and precisely as possible)?

1. Do a stdcm with a real timetable and infra
2. Observe traces

### On which environments the bug occurs?

Production (SNCF)

### On which browser the bug occurs?

Firefox

### OSRD version (top right corner `Account` button > `Informations`)

c9066aa

flomonster:
## STDCM log

| Field  | Size in DB (bjson) | Size Deserialized |
|--------|--------------------|-------------------|
[...]

PR #10527, opened 2025-01-26
Compress editoast cached data

younesschrifi:
The json request and response from the pathfinding and the simulation take too much time to be get from valkey or push in valkey. The solution is to use gzip to compress the payloads.

## Benchmarks

All benchmark were run on dev.

### No compression (`BATCH_SIZE=500`)

- LMR: ~20s
- Get from cache (total): 9.4s
- Get from cache (sum): 14.0s

image

### Zstd -1 (`BATCH_SIZE=500`)

- LMR: ~20s
- Get from cache (total): 13.7s
- Get from cache (sum): 22.6s

image

### LZ4 -1 (`BATCH_SIZE=500`)

- LMR: ~14.4s
- Get from cache (total): 8.8s
- Get from cache (sum): 11.9s

image

### DIFF
--- a/editoast/Cargo.lock
+++ b/editoast/Cargo.lock
@@ -1318,6 +1318,7 @@ dependencies = [
  "itertools",
  "json-patch",
  "lapin",
+ "lz4_flex",
  "mime",
  "mvt",
  "opentelemetry",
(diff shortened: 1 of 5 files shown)


woshilapin (approved):
Hopes this improve what we're looking for. Do you have some numbers? It would be nice to put them in the description so future us can find these when reevaluating the situation.

flomonster:
> Hopes this improve what we're looking for. Do you have some numbers? It would be nice to put them in the description so future us can find these when reevaluating the situation.

We're benchmarking PR. I will add numbers in live :)

leovalais:
Wouldn't reducing the batch size improve the performances since transport is likely the limiting factor here?

flomonster (approved):

flomonster:
We can merge this PR. We've made a slight improvement. The result is not very visible because of other problems I've listed [here](https://github.com/OpenRailAssociation/osrd/issues/10452).

PR #10528, opened 2025-01-27
core: in LoA case, add maintain curve to indication curve before finding intersection with overlay envelope

### LINKED ISSUE #10443: core: etcs: loa bug: add maintain curve to indication curve before calling keepBrakingCurverUnderOverlay
Image

In this case, with the current behavior, indication does not intersect with the mrsp, and the code will throw an exception. What we should do is concatenate the indication curve AND the maintain curve immediately, then call the method `keepBrakingUnderOverlay`

Erashin:
Fixes #10443.

### DIFF
--- a/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
+++ b/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
@@ -13,6 +13,8 @@ import fr.sncf.osrd.envelope_sim.*
 import fr.sncf.osrd.envelope_sim.overlays.EnvelopeDeceleration
 import kotlin.math.max
 import kotlin.math.min
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
 
 /**
  * Formulas are found in `SUBSET-026-3v400.pdf` from the file at
@@ -34,6 +36,8 @@ enum class BrakingType {
     ETCS_SBD,
     ETCS_GUI
 }
+
+val etcsBrakingCurvesLogger: Logger = LoggerFactory.getLogger("EtcsBrakingCurves")
 
 /** Compute braking curves at every end of authority. */
 fun addBrakingCurvesAtEOAs(
@@ -84,12 +88,10 @@ fun addBrakingCurvesAtEOAs(
         assert(fullIndicationCurve.endPos == targetPosition)
         assert(fullIndicationCurve.endSpeed == targetSpeed)
 
-        // Indication curve for EOAs will always have at least 2 points with positive positions.
-        // This method cannot return a null.
         val indicationCurve =
-            keepBrakingCurveUnderOverlay(fullIndicationCurve, envelope, beginPos)!!
-        assert(indicationCurve.beginPos >= beginPos)
-        assert(indicationCurve.endPos == targetPosition)
+            keepBrakingCurveUnderOverlay(Envelope.make(fullIndicationCurve), envelope, beginPos)
+                ?: continue
+        assert(indicationCurve.beginPos >= beginPos && indicationCurve.endPos == targetPosition)
         assert(
             indicationCurve.beginSpeed <= maxSpeedEnvelope &&
                 indicationCurve.endSpeed == targetSpeed
@@ -161,26 +163,11 @@ fun addBrakingCurvesAtLOAs(
 
         val fullIndicationCurve =
             computeIndicationBrakingCurveFromRef(context, ebiCurve, BrakingCurveType.EBI, guiCurve)
-        assert(fullIndicationCurve.endPos <= targetPosition)
+        val endOfIndicationCurve = fullIndicationCurve.endPos
+        assert(endOfIndicationCurve <= targetPosition)
         assert(fullIndicationCurve.endSpeed == targetSpeed)
 
-        val indicationCurve =
-            keepBrakingCurveUnderOverlay(
-                fullIndicationCurve,
-                envelopeWithLoaBrakingCurves,
-                beginPos
-            )
-
-        if (indicationCurve != null) {
-            assert(indicationCurve.beginPos >= 0.0 && indicationCurve.endPos <= targetPosition)
-            assert(
-                indicationCurve.beginSpeed <= maxSpeedEnvelope &&
-                    indicationCurve.endSpeed == targetSpeed
-            )
-            builder.addPart(indicationCurve)
-        }
-
-        val endOfIndicationCurve = indicationCurve?.endPos ?: beginPos
+        val fullIndCurveWithMaintain: Envelope
         if (endOfIndicationCurve < targetPosition) {
             // Maintain target speed until target position, i.e. LOA.
             val maintainTargetSpeedCurve =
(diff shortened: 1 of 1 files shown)


core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -344,21 +341,35 @@ private fun computeIndicationBrakingCurveFromRef(
...
+    if (fullBrakingCurve.endPos <= beginPos)
+        throw IllegalArgumentException(
+            "Expected fullBrakingCurve position range to intersect with [beginPos, +infinity[."
+        )
+    val positions =
+        fullBrakingCurve
+            .map { it.clonePositions() }
+            .reduce { mergedArray, currentArray ->
+                // Remove duplicate part transitions: the last point of the previous array is the
+                // first point of the next array.
```
bougue-pe: Can you mention that it duplicate points break the addStep of the ConstrainedEnvelopePartBuilder, please? 🙏 

nit: I'd also mutualize the comment.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -368,10 +379,11 @@ private fun keepBrakingCurveUnderOverlay(
             EnvelopeConstraint(overlay, EnvelopePartConstraintType.CEILING)
         )
     overlayBuilder.initEnvelopePart(positions[nbPoints - 1], speeds[nbPoints - 1], -1.0)
-    for (i in fullBrakingCurve.pointCount() - 2 downTo 0) {
+    for (i in nbPoints - 2 downTo 0) {
         if (!overlayBuilder.addStep(positions[i], speeds[i], timeDeltas[i])) break
     }
-    if (partBuilder.stepCount() <= 1) return null
+    if (partBuilder.stepCount() <= 1)
```
bougue-pe: `stepCount()` is based on `times`, and there is 1 less value than positions:
```suggestion
    if (partBuilder.isEmpty)
```
So this failed when there was a stop close from the start, and a curve with only 2 positions (19 first meters in my test).

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -344,21 +341,35 @@ private fun computeIndicationBrakingCurveFromRef(
...
     beginPos: Double
-): EnvelopePart? {
-    if (fullBrakingCurve.endPos <= beginPos) {
-        return null
-    }
-    val positions = fullBrakingCurve.clonePositions()
-    val speeds = fullBrakingCurve.cloneSpeeds()
-    val timeDeltas = fullBrakingCurve.cloneTimes()
-    val nbPoints = fullBrakingCurve.pointCount()
+): EnvelopePart {
```
bougue-pe: Especially for EoA and LoA, I would check before calling that there is no need to add a curve if no intersection.

Opinionated (so OK to dismiss): From the `keepBrakingCurveUnderOverlay` fn itself, it's a legitimate use IMO to have no intersection.
I would let the responsibility to the caller to check if it is an expected case.
```suggestion
): EnvelopePart? {
```

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -344,21 +341,35 @@ private fun computeIndicationBrakingCurveFromRef(
...
+    if (fullBrakingCurve.endPos <= beginPos)
+        throw IllegalArgumentException(
+            "Expected fullBrakingCurve position range to intersect with [beginPos, +infinity[."
```
bougue-pe: Why not rejecting already if there is only one position?
```suggestion
    if (fullBrakingCurve.endPos < beginPos)
        throw IllegalArgumentException(
            "Expected fullBrakingCurve position range to intersect with ]beginPos, +infinity[."
```

bougue-pe (commented):
We really need some testing on those cases at some point (unit would be better, could be integration).
Anyway, this patch is fine 👍

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Nice catch!! I just removed it due to another comment further along.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: To check whether or not there is an intersection between 2 envelopes is not something so easily done, especially when they're not both continuous and monotonous. However, yeah, as we talked about it, the current flow rn is not to throw exceptions, so I would rather return null here, and log a warning.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: ~Technically, none, it woulda just thrown later. I'll take it into an account.~
Wait a minute, this is actually correct, what do you mean? If the curve ends at beginPos, the only intersection would be beginPos, which is incorrect. The following check is more of a failsafe. I'd rather keep this one as is, and remove the next one, if that's ok with you.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -344,21 +346,38 @@ private fun computeIndicationBrakingCurveFromRef(
...
  */
 private fun keepBrakingCurveUnderOverlay(
-    fullBrakingCurve: EnvelopePart,
+    fullBrakingCurve: Envelope,
     overlay: Envelope,
     beginPos: Double
 ): EnvelopePart? {
     if (fullBrakingCurve.endPos <= beginPos) {
+        etcsBrakingCurvesLogger.warn(
+            "The ETCS braking curve ending at ($beginPos, ${fullBrakingCurve.endSpeed}) does not intersect with the overlay envelope."
```
bougue-pe: nit (as we had a misunderstanding about that in comments 😅):
```suggestion
            "The position-range of ETCS braking curve ending at ($beginPos, ${fullBrakingCurve.endSpeed}) does not intersect with the overlay envelope's position-range."
```

bougue-pe (approved):
Thanks!
(You'll need to rebase after some other work on `dev` branch)

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
bougue-pe: As discussed, the code was good, not the comment (and I only read the comment 😅).
And agreed and reviewed to remove the final check 👍

eckter (approved):

PR #10529, opened 2025-01-27
Fix gateway otlp config

flomonster:
- `service_name` does not work (we could delete it and use the default env var)
- ~~`figment::merge` seems to work the other way around~~
- We should move `OTEL_SERVICE_NAME` to docker files for all services

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -133,7 +133,6 @@ services:
       OSRDYNE_API_URL: "http://osrd-osrdyne:4242/"
       TELEMETRY_KIND: "opentelemetry"
       TELEMETRY_ENDPOINT: "http://jaeger:4317"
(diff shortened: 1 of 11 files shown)


woshilapin:
I've removed the configuration options for service name in `gateway` and `osrdyne`: let's use the standard `OTEL_SERVICE_NAME` environment variable for that. I've then set up that variable directly in the `Dockerfile` of the three Rust components.

Finally, I've remove the `enable` in `gateway`'s configuration as it's a leftover from b154bad3dc537f62a1a7e757e898896e19cf5295.

I still need to test more this `figment::merge` to understand what is going on since it looks really weird.

bloussou (approved):
LGTM 🚀

woshilapin (approved):

Khoyo (approved):

ghost (approved):

PR #10531, opened 2025-01-27
editoast: force 'otel::tracing=info' for some tests

woshilapin:
The STDCM log feature depends on Opentelemetry span being created in the Axum middleware. If `RUST_LOG=warn`, tests will fail because no span is created and therefore no trace ID is created and tests depends on this trace ID.

The added directives are applied after reading `RUST_LOG`.

This is an alternative solution to what was proposed in #10526.

### DIFF
--- a/editoast/src/views/test_app.rs
+++ b/editoast/src/views/test_app.rs
@@ -177,6 +177,7 @@ impl TestAppBuilder {
         let tracing_config = TracingConfig {
             stream: Stream::Stdout,
             telemetry,
+            directives: vec!["otel::tracing=info".parse().unwrap()],
         };
         let sub = create_tracing_subscriber(
             tracing_config,
--- a/editoast/src/views/test_app.rs
+++ b/editoast/src/views/test_app.rs
@@ -24,6 +24,7 @@ use opentelemetry_sdk::export::trace::SpanData;
 use opentelemetry_sdk::export::trace::SpanExporter;
 use serde::de::DeserializeOwned;
 use tower_http::trace::TraceLayer;
+use tracing_subscriber::filter::Directive;
 use url::Url;
 
 use crate::{
@@ -65,6 +66,7 @@ pub(crate) struct TestAppBuilder {
     enable_authorization: bool,
     enable_stdcm_logging: bool,
     enable_telemetry: bool,
+    telemetry_directives: Vec<Directive>,
     user: Option<UserInfo>,
     roles: HashSet<BuiltinRole>,
 }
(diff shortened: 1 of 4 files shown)


bougue-pe (commented):
✅ works for me a bit better (less logs) than the alternative.

It looks a bit more elegant, but again, not sharp enough on the matter.

bougue-pe (approved):

leovalais (approved):
I think I like this proposition better.

1. Keeping the ability to configure `RUST_LOG` for running tests is useful (to me at least).
2. Tests that require a specific log directives will be emphasized and show explicitly which directives affects the behavior of the tested feature.

editoast/src/views/test_app.rs
```diff
@@ -116,6 +119,11 @@ impl TestAppBuilder {
         self
     }
 
+    pub fn with_directive(mut self, directive: Directive) -> Self {
```
Khoyo: Maybe `with_telemetry_directive()`?

woshilapin: In reality, I'm all in to improve the naming, but now that you mention it, maybe it should be `.with_tracing_directive()`? What do you think?

leovalais: Agreed, maybe `with_rust_log_directive` since the format makes sense even without `tracing`?

(I'm fine with everything proposed here)

woshilapin: Note that `RUST_LOG` is not a rust standard. It is a variable environment used by `env_logger` first, then reused by `tracing-subscriber`: but even `tracing-subscriber`'s version is not fully compatible with `env_logger` ([it's actually more powerful in `tracing-subscriber`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives)). That said, I felt that `with_rust_log_directive()` was still the best proposition (`with_tracing_directive()` was maybe more correct but less explicit, since `tracing` is about a lot of things, not only `RUST_LOG`).

Done in [ff2acc0](https://github.com/OpenRailAssociation/osrd/compare/fa48f4c62fa16528bccf69237d1967aeb9866f13..ff2acc09de7f7ea14827a41549b2ee9198663159).

woshilapin:
Since #10526 has been abandoned in favor of the current PR, let's merge it?

editoast/src/views/test_app.rs
leovalais: Oh I didn't know that the lesser version of `RUST_LOG` wasn't standardized, but that's logic now that you mention why!

.osrd-compose-state
bougue-pe, line 1: Leftover to be removed

woshilapin: Done, thanks for catching that.

PR #10533, opened 2025-01-27
core: pathfinding: fix missing waypoints in edge cases

eckter:
Fix a pathfinding regression, we'd "pass" the waypoints without adding them to the result


JulesPeignier (approved):
Thanks a lot

Erashin (approved):
Agreed, this class has become hard to deal with. Anyway, gj.

Khoyo (approved):
Nice catch

PR #10342, opened 2025-01-13
python: fix maximum Python version

emersion:
Python 3.12 doesn't work:

    # poetry run pflake8 --config ./pyproject.toml
    Traceback (most recent call last):
      File "/root/.cache/pypoetry/virtualenvs/railjson-generator--f9Ha3Rv-py3.12/bin/pflake8", line 5, in <module>
        from pflake8.__main__ import main
      File "/root/.cache/pypoetry/virtualenvs/railjson-generator--f9Ha3Rv-py3.12/lib/python3.12/site-packages/pflake8/__init__.py", line 55, in <module>
        class DivertingSafeConfigParser(ConfigParserTomlMixin, configparser.SafeConfigParser):
                                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    AttributeError: module 'configparser' has no attribute 'SafeConfigParser'. Did you mean: 'RawConfigParser'?

See this upstream bug report:
https://github.com/pydata/pandas-datareader/issues/969

Our CI already checks out Python 3.11, so let's align our maximum version with what CI uses.

PR #10554, opened 2025-01-28
osrd-schema: bump python version

bloussou:
To be consistent with dags, etl and confluo


flomonster (approved):
LGTM

python/osrd_schemas/pyproject.toml
```diff
@@ -11,7 +11,7 @@ authors = ["OSRD <contact@osrd.fr>"]
 [tool.poetry.dependencies]
 geojson-pydantic = "^1.0.0"
 pydantic = "^2.1.1"
-python = ">=3.9,<3.12"
+python = "^3.11,<3.13"
```
emersion, line 14: I think this only matches 3.11.x: `^3.11` is a synonym for `>=3.11,<3.12`.

emersion: Hm, or is it? I suppose it's actually `>=3.11,<4`?

emersion (commented):
In my experience, osrd_schema doesn't work with Python 3.12: https://github.com/OpenRailAssociation/osrd/pull/10342

Have you tested it?

shenriotpro:
> In my experience, osrd_schema doesn't work with Python 3.12: #10342
> 
> Have you tested it?

I missed that PR, it looks like an issue with pflake8 rather than osrd-schemas. https://github.com/csachs/pyproject-flake8/issues/30

flomonster:
Legacy since we moved from poetry to uv.

PR #10557, opened 2025-01-28
misc: add .zed to .gitignore

leovalais:


emersion:
In general I don't like adding lists of editors in `.gitignore`, I recommend setting up a global ignore instead: https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

The reason is that this doesn't scale: each project needs to add each text editor and keep up with new editors.

But I see there are other editors in there already…

leovalais:
I'd agree, that's why I didn't push this sooner. But since others are here and that there's a bunch of us using zed already, why not? 🤷 

I'm completely fine with removing all editors config as well, but consistency and all :)

flomonster (approved):

PR #10561, opened 2025-01-29
front: enable power restrictions merge

### LINKED ISSUE #8850: front: merge several power restriction codes is not possible anymore
### What happened?

In simulation settings, I added different power restriction codes, and can even split a zone in two parts (with the scissors), but can't merge two parts. 

image

I expect to have a button "merge" that allows me to merge two different zone of power restriction.
This functionnality used to exist before but has disappeared.

You can follow the steps in https://github.com/osrd-project/osrd-tests/issues/14 until 13)
Example of rolling stock with power restriction codes : 1TGVDUPL

Recette (SNCF)

Google Chrome

staging bc8b913

clarani:
closes #8850

### DIFF
--- a/front/src/reducers/osrdconf/operationalStudiesConf/powerRestrictionReducer.ts
+++ b/front/src/reducers/osrdconf/operationalStudiesConf/powerRestrictionReducer.ts
@@ -96,6 +96,54 @@ const powerRestrictionReducer = {
     }
   },
 
+  mergePowerRestrictionRanges(
+    state: Draft<OperationalStudiesConfState>,
+    action: PayloadAction<{ from: PathStep; prevTo: PathStep; newTo: PathStep }>
+  ) {
+    const { from, prevTo, newTo } = action.payload;
+
+    if (!state.pathSteps.every((step) => step !== null)) {
+      throw new Error('PathSteps should not have null values at this point');
+    }
+    let newPathSteps: PathStep[] = [...state.pathSteps];
+
+    const powerRestrictionToModify = state.powerRestriction.find(
+      (restriction) => restriction.from === from.id
+    );
+
+    let newPowerRestrictionRanges: PowerRestriction[] = [];
+    if (!powerRestrictionToModify) {
+      // we need to remove the next range if it exists
+      newPowerRestrictionRanges = state.powerRestriction.filter(
+        (restriction) => restriction.from !== prevTo.id && restriction.to !== newTo.id
+      );
+    } else {
+      // replace the previous range by the new one and remove the previous one
+      for (const restriction of state.powerRestriction) {
+        if (restriction.from === from.id && restriction.to === prevTo.id) {
+          newPowerRestrictionRanges.push({ ...restriction, to: newTo.id });
+        } else if (restriction.from !== prevTo.id && restriction.to !== newTo.id) {
+          newPowerRestrictionRanges.push(restriction);
+        }
+      }
+
+      // add the new pathStep if needed
+      const pathIds = compact(state.pathSteps).map((step) => step.id);
+      if (!pathIds.includes(newTo.id)) {
+        const newToIndex = newPathSteps.findIndex(
+          (step) => step.positionOnPath && step.positionOnPath > newTo.positionOnPath!
+        );
+        if (newToIndex === -1) {
+          throw new Error('Can not insert the new pathStep in the pathSteps array');
+        }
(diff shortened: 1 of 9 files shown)


front/src/reducers/osrdconf/operationalStudiesConf/powerRestrictionReducer.ts
```diff
@@ -122,6 +126,51 @@ export function builPowerRestrictionReducer<
...
+      action: PayloadAction<{ from: PathStep; prevTo: PathStep; newTo: PathStep }>
+    ) {
+      const { from, prevTo, newTo } = action.payload;
+
+      const powerRestrictionToModify = state.powerRestriction.find(
+        (restriction) => restriction.from === from.id
+      );
+
+      let newPowerRestrictionRanges: PowerRestriction[] = [];
+      let newPathSteps: PathStep[] = [...state.pathSteps].map((step) => step!);
```
SharglutDev: nit : we can probably add an early check and `throw` if a path step is null because we don't have access to the power restriction tool if origin or destination are null (meaning there is no pathfinding)

SharglutDev (commented):
(was just passing by, didn't review all the code)

RomainValls (commented):
https://github.com/user-attachments/assets/76b8da86-21d2-4453-bfc8-ac155608f706

It seems if you cut the ranges multiple times and merge two of them, you're not able to merge the ranges with a position lower than the first cut. 
It does work for the range positioned after, and also for ranges two steps below the first cut. 

EDIT: tried again, and was able to merge ranges before the first cut (at least 2 steps before the first cut), except for the range that is right before the first cut.

front/src/reducers/osrdconf/operationalStudiesConf/powerRestrictionReducer.ts
clarani: Done :)

clarani:
It's fixed @RomainValls :wink:

front/src/reducers/osrdconf/operationalStudiesConf/powerRestrictionReducer.ts
SharglutDev: Was thinkin more off : `if (!state.pathSteps.every((step) => step !== null))` (like we used in `usePathfinding`), this way typescript knows there won't be any `null` `pathStep` and you don't have to `map` after that.

RomainValls:
> It's fixed @RomainValls 😉

Nice ! 
I think I found another one. When I try to merge the first range of the path (0km -> x km), the first and second range aren't merged. 
The bug doesn't seem to happen if I cut this first range once before trying to merge it. 
Tried on a PLY-MSC path, editing the first range (0-13km)

clarani:
Hum I can't reproduce it, how do you do ?
[Capture vidéo du 2025-02-13 17-45-26.webm](https://github.com/user-attachments/assets/bd9e34cf-739b-47f8-b637-4f8fccccca85)

front/src/reducers/osrdconf/operationalStudiesConf/powerRestrictionReducer.ts
clarani: Ah yes you're right, it's done :)

RomainValls:
> Hum I can't reproduce it, how do you do ? [Capture vidéo du 2025-02-13 17-45-26.webm](https://github.com/user-attachments/assets/bd9e34cf-739b-47f8-b637-4f8fccccca85)

I'm not sure what the expected behavior is on this PR : can we merge ranges that have no power restrictions ? 

I tried merging the first range of a PLY - MSC path with no power restriction, it isn't working. 
If I add a power restriction, the merge isn't working everytime : sometimes, i need to unselect the power restriction first so that I can merge it. 

I'm not able to reproduce it the same way everytime unfortunately. 

https://github.com/user-attachments/assets/21477ea9-74e6-434f-902c-673c9679653e

clarani:
Hum I think it's the expected behaviour, if I merge a empty range with a range with a power restriction, then the 2 ranges are merged and don't have power restriction anymore. But the electrification cuts need to be displayed on empty ranges, that's why nothing appears when you merge 2 empty ranges which were separated by an electrification range

Math-R (approved):
lgtm, i'm okay with the behaviour you indicate

RomainValls (approved):
LGTM ✅

PR #10565, opened 2025-01-29
core: compute ETCS LoA EBD foot and curve more accurately

### LINKED ISSUE #10430: core: etcs: for LOAs, compute EBD foot and EBD curve more accurately
In `ETCSBrakingCurves.computeBrakingCurves`, in the case of an EBD with a non null target speed, we are doing a couple of approximations:
- once we intersect with targetSpeed + dvEbi, we just shift the envelope to the right.
- if the LOA is close to the start of the path, the EBD could just reach the start of the path before reaching targetSpeed + dvEbi: in this case we don't shift it, which is another approximation

This creates a few small problems:
- we're not using the correct gradients for the calculation (by an offset of around ~50m most of the time, since dvEbiMax = 7.5km/h, it's not really high) -> not the biggest problem, since the offset tends to be small, but still not the best.
- if the EBD hits the start of the path, and we shift it, the curve won't start at position = 0.0. Potentially (almost never), that'll mean that the indication curve could start after 0.0. -> in this case, we'll use const gamme to brake from 0.0 till EBD curve, and we won't even see it in the MSP, because we'll skip this entire part with the time it takes for the train to accelerate and reach the MSP. This changes, however, if we make the train start with a beginSpeed (which we almost never do)..

We should really compute the correct curve, i.e., start from the point (LOAPosition, LOASpeed + dvEbi), decelerate to the left until MRSP or position = 0.0, then decelerate to the right until the point where we reache LOASpeed.

Erashin:
Fixes https://github.com/OpenRailAssociation/osrd/issues/10430 (see description for further information).

### DIFF
--- a/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
+++ b/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
@@ -481,7 +481,7 @@ private fun keepBrakingCurveUnderOverlay(
     val points = fullBrakingCurve.iteratePoints().distinct()
     val positions = points.map { it.position }
     val speeds = points.map { it.speed }
-    val timeDeltas = fullBrakingCurve.flatMap { it.cloneTimes().asList() }
+    val timeDeltas = fullBrakingCurve.flatMap { it.getTimeDeltas() }
 
     val partBuilder = EnvelopePartBuilder()
     partBuilder.setAttr(EnvelopeProfile.BRAKING)
--- a/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
+++ b/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
@@ -266,8 +266,7 @@ private fun computeEbdFullIndicationCurve(
     val ebdCurve =
         computeBrakingCurve(context, overhead, targetPosition, targetSpeed, BrakingType.ETCS_EBD)
     assert(ebdCurve.beginPos >= 0.0 && ebdCurve.endPos >= targetPosition)
-    // TODO: uncomment once ebd is not shifted anymore for an LoA
-    // assert((ebdCurve.beginSpeed == maxSpeedEbd || ebdCurve.beginPos == 0.0))
+    assert((ebdCurve.beginSpeed == maxSpeedEbd || ebdCurve.beginPos == 0.0))
 
     val guiCurve =
         computeBrakingCurve(context, overhead, targetPosition, targetSpeed, BrakingType.ETCS_GUI)
@@ -290,7 +289,7 @@ private fun computeBrakingCurve(
     targetPosition: Double,
     targetSpeed: Double,
     brakingType: BrakingType
-): EnvelopePart {
+): Envelope {
     assert(
         brakingType == BrakingType.ETCS_EBD ||
             brakingType == BrakingType.ETCS_SBD ||
@@ -317,40 +316,53 @@ private fun computeBrakingCurve(
             SpeedConstraint(targetSpeed, EnvelopePartConstraintType.FLOOR),
             EnvelopeConstraint(envelope, EnvelopePartConstraintType.CEILING)
         )
-    EnvelopeDeceleration.decelerate(
-        context,
-        targetPosition,
-        targetSpeed,
-        overlayBuilder,
-        -1.0,
-        brakingType
-    )
-    var brakingCurve = partBuilder.build()
-
     if (brakingType == BrakingType.ETCS_EBD && targetSpeed != 0.0) {
-        // TODO: by doing this, there is an approximation on the gradient used. TBD at a later date.
-        // When target is an LOA, EBD reaches target position at target speed + dVEbi: shift
-        // envelope to make it so. See Subset 026: §3.13.8.3.1, figure 40.
+        // When target is an LOA, EBD reaches target position at target speed + dVEbi. See Subset
+        // 026: §3.13.8.3.1, figure 40.
         val dvEbi = dvEbi(targetSpeed)
         val speedAtTargetPosition = targetSpeed + dvEbi
-        // Simplification: if EBD does not reach this limit, do not shift it. Impacts few very
-        // specific cases (LOA/SVL close to 0.0, with an acceleration curve which would hit them
-        // differently if shifted). Manage later on if needed.
-        if (
-            speedAtTargetPosition <= brakingCurve.beginSpeed &&
-                speedAtTargetPosition >= brakingCurve.endSpeed
-        ) {
-            val intersection = brakingCurve.interpolatePosition(targetSpeed + dvEbi)
-            brakingCurve =
-                brakingCurve.copyAndShift(
-                    targetPosition - intersection,
-                    0.0,
-                    Double.POSITIVE_INFINITY
-                )
-        }
+        // Compute deceleration to the left, starting with a speed a little above the LoA point.
+        EnvelopeDeceleration.decelerate(
+            context,
+            targetPosition,
+            speedAtTargetPosition,
+            overlayBuilder,
+            -1.0,
+            brakingType
+        )
+        val leftPart = partBuilder.build()
+        // Complete the curve by computing deceleration from the same point, but to the right (reset
+        // overlayBuilder to compute intersection with targetSpeed).
+        val rightPartBuilder = EnvelopePartBuilder()
+        rightPartBuilder.setAttr(EnvelopeProfile.BRAKING)
+        val rightOverlayBuilder =
+            ConstrainedEnvelopePartBuilder(
+                rightPartBuilder,
+                PositionConstraint(0.0, Double.POSITIVE_INFINITY),
+                SpeedConstraint(targetSpeed, EnvelopePartConstraintType.FLOOR)
+            )
+        EnvelopeDeceleration.decelerate(
+            context,
+            targetPosition,
+            speedAtTargetPosition,
+            rightOverlayBuilder,
+            1.0,
+            brakingType
+        )
+        val rightPart = rightPartBuilder.build()
+        return Envelope.make(leftPart, rightPart)
+    } else {
+        // For every other case, the braking curve reaches the target position at the target speed.
+        EnvelopeDeceleration.decelerate(
+            context,
+            targetPosition,
+            targetSpeed,
+            overlayBuilder,
+            -1.0,
+            brakingType
+        )
+        return Envelope.make(partBuilder.build())
     }
-
-    return brakingCurve
 }
 
 /**
@@ -358,16 +370,19 @@ private fun computeBrakingCurve(
  */
 private fun computeEbiBrakingCurveFromEbd(
     context: EnvelopeSimContext,
-    ebdCurve: EnvelopePart,
+    ebdCurve: Envelope,
     targetSpeed: Double
-): EnvelopePart {
-    val pointCount = ebdCurve.pointCount()
+): Envelope {
+    val ebdPoints = ebdCurve.iteratePoints().distinct()
+    val pointCount = ebdPoints.size
     var newPositions = DoubleArray(pointCount)
     var newSpeeds = DoubleArray(pointCount)
     for (i in 0 until pointCount) {
-        val speed = ebdCurve.getPointSpeed(i)
-        val becParams = computeBecParams(context, ebdCurve, speed, targetSpeed)
-        val newPos = ebdCurve.getPointPos(i) - becParams.dBec
+        val ebdPoint = ebdPoints[i]
+        val position = ebdPoint.position
+        val speed = ebdPoint.speed
+        val becParams = computeBecParams(context, position, speed, targetSpeed)
+        val newPos = position - becParams.dBec
         val newSpeed = speed - becParams.deltaBecSpeed
         newPositions[i] = newPos
         // TODO: unneeded for now: interpolate to not approximate position at 0 m/s.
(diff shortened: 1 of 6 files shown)


tests/tests/test_train_schedule.py
```diff
@@ -391,13 +391,14 @@ def test_etcs_schedule_result_slowdowns(etcs_scenario: Scenario, etcs_rolling_st
         < speed_before_brake_288_to_142
     )
 
-    offset_bending_guidance_point = 38_276_509
+    offset_bending_guidance_point = 38_322_611
     speed_at_bending_guidance_point = _get_current_or_next_speed_at(
         simulation_final_output, offset_bending_guidance_point
     )
-    _assert_equal_speeds(speed_at_bending_guidance_point, kph2ms(235.901_491_880_851_1))
+    # Permitted Speed and Guidance intersect at speed = 65.307_740_6 m/s
```
bougue-pe: nit:
```suggestion
    # Permitted Speed and Guidance intersect at speed ~65.8 m/s, testing a point of the curve close enough where speed is 65.307_740_6 m/s.
```

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -233,44 +233,43 @@ private fun computeBrakingCurve(
...
-        ) {
-            val intersection = brakingCurve.interpolatePosition(targetSpeed + dvEbi)
-            brakingCurve =
-                brakingCurve.copyAndShift(
-                    targetPosition - intersection,
-                    0.0,
-                    Double.POSITIVE_INFINITY
-                )
-        }
+        EnvelopeDeceleration.decelerate(
```
bougue-pe, line 325: nit:
```suggestion
        // Compute deceleration to the left, starting with a speed a little above the LoA point
        EnvelopeDeceleration.decelerate(
```

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -233,44 +233,43 @@ private fun computeBrakingCurve(
...
+            context,
+            targetPosition,
+            speedAtTargetPosition,
+            overlayBuilder,
+            -1.0,
+            brakingType
+        )
+        overlayBuilder.lastIntersection = -1
+        EnvelopeDeceleration.decelerate(
```
bougue-pe: nit:
```suggestion
        // Complete the curve by computing deceleration from the same point, but to the right
        // (reset overlayBuilder to compute intersection with targetSpeed)
        overlayBuilder.lastIntersection = -1
        EnvelopeDeceleration.decelerate(
```

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -233,44 +233,43 @@ private fun computeBrakingCurve(
...
-        -1.0,
-        brakingType
-    )
-    var brakingCurve = partBuilder.build()
-
-    if (brakingType == BrakingType.ETCS_EBD && targetSpeed != 0.0) {
-        // TODO: by doing this, there is an approximation on the gradient used. TBD at a later date.
-        // When target is an LOA, EBD reaches target position at target speed + dVEbi: shift
-        // envelope to make it so. See Subset 026: §3.13.8.3.1, figure 40.
+    if (brakingType != BrakingType.ETCS_EBD || targetSpeed == 0.0) {
```
bougue-pe: nit: for code-contiguity, I like better the condition to be "positively" related to the case computed.
So I'd either change this condition to:
```suggestion
    if (brakingType == BrakingType.ETCS_SBD || brakingType == BrakingType.ETCS_GUI) {
```
Or just revert the if/else parts (and so the condition)... Or just narrow cases for each computing and reject (for now) EBD+0.0?

Looks like we could "silently" trigger SvL computing if we already iterated on the values (which is not the case).

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePartBuilder.java
```diff
@@ -40,10 +40,21 @@ private void addPoint(double position, double speed) {
     }
 
     @Override
+    @SuppressFBWarnings({"FE_FLOATING_POINT_EQUALITY"})
     public void initEnvelopePart(double position, double speed, double direction) {
```
bougue-pe: nit: The naming is not so straightforward now.
I can think of (sorted by preference): `prepareEnvelopePartForExtension`, `initOrReinitEnvelopePart`.

bougue-pe (approved):
Thanks, quite elegant changes (discovering the idea of `initEnvelopePart` changes)

Only nits, so feel free to reply-reject/resolve them, it's fine by me as-is 🙏

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Yeah this wouldn't work. With your condition, we wouldn't go in the if case if it was an svl (EBD + targetSpeed = 0.0). I'll just change the order of the if else parts.

Erashin (commented):
@bougue-pe : everything solved in https://github.com/OpenRailAssociation/osrd/pull/10565/commits/870da53e74954df61667e3ec4fb14b62d19fb1ad.

bougue-pe (approved):

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePartBuilder.java
```diff
@@ -40,10 +40,21 @@ private void addPoint(double position, double speed) {
...
+        } else if (this.direction == direction) {
+            assert this.lastPos == position && this.lastSpeed == speed;
+        } else {
+            assert this.positions.get(0) == position && this.speeds.get(0) == speed;
+            this.lastPos = positions.get(0);
+            this.positions = (DoubleArrayList) positions.reverse();
+            this.lastSpeed = speeds.get(0);
+            this.speeds = (DoubleArrayList) speeds.reverse();
+            this.times = (DoubleArrayList) times.reverse();
+        }
```
eckter: This is used to build an envelope part going forward, then moving back to its start and building further going backwards?

Huh. Do we really want to support this kind of workflow "silently"? It sounds unusual.

Erashin: Quite liked this solution, but yeah, this does complexify the whole code globally.
Alternate solution found while discussing: move this pb back into `ETCSBrakingCurves`: create 2 part builders and 2 respective overlay builders to compute both the left and right sections of the LoA braking curve. Return an Envelope. Use an Envelope everywhere in the rest of the file rather than an EnvelopePart.
Globally, a lot more changes are needed, but the code will be a tad clearer, and people should understand it better.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -233,61 +233,78 @@ private fun computeBrakingCurve(
     val overlayBuilder =
         ConstrainedEnvelopePartBuilder(
             partBuilder,
-            PositionConstraint(0.0, targetPosition),
+            PositionConstraint(0.0, Double.POSITIVE_INFINITY),
```
eckter:
```suggestion
```

I don't think it does anything now

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -604,11 +611,13 @@ private fun getPermittedSpeedPosition(sbiPosition: Double, speed: Double): Doubl
...
     speed: Double,
-    guiCurve: EnvelopePart
+    guiCurve: Envelope
 ): Double {
+    assert(guiCurve.stream().count() == 1L)
+    val guiPart = guiCurve.stream().findFirst().get()
     val guiPosition =
-        if (speed > guiCurve.maxSpeed || speed < guiCurve.minSpeed) Double.POSITIVE_INFINITY
-        else guiCurve.interpolatePosition(speed)
+        if (speed > guiPart.maxSpeed || speed < guiPart.minSpeed) Double.POSITIVE_INFINITY
```
Erashin: PR is not rebased on DEV yet, this change will disappear to become if(...) guipart.beginPos.

eckter (approved):

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -460,21 +479,9 @@ private fun keepBrakingCurveUnderOverlay(
...
-    val speeds =
-        fullBrakingCurve
-            .map { it.cloneSpeeds() }
-            .reduce { mergedArray, currentArray ->
-                mergedArray + currentArray.drop(1).toDoubleArray()
-            }
+    val points = fullBrakingCurve.iteratePoints().distinct()
+    val positions = points.map { it.position }
+    val speeds = points.map { it.speed }
     val timeDeltas = fullBrakingCurve.flatMap { it.cloneTimes().asList() }
```
eckter: Now that we're in Kotlin, those clones could be turned into read-only List viewers on the underlying arrays. The cloned values are actually very rarely modified.

eckter (approved):

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Yep, I'll do that in another commit.
Done in https://github.com/OpenRailAssociation/osrd/pull/10565/commits/32f9dbd3e320df82ff092420640b9d7b4fe5fff2.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -317,57 +316,74 @@ private fun computeBrakingCurve(
...
+        // Complete the curve by computing deceleration from the same point, but to the right (reset
+        // overlayBuilder to compute intersection with targetSpeed).
+        val rightPartBuilder = EnvelopePartBuilder()
+        rightPartBuilder.setAttr(EnvelopeProfile.BRAKING)
+        val rightOverlayBuilder =
+            ConstrainedEnvelopePartBuilder(
+                rightPartBuilder,
+                PositionConstraint(0.0, Double.POSITIVE_INFINITY),
+                SpeedConstraint(targetSpeed, EnvelopePartConstraintType.FLOOR),
+                EnvelopeConstraint(envelope, EnvelopePartConstraintType.CEILING)
```
bougue-pe: Why is envelope a ceiling constraint here?
AFAIU this part is actually _above_ envelope, no? (constraint probably doing nothing the way it works, but it's a bit lucky, if I'm not mistaken)

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -317,57 +316,74 @@ private fun computeBrakingCurve(
...
     var newSpeeds = DoubleArray(pointCount)
     for (i in 0 until pointCount) {
-        val speed = ebdCurve.getPointSpeed(i)
-        val becParams = computeBecParams(context, ebdCurve, speed, targetSpeed)
-        val newPos = ebdCurve.getPointPos(i) - becParams.dBec
+        val ebdPoint = ebdPoints[i]
+        val position = ebdPoint.position
+        val speed = ebdPoint.speed
+        val becParams = computeBecParams(context, position, speed, targetSpeed)
+        val newPos = ebdPoint.position - becParams.dBec
```
bougue-pe: Not sure, it looks like now you can:
```kt
    for (ebdPoint in ebdPoints) {
        val position = ebdPoint.position // maybe even remove this val?
        val speed = ebdPoint.speed // maybe even remove this val?
        val becParams = computeBecParams(context, position, speed, targetSpeed)
        val newPos = position - becParams.dBec
```

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -411,12 +429,13 @@ private fun computeIndicationBrakingCurveFromRef(
...
 
-    val pointCount = refBrakingCurve.pointCount()
+    val refBrakingPoints = refBrakingCurve.iteratePoints().distinct()
+    val pointCount = refBrakingPoints.size
     val newPositions = DoubleArray(pointCount)
     val newSpeeds = DoubleArray(pointCount)
-    for (i in 0 until refBrakingCurve.pointCount()) {
-        val speed = refBrakingCurve.getPointSpeed(i)
-        val sbiPosition = getSbiPosition(refBrakingCurve.getPointPos(i), speed, tBs)
+    for (i in 0 until pointCount) {
```
bougue-pe, line 435: Ditto:
```suggestion
    for (refBrakingPoint in refBrakingPoints) {
```

bougue-pe (commented):
Clearer like this indeed, good work 👍

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Was trying to look into a weird case where this could have an impact but I don't think it's possible so yeah, I can definitely remove it.
See https://github.com/OpenRailAssociation/osrd/pull/10565/commits/850fa3959e99b79b619801199c5ce23c5b810a3c for your comments.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Nope I need  `i` just below to add the values to the DoubleArray. But I can and will modify l386 to `val newPos = position - becParams.dBec`.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Same, I need it just below to add the value to the DoubleArray.

bougue-pe (approved):

bougue-pe (approved):

PR #10597, opened 2025-01-31
front: fix markers display

### LINKED ISSUE #10395: Wrong marker for destination
### What happened?

https://github.com/user-attachments/assets/0ceaf07d-b458-412d-82a8-b7ff53023ef2

The marker should be the one from destination

Reproduce video

Local

Firefox

.

### LINKED ISSUE #10406: The origin marker appears instead of the destination one
### What happened?

When starting with the destination this happens:

https://github.com/user-attachments/assets/f8268492-f981-493b-afd4-90d356772ed7

The marker is however correct once the origin is set.

There should be an orange flag for the destination no matter the order of the actions

1. Only set a destination

Recette (SNCF)

Google Chrome

9151b80

theocrsb:
fix #10406 and fix #10395

### DIFF
--- a/front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
+++ b/front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
@@ -1,3 +1,5 @@
+import { useMemo } from 'react';
+
 import { compact } from 'lodash';
 import { useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
@@ -18,6 +20,10 @@ import { useStoreDataForRollingStockSelector } from 'modules/rollingStock/compon
 import { isElectric } from 'modules/rollingStock/helpers/electric';
 import TimesStopsInput from 'modules/timesStops/TimesStopsInput';
 import { Map } from 'modules/trainschedule/components/ManageTrainSchedule';
+import {
+  MARKER_TYPE,
+  type MarkerInformation,
+} from 'modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/ItineraryMarkers';
 import SimulationSettings from 'modules/trainschedule/components/ManageTrainSchedule/SimulationSettings';
 import TrainSettings from 'modules/trainschedule/components/ManageTrainSchedule/TrainSettings';
 import {
@@ -43,6 +49,29 @@ const ManageTrainSchedule = ({ trainIdToEdit }: ManageTrainScheduleProps) => {
   const origin = useSelector(getOrigin);
   const destination = useSelector(getDestination);
   const pathSteps = useSelector(getPathSteps);
+
+  const markersInformation = useMemo(
+    () =>
+      pathSteps.reduce<MarkerInformation[]>((acc, step, index) => {
+        if (!step) return acc;
+
+        let pointType = MARKER_TYPE.VIA;
+
+        if (index === 0) {
+          pointType = MARKER_TYPE.ORIGIN;
+        } else if (index === pathSteps.length - 1) {
+          pointType = MARKER_TYPE.DESTINATION;
+        }
+        acc.push({
+          ...step,
+          pointType,
+        });
+
+        return acc;
+      }, []),
+    [pathSteps]
+  );
+
   const constraintDistribution = useSelector(getConstraintDistribution);
   const startTime = useSelector(getStartTime);
 
@@ -103,7 +132,7 @@ const ManageTrainSchedule = ({ trainIdToEdit }: ManageTrainScheduleProps) => {
         <div className="floating-itinerary">
           <Itinerary />
         </div>
-        <Map pathProperties={pathProperties} simulationPathSteps={compact(pathSteps)}>
+        <Map pathProperties={pathProperties} simulationPathSteps={markersInformation}>
           <IncompatibleConstraints pathProperties={pathProperties} />
         </Map>
       </div>
--- a/front/src/applications/stdcm/utils/index.ts
+++ b/front/src/applications/stdcm/utils/index.ts
@@ -1,4 +1,7 @@
-import type { MarkerInformation } from 'modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/ItineraryMarkers';
+import {
+  MARKER_TYPE,
+  type MarkerInformation,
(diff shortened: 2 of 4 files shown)


front/src/modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/ItineraryMarkers.tsx
```diff
@@ -68,29 +69,28 @@ const formatPointWithNoName = (
...
           coordinates: cur.coordinates,
-          type: MARKER_TYPE.VIA,
+          type: MARKER_TYPE.DESTINATION,
           marker: cur,
-          imageSource: showStdcmAssets ? stdcmVia : viaSVG,
-          index,
+          imageSource: showStdcmAssets ? stdcmDestination : destinationSVG,
         });
-      } else if (index === pathSteps.length - 1) {
+      } else if (cur.pointType === MARKER_TYPE.VIA)
```
clarani: This last condition is not necessary ;)
```suggestion
      } else {
```

front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
```diff
@@ -43,6 +46,30 @@ const ManageTrainSchedule = ({ trainIdToEdit }: ManageTrainScheduleProps) => {
   const origin = useSelector(getOrigin);
   const destination = useSelector(getDestination);
   const pathSteps = useSelector(getPathSteps);
+
+  const pathStepWithPointType = useMemo(
+    () =>
+      pathSteps.map((step, index, array) => {
```
clarani: Instead of using a map and a compact l133, you can directly use a reduce here. Add in `acc` only the steps that are not null
```suggestion
      pathSteps.reduce((acc, step, index) => {
```

front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
```diff
@@ -43,6 +46,30 @@ const ManageTrainSchedule = ({ trainIdToEdit }: ManageTrainScheduleProps) => {
   const origin = useSelector(getOrigin);
   const destination = useSelector(getDestination);
   const pathSteps = useSelector(getPathSteps);
+
+  const pathStepWithPointType = useMemo(
```
clarani:
```suggestion
  const markersInformation = useMemo(
```

front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
```diff
@@ -43,6 +46,30 @@ const ManageTrainSchedule = ({ trainIdToEdit }: ManageTrainScheduleProps) => {
...
+    () =>
+      pathSteps.map((step, index, array) => {
+        if (!step) return null;
+
+        let pointType;
```
clarani: nit: if you want to save 1 line, you can initiate `pointType` here and remove the last test
```suggestion
        let pointType = MARKER_TYPE.VIA;
```

front/src/applications/stdcm/utils/index.ts
```diff
@@ -13,13 +16,23 @@ export const getTimesInfoFromDate = (date?: Date) =>
...
-  }, [] as MarkerInformation[]);
+export const extractMarkersInfo = (pathSteps: StdcmPathStep[]): MarkerInformation[] =>
+  pathSteps
+    .filter((step) => step.location)
+    .map((step, index, array) => {
+      let pointType;
+
+      if (index === 0) {
+        pointType = MARKER_TYPE.ORIGIN;
+      } else if (index === array.length - 1) {
```
clarani: Why use array instead of pathSteps here ?
```suggestion
      } else if (index === pathSteps.length - 1) {
```

front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
SharglutDev: The `reduce` war continues 🔫

front/src/modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/ItineraryMarkers.tsx
theocrsb: Done

front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
theocrsb: Done

front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
theocrsb: Done

front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
theocrsb: Done

front/src/applications/stdcm/utils/index.ts
theocrsb: Done

front/src/applications/stdcm/utils/index.ts
```diff
@@ -13,13 +16,23 @@ export const getTimesInfoFromDate = (date?: Date) =>
...
-      acc.push({
-        coordinates: step.location.coordinates,
-        name: step.location.name,
-      });
-    }
-    return acc;
-  }, [] as MarkerInformation[]);
+export const extractMarkersInfo = (pathSteps: StdcmPathStep[]): MarkerInformation[] =>
+  pathSteps
+    .filter((step) => step.location)
```
clarani: As you filter before the map, you don't solve the issue here... It would be better either to filter after the map or to use a reduce

theocrsb: Done

clarani:
image

Almost good I think. If I add a via in stdcm but no origin, then the icon is correct but the number in it is not correct

theocrsb:
> image
> 
> Almost good I think. If I add a via in stdcm but no origin, then the icon is correct but the number in it is not correct

I just pushed a solution

clarani (approved):
LGTM :white_check_mark:

Akctarus (approved):
lgtm

PR #10603, opened 2025-01-31
front: add some variables to stdcm store

### LINKED ISSUE #10600: front: add stdcm simulations to the store

clarani:
closes #10600

### DIFF
--- a/front/src/applications/stdcm/views/StdcmView.tsx
+++ b/front/src/applications/stdcm/views/StdcmView.tsx
@@ -11,7 +11,11 @@ import {
   updateLastStdcmResult,
   type StdcmConfSliceActions,
 } from 'reducers/osrdconf/stdcmConf';
-import { getStdcmConf, getStdcmSimulations } from 'reducers/osrdconf/stdcmConf/selectors';
+import {
+  getRetainedSimulationIndex,
+  getStdcmConf,
+  getStdcmSimulations,
+} from 'reducers/osrdconf/stdcmConf/selectors';
 import { useAppDispatch } from 'store';
 
 import StdcmEmptyConfigError from '../components/StdcmEmptyConfigError';
@@ -29,10 +33,10 @@ const StdcmView = () => {
   const currentSimulationInputs = useStdcmForm();
   const stdcmConf = useSelector(getStdcmConf);
   const simulationsList = useSelector(getStdcmSimulations);
+  const retainedSimulationIndex = useSelector(getRetainedSimulationIndex);
 
   const [selectedSimulationIndex, setSelectedSimulationIndex] = useState(-1);
   const [showStatusBanner, setShowStatusBanner] = useState(false);
-  const [retainedSimulationIndex, setRetainedSimulationIndex] = useState(-1);
   const [showBtnToLaunchSimulation, setShowBtnToLaunchSimulation] = useState(false);
   const [isDebugMode, setIsDebugMode] = useState(false);
   const [showHelpModule, setShowHelpModule] = useState(false);
@@ -59,10 +63,8 @@ const StdcmView = () => {
 
   const selectedSimulation = simulationsList[selectedSimulationIndex];
 
-  const handleRetainSimulation = () => setRetainedSimulationIndex(selectedSimulationIndex);
-
   const handleSelectSimulation = (index: number) => {
-    if (retainedSimulationIndex === -1) {
+    if (retainedSimulationIndex === undefined) {
       setSelectedSimulationIndex(index);
       setShowBtnToLaunchSimulation(false);
     }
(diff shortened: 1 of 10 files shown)


front/src/reducers/osrdconf/stdcmConf/index.ts
```diff
@@ -252,6 +232,40 @@ export const stdcmConfSlice = createSlice({
...
+      } = selectedSimulation;
+      state.rollingStockID = consist?.tractionEngine?.id;
+      state.towedRollingStockID = consist?.towedRollingStock?.id;
+      state.totalLength = consist?.totalLength;
+      state.totalMass = consist?.totalMass;
+      state.maxSpeed = consist?.maxSpeed;
+      state.speedLimitByTag = consist?.speedLimitByTag;
+      state.stdcmPathSteps = pathSteps;
+    },
+    retainSimulation(state: Draft<OsrdStdcmConfState>, action: PayloadAction<number>) {
```
SharglutDev, line 265: Should this be called `updateRetainSimulation` ? (same for `selectSimulation` > `updateSelectedSimulation`)

SharglutDev (commented):
Great refacto !

I have one bug : white screen when clicking on new query with current input button

front/src/reducers/osrdconf/types.ts
```diff
@@ -45,6 +46,9 @@ export type OsrdStdcmConfState = OsrdConfState & {
   maxSpeed?: number;
   towedRollingStockID?: number;
   linkedTrains: LinkedTrains;
+  simulations: StdcmSimulation[];
+  selectedSimulation?: StdcmSimulation;
+  retainedSimulationIndex?: number;
```
kmer2016: Storing the selected simulation as a full object can lead to information duplication: if the selected simulation is already in the array, any modification—whether made in the array or in the selected object—must remain consistent across both. 

An alternative is to store only the index for the selected simulation and derive the complete object from the array, which reduces the risk of desynchronization and simplifies update logic. Adopting this approach will also harmonize how we handle both the selected and retained simulation variables.

What do you think about to store only the indexes for both the selected and retained simulations. Then, use selectors to retrieve either the indexes or the full simulation objects as needed.

front/src/reducers/osrdconf/stdcmConf/stdcmConfReducers.spec.ts
kmer2016, line 1: It can be very nice, if you add some tests for the new actions :)

front/src/reducers/osrdconf/stdcmConf/index.ts
kmer2016: I personally think we can leave the names as they are without adding "update" in front. These actions represent business events rather than simple state updates, so keeping them as `selectSimulation` and `retainSimulation` better reflects their intent within the domain.

front/src/applications/stdcm/views/StdcmView.tsx
```diff
@@ -92,27 +98,12 @@ const StdcmView = () => {
...
-    }
-  }, [selectedSimulation]);
-
   useEffect(() => {
     if (!isDebugMode) {
-      setShowBtnToLaunchSimulation(!isEqual(currentSimulationInputs, selectedSimulation?.inputs));
+      setShowBtnToLaunchSimulation(
+        selectedSimulationIndex === undefined ||
+          !isEqual(currentSimulationInputs, simulationsList[selectedSimulationIndex].inputs)
+      );
```
kmer2016: I think we can directely use the `selectedSimulation` here

kmer2016 (commented):
Thank you for the PR! I made some propositions

front/src/reducers/osrdconf/types.ts
clarani: Yes done :white_check_mark:

front/src/applications/stdcm/views/StdcmView.tsx
clarani: Hum no I can't, because I made the selector getSelectedSimulation such as if there is no simulation selected, the app throws an error. Here I need to handle the case where no simulation is selected, so I prefer to use the other selector if it's fine for you

front/src/reducers/osrdconf/stdcmConf/index.ts
clarani: I would rather keep my names, I find them more explicit

achrafmohye (approved):
Thanks for the refacto, LGTM and tested :)

front/src/applications/stdcm/views/StdcmView.tsx
kmer2016: Fine !

kmer2016 (approved):
Thank you forr the refacto and test ! ;)

SharglutDev (approved):
Lgtm and tested

PR #10617, opened 2025-01-31
front: remove react-beautiful-dnd

### LINKED ISSUE #10616: Remove react-beautiful-dnd
### Description and goal

Part of https://github.com/OpenRailAssociation/osrd/issues/10338 (see reason in the issue)

### Acceptance criteria

- [ ] `defaultProps` warning in console on Itinerary tab with vias is gone
- [ ] Vias styles and behavior are the same

SharglutDev:
The lib hasn't been updated since 2022 and now is outdated regarding React 19 new version. The lib uses React "defaultProps" feature which is deprecated in React 19. As the drag and drop feature isn't used by users and can be bypassed, remove the lib and the related feature to enable a future migration to React 19.

close #10616


Synar:
Worth noting that the authors of beautiful dnd have created a new lib called [pragmatic dnd](https://github.com/atlassian/pragmatic-drag-and-drop?tab=readme-ov-file), which they recommend in their [deprecation announcement](https://github.com/atlassian/react-beautiful-dnd/issues/2672) for beautiful dnd. If we're only removing the feature for react 19 support, it may be worth migrating instead

SharglutDev:
> Worth noting that the authors of beautiful dnd have created a new lib called [pragmatic dnd](https://github.com/atlassian/pragmatic-drag-and-drop?tab=readme-ov-file), which they recommend in their [deprecation announcement](https://github.com/atlassian/react-beautiful-dnd/issues/2672) for beautiful dnd. If we're only removing the feature for react 19 support, it may be worth migrating instead

Yes you're right, I saw that as well.

As I tried to explain, this lib was used only for the drag and drop feature which wasn't used by our users. Ex: I have a path Perrigny - Valence - Avignon - Miramas, if I switch Valence and Avignon, the pathfinding breaks, it doesn't make any sense geographically. It will be pretty rare to have a proper use of this feature.

Even if it's not a big deal to migrate with the new lib, @thibautsailly validated that we don't really need the feature and we shouldn't devote time for that.

Synar (approved):
Then everything LGTM ^^

Math-R (approved):
Looks fine on my side thanks !

PR #10620, opened 2025-02-02
Support models with a single primary key in Model derive macro

### LINKED ISSUE #8111: Implement support for single PK field in ModelV2

leovalais:
closes #8111

### DIFF
--- a/editoast/editoast_derive/src/model.rs
+++ b/editoast/editoast_derive/src/model.rs
@@ -110,4 +110,67 @@ mod tests {
             }
         );
     }
+
+    #[test]
+    #[should_panic]
+    fn single_pk_models_do_not_support_updates() {
+        crate::assert_macro_expansion!(
+            model,
+            syn::parse_quote! {
+                #[derive(Clone, Model)]
+                #[model(table = editoast_models::tables::timetable)]
+                #[model(gen(ops = u))]
+                struct Timetable {
+                    id: i64,
+                }
+            }
+        );
+    }
+
+    #[test]
+    #[should_panic]
+    fn single_pk_models_do_not_support_batch_creates() {
+        crate::assert_macro_expansion!(
+            model,
+            syn::parse_quote! {
+                #[derive(Clone, Model)]
+                #[model(table = editoast_models::tables::timetable)]
+                #[model(gen(batch_ops = c))]
+                struct Timetable {
+                    id: i64,
+                }
+            }
+        );
+    }
+
+    #[test]
+    #[should_panic]
+    fn single_pk_models_do_not_support_batch_updates() {
+        crate::assert_macro_expansion!(
+            model,
+            syn::parse_quote! {
+                #[derive(Clone, Model)]
+                #[model(table = editoast_models::tables::timetable)]
+                #[model(gen(batch_ops = u))]
+                struct Timetable {
+                    id: i64,
+                }
+            }
+        );
+    }
+
+    #[test]
+    fn single_pk_model() {
+        crate::assert_macro_expansion!(
+            model,
+            syn::parse_quote! {
+                #[derive(Clone, Model)]
+                #[model(table = editoast_models::tables::timetable)]
+                #[model(gen(ops = crd, batch_ops = rd, list))]
+                struct Timetable {
+                    id: i64,
+                }
+            }
+        );
+    }
 }
--- a/editoast/editoast_derive/src/model/codegen/changeset_decl.rs
+++ b/editoast/editoast_derive/src/model/codegen/changeset_decl.rs
@@ -38,15 +38,24 @@ impl ToTokens for ChangesetDecl {
                 np!(column_name, vis, name, ty)
             })
             .unzip();
-        tokens.extend(quote! {
-            #[derive(Debug, Default, Queryable, AsChangeset, Insertable, #(#additional_derives),*)]
-            #[diesel(table_name = #table)]
-            #vis struct #ident {
-                #(
-                    #[diesel(deserialize_as = #field_type, column_name = #field_column)]
-                    #field_vis #field_name: Option<#field_type>
-                ),*
-            }
-        });
+
+        // If the struct has no fields, Queryable, AsChangeset and Insertable cannot be derived.
+        if fields.is_empty() {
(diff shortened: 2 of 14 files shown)


editoast/src/models/work_schedules.rs
```diff
@@ -22,6 +32,92 @@ pub struct WorkScheduleGroup {
...
+//             > for Name
+//         {
+//             const COLUMN: editoast_models::tables::work_schedule_group::name =
+//                 editoast_models::tables::work_schedule_group::name;
+//         }
+//         Name
+//     }
+// }
+
+// fn lolz() {
```
Castavo: Nice bit of code there

leovalais: Isn't it? Peak programming practices around here.

leovalais: (Thanks, I thought I stashed this 🐒)

editoast/editoast_derive/src/model.rs
```diff
@@ -110,4 +110,46 @@ mod tests {
...
+            model,
+            syn::parse_quote! {
+                #[derive(Clone, Model)]
+                #[model(table = editoast_models::tables::timetable)]
+                #[model(gen(batch_ops = c))]
+                struct Timetable {
+                    id: i64,
+                }
+            }
+        );
```
woshilapin: If I understand correctly, each of this `assert_macro_expansion!()` should panic? But what if the first always panic, and the second doesn't: we would never notice since the entire test will still panic?

flomonster (approved):
LGTM

editoast/editoast_derive/src/model/parsing.rs
```diff
@@ -78,63 +71,145 @@ impl ModelConfig {
...
+                (RawIdentifier::Field(ident.clone()), Some(ident.clone()))
+            }
+            (Some(id), []) => (id.clone(), None),
+            (None, [field]) => (
+                RawIdentifier::Field(field.ident.clone()),
+                Some(field.ident.clone()),
+            ),
+            (None, []) => (primary_field.clone(), None),
+            e => {
+                dbg!(e);
```
woshilapin: Oops (dixit @leovalais)

editoast/editoast_derive/src/model.rs
```diff
@@ -110,4 +110,46 @@ mod tests {
...
+    }
+
+    #[test]
+    fn single_pk_model() {
+        crate::assert_macro_expansion!(
+            model,
+            syn::parse_quote! {
+                #[derive(Clone, Model)]
+                #[model(table = editoast_models::tables::timetable)]
+                #[model(gen(ops = crd, batch_ops = rd))]
```
woshilapin: Also the `list`?

editoast/editoast_derive/src/model/codegen/changeset_decl.rs
```diff
@@ -38,15 +38,22 @@ impl ToTokens for ChangesetDecl {
...
-            #[derive(Debug, Default, Queryable, AsChangeset, Insertable, #(#additional_derives),*)]
-            #[diesel(table_name = #table)]
-            #vis struct #ident {
-                #(
-                    #[diesel(deserialize_as = #field_type, column_name = #field_column)]
-                    #field_vis #field_name: Option<#field_type>
-                ),*
-            }
-        });
+        if fields.is_empty() {
```
woshilapin, line 43: Maybe add a comment to explain why we do need this `if`?

Suggestion:

> If the struct has no fields, `Queryable`, `AsChangeset` and `Insertable` cannot be derived.

It should help future developers to quickly see where the difference is and why.

woshilapin (commented):
Thank you for this! Sorry for the delay of the review. A few comments made mostly by you 😛

editoast/editoast_derive/src/model.rs
leovalais: e0e2876b1

editoast/editoast_derive/src/model/parsing.rs
leovalais: 40e6d811d

editoast/editoast_derive/src/model.rs
leovalais: 9e1334d34

editoast/editoast_derive/src/model/codegen/changeset_decl.rs
leovalais: d2c4bc841

leovalais:
Bonus: 28878075d

woshilapin (approved):

PR #10624, opened 2025-02-03
front: fix track selection after timestops changes

### LINKED ISSUE #10532: Track choice erased when adding a stopping time
### What happened?

I create a train for which I choose the track on which the train stops in Macon. I add a 120s stop at the same point. Then the result does not stop on the expected track. (I don't see the track in the payload)

https://github.com/user-attachments/assets/2daa0ef4-d502-4ce3-b69a-dc3794e98447

Seems that without adding a stopping time, it works

I expected the track to be the one I chose

As in the video

Recette (SNCF)

Google Chrome

fff6686

RomainValls:
Closes #10532

### DIFF
--- a/front/src/reducers/osrdconf/helpers.ts
+++ b/front/src/reducers/osrdconf/helpers.ts
@@ -4,6 +4,7 @@ import { v4 as uuidV4 } from 'uuid';
 
 import { calculateDistanceAlongTrack } from 'applications/editor/tools/utils';
 import type { ManageTrainSchedulePathProperties } from 'applications/operationalStudies/types';
+import type { OperationalPointReference } from 'common/api/osrdEditoastApi';
 import { pathStepMatchesOp } from 'modules/pathfinding/utils';
 import type { SuggestedOP } from 'modules/trainschedule/components/ManageTrainSchedule/types';
 import { addElementAtIndex } from 'utils/array';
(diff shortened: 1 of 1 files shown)


front/src/reducers/osrdconf/helpers.ts
```diff
@@ -90,7 +91,12 @@ export function upsertPathStep(statePathSteps: (PathStep | null)[], op: Suggeste
   if (stepIndex >= 0) {
     // Because of import issues, there can be multiple ops with same position on path
     // To avoid updating the wrong one, we need to find the one that matches the payload
-    newVia = { ...newVia, id: cleanPathSteps[stepIndex].id }; // We don't need to change the id of the updated via
+    newVia = {
+      ...newVia,
+      id: cleanPathSteps[stepIndex].id,
+      track_reference:
+        (cleanPathSteps[stepIndex] as OperationalPointReference).track_reference || undefined,
```
Synar, line 98: Fix LGTM, but we should not have to typecast here. This is a sign the signature of our function is wrong (and so is the signature of the function adding the track_reference) : either we need to add the track_reference property in PathStep, use a supertype of PathStep, or otherwise refacto the addition of the track reference. This may be outside the scope of this PR though

(Also I would rather the front use the case .trackReference until the conversion to PathItemLocation in checkCurrentConfig, but that's kind of an irrelevant nitpick, that also applies to other properties. Making the conversions explicit would help highlight our type abuses though.)

achrafmohye:
I still have the issue but only when adding a stop time, if the via is added without the a stop time the fix seems to work, don't know if it was already there before

achrafmohye:
@RomainValls when i add a stop time i still have **v2** and not **V1BIS** that i made as the track for the train stop in Macon, is it the case or am i missing something ?

RomainValls:
> @RomainValls when i add a stop time i still have **v2** and not **V1BIS** that i made as the track for the train stop in Macon, is it the case or am i missing something ?

I tried again, and it seems fixed for me : 

https://github.com/user-attachments/assets/1b01280b-04f7-4c58-a70e-c82f8ade4d63

I also tried other tracks, and by editing the timestop after I added the train first, and edited it after. 
@Math-R what about you ?

achrafmohye (approved):
Thanks for the fix, it's good for me now, i think you can add a todo if the fix regarding the types proposed by @Synar to be done by someone else in a seperate PR

Math-R (approved):
Looks good to me ! And agreed with @achrafmohye, you shall create a new issue to do the suggestion made by @Synar

Synar (approved):

RomainValls:
Created this issue for the typecast : https://github.com/OpenRailAssociation/osrd/issues/10805

PR #10625, opened 2025-02-03
front: spacetimechart: fix drag behaviour when draging train to the left

### LINKED ISSUE #10511: Stop updating the time window when moving a train around in the GET
### What happened?

When moving a train
- in the past
- in any direction if there's only one train

this happens:

https://github.com/user-attachments/assets/7d6d37e4-ef2a-4ced-b8ec-b9ff8fdb8ca2

While moving it in the future looks like this:

https://github.com/user-attachments/assets/4a8661e5-cc7f-4ada-8044-c732c58fdf9a

The second behavior is way more intuitive. It would be nice to have this when moving a train in any direction.

So the goal is to stop updating the time window when moving a train in the GET, and also not update it if the train has been moved in the past (otherwise the time window will teleport after releasing the clic, which will be confusing).

as above

Recette (SNCF)

Google Chrome

fff6686

Math-R:
close #10511

Previously when you were dragging the earliest train to the left, the spaceTimeChart window was sliding to the left.
Now the SpaceTimeChart window won't move anymore unless you stop the train drag and pan the window

### DIFF
--- a/front/src/applications/operationalStudies/views/SimulationResults.tsx
+++ b/front/src/applications/operationalStudies/views/SimulationResults.tsx
@@ -3,6 +3,7 @@ import { useEffect, useState, useMemo } from 'react';
 import { ChevronLeft, ChevronRight } from '@osrd-project/ui-icons';
 import cx from 'classnames';
 import { useTranslation } from 'react-i18next';
+import { useSelector } from 'react-redux';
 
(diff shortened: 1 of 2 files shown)


kmer2016 (approved):
LGTM !

Akctarus (commented):
tested, lgtm

PR #10626, opened 2025-02-03
front: fix spacetimechart redraw after waypoint visibility change

### LINKED ISSUE #10609: GET : Hiding a PR does not erase the associated horizontal line
### What happened?

When hiding a PR in the GET, the associated horizontal line does not disapper with it.

It diseappear if you navigate through the GET which forces it to repaint the component.

https://github.com/user-attachments/assets/8dad5434-4aaf-4b50-92e7-7fc18343443c

The line should disappear right after the click on the hiding button.

1. Go to the Horizon module, open a study with trains in it
2. Select a train to project the GET on
3. Hide a PR

Recette (SNCF)

Firefox

fff6686

RomainValls:
closes #10609 

I'm not sure if we should fix this in osrd or osrd-ui ?


emersion (commented):
This sounds like an osrd-ui bug. Forcing a redraw is a workaround IMHO.

clarani:
Agree with @emersion, it would be better to fix this in osrd-ui. 

Here is the expected workflow:
- hiding a PR => should trigger the hook useManchetteWithSpaceTimeChart
- useManchetteWithSpaceTimeChart gives a new list of PRs => should trigger the redrawn of the GET

Issue #10627, opened 2025-02-03
Map: rationalize fonts and sprites


Tristramg:
### Description and goal

Currently some fonts and sprites used on the map depend on an external server. We would like integrate all those assets here to avoid an external dependency.

This is also an occasion to check on what fonts we depend to rationalize them.

TODO:

- [ ] reduce the number of used fonts (in particular switch from Open Sans to Roboto)
- [ ] integrate svgs and sprites from https://github.com/nicolaswurtz/signalisation-rfn-svg/
    - [x] Filter to only needed icons (no signals, no bluescreen)
    - [x] Generate the sprite atlas
    - [x] use the assets in the front
    - [x] optional: switch to spreet for smaller atlases
    - [ ] ~~optional: switch to webp for smaller atlases~~
- [ ] fonts
  - [x] save the .pbf fonts in editoast/assets
  - [x] serve the fonts through editoast (in a similar manner as the sprites)
  - [ ] optional: consider pre-compressing the binary .pbf files
  - [x] probably better not: remove unused/empty unicode blocks
  - [x] use the fonts
  - [ ] probably later: change the fonts to plex to be coherent with the rest of the app

### Acceptance criteria

There should be no call to an external server for fonts and sprites

PR #10628, opened 2025-02-03
front: map: use only Roboto not OpenSans

Tristramg:
This reduces the number of fonts

Ref #10627


emersion:
Just to double check: is @thibautsailly aware of this change?

thibautsailly:
> Just to double check: is @thibautsailly aware of this change?

Yes, no problem. I will resume work on the map themes later and will make propositions for fonts then. For now Roboto is the easiest way to go.

Tristramg:
@emersion now that the blueprint has been removed, roboto was only used in dark mode, so I suppose it was just forgotten when migrating to roboto

emersion (approved):
LGTM!

PR #10655, opened 2025-02-03
editoast: adapt timetable endpoint and delete post endpoint

### LINKED ISSUE #10656: editoast: adapt timetable endpoint
Delete `POST /timetable/train_schedule` and adapt `GET /timetable/{id}` to return train_schedules.

younesschrifi:
close [#10656](https://github.com/OpenRailAssociation/osrd/issues/10656)

### DIFF
--- a/editoast/src/views/timetable.rs
+++ b/editoast/src/views/timetable.rs
@@ -12,7 +12,6 @@ use axum::Extension;
 use derivative::Derivative;
 use editoast_authz::BuiltinRole;
 use editoast_derive::EditoastError;
-use editoast_schemas::train_schedule::TrainScheduleBase;
 use itertools::Itertools;
 use serde::Deserialize;
 use serde::Serialize;
@@ -40,15 +39,20 @@ use crate::views::AuthorizationError;
 use crate::AppState;
 use crate::RetrieveBatch;
 use editoast_models::DbConnectionPoolV2;
+use editoast_schemas::train_schedule::TrainScheduleBase;
+
+use super::pagination::PaginatedList as _;
+use super::pagination::PaginationQueryParams;
+use super::pagination::PaginationStats;
 
 crate::routes! {
     "/timetable" => {
         post,
         "/{id}" => {
             delete,
-            get,
+            "/train_schedules" => get,
+            "/train_schedules" => train_schedule,
             "/conflicts" => conflicts,
-            "/train_schedule" => train_schedule,
             &stdcm,
         },
     },
@@ -112,13 +116,22 @@ struct TimetableIdParam {
     id: i64,
 }
 
+#[derive(Serialize, ToSchema, Debug)]
+#[cfg_attr(test, derive(Deserialize))]
+struct ListTrainSchedulesResponse {
+    #[schema(value_type = Vec<TrainScheduleResult>)]
+    results: Vec<TrainScheduleResult>,
+    #[serde(flatten)]
+    stats: PaginationStats,
+}
+
 /// Return a specific timetable with its associated schedules
 #[utoipa::path(
     get, path = "",
     tag = "timetable",
-    params(TimetableIdParam),
+    params(TimetableIdParam, PaginationQueryParams),
     responses(
-        (status = 200, description = "Timetable with train schedules ids", body = TimetableDetailedResult),
+        (status = 200, description = "Timetable with train schedules ids", body = inline(ListTrainSchedulesResponse)),
         (status = 404, description = "Timetable not found"),
     ),
 )]
@@ -126,7 +139,8 @@ async fn get(
     State(db_pool): State<DbConnectionPoolV2>,
     Extension(auth): AuthenticationExt,
     Path(TimetableIdParam { id: timetable_id }): Path<TimetableIdParam>,
-) -> Result<Json<TimetableDetailedResult>> {
+    Query(pagination_params): Query<PaginationQueryParams>,
+) -> Result<Json<ListTrainSchedulesResponse>> {
     let authorized = auth
         .check_roles([BuiltinRole::TimetableRead].into())
         .await
@@ -136,11 +150,20 @@ async fn get(
     }
 
     let conn = &mut db_pool.get().await?;
-    let timetable = TimetableWithTrains::retrieve_or_fail(conn, timetable_id, || {
-        TimetableError::NotFound { timetable_id }
+    Timetable::retrieve_or_fail(conn, timetable_id, || TimetableError::NotFound {
+        timetable_id,
     })
     .await?;
-    Ok(Json(timetable.into()))
+
+    let settings = pagination_params
+        .validate(25)?
+        .into_selection_settings()
+        .filter(move || TrainSchedule::TIMETABLE_ID.eq(timetable_id));
+
+    let (train_schedules, stats) = TrainSchedule::list_paginated(conn, settings).await?;
+    let results = train_schedules.into_iter().map_into().collect();
+
+    Ok(Json(ListTrainSchedulesResponse { stats, results }))
 }
 
 /// Create a timetable
@@ -362,24 +385,17 @@ mod tests {
 
         let timetable = create_timetable(&mut pool.get_ok()).await;
 
-        let request = app.get(&format!("/timetable/{}", timetable.id));
+        let request = app.get(&format!("/timetable/{}/train_schedules", timetable.id));
 
-        let timetable_from_response: TimetableDetailedResult =
+        let timetable_from_response: ListTrainSchedulesResponse =
             app.fetch(request).assert_status(StatusCode::OK).json_into();
-
-        assert_eq!(
-            timetable_from_response,
-            TimetableDetailedResult {
-                timetable_id: timetable.id,
-                train_ids: vec![],
-            }
-        );
+        assert_eq!(timetable_from_response.results.len(), 0);
     }
 
     #[rstest]
     async fn get_unexisting_timetable() {
         let app = TestAppBuilder::default_app();
-        let request = app.get(&format!("/timetable/{}", 0));
+        let request = app.get(&format!("/timetable/{}/train_schedules", 0));
         app.fetch(request).assert_status(StatusCode::NOT_FOUND);
     }
 
--- a/front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
+++ b/front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
@@ -19,11 +19,7 @@ import ImportTrainSchedule from 'applications/operationalStudies/views/ImportTra
 import ManageTrainSchedule from 'applications/operationalStudies/views/ManageTrainSchedule';
 import SimulationResults from 'applications/operationalStudies/views/SimulationResults';
 import { osrdEditoastApi } from 'common/api/osrdEditoastApi';
-import type {
-  InfraWithState,
-  ScenarioResponse,
-  TimetableDetailedResult,
-} from 'common/api/osrdEditoastApi';
+import type { InfraWithState, ScenarioResponse } from 'common/api/osrdEditoastApi';
 import ScenarioLoaderMessage from 'modules/scenario/components/ScenarioLoaderMessage';
 import TimetableManageTrainSchedule from 'modules/trainschedule/components/ManageTrainSchedule/TimetableManageTrainSchedule';
 import Timetable from 'modules/trainschedule/components/Timetable/Timetable';
@@ -35,14 +31,12 @@ import MacroEditorState from '../MacroEditor/MacroEditorState';
 
 type ScenarioDescriptionProps = {
   scenario: ScenarioResponse;
-  timetable: TimetableDetailedResult;
   infra: InfraWithState;
   infraMetadata: { isInfraLoaded: boolean; reloadCount: number };
 };
 
 const ScenarioContent = ({
   scenario,
-  timetable,
   infra,
   infraMetadata: { isInfraLoaded, reloadCount },
 }: ScenarioDescriptionProps) => {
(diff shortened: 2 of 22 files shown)


editoast/src/views/timetable.rs
```diff
@@ -126,21 +130,41 @@ async fn get(
...
     Extension(auth): AuthenticationExt,
     Path(TimetableIdParam { id: timetable_id }): Path<TimetableIdParam>,
-) -> Result<Json<TimetableDetailedResult>> {
+    Query(pagination_params): Query<PaginationQueryParams>,
+) -> Result<Json<ListTrainSchedulesResponse>>  {
     let authorized = auth
-        .check_roles([BuiltinRole::TimetableRead].into())
-        .await
-        .map_err(AuthorizationError::AuthError)?;
+    .check_roles([BuiltinRole::InfraRead, BuiltinRole::TimetableRead].into())
```
flomonster: Why should we add `InfraRead` role?

editoast/src/views/timetable.rs
```diff
@@ -114,7 +114,7 @@ struct TimetableIdParam {
 
 /// Return a specific timetable with its associated schedules
 #[utoipa::path(
-    get, path = "",
+    get, path = "/train_schedule",
```
```suggestion
    get, path = "/train_schedules",
```

we mostly pluralize paths (not consistent though)

editoast/src/views/timetable.rs
```diff
@@ -126,21 +126,41 @@ async fn get(
...
-    let timetable = TimetableWithTrains::retrieve_or_fail(conn, timetable_id, || {
+    Timetable::retrieve_or_fail(conn, timetable_id, || {
         TimetableError::NotFound { timetable_id }
     })
     .await?;
-    Ok(Json(timetable.into()))
+
+    let settings = pagination_params
+    .validate(1000)?
+    .warn_page_size(100)
```
leovalais: Let's enforce pagination for new endpoints. What about

```suggestion
    .validate(25)?
```

?

editoast/src/views/timetable.rs
```diff
@@ -202,50 +222,6 @@ async fn delete(
     Ok(StatusCode::NO_CONTENT)
 }
 
-/// Create train schedule by batch
```
leovalais: Is this the right post endpoint? 🤔

editoast/src/models/timetable.rs
```diff
@@ -58,6 +58,20 @@ impl Timetable {
...
+
+
+        dsl::train_schedule.filter(dsl::timetable_id.eq(timetable_id))
+        .load_stream::<Row<TrainSchedule>>(conn.write().await.deref_mut())
+            .await?
+            .map_ok(|ts| ts.into())
+            .try_collect::<Vec<TrainSchedule>>()
+            .await.map_err(Into::into)
+    }
+
```
leovalais: I can't see where this function is used in this PR. If it's used, you can use `models::prelude::List` for that, or better delete it if it's unused ^^

editoast/src/views/timetable.rs
younesschrifi: oups

editoast/src/models/timetable.rs
younesschrifi: It was a first try that was not conclusive ^^

editoast/src/views/timetable.rs
```diff
@@ -114,7 +118,7 @@ struct TimetableIdParam {
 
 /// Return a specific timetable with its associated schedules
 #[utoipa::path(
-    get, path = "",
+    get, path = "/train_schedules",
```
woshilapin: Does this really work (cc @leovalais)? I was under the impression that with our macro's system, paths need to be defined in `crate::routes!()` (the macro defines both OpenAPI and `axum`'s router, so you might be missing the `axum` part?). At least, when testing, `/timetable/1` returns the entire payload, but `/timetable/1/train_schedules` returns a 404.

editoast/src/views/timetable.rs
```diff
@@ -126,21 +130,40 @@ async fn get(
...
+}
+
+#[derive(Serialize, ToSchema)]
+#[cfg_attr(test, derive(Deserialize))]
+struct ListTrainSchedulesResponse {
+    #[schema(value_type = Vec<TrainScheduleResult>)]
+    results: Vec<TrainScheduleResult>,
+    #[serde(flatten)]
+    stats: PaginationStats,
 }
```
woshilapin: nitpick: how about moving the `struct` above the function where it is used (I believe I can usually expect to find the `struct` used in a function by looking before the function).

woshilapin (commented):
You might be missing something about the new endpoint?

editoast/src/views/timetable.rs
leovalais: No it doesn't you're completely right, I should stop reviewing PRs late... 🤦 

It should indeed be `""` and put the path suffix in the `routes!` macro. Not sure how easy it would be, but triggering an error when the path is non-empty may be a good idea.

front/src/modules/trainschedule/components/Timetable/TimetableTrainCard.tsx
```diff
@@ -106,14 +106,14 @@ const TimetableTrainCard = ({
...
+    const trainDetail = await getTrainSchedule({ id: train.id })
       .unwrap()
       .catch((e) => {
         dispatch(setFailure(castErrorToFailure(e)));
       });
 
-    if (trainsResults) {
-      const trainDetail = trainsResults[0];
+    if (trainDetail) {
+      // const trainDetail = trainsResults[0];
```
emersion: Nit: this can be dropped.

front/src/modules/scenario/components/ScenarioExplorer/ScenarioExplorer.tsx
```diff
@@ -176,7 +164,7 @@ const ScenarioExplorer = ({
                 </span>
 
                 <span className="scenario-explorator-card-head-scenario-traincount">
-                  {timetable && trainCount(timetable.train_ids)}
+                  {timetable && trainCount(timetable.results.map((r) => r.id))}
```
emersion: TBH I think we can just drop the `trainCount()` function. `timetable.results.length` should work fine.

(Alternatively, if keeping it is deemed better, change its type.)

front/src/applications/operationalStudies/hooks/useScenarioData.ts
```diff
@@ -22,18 +22,20 @@ import formatTrainScheduleSummaries from '../helpers/formatTrainScheduleSummarie
 
 const useScenarioData = (
   scenario: ScenarioResponse,
-  timetable: TimetableDetailedResult,
+  timetable: GetTimetableByIdTrainSchedulesApiResponse,
```
emersion: Could we use `TrainScheduleResult[]` instead?

front/src/modules/scenario/components/ScenarioExplorer/ScenarioExplorer.tsx
```diff
@@ -60,23 +60,11 @@ const ScenarioExplorer = ({
       }
     );
 
-  const { data: timetable } = osrdEditoastApi.endpoints.getTimetableById.useQuery(
+  const { data: timetable } = osrdEditoastApi.endpoints.getTimetableByIdTrainSchedules.useQuery(
```
emersion: Can be simplified with further destructuring:

```ts
const { data: { results: trainsDetails } } = …
```

(I used `trainsDetails` to re-use the same name as the original code and reduce the diff, but `trainSchedules` would probably be better. `timetable` I'm less a fan of because a timetable will contain stuff which isn't train schedules soon: paced trains.)

The same comment applies elsewhere in the PR.

front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
```diff
@@ -68,28 +68,22 @@ const ScenarioContent = ({
...
-        { id: timetable.timetable_id },
-        { forceRefetch: true, subscribe: false }
-      )
-    );
-    const { train_ids } = await timetablePromise.unwrap();
     const trainSchedulesPromise = dispatch(
-      osrdEditoastApi.endpoints.postTrainSchedule.initiate(
-        { body: { ids: train_ids } },
+      osrdEditoastApi.endpoints.getTimetableByIdTrainSchedules.initiate(
+        { id: timetable.results[0].timetable_id }, // TODO
```
emersion: I think we can use `scenario` here instead to grab the timetable ID. That way, it doesn't break with empty timetables.

front/src/applications/operationalStudies/hooks/useScenario.ts
```diff
@@ -49,7 +49,7 @@ const useScenario = () => {
     }
   );
 
-  const { data: timetable } = osrdEditoastApi.endpoints.getTimetableById.useQuery(
+  const { data: timetable } = osrdEditoastApi.endpoints.getTimetableByIdTrainSchedules.useQuery(
```
emersion: Ditto, it'll be cleaner to extract the `results` here directly.

woshilapin (approved):

editoast/src/views/timetable.rs
```diff
@@ -30,25 +29,30 @@ use crate::models::prelude::*;
...
 
 crate::routes! {
     "/timetable" => {
         post,
         "/{id}" => {
             delete,
-            get,
-            "/conflicts" => conflicts,
+            "/train_schedules" => get,
             "/train_schedule" => train_schedule,
```
leovalais: Not sure how easy it would be, but it'd be great to have both GET and POST routes under `/train_schedules`. That'll probably imply some changes to the frontend since the generated function name will change a bit, but since there's already front changes in this PR, maybe this could be done here?

editoast/src/views/timetable.rs
leovalais: The doc comment hasn't been restored

flomonster (approved):

shenriotpro (approved):
LGTM for tests/

Math-R (approved):
lgtm !

front/src/common/api/osrdEditoastApi.ts
```diff
@@ -1,136 +1,178 @@
...
+          }
+          return { data: result };
+        },
+        providesTags: (result) =>
+          result
+            ? [
+                ...result.map(({ id }) => ({ type: 'train_schedule' as const, id })),
+                { type: 'train_schedule', id: 'LIST' },
+              ]
+            : [{ type: 'train_schedule', id: 'LIST' }],
```
emersion: We don't use fine-grained tags for train schedules: mutations don't invalidate the tags mentioned here. I think this should just be `providesTags: ['timetable']`

Maymanaf (approved):

front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
```diff
@@ -69,27 +69,20 @@ const ScenarioContent = ({
...
+    const schedules = (await timetablePromise.unwrap()).filter(
       (trainSchedule) => trainSchedule.path.length >= 2
     );
     const state = new MacroEditorState(scenario, schedules || []);
     await loadAndIndexNge(state, dispatch);
     const dto = getNgeDto(state);
     macroEditorState.current = state;
     setNgeDto(dto);
-  }, [dispatch, scenario, timetable.timetable_id]);
+  }, [dispatch, scenario, scenario.timetable_id, timetable]);
```
emersion: This callback doesn't depend on `timetable`.

front/src/applications/operationalStudies/hooks/useScenarioData.ts
```diff
@@ -271,7 +273,7 @@ const useScenarioData = (
       // fetch conflicts
       refetchConflicts();
     },
-    [trainSchedules, rollingStocks]
+    [trainSchedulesResults, rollingStocks]
```
emersion: This callback only depends on `trainSchedules`, it never reads from `trainSchedulesResults`.

front/src/applications/operationalStudies/hooks/useScenarioData.ts
```diff
@@ -110,14 +111,16 @@ const useScenarioData = (
...
+    trainScheduleSummaries
+  );
 
   useEffect(() => {
     const fetchTrains = async () => {
-      const rawTrainSchedules = await fetchTrainSchedules({
-        body: {
-          ids: timetable.train_ids,
-        },
+      const rawTrainSchedules = await fetchAllTimetableByIdTrainSchedules({
```
emersion: This hook now takes the list of train schedules as input. Here we're fetching the train schedules again.

There are two ways to fix this: either fetch the train schedules once in `useScenario()`, either fetch them once in `useScenarioData()`. I have a small preference for fetching them in `useScenarioData()` personally.

To fetch train schedules in `useScenarioData()`, we can drop `timetable` from the values returned by `useScenario()`, remove `trainSchedulesResults` from the props here, and then use `trainSchedules` instead of `trainSchedulesResults`.

front/src/applications/operationalStudies/hooks/useScenarioData.ts
```diff
@@ -110,14 +111,16 @@ const useScenarioData = (
   );
 
   // TODO Paced train : Adapt this to accept paced trains in issue https://github.com/OpenRailAssociation/osrd/issues/10613
-  useAutoUpdateProjection(infra, timetable.train_ids, trainScheduleSummaries);
+  useAutoUpdateProjection(
+    infra,
+    trainSchedulesResults.map((t) => t.id),
```
emersion: Nit: this passes a new value to the hook each time a re-render happens. This happens to be mostly fine by chance because `useAutoUpdateProjection()` doesn't perform any heavyweight action when `editoastTrainIds` changes (apart from checking whether the selected/projected train still exist), but that may change in the future and will not be easily noticeable since it's a completely separate file.

I'd suggest a `useMemo()` to make sure the value is stable when the train schedules don't change.

emersion (commented):
Sorry, I had one last look and got a few more comments!

emersion (approved):
LGTM, thanks!

emersion:
Please make sure to squash the commits before merging!

PR #10657, opened 2025-02-04
tests: add infra to test stdcm takeover setups

### LINKED ISSUE #10570: stdcm: test that an extra fake track actually works at faking stops
See https://github.com/osrd-project/osrd-dags/issues/675

Before tweaking with the data, we need to see if it actually works.

We'd need to:

Create an infra with one long line and several signals, and one side track in the middle:

```
          ___
         /    \ 
________/______\_________
```

The tracks before and after need to be very long with several signals every 2km.

We then add signals and detectors: one at the start of the side-track, then 3 at the end, 1m apart.

We create a timetable where lots of close trains on the main track, too close for a new train. One opening on the first half and one opening on the second, but far apart, so we need to stop.

Then, we need to find a solution where it goes on the side track.

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/10570

I'll write the conclusions in the original issue

### DIFF
--- a/tests/data/infras/takeover/infra.json
+++ b/tests/data/infras/takeover/infra.json
@@ -1,0 +1,1306 @@
+{
+    "version": "3.4.13",
+    "operational_points": [
+        {
+            "id": "op.start",
+            "parts": [
+                {
+                    "track": "t_1",
+                    "position": 0.0
+                }
+            ],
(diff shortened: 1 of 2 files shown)


shenriotpro (approved):
Thanks

PR #10659, opened 2025-02-04
core: stdcm: increase min allowance speed

eckter:
That value was higher to avoid issues caused by a different bug that was fixed since then.

Required for https://github.com/osrd-project/osrd-dags/issues/675


Khoyo (approved):
I believe in some case the speed can be even lower (during certain maneuvers), but I don't think it applies here

PR #10671, opened 2025-02-04
front: switch package type to module

emersion:
_See individual commits._

Modules are the standard way to handle JavaScript imports. osrd-ui already uses modules everywhere, but osrd is lagging behind.

Modernize our configuration to use ECMAScript Modules instead of legacy CommonJS.


front/tests/004-scenario-management.spec.ts
```diff
@@ -2,15 +2,16 @@ import { expect } from '@playwright/test';
...
 import { infrastructureName } from './assets/project-const';
 import test from './logging-fixture';
 import ScenarioPage from './pages/scenario-page-model';
-import { generateUniqueName } from './utils';
+import { generateUniqueName, readJsonFile } from './utils';
 import { deleteApiRequest, getProject, getStudy, setElectricalProfile } from './utils/api-setup';
 import createScenario from './utils/scenario';
 import { deleteScenario } from './utils/teardown-utils';
 
+const scenarioData = readJsonFile('tests/assets/operationStudies/scenario.json');
```
SharglutDev, line 13: Shouldn't we do that for all json assets we import ? Like in `OSM.tsx` or it's only for some special kind of data ?

emersion: I think it depends. Stuff in `src/` is built via vite, which handles JSON imports. Stuff in `tests/` uses Nodejs directly, where JSON imports are more complicated to handle (and break here). Also, in `src/` any `import` will include the target in the final bundle, which is good for small files and bad for big files.

SharglutDev: Ok got it :)

SharglutDev (approved):
Lgtm and tested (didn't launch the e2e tests)

front/tests/013-stdcm-simulation-sheet.spec.ts
```diff
@@ -13,6 +13,30 @@ import { getInfra } from './utils/api-setup';
 import { findFirstPdf, verifySimulationContent } from './utils/simulationSheet';
 import type { ConsistFields, Simulation } from './utils/types';
 
+const parsePdfText = async (buffer: Buffer) => {
```
Maymanaf: This should be a utility function, not defined inside the test file

Maymanaf (commented):
Good job ! 
However, the test appears to be flaky when using the `pdfjs-dist` library
image

front/tests/013-stdcm-simulation-sheet.spec.ts
emersion: Done

Maymanaf (approved):
LGTM ✅

PR #10675, opened 2025-02-04
front: upgrade @rtk-query/codegen-openapi to v2.0.0 

emersion:
_See individual commit messages._

This required getting rid of our fork.

### DIFF
--- a/front/src/common/api/generatedEditoastApi.ts
+++ b/front/src/common/api/generatedEditoastApi.ts
@@ -69,7 +69,9 @@ const injectedRtkApi = api
           url: `/documents`,
           method: 'POST',
           body: queryArg.body,
-          headers: { content_type: queryArg.contentType },
+          headers: {
+            content_type: queryArg.contentType,
+          },
         }),
         invalidatesTags: ['documents'],
       }),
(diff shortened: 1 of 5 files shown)


gateway/openapi.yaml
```diff
@@ -86,8 +86,6 @@ components:
       oneOf:
         - $ref: "#/components/schemas/LoginSuccess"
         - $ref: "#/components/schemas/LoginRedirect"
-      discriminator:
```
anisometropie, line 89: This doesn’t seem to matter on typeScript side, but I’m not sure what consequences it has for the gateway itself

emersion: The OpenAPI schema is only used to generate TypeScript AFAIK.

anisometropie (approved):
Tested and looking good

front/src/common/api/generatedEditoastApi.ts
SharglutDev, line 1: Does the re-indenting in this file comes from the bump of the lib ? Or you did a ctrl+s in there ?

SharglutDev (approved):
Lgtm and tested, just one question

front/src/common/api/generatedEditoastApi.ts
emersion: It's due to the code generator upgrade, I believe. I didn't edit this file by hand (and CI would catch this).

ghost (approved):
LGTM

PR #10676, opened 2025-02-04
front: fix graou schedules import

### LINKED ISSUE #10501: Error during import - Failed to parse duration
### What happened?

When importing data  from open data with PLY as origin and LYD as destination for 23/01/2025, I received an error in the console: 
"Failed to deserialize the JSON body into the target type: [6].schedule[0].arrival: Failed to parse duration: -PT30S at line 1 column 2584."
This error seems to mainly affect TER trains, where a negative duration (-PT30S) was not parsed correctly.

I expected the import to complete successfully without any errors.

1. Go to scenario page.
2. Select PLY as origin and LYD as destination.
3. Set the date and time range to 23/01/2025 from 00:00 to 23:59.
4. Click the magnifying glass to fetch the data.
5. Click "Démarrer l'importation"
6. The error will appear in the console.

Local

Google Chrome

8d71c730

Synar:
Close #10501

### DIFF
--- a/front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
+++ b/front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
@@ -53,6 +53,20 @@ const ImportTrainScheduleConfig = ({
   const dispatch = useAppDispatch();
   const { openModal, closeModal } = useContext(ModalContext);
 
+  function filterInvalidSteps(
+    importedTrainSchedules: ImportedTrainSchedule[]
+  ): ImportedTrainSchedule[] {
+    return importedTrainSchedules.map((trainSchedule) => ({
+      ...trainSchedule,
+      steps: trainSchedule.steps.filter(
+        (step, i) =>
+          i === 0 ||
+          new Date(step.arrivalTime).getTime() >=
+            new Date(trainSchedule.steps[i - 1].departureTime).getTime()
+      ),
+    }));
+  }
+
   function validateImportedTrainSchedules(
     importedTrainSchedules: Record<string, unknown>[]
   ): ImportedTrainSchedule[] | null {
@@ -81,7 +95,7 @@ const ImportTrainScheduleConfig = ({
       );
       return null;
     }
-    return importedTrainSchedules as ImportedTrainSchedule[];
+    return filterInvalidSteps(importedTrainSchedules as ImportedTrainSchedule[]);
   }
 
   function updateTrainSchedules(importedTrainSchedules: ImportedTrainSchedule[]) {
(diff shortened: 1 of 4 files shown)


front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
```diff
@@ -53,6 +53,20 @@ const ImportTrainScheduleConfig = ({
   const dispatch = useAppDispatch();
   const { openModal, closeModal } = useContext(ModalContext);
 
+  function filterInvalidSteps(
+    importedTrainSchedules: ImportedTrainSchedule[]
+  ): ImportedTrainSchedule[] {
+    return importedTrainSchedules.map((trainSchedule) => ({
+      ...trainSchedule,
+      steps: trainSchedule.steps.filter(
```
emersion: Should we really completely drop these steps? Or should we drop only the `arrival` field?

Synar: Good question. In the graou import it seems that the issue is not with the arrival times themselves, but with repeated or malformed, extraneous steps. We could also instead try and detect that, but the acceptance criteria would not be clear (do we start rejecting any step with empty uic for example?). We could also just skip the train entirely, since it clearly contains bad data. At the end of the day the bug comes from bad data from Graou's side, so I don't see a perfect fix on our end.

In any case, maybe we should display a warning for the users to know that some trains/steps needed to be modified/discarded?

clarani: Yes displaying a warning may be a good thing in any case.
Deleting problematic steps is okay IMHO

clarani (approved):
Code LGTM :white_check_mark:

RomainValls (approved):
Lgtm, tested ✅

front/src/modules/trainschedule/components/ImportTrainSchedule/ImportTrainScheduleConfig.tsx
emersion: Works for me!

Synar: Warning added!

Synar: (Also technically we could try to recover bad data by looking at similar schedules in the same import and extrapolating from them, it may be worth doing if this is a critical feature, but otherwise...)

clarani (approved):
LGTM :white_check_mark:

PR #10677, opened 2025-02-04
front: refacto: split simulation report sheet files to improve readability

### LINKED ISSUE #9423: Split simulation sheet (STDCM/Scenario)

Software developers

wait until [PR](https://github.com/OpenRailAssociation/osrd/pull/9287) is finished

Split the huge simulation file into several smaller components

-easier reading
-simpler review

Akctarus:
closes #9423 

I have some questions : 
- Are the names of my folders and new components correct? 
- Are the report sheet cut out enough/too much? 
- Do I also split the style file?


front/src/applications/stdcm/components/StdcmResults/SimulationReportSheet/modules/LogoLMR.tsx
```diff
@@ -0,0 +1,18 @@
+import { Text, Image } from '@react-pdf/renderer';
+import type { TFunction } from 'i18next';
+
+import styles from '../SimulationReportStyleSheet';
+
+const LogoLMR = ({ logo, t }: { logo: string | undefined; t: TFunction }) => {
```
emersion: "LMR" is a SNCF-specific term, we should use "STDCM" in the open-source application.

emersion (commented):
Is this PR just moving things around, or also doing other things? I see some "convoy" occurences are renamed to "consist" (which seems to be the correct term). Are there other changes as well? (It's in general difficult to review commits with both code block moves and logic changes, I'd like to know if I should be paying attention.)

I don't think we should nest the components that deep in directories. Suggestion: put everything in a `front/src/applications/stdcm/components/SimulationReportSheet/` directory.

Not sure `Footer`/`Map` is worth splitting off, since it's just 3 lines of code and used only once.

front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
```diff
@@ -19,7 +19,7 @@ import {
 } from 'reducers/osrdconf/stdcmConf/selectors';
 import useDeploymentSettings from 'utils/hooks/useDeploymentSettings';
 
-import SimulationReportSheet from './SimulationReportSheet';
+import SimulationReportSheet from './SimulationReportSheet/SimulationReportSheet';
```
clarani: You can add an index to avoid repetitions in the import

clarani (commented):
Nice work :)

I think you can create a new folder `src/modules/simulationResults/components/SimulationReportSheet` and put all the code related to the ReportSheet in it (the SimulationReportSheet is also used in the operationalStudies part right ?)

I agree with Simont's comments

Akctarus:
> Is this PR just moving things around, or also doing other things? I see some "convoy" occurences are renamed to "consist" (which seems to be the correct term). Are there other changes as well? (It's in general difficult to review commits with both code block moves and logic changes, I'd like to know if I should be paying attention.)
> 
> I don't think we should nest the components that deep in directories. Suggestion: put everything in a `front/src/applications/stdcm/components/SimulationReportSheet/` directory.
> 
> Not sure `Footer`/`Map` is worth splitting off, since it's just 3 lines of code and used only once.

Ty for your comments ! In this PR I just move things around, I did rename convoy to consist, but that's the only change I made.

front/src/applications/stdcm/components/StdcmResults/SimulationReportSheet/modules/LogoLMR.tsx
Akctarus: fixed !

front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
Akctarus: Fixed ! 
unfortunately, the components are not identical between the 2 simulation sheets (especially the results table, which is the longest), so I can't define just one sheet and call it up in operational studies and stdcm.

front/src/applications/stdcm/components/SimulationReportSheet/components/LogoSTDCM.tsx
```diff
@@ -0,0 +1,18 @@
...
+import styles from '../SimulationReportStyleSheet';
+
+const LogoSTDCM = ({ logo, t }: { logo: string | undefined; t: TFunction }) => {
+  if (logo) {
+    return <Image src={logo} style={styles.header.lmrLogo} />;
+  }
+  return (
+    <>
+      <Text style={styles.header.title}>{t('stdcm')}</Text>
+      <Text style={styles.header.creation}>{t('stdcmCreation')}</Text>
```
Caracol3: I think 'creation' should be changed to 'simulation' here. Same in the two corresponding JSON.

Caracol3 (commented):
Nice work ! 
Juste a small change.

front/src/applications/stdcm/components/SimulationReportSheet/components/LogoSTDCM.tsx
Akctarus: fixed !

emersion (commented):
I still see some pretty long paths, e.g. `front/src/applications/stdcm/components/SimulationReportSheet/components/Consist.tsx`. My earlier suggestion was to drop the intermediary `components` path: `front/src/applications/stdcm/components/SimulationReportSheet/Consist.tsx`.

Akctarus:
> I still see some pretty long paths, e.g. `front/src/applications/stdcm/components/SimulationReportSheet/components/Consist.tsx`. My earlier suggestion was to drop the intermediary `components` path: `front/src/applications/stdcm/components/SimulationReportSheet/Consist.tsx`.

Fixed in : 7316b6702f72d0f9f513512ba5cd65ca5d778c2f and 2281f0c5abd945ec1e73dcfb54c4ab9ba5acfe57

clarani (commented):
Some files are duplicated but it should not be necessary. Can you try to merge the components which are both in `stdcm/components/SimulationReportSheet` and in `modules/SimulationResults/components/SimulationReportSheetScenario` ?

Also, can you rename the `SimulationReportSheetScenario` -> `SimulationReportSheet` in the components of the module SimulationResults ?

front/src/modules/simulationResult/types.ts
```diff
@@ -69,3 +78,21 @@ export type AspectLabel =
   | '160A'
   | '080A'
   | '000';
+
+export type SimulationReportSheetScenarioProps = {
```
emersion, line 82: Props should be near the component definition.

front/src/modules/simulationResult/components/SimulationReportSheet/SimulationTable.tsx
```diff
@@ -0,0 +1,208 @@
...
+                        }
+                      : !isViaWithoutStop
+                        ? styles.simulation.stopColumn
+                        : { ...styles.simulation.stopColumn, marginLeft: '' }),
+                  }}
+                >
+                  {isNotExtremity
+                    ? step.duration !== 0
+                      ? getStopDurationTime(
+                          step.duration instanceof Duration
```
emersion, line 157: This isn't the original file's code.

I think some changes got lost in a rebase.

emersion:
I am a bit worried about loosing some changes in this PR. This PR is especially difficult to review because the new files need to be compared line-by-line with the old files to ensure the logic didn't change.

Akctarus:
> I am a bit worried about loosing some changes in this PR. This PR is especially difficult to review because the new files need to be compared line-by-line with the old files to ensure the logic didn't change.

I totally agree, I'm trying my best not to have any regression, but the OP types of operationalStudies and STDCM are not the same, so I need to find a solution to combine the identical code, while keeping a differentiation between the two.

Akctarus:
Too old PR. New one here : https://github.com/OpenRailAssociation/osrd/pull/11820

PR #10678, opened 2025-02-04
Editoast improve valkey connection

flomonster:
What is done:

1. Remove cluster client
2. Use a pool of connection

Performances are unchanged.

[traces.zip](https://github.com/user-attachments/files/18672744/traces.zip)


editoast/src/valkey_utils.rs
```diff
@@ -286,30 +282,22 @@ impl ValkeyClient {
...
-                client.get_connection_manager().await?,
-            )),
+            ValkeyClient::Tokio(pool) => Ok(ValkeyConnection::Tokio(pool.get().await?)),
             ValkeyClient::NoCache => Ok(ValkeyConnection::NoCache),
         }
     }
 
     #[tracing::instrument(skip_all)]
-    pub async fn ping_valkey(&self) -> RedisResult<()> {
+    pub async fn ping_valkey(&self) -> anyhow::Result<()> {
```
leovalais, line 300: Wouldn't this work?

```suggestion
    pub async fn ping_valkey(&self) -> Result<(), Either<PoolError, RedisError>> {
```

leovalais:
Shouldn't we add a parameter for the number of connections in the pool? The default seems to be `num_physical_cpus * 4`, but maybe that's too low for what we're trying to do? https://docs.rs/deadpool-redis/latest/deadpool_redis/struct.PoolConfig.html#structfield.max_size

I'm baffled that parallelizing redis calls do not save us transport time 😮 What does the trace timings look like?

hamz2a (approved):
LGTM, thanks!

leovalais (approved):

PR #10679, opened 2025-02-04
change trainid format

### LINKED ISSUE #10601: Change how trainschedule id is stored in the front part
### Description and goal

With the train service / paced train feature incoming, we will now manipulate 2 sorts of trains in the app : "unique" trains and "paced trains".

Paced trains regroup many trains with a `start_time` equally staggered. We gonna need to be able to select every of the paced train's children on the timetable or on the space time chart and manipulate (drag on space time chart).

The back-end will provide an id for a paced train but not for its "children". To be aware of which one of them we manipulate, we need to change how we store their ids in the `TrainScheduleWithDetails` type.
The new id will now be of type `string` instead of `number` and will look like this : 
- train id : `trainschedule-{id}`
- occurrence (paced train child) : `occurrence-{occurrenceIndex}-paced-{id}`

We also need to have another type of id for paced trains that will be used for the projection to differentiate it with a train schedule id.
paced train id : `paced-{id}`

To improve safety, we should also be able to distinguish these two kinds of ids.
Goal is to use the logic describe in this issue : https://github.com/OpenRailAssociation/osrd/issues/8816.
These news types should look like this : 
```
type TrainScheduleId = string & { readonly __type: unique symbol }; // format: "trainschedule-{id}"
type OccurrenceId = string & { readonly __type: unique symbol }; // format: "occurrence-{occurrenceIndex}-paced-{id}"
type PacedTrainId = string & { readonly __type: unique symbol }; // format: "paced-{id}"

type TrainId = TrainScheduleId | OccurrenceId; // TODO: find a better name
type TimetableItemId = TrainScheduleId | PacedTrainId;
```

> [!WARNING]
> You might need to reset your store / local storage to test the PR and after testing it

### Acceptance criteria

- [ ] The new train id types have been created
- [ ] `TrainScheduleWithDetails` type `id` property is now of type `TrainId`
- [ ] When fetching current `trainSchedule`, the editoast `id` is transformed to match the new id string format
- [ ] Everywhere we use or compare train ids has been updated
- [ ] When a train is selected, the url displays the new id format for selection and projection
- [ ] When "CRUD"ing a `trainSchedule`, the "string" id is properly transformed into the editoast "number" id
- [ ] Macro > micro and micro > macro should still work
- [ ] Deleting the selected train and/or the projected one should still work as before (update url and fallback into another train)
- [ ] Selecting a train on the space time chart should still work
- [ ] Clicking on a conflict card should still change the selected train (if the conflicted one wasn't selected)
- [ ] Stdcm debug mode still works

SharglutDev:
close #10601

### DIFF
--- a/front/src/applications/operationalStudies/__tests__/sampleData.ts
+++ b/front/src/applications/operationalStudies/__tests__/sampleData.ts
@@ -13,6 +13,7 @@ import type {
   SimulationSummaryResult,
   TrainScheduleResult,
 } from 'common/api/osrdEditoastApi';
+import type { TrainScheduleId, TrainScheduleResultWithTrainId } from 'reducers/osrdconf/types';
 
 export const pathLength = 4000;
 
@@ -351,8 +352,8 @@ export const electrificationRangesLarge: ElectrificationRange[] = [
   },
 ];
 
-export const trainScheduleTooFast: TrainScheduleResult = {
-  id: 98,
+export const trainScheduleTooFast: TrainScheduleResultWithTrainId = {
+  id: 'trainschedule-98' as TrainScheduleId,
   timetable_id: 10,
   train_name: 'tooFast',
   labels: [],
@@ -418,8 +419,8 @@ export const trainSummaryTooFast: Extract<SimulationSummaryResult, { status: 'su
   path_item_times_base: [0, 1444453, 2491479],
 };
 
-export const trainScheduleNotHonored: TrainScheduleResult = {
-  id: 96,
+export const trainScheduleNotHonored: TrainScheduleResultWithTrainId = {
+  id: 'trainschedule-96' as TrainScheduleId,
   timetable_id: 10,
   train_name: 'notHonored',
   labels: [],
@@ -485,8 +486,8 @@ export const trainSummaryNotHonored: Extract<SimulationSummaryResult, { status: 
   path_item_times_base: [0, 1425534, 2186885],
 };
 
-export const trainScheduleHonored: TrainScheduleResult = {
-  id: 95,
+export const trainScheduleHonored: TrainScheduleResultWithTrainId = {
+  id: 'trainschedule-95' as TrainScheduleId,
   timetable_id: 10,
   train_name: 'normal',
   labels: [],
@@ -529,12 +530,12 @@ export const trainScheduleHonored: TrainScheduleResult = {
   },
 };
 
-export const trainScheduleNoSchedule: TrainScheduleResult = {
+export const trainScheduleNoSchedule: TrainScheduleResultWithTrainId = {
   ...trainScheduleHonored,
   schedule: undefined,
 };
 
-export const trainScheduleNoMatch: TrainScheduleResult = {
+export const trainScheduleNoMatch: TrainScheduleResultWithTrainId = {
   ...trainScheduleHonored,
   schedule: [
     {
--- a/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
@@ -29,8 +29,6 @@ import type {
   LabelDto,
 } from '../NGE/types';
 
-const createdTrainrun = new Map<number, number>();
-
 const getNodeById = (nodes: NodeDto[], nodeId: number | string) =>
   nodes.find((node) => node.id === nodeId);
 
@@ -268,16 +266,18 @@ const handleUpdateTrainSchedule = async ({
   dispatch,
   infraId,
   addUpsertedTrainSchedules,
+  state,
 }: {
   netzgrafikDto: NetzgrafikDto;
   trainrun: TrainrunDto;
   dispatch: AppDispatch;
   infraId: number;
   addUpsertedTrainSchedules: (trainSchedules: TrainScheduleResult[]) => void;
+  state: MacroEditorState;
 }) => {
   const { nodes, labels } = netzgrafikDto;
   const trainrunSectionsByTrainrunId = getTrainrunSectionsByTrainrunId(netzgrafikDto, trainrun.id);
-  const trainrunIdToUpdate = createdTrainrun.get(trainrun.id) || trainrun.id;
+  const trainrunIdToUpdate = state.trainScheduleIdByNgeId.get(trainrun.id);
   const trainSchedule = await dispatch(
     osrdEditoastApi.endpoints.getTrainScheduleById.initiate({
       id: trainrunIdToUpdate,
@@ -317,6 +317,7 @@ const handleTrainrunOperation = async ({
   netzgrafikDto,
   addUpsertedTrainSchedules,
   addDeletedTrainIds,
+  state,
 }: {
   type: NGEEvent['type'];
   trainrunId: number;
@@ -326,6 +327,7 @@ const handleTrainrunOperation = async ({
   netzgrafikDto: NetzgrafikDto;
   addUpsertedTrainSchedules: (trainSchedules: TrainScheduleResult[]) => void;
   addDeletedTrainIds: (trainIds: number[]) => void;
+  state: MacroEditorState;
 }) => {
   const { nodes, labels, trainruns } = netzgrafikDto;
   const trainrun = trainruns.find((tr) => tr.id === trainrunId);
@@ -355,19 +357,19 @@ const handleTrainrunOperation = async ({
           ],
         })
       ).unwrap();
-      createdTrainrun.set(trainrunId, newTrainSchedules[0].id);
+      state.trainScheduleIdByNgeId.set(trainrunId, newTrainSchedules[0].id);
       addUpsertedTrainSchedules(newTrainSchedules);
       break;
     }
     case 'delete': {
-      const trainrunIdToDelete = createdTrainrun.get(trainrunId) || trainrunId;
+      const trainrunIdToDelete = state.trainScheduleIdByNgeId.get(trainrunId)!;
       await dispatch(
         osrdEditoastApi.endpoints.deleteTrainSchedule.initiate({
           body: { ids: [trainrunIdToDelete] },
         })
       ).unwrap();
-      createdTrainrun.delete(trainrunId);
       addDeletedTrainIds([trainrunIdToDelete]);
+      state.trainScheduleIdByNgeId.delete(trainrunId);
       break;
     }
     case 'update': {
@@ -377,6 +379,7 @@ const handleTrainrunOperation = async ({
         dispatch,
         infraId,
         addUpsertedTrainSchedules,
+        state,
       });
       break;
     }
@@ -478,6 +481,7 @@ const handleLabelOperation = async ({
   dispatch,
   infraId,
   addUpsertedTrainSchedules,
+  state,
 }: {
   type: NGEEvent['type'];
   label: LabelDto;
@@ -485,6 +489,7 @@ const handleLabelOperation = async ({
   dispatch: AppDispatch;
   infraId: number;
   addUpsertedTrainSchedules: (trainSchedules: TrainScheduleResult[]) => void;
+  state: MacroEditorState;
 }) => {
   const { trainruns } = netzgrafikDto;
   switch (type) {
@@ -499,6 +504,7 @@ const handleLabelOperation = async ({
           dispatch,
           infraId,
           addUpsertedTrainSchedules,
+          state,
         });
       });
       break;
@@ -543,6 +549,7 @@ const handleOperation = async ({
         netzgrafikDto,
         addUpsertedTrainSchedules,
         addDeletedTrainIds,
+        state,
       });
       break;
     }
@@ -554,6 +561,7 @@ const handleOperation = async ({
         dispatch,
         infraId,
         addUpsertedTrainSchedules,
+        state,
       });
       break;
     default:
--- a/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
@@ -4,11 +4,19 @@ import {
   osrdEditoastApi,
   type SearchResultItemOperationalPoint,
   type TrainScheduleBase,
-  type TrainScheduleResult,
 } from 'common/api/osrdEditoastApi';
+import type {
+  TimetableItemId,
+  TrainScheduleId,
+  TrainScheduleResultWithTrainId,
+} from 'reducers/osrdconf/types';
 import type { AppDispatch } from 'store';
 import { formatToIsoDate } from 'utils/date';
 import { Duration } from 'utils/duration';
+import {
+  formatEditoastTrainIdToTrainScheduleId,
+  formatTrainScheduleIdToEditoastTrainId,
+} from 'utils/trainId';
 
 import { DEFAULT_TRAINRUN_FREQUENCIES, DEFAULT_TRAINRUN_FREQUENCY } from './consts';
 import type MacroEditorState from './MacroEditorState';
@@ -272,15 +280,18 @@ const handleUpdateTrainSchedule = async ({
   trainrun: TrainrunDto;
   dispatch: AppDispatch;
   infraId: number;
-  addUpsertedTrainSchedules: (trainSchedules: TrainScheduleResult[]) => void;
+  addUpsertedTrainSchedules: (trainSchedules: TrainScheduleResultWithTrainId[]) => void;
   state: MacroEditorState;
 }) => {
   const { nodes, labels } = netzgrafikDto;
   const trainrunSectionsByTrainrunId = getTrainrunSectionsByTrainrunId(netzgrafikDto, trainrun.id);
-  const trainrunIdToUpdate = state.trainScheduleIdByNgeId.get(trainrun.id);
+  // TODO Paced train : Adapt to handle paced trains in issue https://github.com/OpenRailAssociation/osrd/issues/10612
+  const editoastTrainScheduleIdToUpdate = formatTrainScheduleIdToEditoastTrainId(
+    state.trainScheduleIdByNgeId.get(trainrun.id) as TrainScheduleId
+  );
   const trainSchedule = await dispatch(
     osrdEditoastApi.endpoints.getTrainScheduleById.initiate({
-      id: trainrunIdToUpdate,
+      id: editoastTrainScheduleIdToUpdate,
     })
   ).unwrap();
   const startDate = new Date(trainSchedule.start_time);
@@ -296,7 +307,7 @@ const handleUpdateTrainSchedule = async ({
   });
   const newTrainSchedule = await dispatch(
     osrdEditoastApi.endpoints.putTrainScheduleById.initiate({
-      id: trainrunIdToUpdate,
+      id: editoastTrainScheduleIdToUpdate,
       trainScheduleForm: {
         ...trainSchedule,
         ...trainSchedulePayload,
@@ -305,7 +316,11 @@ const handleUpdateTrainSchedule = async ({
       },
     })
   ).unwrap();
-  addUpsertedTrainSchedules([newTrainSchedule]);
+  const formattedNewTrainSchedule: TrainScheduleResultWithTrainId = {
+    ...newTrainSchedule,
+    id: formatEditoastTrainIdToTrainScheduleId(newTrainSchedule.id),
+  };
+  addUpsertedTrainSchedules([formattedNewTrainSchedule]);
 };
 
 const handleTrainrunOperation = async ({
@@ -325,8 +340,8 @@ const handleTrainrunOperation = async ({
   infraId: number;
   timeTableId: number;
   netzgrafikDto: NetzgrafikDto;
-  addUpsertedTrainSchedules: (trainSchedules: TrainScheduleResult[]) => void;
-  addDeletedTrainIds: (trainIds: number[]) => void;
+  addUpsertedTrainSchedules: (trainSchedules: TrainScheduleResultWithTrainId[]) => void;
+  addDeletedTrainIds: (trainIds: TimetableItemId[]) => void;
   state: MacroEditorState;
 }) => {
   const { nodes, labels, trainruns } = netzgrafikDto;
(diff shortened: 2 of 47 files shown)


front/src/utils/trainId.ts
```diff
@@ -0,0 +1,95 @@
+import type { OccurrenceId, PacedTrainId, TrainScheduleId } from 'reducers/osrdconf/types';
+
+export const isPacedTrain = (id: TrainScheduleId | PacedTrainId): id is PacedTrainId =>
```
clarani:
```suggestion
export const isPacedTrain = (id: TimetableItemId): id is PacedTrainId =>
```

front/src/utils/trainId.ts
```diff
@@ -0,0 +1,95 @@
...
+  id.startsWith('paced-');
+
+export const isOccurrence = (id: TrainScheduleId | OccurrenceId): id is OccurrenceId => {
```
clarani:
```suggestion
export const isOccurrence = (id: TrainId): id is OccurrenceId => {
```

front/src/utils/__tests__/trainId.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+describe('formatEditoastTrainIdToTrainScheduleId', () => {
+  it('should format to a TrainScheduleId', () => {
+    const trainId = 123;
+    const result = formatEditoastTrainIdToTrainScheduleId(trainId);
+    expect(result).toEqual(`train-${trainId}`);
+  });
+});
+
+describe('formatEditoastTrainIdToPacedTrainId', () => {
+  it('should format to a TrainScheduleId', () => {
```
clarani:
```suggestion
  it('should format to a PacedTrainId', () => {
```

front/src/utils/__tests__/trainId.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+    const pacedTrainId = 123;
+    const occurrenceIndex = 1;
+    const result = formatEditoastTrainIdToOccurrenceId({ pacedTrainId, occurrenceIndex });
+    expect(result).toBe(`paced-${pacedTrainId}-occurrence-${occurrenceIndex}`);
+  });
+});
+
+describe('formatTrainScheduleIdToEditoastTrainId', () => {
+  it('should return a valid editoast id', () => {
+    const trainScheduleId: TrainScheduleId = 'train-123' as TrainScheduleId;
```
clarani: You don't need the 2 casts here :)
```suggestion
    const trainScheduleId = 'train-123' as TrainScheduleId;
```

It's the same in the other functions

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
```diff
@@ -277,10 +282,13 @@ const handleUpdateTrainSchedule = async ({
 }) => {
   const { nodes, labels } = netzgrafikDto;
   const trainrunSectionsByTrainrunId = getTrainrunSectionsByTrainrunId(netzgrafikDto, trainrun.id);
-  const trainrunIdToUpdate = state.trainScheduleIdByNgeId.get(trainrun.id);
+  // TODO Paced train : Adapt to handle paced trains in issue https://github.com/OpenRailAssociation/osrd/issues/10612
+  const formattedTrainrunIdToUpdate = formatTrainScheduleIdToEditoastTrainId(
```
clarani: Hum I think you manipulate here a `trainScheduleId` rather than a `trainrunId` no ?

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
```diff
@@ -357,19 +365,24 @@ const handleTrainrunOperation = async ({
...
+      // TODO Paced train : Adapt to handle paced trains in issue https://github.com/OpenRailAssociation/osrd/issues/10612
+      const formattedTrainId = formatEditoastTrainIdToTrainScheduleId(newTrainSchedules[0].id);
+      state.trainScheduleIdByNgeId.set(trainrunId, formattedTrainId);
       addUpsertedTrainSchedules(newTrainSchedules);
       break;
     }
     case 'delete': {
-      const trainrunIdToDelete = state.trainScheduleIdByNgeId.get(trainrunId)!;
+      // TODO Paced train : Adapt to handle paced trains in issue https://github.com/OpenRailAssociation/osrd/issues/10612
+      const formattedTrainrunIdToDelete = formatTrainScheduleIdToEditoastTrainId(
```
clarani: Same here

front/src/applications/operationalStudies/hooks/useAutoUpdateProjection.ts
```diff
@@ -29,6 +30,8 @@ const useAutoUpdateProjection = (
   const currentTrainIdForProjection = useSelector(getTrainIdUsedForProjection);
   const selectedTrainId = useSelector(getSelectedTrainId);
 
+  const formattedTrainIds = trainIds.map((id) => formatEditoastTrainIdToTrainScheduleId(id));
```
clarani: Maybe it would be clearer if we keep this variable as `trainIds` and rename the argument `trainIds` -> `rawTrainIds` ?

front/src/applications/operationalStudies/hooks/useScenarioData.ts
```diff
@@ -210,14 +243,14 @@ const useScenarioData = (
...
+          ids: [editoastTrainId],
           electrical_profile_set_id: electricalProfileSetId,
         },
       }).unwrap();
       const summaries = formatTrainScheduleSummaries(
         [trainId],
         rawSummaries,
-        mapBy([trainScheduleResult], 'id'),
+        mapBy([updatedTrainScheduleResult], 'id'),
         rollingStocks!
```
clarani: I think rawSummaries here is a Map<number, SimulationSummaryResult> and not a Map<TrainScheduleId, SimulationSummaryResult>

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -205,12 +208,31 @@ const ManchetteWithSpaceTimeChartWrapper = ({
...
+      apiTmpSelectedTrain = formatTrainScheduleIdToEditoastTrainId(
+        tmpSelectedTrain as TrainScheduleId
+      );
+    }
+
+    return {
+      formattedCutProjectedTrains: apiCutProjectedTrains,
+      formattedTmpSelectedTrain: apiTmpSelectedTrain,
+    };
+  }, [cutProjectedTrains, tmpSelectedTrain]);
```
clarani: Why do you compute formattedCutProjectedTrains and formattedTmpSelectedTrain in the same useMemo ? I think they can have their own useMemo, which will avoid to recompute both of them when we need to re-compute only one

front/src/utils/trainId.ts
clarani, line 1: It's a bit confusing that some functions accept undefined values and others don't :thinking:

front/src/modules/simulationResult/components/SpaceTimeChart/useLazyProjectTrains.ts
```diff
@@ -65,24 +74,36 @@ const useLazyProjectTrains = ({
   useEffect(() => {
     const projectNextPackage = async (
       _path: PathfindingResultSuccess,
-      packageToProject: number[]
+      packageToProject: TrainId[]
     ) => {
       packageToProject.forEach((trainId) => requestedProjectedTrainIds.current.add(trainId));
 
+      // Format train ids back to editoast format
+      const formattedPackageToProject = packageToProject.map(
```
clarani:
```suggestion
      const editoastIds = packageToProject.map(
```

front/src/applications/operationalStudies/hooks/usePathProjection.ts
```diff
@@ -4,19 +4,26 @@ import { useSelector } from 'react-redux';
...
       {
-        id: trainIdUsedForProjection!,
+        id: editoastTrainIdUsedForProjection!,
         infraId: infra.id,
       },
       {
-        skip: !trainIdUsedForProjection,
+        skip: !editoastTrainIdUsedForProjection,
       }
     );
```
clarani: I suppose that you made formatTrainScheduleIdToEditoastTrainId accept undefined value for this use case. It might be better to make it do not accept undefined value and rewrite the code like this, what do you think ? (I'm not 100% sure it will work with typescript)
```suggestion
  const { data: projectionPathResult } =
    osrdEditoastApi.endpoints.getTrainScheduleByIdPath.useQuery(
      {
        id: formatTrainScheduleIdToEditoastTrainId(trainIdUsedForProjection as TrainScheduleId),
        infraId: infra.id,
      },
      {
        skip: !trainIdUsedForProjection,
      }
    );
```

front/src/utils/trainId.ts
SharglutDev: Done

front/src/utils/trainId.ts
SharglutDev: Done. By the way I think your branch was a bit outdated, because these functions can also accept `string` (for the `useScenarioQueryParams` file)

front/src/utils/__tests__/trainId.spec.ts
SharglutDev: Good catch ! Done

front/src/utils/__tests__/trainId.spec.ts
SharglutDev: Oups, done

front/src/utils/trainId.ts
```diff
@@ -0,0 +1,107 @@
+import type { OccurrenceId, PacedTrainId, TrainScheduleId } from 'reducers/osrdconf/types';
+
+export const isPacedTrain = (id: string | TrainScheduleId | PacedTrainId): id is PacedTrainId =>
+  id.startsWith('paced-');
```
emersion: This is not enough: this will return true for `paced-42-occurence-0` even though it's not a paced train.

Maybe it would be best to use a difference format for occurences? Maybe something like `occurence-42-0`? That way the prefix alone is enough to tell the type.

front/src/utils/trainId.ts
```diff
@@ -0,0 +1,107 @@
...
+export const isTrainSchedule = (
+  id: string | TrainScheduleId | OccurrenceId | PacedTrainId
+): id is TrainScheduleId => id.startsWith('train-');
+
+/**
+ * Given a train id in the Editoast format (used for api),
+ * returns the train id with a TrainScheduleId format (used across the front).
+ */
+export const formatEditoastTrainIdToTrainScheduleId = (trainId: number): TrainScheduleId =>
+  `train-${trainId}` as TrainScheduleId;
```
emersion: Maybe we should use a different prefix for train schedule? For instance `trainschedule-`? `train-` is a bit confusing/unclear.

front/src/utils/trainId.ts
```diff
@@ -0,0 +1,107 @@
+import type { OccurrenceId, PacedTrainId, TrainScheduleId } from 'reducers/osrdconf/types';
+
+export const isPacedTrain = (id: string | TrainScheduleId | PacedTrainId): id is PacedTrainId =>
```
emersion: I think `string | TrainScheduleId | PacedTrainId` is the same as `string`.

Why do we need `string` in the list? Because it might come from the URL?

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
SharglutDev: Yes you're right, wasn't sure which term to keep because before we used `trainrunIdToUpdate` but it's less confusing like that you're right. Done.

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
SharglutDev: Done

front/src/applications/operationalStudies/hooks/useAutoUpdateProjection.ts
SharglutDev: Done. To be consistent with the rest of the PR, I used `editoastTrainIds` instead of `rawTrainIds`, is that ok for you ?

front/src/applications/operationalStudies/hooks/useScenarioData.ts
SharglutDev: Oh very nice catch, it's bad that typescript doesn't catch that :/ Done

emersion: Any idea why TypeScript didn't catch this?

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx

front/src/utils/trainId.ts
SharglutDev: Yeah that was a recent change. All formatting functions from new format to editoast can accept undefined because in many places we manipulate `selectors` that can be `undefined`. 
I try to change the current behavior as less as possible and thought it was better to handle that here in the function and allow to accept/return undefined instead of checking everywhere it's called if the `selector` is defined.

front/src/modules/simulationResult/components/SpaceTimeChart/useLazyProjectTrains.ts
SharglutDev: Done

front/src/utils/trainId.ts
SharglutDev: Since we can pass a `PacedTrainId` and not a `OccurenceId`, I thought it was enough at first.
But now that I added `string`, you're right it becomes flaky.

front/src/utils/trainId.ts
SharglutDev: Fine by me :)

front/src/utils/trainId.ts
SharglutDev: Yes exactly. Do you have a suggestion for that ?

front/src/applications/operationalStudies/hooks/usePathProjection.ts
SharglutDev: Seems to be working, I will give it a try :)

emersion: Can we avoid `as`, and use `!` instead?

front/src/utils/trainId.ts
SharglutDev: Edit : fixed related to your last comment

front/src/applications/operationalStudies/hooks/usePathProjection.ts
SharglutDev: Done (with some adaptation to your suggestion)

front/src/utils/trainId.ts
SharglutDev: Changed for format `occurence-23-paced-0`

emersion: It sounds a bit weird to put the word "paced" in the middle IMHO.

theocrsb (approved):
lgtm, I didn't see any problem after testing

front/src/utils/trainId.ts
SharglutDev: example was bad but the format is `occurrence-{occurenceIndex}-paced-{pacedId]`

front/src/utils/trainId.ts
SharglutDev: Kept only `string` in params

front/src/utils/trainId.ts
emersion: I see! This can later be easy adjusted to something else anyways.

front/src/applications/operationalStudies/hooks/useScenarioData.ts
emersion: Aha, it seems like `Record` type checking is pretty loose! See:

- https://www.typescriptlang.org/play/?#code/C4TwDgpgBAKgTgQwJYDsDKBjAFhAJgVwBsIBJXKAXigGdg5UBzKAMigG8o4IFcB7FQiCgBbBHADWEOAC4o+FEgCO+aNRDCARr0JQAvgG4AUIYBm8jMCT8oJ3rwAUw2QCUIGXnFwAeeMnTY8IlJcABooFHxNKQA+AEp2XVNzS2sNMUcXNw9vWnoUBjCIqLg4hON3FFoRTPdPLyKNKULIxpLKdigAbQAWACYAXVk+vSNbB2FYozS4R0njebMUCysUGztejKgAWQQwH0RUTBwCYjJm4tK2RIqq4V7ZHb2GpvCWmPaUCAB3bd37ObGGzukyAA
- https://github.com/microsoft/TypeScript/issues/51680

We should prefer `Map` over `Record`.

front/src/applications/operationalStudies/hooks/useAutoUpdateProjection.ts
clarani: Yes it's perfect for me :+1:

front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
```diff
@@ -68,8 +68,8 @@ const useLazyLoadTrains = ({
         const packageToFetch = getBatchPackage(i, _trainToFetchIds, BATCH_SIZE);
 
         // Format train ids back to editoast format
-        const formattedPackageToFetch = packageToFetch.map(
-          (trainId) => formatTrainScheduleIdToEditoastTrainId(trainId as TrainScheduleId)!
+        const formattedPackageToFetch = packageToFetch.map((trainId) =>
```
clarani:
```suggestion
        const editoastIds = packageToFetch.map((trainId) =>
```

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
```diff
@@ -286,12 +286,12 @@ const handleUpdateTrainSchedule = async ({
   const { nodes, labels } = netzgrafikDto;
   const trainrunSectionsByTrainrunId = getTrainrunSectionsByTrainrunId(netzgrafikDto, trainrun.id);
   // TODO Paced train : Adapt to handle paced trains in issue https://github.com/OpenRailAssociation/osrd/issues/10612
-  const formattedTrainrunIdToUpdate = formatTrainScheduleIdToEditoastTrainId(
+  const formattedTrainScheduleIdToUpdate = formatTrainScheduleIdToEditoastTrainId(
```
clarani: nit (editoastIdToUpdate may be enough)
```suggestion
  const editoastTrainScheduleIdToUpdate = formatTrainScheduleIdToEditoastTrainId(
```

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
```diff
@@ -385,12 +385,12 @@ const handleTrainrunOperation = async ({
     }
     case 'delete': {
       // TODO Paced train : Adapt to handle paced trains in issue https://github.com/OpenRailAssociation/osrd/issues/10612
-      const formattedTrainrunIdToDelete = formatTrainScheduleIdToEditoastTrainId(
+      const formattedTrainScheduleIdToDelete = formatTrainScheduleIdToEditoastTrainId(
```
clarani:
```suggestion
      const editoastTrainScheduleIdToDelete = formatTrainScheduleIdToEditoastTrainId(
```

front/src/applications/operationalStudies/hooks/useAutoUpdateProjection.ts
```diff
@@ -23,17 +23,19 @@ import { formatEditoastTrainIdToTrainScheduleId, isTrainSchedule } from 'utils/t
...
-  trainIds: number[],
+  editoastTrainIds: number[],
   trainScheduleSummaries: TrainScheduleWithDetails[]
 ) => {
   const dispatch = useAppDispatch();
   const currentTrainIdForProjection = useSelector(getTrainIdUsedForProjection);
   const selectedTrainId = useSelector(getSelectedTrainId);
 
-  const formattedTrainIds = trainIds.map((id) => formatEditoastTrainIdToTrainScheduleId(id));
+  const formattedTrainIds = editoastTrainIds.map((editoastTrainId) =>
```
clarani, line 33: nit (back to the original name of the const)
```suggestion
  const trainIds = editoastTrainIds.map((editoastTrainId) =>
```

front/src/applications/operationalStudies/hooks/useScenarioData.ts
```diff
@@ -247,9 +248,17 @@ const useScenarioData = (
           electrical_profile_set_id: electricalProfileSetId,
         },
       }).unwrap();
+
+      // TODO Paced train : Adapt this for the add paced train issue : https://github.com/OpenRailAssociation/osrd/issues/10615
+      const formattedRawSummaries: { [key: TrainScheduleId]: SimulationSummaryResult } = {};
```
clarani, line 253: Hum do you want to use a Map instead of a dict like this ?

front/src/applications/operationalStudies/__tests__/sampleData.ts
```diff
@@ -351,8 +352,8 @@ export const electrificationRangesLarge: ElectrificationRange[] = [
   },
 ];
 
-export const trainScheduleTooFast: TrainScheduleResult = {
-  id: 98,
+export const trainScheduleTooFast: TrainScheduleResultWithTrainId = {
+  id: 'train-98' as TrainScheduleId,
```
clarani: It should be trainschedule now I think (you can grep `' as TrainScheduleId` to find some ids you forget to update :wink:)
```suggestion
  id: 'trainschedule-98' as TrainScheduleId,
```

front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
SharglutDev: Done

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
SharglutDev: Done

front/src/applications/operationalStudies/components/MacroEditor/ngeToOsrd.ts
SharglutDev: Done

front/src/applications/operationalStudies/hooks/useScenarioData.ts
SharglutDev: Was planning to change the accepted type in `formatTrainScheduleSummaries` in a different PR. What do you think ?

front/src/applications/operationalStudies/__tests__/sampleData.ts
SharglutDev: Good catch !

front/src/applications/operationalStudies/hooks/useAutoUpdateProjection.ts
SharglutDev: Hmm I think keeping the term `formatted` would easier alert someone, who isn't familiar with the paced train issues, that we are manipulating a different kind of id.

front/src/applications/operationalStudies/hooks/useScenarioData.ts
clarani: Fine ! :)

front/src/applications/operationalStudies/__tests__/sampleData.ts
clarani: I can still find some `'train-98' as TrainScheduleId`

SharglutDev: Oups. Should be ok now.

clarani (approved):

PR #10680, opened 2025-02-04
Import all sprites in the project

Tristramg:
Ref #10627

## What does this pull request do

This pull requests imports the sprites to have them locally instead of a remote server (4f6659c5530beadbcf84433c0ad0c58d0501903a)

We also change the tool to generate. This reduces a bit the size of the generated atlas and is — arguably — a little bit easier to use 92a75ff00296c154c8ab8fc006516d1bc8ae664e

The front now uses those new sprites f60ec385df90aa988b46f1dc9aa5c1f2df13411e

## Benefits

* Less dependency to an external server
* Smaller files (~1Mb saved per map view)

## How to test

There should be no behavior change.

Navigate through the map with the signal layer on. See in the console if there are no missing sprites and in the network tools that all `sprites` files are correctly loaded from the osrd server and no longer https://static.osm.osrd.fr/sprites/sprites

In editoast, run generate-atlas.sh (it shouldn’t change anything, but output what atlases have been generated). This requires spreet (`cargo install spreet`)


emersion:
How large are these new assets?

Tristramg:
When taking the largest on my computer (@2x, I suppose this would be the same for most users), that used to be hosted separately
- Before : 1.28 Mb
- After : 409 kb

You can see in the diff on github the size reduction for the existing sprites (~40%)

flomonster (changes requested):
You are using way too much sprites.
For example all "Arret *" are not used at all (referenced only in `SignalsName.ts`).

It's annoying to put all sprites under `signal_sprites` when some are not signals (e.g. “HEURTOIR”). 
- We could rename this folder to `sprites.
- We could move the default sprites in another folder.

Tristramg:
Thank you for the feedback

1. I move `signal_sprites` to `sprites` (folder with the assets and its path in `sprites.rs`) : 671e696e
2. Instead of just dumping all the icons of the previous sprite, I looked in the code : 49bb8e29
  *  grep `icon-image`
    * MATD
    * MATG
    * HEURTOIR
    * UNKNOWN2 (for holyland)
    * UNKNOWN
    * TIV D FIXE *
    * TIVD B FIX *
    * R
    * Z
    * RIGHT
  * select distinct(data->'type') from infra_layer_neutral_sign;
    * CC FIN 
    * REV 
    * BP DIS
    * CC EXE
    * BP EXE
    * SECT (this one is missing in the blueprint)
    * BP FIN
  * Actually the blue icons are used in the blueprint style for neutral_sign. Sorry for that, I added them
  * I just kept used images and their blue counterpart
 
As a result, the served PNG is now less than 300kb big, and of course we need less space in the repo

editoast/assets/sprites/default/SECT.svg
flomonster, line 1: Where this is used?

flomonster (approved):
LGTM, I wonder if we want to keep the blueprint style. It was there to visualize the infra schematic (which no longer exists).

We should ask @thibautsailly. If it's not essential to keep this style, we could drop all the blue sprites.

editoast/assets/sprites/default/SECT.svg
thibautsailly: I'm not aware of a real usage today for this display.

editoast/assets/sprites/generate-atlas.sh
```diff
@@ -1,4 +1,8 @@
-#!/bin/sh
+#!/bin/bash
+
+# Allows to use **/*.svg to find files in the directory and subdirectory
+# Requires at leas bash4, release in 2009
```
woshilapin:
```suggestion
# Requires at least bash4, release in 2009
```

editoast/assets/sprites/generate-atlas.sh
```diff
@@ -1,4 +1,8 @@
-#!/bin/sh
```
woshilapin: Maybe we want to stay POSIX compatible? If I understand correctly, the move to `bash` is about the `globstar` option? You might be able to change the `cp` command with something along these lines:

```sh
find "${signaling_system}" -type f -name "*.svg" -exec cp "{}" "${svg_dir}" ";"
```

editoast/assets/sprites/default/sprites@3x.png
woshilapin, line 1: Since this file is generated from the other files (if I understand correctly), should it be commited? Maybe it can be generated in the `Dockerfile` instead?

woshilapin (approved):

editoast/assets/sprites/default/SECT.svg
Tristramg: If I understand the code correctly, it might be used in https://github.com/OpenRailAssociation/osrd/blob/dev/front/src/common/Map/Layers/InfraObjectLayers/extensions/SNCF/NeutralSectionSigns.tsx#L23
Where type comes from the table `infra_layer_neutral_sign` where SECT exists

editoast/assets/sprites/default/sprites@3x.png
Tristramg: Until now, the sprites where committed. We can rethink the existing way to do it.
What about people not using Docker when running editoast? How can we make sure that they won’t blame the front when some signals are missing on the front?

editoast/assets/sprites/generate-atlas.sh
Tristramg: Sorry, I got mixed up in my versions.

By switching to spreet, we now have smaller png, and it can crawl subdirectories, so no need to switch to 15 year young stuff, and no need for a syntax that always needs to be googled ;)

SharglutDev (approved):
Lgtm and tested, sprites are still there :) (except the remaining error for etcs)

editoast/assets/sprites/default/sprites@3x.png
woshilapin: If I understand correctly, it's a deployment problem. And the only way we document how to deploy `editoast` for now is through `docker` (and when I mean documentation, I mostly mean a `Dockerfile` and `osrd-chart`... so mostly code). So I would say that this is ok to make it part of the `Dockerfile`. We could also add a section in some `README`.

Tristramg: All right, so I added this commit: b96f98b333d06d16d101f3c97f683dd3c67d8694 to generate the sprites instead of comiting them.
This should make @emersion  happy to have less data in the repository

@flomonster do you have an opinion?

If every body agrees, I will squash the commit to not have the generated sprite in the history and I think we would be good to go

flomonster: Generating sprites during the docker build is a good idea. The drawback is for editoast developers. When we run it without docker, it doesn't generate the static files.

Tristramg: Yes, I added the information to the readme, but some might still be surprised
I don’t think there is a silverbullet here. Maybe we could add a justfile at some point that runs everything, but that’s clearly far from the scope of this pull request

PR #10681, opened 2025-02-05
core: editoast: add flag to ignore infra speed limits

### LINKED ISSUE #10587: Add a flag to enable ignoring speed limits
# Description

Add a flag in simulations, allowing us to ignore speed limits if we want to.

# Goal

Allow fixing BASIC trains that are late because of broken speed limits.

# AC

- [ ] In the simulation part, there's now a flag to take into account speed limits or not
- [ ] All the calls to simulation are using the speed limits

axrolld:
Add a flag to ignore infra speed limits in a standalone simulation.

This will be useful in BASIC import to fix late trains for the moment.

Closes #10587


core/src/main/kotlin/fr/sncf/osrd/api/api_v2/standalone_sim/SimulationRequest.kt
```diff
@@ -143,5 +143,6 @@ class SimulationPowerRestrictionItem(
 )
 
 class TrainScheduleOptions(
-    @Json(name = "use_electrical_profiles") val useElectricalProfiles: Boolean
+    @Json(name = "use_electrical_profiles") val useElectricalProfiles: Boolean,
+    @Json(name = "ignore_infra_speed_limits") val ingoreInfraSpeedLimits: Boolean
```
eckter: typo

```suggestion
    @Json(name = "ignore_infra_speed_limits") val ignoreInfraSpeedLimits: Boolean
```

eckter: We may want to invert the logic to be consistent with `use_electrical_profiles`. We could make it optional as well, and use them if unspecified.

core/src/main/kotlin/fr/sncf/osrd/envelope_sim_infra/MRSP.kt
```diff
@@ -59,6 +59,8 @@ fun computeMRSP(
  * @param rsLength length of the rolling stock (m)
  * @param addRollingStockLength whether the rolling stock length should be taken into account in the
  *   computation.
+ * @param ignoreInfraSpeedLimits whether the speed limits coming from the infrastructure should be ignored in the
+ *   computaion
```
eckter:
```suggestion
 *   computation
```

core/src/main/kotlin/fr/sncf/osrd/envelope_sim_infra/MRSP.kt
```diff
@@ -72,36 +74,39 @@ fun computeMRSP(
...
-        }
-        if (attrs.any { it is UnknownTag }) attrs.add(HasMissingSpeedTag)
-        if (speed != 0.0) {
-            // Add the envelope part corresponding to the restricted speed section
-            builder.addPart(
-                EnvelopePart.generateTimes(
-                    attrs,
-                    doubleArrayOf(start, end),
-                    doubleArrayOf(speed, speed)
+    if (!ignoreInfraSpeedLimits) { //if we take into account speedlimits coming from the infrastructure
```
eckter: I'd have preferred a way to avoid this large `if` block.

Maybe something like this, what do you think?

```kotlin
    val speedLimitProperties =
        if (ignoreInfraSpeedLimits) distanceRangeMapOf()
        else path.getSpeedLimitProperties(trainTag, temporarySpeedLimitManager)
    for (speedLimitPropertyRange in speedLimitProperties) {
        // ...
    }
```

core/src/main/kotlin/fr/sncf/osrd/envelope_sim_infra/MRSP.kt
```diff
@@ -72,36 +74,39 @@ fun computeMRSP(
     trainTag: String?,
     temporarySpeedLimitManager: TemporarySpeedLimitManager?,
     safetySpeedRanges: DistanceRangeMap<Speed>? = null,
+    ignoreInfraSpeedLimits: Boolean = false,
```
eckter: The parameter isn't set.

It should be done in `StandaloneSimulation.kt:71`. And it should also be added to the other version of the method which takes a rolling stock as parameter (this file, line 35).

eckter (commented):
Nice job!

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/standalone_sim/SimulationRequest.kt
axrolld: Good catch thanks !

python/osrd_schemas/osrd_schemas/train_schedule.py
```diff
@@ -170,12 +170,19 @@ class PowerRestrictionRanges(RootModel):
 
 
 class TrainScheduleOptions(BaseModel):
-    """Optional arguments for the standalone simulation."""
+    """Optional arguments :
+    - `ignore_electrical_profiles` : ignore the electrical profiles for the standalone simulation
+    - `ignore_infra_speed_limits` : ignore the speed limits coming from the infrastructure for the standalone simulation
+    """
 
     ignore_electrical_profiles: bool = Field(
```
woshilapin: Hmm, something is fishy. In `editoast`, this schema is defined as [`use_electrical_profiles`](https://github.com/OpenRailAssociation/osrd/blob/66a9cade203fbb3ecccf5182bc9577dd2eb43b0c/editoast/editoast_schemas/src/train_schedule/train_schedule_options.rs#L16)... 

It was added in `osrd_schemas` in https://github.com/OpenRailAssociation/osrd/commit/65936d6aca05d1c7bd2025fbb4094afddf5166bc#diff-d5864b67bb112b9568c11e82046affbf1336af068b01ea186f7f7e39bf39b20eR139-R142 as `ignore_electrical_profiles` and was added in `editoast` in https://github.com/OpenRailAssociation/osrd/commit/79c5caa5f5e6595877c05ea9f55c1ff4f7d9264d#diff-64eae74a5b61bec4f050765098f382e536496ec9b5917caa1385722f99ffdf9fR157 as `use_electrical_profiles`.

cc @flomonster @Castavo any idea what might have happened and what is the best way to fix that (sorry, both of your names showed up in `git blame`)?

woshilapin (approved):
Looks ok for `editoast` but something weird showed up. Since it's not part of your PR, feel free to discard it, but if it could be fixed with it, that'd be great.

editoast/editoast_schemas/src/train_schedule/train_schedule_options.rs
```diff
@@ -14,6 +14,8 @@ pub struct TrainScheduleOptions {
     #[derivative(Default(value = "true"))]
     #[serde(default = "default_use_electrical_profiles")]
     use_electrical_profiles: bool,
+    #[serde(default)]
+    ignore_infra_speed_limits: bool,
```
flomonster: Does this flag also ignore speed limits for the pathfinding (or just for simulation)?
If so we should rename it to something like: `ignore_speed_limits_during_simulation`.

python/osrd_schemas/osrd_schemas/train_schedule.py
flomonster: There are no sync checks between editoast and the osrd_schema repo. @Castavo suggested to generate osrd_schema from openapi.
My opinion is that osrd_schema should be removed and replaced by a rust library (with python bindings)

editoast/editoast_schemas/src/train_schedule/train_schedule_options.rs
axrolld: It's only for the simulation, i'll rename the editoast side, core side isn't being used in pathfinding

python/osrd_schemas/osrd_schemas/train_schedule.py
woshilapin: OK, let's not address this question now. But at least, we might be able to fix this error for the moment. Do you have any recall on which of `use_electrical_profiles` and `ignore_electrical_profiles` should be used? From [@eckter's comment](https://github.com/OpenRailAssociation/osrd/pull/10681#discussion_r1944990676), we might prefer `use_electrical_profiles` (also, I tend to prefer positive forms instead of negative forms which creates double negation when the value is `false`). Because, if I understand correctly, whatever is given to import through Python would be ignored by `editoast` for the moment.

core/src/main/kotlin/fr/sncf/osrd/envelope_sim_infra/MRSP.kt
axrolld: Yep, i went for this:
```kotlin
val speedLimitProperties =
        if (useInfraSpeedLimits) path.getSpeedLimitProperties(trainTag, temporarySpeedLimitManager)
        else distanceRangeMapOf<SpeedLimitProperty>()
```

python/osrd_schemas/osrd_schemas/train_schedule.py
axrolld: Not quite sure what I was supposed to do here but I renamed it as `use_electrical_profiles` for more consistency

flomonster: Like @woshilapin said we can address this in another PR.

editoast/editoast_schemas/src/train_schedule/train_schedule_options.rs
```diff
@@ -14,8 +14,15 @@ pub struct TrainScheduleOptions {
     #[derivative(Default(value = "true"))]
     #[serde(default = "default_use_electrical_profiles")]
     use_electrical_profiles: bool,
+    #[derivative(Default(value = "true"))]
+    #[serde(default = "default_use_infra_speed_limits_for_simulation")]
+    use_electrical_profiles_for_simulation: bool,
```
flomonster:
```suggestion
    #[serde(default = "default_use_speed_limits_for_simulation")]
    use_speed_limits_for_simulation: bool,
```

editoast/editoast_schemas/src/train_schedule/train_schedule_options.rs
```diff
@@ -14,8 +14,15 @@ pub struct TrainScheduleOptions {
...
 }
 
 fn default_use_electrical_profiles() -> bool {
     true
 }
+
+fn default_use_infra_speed_limits_for_simulation() -> bool {
```
flomonster:
```suggestion
fn default_use_speed_limits_for_simulation() -> bool {
```

editoast/openapi.yaml
flomonster, line 1: It seems like you edited this file manually. This file is autogenerated using the following command: `editoast openapi > openapi.yaml`

> [!NOTE]
> The CI won't pass if it does not match the generated file.

flomonster (changes requested):

python/osrd_schemas/osrd_schemas/train_schedule.py
woshilapin: Oh, thank you @axrolld for making the change 🙏

woshilapin (approved):

flomonster (approved):
LGTM

core/src/main/kotlin/fr/sncf/osrd/envelope_sim_infra/MRSP.kt
```diff
@@ -72,12 +78,15 @@ fun computeMRSP(
...
+    useInfraSpeedLimits: Boolean = true,
 ): Envelope {
     val builder = MRSPEnvelopeBuilder()
     val pathLength = toMeters(path.getLength())
 
     val offset = if (addRollingStockLength) rsLength else 0.0
-    val speedLimitProperties = path.getSpeedLimitProperties(trainTag, temporarySpeedLimitManager)
+    val speedLimitProperties =
+        if (useInfraSpeedLimits) path.getSpeedLimitProperties(trainTag, temporarySpeedLimitManager)
+        else distanceRangeMapOf<SpeedLimitProperty>()
```
eckter, line 89: nit

```suggestion
        else distanceRangeMapOf()
```

eckter (approved):
LGTM

core/src/main/kotlin/fr/sncf/osrd/envelope_sim_infra/MRSP.kt
axrolld: This was making some tests fail on my side (idk why) so i set <SpeedLimitProperty>

PR #10682, opened 2025-02-05
front: remove mlg data

Tristramg:
We use data from http://www.mlgtraffic.net/ to display rolling stock

The rolling stock images are now imported through the API and are we no longer need hard coded things. This pull request removes all unnecessary data to manipulate mlg assets.

I left the credits in https://github.com/OpenRailAssociation/osrd/blob/dev/front/src/common/ReleaseInformations/LicenseAttributions.tsx#L43-L50 as we still use those images


emersion (approved):
Nice, LGTM!

clarani (approved):
LGTM :broom:

Issue #10684, opened 2025-02-05
Editoast: adapt conflict endpoint to paced trains


younesschrifi:
the aim of this issue is to adapt the conflict endpoint to paced trains. The paced train field is still not used, as the paced train model does not yet exist and is currently an empty vector. 
The aim is to enable the front end to make the necessary adaptations.

CF : https://github.com/osrd-project/osrd-confidential/issues/779

PR #10685, opened 2025-02-05
editoast: adapt conflicts endpoint to paced trains

younesschrifi:
Close #[10684](https://github.com/OpenRailAssociation/osrd/issues/10684)

### DIFF
--- a/editoast/src/core/conflict_detection.rs
+++ b/editoast/src/core/conflict_detection.rs
@@ -76,9 +76,20 @@ pub struct ConflictDetectionResponse {
 }
 
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize, ToSchema)]
+pub struct PacedTrainOccurrenceId {
+    paced_train_id: i64,
+    index: i64,
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, ToSchema)]
 pub struct Conflict {
-    /// List of train ids involved in the conflict
-    pub train_ids: Vec<i64>,
+    /// List of train schedule ids involved in the conflict
+    pub train_schedule_ids: Vec<i64>,
+    /// List of paced train occurrences involved in the conflict.
+    /// Each occurrence is identified by a `paced_train_id` and its `index`
+    #[serde(default)]
+    #[schema(inline)]
+    pub paced_train_occurrence_ids: Vec<PacedTrainOccurrenceId>,
     /// List of work schedule ids involved in the conflict
     pub work_schedule_ids: Vec<i64>,
     /// Datetime of the start of the conflict
--- a/front/src/common/api/generatedEditoastApi.ts
+++ b/front/src/common/api/generatedEditoastApi.ts
@@ -3625,12 +3625,18 @@ export type Conflict = {
   conflict_type: 'Spacing' | 'Routing';
   /** Datetime of the end of the conflict */
   end_time: string;
+  /** List of paced train occurrences involved in the conflict.
+    Each occurrence is identified by a `paced_train_id` and its `index` */
+  paced_train_occurrence_ids?: {
+    index: number;
+    paced_train_id: number;
+  }[];
   /** List of requirements causing the conflict */
   requirements: ConflictRequirement[];
   /** Datetime of the start of the conflict */
   start_time: string;
-  /** List of train ids involved in the conflict */
-  train_ids: number[];
+  /** List of train schedule ids involved in the conflict */
+  train_schedule_ids: number[];
   /** List of work schedule ids involved in the conflict */
   work_schedule_ids: number[];
 };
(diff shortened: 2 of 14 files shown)


editoast/src/core/conflict_detection.rs
```diff
@@ -75,10 +75,20 @@ pub struct ConflictDetectionResponse {
...
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, ToSchema)]
+pub struct PacedTrainsConflicts {
+    paced_train_id: i64,
+    index: i64,
+}
+
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize, ToSchema)]
 pub struct Conflict {
     /// List of train ids involved in the conflict
     pub train_ids: Vec<i64>,
```
flomonster: We could rename the other field to be more consistent.
We can also adapt the doc.

```suggestion
    pub train_schedule_ids: Vec<i64>,
```

flomonster (approved):
LGTM

editoast/src/core/conflict_detection.rs
```diff
@@ -75,10 +75,20 @@ pub struct ConflictDetectionResponse {
     pub conflicts: Vec<Conflict>,
 }
 
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, ToSchema)]
+pub struct PacedTrainsConflicts {
```
woshilapin: This designates a paced train occurence, not a conflict (it's only used in `Conflict` to designate an occurence). What do you think of renaming this `PacedTrainOccurrenceId` (it's basically the way to identify an occurrence of paced train)?

editoast/src/core/conflict_detection.rs
```diff
@@ -75,10 +75,20 @@ pub struct ConflictDetectionResponse {
...
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize, ToSchema)]
 pub struct Conflict {
-    /// List of train ids involved in the conflict
-    pub train_ids: Vec<i64>,
+    /// List of train schedule ids involved in the conflict
+    pub train_schedule_ids: Vec<i64>,
+    /// List of paced train ids involved in the conflict
+    #[serde(default)]
+    #[schema(inline)]
+    pub paced_trains: Vec<PacedTrainsConflicts>,
```
woshilapin: We have `train_schedule_ids` (above), that reference a train schedule (it is not the train schedule data, only a reference to one). `paced_trains` is also only a reference to an occurrence of a paced train (it's not the data of this paced train). So I would keep the naming similar. What about `paced_train_occurrence_ids`?

leovalais (approved):

shenriotpro (approved):
LGTM for tests/

eckter (approved):
For core: seems like a simple renaming, I'm not checking further (I trust that this matches the new names used in the payload)

Erashin (commented):
Not that this shouldn't be merged, I trust everyone else's opinion on the matter (especially if you know the subject). But from the pov of someone who has no idea what's going on, I clearly don't have enough info on this pr: i don't know what a paced train is, the commit has no additional info, description same, linked issue same, and no additional linked issue for additional info.
I'm guessing this is a first PR to pave the way for the following paced train conflict logic that will come in another PR. But yeah, I wouldn't be against having a bit more info somewhere for this sort of PR :)
As for the core part, lgtm, the rename works, and the added field isn't used yet anyhow.

emersion:
Yeah, agreed - I kind of realize we should've written down a design doc on the website for paced trains…

Khoyo:
There's a confidential issue (https://github.com/osrd-project/osrd-confidential/issues/779). It should be cleaned up and put as a real osrd issue, and ideally we should have a design document on the website indeed.

younesschrifi:
I agree with everyone and I apologize for the lack of information. 
I added a description in the issue.
I already made a PR in osrd_website (CF Timetable V2) to describe the paced train model and I'm going to add more information on it. 
Thanks for the reviews. 

@eckter @emersion @Erashin @Khoyo

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -3577,12 +3577,17 @@ export type Conflict = {
   conflict_type: 'Spacing' | 'Routing';
   /** Datetime of the end of the conflict */
   end_time: string;
+  /** List of paced train occurence ids involved in the conflict */
+  paced_train_occurrence_ids?: {
+    index: number;
+    paced_train_id: number;
+  }[];
```
SharglutDev: If I understand correctly, this gives, for each paced train involved in a conflict, its id and the index of the occurrence responsible of the conflict ?
Maybe the doc above should reflect that instead of only mentioning "train occurence ids" (by the way, occurrence comes with 2 `r` in english 🙂).
I also think that the property name isn't accurate enough. Occurrences don't have ids but indexes.
```suggestion
  /** List of paced train occurrence ids involved in the conflict */
  paced_train_occurrence_ids?: {
    index: number;
    paced_train_id: number;
  }[];
```

Also what if multiple occurrences are responsible for the conflict ? (can this happen ?) Do we want to provide the first one with conflict or all of them ?

younesschrifi: I'm adding the info. Thanks :)
If multiple occurrences are responsible for the conflict, then all of them will be provided 
for exemple 

[{index: 1, paced_train_id: 4}, {index:4, paced_train_id:4}]

Erashin (approved):

SharglutDev (approved):
Lgtm (for front part) and tested :)

PR #10692, opened 2025-02-05
Speedup stdcm editoast endpoint :rocket: 

### LINKED ISSUE #10452: STDCM queries take too much time

flomonster:
close #10452

What si done (review commits individual):

1. Spawn tasks to dispatch cpu charge
2. Reduce train batch size
3. Avoid useless psql queries
4. Optimize a postgresql queries. Limit the number of elements in the response. Moving a filter from rust to postgres. (check perf here: https://explain.dalibo.com/plan/e7d4427b7a8ea7d5)

### Local benchmark

|               | Before | After |
|---------------|--------|-------|
| Without cache |  106s  |  108s |
| With cache    |  9.64s | 3.15s |

### Dev benchmark

|               | Before (Compression) | After (No-Compression) | After (Compression) |
|---------------|----------------------|------------------------|---------------------|
| Without cache | 90s                  |          111s          |         109s        |
| With cache    | 17s                  |          12.7s         |         8.3s        |

### DIFF
--- a/editoast/src/models/timetable.rs
+++ b/editoast/src/models/timetable.rs
@@ -4,6 +4,7 @@ use diesel::prelude::*;
 use diesel::sql_query;
 use diesel::sql_types::Array;
 use diesel::sql_types::BigInt;
+use diesel::sql_types::Timestamptz;
 use diesel_async::RunQueryDsl;
 use editoast_derive::Model;
 use futures_util::stream::TryStreamExt;
@@ -55,27 +56,42 @@ impl Timetable {
             .map_err(Into::into)
     }
 
-    pub async fn schedules_before_date(
+    /// This function will return all train schedules in a timetable that runs within the time window.
+    ///
+    /// **IMPORTANT**: The filter is based on the scheduled arrival time and not the actual simulated arrival time.
+    ///
+    /// The diagram below shows a list of trains in a timetable:
+    /// `?`: unscheduled arrival times.
+    /// `|`: scheduled arrival times.
+    ///
+    /// ```
+    ///                       min_time               max_time
+    ///     Time Window           |----------------------|
+    /// Train 1 ✅         |-------------|
+    /// Train 2 ❌    |-------|
+    /// Train 3 ✅                              |----------|
+    /// Train 4 ✅   |-------?
+    /// Train 5 ✅                        |---------?
+    /// Train 6 ❌                                           |------?
+    /// ```
+    pub async fn schedules_in_time_window(
         self,
         conn: &mut DbConnection,
-        time: DateTime<Utc>,
+        min_time: DateTime<Utc>,
+        max_time: DateTime<Utc>,
     ) -> Result<Vec<TrainSchedule>> {
-        use diesel::prelude::*;
-        use diesel_async::RunQueryDsl;
-        use editoast_models::tables::train_schedule::dsl;
-
-        let train_schedules = dsl::train_schedule
-            .filter(dsl::start_time.le(time))
-            .filter(dsl::timetable_id.eq(self.id))
-            .load_stream::<Row<TrainSchedule>>(conn.write().await.deref_mut())
-            .await?
-            .map_ok(|ts| ts.into())
-            .try_collect::<Vec<TrainSchedule>>()
-            .await;
-        match train_schedules {
-            Ok(train_schedules) => Ok(train_schedules),
-            Err(err) => Err(err.into()),
-        }
+        let train_schedules = sql_query(include_str!(
+            "timetable/sql/get_train_schedules_in_time_window.sql"
+        ))
+        .bind::<BigInt, _>(self.id)
+        .bind::<Timestamptz, _>(max_time)
+        .bind::<Timestamptz, _>(min_time)
+        .load_stream::<Row<TrainSchedule>>(conn.write().await.deref_mut())
+        .await?
+        .map_ok(|ts| ts.into())
+        .try_collect::<Vec<TrainSchedule>>()
+        .await;
+        train_schedules.map_err(|e| e.into())
     }
 }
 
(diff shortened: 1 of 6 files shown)


editoast/src/views/train_schedule.rs
```diff
@@ -404,30 +405,37 @@ pub async fn train_simulation_batch(
...
-                    core.clone(),
-                    infra,
-                    chunk,
-                    consists_ref,
-                    electrical_profile_set_id,
-                )
-                .await
-            }
+            let consists = consists.clone();
+            let infra = <Infra as Clone>::clone(infra);
```
woshilapin, line 414: `infra.clone()` doesn't work? 🤔

editoast/src/models/timetable.rs
```diff
@@ -55,27 +56,64 @@ impl Timetable {
...
+    ///
+    /// ```
+    ///                       min_time               max_time
+    ///     Time Window           |----------------------|
+    /// Train 1 ✅         |-------------|
+    /// Train 2 ❌    |-------|
+    /// Train 3 ✅                              |----------|
+    /// Train 4 ✅   |-------?
+    /// Train 5 ✅                        |---------?
+    /// Train 6 ❌                                           |------?
```
woshilapin, line 75: Nice ASCII art. Since there is a lot edge case we want to support, this looks like it'd be nice to make some tests for all these edge cases.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -216,65 +216,10 @@ async fn stdcm(
...
-    // For trains departing before the LMR train departure (tA):
-
-    // If the train's arrival time is unscheduled (?), the train is kept (e.g., Train 4 and Train 5).
-    // If the train's arrival time is scheduled (|), the train is kept only if its arrival time is after the LMR train's earliest departure time.
-    // Train 1 is kept and train 2 is filtered out.
-
-    // Step 1
-    let mut train_schedules = timetable
-        .schedules_before_date(&mut conn, latest_simulation_end)
+    let train_schedules = timetable
```
woshilapin, line 219: `// Step 1` has been removed as well as `// Step 2 and 3` below. Other steps (comments) of the function might need to be updated.

editoast/src/models/timetable.rs
```diff
@@ -55,27 +56,64 @@ impl Timetable {
...
-            .await?
-            .map_ok(|ts| ts.into())
-            .try_collect::<Vec<TrainSchedule>>()
-            .await;
-        match train_schedules {
-            Ok(train_schedules) => Ok(train_schedules),
-            Err(err) => Err(err.into()),
-        }
+        let train_schedules = sql_query("
+        WITH last_path_step AS (
```
woshilapin: For big SQL queries, we try to put them in external files (helps also with syntax highlighting in our editors). See [here](https://github.com/OpenRailAssociation/osrd/blob/14fa21bf67cf757bd42881dbf5459adab5281a8c/editoast/src/models/infra/voltage.rs#L27) for example.

woshilapin (approved):
Nice improvement of performance 🚀 Only minor comments.

editoast/src/views/train_schedule.rs
flomonster: Nope `Infra::clone` exists

editoast/src/views/timetable/stdcm.rs
flomonster: These steps were only for filtering the timetable. No need to update other comments.

editoast/src/views/train_schedule.rs
woshilapin: What an unfortunate naming of the function 😄

editoast/src/models/timetable.rs
flomonster: Done

editoast/src/views/train_schedule.rs
flomonster: Yeah, it should probably be renamed `Infra::duplicate` or something like that.

editoast/src/models/timetable.rs
flomonster: Test added

PR #10696, opened 2025-02-05
front: fix power restrictions get reset on editing a train

### LINKED ISSUE #10494: Power restrictions get reset on first opening of train schedule
### What happened?

When a train has some power restrictions set, when you hard reload the page (ctrl + R) and you edit it, without changing anything the power restrictions are reset

https://github.com/user-attachments/assets/f146c40d-069a-4089-9d4e-16f2dba34d4c

I expect the power restrictions not be reset if we didn't change any parameter of the path

1. Create a train with power restrictions
2. Hard reload your browser (I guess it's emptying the cache)
3. Click on the edit button for this train
4. Witness the toast

Recette (SNCF)

Firefox

c9066aa

clarani:
closes #10494 

~After a hard refresh, the pathfinding was relaunched by the useEffect depending on the rolling stock in usePathfinding. Indeed, the trick with `isInitialized` did not work because it was set to true before the fetch of the rolling stock. Thus, when the rolling stock was received, the useEffect was triggered and the power restrictions reset.~

~Now, the hook usePathfinding is considered initialized only if there is no rollingStock selected or if the selected rolling stock has been already received.~

See commits


RomainValls (approved):
LGTM, tested ✅

emersion (commented):
Hmm, I'm not a fan of adding even more workarounds to the current workaround…

Why does the workaround exist in the first place? What are we trying to achieve with `isInitialized`?

If we want to launch pathfinding when the hook is mounted and when the infra state is cached and when the rolling stock is loaded, I don't think we need this `isInitialized` state?

Or do we want a different behavior?

front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
```diff
@@ -86,15 +81,14 @@ const ManageTrainSchedule = ({ trainIdToEdit }: ManageTrainScheduleProps) => {
...
-
-  const onSelectRollingStock = useCallback((_rollingStockId: number, comfort: Comfort) => {
-    dispatch(updateRollingStockID(_rollingStockId));
-    dispatch(updateRollingStockComfort(comfort));
-  }, []);
+  const onSelectRollingStock = useCallback(
+    (_rollingStockId: number, comfort: Comfort) => {
+      dispatch(updateRollingStockID(_rollingStockId));
+      dispatch(updateRollingStockComfort(comfort));
+      launchPathfinding(pathSteps);
```
emersion: Hm, `dispatch()` runs asynchronously so `launchPathfinding()` might get called before the rolling stock ID has actually been updated in the store… We might want to either pass a rolling stock ID to `launchPathfinding()` (the same way we pass `pathSteps`?), or `await` these two `dispatch()` calls?

clarani: Done :white_check_mark:

emersion (approved):
LGTM :)

Issue #10574, opened 2025-01-29
Adding categories to rolling stock model


hamz2a:
### Description and goal
This feature is the first step in managing the concept of "rolling stock categories." There are many ways to define these categories depending on the industry and usage. For the needs of OSRD, we have chosen the following categories (not specific to SNCF):
- high-speed-train
- intercity-train
- regional-train
- night-train
- commuter-train
- freight-train
- fast-freight-train
- tram-train
- touristic-train
- work-train

### Tasks:
- [x] [task 1](https://github.com/OpenRailAssociation/osrd/pull/10701):
1. Add two new columns to the model:
     - `primary_category`: `RollingStockCategory` (Required)
     - `other_categories`: `Vec<RollingStockCategory>` (Default: `Vec::default()`)
2. Add the columns to the table.
3. Update the model in `osrd-schemas`.
- [x] [task 2](https://github.com/OpenRailAssociation/osrd/pull/10840):
1. Add the columns to the endpoints (at a minimum, the edit endpoint).
    - Check that `primary_category` is not included in `other_categories`.

PR #10701, opened 2025-02-06
Add rolling stocks categories

hamz2a:
Part of https://github.com/OpenRailAssociation/osrd/issues/10574

**Editoast**:
1. A new migration file was added to create a `rolling_stock_category` enum and update the `rolling_stock` table with the new fields `primary_category` and `other_categories`.
2. New models (`rolling_stock_category` and `rolling_stock_categories`) were added to the `editoast_models` crate, as the goal is to migrate all models to this crate for better organization.
3. The `RollingStockModel` struct was updated to include `primary_category` and `other_categories`.
4. New test cases were added to check if `primary_category` and `other_categories` are correctly set when creating rolling stock.
5. `openapi.yaml` was updated to reflect the new `primary_category` and `other_categories` fields in `RollingStock`.

**Front**:
1. `generatedEditoastApi.ts` was updated to include the new `RollingStockCategory` and `RollingStockCategories` types.

**osrd_schemas**
1. New enum `RollingStockCategory` was added
2. `RollingStock` class was updated to include `primary_category` and `other_categories`.
3. Update the test files

**core**
1. updating rolling stock railjson version


hamz2a:
Hi @flomonster, I removed the `generate_missing_sql_type_definitions = false` configuration to allow Diesel to generate sql types automatically. Is that okay with you? Was there any particular reason for setting it to false?

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
```diff
@@ -0,0 +1,64 @@
...
+    fn from_sql(value: PgValue) -> diesel::deserialize::Result<Self> {
+        let s = std::str::from_utf8(value.as_bytes()).map_err(|_| "Invalid UTF-8 data")?;
+        RollingStockCategory::from_str(s)
+            .map_err(|_| "Unrecognized enum variant for RollingStockCategory".into())
+    }
+}
+
+impl ToSql<crate::tables::sql_types::RollingStockCategory, Pg> for RollingStockCategory {
+    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Pg>) -> diesel::serialize::Result {
+        let variant = self.to_string();
```
woshilapin: `.to_string()` does allocate memory, but we might be able to avoid that by implementing `strum::IntoStaticStr` and then do
```suggestion
        let variant: &str = self.into();
```

editoast/src/models/rolling_stock_model.rs
```diff
@@ -100,6 +102,9 @@ pub struct RollingStockModel {
     #[schema(value_type = Vec<String>)]
     #[model(remote = "Vec<Option<String>>")]
     pub supported_signaling_systems: RollingStockSupportedSignalingSystems,
+    pub primary_category: RollingStockCategory,
+    #[model(remote = "Vec<Option<RollingStockCategory>>")]
```
woshilapin, line 106: Why can't we have a `Vec<RollingStockCategory>` and need the `Option` nesting?

editoast/src/models/rolling_stock_model.rs
```diff
@@ -317,4 +325,47 @@ pub mod tests {
...
+            created_fast_rolling_stock.other_categories,
+            RollingStockCategories(vec![])
+        );
+    }
+
+    #[rstest]
+    async fn create_rolling_stock_with_categories() {
+        let db_pool = DbConnectionPoolV2::for_tests();
+
+        let rolling_stock = fast_rolling_stock_changeset("fost_rolling_stock")
```
woshilapin:
```suggestion
        let rolling_stock = fast_rolling_stock_changeset("fast_rolling_stock")
```

woshilapin (commented):
I'm not sure that I understand why we need to nest into `Option`? Apart from that, nice PR.

editoast/src/models/rolling_stock_model.rs
leovalais: That's because SQL array elements are always nullable, and SQL provides no way to ensure they're not. So the `Option` "leaks" from `diesel` trying to comply. But since we know that we won't have `NULL` in this array, we can strip it like this to manipulate a better type.

I've been thinking that this problem is occurring for a few models now, we probably can have `Model` handle this detail.

editoast/src/models/rolling_stock_model.rs
leovalais: That's worrying 😨 

image

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
hamz2a: done.

editoast/src/models/rolling_stock_model.rs
hamz2a: Done.

woshilapin: Thank you. Can you investigate why the test does not fail when the name is incorrect?

editoast/src/models/rolling_stock_model.rs
woshilapin: Let's not address that in the current PR then. I'm resolving.

editoast/src/models/rolling_stock_model.rs
hamz2a: There is no verification on the name, so we can enter whatever we want.

leovalais: Oh you're right that's just the name of the rs which isn't used anywhere... 🤦

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
leovalais, line 1: Could we merge `rolling_stock_categories.rs` and this file? I feel like there's no need for different modules when one is just "a `Vec` of the other".

woshilapin (approved):

editoast/migrations/2025-02-03-094532_add_primary_and_other_categories_to_rolling_stock/up.sql
```diff
@@ -0,0 +1,17 @@
...
+    'HIGH_SPEED_TRAIN',
+    'INTERCITY_TRAIN',
+    'REGIONAL_TRAIN_MU',
+    'NIGHT_TRAIN',
+    'COMMUTER_TRAIN',
+    'FREIGHT_TRAIN',
+    'FAST_FREIGHT_TRAIN',
+    'TRAM_TRAIN',
+    'TOURISTIC_TRAIN',
+    'WORK_TRAIN'
```
leovalais: nit: I think the issue writes categories name in `kebab-case`, which would be my preference as well. Is that required by postgres enums (I hope not 😨)?

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
```diff
@@ -0,0 +1,66 @@
...
+    FromSqlRow,
+    AsExpression,
+    EnumString,
+    IntoStaticStr,
+    Display,
+)]
+#[diesel(sql_type = crate::tables::sql_types::RollingStockCategory)]
+#[strum(serialize_all = "SCREAMING_SNAKE_CASE")]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+pub enum RollingStockCategory {
```
leovalais: Could you add a comment mentioning that this enum maps a Postgres enum and that any change made to this struct must be reflected in the DB and vice versa?

leovalais (commented):
Thanks for adding things in `editoast_models`, less things to do in the future!

Shouldn't the `-Form`s be updated as well with the categories?

> [!NOTE]
> The commit and the PR title use "categorie" instead of "category/ies"

front/src/modules/rollingStock/components/RollingStockSelector/sampleData.ts
SharglutDev, line 1: Looks like we don't use `ROLLING_STOCK_SAMPLE_DATA` anywhere ?

You could remove the file in a separate commit.

editoast/migrations/2025-02-03-094532_add_primary_and_other_categories_to_rolling_stock/up.sql
hamz2a: I just wanted to keep consistency with the frontend since we're using `SCREAMING_SNAKE_CASE`.

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
hamz2a: Done.

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
hamz2a: Done.

editoast/migrations/2025-02-03-094532_add_primary_and_other_categories_to_rolling_stock/up.sql
leovalais: Imo there's no consistency to maintain the frontend and the DB: separation of concerns and all. The frontend is free to define enums however they want and that shouldn't impact any decision made at the database level. Regarding the frontend, I'd say the only consistency we have to maintain is at the API level, so if any case change is required, I'd argue it should happen in the views.

front/src/modules/rollingStock/components/RollingStockSelector/sampleData.ts
hamz2a: Done.

hamz2a:
> The commit and the PR title use "categorie" instead of "category/ies"

Thanks for noticing that! I’ve fixed it now.

editoast/migrations/2025-02-03-094532_add_primary_and_other_categories_to_rolling_stock/up.sql
hamz2a: Done.

front/src/modules/rollingStock/components/RollingStockSelector/sampleData.ts
SharglutDev: I would have removed the file before your editoast commit so you don't have to make the change in it at all

hamz2a:
> Shouldn't the `-Form`s be updated as well with the categories?

As mentioned in this [issue](https://github.com/OpenRailAssociation/osrd/issues/10574), this PR addresses **Task 1**, which involves adding two new columns to the model and updating the model in `schemas`. **Task 2**, which involves adding these columns to the endpoints, will be handled in a separate PR.

front/src/modules/rollingStock/components/RollingStockSelector/sampleData.ts
hamz2a: Sorry, I don't understand. On the [dev branch,](https://github.com/OpenRailAssociation/osrd/blob/dev/front/src/modules/rollingStock/components/RollingStockSelector/sampleData.ts) we still have the file. Has your PR already been merged?

SharglutDev: Not sure what PR you mean :)

What I meant was : on your PR, instead of updating the sample data on your editoast commit and then removing it on the next front commit, you could make a first front commit to remove the sample data and then add your editoast commit that won't need any update on the sample data at all.

You can call me / pm me if its not very clear.

hamz2a: Oh, okay, sorry, I didn't understand the previous comment.

hamz2a: Done.

SharglutDev (approved):
Lgtm for front (not tested)

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
```diff
@@ -0,0 +1,84 @@
...
+use utoipa::ToSchema;
+
+editoast_common::schemas! {
+    RollingStockCategory,
+    RollingStockCategories,
+}
+
+/// This enum maps to a Postgres enum type, specifically `rolling_stock_category`.
+/// Any changes made to this enum must be reflected in the corresponding Postgres enum,
+/// and vice versa, to ensure consistency between the application and the database.
```
leovalais:
```suggestion
// This enum maps to a Postgres enum type, specifically `rolling_stock_category`.
// Any changes made to this enum must be reflected in the corresponding Postgres enum,
// and vice versa, to ensure consistency between the application and the database.
```

It's not really a doc comment, and doing so makes it appear in the OpenAPI

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
```diff
@@ -0,0 +1,84 @@
...
+    Display,
+)]
+#[diesel(sql_type = crate::tables::sql_types::RollingStockCategory)]
+#[strum(serialize_all = "kebab-case")]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+pub enum RollingStockCategory {
+    Unknown,
+    HighSpeedTrain,
+    IntercityTrain,
+    RegionalTrainMU,
```
leovalais: What does `MU` stands for? I'd suggest to respect CamelCase even for acronyms:

```suggestion
    RegionalTrainMu,
```

editoast/openapi.yaml
```diff
@@ -9494,6 +9500,28 @@ components:
...
+      type: string
+      description: |-
+        This enum maps to a Postgres enum type, specifically `rolling_stock_category`.
+        Any changes made to this enum must be reflected in the corresponding Postgres enum,
+        and vice versa, to ensure consistency between the application and the database.
+      enum:
+      - UNKNOWN
+      - HIGH_SPEED_TRAIN
+      - INTERCITY_TRAIN
+      - REGIONAL_TRAIN_M_U
```
leovalais: Yep weird things happen

leovalais (commented):
> Task 2, which involves adding these columns to the endpoints, will be handled in a separate PR.

Oh okay (I'm curious why this is divided since this is a rather atomic change, but it doesn't really matter much)

> Update the model in osrd-schemas.

I believe this part is missing then ^^

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
woshilapin: I believe it stands for "Multiple Unit" 🤷 Do you think it would make sense to have the long version of it

```suggestion
    RegionalTrainMultipleUnit
```

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
hamz2a: Done.

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
hamz2a: Done.

editoast/openapi.yaml
hamz2a: https://github.com/OpenRailAssociation/osrd/pull/10701#discussion_r1946628656

python/osrd_schemas/osrd_schemas/rolling_stock.py
shenriotpro, line 1: Please update the package version as well

shenriotpro (approved):
LGTM for tests

python/osrd_schemas/osrd_schemas/rolling_stock.py
hamz2a: Done.

editoast/migrations/2025-02-03-094532_add_primary_and_other_categories_to_rolling_stock/up.sql
Synar, line 1: Are you sure we want to add an unknown value?

python/osrd_schemas/osrd_schemas/rolling_stock.py
```diff
@@ -12,10 +12,28 @@
...
 RAILJSON_ROLLING_STOCK_VERSION = get_args(RAILJSON_ROLLING_STOCK_VERSION_TYPE)[0]
 
 
+class RollingStockCategory(str, Enum):
+    """
+    This enum defines the rolling stock category that a train can belong to.
+    """
+
+    UNKNOWN = "Unknown"
+    HIGH_SPEED_TRAIN = "HighSpeedTrain"
```
Synar: Nitpick : If there is no other more important reason to use this case, I would rather we use human readable case (with spaces) for the strings

Or if impossible, then snake case.

Synar: (Feel free to resolve if you disagree)

hamz2a: Done.

python/osrd_schemas/osrd_schemas/rolling_stock.py
```diff
@@ -12,10 +12,28 @@
...
 from .infra import LoadingGaugeType
 
-RAILJSON_ROLLING_STOCK_VERSION_TYPE = Literal["3.2"]
+RAILJSON_ROLLING_STOCK_VERSION_TYPE = Literal["3.3"]
 RAILJSON_ROLLING_STOCK_VERSION = get_args(RAILJSON_ROLLING_STOCK_VERSION_TYPE)[0]
 
 
+class RollingStockCategory(str, Enum):
+    """
+    This enum defines the rolling stock category that a train can belong to.
```
leovalais:
```suggestion
    The list of categories that can be assigned to a rolling stock.
```

The term "train" can refer to too many things in OSRD's context, we should avoid using it on its own as much as possible imo. (In most contexts it actually refers to a train schedule instance.)

python/osrd_schemas/osrd_schemas/rolling_stock.py
```diff
@@ -12,10 +12,28 @@
...
+    HIGH_SPEED_TRAIN = "High Speed Train"
+    INTERCITY_TRAIN = "Intercity Train"
+    REGIONAL_TRAIN_MULTIPLE_UNIT = "Regional Train Multiple Unit"
+    NIGHT_TRAIN = "Night Train"
+    COMMUTER_TRAIN = "Commuter Train"
+    FREIGHT_TRAIN = "Freight Train"
+    FAST_FREIGHT_TRAIN = "Fast Freight Train"
+    TRAM_TRAIN = "Tram Train"
+    TOURISTIC_TRAIN = "Touristic Train"
+    WORK_TRAIN = "Work Train"
```
leovalais: Nit: could we keep kebab-case here as well? Otherwise we'll require some conversion to be done in the import code.

leovalais (approved):
Thanks for the changes!

python/osrd_schemas/osrd_schemas/rolling_stock.py
hamz2a: @shenriotpro what do you think ?

python/osrd_schemas/osrd_schemas/rolling_stock.py
hamz2a: Done.

python/osrd_schemas/osrd_schemas/rolling_stock.py
shenriotpro: I don't really mind either way, but kebab-case is surprising to me, doesn't seem very consistent with other enums.

leovalais: My point was: `editoast` usually accepts instances of the schemas defined in `osrd-schemas` (especially in the command line). If the schema has string variants defined like this, they'll be transformed into kebab-case anyway since that's what we store in the DB. So we'll have to keep that in mind when debugging and writing tests. I feel like it's needlessly confusing 🤷 Besides keeping this custom case rule makes `serde` unable to attribute values to each `enum` variants, unlike for "kebab-case" or "CamelCase".

It's not the end of the world of course, but the benefits are unclear to me :/

(Btw most enums in `osrd-schemas` use SCREAMING_SNAKE_CASE.)

eckter (approved):
LGTM for core

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
```diff
@@ -0,0 +1,84 @@
...
+    ToSchema,
+    FromSqlRow,
+    AsExpression,
+    EnumString,
+    IntoStaticStr,
+    Display,
+)]
+#[diesel(sql_type = crate::tables::sql_types::RollingStockCategory)]
+#[strum(serialize_all = "kebab-case")]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
```
flomonster: Why do we need a different format between serialization and display?

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
```diff
@@ -0,0 +1,84 @@
...
+pub enum RollingStockCategory {
```
flomonster: @leovalais Why this enum is defined in the model and not in the `osrd-schema`.

I don't see how different this enum is from `LoadingGaugeType`. The enum should be defined in the schema and used in models with the annotation `to_enum`.

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
```diff
@@ -0,0 +1,84 @@
...
+    Unknown,
```
flomonster: I don't think that adding an `Unknown` value is useful. I get that we need a value to initialize the primary_category, but in the end, this variant is not permitted.

If you plan to do another migration and remove this field you will have to handle RS that have this `Unknown` category assigned anyway.

In my opinion, we should assign an arbitrary legit value.

flomonster (changes requested):
Thanks for your PR. There are a few things I'd like to clarify before I approve.

editoast/migrations/2025-02-03-094532_add_primary_and_other_categories_to_rolling_stock/up.sql
hamz2a: @flomonster please take a look at the comments above to understand why there is a difference between the DB serialization and `serde` serialization.

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
hamz2a: https://github.com/OpenRailAssociation/osrd/pull/10701#discussion_r1950522505

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
leovalais: A random category would work as well. But removing an unused enum variant is cheap. The benefit is that if a rolling stock gets forgotten and isn't assigned a category, the migration won't succeed and we'll spot the mistake early and easily. A remaining randomly assigned category could stay unnoticed for quite a while.

I don't mind the other way, but if we can cheaply have a small safety net at the cost of a very straightforward additional migration, why not? 🤷

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
leovalais: Agreed that this enum should be defined in `editoast-schemas`.

Disagreed about enums being stored as integers instead of PG enums (cc @woshilapin who suggested this). I believe we should allow a `#[model(enum = <type>)]` where `type` is either `u8/16/64` or `&str` by default. (We're not proposing to store variants as strings in db, we just convert them to string to write the query.)

python/osrd_schemas/osrd_schemas/rolling_stock.py
leovalais: Closing this to continue the discussion in https://github.com/OpenRailAssociation/osrd/pull/10701#discussion_r1950456609

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
leovalais: We don't, I probably wasn't explicit enough sorry. I suggested to use `kebab-case` for *both* the Postgres enum and the schema. (The point is to have the same format for everything to avoid confusion. I'd be fine as well with screaming snake case, but I think kebab case makes more sense here since the values are nouns and not acronyms.)

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
hamz2a: If we move the categories enum into `editoast-schemas`, we need to add `diesel` as a dependency in `editoast-schemas`, and I don't think it's desirable. what do you think?

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
hamz2a: `ReceptionSignal`, `InfraState`, `Comfort`, `Direction`, and `ApplicableDirections` are in `SCREAMING_SNAKE_CASE` in OpenAPI. It would be better to keep the same format for `RollingStockCategory` to stay consistent.

leovalais: There's already inconsistencies (`FlagSignalParameter`, `StudyState`, `StudyType`). Either we agree on fixing them (not necessarily in this PR ^^), or adding another case isn't really a issue.

Imo, if we accept different cases, we should choose values that:

* are the most convenient to use later
* are the easiest to read (kebab-case or snake_case for words, SCREAMING SNAKE for acronyms imo)

These enum values should be parsed and forgotten as early as possible, so the content doesn't really matter ultimately.

On the other hand this case change breaks the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). But since RailJSON today is more a transport schema rather than a user-facing format designed for manual edition, maybe this principle isn't really applicable?

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
leovalais: You're right we should try to keep `diesel` only in `editoast_models`, you'll need an `impl From<editoast_schemas::RSCategory> for editoast_models::RsCategory`

hamz2a: If I understand correctly, we need to define another enum for categories in `editoast_schemas`. If so, why is it necessary since the enum is already in `editoast_models`?

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
hamz2a: For now, it's mainly the frontend team using these enum formats. Should we ask for their opinion? What do you think?

leovalais: Unless I'm wrong, the frontend use its own TypeScript enums and never deal with the actual values as strings (besides the actual `enum` definition of course). So the case shouldn't matter to the frontend either.

> For now, it's mainly the frontend team using these enum formats

Dags use them as well via the `osrd-schemas` directly ^^

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
leovalais: Because `editoast-schemas` aims to be the Rust counterpart of `osrd-schemas`

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
hamz2a: The actual process is as follows, or am I missing something?
`#[serde(rename_all = "SCREAMING_SNAKE_CASE")]` -> `openapi.ymal` -> `generatedEditoastApi.ts` 

Perhaps we should ask the dugs' team for their opinion.

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
hamz2a: So, I need to define another enum in `editoast-schemas`, right?

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
leovalais: (Again, I'm fine with SCREAMING SNAKE as well, I'm just not against kebab case since we're already mixing cases in the schemas.)

hamz2a: In my opinion, it’s better to gradually move towards a single serialization format. That’s why I chose `SCREAMING_SNAKE_CASE`, as it’s the format mostly used for enums in dags and frontend.

Khoyo: `FlagSignalParameter` is *special* (as in, it doesn't come from an enum and is only there to please the frontend form generator), changing it will requires non obvious changes...

I'd advise keeping it out of any consistency change.

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
flomonster: A migration should not fail. From an open source point of view, we don't expect people to adapt the rolling stock to bump OSRD.

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
flomonster: Since the enum is now defined in `osrd-schemas` we could do something like this to avoid duplication:

```suggestion
use osrd_schemas::rolling_stock::RollingStockCategory as RollingStockCategorySchema;
pub struct RollingStockCategory(RollingStockCategorySchema);
```

flomonster (commented):
I still do not agree with the `unknown` variant. If the other maintainers agree with this, then I'll yield.

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
hamz2a: Done.

hamz2a:
> I still do not agree with the `unknown` variant. If the other maintainers agree with this, then I'll yield.

@leovalais @woshilapin what do you think? 
https://github.com/OpenRailAssociation/osrd/pull/10701#issuecomment-2659620246

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
leovalais: > we don't expect people to adapt the rolling stock to bump OSRD

That's exactly what we're doing though. We're adding a new required field to an existing schema. So if a user with an existing rolling stock DB wants to use the new OSRD version, they **have to** fill in the new data to keep the app in a working state. Assigning random data works from the technical perspective, but I'd hardly qualify that state of "working" since OSRD might then produce nonsensical results.

hamz2a:
As discussed with @flomonster, @leovalais and @woshilapin: we will use `SCREAMING_SNAKE_CASE` format, remove the `unknown`, and set a default category chosen randomly.

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
hamz2a: https://github.com/OpenRailAssociation/osrd/pull/10701#issuecomment-2659620246

editoast/editoast_models/src/rolling_stock/rolling_stock_category.rs
hamz2a: https://github.com/OpenRailAssociation/osrd/pull/10701#issuecomment-2659620246

flomonster (approved):
LGTM

PR #10703, opened 2025-02-06
Add tls feature to redis

flomonster:
Prod requires connecting to Elasticache with tls enabled.


woshilapin (commented):
Hmm, annoying that `deadpool_redis` doesn't expose that as a feature, but since that's what they document for now, I guess this is the way to go.

woshilapin (approved):

woshilapin (commented):
Would it make sense to use `tls-rustls` instead?

flomonster:
> Would it make sense to use `tls-rustls` instead?

Tokio does not compile if we use tls-rustls.

PR #10706, opened 2025-02-06
front: fix cached power restrictions on GEV

### LINKED ISSUE #10493: Power restriction in GEV are the ones of another schedule
### What happened?

When displaying the GEV of the second train (which has no power restriction), I see the restrictions of the first train

https://github.com/user-attachments/assets/66ae28f0-fe50-4413-95ba-c21b46fa31c5

I expect no power restriction in the GEV of the second train

1. Get a small infra scenario
2. Load [these train schedules](https://github.com/user-attachments/files/18517543/train_schedules.2.json)
3. Look at the GEVs of two trains, while displaying the power restrictions

Recette (SNCF)

Firefox

c9066aa

clarani:
closes #10493 

The power restrictions were computed only when the path of the selected train changed, and not when the selected train himself changed. Thus, if 2 trains had exactly the same path, then the power restrictions were not re-computed when switching from one to another.

Now, the power restrictions are computed each time the selected train or the pathfinding changes.

### DIFF
--- a/front/src/modules/simulationResult/components/SpeedSpaceChart/useSpeedSpaceChart.ts
+++ b/front/src/modules/simulationResult/components/SpeedSpaceChart/useSpeedSpaceChart.ts
@@ -72,20 +72,23 @@ const useSpeedSpaceChart = (
         );
 
         setFormattedPathProperties(formattedPathProps);
-
-        // Format power restrictions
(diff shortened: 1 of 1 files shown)


anisometropie (approved):
Tested

Math-R (approved):
lgtm

PR #10707, opened 2025-02-06
front: enable import/no-extraneous-dependencies lint

emersion:
This is a valuable lint to figure out cases where we're missing a
dependency (and only get it by chance because another dependency
pulls it in).

_See individual commits._


front/.eslintrc
Synar, line 1: I believe we should include the e2e tests in front/test as allowed to use dev dependencies

Synar, line 1: Could we change all the numerical values to off/warning/error? I think it would improve readability and consistency

front/src/utils/__tests__/geometry.ts
Synar, line 1: I believe this will skip the test, as vitest is configured to ```include: ['src/**/*.{test,spec},{js,mjs,cjs,ts,mts,cts,jsx,tsx}']```

Removing .test.ts as a possibility in the config would make sense with your changes, but I'm not sure about including all \_\_tests\_\_, are we sure that they couldn't include test utils? But perhaps running vitest over a couple non test files would not be the biggest issue (as long as that doesn't cause it to fail).

Synar (commented):
Overall nice changes, just a couple small issues.

(Perhaps it would be a good idea to have a short meeting at some point to more generally review and update the linting rules we are using)

front/.eslintrc
emersion: Well, the linter didn't complain, so I'm not sure it's really necessary.

emersion: Sure, thought the same, but that's orthogonal to this PR I think.

front/src/utils/__tests__/geometry.ts
emersion: Ouch, good catch! I thought multiple tests were already missing the `.spec.ts` extension in `__tests__/`, but it seems it's all helpers. I'll revert that change.

front/.eslintrc
Synar: Ah, I guess we are in fact not linting front/tests at all! 

Perhaps we should, but this is outside the scope of this pr

Synar: Sure ^^

emersion: cc @Maymanaf

Maymanaf: we can include the e2e tests, I checked no issues with the linter

front/src/utils/__tests__/geometry.ts
emersion: Fixed!

Synar (approved):
LGTM ^^

front/.eslintrc
Synar: @Maymanaf When running eslint on front/tests, I do get a few warnings, even on dev :

```bash
❯ npx eslint --cache --ext .ts,.tsx,.js,.jsx tests --max-warnings 0
(node:30578) ExperimentalWarning: CommonJS module /nix/store/cgkpjpl1hspg30dlmv4hhvlhbvik1bn1-nodejs-23.2.0/lib/node_modules/npm/node_modules/debug/src/node.js is loading ES Module /nix/store/cgkpjpl1hspg30dlmv4hhvlhbvik1bn1-nodejs-23.2.0/lib/node_modules/npm/node_modules/supports-color/index.js using require().
Support for loading ES Module in require() is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)

/home/alice/dev/sncf/osrd/front/tests/utils/api-setup.ts
   54:3   warning  Unsafe return of a value of type `Promise<any>`                               @typescript-eslint/no-unsafe-return
   89:3   warning  Unsafe return of a value of type `Promise<any>`                               @typescript-eslint/no-unsafe-return
  213:10  warning  Use a ! assertion to more succinctly remove null and undefined from the type  @typescript-eslint/non-nullable-type-assertion-style

/home/alice/dev/sncf/osrd/front/tests/utils/index.ts
  78:47  warning  Unsafe return of a value of type `any`  @typescript-eslint/no-unsafe-return

✖ 4 problems (0 errors, 4 warnings)
  0 errors and 1 warning potentially fixable with the `--fix` option.

ESLint found too many warnings (maximum: 0).
```

Linting tests too would be a nice change though, I guess I'll open a PR for that later

Maymanaf: I thought we were talking about the dependency, but yeah, there are some lint warnings we can handle in a different PR

clarani (approved):
LGTM, thanks for this PR :pray:

PR #10708, opened 2025-02-06
front: do not display errors layer by default in the infra editor

### LINKED ISSUE #6711: When opening the editor don't show the warnings

End users

Currently when you open the editor with an infra there is a focus on warnings on the map. I my case I don't want to see these so I have to click to remove them.
Each Time I change the object id in the url to focus on a new object I want to check I have to remove the focus on warning... It is a bit painful.

Less click and a happier user.

I don't think it is too complicated and it would improve a lot my use of the editor

clarani:
closes #6711


emersion (approved):
LGTM!

achrafmohye (approved):
LGTM thanks :)

PR #10711, opened 2025-02-06
front: import osrd-ui

### LINKED ISSUE #451: Move osrd-ui into main osrd repository
### Description and goal

A pain point when working on osrd-ui is that it's tricky to work on osrd-ui and test the changes in OSRD in a real-world scenario instead of a storybook. Additionally, osrd-ui changes are costly: each time we want to update osrd-ui (e.g. for a bugfix or small enhancement) we need to create a new osrd-ui release and then bump the osrd-ui dependency in OSRD's `package.json`.

We could move osrd-ui into a subdirectory of the main OSRD repository. osrd-ui's NPM packages would continue to work as usual. OSRD's front could pull osrd-ui from a local directory instead of the NPM registry.

Possible roadblocks:

- How do we integrate osrd-ui's multi-package/lerna setup into OSRD's front build process ?
- Do we want to retain git history? (It's possible)
- yarn vs npm?
- Release process? Right now we use git tags, GitHub releases and a GitHub actions workflow.

### Acceptance criteria

.

emersion:
_See individual commits._

TODO:

- [x] Basic NPM workspace setup
- [x] Build system
- [x] Docker integration
  - [x] Base flavor
  - [x] Front flavor
- [x] Storybook publishing
- [x] Releases
- [x] Adjust docs

Closes: https://github.com/OpenRailAssociation/osrd-ui/issues/451

### Why move `front/ui/package.json` to `front/ui/base/package.json`?

NPM doesn't correctly handle nested sub-packages (`front/ui/` and `front/ui/storybook/`, for instance). It doesn't install dev dependencies of nested sub-packages (storybook dependencies are missing).

### Why keep `front/ui/base/package.json`?

`front/ui/` contains config files for rollup, ESLint, postcss used by all ui-* sub-packages. We need to specify dependencies for these somewhere.

We could specify them in `front/package.json`, but that would mix up osrd and osrd-ui deps. Keeping shared osrd-ui deps inside `front/ui/base/package.json` makes it clearer what's required by osrd itself and what's required by osrd-ui.

### Should we build osrd-ui when running `npm run build`?

Another possible solution would've been to only build osrd when running `npm run build`, and require users to run `npm run build --workspaces` to build osrd-ui. The upside would be that less work is performed when only touching osrd.

However, this would make it so `npm run build` would fail if osrd-ui hasn't been built before, or has an outdated build. Also, building osrd-ui requires less changes to the CI and tooling.

### Should we build osrd-ui's storybook when building osrd's frontend?

The osrd-ui storybook is built when running `npm run build` in `front/`, but is not required for the frontend.

This simplifies a bit CI, because there is a single build/test/lint command to execute instead of separate commands for osrd itself and the storybook. If we choose to only execute the bare minimum there might be some inconsistencies between the scripts: build will need to build osrd-ui (without the storybook) but test/lint would not operate on osrd-ui. Additionally, `npm run build` in `front/` is seldom used outside of CI, and the current approach requires less changes in the CI manifests.

One could argue that we should not build the storybook when running `npm run build`. There are already some discrepancies in the scripts, e.g. `compile` and `lint-fix` don't operate on osrd-ui, and `start` requires osrd-ui to be already built.

### Why not build osrd-ui on startup for the front flavor of the Compose setup?

Right now, the container might spew some errors the first time it's started because it's missing an osrd-ui build. The osrd-ui build happens in parallel, so the errors should eventually go away, except for ESLint which unfortunately cannot watch osrd-ui. Restarting the container should fix the ESLint issues. This should only happen the very first time the front container is started.

We could build osrd-ui before starting osrd, but this would add 30s or 1min of extra delay every time the container starts up. This was deemed too annoying, especially if restarting the container fixes it.

### There are ESLint errors when I start the front flavor of the Compose setup

This is due to osrd-ui and osrd being built in parallel when the container starts, and ESLint doesn't notice when osrd-ui has been built. Indeed, ESLint doesn't install a watcher on files in `node_modules/`, which osrd-ui deps point to.

To fix, run `npm run build --workspaces` before starting the container, or restart the container (either should work).

### How will we merge this?

We will need to coordinate a bit to not loose anything in osrd-ui. Easiest would be to get this reviewed and ready to merge, freeze osrd-ui, release a new osrd-ui version and bump osrd to use that new version (if there are breaking API changes), adjust the "import osrd-ui" commit to import that version, and merge this PR.

Once that's done, all opened osrd-ui PRs will need to be manually migrated to the osrd repository.

<a name="migrate-pr"></a>
One can migrate a PR from osrd-ui to osrd like so:

- Create a new branch in osrd
- Run the command below to apply the commits locally
- Push to GitHub
- Open a new PR with a link to the old osrd-ui PR in the description
- Assign reviewers

```sh
pr=976
curl -L "https://github.com/OpenRailAssociation/osrd-ui/pull/$pr.patch" | git am --directory=front/ui
```

### Why `front/ui/` instead of `front/osrd-ui/`?

Because "osrd" is already in the name of this repo  ¯\\\_(ツ)\_/¯


.github/workflows/ui-icons-optimize.yml
```diff
@@ -0,0 +1,39 @@
+on:
+  # This workflow will be laucnhed manually and will trigger a new commit
+  # on the branc that is specified.
```
Yohh, line 3:
```suggestion
  # on the branch that is specified.
```

front/README.md
```diff
@@ -2,9 +2,26 @@
...
+run the frontend in watch mode together with the rest of the OSRD stack. The
+osrd-compose script can be used to start OSRD in this mode:
+
+    ./osrd-compose dev-front build
+    ./osrd-compose up -d
+
+The first time the container starts up, the osrd-ui library will be missing.
+This will trigger some build errors, which should go away as soon as osrd-ui
+gets built. Restarting the front container helps getting rid of lingering
+ESLint errors.
```
Yohh, line 17: not sure to understand this, will it be the same every time OSRD is fully setup?

front/package.json
```diff
@@ -145,20 +153,26 @@
...
+    "lint": "npm run lint --workspaces --if-present && eslint --ext .ts,.tsx,.js,.jsx src scripts tests *.ts --max-warnings 0",
+    "lint-fix": "eslint --ext .ts,.tsx,.js,.jsx src scripts tests *.ts --fix",
     "test-coverage": "vitest run --coverage",
     "generate-types": "exec scripts/generate-types.sh",
     "generate-licenses": "license-checker-rseidelsohn --json --production --direct --excludePackagesStartingWith '@types' --excludePrivatePackages --customPath src/common/ReleaseInformations/json/licenseCustomFormat.json > src/common/ReleaseInformations/json/licenses.json",
     "e2e-tests": "playwright test",
     "i18n-api-errors": "tsx ./scripts/i18n-api-errors.ts",
-    "i18n-checker": "tsx ./scripts/i18n-checker.ts"
+    "i18n-checker": "tsx ./scripts/i18n-checker.ts",
+    "build-ui": "npm run build --workspace ui/ui-icons --workspace ui/ui-core --workspace ui/ui-charts --workspace ui/ui-warped-map"
```
Yohh, line 168:
```suggestion
      "build-ui": "npm run build --workspaces --if-present --exclude ui/base --exclude ui/storybook"
```

front/ui/.gitignore
```diff
@@ -0,0 +1,13 @@
+# Common
+dist/
+node_modules/
+*.tgz
+storybook-static
+
+# Direnv
+.envrc
+.direnv
```
Yohh: already in root gitignore, don't think it,s needed here

front/ui/README.md
```diff
@@ -0,0 +1,66 @@
...
+}
+```
+
+If this section is not added, the package will not be published. Having a `"private": true` in the
+package.json will also prevent the package from being published.
+
+# Build Instructions
+
+1. Run `npm install`
+2. Run `npm run build`
```
Yohh: I didn't need it after running `/front` `npm run build-ui`, I also think this section needs some extra comments to explain how it works now it's integrated in osrd main repo

front/vite.config.ts
```diff
@@ -25,8 +25,7 @@ export default defineConfig(({ mode }) => {
           buildMode: true,
         },
         eslint: {
-          lintCommand:
-            'eslint --cache --ext .ts,.tsx,.js,.jsx src scripts tests *.ts --max-warnings 0',
+          lintCommand: 'eslint --ext .ts,.tsx,.js,.jsx src scripts tests *.ts --max-warnings 0',
```
Yohh, line 28: why did we add it?

front/README.md
Yohh: just tried it, I have no errors, no missing library

Yohh:
> To fix, run `npm run build --workspaces` before starting the container, or restart the container (either should work).

what about updating `osrd-compose` script for `dev-front` mode and run this command?

emersion:
> what about updating osrd-compose script for dev-front mode and run this command?

This would delay the container startup by 30s or 1min every time it's started, which was deemed more annoying.

front/vite.config.ts
emersion: See the commit description for "front: disable ESLint cache". I added `--cache` a while back to improve ESLint's startup time.

front/README.md
emersion: This only happens the very first time, with a freshly cloned osrd repository. Maybe you were lucky and osrd-ui finished building before ESLint kicked in. (It's all done in parallel, so either may win the race.)

.github/workflows/ui-icons-optimize.yml
emersion: I'd prefer to not fix any unrelated issues in files moved around. This file comes from osrd-ui and is only slightly adjusted to accommodate for the new file paths.

front/package.json
emersion: `npm run` doesn't have an `--exclude` flag.

front/ui/.gitignore
emersion: Indeed, dropped!

front/ui/README.md
emersion: Indeed, I changed it to `npm run build-ui` here to only build osrd-ui instead of osrd + storybook.

What exactly would you want to add here? I think with these updated commands it should work the same as before?

Yohh: I just wonder if it's a good thing to keep those instructions here, this script is from upper package.json, in `front/` which already has a ReadMe with setup instructions...

emersion: I think it's still useful to have instructions to work on osrd-ui alone, without touching/building osrd itself, especially for external contributors. This README is a good place for that IMHO, to avoid having these users have to go through a lot of other osrd-specific instructions and figure out which they should ignore.

Yohh: ok fine

front/ui/README.md
```diff
@@ -10,8 +10,7 @@ Development will leverage preconstruct. **Documentation yet to come**.
 
 ## Publishing versions
 
-In a nutshell: we consider our version numbers metadata of the project. We use
-[Semantic Versioning](https://semver.org/).
+osrd-ui releases are published automatically together with OSRD releases. osrd-ui's latest version is always in sync with OSRD.
```
clarani: This might affect our external users :thinking: But I remember the discussions about it :+1:

front/docker/Dockerfile.devel
```diff
@@ -23,4 +23,4 @@ ENV VITE_OSRD_GIT_DESCRIBE=${OSRD_GIT_DESCRIBE}
 # Start the app
 COPY docker/dev-entrypoint.sh /
 ENTRYPOINT ["/dev-entrypoint.sh"]
-CMD ["sh", "-c", "id && npm install && exec npm run start-container"]
```
clarani, line 26: Only small questions by curiosity :) 
Why do you remove the `exec` ? Adding the `()` run the 2 commands in parallel ?

front/package.json
```diff
@@ -3,15 +3,23 @@
   "version": "0.1.0",
   "private": true,
   "type": "module",
+  "workspaces": [
+    "ui/base",
```
clarani, line 7: What `ui/base` is used for ?

front/docker/Dockerfile.devel
emersion: `exec` is a shell built-in utility which replaces the shell with another command: the shell's process is re-used for executing the supplied command. With `npm run start`, there are two processes, the shell and npm as a child process. With `exec npm run start`, there is only a single process, npm.

`&` runs the previous command in parallel with the rest. When evaluating `a & b`, the shell forks, evaluates `a` in the child, then evaluates `b` in the parent.

`exec` can only be used to execute a single process. To execute two commands in parallel, the shell needs to remain (not be replaced by another command) to coordinate everything.

The `()` is to disambiguate what runs in parallel. With `a && b & c` it's unclear whether `a && b` runs in parallel with `c`, or if `a` is executed and then `b` and `c` run in parallel. (Only folks who wrote a POSIX shell implementation remember that without the parentheses, `a && b` runs in parallel with `c`.)

front/package.json
emersion: See the section "Why keep front/ui/base/package.json?" in the PR description :P

front/ui/README.md
emersion: Yeah. We can change how releases are managed later on if needed.

front/ui/README.md
```diff
@@ -0,0 +1,77 @@
...
+  },
+  // ...
+}
+```
+
+If this section is not added, the package will not be published. Having a `"private": true` in the
+package.json will also prevent the package from being published.
+
+# Build Instructions
+
```
theocrsb, line 54: I don't know if this is very useful, but perhaps we could specify the folder in which to run these commands, as in the `README` of `OSRD`

theocrsb:
We have an error + white screen on `Docs` pages

image

image

front/ui/README.md
```diff
@@ -0,0 +1,66 @@
...
+1. Run `npm install`
+2. Run `npm run build-ui`
+3. Run `cd storybook` and `npm start`
+
+While developing on a project, you need to run the following commands to see the css modifications:
+
+1. Run `cd ui-<name-of-the-project>`
+2. Run `npm run watch`
```
theocrsb, line 63: Personally, I didn't need to run the `watch` command to see the changes :)

kmer2016 (commented):
I started the stack using Docker:
```
./osrd-compose dev-front build    
./osrd-compose up -d
```
Hot reload works for changes in the osrd codebase, but it doesn't for changes in osrd-ui.

emersion:
> Hot reload works for changes in the osrd codebase, but it doesn't for changes in osrd-ui.

This is expected. Hot reload for osrd-ui requires more changes and is trickier to implement (because of the relative dependencies between osrd-ui packages).

Hot reload should work if you start `npm run watch` in the osrd-ui package you're working on.

front/ui/README.md
emersion: I don't understand how that could be. Without `watch`, the files in `dist/` are never updated.

front/docker/Dockerfile.devel
clarani: Ok thanks :)

front/package.json
```diff
@@ -145,13 +153,18 @@
...
+    }
+  },
   "scripts": {
     "compile": "tsc -b",
     "start": "vite",
     "start-container": "vite --host",
-    "build": "vite build",
-    "test": "vitest",
-    "lint": "eslint --cache --ext .ts,.tsx,.js,.jsx src scripts tests *.ts --max-warnings 0",
+    "build": "npm run build --workspaces --if-present && vite build",
```
clarani, line 159: Do we need to build storybook here too ?

front/ui/README.md
theocrsb: Yet we did the test with @clarani. but only in storybook

front/package.json
emersion: Ah, good point, I didn't explain the motivation. I've added a new section "Should we build osrd-ui's storybook when building osrd's frontend?" in the PR description. Happy to change it if you prefer.

emersion:
> We have an error + white screen on Docs pages 

I could not reproduce this bug. What I did, from a clean slate:

```
cd front
npm install
npm run build-ui
cd ui/storybook
npm run start
```

front/ui/README.md
emersion: What I did:

- `cd front/ui/storybook` then `npm run start`
- Open the manchette story in the browser
- Edit `ui/ui-charts/src/manchette/components/Waypoint.tsx` to add "SALUT THÉO" in `.waypoint-name`
- The story isn't updated
- In another terminal, `cd front/ui/ui-charts` then `npm run watch`
- The story is updated

flomonster (approved):
LGTM.

Note: I found some refs to "https://github.com/osrd-project/osrd-ui/". Do you think we should replace it in this PR?

We will need some follow-up PR to clean up CI files.

front/ui/README.md
theocrsb: indeed, thank you!

front/ui/README.md
emersion: Sure, I've added an extra bullet point to make this clearer!

emersion:
> I found some refs to "https://github.com/osrd-project/osrd-ui/". Do you think we should replace it in this PR?

Good catch, I've added a commit to fixup these.

front/ui/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+A [live storybook](https://openrailassociation.github.io/osrd-ui/) showcases all components.
+
+## Development
+
+Development will leverage preconstruct. **Documentation yet to come**.
+
+## Publishing versions
+
+osrd-ui releases are published automatically together with OSRD releases. osrd-ui's latest version is always in sync with OSRD.
```
SharglutDev: This has probably been answered already but what if, when publishing a new osrd release, the last osrd-ui release hasn't been bumped on osrd ? (or, if we don't release osrd-ui individually anymore, the last osrd-ui changes haven't been added to osrd)

emersion: OSRD always uses the local osrd-ui directory instead of the NPM package. If a breaking change is made to osrd-ui, OSRD needs to be updated to account for the change or else the build fails.

ghost (approved):

ghost:
Thank you very much, I mostly checked the GitHub CI part, but overall it looks good to me.

SharglutDev (approved):
Thank you for the great work ! (not tested)

kmer2016 (approved):
Very good ! Thank you for the PR !

emersion:
Rebased and updated for osrd-ui v0.0.73.

I've also appended a commit to temporarily disable automatic releases for now:
https://github.com/OpenRailAssociation/osrd/pull/10711/commits/4ab6c34b362561048fffe6a14bfb2ee23b8e4085

clarani (approved):

emersion:
Trivial rebase to bump osrd-ui to v0.0.74.

PR #10716, opened 2025-02-06
editoast: update `modified` field of `infra` upon modification

leovalais:
Part of `editoast-statful` scalability work.

The field used to default to `NOW()` on `INSERT`, but was never updated.

Naive datetimes are also replaced by `DateTime<Utc>`.

### DIFF
--- a/editoast/src/infra_cache/operation/mod.rs
+++ b/editoast/src/infra_cache/operation/mod.rs
@@ -4,6 +4,7 @@ mod update;
 
 use std::ops::Deref as _;
 
+use chrono::Utc;
 use editoast_derive::EditoastError;
 use editoast_schemas::primitives::OSRDObject as _;
 use json_patch::Patch;
@@ -16,6 +17,8 @@ use utoipa::ToSchema;
 pub use self::delete::DeleteOperation;
 use crate::error::Result;
 use crate::infra_cache::ObjectCache;
+use crate::models::prelude::*;
+use crate::models::Infra;
 use editoast_models::DbConnection;
 use editoast_schemas::infra::InfraObject;
 use editoast_schemas::primitives::ObjectRef;
@@ -142,10 +145,10 @@ impl Operation {
         infra_id: i64,
         conn: &mut DbConnection,
     ) -> Result<Option<InfraObject>> {
-        match self {
+        let res = match self {
             Operation::Delete(deletion) => {
                 deletion.apply(infra_id, conn).await?;
-                Ok(None)
+                Ok::<_, crate::error::InternalError>(None)
             }
             Operation::Create(railjson_object) => {
                 create::apply_create_operation(railjson_object, infra_id, conn).await?;
@@ -155,7 +158,12 @@ impl Operation {
                 let railjson_object = update.apply(infra_id, conn).await?;
                 Ok(Some(railjson_object))
             }
-        }
+        }?;
+        Infra::changeset()
+            .modified(Utc::now())
+            .update(conn, infra_id)
+            .await?;
+        Ok(res)
     }
 }
 
@@ -165,7 +173,7 @@ pub fn patch_infra_object(infra_object: &InfraObject, json_patch: &Patch) -> Res
     // (1) transform `RailjsonObject` into `serde_json::Value`,
     // (2) patch the object,
     // (3) transform `serde_json::Value` back into a `RailjsonObject`.
-    // The code below is explicitely typed (even if not needed) to help understand this dance.
+    // The code below is explicitly typed (even if not needed) to help understand this dance.
     let object_type = infra_object.get_type();
     let mut value: serde_json::Value = match &infra_object {
         InfraObject::TrackSection { railjson } => serde_json::to_value(railjson)?,
--- a/editoast/src/models/infra.rs
+++ b/editoast/src/models/infra.rs
@@ -8,7 +8,7 @@ mod voltage;
 
 use std::ops::DerefMut;
 
-use chrono::NaiveDateTime;
+use chrono::DateTime;
 use chrono::Utc;
 use derivative::Derivative;
 use diesel::delete;
@@ -69,9 +69,9 @@ pub struct Infra {
     #[schema(required)]
     pub generated_version: Option<String>,
     pub locked: bool,
-    pub created: NaiveDateTime,
-    #[derivative(Default(value = "Utc::now().naive_utc()"))]
-    pub modified: NaiveDateTime,
+    pub created: DateTime<Utc>,
+    #[derivative(Default(value = "Utc::now()"))]
+    pub modified: DateTime<Utc>,
 }
 
 impl InfraChangeset {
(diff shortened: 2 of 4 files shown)


editoast/src/views/infra/edition.rs
```diff
@@ -1027,6 +1027,41 @@ pub mod tests {
...
+                    [PatchOperation::Replace(ReplaceOperation {
+                        path: "/length".to_string().parse().unwrap(),
+                        value: json!(1234),
+                    })]
+                    .to_vec(),
+                ),
+            }),
+        ]
+        .to_vec();
+        apply_edit(conn, &mut small_infra, &operations, &mut infra_cache)
```
hamz2a:
```suggestion
        apply_edit(&mut db_pool.get_ok(), &mut small_infra, &operations, &mut infra_cache)
```

editoast/src/views/infra/edition.rs
```diff
@@ -1027,6 +1027,41 @@ pub mod tests {
         assert_eq!(errors_without_routes.len() - init_errors.len(), 0);
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn infra_edition_updates_modification_date() {
+        let app = TestAppBuilder::default_app();
+        let db_pool = app.db_pool();
+        let conn = &mut db_pool.get().await.unwrap();
```
hamz2a:
```suggestion
```

hamz2a (approved):
Thanks for the PR! I left a small suggestion for a non-blocking change.

editoast/src/models/infra.rs
```diff
@@ -126,8 +127,8 @@ impl Infra {
             let cloned_infra = <Self as Clone>::clone(self)
                 .into_changeset()
                 .name(new_name)
-                .created(Utc::now().naive_utc())
-                .modified(Utc::now().naive_utc())
+                .created(Utc::now())
+                .modified(Utc::now())
```
woshilapin: nitpick: there is non-zero chance that these 2 values are different and we probably don't want that. How about creating a variable first?

editoast/src/models/infra.rs
```diff
@@ -386,6 +389,20 @@ pub mod tests {
         assert_eq!(res.unwrap_err().get_type(), expected_error.get_type());
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
```
woshilapin, line 393: Why do we need that? Maybe a comment might help understand?

editoast/src/infra_cache/operation/mod.rs
```diff
@@ -155,7 +158,12 @@ impl Operation {
                 let railjson_object = update.apply(infra_id, conn).await?;
                 Ok(Some(railjson_object))
             }
-        }
+        }?;
+        Infra::changeset()
+            .modified(Utc::now())
```
woshilapin, line 163: That's sad that Postgres doesn't support this ([MySQL seems to support it](https://stackoverflow.com/a/50604097)). There is way to make all of that on Postgres's side but [it implies triggers and functions](https://x-team.com/magazine/automatic-timestamps-with-postgresql). What do you think?

woshilapin (commented):
How about moving that logic to the DB so we don't forget to put it in every places in the code that needs it? I usually don't like putting logic in the DB, but in this case, it looks almost like a missing feature of Postgres (MySQL seems to have syntactic sugar for it).

editoast/src/views/infra/edition.rs
leovalais: 67b2180ad

editoast/src/views/infra/edition.rs
leovalais: 67b2180ad

editoast/src/models/infra.rs
leovalais: ~~67b2180ad~~

~~Removed these lines as we can use the default value Postgres sets on insertion.~~

Nope 🤦 done as you asked 23c96c1ee

editoast/src/infra_cache/operation/mod.rs
leovalais: MySQL's feature (albeit very nice 😄) wouldn't help us here because we want to update `modified` when infra objects are changed as well.

I thought about using triggers for that, but quickly changed my mind for several reasons.

First, "every places in the codes that need it" is really two places atm. (The update in `bump_version` is redundant: I kept it because it makes sense in the scope of the function.) I understand that this approach isn't really future-proof, but neither are triggers tbh (that's why we need incremental computation!).

I'm against triggers because:

1. **Volume**: we need `INSERT`/`UPDATE` triggers for each infra object table, and an `UPDATE` trigger for the `infra` table itself: that's 27 triggers! (We may be able to share some trigger functions, but I wouldn't bet on it.)
2. **Performance**: talking from experience with the search engine, triggers are very costly for infra objects since there are many of them. That's why we have such a [convoluted implementation of `Infra::clone`](https://github.com/OpenRailAssociation/osrd/blob/14fa21bf67cf757bd42881dbf5459adab5281a8c/editoast/src/models/infra.rs#L123-L217). (The same goes for `fast_delete_static` below.)
3. **`FOR EACH STATEMENT` limitations**: we could setup our triggers so that they run once `FOR EACH STATEMENT` instead of `FOR EACH ROW` to help with performance. But then we loose the `NEW` and `OLD` magic variables cf. [doc](https://www.postgresql.org/docs/current/plpgsql-trigger.html#PLPGSQL-DML-TRIGGER). Then we'd need to setup another mechanism to find the `infra` IDs to update.
4. Every opinion I heard about triggers is that while they seem like a good idea they always come back to bite you. We experienced this ourselves with the search engine (so much that we want to rewrite it for that reason! 🙃).
5. Not much of an argument I agree, but that goes against the paradigm we slowly try to establish of limiting the amount of business logic in the DB.

editoast/src/models/infra.rs
leovalais: `#[tokio::test]` just sets up an executor with no worker threads. But it panics (and rightfully so) when we're trying to `block_on` on the executor's thread. We do that for example with `db_pool.get_ok()`. So we need one additional worker thread that we can block.

I think we can replace almost every occurrence of `#[rstest]` with this declaration to get rid of `async-std`. But then I'm not sure where to put the comment explaining that?

editoast/src/infra_cache/operation/mod.rs
woshilapin: Oh, thanks for the explanation. Indeed, I didn't realize we want to change the infra table when other objects in other tables are changed. So I also do prefer your solution in this case.

editoast/src/models/infra.rs
woshilapin: ok, I think I understand. I'd say that if it starts to be everywhere, then it starts to be the norm in our tests, and I'd feel it's ok for making it the charge of the developer to understand what is behind it. Anyway, I don't have a good proposition and getting rid of `async-std` feels like a better thing to focus on 😄

woshilapin (approved):

PR #10718, opened 2025-02-07
front: don't use front/tests/ to write PDF files

### LINKED ISSUE #10492: front: adapt playwright docker script to new stdcm tests
### Description and goal

in the last STDCM tests, when the simulation sheet should be downloaded, we have a permission error

### Acceptance criteria

the tests should be fully available with the script

emersion:
Closes #10492 
front/tests/ is not a directory meant to be used for test results. It should only contain source files.

Playwright sets up a specific directory for test outputs, so we can just use that instead. As a bonus, the PDF is now included in CI artifacts and can be inspected manually if needed.

Fixes permission errors when running tests in a container for me.

### DIFF
--- a/front/tests/013-stdcm-simulation-sheet.spec.ts
+++ b/front/tests/013-stdcm-simulation-sheet.spec.ts
@@ -48,8 +48,10 @@ test.describe('Verify stdcm simulation page', () => {
     await waitForInfraStateToBeCached(infra.id);
   });
 
+  let downloadDir: string | undefined;
+
   /** *************** Test 1 **************** */
-  test('Verify STDCM stops and simulation sheet', async ({ browserName, context }) => {
+  test('Verify STDCM stops and simulation sheet', async ({ browserName, context }, testInfo) => {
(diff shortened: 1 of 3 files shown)


front/tests/013-stdcm-simulation-sheet.spec.ts
```diff
@@ -77,7 +79,8 @@ test.describe('Verify stdcm simulation page', () => {
     await stdcmPage.displayAllOperationalPoints();
     await stdcmPage.verifyTableData('./tests/assets/stdcm/stdcmWithAllVia.json');
     await stdcmPage.retainSimulation();
-    await stdcmPage.downloadSimulation(browserName);
+    downloadDir = testInfo.outputDir;
+    await stdcmPage.downloadSimulation(testInfo.outputDir);
```
Maymanaf: Use `downloadDir` directly

Maymanaf (approved):
Good job ! LGTM 
just left a signle comment you can resolve it later on

PR #10721, opened 2025-02-07
dev: rename front to front-dev in osrd-compose

ghost:
In order to avoid confusions on commands like `osrd-compose front up front`.

After those changes it would be `osrd-compose dev-front up front`


flomonster (approved):
LGTM (not tested)

SharglutDev (approved):
Nice thank you

PR #10724, opened 2025-02-07
osrd_schemas: update to 'poetry:2.0' and use 'ruff' for linting

woshilapin:
2 commits because bumping to Python 3.12 was problematic with some of the linting tools. So first move to `ruff`, then move to `pyright` instead of `pytype` because `pytype` doesn't work on Python 3.12., then move all the project to the future (newer Python's version and modern `pyproject.toml`).


python/osrd_schemas/pyproject.toml
```diff
@@ -1,40 +1,14 @@
...
-version = "0.8.16"
 description = ""
-authors = ["OSRD <contact@osrd.fr>"]
-
-[tool.poetry.dependencies]
-geojson-pydantic = "^1.0.0"
-pydantic = "^2.1.1"
-python = ">=3.9,<3.12"
+authors = [{ name = "OSRD Contributors", email = "contact@osrd.fr" }]
+requires-python = "~= 3.12"
```
shenriotpro: Should we keep supporting 3.11?

woshilapin: Good point, made a change to relate to this [other PR](https://github.com/OpenRailAssociation/osrd/pull/10554).

.github/workflows/build.yml
```diff
@@ -379,37 +379,25 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Install poetry
-        run: pipx install 'poetry<2.0'
+        run: pipx install poetry
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
+          python-version: "3.12"
```
shenriotpro: nit: since our team will likely use 3.12, I'd like to keep 3.11 in the CI so we avoid silent breaking changes

woshilapin: Seems reasonnable. Done.

python/osrd_schemas/.python-version
woshilapin, line 1: For libraries, it doesn't make sense to impose a version, and `pyproject.toml` already defines which versions of Python are supposed to be supported through `requires-python`.

woshilapin, line 1: For libraries, it doesn't make sense to impose a version, and `pyproject.toml` already defines which versions of Python are supposed to be supported through `requires-python`.

python/railjson_generator/.python-version
```diff
@@ -1 +1,6 @@
+<<<<<<< Conflict 1 of 1
++++++++ Contents of side #1
 3.9
+%%%%%%% Changes from base to side #2
+-3.12.7
+>>>>>>> Conflict 1 of 1 ends
```
Khoyo: I think that's a mistake :D (Also, the current python version is 3.13.2, not 3.12.7)

python/osrd_schemas/pyproject.toml
```diff
@@ -1,22 +1,11 @@
...
-version = "0.8.16"
 description = ""
-authors = ["OSRD <contact@osrd.fr>"]
-
-[tool.poetry.dependencies]
-geojson-pydantic = "^1.0.0"
-pydantic = "^2.1.1"
-python = ">=3.9,<3.12"
+authors = [{ name = "OSRD Contributors", email = "contact@osrd.fr" }]
+requires-python = ">= 3.11, < 3.13"
```
Khoyo: Can we remove the upper bound actually? Poetry didn't like it in `tool.poetry.python` but it is allowed by the `project.requires-python` spec. (Especially considering 3.13 is in fact the current version)

woshilapin: We can. Done in https://github.com/OpenRailAssociation/osrd/compare/550500abb5a38eb4924ed964895fbf47db65d0fa..adf13791c00c1115ecdc09d8f9b2e641998be156.

python/railjson_generator/.python-version
woshilapin: I think I begin to not like so much `jj` for some use cases 😱

I actually removed `.python-version`. For libraries, it doesn't makes sense to much since we already specify a range of supported versions in `pyproject.toml`. Also, maybe we should let this file outside of the repository so developers can set it up the way they want (for example, if you use `pyenv virtualenvs`, you may actually have a label in it, not a version).

flomonster (approved):

Khoyo (approved):

PR #10727, opened 2025-02-07
editoast: cli: print infra object errors on import

Castavo:
Useful to know the earliest whether your infra is broken


woshilapin (approved):
Looks reasonable 😄 Thank you.

PR #10728, opened 2025-02-07
editoast: sync infra caches when out of date

leovalais:
Works, but since the France infra takes about 7 seconds to load,
this commit isn't really working UX-wise.

Signed-off-by: Leo Valais <leo.valais97@gmail.com>


leovalais:
We changed our approach.

PR #10729, opened 2025-02-07
front: fix some trigrams are not recognized

### LINKED ISSUE #10668: Some trigrams aren't recognized
### What happened?

In stdcm, when typing some existing trigrams the search endpoint doesn't return the corresponding operational point in the list (even though it works in operational studies with the rocket search).

Exemple of trigrams : 
- LE (Lille Flandres)
- DE (Dieppe)
- RN (Rouen rive droite)

All trigrams should be recognized in stdcm

1. Go to stdcm
2. Type one of these trigrams in the CI input

Recette (SNCF)

Firefox

.

clarani:
closes #10668 

Since the search endpoint response is paginated, when the user searches a short string, the search endpoint will return some of the operational points which match either the trigram or the name. As the search endpoint response is not ordered by trigram first, we could miss some responses, and especially the operational points matching by trigram.

Now, if the user searches a short string, we will first get all the operational points matching the trigram and then the ones matching the name. This way, we won't miss the reponse by trigram anymore.

### DIFF
--- a/front/src/common/Map/Search/useSearchOperationalPoint.tsx
+++ b/front/src/common/Map/Search/useSearchOperationalPoint.tsx
@@ -4,7 +4,11 @@ import { isEmpty } from 'lodash';
 import { useSelector } from 'react-redux';
 
 import DPY_TO_MAS_OPERATIONAL_POINTS from 'assets/operationStudies/DPYToMASOperationalPoints';
-import { type SearchResultItemOperationalPoint, osrdEditoastApi } from 'common/api/osrdEditoastApi';
+import {
+  type SearchQuery,
+  type SearchResultItemOperationalPoint,
+  osrdEditoastApi,
+} from 'common/api/osrdEditoastApi';
 import { useInfraID } from 'common/osrdContext';
 import { setFailure } from 'reducers/main';
 import { getIsSuperUser } from 'reducers/user/userSelectors';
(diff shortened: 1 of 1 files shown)


flomonster (approved):
LGTM

theocrsb (approved):
LGTM and tested

front/src/common/Map/Search/useSearchOperationalPoint.tsx
```diff
@@ -36,46 +44,64 @@ export default function useSearchOperationalPoint({
...
+              ['search', ['name'], debouncedSearchTerm],
+              ['like', ['to_string', ['uic']], `%${debouncedSearchTerm}%`],
+            ],
+            ['=', ['infra_id'], infraID],
+            dpyToMasOperationalpointsFilter,
+          ],
+        },
+        pageSize: 101,
+      }).unwrap()) as SearchResultItemOperationalPoint[];
+      setSearchResults([...trigramResults, ...results]);
```
emersion, line 104: Do we need to de-duplicate results here? e.g. if both search queries return the same OP, it would appear twice in the list?

SharglutDev (approved):
Lgtm and tested, thank you !

front/src/common/Map/Search/useSearchOperationalPoint.tsx
clarani: We have a post-processing functions that filter the duplicates :) So I think it's ok like this

Issue #5254, opened 2023-10-09
Make osrd-schemas extension type visible 


Castavo:
### Who would benefit from this feature?

Both

### What is this feature about?

The dynamically created extension types are not easily visible from outside the package.
To instantiate an extension object, we need to do:

```
op_extension = OperationalPointPart.model_fields["extensions"].annotation(
    sncf=OperationalPointPartSncfExtension(kp=location["pk"])
)
```

If we could do something like below, or through a method, it would be cleaner.
```
op_extension = OperationalPointParExtensions(
    sncf=OperationalPointPartSncfExtension(kp=location["pk"])
)
```

### Why is this feature valuable?

It will make using osrd-schemas easier, and it will be less bug-prone

PR #10730, opened 2025-02-07
railjson_generator: update to 'poetry:2.0' and use 'ruff' for linting

woshilapin:
Needs #10724 first since `osrd_schemas` needs to support a recent version of Python first.

3 commits:
- first replace `pyflake`, `isort` and `black` with `ruff` (all modified Python's file are only reformatting)
- second replace `pytype` with `pyright` (needed a bunch of corrections, might need an attentive look since I'm not the best Python developer)
- third make it compatible for `poetry:2.0` and a modern version of Python


python/railjson_generator/.python-version
woshilapin, line 1: For libraries, it doesn't make sense to impose a version, and `pyproject.toml` already defines which versions of Python are supposed to be supported through `requires-python`.

python/railjson_generator/pyproject.toml
```diff
@@ -1,41 +1,12 @@
...
-exclude = [".venv/"]
-
-[tool.black]
-line-length = 120
-
-[tool.isort]
-profile = "black"
-multi_line_output = 3
+authors = [{ name = "OSRD Contributors", email = "contact@osrd.fr" }]
+requires-python = ">= 3.11, < 3.13"
```
Khoyo: Can we remove the max version? (for the same reasons as schemas)

Khoyo (approved):
Otherwise LGTM :)

leovalais (approved):

python/railjson_generator/pyproject.toml
woshilapin: Done.

.github/workflows/build.yml
```diff
@@ -250,26 +250,11 @@ jobs:
...
-          cd python/railjson_generator
-          poetry run black . --check
-
-      - name: Isort
+      - name: Ruff
         run: |
           cd python/railjson_generator
-          poetry run isort . --check
+          poetry run ruff check
+          poetry run ruff format --check
```
Castavo: Using the following allows for ruff lints to be included in the PR files. I like it, maybe it's not consensual ?

```yml
        run: |
          poetry run ruff check --output-format github
```

python/railjson_generator/pyproject.toml
```diff
@@ -1,41 +1,12 @@
...
-[tool.pytype]
-inputs = ["railjson_generator"]
+dependencies = ["osrd-schemas @ ../osrd_schemas"]
 
-[build-system]
-requires = ["poetry-core>=1.0.0"]
-build-backend = "poetry.core.masonry.api"
+[project.optional-dependencies]
+check = ["pytest ~= 7.4.3", "pytest-cov ~= 4.1.0"]
+dev = ["pyright ~= 1.1.393", "ruff ~= 0.9.5"]
```
Castavo: Aren't `pyright` and `ruff` also checking dependencies ?
```suggestion
check = ["pyright ~= 1.1.393", "pytest ~= 7.4.3", "pytest-cov ~= 4.1.0", "ruff ~= 0.9.5"]
```

python/railjson_generator/pyproject.toml
```diff
@@ -1,41 +1,12 @@
...
-pyproject-flake8 = "^6.0.0.post1"
-pytype = { platform = "linux", version = "^2023.10.17" }
-
-[tool.flake8]
-ignore = "W503,E203"
-max-line-length = 120
-exclude = [".venv/"]
-
-[tool.black]
-line-length = 120
```
Castavo, line 30: Why not keep the line length of 120 ? It seems to be the norm throughout our repos (and it's more pleasant to use in my opinion)

Castavo (approved):
Looks neat !

We should address the weird things we do with extensions one day. We have #5254 on somewhat the same subject.

python/railjson_generator/pyproject.toml
woshilapin: In my mind, one is for tests, the other is for development tools. But I'm fine putting all of them in the same bag and agree the line is thin. I just followed a convention that was already here in other python projects. I merged them into the `check` category in https://github.com/OpenRailAssociation/osrd/pull/10730/commits/e83e5aba8a8e243774dd47c966d147393e15e927 and I also added https://github.com/OpenRailAssociation/osrd/pull/10730/commits/5234b796c6fe76e4daef6172ac5f0344ee4d0da5 to follow the same convention in `osrd_schemas`.

python/railjson_generator/pyproject.toml
woshilapin: Because `black` is not there anymore and because I like not customizing the formatter to avoid subjective debat 🤷 But if I'm the only one, [it can be put back in `ruff`'s configuration](https://docs.astral.sh/ruff/settings/#line-length).

.github/workflows/build.yml
woshilapin: Didn't know about it and kind of nice. Unless someone else complains, Done in [f47f109f5a938ff5b9cc4e52cba43ad9d0b65fcf](https://github.com/OpenRailAssociation/osrd/compare/7fab4f32c5b6c6c33700fa1abb0495f754fa9fbb..f47f109f5a938ff5b9cc4e52cba43ad9d0b65fcf).

python/railjson_generator/pyproject.toml
Castavo: Ah, I would tend to use the group `dev` for dependencies a dev would use but not the ci 
In this instance there aren't any but for example `jupyter` would fit in `dev` more than in `checks`

python/railjson_generator/pyproject.toml
Castavo: The others reviewers didn't seem to care so it seems it's me that's lonely
Feel free to resolve this conversation

Khoyo: I think we can keep it at `120`, if only because it was the config already and thus avoid diff (and blame) pollution - and also keeping the status quo causes less debates

(But I actually prefer 88, so feel free to resolve either way :D)

woshilapin: OK, I feel 🆓

PR #10732, opened 2025-02-08
dev: resilient gateway build for dev front

ghost:
When building the Gateway in development, if the front is broken we should not fail since it can happens a lot for front-end developers.

Should be tested by front-end dev to see if it fits their development workflows.

### DIFF
--- a/gateway/Dockerfile
+++ b/gateway/Dockerfile
@@ -63,18 +63,28 @@ CMD ["/usr/local/bin/osrd_gateway"]
 FROM node:23-bookworm AS front_build
 WORKDIR /app
 
-# Build dependencies
-COPY --from=front_src package.json package-lock.json /app/
-RUN npm install --frozen-lockfile
+# Add build argument for skipping failed front
+ARG SKIP_FRONT=0
 
 # Fill build time values
 ARG OSRD_GIT_DESCRIBE
 ENV VITE_OSRD_GIT_DESCRIBE=${OSRD_GIT_DESCRIBE}
 
+# Build dependencies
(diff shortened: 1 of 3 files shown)


emersion:
Hm, maybe we can just completely skip the front build in that case, instead of trying it and ignoring errors?

ghost:
> Hm, maybe we can just completely skip the front build in that case, instead of trying it and ignoring errors?

Guess we could, no strong opinion on that. I'll let front developers go one way or another since I will not be available for this (feel free to modify the branch as you please).

gateway/Dockerfile
```diff
@@ -74,7 +77,9 @@ ENV VITE_OSRD_GIT_DESCRIBE=${OSRD_GIT_DESCRIBE}
 # Generate the licenses file and build
 COPY --from=front_src . /app
 ENV NODE_OPTIONS="--max_old_space_size=4096"
-RUN npm run generate-licenses && npm run build
+RUN (npm run generate-licenses && npm run build) || \
+    (test "$SKIP_BAD_FRONT" = "1" && mkdir -p build && \
+    echo '<html><body><h1>Front build failed : check logs.</h1></body></html>' > build/index.html)
```
Khoyo: nit
```suggestion
     echo '<html><body><h1>Front build failed : check logs.</h1></body></html>' > build/index.html)
```

Khoyo (approved):
lgtm

ghost:
@emersion I've followed what you've suggested, I will merge it so there will be less issues for developers going forward.

PR #10735, opened 2025-02-09
front: add linting to e2e tests and scripts

Synar:
- Include front/tests and front/scripts in local and ci eslint commands
- Fix preexisting warnings

The cost of extending the linting scope should be minimal.

Warnings related to the import of i18next-parser and its types in one of the script on could not be solved. [A related issue was already open on their repository](https://github.com/i18next/i18next-parser/issues/1000)


Synar:
And now the ci properly fails on import/no-extraneous-dependencies because we don't include tests and scripts in the exceptions, as I thought it would last time.

Edit : fixed, but at least we know now that the check works fine ^^

Maymanaf (approved):

front/.eslintrc
```diff
@@ -78,7 +78,7 @@
     "import/no-extraneous-dependencies": [
       "error",
       {
-        "devDependencies": ["**/*.spec.ts", "**/__tests__/**"]
+        "devDependencies": ["**/*.spec.ts", "**/__tests__/**", "**/tests/**"]
```
emersion: I think `tests/**` should be enough here: we only want to allow dev deps in the top-level tests directory. (Same for scripts.)

front/scripts/i18n-api-errors.ts
```diff
@@ -3,16 +3,28 @@ import i18next from 'i18next';
 import { noop } from 'lodash';
 import type { OpenAPI } from 'openapi-types';
 
+import i18nData from '../public/locales/fr/errors.json' assert { type: 'json' };
```
emersion: `assert` has been deprecated and removed in newer Node versions. I'd prefer to not use imports to load data files: https://github.com/OpenRailAssociation/osrd/pull/10671/commits/d11d1ffa0f4d7dd628ea952bbb1d407a0aa60d5f

front/scripts/i18n-api-errors.ts
```diff
@@ -3,16 +3,28 @@ import i18next from 'i18next';
...
+
 // relative the project's root
 const openapi_path = '../editoast/openapi.yaml';
 // relative to this file
 const i18n_error_path = '../public/locales/fr/errors.json';
 
+type OpenAPIErrorSchema = {
```
emersion: Instead of defining our own type here, can't we use something from `openapi-types`? Would be better to runtime assert that the OpenAPI data conforms to the shape we assume, rather than cast.

emersion (commented):
Very nice!

Could we also set the ESLint command to `eslint --cache --ext .ts,.tsx,.js,.jsx .` (lint the whole current directory), instead of listing directories? We can supply files to ignore in `.eslintrc` instead of replicating these across all command invocations. (Replicating was okay before because it was a single directory, but now it's a list that we need to keep in sync.)

Synar:
> Very nice!
> 
> Could we also set the ESLint command to `eslint --cache --ext .ts,.tsx,.js,.jsx .` (lint the whole current directory), instead of listing directories? We can supply files to ignore in `.eslintrc` instead of replicating these across all command invocations. (Replicating was okay before because it was a single directory, but now it's a list that we need to keep in sync.)

Sure, good idea!
I suppose we mostly want to exclude build, node_modules and playwright_report.
I'll need to fix warnings for a couple more files though ^^

front/.eslintrc
Synar: Sure! I was afraid of things failing when mounted to docker in the ci, but I guess I'll at least check.

emersion:
`node_modules` is already ignored by default: https://eslint.org/docs/latest/use/configure/ignore#ignoring-files

If you don't want to fix everything, we can leave more stuff in the ignore list and shrink it later: I don't want to make you responsible for doing all of this work :P

front/scripts/i18n-api-errors.ts
Synar: Mb! I also did not realize you had a conflicting pr on this, do you want me to put this one on hold? Or perhaps I should just cherrypick your commit? ^^

Synar:
> `node_modules` is already ignored by default: https://eslint.org/docs/latest/use/configure/ignore#ignoring-files

That seems logical ^^

> If you don't want to fix everything, we can leave more stuff in the ignore list and shrink it later: I don't want to make you responsible for doing all of this work :P

Don't worry, there's barely anything more =p

I'm unsure about fixing the "Prefer default export on a file with single export.eslint" rule though on vitest.global-setup.ts, guess we'll see if the tests catch fire ^^ More generally though this is by far the most ignored linting rule in the project, I think it would be more coherent if we either removed all the exceptions, or removed the rule globally.

front/scripts/i18n-api-errors.ts
Synar: Done, but in quite an ugly way. I'll guess you'll tell me if you're fine with that =p

front/.eslintrc
Synar: Works fine, thanks for the catch!

front/scripts/i18n-api-errors.ts
Synar: I ended up dastardly borrowing that line, hope you'll forgive me (but I did not use the url part, I'm not sure why we would need it here)

emersion: https://github.com/OpenRailAssociation/osrd/pull/10671 has been approved by Paul, so I think it can be cherry-picked.

> but I did not use the url part, I'm not sure why we would need it here

We need it to resolve paths relative to the script, rather than the current directory.

emersion: https://github.com/OpenRailAssociation/osrd/pull/10671 has been merged! You should be able to rebase.

Synar: Thanks for the explanation, done.

However, rebasing creates an absolute mess of typescript errors, since all the files are now imported dynamically. ^^'
As well as a cyclic dep
I'm going to have fun =p

Synar:
It would be nice to set the import/extensions eslint rule correctly for front/tests, but since activating it gave me 180 statements to modify and this pr is already long enough, this will be a task for next time. ^^

front/tests/011-op-times-and-stops-tab.spec.ts
```diff
@@ -9,16 +9,29 @@ import OperationalStudiesOutputTablePage from './pages/op-output-table-page-mode
...
 import { cleanWhitespace, cleanWhitespaceInArray } from './utils/dataNormalizer';
+import readJsonFile from './utils/file-utils';
 import createScenario from './utils/scenario';
 import scrollContainer from './utils/scrollHelper';
 import { deleteScenario } from './utils/teardown-utils';
 import type { StationData } from './utils/types';
 
-const enTranslations = readJsonFile('public/locales/en/timesStops.json');
-const frTranslations = readJsonFile('public/locales/fr/timesStops.json');
+type timeStopsTranslation = {
```
emersion: Nit: first letter should be upper-case.

Hm, it's a bit unfortunate to have to define types for each translation file, but I don't see a good way around it...

front/tests/pages/op-route-page-model.ts
```diff
@@ -1,11 +1,18 @@
 import { expect, type Locator, type Page } from '@playwright/test';
 
-import { getTranslations, readJsonFile } from '../utils';
+import { getTranslations } from '../utils';
+import readJsonFile from '../utils/file-utils';
 
-const enTranslations = readJsonFile(
+type manageTrainScheduleTranslations = {
```
emersion: Ditto

front/tests/011-op-times-and-stops-tab.spec.ts
Synar: An alternative would be to force any types I suppose. The other alternative I can see would be to go back to importing statically in a way ts can infer the type.

But at least this way we can clearly see which keys we're checking I suppose?

(Good catch for the casing, will fix ^^)

front/scripts/i18n-api-errors.ts
```diff
@@ -11,10 +12,42 @@ const i18n_error_path = '../public/locales/fr/errors.json';
...
  * Check if the given error is well filled in the i18n.
  */
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-async function checkI18N(error: any): Promise<string[]> {
+async function checkI18N(
+  error:
+    | OpenAPIV3_1.SchemaObject
+    | OpenAPIV3.ReferenceObject
+    | OpenAPIV3.SchemaObject
+    | OpenAPIV3_1.ReferenceObject
```
emersion, line 20: Ew, it's painful to handle all of these cases… I don't see a better way though!

emersion (approved):
LGTM once the two nits are fixed!

PR #10736, opened 2025-02-09
front: fix prettier formatting of jsonc

Synar:
- Fix prettier adding trailing commas to jsonc, which is incorrect for jsonc specs
- Thus formats .eslintrc with prettier, without any commas addition

See the [related prettier issue](https://github.com/prettier/prettier/issues/15956)


emersion (approved):
LGTM!

PR #10738, opened 2025-02-10
core: add error on zero length simulations

eckter:
We sometimes have 500 errors in prod because of this

### DIFF
--- a/core/osrd-reporting/src/main/java/fr/sncf/osrd/reporting/exceptions/ErrorType.java
+++ b/core/osrd-reporting/src/main/java/fr/sncf/osrd/reporting/exceptions/ErrorType.java
@@ -171,6 +171,7 @@ public enum ErrorType {
             "missing_route_from_chunk_path",
             "couldn't find a route matching the given chunk list",
             ErrorCause.INTERNAL),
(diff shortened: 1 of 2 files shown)


core/osrd-reporting/src/main/java/fr/sncf/osrd/reporting/exceptions/ErrorType.java
```diff
@@ -171,6 +171,7 @@ public enum ErrorType {
             "missing_route_from_chunk_path",
             "couldn't find a route matching the given chunk list",
             ErrorCause.INTERNAL),
+    ZeroLengthPath("zero_length_path", "can't simulate an zero length path", ErrorCause.USER),
```
Erashin:
```suggestion
    ZeroLengthPath("zero_length_path", "can't simulate a zero length path", ErrorCause.USER),
```

Erashin (approved):

emersion:
Weird that we're still seeing this, editoast should avoid sending such requests now: https://github.com/OpenRailAssociation/osrd/pull/9931

I suppose more safety is always a good thing :P

eckter:
Yeah we should investigate further, I'll create a different issue: https://github.com/OpenRailAssociation/osrd/issues/10762

But when it does happen, an exception is better than an assert error very far down the call stack

PR #10740, opened 2025-02-10
front: stdcm - show consist tooltip on focus out only

### LINKED ISSUE #9374: Handle invalid values in stdcm inputs

SharglutDev:
close #9374 

fix this comment in the issue : https://github.com/OpenRailAssociation/osrd/issues/9374#issuecomment-2639980286

- Update ConsistError type to be able to display them or not depending on
some conditions
- Transform consistErrors useMemo into a state and move it in StdcmConsist component

To test the PR : 
- the tooltip error should not pop unless the input is focus out if the value is invalid (doesn't match the range)
- the tooltip should hide if the value becomes correct but not show again if it was corrected and then invalid (unless focus out)

### DIFF
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
@@ -5,25 +5,15 @@ import cx from 'classnames';
 import { useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
 
-import useStdcmTowedRollingStock from 'applications/stdcm/hooks/useStdcmTowedRollingStock';
 import { extractMarkersInfo } from 'applications/stdcm/utils';
-import {
-  validateMaxSpeed,
-  validateTotalLength,
-  validateTotalMass,
-} from 'applications/stdcm/utils/consistValidation';
 import DefaultBaseMap from 'common/Map/DefaultBaseMap';
 import { useOsrdConfSelectors } from 'common/osrdContext';
 import useInfraStatus from 'modules/pathfinding/hooks/useInfraStatus';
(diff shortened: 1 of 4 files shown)


clarani (approved):
LGTM :white_check_mark:

Akctarus (approved):
thank you for the test instructions, LGTM !

PR #10766, opened 2025-02-11
front: fix rounded hours

### LINKED ISSUE #9179: Departure and arrival times are rounded down in the train list
### What happened?

All those trains are arriving at 11:55:59 or 11:56:00, but the times are rounded down:
image

image

I'm expecting the times (departure too) to be rounded.

10:00:01 -> 10:00
11:55:59 -> 11:56
22:40:30 -> 22:40
22:40:31 -> 22:41

1. Create a train making it departure at 10:00:01
2. Try 10:00:59

Recette (SNCF)

Google Chrome

1482e8c

Akctarus:
closes #9179

### DIFF
--- a/front/src/modules/trainschedule/components/Timetable/TrainScheduleItem.tsx
+++ b/front/src/modules/trainschedule/components/Timetable/TrainScheduleItem.tsx
@@ -44,7 +44,10 @@ type TrainScheduleItemProps = {
 };
 
 const formatFullDate = (d: Date) => dayjs(d).format('D/MM/YYYY HH:mm:ss');
-const formatDateHours = (d: Date) => dayjs(d).format('HH:mm');
+const formatDateHours = (d: Date) =>
+  dayjs(d)
+    .add(d.getSeconds() >= 30 ? 1 : 0, 'minute')
+    .format('HH:mm');
 
 const TrainScheduleItem = ({
   isInSelection,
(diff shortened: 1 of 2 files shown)


emersion (commented):
Weird, I think dayjs is already expected to handle this case?

https://day.js.org/docs/en/customization/relative-time#relative-time-thresholds-and-rounding

Akctarus:
> Weird, I think dayjs is already expected to handle this case?
> 
> https://day.js.org/docs/en/customization/relative-time#relative-time-thresholds-and-rounding

dayjs has built-in rounding behaviors, but by default, it does not round up minutes when seconds are >=30

emersion:
Right, but it can be configured to do so, right? (See above link)

Akctarus:
> Right, but it can be configured to do so, right? (See above link)

you mean something like that ? https://github.com/OpenRailAssociation/osrd/pull/10766/commits/fc1ebcd8fd570fc5793ddf4d8575e69121ddf9e5

emersion:
Ew, my bad, it seems like the options I've linked to only apply to the relative time plugin, which we don't use.

Ideally we'd use a rounding function ([just like Temporal has](https://tc39.es/proposal-temporal/docs/instant.html#round)) but it seems like the library is missing that.

Oh well, let's just go with your original solution then, sorry for the noise!

Akctarus:
@emersion Done !

emersion (approved):
LGTM!

Ideally these commits would be squashed together so that the tests don't regress in-between the commits (useful when bisecting).

Maymanaf (approved):

PR #10768, opened 2025-02-11
core: only apply driver behavior to BAL and BAPR

### LINKED ISSUE #10683: core: limit driver behavior to BAL + BAPR
The "driver behavior" we currently have should only be applied to some signaling systems

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/10683


core/envelope-sim/src/main/kotlin/fr/sncf/osrd/DriverBehaviour.kt
```diff
@@ -0,0 +1,51 @@
...
+
+data class DriverBehaviour(
+    val acceleratingPostponementOffset: Double = 50.0,
+    val brakingAnticipationOffset: Double = 100.0,
+    val signalingSystems: List<String> = listOf("BAL", "BAPR")
+) {
+    /** Applies the driver behavior to the MRSP, adding reaction time for MRSP changes */
+    fun applyToMRSP(
+        mrsp: Envelope,
+        optSignalingSystemRanges: DistanceRangeMap<String>? = null
```
Erashin, line 21: You could just :
```suggestion
        optSignalingSystemRanges: DistanceRangeMap<String>? = distanceRangeMapOf()
```
Would avoid having to put null when unneeded.

Erashin (approved):
LGTM!
Btw, this should fail the newly done tests in this pr: https://github.com/OpenRailAssociation/osrd/pull/10769#

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/DriverBehaviour.kt
eckter: This is still called from java with v1, where the optional parameters still have to be called explicitly (and IIRC it's more verbose than just `distanceRangeMapOf()` in java)

Erashin: Oh right, true true. I've gotta drop the v1 endpoints x)

PR #10770, opened 2025-02-11
front: fix itinerary marker not always defined

### LINKED ISSUE #10414: Itinerary marker not always defined
### What happened?

In some cases, vias information is not displayed
Image

This information should be displayed

Go to scenario
Edit one train
Go to itinerary section
See the bug

Local

Google Chrome

b1036ac

theocrsb:
closes #10414


Math-R:
Sorry but the bug still occure here it's a footage

https://github.com/user-attachments/assets/42e20162-400c-4e20-b346-2b95a79cd225

axrolld:
@Math-R i think when there's no pathfinding we should not display the small cards with the tracks, only the pins with numbers 1, 2, 3, etc. @theocrsb agrees with that.

theocrsb:
I just pushed with the modifications

Akctarus (approved):
Tested and lgtm ! I find it much cleaner and easier to understand

Math-R (approved):
Okay till the comportement as been discuss, it looks good to me

PR #10771, opened 2025-02-11
front: remove blueprint map style

Tristramg:
This style existed for a schematic representation that no longer exists.

It added some complexity to handle specific sprites and one sprite was missing (SECT) and nobody ever noticed it.


front/src/common/Map/Layers/InfraObjectLayers/extensions/SNCF/PSLSigns.tsx
```diff
@@ -30,14 +27,10 @@ export function getPSLSignsLayerProps({
...
-        ['==', ['get', 'type'], 'R'],
-        ['concat', prefix, 'R'],
-        ['==', ['get', 'type'], 'Z'],
-        ['concat', prefix, 'Z'],
+        ['concat', 'TIV D FIXE ', ['get', 'value']],
         ['==', ['get', 'type'], 'TIV_B'],
-        ['concat', prefix, 'TIVD B FIX ', ['get', 'value']],
-        'none',
+        ['concat', 'TIVD B FIX ', ['get', 'value']],
+        ['get', 'type'],
```
emersion, line 33: Is this an unrelated change by any chance? Or is there something specific to the blueprint style which changes how signals are displayed?

emersion: Oh, the prefix is dropped. Makes sense, sorry for the noise!

Tristramg: The code before:
```json
['==', ['get', 'type'], 'R'],
        ['concat', prefix, 'R'],
        ['==', ['get', 'type'], 'Z'],
        ['concat', prefix, 'Z']
```

When considering speed limits, there are `TIV D 60`, `TIVD B 60` (the 60 being variable) and R and Z signals.

The `prefix` part is only for the blueprint.

So once the prefix part is removed, in pseudo code it would look like:

```js
if (type =='R') return 'R'
else if (type == 'Z') return 'Z'
else return none
```

So there is arguably a slight change in behavior, but the data comes from the `psl_sign` layer and:

```
select distinct(data->'type') from infra_layer_psl_sign;
 ?column?
----------
 "TIV_C"
 "TIV_B"
 "R"
 "TIV_D"
 "Z"
```

Which raises the question about TIV_C types

Uriel-Sautron (approved):
LGTM

Math-R (approved):
Looks good to me ! thanks for the cleanup :D

PR #10772, opened 2025-02-11
front: add paced trains

### LINKED ISSUE #10607: Add new paced train
### Description and goal

Update the current interface to add train to be able to add a unique train or a paced train.

Ppt mockup here : https://sncf.sharepoint.com/:p:/r/sites/OSRD644GrpO365/Documents%20partages/PO/Groupes%20et%20familles%20de%20trains/Maquette%20mission%20PI14.pptx?d=wfceedf5d80034b0aaed5e9ba4dff9dfd&csf=1&web=1&e=Tbb4nc

> [!NOTE]
> The ppt mockup doesn't provide any specific style, it's more of a functional mockup

This interface should also have end to end tests.

### End to end tests strategy (details here : https://github.com/osrd-project/osrd-tests/issues/89)
- The old test `005-operational-studies.spec.ts` should be replaced by the one in the linked issue
- Every requirements of the tests should be done until the 6. part included

### Acceptance criteria

- [ ] When service mode is off, we can still create a regular train
- [ ] There is a new checkbox below the submit button to switch the form between unique train and paced train
- [ ] The checkbox is unchecked by default
- [ ] In service mode, there is a new field `Duration of the repetition time range` ("Durée de la plage de répétition")
- [ ] In service mode, there is a new field `Cadence of the service` ("Cadence de la mission")
- [ ] The duration field is in minutes and is filled with 120min by default
- [ ] The cadence field is in minutes and is filled with 60min by default
- [ ] Both these fields are saved in the store
- [ ] When the checkbox is checked, the main "Add" button, shows `Add a service` ("Ajouter une mission")
- [ ] New e2e tests have been added and follow the strategy in description

SharglutDev:
close #10607


kmer2016 (approved):
Well done !

front/tests/005-operational-studies.spec.ts
```diff
@@ -1,24 +1,35 @@
...
+  addPacedTrain: string;
+  addTrainSchedule: string;
+  addTrainSchedules: string;
+  pacedTrains: {
+    cadence: string;
+    defineService: string;
+    timeRangeDuration: string;
+  };
+  trainLabels: string;
+};
```
Yohh: needs to be set in `tests/utils/types.ts`

front/src/modules/trainschedule/components/ManageTrainSchedule/PacedTrainSettings.tsx
```diff
@@ -0,0 +1,71 @@
...
+            </>
+          }
+          id="paced-train-time-range-duration"
+          onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
+            dispatch(updateTimeRangeDuration(+e.target.value));
+          }}
+          value={timeRangeDuration}
+          noMargin
+          isInvalid={timeRangeDuration < 1}
+          errorMsg={timeRangeDuration < 1 ? t('errorMessages.noTimeRangeDuration') : undefined}
```
Yohh: the button `add a service` should be disabled if this message appear

same for cadence

front/src/modules/trainschedule/components/ManageTrainSchedule/PacedTrainSettings.tsx
```diff
@@ -0,0 +1,71 @@
...
+const PacedTrainSettings = () => {
+  const timeRangeDuration = useSelector(getTimeRangeDuration);
+  const cadence = useSelector(getCadence);
+  const { t } = useTranslation(['operationalStudies/manageTrainSchedule']);
+  const dispatch = useAppDispatch();
+
+  return (
+    <div className="d-flex px-3 mt-2">
+      <span className="mr-3">
+        <InputSNCF
```
Yohh, line 22: I see it's convient with the mockup, but it seems weird that the icon and the text are not vertically aligned, @thibautsailly ?

Capture d’écran 2025-02-21 à 17 47 25

front/src/modules/trainschedule/components/ManageTrainSchedule/PacedTrainSettings.tsx
```diff
@@ -0,0 +1,71 @@
...
+          errorMsg={cadence < 1 ? t('errorMessages.noCadence') : undefined}
+          min={1}
+          unit="min"
+          textRight
+          sm
+        />
+      </span>
+    </div>
+  );
+};
```
Yohh:
```suggestion
const PacedTrainSettings = () => {
  const pacedTrainOptions = {
    timeRangeDuration: {
      initialValue: useSelector(getTimeRangeDuration),
      label: 'pacedTrains.timeRangeDuration',
      id: 'paced-train-time-range-duration',
      errorMessage: 'errorMessages.noTimeRangeDuration',
      action: updateTimeRangeDuration,
    },
    cadence: {
      initialValue: useSelector(getCadence),
      label: 'pacedTrains.cadence',
      id: 'paced-train-cadence',
      errorMessage: 'errorMessages.noCadence',
      action: updateCadence,
    },
  };

  const { t } = useTranslation(['operationalStudies/manageTrainSchedule']);
  const dispatch = useAppDispatch();

  return (
    <div className="d-flex px-3 mt-2">
      {Object.entries(pacedTrainOptions).map(([key, value]) => (
        <span className="mr-3">
          <InputSNCF
            type="number"
            label={
              <>
                <ArrowBoth />
                <small className="text-nowrap">{t(value.label)}</small>
              </>
            }
            id={value.id}
            onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
              dispatch(value.action(+e.target.value));
            }}
            value={value.initialValue}
            noMargin
            isInvalid={value.initialValue < 1}
            errorMsg={value.initialValue < 1 ? t(value.errorMessage) : undefined}
            min={1}
            unit="min"
            textRight
            sm
          />
        </span>
      ))}
    </div>
  );
};
```

Yohh (commented):
just a proposal

I need more time for e2e tests

front/src/modules/trainschedule/components/ManageTrainSchedule/PacedTrainSettings.tsx
thibautsailly: Thank you for spotting this. I think this is a powerpoint display issue, on my end it looks fine. I trust you to align it correctly :-)

front/tests/005-operational-studies.spec.ts
SharglutDev: I reproduced the way it was done in other tests (001, 003...). But yes, I can move it since it's used in different files.

front/src/modules/trainschedule/components/ManageTrainSchedule/PacedTrainSettings.tsx
SharglutDev: Atm, we never disable the button (expect if no infra) even for "old" settings like name etc so I didn't want to change the behavior. But this could be discussed yes. What do you think @maelysLeratRosso ?

maelysLeratRosso: Yes I think we should remain consistent for this PR and not disable the button, since we don't disable it for other fields. However maybe disabling the button when there is a missing field in the creation/edition window of a train/service could be an enhancement, what to you think @thibautsailly ?

front/src/modules/trainschedule/components/ManageTrainSchedule/PacedTrainSettings.tsx
SharglutDev: Seems like a weird behavior for osrd-ui icons. They are wrapped in a span making the default height of the icon 22px instead of the 16px height of the svg.
Added a little hack for now with margin-bottom to center both icons only here.

I also recentered the checkbox label for "Define a service"

front/src/modules/trainschedule/components/ManageTrainSchedule/PacedTrainSettings.tsx
SharglutDev: Totally debatable but I find it easier to read with the 2 inputs as they are. It makes us win only 6 lines of code (agreed that it's more relevant for a large form like the rolling stock editor)

Yohh: as you wish, it's just a suggestion

front/src/modules/trainschedule/components/ManageTrainSchedule/PacedTrainSettings.tsx
Yohh: I asked for it because the old behavior was to display an error message, for the new behavior I didn't see anyone, clicking on the add button closes the creation view without any information, so basically at this point it's not consistent already

front/tests/005-operational-studies.spec.ts
SharglutDev: Forced pushed to fix conflicts after the merge of this [PR](https://github.com/OpenRailAssociation/osrd/pull/10900).
I re-used the new way of defining translations.

front/src/modules/trainschedule/components/ManageTrainSchedule/PacedTrainSettings.tsx
thibautsailly: It's always better to let the user click a button and provide actionable feedback to get unstuck.

SharglutDev: > I asked for it because the old behavior was to display an error message, for the new behavior I didn't see anyone, clicking on the add button closes the creation view without any information, so basically at this point it's not consistent already

You're right, I'll add a toast message like we do for other fields

SharglutDev: Fixed (good catch !)

front/public/locales/en/operationalStudies/manageTrainSchedule.json
```diff
@@ -25,7 +25,8 @@
     "invalidInitialSpeed": "The initial speed must be rounded to the nearest tenth",
     "mandatoryField": "Required field",
     "nameLengthLimit": "The name must not exceed 128 characters",
-    "noCadence": "Must be > 0",
+    "noCadence": "The cadence must be must be greater than 0",
+    "noCadenceInput": "Must be > 0",
```
Yohh: nit: some translations are very similar or just exactly the same, do you think you could use only one generic translation when possible, a generic translation with a variable for the rest?

SharglutDev: Done for the `Must be > 0`.
For the error message when submitting, done only for cadence and time range duration. 
For the legacy inputs (count, delta, step that will soon be removed), as their translations are different between the error message and the label one, it would have been too much trouble I think (would have required to add some new translations only for this case).

front/tests/005-operational-studies.spec.ts
```diff
@@ -50,68 +53,52 @@ test.describe('Verify simulation configuration in operational studies', () => {
...
-  test('Pathfinding with rolling stock and composition code', async ({ page }) => {
-    // Page models
-
+  test('Add a paced train', async ({ page }) => {
     // Navigate to the scenario page for the given project and study
     await page.goto(
       `/operational-studies/projects/${project.id}/studies/${study.id}/scenarios/${scenario.id}`
     );
 
+    await operationalStudiesPage.checkPacedTrainSwitch();
```
Yohh, line 74: as the old behavior still exists, we should keep the actual tests, and add the new test for paced train

front/tests/pages/operational-studies-page-model.ts
```diff
@@ -121,19 +151,113 @@ class OperationalStudiesPage extends CommonPage {
...
+  async checkPacedTrainSwitch() {
+    await expect(this.operationStudiesSettings).toBeVisible();
+    await this.operationStudiesSettings.click();
+
+    await expect(this.userSettings).toBeVisible();
+    await this.userSettings.click();
+
+    await expect(this.pacedTrainSwitch).toBeVisible();
+    await expect(this.pacedTrainSwitch).not.toBeChecked();
+    await this.pacedTrainSwitch.click();
```
Yohh, line 166: you should check the state of the switch after clicking on it

front/tests/pages/operational-studies-page-model.ts
```diff
@@ -121,19 +151,113 @@ class OperationalStudiesPage extends CommonPage {
...
+    await expect(this.returnSimulationResultButton).toBeVisible();
+    await expect(this.trainNameInput).toBeVisible();
+
+    await expect(this.startTimeField).toBeVisible();
+    const startTimeDate = new Date(await this.startTimeField.inputValue());
+    const scenarioCreationDateDate = new Date(scenarioCreationDate);
+    const isSameDate =
+      startTimeDate.getFullYear() === scenarioCreationDateDate.getFullYear() &&
+      startTimeDate.getMonth() === scenarioCreationDateDate.getMonth() &&
+      startTimeDate.getDate() === scenarioCreationDateDate.getDate();
```
Yohh:
```suggestion
    date: string
  ) {
    await expect(this.addTrainButton).toBeVisible();
    await expect(this.addTrainButton).toHaveText(translations.addTrainSchedule);
    await expect(this.definePacedTrainCheckboxLabel).toBeVisible();
    await expect(this.definePacedTrainCheckbox).not.toBeChecked();
    await expect(this.returnSimulationResultButton).toBeVisible();
    await expect(this.trainNameInput).toBeVisible();

    await expect(this.startTimeField).toBeVisible();
    const startTimeDate = new Date(await this.startTimeField.inputValue());
    const scenarioCreationDate = new Date(date);
    const isSameDate =
      startTimeDate.getFullYear() === scenarioCreationDate.getFullYear() &&
      startTimeDate.getMonth() === scenarioCreationDate.getMonth() &&
      startTimeDate.getDate() === scenarioCreationDate.getDate();
```

front/tests/pages/operational-studies-page-model.ts
```diff
@@ -121,19 +151,113 @@ class OperationalStudiesPage extends CommonPage {
...
+    await this.verifyTabWarningPresence();
+  }
+
+  async checkPacedTrainModeAndVerifyInputs(translations: ManageTrainScheduleTranslations) {
+    await this.definePacedTrainCheckboxLabel.click();
+    await expect(this.addTrainButton).toHaveText(translations.addPacedTrain);
+    await expect(this.pacedTrainTimeRangeDurationInput).toBeVisible();
+    await expect(this.pacedTrainTimeRangeDurationInput).toHaveValue('120');
+    await expect(this.pacedTrainCadenceInput).toBeVisible();
+    await expect(this.pacedTrainCadenceInput).toHaveValue('60');
```
Yohh: magic numbers

front/tests/005-operational-studies.spec.ts
```diff
@@ -50,68 +53,52 @@ test.describe('Verify simulation configuration in operational studies', () => {
...
+    // Test the paced train mode behavior
+    await operationalStudiesPage.testPacedTrainMode(translations);
 
-    // Confirm rolling stock selection by clicking the button on the card
-    await rollingstockCard.locator('button').click();
+    // Set the paced train inputs
+    await operationalStudiesPage.setTimeRangeDuration('150');
+    await operationalStudiesPage.setCadence('20');
+    await operationalStudiesPage.setTrainScheduleName('Paced train test');
+    await operationalStudiesPage.setTrainStartTime('08:35:40');
```
Yohh: you can set it in a method, with a const to avoid magic numbers, add `120` and `60` in the same const

front/tests/pages/operational-studies-page-model.ts
```diff
@@ -121,19 +151,113 @@ class OperationalStudiesPage extends CommonPage {
...
+    await rollingStockSelectorPage.openRollingstockModal();
+    const rollingstockCard = rollingStockSelectorPage.getRollingstockCardByTestID(
+      `rollingstock-${dualModeRollingStockName}`
+    );
+    await rollingstockCard.click();
+    await rollingstockCard.locator('button').click();
+    expect(
+      await rollingStockSelectorPage.getRollingStockMiniCardInfo().first().textContent()
+    ).toMatch(dualModeRollingStockName);
+  }
```
Yohh: you could import `selectRollingStock` from `rolling-selector-page-model`

front/tests/pages/operational-studies-page-model.ts
```diff
@@ -44,12 +65,21 @@ class OperationalStudiesPage extends CommonPage {
...
     this.trainCountInput = page.locator('#osrdconf-traincount');
-    this.deltaInput = page.locator('#osrdconf-delta');
-    this.addTrainScheduleButton = page.getByTestId('add-train-schedules');
-    this.trainScheduleNameInput = page.locator('#trainSchedule-name');
+    this.operationStudiesSettings = page.getByTestId('dropdown-sncf');
+    this.userSettings = page.getByTestId('user-settings-btn');
+    this.modalCloseButton = page.getByTestId('modal-close-button');
+    this.pacedTrainSwitch = page.getByTestId('paced-train-switch');
+    this.definePacedTrainCheckbox = page.locator('#define-paced-train');
+    this.definePacedTrainCheckboxLabel = page.locator('label[for="define-paced-train"]');
```
Yohh, line 78: you could verify the checkbox innerText

front/tests/005-operational-studies.spec.ts
SharglutDev: Done

front/tests/pages/operational-studies-page-model.ts
SharglutDev: Done

front/tests/pages/operational-studies-page-model.ts
SharglutDev: Done

front/tests/pages/operational-studies-page-model.ts
SharglutDev: Done

front/tests/005-operational-studies.spec.ts
SharglutDev: Done (added some stuff in types.ts and a new consts.ts file)

front/tests/pages/operational-studies-page-model.ts
SharglutDev: Good point ! Done.

front/tests/pages/operational-studies-page-model.ts
SharglutDev: I re-used the way we select `CheckboxSNCF` label in rolling stock editor tests (since the label is hidden in the component). But I added an extra check for the text of the label in the `checkInputsAndButtons` method if it's what you meant.

Yohh (approved):

PR #10773, opened 2025-02-11
front: remove @vitejs/plugin-react

emersion:
Two alternatives exist to integrate Vite with React: @vitejs/plugin-react and @vitejs/plugin-react-swc. The former uses Babel, the latter uses SWC.

We use @vitejs/plugin-react-swc, so we have no use for @vitejs/plugin-react. Drop it from our dependencies.


kmer2016 (approved):
LGTM !

Akctarus (approved):
tested, lgtm !

PR #10775, opened 2025-02-11
front: de-duplicate extractMarkerInformation()

emersion:
_See individual commits._

All branches of the if are very similar here.

Unfortunately because of the MarkerProperties type definition we need this ternary and spread operator.


theocrsb (approved):
LGTM and tested

SharglutDev (approved):
Nice, thank you

PR #10782, opened 2025-02-12
front: refacto useSearchOperationalPoints and get all ch options when selecting a op in stdcm

### LINKED ISSUE #10670: stdcm: search returns different ch codes for same ci using name and trigram
### What happened?

The OP search does not return the same ch results for the same OP, depending on if I write full name or trigram

https://github.com/user-attachments/assets/8e4305fb-d00b-42d1-a687-489543c0dc23

endpoint returning all the ch

1. follow video

Recette (SNCF)

Firefox

dev

clarani:
closes #10670

You can review commit by commit and hide whitespaces ;)

### DIFF
--- a/front/src/common/Map/Search/useSearchOperationalPoint.tsx
+++ b/front/src/common/Map/Search/useSearchOperationalPoint.tsx
@@ -154,18 +154,19 @@ export default function useSearchOperationalPoint({
     [infraID, isStdcm, isSuperUser]
   );
 
-  const filteredAndSortedSearchResults = useMemo(
-    () =>
-      searchResults.filter((result) => {
-        if (mainOperationalPointsOnly || (chCodeFilter && MAIN_OP_CH_CODES.includes(chCodeFilter)))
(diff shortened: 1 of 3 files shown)


front/src/common/Map/Search/useSearchOperationalPoint.tsx
```diff
@@ -44,135 +40,138 @@ export default function useSearchOperationalPoint({
...
+          ['=', ['infra_id'], infraID],
+          dpyToMasOperationalpointsFilter,
+        ],
+      };
+      try {
+        const results = (await postSearch({
+          searchPayload: payload,
+          pageSize: 101,
+        }).unwrap()) as SearchResultItemOperationalPoint[];
+        results.sort(sortOperationalPoints(searchQuery));
```
kmer2016: It seems that the trigram search is broken, I have this bug there whenI am trying to search an OPby trigram
image

kmer2016 (approved):
Thank you for the refacto. !

Synar (approved):

PR #10784, opened 2025-02-12
editoast: simplify the continuation state for streams

woshilapin:
Suggestion to simplify all of these `ContinuationUnfolder`, `UnfoldArgs`, `UnfoldNextState` and `Iter` into a single struct used everywhere. To be honest, I don't really understand the existence of 4 different structs. At the very least, `futures::stream::try_unfold` does use the same struct as input and output of the closure so to me, `UnfoldArgs`, `UnfoldNextState` and `Iter` are only one struct. Tell me if you like it.

### DIFF
--- a/fga/src/client.rs
+++ b/fga/src/client.rs
@@ -142,19 +142,15 @@ impl Client {
     }
 
     pub fn stores(&self) -> impl stream::TryStream<Ok = Store, Error = RequestFailure> + '_ {
-        ContinuationUnfolder::new(()).stream(
-            move |UnfoldArgs {
-                      continuation,
-                      ctx: _ctx,
-                  }| async move {
-                let (stores, continuation) = self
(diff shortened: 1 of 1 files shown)


leovalais:
Much better indeed thanks! If the `ctx` disappears, we even might be able to the `enum Continuation` directly instead of `ContinuationState` :)

Using the same struct for both the argument and the returned state was the first thing I did but moved away from that while trying to abstract away the looping process: [`28052f1` (#10504)](https://github.com/OpenRailAssociation/osrd/pull/10504/commits/28052f17d59f5864081ee5abd0567ce9612f4fca) 

As a great regional manager once said:

> Well, well, well. How the turntables...

leovalais (approved):

leovalais:
I'm merging since I want to use it for `list-objects`. Feel free to commit directly on the target branch if there was changes you still wanted to make :)

PR #10785, opened 2025-02-12
core: stdcm: handle engineering allowances at low speed

eckter:
Low speeds would mess with the envelope operations, we'd stop before the first step

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/EngineeringAllowanceManager.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/EngineeringAllowanceManager.kt
@@ -130,8 +130,8 @@ class EngineeringAllowanceManager(private val graph: STDCMGraph) {
                 -1.0
             )
             val builder = OverlayEnvelopeBuilder.backward(maxEffort)
-            if (speedupPartBuilder.stepCount() > 1) {
-                val speedupPart = speedupPartBuilder.build()
+            val speedupPart = speedupPartBuilder.build()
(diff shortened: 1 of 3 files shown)


Erashin (approved):
Fun

Erashin (approved):

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/EngineeringAllowanceManager.kt
```diff
@@ -112,6 +112,14 @@ class EngineeringAllowanceManager(private val graph: STDCMGraph) {
...
             }
             if (beginSpeed == 0.0 || endSpeed == 0.0) return Double.POSITIVE_INFINITY
+            if (maxEffort.maxSpeed < 6.0 && maxEffort.endPos > 120.0) {
+                // Low speeds mess with the envelope computations, we end up having less than one
+                // step. As long as there's *some* length, we can be optimistic.
+                // The max speed isn't random, generated tracks for takeovers are set to
+                // 20km/h = 5.5m/s. With 0.5ms-2 average deceleration and acceleration,
+                // we need 60m to reach a stop and go back to 6m/s.
+                return Double.POSITIVE_INFINITY
+            }
```
bougue-pe: This is not required anymore, is it?

eckter: Oh it seems I messed up with git, thanks for the catch

PR #10787, opened 2025-02-12
front: tidy map layers folder

clarani:


kmer2016 (approved):
LGTM ! Thank you for cleaning

PR #10788, opened 2025-02-12
editoast, front: host fonts for the map in editoast

### LINKED ISSUE #10627: Map: rationalize fonts and sprites

Tristramg:
The glyphs are generated during the build step

Closes #10627

How to test: open the map, filter network requests by `roboto` and make sure that everything is 200 and is now on localhost.

(Should probably be merged after #10680 and #10628)


editoast/Dockerfile
```diff
@@ -23,11 +23,15 @@ COPY --from=static_assets . /assets
 #######################
 # Build assets        #
 #######################
-FROM alpine:latest AS editoast_assets
+FROM ubuntu:latest AS editoast_assets
```
woshilapin, line 26: Why the switch of image from `alpine` to `ubuntu`?

front/public/locales/fr/errors.json
```diff
@@ -61,6 +61,9 @@
     "electrical_profiles": {
       "NotFound": "Profil électrique '{{electrical_profile_set_id}}' non trouvé"
     },
+    "fonts": {
+      "FileNotFound": "Le fichier pour la police de charactère demandée non trouvé"
```
woshilapin:
```suggestion
      "FileNotFound": "Le fichier pour la police de charactère demandée n'a pas été trouvé"
```

or

```suggestion
      "FileNotFound": "Fichier pour la police de charactère demandée non trouvé"
```

front/public/locales/en/errors.json
```diff
@@ -61,6 +61,9 @@
     "electrical_profiles": {
       "NotFound": "Electrical Profile Set '{{electrical_profile_set_id}}', could not be found"
     },
+    "fonts": {
+      "FileNotFound": "The file for the request font not found"
```
woshilapin:
```suggestion
      "FileNotFound": "The file for the requested font was not found"
```

or 

```suggestion
      "FileNotFound": "File for the requested font not found"
```

woshilapin (approved):
Nice follow up of the sprite PR. Thank you.

editoast/Dockerfile
Tristramg: Because the binary has some library bindings that don’t work out of the box on alpine. I don’t know if it’s a musl/gnu thing or some library that is install on the ubuntu image

flomonster: We could also use `FROM chef` and then simply install it with `cargo`.

editoast/src/views/fonts.rs
```diff
@@ -0,0 +1,91 @@
...
+    Path((font, file_name)): Path<(String, String)>,
+    request: Request,
+) -> Result<impl IntoResponse> {
+    let authorized = auth
+        .check_roles([BuiltinRole::MapRead].into())
+        .await
+        .map_err(AuthorizationError::AuthError)?;
+    if !authorized {
+        return Err(AuthorizationError::Forbidden.into());
+    }
```
flomonster: Assets are open, I don't think we need to expect users to have the `MapRead` role

flomonster (approved):
LGTM

editoast/src/views/fonts.rs
Tristramg: This was also for the sprite assets, should I simplify both?

editoast/Dockerfile
Tristramg: I’m fine with either, do you have any preference?

editoast/src/views/fonts.rs
flomonster: We should simplify this. We can do it here or in another PR.

editoast/assets/fonts/generate-glyphs.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+# This script should be used to generate glyphs from ttf fonts.
+# Those glyphs are used to display text on the map
+# You will need build_pbf_glyphs, you can install it with:
+# `$ cargo install build_pbf_glyphs`
+
```
flomonster, line 8:
```suggestion
set -e
```

editoast/src/views/fonts.rs
Tristramg: I removed it for fonts.

For the sprites see https://github.com/OpenRailAssociation/osrd/pull/10828

front/src/common/Map/Layers/blankStyle.ts
```diff
@@ -66,7 +66,7 @@ export const useMapBlankStyle = (): MapProps['mapStyle'] => {
       name: 'Blank',
       sources: {},
       sprite,
-      glyphs: 'https://static.osm.osrd.fr/fonts/{fontstack}/{range}.pbf',
+      glyphs: `${window.location.origin}${baseURL}/fonts/{fontstack}/{range}.pbf`,
```
emersion, line 69: Hm, I don't think we should be using `window.location.origin` here, `baseURL` alone should be enough. `baseURL` is taken from `PROXY_EDITOAST` from the environment and can contain a host.

That said, this is copy-pasted from elsewhere so orthogonal to this PR. (Feel free to resolve.)

emersion (approved):
LGTM for the frontend

(I'm not sure the per-commit split makes sense: the script and translation file changes belong to the first commit IMHO.)

front/src/common/Map/Layers/blankStyle.ts
Tristramg: I’ll do a follow-up PR to change that in this place and the one or 2 other places

Tristramg:
I re-arranged the commits

front/public/locales/fr/errors.json
```diff
@@ -61,6 +61,9 @@
     "electrical_profiles": {
       "NotFound": "Profil électrique '{{electrical_profile_set_id}}' non trouvé"
     },
+    "fonts": {
+      "FileNotFound": "Fichier pour la police de charactère demandée non trouvé"
```
Akctarus, line 65: there is a typo : caractère instead of charactère

Akctarus (commented):
there is a small typo, otherwise lgtm

PR #10790, opened 2025-02-13
front: add rolling stock categories

### LINKED ISSUE #10691: front: Ajouter des catégories dans l'éditeur de MR
### Description and goal

Front part of [#662](https://github.com/osrd-project/osrd-confidential/issues/662) (sncf login required)

Cette feature est la première brique permettant de gérer le concept de "catégorie de matériel roulant". Il existe de nombreuses manières selon les métiers et les usages de définir ces catégories. Pour le besoin d'OSRD, nous avons choisi de retenir les catégories suivantes (non spécifiques à SNCF) :

    high-speed-train
    intercity-train
    regional-train-mu
    night-train
    commuter-train
    freight-train
    fast-freight-train
    tram-train
    touristic-train
    work-train

- [x]     Les catégories sont contenues dans un Enum -> dict actually, enables easier type checking consistency with the API
- [x]     Intégrer des nouveaux champs de formulaire avec des checkboxes (une checkbox par catégorie) en bouclant sur cette Enum
- [x]     Lorsqu'une box est cliquée, on ajoute une catégorie dans le tableau de catégories du rolling stock
- [x]     Lorsqu'on clique sur "enregistrer" on utilise l'endpoint d'édition/update de rolling stock
- [x]     Il faut trouver un systeme pour définir la catégorie primaire. Demander à @thibautsailly 
- [x]     Traduction des catégories en français

    permet de https://github.com/osrd-project/osrd-confidential/issues/776
    permet de https://github.com/osrd-project/osrd-confidential/issues/779
    permet d'https://github.com/osrd-project/osrd-confidential/issues/657
    permet d'https://github.com/osrd-project/osrd-confidential/issues/698
    permet de https://github.com/osrd-project/osrd-confidential/issues/672

### Acceptance criteria

- [x] Chaque MR a au moins une catégorie
- [x] Chaque MR a exactement une catégorie principale
- [x] Chaque MR peut avoir autant de catégorie "autres" que voulu par l'utilisateur

- Il n'y a pas de vérification de cohérence sur ces valeurs (par exemple un "high-speed-train" peut aussi être "freight-train" sans alerte pour l'utilisateur)
- Pas d'affichage dans la carte qui représente le matériel roulant, uniquement dans le détail
- Pas de filtre prévu pour le moment

Synar:
Close #10691
Depend on #10701 

Behavior and translations should be fine, but the css could still use a little improvement ^^


front/public/locales/en/rollingstock.json
```diff
@@ -63,6 +76,7 @@
     "failure": "Operation failed",
     "invalidForm": "Incomplete form",
     "missingName": "Please fill out a name",
+    "missingPrimaryCategory": "Please fill out a primary category",
```
SharglutDev, line 82: nit : I know it's not your fault, but would you mind to order these `messages` key ? :)

front/public/locales/fr/rollingstock.json
```diff
@@ -5,6 +5,19 @@
   "basePowerClass": "Classe de puissance",
   "cancelAction": "Souhaitez-vous annuler cette action ?",
   "cancelUpdateRollingStock": "Annuler la mise à jour du matériel roulant?",
+  "categoriesOptions": {
+    "UNKNOWN": "Inconnu",
+    "HIGH_SPEED_TRAIN": "Train à grande vitesse",
+    "INTERCITY_TRAIN": "Train interurbain",
```
SharglutDev, line 11: I think we use `Train intercités` in France ?

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -3138,7 +3139,22 @@ export type EffortCurves = {
     [key: string]: ModeEffortCurves;
   };
 };
+export type RollingStockCategory =
+  | 'UNKNOWN'
```
SharglutDev: What is this category used for ?

front/src/common/api/generatedEditoastApi.ts
SharglutDev, line 1: Are these changes hard coded for now ?

front/src/modules/rollingStock/types.ts
```diff
@@ -73,6 +68,8 @@ export type RollingStockParametersValues = {
   basePowerClass: string | null;
   powerRestrictions: RollingStock['power_restrictions'];
   supportedSignalingSystems: string[];
+  primaryCategory: RollingStockCategory | undefined;
```
SharglutDev:
```suggestion
  primaryCategory?: RollingStockCategory;
```

front/src/modules/rollingStock/components/RollingStockSelector/sampleData.ts
```diff
@@ -1,6 +1,4 @@
-import type { RollingStockWithLiveries } from 'common/api/osrdEditoastApi';
```
SharglutDev: We could use the type `RollingStock`

front/src/modules/rollingStock/consts.ts
```diff
@@ -77,6 +82,7 @@ export const RS_REQUIRED_FIELDS = Object.freeze({
   rollingResistanceC: newRollingStockValues.rollingResistanceC,
   electricalPowerStartupTime: 0,
   raisePantographTime: 15,
+  primaryCategory: 'FREIGHT_TRAIN',
```
SharglutDev: This seems to indicate that if missing, primary category will be filled with `FREIGHT_TRAIN`. Is it an AC ? I'm not sure we want that.
Also if I try to submit the form without any primary category, I get a tooltip error saying it's compulsory and it's not autofilled.

front/src/modules/rollingStock/consts.ts
```diff
@@ -337,3 +343,29 @@ export const EP_BY_MODE: ElectricalProfileByMode = {
...
+  INTERCITY_TRAIN = 'INTERCITY_TRAIN',
+  REGIONAL_TRAIN = 'REGIONAL_TRAIN',
+  NIGHT_TRAIN = 'NIGHT_TRAIN',
+  COMMUTER_TRAIN = 'COMMUTER_TRAIN',
+  FREIGHT_TRAIN = 'FREIGHT_TRAIN',
+  FAST_FREIGHT_TRAIN = 'FAST_FREIGHT_TRAIN',
+  TRAM_TRAIN = 'TRAM_TRAIN',
+  TOURISTIC_TRAIN = 'TOURISTIC_TRAIN',
+  WORK_TRAIN = 'WORK_TRAIN',
+}
```
SharglutDev: Wouldn't it work to just use an object like `const RollingStockCategoryDict: Record<RollingStockCategory, RollingStockCategory>`.
The value would be the translation key. If we forget a category, typescript will tell us so we don't need the extra type `_CheckRollingStockCategoryEnumMatchesAPI` below. What do you think ?

front/src/modules/rollingStock/components/RollingStockEditor/RollingStockEditorForm.tsx
```diff
@@ -157,6 +158,11 @@ const RollingStockEditorForm = ({
         name: t('messages.invalidForm'),
         message: t('messages.missingName'),
       };
+    } else if (!data.primaryCategory) {
+      error = {
+        name: t('messages.invalidForm'),
+        message: t('messages.missingPrimaryCategory'),
+      };
```
SharglutDev, line 165: Related to the comment in consts.ts. If we keep this, we won't have the default primary category set.

front/src/modules/rollingStock/components/RollingStockEditor/RollingStockEditorFormHelpers.tsx
```diff
@@ -372,3 +374,87 @@ export const RollingStockEditorOnboardSystemEquipmentForm = ({
...
+        ...prevValues,
+        categories: updatedCategories,
+      }));
+    };
+
+  return (
+    <div className="rollingstock-editor-input-container px-1 pb-3">
+      {/* Primary Category Selection */}
+      <div className="d-flex align-items-center justify-content-between rollingstock-editor-select mb-4">
+        <SelectSNCF
```
SharglutDev, line 422: Even though the mockup uses the new design system components, I think it's fine to keep an harmonized UI with the old components while waiting for a full rework on the rs editor.

Maybe we could just show to @thibautsailly what it looks like with the wrapped checkbox to see it's fine for him.

front/src/modules/rollingStock/components/RollingStockEditor/RollingStockEditorFormHelpers.tsx
```diff
@@ -372,3 +374,87 @@ export const RollingStockEditorOnboardSystemEquipmentForm = ({
...
+        </label>
+        <div className="d-flex flex-wrap" id="rs_category_checkboxes">
+          {Object.values(RollingStockCategoryEnum).map((category) => (
+            <div key={category} className={cx('col-6', 'col-xl-4', 'mb-2')}>
+              <CheckboxRadioSNCF
+                type="checkbox"
+                id={`other_category_${category}`}
+                name="other_categories"
+                label={t(`categoriesOptions.${category}`)}
+                checked={rollingStockValues.categories.includes(category)}
```
SharglutDev: If the idea is indeed to display a default primary category, when it does, the associated "other category" is disabled but not checked.

SharglutDev (commented):
Thank you for the PR. Left some comments. I'm a bit confused on some AC, maybe there need to be some extra discussions with @maelysLeratRosso and/or @thibautsailly regarding the "requiredness" of the primary category :)

front/public/locales/fr/rollingstock.json
maelysLeratRosso: We chose a translation which was less SNCF dependent and was more generic, that's why it was Train interurbain

front/src/modules/rollingStock/consts.ts
maelysLeratRosso: The primary category should not be filled by default. However, one should not be able to validate the form if no primary category is defined.

front/public/locales/en/rollingstock.json
Synar: Sure ^^

front/public/locales/fr/rollingstock.json
Synar: @sharglut that was my first guess too, but it wasn't the final choice ^^

front/src/common/api/generatedEditoastApi.ts
Synar: It was to be temporarily used for the editoast migration. In the end it was removed before merging the pr, so I'll remove it here too.

front/src/common/api/generatedEditoastApi.ts
Synar: Yes, hence the name of the commit, "front: temp temp temp waiting for api update" ^^

Synar: I'll of course delete the commit when rebasing

front/src/modules/rollingStock/types.ts
Synar: Good catch, thanks!

front/src/modules/rollingStock/consts.ts
Synar: See the comment in the code.

There is this weird quirk in the code where the object Required fields comes with a list of default values, and there is a different, unrelated check at form submission where only the presence of name, effort curve and (now) primary category are checked.

So the form indeed can not be submitted without a primary category.

The simplest way of avoiding adding an unused default value like I've done here would be to let undefined be a valid value for primary_category in RollingStockParametersValidValues, and not add primary category to the list of required fields, but still require it on form validation. I think it's a bit unintuitive too.

front/src/modules/rollingStock/consts.ts
Synar: Great idea! I'll check, but I believe it should work ^^

front/src/modules/rollingStock/components/RollingStockEditor/RollingStockEditorForm.tsx
Synar: See my other response. Indeed the default value is not intended to be exposed to the user. I tried to follow what was already done for name, which follows the same pattern.

front/src/modules/rollingStock/components/RollingStockEditor/RollingStockEditorFormHelpers.tsx
Synar: values.categories should always contain the primary category, unless I messed up. Did you get to such a state while testing? I'll check

SharglutDev: No not at all.

I'm also a bit confused by the editoast naming : `other_categories` can't be undefined so it means if I select only one primary category, it's added in other_categories but it's not another one it's the same.

Also what should be the behavior if the user start by putting "freight train" for the primary, so it's added on "others", then he changes for "tourist train". We now have 2 other categories but they are kind of side effects of the primary selection. Do we want to keep "freight train" in that case ?

Synar: other_categories can't be undefined, but it can be empty. It should never contain the primary category. So when we display in details, we display on one line the primary category, and on another the others.

On the other hand, categories which is used in the edition form always contains the primary category, as it makes the checkboxes more intuitive for the user.

The primary category is (supposedly) removed from the list when converting from RSValidValues.categories to openapi RSForm.other_categories, and added when going from openapi RS.other_categories to RSValues.categories. Perhaps I should had a comment to clarify this?

For the question of choosing primary category A then B, I think A should remain in categories until it's manually unchecked. I think that's how the proof-of-concept mockup worked. We can ask Thibaut for confirmation though, it should be easy to change.

SharglutDev: > other_categories can't be undefined, but it can be empty. It should never contain the primary category. So when we display in details, we display on one line the primary category, and on another the others.

You're totally right !

> The primary category is (supposedly) removed from the list when converting from RSValidValues.categories to openapi RSForm.other_categories, and added when going from openapi RS.other_categories to RSValues.categories. Perhaps I should had a comment to clarify this?

Didn't pay attention about that part as connection to back wasn't up but you're right

> For the question of choosing primary category A then B, I think A should remain in categories until it's manually unchecked. I think that's how the proof-of-concept mockup worked. We can ask Thibaut for confirmation though, it should be easy to change.

Agreed :)

front/public/locales/en/rollingstock.json
Synar: Done, but now I think we should have a script for that =p
Perhaps I'll do a pr when I have time

front/src/modules/rollingStock/consts.ts
Synar: I've added bonus comments to clarify

front/src/modules/rollingStock/components/RollingStockSelector/sampleData.ts
Synar: (For anyone wondering what this is about, the change to this file was temporary, and the file ended up entirely deleted in a previous PR)

front/src/modules/rollingStock/components/RollingStockEditor/RollingStockEditorFormHelpers.tsx
```diff
@@ -372,3 +374,87 @@ export const RollingStockEditorOnboardSystemEquipmentForm = ({
...
   );
 };
+
+export const RollingStockEditorCategoryForm = ({
+  rollingStockValues,
+  setRollingStockValues,
+}: RollingStockEditorParameterFormProps) => {
+  const { t } = useTranslation(['rollingstock']);
+
+  type CategoryOption = { id: RollingStockCategory | ''; label: string };
```
SharglutDev: We can move this type in type.ts (or just outside of the function at least)

front/src/modules/rollingStock/components/RollingStockEditor/RollingStockEditorFormHelpers.tsx
```diff
@@ -372,3 +374,87 @@ export const RollingStockEditorOnboardSystemEquipmentForm = ({
...
+  const handlePrimaryCategoryChange = (selectedCategory: CategoryOption | undefined) => {
+    if (!selectedCategory) return;
+    setRollingStockValues((prevValues) => ({
+      ...prevValues,
+      primaryCategory: selectedCategory.id || undefined,
+      categories: selectedCategory.id
+        ? [...rollingStockValues.categories, selectedCategory.id]
+        : rollingStockValues.categories,
+    }));
+  };
```
SharglutDev: When updating a primary category, I don't think we can select an empty one
```suggestion
  const handlePrimaryCategoryChange = (selectedCategory: CategoryOption) => {
    setRollingStockValues((prevValues) => ({
      ...prevValues,
      primaryCategory: selectedCategory.id,
      categories: [...rollingStockValues.categories, selectedCategory.id]
    }));
  };
```

front/src/modules/rollingStock/components/RollingStockEditor/RollingStockEditorFormHelpers.tsx
```diff
@@ -372,3 +374,87 @@ export const RollingStockEditorOnboardSystemEquipmentForm = ({
...
+    { id: '', label: '' },
+  ];
+
+  const handlePrimaryCategoryChange = (selectedCategory: CategoryOption | undefined) => {
+    if (!selectedCategory) return;
+    setRollingStockValues((prevValues) => ({
+      ...prevValues,
+      primaryCategory: selectedCategory.id || undefined,
+      categories: selectedCategory.id
+        ? [...rollingStockValues.categories, selectedCategory.id]
```
SharglutDev: This should maybe be a `Set` here rather than a regular array. If I select prima cat A, then B, then A again, then C, I'll end up with 2 cat A in `other_categories` and they are not filtered before sending the endpoint (only the primary one is remove from the array)

front/src/modules/rollingStock/components/RollingStockEditor/RollingStockEditorFormHelpers.tsx
Synar: Seen with thibaut ^^

front/src/modules/rollingStock/components/RollingStockEditor/RollingStockEditorFormHelpers.tsx
Synar: I don't like the idea of exporting it in rollingstock/types, as it's a very local 'private' type. I agree with putting it outside of the function thougn

front/src/modules/rollingStock/components/RollingStockEditor/RollingStockEditorFormHelpers.tsx
Synar: Sadly the parameter onChange of the select component has the following signature : onChange: (e?: T | undefined) => void, where T is the type of the provided options. So even if if it should not happen (though I did not check the implementation of select to know if perhaps it happens when handling some kind of events), we have to deal with the possibility for typescript to be happy.
 
 There is another check here though for the truth value of selectedCategory.id (in primaryCategory: selectedCategory.id || undefined), which happens when selecting the blank value back for the primary category. We could absolutely disable this by removing the blank string from the options as long as a value was selected for primary category. Would you like me to make this change? I'll ask Thibaut what he thinks

front/src/modules/rollingStock/components/RollingStockEditor/RollingStockEditorFormHelpers.tsx
Synar: Oh my bad, you're totally right!

front/src/modules/rollingStock/components/RollingStockEditor/RollingStockEditorFormHelpers.tsx
Synar: Simplified things a bit

I also discussed with Thibaut and now we have a Choose label.

front/src/modules/rollingStock/components/RollingStockEditor/RollingStockEditorFormHelpers.tsx
```diff
@@ -372,3 +374,90 @@ export const RollingStockEditorOnboardSystemEquipmentForm = ({
...
+
+  const categoryOptions: CategoryOption[] = [
+    { label: t('categoriesOptions.choose') },
+    ...Object.values(RollingStockCategoryDict).map((category) => ({
+      id: category,
+      label: t(`categoriesOptions.${category}`),
+    })),
+  ];
+
+  const handlePrimaryCategoryChange = (selectedCategory: CategoryOption | undefined) => {
```
SharglutDev:
```suggestion
  const handlePrimaryCategoryChange = (selectedCategory?: CategoryOption) => {
```

front/src/modules/rollingStock/components/RollingStockEditor/RollingStockEditorFormHelpers.tsx
SharglutDev: Looks great !

SharglutDev (approved):
Lgtm and tested, great job !
You can resolve the last comment when fixed (and don't forget to squash)

front/src/modules/rollingStock/components/RollingStockEditor/RollingStockEditorFormHelpers.tsx
Synar: Done thanks! ^^

Maymanaf (commented):
Good job! But can you enhance the E2E tests?
**For the "Create a new rolling stock" test:**

- Verify that the primary category is checked and disabled in other categories
- Check additional categories and verify that they are selected
- Verify the categories after rolling stock creation

**For the "Duplicate and modify a rolling stock" test:**

- Modify the primary category
- Select additional categories
- Verify that the categories are updated after saving

front/tests/pages/rollingstock-editor-page-model.ts
```diff
@@ -51,6 +75,18 @@ class RollingstockEditorPage extends CommonPage {
...
+    });
+    this.categoriesCheckboxes = Object.fromEntries(
+      Object.keys(RollingStockCategoryDict).map((category) => [
+        category,
+        page.getByRole('checkbox', {
+          name: this.translations.categoriesOptions[category],
+          exact: true,
+        }),
+      ])
+    ) as Record<RollingStockCategory, Locator>;
```
Maymanaf:
```suggestion
  private getCategoryCheckbox(category: string): Locator {
    return this.page.locator(`#other_category_${category}`);
  }

```
why not use the ID or even add data-testid, it's faster than querying by role and name

front/tests/pages/rollingstock-editor-page-model.ts
```diff
@@ -176,6 +212,30 @@ class RollingstockEditorPage extends CommonPage {
...
+  // Select primary category
+  async selectPrimaryCategory(value: RollingStockCategory) {
+    await this.primaryCategorySelector.selectOption(value);
+    await expect(this.primaryCategorySelector).toHaveValue(value);
+    await expect(this.categoriesCheckboxes[value]).toBeChecked();
+    await expect(this.categoriesCheckboxes[value]).toBeDisabled();
+  }
+
+  // Select an unselected category checkbox
+  async selectUnselectedCategory(category: RollingStockCategory) {
```
Maymanaf: somewhat redundant
 
```suggestion
  async selectCategory(category: RollingStockCategory) {
```
or 
```suggestion
  async checkCategoryCheckbox(category: RollingStockCategory) {
```

front/tests/pages/rollingstock-editor-page-model.ts
```diff
@@ -176,6 +212,30 @@ class RollingstockEditorPage extends CommonPage {
...
+    const checkbox = this.categoriesCheckboxes[category];
+    await expect(checkbox).not.toBeChecked();
+    await checkbox.locator('..').locator('label').click();
+    await expect(checkbox).toBeChecked();
+  }
+
+  // Deselect a selected category checkbox
+  async deselectSelectedCategory(category: RollingStockCategory) {
```
Maymanaf: same

front/tests/009-rollingstock-editor.spec.ts
```diff
@@ -55,6 +55,16 @@ test.describe('Rollingstock editor page tests', () => {
     }
     await rollingStockEditorPage.selectLoadingGauge('GA'); // Select loading gauge
 
+    // Select a primary category and other categories, checking each time the state of selector and checkboxes are correct
+    await rollingStockEditorPage.selectPrimaryCategory('WORK_TRAIN');
+    await rollingStockEditorPage.selectPrimaryCategory('NIGHT_TRAIN');
+    await rollingStockEditorPage.deselectSelectedCategory('WORK_TRAIN');
```
Maymanaf: Selecting a new primary category shouldn't automatically deselect the previous one?

front/tests/pages/rollingstock-editor-page-model.ts
Synar: Playwright recommends using role locators over .locators using id, to be closer to actual user interaction.

> Playwright comes with multiple built-in locators. To make tests resilient, we recommend prioritizing user-facing attributes and explicit contracts such as page.getByRole().

(...)

> CSS and XPath are not recommended as the DOM can often change leading to non resilient tests. Instead, try to come up with a locator that is close to how the user perceives the page such as role locators or define an explicit testing contract using test ids.

Using test ids could also work though

front/tests/009-rollingstock-editor.spec.ts
Synar: No, it should change the primary category, but leave the previous category in the category list. We thus check that we can indeed remove it from the list, meaning the checbox is both still ticked and no longer disabled.

front/tests/pages/rollingstock-editor-page-model.ts
Synar: Sure! I wanted to make it clear that trying to select an already selected category would cause test failure, but perhaps that should be implicit.

front/tests/009-rollingstock-editor.spec.ts
Maymanaf: It wasn't a question about the test code but about the functionality. Thanks for clarifying!

Synar: Np 💜

front/tests/pages/rollingstock-editor-page-model.ts
Maymanaf: Yes, Playwright's built-in locators are recommended, but using `IDs` if they are unique and stable is also a valid choice However, let's go with` page.getByTestId()` since it's a built-in locator and performs a range of actionability checks

Synar: Sure ^^

front/tests/pages/rollingstock-editor-page-model.ts
Synar: Done ^^

front/tests/pages/rollingstock-editor-page-model.ts
Synar: Done ^^

front/tests/pages/rollingstock-editor-page-model.ts
Synar: Added data-testid to primary category selector and category checkboxes, and updated the locators ^^

Maymanaf (approved):
Great job ! 
LGTM for e2e tests ✅

PR #10795, opened 2025-02-13
tests: fuzzer: print error message instead of raise_for_status

eckter:
The tests are flaky because of something related to infra loading, but I don't even know what the underlying error actually is, just that /infra/1/load returns a 500


shenriotpro (approved):
Thanks

PR #10803, opened 2025-02-13
front: no longer display destination on single step path

### LINKED ISSUE #10115: Train schedules with one path step handled weirdly
### What happened?

I have an imported train with only one step : [train_schedules (17).json](https://github.com/user-attachments/files/18168415/train_schedules.17.json)
I see the corresponding step twice, which makes me think that I have twice the same step, which is not the same

Image

Furthermore, if I click on the blue cross to correct the origin for example, both steps are erased 

Image

I guess I would except to have only the origin filled with the corresponding step, and an undefined arrival, but it can be discussed

Import the previous train

Dev (SNCF)

Google Chrome

fe3b339

Synar:
Close #10115 

(Although single step schedules seem kind of nonsensical to me in the first place)


SharglutDev (approved):
Lgtm and tested :)

Caracol3 (approved):
lgtm (tested)

PR #10806, opened 2025-02-14
front: use Duration in TrainScheduleWithDetails.duration

emersion:
This is more strongly typed.

### DIFF
--- a/front/src/modules/trainschedule/components/Timetable/TimetableTrainCard.tsx
+++ b/front/src/modules/trainschedule/components/Timetable/TimetableTrainCard.tsx
@@ -269,7 +269,7 @@ const TimetableTrainCard = ({
             </div>
             <div className="duration-time">
               <span data-testid="train-duration">
-                {dayjs.duration(train.duration).format('HH[h]mm')}
+                {dayjs.duration(train.duration!.ms).format('HH[h]mm')}
               </span>
             </div>
           </div>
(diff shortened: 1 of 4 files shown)


Akctarus (approved):
tested and lgtm !

front/src/modules/trainschedule/components/Timetable/TimetableTrainCard.tsx
```diff
@@ -269,7 +269,7 @@ const TimetableTrainCard = ({
             </div>
             <div className="duration-time">
               <span data-testid="train-duration">
-                {dayjs.duration(train.duration).format('HH[h]mm')}
+                {dayjs.duration(train.duration?.ms ?? 0).format('HH[h]mm')}
```
anisometropie: why don’t you keep 0 as default value in `formatTrainScheduleSummaries.ts`
 if you end up using 0 anyway if it’s null?

emersion: I prefer to have an explicit indicator when the value is non-sensical. `null` means "I don't know" and forces users to handle the case somehow.

But probably we should be using `train.duration!.ms` instead, since this component can never be rendered with an invalid train schedule?

anisometropie: makes sense

i guess you can go for `train.duration!.ms`. But this assumes that we enforce long term that we only ever create valid train with non-null duration.

emersion: Done!

> this assumes that we enforce long term that we only ever create valid train with non-null duration.

This should always be the case: if the train is valid we can simulate it, if we can simulate it then we have an arrival time and total duration. BTW, `arrivalTime` is already nullable.

SharglutDev (approved):
Lgtm and tested ✅

anisometropie (approved):
Big fan of the `Duration` class 
(reviewed only)

PR #10808, opened 2025-02-14
front: use Map instead of Record for train summaries/projections

emersion:
TypeScript implicitly converts Record keys to strings. Map is safer.

### DIFF
--- a/front/src/applications/operationalStudies/helpers/formatTrainScheduleSummaries.ts
+++ b/front/src/applications/operationalStudies/helpers/formatTrainScheduleSummaries.ts
@@ -19,7 +19,7 @@ import { isScheduledPointsNotHonored, isTooFast } from '../utils';
 
 const formatTrainScheduleSummaries = (
   trainIds: TrainId[],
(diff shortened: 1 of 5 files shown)


SharglutDev (approved):
Thanks !

anisometropie:
I’m not against using map here, but Why is it safer ?

The record is using `TrainScheduleId = string & { readonly __type: unique symbol }` as key, so there is no way to put anything else but a `TrainScheduleId` string as the key. We never coerced number into string in there.

emersion:
See the discussion and TypeScript playground link here: https://github.com/OpenRailAssociation/osrd/pull/10679#discussion_r1952347242

tl;dr `Record` is a footgun when the key isn't a plain `string` :o

anisometropie:
> See the discussion and TypeScript playground link here: [#10679 (comment)](https://github.com/OpenRailAssociation/osrd/pull/10679#discussion_r1952347242)
> 
> tl;dr `Record` is a footgun when the key isn't a plain `string` :o

Nice

anisometropie (approved):

PR #10811, opened 2025-02-14
front: fix map markers react warning for duplicate keys

### LINKED ISSUE #10810: React key warning for MapMarkers
### What happened?

Image

No warning in console

Go to a scenario with train and open the console

Recette (SNCF)

Firefox

.

SharglutDev:
fix #10810

### DIFF
--- a/front/src/common/Map/components/MapMarkers.tsx
+++ b/front/src/common/Map/components/MapMarkers.tsx
@@ -38,6 +38,7 @@ const MapMarkers = ({ markers }: MapMarkersProps) =>
         latitude={coordinates[1]}
         anchor="bottom"
         offset={MARKER_OFFSET}
+        key={index}
       >
         <img src={imgSrc} alt={imgAlt} />
         {pointType === MARKER_TYPE.VIA && (


front/src/common/Map/components/MapMarkers.tsx
```diff
@@ -38,6 +38,7 @@ const MapMarkers = ({ markers }: MapMarkersProps) =>
         latitude={coordinates[1]}
         anchor="bottom"
         offset={MARKER_OFFSET}
+        key={`${coordinates[0]}-${coordinates[1]}-${index}`}
```
emersion: The index should be enough no?

SharglutDev: Safer the better ? :) Thought we should avoid to use `index` only when possible.

Synar (approved):
LGTM ^^

(I don't have an opinion on the index vs index+coord key, either choice seems fine to me)

front/src/common/Map/components/MapMarkers.tsx
emersion: Why is that?

What if `coordinates` stay the same but other properties change?

SharglutDev: Fair enough :)

SharglutDev: Changed for `key={index}`

emersion (approved):
Thank you!

PR #10812, opened 2025-02-14
front: drop addDurationToDate() and subtractDurationFromDate() from utils/date

emersion:
Replace them with Duration.

_See individual commits._


front/src/utils/duration.ts
```diff
@@ -9,20 +9,20 @@ export class Duration {
...
-  constructor(ms: number) {
-    this.ms = ms;
+  constructor({ hours = 0, minutes = 0, seconds = 0, milliseconds = 0 }) {
+    this.ms = ((hours * 60 + minutes) * 60 + seconds) * 1000 + milliseconds;
   }
 
-  static zero = new Duration(0);
+  static zero = new Duration({});
 
   /** Parse an ISO 8601 duration string. */
```
kmer2016, line 18: As the function does not only accept durations in ISO 8601 format
```suggestion
  /** Parse a duration string using dayjs. */
```

emersion: This function is not changed in this PR.

The intent of the function is to only support ISO 8601 duration strings, and nothing else. I don't want to tie the `Duration` API to a particular library.

AFAIK since the parameter is a `string` dayjs will not accept other formats.

kmer2016 (commented):
LGTM ! Thank you for the refacto.

kmer2016 (approved):

front/src/modules/pathfinding/helpers/__tests__/reversePathSteps.spec.ts
```diff
@@ -35,7 +35,7 @@ describe('reversePathSteps', () => {
       positionOnPath: 10402000,
       name: 'Pont-Hébert',
       kp: '10+396',
-      arrival: new Duration(1500000),
+      arrival: new Duration({ milliseconds: 1500000 }),
```
anisometropie, line 38: I guess, for readability, we can favor using seconds or minutes now that the constructor allows it

front/src/applications/stdcm/utils/computeOpSchedules.ts
```diff
@@ -29,12 +30,12 @@ const computeOpSchedules = (startTime: Date, msFromStartTime: number) => {
     origin: {
       date: originDate,
       time: originTime,
-      arrivalDate: subtractDurationFromDate(startTime, 1800),
+      arrivalDate: addDurationToDate(startTime, new Duration({ minutes: -30 })),
```
anisometropie, line 33: we could also use this expression (with `valueOf` defined in `Duration` to return the ms value).

`addDurationToDate(startTime, -new Duration({ minutes: 30 }))`

could this make things clearer ?

It could make it read more like `startTime - new Duration({ minutes: 30 })`, infortunately impossible in JS

anisometropie (approved):
Behavior is looking good on the few functions impacted by the change

front/src/modules/pathfinding/helpers/__tests__/reversePathSteps.spec.ts
emersion: I decided to keep the previous number to make it easier to check that the value is still the same, and because readability doesn't matter very much here since it's a bunch of arbitrary numbers. I can change it if you prefer :P

front/src/applications/stdcm/utils/computeOpSchedules.ts
emersion: I don't think this works: once the value is converted to a `number`, it's not a `Duration` object anymore. `addDurationToDate()` accepts only `Duration` objects, not `number`s. (Same with the `-` operator: this would give a `number` instead of a `Date`.)

TypeScript/JavaScript don't have operator overloading.

front/src/modules/pathfinding/helpers/__tests__/reversePathSteps.spec.ts
anisometropie: fine to leave them too

PR #10825, opened 2025-02-14
front: fix unnecessary type casting in upsert path step

### LINKED ISSUE #10805: Use better types in upsertPathSteps
### What happened?

in the UpsertPathSteps function, we have to typecast after adding the track_reference to the statePathSteps array. 
```  newVia = {
      ...newVia,
      id: cleanPathSteps[stepIndex].id,
      track_reference:
        (cleanPathSteps[stepIndex] as OperationalPointReference).track_reference || undefined,
    };
```

this is weird as OperationalPointReference comes from PathItemLocation, included in PathStep. 

Either we need to add the track_reference property in PathStep, use a supertype of PathStep, or otherwise refacto the addition of the track reference.

See the helpers.ts file from osrdConf.

Recette (SNCF)

Firefox

.

Synar:
Close #10805


emersion (approved):
Nice!

Math-R (approved):
lgtm

PR #10830, opened 2025-02-17
front: stop heating the room when selecting a train

emersion:
Spring is coming.

We were generating the PDF report (rendering and all) each time a train was selected. Instead, only generate the PDF when the user clicks the download button.

image


clarani (approved):

Uriel-Sautron (approved):
LGTM, tested

PR #10833, opened 2025-02-17
front: centered the space time chart on the projected train

### LINKED ISSUE #9039: front: stdcm: in new interface with debug data, GET is often not centered to the right time frame
### What happened?

When testing stdcm requests at random, the output train sometimes appeared to be missing from the space-time graph. But it was actually on a different day. For example the solution would span from 8:00 to 10:00 on Tuesday, the output would be centered on 8:00 to 10:00 on Monday.

The output should be centered on the new train

I don't have an exact request payload (but I can probably find one). Try random paths on realistic timetables.

What I **feel** improve the chances of it happening is having the output train span over two days, and not picking the first available date of the timetable. 

Local

_No response_

c1d34dd4be266e40e8932d789a3e7c73fef662b4

### LINKED ISSUE #9636: GET: The graph doesn't set on the right day
### What happened?

When you have trains in several days, the time-space graph doesn't always fit on the right day

It seems to always set on the day of the earlier train of the scenario.

https://github.com/user-attachments/assets/fedfa40e-f6d3-41f2-8d44-7b5b4e4efbec

Expected behavior is not completely set when the graph is fit. Waiting for more information to complete.

1. Open a project and a scenario
2. Import the two trains [train_schedules.20.json](https://github.com/user-attachments/files/17662746/train_schedules.20.json)
[train_schedules.json](https://github.com/user-attachments/files/17662745/train_schedules.json)
3. Try to use path of both trains to show time-space graph and see what happens

Recette (SNCF)

Firefox

e9f3b96

clarani:
closes #9636
closes #9039

### DIFF
--- a/front/src/applications/operationalStudies/views/SimulationResults.tsx
+++ b/front/src/applications/operationalStudies/views/SimulationResults.tsx
@@ -215,7 +215,7 @@ const SimulationResults = ({
                       height={manchetteWithSpaceTimeChartHeight - MANCHETTE_HEIGHT_DIFF}
                       handleTrainDrag={handleTrainDrag}
                       onTrainClick={(trainId) => dispatch(updateSelectedTrainId(trainId))}
-                      selectedProjection={trainIdUsedForProjection}
+                      selectedProjectionId={trainIdUsedForProjection}
                     />
                   )}
                 </div>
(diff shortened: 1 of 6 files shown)


theocrsb (approved):
LGTM and tested !

Math-R (approved):
lgtm

PR #10835, opened 2025-02-17
docker: use cache mount for npm install

emersion:
This should avoid re-downloading dependencies when re-building the container.

### DIFF
--- a/gateway/Dockerfile
+++ b/gateway/Dockerfile
@@ -74,12 +74,13 @@ ENV VITE_OSRD_GIT_DESCRIBE=${OSRD_GIT_DESCRIBE}
 COPY --from=front_src package.json package-lock.json /app/
 
 # If SKIP_FRONT is set to 1, skip the front build
-RUN if [ "$SKIP_FRONT" = "1" ]; then exit 0; else npm ci; fi
+RUN --mount=type=cache,target=/root/.npm \
+    if [ "$SKIP_FRONT" = "1" ]; then exit 0; else npm ci; fi
 
 # Generate the licenses file and build
 COPY --from=front_src . /app
 ENV NODE_OPTIONS="--max_old_space_size=4096"
-RUN if [ "$SKIP_FRONT" = "1" ]; then \ 
+RUN if [ "$SKIP_FRONT" = "1" ]; then \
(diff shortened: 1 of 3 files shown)


ghost (dismissed):
We should sed the cache out on GitHub Actions due to its instability. We already do it for other components.

Khoyo:
Yeah, you just need to adapt the pattern on the find to include `Dockerfile.playwright`

gateway/Dockerfile
```diff
@@ -74,12 +74,13 @@ ENV VITE_OSRD_GIT_DESCRIBE=${OSRD_GIT_DESCRIBE}
...
 # If SKIP_FRONT is set to 1, skip the front build
-RUN if [ "$SKIP_FRONT" = "1" ]; then exit 0; else npm ci; fi
+RUN --mount=type=cache,target=/root/.npm \
+    if [ "$SKIP_FRONT" = "1" ]; then exit 0; else npm ci; fi
 
 # Generate the licenses file and build
 COPY --from=front_src . /app
 ENV NODE_OPTIONS="--max_old_space_size=4096"
-RUN if [ "$SKIP_FRONT" = "1" ]; then \ 
+RUN if [ "$SKIP_FRONT" = "1" ]; then \
```
Khoyo, line 83: Doesn't this need the cache too?

emersion:
> We should sed the cache out on GitHub Actions due to its instability. We already do it for other components.

Oh right! I thought this was already performed for all Dockerfiles, but the `find` invocation was missing `Dockerfile.*`. Fixed!

gateway/Dockerfile
emersion: I don't believe so: the cache is used for downloaded packages, and building doesn't use the network.

Khoyo: Sure, but doesn't building need the downloaded packages?

emersion: Yes, but the installed packages are in `node_modules`. `~/.npm` contains a cache.

Khoyo: Alright then :)

Khoyo (approved):

kmer2016 (approved):

PR #10837, opened 2025-02-17
front: add speedlimits boundaries to simulation export to csv

### LINKED ISSUE #9346: simulation result as csv export missing speed limit transitions
### What happened?

When exporting a simulation result as csv, speed limit transitions aren't included in the data points.

Therefore, the MRSP isn't correct inside the csv.

Image

The csv should contain points for speed transitions: two identical lines except for the speed limit value (with position = the offset at which the speed limit changes) 

1. Create a scenario on small infra
2. Load this [train schedule](https://github.com/user-attachments/files/17377214/train_schedules.19.json)
3. Download the csv Image
4. In a sheet tool, you can see as I did that the MRSP looks weird

Dev (SNCF)

Firefox

3ac680e

Synar:
Close #9346

### DIFF
--- a/front/src/modules/simulationResult/SimulationResultExport/__tests__/utils.spec.ts
+++ b/front/src/modules/simulationResult/SimulationResultExport/__tests__/utils.spec.ts
@@ -1,32 +1,32 @@
 import { describe, it, expect } from 'vitest';
 
-import { findActualVmax } from '../utils';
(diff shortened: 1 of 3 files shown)


theocrsb (approved):
LGTM and tested

clarani (approved):
Code LGTM, not tested :white_check_mark:

PR #10840, opened 2025-02-18
editoast: add categories to rolling stock edit endpoint

hamz2a:
Part of https://github.com/OpenRailAssociation/osrd/issues/10574

- [x] Add the columns to edit endpoints
- [x] Check that `primary_category` is not included in `other_categories`.

### DIFF
--- a/editoast/src/models/rolling_stock_model.rs
+++ b/editoast/src/models/rolling_stock_model.rs
@@ -212,6 +212,15 @@ impl From<RollingStockModel> for RollingStock {
             electrical_power_startup_time: rolling_stock_model.electrical_power_startup_time,
             raise_pantograph_time: rolling_stock_model.raise_pantograph_time,
             supported_signaling_systems: rolling_stock_model.supported_signaling_systems,
+            primary_category: rolling_stock_model.primary_category.0,
+            other_categories: editoast_schemas::rolling_stock::RollingStockCategories(
+                rolling_stock_model
+                    .other_categories
+                    .0
+                    .into_iter()
+                    .map(|x| x.0)
+                    .collect::<Vec<_>>(),
+            ),
         }
     }
 }
@@ -241,6 +250,15 @@ impl From<RollingStock> for RollingStockModelChangeset {
             .electrical_power_startup_time(rolling_stock.electrical_power_startup_time)
             .raise_pantograph_time(rolling_stock.raise_pantograph_time)
             .supported_signaling_systems(rolling_stock.supported_signaling_systems)
+            .primary_category(RollingStockCategory(rolling_stock.primary_category))
+            .other_categories(RollingStockCategories(
+                rolling_stock
+                    .other_categories
+                    .0
+                    .into_iter()
+                    .map(RollingStockCategory)
+                    .collect::<Vec<_>>(),
+            ))
     }
 }
 
@@ -327,7 +345,7 @@ pub mod tests {
     }
 
     #[rstest]
-    async fn test_primary_category_is_unknown_with_other_categories_are_empty() {
+    async fn test_primary_category_with_empty_other_categories() {
         let db_pool = DbConnectionPoolV2::for_tests();
 
         let created_fast_rolling_stock =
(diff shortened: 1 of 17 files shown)


editoast/src/models/rolling_stock_model.rs
```diff
@@ -138,53 +138,78 @@ impl From<model::Error> for Error {
...
-            Some("raise_pantograph_time is required for electric rolling stocks".into());
-        return Err(error);
+    fn validate_primary_category(&self, validation_errors: &mut ValidationErrors) {
+        if let Some(primary_category) = &self.primary_category {
+            if let Some(other_categories) = &self.other_categories {
+                if other_categories
+                    .iter()
+                    .flatten()
+                    .collect::<Vec<_>>()
+                    .contains(&primary_category)
```
woshilapin: You should be able to use [`.any()`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any) to avoid collecting?

```suggestion
                if other_categories
                    .iter()
                    .flatten()
                    .any(|category| category == &primary_category)
```

editoast/src/models/rolling_stock_model.rs
```diff
@@ -138,53 +138,78 @@ impl From<model::Error> for Error {
...
+                {
+                    let mut error = ValidationError::new("primary_category");
+                    error.message = Some("The primary_category cannot be listed in other_categories for rolling stocks.".into(),);
```
woshilapin:
```suggestion
                    error.message = Some("The primary_category cannot be listed in other_categories for rolling stocks.".into());
```

I'm actually surprised the formatter didn't catch that. I've seen the formatter having problem with long string but it will [soon](https://doc.rust-lang.org/edition-guide/rust-2024/index.html) (in 2 days 🤞) be over with [the coming 2024's edition](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/rustfmt-formatting-fixes.html#long-strings-dont-prevent-formatting-expressions).

woshilapin (approved):
Nice bug fix!

editoast/src/models/rolling_stock_model.rs
hamz2a: Done.

editoast/src/models/rolling_stock_model.rs
hamz2a: Done.

editoast/src/models/rolling_stock_model.rs
```diff
@@ -212,6 +236,15 @@ impl From<RollingStockModel> for RollingStock {
             electrical_power_startup_time: rolling_stock_model.electrical_power_startup_time,
             raise_pantograph_time: rolling_stock_model.raise_pantograph_time,
             supported_signaling_systems: rolling_stock_model.supported_signaling_systems,
+            primary_category: rolling_stock_model.primary_category.0,
+            other_categories: editoast_schemas::rolling_stock::RollingStockCategories(
+                rolling_stock_model
+                    .other_categories
+                    .0
+                    .into_iter()
+                    .map(|x| x.0)
```
Synar: Nitpick: Correct me if I'm wrong, but I believe we could remove some of this boilerplate by implementing ops::{Deref, DerefMut} in RollingStockCategory and RollingStockCategories, as done elsewhere in the code.

Synar (approved):
Overall LGTM, just one nitpick ^^

editoast/src/models/rolling_stock_model.rs
hamz2a: Done.

front/src/modules/rollingStock/consts.ts
```diff
@@ -60,6 +60,8 @@ export const newRollingStockValues: RollingStockParametersValues = {
   basePowerClass: null,
   powerRestrictions: {},
   supportedSignalingSystems: DEFAULT_SIGNALING_SYSTEMS,
+  primary_category: 'FREIGHT_TRAIN',
```
SharglutDev, line 63: We don't want to default a value here. Maybe we can accept undefined for this property in `RollingStockParametersValues` type ? 
@Synar will do the proper changes to ensure we can't provide an undefined primary category.

Synar: Ah yes this indeed differs from what is done in my pr, but I don't think it's a big deal. I'll just overwrite this when rebasing.

Maymanaf (approved):

front/src/modules/rollingStock/consts.ts
SharglutDev: Totally fine by me :)

SharglutDev (approved):
Lgtm, thanks !

leovalais (approved):
Thanks for the fix!

PR #10792, opened 2025-02-13
editoast: migrations: add missing indexes on foreign keys

Khoyo:
Mostly generated with pgassistant, but can also be done with

```sql
WITH fkeys_without_indexes AS
  (SELECT conname AS fk_name,
          conrelid::regclass AS TABLE_NAME,
          a.attname AS COLUMN_NAME,
          n.nspname AS SCHEMA_NAME,
          condeferrable AS is_deferrable,
          confupdtype AS update_action,
          confdeltype AS delete_action,
          conkey
   FROM pg_constraint c
   JOIN pg_namespace n ON n.oid = c.connamespace
   JOIN pg_class r ON r.oid = c.conrelid
   JOIN unnest(c.conkey) AS col_num ON TRUE
   JOIN pg_attribute a ON a.attnum = col_num
   AND a.attrelid = r.oid
   LEFT JOIN pg_index i ON i.indrelid = r.oid
   AND col_num = ANY (i.indkey)
   WHERE c.contype = 'f'
     AND i.indexrelid IS NULL),
     suggested_indexes AS
  (SELECT SCHEMA_NAME,
          TABLE_NAME,
          COLUMN_NAME,
          'CREATE INDEX pga_idx_fk_' || TABLE_NAME || '_' || COLUMN_NAME || ' ON ' || SCHEMA_NAME || '.' || TABLE_NAME || '(' || COLUMN_NAME || ');' || 'ANALYZE ' || SCHEMA_NAME || '.' || TABLE_NAME || ';' AS pga_suggestion
   FROM fkeys_without_indexes)
SELECT SCHEMA_NAME,
       TABLE_NAME,
       COLUMN_NAME,
       pga_suggestion
FROM suggested_indexes
ORDER BY SCHEMA_NAME,
         TABLE_NAME,
         COLUMN_NAME;
```

PR #10842, opened 2025-02-18
Create paced trains model and add post and delete endpoints

### LINKED ISSUE #10888: editoast: create paced train model and add Post / Delete endpoints
### Description and goal

Write the migration to create the table in database and then write the first endpoints: 
- Post a list of paced trains ( ` POST /timetable/{id}/paced_trains`)
- Delete a list of paced trains (`DELETE /paced_train`)

CF : https://github.com/osrd-project/osrd-confidential/issues/779

### Acceptance criteria

- [ ] there is an endpoint that can add a list of paced trains
- [ ] There is an endpoint that can delete a list of paced trains

Part of

younesschrifi:
Close #10888


editoast/migrations/2025-02-13-115126_create_paced_trains/up.sql
woshilapin, line 1: You may want to take a look at https://github.com/OpenRailAssociation/osrd/pull/10792 to add the necessary indexes.

editoast/editoast_schemas/src/paced_train/paced_train_base.rs
woshilapin, line 1: nit: maybe this file can be inlined in `paced_train.rs`? Or do we expect more in the future in either one of the files?

editoast/editoast_schemas/src/paced_train.rs
woshilapin, line 1: `osrd_schemas` (the python module) is going to need an update too.

front/public/locales/fr/errors.json
woshilapin, line 1: `front/public/locales/en/errors.json` is going to need an update too.

editoast/src/views/timetable.rs
```diff
@@ -53,12 +60,14 @@ crate::routes! {
             "/train_schedules" => get,
             "/train_schedules" => train_schedule,
             "/conflicts" => conflicts,
+            "/paced_trains" => paced_train,
```
woshilapin, line 62: Don't we want also a `GET` method that would return a list (like we have for `/timetable/<id>/train_schedules`?

editoast/src/views/paced_train.rs
```diff
@@ -0,0 +1,128 @@
...
+use super::AuthenticationExt;
+use crate::views::train_schedule::BatchRequest;
+use crate::{
+    models::paced_train::{PacedTrain, PacedTrainChangeset},
+    views::AuthorizationError,
+};
+
+crate::routes! {
+    "/paced_train" => {
+        delete,
```
woshilapin, line 20: Don't we want also a `GET` method like we already have on `/train_schedule/<id>`? That would also allow for improving the integration test to verify after that the paced train is gone after the deletion.

editoast/src/models/paced_train.rs
```diff
@@ -0,0 +1,74 @@
...
+use editoast_schemas::train_schedule::PowerRestrictionItem;
+use editoast_schemas::train_schedule::ScheduleItem;
+use editoast_schemas::train_schedule::TrainScheduleOptions;
+
+use crate::models::prelude::*;
+
+#[derive(Debug, Clone, Model)]
+#[model(table = editoast_models::tables::paced_train)]
+#[model(gen(ops = crud, batch_ops = crd, list))]
+#[model(row(derive(diesel::QueryableByName)))]
```
leovalais: Unless you write an SQL query by hand, you don't need to derive this.

```suggestion
```

editoast/src/models/paced_train.rs
```diff
@@ -0,0 +1,74 @@
...
+use editoast_schemas::train_schedule::PathItem;
+use editoast_schemas::train_schedule::PowerRestrictionItem;
+use editoast_schemas::train_schedule::ScheduleItem;
+use editoast_schemas::train_schedule::TrainScheduleOptions;
+
+use crate::models::prelude::*;
+
+#[derive(Debug, Clone, Model)]
+#[model(table = editoast_models::tables::paced_train)]
+#[model(gen(ops = crud, batch_ops = crd, list))]
```
leovalais: Unless I'm wrong not all these operations are needed

editoast/src/views/timetable.rs
```diff
@@ -269,6 +278,50 @@ async fn train_schedule(
...
+        .check_roles([BuiltinRole::TimetableWrite].into())
+        .await
+        .map_err(AuthorizationError::AuthError)?;
+    if !authorized {
+        return Err(AuthorizationError::Forbidden.into());
+    }
+
+    let conn = &mut db_pool.get().await?;
+
+    TimetableWithTrains::retrieve_or_fail(conn, timetable_id, || TimetableError::NotFound {
```
woshilapin: If you only want to test for existence, you can use `exists` instead of `retrieve_or_fail`. Because it looks weird to use a `retrieve` but not storing the value.

editoast/src/models/paced_train.rs
```diff
@@ -0,0 +1,74 @@
...
+#[model(row(derive(diesel::QueryableByName)))]
+pub struct PacedTrain {
+    pub id: i64,
+    pub train_name: String,
+    pub labels: Vec<Option<String>>,
```
leovalais: I understand that the `Option` comes from Postgres, but since we don't allow `NULL` labels, can we have a plain `Vec<String>` instead? You can do the same trick as other models using `#[model(remote = "Vec<Option<Strng>>")]` and a custom local type `Labels` (if it doesn't already exist for train schedules).

(Model should handle this case eventually.)

editoast/src/views/paced_train.rs
```diff
@@ -0,0 +1,128 @@
...
+editoast_common::schemas! {
+    PacedTrainBase,
+}
+
+#[derive(Debug, Error, EditoastError)]
+#[editoast_error(base_id = "paced_train")]
+pub enum PacedTrainError {
+    #[error("{number} paced train(s) could not be found")]
+    #[editoast_error(status = 404)]
+    BatchPacedTrainNotFound { number: usize },
```
```suggestion
    BatchPacedTrainNotFound { count: usize },
```

nit: "number" may be mistaken for an ID

editoast/src/views/paced_train.rs
```diff
@@ -0,0 +1,128 @@
...
+) -> Result<impl IntoResponse> {
+    let authorized = auth
+        .check_roles([BuiltinRole::InfraRead, BuiltinRole::TimetableWrite].into())
+        .await
+        .map_err(AuthorizationError::AuthError)?;
+    if !authorized {
+        return Err(AuthorizationError::Forbidden.into());
+    }
+
+    use crate::models::DeleteBatch;
```
leovalais: You can `use crate::models::prelude::*;` for the whole module for consistency.

editoast/src/views/timetable.rs
```diff
@@ -53,12 +60,14 @@ crate::routes! {
...
             "/train_schedules" => train_schedule,
             "/conflicts" => conflicts,
+            "/paced_trains" => paced_train,
             &stdcm,
         },
     },
 }
 
 editoast_common::schemas! {
+    PacedTrainResult,
```
leovalais: nit: can it be added in the `schemas!` of the module it's defined in?

editoast/src/views/paced_train.rs
```diff
@@ -0,0 +1,128 @@
...
+    },
+}
+
+editoast_common::schemas! {
+    PacedTrainBase,
+}
+
+#[derive(Debug, Error, EditoastError)]
+#[editoast_error(base_id = "paced_train")]
+pub enum PacedTrainError {
```
```suggestion
enum PacedTrainError {
```

editoast/src/views/paced_train.rs
```diff
@@ -0,0 +1,128 @@
...
+    Database(#[from] editoast_models::model::Error),
+}
+
+#[derive(Debug, Clone, Serialize, ToSchema)]
+pub struct PacedTrainForm {
+    /// Timetable attached to the train schedule
+    pub timetable_id: Option<i64>,
+    #[serde(flatten)]
+    pub paced_train: PacedTrainBase,
+}
```
```suggestion
pub(in crate::views) struct PacedTrainForm {
    /// Timetable attached to the train schedule
    pub(in crate::views) timetable_id: Option<i64>,
    #[serde(flatten)]
    pub(in crate::views) paced_train: PacedTrainBase,
}
```

editoast/src/views/paced_train.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct PacedTrainResult {
+    id: i64,
+    timetable_id: i64,
+    #[serde(flatten)]
+    pub paced_train: PacedTrainBase,
+}
```
```suggestion
pub(in crate::views) struct PacedTrainResult {
    id: i64,
    timetable_id: i64,
    #[serde(flatten)]
    paced_train: PacedTrainBase,
}
```

editoast/src/views/train_schedule.rs
```diff
@@ -197,8 +197,8 @@ async fn get(
 }
 
 #[derive(Debug, Deserialize, ToSchema)]
-struct BatchRequest {
-    ids: HashSet<i64>,
+pub struct BatchRequest {
+    pub ids: HashSet<i64>,
 }
```
leovalais: Could this be duplicated in paced_trains.rs to avoid needing a visibility change? I think that for simple definitions like this, locality is more important than deduplication.

But maybe we'd better have a different module tree since there's several visibility changes needed?

leovalais (commented):
Cursory review.

I wonder if `mod paced_train` shouldn't be included in `mod timetable` since there's a bunch of visibility changes of existing definitions (that usually is a signal that something's wrong). (I know that we have `DELETE /paced_trains` and not `DELETE /timetable/paced_trains`, but URL path hierarchy and rust module trees are two different things that do not necessarily require correlation imo.)

Wdyt?

editoast/src/views/paced_train.rs
```diff
@@ -0,0 +1,128 @@
+use crate::error::Result;
+use axum::extract::Json;
+use axum::extract::State;
+use axum::{response::IntoResponse, Extension};
+use editoast_authz::BuiltinRole;
+use editoast_derive::EditoastError;
+use editoast_models::DbConnectionPoolV2;
+use editoast_schemas::{paced_train::PacedTrainBase, train_schedule::TrainScheduleBase};
```
woshilapin: nit: Can you expand the `use` with one per line please?

editoast/src/views/train_schedule.rs
woshilapin: Maybe this `struct` needs to be moved to `src/views` since it needs to be used in multiple sub-modules.

younesschrifi:
> Cursory review.
> 
> I wonder if `mod paced_train` shouldn't be included in `mod timetable` since there's a bunch of visibility changes of existing definitions (that usually is a signal that something's wrong). (I know that we have `DELETE /paced_trains` and not `DELETE /timetable/paced_trains`, but URL path hierarchy and rust module trees are two different things that do not necessarily require correlation imo.)
> 
> Wdyt?

 don't totally agree. We'd like to use paced trains in cases where the timetable is decorelated, as will perhaps be the case with train schedules. Maybe there won't be such a case, and if it's not the case I think it's a good idea to put paced train in timetable, but we'd have to put train schedule there too, wouldn't we?

leovalais:
> We'd like to use paced trains in cases where the timetable is decorelated

I'm not saying we should change the URL scheme, `DELETE /paced_train` is fine for the reasons you mentioned. However nothing prevents us from putting `mod paced_train` in `mod timetable` as it can be convenient to share some definitions without compromising visibility (and locality).

What I'm trying to say is that the module tree doesn't necessarily have to match our URLs paths. We're currently a bit forced to do that because of the `routes!` macro, but if we had all the paths defined in a single file, we'd have no constraint on module trees whatsoever.

editoast/src/views/timetable.rs
```diff
@@ -269,6 +278,50 @@ async fn train_schedule(
...
+    })
+    .await?;
+    let changesets: Vec<PacedTrainChangeset> = paced_trains
+        .into_iter()
+        .map(|pt| PacedTrainForm {
+            timetable_id: Some(timetable_id),
+            paced_train: pt,
+        })
+        .map_into()
+        .collect();
```
woshilapin: I'm not sure I agree with the existence of `struct PacedTrainForm`. With the suffix `Form`, I expect a payload that the `front` is sending. But in this case, the `front` creates paced trains with a `PacedTrainBase` on an endpoint that have a timetable identifier: `PacedTrainForm` is only constructed from these only on the backend.

Instead, maybe you can just do the following?

```suggestion
    let changesets = paced_trains
        .into_iter()
        .map(PacedTrainChangeset::from)
        .map(|cs| cs.timetable_id(timetable_id))
        .collect::<Vec<_>>();
```

Then you should be able to remove the `struct PacedTrainForm` and all related code.

tests/tests/test_timetable.py
```diff
@@ -247,3 +247,42 @@ def _get_train_schedule_simulation_response(
...
+            "train_name": "with_stop",
+            "duration": "PT2H",
+            "step": "PT15M",
+        }
+    ]
+    pt_response = requests.post(f"{EDITOAST_URL}/timetable/{timetable_id}/paced_train", json=payload)
+    pt_response.raise_for_status()
+    paced_train_id = pt_response.json()[0]["id"]
+    r = requests.delete(f"{EDITOAST_URL}paced_train/", json={"ids": [paced_train_id]})
+    r.raise_for_status()
```
woshilapin: If a `GET /paced_train/<id>` endpoint were to exist, you could actually verify that the deletion worked.

woshilapin (commented):
A first review, a few things are missing like indexes in the DB migration or the update of `osrd_schemas`.

editoast/src/views/train_schedule.rs
woshilapin: Since it's relatively agnostic of which identifier it's about, how about moving it in the parent module instead (see https://github.com/OpenRailAssociation/osrd/pull/10842#discussion_r1965229758).

editoast/src/views/timetable.rs
younesschrifi: Yes it's comming in a different PR

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -3736,6 +3752,27 @@ export type TrainScheduleBase = {
   start_time: string;
   train_name: string;
 };
+export type PacedTrainBase = TrainScheduleBase & {
+  duration: string;
+  step: string;
```
SharglutDev: Would it be possible to add doc explaining we expect an ISO format and not something like "120min"

editoast/src/views/train_schedule.rs
leovalais: I don't mind it either way, but if moved to `mod views` directly, I think it should be inlined every time in the OpenAPI (if not already) as "BatchRequest" isn't very explanatory. (Maybe `IdList` would be more suitable then?)

editoast/src/views/timetable.rs
younesschrifi: Nice catch :)

tests/tests/test_timetable.py
younesschrifi: You are right, I'll write the integration tests in another PR when all the endpoints will be written to test all cases. Wdyt?

editoast/src/views/paced_train.rs
younesschrifi: Yes it will be done in another PR :)

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -3736,6 +3752,29 @@ export type TrainScheduleBase = {
   start_time: string;
   train_name: string;
 };
+export type PacedTrainBase = TrainScheduleBase & {
+  /** Duration of the mission, an ISO 8601 format is expected */
+  duration: string;
+  /** Time between two trains, an ISO 8601 format is expected */
+  step: string;
+};
```
SharglutDev: We should use the same name convention for docs
```suggestion
export type PacedTrainBase = TrainScheduleBase & {
  /** Duration of the paced train (or service), an ISO 8601 format is expected */
  duration: string;
  /** Time between two occurrences, an ISO 8601 format is expected */
  step: string;
};
```

editoast/editoast_schemas/src/paced_train.rs
```diff
@@ -0,0 +1,21 @@
...
+use serde::Deserialize;
+use serde::Serialize;
+use utoipa::ToSchema;
+
+editoast_common::schemas! {
+    PacedTrainBase,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct PacedTrainBase {
```
Wadjetz: It look like a duplicate with the file in `/src/paced_train/paced_train.rs`

editoast/src/views/timetable.rs
```diff
@@ -269,6 +276,44 @@ async fn train_schedule(
...
+        .check_roles([BuiltinRole::TimetableWrite].into())
+        .await
+        .map_err(AuthorizationError::AuthError)?;
+    if !authorized {
+        return Err(AuthorizationError::Forbidden.into());
+    }
+
+    let conn = &mut db_pool.get().await?;
+
+    Timetable::exists(conn, timetable_id).await?;
```
Wadjetz: I think exists returns a boolean with the result
```suggestion
    let is_timetable_exists = Timetable::exists(conn, timetable_id).await?;
    if !is_timetable_exists {
        return Err(TimetableError::NotFound { timetable_id: timetable_id }.into())
    }
```

editoast/migrations/2025-02-13-115126_create_paced_trains/up.sql
```diff
@@ -0,0 +1,22 @@
...
+    constraint_distribution smallint NOT NULL,
+    speed_limit_tag varchar(128),
+    power_restrictions jsonb NOT NULL,
+    options jsonb NOT NULL,
+    duration interval NOT NULL,
+    step interval NOT NULL
+);
+
+CREATE INDEX pga_idx_fk_paced_train_timetable_id ON public.paced_train (timetable_id);
+ANALYZE public.paced_train;
```
Khoyo: nit: That analyze isn't necessary, since we've just created the table.

SharglutDev (approved):
Lgtm (not tested)

python/railjson_generator/railjson_generator/schema/simulation/paced_train.py
Khoyo, line 1: The simulation part of railjson_simulator is there to generate "sim raijson", that is now only accepted by core CLI (and is obsolete since it's stuck on TSv1, we will see if we keep it once TSv1 apis get dropped).

editoast/editoast_schemas/src/paced_train.rs
Khoyo: No. Sim railjson is only used by core cli, it should only allow for core supported features.   We should probably have a split in editoast between what's part of general editoast "schemas", and what's actually part of railjson

editoast/src/views/paced_train.rs
```diff
@@ -0,0 +1,183 @@
...
+)]
+async fn delete(
+    State(db_pool): State<DbConnectionPoolV2>,
+    Extension(auth): AuthenticationExt,
+    Json(ListId {
+        ids: paced_train_ids,
+    }): Json<ListId>,
+) -> Result<impl IntoResponse> {
+    let authorized = auth
+        .check_roles([BuiltinRole::InfraRead, BuiltinRole::TimetableWrite].into())
```
flomonster:
```suggestion
        .check_roles([BuiltinRole::TimetableWrite].into())
```

flomonster (approved):

Wadjetz (approved):

PR #10845, opened 2025-02-18
front: drop formatToIsoDate()

emersion:
_See individual commits._

This function was a bit weird, because it had two unrelated purposes: format a human-readable string with a fixed frenchy format, or format a machine-readable string for API purposes. The function also accepted strings and numbers as input (implicitly converting these to dates), making it so callers are less type-safe.


Synar (approved):

SharglutDev (approved):
Lgtm and tested

PR #10847, opened 2025-02-18
editoast: stdcm_log: store requests that cause core errors

### LINKED ISSUE #10841: editoast: stdcm request is not logged when core fails
### What happened?

When core fails on an STDCM request, we don't write the stdcm_log. That's not ideal, as those tend to be the interesting request...

The request is logged when core fails

Make an STDCM request that fails, inspect the stdcm logs table

Production (SNCF), Local

Firefox

3e1768d

Khoyo:
Fixes #10841.

### DIFF
--- a/editoast/src/models/stdcm_log.rs
+++ b/editoast/src/models/stdcm_log.rs
@@ -42,7 +42,7 @@ impl StdcmLog {
         let stdcm_log_changeset = StdcmLog::changeset()
             .trace_id(trace_id)
             .request(request)
-            .response(response.clone())
+            .response(response)
             .user_id(user_id);
         if let Err(e) = stdcm_log_changeset.create(&mut conn).await {
             tracing::error!("Failed during log operation: {e}");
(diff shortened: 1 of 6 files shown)


editoast/src/models/stdcm_log.rs
```diff
@@ -8,10 +8,23 @@ use utoipa::ToSchema;
...
+
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "snake_case")]
+// We accepted the difference of memory size taken by variants
+// Since there is only on success and others are error cases
+#[allow(clippy::large_enum_variant)]
+pub enum StdcmResponseOrError {
+    #[schema(value_type = StdcmResponse)]
+    Response(Response),
+    RequestError(InternalError),
```
```suggestion
    RequestError(serde_json::Value),
```

for the split in crates plz

leovalais (approved):
lgtm

flomonster (approved):
LGTM

PR #10848, opened 2025-02-18
editoast: fix glyph generation on docker on mac-arm

Tristramg:


SharglutDev (approved):
Works on a mac M3 Pro Sequoia 15.0.1

theocrsb (approved):
Tested

editoast/Dockerfile
```diff
@@ -23,12 +23,11 @@ COPY --from=static_assets . /assets
...
 #######################
-FROM ubuntu:latest AS editoast_assets
-RUN apt update && apt install --yes wget
-RUN wget https://github.com/flother/spreet/releases/download/v0.11.0/spreet-x86_64-unknown-linux-musl.tar.gz
-RUN tar xvf spreet-x86_64-unknown-linux-musl.tar.gz --directory /usr/bin
-RUN wget http://github.com/stadiamaps/sdf_font_tools/releases/download/cli-v1.4.2/build_pbf_glyphs.x86_64-unknown-linux-gnu -O /usr/bin/build_pbf_glyphs
-RUN chmod +x /usr/bin/build_pbf_glyphs
+FROM chef AS editoast_assets
+# build_pbf_glyphs requires a c++ compiler
+RUN apk add build-base
```
emersion: `--no-cache` should be used to avoid working with an outdated local cache (which would result in 404s).

woshilapin:
Just a suggestion, do you think this is where we could try `binstall`. `cargo-binstall` does ship a musl release (which should not be impacted by the problem we have?), and then, maybe (maybe ?) `cargo binstall` will be able to realize if it needs to build or if it can download from Github's release?

Because to be honest, `editoast` build is already the longer one and using `cargo install` will make it even longer.

flomonster (approved):

flomonster:
> Because to be honest, `editoast` build is already the longer one and using `cargo install` will make it even longer.

I don't think it's going to make the editoast build any longer. The internship can be done completely in parallel with editoast.

editoast/Dockerfile
flomonster: Done

PR #10851, opened 2025-02-18
core: stdcm: fix initialization of fixed time points in postproc

eckter:
Fix two issues: we'd accidentally delete the last point (for allowance), and the points at the *start* of allowances can cause issues.

Fix at least some cases of "mismatch during post-processing".

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/PostProcessingSimulation.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/PostProcessingSimulation.kt
@@ -118,13 +118,12 @@ fun buildFinalEnvelope(
                     isMareco,
                 )
             }
-            postProcessingLogger.info(
-                "Conflict when running final stdcm simulation at offset {}, adding a fixed time point",
-                conflictOffset
-            )
-            fixedPoints.add(
+            val newPoint =
                 makeFixedPoint(fixedPoints, edges, conflictOffset, pathLength, updatedTimeData)
+            postProcessingLogger.info(
+                "Conflict when running final stdcm simulation at offset $conflictOffset, adding a fixed time point: $newPoint",
             )
+            fixedPoints.add(newPoint)
         } catch (e: OSRDError) {
             if (e.osrdErrorType == ErrorType.AllowanceConvergenceTooMuchTime) {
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/PostProcessingSimulation.kt
```diff
@@ -198,11 +199,28 @@ private fun initFixedPoints(
...
         prevStopTime += stop.duration
     }
+
+    // Add one point at the end to match the standard allowance (if any)
     if (hasStandardAllowance && res.none { it.offset == length })
         res.add(makeFixedPoint(res, edges, length, length, updatedTimeData, 0.0))
 
     // Add points at the end of each engineering allowance
-    fun addFixedPointAvoidingDuplicates(offset: Distance) {
+    var prevEdgeLength = 0.meters
```
bougue-pe: How about renaming it to:
```suggestion
    var edgeStartOffset = 0.meters
```

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/PostProcessingSimulation.kt
```diff
@@ -215,20 +233,8 @@ private fun initFixedPoints(
...
-            // So we remove any point in the current allowance range.
-            res.removeIf { it.offset.distance > engineeringAllowanceStart && it.stopTime == null }
-
-            addFixedPointAvoidingDuplicates(prevEdgeLength)
-            addFixedPointAvoidingDuplicates(engineeringAllowanceStart)
-        }
-        prevEdgeLength += edge.length.distance
-    }
+    logger.info("initial fixed time points:")
+    for (p in res) logger.info("    $p")
```
bougue-pe, line 237: Do you want to keep this log?

bougue-pe (approved):
If you can easily add a test, it's welcome, but LGTM as-is.

Erashin (approved):

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/PostProcessingSimulation.kt
eckter: It can be useful and it shouldn't add too many lines

PR #10852, opened 2025-02-18
front: move the deployment settings in high level context

### LINKED ISSUE #10834: front: logo is blinking on deployed environments
### What happened?

The LMR or Horizon logos are blinking when changing page

No blinking

1. Go to the home page
2. Click on a project
3. See the logo blinking in the navbar

Recette (SNCF), Local

Firefox

.

clarani:
closes #10834

### DIFF
--- a/front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
+++ b/front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
@@ -46,7 +46,7 @@ const StcdmResults = ({
 }: StcdmResultsProps) => {
   const infraId = useInfraID();
   const { t } = useTranslation('stdcm', { keyPrefix: 'simulation.results' });
-  const { stdcmName } = useDeploymentSettings();
+  const deploymentSettings = useDeploymentSettings();
 
   const selectedSimulation = useSelector(getSelectedSimulation);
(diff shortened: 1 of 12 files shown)


front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
```diff
@@ -111,9 +111,10 @@ const StcdmResults = ({
                             consist={selectedSimulation.inputs.consist}
                             simulationReportSheetNumber={simulationReportSheetNumber}
                             operationalPointsList={operationalPointsList}
+                            deploymentSettings={deploymentSettings}
```
kmer2016: I think it is better to pass only the `stdcmSimulationSheetLog` field to `SimulationReportSheet` component   instead of the entire `deploymentSettings` object.

kmer2016 (commented):
Small suggestion, but very nice solution !

front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
clarani: You're right, it will be cleaner like this ! Fixed :)

kmer2016 (approved):

PR #10853, opened 2025-02-18
front: improves common conf selectors management

jacomyal:
Fixes osrd-project/osrd-confidential#882

This commit:
- Moves some properties from `OsrdConfState` to `OsrdStdcmConfState`
- In STDCM, ensures `OsrdStdcmConfState` is valid before mounting the app, and improves typings so that `infraID`, `timetableID` and `searchDatetimeWindow` are never null in the code
- Improves the way we must retrieve the proper conf selectors from the application codebases, and removes all remaining calls to `useOsrdConfSelectors` from the application codebases
- Improves common selectors generation, to make it easier later to completely remove all selectors from the context later

### DIFF
--- a/front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
+++ b/front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
@@ -1,10 +1,7 @@
 /* eslint-disable no-restricted-syntax, no-await-in-loop */
 import { useEffect, useState, type Dispatch, type SetStateAction, useMemo } from 'react';
 
-import { useSelector } from 'react-redux';
-
 import { osrdEditoastApi, type SimulationSummaryResult } from 'common/api/osrdEditoastApi';
-import { useOsrdConfSelectors } from 'common/osrdContext';
 import type { TrainScheduleWithDetails } from 'modules/trainschedule/components/Timetable/types';
 import type {
   TrainId,
@@ -24,6 +21,7 @@ const BATCH_SIZE = 10;
 
 type UseLazyLoadTrainsProps = {
   infraId?: number;
+  electricalProfileSetId?: number;
   trainIdsToFetch?: TrainId[];
   trainSchedules?: TrainScheduleResultWithTrainId[];
   setTrainIdsToFetch?: Dispatch<SetStateAction<TrainId[] | undefined>>;
@@ -39,13 +37,11 @@ type UseLazyLoadTrainsProps = {
  */
 const useLazyLoadTrains = ({
   infraId,
+  electricalProfileSetId,
   trainIdsToFetch,
   trainSchedules,
   setTrainIdsToProject,
 }: UseLazyLoadTrainsProps) => {
-  const { getElectricalProfileSetId } = useOsrdConfSelectors();
-  const electricalProfileSetId = useSelector(getElectricalProfileSetId);
-
   const [trainScheduleSummariesById, setTrainScheduleSummariesById] = useState<
     Map<TrainId, TrainScheduleWithDetails>
   >(new Map());
--- a/front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
+++ b/front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
@@ -13,10 +13,10 @@ import {
 } from 'applications/stdcm/utils/formatSimulationReportSheet';
 import { hasConflicts, hasResults } from 'applications/stdcm/utils/simulationOutputUtils';
 import DefaultBaseMap from 'common/Map/DefaultBaseMap';
-import { useInfraID } from 'common/osrdContext';
 import {
   getRetainedSimulationIndex,
   getSelectedSimulation,
+  getStdcmInfraID,
 } from 'reducers/osrdconf/stdcmConf/selectors';
 import useDeploymentSettings from 'utils/hooks/useDeploymentSettings';
 
@@ -44,7 +44,7 @@ const StcdmResults = ({
   buttonsVisible,
   showStatusBanner,
 }: StcdmResultsProps) => {
-  const infraId = useInfraID()!;
+  const infraId = useSelector(getStdcmInfraID);
 
   const { t } = useTranslation('stdcm', { keyPrefix: 'simulation.results' });
   const deploymentSettings = useDeploymentSettings();
--- a/front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
+++ b/front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
@@ -44,7 +44,8 @@ const StcdmResults = ({
   buttonsVisible,
   showStatusBanner,
 }: StcdmResultsProps) => {
-  const infraId = useInfraID();
+  const infraId = useInfraID()!;
+
   const { t } = useTranslation('stdcm', { keyPrefix: 'simulation.results' });
   const deploymentSettings = useDeploymentSettings();
 
--- a/front/src/applications/stdcm/hooks/useProjectedTrainsForStdcm.ts
+++ b/front/src/applications/stdcm/hooks/useProjectedTrainsForStdcm.ts
@@ -55,19 +55,16 @@ const keepTrainsRunningDuringStdcm = (
 };
 
 const useProjectedTrainsForStdcm = (stdcmResponse?: StdcmSuccessResponse) => {
-  const infraId = useInfraID();
   const { getTimetableID } = useOsrdConfSelectors();
-  const timetableId = useSelector(getTimetableID);
+  const infraId = useInfraID()!;
+  const timetableId = useSelector(getTimetableID)!;
 
   const [spaceTimeData, setSpaceTimeData] = useState<TrainSpaceTimeData[]>([]);
   const [trainIdsToProject, setTrainIdsToProject] = useState<Set<TrainId>>(new Set());
 
-  const { data: timetable } = osrdEditoastApi.endpoints.getAllTimetableByIdTrainSchedules.useQuery(
-    { timetableId: timetableId! },
-    {
-      skip: !timetableId,
-    }
-  );
+  const { data: timetable } = osrdEditoastApi.endpoints.getAllTimetableByIdTrainSchedules.useQuery({
+    timetableId,
+  });
 
   const trainIds = useMemo(() => timetable?.map((t) => t.id) || [], [timetable]);
 
@@ -132,7 +129,7 @@ const useProjectedTrainsForStdcm = (stdcmResponse?: StdcmSuccessResponse) => {
     setSpaceTimeData(newSpaceTimeData);
   }, [projectedTrainsById]);
 
-  if (!infraId || !stdcmResponse) return null;
+  if (!stdcmResponse) return null;
 
   return {
     spaceTimeData,
--- a/front/src/applications/stdcm/hooks/useProjectedTrainsForStdcm.ts
+++ b/front/src/applications/stdcm/hooks/useProjectedTrainsForStdcm.ts
@@ -6,9 +6,13 @@ import useLazyLoadTrains from 'applications/operationalStudies/hooks/useLazyLoad
 import type { TrainSpaceTimeData } from 'applications/operationalStudies/types';
 import type { StdcmSuccessResponse } from 'applications/stdcm/types';
 import { osrdEditoastApi } from 'common/api/osrdEditoastApi';
-import { useInfraID, useOsrdConfSelectors } from 'common/osrdContext';
 import useLazyProjectTrains from 'modules/simulationResult/components/SpaceTimeChart/useLazyProjectTrains';
 import type { TrainScheduleWithDetails } from 'modules/trainschedule/components/Timetable/types';
+import {
+  getStdcmElectricalProfileSetId,
+  getStdcmInfraID,
+  getStdcmTimetableID,
+} from 'reducers/osrdconf/stdcmConf/selectors';
 import type { TrainId, TrainScheduleResultWithTrainId } from 'reducers/osrdconf/types';
 import { Duration, addDurationToDate } from 'utils/duration';
 import { formatEditoastTrainIdToTrainScheduleId } from 'utils/trainId';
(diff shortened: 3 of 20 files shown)


front/src/applications/stdcm/views/StdcmView.tsx
```diff
@@ -237,4 +242,42 @@ const StdcmView = () => {
...
+    }
+  }, [isDebugMode]);
+
+  // If we've got an error during the loading of the stdcm env which is not the "no config error" message,
+  // we let the error boundary manage it
+  if (error && error.message !== NO_CONFIG_FOUND_MSG) throw error;
+
+  // When the STDCM environment is being loaded and the conf is not valid, we do not mount the actual STDCM
+  // view content yet:
+  if (loading && !isStdcmConfValid) return null;
```
SharglutDev, line 269: Is there a point to return null if `!isStdcmConfValid` if we throw after anyway ?

front/src/applications/stdcm/views/StdcmView.tsx
```diff
@@ -237,4 +242,42 @@ const StdcmView = () => {
   );
 };
 
+const StdcmView = () => {
+  const [isDebugMode, setIsDebugMode] = useState(false);
+  const { loading, error, loadStdcmEnvironment } = useStdcmEnvironment();
+  const stdcmConf = useSelector(getStdcmConf);
```
SharglutDev, line 250: nit : Maybe we could pass `stdcmconf` as prop to avoid recalling it in `StdcmViewContent` ?

front/src/applications/stdcm/hooks/useStdcmResults.ts
```diff
@@ -27,7 +27,7 @@ const useStdcmResults = (
   stdcmTrainResult: TrainScheduleResult | undefined,
   setPathProperties: (pathProperties?: StdcmPathProperties) => void
 ) => {
-  const infraId = useInfraID();
+  const infraId = useInfraID()!;
```
SharglutDev: Can't we have a way where typescript infer that we are in stdcm and that `infraId` will be defined here ?

front/src/reducers/osrdconf/stdcmConf/selectors.ts
```diff
@@ -70,15 +69,31 @@ const buildStdcmConfSelectors = () => {
     getSelectedSimulationIndex,
     getSelectedSimulation,
     getRetainedSimulationIndex: makeOsrdConfSelector('retainedSimulationIndex'),
+    getWorkScheduleGroupId: makeOsrdConfSelector('workScheduleGroupId'),
+
+    // For some selectors, if data were missing, errors would have been thrown earlier, at startup.
+    // The useStdcmEnv hook ensures that.
+    getSearchDatetimeWindow: makeOsrdConfSelector('searchDatetimeWindow', { nonNullable: true }),
```
SharglutDev, line 76: In the `checkStdcmConf` function, we also use `searchDatetimeWindow`, we could maybe update the code overthere as well.

Same in `isArrivalDateInSearchTimeWindow` function, you can probably remove the `undefined` check

front/src/applications/stdcm/hooks/useProjectedTrainsForStdcm.ts
```diff
@@ -57,18 +57,16 @@ const keepTrainsRunningDuringStdcm = (
...
   const [spaceTimeData, setSpaceTimeData] = useState<TrainSpaceTimeData[]>([]);
   const [trainIdsToProject, setTrainIdsToProject] = useState<Set<TrainId>>(new Set());
 
   const { data: timetable } = osrdEditoastApi.endpoints.getTimetableById.useQuery(
-    { id: timetableId! },
-    {
-      skip: !timetableId,
-    }
+    { id: timetableId },
+    { skip: false }
```
SharglutDev: We could just remove the `skip` ?

SharglutDev (commented):
Thank you for the PR, it looks very nice ! Review only the first commit for now (sorry if there are comments irrelevant that are solved in the 2nd one)

front/src/applications/stdcm/views/StdcmView.tsx
jacomyal: Yes:
- `loading && !isStdcmConfValid` is the normal expected default setup. It is like this until the conf state is properly initialized.
- `loading && isStdcmConfValid` can happen when transitionning between conf states. In which case, a loader is rendered by `StdcmViewContent`
- `!loading && isStdcmConfValid` is what happens normally, when we simply display the expected UI
- `!loading && !isStdcmConfValid`, though, is unexpected, and is the only case where we throw an error

front/src/applications/stdcm/hooks/useStdcmResults.ts
jacomyal: That's exactly what the next commit is about :)

front/src/reducers/osrdconf/stdcmConf/selectors.ts
jacomyal: Hum, at the moment, I wouldn't dare remove those checks from these helper functions: I only made sure that in all the components mounted under `StdcmViewContent`, the conditions for `isStdcmConfValid` are respected. But I also now that at some point, before the proper initialization of the state, the conditions are not met. And some devs might want to call these helper functions in some place that is not under `StdcmViewContent`.

I think at some point, these new "better typings" should actually be enforced directly in `OsrdStdcmConfState` (which cannot be cleanly done yet, because of the shared part of the state). At that point, yes, I think we can remove all the checks everywhere.

front/src/applications/stdcm/views/StdcmView.tsx
```diff
@@ -237,4 +242,42 @@ const StdcmView = () => {
...
+  // we let the error boundary manage it
+  if (error && error.message !== NO_CONFIG_FOUND_MSG) throw error;
+
+  // When the STDCM environment is being loaded and the conf is not valid, we do not mount the actual STDCM
+  // view content yet:
+  if (loading && !isStdcmConfValid) return null;
+
+  if (!isStdcmConfValid) {
+    throw new Error('STDCM conf is not valid.');
+  }
```
clarani, line 273: With this check, if we have no config, then I believe that the stdcmConf will not be valid right ? So will we display the `StdcmEmptyConfigError` ?

front/src/applications/stdcm/views/StdcmView.tsx
SharglutDev: Ok got it

front/src/reducers/osrdconf/stdcmConf/selectors.ts
SharglutDev: Fine by me :)

front/src/applications/stdcm/views/StdcmView.tsx
clarani: Hum currently, while loading, we display the header and the form, with a header above. Do you think we can keep this behavior ? Or at least the header and the loader ?

front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
```diff
@@ -43,8 +43,7 @@ const useLazyLoadTrains = ({
   trainSchedules,
   setTrainIdsToProject,
 }: UseLazyLoadTrainsProps) => {
-  const { getElectricalProfileSetId } = useOsrdConfSelectors();
-  const electricalProfileSetId = useSelector(getElectricalProfileSetId);
+  const electricalProfileSetId = useSelector(getOperationalStudiesElectricalProfileSetId);
```
clarani: This one is tricky, but this hook is used actually in stdcm too

front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
```diff
@@ -44,7 +44,8 @@ const StcdmResults = ({
   buttonsVisible,
   showStatusBanner,
 }: StcdmResultsProps) => {
-  const infraId = useInfraID();
+  const infraId = useInfraID()!;
```
clarani: Why not this ?
```suggestion
  const infraId = useSelector(getStdcmInfraID);
```

front/src/applications/stdcm/hooks/useStdcmResults.ts
clarani:
```suggestion
  const infraId = useSelector(getStdcmInfraID);
```

front/src/reducers/osrdconf/osrdConfCommon/selectors.ts
```diff
@@ -1,24 +1,55 @@
...
-    getElectricalProfileSetId: makeOsrdConfSelector('electricalProfileSetId'),
-    getWorkScheduleGroupId: makeOsrdConfSelector('workScheduleGroupId'),
-    getSearchDatetimeWindow: makeOsrdConfSelector('searchDatetimeWindow'),
-    getRollingStockID: makeOsrdConfSelector('rollingStockID'),
-    getSpeedLimitByTag: makeOsrdConfSelector('speedLimitByTag'),
-  };
+type Selectors<Name extends string> = {
+  [Result in keyof SelectorResults as `get${Name}${string & Result}`]: (
+    state: RootState
+  ) => SelectorResults[Result];
```
SharglutDev, line 22: This and `buildCommonConfSelectors` function are very nice ! But this could maybe use some doc to explain what it does

front/src/reducers/osrdconf/osrdConfCommon/selectors.ts
```diff
@@ -1,24 +1,55 @@
...
 };
 
+type CommonSelectors<Name extends string> = Selectors<Name> & InfraStateSelectors<Name>;
+
+function buildCommonConfSelectors(
```
SharglutDev, line 35: What is the plan for that ? When would be a good time to use it for operational studies ?

In the commit message, you say you removed the call of `useOsrdConfSelectors` but it remains in some components.

clarani (commented):
Nice work, thanks :pray:

front/src/reducers/osrdconf/osrdConfCommon/selectors.ts
jacomyal: Good question! First, about your second observation: I removed all calls to `useOsrdConfSelectors` from the `src/applications` folder, because in all cases there were now a selector with a better name to import instead. But there are still various common components, that can be used in either STDCM or Operational Studies, and that depend on that common part of the state.

There is a big problem with that hook `useOsrdConfSelectors`: It says it returns [`StdcmConfSelectors | OperationalStudiesConfSelectors`](https://github.com/OpenRailAssociation/osrd/blob/477b988345f5e1aed800da7f7dcdc3f73f5e0ebd/front/src/reducers/osrdconf/osrdConfCommon/index.ts#L69), when it actually returns a [`StdcmConfSelectors | OperationalStudiesConfSelectors | MapViewerSelectors | EditorSelectors`](https://github.com/OpenRailAssociation/osrd/blob/ac74e97736df917f9331b651fa916931fda2757f/front/src/common/osrdContext.tsx#L13). So, a dev that doesn't know that the "common part of the conf state" is only common to STDCM and Operational Studies might try to use it somewhere else at some point. It doesn't feel very clear, nor very safe.

My recommendation would be:
- In each common component that depends on it, we move what it reads from it to the props.
- Now, we can safely remove that hook at all. We can probably even remove the selectors from the context at all.
- Finally, we use the code I added to `buildCommonConfSelectors` to add clear prefixes to the selectors, instead of doing it [by hand](https://github.com/OpenRailAssociation/osrd/blob/3140e65a493f3d753931fbc512834be93523b288/front/src/reducers/osrdconf/operationalStudiesConf/selectors.ts#L64-L71). Then, it becomes easier for new devs to understand when they should use `getOperationalStudiesTimetableID` or `getStdcmTimetableID`

We might also simply remove `buildCommonConfSelectors` at all, actually. Especially since they are no more exactly the same (since some things are nullable on one side and not in the other).

front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
jacomyal: Oh, good catch :) I'll move `electricalProfileSetId` to the props, then.

front/src/applications/stdcm/views/StdcmView.tsx
jacomyal: Actually, now, if I remove that check and at some point the STDCM conf is not valid, then the call to `useStdcm` from `StdcmViewContent` will probably fail at some point. For instance, [this line](https://github.com/OpenRailAssociation/osrd/blob/3140e65a493f3d753931fbc512834be93523b288/front/src/applications/stdcm/hooks/useStdcm.ts#L56) will throw a `Value timetableID of RootState is undefined, while the selector should remain non-nullable` error (now thrown by [`makeSubSelector`](https://github.com/OpenRailAssociation/osrd/blob/41fa8386154fb31939c4f0162ee831f9b3249dc3/front/src/utils/selectors.ts#L26-L28)).

front/src/reducers/osrdconf/osrdConfCommon/selectors.ts
SharglutDev: Very clear thanks :)

front/src/reducers/osrdconf/osrdConfCommon/selectors.ts
jacomyal: You're right, I added some comments.

front/src/applications/operationalStudies/hooks/useLazyLoadTrains.ts
jacomyal: Fixed

front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
jacomyal: Fixed

front/src/applications/stdcm/hooks/useStdcmResults.ts
jacomyal: Fixed

clarani (approved):
LGTM :white_check_mark:

front/src/applications/stdcm/views/StdcmView.tsx
clarani: OK :+1:

SharglutDev (approved):
Lgtm and tested as well :)

PR #10857, opened 2025-02-19
core: prioritize signaling system during pathfinding

### LINKED ISSUE #10296: core: prioritize signaling systems for blocks of pathfinding
Priorities: ETCS>TVM430>TVM300>BAL>BAPR

bougue-pe:
When everything else is equal, use the preferred signaling system.
Also migrate pathfinding signaling tests to API v2 (and add new tests).

🔍 please review by commit.

Fix #10296

### DIFF
--- a/core/kt-osrd-signaling/src/main/kotlin/fr/sncf/osrd/signaling/impl/SigSystemManagerImpl.kt
+++ b/core/kt-osrd-signaling/src/main/kotlin/fr/sncf/osrd/signaling/impl/SigSystemManagerImpl.kt
@@ -9,13 +9,17 @@ import fr.sncf.osrd.utils.indexing.StaticPool
 class SigSystemManagerImpl : SigSystemManager {
     private val sigSystemMap = mutableMapOf<String, SignalingSystemId>()
     private val sigSystemPool = StaticPool<SignalingSystem, SignalingSystemDriver>()
+    private val sigSystemCost = mutableMapOf<SignalingSystemId, Double>()
     private val driverMap =
         mutableMapOf<Pair<SignalingSystemId, SignalingSystemId>, SignalDriverId>()
     private val driverPool = StaticPool<SignalDriver, fr.sncf.osrd.signaling.SignalDriver>()
 
-    fun addSignalingSystem(sigSystem: SignalingSystemDriver): SignalingSystemId {
+    // cost must be in [0; 1] (used to choose block when everything else is equal)
+    fun addSignalingSystem(sigSystem: SignalingSystemDriver, cost: Double): SignalingSystemId {
         val res = sigSystemPool.add(sigSystem)
         sigSystemMap[sigSystem.id] = res
+        assert(cost in 0.0..1.0) { "Signaling system costs must be normalized in [0; 1]" }
+        sigSystemCost[res] = cost
         return res
     }
 
@@ -71,6 +75,11 @@ class SigSystemManagerImpl : SigSystemManager {
 
     override fun getName(sigSystem: SignalingSystemId): String {
         return sigSystemPool[sigSystem].id
+    }
+
+    override fun getCost(sigSystem: SignalingSystemId): Double {
+        return sigSystemCost[sigSystem]
+            ?: throw RuntimeException("signaling system does not have an assigned cost")
     }
 
     override val drivers
--- a/core/src/main/java/fr/sncf/osrd/api/SignalingSimulator.kt
+++ b/core/src/main/java/fr/sncf/osrd/api/SignalingSimulator.kt
@@ -1,6 +1,7 @@
 package fr.sncf.osrd.api
 
 import fr.sncf.osrd.signaling.SignalingSimulator
+import fr.sncf.osrd.signaling.SignalingSystemDriver
 import fr.sncf.osrd.signaling.bal.*
 import fr.sncf.osrd.signaling.bapr.*
 import fr.sncf.osrd.signaling.etcs_level2.*
@@ -9,17 +10,27 @@ import fr.sncf.osrd.signaling.impl.SignalingSimulatorImpl
 import fr.sncf.osrd.signaling.tvm300.*
 import fr.sncf.osrd.signaling.tvm430.*
 
+val signalingSystemCost =
+    mapOf(BAPR to 0.50, BAL to 0.40, TVM300 to 0.30, TVM430 to 0.20, ETCS_LEVEL2 to 0.10)
+
+fun addSignalingSystem(sigSystemManager: SigSystemManagerImpl, sigSystem: SignalingSystemDriver) {
+    assert(signalingSystemCost.containsKey(sigSystem)) {
+        "Trying to add a signaling system without cost"
+    }
+    sigSystemManager.addSignalingSystem(sigSystem, signalingSystemCost[sigSystem]!!)
+}
+
 /**
  * Configure the signaling simulator for all the supported signaling systems Mainly useful because
  * we can't do it directly from java due to compiler issues
  */
 fun makeSignalingSimulator(): SignalingSimulator {
     val sigSystemManager = SigSystemManagerImpl()
-    sigSystemManager.addSignalingSystem(BAL)
-    sigSystemManager.addSignalingSystem(BAPR)
-    sigSystemManager.addSignalingSystem(TVM300)
-    sigSystemManager.addSignalingSystem(TVM430)
-    sigSystemManager.addSignalingSystem(ETCS_LEVEL2)
+    addSignalingSystem(sigSystemManager, BAL)
+    addSignalingSystem(sigSystemManager, BAPR)
+    addSignalingSystem(sigSystemManager, TVM300)
+    addSignalingSystem(sigSystemManager, TVM430)
+    addSignalingSystem(sigSystemManager, ETCS_LEVEL2)
 
     sigSystemManager.addSignalDriver(BALtoBAL)
     sigSystemManager.addSignalDriver(BALtoBAPR)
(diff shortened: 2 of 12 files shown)


core/kt-osrd-signaling/src/main/kotlin/fr/sncf/osrd/signaling/impl/SigSystemManagerImpl.kt
```diff
@@ -9,13 +9,21 @@ import fr.sncf.osrd.utils.indexing.StaticPool
 class SigSystemManagerImpl : SigSystemManager {
     private val sigSystemMap = mutableMapOf<String, SignalingSystemId>()
     private val sigSystemPool = StaticPool<SignalingSystem, SignalingSystemDriver>()
+    private val sigSystemPrio = mutableMapOf<String, Double>()
```
Khoyo: Priority. And it should be a map of `SignalingSystemId => Double`.

core/kt-osrd-signaling/src/main/kotlin/fr/sncf/osrd/signaling/impl/SigSystemManagerImpl.kt
```diff
@@ -73,6 +81,10 @@ class SigSystemManagerImpl : SigSystemManager {
         return sigSystemPool[sigSystem].id
     }
 
+    override fun getPriority(sigSystem: SignalingSystemId): Double {
+        return sigSystemPrio[getName(sigSystem)] ?: 1.0
```
Khoyo: That should be an assertion if we disallow missing priorities everywhere else

core/kt-osrd-signaling/src/main/kotlin/fr/sncf/osrd/signaling/impl/SigSystemManagerImpl.kt
```diff
@@ -9,13 +9,21 @@ import fr.sncf.osrd.utils.indexing.StaticPool
...
+    private val sigSystemPrio = mutableMapOf<String, Double>()
     private val driverMap =
         mutableMapOf<Pair<SignalingSystemId, SignalingSystemId>, SignalDriverId>()
     private val driverPool = StaticPool<SignalDriver, fr.sncf.osrd.signaling.SignalDriver>()
 
-    fun addSignalingSystem(sigSystem: SignalingSystemDriver): SignalingSystemId {
+    // priority must be in [0; 1[, the lower is the most prior (which block to choose when
+    // everything else is equal)
+    // 1 is reserved for the default value (should not be used)
+    fun addSignalingSystem(sigSystem: SignalingSystemDriver, priority: Double): SignalingSystemId {
```
Khoyo: Maybe we should call it cost everywhere to avoid confusion? Or antipriority?

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
```diff
@@ -165,13 +167,18 @@ private fun computePaths(
...
             .setTimeout(timeout)
             .setEdgeToLength { edge -> Offset(edge.length.distance) }
             .setRangeCost { range ->
                 mrspBuilder.getBlockTime(range.edge.block, Offset(range.end.distance)) -
-                    mrspBuilder.getBlockTime(range.edge.block, Offset(range.start.distance))
+                    mrspBuilder.getBlockTime(range.edge.block, Offset(range.start.distance)) +
+                    SIGNALING_SYSTEM_COST_WEIGHTING *
+                        infra.signalingSimulator.sigModuleManager.getPriority(
+                            infra.blockInfra.getBlockSignalingSystem(range.edge.block)
+                        )
```
eckter: Adding a flat value instead of scaling the result may have unexpected results, it means we discourage paths with shorter edges.

We'd also need to be careful about float inaccuracies here, at a certain point the small values would be "absorbed". We need to check when that would happen and if the values we pick make it unlikely. 

(nit: this code segment would probably be clearer with a distinct function)

core/kt-osrd-signaling/src/main/kotlin/fr/sncf/osrd/signaling/impl/SigSystemManagerImpl.kt
bougue-pe: Done https://github.com/OpenRailAssociation/osrd/pull/10857/commits/b386afa4a4b69a4cb2e52d4977a5c1cde40a8d08#diff-f5b70c57f844cded21af5a59518d2fc6a50b2a212f1118a4f2a470f7901827e6R12
(renamed to `cost` as suggested)

core/kt-osrd-signaling/src/main/kotlin/fr/sncf/osrd/signaling/impl/SigSystemManagerImpl.kt
bougue-pe: Done https://github.com/OpenRailAssociation/osrd/pull/10857/commits/b386afa4a4b69a4cb2e52d4977a5c1cde40a8d08#diff-f5b70c57f844cded21af5a59518d2fc6a50b2a212f1118a4f2a470f7901827e6R82

core/kt-osrd-signaling/src/main/kotlin/fr/sncf/osrd/signaling/impl/SigSystemManagerImpl.kt
bougue-pe: Done in https://github.com/OpenRailAssociation/osrd/pull/10857/commits/b386afa4a4b69a4cb2e52d4977a5c1cde40a8d08 (also `penalty` used when building range cost)

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
bougue-pe: Nice idea/catch, much better indeed in https://github.com/OpenRailAssociation/osrd/pull/10857/commits/b386afa4a4b69a4cb2e52d4977a5c1cde40a8d08#diff-c84bcacb1095e472132ce531ecdfd83d69df5d2e5046a02b567314eef4f8b7d6R205-R222

core/src/test/kotlin/fr/sncf/osrd/pathfinding/PathfindingSignalingTest.kt
```diff
@@ -147,4 +148,33 @@ class PathfindingSignalingTest {
                 )
             )
     }
+
+    @Test
+    fun shouldPriorUseBalPathForBaprBalTrain() {
```
Erashin: If we want to be absolutely sure it's not using the correct path randomly, we could check that doing the opposite makes the train go south. This could be overkill, feel free to disregard.

bougue-pe: It's an easy one to add, so I'll add it if you thought of it too.
For the record I didn't go that way first as the initial version of test was "failing" and then corrected, which is good for TDD, but not as future-proof as your suggestion, so let's go that way :)
Added in https://github.com/OpenRailAssociation/osrd/pull/10857/commits/0b9a06c335c986406be251a3b0c7d2f555464d23

Erashin: Probably could be put into a parametrised test, but its value would only be high if you test all the signalling cost cases. Which wouldn't be too bad tbh.

bougue-pe: Indeed, I digged on that, and it turns out we also win a test ETCS rolling stock, which might be useful: https://github.com/OpenRailAssociation/osrd/pull/10857/commits/99f4c5241f535664ede4ef140ab7f81d6f28f160

Erashin (approved):
Good job!

core/kt-osrd-signaling/src/main/kotlin/fr/sncf/osrd/signaling/impl/SigSystemManagerImpl.kt
```diff
@@ -73,6 +77,12 @@ class SigSystemManagerImpl : SigSystemManager {
         return sigSystemPool[sigSystem].id
     }
 
+    override fun getCost(sigSystem: SignalingSystemId): Double {
+        val cost = sigSystemCost[sigSystem]
+        assert(cost != null) { "signaling system does not have an assigned cost" }
+        return cost!!
```
eckter: (nit)

```suggestion
        return sigSystemCost[sigSystem]
            ?: throw RuntimeException("signaling system does not have an assigned cost")
```

core/src/main/java/fr/sncf/osrd/api/SignalingSimulator.kt
```diff
@@ -9,17 +10,33 @@ import fr.sncf.osrd.signaling.impl.SignalingSimulatorImpl
 import fr.sncf.osrd.signaling.tvm300.*
 import fr.sncf.osrd.signaling.tvm430.*
 
+val signalingSystemNameToCost =
```
eckter:
```suggestion
val signalingSystemToCost =
```

Or even just `signalingSystemCost`

eckter (approved):
LGTM, good job!

Just a couple nitpicks that may be ignored

core/kt-osrd-signaling/src/main/kotlin/fr/sncf/osrd/signaling/impl/SigSystemManagerImpl.kt
bougue-pe: Thanks, done: https://github.com/OpenRailAssociation/osrd/pull/10857/commits/2789daa024e4b073e32f6858d8b7c924f7f46f65

core/src/main/java/fr/sncf/osrd/api/SignalingSimulator.kt
bougue-pe: I also got rid of the `id` on the way: https://github.com/OpenRailAssociation/osrd/pull/10857/commits/2789daa024e4b073e32f6858d8b7c924f7f46f65

PR #10859, opened 2025-02-19
front: take into account new flag for using speed limits

### LINKED ISSUE #10858: front: editing train schedule doesn't take into account the new option `use_speed_limits_for_simulation`
### What happened?

Editing a train doesn't take into account if the option `use_speed_limits_for_simulation` is set to false, and set it back to its default value : true.

I would like to edit a train and let it keep its options if i don't edit them.

Import a json with the option `use_speed_limits_for_simulation` set to false, then edit it.

Recette (SNCF), Dev (SNCF)

Google Chrome

8aaefd4

axrolld:
The idea is to make front end keep this flag when it is present (from a json import for example), and keep its default value otherwise. Not adding anything in the interface.

Fixes #10858 

You can test with this [trainschedule](https://github.com/user-attachments/files/18889025/train_schedules.27.json):
- import the trainSchedule, change the rollingStock for a TGV and see that the train can hit 300km
- change in the json the option usingSpeedLimits, import again the trainSchedule, change the rollingStock for the same TGV. The GEV should not be the same
- check that you can't create a train which will not use the speed limits:
  - click on edit on the train which does not use the speed limits
  - close the manageTrainSchedule page immediately and create a new train
  - the new train should use the speed limits


axrolld (commented):
lgtm, thank you 🙏

SarahBellaha (approved):
LGTM :)

front/src/applications/operationalStudies/views/ManageTrainSchedule.tsx
```diff
@@ -198,6 +201,15 @@ const ManageTrainSchedule = () => {
...
   };
 
+  // reset usingSpeedLimits when unmounting, to prevent user from being able to create a train
+  // without speed limits
+  useEffect(
+    () => () => {
+      dispatch(resetUsingSpeedLimits());
+    },
+    []
+  );
```
SharglutDev, line 211: 🔥 *PO mode on* 🔥 

This is not enough.

Weird case (but not impossible) : 
1. Import the train in the .json and add the train
2. Edit the train (`using_speed_limit` goes `false` in the store as expected)
3. Stay in edit mode
4. Open a new tab, go to a scenario and add a train
5. He will be added with `using_speed_limit` at `false`

Capture d’écran 2025-02-21 à 10 48 40

clarani:
Nice catch ! I remove the field from the persist config, so it is not persisted anymore and if a  user opens a new tab, the config will have `usingSpeedLimit = true`

Math-R (approved):
looks good to me, just wait @SharglutDev  to resolve :)

SharglutDev:
> looks good to me, just wait @SharglutDev to resolve :)

Was waiting for @clarani to rebase but I ended up rebasing in local to test :)

SharglutDev (approved):
Lgtm and tested ✅

PR #10860, opened 2025-02-19
front: fix clear ci input does not clear ch input in map op search

### LINKED ISSUE #9519: Operational point search: clear filters does not work
### What happened?

When the user clear the "Name" field it also clear the "CH" field without removing the text.

We expect each clear button to clear only the associated input text.

1. Fill the search and clear the "name" (or  "CH") field.
Image

> [!CAUTION]
> Both field are cleared, even if the CH text remains.

1. Refill the name
Image

> [!CAUTION]
> The CH field is considered empty even if the text "bv" still here.

Local

Firefox

7f95669451694f5a9fe661fddd8f0b2e500ec077

clarani:
closes #9519


SharglutDev (approved):
Lgtm and tested ✅

PR #10863, opened 2025-02-19
front: fix various problems related to long rolling stock names

### LINKED ISSUE #7278: non-responsive display for a rolling stock with a long name
### What happened?

When creating a rolling stock with a long name , subseries and details ... the created field is not correctly viewed 
image

Add ellipsis for fields with long value 

1. Go to rolling stock editor
2. Create a new rolling stock with long value in the different fields
3. Verify the display of the fields 

- **Browser**: Chrome
- **OS**: W11
- **Env**: rec

1482e8c

### LINKED ISSUE #8439: Long rolling stock name is tightening other tabs during train creation
### What happened?

image

An ellipsis should be fine

1. Try a very long rolling stock name in the train creation window

Dev (SNCF)

Google Chrome

clarani:
closes #8439 
closes #7278

image

### DIFF
--- a/front/src/styles/scss/applications/operationalStudies/_managetrainschedule.scss
+++ b/front/src/styles/scss/applications/operationalStudies/_managetrainschedule.scss
@@ -3,19 +3,29 @@
     display: flex;
     align-items: center;
     width: 100%;
-    .rolling-stock {
+
+    .rolling-stock-img {
       width: 48px;
-      overflow: hidden;
       padding-bottom: 4px;
+      overflow: hidden;
+      mask-image: linear-gradient(90deg, black 90%, rgba(0, 0, 0, 0));
+      -webkit-mask-image: linear-gradient(90deg, black 90%, rgba(0, 0, 0, 0));
+
       img {
         height: 16px;
         object-fit: cover;
         object-position: left;
-        transform: scaleX(-1);
-        mask-image: linear-gradient(90deg, black 80%, rgba(0, 0, 0, 0));
-        -webkit-mask-image: linear-gradient(90deg, black 80%, rgba(0, 0, 0, 0));
       }
     }
+
+    .rolling-stock-name {
+      margin-left: 8px;
+      width: calc(100% - 48px - 8px); // 48px for the rs image width, 8px for the left margin
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: nowrap;
+    }
+
     img {
       height: 24px;
     }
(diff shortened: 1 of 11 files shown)


front/src/styles/scss/applications/operationalStudies/_managetrainschedule.scss
```diff
@@ -3,9 +3,11 @@
     display: flex;
     align-items: center;
     width: 100%;
+    overflow: hidden;
+    text-overflow: ellipsis;
```
SharglutDev: Do we need these with the one in `.rolling-stock-name` class ?

front/src/styles/scss/applications/operationalStudies/_managetrainschedule.scss
```diff
@@ -16,6 +18,14 @@
         -webkit-mask-image: linear-gradient(90deg, black 80%, rgba(0, 0, 0, 0));
       }
     }
+
+    .rolling-stock-name {
+      margin-left: 8px;
+      width: calc(100% - 48px - 8px); // 48px for the rs image width, 8px for the left margin
```
SharglutDev, line 23: Do we keep magic numbers (perfectly fine with your comment) or do we create css variables in the file to avoid the comment ?

SharglutDev (commented):
Great fix thanks ! Left just some nit comments.

Also, we open the pandora box with this issue :upside_down_face: 

Can't see the whole rs name
image

Weird style in the rs card
Capture d’écran du 2025-02-19 17-55-07

Horizontal scroll if the rs name is too long (hiding the action buttons)
Capture d’écran du 2025-02-19 17-53-44

front/src/styles/scss/applications/operationalStudies/_managetrainschedule.scss
clarani: I think it's ok to keep the magic numbers here since they are explained in the comment :)

SharglutDev: Fine by me :)

clarani:
I fixed some of the issues:
- the trains are flipped in the timetable cards
  image

- the names are not wrapped anymore
  image

- ellipsis were added in the rolling stock editor as well
  image

I won't go further since these components will soon be removed, if it's fine for you :+1:

front/src/styles/scss/applications/operationalStudies/_managetrainschedule.scss
clarani: Fixed :)

theocrsb:
I have the impression that there is still some strange behavior when the train name is really long : 
image

SharglutDev (approved):
Lgtm and tested, thank you for these painfull fixes :)
(There are still some weird edge cases but it's unlikely that they happen in a real life use of the app)

clarani:
@theocrsb can you test again ? I can't reproduce anymore

theocrsb (approved):
Sounds good to me. Paul is right about the other use cases.

PR #10864, opened 2025-02-19
editoast: update enums to use `SCREAMING_SNAKE_CASE` for consistency

hamz2a:


flomonster (commented):
What is the motivation behind this change, since in practice it changes nothing?

hamz2a:
> What is the motivation behind this change, since in practice it changes nothing?

Nothing special, since I did this for the enum in my previous PR, I just wanted the same for these enums too.

flomonster (approved):
ok

PR #10866, opened 2025-02-19
editoast: add enabled window for stdcm search env table

### LINKED ISSUE #10838: lmr: time window for possible simulations is incorrect
### What happened?

Now : one can start a simulation with a departure between J+1 17h and J+7 (anytime), no matter when the simulation is done.

The expected behavior is : 
Before 17h, one can start a simulation between J+1 and J+7 (anytime for both)
After 17h, one can start a simulation between J+2 and J+7 (anytime for both)

-

Recette (SNCF)

Google Chrome

8aaefd4

flomonster:
close #10838

### DIFF
--- a/editoast/editoast_models/src/tables.rs
+++ b/editoast/editoast_models/src/tables.rs
@@ -671,6 +671,8 @@ diesel::table! {
         search_window_begin -> Timestamptz,
         search_window_end -> Timestamptz,
         temporary_speed_limit_group_id -> Nullable<Int8>,
+        enabled_from -> Timestamptz,
+        enabled_until -> Timestamptz,
     }
 }
 
--- a/editoast/migrations/2025-02-19-173042_adapt_search_environment_adding_enabled_window/up.sql
+++ b/editoast/migrations/2025-02-19-173042_adapt_search_environment_adding_enabled_window/up.sql
@@ -1,0 +1,5 @@
+ALTER TABLE stdcm_search_environment
+    ADD COLUMN enabled_from TIMESTAMP WITH TIME ZONE DEFAULT '1970-01-01 00:00:00' NOT NULL,
+    ADD COLUMN enabled_until TIMESTAMP WITH TIME ZONE DEFAULT '1970-01-02 00:00:00' NOT NULL;
+CREATE INDEX idx_search_environment_enabled_from ON stdcm_search_environment (enabled_from);
+CREATE INDEX idx_search_environment_enabled_until ON stdcm_search_environment (enabled_until);
--- a/editoast/src/models/stdcm_search_environment.rs
+++ b/editoast/src/models/stdcm_search_environment.rs
@@ -4,15 +4,15 @@ use diesel::ExpressionMethods;
 use diesel::QueryDsl;
 use diesel_async::RunQueryDsl;
 use editoast_derive::Model;
-use editoast_models::model;
 use editoast_models::DbConnection;
 use serde::Serialize;
 use std::ops::DerefMut;
-use std::result::Result;
 use utoipa::ToSchema;
 
 use crate::models::prelude::*;
 
+#[cfg(test)]
+use editoast_models::model;
 #[cfg(test)]
 use serde::Deserialize;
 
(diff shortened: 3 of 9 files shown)


editoast/editoast_models/src/tables.rs
```diff
@@ -671,6 +671,8 @@ diesel::table! {
         search_window_begin -> Timestamptz,
         search_window_end -> Timestamptz,
         temporary_speed_limit_group_id -> Nullable<Int8>,
+        enabled_from -> Timestamptz,
+        enabled_to -> Timestamptz,
```
Castavo: We could solve our problem with only an expiration date, I don't know if that was considered in the refinement

flomonster: We didn't think of that. But would the criterion be to take all environments that haven't expired and take the one with the expiration date closest to `now()`?

Castavo: ~~You could use the same criterion as today: the one that hasn't expired with the largest id~~

Yes, taking the one closest to now() would be necessary, I didn't think of that, that feels flaky for some reason.

In our use of the current model, the `enabled_to` would be equal to the `enabled_from` of the next search env right ?
We could acknowledge this redundancy  ?

It might not actually be that flaky... when thinking about those dates simply as points along the timeline it seems okay.

I don't really know, the idea of the single date crossed my mind, I don't know if it's better, it was simply to make sure it had been considered, the current model is okay

editoast/src/views/stdcm_search_environment.rs
```diff
@@ -250,27 +284,19 @@ pub mod tests {
...
     }
 
     #[rstest]
     async fn retrieve_stdcm_search_env_not_found() {
+        use editoast_models::tables::stdcm_search_environment::dsl::*;
         // GIVEN
         let app = TestAppBuilder::default_app();
-
-        let _ = StdcmSearchEnvironment::delete_all(&mut app.db_pool().get_ok()).await;
+        diesel::delete(stdcm_search_environment)
```
Castavo: Isn't using diesel in the `views` crate frowned upon these days ?

editoast/src/models/stdcm_search_environment.rs
```diff
@@ -30,46 +28,54 @@ pub struct StdcmSearchEnvironment {
...
+    /// The end of the search time window.
     pub search_window_end: DateTime<Utc>,
     #[schema(nullable = false)]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub temporary_speed_limit_group_id: Option<i64>,
+    /// The time window start point where the environment is enabled.
+    pub enabled_from: DateTime<Utc>,
+    /// The time window end point where the environment is enabled.
+    /// This value is usually lower than the `search_window_begin`, since a search is performed before the train rolls.
+    pub enabled_to: DateTime<Utc>,
```
Castavo: `enabled_until` seems more idiomatic

Castavo (approved):
`diesel migration redo` ✅ 
Tested sending several envs and retrieving them ✅ 
The front behaves correctly ✅ 
When adding a new env, reloading the front does pull the new env.

editoast/migrations/2025-02-19-173042_adapt_search_environment_adding_enabled_window/up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE stdcm_search_environment
+ADD COLUMN enabled_from TIMESTAMP WITH TIME ZONE DEFAULT '1970-01-01 00:00:00' NOT NULL,
+    ADD COLUMN enabled_to TIMESTAMP WITH TIME ZONE DEFAULT '1970-01-02 00:00:00' NOT NULL;
```
woshilapin: Should we index these new columns?

editoast/migrations/2025-02-19-173042_adapt_search_environment_adding_enabled_window/up.sql
woshilapin, line 1: Formatting of the file looks weird. Maybe the second line can end with a `;` instead?

editoast/src/models/stdcm_search_environment.rs
```diff
@@ -30,46 +28,54 @@ pub struct StdcmSearchEnvironment {
...
+        // Search for the most recent env
+        tracing::warn!("No STDCM search environment enabled");
         stdcm_search_environment
-            .order_by((search_window_end.desc(), search_window_begin.asc()))
+            .order_by((enabled_to.desc(), id.desc()))
             .first::<Row<StdcmSearchEnvironment>>(conn.write().await.deref_mut())
             .await
             .map(Into::into)
             .ok()
     }
```
woshilapin: nitpick: Maybe this would be more idiomatic?

```rust
enabled_env.or_else(|| {
    tracing::warn!("No STDCM search environment enabled");
    stdcm_search_environment
        .order_by((enabled_to.desc(), id.desc()))
        .first::<Row<StdcmSearchEnvironment>>(conn.write().await.deref_mut())
        .await
        .map(Into::into)
        .ok()
})

File: editoast/src/views/stdcm_search_environment.rs [hunk 6]
```diff
@@ -106,7 +121,7 @@ impl From<StdcmSearchEnvironmentCreateForm> for Changeset<StdcmSearchEnvironment
         (status = 201, body = StdcmSearchEnvironment),
     )
 )]
-async fn overwrite(
+async fn create(
```
woshilapin, line 124: Does this mean that we're going to create data that will never be deleted anymore?

editoast/src/views/stdcm_search_environment.rs
flomonster: Yeah true. I'll fix it

editoast/migrations/2025-02-19-173042_adapt_search_environment_adding_enabled_window/up.sql
flomonster: Yeah, this could avoid a full scan when retrieving an env.

editoast/src/views/stdcm_search_environment.rs
flomonster: We could add a delete endpoint so the user can handle and clean the history.

editoast/editoast_models/src/tables.rs
flomonster: IMO, even if it's okay semantically, It makes the sql more flaky

editoast/migrations/2025-02-19-173042_adapt_search_environment_adding_enabled_window/up.sql
flomonster: I did format this file (I'm using the default vs code formatter). Ending with a `;` won't be valid sql.

editoast/src/models/stdcm_search_environment.rs
flomonster: It requires an async closure. So IMO it's simpler as it is.

editoast/editoast_models/src/tables.rs
Castavo: Okay 😁

editoast/src/models/stdcm_search_environment.rs
```diff
@@ -30,25 +29,50 @@ pub struct StdcmSearchEnvironment {
...
-            .order_by((search_window_end.desc(), search_window_begin.asc()))
+            .order_by((enabled_until.desc(), id.desc()))
             .first::<Row<StdcmSearchEnvironment>>(conn.write().await.deref_mut())
             .await
             .map(Into::into)
             .ok()
     }
 
+    /// Delete all existing search environments.
     pub async fn delete_all(conn: &mut DbConnection) -> Result<(), model::Error> {
```
leovalais, line 78:
```suggestion
	#[cfg(test)]
    pub async fn delete_all(conn: &mut DbConnection) -> Result<(), model::Error> {
```

leovalais (approved):
LGTM

editoast/src/views/stdcm_search_environment.rs
flomonster: We also require an endpoint to list search environments. I suggest we do this in another PR.

Maymanaf (approved):

PR #10869, opened 2025-02-20
front: consolidate consecutive fake tracks overtake steps to single stops in stdcm report

### LINKED ISSUE #10566: front: stdcm: identify fake service tracks and report takeover stops
See https://github.com/osrd-project/osrd-confidential/issues/946 https://github.com/osrd-project/osrd-dags/issues/675

We add fake service track as a *temporary solution* to get takeover stops.

We need to identify when the train enters and leave this fake track (using fake operational points). We'd then remove the fake OPs and replace them with an extra stop. 

**EDIT: as the output table is exclusively built from the front-end, this is actually a front issue**

The back-end would return data that would render like this:

  OP              | CH | end | passage   | start |
 |-----------------|----|-----|-----------|-------|
 | =              | BV  |     | 02:45     |       |
 | =              | P2  |     | 02:45     |       |
 | =              | RL |     | 02:45     |       |
 | =              | TR |     | 02:45     |       |
 | OVERTAKE_0_A;Ambérieu| 0  |     | 02:45     |       |
 | OVERTAKE_1_B;Ambérieu| 0  |     | 03:08     |       |
 | Ambérieu        | P3 |   | 03:08     |       |
 | =              | FR |     | 03:09     |       |
 | =              | FR |     | 03:09     |       |

We need it to look like that, like a single stop:

  OP              | CH | end | passage   | start |
 |-----------------|----|-----|-----------|-------|
 | =              | BV  |     | 02:45     |       |
 | =              | P2  |     | 02:45     |       |
 | =              | RL |     | 02:45     |       |
 | =              | TR |     | 02:45     |       |
 | Ambérieu| 0  |   02:45  |  (23 min)    |   03:08    |
 | Ambérieu        | P3 |   | 03:08     |       |
 | =              | FR |     | 03:09     |       |
 | =              | FR |     | 03:09     |       |

We can discuss the formatting of the initial `OP` field further, it's not set in stone. The way I imagine it would be to match any consecutive `OVERTAKE.*`, and then only use what's after a `;`. 

**Note**: this is all a temporary solution, it would be best to have a lightweight commit that can be `git reverted`.

Synar:
Close #10566

(If you want to test this PR, you will need an infra with such overtake tracks, feel free to ask me for a private infra.json ^^)

Before :
fake_stops_overtaking_before
After :
fake_stops_overtaking_after

Some notes : 

- we may not want to add a stopFor value, as it only seems to be used for user requested path steps
- the existing duration computation would not work as steps to be consolidated do not share a position, and our computatio0n for duration is only precise to the minute. However, it's only required at that precision level by all components that use it, so it should be fine
- this change does not affect the GET/GEV. If we want to modify them, other changes would be required
- depending on how temporary this change is supposed to be, we may want to add a specific stop type for overtaking and/or a translation, to avoid this ugly mention in the report sheet : 

image

### DIFF
--- a/front/public/locales/en/stdcm.json
+++ b/front/public/locales/en/stdcm.json
@@ -142,6 +142,7 @@
     "stopType": {
       "driverSwitch": "driver switch",
       "passageTime": "passage time",
+      "overtake": "overtake",
       "serviceStop": "service stop"
     },
     "time": "Time",
--- a/front/public/locales/fr/stdcm.json
+++ b/front/public/locales/fr/stdcm.json
@@ -142,6 +142,7 @@
     "stopType": {
       "driverSwitch": "relève conducteur",
       "passageTime": "passage",
+      "overtake": "dépassement",
       "serviceStop": "arrêt de service"
     },
     "time": "Heure",
--- a/front/src/applications/stdcm/components/StdcmResults/StdcmResultsTable.tsx
+++ b/front/src/applications/stdcm/components/StdcmResults/StdcmResultsTable.tsx
@@ -62,7 +62,8 @@ const StcdmResultsTable = ({
               ({ location }) =>
                 location && location.name === step.name && location.secondary_code === step.ch
             );
-            const shouldRenderRow = isFirstStep || isRequestedPathStep || isLastStep;
+            const shouldRenderRow =
+              isFirstStep || isRequestedPathStep || isLastStep || step.duration;
             const isPathStep =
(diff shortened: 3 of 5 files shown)


clarani (commented):
Code LGTM, tested and everything works well ! We just need to add the translation for the simulation report sheet before merging :white_check_mark:

Akctarus (approved):
tested, LGTM !

clarani (approved):
LTGM :white_check_mark:

PR #10871, opened 2025-02-20
front: collapse sidebar when infra load

### LINKED ISSUE #10862: Can't uncollapse sidebar while infra is loading
### What happened?

When arriving on a scenario and the infra is loading, if you click on the sidebar collapse button, you can't uncollapse it until the infra is loaded

Image

I expect the uncollapse button to still be accessible

1. Go to a scenario with an unloaded infra
2. Click the sidebar collapse button while the infra is still loading
3. Witness

Recette (SNCF), Dev (SNCF)

Firefox

7eeaeca

theocrsb:
closes #10862


Yohh (approved):

kmer2016 (approved):
LGTM

PR #10884, opened 2025-02-20
editoast: remove 'async-std'

woshilapin:
This removes about half a dozen of dependencies that are not needed.

### DIFF
--- a/editoast/src/views/infra/edition.rs
+++ b/editoast/src/views/infra/edition.rs
@@ -923,7 +923,6 @@ enum EditionError {
 pub mod tests {
     use axum::http::StatusCode;
     use pretty_assertions::assert_eq;
-    use rstest::rstest;
 
     use super::*;
     use crate::generated_data::infra_error::InfraError;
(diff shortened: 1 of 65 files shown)


editoast/src/views/infra/edition.rs
```diff
@@ -988,9 +987,11 @@ pub mod tests {
         app.fetch(request).assert_status(StatusCode::BAD_REQUEST);
     }
 
-    #[rstest]
+    #[rstest::rstest]
     #[case("TA0", 1000000)]
     #[case("TD1", 15500000)]
+    #[tokio::test(flavor = "multi_thread")]
+    #[ignore]
```
woshilapin, line 994: I'm having a problem with this function that seems to deadlock.

leovalais: With worker_threads = 1 or 2 maybe?

woshilapin: Nope, already tried that. At some point of the test, it requires a writable connection to the DB through `.write_owned()`. Since it's behind a `RwLock`, it must wait for all readers to be released and it seems it never happen. I actually tried replacing the `.write_owned()` with `try_write_owned()` which may fail with `TryLockError` and it's exactly what happened.

woshilapin: The part I can't explain is why the behavior is different between `async-std` and `tokio` 🤯

flomonster: Do you think it's a bug from tokio?

woshilapin: No idea. Didn't have time to investigate this since, and probably not in the foreseeable future.

PR #10885, opened 2025-02-20
front: allow e2e tests in no-extraneous-dependencies

Maymanaf:


front/.eslintrc
```diff
@@ -84,7 +84,7 @@
         "devDependencies": [
           "**/*.spec.ts",
           "**/__tests__/**",
-          "tests/**",
+         "**/tests/**",
```
emersion: This would allow any directory named "tests", not just the E2E directory located at `front/tests/`.

Maymanaf: Updated to` front/tests/**` to scope the rule strictly to the E2E directory and prevent other `tests/` directories from being affected

emersion: The eslint file is already located in `front/`.

Maymanaf: yes but using `"tests/**"` does not work, it seems that eslint resolves paths from project root.
Only `"front/tests/**"` and `"**/tests/**"` are working for me

emersion: Probably because you're running eslint from the root of the project, instead of from `front/`?

Maymanaf: it was do to NPM corruption, all works fine after clean up

PR #10890, opened 2025-02-20
editoast: add rustfmt.toml to .gitignore

leovalais:
# Proposition

We prefer using un-merged `use` statements for clarity and ease of
review. `rustfmt` has a `nightly` option to do so automatically, but
we don't want to ask everyone to use `nightly` rust.

Ignoring `rustfmt.toml` allows to define it locally and still benefit
from the auto-formatting of `use` statements, without the risk of
accidentally committing it.

```
$ cat rustfmt.toml
imports_granularity = "item"
```

If one day we want to commit a mandatory `rustfmt.toml`, we can always
revert this commit.

### DIFF
--- a/editoast/.gitignore
+++ b/editoast/.gitignore
@@ -2,3 +2,6 @@ target
 *.snap.new
 assets/sprites/*/sprites.*
 assets/fonts/glyphs/*
+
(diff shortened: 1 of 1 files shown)


woshilapin (approved):
Looks reasonable.

PR #10896, opened 2025-02-21
core: safety speeds: fix route path start offset

eckter:
We'd end up adding a value instead of removing it, sometimes resulting in values outside the route range.

It caused an out of bounds at some point in the benchmark.

### DIFF
--- a/core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
+++ b/core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
@@ -7,6 +7,7 @@ import fr.sncf.osrd.sim_infra.api.*
 import fr.sncf.osrd.sim_infra.impl.ChunkPath
 import fr.sncf.osrd.utils.DistanceRangeMap
 import fr.sncf.osrd.utils.distanceRangeMapOf
+import fr.sncf.osrd.utils.getRoutePathStartOffset
 import fr.sncf.osrd.utils.indexing.StaticIdx
 import fr.sncf.osrd.utils.indexing.StaticIdxList
 import fr.sncf.osrd.utils.units.Offset
@@ -36,7 +37,7 @@ fun makeSafetySpeedRanges(
 ): DistanceRangeMap<Speed> {
     val rawInfra = infra.rawInfra
     val zonePaths = routes.flatMap { rawInfra.getRoutePath(it) }
-    val zonePathStartOffset = getRoutePathStartOffset(rawInfra, chunkPath, zonePaths)
+    val zonePathStartOffset = getRoutePathStartOffset(rawInfra, chunkPath, routes)
     val signalOffsets = getSignalOffsets(infra, zonePaths, zonePathStartOffset)
 
     val stopsWithSafetySpeed =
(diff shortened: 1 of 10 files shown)


Erashin (approved):
This works. But this method is a pain to understand.

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
```diff
@@ -127,19 +127,22 @@ fun getRoutePathStartOffset(
...
     }
 
     for (chunk in routeChunks) {
         if (chunk == firstChunk) {
-            return prevChunksLength + chunkPath.beginOffset.distance
+            return prevChunksLength + pathBeginOffset
         }
-        prevChunksLength += infra.getTrackChunkLength(chunk.value).distance
+        val len = infra.getTrackChunkLength(chunk.value).distance
+        prevChunksLength += len
```
bougue-pe: Can it be clearer with something along this:
```kt
    val firstChunk = Pair(chunkPath.chunks[0], chunkPath.beginOffset)
    val startChunkCandidates = arrayListOf(firstChunk)

    // If the path starts precisely at the end of the first chunk, that first chunk may not be
    // present in the route path.
    // We can look for the next chunk instead.
    if (
        infra.getTrackChunkLength(firstChunk.first.value) == firstChunk.second &&
            chunkPath.chunks.size > 1
    ) {
        startChunkCandidates.add(Pair(chunkPath.chunks[1], Offset.zero()))
    }

    for (chunk in routeChunks) {
        for (startCandidate in startChunkCandidates) if (chunk == startCandidate.first) {
            return prevChunksLength + startCandidate.second.distance
        }
        prevChunksLength += infra.getTrackChunkLength(chunk.value).distance
    }
```
(maybe more robust to corner cases where there is only the first or only the second chunk in the routePath ?)

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
```diff
@@ -127,19 +127,22 @@ fun getRoutePathStartOffset(
     val routeChunks = zonePaths.flatMap { infra.getZonePathChunks(it) }
```
bougue-pe: Looking for other possible identical bugs:
Can we get that same issue on blockPath to chunkPath? I see [ScheduleMetadatExtractor](https://github.com/OpenRailAssociation/osrd/blob/33f63cff3800e39d301de27c46b0350246d81d04/core/src/main/java/fr/sncf/osrd/standalone_sim/ScheduleMetadataExtractor.kt#L638-L665) (v1, used in signal projection v2)

Working on chunkPath for PathPropertiesImpl [here](https://github.com/OpenRailAssociation/osrd/blob/33f63cff3800e39d301de27c46b0350246d81d04/core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/PathPropertiesImpl.kt#L42-L46) seems safe (ctrl+F `beginOffset`).

bougue-pe (commented):
🙏 Thanks, tricky bug.

It may be nice to add a test, but a unit-test might be over-tied to the implem... Not sure about the cost.

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
eckter: Good point, the two methods do almost the same thing. I ended up moving them into a single utils file, then merging what could be merged (which was almost all of it).

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
eckter: I rewrote the method to get closer to this, it's not exactly the same though

Erashin (approved):
Waaaaayyyy better, love this!

core/src/test/kotlin/fr/sncf/osrd/utils/OffsetConversionTests.kt
```diff
@@ -0,0 +1,36 @@
...
+        val res = trainPathZonePathOffset(infra, zones, chunkPath)
+        assertEquals(Offset(1_500.meters), res)
+    }
+
+    @Test
+    fun testStartEndOfZone() {
+        val chunkPath = ChunkPath(chunks.toIdxList(), Offset(1_000.meters), Offset(10_000.meters))
+        val res = trainPathZonePathOffset(infra, zones, chunkPath)
+        assertEquals(Offset(1_000.meters), res)
+    }
```
bougue-pe, line 35: We may also add the corresponding test for the initial fix:
```kt
    @Test
    fun testStartBeginOfZone() {
        val chunkPath = ChunkPath(chunks.toIdxList(), Offset(1_000.meters), Offset(10_000.meters))
        val res = trainPathZonePathOffset(infra, zones.subList(1, zones.size), chunkPath)
        assertEquals(Offset(0.meters), res)
    }
```

bougue-pe (approved):
Nice!
Bugfix, tests, mutualization and clarity 🎉 
Thank you!

core/src/test/kotlin/fr/sncf/osrd/utils/OffsetConversionTests.kt
eckter: We're supposed to have all the zones but possibly parts of the chunks, so it's already one of the new unit tests

eckter: Actually I misunderstood: we do need to support cases where the path starts at the end of a chunk that isn't in the zone path, that's why we need that extra complexity in the first place.

I've added new tests in 8882f0323dcf8fb8e0eb81e51a72b24dcccdff74

bougue-pe (approved):

PR #10902, opened 2025-02-21
front: fix number of steps color in import train schedule

### LINKED ISSUE #10898: viriato import: trains with no intermediate steps should not have a red dot
### What happened?

when importing a Viriato file, the number of intermediate steps is displayed

Image

Here the 3rd train has no intermediate step (so only origin and destination) and its dot is in red -> it should be blue

The case where the dot should be red is if there is less at two path steps, so if the number of intermediate steps is <0 (and not <=0 currently)

as above

Dev (SNCF)

Google Chrome

6edea6d

RomainValls:
close #10898


SharglutDev (approved):
Lgtm and tested ✅

PR #10904, opened 2025-02-21
front: fix scroll in help module

RomainValls:
close https://github.com/osrd-project/osrd-confidential/issues/989

### DIFF
--- a/front/src/styles/scss/applications/stdcm/_helpModule.scss
+++ b/front/src/styles/scss/applications/stdcm/_helpModule.scss
@@ -13,6 +13,7 @@
   padding: 72px 48px 72px 56px;
   display: flex;
   flex-direction: column;
+  overflow: auto;
   &.active {
     right: 0;
   }


SharglutDev (commented):
Need to be discussed with @thibautsailly but with this change if the help module is open, I can't scroll anymore in the stdcm form.

thibautsailly:
I can't see the change you're talking about, but you should be able to scroll the form while reading the help content. Scrolling over the help section should not affect the form, and scrolling the form should not affect the help section.

RomainValls:
Tried a different approach, now the help module and the stdcm body can be scrolled independantly. 
@thibautsailly since you can't test, you can check this video 

https://github.com/user-attachments/assets/acef6ef6-4470-4cd9-8826-a41714de2b7d

should I modify something ?

SharglutDev:
> Tried a different approach, now the help module and the stdcm body can be scrolled independantly. @thibautsailly since you can't test, you can check this video
> should I modify something ?

The scroll in the help module seems to be only in the tab block. What if the screen hides the footer of the help module (phone number etc), is it possible to scroll to see it ?

RomainValls:
> > Tried a different approach, now the help module and the stdcm body can be scrolled independantly. @thibautsailly since you can't test, you can check this video
> > should I modify something ?
> 
> The scroll in the help module seems to be only in the tab block. What if the screen hides the footer of the help module (phone number etc), is it possible to scroll to see it ?

I'm not sure I get your point, but what happens when we scroll more, is that we'll only see the footer, that will hide the above content

clarani (commented):
Hum I think that adding "overflow: auto" in the class `stdcm__help-module` is enough (you can scroll in the help module with this)

RomainValls:
> Hum I think that adding "overflow: auto" in the class `stdcm__help-module` is enough (you can scroll in the help module with this)

Thanks, it's done :)

clarani (approved):
LGTM :white_check_mark:

PR #10911, opened 2025-02-24
front: removes useOsrdConfSelectors hook

jacomyal:
This PR keeps removing selectors from the context, and is a continuation of #10853.

Details:
- In each hook or component that used `useOsrdConfSelectors` for some parts of the conf, moves these parts as props
- Updates accordingly all uses of these hooks and components
- Finally safely deletes `useOsrdConfSelectors`

### DIFF
--- a/front/src/applications/stdcm/hooks/useStdcm.ts
+++ b/front/src/applications/stdcm/hooks/useStdcm.ts
@@ -68,7 +68,10 @@ const useStdcm = ({
       { skip: !osrdconf.rollingStockID }
     );
 
-  const { speedLimitByTag } = useStoreDataForSpeedLimitByTagSelector({ isStdcm: true });
+  useStoreDataForSpeedLimitByTagSelector({
+    isStdcm: true,
+    speedLimitByTag: osrdconf.speedLimitByTag,
+  });
 
   const resetStdcmState = () => {
     setStdcmTrainResult(undefined);
@@ -108,7 +111,7 @@ const useStdcm = ({
             ...response,
             rollingStock: stdcmRollingStock,
             creationDate: new Date(),
-            speedLimitByTag,
+            speedLimitByTag: osrdconf.speedLimitByTag,
             simulationPathSteps: osrdconf.stdcmPathSteps,
           } as StdcmSuccessResponse);
 
@@ -129,7 +132,7 @@ const useStdcm = ({
             ...response,
             rollingStock: stdcmRollingStock,
             creationDate: new Date(),
-            speedLimitByTag,
+            speedLimitByTag: osrdconf.speedLimitByTag,
             simulationPathSteps: osrdconf.stdcmPathSteps,
             path: response.pathfinding_result,
           } as StdcmConflictsResponse);
--- a/front/src/common/SpeedLimitByTagSelector/useStoreDataForSpeedLimitByTagSelector.ts
+++ b/front/src/common/SpeedLimitByTagSelector/useStoreDataForSpeedLimitByTagSelector.ts
@@ -2,21 +2,24 @@ import { useEffect, useMemo } from 'react';
 
 import { compact, concat, uniq } from 'lodash';
 import { useTranslation } from 'react-i18next';
-import { useSelector } from 'react-redux';
 
 import { COMPOSITION_CODES, DEFAULT_COMPOSITION_CODE } from 'applications/stdcm/consts';
 import { osrdEditoastApi } from 'common/api/osrdEditoastApi';
-import { useOsrdConfSelectors, useOsrdConfActions, useInfraID } from 'common/osrdContext';
+import { useOsrdConfActions, useInfraID } from 'common/osrdContext';
 import { setFailure } from 'reducers/main';
 import { useAppDispatch } from 'store';
 import { castErrorToFailure } from 'utils/error';
 
-export const useStoreDataForSpeedLimitByTagSelector = ({ isStdcm } = { isStdcm: false }) => {
+export const useStoreDataForSpeedLimitByTagSelector = ({
+  isStdcm,
+  speedLimitByTag,
+}: {
+  isStdcm?: boolean;
+  speedLimitByTag: string | undefined;
+}) => {
   const dispatch = useAppDispatch();
   const { t } = useTranslation(['operationalStudies/manageTrainSchedule']);
 
-  const { getSpeedLimitByTag } = useOsrdConfSelectors();
-  const speedLimitByTag = useSelector(getSpeedLimitByTag);
   const infraID = useInfraID();
 
   const { updateSpeedLimitByTag } = useOsrdConfActions();
@@ -57,7 +60,6 @@ export const useStoreDataForSpeedLimitByTagSelector = ({ isStdcm } = { isStdcm: 
   const speedLimitsByTagsOrdered = useMemo(() => speedLimitsByTags.sort(), [speedLimitsByTags]);
 
   return {
-    speedLimitByTag,
     speedLimitsByTags: speedLimitsByTagsOrdered,
     dispatchUpdateSpeedLimitByTag,
   };
--- a/front/src/modules/pathfinding/components/Pathfinding/Pathfinding.tsx
+++ b/front/src/modules/pathfinding/components/Pathfinding/Pathfinding.tsx
@@ -8,7 +8,6 @@ import InfraLoadingState from 'applications/operationalStudies/components/Scenar
 import { useManageTrainScheduleContext } from 'applications/operationalStudies/hooks/useManageTrainScheduleContext';
 import infraLogo from 'assets/pictures/components/tracks.svg';
 import { Spinner } from 'common/Loaders';
-import { useOsrdConfSelectors } from 'common/osrdContext';
 import { isPathStepInvalid } from 'modules/pathfinding/utils';
 import {
   getPathSteps,
@@ -19,16 +18,13 @@ import { conditionalStringConcat, formatKmValue } from 'utils/strings';
 
 import { InfraHardError, InfraSoftError } from './InfraError';
 
-const Pathfinding = () => {
+const Pathfinding = ({ rollingStockId }: { rollingStockId: number | undefined }) => {
   const { t } = useTranslation(['operationalStudies/manageTrainSchedule']);
 
   const pathSteps = useSelector(getPathSteps);
   const hasInvalidPathStep = pathSteps.some((pathStep) => isPathStepInvalid(pathStep));
   const origin = useSelector(getOrigin, isEqual);
   const destination = useSelector(getDestination, isEqual);
-
-  const { getRollingStockID } = useOsrdConfSelectors();
-  const rollingStockId = useSelector(getRollingStockID);
 
   const {
     pathProperties,
--- a/front/src/modules/pathfinding/components/Pathfinding/TypeAndPath.tsx
+++ b/front/src/modules/pathfinding/components/Pathfinding/TypeAndPath.tsx
@@ -5,7 +5,6 @@ import { Alert, TriangleRight } from '@osrd-project/ui-icons';
 import cx from 'classnames';
 import { useTranslation } from 'react-i18next';
 import nextId from 'react-id-generator';
-import { useSelector } from 'react-redux';
 
 import { useManageTrainScheduleContext } from 'applications/operationalStudies/hooks/useManageTrainScheduleContext';
 import type {
@@ -14,7 +13,7 @@ import type {
 } from 'common/api/osrdEditoastApi';
 import { osrdEditoastApi } from 'common/api/osrdEditoastApi';
 import { MAIN_OP_CH_CODES } from 'common/Map/Search/useSearchOperationalPoint';
-import { useInfraID, useOsrdConfSelectors } from 'common/osrdContext';
+import { useInfraID } from 'common/osrdContext';
 import { useDebounce } from 'utils/helpers';
 import {
   isCursorSurroundedBySpace,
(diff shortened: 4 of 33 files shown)


front/src/applications/stdcm/hooks/useStdcm.ts
```diff
@@ -68,7 +68,10 @@ const useStdcm = ({
       { skip: !osrdconf.rollingStockID }
     );
 
-  const { speedLimitByTag } = useStoreDataForSpeedLimitByTagSelector({ isStdcm: true });
+  useStoreDataForSpeedLimitByTagSelector({
```
jacomyal, line 71: I have to confess I'm not fully convinced of what this hook call does now.

front/src/applications/operationalStudies/components/Scenario/ScenarioDescription.tsx
```diff
@@ -116,7 +119,7 @@ const ScenarioDescription = ({
           aria-label={t('editScenario')}
           onClick={() =>
             openModal(
-              <AddAndEditScenarioModal editionMode scenario={scenario} />,
+              <AddAndEditScenarioModal editionMode timetableId={timetableId} scenario={scenario} />,
```
clarani: If you change the type of scenario in `AddAndEditScenarioModal` (`ScenarioResponse` instead of `ScenarioForm`), then you can simply get the `timetableId` with `scenario.timetable_id`

front/src/applications/stdcm/hooks/useStdcm.ts
clarani: Hum me too. It's only used to dispatch the default speed limit by tag. Maybe we can remove it in another commit/PR ?

clarani (commented):
Code LGTM, I just need to test

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -43,16 +42,19 @@ const initialPathfindingState = {
...
+}: {
+  rollingStockId: number | undefined;
+  setPathProperties: (pathProperties?: ManageTrainSchedulePathProperties) => void;
+}) => {
   const { t } = useTranslation(['operationalStudies/manageTrainSchedule']);
   const dispatch = useAppDispatch();
   const pathSteps = useSelector(getPathSteps);
   const powerRestrictions = useSelector(getPowerRestrictions);
-  const { infra, reloadCount, setIsInfraError } = useInfraStatus();
-  const { getRollingStockID } = useOsrdConfSelectors();
```
SharglutDev, line 54: `usePathfinding` is used only in operational studies (we have `useStaticPathfinding` for stdcm), it should probably be safe to call `useSelector(getOperationalStudiesRollingStockID)` here directly ?

front/src/modules/simulationResult/components/SpaceTimeChart/useLazyProjectTrains.ts
```diff
@@ -30,6 +27,7 @@ const BATCH_SIZE = 5;
 
 type useLazyLoadTrainsProp = {
   infraId?: number;
+  electricalProfileSetId: number | undefined;
```
SharglutDev, line 30:
```suggestion
  electricalProfileSetId?: number;
```

front/src/modules/rollingStock/components/RollingStockSelector/useStoreDataForRollingStockSelector.ts
```diff
@@ -1,12 +1,13 @@
...
-import { useOsrdConfSelectors } from 'common/osrdContext';
 import { getRollingStockComfort } from 'reducers/osrdconf/operationalStudiesConf/selectors';
 
-export const useStoreDataForRollingStockSelector = () => {
-  const { getRollingStockID } = useOsrdConfSelectors();
-  const rollingStockId = useSelector(getRollingStockID);
+export const useStoreDataForRollingStockSelector = ({
+  rollingStockId,
+}: {
+  rollingStockId: number | undefined;
```
SharglutDev, line 9:
```suggestion
  rollingStockId?: number;
```

front/src/common/SpeedLimitByTagSelector/useStoreDataForSpeedLimitByTagSelector.ts
```diff
@@ -2,21 +2,24 @@ import { useEffect, useMemo } from 'react';
...
 import { useAppDispatch } from 'store';
 import { castErrorToFailure } from 'utils/error';
 
-export const useStoreDataForSpeedLimitByTagSelector = ({ isStdcm } = { isStdcm: false }) => {
+export const useStoreDataForSpeedLimitByTagSelector = ({
+  isStdcm,
+  speedLimitByTag,
+}: {
+  isStdcm?: boolean;
+  speedLimitByTag: string | undefined;
```
SharglutDev, line 18:
```suggestion
  speedLimitByTag?: string;
```

front/src/modules/rollingStock/components/RollingStockSelector/RollingStockModal.tsx
```diff
@@ -1,27 +1,28 @@
...
 import ModalBodySNCF from 'common/BootstrapSNCF/ModalSNCF/ModalBodySNCF';
 import { ModalContext } from 'common/BootstrapSNCF/ModalSNCF/ModalProvider';
 import { Loader } from 'common/Loaders';
-import { useOsrdConfSelectors } from 'common/osrdContext';
 import RollingStockCard from 'modules/rollingStock/components/RollingStockCard/RollingStockCard';
 import SearchRollingStock from 'modules/rollingStock/components/RollingStockSelector/SearchRollingStock';
 import useFilterRollingStock from 'modules/rollingStock/hooks/useFilterRollingStock';
 
 type RollingStockModal = {
+  rollingStockId: number | undefined;
```
SharglutDev, line 14:
```suggestion
  rollingStockId?: number;
```

front/src/modules/pathfinding/components/Pathfinding/Pathfinding.tsx
```diff
@@ -19,17 +18,14 @@ import { conditionalStringConcat, formatKmValue } from 'utils/strings';
 
 import { InfraHardError, InfraSoftError } from './InfraError';
 
-const Pathfinding = () => {
+const Pathfinding = ({ rollingStockId }: { rollingStockId: number | undefined }) => {
```
SharglutDev, line 21:
```suggestion
const Pathfinding = ({ rollingStockId }: { rollingStockId?: number }) => {
```

front/src/modules/pathfinding/components/Pathfinding/Pathfinding.tsx
```diff
@@ -19,17 +18,14 @@ import { conditionalStringConcat, formatKmValue } from 'utils/strings';
...
   const { t } = useTranslation(['operationalStudies/manageTrainSchedule']);
 
   const pathSteps = useSelector(getPathSteps);
   const hasInvalidPathStep = pathSteps.some((pathStep) => isPathStepInvalid(pathStep));
   const origin = useSelector(getOrigin, isEqual);
   const destination = useSelector(getDestination, isEqual);
 
-  const { getRollingStockID } = useOsrdConfSelectors();
```
SharglutDev, line 30: This component is also used only in operational studies, we could call the selector here (agreed that's is like an anti-pattern with the module structure)

front/src/modules/pathfinding/components/Pathfinding/TypeAndPath.tsx
```diff
@@ -56,9 +55,10 @@ function OpTooltips({ opList }: { opList: SearchResultItemOperationalPoint[] })
 }
 type TypeAndPathProps = {
   setDisplayTypeAndPath: React.Dispatch<React.SetStateAction<boolean>>;
+  rollingStockId: number | undefined;
```
SharglutDev, line 58:
```suggestion
  rollingStockId?: number;
```

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
```diff
@@ -2,25 +2,27 @@ import { useEffect, useMemo, useState } from 'react';
...
-  infraId?: number
-) => {
+const useGetProjectedTrainOperationalPoints = ({
+  infraId,
+  timetableId,
+  trainScheduleUsedForProjection,
+  trainIdUsedForProjection,
+}: {
+  infraId: number | undefined;
+  timetableId: number | undefined;
```
SharglutDev, line 21:
```suggestion
  infraId?: number;
  timetableId?: number;
```

front/src/modules/simulationResult/types.ts
```diff
@@ -42,6 +42,7 @@ export type ProjectionData = {
 };
 
 export type WaypointsPanelData = {
+  timetableId: number | undefined;
```
SharglutDev, line 45:
```suggestion
  timetableId?: number;
```

front/src/modules/scenario/components/ScenarioExplorer/ScenarioExplorer.tsx
```diff
@@ -25,14 +23,14 @@ const ScenarioExplorer = ({
   globalStudyId,
   globalScenarioId,
   displayImgProject = true,
+  timetableId,
 }: ScenarioExplorerProps & {
   displayImgProject?: boolean;
+  timetableId: number | undefined;
```
SharglutDev, line 29:
```suggestion
  timetableId?: number;
```

front/src/modules/scenario/components/AddOrEditScenarioModal.tsx
```diff
@@ -40,6 +39,7 @@ export type ScenarioForm = ScenarioPatchForm & {
 };
 
 type AddOrEditScenarioModalProps = {
+  timetableId: number | undefined;
```
SharglutDev:
```suggestion
  timetableId?: number;
```

front/src/modules/trainschedule/components/ManageTrainSchedule/ManageTrainScheduleMap/AddPathStepPopup.tsx
```diff
@@ -25,20 +24,20 @@ import type { FeatureInfoClick } from '../types';
 import OperationalPointPopupDetails from './OperationalPointPopupDetails';
 
 type AddPathStepPopupProps = {
+  infraId: number | undefined;
```
SharglutDev, line 27:
```suggestion
  infraId?: number;
```

SharglutDev (commented):
Thank you for the refacto ! Tested and everything works fine.

Sorry for all the comments for `undefined`, preferred to be exhaustive

front/src/applications/operationalStudies/components/Scenario/ScenarioDescription.tsx
jacomyal: Oh nice, thanks! This one was indeed a bit painful to write ^^

front/src/modules/pathfinding/hooks/usePathfinding.ts
jacomyal: Hum, I don't really like removing it, because this component is in the `modules` directory, not the `applications` directory, and so shouldn't depend on one application's state. What do you think about it?

jacomyal:
Hi @SharglutDev, thanks for the review! I have two questions for you:

1. About the `undefined`, I wrote them on purpose: The value can be `undefined`, but the prop has to be given. With your notation `something?: someType`, the prop is not required. Why do you want to switch exactly?

2. Why do you prefer calling directly the application-specific selectors in some modules? This makes these modules ambiguous and not future proof. I'd say ok to use the application-specific selectors only if we move the code to the `application` directory, personally.

front/src/applications/operationalStudies/components/Scenario/ScenarioDescription.tsx
jacomyal: Fixed

front/src/modules/pathfinding/hooks/usePathfinding.ts
SharglutDev: That was my thought also by writing this [comment](https://github.com/OpenRailAssociation/osrd/pull/10911#discussion_r1969555726). The module structure is a bit misleading because everything in it isn't reused elsewhere.

Here the `usePathfinding` hook already calls some selectors that exist only in operational studies (which is kind of a non-sense for a module component too) which is why I wrote this.
So if we chose to remove `rollingStockId` from here, we should also do it for the rest of the selectors.

In another PR, we could change this behavior for all modules. What do you think @clarani ?

clarani: Yes I agree :+1: We can leave it for another PR :+1:

clarani:
To answer to your questions @jacomyal (IMO):
1. I think that there is not much difference between making a prop optional or mandatory but possibly undefined, especially for components, since the order of the props doesn't matter. By default, for components, I would make the props optional to avoid forcing a user to pass undefined as a prop in the future.
  I also find that `infraId?: number;` is more concise and more readable than `infraId: number | undefined;` (but that's purely subjective).

3. Currently, some files/components/hooks are not in the right folder, which leads us to these questions. I think the main issue is that the files are not properly organized, but this could be tackled in another issue. I'm okay with the change proposed by this merge request

clarani (approved):
LGTM. We just need to tackle Paul's comments :+1:

SharglutDev (approved):
Lgtm, thank you for the changes !

> Hi @SharglutDev, thanks for the review! I have two questions for you:
> 
>     1. About the `undefined`, I wrote them on purpose: The value can be `undefined`, but the prop has to be given. With your notation `something?: someType`, the prop is not required. Why do you want to switch exactly?
> 
>     2. Why do you prefer calling directly the application-specific selectors in some modules? This makes these modules ambiguous and not future proof. I'd say ok to use the application-specific selectors only if we move the code to the `application` directory, personally.

1. You're right. It's a better type saftey for theses cases. We will address your point in the next front meeting to see if we could do that in other components.

2. Agreed to make some changes in another PR because our structure has to be enhanced and is misleading right now :)

PR #10912, opened 2025-02-24
front: invert signs on the tolerance range in the simulation sheet

### LINKED ISSUE #10899: STDCM: inverted signs on tolerance range in simulation sheet
### What happened?

The signs on the tolerance range are reversed in the simulation sheet compared to those requested:
  Image

Image

the same signs in the simulation sheet as in the STDCM request form (in the example -30/+40)

Go to STDCM page and ask for a train slot with a fixed departure time +/- a certain tollerance (put for example -30/+40 as tollerance range). When the simulation returns a result, ask for dowloading the simulation sheet and check the tollerance range.

Production (SNCF)

Google Chrome

v0.3.12 8aaefd4

theocrsb:
closes #10899

### DIFF
--- a/front/src/applications/stdcm/components/StdcmResults/SimulationReportSheet.tsx
+++ b/front/src/applications/stdcm/components/StdcmResults/SimulationReportSheet.tsx
@@ -201,13 +201,13 @@ const SimulationReportSheet = ({
                           {isLastStep && !step.isVia && step.arrivalType === 'preciseTime' && (
                             <View style={styles.convoyAndRoute.tolerancesWidth}>
                               <Text style={styles.convoyAndRoute.tolerancesText}>
-                                {step.tolerances?.before
-                                  ? `+${step.tolerances.before.total('minute')}`
+                                {step.tolerances?.after
+                                  ? `+${step.tolerances.after.total('minute')}`
                                   : ''}
                               </Text>
                               <Text style={styles.convoyAndRoute.tolerancesText}>
-                                {step.tolerances?.after
-                                  ? `-${step.tolerances.after.total('minute')}`
+                                {step.tolerances?.before
+                                  ? `-${step.tolerances.before.total('minute')}`
                                   : ''}
                               </Text>
                             </View>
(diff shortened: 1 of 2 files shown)


emersion (approved):

SharglutDev (approved):
Lgtm and tested

Caracol3 (approved):
lgtm (tested)

front/tests/assets/simulation-sheet-const.ts
```diff
@@ -42,8 +42,8 @@ const simulationSheetDetails = (): Simulation => {
       station1: {
         name: '1 North_West_station',
         ch: 'BV',
-        plusTolerance: '+60',
         minusTolerance: '-15',
+        plusTolerance: '+60',
```
Yohh: have you missed to change operators?

theocrsb: I'm trying to get the E2E test to work, but it's not working :/

Yohh: ok, if needed we can peer if you wish, or you can also ask for help on element `test` channel

Maymanaf:
```suggestion
        minusTolerance: '-60',
        plusTolerance: '+15',
```

theocrsb: I thought it was a problem of order. Thanks, I've just added the changes

Maymanaf (approved):
LGTM ✅ 
You can simply merge after reverting the order

theocrsb:
I'm thinking of leaving this order so that I always have the `before` then the `after`. Is that ok with you?

front/src/applications/stdcm/components/StdcmResults/SimulationReportSheet.tsx
```diff
@@ -202,12 +202,12 @@ const SimulationReportSheet = ({
                             <View style={styles.convoyAndRoute.tolerancesWidth}>
                               <Text style={styles.convoyAndRoute.tolerancesText}>
                                 {step.tolerances?.before
-                                  ? `+${step.tolerances.before.total('minute')}`
+                                  ? `-${secToMin(step.tolerances.before)}`
```
emersion: Please keep using `total()` instead of `secToMin()` :P

theocrsb: Rebasing error :/

theocrsb: Done, but we have an eslint error : `Unsafe call of a(n) `error` type typed value.eslint[@typescript-eslint/no-unsafe-call](https://typescript-eslint.io/rules/no-unsafe-call)`

emersion (approved):

front/src/applications/stdcm/components/StdcmResults/SimulationReportSheet.tsx
SharglutDev: If you reload vscode, the warning will go away (not picked by the CI)

theocrsb:
After speaking with @thibautsailly, I had to reverse the tolerance layout.

SharglutDev (approved):
Thanks for the re-change

PR #10913, opened 2025-02-24
lmr: add checks in the form to verify if some fields are left empty

### LINKED ISSUE #9707: lmr: front: catch missing fields and adapt warning message

Caracol3:
close #9707

### DIFF
--- a/front/public/locales/en/stdcm.json
+++ b/front/public/locales/en/stdcm.json
@@ -127,6 +127,17 @@
   "stdcmErrors": {
     "bothPointAreScheduled": "The calculation cannot take into account both an origin and a destination schedule. You have to choose one or the other:",
     "infraNotLoaded": "Infrastructure is loading",
+    "missingFields": {
+      "destination": "the destination",
+      "origin": "the origin",
+      "timeDestination": "the arrival time at the destination",
+      "timeOrigin": "the departure time at the origin",
+      "tractionEngine": "the traction engine",
+      "totalMass": "the total mass of the consist",
+      "totalLength": "the total length of the consist",
+      "maxSpeed": "the maximum speed of the consist"
+    },
+    "missingInformations": "Missing informations for calculation:",
     "missingLocation": "Missing location",
     "noPaths": "Incompatibility with other train paths.",
     "noResults": "No path found",
--- a/front/public/locales/fr/stdcm.json
+++ b/front/public/locales/fr/stdcm.json
@@ -127,6 +127,17 @@
   "stdcmErrors": {
     "bothPointAreScheduled": "Le calcul ne peut pas prendre en compte à la fois un horaire d'origine et un horaire de destination. Il faut choisir l'un ou l'autre :",
     "infraNotLoaded": "Infrastructure en cours de chargement",
+    "missingFields": {
+      "destination": "la destination",
+      "origin": "l'origine",
+      "timeDestination": "l'heure d'arrivée à la destination",
+      "timeOrigin": "l'heure de départ à l'origine",
+      "tractionEngine": "l'engin de traction",
+      "totalMass": "le poids total du convoi",
+      "totalLength": "la longueur totale du convoi",
+      "maxSpeed": "la vitesse maximale du convoi"
+    },
+    "missingInformations": "Informations manquantes pour le calcul :",
     "missingLocation": "Localisation manquante",
     "noPaths": "Incompatibilité avec d'autres sillons.",
     "noResults": "Aucun sillon trouvé",
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
@@ -5,16 +5,21 @@ import cx from 'classnames';
 import { useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
 
+import useStdcmConsist from 'applications/stdcm/hooks/useStdcmConsist';
 import { extractMarkersInfo } from 'applications/stdcm/utils';
+import filterMissingFields from 'applications/stdcm/utils/filterMissingFields';
 import DefaultBaseMap from 'common/Map/DefaultBaseMap';
 import useInfraStatus from 'modules/pathfinding/hooks/useInfraStatus';
+import { useStoreDataForRollingStockSelector } from 'modules/rollingStock/components/RollingStockSelector/useStoreDataForRollingStockSelector';
 import { resetMargins, restoreStdcmConfig, updateStdcmPathStep } from 'reducers/osrdconf/stdcmConf';
 import {
+  getStdcmConf,
   getStdcmDestination,
   getStdcmInfraID,
   getStdcmOrigin,
   getStdcmPathSteps,
   getStdcmProjectID,
+  getStdcmRollingStockID,
   getStdcmScenarioID,
   getStdcmSelectedLayers,
   getStdcmStudyID,
@@ -76,9 +81,14 @@ const StdcmConfig = ({
 
   const dispatch = useAppDispatch();
 
+  const stdcmConf = useSelector(getStdcmConf);
+
   const origin = useSelector(getStdcmOrigin);
   const pathSteps = useSelector(getStdcmPathSteps);
   const destination = useSelector(getStdcmDestination);
+  const rollingStockId = useSelector(getStdcmRollingStockID);
+  const { rollingStock } = useStoreDataForRollingStockSelector({ rollingStockId });
+  const { totalMass, totalLength, maxSpeed } = useStdcmConsist();
   const projectID = useSelector(getStdcmProjectID);
   const studyID = useSelector(getStdcmStudyID);
   const scenarioID = useSelector(getStdcmScenarioID);
@@ -98,7 +108,16 @@ const StdcmConfig = ({
   const markersInfo = useMemo(() => extractMarkersInfo(pathSteps), [pathSteps]);
 
   const startSimulation = async () => {
-    const formErrorsStatus = checkStdcmConfigErrors(pathSteps, t, pathfinding?.status);
+    const formErrorsStatus = checkStdcmConfigErrors({
+      t,
+      pathfindingStatus: pathfinding?.status,
+      stdcmConf,
+      prevFormErros: formErrors,
+      shouldCheckMandatoryFields: true,
+    });
+
+    setFormErrors(formErrorsStatus);
+
     if (pathfinding?.status === 'success' && !formErrorsStatus) {
       launchStdcmRequest();
     } else {
@@ -129,16 +148,18 @@ const StdcmConfig = ({
     return t('pathfindingStatus.success');
   };
 
+  // Checks for live warnings regarding pathSteps
   useEffect(() => {
-    const formErrorsStatus = checkStdcmConfigErrors(pathSteps, t, pathfinding?.status);
+    if (!origin.location || !destination.location) return;
+
+    const formErrorsStatus = checkStdcmConfigErrors({
+      t,
+      pathfindingStatus: pathfinding?.status,
+      stdcmConf,
+      prevFormErros: formErrors,
+    });
     setFormErrors(formErrorsStatus);
   }, [pathfinding, pathSteps, t]);
-
-  useEffect(() => {
-    if (!isDebugMode) {
-      dispatch(resetMargins());
-    }
-  }, [isDebugMode]);
 
   useEffect(() => {
     if (!infra || infra.state === 'CACHED') {
@@ -147,6 +168,12 @@ const StdcmConfig = ({
       setFormErrors({ errorType: StdcmConfigErrorTypes.INFRA_NOT_LOADED });
     }
   }, [infra]);
+
+  useEffect(() => {
+    if (!isDebugMode) {
+      dispatch(resetMargins());
+    }
+  }, [isDebugMode]);
 
   useEffect(() => {
     const state = window.osrdStdcmConfState;
@@ -179,6 +206,50 @@ const StdcmConfig = ({
       pathfindingBannerRef.current?.removeEventListener('animationend', handleAnimationEnd);
     };
   }, [showMessage, formErrors]);
+
+  useEffect(() => {
+    if (
+      formErrors?.errorType !== StdcmConfigErrorTypes.MISSING_INFORMATIONS ||
+      !formErrors.errorDetails?.missingFields
+    ) {
+      return;
+    }
+
+    if (
+      rollingStockId &&
+      totalMass &&
+      totalLength &&
+      maxSpeed &&
+      origin.location &&
+      destination.location
+    ) {
+      return;
+    }
+
+    const updatedMissingFields = filterMissingFields({
+      missingFields: formErrors.errorDetails.missingFields,
+      rollingStock,
+      totalMass,
+      totalLength,
+      maxSpeed,
+      origin,
+      destination,
+    });
+
+    const previousFields = formErrors.errorDetails.missingFields;
+    const hasChanged =
+      previousFields.length !== updatedMissingFields.length ||
+      previousFields.some((field, i) => field !== updatedMissingFields[i]);
+
+    if (!hasChanged) return;
+
+    setFormErrors({
+      errorType: StdcmConfigErrorTypes.MISSING_INFORMATIONS,
+      errorDetails: {
+        missingFields: updatedMissingFields,
+      },
+    });
+  }, [rollingStockId, totalMass, totalLength, maxSpeed, origin, destination]);
 
   return (
     <div className="stdcm__body">
@@ -219,6 +290,13 @@ const StdcmConfig = ({
                 })}
                 ref={launchButtonRef}
               >
+                {formErrors && (
+                  <StdcmWarningBox
+                    errorInfos={formErrors}
+                    removeOriginArrivalTime={removeOriginArrivalTime}
+                    removeDestinationArrivalTime={removeDestinationArrivalTime}
+                  />
+                )}
                 <Button
                   data-testid="launch-simulation-button"
                   className={cx({
@@ -232,13 +310,6 @@ const StdcmConfig = ({
                     formErrors?.errorType === StdcmConfigErrorTypes.INFRA_NOT_LOADED
                   }
                 />
-                {formErrors && (
-                  <StdcmWarningBox
-                    errorInfos={formErrors}
-                    removeOriginArrivalTime={removeOriginArrivalTime}
-                    removeDestinationArrivalTime={removeDestinationArrivalTime}
-                  />
-                )}
               </div>
 
               {!formErrors && showMessage && (
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
@@ -215,17 +215,6 @@ const StdcmConfig = ({
       return;
     }
 
-    if (
-      rollingStockId &&
-      totalMass &&
-      totalLength &&
-      maxSpeed &&
-      origin.location &&
-      destination.location
-    ) {
-      return;
-    }
-
     const updatedMissingFields = filterMissingFields({
       missingFields: formErrors.errorDetails.missingFields,
       rollingStock,
@@ -241,15 +230,29 @@ const StdcmConfig = ({
       previousFields.length !== updatedMissingFields.length ||
       previousFields.some((field, i) => field !== updatedMissingFields[i]);
 
-    if (!hasChanged) return;
+    if (updatedMissingFields.length === 0) {
+      setFormErrors(undefined);
+      return;
+    }
 
-    setFormErrors({
-      errorType: StdcmConfigErrorTypes.MISSING_INFORMATIONS,
-      errorDetails: {
-        missingFields: updatedMissingFields,
-      },
-    });
-  }, [rollingStockId, totalMass, totalLength, maxSpeed, origin, destination]);
+    if (hasChanged) {
+      setFormErrors({
+        errorType: StdcmConfigErrorTypes.MISSING_INFORMATIONS,
+        errorDetails: {
+          missingFields: updatedMissingFields,
+        },
+      });
+    }
+  }, [
+    formErrors,
+    rollingStockId,
+    rollingStock,
+    totalMass,
+    totalLength,
+    maxSpeed,
+    origin,
+    destination,
+  ]);
 
   return (
     <div className="stdcm__body">
--- a/front/src/applications/stdcm/components/StdcmWarningBox.tsx
+++ b/front/src/applications/stdcm/components/StdcmWarningBox.tsx
@@ -48,17 +48,24 @@ const StdcmWarningBox = ({
       >
         {t(`stdcmErrors.${errorType}`)}
       </p>
-      {errorType === 'bothPointAreScheduled' && errorDetails && (
+      {errorType === StdcmConfigErrorTypes.MISSING_INFORMATIONS && errorDetails?.missingFields && (
+        <div>
+          {errorDetails.missingFields.map((field) => (
+            <div key={field}>&bull;&nbsp;{t(`stdcmErrors.missingFields.${field}`)}</div>
+          ))}
+        </div>
+      )}
+      {errorType === StdcmConfigErrorTypes.BOTH_POINT_SCHEDULED && errorDetails && (
         <div className="stdcm-warning-buttons">
           <Button
             type="button"
             onClick={removeDestinationArrivalTime}
-            label={errorDetails.originTime}
+            label={errorDetails.originTime!}
           />
           <Button
             type="button"
             onClick={removeOriginArrivalTime}
-            label={errorDetails.destinationTime}
+            label={errorDetails.destinationTime!}
           />
         </div>
       )}
--- a/front/src/applications/stdcm/types.ts
+++ b/front/src/applications/stdcm/types.ts
@@ -188,11 +188,24 @@ export enum StdcmConfigErrorTypes {
   BOTH_POINT_SCHEDULED = 'bothPointAreScheduled',
   NO_SCHEDULED_POINT = 'noScheduledPoint',
   ZERO_LENGTH_PATH = 'zeroLengthPath',
+  MISSING_INFORMATIONS = 'missingInformations',
 }
+
+export type MissingFields =
+  | 'tractionEngine'
+  | 'totalMass'
+  | 'totalLength'
+  | 'maxSpeed'
+  | 'origin'
+  | 'destination';
 
 export type StdcmConfigErrors = {
   errorType: StdcmConfigErrorTypes;
-  errorDetails?: { originTime: string; destinationTime: string };
+  errorDetails?: {
+    originTime?: string;
+    destinationTime?: string;
+    missingFields?: MissingFields[];
+  };
 };
 
 export type ScheduleConstraint = {
--- a/front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
+++ b/front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
@@ -1,21 +1,56 @@
 import type { TFunction } from 'i18next';
+import { isNil } from 'lodash';
 
-import type { StdcmPathStep } from 'reducers/osrdconf/types';
+import type { OsrdStdcmConfState } from 'reducers/osrdconf/types';
 import { dateToHHMMSS } from 'utils/date';
 
-import { StdcmConfigErrorTypes, ArrivalTimeTypes, type StdcmConfigErrors } from '../types';
+import {
+  StdcmConfigErrorTypes,
+  ArrivalTimeTypes,
+  type StdcmConfigErrors,
+  type ConsistErrors,
+  type MissingFields,
+} from '../types';
 
-const checkStdcmConfigErrors = (
-  pathSteps: StdcmPathStep[],
-  t: TFunction,
-  pathfindingStatus?: 'success' | 'failure'
-): StdcmConfigErrors | undefined => {
-  if (pathSteps.some((step) => !step.location)) {
-    return { errorType: StdcmConfigErrorTypes.MISSING_LOCATION };
+const checkStdcmConfigErrors = ({
+  t,
+  pathfindingStatus,
+  stdcmConf,
+  prevFormErros,
+  shouldCheckMandatoryFields,
+}: {
+  t: TFunction;
+  pathfindingStatus?: 'success' | 'failure';
+  stdcmConf?: OsrdStdcmConfState;
+  prevFormErros?: StdcmConfigErrors;
+  consistErrors?: ConsistErrors;
+  shouldCheckMandatoryFields?: boolean;
+}): StdcmConfigErrors | undefined => {
+  const { stdcmPathSteps, rollingStockID, totalMass, totalLength, maxSpeed } = stdcmConf!;
+  const origin = stdcmPathSteps.at(0)!;
+  const destination = stdcmPathSteps.at(-1)!;
+  const missingFields: MissingFields[] = [];
+
+  if (shouldCheckMandatoryFields) {
+    if (!rollingStockID) missingFields.push('tractionEngine');
+    if (isNil(totalMass)) missingFields.push('totalMass');
+    if (isNil(totalLength)) missingFields.push('totalLength');
+    if (isNil(maxSpeed)) missingFields.push('maxSpeed');
+    if (!origin.location) {
+      missingFields.push('origin');
+    }
+    if (!destination.location) {
+      missingFields.push('destination');
+    }
+
+    if (missingFields.length > 0) {
+      return {
+        errorType: StdcmConfigErrorTypes.MISSING_INFORMATIONS,
+        errorDetails: { missingFields },
+      };
+    }
   }
 
-  const origin = pathSteps.at(0)!;
-  const destination = pathSteps.at(-1)!;
   if (origin.isVia) {
     throw new Error('First step can not be a via');
   }
@@ -24,8 +59,8 @@ const checkStdcmConfigErrors = (
   }
 
   if (
-    origin.location!.uic === destination.location!.uic &&
-    origin.location!.secondary_code === destination.location!.secondary_code
+    origin.location?.uic === destination.location?.uic &&
+    origin.location?.secondary_code === destination.location?.secondary_code
   ) {
     return { errorType: StdcmConfigErrorTypes.ZERO_LENGTH_PATH };
   }
@@ -70,7 +105,9 @@ const checkStdcmConfigErrors = (
   if (isOnePointScheduledWithoutTime) {
     return { errorType: StdcmConfigErrorTypes.NO_SCHEDULED_POINT };
   }
-  return undefined;
+  return prevFormErros?.errorType === StdcmConfigErrorTypes.MISSING_INFORMATIONS
+    ? prevFormErros
+    : undefined;
 };
 
 export default checkStdcmConfigErrors;
--- a/front/src/applications/stdcm/utils/filterMissingFields.ts
+++ b/front/src/applications/stdcm/utils/filterMissingFields.ts
@@ -1,0 +1,35 @@
+import type { RollingStockWithLiveries } from 'common/api/osrdEditoastApi';
+import type { StdcmPathStep } from 'reducers/osrdconf/types';
+
+import type { MissingFields } from '../types';
+
+type FilterMissingFields = {
+  missingFields: MissingFields[];
+  rollingStock?: RollingStockWithLiveries;
+  totalMass?: number;
+  totalLength?: number;
+  maxSpeed?: number;
+  origin?: StdcmPathStep;
+  destination?: StdcmPathStep;
+};
+
+const filterMissingFields = ({
+  missingFields,
+  rollingStock,
+  totalMass,
+  totalLength,
+  maxSpeed,
+  origin,
+  destination,
+}: FilterMissingFields): MissingFields[] =>
+  missingFields.filter((field) => {
+    if (field === 'tractionEngine' && rollingStock) return false;
+    if (field === 'totalMass' && !!totalMass) return false;
+    if (field === 'totalLength' && !!totalLength) return false;
+    if (field === 'maxSpeed' && !!maxSpeed) return false;
+    if (field === 'origin' && origin?.location) return false;
+    if (field === 'destination' && destination?.location) return false;
+    return true;
+  });
+
+export default filterMissingFields;
--- a/front/src/styles/scss/applications/stdcm/_home.scss
+++ b/front/src/styles/scss/applications/stdcm/_home.scss
@@ -184,8 +184,13 @@
         }
 
         .warning-box {
+          width: calc(100% + 16px);
+          transform: translateX(-8px);
           border-radius: 8px;
           box-shadow:
+            0 3px 5px -2px var(--black10),
+            0 16px 30px -5px rgba(0, 0, 0, 0.16),
+            0 6px 21px -5px rgba(24, 68, 239, 0.26),
             0 0 0 4px rgba(255, 255, 255, 1) inset,
             0 0 0 5px rgba(234, 167, 43, 1) inset,
             0 3px 7px -3px rgba(125, 82, 30, 0.55);
@@ -194,12 +199,11 @@
           display: flex;
           flex-direction: column;
           padding: 16px 29px 32px 29px;
-          margin-top: 12px;
+          margin-bottom: 30px;
 
           span {
             color: #eaa72b;
             align-self: center;
-            margin-bottom: 6px;
           }
         }
 
@@ -314,6 +318,9 @@
         background-color: var(--info5);
         padding-block: 22px 26px;
         box-shadow:
+          0 3px 5px -2px var(--black10),
+          0 16px 30px -5px rgba(0, 0, 0, 0.16),
+          0 6px 21px -5px rgba(24, 68, 239, 0.26),
           0 0 0 4px rgba(255, 255, 255, 1) inset,
           0 0 0 5px rgb(112, 193, 229) inset,
           0 3px 7px -3px rgba(28, 28, 217, 0.4);
@@ -322,6 +329,9 @@
           background-color: var(--success5);
           opacity: 0;
           box-shadow:
+            0 3px 5px -2px var(--black10),
+            0 16px 30px -5px rgba(0, 0, 0, 0.16),
+            0 6px 21px -5px rgba(24, 68, 239, 0.26),
             0 0 0 4px rgba(255, 255, 255, 1) inset,
             0 0 0 5px rgb(60, 202, 128) inset,
             0 3px 7px -3px rgba(11, 114, 60, 0.4);
--- a/front/tests/017-stdcm-missing-fields.spec.ts
+++ b/front/tests/017-stdcm-missing-fields.spec.ts
@@ -1,0 +1,96 @@
+import type { Infra } from 'common/api/osrdEditoastApi';
+
+import {
+  allMissingFieldsKeys,
+  getFieldsLabel,
+  partialMissingFieldsKeys,
+} from './assets/constants/missing-fields';
+import { electricRollingStockName } from './assets/constants/project-const';
+import test from './logging-fixture';
+import ConsistSection from './pages/stdcm/consist-section';
+import DestinationSection from './pages/stdcm/destination-section';
+import OriginSection from './pages/stdcm/origin-section';
+import SimulationResultPage from './pages/stdcm/simulation-results-page';
+import STDCMPage from './pages/stdcm/stdcm-page';
+import { getTranslations, waitForInfraStateToBeCached } from './utils';
+import { getInfra } from './utils/api-utils';
+import readJsonFile from './utils/file-utils';
+import type { StdcmTranslations } from './utils/types';
+
+const enTranslations: StdcmTranslations = readJsonFile('public/locales/en/stdcm.json');
+const frTranslations: StdcmTranslations = readJsonFile('public/locales/fr/stdcm.json');
+
+test.describe('Verify stdcm missing fields', () => {
+  test.slow(); // Mark test as slow due to multiple steps
+  test.use({ viewport: { width: 1920, height: 1080 } });
+
+  let stdcmPage: STDCMPage;
+  let consistSection: ConsistSection;
+  let originSection: OriginSection;
+  let destinationSection: DestinationSection;
+  let simulationResultPage: SimulationResultPage;
+  let translations: StdcmTranslations;
+  let infra: Infra;
+
+  test.beforeAll('Fetch infrastructure', async () => {
+    infra = await getInfra();
+  });
+
+  test.beforeEach('Navigate to the STDCM page', async ({ page }) => {
+    [stdcmPage, consistSection, originSection, destinationSection, simulationResultPage] = [
+      new STDCMPage(page),
+      new ConsistSection(page),
+      new OriginSection(page),
+      new DestinationSection(page),
+      new SimulationResultPage(page),
+    ];
+    translations = getTranslations({ en: enTranslations, fr: frTranslations });
+
+    await page.goto('/stdcm');
+    await page.waitForLoadState('networkidle');
+    await stdcmPage.removeViteOverlay();
+
+    // Wait for infra to be in 'CACHED' state before proceeding
+    await waitForInfraStateToBeCached(infra.id);
+  });
+
+  /** *************** Test 1 **************** */
+  test('Verify missing fields warnings when launching simulation', async () => {
+    // Step 1 — verify all missing fields
+    const allMissingLabels = getFieldsLabel(allMissingFieldsKeys, translations);
+    await stdcmPage.launchSimulationWithEmptyFields();
+    await stdcmPage.expectWarningBoxVisible();
+    await stdcmPage.expectWarningBoxContains(allMissingLabels);
+
+    // Step 2 — fill origin & destination and relaunch
+    await originSection.fillOriginDetailsLight();
+    await destinationSection.fillDestinationDetailsLight();
+    await stdcmPage.launchSimulationWithEmptyFields();
+    await stdcmPage.expectWarningBoxVisible();
+
+    const partialMissingLabels = getFieldsLabel(partialMissingFieldsKeys, translations);
+    const nonMissingLabels = getFieldsLabel(['origin', 'destination'], translations);
+    await stdcmPage.expectWarningBoxContains(partialMissingLabels, nonMissingLabels);
+  });
+
+  /** *************** Test 2 **************** */
+  test('Launch simulation with all mandatory fields filled', async () => {
+    await originSection.fillOriginDetailsLight();
+    await destinationSection.fillDestinationDetailsLight();
+    await consistSection.fillAndVerifyConsistDetails(
+      { tractionEngine: electricRollingStockName },
+      '900',
+      '400'
+    );
+
+    await stdcmPage.expectWarningBoxHidden();
+    await stdcmPage.launchSimulation();
+    await simulationResultPage.verifySimulationDetails({
+      simulationIndex: 0,
+      simulationLengthAndDuration: '51 km — 48min',
+      validSimulationNumber: 1,
+    });
+  });
+
+  // TODO: Add more tests to cover invalid fields
+});
--- a/front/tests/017-stdcm-missing-fields.spec.ts
+++ b/front/tests/017-stdcm-missing-fields.spec.ts
@@ -1,9 +1,10 @@
 import type { Infra } from 'common/api/osrdEditoastApi';
 
 import {
-  allMissingFieldsKeys,
+  ALL_MISSING_FIELDS_KEY,
   getFieldsLabel,
-  partialMissingFieldsKeys,
+  PARTIAL_MISSING_FIELDS_KEYS,
+  REMOVED_MISSING_FIELDS_KEYS,
 } from './assets/constants/missing-fields';
 import { electricRollingStockName } from './assets/constants/project-const';
 import test from './logging-fixture';
@@ -56,35 +57,29 @@ test.describe('Verify stdcm missing fields', () => {
 
   /** *************** Test 1 **************** */
   test('Verify missing fields warnings when launching simulation', async () => {
-    // Step 1 — verify all missing fields
-    const allMissingLabels = getFieldsLabel(allMissingFieldsKeys, translations);
-    await stdcmPage.launchSimulationWithEmptyFields();
+    // Step 1 — Launch simulation with all fields empty and expect all missing field warnings
+    const allMissingLabels = getFieldsLabel(ALL_MISSING_FIELDS_KEY, translations);
+    await stdcmPage.verifyInvalidSimulationLaunch();
     await stdcmPage.expectWarningBoxVisible();
     await stdcmPage.expectWarningBoxContains(allMissingLabels);
 
-    // Step 2 — fill origin & destination and relaunch
+    // Step 2 — Fill origin and destination, launch again and expect only partial missing field warnings
     await originSection.fillOriginDetailsLight();
     await destinationSection.fillDestinationDetailsLight();
-    await stdcmPage.launchSimulationWithEmptyFields();
+    await stdcmPage.verifyInvalidSimulationLaunch();
     await stdcmPage.expectWarningBoxVisible();
-
-    const partialMissingLabels = getFieldsLabel(partialMissingFieldsKeys, translations);
-    const nonMissingLabels = getFieldsLabel(['origin', 'destination'], translations);
+    const partialMissingLabels = getFieldsLabel(PARTIAL_MISSING_FIELDS_KEYS, translations);
+    const nonMissingLabels = getFieldsLabel(REMOVED_MISSING_FIELDS_KEYS, translations);
     await stdcmPage.expectWarningBoxContains(partialMissingLabels, nonMissingLabels);
-  });
 
-  /** *************** Test 2 **************** */
-  test('Launch simulation with all mandatory fields filled', async () => {
-    await originSection.fillOriginDetailsLight();
-    await destinationSection.fillDestinationDetailsLight();
+    // Step 3 — Launch simulation with all mandatory fields filled
     await consistSection.fillAndVerifyConsistDetails(
       { tractionEngine: electricRollingStockName },
       '900',
       '400'
     );
-
+    await stdcmPage.verifyValidSimulationLaunch();
     await stdcmPage.expectWarningBoxHidden();
-    await stdcmPage.launchSimulation();
     await simulationResultPage.verifySimulationDetails({
       simulationIndex: 0,
       simulationLengthAndDuration: '51 km — 48min',
--- a/front/tests/assets/constants/missing-fields.ts
+++ b/front/tests/assets/constants/missing-fields.ts
@@ -1,0 +1,22 @@
+import type { StdcmTranslations } from '../../utils/types';
+
+export const allMissingFieldsKeys = [
+  'tractionEngine',
+  'totalMass',
+  'totalLength',
+  'maxSpeed',
+  'origin',
+  'destination',
+] as const;
+
+export const partialMissingFieldsKeys = [
+  'tractionEngine',
+  'totalMass',
+  'totalLength',
+  'maxSpeed',
+] as const;
+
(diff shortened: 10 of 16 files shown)


front/src/applications/stdcm/components/StdcmWarningBox.tsx
SharglutDev, line 1: Did not review the code yet but when starting to test starting with an empty form, when I fill the origin, I'm auto-scrolled to the warning box. I think the issue mentions to display the warnings only when clicking on the launch simulation button ?

front/public/locales/en/stdcm.json
SharglutDev, line 1: Some weird bullet point in english only : 
image

front/src/applications/stdcm/components/StdcmWarningBox.tsx
SharglutDev, line 1: The box sometimes disappears even if there is still a missing field (but I guess this shouldn't happen anyway because of first comment)

[Capture vidéo du 26-02-2025 16:46:26.webm](https://github.com/user-attachments/assets/141dee70-e1fa-4f86-bdcd-48b428856ad3)

front/public/locales/fr/stdcm.json
```diff
@@ -114,6 +114,18 @@
   "stdcmErrors": {
     "bothPointAreScheduled": "Le calcul ne peut pas prendre en compte à la fois un horaire d'origine et un horaire de destination. Il faut choisir l'un ou l'autre :",
     "infraNotLoaded": "Infrastructure en cours de chargement",
+    "missingFields": {
+      "departureTime": "L'horaire de départ",
+      "destination": "La destination",
+      "origin": "L'origine",
+      "timeDestination": "l'heure de départ à la destination",
```
SharglutDev: I guess ? (same in english)
```suggestion
      "timeDestination": "l'heure d'arrivée à la destination",
```

front/public/locales/fr/stdcm.json
```diff
@@ -114,6 +114,18 @@
   "stdcmErrors": {
     "bothPointAreScheduled": "Le calcul ne peut pas prendre en compte à la fois un horaire d'origine et un horaire de destination. Il faut choisir l'un ou l'autre :",
     "infraNotLoaded": "Infrastructure en cours de chargement",
+    "missingFields": {
+      "departureTime": "L'horaire de départ",
```
SharglutDev: Same as `timeOrigin` I think ?
```suggestion
```

front/public/locales/en/stdcm.json
```diff
@@ -114,6 +114,18 @@
...
     "infraNotLoaded": "Infrastructure is loading",
+    "missingFields": {
+      "departureTime": "The departure time",
+      "destination": "The destination",
+      "origin": "The origin",
+      "timeDestination": "the arrival time at the destination",
+      "timeOrigin": "the departure time at the origin",
+      "tractionEngine": "• The traction engine",
+      "totalMass": "• The total mass of the consist",
+      "totalLength": "• The total length of the consist",
```
```suggestion
      "tractionEngine": "The traction engine",
      "totalMass": "The total mass of the consist",
      "totalLength": "The total length of the consist",
```

front/public/locales/fr/stdcm.json
SharglutDev, line 1: All texts should start with a small case like in the mockup

front/public/locales/fr/stdcm.json
```diff
@@ -114,6 +114,18 @@
...
+      "destination": "La destination",
+      "origin": "L'origine",
+      "timeDestination": "l'heure de départ à la destination",
+      "timeOrigin": "l'heure de départ à l'origine",
+      "tractionEngine": "L'engin de traction",
+      "totalMass": "Le poids total du convoi",
+      "totalLength": "La longueur totale du convoi",
+      "maxSpeed": "La vitesse maximale du convoi"
+    },
+    "missingInformations": "Informations manquantes pour le calcul :",
```
SharglutDev, line 140: Regarding this [comment](https://github.com/OpenRailAssociation/osrd/issues/9707#issuecomment-2580312374) in the issue, what has been decided ?

SharglutDev (commented):
First part of code review (trads only)

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -90,15 +96,47 @@ const StdcmConfig = ({
...
+  const [missingMandatoryFields, setMissingMandatoryFields] = useState<string[]>([]);
   const [formErrors, setFormErrors] = useState<StdcmConfigErrors>();
 
   const currentSimulationInputs = useStdcmForm();
+  const { rollingStock } = useStoreDataForRollingStockSelector();
+  const { totalMass, totalLength, maxSpeed } = useStdcmConsist();
+
+  const updateMissingFields = useCallback((fields: string[], action?: 'add' | 'remove') => {
+    setMissingMandatoryFields((prevFields) =>
+      produce(prevFields, (draft) => {
```
SharglutDev: Is there a specific reason why you used `produce` here ? 

My understanding is that it allows to mutate the state but when you use `setState((prev) => {})`, `prev` can already be mutated (see [react docs](https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state))

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -182,6 +220,30 @@ const StdcmConfig = ({
...
+  useEffect(() => {
+    if (missingMandatoryFields.length === 0) {
+      setFormErrors(undefined);
+    } else {
+      setFormErrors({
+        errorType: StdcmConfigErrorTypes.MISSING_INFORMATIONS,
+        errorDetails: { missingFields: missingMandatoryFields },
+      });
+    }
+  }, [missingMandatoryFields]);
```
SharglutDev: The issue mentionned in this [comment](https://github.com/OpenRailAssociation/osrd/pull/10913#discussion_r1971844231) comes from this. `missingMandatoryFields` is updated each time a mandatory files is updated, firing this `useEffect` who will `setFormErrors` and display the warning even if the launch button hasn't been clicked.

However, we still want to directly update the warning if a missing field has been filled.
Expected behavior is the same as for consist tooltip errors (validated by @thibautsailly) : 
1. click on launch button : display warning box if missing fields
2. fill one of the empty fields : warning box is directly updated.
3. empty the field again : warning box updates only if we click on launch button.

There are probably many ways to achieve that. One could be to check the previous state of missing fields : if the field was a missing field and isn't anymore, we `setFormErrors` and if the field was ok before we don't `setFormErrors` (will be updated when clicking on the launch button anyway)

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -90,15 +96,47 @@ const StdcmConfig = ({
...
   const markersInfo = useMemo(() => extractMarkersInfo(pathSteps), [pathSteps]);
 
   const startSimulation = async () => {
+    if (missingMandatoryFields.length > 0) {
+      setFormErrors({
+        errorType: StdcmConfigErrorTypes.MISSING_INFORMATIONS,
+        errorDetails: { missingFields: missingMandatoryFields },
+      });
+      return;
+    }
```
SharglutDev: For better maintainability, all form error logic when clicking the launch button should stay in the `checkStdcmConfigErrors`function. 

Regarding this [comment](https://github.com/OpenRailAssociation/osrd/pull/10913#discussion_r1971878329), if it has been decided to handle invalid fields in another PR, and if you want to display missing field warning in priority, you can put all the logic for that at the beginning of the function.

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
```diff
@@ -74,3 +76,28 @@ const checkStdcmConfigErrors = (
...
+}): string[] => {
+  const missingFields = [];
+  if (!rollingStock) missingFields.push('tractionEngine');
+  if (isNil(totalMass)) missingFields.push('totalMass');
+  if (isNil(totalLength)) missingFields.push('totalLength');
+  if (isNil(maxSpeed)) missingFields.push('maxSpeed');
+  if (!origin.location?.name) missingFields.push('origin');
+  if (!destination.location?.name) missingFields.push('destination');
+  return missingFields;
+};
```
SharglutDev: Regarding the previous comment, you can call this function in `checkStdcmConfigErrors` and return `errorType` and `errorDetails` if `missingFields` is not empty.

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
```diff
@@ -74,3 +76,28 @@ const checkStdcmConfigErrors = (
 };
 
 export default checkStdcmConfigErrors;
+
+export const getMissingMandatoryFields = ({
```
SharglutDev: nit : Would you mind to put the `export default` at the end of the flle ?

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
```diff
@@ -74,3 +76,28 @@ const checkStdcmConfigErrors = (
...
+  origin: StdcmPathStep;
+  destination: StdcmPathStep;
+}): string[] => {
+  const missingFields = [];
+  if (!rollingStock) missingFields.push('tractionEngine');
+  if (isNil(totalMass)) missingFields.push('totalMass');
+  if (isNil(totalLength)) missingFields.push('totalLength');
+  if (isNil(maxSpeed)) missingFields.push('maxSpeed');
+  if (!origin.location?.name) missingFields.push('origin');
+  if (!destination.location?.name) missingFields.push('destination');
```
SharglutDev: It can't really happen but the issue also asks to display the warning if the start time (for origin) or arrival time (destination) are missing.
This way in the future, if the form evolves we might catch this error easily.

front/src/applications/stdcm/components/StdcmWarningBox.tsx
SharglutDev, line 1: @Maymanaf, since we recently started to implement the e2e test at the same time of the feature, should we also at some tests for these missing fields ?

front/src/applications/stdcm/components/StdcmForm/StdcmOrigin.tsx
SharglutDev, line 1: I feel like we go to deep for these missing origin/destination checks.

When you'll add the missing field check in `checkStdcmConfigErrors`, you can update `pathSteps` `if` to check if `origin` or `destination` are missing.
We can still keep the "missing location" check if a via has been added without any location.

This way, you can remove all logic from origin, destination, opschedule and operationalpoint files.

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -194,13 +201,17 @@
...
+          margin-bottom: 32px;
 
           span {
             color: #eaa72b;
             align-self: center;
             margin-bottom: 6px;
           }
+          & ul {
+            list-style: disc;
+            margin-left: 15px;
```
SharglutDev: We need to reset the `margin-bottom` as well, otherwise the cumulated `margin-bottom` will be higher than the 32px of the mockup.
```suggestion
            margin-left: 15px;
            margin-bottom: 0;
```

SharglutDev: Using `ul` and `li` is a good idea for semantic but we can't use the default disc provided by the api, it doesn't match the one in the mockup. The one used is probably just a regular ascii character like this one : • .

You can also reset the `margin-left` and just add a space (`&nbsp;` to put in the `jsx`) between the bullet point and the text.

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -149,6 +149,8 @@
 
         /*TODO Waiting to fix the button in ui-core...*/
         .stdcm-launch-request {
+          display: flex;
+          flex-direction: column-reverse;
```
SharglutDev: We could just move the warning box before the launch button in the jsx.
```suggestion
```

front/src/applications/stdcm/components/StdcmWarningBox.tsx
Maymanaf: Yes it should be done

SharglutDev: Would you be so kind to prepare for @Caracol3 a list of what should be updated in the stdcm test ? :)

front/public/locales/fr/stdcm.json
Caracol3: I proposed to take care of invalid values in this PR too. At least prevent from launching a simulation when there is one

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
Caracol3: Yes, I used `prev` in the first version of my code, but it didn't work as expected. I had an issue where some missing fields were overwriting others. So, I searched for an alternative, and `produce` solved the problem.

The code has since changed a bit, so I might try using `prev` again.

front/src/applications/stdcm/components/StdcmWarningBox.tsx
Caracol3: With the new changes it shouldn't appear again

front/public/locales/fr/stdcm.json
Caracol3: done

front/public/locales/fr/stdcm.json
Caracol3: Yes, i deleted it

front/public/locales/en/stdcm.json
Caracol3: done

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
Caracol3: I put `prev`back and it now works as expected

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
Caracol3: With a default value, are we sure that adding start time and arrival time when we won’t use them is a good idea, knowing that we will also have to add this in the E2E tests?

SharglutDev: I'm totally fine not checking the times. You could just ask the new PO in charge of this issue if it's ok for him/her :)

front/src/applications/stdcm/components/StdcmForm/StdcmOrigin.tsx
Caracol3: done

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
Caracol3: @Melissa266 Are you ok to put the checks for start time and arrival time out of the scope ? Now that it is a default value it seems meaningless to add it

front/src/styles/scss/applications/stdcm/_home.scss
Caracol3: done

front/src/styles/scss/applications/stdcm/_home.scss
Caracol3: Yes, i put it that way because i thought that we would have two different behavior for the warning-box. I changed it

Caracol3:
@thibautsailly @Melissa266 
Now that we have the warning box displaying the absence of an origin or destination, is the mention of 'missing location' still relevant? Considering that it appears before we have even clicked the button to start the simulation, which is no longer consistent with the display of other required fields.

front/src/styles/scss/applications/stdcm/_home.scss
Caracol3: done

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
SharglutDev: Expected behavior doesn't work.

1. click on launch button : display warning box if missing fields ✅ 
2. fill one of the empty fields : warning box is directly updated. ✅ 
3. empty the field again : warning box updates only if we click on launch button. ❌ 

If I empty the field again, the warning comes back directly (should come back only if I click on the launch button)

front/src/applications/stdcm/components/StdcmWarningBox.tsx
```diff
@@ -48,17 +48,28 @@ const StdcmWarningBox = ({
...
+      {errorType === StdcmConfigErrorTypes.MISSING_INFORMATIONS &&
+        errorDetails?.missingFields?.length && (
+          <div className="warning-missing-fields">
+            {errorDetails.missingFields.map((field) => (
+              <div key={field} className="warning-item">
+                <span className="bullet">&bull;</span>&nbsp;
+                {t(`stdcmErrors.missingFields.${field}`)}
+              </div>
+            ))}
+          </div>
```
SharglutDev: It seems like we don't need any class or `span` here
```suggestion
          <div>
            {errorDetails.missingFields.map((field) => (
              <div key={field}>&bull;&nbsp;{t(`stdcmErrors.missingFields.${field}`)}</div>
            ))}
          </div>
```

front/src/styles/scss/applications/stdcm/_home.scss
```diff
@@ -194,12 +199,27 @@
           display: flex;
           flex-direction: column;
           padding: 16px 29px 32px 29px;
-          margin-top: 12px;
+          margin-bottom: 32px;
```
SharglutDev: Looks like it's 30px afterall ?
Capture d’écran 2025-03-06 à 12 17 08

```suggestion
          margin-bottom: 30px;
```

front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
```diff
@@ -152,7 +152,7 @@ const StdcmOperationalPoint = ({
   };
 
   useEffect(() => {
-    if (location) {
+    if (location && location.name) {
```
SharglutDev: `name` always exists in `location`
```suggestion
    if (location) {
```

front/src/applications/stdcm/components/StdcmForm/StdcmOrigin.tsx
```diff
@@ -28,7 +28,7 @@ const StdcmOrigin = ({ disabled = false, onItineraryChange }: StdcmItineraryProp
         disabled={disabled}
         onItineraryChange={onItineraryChange}
       />
-      <StdcmOpSchedule pathStep={origin} disabled={disabled} opId="origin-arrival" isOrigin />
+      <StdcmOpSchedule pathStep={origin} disabled={disabled} opId="origin-arrival" />
```
SharglutDev: We need to keep `isOrigin` otherwise we won't have the proper option in the select : 

Capture d’écran 2025-03-06 à 12 26 13

```suggestion
      <StdcmOpSchedule pathStep={origin} disabled={disabled} opId="origin-arrival" isOrigin/>
```

front/src/applications/stdcm/components/StdcmWarningBox.tsx
```diff
@@ -48,17 +48,28 @@ const StdcmWarningBox = ({
       >
         {t(`stdcmErrors.${errorType}`)}
       </p>
+      {errorType === StdcmConfigErrorTypes.MISSING_INFORMATIONS &&
+        errorDetails?.missingFields?.length && (
```
SharglutDev: `missingFields` is always > 0 if `errorType === StdcmConfigErrorTypes.MISSING_INFORMATIONS`
```suggestion
      {errorType === StdcmConfigErrorTypes.MISSING_INFORMATIONS &&
        errorDetails?.missingFields && (
```

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -185,6 +219,49 @@ const StdcmConfig = ({
...
+  }, [
+    pathSteps,
+    t,
+    pathfinding?.status,
+    rollingStock,
+    totalMass,
+    totalLength,
+    maxSpeed,
+    isSimulationLaunched,
+  ]);
```
SharglutDev: Regarding this [comment](https://github.com/OpenRailAssociation/osrd/pull/10913#discussion_r1983151362), I think if you drop this `useEffect`, it will be fixed

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -185,6 +219,49 @@ const StdcmConfig = ({
...
+
+  useEffect(() => {
+    const prevMissingFields = prevMissingFieldsRef.current;
+
+    if (missingMandatoryFields.length === 0) {
+      setFormErrors(undefined);
+    } else if (
+      isSimulationLaunched &&
+      prevMissingFields.length > 0 &&
+      prevMissingFields.length !== missingMandatoryFields.length
```
SharglutDev: Probably because of this check, if I have an empty length, then I empty the mass, the warning box will be updated even if I don't click on the launch button.
Not sure if there is another way but we should rather comparer the old value and new value for each "watched" field. If the prev was empty and the new isn't, we update `formErrors` otherwise we don't do anything.

edit : see this related [comment](https://github.com/OpenRailAssociation/osrd/pull/10913#discussion_r1983281637)

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
```diff
@@ -8,14 +10,42 @@ import { StdcmConfigErrorTypes, ArrivalTimeTypes, type StdcmConfigErrors } from
 const checkStdcmConfigErrors = (
   pathSteps: StdcmPathStep[],
   t: TFunction,
-  pathfindingStatus?: 'success' | 'failure'
+  pathfindingStatus?: 'success' | 'failure',
+  rollingStock?: RollingStockWithLiveries,
+  totalMass?: number | undefined,
+  totalLength?: number | undefined,
+  maxSpeed?: number | undefined,
```
SharglutDev: Instead of listing (and having to pass) everything, we could pass `stdcmConf` to the function and destructure all the fields we need here. 
With this we can remove `pathSteps` from the params (we have `stdcmPathSteps` in stdcm conf) and we could use `rollingStockID` from the conf instead of `rollingStock`

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
```diff
@@ -8,14 +10,42 @@ import { StdcmConfigErrorTypes, ArrivalTimeTypes, type StdcmConfigErrors } from
...
+  isSimulationLaunched?: boolean
```
SharglutDev: This way it's more obvious when we use the function
```suggestion
  options?: { isSimulationLaunched: boolean }
```

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -88,21 +91,52 @@ const StdcmConfig = ({
 
   const [showMessage, setShowMessage] = useState(false);
 
+  const [isSimulationLaunched, setIsSimulationLaunched] = useState(false);
```
SharglutDev: We don't need this state, we can just pass `{
      isSimulationLaunched: true,
    }` to `checkStdcmConfigErrors` when we call it when launching the simulation
```suggestion
```

front/src/applications/stdcm/types.ts
```diff
@@ -185,11 +186,12 @@ export enum StdcmConfigErrorTypes {
   BOTH_POINT_SCHEDULED = 'bothPointAreScheduled',
   NO_SCHEDULED_POINT = 'noScheduledPoint',
   ZERO_LENGTH_PATH = 'zeroLengthPath',
+  MISSING_INFORMATIONS = 'missingInformations',
 }
 
 export type StdcmConfigErrors = {
   errorType: StdcmConfigErrorTypes;
-  errorDetails?: { originTime: string; destinationTime: string };
+  errorDetails?: { originTime?: string; destinationTime?: string; missingFields?: string[] };
```
SharglutDev: We should reinforce the `missingFields` type by creating a type : 
```
export type MissingField =
  | 'tractionEngine'
  | 'totalMass'
  | 'totalLength'
  | 'maxSpeed'
  | 'origin'
  | 'destination';
```
```suggestion
  errorDetails?: { originTime?: string; destinationTime?: string; missingFields?: MissingField[] };
```

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -185,6 +219,49 @@ const StdcmConfig = ({
...
+      prevMissingFields.length !== missingMandatoryFields.length
+    ) {
+      setFormErrors({
+        errorType: StdcmConfigErrorTypes.MISSING_INFORMATIONS,
+        errorDetails: { missingFields: missingMandatoryFields },
+      });
+    }
+
+    prevMissingFieldsRef.current = missingMandatoryFields;
+  }, [missingMandatoryFields]);
```
SharglutDev: To prevent some extra rerenders because of `missingMandatoryFields` state, we could do this `useEffect` like this : 
```suggestion
  useEffect(() => {
    if (formErrors && formErrors.errorType === StdcmConfigErrorTypes.MISSING_INFORMATIONS) {
      if (rollingStockId && totalMass && totalLength && maxSpeed && origin && destination) {
        setFormErrors(undefined);
      }
      setFormErrors((prevErrors) => {
        if (prevErrors?.errorDetails?.missingFields?.includes('totalMass') && !!totalMass) {
          return {
            errorType: StdcmConfigErrorTypes.MISSING_INFORMATIONS,
            errorDetails: {
              missingFields: prevErrors.errorDetails.missingFields.filter(
                (field) => field !== 'totalMass'
              ),
            },
          };
        }
        if (prevErrors?.errorDetails?.missingFields?.includes('totalLength') && !!totalLength) {
          return {
            errorType: StdcmConfigErrorTypes.MISSING_INFORMATIONS,
            errorDetails: {
              missingFields: prevErrors.errorDetails.missingFields.filter(
                (field) => field !== 'totalLength'
              ),
            },
          };
        }
        return prevErrors;
      });
    }
  }, [rollingStockId, totalMass, totalLength, maxSpeed, origin, destination]);
```
It's messy for now but it was just to test that it worked and it also fixes the other comment regarding this `useEffect`. We should do the same checks for all fields and it would probably better to extract this checks in a function (which would receive `prevErrors` and all mandatory fields, or just a generic field and do the checks inside) somewhere else not to pollute this file which is already quite big.

front/src/applications/stdcm/components/StdcmForm/StdcmOrigin.tsx
Caracol3: I put it back, it was a mistake

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
Melissa-LASSAL: I completely agree with both of you about not checking the time. At worst, if the user realizes that the time of their simulation is set to 00:00, they can start another one.

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
```diff
@@ -8,14 +10,42 @@ import { StdcmConfigErrorTypes, ArrivalTimeTypes, type StdcmConfigErrors } from
...
 ): StdcmConfigErrors | undefined => {
+  const destination = pathSteps.at(-1)!;
+  const origin = pathSteps.at(0)!;
+
+  if (!origin || !destination) {
+    return { errorType: StdcmConfigErrorTypes.MISSING_LOCATION };
+  }
```
Synar: This check will succeed as long as `pathSteps `has at least one step (with `origin === destination`), and is equivalent to `if (!origin)`. 

Instead we could check that `length(pathSteps)>=2`.

Synar: It would be cleaner to do it before our origin and destination definitions too, because we are currently using ! to declare they are defined before checking whether they are.

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
Caracol3: done

front/src/applications/stdcm/components/StdcmWarningBox.tsx
Caracol3: done

front/src/styles/scss/applications/stdcm/_home.scss
Caracol3: done

front/src/applications/stdcm/components/StdcmForm/StdcmOperationalPoint.tsx
Caracol3: done

front/src/applications/stdcm/components/StdcmWarningBox.tsx
Caracol3: done

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
Caracol3: done

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
Caracol3: seen with @thibautsailly , i deleted the `MISSING.LOCATION` warning. It isn't relevant now we have the check on every input including origin and destination

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
Caracol3: The logic is different now

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
Caracol3: Good catch !
Done

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
Caracol3: done

front/src/applications/stdcm/types.ts
Caracol3: done

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
Caracol3: At the time of the review, `isSimulationLaunched` and `hasAttemptedSimulation` were only used to call checkStdcmConfigErrors, so we could pass them directly without using state.
Now, they are used in multiple parts of the component logic. I think they must remain as useState to trigger updates.

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
Caracol3: I put the logic of the missingFields checks in a util. 
Not exactly like you did but works the same way

front/src/applications/stdcm/components/StdcmWarningBox.tsx
SharglutDev: Looks like they are still there for the invalid informations

Caracol3: Yes, i'm dealing with invalidFields now. It will be cleaner after the next push

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
```diff
@@ -48,9 +49,21 @@ const ConsistCardTitle = ({
 export type StdcmConsistProps = {
   isDebugMode: boolean;
   disabled?: boolean;
+  setMissingMandatoryFields?: (
```
SharglutDev: This is always defined (you can also remove the `if` l.156)

```suggestion
  setMissingMandatoryFields: (
```

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
```diff
@@ -24,8 +68,10 @@ const checkStdcmConfigErrors = (
   }
 
   if (
-    origin.location!.uic === destination.location!.uic &&
-    origin.location!.secondary_code === destination.location!.secondary_code
+    origin.location &&
+    destination.location &&
+    origin.location.uic === destination.location.uic &&
+    origin.location.secondary_code === destination.location.secondary_code
```
SharglutDev:
```suggestion
    origin.location?.uic === destination.location?.uic &&
    origin.location?.secondary_code === destination.location?.secondary_code
```

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
```diff
@@ -24,8 +68,10 @@ const checkStdcmConfigErrors = (
...
   ) {
     return { errorType: StdcmConfigErrorTypes.ZERO_LENGTH_PATH };
```
SharglutDev, line 65: The warning box for this use case doesn't work anymore. The box seems to blink even though the `formErrors` is properly set with the good `errorType` (see console)

https://github.com/user-attachments/assets/b5be4c8e-f711-4104-8af0-cff117908d4f

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -76,29 +82,114 @@ const StdcmConfig = ({
...
   const [formErrors, setFormErrors] = useState<StdcmConfigErrors>();
 
+  const [consistErrors, setConsistErrors] = useState<ConsistErrors>({
+    totalMass: { message: undefined, display: false },
+    totalLength: { message: undefined, display: false },
+    maxSpeed: { message: undefined, display: false },
+  });
+
+  const [persistedErrors, setPersistedErrors] = useState<MissingFields[]>([]);
+  const [hasAttemptedSimulation, setHasAttemptedSimulation] = useState(false);
```
SharglutDev: This state is a bit weird, we never set it to false

front/src/applications/stdcm/utils/filterMissingFields.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (field === 'maxSpeed' && maxSpeed !== undefined) return false;
+    if (field === 'origin' && origin?.location) return false;
+    if (field === 'destination' && destination?.location) return false;
+    return true;
+  });
+
+  return Array.from(
+    new Set([
+      ...updatedMissingFields,
+      ...updatedPersistedErrors.filter((field) => !updatedMissingFields.includes(field)),
```
SharglutDev: You shouldn't have to `filter` here since `Set` already stores unique values.

front/src/applications/stdcm/types.ts
```diff
@@ -168,6 +168,10 @@ export type StdcmConfigCardProps = {
   disabled?: boolean;
   consistErrors?: ConsistErrors;
   isDebugMode?: boolean;
+  setMissingMandatoryFields?: (
+    missingMandatoryFields: MissingFields[],
+    action?: 'add' | 'remove'
+  ) => void;
```
SharglutDev: Looks like we never use this prop in `StdcmOrigin` and `StdcmDestination`

```suggestion
```

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
```diff
@@ -1,21 +1,65 @@
...
+} from '../types';
 
 const checkStdcmConfigErrors = (
-  pathSteps: StdcmPathStep[],
   t: TFunction,
-  pathfindingStatus?: 'success' | 'failure'
+  pathfindingStatus?: 'success' | 'failure',
+  stdcmConf?: OsrdStdcmConfState,
+  consistErrors?: ConsistErrors,
+  options: { isSimulationLaunched?: boolean; hasAttemptedSimulation?: boolean } = {}
```
SharglutDev: This function starts to have many parameters, we should put them in an object to avoid errors when we call it

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
SharglutDev: I still think we can use this to simplify a lot the code, we should talk about it

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
Caracol3: It's now deleted

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
Caracol3: Good catch ! Done

front/src/applications/stdcm/types.ts
Caracol3: done

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -140,14 +174,6 @@ const StdcmConfig = ({
     }
   }, [isDebugMode]);
 
-  useEffect(() => {
-    if (!infra || infra.state === 'CACHED') {
-      setFormErrors(undefined);
-    } else {
-      setFormErrors({ errorType: StdcmConfigErrorTypes.INFRA_NOT_LOADED });
-    }
-  }, [infra]);
```
SharglutDev: This one ended up being tossed in the refacto, we should keep it

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -93,12 +103,28 @@ const StdcmConfig = ({
 
   const [formErrors, setFormErrors] = useState<StdcmConfigErrors>();
 
+  const [consistErrors, setConsistErrors] = useState<ConsistErrors>({
+    totalMass: { message: undefined, display: false },
+    totalLength: { message: undefined, display: false },
+    maxSpeed: { message: undefined, display: false },
+  });
```
SharglutDev: Since we decided to handle consist errors in another PR, could you remove the changes you made for that ?

front/src/applications/stdcm/components/StdcmWarningBox.tsx
```diff
@@ -19,12 +19,14 @@ type StdcmWarningBoxProps = {
   };
   removeOriginArrivalTime: () => void;
   removeDestinationArrivalTime: () => void;
+  // persistedErrors?: string[];
```
SharglutDev: You can remove this commented line (same on l.29 and l.56)

front/src/applications/stdcm/utils/filterMissingFields.ts
```diff
@@ -0,0 +1,38 @@
...
+    if (field === 'tractionEngine' && rollingStock) return false;
+    if (field === 'totalMass' && !!totalMass) return false;
+    if (field === 'totalLength' && !!totalLength) return false;
+    if (field === 'maxSpeed' && !!maxSpeed) return false;
+    if (field === 'origin' && origin?.location) return false;
+    if (field === 'destination' && destination?.location) return false;
+    return true;
+  });
+
+  return Array.from(new Set([...updatedMissingFields]));
```
SharglutDev: Since for now we handle only missing fields, we don't need the `Set`, there won't be any duplicates. This function can only handle the `.filter()`

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
Caracol3: Oh indeed !

front/src/applications/stdcm/utils/filterMissingFields.ts
Caracol3: done

front/src/applications/stdcm/components/StdcmWarningBox.tsx
Caracol3: done

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
Caracol3: done, the state is back in `stdcmConsist`

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
SharglutDev, line 1: The diffs in this file should be removed

front/src/applications/stdcm/utils/filterMissingFields.ts
```diff
@@ -0,0 +1,38 @@
...
+    if (field === 'totalMass' && !!totalMass) return false;
+    if (field === 'totalLength' && !!totalLength) return false;
+    if (field === 'maxSpeed' && !!maxSpeed) return false;
+    if (field === 'origin' && origin?.location) return false;
+    if (field === 'destination' && destination?.location) return false;
+    return true;
+  });
+
+  return updatedMissingFields;
+};
```
SharglutDev:
```suggestion
}: FilterMissingFields): MissingFields[] => missingFields.filter((field) => {
    if (field === 'tractionEngine' && rollingStock) return false;
    if (field === 'totalMass' && !!totalMass) return false;
    if (field === 'totalLength' && !!totalLength) return false;
    if (field === 'maxSpeed' && !!maxSpeed) return false;
    if (field === 'origin' && origin?.location) return false;
    if (field === 'destination' && destination?.location) return false;
    return true;
  });
```

front/tests/016-stdcm-missing-fields.spec.ts
SharglutDev, line 1: Test 016 is already used since the merge of this [PR](https://github.com/OpenRailAssociation/osrd/pull/10966), you can change it for 17

front/src/applications/stdcm/components/StdcmWarningBox.tsx
```diff
@@ -48,17 +48,35 @@ const StdcmWarningBox = ({
...
+        errorDetails?.invalidFields?.length && (
+          <div className="warning-invalid-fields">
+            {errorDetails.invalidFields.map((field) => (
+              <div key={field} className="warning-item">
+                <span className="bullet">&bull;</span>&nbsp;
+                {t(`stdcmErrors.missingFields.${field}`)}
+              </div>
+            ))}
+          </div>
+        )}
```
SharglutDev: This can go for now I think

front/src/applications/stdcm/types.ts
```diff
@@ -188,11 +188,26 @@ export enum StdcmConfigErrorTypes {
   BOTH_POINT_SCHEDULED = 'bothPointAreScheduled',
   NO_SCHEDULED_POINT = 'noScheduledPoint',
   ZERO_LENGTH_PATH = 'zeroLengthPath',
+  MISSING_INFORMATIONS = 'missingInformations',
+  INVALID_INFORMATIONS = 'invalidInformations',
```
SharglutDev: This can go for now
```suggestion
```

front/src/applications/stdcm/types.ts
```diff
@@ -188,11 +188,26 @@ export enum StdcmConfigErrorTypes {
...
+  | 'destination';
+
 export type StdcmConfigErrors = {
   errorType: StdcmConfigErrorTypes;
-  errorDetails?: { originTime: string; destinationTime: string };
+  errorDetails?: {
+    originTime?: string;
+    destinationTime?: string;
+    missingFields?: MissingFields[];
+    invalidFields?: string[];
```
SharglutDev:
```suggestion
```

front/src/applications/stdcm/components/StdcmWarningBox.tsx
```diff
@@ -48,17 +48,35 @@ const StdcmWarningBox = ({
...
+      {errorType === StdcmConfigErrorTypes.BOTH_POINT_SCHEDULED && errorDetails && (
```
SharglutDev, line 58: One issue where this error doesn't pop : 
1. fill all fields for a stdcm request
2. remove the destination
3. launch the request (the missing destination warning shows)
4. change the destination for the same one as the origin
5. the warning for missing field disappears but we don't have the "both schedulded" one

front/tests/assets/constants/missing-fields.ts
```diff
@@ -0,0 +1,35 @@
...
+export const getMissingFields = (fields: readonly MissingFields[]) => {
+  const translations = getTranslations({
+    en: enTranslations,
+    fr: frTranslations,
+  });
+
+  return fields.map((field) => translations.stdcmErrors.missingFields[field]);
+};
+
+export default getMissingFields;
```
SharglutDev:
```suggestion
```

front/tests/assets/constants/missing-fields.ts
```diff
@@ -0,0 +1,35 @@
...
+  'maxSpeed',
+] as const;
+
+export type MissingFields = (typeof allMissingFieldsKeys)[number];
+
+export const getMissingFields = (fields: readonly MissingFields[]) => {
+  const translations = getTranslations({
+    en: enTranslations,
+    fr: frTranslations,
+  });
```
SharglutDev: To setup the translations in the test, we should rather harmonize with the other test (like the other 16 for example, see l.62)

front/tests/pages/stdcm/stdcm-page.ts
```diff
@@ -132,10 +132,35 @@ class STDCMPage extends HomePage {
     await this.simulationStatus.waitFor({ timeout: STDCM_SIMULATION_TIMEOUT });
   }
 
+  // Launch the simualtion with mandatory fields empty
+  async launchSimulationWithEmptyFields() {
+    await this.launchSimulationButton.click();
+    await expect(this.launchSimulationButton).toBeEnabled();
+    await this.launchSimulationButton.click({ force: true });
```
SharglutDev: Why do we click 2 times on the launch button ?

front/tests/pages/stdcm/stdcm-page.ts
```diff
@@ -132,10 +132,35 @@ class STDCMPage extends HomePage {
...
+    await expect(this.page.locator('[data-testid="simulation-status"]')).not.toBeVisible();
```
SharglutDev: You can create a variable in the constructor for this : 
```
this.simulationStatus = page.locator('[data-testid="simulation-status"]');
```

front/tests/pages/stdcm/stdcm-page.ts
```diff
@@ -132,10 +132,35 @@ class STDCMPage extends HomePage {
...
+  }
+
   async mapMarkerVisibility() {
     await expect(this.originMarker).toBeVisible();
     await expect(this.destinationMarker).toBeVisible();
     await expect(this.viaMarker).toBeVisible();
   }
+
+  async expectWarningBoxVisible() {
+    await expect(this.page.getByTestId('warning-box')).toBeVisible();
```
SharglutDev: Add a variable in constructor for the warning box and use it here, l.154 and in l.159

front/tests/016-stdcm-missing-fields.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    await stdcmPage.expectWarningBoxContains(allMissingLabels);
+
+    // Step 2 — fill origin & destination and relaunch
+    await originSection.fillOriginDetailsLight();
+    await destinationSection.fillDestinationDetailsLight();
+    await stdcmPage.launchSimulationWithEmptyFields();
+    await stdcmPage.expectWarningBoxVisible();
+
+    const partialMissingLabels = getMissingFields(partialMissingFieldsKeys);
+    await stdcmPage.expectWarningBoxContains(partialMissingLabels);
```
SharglutDev: Hmm this tests that the remaining missing fields are still there but not that origin and destination are not there anymore.
Maybe we could pass another param to this method to check which field the box doesn't contain (and skip this part if the param is undefined)

front/tests/016-stdcm-missing-fields.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+  });
+
+  /** *************** Test 2 **************** */
```
SharglutDev: Except if it's a request from @Maymanaf but part 3 of the request [tests](https://github.com/osrd-project/osrd-tests/issues/106) specifies that after filling the consist part, we should check that the warning box disappeared (before launching the simulation).
You could move l.73 to l.83 in the previous test.

front/tests/pages/stdcm/stdcm-page.ts
Caracol3: I updated the first one. It was a `waitFor()` not a `click()`

front/src/applications/stdcm/components/StdcmWarningBox.tsx
Caracol3: Good catch (but it was actually the `ZERO_LENGTH_PATH` error that should have been triggered).
I’ve updated the useEffect to return earlier so it doesn’t overwrite other errors. Now it works has expected.

front/tests/pages/stdcm/stdcm-page.ts
```diff
@@ -86,6 +86,8 @@ class STDCMPage extends HomePage {
     this.closeTimePickerButton = page.locator('.time-picker .close-button');
     this.warningBox = page.getByTestId('warning-box');
     this.incrementButton = page.locator('.minute-button', { hasText: '+1mn' });
+    this.simulationStatus = page.locator('[data-testid="simulation-status"]');
+    this.warningBox = page.locator(`[data-testid="warning-box"]`);
```
Maymanaf: duplicated

front/tests/pages/stdcm/stdcm-page.ts
```diff
@@ -132,10 +134,39 @@ class STDCMPage extends HomePage {
...
     await expect(this.destinationMarker).toBeVisible();
     await expect(this.viaMarker).toBeVisible();
   }
+
+  async expectWarningBoxVisible() {
+    await expect(this.warningBox).toBeVisible();
+  }
+
+  async expectWarningBoxHidden() {
+    await expect(this.page.getByTestId('warning-box')).toBeHidden();
```
Maymanaf:
```suggestion
    await expect(this.warningBox).toBeHidden();
```

front/tests/pages/stdcm/stdcm-page.ts
```diff
@@ -86,6 +86,8 @@ class STDCMPage extends HomePage {
     this.closeTimePickerButton = page.locator('.time-picker .close-button');
     this.warningBox = page.getByTestId('warning-box');
     this.incrementButton = page.locator('.minute-button', { hasText: '+1mn' });
+    this.simulationStatus = page.locator('[data-testid="simulation-status"]');
```
Maymanaf: duplicated

front/tests/pages/stdcm/stdcm-page.ts
```diff
@@ -132,10 +134,39 @@ class STDCMPage extends HomePage {
...
   }
 
+  // Launch the simualtion with mandatory fields empty
+  async launchSimulationWithEmptyFields(): Promise<void> {
+    await this.launchSimulationButton.waitFor();
+    await expect(this.launchSimulationButton).toBeEnabled();
+    await this.launchSimulationButton.click({ force: true });
+    await expect(this.simulationStatus).not.toBeVisible();
+  }
+
```
Maymanaf: Extract the duplicated logic from launchSimulation() and launchSimulationWithEmptyFields()
```suggestion
async launchSimulation(): Promise<void> {
  await this.launchSimulationButton.waitFor();
  await expect(this.launchSimulationButton).toBeEnabled();
  await this.launchSimulationButton.click({ force: true });
}

async verifyValidSimulationLaunch(): Promise<void> {
  await this.launchSimulation();
  await this.simulationStatus.waitFor({ timeout: STDCM_SIMULATION_TIMEOUT });
}

async verifyInvalidSimulationLaunch(): Promise<void> {
  await this.launchSimulation();
  await expect(this.simulationStatus).not.toBeVisible();
}
```

front/tests/016-stdcm-missing-fields.spec.ts
Maymanaf: Like @SharglutDev said, it should be 1 test that covers the 3 steps

front/tests/017-stdcm-missing-fields.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+    await stdcmPage.removeViteOverlay();
+
+    // Wait for infra to be in 'CACHED' state before proceeding
+    await waitForInfraStateToBeCached(infra.id);
+  });
+
+  /** *************** Test 1 **************** */
+  test('Verify missing fields warnings when launching simulation', async () => {
+    // Step 1 — verify all missing fields
+    const allMissingLabels = getFieldsLabel(allMissingFieldsKeys, translations);
```
SharglutDev: I believe for const we use capital cases in snake case (same for `partialMissingLabels`)
```suggestion
    const allMissingLabels = getFieldsLabel(ALL_MISSING_FIELDS_KEY, translations);
```

front/tests/017-stdcm-missing-fields.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+    await stdcmPage.expectWarningBoxContains(allMissingLabels);
+
+    // Step 2 — fill origin & destination and relaunch
+    await originSection.fillOriginDetailsLight();
+    await destinationSection.fillDestinationDetailsLight();
+    await stdcmPage.launchSimulationWithEmptyFields();
+    await stdcmPage.expectWarningBoxVisible();
+
+    const partialMissingLabels = getFieldsLabel(partialMissingFieldsKeys, translations);
+    const nonMissingLabels = getFieldsLabel(['origin', 'destination'], translations);
```
SharglutDev: To harmonize with the rest of the test, we should also put `['origin', 'destination']` in a const

front/tests/017-stdcm-missing-fields.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+    await stdcmPage.expectWarningBoxContains(partialMissingLabels, nonMissingLabels);
+
+    // Step 3 — Launch simulation with all mandatory fields filled
+    await consistSection.fillAndVerifyConsistDetails(
+      { tractionEngine: electricRollingStockName },
+      '900',
+      '400'
+    );
+    await stdcmPage.expectWarningBoxHidden();
+    await stdcmPage.verifyValidSimulationLaunch();
```
Maymanaf:
```suggestion
    await stdcmPage.verifyValidSimulationLaunch();
    await stdcmPage.expectWarningBoxHidden();
```

front/tests/pages/stdcm/stdcm-page.ts
```diff
@@ -128,14 +128,45 @@ class STDCMPage extends HomePage {
...
-    // Wait for simulation message "Calculation completed"
+  }
+
+  // Wait for simulation message "Calculation completed"
+  async verifyValidSimulationLaunch(): Promise<void> {
+    await this.launchSimulation();
     await this.simulationStatus.waitFor({ timeout: STDCM_SIMULATION_TIMEOUT });
   }
 
+  // Launch the simualtion with mandatory fields empty
```
Maymanaf: unnecessary comment

front/tests/pages/stdcm/stdcm-page.ts
```diff
@@ -128,14 +128,45 @@ class STDCMPage extends HomePage {
     await this.launchSimulationButton.waitFor();
     await expect(this.launchSimulationButton).toBeEnabled();
     await this.launchSimulationButton.click({ force: true });
-    // Wait for simulation message "Calculation completed"
+  }
+
+  // Wait for simulation message "Calculation completed"
```
Maymanaf: unnecessary comment

front/tests/017-stdcm-missing-fields.spec.ts
Caracol3: done

front/tests/pages/stdcm/stdcm-page.ts
Caracol3: done

front/tests/017-stdcm-missing-fields.spec.ts
Caracol3: done

front/tests/pages/stdcm/stdcm-page.ts
Caracol3: done

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -180,6 +207,50 @@ const StdcmConfig = ({
     };
   }, [showMessage, formErrors]);
 
+  useEffect(() => {
```
SharglutDev, line 210: We can simplify the useEffect like this (it seems to work fine)

```suggestion
  useEffect(() => {
    if (
      formErrors?.errorType !== StdcmConfigErrorTypes.MISSING_INFORMATIONS ||
      !formErrors.errorDetails?.missingFields
    ) {
      return;
    }

    const updatedMissingFields = filterMissingFields({
      missingFields: formErrors.errorDetails.missingFields,
      rollingStock,
      totalMass,
      totalLength,
      maxSpeed,
      origin,
      destination,
    });

    if (updatedMissingFields.length === 0) {
      setFormErrors(undefined);
      return;
    }

    setFormErrors({
      errorType: StdcmConfigErrorTypes.MISSING_INFORMATIONS,
      errorDetails: {
        missingFields: updatedMissingFields,
      },
    });
  }, [rollingStockId, totalMass, totalLength, maxSpeed, origin, destination]);
```

To only thing we have to think about is what behavior we want when we have both error types at the same time (missing field + other). I asked @thibautsailly about that, waiting for his answer

thibautsailly: Warnings should be displayed as different sections of the same message bloc, in this priority order:
1. Origin and destination timing preferences incompatibility
2. Information invalidity problems
3. Missing informations

Each section should order their warning in the reading order of the form: consist, origin, intermediate OPs, destination.

SharglutDev: This will be handle in another PR. Could you create the issue @Caracol3 ?

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
```diff
@@ -180,6 +207,53 @@ const StdcmConfig = ({
...
+      totalLength,
+      maxSpeed,
+      origin,
+      destination,
+    });
+
+    const previousFields = formErrors.errorDetails.missingFields;
+    const hasChanged =
+      previousFields.length !== updatedMissingFields.length ||
+      previousFields.some((field, i) => field !== updatedMissingFields[i]);
```
SharglutDev, line 231: Do we need the first check ?
```suggestion
    const hasChanged = previousFields.some((field, i) => field !== updatedMissingFields[i]);
```

SharglutDev (approved):
Lgtm and tested, great work :)

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
Caracol3: `some()`only compares items at the same indexes, so if a new field is added at the end of the list, it won’t detect any difference.

front/src/applications/stdcm/components/StdcmForm/StdcmConfig.tsx
Caracol3: Yes, it will be handled in https://github.com/OpenRailAssociation/osrd/issues/11278

Maymanaf (approved):
LGTM for e2e tests ✅

PR #10914, opened 2025-02-24
front: use Duration in StdcmPathStep tolerances

emersion:
Improves type safety.

### DIFF
--- a/front/src/applications/stdcm/utils/formatStdcmConf.ts
+++ b/front/src/applications/stdcm/utils/formatStdcmConf.ts
@@ -165,8 +165,8 @@ export const checkStdcmConf = (
       if (arrivalType === 'preciseTime' && arrival) {
         timingData = {
           arrival_time: arrival.toISOString(),
-          arrival_time_tolerance_before: tolerances?.before?.ms ?? 0,
-          arrival_time_tolerance_after: tolerances?.after?.ms ?? 0,
+          arrival_time_tolerance_before: tolerances.before.ms,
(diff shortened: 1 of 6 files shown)


Akctarus (approved):
tested, works well, lgtm !

front/src/applications/stdcm/utils/formatStdcmConf.ts
```diff
@@ -165,8 +165,8 @@ export const checkStdcmConf = (
       if (arrivalType === 'preciseTime' && arrival) {
         timingData = {
           arrival_time: arrival.toISOString(),
-          arrival_time_tolerance_before: sec2ms(tolerances?.before ?? 0),
-          arrival_time_tolerance_after: sec2ms(tolerances?.after ?? 0),
+          arrival_time_tolerance_before: tolerances?.before?.ms ?? 0,
+          arrival_time_tolerance_after: tolerances?.after?.ms ?? 0,
```
SharglutDev: Not part of your PR but here `tolerances` (and `before` / `after`) are always defined, we don't need the `?` (we just put it in the origin or destination path item depending of which one has `arrivalType === 'preciseTime'`)
```suggestion
          arrival_time_tolerance_before: tolerances.before.ms,
          arrival_time_tolerance_after: tolerances.after.ms,
```

emersion: Added a commit to clean this up!

SharglutDev (approved):
Lgtm and tested, thanks !

PR #10918, opened 2025-02-24
front: fix type for macro nodes full name

### LINKED ISSUE #10390: nge: nodes have no name
### What happened?

Image

Recognized OP have no name, but we still can set it manually:

Image

_No response_

1. create train schedule from osrd

Recette (SNCF)

Firefox

-

### LINKED ISSUE #10433: nge: full name of OPs recognized in OSRD not displayed
### What happened?

Prevents feature https://github.com/osrd-project/osrd-confidential/issues/597 from being validated

If a node in NGE corresponds to an operational point recognized in the infrastructure, the full_name should be displayed in the name, but it is not : 

Image

As a reminder, the ACs of the feature https://github.com/osrd-project/osrd-confidential/issues/597 were :  

- If a node in NGE has a name stored in the nodes table, then it is displayed
- Otherwise, if the node corresponds to a recognized OP in OSRD, then the full_name attribute of the OP is displayed in the name field in NGE (without it being stored in the table)
- Else, nothing is displayed in the name field
- If the user changes the name attribute in NGE, then it is stored in the table and displayed in NGE (even if the node corresponds to a recognized OP in OSRD for which we previously displayed the full_name by default)

Note : the first AC in the previous list seems ok.

Create a train in OSRD going through OPs, and go to NGE, double click on one of the nodes.

Recette (SNCF)

Google Chrome

c9066aa

louisgreiner:
Close https://github.com/OpenRailAssociation/osrd/issues/10433
Close https://github.com/OpenRailAssociation/osrd/issues/10390

### DIFF
--- a/front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
@@ -180,8 +180,8 @@ export const loadAndIndexNge = async (
   // Enhance nodes by calling the search API
   const searchResults = await executeSearch(state, dispatch);
   searchResults.forEach((searchResult) => {
-    const macroNode = {
-      fullName: searchResult.name,
+    const macroNode: Pick<NodeIndexed, 'full_name' | 'trigram' | 'geocoord'> = {
+      full_name: searchResult.name,
       trigram: searchResult.trigram + (searchResult.ch ? `/${searchResult.ch}` : ''),
       geocoord: {
         lng: searchResult.geographic.coordinates[0],
--- a/front/src/applications/operationalStudies/components/MacroEditor/utils.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/utils.ts
@@ -68,6 +68,7 @@ export const createMacroNode = async (
     console.error(e);
   }
 };
+
 export const updateMacroNode = async (
   state: MacroEditorState,
   dispatch: AppDispatch,


front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
```diff
@@ -180,8 +180,8 @@ export const loadAndIndexNge = async (
   // Enhance nodes by calling the search API
   const searchResults = await executeSearch(state, dispatch);
   searchResults.forEach((searchResult) => {
-    const macroNode = {
-      fullName: searchResult.name,
+    const macroNode: Partial<NodeIndexed> = {
```
SharglutDev: nit: `Partial` gives all type keys an `undefined` possibility, even the one we used. It's not the case here but this would allow one of the 3 properties we use to accept an `undefined` value. If you think this can lead to a unwanted behavior in the future, you can replace `Partial` by `Pick<NodeIndexed, 'full_name' | 'trigram' | 'geocoord'>` (down side is you have to list all the keys you want)

louisgreiner: Thanks for the explanation, I don't know for now, but it surely could lead to unwanted behavior in the future -> fixed

emersion (approved):
Good catch!

Uriel-Sautron (approved):
Nice, thank you

PR #10945, opened 2025-02-25
editoast: update paced train endpoint

### LINKED ISSUE #10591: Update a paced train endpoint
### Description and goal

Add new endpoint to update a paced train
`PUT paced_train/{id}`

### Acceptance criteria

- [ ] Enable updating of paced trains

Wadjetz:
fix #10591

part of https://github.com/osrd-project/osrd-confidential/issues/779

### DIFF
--- a/editoast/editoast_schemas/src/primitives/duration.rs
+++ b/editoast/editoast_schemas/src/primitives/duration.rs
@@ -57,7 +57,7 @@ pub enum PositiveDurationError {
 /// let err_s = r#"{"duration":"P1M"}"#; // 1 month
 /// assert!(serde_json::from_str::<MyStruct>(err_s).is_err());
 /// ```
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct PositiveDuration(ChronoDuration);
 
 impl TryFrom<ChronoDuration> for PositiveDuration {
--- a/editoast/src/models/fixtures.rs
+++ b/editoast/src/models/fixtures.rs
@@ -14,7 +14,7 @@ use editoast_schemas::infra::DirectionalTrackRange;
 use editoast_schemas::infra::InfraObject;
 use editoast_schemas::infra::RailJson;
 use editoast_schemas::paced_train::Paced;
-use editoast_schemas::paced_train::PacedTrainBase;
+use editoast_schemas::paced_train::PacedTrain;
 use editoast_schemas::primitives::OSRDObject;
 use editoast_schemas::rolling_stock::EffortCurves;
 use editoast_schemas::rolling_stock::LoadingGaugeType;
@@ -31,6 +31,7 @@ use serde::Deserialize;
 use serde_json::Value;
 
 use crate::infra_cache::operation::create::apply_create_operation;
+use crate::models;
 use crate::models::electrical_profiles::ElectricalProfileSet;
 use crate::models::prelude::*;
 use crate::models::rolling_stock_livery::RollingStockLiveryModel;
@@ -46,7 +47,6 @@ use crate::models::Scenario;
 use crate::models::Study;
 use crate::models::Tags;
 
-use super::paced_train::PacedTrain;
 use super::temporary_speed_limits::TemporarySpeedLimitGroup;
 
 pub fn project_changeset(name: &str) -> Changeset<Project> {
@@ -96,11 +96,11 @@ pub fn simple_train_schedule_base() -> TrainScheduleBase {
         .expect("Unable to parse test train schedule")
 }
 
-pub fn simple_paced_train_base() -> PacedTrainBase {
+pub fn simple_paced_train_base() -> PacedTrain {
     let train_schedule_base =
         serde_json::from_str(include_str!("../tests/train_schedules/simple.json"))
             .expect("Unable to parse test train schedule");
-    PacedTrainBase {
+    PacedTrain {
         train_schedule_base,
         paced: Paced {
             duration: ChronoDuration::hours(2).try_into().unwrap(),
@@ -113,8 +113,8 @@ pub fn simple_train_schedule_changeset(timetable_id: i64) -> Changeset<TrainSche
     Changeset::<TrainSchedule>::from(simple_train_schedule_base()).timetable_id(timetable_id)
 }
 
-pub fn simple_paced_train_changeset(timetable_id: i64) -> Changeset<PacedTrain> {
-    Changeset::<PacedTrain>::from(simple_paced_train_base()).timetable_id(timetable_id)
+pub fn simple_paced_train_changeset(timetable_id: i64) -> Changeset<models::PacedTrain> {
+    Changeset::<models::PacedTrain>::from(simple_paced_train_base()).timetable_id(timetable_id)
 }
 
 pub async fn create_simple_train_schedule(
@@ -127,7 +127,10 @@ pub async fn create_simple_train_schedule(
         .expect("Failed to create train schedule")
 }
 
-pub async fn create_simple_paced_train(conn: &mut DbConnection, timetable_id: i64) -> PacedTrain {
+pub async fn create_simple_paced_train(
+    conn: &mut DbConnection,
+    timetable_id: i64,
+) -> models::PacedTrain {
     simple_paced_train_changeset(timetable_id)
         .create(conn)
         .await
--- a/editoast/src/models/paced_train.rs
+++ b/editoast/src/models/paced_train.rs
@@ -2,8 +2,8 @@ use chrono::DateTime;
 use chrono::Duration as ChronoDuration;
 use chrono::Utc;
 use editoast_derive::Model;
+use editoast_schemas::paced_train;
 use editoast_schemas::paced_train::Paced;
-use editoast_schemas::paced_train::PacedTrainBase;
 use editoast_schemas::train_schedule::Comfort;
 use editoast_schemas::train_schedule::Distribution;
 use editoast_schemas::train_schedule::Margins;
@@ -18,6 +18,7 @@ use crate::models::prelude::*;
 use crate::models::train_schedule::TrainSchedule;
 
 #[derive(Debug, Clone, Model)]
+#[cfg_attr(test, derive(PartialEq))]
 #[model(table = editoast_models::tables::paced_train)]
 #[model(gen(ops = crud, batch_ops = crd, list))]
 pub struct PacedTrain {
@@ -72,12 +73,12 @@ impl PacedTrain {
     }
 }
 
-impl From<PacedTrainBase> for PacedTrainChangeset {
+impl From<paced_train::PacedTrain> for PacedTrainChangeset {
     fn from(
-        PacedTrainBase {
+        paced_train::PacedTrain {
             train_schedule_base,
             paced,
-        }: PacedTrainBase,
+        }: paced_train::PacedTrain,
     ) -> Self {
         PacedTrain::changeset()
             .comfort(train_schedule_base.comfort)
@@ -98,7 +99,7 @@ impl From<PacedTrainBase> for PacedTrainChangeset {
     }
 }
 
-impl From<PacedTrain> for PacedTrainBase {
+impl From<PacedTrain> for paced_train::PacedTrain {
     fn from(paced_train: PacedTrain) -> Self {
         Self {
             train_schedule_base: TrainScheduleBase {
--- a/editoast/src/views/paced_train.rs
+++ b/editoast/src/views/paced_train.rs
@@ -10,7 +10,7 @@ use axum::Extension;
 use editoast_authz::Role;
 use editoast_derive::EditoastError;
 use editoast_models::DbConnectionPoolV2;
-use editoast_schemas::paced_train::PacedTrainBase;
+use editoast_schemas::paced_train::PacedTrain;
 use serde::Deserialize;
 use serde::Serialize;
 use thiserror::Error;
@@ -28,7 +28,8 @@ use super::InfraIdQueryParam;
 use super::SimulationSummaryResult;
 use crate::core::simulation::SimulationResponse;
 use crate::error::Result;
-use crate::models::paced_train::PacedTrain;
+use crate::models;
+use crate::models::paced_train::PacedTrainChangeset;
 use crate::models::prelude::*;
 use crate::models::train_schedule::TrainSchedule;
 use crate::models::Infra;
@@ -51,9 +52,9 @@ crate::routes! {
 }
 
 editoast_common::schemas! {
-    PacedTrainResult,
+    PacedTrainResponse,
     PacedTrainForm,
-    PacedTrainBase,
+    PacedTrain,
 }
 
 #[derive(Debug, Error, EditoastError)]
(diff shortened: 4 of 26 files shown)


editoast/src/views/paced_train.rs
```diff
@@ -128,6 +129,43 @@ async fn get_by_id(
     Ok(Json(paced_train))
 }
 
+/// Update a train schedule and its result
```
```suggestion
/// Update a paced train and its result
```

what is the "result" referring to?

editoast/src/views/paced_train.rs
```diff
@@ -128,6 +129,43 @@ async fn get_by_id(
...
+#[utoipa::path(
+    put, path = "",
+    tag = "timetable,paced_train",
+    params(PacedTrainIdParam),
+    request_body = inline(PacedTrainBase),
+    responses(
+        (status = 204, body = PacedTrainResult, description = "Paced train have been updated")
```
leovalais: 204 with a content? 🤔

editoast/src/views/paced_train.rs
```diff
@@ -128,6 +129,43 @@ async fn get_by_id(
...
+    let authorized = auth
+        .check_roles([BuiltinRole::InfraRead, BuiltinRole::TimetableWrite].into())
+        .await
+        .map_err(AuthorizationError::AuthError)?;
+    if !authorized {
+        return Err(AuthorizationError::Forbidden.into());
+    }
+
+    let conn = &mut db_pool.get().await?;
+    let paced_train_changeset: PacedTrainChangeset = paced_train_base.into();
```
leovalais, line 167:
```suggestion
    let paced_train_changeset = paced_train_base.into_changeset();
```

Wadjetz: `paced_train_base` is not a model but a schema, it do not implement `into_changeset`

leovalais: Oh right indeed...

woshilapin (approved):
Note, the commit's title has a typo ("endpont" instead of "endpoint"), could you fix that please? 🙏

editoast/editoast_schemas/src/primitives/duration.rs
```diff
@@ -57,7 +57,7 @@ pub enum PositiveDurationError {
 /// let err_s = r#"{"duration":"P1M"}"#; // 1 month
 /// assert!(serde_json::from_str::<MyStruct>(err_s).is_err());
 /// ```
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
```
leovalais:
```suggestion
#[derive(Debug, Clone)]
#[cfg_attr(test, derive(PartialEq, Eq))]
```

editoast/src/views/paced_train.rs
```diff
@@ -137,6 +137,43 @@ async fn get_by_id(
...
+    responses(
+        (status = 200, body = PacedTrainResult, description = "Paced train have been updated")
+    )
+)]
+async fn update_paced_train(
+    State(db_pool): State<DbConnectionPoolV2>,
+    Extension(auth): AuthenticationExt,
+    Path(PacedTrainIdParam { id: paced_train_id }): Path<PacedTrainIdParam>,
+    Json(paced_train_base): Json<PacedTrainBase>,
+) -> Result<impl IntoResponse> {
```
leovalais, line 157:
```suggestion
) -> Result<PacedTrainResult> {
```

editoast/src/views/paced_train.rs
```diff
@@ -137,6 +137,43 @@ async fn get_by_id(
...
+    let authorized = auth
+        .check_roles([BuiltinRole::OperationalStudies, BuiltinRole::Stdcm].into())
```
leovalais: why Stcdm?

editoast/src/views/paced_train.rs
```diff
@@ -446,6 +463,32 @@ mod tests {
...
+            .json(&json!(&paced_train_base));
+
+        let response: PacedTrainResult =
+            app.fetch(request).assert_status(StatusCode::OK).json_into();
+
+        assert_eq!(
+            response.paced_train.paced.duration,
+            paced_train_base.paced.duration
+        );
+        assert_eq!(response.paced_train.paced.step, paced_train_base.paced.step);
```
leovalais: A stronger check would retrieve the updated PacedTrain from the DB and check that it equals to `paced_train_base`. With such tests it's easy to break the endpoint while still making the test pass.

editoast/src/views/paced_train.rs
```diff
@@ -137,6 +137,43 @@ async fn get_by_id(
...
 }
 
+/// Update a paced train
+#[utoipa::path(
+    put, path = "",
+    tag = "timetable,paced_train",
+    params(PacedTrainIdParam),
+    request_body = inline(PacedTrainBase),
+    responses(
+        (status = 200, body = PacedTrainResult, description = "Paced train have been updated")
```
```suggestion
        (status = 204, description = "The paced train has been updated")
```

no need to return the request body unchanged

editoast/src/views/paced_train.rs
leovalais: This is the definition of `PacedTrainResult` which *I suppose* is what the "result" refers to here. Can't it just be named `PacedTrain`? It would make the OpenAPI schema clearer as well. Wdyt? @Wadjetz @younesschrifi

editoast/editoast_schemas/src/primitives/duration.rs
Wadjetz: It’s a different crate, and #[cfg_attr(test, derive(PartialEq, Eq))] does not work because the attributes only apply within the same crate.

leovalais: Right... the drawbacks of splitting into different crates... 😑

editoast/src/views/paced_train.rs
Wadjetz: Without `body = PacedTrainResult` we lose the type in openapi and TS types

leovalais: I'm suggesting to not return anything for this endpoint as it's a PUT and not a PATCH, so its request's payload won't differ from its response's

editoast/src/views/paced_train.rs
Wadjetz: It would be nice to have a naming convention, such as *Model/*Row for database structs, just the name for OpenAPI structs, and *Railjson for RailJSON for examples ?

leovalais: My take is that RailJson objects are supposed to represent the objects completely, so I'd strip the *Railjson suffix for those. But if they are augmented with additional information (for a view response, like `PacedTRainResult` here), I'd say that's where we need a suffix. Since `PAcedTrainResult` only exists for view responses, I'd suggest the following names, **especially for the OpenAPI**:

* `PacedTrain` instead of `PacedTrainBase`, since the `Base` suffix doesn't mean anything in the `editoast_schemas` context
* `PacedTrainResponse` instead of `PacedTrainResult`
* We can do the same of `TrainSchedule` definitions

Wdyt? @younesschrifi @flomonster

editoast/src/views/paced_train.rs
leovalais: :/

editoast/src/views/paced_train.rs
```diff
@@ -472,6 +488,36 @@ mod tests {
...
+        let updated_paced_train = PacedTrain::retrieve(&mut pool.get_ok(), paced_train.id)
+            .await
+            .expect("Failed to retrieve updated paced train")
+            .expect("Updated paced train not found");
+
+        assert_eq!(
+            updated_paced_train.duration,
+            paced_train_base.paced.duration.into()
+        );
+        assert_eq!(updated_paced_train.step, paced_train_base.paced.step.into());
```
```suggestion
        assert_eq!(
            updated_paced_train,
            paced_train_base.into()
        );
```

editoast/src/views/paced_train.rs
flomonster: What you are suggesting sound good.

Wadjetz: If we rename PacedTrainBase to PacedTrain, what should we name the current PacedTrain (model)? PacedTrainModel ? @flomonster @leovalais @younesschrifi

woshilapin: We can keep `PacedTrain` too, the namespace is different 🤷

editoast/src/models/paced_train.rs
```diff
@@ -18,6 +18,7 @@ use crate::models::prelude::*;
 use crate::models::train_schedule::TrainSchedule;
 
 #[derive(Debug, Clone, Model)]
+#[cfg_attr(test, derive(PartialEq))]
```
woshilapin, line 21: Note that if `PartialEq` is derived only for tests, then the same should be done for all the other objects on which deriving `PartialEq` (and `Eq`) has been done.

editoast/src/views/paced_train.rs
woshilapin: Why is this resolved since it's not been fixed?

woshilapin (approved):

editoast/src/views/paced_train.rs
leovalais: > We can keep PacedTrain too, the namespace is different 🤷

Agreed for code, but we still need to tell apart the model and the schema in the OpenAPI. I can see two ways to handle this:

1. Never return the model the endpoints and always convert it into the schema. That works well with the concept of "editoast_schemas". Not sure how well this applies to all our models though.
2. Have a prefix like `Stored-` or `Persisted-` (`Model` doesn't really makes sense in the OpenAPI imo, I see it more as a programming class of objects)

editoast/src/views/paced_train.rs
Wadjetz: It’s fixed, but not in this line. We no longer send PacedTrainResult, but a no-content code.

editoast/src/views/paced_train.rs
Wadjetz: Sorry, I didn’t respond. I actually assumed that train schedules have the Stdcm role and that paced trains should too, but I’m not sure if paced trains are used in Stdcm.

editoast/src/models/paced_train.rs
Wadjetz: We can’t, the others objects are part of a different crate, editoast_schema. It’s the drawback of splitting into different crates.

editoast/src/views/paced_train.rs
Wadjetz: I find it quite confusing to use the same name, even if it’s in a different namespace. Personally, I would prefer adding a prefix like ‘Stored’ or ‘Persisted’.

Would you like me to make this change? If so, should I adjust the prefix or the different namespaces?

younesschrifi (approved):

editoast/src/views/paced_train.rs
leovalais: The names are defined in two different namespaces, `models::PacedTrainModels` is redundant imo, whereas `models::PacedTrain` conveys all the information we need. And if we're in a context without ambiguity, we can always `use models::PacedTrain`.

But we need the prefix/suffix in the OpenAPI since all schemas are defined in the same namespace though.

Wadjetz: Hello @leovalais, I’ve made the changes. Could you please review them?

editoast/src/models/fixtures.rs
```diff
@@ -113,8 +113,8 @@ pub fn simple_train_schedule_changeset(timetable_id: i64) -> Changeset<TrainSche
     Changeset::<TrainSchedule>::from(simple_train_schedule_base()).timetable_id(timetable_id)
 }
 
-pub fn simple_paced_train_changeset(timetable_id: i64) -> Changeset<PacedTrain> {
-    Changeset::<PacedTrain>::from(simple_paced_train_base()).timetable_id(timetable_id)
+pub fn simple_paced_train_changeset(timetable_id: i64) -> Changeset<PacedTrainModel> {
```
leovalais:
```suggestion
pub fn simple_paced_train_changeset(timetable_id: i64) -> Changeset<models::PacedTrain> {
```

conveys the same information and avoids the indirection caused by the aliasing

editoast/src/models/paced_train.rs
```diff
@@ -3,7 +3,7 @@ use chrono::Duration as ChronoDuration;
 use chrono::Utc;
 use editoast_derive::Model;
 use editoast_schemas::paced_train::Paced;
-use editoast_schemas::paced_train::PacedTrainBase;
+use editoast_schemas::paced_train::PacedTrain as PacedTrainSchema;
```
leovalais: I'd argue that using the fully qualified path enhances readability in this context

editoast/src/views/timetable.rs
```diff
@@ -39,7 +39,7 @@ use crate::core::conflict_detection::TrainRequirements;
 use crate::core::simulation::SimulationResponse;
 use crate::core::AsCoreRequest;
 use crate::error::Result;
-use crate::models::paced_train::PacedTrain;
+use crate::models::paced_train::PacedTrain as PacedTrainModel;
```
```suggestion
use crate::models;
```
likewise

leovalais (commented):
Thanks for the renaming! Though I think aliasing is superfluous in most cases (I believe @woshilapin agrees 👀). Could you open an enhancement issue for doing the same renaming work for TrainSchedules?

front/src/common/api/generatedEditoastApi.ts
```diff
@@ -3255,22 +3261,18 @@ export type TrainScheduleBase = {
...
+export type PacedTrain = TrainScheduleBase & {
   paced: {
     /** Duration of the paced train, an ISO 8601 format is expected */
     duration: string;
     /** Time between two occurrences, an ISO 8601 format is expected */
     step: string;
   };
 };
-export type PacedTrainResult = PacedTrainBase & {
+export type PacedTrainResponse = PacedTrain & {
```
SharglutDev, line 3272: Hmm why do we want to change this naming ? It was handy to have the same syntax for both paced trains and train schedules.
Now we will have : 
- PacedTrain / TrainScheduleBase
- PacedTrainResponse / TrainScheduleResult
I think we should harmonize this

front/src/modules/powerRestriction/helpers/formatPowerRestrictionRangesWithHandled.ts
```diff
@@ -64,7 +64,7 @@ export const convertPowerRestrictionsAndCheckCompatibility = (
 
         // add the restriction corresponding to the voltage range, format the position to km
         restrictionsWithHandled.push({
-          position: { start: mmToKm(start), end: mmToKm(end!) },
+          position: { start: mmToKm(start), end: end ? mmToKm(end) : undefined },
```
SharglutDev: Why do we change this ? Does this fix a bug ?

Wadjetz: It appeared as a type compilation error, which I fixed, but I believe it was just a merge conflict that went wrong.

front/src/common/api/generatedEditoastApi.ts
Wadjetz: I created a [ticket](https://github.com/OpenRailAssociation/osrd/issues/11249) to harmonize it, and it follows the resolution of this comment: https://github.com/OpenRailAssociation/osrd/pull/10945#discussion_r1979402676.

Maymanaf (approved):

front/tests/utils/paced-train.ts
```diff
@@ -9,12 +9,12 @@ import { getApiContext, handleErrorResponse } from './api-utils';
  *
  * @param timetableId - The ID of the timetable for which the paced trains are being sent.
  * @param body - The request payload containing paced train data.
- * @returns {Promise<PacedTrainResult[]>} - The API response containing the train schedule results.
+ * @returns {Promise<PacedTrainResponse[]>} - The API response containing the train schedule results.
```
SharglutDev:
```suggestion
 * @returns {Promise<PacedTrainResponse[]>} - The API response containing the paced train response.
```

front/src/modules/trainschedule/components/Timetable/PacedTrain/PacedTrainItem.tsx
```diff
@@ -142,7 +142,7 @@ const PacedTrainItem = ({
       return;
     }
 
-    const formattedTrainScheduleResult: PacedTrainResultWithPacedTrainId = {
+    const formattedTrainScheduleResult: PacedTrainResponseWithPacedTrainId = {
```
SharglutDev:
```suggestion
    const formattedPacedTrainResponse: PacedTrainResponseWithPacedTrainId = {
```

SharglutDev (approved):

leovalais (approved):
LGTM, thanks!

PR #10947, opened 2025-02-25
front: fix useEffet warning in useScenarioData

### LINKED ISSUE #10917: useEffect warning when entering a scenario
### What happened?

When entering a scenario with train, we can see this warning in the console : 

Image

It probably comes from this [PR](https://github.com/OpenRailAssociation/osrd/pull/10655)

No warning

Go to a scenario with trains and open the console

Local

Firefox

.

SharglutDev:
The call to get the train schedule results was recently moved into a useQuery. By defaulting the respone with an empty array and using it in some useEffect dependencies, the useEffects were called multiple times (since the ref of the array changes at each rerender) while the endpoint was loading.

fix #10917


theocrsb (approved):
LGTM and tested

emersion (approved):
LGTM!

PR #10949, opened 2025-02-25
front: fix operational points names display in itinerary

### LINKED ISSUE #10117: Name of waypoints not properly displayed afterViriato import
### What happened?

I import this train [train_schedules.19.json](https://github.com/user-attachments/files/19587265/train_schedules.19.json)

Then I open it and put a rolling stock on it. Then I go the itinerary tab. The names of waypoints are not displayed :  
Image

Then if I save the train and reopen it, I have the correct names :  
Image

I expect to have the correct names at the first step, without having to save and reopen my train.

1. Import the json file
2. Open the MSCAXP06-1 train
3. Put a rolling stock on it
4. Go to the itinerary tab -> observe the problem
5. Save and reopen the train and go to the itinerary tab -> observe the expected behavior from the beginning

Dev (SNCF)

Google Chrome

fe3b339

RomainValls:
close #10117 

The title says the issue happened after a viriato import. 
In fact, it was happening with all types of files during import when the name wasn't given in the file.

### DIFF
--- a/front/src/modules/operationalPoint/helpers/buildOpSearchQuery.ts
+++ b/front/src/modules/operationalPoint/helpers/buildOpSearchQuery.ts
@@ -1,18 +1,15 @@
-import type { SearchPayload, SearchQuery, TrainScheduleResult } from 'common/api/osrdEditoastApi';
+import type { SearchPayload, SearchQuery } from 'common/api/osrdEditoastApi';
+import type { PathStep } from 'reducers/osrdconf/types';
 
 /**
  * Build a search query to fetch all operational points from their UICs,
  * trigrams and IDs.
  */
-const buildOpSearchQuery = (
-  infraId: number,
-  trainSchedules: TrainScheduleResult[]
-): SearchPayload | null => {
-  const pathItems = trainSchedules.map((schedule) => schedule.path).flat();
+const buildOpSearchQuery = (infraId: number, pathSteps: PathStep[]): SearchPayload | null => {
   const pathItemQueries = [];
   const pathItemSet = new Set<string>();
 
-  for (const item of pathItems) {
+  for (const item of pathSteps) {
     let query: SearchQuery;
     if ('uic' in item) {
       query = ['=', ['uic'], item.uic];
--- a/front/src/modules/pathfinding/hooks/usePathfinding.ts
+++ b/front/src/modules/pathfinding/hooks/usePathfinding.ts
@@ -11,8 +11,10 @@ import type {
   PathfindingInputError,
   PathfindingResultSuccess,
   PostInfraByInfraIdPathPropertiesApiArg,
+  SearchResultItemOperationalPoint,
 } from 'common/api/osrdEditoastApi';
 import { osrdEditoastApi } from 'common/api/osrdEditoastApi';
+import buildOpSearchQuery from 'modules/operationalPoint/helpers/buildOpSearchQuery';
 import {
   formatSuggestedOperationalPoints,
   getPathfindingQuery,
@@ -55,7 +57,6 @@ const usePathfinding = ({
   const powerRestrictions = useSelector(getPowerRestrictions);
   const { infraId, getTrackSectionsByIds } = useScenarioContext();
   const { infra, reloadCount, setIsInfraError } = useInfraStatus({ infraId });
-
   const [pathfindingState, setPathfindingState] =
     useState<PathfindingState>(initialPathfindingState);
 
@@ -65,6 +66,53 @@ const usePathfinding = ({
     osrdEditoastApi.endpoints.postInfraByInfraIdPathfindingBlocks.useLazyQuery();
   const [postPathProperties] =
     osrdEditoastApi.endpoints.postInfraByInfraIdPathProperties.useLazyQuery();
+
+  const [postSearch] = osrdEditoastApi.endpoints.postSearch.useMutation();
+
+  const fetchOperationalPoints = useCallback(
+    async (steps: PathStep[]) => {
+      if (!infraId || !steps.length) return;
+
+      const searchPayload = buildOpSearchQuery(infraId, steps);
+
+      if (!searchPayload) return;
+      let results: SearchResultItemOperationalPoint[];
+      try {
+        results = (await postSearch({
+          searchPayload,
+          pageSize: 101,
+        }).unwrap()) as SearchResultItemOperationalPoint[];
+      } catch (error) {
+        console.error('Error fetching operational points:', error);
+        return;
+      }
+
+      const updatedSteps = steps.map((step) => {
+        let matchedOp: SearchResultItemOperationalPoint | undefined;
+        // TODO we should handle the case where the step is missing a secondary_code, as a path step who doesn't specify a secondary_code matches any ch.
+        if ('uic' in step && step.uic) {
+          matchedOp = results.find(
+            (op) => op.uic === step.uic && (!step.secondary_code || op.ch === step.secondary_code)
+          );
+        } else if ('trigram' in step && step.trigram) {
+          matchedOp = results.find(
+            (op) =>
+              op.trigram === step.trigram && (!step.secondary_code || op.ch === step.secondary_code)
+          );
+        } else if ('operational_point' in step && step.operational_point) {
+          matchedOp = results.find((op) => op.obj_id === step.operational_point);
+        }
+
+        return {
+          ...step,
+          name: matchedOp?.name,
+        };
+      });
+
+      dispatch(updatePathSteps(updatedSteps));
+    },
+    [infraId]
+  );
 
   const setIsMissingParam = () =>
     setPathfindingState({ ...initialPathfindingState, isMissingParam: true });
@@ -205,6 +253,7 @@ const usePathfinding = ({
 
       if (!pathfindingInput) {
         setIsMissingParam();
+        await fetchOperationalPoints(pathSteps.filter((step) => step !== null));
         return;
       }
 
--- a/front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
+++ b/front/src/applications/operationalStudies/components/MacroEditor/osrdToNge.ts
@@ -37,7 +37,8 @@ const executeSearch = async (
   state: MacroEditorState,
   dispatch: AppDispatch
 ): Promise<SearchResultItemOperationalPoint[]> => {
-  const searchPayload = buildOpSearchQuery(state.scenario.infra_id, state.trainSchedules);
+  const pathSteps = state.trainSchedules.flatMap((ts) => ts.path);
+  const searchPayload = buildOpSearchQuery(state.scenario.infra_id, pathSteps);
   if (!searchPayload) {
     return [];
   }


clarani (commented):
Thank you for your PR @RomainValls :pray: 

It's exactly what should be done, but I think it might be better to move this code in usePathfinding, what do you think ?
If there is no pathfindingInput, then we should call your code to retrieve the additional data of the path steps, even if we can't launch a pathfinding

front/src/modules/pathfinding/components/Itinerary/DisplayItinerary/Vias.tsx
```diff
@@ -19,6 +28,74 @@ const Vias = ({ zoomToFeaturePoint }: ViasProps) => {
...
+    const validPathSteps = pathSteps.filter(
+      (step): step is PathStep & { uic: number; secondary_code?: string } =>
+        step !== null && 'uic' in step
+    );
+
+    if (!validPathSteps.length) return [];
+
+    const uicChConditions = validPathSteps.map(({ uic, secondary_code }) => [
+      'and',
+      ['=', ['uic'], uic],
```
emersion: This handles the UIC case, but not other cases. Other cases might be used, e.g. trigram.

`buildOpSearchQuery()` already handles all of the cases, maybe we could just use that?

front/src/modules/pathfinding/components/Itinerary/DisplayItinerary/Vias.tsx
```diff
@@ -19,6 +28,74 @@ const Vias = ({ zoomToFeaturePoint }: ViasProps) => {
...
+  const [postSearch] = osrdEditoastApi.endpoints.postSearch.useMutation();
+  const dispatch = useAppDispatch();
+
+  const hasUpdatedNames = useRef(false);
+
+  const searchOperationalPointsByUIC = useCallback(async () => {
+    if (!infraId || !pathSteps.length) return [];
+
+    const validPathSteps = pathSteps.filter(
+      (step): step is PathStep & { uic: number; secondary_code?: string } =>
```
Synar: I would prefer something like 'step is Extract<PathStep, { uic: number }>' to assume as little as possible from what we actually check, but in this case I believe the cast is entirely unnecessary and typescript will do the correct assumptions by itself.
```suggestion
      (step) =>
```

front/src/modules/pathfinding/components/Itinerary/DisplayItinerary/Vias.tsx
```diff
@@ -19,6 +28,74 @@ const Vias = ({ zoomToFeaturePoint }: ViasProps) => {
...
+      const opMap = results.reduce(
+        (acc, op) => {
+          acc[`87${op.ci}`] = op.name;
+          return acc;
+        },
+        {} as Record<string, string>
+      );
+
+      const updatedSteps = pathSteps
+        .filter((step): step is PathStep => step !== null)
```
Synar:
```suggestion
      (step) =>
```

Synar (commented):
Nice work, this was an annoying issue indeed ^^

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -66,6 +68,58 @@ const usePathfinding = ({
...
+        } else if ('operational_point' in step && step.operational_point) {
+          name = opMap[step.operational_point];
+        }
+
+        return {
+          ...step,
+          name,
+        };
+      });
+      dispatch(updatePathSteps(updatedSteps));
```
emersion, line 112: As a more general comment: I don't really like the pattern where a `useEffect()` listens for `pathSteps` changes, then transforms the data and overwrites path steps again by calling `updatePathSteps()`. It's very complicated to reason about the data flow (since `pathSteps` may be mutated basically anywhere in the app) and it might result in infinite feedback loops.

In fact, doesn't this PR make the app search for OPs in an infinite loop?

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -66,6 +68,58 @@ const usePathfinding = ({
   const [postPathProperties] =
     osrdEditoastApi.endpoints.postInfraByInfraIdPathProperties.useLazyQuery();
 
+  const [postSearch] = osrdEditoastApi.endpoints.postSearch.useMutation();
+
+  const fetchOperationalPoints = useCallback(async () => {
+    if (!infraId || !pathSteps.length) return;
+
+    const searchPayload = buildOpSearchQuery(infraId, [
+      { path: pathSteps.filter((step): step is PathStep => step !== null) } as TrainScheduleResult,
```
emersion: I think instead of casting (and sending a value which isn't actually a `TrainScheduleResult`), we can adapt `buildOpSearchQuery()` to take as input a list of path steps instead of `TrainScheduleResult[]` (ideally in a separate commit before this call is added here).

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -66,6 +68,58 @@ const usePathfinding = ({
...
+        pageSize: 101,
+      }).unwrap()) as SearchResultItemOperationalPoint[];
+
+      if (!results.length) return;
+
+      const opMap = results.reduce(
+        (acc, op) => {
+          acc[`87${op.ci}`] = op.name;
+          acc[op.trigram] = op.name;
+          acc[op.obj_id] = op.name;
```
emersion: This can result in conflicts: OP IDs might collide with trigrams and UICs.

Moreover, the input list may contain multiple path steps with the same trigram/UIC, but with a different `secondary_code`.

front/src/modules/pathfinding/hooks/usePathfinding.ts
Synar: I see a useCallback, not a useEffect, so it should be fine, beyond redefining the function often. Unless you're talking about a different useEffect deeper in the code.

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -66,6 +68,58 @@ const usePathfinding = ({
...
+        return {
+          ...step,
+          name,
+        };
+      });
+      dispatch(updatePathSteps(updatedSteps));
+    } catch (error) {
+      console.error('Error fetching operational points:', error);
+    }
+  }, [infraId, pathSteps]);
```
Synar: The caller of this function takes pathSteps as an argument, not a dependency. I believe we should either do the same (using the functional form of set state if we really want to be cautious), or update the dependency array of the caller.

front/src/modules/pathfinding/hooks/usePathfinding.ts
Synar: I agree that having 3 dicts/a struct of dicts would be cleaner to avoid collisions, though I'm unsure they're possible in practice.

For the same primary identifier but different secondary code case, won't these steps share the same name though?

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -66,6 +68,58 @@ const usePathfinding = ({
...
+      const results = (await postSearch({
+        searchPayload,
+        pageSize: 101,
+      }).unwrap()) as SearchResultItemOperationalPoint[];
+
+      if (!results.length) return;
+
+      const opMap = results.reduce(
+        (acc, op) => {
+          acc[`87${op.ci}`] = op.name;
```
Synar: Feel free to ignore, but I'm curious, is the 87 in front of a ci to give an uic a sncf thing? Is it something general?

front/src/modules/pathfinding/hooks/usePathfinding.ts
emersion: I thought the callback was invoked from a `useEffect()`, but it seems like that's not the case.

front/src/modules/pathfinding/hooks/usePathfinding.ts
emersion: They're maybe not happening with the France infrastructure, but may happen if another country decides to use numerical IDs for OP IDs for instance.

> For the same primary identifier but different secondary code case, won't these steps share the same name though?

I don't believe that's guaranteed by the data model. Each OP object is independent in the database and has its own name.

emersion: (I don't think we need any kind of map/record here. We can just iterate over all results instead. The number of results should be low™: it cannot exceed the number of path steps.)

front/src/modules/pathfinding/hooks/usePathfinding.ts
clarani: Yes it's a sncf thing :/ the pr's uic is 87 + the pr's ci in France (87 is an id for SNCF, see https://fr.wikipedia.org/wiki/Liste_des_codes_pays_UIC)

Synar: Thanks! So ideally we would need to know the op's country in order to correctly match with a path step, but that seems painful.

Another stupid question way beyond this pr, but why do we use a ci to identify each op and not directly an uic in the first place?

front/src/modules/pathfinding/hooks/usePathfinding.ts
RomainValls: Should I change anything or is keeping the useCallback ok for you ?

emersion: I'm still not a fan of how we overwrite `pathSteps` in a lot of places in osrd, but fixing that would require a much larger refactoring and is out of scope for your PR. Let's track this separately in https://github.com/OpenRailAssociation/osrd/issues/11217.

front/src/modules/operationalPoint/helpers/buildOpSearchQuery.ts
```diff
@@ -1,17 +1,16 @@
...
-const buildOpSearchQuery = (
-  infraId: number,
-  trainSchedules: TrainScheduleResult[]
-): SearchPayload | null => {
-  const pathItems = trainSchedules.map((schedule) => schedule.path).flat();
+const buildOpSearchQuery = (infraId: number, pathSteps: PathStep[]): SearchPayload | null => {
   const pathItemQueries = [];
   const pathItemSet = new Set<string>();
 
+  const pathItems = pathSteps.map((step) => step).flat();
```
emersion: This `.map()` seems to be a no-op: it creates a new array identical to the previous one. I don't think we need the `flat()` anymore.

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -205,6 +269,7 @@ const usePathfinding = ({
 
       if (!pathfindingInput) {
         setIsMissingParam();
+        await fetchOperationalPoints(pathSteps.filter((step): step is PathStep => step !== null));
```
emersion: Why do we need the `step is PathStep`?

I think this is dangerous in case the type of `pathSteps` is changed to something else.

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -66,6 +67,69 @@ const usePathfinding = ({
...
+          }
+
+          return {
+            ...step,
+            name: matchedOp?.name,
+          };
+        });
+
+        dispatch(updatePathSteps(updatedSteps));
+      } catch (error) {
```
emersion, line 85: I think we only need to wrap the `postSearch()` in the `try`/`catch` block:

```ts
let results: SearchResultItemOperationalPoint[];
try {
  results = (await postSearch({
    searchPayload,
    pageSize: 101,
  }).unwrap()) as SearchResultItemOperationalPoint[];
} catch (error) {
  console.error('Error fetching operational points:', error);
  return;
}

// rest of the code
```

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -66,6 +67,69 @@ const usePathfinding = ({
...
 
+  const [postSearch] = osrdEditoastApi.endpoints.postSearch.useMutation();
+
+  const fetchOperationalPoints = useCallback(
+    async (steps: PathStep[]) => {
+      if (!infraId || !steps.length) return;
+
+      const searchPayload = buildOpSearchQuery(
+        infraId,
+        steps.filter((step) => step !== null)
```
emersion: This filter shouldn't be required since the type of `steps` is `PathStep[]`.

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -66,6 +67,69 @@ const usePathfinding = ({
...
+      try {
+        const results = (await postSearch({
+          searchPayload,
+          pageSize: 101,
+        }).unwrap()) as SearchResultItemOperationalPoint[];
+
+        if (!results.length) return;
+
+        const updatedSteps = steps.map((step) => {
+          if (!step) return step;
```
emersion: Same here: `step` cannot be null here.

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -66,6 +67,69 @@ const usePathfinding = ({
...
+
+          let matchedOp: SearchResultItemOperationalPoint | undefined;
+
+          const stepUICString = 'uic' in step && step.uic ? String(step.uic) : null;
+
+          if (stepUICString) {
+            matchedOp = results.find(
+              (op) =>
+                `87${op.ci}` === stepUICString &&
```
emersion: I'd prefer to avoid hardcoding French-specific assumptions if we can. We should be able to use `op.uic` here instead.

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -66,6 +67,69 @@ const usePathfinding = ({
...
+                'secondary_code' in step &&
```
emersion: I think `'secondary_code' in step` is always true when `'uic' in step`?

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -66,6 +67,69 @@ const usePathfinding = ({
...
+                (!step.secondary_code || op.ch === step.secondary_code)
```
emersion: So, in case the step is missing a secondary code, things are a bit tricky… Let's say for instance the path steps are:

```ts
[
  { uic: 87042, secondary_code: "TR" },
  { uic: 87042, secondary_code: null }
]
```

and then the backend returns this:

```ts
[
  { uic: 87042, ch: "TR", name: "Romain Triage" },
  { uic: 87042, ch: "BV", name: "Romain-Ville" }
]
```

Then we'd pick "TR" for both path steps, even if the pathfinding algorithm may pick "TR" for the first one and "BV" for the second one.

(Note, maybe instead of BV it could be something else entirely… technically when a path step doesn't specify a `secondary_code` it matches any `ch`.)

I'd be fine leaving this as a TODO for later. Ideally we'd get a new API endpoint to get OPs or OP IDs from path steps and all of these issues would be resolved.

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -66,6 +67,69 @@ const usePathfinding = ({
...
+            );
+          }
+
+          if (!matchedOp && 'trigram' in step && step.trigram) {
```
emersion: If we use `else if`, we don't need to repeat `!matchedOp` each time.

front/src/modules/pathfinding/hooks/usePathfinding.ts
emersion: Not a stupid question at all! I've come to the same conclusion as you it seems: https://github.com/OpenRailAssociation/osrd/pull/10949#discussion_r2007387063

front/src/modules/pathfinding/hooks/usePathfinding.ts
RomainValls: Good point !
I agree on leaving this as a TODO, I see you created a ticket mentioning this one , thanks 👍

emersion: https://github.com/OpenRailAssociation/osrd/issues/11217 is unrelated to the problem described in this comment. I'd suggest leaving a TODO comment.

front/src/modules/operationalPoint/helpers/buildOpSearchQuery.ts
emersion: It seems like the `map()` call is still there? Does removing it cause issues?

front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -66,6 +67,52 @@ const usePathfinding = ({
...
+
+      if (!searchPayload) return;
+      let results: SearchResultItemOperationalPoint[];
+      try {
+        results = (await postSearch({
+          searchPayload,
+          pageSize: 101,
+        }).unwrap()) as SearchResultItemOperationalPoint[];
+      } catch (error) {
+        console.error('Error fetching operational points:', error);
```
emersion, line 86: We should `return` after printing the error, or else we `results` is left undefined and will cause a runtime error below when calling `results.find()`.

I'm actually surprised TypeScript doesn't complain about this.

emersion (commented):
Overall looks like this is almost good to go, thanks for bearing with me :)

front/src/modules/operationalPoint/helpers/buildOpSearchQuery.ts
RomainValls: My bad ! Removed it :)

emersion (approved):
LGTM, thanks!

PR #10955, opened 2025-02-25
front: handle and display non op stops in stdcm simulation report 

### LINKED ISSUE #10723: front: we don't report stops when they're not on OP
### What happened?

When a stop isn't located precisely at an operational point, we don't report it in the output sheet.

It never *usually* happen (for STDCM), until https://github.com/OpenRailAssociation/osrd/pull/10200 happens.

_No response_

Easiest way is probably to checkout on https://github.com/OpenRailAssociation/osrd/pull/10200

Local

Firefox

14fa21bf67cf757bd42881dbf5459adab5281a8c

Synar:
Close [#10723](https://github.com/OpenRailAssociation/osrd/issues/10723)

### DIFF
--- a/front/src/applications/stdcm/types.ts
+++ b/front/src/applications/stdcm/types.ts
@@ -68,8 +68,7 @@ export type StdcmResultsOperationalPoint = {
   time: string | null;
   name?: string;
   ch?: string;
-  stop?: string | null;
-  duration: number;
+  duration: number | null;
   stopEndTime: string;
   trackName?: string;
   stopType?: StdcmStopTypes;
--- a/front/src/applications/stdcm/types.ts
+++ b/front/src/applications/stdcm/types.ts
@@ -73,6 +73,7 @@ export type StdcmResultsOperationalPoint = {
   stopEndTime: string;
   trackName?: string;
   stopType?: StdcmStopTypes;
+  stopFor?: number;
 };
 
 export type ConsistErrors = {
--- a/front/src/applications/stdcm/types.ts
+++ b/front/src/applications/stdcm/types.ts
@@ -63,7 +63,7 @@ export type SimulationReportSheetProps = {
 };
 
 export type StdcmResultsOperationalPoint = {
-  opId: string;
+  opId?: string;
   positionOnPath: number;
   time: string | null;
   name?: string;
--- a/front/src/applications/stdcm/types.ts
+++ b/front/src/applications/stdcm/types.ts
@@ -73,7 +73,7 @@ export type StdcmResultsOperationalPoint = {
   stopEndTime: string;
   trackName?: string;
   stopType?: StdcmStopTypes;
-  stopFor?: number;
+  stopRequested: boolean;
 };
 
 export type ConsistErrors = {
--- a/front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
+++ b/front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
@@ -111,7 +111,7 @@ function formatOperationalPointWithTimes(
   const stopBegin = getTimeAtPosition(op.positionOnPath, train);
 
   const duration = getStopDurationAtPosition(op.positionOnPath, train.positions, train.times);
-  const durationInSeconds = duration !== null ? duration.total('second') : 0;
+  const durationInSeconds = duration !== null ? duration.total('second') : null;
   const stopEnd = stopBegin.add(duration || Duration.zero);
   // Find the corresponding stopType from pathSteps
   const correspondingStep = simulationPathSteps.find(
--- a/front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
+++ b/front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
@@ -100,19 +100,40 @@ export function getStopDurationAtPosition(
 /**
  * @param op Operational point to format
  * @param train Object containing simulated train positions, times, and departure time
- * @param simulationPathSteps List of simulation path steps
- * @returns A formatted operational point with calculated stop duration and departure time
+ * @returns A formatted operational point with required fields only, including calculated stop duration and departure time
  */
-function formatOperationalPointWithTimes(
-  op: SuggestedOP,
-  train: TrainSimulation,
-  simulationPathSteps: StdcmPathStep[]
+function formatMinimalOperationalPointWithTimes(
+  op: Pick<SuggestedOP, 'positionOnPath' | 'opId'>,
+  train: TrainSimulation
 ): StdcmResultsOperationalPoint {
   const stopBegin = getTimeAtPosition(op.positionOnPath, train);
 
   const duration = getStopDurationAtPosition(op.positionOnPath, train.positions, train.times);
   const durationInSeconds = duration !== null ? duration.total('second') : null;
   const stopEnd = stopBegin.add(duration || Duration.zero);
+
+  return {
+    opId: op.opId,
+    positionOnPath: op.positionOnPath,
+    time: durationToHHMM(stopBegin),
+    duration: durationInSeconds,
+    stopEndTime: durationToHHMM(stopEnd),
+    stopRequested: false,
+  };
+}
+
+/**
+ * @param op Operational point to format
+ * @param train Object containing simulated train positions, times, and departure time
+ * @param simulationPathSteps List of simulation path steps
+ * @returns A fully formatted operational point with calculated stop duration and departure time
+ */
+function formatOperationalPointWithTimes(
+  op: SuggestedOP,
+  train: TrainSimulation,
+  simulationPathSteps: StdcmPathStep[]
+): StdcmResultsOperationalPoint {
+  const partiallyFormattedOp = formatMinimalOperationalPointWithTimes(op, train);
   // Find the corresponding stopType from pathSteps
   const correspondingStep = simulationPathSteps.find(
     (step) => step.location && matchPathStepAndOp(step.location, op)
@@ -127,13 +148,9 @@ function formatOperationalPointWithTimes(
     correspondingStep.stopFor !== undefined;
 
   return {
-    opId: op.opId!,
-    positionOnPath: op.positionOnPath,
-    time: durationToHHMM(stopBegin),
+    ...partiallyFormattedOp,
     name: op.name,
     ch: op.ch,
-    duration: durationInSeconds,
-    stopEndTime: durationToHHMM(stopEnd),
     trackName: op.metadata?.trackName,
     stopType,
     stopRequested,
@@ -163,8 +180,7 @@ export function findAllStops(positions: number[]): number[] {
 export function insertMissingStopsInOperationalPointsWithTimes(
   formatedOps: StdcmResultsOperationalPoint[],
   stopPositions: number[],
-  train: TrainSimulation,
-  simulationPathSteps: StdcmPathStep[]
+  train: TrainSimulation
 ): StdcmResultsOperationalPoint[] {
   const formatedOpsWithAllStops: StdcmResultsOperationalPoint[] = [];
   let opIndex = 0;
@@ -182,14 +198,12 @@ export function insertMissingStopsInOperationalPointsWithTimes(
 
     // At least the departure with pos 0 should have been added, so updatedOperationalPointsWT.length > 1
     const lastAddedOp = formatedOpsWithAllStops.at(-1)!;
-    const formattedStop = formatOperationalPointWithTimes(
+    const formattedStop = formatMinimalOperationalPointWithTimes(
       {
         positionOnPath: stopPosition,
-        offsetOnTrack: NaN,
-        track: '',
+        opId: `unplanned_stop_at_${stopPosition}`,
       },
-      train,
-      simulationPathSteps
+      train
     );
     if (lastAddedOp.stopRequested && !lastAddedOp.duration) {
       // If a stop was requested at the last op and no stop was performed,
@@ -255,7 +269,7 @@ export function getOperationalPointsWithTimes(
   simulationPathSteps: StdcmPathStep[],
   departureTime: Date
 ): StdcmResultsOperationalPoint[] {
-  const { positions, times } = simulation.final_output;
+  const { positions, times, speeds } = simulation.final_output;
 
   const departureHour = departureTime.getHours();
   const departureMinute = departureTime.getMinutes();
@@ -269,12 +283,11 @@ export function getOperationalPointsWithTimes(
     )
   );
 
-  const stopPositions = findAllStops(simulation.final_output.positions);
+  const stopPositions = findAllStops(positions);
   const formattedOpsWithAllStops = insertMissingStopsInOperationalPointsWithTimes(
     formattedOps,
     stopPositions,
-    { positions, times, departureHour, departureMinute },
-    simulationPathSteps
+    { positions, times, departureHour, departureMinute }
   );
   return consolidateOvertakesToSingleSteps(formattedOpsWithAllStops);
 }
--- a/front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
+++ b/front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
@@ -1,4 +1,4 @@
-import type { ReportTrain, SimulationResponse } from 'common/api/osrdEditoastApi';
+import type { SimulationResponse } from 'common/api/osrdEditoastApi';
 import { matchPathStepAndOp } from 'modules/pathfinding/utils';
 import { interpolateValue } from 'modules/simulationResult/SimulationResultExport/utils';
 import type { SuggestedOP } from 'modules/trainschedule/components/ManageTrainSchedule/types';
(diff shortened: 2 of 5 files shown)


louisgreiner (approved):
Tested using this soon-to-be feature : https://github.com/OpenRailAssociation/osrd/pull/10200 (on this branch : `ali-lgr/fix-train-switch-overflow-w-front`)

It works very well 
Capture d'écran 2025-03-04 112319
Capture d'écran 2025-03-04 112512

Nice documentation as well, which made it easier to dive in the code ^^

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
```diff
@@ -67,33 +79,45 @@ export function addMinutesToTime(
...
-  trainDepartureHour: number,
-  trainDepartureMinute: number
-): string {
+/**
+ * @property positions List of positions of a train in mm
+ * @property times List of times in milliseconds corresponding to the train positions
+ * @property departureHour Hour of the train departure (24h format)
+ * @property departureMinute Minute of the train departure
+ */
+interface TrainSimulationProps {
```
kmer2016: The name `TrainSimulationProps` is misleading as it implies these are React component props, which they are not. What do you think about renaming it to `TrainSimulation` ? Additionally, please use `type` instead of `interface` in according to our guidelines.
```suggestion
type TrainSimulation = {
```

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
```diff
@@ -67,33 +79,45 @@ export function addMinutesToTime(
...
- * @param position format: Distance from the beginning of the path in mm
- * @param positionsList format: List of positions of a train in mm.
- * @param timesList format: List of times in milliseconds corresponding to the positions in trainPositions.
+ * @param position Distance from the beginning of the path in mm
+ * @param positionsList List of positions of a train in mm.
+ * @param timesList List of times in milliseconds corresponding to the positions in trainPositions.
  * @returns The duration in milliseconds between the first and last occurrence of the position in the trainPositions array
  */
-export function getStopDurationBetweenTwoPositions(
+export function getStopDurationAtPosition(
```
kmer2016, line 87: I think we could make the function `getStopDurationAtPosition` more intuitive by returning `0` instead of `null` when there's no stop.

Benefits would be:
- It's more semantically correct - no stop means `0` duration
- Simplifies the return type to just `number` 
- Makes the code using this function cleaner by avoiding `null` checks

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
```diff
@@ -106,6 +130,121 @@ export function getStopDurationBetweenTwoPositions(
...
+  const durationInSeconds = duration !== null ? duration / 1000 : 0;
+  const durationToString = secondsToTimeString(durationInSeconds);
+  const stopEndTime = computeStopDepartureTime(formattedTime, durationToString);
+  // Find the corresponding stopType from pathSteps
+  const correspondingStep = simulationPathSteps.find(
+    (step) => step.location && matchPathStepAndOp(step.location, op)
+  );
+  let stopType;
+  if (correspondingStep) {
+    stopType = !correspondingStep.isVia ? 'serviceStop' : correspondingStep.stopType;
```
kmer2016: The ternary could be more readable if we reverse the condition. Also, let's take advantage of TypeScript's type safety by using the `StdcmStopTypes` enum instead of string literals:
```suggestion
    stopType = correspondingStep.isVia ? correspondingStep.stopType : StdcmStopTypes.SERVICE_STOP;
```

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
```diff
@@ -106,6 +130,121 @@ export function getStopDurationBetweenTwoPositions(
...
+}
+
+/**
+ * @param finalOutput Final simulation report containing lists of positions and times of all simulated points
+ * @returns A list of all positions at which the trains stops
+ */
+function findAllStops(finalOutput: ReportTrain): number[] {
+  return finalOutput.positions.filter(
+    (position, index) =>
+      index !== finalOutput.positions.length &&
```
kmer2016: The array length check needs to be fixed to properly handle the last position:
```suggestion
      index !== finalOutput.positions.length -1 &&
```

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
```diff
@@ -106,6 +130,121 @@ export function getStopDurationBetweenTwoPositions(
...
+}
+
+/**
+ * @param formatedOps List of operational points with times
+ * @param stopPositions List of all detected stop positions
+ * @param train Object containing simulated train positions, times, and departure time
+ * @param simulationPathSteps List of simulation path steps
+ * @returns A list of operational points including detected missing stops
+ */
+function insertsMissingStopsInOperationalPointsWithTimes(
```
kmer2016: I noticed this function handles some interesting business cases about merging detected stops with operational points. Some tests would help understand and document how it works!

I see these test cases : 
- It should insert a detected stop between two operational points
- It should update duration of previous op when needed
- It should manages the creation of new OP for unplanned stops

What do you think ?

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
```diff
@@ -140,6 +279,13 @@ export function consolidateOvertakesToSingleSteps(
...
 }
 
+/**
+ * @param operationalPoints List of operational points to be formated and enriched
+ * @param simulation Simulation response containing final output positions and times
+ * @param simulationPathSteps List of simulation path steps
+ * @param departureTime Departure time in hh:mm format
+ * @returns A list of formated operational points with times and stop durations
+ */
 export function getOperationalPointsWithTimes(
```
kmer2016, line 267: Really nice work on making the code more readable!

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
Synar: Sure good idea! ^^

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
Synar: Good catch my bad!

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
Synar: Sure ^^

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
Synar: Sure! I'll doublecheck since I think the function is used in one other place, but I agree this would be a good change, thanks ^^

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
Synar: Sure! I was not satisfied with the name either, but I feel TrainSimulation implies this contains more than it does. I've considered TrainSimulationPartial, TrainSimulationSimplified, TrainSimulationSpacetime, TrainTimesPositions, TrainKinematics, and others, but I'm not sure I'm convinced by any. Perhaps you're right and simply TrainSimulation works best.

Btw, why do we prefer type over interface? And are these guidelines written somewhere? If so I've been making prs blind this whole time haha

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
Synar: Thanks ^^

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
emersion: Unfortunately I think a zero duration is semantically different from null here. null means that the train doesn't stop, zero means that the train stops and then immediately restarts.

front/src/applications/stdcm/types.ts
```diff
@@ -72,6 +72,7 @@ export type StdcmResultsOperationalPoint = {
   stopEndTime: string;
   trackName?: string;
   stopType?: string;
+  stopFor?: number;
```
emersion: Where was this set and used?

It seems like this addition is unrelated to `getStopDurationAtPosition()`, am I missing something?

emersion: Seems like `getOperationalPointsWithTimes()` is setting this. Do you know if it's used anywhere? If it's missing from the type, then it shouldn't be usable anywhere without bypassing the type system. In that case, maybe we can just drop it?

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
```diff
@@ -106,6 +106,123 @@ export function getStopDurationAtPosition(
...
+      opIndex += 1;
+    }
+
+    // If there is already an operational point at the stop position, skip
+    if (opIndex < formatedOps.length && formatedOps[opIndex].positionOnPath === stopPosition)
+      return;
+
+    // At least the departure with pos 0 should have been added, so updatedOperationalPointsWT.length > 1
+    const lastAddedOpIndex = formatedOpsWithAllStops.length - 1;
+    const lastAddedOp = formatedOpsWithAllStops[lastAddedOpIndex];
```
emersion: Nit: could be simplified to `.at(-1)`

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
```diff
@@ -106,6 +106,123 @@ export function getStopDurationAtPosition(
...
+    if (
+      lastAddedOp.stopFor &&
+      !lastAddedOp.duration &&
+      formattedStop.duration &&
+      !formattedStop.stopFor
+    ) {
+      // If a stop was requested at the last op and no stop was performed,
+      // we assume the current stop actually corresponds to the last op
+      formatedOpsWithAllStops[lastAddedOpIndex].duration = formattedStop.duration;
+      formatedOpsWithAllStops[lastAddedOpIndex].stopEndTime = formattedStop.stopEndTime;
```
emersion: Nit: could be simplified by mutating `lastAddedOp` instead of accessing the array again.

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
```diff
@@ -106,6 +106,123 @@ export function getStopDurationAtPosition(
...
+function findAllStops(finalOutput: ReportTrain): number[] {
+  return finalOutput.positions.filter(
+    (position, index) =>
+      index !== finalOutput.positions.length &&
+      position === finalOutput.positions[index + 1] &&
+      (!index || position !== finalOutput.positions[index - 1])
+  );
+}
+
+function insertsMissingStopsInOperationalPointsWithTimes(
```
emersion: Nit: should maybe be named "insert" instead of "inserts"?

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
Synar: Sure!

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
Synar: TIL, thanks! I didn't know ts supported negative indexing.

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
Synar: Sure!

front/src/applications/stdcm/types.ts
Synar: Hmm, since before my pr getOperationalPointsWithTimes set it and SimulationReportSheet used a .stopFor, I assumed it was where it was used. But in fact SimulationReportSheet simply reads the path step again.

On the other hand I ended up using it in my code later to know if a stop was requested. I don't really need the actual stopFor field though, a boolean "stopRequested" field would make more sense.

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
Synar: In theory I would agree, but the only caller of this function actually immediatly converts null back to 0 (which is why I wanted to doublecheck if there was another caller). Zero seconds stops are then treated exctly the same as the absence of stops. 

We may want to maintain the difference between null and 0 just in case we want to use this function somewhere else, but considering where it is in the code I doubt this would happen.

On the other hand, we could also modify the behavior of the output table/sheet, and actually display 0 second stops as stops, which would justify keeping a difference between 0 and null here. This would make sense to me as a change, but I believe it would be a behavior modification that would require PO approval.

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
kmer2016: I think `TrainTimesPositions` is a good name, it's descriptive enough.

Yes you can find the guidelines [here](https://github.com/osrd-project/osrd-confidential/issues/623).

I think we decided to use `type` over `interface` because `interface` is best for defining contracts (like module APIs, public methods, class shapes) and fits better in OOP paradigm where we define behaviors. while `type` is better for simple data structures

Synar: Thanks a lot! There were definitely some conventions I was not aware of.

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
kmer2016: I initially suggested returning 0 based on current usage,but @emersion makes a good point about semantic difference.

@Synar, feel free to choose either approach - returning 0 directly would make the code cleaner, go for it! Otherwise, keeping null for semantic clarity is fine too.

Not blocking for me.

Synar:
Sorry I was busy with other things, but I finally got around to making the requested changes. 

The change to make 0 seconds stops displayed as stops was PO approved ^^

I haven't tested the pr again yet though.

Synar:
Up

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
```diff
@@ -106,6 +130,123 @@ export function getStopDurationBetweenTwoPositions(
...
+}
+
+/**
+ * @param positions Lists of all positions of simulated points of a train simulation report
+ * @returns A list of all positions at which the train stops
+ */
+export function findAllStops(positions: number[]): number[] {
+  return positions.filter(
+    (position, index) =>
+      index !== positions.length - 1 &&
```
emersion: Isn't the last position also a stop?

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
```diff
@@ -106,6 +130,123 @@ export function getStopDurationBetweenTwoPositions(
...
+      position === positions[index + 1] &&
```
emersion: Is this really a reliable way to find out whether a train stops somewhere? Won't this break with zero duration stops?

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
```diff
@@ -106,6 +130,123 @@ export function getStopDurationBetweenTwoPositions(
...
+      return;
+
+    // At least the departure with pos 0 should have been added, so updatedOperationalPointsWT.length > 1
+    const lastAddedOp = formatedOpsWithAllStops.at(-1)!;
+    const formattedStop = formatOperationalPointWithTimes(
+      {
+        positionOnPath: stopPosition,
+        offsetOnTrack: NaN,
+        track: '',
+      },
```
emersion: Hm, could we avoid passing an invalid type to `formatOperationalPointWithTimes()`?

Not only this is surprising to deal with in `formatOperationalPointWithTimes()` (could make it optional to make this more obvious?), but also this results in an object with a missing `opId` field even though the type `StdcmResultsOperationalPoint` marks it as mandatory.

emersion (commented):
Still not sure I understand this whole workaround (not a huge fan of the whole concept I must admit)… Here are a few comments regardless.

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
Synar: Added a speed check too ^^

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
Synar: Sure!

front/src/applications/stdcm/utils/formatSimulationReportSheet.ts
Synar: I've split the logic of formatOperationalPointWithTimes in two, which should make things cleaner. Not sure about the name of the first part though.

Also opId is never actually used and optional in the incoming type, so I've made it optional. Still, I've provided an id for the new steps, in case we ever want to use the id field for debugging purpose.

Synar:
Everything should be resolved, and I have a clean, up to date branch with both PRs at ali-lgr/test-stdcm-stop-next-signal.

~~@louisgreiner It does not seem the train actually stops at the next signal anymore though (I remember it working previously) no matter how many flags I flip~~

Edit: Seen with louis, everything now working on the fused branch as expected ^^

kmer2016 (approved):
LGTM

PR #10961, opened 2025-02-26
editoast: doc: advertize migration with locked schema

bougue-pe:
To be used as much as possible to avoid triggering rebuild (triggered by file's modification-date).
Should improve dev-ex.


flomonster (approved):
Nice catch

woshilapin (approved):
Seems a good idea. Thank you.

younesschrifi (approved):
LGTM

PR #10962, opened 2025-02-26
front: reorder and redefine map layer groups

### LINKED ISSUE #8302: Infra editor : Speed restriction edition switches not visible
### What happened?

When I create or edit a speed restriction and I want to select a switch, I don't see the switches on the map (even if the layer is activated) because the switches are hidden by speed limits
image
Not major because I can still select it, but it is confusing : 
image

I expected to see the switches above the speed sections

Edit a speed restriction in the infra editor

Recette (SNCF)

Google Chrome

Synar:
Close #8302


front/src/modules/simulationResult/components/SimulationResultsMap/VirtualLayers.tsx
```diff
@@ -1,7 +1,7 @@
 import { range } from 'lodash';
 import { Layer } from 'react-map-gl/maplibre';
 
-export const VIRTUAL_LAYERS = 8;
+export const VIRTUAL_LAYERS = 12;
```
jacomyal: This looks like a magic number (and already felt like this before). It feels tricky to guess that this should be increased, for a new developer who would update the layers order again later.

Do you think we could use something like `const VIRTUAL_LAYERS = Object.keys(LAYER_GROUPS_ORDER).length;` instead?

jacomyal (changes requested):

front/src/modules/simulationResult/components/SimulationResultsMap/VirtualLayers.tsx
Synar: Good idea thanks! As an example myself of an aforementioned new developer on this component, it was indeed tricky to guess. Changed ^^

jacomyal (approved):

jacomyal:
LGTM :+1:

clarani (commented):
Nice refacto :pray: 

However, I'm not sure it solves the bug :/
I can still have the speed limits over the punctual elements
[Capture vidéo du 2025-02-27 11-47-49.webm](https://github.com/user-attachments/assets/51145bbb-ca43-404d-96e2-dba41a53c811)

Synar:
Sorry, my bad, I was so convinced the above suggestion would work right out of the gate I did not test it. Turns out symbols are not keys, and so Object.keys does not return them. So the map had 0 virtual layers, and so ordering was completely broken. Replacing Object.keys with Object.getOwnPropertySymbols works as intended. '^^

Can you try again @clarani ?

clarani (approved):
LGTM :white_check_mark:  
It works very well, thanks !!

PR #10963, opened 2025-02-26
front: fix results display when entering a scenario

### LINKED ISSUE #10946: No simulation results when entering a scenario
### What happened?

Sometimes when entering a scenario for the first time, we can see only the space time chart but not the rest of the simulation results.

Image

If we go back to the scenario list and back to the same scenario, the results are properly displayed. Same if switch selection to another train.

To see the simulation results all the time

1. Go to a scenario with some trains
2. If it doesn't work, go to the scenarios list, refresh the page and enter the scenario again

Local

Firefox

.

theocrsb:
closes #10946

Some `getPathProperties` conditions were not in the useEffect depths, so changing these values did not update `trainSimulation` (which is undefined by default when entering a scenario), preventing the simulation results to be displayed

### DIFF
--- a/front/src/modules/simulationResult/components/SpeedSpaceChart/useSpeedSpaceChart.ts
+++ b/front/src/modules/simulationResult/components/SpeedSpaceChart/useSpeedSpaceChart.ts
@@ -76,7 +76,14 @@ const useSpeedSpaceChart = (
     };
 
     getPathProperties();
-  }, [pathProperties, infraId, rollingStock]);
+  }, [
+    infraId,
+    trainScheduleResult,
+    rollingStock,
+    pathfindingResult,
+    simulation?.status,
+    pathProperties,
+  ]);
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


front/src/modules/simulationResult/components/SpeedSpaceChart/useSpeedSpaceChart.ts
Akctarus, line 1: Are you sure trainScheduleResult and pathfindingResult are mandatory ?

Akctarus (approved):
Tested, LGTM but left a comment

front/src/modules/simulationResult/components/SpeedSpaceChart/useSpeedSpaceChart.ts
theocrsb: Since they are used in condition, I think it's a good idea to keep them in dependency

SharglutDev (approved):
Lgtm and tested, good job ! Could you put a description in the commit to explain the bug ?

theocrsb:
I just added the description to the commit :)

PR #10965, opened 2025-02-26
front: fix origin and destination are not displayed by default

### LINKED ISSUE #10883: front: manchette origin and destination aren't displayed by default when importing train
### What happened?

I imported 

[2-train_schedules with-speed-limits.json](https://github.com/user-attachments/files/18888273/2-train_schedules.with-speed-limits.json)

 in my scenario and got this manchette.

Image

The origin and destination where not selected by default so only part of my train was displayed.

Image

I'm expecting origin and destination to be displayed by default

Reproduce description. 

Recette (SNCF)

Google Chrome

776f9b3

theocrsb:
closes #10883

### DIFF
--- a/front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
+++ b/front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
@@ -90,9 +90,15 @@ const useGetProjectedTrainOperationalPoints = ({
         } else {
           // If the manchette hasn't been saved, we want to display by default only
           // the waypoints with CH BV/00/'' and the path steps (origin, destination, vias)
-          operationalPointsWithUniqueIds = operationalPointsWithUniqueIds.filter((op) =>
-            op.extensions?.sncf ? isStation(op.extensions.sncf.ch) || op.weight === 100 : true
-          );
+
+          const lastIndex = operationalPointsWithUniqueIds.length - 1;
+          operationalPointsWithUniqueIds = operationalPointsWithUniqueIds.filter((op, i) => {
+            if (i === 0 || i === lastIndex) return true;
+            // handle waypoints added from the map
+            if (!op.extensions?.sncf) return true;
+            // handle waypoints added from the pathfinding or operational points on path
+            return isStation(op.extensions.sncf.ch) || op.weight === 100;
+          });
         }
 
         setFilteredOperationalPoints(operationalPointsWithUniqueIds);


Caracol3 (commented):
Good job. Tested and works fine

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
```diff
@@ -88,8 +88,11 @@ const useGetProjectedTrainOperationalPoints = (
...
           // If the manchette hasn't been saved, we want to display by default only
           // the waypoints with CH BV/00/'' and the path steps (origin, destination, vias)
-          operationalPointsWithUniqueIds = operationalPointsWithUniqueIds.filter((op) =>
-            op.extensions?.sncf ? isStation(op.extensions.sncf.ch) || op.weight === 100 : true
+          operationalPointsWithUniqueIds = operationalPointsWithUniqueIds.filter(
+            (op, i) =>
+              i === 0 ||
+              i === operationalPointsWithUniqueIds.length - 1 ||
+              (op.extensions?.sncf ? isStation(op.extensions.sncf.ch) || op.weight === 100 : true)
           );
```
clarani: Hum it's a bit hard to read. Can you write a more explicit function like this ?
```suggestion
           operationalPointsWithUniqueIds = operationalPointsWithUniqueIds.filter((op, i) => {
            if (i === 0 || i === operationalPointsWithUniqueIds.length - 1) return true;
            // handle waypoints added from the map
            if (!op.extensions?.sncf) return true;
            // handle waypoints added from the pathfinding or operational points on path
            return isStation(op.extensions.sncf.ch) || op.weight === 100;
          });
```

theocrsb: Done :)

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
```diff
@@ -88,9 +88,13 @@ const useGetProjectedTrainOperationalPoints = (
...
-          operationalPointsWithUniqueIds = operationalPointsWithUniqueIds.filter((op) =>
-            op.extensions?.sncf ? isStation(op.extensions.sncf.ch) || op.weight === 100 : true
-          );
+          operationalPointsWithUniqueIds = operationalPointsWithUniqueIds.filter((op, i) => {
+            if (i === 0 || i === operationalPointsWithUniqueIds.length - 1) return true;
+            // handle waypoints added from the map
+            if (!op.extensions?.sncf) return true;
+            // handle waypoints added from the pathfinding or operational points on path
+            return isStation(op.extensions.sncf.ch) || op.weight === 100;
+          });
```
Caracol3:
```suggestion
          if (operationalPointsWithUniqueIds.length > 0) {
  const lastIndex = operationalPointsWithUniqueIds.length - 1;

  operationalPointsWithUniqueIds = operationalPointsWithUniqueIds.filter((op, i) => {
    if (i === 0 || i === lastIndex) return true;
    if (!op.extensions?.sncf) return true;
    return isStation(op.extensions.sncf.ch) || op.weight === 100;
  });
}
```

We could maybe use a const like this to avoid recalculation on each iteration of the filter

theocrsb: Done

Caracol3 (approved):
lgtm

clarani (approved):
LGTM :white_check_mark:

PR #10968, opened 2025-02-26
front: cut down ESLint watch list

emersion:
When ESLint starts up, it sets up a recursive file watcher on all files/directories passed as arguments, regardless of ignorePatterns. Passing "." causes performance issues because it makes ESLint look at all files.

Instead, go back to the old approach: specify a list of files on the CLI instead of ".".

Ref https://github.com/fi3ework/vite-plugin-checker/blob/ef923397d70a54de10a76882d561f816027794e4/packages/vite-plugin-checker/src/checkers/eslint/main.ts#L127


Synar (approved):
LGTM ^^

kmer2016 (approved):
LGTM

PR #10969, opened 2025-02-27
editoast: Remove unused `RabbitMQ` call and core client from `test_app`

hamz2a:


ghost (changes requested):
We have kept this method here for API completion's sake. Even if this API is not yet used, I do not see why we should delete it.

Even if we wanted this change it should be on a PR by itself not be grouped in another PR.

hamz2a:
> We have kept this method here for API completion's sake. Even if this API is not yet used, I do not see why we should delete it.

Since the method is not being used, it's a reason to remove it. However, if the Editoast maintainers agree to keep it, it's ok for me. @woshilapin @flomonster @leovalais

hamz2a:
> Even if we wanted this change it should be on a PR by itself not be grouped in another PR.

I can separate the two changes into 2 commits, if it's OK for you ?

ghost:
> Since the method is not being used, it's a reason to remove it. However, if the Editoast maintainers agree to keep it, it's ok for me. @woshilapin @flomonster @leovalais

I do not agree with this statement. The fact that there is the "allow dead code" upon the method is an indication that it should stay here. This method might be useful in the future and it is literally not doing anything : its not like its creating an attack surface or anything else.

I do not believe we should remove it on the sole case of not being used. It's there to cover for a potential use and is required to do RPC-call without expecting responses from the service.

ghost:
> I can separate the two changes into 2 commits, if it's OK for you ?

I'd rather have it in another PR, if you want to discuss the remove of this method let's open a PR for this and discuss it there: it has no place in this PR.

leovalais:
I don't mind unused definitions (as long as the annotation is justified in a comment :eyes:). API completeness is a valid reason to me, but I believe @flomonster and @woshilapin prefer avoiding dead code. I don't mind it either way, but I'd tend to agree with @ElysaSrc, **except** on:

> It's there to cover for a potential use and is required to do RPC-call without expecting responses from the service.

That use case was already covered by the client prior to the RabbitMQ support, cf. `InfraLoadRequest`:

```rust
impl AsCoreRequest<()> for InfraLoadRequest {
    const METHOD: reqwest::Method = reqwest::Method::POST;
    const URL_PATH: &'static str = "/infra_load";

    fn infra_id(&self) -> Option<i64> {
        Some(self.infra)
    }
}
```

Then is there a use-case where we'd want to provide this ⬇️ information manually and not through the usual interface?

```
routing_key: String,
path: &str,
published_payload: &T,
mandatory: bool,
correlation_id: Option<String>,
```

Khoyo:
The current call with response expects a response on the reply-to queue. A response with an empty payload, but we will still wait for it. That's obvious for an HTTP request, but not necessarily for a message queue. We *might* want to change the semantic of load to not wait for the completion of the loading before returning for example.

ghost:
> I don't mind unused definitions (as long as the annotation is justified in a comment 👀).

We can add a comment here, back when we implemented this method we indeed had the case like infra load in the head.

leovalais:
Oh right, you don't don't necessarily have to enqueue a response each time. However since core rpcs are implemented with an http server, can this case ever happen?

ghost:
> Oh right, you don't don't necessarily have to enqueue a response each time. However since core rpcs are implemented with an http server, can this case ever happen?

Isn't the core HTTP API on the path to deprecation @Khoyo ?

Khoyo:
> Isn't the core HTTP API on the path to deprecation @Khoyo ?

Yes. An even right now, editoast can perfectly choose not to provide a reply-to path (ie. use `call()`), in which case core will not reply.

Note that we don't use the HTTP server in worker mode. We do use two interface of it, but they are pretty thin.

leovalais:
Ok. In this case I'm fine with keeping `call()` for now.

However I feel like the `call` signature kinda breaks the paradigm we established with `AsCoreRequest` by explicitly requiring low-level details (osrdyne protocol values) that the client is expected to fill in. Maybe we could have the no reply behavior using `impl AsCoreRequest<NoResponse>` or change `impl AsCoreRequest<()>` to have this semantic?

But that's for later if the need arises.

Khoyo:
I agree, `call()` is at the same level than `call_with_response()`, it's not (yet?) exposed in the `AsCoreRequest` level api

woshilapin:
Are we arguing about changing a method that is `#[allow(dead_code)]`? Keeping something around

> to cover for a potential use

sounds like not so much a good reason to me. Because "potential" sometimes never happen, and you end up with dead code forever. To me, let's bring back this function from the dead **the day we need it**.

At the very least, if I'm the only one thinking it's fine to remove, but everyone else want to keep it around, let's add a comment justifying why we have `#[allow(dead_code)]`, I believe `#[allow]` should be justified.

flomonster:
I do not have an opinion on it, I think both arguments are valid.
I agree that if we chose to keep it at least we should explain why in a comment.

> [!NOTE]
> Since core is stateless I don't see a case where we will call it without expecting any response. So maybe we should remove it?

ghost:
> Since core is stateless I don't see a case where we will call it without expecting any response. So maybe we should remove it?

I have a usecase in mind (discussed with @Khoyo) : Core is not completely stateless if you take the loaded infra as a kind of state. Sending a message from editoast to preemptively ask the core to refresh its infrastructure can be a use case for such call.

woshilapin:
@hamz2a or @ElysaSrc, can you provide some comment to justify the presence of `#[allow]` so we can move on, on this topic? 🙏

ghost:
> @hamz2a or @ElysaSrc, can you provide some comment to justify the presence of `#[allow]` so we can move on, on this topic? 🙏

Already done before, but basically we should keep it because :
- its isolated, its not increasing the surface of attack, its not making the maintenance more complex since it is  small and the API is complete with it since it is part of the API we designed around RMQ.

I'm not sure how to put it in a comment for the code ?

leovalais:
@ElysaSrc 

Like that 😛 

```rust
// Left here because its isolated, its not increasing the surface of attack, its not making the maintenance more complex since it is small and the API is complete with it since it is part of the API we designed around RMQ.
#[expect(unused)]
```

hamz2a:
> ```rust
> // Left here because its isolated, its not increasing the surface of attack, its not making the maintenance more complex since it is small and the API is complete with it since it is part of the API we designed around RMQ.
> #[expect(unused)]
> ```

Do you want me to add the comment? Should I keep everything in the same PR?

woshilapin:
> Do you want me to add the comment? Should I keep everything in the same PR?

I believe there is 2 different topic in your PR. So you can either do 2 commits in the current PR, or do another PR 🤷 both are fine with me.

hamz2a:
https://github.com/OpenRailAssociation/osrd/pull/11065
https://github.com/OpenRailAssociation/osrd/pull/11066

PR #10972, opened 2025-02-27
front: handle long file names in infra editor

### LINKED ISSUE #10892: Infrastructure import : long file names not handled
### What happened?

I imported the railjson file with a long name
Image
The names goes out of the box

Use ellipsis to stay in the box

as above

Recette (SNCF)

Google Chrome

776f9b3

theocrsb:
closes #10892

### DIFF
--- a/front/src/modules/infra/styles/_infraSelector.scss
+++ b/front/src/modules/infra/styles/_infraSelector.scss
@@ -213,12 +213,17 @@
         input {
           background: none;
         }
-        span {
-          display: flex;
+        .infra-name {
+          width: 100px;
+          overflow: hidden;
+          position: relative;
+          display: inline-block;
+          text-overflow: ellipsis;
+          white-space: nowrap;
         }
         .infra-lock {
           display: flex;
-          align-items: center;
+          justify-content: end;
           color: var(--coolgray7);
           padding: 0 4px 0 4px;
           text-transform: uppercase;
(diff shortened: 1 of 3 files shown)


front/src/modules/infra/styles/_infraSelector.scss
```diff
@@ -213,15 +213,22 @@
...
         }
-        span {
+        .infra-name {
           display: flex;
+          overflow: hidden;
+          -webkit-line-clamp: 1;
+          text-overflow: ellipsis;
+          -webkit-box-orient: vertical;
+          display: -webkit-box;
+          min-width: 60%;
```
kmer2016: I have the impression that you intended to apply ellipsis to the span that displays the infrastructure names already stored in the database, correct? However, the ellipsis does not seem to work as expected because a scrollbar appears underneath. Could you please check this behavior?

image

theocrsb: You're right, it's done. Thank you

kmer2016 (changes requested):
the infra cards are not aligned anymore. This seems to be caused by the scrollbar.

image

theocrsb:
> the infra cards are not aligned anymore. This seems to be caused by the scrollbar.
> 
> image

Done

kmer2016 (approved):
LGTM !

PR #10975, opened 2025-02-27
Front: fix stdcm config crash when changing scenario in stdcm debug

### LINKED ISSUE #10973: stdcm: crash when selecting scenario in debug "STDCM conf is not valid"
### What happened?

When I try to select a scenario in STDCM debug mode, it crashes

https://github.com/user-attachments/assets/64df9cc7-3448-474c-9f35-23640bf89cbd

not crash

1. video

Recette (SNCF)

Firefox

-

clarani:
closes #10973

Thanks to the first commit, from now on, if a user selects an empty timetable in STDCM, then he will be able to request a train with a departure within the current day.

### DIFF
--- a/front/src/applications/stdcm/views/StdcmView.tsx
+++ b/front/src/applications/stdcm/views/StdcmView.tsx
@@ -6,6 +6,7 @@ import {
   type Dispatch,
   type SetStateAction,
   useMemo,
+  useCallback,
 } from 'react';
 
 import { isEqual, isNil } from 'lodash';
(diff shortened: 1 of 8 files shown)


front/src/modules/scenario/components/ScenarioExplorer/ScenarioExplorer.tsx
```diff
@@ -89,7 +89,7 @@ const ScenarioExplorer = ({
         })
       );
     }
-  }, [scenario, timetable]);
+  }, [timetable]);
```
SharglutDev: Hmm if we remove `scenario` from the depths, if we switch from a scenario to another, stdcm env won't be updated with the good values no ?

front/src/applications/stdcm/views/StdcmView.tsx
```diff
@@ -266,20 +254,41 @@ const StdcmView = () => {
...
+        <StdcmHeader
+          isDebugMode={isDebugMode}
+          onDebugModeToggle={setIsDebugMode}
+          toggleHelpModule={toggleHelpModule}
+          showHelpModule={showHelpModule}
+        />
+        <LoaderFill />
+        <StdcmHelpModule showHelpModule={showHelpModule} toggleHelpModule={toggleHelpModule} />
+      </div>
+    );
```
SharglutDev: Same `StdcmHeader` and `StdcmHelpModule` are used twice here. Could we just have a function which render either `LoaderFill`, `StdcmEmptyConfigError` or `StdcmViewContent` depending on the conditions ?

SharglutDev (approved):
Lgtm and tested, thank you !

PR #10977, opened 2025-02-27
front: create occurrence item

### LINKED ISSUE #10611: Create OccurrenceItem
### Description and goal

This task is about to create the train service item for the timetable: the component TimetableTrainServiceItem will be wrapped by the component TimetablePacedTrainItem

Image

### Acceptance criteria

Each occurrence card contains, from left to right :

  - [x] A gray dot
  - [x] The name of the occurrence
  - [x] The livery of the rollingstock
  - [x] The departure and arrival time
  - [ ] Occurrence is selectable
  - [ ] If an occurrence is selected and the user folds the mission, then the occurrence remains selected

Math-R:
close #10611

To test the PR mind to activate the show mission option in user parameters. Then go to a timetable make sur to create at least one train, and you will see a PacedTrain that you can unwrap and see the occurrences.
You can toggle the selection style of an occurence using react dev tools 

https://www.sketch.com/s/2fb064cd-c483-46ac-bd30-72705fd63e75/p/4D0E36FD-BB1D-4D86-A18A-2CAC08EBBADB/canvas#Inspect

### DIFF
--- a/front/src/modules/trainschedule/components/Timetable/PacedTrain/OccurrenceItem.tsx
+++ b/front/src/modules/trainschedule/components/Timetable/PacedTrain/OccurrenceItem.tsx
@@ -1,0 +1,77 @@
+import { Dot, Moon } from '@osrd-project/ui-icons';
+import cx from 'classnames';
+import dayjs from 'dayjs';
+
+import RollingStock2Img from 'modules/rollingStock/components/RollingStock2Img';
+
+import type { Occurrence } from '../types';
+
+const ConsecutiveDayDateDisplay = ({
+  departureTime,
+  nextDepartureTime,
+}: {
+  departureTime: Date;
+  nextDepartureTime: Date;
+}) => (
+  <div className="consecutive-day-display">
+    <div>
+      <div className="date-display before-midnight">{dayjs(departureTime).format('DD')}</div>
+      <div className="date-display after-midnight">{dayjs(nextDepartureTime).format('DD')}</div>
+    </div>
+    <div className="date-display">/{dayjs(nextDepartureTime).format('MM')}</div>
+  </div>
+);
+
+type OccurrenceItemProps = {
+  occurrence: Occurrence;
+  isSelected: boolean;
+  nextOccurrence?: Occurrence;
+};
+
+const OccurrenceItem = ({ occurrence, isSelected, nextOccurrence }: OccurrenceItemProps) => {
+  const { trainName, rollingStock, startTime, arrivalTime } = occurrence;
+  const isAfterMidnight = dayjs(occurrence.arrivalTime).isAfter(occurrence.startTime, 'day');
+  const isNextAfterMidnight = nextOccurrence
+    ? dayjs(nextOccurrence.startTime).isAfter(occurrence.startTime, 'day')
+    : false;
+
+  return (
+    <div
+      className={cx('occurrence-item', {
+        'after-midnight': isAfterMidnight,
+        'next-after-midnight': isNextAfterMidnight,
+        selected: isSelected,
+      })}
+    >
+      <div className="occurrence-item-dot">
+        <Dot variant="fill" />
+      </div>
+      <div className="occurrence-item-name" title={trainName}>
+        {trainName}
+      </div>
+      <div className="rolling-stock">
+        {rollingStock && <RollingStock2Img rollingStock={rollingStock} />}
+      </div>
+
+      <div className="occurrence-item-horaries">
+        <div className="status-icon after-midnight">
+          {isAfterMidnight && <Moon iconColor="rgba(33, 100, 130, 0.7)" />}
+        </div>
+        <div className="occurrence-item-time departure-time">
+          {dayjs(startTime).format('HH:mm')}
+        </div>
+        <div className="occurrence-item-time arrival-time">
+          {dayjs(arrivalTime).format('HH:mm')}
+        </div>
+      </div>
+      {nextOccurrence && isNextAfterMidnight && (
+        <ConsecutiveDayDateDisplay
+          departureTime={startTime}
+          nextDepartureTime={nextOccurrence?.startTime}
+        />
+      )}
+    </div>
+  );
+};
+
+export default OccurrenceItem;
--- a/front/src/modules/trainschedule/components/Timetable/PacedTrain/PacedTrainItem.tsx
+++ b/front/src/modules/trainschedule/components/Timetable/PacedTrain/PacedTrainItem.tsx
@@ -6,11 +6,13 @@ import cx from 'classnames';
 import dayjs from 'dayjs';
 import { useTranslation } from 'react-i18next';
 
-import type { PacedTrainId } from 'reducers/osrdconf/types';
+import type { PacedTrainId, TrainId } from 'reducers/osrdconf/types';
 import { ms2min } from 'utils/timeManipulation';
 
 import TimetableItemActions from '../TimetableItemActions';
 import type { PacedTrainWithResult } from '../types';
+import useOccurrences from './hooks/useOccurrences';
+import OccurrenceItem from './OccurrenceItem';
 
 type PacedTrainItemProps = {
   isInSelection: boolean;
@@ -19,6 +21,7 @@ type PacedTrainItemProps = {
   isOnEdit: boolean;
   isProjectionPathUsed: boolean;
   selectPacedTrainToEdit: (pacedTrain: PacedTrainWithResult) => void;
+  selectedTimeTableItemId: TrainId | undefined;
 };
 
 const PacedTrainItem = ({
@@ -28,19 +31,18 @@ const PacedTrainItem = ({
   isOnEdit,
   isProjectionPathUsed,
   selectPacedTrainToEdit,
+  selectedTimeTableItemId,
 }: PacedTrainItemProps) => {
   const { t } = useTranslation(['operationalStudies/scenario']);
 
   const [isOccurrencesListOpen, setIsOccurrencesListOpen] = useState(false);
+  const { occurrences, occurrencesCount } = useOccurrences(pacedTrain);
 
   const toggleOccurrencesList = () => setIsOccurrencesListOpen((open) => !open);
   const selectPathProjection = async () => {};
   const duplicatePacedTrain = async () => {};
   const deletePacedTrain = async () => {};
 
-  const pacedTrainCadence = pacedTrain.paced.step;
-
-  const occurrencesCount = Math.ceil(pacedTrain.paced.duration.ms / pacedTrain.paced.step.ms);
   return (
     <div
       data-testid="scenario-timetable-train"
@@ -92,7 +94,9 @@ const PacedTrainItem = ({
 
         {!pacedTrain.invalidReason ? (
           <div className="paced-train-right-zone">
-            {pacedTrain.isValid && <div>&mdash;&nbsp;{`${ms2min(pacedTrainCadence.ms)}min`}</div>}
+            {pacedTrain.isValid && (
+              <div>&mdash;&nbsp;{`${ms2min(pacedTrain.paced.step.ms)}min`}</div>
+            )}
             <div
               className={cx('status-icon', {
                 'not-honored-or-too-fast': pacedTrain.notHonoredReason,
@@ -120,7 +124,16 @@ const PacedTrainItem = ({
         editTimetableItem={() => selectPacedTrainToEdit(pacedTrain)}
         deleteTimetableItem={deletePacedTrain}
       />
-      <div className="occurrences" />
+      <div className="occurrences">
+        {occurrences.map((occurrence, index) => (
+          <OccurrenceItem
+            occurrence={occurrence}
+            key={occurrence.id}
+            isSelected={selectedTimeTableItemId === occurrence.id}
+            nextOccurrence={occurrences[index + 1]}
+          />
+        ))}
+      </div>
       {pacedTrain.isValid && (
         <div className="more-info">
           <div className="more-info-left">
--- a/front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
+++ b/front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
@@ -1,0 +1,60 @@
+import { useMemo } from 'react';
+
+import dayjs from 'dayjs';
+
+import type { OccurrenceId } from 'reducers/osrdconf/types';
+
+import type { Occurrence, PacedTrainWithResult } from '../../types';
+
+export const computeOccurrenceName = (pacedTrainName: string, index: number): string => {
+  const endByNumber = /\b\w+\s\d+$/;
+
+  if (endByNumber.test(pacedTrainName)) {
+    const endOfPacedTrainName = Number(pacedTrainName.split(' ').pop());
+    return `${pacedTrainName.replace(/\s\d+$/, '')} ${endOfPacedTrainName + 2 * index}`;
+  }
+  if (!Number.isNaN(+pacedTrainName)) {
+    return `${+pacedTrainName + 2 * index}`;
+  }
+  return `${pacedTrainName} ${2 * index + 1}`;
+};
+
+type OccurrencesState = {
+  occurrences: Occurrence[];
+  occurrencesCount: number;
+};
+
+const useOccurrences = ({
+  id,
+  paced,
+  startTime,
+  arrivalTime,
+  trainName,
+  rollingStock,
+}: PacedTrainWithResult) => {
+  const occurrencesState = useMemo<OccurrencesState>(() => {
+    const occurrencesCount = Math.ceil(paced.duration.ms / paced.step.ms);
+    const computedOccurrences: Occurrence[] = [];
+
+    for (let i = 0; i < occurrencesCount; i += 1) {
+      const occurrenceStartTime = dayjs(startTime)
+        .add(i * paced.step.ms, 'ms')
+        .toDate();
+      const occurrenceArrivalTime = dayjs(arrivalTime)
+        .add(i * paced.step.ms, 'ms')
+        .toDate();
+      computedOccurrences.push({
+        id: `occurrence-${i}-${id}` as OccurrenceId,
+        trainName: computeOccurrenceName(trainName, i),
+        rollingStock,
+        startTime: occurrenceStartTime,
+        arrivalTime: occurrenceArrivalTime,
+      });
+    }
+    return { occurrencesCount, occurrences: computedOccurrences };
+  }, [paced.duration, paced.step, startTime, arrivalTime, trainName, id, rollingStock]);
+
+  return occurrencesState;
+};
+
+export default useOccurrences;
--- a/front/src/modules/trainschedule/components/Timetable/types.ts
+++ b/front/src/modules/trainschedule/components/Timetable/types.ts
@@ -5,7 +5,7 @@ import type {
   SimulationSummaryResult,
   TrainScheduleResult,
 } from 'common/api/osrdEditoastApi';
-import type { PacedTrainId, TrainScheduleId } from 'reducers/osrdconf/types';
+import type { OccurrenceId, PacedTrainId, TrainScheduleId } from 'reducers/osrdconf/types';
 import type { Duration } from 'utils/duration';
 
 export type ValidityFilter = 'both' | 'valid' | 'invalid';
@@ -76,4 +76,12 @@ export type TimetableFilters = {
   setTrainTypeFilter: (trainType: TrainTypeFilter) => void;
   selectedTags: Set<string | null>;
   setSelectedTags: React.Dispatch<React.SetStateAction<Set<string | null>>>;
+};
+
+export type Occurrence = {
+  id: OccurrenceId;
+  trainName: string;
+  rollingStock?: LightRollingStockWithLiveries;
+  startTime: Date;
+  arrivalTime: Date;
 };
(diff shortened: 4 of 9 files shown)


front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
```diff
@@ -0,0 +1,62 @@
...
+        });
+      }
+      return { occurrencesCount, occurrences: computedOccurrences };
+    };
+
+    setOccurrencesState(computeOccurrences());
+  }, [pacedTrain]);
+
+  return occurrencesState;
+};
```
Yohh:
```suggestion
const useOccurrences = (pacedTrain: PacedTrainWithResult, stepDuration: Duration) => {
  const { id, paced, startTime, trainName, rollingStock } = pacedTrain;
  const [occurrencesState, setOccurrencesState] = useState<OccurrencesState>({
    occurrences: [],
    occurrencesCount: 0,
  });

  const computeOccurrenceName = (name: string, index: number): string => {
    const endByNumber = /\b\w+\s\d+$/;

    if (endByNumber.test(name)) {
      const endOfTrainName = Number(name.split(' ').pop());
      return `${name.replace(/\s\d+$/, '')} ${endOfTrainName + 2 * index}`;
    }
    if (!Number.isNaN(+name)) {
      return `${+name + 2 * index}`;
    }
    return `${name} ${2 * index + 1}`;
  };

  useEffect(() => {
    const computeOccurrences = (): OccurrencesState => {
      const occurrencesCount = Math.floor(paced.duration.ms / paced.step.ms);
      const computedOccurrences = [];
      for (let i = 0; i < occurrencesCount; i += 1) {
        const formatedStartTime = dayjs(startTime)
          .add(i * stepDuration.ms, 'ms')
          .toDate();
        const formatedArrivalTime = dayjs(startTime)
          .add((i + 1) * stepDuration.ms, 'ms')
          .toDate();
        computedOccurrences.push({
          id: `occurrence-${i}-${id}`,
          trainName: computeOccurrenceName(trainName, i),
          rollingStock,
          startTime: fomartedStartTime,
          arrivalTime: fomartedArrivalTime,
        });
      }
      return { occurrencesCount, occurrences: computedOccurrences };
    };

    setOccurrencesState(computeOccurrences());
  }, [pacedTrain]);

  return occurrencesState;
};
```

Yohh: just a little destructuring to make it more readable, feel free to avoid it if it seems to you overkill

front/src/modules/trainschedule/components/Timetable/PacedTrain/OccurrenceItem.tsx
```diff
@@ -0,0 +1,71 @@
...
+  </div>
+);
+
+export type Occurrence = {
+  id: string;
+  trainName: string;
+  rollingStock?: LightRollingStockWithLiveries;
+  startTime: Date;
+  arrivalTime: Date;
+};
```
Yohh: could be exported from `types.ts`

front/src/styles/scss/applications/operationalStudies/_pacedTrain.scss
```diff
@@ -121,8 +131,134 @@
...
+      }
+
+      &-time {
+        width: 44px;
+        text-align: right;
+        &.arrival-time {
+          margin-left: 32px;
+        }
+      }
+      .consecutive-day-date-display {
```
Yohh:
```suggestion
      .consecutive-day-timeline-display {
```

just because it's a bit weird to have `date-display` nested, it could be a source of error during future development

Yohh (commented):
while trying to modify the train name, I have this error:
image

don't know if it's better to add `trainschedule-` in `computedOccurrences`ids or change to modifier conditions

Yohh:
can't duplicate or delete an occurrences list, is it for further development ?

front/src/styles/scss/applications/operationalStudies/_pacedTrain.scss
```diff
@@ -121,8 +131,134 @@
...
+          margin: 0 7px 0 5px;
+        }
+        &::after {
+          content: '';
+          position: absolute;
+          top: 0;
+          left: 0;
+          width: 24px;
+          height: 22px;
+          background: linear-gradient(270deg, rgba(247, 246, 238, 0), rgba(247, 246, 238, 1));
```
Yohh, line 207: gradient color doesn't follow the background, @thibautsailly it's the colors mentioned in the mockup, is it on purpose?

front/src/styles/scss/applications/operationalStudies/_pacedTrain.scss
Math-R: There is no day's timeline here just date of the days and number of month maybe I don't get your comment

front/src/modules/trainschedule/components/Timetable/PacedTrain/OccurrenceItem.tsx
Math-R: sure !

front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
Math-R: No I'm okay with you destructuring is better in that case for readability

front/src/modules/trainschedule/components/Timetable/PacedTrain/PacedTrainItem.tsx
```diff
@@ -120,7 +123,15 @@ const PacedTrainItem = ({
         editTimetableItem={() => selectPacedTrainToEdit(pacedTrain)}
         deleteTimetableItem={deletePacedTrain}
       />
-      <div className="occurrences" />
+      <div className="occurrences">
+        {occurrences.map((occurrence, index) => (
+          <OccurrenceItem
+            occurrence={occurrence}
+            key={index}
```
kmer2016:
```suggestion
            key={occurrence.id}
```

Math-R:
> while trying to modify the train name, I have this error:
> 
> image
> 
> 
> 
> don't know if it's better to add `trainschedule-` in `computedOccurrences`ids or change to modifier conditions

Is it because the pacedtrain is actually a duplicate from the first train, you can make the paced in édition statue to visualise the effectivement but you can Really edit it at this time

Math-R:
> can't duplicate or delete an occurrences list, is it for further development ?

Yes it is

front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
```diff
@@ -0,0 +1,62 @@
...
+          rollingStock: pacedTrain.rollingStock,
+          startTime,
+          arrivalTime,
+        });
+      }
+      return { occurrencesCount, occurrences: computedOccurrences };
+    };
+
+    setOccurrencesState(computeOccurrences());
+  }, [pacedTrain]);
```
kmer2016: No need to pass `stepDuration` in useEffect dependencies ?
```suggestion
  }, [pacedTrain, stepDuration]);
```

Math-R: We remove the second hook parameters so we don't need to watch it anymore.

front/src/modules/trainschedule/components/Timetable/PacedTrain/PacedTrainItem.tsx
Math-R: done

front/src/styles/scss/applications/operationalStudies/_pacedTrain.scss
Math-R: But i don't have passion so let's go

Yohh: hum ok it's fine IMHO

kmer2016 (commented):
According to the acceptance criteria, it should be possible to select an occurrence, but it seems to me that this is not the case.

Yohh (approved):

front/src/modules/trainschedule/components/Timetable/types.ts
```diff
@@ -73,3 +73,11 @@ export type TimetableFilters = {
   selectedTags: Set<string | null>;
   setSelectedTags: React.Dispatch<React.SetStateAction<Set<string | null>>>;
 };
+
+export type Occurrence = {
+  id: string;
```
SharglutDev: Shouldn't this be of type `OccurrenceId` ?

kmer2016 (approved):
LGTM !

front/src/modules/trainschedule/components/Timetable/types.ts
Math-R: done

front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
```diff
@@ -0,0 +1,62 @@
...
+      const computedOccurrences = [];
+      for (let i = 0; i < occurrencesCount; i += 1) {
+        const occurrenceStartTime = dayjs(startTime)
+          .add(i * paced.step.ms, 'ms')
+          .toDate();
+        const occurrenceArrivalTime = dayjs(startTime)
+          .add((i + 1) * paced.step.ms, 'ms')
+          .toDate();
+        computedOccurrences.push({
+          id: `occurrence-${i}-${id}` as OccurrenceId,
```
SharglutDev: Here you can use `formatEditoastTrainIdToOccurrenceId` function

SharglutDev: Or maybe it's not exactly what we need actually since you already receive a `PacedTrainId`. Maybe we could add another util for type safety.

front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
```diff
@@ -0,0 +1,62 @@
...
+    }
+    if (!Number.isNaN(+pacedTrainName)) {
+      return `${+pacedTrainName + 2 * index}`;
+    }
+    return `${pacedTrainName} ${2 * index + 1}`;
+  };
+
+  useEffect(() => {
+    const computeOccurrences = (): OccurrencesState => {
+      const occurrencesCount = Math.floor(paced.duration.ms / paced.step.ms);
```
SharglutDev:
```suggestion
      const occurrencesCount = Math.ceil(paced.duration.ms / paced.step.ms);
```

front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
```diff
@@ -0,0 +1,62 @@
...
+      const computedOccurrences = [];
+      for (let i = 0; i < occurrencesCount; i += 1) {
+        const occurrenceStartTime = dayjs(startTime)
+          .add(i * paced.step.ms, 'ms')
+          .toDate();
+        const occurrenceArrivalTime = dayjs(startTime)
+          .add((i + 1) * paced.step.ms, 'ms')
+          .toDate();
```
SharglutDev: Shouldn't we use `arrivalTime` here ?

front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
```diff
@@ -0,0 +1,62 @@
...
+      const endOfPacedTrainName = Number(pacedTrainName.split(' ').pop());
+      return `${pacedTrainName.replace(/\s\d+$/, '')} ${endOfPacedTrainName + 2 * index}`;
+    }
+    if (!Number.isNaN(+pacedTrainName)) {
+      return `${+pacedTrainName + 2 * index}`;
+    }
+    return `${pacedTrainName} ${2 * index + 1}`;
+  };
+
+  useEffect(() => {
```
SharglutDev: Do we need a `useEffect` + `useState` here ? Wouldn't it work with just a `const occurrenceData = useMemo(...` ?

front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
```diff
@@ -0,0 +1,62 @@
...
+};
+
+const useOccurrences = (pacedTrain: PacedTrainWithResult) => {
+  const { id, paced, startTime, trainName, rollingStock } = pacedTrain;
+  const [occurrencesState, setOccurrencesState] = useState<OccurrencesState>({
+    occurrences: [],
+    occurrencesCount: 0,
+  });
+
+  const computeOccurrenceName = (pacedTrainName: string, index: number): string => {
```
SharglutDev: Good job with that function ! Would be nice to have some unit tests for all cases.

front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
Yohh: I wonder if it's a good idea to compare useEffect() + useState() and useMemo() every time, I know it's something we do many times in the project, but each solution has its own purpose, and memoization is not a thing we should use to replace a syntaxe everywhere, it's about keeping a result in memory to improve perfs. does this component really need memoization?

emersion: `useMemo()` is effectively the same as `useState()` + `useEffect()`, except more concise and with a clearer _data flow_. With `useMemo()`, the inputs (dependencies) and the outputs (return value) are clearly defined in a single place. With `useState()` and `useEffect()`, these two are often times far apart, one needs to check for other places which might call the setter to understand where the data is coming from, and it's tempting to break the flow by calling the setter somewhere else.

More generally, `useEffect()` should be avoided when possible, because it makes the state harder to reason about. I believe we over-use `useEffect()` in OSRD. More details on this topic here: https://react.dev/learn/you-might-not-need-an-effect

Yohh: ok for that, and for our over-use of useEffect(), but the link you share is about catching unnecessary useEffect(), and useMemo() is one solution, with one specific purpose: cache (or memoize) an expensive calculation

my question is about this specific case, for this hook that is not hard to read, with one prop and one useState(), does this hook really need memoization?

front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
Math-R: nice catch, i miss this one since it was already moved from the other file

front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
Math-R: yes you're completly true this calculation was so wrong thanks

front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
Math-R: will we need it somewhere else, has the occurrences solely lives in the pacedTrain context (for instance) ?

front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
Math-R: Here, useMemo is a better option because, there could possibly be many occurrences (a lot), and in a long timetable with also many pacedTrain, memoize some of the calculation (who can be heavy) is a good point.

It's not about readability, but prevent some recalculation.

I missed this one, thanks !

front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
SharglutDev: Maybe we will need it also when we add occurrence projection data and and simulation results (depending on how you plan to handle this)

front/src/modules/trainschedule/components/Timetable/PacedTrain/OccurrenceItem.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+      <div className="occurrence-item-horaries">
+        <div className="status-icon after-midnight">
+          {isAfterMidnight && <Moon iconColor="rgba(33, 100, 130, 0.7)" />}
+        </div>
+        <div className="occurrence-item-time departure-time">
+          {dayjs(startTime).format('HH:mm')}
+        </div>
+        <div className="occurrence-item-time arrival-time">
+          {dayjs(arrivalTime).format('HH:mm')}
```
SharglutDev, line 64: Capture d’écran 2025-03-04 à 16 34 27

Feels weird not to have the same arrival time between the first train and the first occurrence (but this will probably go away when the paced trains aren't mocked anymore)

front/src/modules/trainschedule/components/Timetable/PacedTrain/OccurrenceItem.tsx
```diff
@@ -0,0 +1,77 @@
+import { Dot, Moon } from '@osrd-project/ui-icons';
+import cx from 'classnames';
+import dayjs from 'dayjs';
+
+import RollingStock2Img from 'modules/rollingStock/components/RollingStock2Img';
+
+import type { Occurrence } from '../types';
+
+const ConsecutiveDayDateDisplay = ({
```
SharglutDev, line 9: Capture d’écran 2025-03-04 à 16 34 27

Has it been validated by @maelysLeratRosso for the edge case when changing month ? (here we display the next month, which can be deduced, but it can surprise someone who doesn't know how it works)

front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
SharglutDev, line 1: Not sure it can happen irl but it feels hard to know how long is the time range duration for the paced train for a large one like this (3 days) (@maelysLeratRosso @thibautsailly)

Capture d’écran 2025-03-04 à 16 42 51

front/src/modules/trainschedule/components/Timetable/PacedTrain/OccurrenceItem.tsx
maelysLeratRosso: Hmmm no I don't know how it should be represented... If I remember well, initially @thibautsailly imagined two full dates in all cases, but then it seemed to heavy so he grouped it with the month in common, as shown here. But maybe we should display both dates in that case. 

Is it technically long to add this special case for changing months ?

Math-R: We had a discussion with @thibautsailly saying to keep the same disposition and put the number of the new month instead of the previous month

maelysLeratRosso: Ok great then !

front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
Uriel-Sautron: It's just fake data to illustrate the situation over several days, but what's done is what's requested on the sketch (min)

SharglutDev: Discussed with @maelysLeratRosso, she's also fine leaving it as it is :)

front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
Math-R: Maybe if so we (@Uriel-Sautron and me) will make the change at that time, as we're in charge of this in the next issue, but right now it's not obvious so i prefer to see the needs

front/src/modules/trainschedule/components/Timetable/PacedTrain/hooks/useOccurrences.tsx
Uriel-Sautron: Done

front/src/modules/trainschedule/components/Timetable/PacedTrain/OccurrenceItem.tsx
SharglutDev: Tested again and even though the path duration shows `1h40`, the arrival for the occurrences is 1h41min after the departure. We should do something to match the round between duration and arrivals.

image

Uriel-Sautron: The function `formatDateHours` is used for arrival time and departure time in `TrainScheduleItem.tsx ` which rounds up to the nearest lower or higher minutes. Is this the expected behavior?
In this case, an occurrence may have a start time of 12h24 and 50s and an arrival time of 14h05 and 20s, that is a duration of 1h40 and 30s.

SharglutDev: Oh you're totally right, we probably have the same issue for train schedules so we can think a more global fix in another PR.

SharglutDev (approved):
Lgtm and tested, nice job !

kmer2016 (approved):

PR #10984, opened 2025-02-28
Use OpenFGA to store and check roles

sim51:
A single big commit, sorry, but I couldn't find a way to split it nicely since this change is rather atomic tbh.

# What's done here

- **New dependency**: `stdext`, used to retrieve the name of the unit test function in order to create an OpenFGA store with the same name
- Migrations removing the `authn_group_membership` and `authz_role` tables
- The column `name` of `authn_user` is now non-nullable.

### `editoast_authz`

- Removal of `trait BuiltinRoleSet`, it's not necessary as we can use `BuiltinRole` directly in unit tests. It also simplifies some type signatures.
- Modelization of the *required subset* of our authorization model
- **New object**: **`Regulator`** (the name can still change but please let me know quickly if that's necessary 🥺)
    - It's the aggregation of an auth driver querying the DB and an OpenFGA client
    - Meant to be used directly for operations not related to a single user (listing groups for instance)
    - It's used in the `Authorizer`, which API now solely focuses on operations targeting a *single user* (the issuer of the request)
- `Authorizer` changes:
	- Most of its API has been moved in the `Regulator`
	- All roles-related functions have been rewritten to use OpenFGA instead of Postgres
	- No more superuser handling. An `Authorizer` object now has to be tied to an existing user, without corner cases.
- Tests have been fully rewritten
- Some duplication of the OpenFGA setup for tests is necessary because the crates splitting.

### `models`

- Removed role-related functions (not part of the `StorageDriver` anymore)
- Add implementations of functions that were previously inlined in the CLI:
	- `get_group_id`
	- `ensure_group`
	- `list_users`
	- `list_groups`

### `views`

- Rewrote roles routes
- Put user ID deduction code in the `Authentication` (simplifies a bit STDCM log logic)
- Twists the meaning of `check_roles` which now returns ok if the `Authorizer`'s user has *one of* the required roles (instead of *all of them* previously).
- When authorization is disabled, inject `SkipAuthorization` instead of a fake `Authorizer` with `user_id = -1` like before. Simplifies the process and cuts corner cases.
- The `AppState` now includes a `Regulator` instead of an `fga::Client` (now embedded in the `Regulator`).
- `TestApp` changes
	- Adds a `test_name` parameter which is used to configure the name of the OpenFGA store dedicated to a unit test.
	- Adds a `test_app!` macro returning a `TestAppBuilder` with the `test_name` preconfigured using `stdext`.
	- It is no longer required to provide a DB pool or a core client, if none were configured, they are created by default.
	- Fixes the OpenFGA configuration if the store name exceeds 64 characters.
	- Ensures an authorization model if one is provided (by default if authorization is configured in the `TestAppBuilder`).
	- Test users creation responsibility now fall back to the `TestApp` instead of its builder. It allows us to configure multiple users for a single test. We'll enhance the `UserBuilder` interface when we'll support grants.
- Fixes the `project` test checking authorization.
- Simplifies STDCM log testing setup and adapts it to the new `TestApp` (these tests require configuration tweaks of the telemetry and some mocked users).

### `client`

- Adapt most `role` and `group` subcommands to use the `Regulator` instead of querying the DB directly.
- The latest authorization model is also ensured.

# Left to do

- **Extensive testing**
	- [X] With authorization
		- [X] No roles
		- [X] `Stdcm` only
		- [X] `OperationalStudies`
	- [x] Without authorization
	- [x] With authorization disabled
	- [x] With an `Admin` user with authorization enabled
- Some renaming of both definitions and modules. Not done yet to ease the review process. Check out the `TODO`s here and there in the diff.


flomonster (changes requested):
LGTM.

Found a bug I can't remove a role from a group. `editoast roles remove 44 Admin`

leovalais:
Should be fixed now, thanks

flomonster (approved):
LGTM

kmer2016 (approved):

hamz2a (approved):
Incredible, thank you!

PR #10987, opened 2025-03-03
front: fix icon change when locking an infra

### LINKED ISSUE #10880: No icon change when unlocking an infrastructure
### What happened?

When you unlock an infrastructure from the infrastructure list, if it is already open in the infrastructure editor the padlock icon only disappears when refreshed

Image

Image

The same for the opposite case (lock an unlocked infra)

the padlock icon should automatically disappear

Go to the editor and select a locked infra
go to the list of infras and unlock the selected infra

Recette (SNCF)

Firefox, Google Chrome

staging 776f9b3

theocrsb:
closes #10880

### DIFF
--- a/front/src/modules/infra/useInfra.ts
+++ b/front/src/modules/infra/useInfra.ts
@@ -1,42 +1,16 @@
-import { useCallback, useEffect, useState } from 'react';
-
 import { isNil } from 'lodash';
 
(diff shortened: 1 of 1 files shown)


Akctarus (approved):
tested, good fix, LGTM !

kmer2016 (approved):
LGTM!

Math-R (approved):
lgtm !

PR #10989, opened 2025-03-03
core: stdcm: reorganize stop data

### LINKED ISSUE #8974: refacto: stdcm: unify stop data
Currently, the stop data are spread all over the place. They are:

1. In [`TimeData`](https://github.com/OpenRailAssociation/osrd/blob/dev/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/TimeData.kt#L51) (min/max/current stop durations)
2. In [`InfraExplorerWithEnvelope`](https://github.com/OpenRailAssociation/osrd/blob/dev/core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt#L30) (position / current duration)
3. In [`InfraExplorer`](https://github.com/OpenRailAssociation/osrd/blob/dev/core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorer.kt#L153) (list of all possible stop locations)

This is a mess. It happened because we couldn't plan early enough with a clear vision of how those features would evolve. We should *ideally* merge the 1 and 2. The third one is a bit different as it's not just about the explored path, but it can still make new stop-related features more difficult.

This refactoring would take some planning workshops, the end goal isn't clear yet.

Note: the current situation is *functional* for now. It's just some tech debt, debugging is more difficult and new features may be limited.

---

Likely a requirement for:

1. Adding stops to avoid conflicts (https://github.com/osrd-project/osrd-confidential/issues/338)
2. Proper safety speed management (https://github.com/OpenRailAssociation/osrd/issues/9282)

## Definition of ready

- **Technical**

- **General**
  - [x] WSJF values are filled
  - [x] Validated by the PM

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/8974

* A new class provides the stop locations to the `InfraExplorer`. The actual behavior doesn't change much, it's just to split the responsibilities. We'll eventually tweak this class to provide overtake stops. The intent is to eventually provide "optional" stops that generate two explorers (with and without), but that behavior isn't supported yet, to avoid cluttering this PR.
* `InfraExplorer` now provides an accessor to list stops locations on the path, though it doesn't keep track of any temporal data. The stop data itself isn't stored in the explorer, it's just an accessor for the underlying `IncrementalPath`. 
* `InfraExplorerWithEnvelope` does not store anything itself (besides what's in the underlying `InfraExplorer`). It matches data from `InfraExplorer` stop locations and `TimeData` stop data to build the resulting envelope. No more redundant data there.

With these changes, it should be easier to implement proper overtakes. Safety speeds as well, though we'd need to be careful about lookahead. We need to know about the stop before entering the safety speed area. 

---

EDIT: back to draft, there's some bugs around

So I missed some complexity:

1. We often refer to `graph.steps` to figure out if/when we should stop. That needs to go as well.
2. When the input says to stop at some given operational point, we may encounter several parts of that OP on the path. But we only need to stop once. Which involves keeping track of where we're at in the path.

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/api/api_v2/pathfinding/PathfindingBlocksEndpointV2.kt
@@ -16,6 +16,7 @@ import fr.sncf.osrd.reporting.warnings.DiagnosticRecorderImpl
 import fr.sncf.osrd.sim_infra.api.*
 import fr.sncf.osrd.stdcm.graph.extendLookaheadUntil
 import fr.sncf.osrd.stdcm.infra_exploration.initInfraExplorer
+import fr.sncf.osrd.stdcm.infra_exploration.stopProviderFromArrival
 import fr.sncf.osrd.utils.*
 import fr.sncf.osrd.utils.indexing.*
 import fr.sncf.osrd.utils.units.Length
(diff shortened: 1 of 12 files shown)


eckter:
Closing, there's a different branch (and thus different PR): https://github.com/OpenRailAssociation/osrd/pull/11109

PR #10990, opened 2025-03-03
front: fix MacroEditorState.getPathKeys() when missing secondary code

emersion:
getPathKey() (right above) produces different keys when an OP is missing a secondary code: it omits the "/code" suffix. Do the same here so that positions don't get lost for these OPs.


Uriel-Sautron (approved):
Tested, thank you

kmer2016 (approved):
LGTM !

PR #10991, opened 2025-03-03
core: compute min between 2 envelope parts and use it in ETCS

Erashin:
Compute the min curve between 2 envelope parts and use it in ETCS to compute min between indicationSvL and indicationEoA more precisely.
Disclaimer: with the simplification of the curves post calculation, there isn't actually a very distinctive improvement in the curve's precision..
image


core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.java
```diff
@@ -616,6 +613,65 @@ public EnvelopePart copyAndShift(double positionDelta, double minPosition, doubl
                 new HashMap<>(attrs), newPositions.toArray(), newSpeeds.toArray(), newTimeDeltas.toArray());
     }
 
+    /**
+     * Compute the minimum EnvelopePart between 2 envelope parts with an intersecting range [a, b].
+     * The resulting envelope part starts at the envelope with the minimum speed at a and ends at b.
+     */
+    public static EnvelopePart min(
+            EnvelopePart envelopePartA, EnvelopePart envelopePartB, Iterable<SelfTypeHolder> attrs) {
```
eckter: I would have thought it would have been easier to use if it returned the whole range covered by either. And possibly dealing with Envelopes instead of Parts. But you're the caller, you probably know what's most convenient.

eckter: Side note: we *could* just iterate meter per meter, interpolate each envelope, and pick the min (and *maybe* simplify it). It would probably be slower, but the difference may not be important enough to be an issue.

Though this method with the key position doesn't add that much complexity either.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.java
```diff
@@ -616,6 +613,65 @@ public EnvelopePart copyAndShift(double positionDelta, double minPosition, doubl
...
+        assert (beginPosA < endPosB && endPosA > beginPosB);
+        var startIntersectingRange = Math.max(beginPosA, beginPosB);
+        var start = envelopePartA.interpolateSpeed(startIntersectingRange)
+                        <= envelopePartB.interpolateSpeed(startIntersectingRange)
+                ? beginPosA
+                : beginPosB;
+        var end = Math.min(endPosA, endPosB);
+
+        TreeSet<Double> keyPositions =
+                Arrays.stream(envelopePartA.positions).boxed().collect(Collectors.toCollection(TreeSet::new));
```
eckter: I feel like we should really convert that file to kotlin if we write that kind of stuff :smile:

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.java
```diff
@@ -616,6 +613,65 @@ public EnvelopePart copyAndShift(double positionDelta, double minPosition, doubl
...
+        keyPositions.addAll(Arrays.stream(envelopePartB.positions).boxed().collect(Collectors.toSet()));
+        keyPositions = new TreeSet<>(keyPositions.subSet(start, true, end, true));
```
eckter:
```suggestion
        NavigableSet<Double> keyPositions =
                Arrays.stream(envelopePartA.positions).boxed().collect(Collectors.toCollection(TreeSet::new));
        keyPositions.addAll(Arrays.stream(envelopePartB.positions).boxed().collect(Collectors.toSet()));
        keyPositions = keyPositions.subSet(start, true, end, true);
```

Avoids instantiating a new TreeSet from the subset

eckter (commented):
Nice job

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.java
Erashin: I was actually in the midst of doing it and there were a few pbs while i was converting it. I'll give it another shot though, we both agree it's better.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.java
Erashin: We could. It's not a bad idea. But i don't feel like the complexity added is insane here. Tell me if you feel strongly about it, but I'd personally keep it as is.

eckter: No it's fine as it is

Erashin: Here were my thoughts on this, tell me what you think about it:
- if by "the whole range covered by either" (probably not what you mean) you mean the whole range, even after the end of one, well the envelope part would have discontinuities, so this isn't doable.
- if by "the whole range covered by either" you mean just the intersecting range of the two curves, then yeah that is a possibility. That would mean moving the part of the code that creates the first non intersecting part of the curve into EtcsBrakingCurves, which is clearly doable.
- for envelopes instead of envelope parts: the pb is how we're going to manage the discontinuities between the successive envelope parts, because they're not necessarily continuous (in our case they are). Except if you have a sol for that, it would work if we assert it is continuous for a min method between 2 Envelopes. The changes would be: putting the current method into a utils, taking positions + speeds as arguments. Call it in both EnvelopePart.min and Envelope.min (2 birds with one stone). EDIT: even then, how would we be managing the different EnvelopeParts' attributes?

eckter: We can and do have discontinuities in envelopes, it's just in the final simulation output that they're forbidden.

I meant the whole range, or possibly the whole range given by one of the inputs. 

I'm just a little surprised to see a method where we don't really know what range will be covered by the output. Generally speaking we have a reference envelope and we modify parts of it, so having the solution defined on the intersection of two parts feels a little off. 

It would probably be easier to understand if you describe where it would be called: what are the two parts that will be used as input? And why are we only interested in the intersection? (and why can't they be envelopes?)

Erashin: What I meant by Envelopes with discontinuities not being possible to use the method on is that we can't use `interpolateSpeed` and `intersectStep` on them. Imagine envelopeA going from a constant speed of 100 to a constant speed of 50, and envelopeB being at 75 at the time: how do we manage the intersection between the 2 curves? Would you just cut it off, create a new EnvelopePart starting at this discontinuity? How do we manage this without the method having quite a few specific cases and ending up being quite complex?


The 2 parts used as input are just 2 random parts in my head, except they have to have a common range of position because otherwise how would we determine which of them is the min? Only alternative i see would be to return an Envelope combining them both in that case, but that would mean the method shouldn't be called `min`, but probably something like `mergeMin` or something. And it would also mean the output can be discontinuous, which leads to the following point.


Why are we only interested in the intersection: the objective of this method, and also the reason why I'm working with EnvelopeParts rather than Envelopes, is to keep the continuity property of the EnvelopePart: the output will be continuous (which is handy when creating braking curves). The only way to have the output be continuous is:
- taking the min of the 2 curves at least in the position-range intersection
- following the curve before the intersection if it is the min at the start of the intersection
- following the curve after the intersection if it is the min at the end of the intersection (which i could have coded actually)

However, as you've said, a good alternative would be to have the position-range given as an input: only pb would be that, if the output is still an EnvelopePart, you'd have cases when the output's range would be smaller than the input range for continuity reasons. Which i think is ok-ish kinda.

eckter: So I feel like we're kind of dodging some important questions by only returning the intersection. The questions being what we're going to do at those borders. I wouldn't want to have the caller of this method reassemble some of the input parts to build a consistent final result, this method wouldn't be usable on its own. 

What I'm really uncomfortable about is not really knowing the "domain" of the output. We expect a set of points but without knowing where they'll be, it feels like we'd need a lot of weird post-processing (and pre-processing to even have a valid overlap). 

I'd really like if we could:

1. have a clear output domain that can't be ill-defined (such as "all of the range covered by the leftmost parameter")
2. Handle discontinuities in this method, or possibly in an auxiliary one. I don't think handling envelope parts really helps here. 

Actually we've had very similar use cases with mareco. Generally we tolerate discontinuities during intermediate steps, and then add accelerations or decelerations to fix them. But we need to plan carefully to avoid having weird errors or ill-defined results, the mareco module took a lot of rewrites (and isn't even stable today). 

---

I know I'm not really helping much here, I don't have any clear path to offer. If possible I'd like some fresh eyes (@bougue-pe :eyes:). Just to be clear: I'm not strongly against the current solution, just a little afraid of what's coming next

core/envelope-sim/src/test/java/fr/sncf/osrd/envelope/EnvelopePartTest.java
```diff
@@ -109,4 +114,43 @@ void testGetMechanicalEnergyConsumed() {
...
+    }
+
+    static Stream<Arguments> minEnvelopePartsArgs() {
+        return Stream.of(
+                Arguments.of(
+                        EnvelopeTestUtils.generateTimes(new double[] {0, 10, 20}, new double[] {100, 50, 0}),
+                        EnvelopeTestUtils.generateTimes(new double[] {5, 14, 16}, new double[] {150, 50, 0}),
+                        Stream.of(4),
+                        new double[] {0, 5, 10, 14, 15, 16},
+                        new double[] {100, 79.05694150420949, 50, 38.72983346207417, 35.35533905932738, 0}),
```
bougue-pe: This looks bugged, no?
I would expect the following results (easy to check for 5 and 15):
```suggestion
                        new double[] {0, 5, 10, 14, 15, 16},
                        new double[] {100, 75, 50, 30, 25, 0}),
```

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.java
```diff
@@ -616,6 +613,65 @@ public EnvelopePart copyAndShift(double positionDelta, double minPosition, doubl
...
+        var beginPosA = envelopePartA.getBeginPos();
+        var beginPosB = envelopePartB.getBeginPos();
+        var endPosA = envelopePartA.getEndPos();
+        var endPosB = envelopePartB.getEndPos();
+        assert (beginPosA < endPosB && endPosA > beginPosB);
+        var startIntersectingRange = Math.max(beginPosA, beginPosB);
+        var start = envelopePartA.interpolateSpeed(startIntersectingRange)
+                        <= envelopePartB.interpolateSpeed(startIntersectingRange)
+                ? beginPosA
+                : beginPosB;
```
bougue-pe: (if keeping the current behaviour) If speeds are **equal** at the beginning of pos-intersection, we probably want the `min(beginPosA, beginPosB)` (or `max`).
Otherwise, the result won't be identical if we exchange `envelopePartA` and `envelopePartB`.

core/envelope-sim/src/test/java/fr/sncf/osrd/envelope/EnvelopePartTest.java
```diff
@@ -109,4 +114,43 @@ void testGetMechanicalEnergyConsumed() {
...
+                Arguments.of(
+                        EnvelopeTestUtils.generateTimes(new double[] {5, 14, 16}, new double[] {100, 50, 0}),
+                        EnvelopeTestUtils.generateTimes(new double[] {0, 10, 20}, new double[] {150, 50, 0}),
+                        Stream.of(1, 4),
+                        new double[] {5, 7.142857142857143, 10, 14, 15, 16},
+                        new double[] {100, 90.63269671749657, 50, 38.72983346207417, 35.35533905932738, 0}));
```
bougue-pe: Unexpected output here too, if using `min(beginPositions)` in case of equality (both speeds are 100 at position 5):
```suggestion
                        new double[] {0, 5, 10, 14, 15, 16},
                        new double[] {150, 100, 50, 30, 25, 0}));
```

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.java
bougue-pe: 😅 This is a bit tough, I hope I understood the stakes.

The first use-case:
* computing the min of 2 braking curves that are continuous, monotonous (maybe not strictly if we consider the maintain-speed part of SvL).
* the result is expected to start from the `[min(beginPositions); maxSpeed]` point (both curves start -or can start- at `maxSpeed`).
* the end of the result is debatable:
  * can finish at `min(endPositions)`, then external work to add final parts: maintain-speed to reach second curve, then use the second curve to its `endPosition`.
  * can finish at `max(endPositions)`: internally add a maintain-speed part to link to second curve when `min(endPositions)` is reached.

I see 4 valid options here (detailed by my order of preference).

> [!IMPORTANT]
> **TLDR**
> * Working on `Envelopes` (and assume discontinuities) has my preference (on the expected RoI also).
> * We can easily manage continuity and size of position-ranges for the described case by adding missing parts, so any solution looks valid.

## 1 - work on `Envelopes` (force one intersection, return union of position ranges)

That would mean assuming discontinuities sometimes, indeed.
But forcing monotonous (and in the same direction) inputs guarantees monotonous output on the intersecting range.
And forcing continuous inputs guarantees continuous output on the intersecting range.
Monotony and/or continuity are quite easy to obtain on the first use-case we have (adding constant parts at the beginning and the end of curves so that their position range is equal), so it wouldn't be a major issue.

That would also mean applying attributes given in inputs, which is better IMO (allows for finer result).

That would require a clear choice of what attributes are applied when strictly equal (the previous min curve, then the next min curve, then the leftmost param if no previous?).

That means that going back to working on EnvelopePart is a bit tricky if required (converting `Envelope` slice to a `Part` while checking that `Part` requirements are respected).

That may require some kind of "final cleanup" (optional?) to regroup parts that are "identical" to obtain a clean single braking curve when possible (in our use case, it is not mandatory to avoid mutliple consecutive braking parts in a single ETCS braking, is it?).

That may require a bit more work to implement and to fit it in current code.

From the point-of-view of an external user of this method, I think we would have more use of this: it can more safely be used to add braking and acceleration curves (the extra cost on perf may be an issue?). Superposing them blindly would just work, and the code would be much clearer IMO.

## 2 - work on `EnvelopeParts` and return a union of position ranges (force one intersection, identical monotony)

TBD if the result should be an `Envelope` (my preference) or an `EnvelopePart`, see below.

The logic of the method would actually be a bit "reversed": return a curve that's the min position for all the speed range, and keep the min speed when 2 positions lead to different speed.

That would also mean assuming discontinuities sometimes (but it would not happen for the first use-case).
That would mean adding a maintain-speed piece:
* either a `point` to the `EnvelopePart` result if allowed (post-processing to change it to a separate EnvelopePart?)
* or just a dedicated `EnvelopePart` to the `Envelope` result (better IMO)

The amount of work is OK I think to have it fit in the current code.

Less external "unclear" code to assemble parts, the logic is contained and **a bit** more easy to explain IMO.

## 3 - Keep the current code

Staying with the current proposition is OK IMO, but we should probably move it back to the ETCSBrakingCurve.kt to avoid any misuse, as I agree it's a bit of a trap (for 2 acceleration curves, the top part would be broken).

Or keep it here, add asserts on braking and TODO for the rest to ensure future mutualization.

## 4 - work on `EnvelopeParts` and return intersection of position ranges (force one intersection, continuity, identical monotony)

This guarantees a continuous and monotonous output.
We may also force identical attributes of inputs to use the same one in output?
In the first use-case, we could add an extra first-point (pos: `0`, speed: `beginSpeed`) to make sure that intersection is enough, I think.

This would work for our first case, cover the concern on the output position-range and the amount of work is OK I guess.

It would keep the "unclear" external code that's assembling the parts.

Erashin: What was decided in the end:
- keep a min for `EnvelopeParts`, which computes the minimum between 2 envelope parts with the exact same position ranges ~on their intersecting range (if there is no intersecting range, throws)~.
- create a min for `Envelopes`, which computes the minimum envelope between 2 envelopes with the exact same position ranges. Use min for envelope parts in the method.
- use the min for `Envelopes` in `ETCSBrakingCurves`.

EDITED by @bougue-pe for range on envelope parts after reviews.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.java
Erashin: This isn't relevant anymore.

core/envelope-sim/src/test/java/fr/sncf/osrd/envelope/EnvelopePartTest.java
Erashin: Nah, check out the intersect method. Acceleration is supposed constant  between 2 following points, so we find the intersection between 2 parabolas, not 2 linear curves.
This is proven right by the assertions that check that the 2 compared envelopes have the same speed at the intersection.

EDIT: ~~might have talked too fast I'm getting a very bad output in the integration tests, holy.~~ nah juste a bad manipulation of the data on my part, it's crispy clean :D

core/envelope-sim/src/test/java/fr/sncf/osrd/envelope/EnvelopePartTest.java
Erashin: They aren't. See comment above (yes it's kinda cursed but w/e). Trust the assertions at intersection.

EDIT: ~~o.o this doesn't look good~~ nvm it was correct as expected.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/Envelope.kt
```diff
@@ -0,0 +1,414 @@
...
+/** Build the min envelopes between 2 envelopes starting and ending at the exact same positions. */
+fun minEnvelopes(envelopeA: Envelope, envelopeB: Envelope): Envelope {
+    val minParts = mutableListOf<EnvelopePart>()
+    assert(envelopeA.beginPos == envelopeB.beginPos)
+    assert(envelopeA.endPos == envelopeB.endPos)
+
+    val keyPositions =
+        Arrays.stream(envelopeA.partPositions)
+            .boxed()
+            .collect(Collectors.toCollection { TreeSet() })
```
eckter:
```suggestion
    val keyPositions = TreeSet(envelopeA.partPositions.toList())
```

It does have a small performance overhead compared to the stream (hoping that the stream version avoids copies), it's just a lot less clutter. Otherwise we could create a small utility function to hide it away.

(same for the line just below)

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
```diff
@@ -0,0 +1,729 @@
...
+    /** A list of N spacial offsets */
+    val positions: DoubleArray,
+    /** A list of N speeds, one per position */
+    val speeds: DoubleArray,
+    /** A list of N - 1 time deltas between positions */
+    private val timeDeltas: DoubleArray
+) : SearchableEnvelope {
+    // region INTRINSIC DATA FIELDS
+
+    /* !!! These arrays must stay private, as even public final arrays are mutable !!! */
```
eckter: Yeah those arrays did not stay private.

We can use [`DoubleArray.asList()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/as-list.html) to create a read-only getter of the arrays. I'm like 95% sure it would not copy the underlying array. 

Oh this would *also* fix the other issue with the boxed streams clutter.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
```diff
@@ -0,0 +1,729 @@
...
+    val beginPos = max(envelopePartA.beginPos, envelopePartB.beginPos)
+    val endPos = min(envelopePartA.endPos, envelopePartB.endPos)
+    assert(beginPos < endPos)
+
+    var keyPositions =
+        Arrays.stream(envelopePartA.positions)
+            .boxed()
+            .collect(Collectors.toCollection { TreeSet() })
+    keyPositions.addAll(Arrays.stream(envelopePartB.positions).boxed().collect(Collectors.toSet()))
+    keyPositions = TreeSet(keyPositions.subSet(beginPos, true, endPos, true))
```
eckter: The subset does not sound useful there

eckter (commented):
A few comments but mostly LGTM

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
Erashin: It actually is. This method only returns the min envelope part on the INTERSECTION between 2 input envelope parts. If i don't do this subset, there are positions outside of the intersection in `keyPositions`.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
Erashin: I'll do the same for `partPositions` in `Envelope.kt`.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
eckter: Wait I got confused and assumed that their position ranges were identical. Or that's just for envelopes?

Erashin: That's for Envelopes yeah :) I could as well assert it here ngl, I don't mind. Only changes are add assertion, modify method description, remove `.subset()`. You tell me, either way is fine for me.

eckter: If we expect the parts to cover an identical range each, yeah I think we should assert it and remove the subset. 

(I don't generally like asserts that restrict a function domain more than necessary, but returning the intersection could be confusing, see the debate above)

Erashin: Done :)

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
```diff
@@ -675,3 +676,60 @@ class EnvelopePart(
...
+    val keyPosList = ArrayList(keyPositions)
```
bougue-pe, line 692: We could iterate over `keyPositions`, keeping memory (nullable) of `prevPos`, `prevSpeedA` and `prevSpeedB` to avoid copy and double computation of prevSpeeds.
Probably wouldn't change the clarity of code, so I'd say it's good to get.

core/envelope-sim/src/test/java/fr/sncf/osrd/envelope/EnvelopeTest.kt
```diff
@@ -0,0 +1,204 @@
...
+            val partA =
+                EnvelopeTestUtils.generateTimes(
+                    doubleArrayOf(0.0, 20.0),
+                    doubleArrayOf(150.0, 50.0)
+                )
+            val partB =
+                EnvelopeTestUtils.generateTimes(doubleArrayOf(0.0, 20.0), doubleArrayOf(200.0, 0.0))
+            val partC =
+                EnvelopeTestUtils.generateTimes(doubleArrayOf(0.0, 10.0), doubleArrayOf(50.0, 0.0))
+            val partD =
```
bougue-pe: nit:
```suggestion
            val partC1 =
                EnvelopeTestUtils.generateTimes(doubleArrayOf(0.0, 10.0), doubleArrayOf(50.0, 0.0))
            val partC2 =
```

core/envelope-sim/src/test/java/fr/sncf/osrd/envelope/EnvelopeTest.kt
```diff
@@ -0,0 +1,204 @@
...
+                EnvelopeTestUtils.generateTimes(
+                    doubleArrayOf(10.0, 20.0),
+                    doubleArrayOf(200.0, 0.0)
+                )
+            val envelopeA = make(partA)
+            val envelopeB = make(partB)
+            val envelopeC = make(partC, partD)
+            val intersection1 = 17.5
```
bougue-pe:
```suggestion
            val intersectionAB = 17.5
```

core/envelope-sim/src/test/java/fr/sncf/osrd/envelope/EnvelopeTest.kt
```diff
@@ -0,0 +1,204 @@
...
+            val speedIntersection1 = envelopeA.interpolateSpeed(intersection1)
+            val intersection2 = 19.166666666666668
```
bougue-pe:
```suggestion
            val intersectionAC = 19.166666666666668
```

core/envelope-sim/src/test/java/fr/sncf/osrd/envelope/EnvelopeTest.kt
```diff
@@ -0,0 +1,204 @@
...
+        val resultingPoints = minEnvelope.iteratePoints()
+        val resultingPositions = resultingPoints.map { it.position }
+        val resultingSpeeds = resultingPoints.map { it.speed }
+
+        // Check that both envelopes have an equal speed at their intersection.
+        val intersectionPosition = resultingPositions[intersectionIndex]
+        val intersectionSpeed = resultingSpeeds[intersectionIndex]
+        assertTrue(
+            areSpeedsEqual(
+                envelopeA.interpolateSpeedLeftDir(intersectionPosition, 1.0),
```
bougue-pe: nit: as I wondered what the 1.0 was (same below)
```suggestion
        directionAscending = 1.0
        assertTrue(
            areSpeedsEqual(
                envelopeA.interpolateSpeedLeftDir(intersectionPosition, directionAscending),
```

eckter (approved):

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -103,80 +100,61 @@ private fun computeMinSvlEoaIndCurve(
...
                 doubleArrayOf(NATIONAL_RELEASE_SPEED, NATIONAL_RELEASE_SPEED)
             )
-        return Envelope.make(firstIndicationPart, maintainReleaseSpeedPart, lastIndicationPart)
-    } else {
-        return Envelope.make(firstIndicationPart, lastIndicationPart)
+        )
+
+    // Last part of the indication curve is the minimum of the SvL curve maintaining release speed
+    // and the EoA curve. It starts at the max between their beginning positions, and ends at target
+    // position.
```
bougue-pe:
```suggestion
    // Last part of the indication curve is the minimum of the SvL curve maintaining release speed
    // and the EoA curve. Working on their (extended) intersecting position range is enough as they both cover the targeted speed range.
```

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
```diff
@@ -103,80 +100,61 @@ private fun computeMinSvlEoaIndCurve(
...
+    val startIntersectingRange =
+        max(fullIndicationCurveEoa.beginPos, fullIndicationCurveSvl.beginPos)
+    val indCurveSvlOnIntersection =
+        Envelope(indCurveSvl.slice(startIntersectingRange, targetPosition))
+    val indCurveEoaOnIntersection =
+        Envelope.make(fullIndicationCurveEoa.slice(startIntersectingRange, targetPosition)!!)
+    val minIndCurveOnIntersection =
+        minEnvelopes(indCurveEoaOnIntersection, indCurveSvlOnIntersection)
+
+    // Add the start of the corresponding curve if needed.
```
bougue-pe: in that context, we could just "maintain" (advance) the initial speed of the latest starting curve and just apply the minimum.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
```diff
@@ -0,0 +1,728 @@
...
+        fun generateTimes(
+            attrs: Iterable<SelfTypeHolder>,
+            positions: DoubleArray,
+            speeds: DoubleArray
+        ): EnvelopePart {
+            return EnvelopePart(attrs, positions, speeds, computeTimes(positions, speeds))
+        }
+
+        // endregion
+        // region ATTRS
```
bougue-pe, line 617: nit: leftovers to be removed? (ditto for all)

Erashin: Not my code.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingCurves.kt
Erashin: Kinda modified that but yeah all good :)

bougue-pe (approved):
Small comments to big PR and more kotlin 🎉 
Thanks !

bougue-pe (approved):

PR #11022, opened 2025-03-04
editoast: PacedTrain: add simulation summary and simulation endpoints

younesschrifi:
Add simulation summary and simulation endpoints for paced trains.

### DIFF
--- a/editoast/src/views/mod.rs
+++ b/editoast/src/views/mod.rs
@@ -77,6 +77,7 @@ use crate::client::get_app_version;
 use crate::core::mq_client;
 use crate::core::pathfinding::PathfindingInputError;
 use crate::core::pathfinding::PathfindingNotFound;
+use crate::core::simulation::SimulationResponse;
 use crate::core::version::CoreVersionRequest;
 use crate::core::AsCoreRequest;
 use crate::core::CoreClient;
@@ -93,6 +94,7 @@ use crate::map::MapLayers;
 use crate::models;
 use crate::models::auth::PgAuthDriver;
 use crate::valkey_utils::ValkeyConfig;
+use crate::views::path::pathfinding::PathfindingFailure;
 use crate::ValkeyClient;
 
 crate::routes! {
@@ -171,6 +173,7 @@ pub struct InfraIdQueryParam {
 }
 
 #[derive(Debug, Serialize, ToSchema)]
+#[cfg_attr(test, derive(PartialEq, Deserialize))]
 #[serde(tag = "status", rename_all = "snake_case")]
 enum SimulationSummaryResult {
     /// Minimal information on a simulation's result
@@ -199,6 +202,47 @@ enum SimulationSummaryResult {
     SimulationFailed { error_type: String },
     /// InputError
     PathfindingInputError(PathfindingInputError),
+}
+
+impl From<core::simulation::SimulationResponse> for SimulationSummaryResult {
+    fn from(sim: SimulationResponse) -> Self {
+        match sim {
+            SimulationResponse::Success {
+                final_output,
+                provisional,
+                base,
+                ..
+            } => {
+                let report = final_output.report_train;
+                Self::Success {
+                    length: *report.positions.last().unwrap(),
+                    time: *report.times.last().unwrap(),
+                    energy_consumption: report.energy_consumption,
+                    path_item_times_final: report.path_item_times.clone(),
+                    path_item_times_provisional: provisional.path_item_times.clone(),
+                    path_item_times_base: base.path_item_times.clone(),
+                }
+            }
+            SimulationResponse::PathfindingFailed { pathfinding_failed } => {
+                match pathfinding_failed {
+                    PathfindingFailure::InternalError { core_error } => {
+                        Self::PathfindingFailure { core_error }
+                    }
+
+                    PathfindingFailure::PathfindingInputError(input_error) => {
+                        Self::PathfindingInputError(input_error)
+                    }
+
+                    PathfindingFailure::PathfindingNotFound(not_found) => {
+                        Self::PathfindingNotFound(not_found)
+                    }
+                }
+            }
+            SimulationResponse::SimulationFailed { core_error } => Self::SimulationFailed {
+                error_type: core_error.get_type().into(),
+            },
+        }
+    }
 }
 
 /// Represents the bundle of information about the issuer of a request
@@ -661,6 +705,71 @@ mod tests {
 
     use super::test_app::TestAppBuilder;
     use crate::core::mocking::MockingClient;
+
+    #[cfg(test)]
+    pub fn mocked_core_pathfinding_sim_and_proj(train_id: i64) -> MockingClient {
+        let mut core = MockingClient::new();
+        core.stub("/v2/pathfinding/blocks")
+            .method(reqwest::Method::POST)
+            .response(StatusCode::OK)
+            .json(json!({
+                "blocks":[],
+                "routes": [],
+                "track_section_ranges": [],
+                "path_item_positions": [0,1,2,3],
+                "length": 1,
+                "status": "success"
+            }))
+            .finish();
+        core.stub("/v2/standalone_simulation")
+            .method(reqwest::Method::POST)
+            .response(StatusCode::OK)
+            .json(json!({
+                "status": "success",
+                "base": {
+                    "positions": [],
+                    "times": [],
+                    "speeds": [],
+                    "energy_consumption": 0.0,
+                    "path_item_times": [0, 1000, 2000, 3000]
+                },
+                "provisional": {
+                    "positions": [],
+                    "times": [],
+                    "speeds": [],
+                    "energy_consumption": 0.0,
+                    "path_item_times": [0, 1000, 2000, 3000]
+                },
+                "final_output": {
+                    "positions": [0],
+                    "times": [0],
+                    "speeds": [],
+                    "energy_consumption": 0.0,
+                    "path_item_times": [0, 1000, 2000, 3000],
+                    "signal_critical_positions": [],
+                    "zone_updates": [],
+                    "spacing_requirements": [],
+                    "routing_requirements": []
+                },
+                "mrsp": {
+                    "boundaries": [],
+                    "values": []
+                },
+                "electrical_profiles": {
+                    "boundaries": [],
+                    "values": []
+                }
+            }))
+            .finish();
+        core.stub("/v2/signal_projection")
+            .method(reqwest::Method::POST)
+            .response(StatusCode::OK)
+            .json(json!({
+                "signal_updates": {train_id.to_string(): [] },
+            }))
+            .finish();
+        core
+    }
 
     #[rstest]
     async fn health() {
--- a/editoast/src/views/paced_train.rs
+++ b/editoast/src/views/paced_train.rs
@@ -1,6 +1,13 @@
 use std::collections::HashMap;
 use std::collections::HashSet;
 
+use crate::core::simulation::SimulationResponse;
+use crate::error::Result;
+use crate::models::prelude::*;
+use crate::models::train_schedule::TrainSchedule;
+use crate::models::Infra;
+use crate::views::projection::ProjectPathTrainResult;
+use crate::views::ListId;
 use axum::extract::Json;
 use axum::extract::Path;
 use axum::extract::Query;
@@ -11,6 +18,7 @@ use editoast_authz::BuiltinRole;
 use editoast_derive::EditoastError;
 use editoast_models::DbConnectionPoolV2;
 use editoast_schemas::paced_train::PacedTrainBase;
+
 use serde::Deserialize;
 use serde::Serialize;
 use thiserror::Error;
@@ -19,17 +27,13 @@ use utoipa::ToSchema;
 
 use super::path::pathfinding::PathfindingResult;
 use super::projection::ProjectPathForm;
+use super::train_schedule::train_simulation_batch;
 use super::AppState;
 use super::AuthenticationExt;
 use super::InfraIdQueryParam;
 use super::SimulationSummaryResult;
-use crate::core::simulation::SimulationResponse;
-use crate::error::Result;
 use crate::models::paced_train::PacedTrain;
-use crate::models::prelude::*;
-use crate::views::projection::ProjectPathTrainResult;
 use crate::views::AuthorizationError;
-use crate::views::ListId;
 
 crate::routes! {
     "/paced_train" => {
@@ -56,10 +60,13 @@ editoast_common::schemas! {
 enum PacedTrainError {
     #[error("{count} paced train(s) could not be found")]
     #[editoast_error(status = 404)]
-    BatchPacedTrainNotFound { count: usize },
+    BatchNotFound { count: usize },
+    #[error("Paced train '{paced_train_id}', could not be found")]
     #[editoast_error(status = 404)]
-    #[error("Paced train {paced_train_id} does not exist")]
-    PacedTrainNotFound { paced_train_id: i64 },
+    NotFound { paced_train_id: i64 },
+    #[error("Infra '{infra_id}', could not be found")]
+    #[editoast_error(status = 404)]
+    InfraNotFound { infra_id: i64 },
     #[error(transparent)]
     #[editoast_error(status = 500)]
     Database(#[from] editoast_models::model::Error),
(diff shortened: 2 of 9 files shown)


editoast/src/views/paced_train.rs
```diff
@@ -202,19 +211,62 @@ struct SimulationBatchForm {
...
-        electrical_profile_set_id: _electrical_profile_set_id,
-        ids: _paced_train_ids,
+        infra_id,
+        electrical_profile_set_id,
+        ids: paced_train_ids,
     }): Json<SimulationBatchForm>,
 ) -> Result<Json<HashMap<i64, SimulationSummaryResult>>> {
-    todo!();
+    let authorized = auth
+        .check_roles([BuiltinRole::InfraRead, BuiltinRole::TimetableRead].into())
```
leovalais: You'll need to rebase to use the new role set

editoast/src/views/paced_train.rs
```diff
@@ -19,17 +28,14 @@ use utoipa::ToSchema;
 
 use super::path::pathfinding::PathfindingResult;
 use super::projection::ProjectPathForm;
+use super::train_schedule::simulation_response;
+use super::train_schedule::train_simulation_batch;
```
leovalais: Module tree is definitely wrong... (nothing to do in this PR, just commenting ^^)

editoast/src/views/paced_train.rs
```diff
@@ -202,19 +211,62 @@ struct SimulationBatchForm {
...
+
+    let paced_trains: Vec<PacedTrain> =
+        PacedTrain::retrieve_batch_or_fail(conn, paced_train_ids, |missing| {
+            PacedTrainError::BatchPacedTrainNotFound {
+                count: missing.len(),
+            }
+        })
+        .await?;
+    let paced_trains_to_ts: Vec<TrainSchedule> =
+        paced_trains.clone().into_iter().map_into().collect();
```
leovalais: This means we're always simulating at the time of the first occurrence right?

editoast/src/views/paced_train.rs
```diff
@@ -57,17 +63,20 @@ enum PacedTrainError {
     #[error("{count} paced train(s) could not be found")]
     #[editoast_error(status = 404)]
     BatchPacedTrainNotFound { count: usize },
+    #[error("Paced train '{paced_train_id}', could not be found")]
+    #[editoast_error(status = 404)]
+    NotFound { paced_train_id: i64 },
```
leovalais, line 66: Much better! No need to repeat the "PacedTrain". Maybe the same could be done for BatchPacedTrainNotFound for consistency?

editoast/src/views/timetable.rs
```diff
@@ -313,7 +313,7 @@ async fn post_paced_train(
     let changesets = paced_trains
         .into_iter()
         .map(PacedTrainChangeset::from)
-        .map(|cs| cs.timetable_id(timetable_id))
+        .map(|cs: PacedTrainChangeset| cs.timetable_id(timetable_id))
```
leovalais: Is this really required?

editoast/src/views/paced_train.rs
```diff
@@ -406,4 +496,60 @@ mod tests {
...
+        let (app, infra_id, train_schedule_id) =
+            app_infra_id_paced_train_id_for_simulation_tests().await;
+        let request = app.get(
+            format!(
+                "/paced_train/{}/simulation/?infra_id={}",
+                train_schedule_id, infra_id
+            )
+            .as_str(),
+        );
+        app.fetch(request).assert_status(StatusCode::OK);
```
leovalais: This test isn't really testing anything really. If we break the simulation processing at some point, but still return OK, this test won't catch anything even though there is indeed a regression. Better assert on the output of the endpoint, or at least check for some properties on it if it's too lengthy. Otherwise you could use [`insta`](https://insta.rs) for snapshot testing (it's already used in editoast to test the output of derive macros).

editoast/src/views/paced_train.rs
```diff
@@ -406,4 +496,60 @@ mod tests {
...
+    #[rstest]
+    async fn paced_train_simulation_summary() {
+        let (app, infra_id, paced_train_id) =
+            app_infra_id_paced_train_id_for_simulation_tests().await;
+        let request = app.post("/paced_train/simulation_summary").json(&json!({
+            "infra_id": infra_id,
+            "ids": vec![paced_train_id],
+        }));
+        app.fetch(request).assert_status(StatusCode::OK);
+    }
```
leovalais, line 630: Error cases could be tested as well.

editoast/src/views/train_schedule.rs
```diff
@@ -1105,7 +1107,7 @@ mod tests {
         )
     }
 
-    fn mocked_core_pathfinding_sim_and_proj(train_id: i64) -> MockingClient {
+    pub fn mocked_core_pathfinding_sim_and_proj(train_id: i64) -> MockingClient {
```
leovalais:
```suggestion
    pub(in crate::views) fn mocked_core_pathfinding_sim_and_proj(train_id: i64) -> MockingClient {
```

We may want to use a pattern that's easy to grep for when we'll have reworked the module tree. This way we can get all placed where the visibility may need to change. Wdyt?

leovalais (changes requested):
Thanks! Left a few comments, but nothing really blocking apart the ones about tests

editoast/src/views/paced_train.rs
younesschrifi: Yes, if the front needs the following occurrences, it must simulate the following trains by adding the step to the train start_time.

editoast/src/views/mod.rs
```diff
@@ -172,7 +172,7 @@ pub struct InfraIdQueryParam {
 
 #[derive(Debug, Serialize, ToSchema)]
 #[serde(tag = "status", rename_all = "snake_case")]
-enum SimulationSummaryResult {
+pub enum SimulationSummaryResult {
```
woshilapin: Why do you need to change the visibility? If I understand correctly, it's only used in submodules?

editoast/src/views/paced_train.rs
```diff
@@ -202,19 +211,62 @@ struct SimulationBatchForm {
 )]
 async fn simulation_summary(
     State(AppState {
-        db_pool: _db_pool,
-        valkey: _valkey_client,
-        core_client: _core,
+        db_pool,
+        valkey: valkey_client,
+        core_client: core,
```
woshilapin: So much for consistency. The `AppState` is not consistent, sometimes using the suffix `_client` sometimes not... But maybe we should avoid to actually invert it completely in our functions? Let's keep the suffix everywhere for example?

```suggestion
        valkey: valkey_client,
        core_client,
```

editoast/src/views/paced_train.rs
woshilapin: Maybe a comment would be nice to actually state the non-obvious here 🤷

editoast/src/views/paced_train.rs
```diff
@@ -262,23 +314,53 @@ pub struct ElectricalProfileSetIdQueryParam {
...
+            PacedTrainError::NotFound { paced_train_id }
+        })
+        .await?;
+
+    // Compute simulation of a paced_train
+    let (simulation, _) = train_simulation_batch(
+        &mut db_pool.get().await?,
+        valkey_client,
+        core_client,
+        &[paced_train.into()],
```
woshilapin: nitpick: I often don't like too much `.into()` since it hides the "into what" (you'll have to look somewhere else to figure out the types). You can use the `From` instead, which makes things more obvious. What do you think?

```suggestion
        &[TrainSchedule::from(paced_train)],
```

editoast/src/views/paced_train.rs
leovalais: Ok, how about not using `From` and `into` for this case? Since this is a rather strong semantic change, maybe a function `into_first_occurrence(self) -> TrainSchedule` would be more explicit?

editoast/src/views/train_schedule.rs
```diff
@@ -1105,7 +1107,7 @@ mod tests {
         )
     }
 
-    fn mocked_core_pathfinding_sim_and_proj(train_id: i64) -> MockingClient {
+    pub(in crate::views) fn mocked_core_pathfinding_sim_and_proj(train_id: i64) -> MockingClient {
```
woshilapin: Instead of changing the visibility, how about moving this function inside `src/views/mod.rs` and exposing it only for tests. Because if it's a function that is supposed to be used by 2 different submodules of `views`, then it's now a utility of `views`, not anymore a simple function of `view::train_schedule`.

```suggestion
    #[cfg(test)]
    pub fn mocked_core_pathfinding_sim_and_proj(train_id: i64) -> MockingClient {
```

editoast/src/views/paced_train.rs
leovalais: We should do an homogenization pass at some point

editoast/src/views/mod.rs
younesschrifi: I also need it for tests.

front/src/config/openapi-editoast-config.cts
```diff
@@ -15,8 +15,10 @@ const config: ConfigFile = {
         'postInfraByInfraIdPathfinding',
         'postInfraByInfraIdPathfindingBlocks',
         'postInfraByInfraIdPathProperties',
+        'postPacedTrainSimulationSummary',
```
SharglutDev, line 18:
```suggestion
```

front/src/config/openapi-editoast-config.cts
```diff
@@ -15,8 +15,10 @@ const config: ConfigFile = {
...
         'postTrainSchedule',
```
SharglutDev: Not (directly) related to your PR, this one doesn't exist anymore since this [PR](https://github.com/OpenRailAssociation/osrd/pull/10655).

editoast/src/views/mod.rs
woshilapin: I'm not sure I understand, so I pulled the code and tried it. The warning (it's actually not even an error) is about the function `views::train_schedule::simulation_response` that uses `SimulationSummaryResult` as a return type. The warning mentions that the function has a broader visibility than the type. But the function, now that it has been used by both `views::train_schedule` and `views::paced_train` should probably be moved one module up, in `views`.

[Here](https://github.com/OpenRailAssociation/osrd/pull/11059) is a proposition to solve the problem (and we can even loose the `pub` on `simulation_response`). I made this tiny PR target directly your branch so it can be merged into yours.

woshilapin: I'm also thinking, we might start to clutter the `views::mod.rs` file with a lot of thing. Maybe we need to change the hierarchy of modules. Moving from

```
views
- paced_train
- train_schedule
```

to something like

```
views
- train
  - paced_train
  - train_schedule
```
and then put all the common functionalities for both modules inside the `train` module. What do you think (ok for doing that in some other PR)? (cc @leovalais)

younesschrifi: Thank you for your PR ! I 
We talked about it with @leovalais and @flomonster. A rework will be done after merging all the endpoints.

leovalais: @woshilapin we thought about putting these modules in `timetable` (`train` is a bit too generic in the context imo)

editoast/src/views/paced_train.rs
```diff
@@ -202,19 +210,65 @@ struct SimulationBatchForm {
...
+            PacedTrainError::BatchNotFound {
+                count: missing.len(),
+            }
+        })
+        .await?;
+    let paced_trains_to_ts: Vec<TrainSchedule> = paced_trains
+        .clone()
+        .into_iter()
+        .map(|pt| pt.into_first_occurrence())
+        .collect();
```
woshilapin: nitpick: just clone what you need (the object inside the `Vec`, not the entire `Vec`). And you should be able to use a non-lambda version.

```suggestion
    let paced_trains_to_ts: Vec<TrainSchedule> = paced_trains
        .iter()
        .cloned()
        .map(PacedTrain::into_first_occurrence)
        .collect();
```

editoast/src/views/mod.rs
```diff
@@ -170,7 +172,7 @@ pub struct InfraIdQueryParam {
     infra_id: i64,
 }
 
-#[derive(Debug, Serialize, ToSchema)]
+#[derive(Debug, PartialEq, Serialize, Deserialize, ToSchema)]
```
```suggestion
#[derive(Debug, Serialize, ToSchema)]
#[cfg_attr(test, derive(PartialEq, Eq, Deserialize)]
```

editoast/src/views/mod.rs
```diff
@@ -201,6 +203,45 @@ enum SimulationSummaryResult {
     PathfindingInputError(PathfindingInputError),
 }
 
+fn simulation_response(sim: SimulationResponse) -> SimulationSummaryResult {
```
leovalais: `impl From<core::simulation::SimulationResponse> for SimulationSummaryResult`?

editoast/src/views/paced_train.rs
```diff
@@ -202,19 +210,65 @@ struct SimulationBatchForm {
...
+        PacedTrain::retrieve_batch_or_fail(conn, paced_train_ids, |missing| {
+            PacedTrainError::BatchNotFound {
+                count: missing.len(),
+            }
+        })
+        .await?;
+    let paced_trains_to_ts: Vec<TrainSchedule> = paced_trains
+        .clone()
+        .into_iter()
+        .map(|pt| pt.into_first_occurrence())
```
```suggestion
        .map(PacedTrain::into_first_occurrence)
```

editoast/src/views/paced_train.rs
```diff
@@ -315,19 +399,35 @@ async fn project_path(
 
 #[cfg(test)]
 mod tests {
-    use axum::http::StatusCode;
-    use pretty_assertions::assert_eq;
```
leovalais, line 319: :'(

It's useful for equality on deep structs like here

editoast/src/views/paced_train.rs
```diff
@@ -202,19 +210,65 @@ struct SimulationBatchForm {
...
+    )
+    .await?;
+
+    // Transform simulations to simulation summary
+    let mut simulation_summaries = HashMap::new();
+    for (paced_train, sim) in paced_trains.into_iter().zip(simulations) {
+        let (sim, _) = sim;
+        let simulation_summary_result = simulation_response(sim);
+        simulation_summaries.insert(paced_train.id, simulation_summary_result);
+    }
```
woshilapin: nitpick: not mutable, only one destructuration.

```suggestion
    let simulation_summaries = paced_trains
        .into_iter()
        .zip(simulations)
        .map(|(paced_train, (sim, _))| (paced_train.id, simulation_response(sim)))
        .collect();
```

front/public/locales/en/errors.json
```diff
@@ -241,8 +241,9 @@
     },
     "paced_train": {
       "Database": "Internal error (database)",
-      "BatchPacedTrainNotFound": "Some Paced trains could not be found",
-      "PacedTrainNotFound": "Paced train '{{paced_train_id}}' not found"
+      "BatchNotFound": "'{{count}}' paced trains could not be found",
```
leovalais:
```suggestion
      "BatchNotFound": "'{{count}}' paced train(s) could not be found",
```

nit++

editoast/src/views/paced_train.rs
```diff
@@ -315,19 +399,35 @@ async fn project_path(
...
+    use crate::views::tests::mocked_core_pathfinding_sim_and_proj;
+    use crate::views::InternalError;
+    use crate::views::SimulationSummaryResult;
+    use crate::{
+        models::{
+            fixtures::{create_simple_paced_train, create_timetable, simple_paced_train_base},
+            paced_train::PacedTrain,
+        },
+        views::{paced_train::PacedTrainResult, test_app::TestAppBuilder},
+    };
```
woshilapin: nitpick: can you expand all of `use` (one line for each). It's easier to review diff when they're flatten. You might be able to configure `rustfmt` to do that automatically (need the nightly formatter though).

front/public/locales/fr/errors.json
```diff
@@ -238,8 +238,9 @@
     },
     "paced_train": {
       "Database": "Erreur interne (base de données)",
-      "BatchPacedTrainNotFound": "Certaines missions sont introuvables",
-      "PacedTrainNotFound": "Mission '{{paced_train_id}}' non trouvée"
+      "BatchNotFound": "'{{count}}' missions sont introuvables",
```
leovalais:
```suggestion
      "BatchNotFound": "'{{count}}' mission(s) sont introuvable(s)",
```

nit++, but I really dislike seeing things like "1 objets créés"

woshilapin (approved):
Thanks for all the improvements. I left a bunch of nitpick comments.

SharglutDev (approved):
Lgtm for front (not tested)

leovalais (approved):
LGTM (not tested), a few last comments, thanks for the changes!

PR #11028, opened 2025-03-04
editoast: remove 'async_trait'

woshilapin:
It's not entirely needed since 'rust:1.75'. There is some limitations, but the project is not impacted by them.

This is what the warning looks like (and the reason why I chose to deactivate it in the `[lint.rust]` section in `Cargo.toml`).

warning for async_fn_in_trait


leovalais (approved):
🥳🥳🥳🥳🥳🥳🥳🥳

PR #11032, opened 2025-03-04
editoast: add an endpoint to list infra objects

Khoyo:


editoast/src/views/infra/objects.rs
```diff
@@ -114,6 +119,47 @@ async fn get_objects(
...
+struct ListObjectsResponse {
+    ids: Vec<String>,
+}
+
+#[utoipa::path(
+    get, path = "",
+    tag = "infra",
+    params(InfraIdParam, ObjectTypeParam),
+    responses(
+        (status = 200, description = "The list of objects", body = ListObjectsResponse),
```
leovalais:
```suggestion
        (status = 200, description = "The list of objects", body = inline(ListObjectsResponse)),
```

or 

```suggestion
        (status = 200, description = "The list of objects", body = Vec<String>),
```

directly

editoast/src/views/infra/objects.rs
```diff
@@ -114,6 +119,47 @@ async fn get_objects(
...
+        (status = 404, description = "Infra ID invalid")
+    )
+)]
+async fn list_objects(
+    Path(infra_id_param): Path<InfraIdParam>,
+    Path(object_type_param): Path<ObjectTypeParam>,
```
leovalais:
```suggestion
    Path(InfraIdParam { infra_id }): Path<InfraIdParam>,
    Path(ObjectTypeParam { object_type }): Path<ObjectTypeParam>,
```

editoast/src/views/infra/objects.rs
```diff
@@ -114,6 +119,47 @@ async fn get_objects(
...
+    ids: Vec<String>,
+}
+
+#[utoipa::path(
+    get, path = "",
+    tag = "infra",
+    params(InfraIdParam, ObjectTypeParam),
+    responses(
+        (status = 200, description = "The list of objects", body = ListObjectsResponse),
+        (status = 404, description = "Infra ID invalid")
```
leovalais:
```suggestion
```

for consistency with the rest of the routes (that'll come with ViewErrors)

editoast/src/views/infra/objects.rs
```diff
@@ -114,6 +119,47 @@ async fn get_objects(
...
+
+    let objects = infra
+        .list_objects(&mut db_pool.get().await?, object_type_param.object_type)
+        .await?;
+
+    Ok(Json(ListObjectsResponse { ids: objects }))
+}
+
 #[cfg(test)]
 mod tests {
```
leovalais, line 158: a test would be nice

editoast/src/views/infra/objects.rs
Khoyo: Generally, it's a "good idea" to return json objects as a top level response... (To allow adding stuff later without making it a breaking change) But i could go either way, wdyt

leovalais: I don't mind an object at all, but in this case since the schema is only used at this one place better inline it

editoast/src/views/infra/objects.rs
```diff
@@ -22,6 +22,11 @@ use crate::Retrieve;
 
 crate::routes! {
     "/objects/{object_type}" => get_objects,
+    "/objects/{object_type}/list" => list_objects,
```
woshilapin: It's not clear to me why we need both a `get_objects` and a `list_objects`. Couldn't we put the functionality directly in `/objects/{object_type}` (if no ID is provided in the body, it defaults to returning all of them?). In any case, it'd be nice to add a description in the PR to explain why this PR.

woshilapin (commented):
Not sure I understand yet the use case and/or the solution.

editoast/src/views/infra/objects.rs
Morgane-SJK: When we import the work schedules in OSRD, we look for their associated tracks, using GAIA. Sometimes, the track ids returned are in GAIA but not in the OSRD infrastructure. We don't want to keep these extra track ids.
To check this, we need an endpoint that returns track ids from the OSRD infrastructure. (Alternatively, the other solution would be to add another time-consuming track geometry verification step in the work schedule pipeline.)

Your suggestion would also be appropriate for this need

Morgane-SJK: (The problem with the actual endpoint is that it crashes when at least one track id in the given payload is not in the OSRD infrastructure)

Khoyo: Re why not return everything with an empty payload... That's mainly because that felt like a potential mistake waiting to happen, and that you probably do not want to download all tracksections just to get a list of ids

Morgane-SJK: Yes we only need the list of ids, i don't know if it would be possible to return this as default behavior of the existing endpoint ?

woshilapin: > Re why not return everything with an empty payload... That's mainly because that felt like a potential mistake waiting to happen, and that you probably do not want to download all tracksections just to get a list of ids

Sure, both default would make sense and I agree that it might lead to misuse. But having both `GET /objects/{object_type}` (which I expect to return a list), and `GET /objects/{object_type}/list` is confusing too. So it might just be a naming problem. What do you think of calling the new endpoint `GET /objects/{object_type}/ids` if it returns only identifiers?

leovalais (approved):

editoast/src/views/infra/objects.rs
```diff
@@ -22,6 +22,11 @@ use crate::Retrieve;
 
 crate::routes! {
     "/objects/{object_type}" => get_objects,
+    "/objects/{object_type}/ids" => list_objects_ids,
+}
+
+editoast_common::schemas! {
+    ListObjectsResponse,
 }
```
leovalais:
```suggestion
```

already inlined

PR #11033, opened 2025-03-04
front: fix trad no item in timetabletoolbar

Uriel-Sautron:
When our timetable is empty, we must display No item and not No train. During development of PacedTrainItem, we forgot to delete part of the code

### DIFF
--- a/front/public/locales/de/operational-studies.json
+++ b/front/public/locales/de/operational-studies.json
@@ -178,7 +178,6 @@
       "invalidTrains": "Einige Züge sind ungültig",
       "noItem": "Kein Item",
       "noSpeedLimitTagsShort": "Kein Gesch.-Schl.",
-      "noTrain": "Kein Zug",
       "punctuality": "Pünktlichkeit",
       "rollingStockFilterPlaceholder": "Baureihe, Serie, Detail",
       "selectAll": "Alles auswählen",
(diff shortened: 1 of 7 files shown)


SharglutDev:
While paced train is in dev mode, should we keep the `No train` when switch is off ?

Uriel-Sautron:
Even if the switch is off in the timetable, a train can be considered as an item.
But I understand your questioning.

theocrsb (approved):
LGTM and tested

maelysLeratRosso:
I think we should keep "Aucun train" in the classic mode, if it is not too long to implement, in order not to disturb users

Uriel-Sautron:
Ok we can leave this PR for the end of the paced train development.

flomonster:
> Ok we can leave this PR for the end of the paced train development.

I think we are there. What should we do with this PR?

SharglutDev:
Probably the only needed change should now be : 
```
label={computedItemLabel()}
```
in the checkbox label in `TimetableToolbar.tsx`. This way we will have the "No item".

Uriel-Sautron:
> > Ok we can leave this PR for the end of the paced train development.
> 
> I think we are there. What should we do with this PR?

I just updated the PR.

flomonster (approved):

clarani:
@Uriel-Sautron can you fix the build ?

theocrsb (approved):
Tested

SharglutDev (approved):

clarani (approved):

Synar (approved):

PR #11041, opened 2025-03-06
editoast: refactor core error

### LINKED ISSUE #10868: editoast: refactor `CoreError`

hamz2a:
closes https://github.com/OpenRailAssociation/osrd/issues/10868

### DIFF
--- a/editoast/src/core/mocking.rs
+++ b/editoast/src/core/mocking.rs
@@ -24,7 +24,6 @@ impl From<MockingClient> for CoreClient {
 #[derive(Debug)]
 pub struct MockingError {
     pub bytes: Vec<u8>,
-    pub status: StatusCode,
     pub url: String,
 }
 
@@ -96,13 +95,12 @@ impl MockingClient {
                 )
                 .expect("mocked response body should deserialize faultlessly"),
             )),
-            Some(StubResponse { code, .. }) => {
+            Some(StubResponse { .. }) => {
                 let err = response
                     .expect("mocked response body should not be empty when specified")
                     .to_vec();
                 Err(MockingError {
                     bytes: err,
-                    status: code,
                     url: req_path,
                 })
             }
--- a/editoast/src/core/mod.rs
+++ b/editoast/src/core/mod.rs
@@ -15,7 +15,6 @@ use std::fmt::Display;
 use std::marker::PhantomData;
 
 use axum::http::StatusCode;
-use editoast_derive::EditoastError;
 use mq_client::MqClientError;
 use serde::de::DeserializeOwned;
 use serde::Deserialize;
@@ -27,8 +26,6 @@ use tracing::trace;
 
 #[cfg(test)]
 use crate::core::mocking::MockingError;
-use crate::error::InternalError;
-use crate::error::Result;
 
 pub use mq_client::RabbitMQClient;
 
@@ -47,38 +44,19 @@ pub enum CoreClient {
 }
 
 impl CoreClient {
-    pub async fn new_mq(options: mq_client::Options) -> Result<Self> {
+    pub async fn new_mq(options: mq_client::Options) -> Result<Self, Error> {
         let client = RabbitMQClient::new(options)
             .await
-            .map_err(CoreError::MqClientError)?;
+            .map_err(Error::MqClientError)?;
 
         Ok(Self::MessageQueue(client))
     }
 
-    fn handle_error(&self, bytes: &[u8], url: String) -> InternalError {
-        // We try to deserialize the response as an StandardCoreError in order to retain the context of the core error
-        if let Ok(mut core_error) = <Json<StandardCoreError>>::from_bytes(bytes) {
-            let status: u16 = match core_error.cause {
-                CoreErrorCause::Internal => 500,
-                CoreErrorCause::User => 400,
-            };
-            core_error.context.insert("url".to_owned(), url.into());
-            let mut internal_error: InternalError = core_error.into();
-            internal_error.set_status(StatusCode::from_u16(status).unwrap());
-            return internal_error;
-        }
-
-        let error: InternalError = CoreError::UnparsableErrorOutput.into();
-        let mut error = error.with_context("url", url);
-        error.set_status(StatusCode::INTERNAL_SERVER_ERROR);
-        error
-    }
-
     #[tracing::instrument(name = "ping_core", skip_all)]
-    pub async fn ping(&self) -> Result<bool, CoreError> {
+    pub async fn ping(&self) -> Result<bool, Error> {
         match self {
             CoreClient::MessageQueue(mq_client) => {
-                mq_client.ping().await.map_err(|_| CoreError::BrokenPipe)
+                mq_client.ping().await.map_err(|_| Error::BrokenPipe)
             }
             #[cfg(test)]
             CoreClient::Mocked(_) => Ok(true),
@@ -97,7 +75,7 @@ impl CoreClient {
         path: &str,
         body: Option<&B>,
         infra_id: Option<i64>,
-    ) -> Result<R::Response> {
+    ) -> Result<R::Response, Error> {
         trace!(
             target: "editoast::coreclient",
             body = body.and_then(|b| serde_json::to_string_pretty(b).ok()).unwrap_or_default(),
@@ -112,14 +90,14 @@ impl CoreClient {
                 let response = client
                     .call_with_response(infra_id.to_string(), path, &body, true, None)
                     .await
-                    .map_err(CoreError::MqClientError)?;
+                    .map_err(Error::MqClientError)?;
 
                 if response.status == b"ok" {
                     return R::from_bytes(&response.payload);
                 }
 
                 if response.status == b"core_error" {
-                    return Err(self.handle_error(&response.payload, path.to_string()));
+                    return Err(Error::parse(&response.payload, path.to_string()));
                 }
 
                 todo!("TODO: handle protocol errors")
@@ -128,12 +106,8 @@ impl CoreClient {
             CoreClient::Mocked(client) => {
                 match client.fetch_mocked::<_, B, R>(method, path, body) {
                     Ok(Some(response)) => Ok(response),
-                    Ok(None) => Err(CoreError::NoResponseContent.into()),
-                    Err(MockingError { bytes, status, url }) => Err({
-                        let mut err = self.handle_error(&bytes, url);
-                        err.set_status(status);
-                        err
-                    }),
+                    Ok(None) => Err(Error::NoResponseContent),
+                    Err(MockingError { bytes, url }) => Err(Error::parse(&bytes, url)),
                 }
             }
         }
@@ -199,12 +173,12 @@ where
 
     /// Sends this request using the given [CoreClient] and returns the response content on success
     ///
-    /// Raises a [CoreError] if the request is not a success.
+    /// Raises a [enum@Error] if the request is not a success.
     ///
     /// TODO: provide a mechanism in this trait to allow the implementer to
     /// manage itself its expected errors. Maybe a bound error type defaulting
     /// to CoreError and a trait function handle_errors would suffice?
-    async fn fetch(&self, core: &CoreClient) -> Result<R::Response> {
+    async fn fetch(&self, core: &CoreClient) -> Result<R::Response, Error> {
         core.fetch::<Self, R>(
             self.method(),
             self.url(),
@@ -221,7 +195,7 @@ pub trait CoreResponse {
     type Response;
 
     /// Reads the content of `bytes` and produces the response object
-    fn from_bytes(bytes: &[u8]) -> Result<Self::Response>;
+    fn from_bytes(bytes: &[u8]) -> Result<Self::Response, Error>;
 }
 
 /// Indicates that the response that deserializes to `T` is expected to have a Json body
@@ -233,12 +207,9 @@ pub struct Bytes;
 impl<T: DeserializeOwned> CoreResponse for Json<T> {
     type Response = T;
 
-    fn from_bytes(bytes: &[u8]) -> Result<Self::Response> {
-        serde_json::from_slice(bytes).map_err(|err| {
-            CoreError::CoreResponseFormatError {
-                msg: err.to_string(),
-            }
-            .into()
+    fn from_bytes(bytes: &[u8]) -> Result<Self::Response, Error> {
+        serde_json::from_slice(bytes).map_err(|err| Error::CoreResponseFormatError {
+            msg: err.to_string(),
         })
     }
 }
@@ -246,7 +217,7 @@ impl<T: DeserializeOwned> CoreResponse for Json<T> {
 impl CoreResponse for Bytes {
     type Response = Vec<u8>;
 
-    fn from_bytes(bytes: &[u8]) -> Result<Self::Response> {
+    fn from_bytes(bytes: &[u8]) -> Result<Self::Response, Error> {
         Ok(Vec::from_iter(bytes.iter().cloned()))
     }
 }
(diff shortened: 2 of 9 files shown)


editoast/src/core/mod.rs
```diff
@@ -279,25 +256,24 @@ pub enum CoreError {
...
-    #[editoast_error(status = 500)]
     BrokenPipe,
 
     #[error(transparent)]
-    #[editoast_error(status = "500")]
-    MqClientError(MqClientError),
+    MqClientError(#[from] MqClientError),
+
+    #[error(transparent)]
+    StandardCoreError(#[from] StandardCoreError),
```
woshilapin, line 249: "Standard" feels vague, especially since there is also a variant "Generic". Not sure what each does.

editoast/src/views/timetable/stdcm.rs
```diff
@@ -292,7 +294,10 @@ async fn stdcm(
             .collect(),
     };
 
-    let stdcm_response = stdcm_request.fetch(core_client.as_ref()).await;
+    let stdcm_response: Result<core::stdcm::Response, InternalError> = stdcm_request
+        .fetch(core_client.as_ref())
+        .await
+        .map_err(Into::into);
```
woshilapin, line 303: Here, we convert the error side of `stdcm_response` into an `InternalError`... but only to transform it again L317 below into again something else (note, it's not part of your PR, the code was already like this before, it's just that looking at the modification, it highlighted it). Why not doing all the transformation right away, something like the following?

```suggestion
    let stdcm_response = stdcm_request
        .fetch(core_client.as_ref())
        .await
        .map(StdcmResponseOrError::Response)
        .map_err(|error| {
                let error: InternalError = error.into();
                StdcmResponseOrError::RequestError(serde_json::to_value(error).unwrap_or(
                    serde_json::Value::String("Failed to serialize the error".to_string()),
                ))
            });
```

You should be able to get rid of the lines below (317-325).

woshilapin (approved):
Thank you. Some comment, but mostly unrelated to the work of this PR, could be addressed in another PR.

editoast/src/core/mocking.rs
```diff
@@ -24,7 +24,6 @@ impl From<MockingClient> for CoreClient {
 #[derive(Debug)]
 pub struct MockingError {
     pub bytes: Vec<u8>,
-    pub status: StatusCode,
```
leovalais, line 27: Why does this have to go?

editoast/src/core/mod.rs
leovalais: Unless I'm wrong I think we could remove the other `GenericCoreError` and only keep the `StandardCoreError` one (maybe simply `Core` would suffice as a name?)

editoast/src/views/core_error_response.rs
```diff
@@ -0,0 +1,69 @@
+use axum::http::StatusCode;
+
+use crate::core::CoreErrorCause;
+use crate::core::Error;
+use crate::error::InternalError;
+
+impl From<Error> for InternalError {
```
leovalais: We should make a new type conveying the "core error as a response" semantics instead of relying on `InternalError` which is bound to go.

editoast/src/core/mod.rs
```diff
@@ -306,10 +282,10 @@ pub struct StandardCoreError {
     context: HashMap<String, Value>,
     message: String,
     #[serde(default = "CoreErrorCause::default")]
-    cause: CoreErrorCause,
+    pub(crate) cause: CoreErrorCause,
```
leovalais:
```suggestion
    pub cause: CoreErrorCause,
```

since this becomes part of the public `crate core` interface (used in views), let's make all the fields `pub` for api consistency

leovalais (changes requested):
Thanks for resuming the work on the split in crates!

editoast/src/core/mod.rs
hamz2a: Done.

editoast/src/views/timetable/stdcm.rs
woshilapin: Oh, after discussion, it is indeed not the same variable. The second one is scoped inside an `if` and the first one is still used after the `if`. So what I propose is not possible.

editoast/src/views/core_error_response.rs
hamz2a: Done.

editoast/src/core/mocking.rs
hamz2a: We needed this in `CoreClient.fetch()` because before, `fetch` returned an `InternalError` and we needed the status. Now, [`fetch` returns `core::Error`](https://github.com/OpenRailAssociation/osrd/blob/hai/editoast-refactor-core-error/editoast/src/core/mod.rs#L81), and we handle the status elsewhere.

editoast/src/core/mod.rs
hamz2a: For `GenericCoreError`, we serialize the error without a specific format `raw_error` ([example](https://github.com/OpenRailAssociation/osrd/blob/hai/editoast-refactor-core-error/editoast/src/core/mod.rs#L337)), while for `StandardCoreError`, we use it to parse the error and get a specific format ([example](https://github.com/OpenRailAssociation/osrd/blob/hai/editoast-refactor-core-error/editoast/src/core/mod.rs#L57)).
Both are used.

leovalais: I understand the difference but since they have the same meaning I believe we could replace `GenericCoreError` (it was created at a time when Core errors weren't homogeneous if I recall correctly)

editoast/src/core/mod.rs
```diff
@@ -456,10 +435,13 @@ mod tests {
...
+        let expected_error: StandardCoreError = serde_json::from_value(error).unwrap();
         let result = Req.fetch(&core.into()).await;
-        let expected_err: InternalError = error_with_status;
-        assert_eq!(result, Err(expected_err));
+        let result = result.err().unwrap();
+        if let Error::StandardCoreError(standard_core_error) = result {
+            assert_eq!(standard_core_error, expected_error);
+        } else {
+            panic!("Unexpected error type: {:?}", result);
+        }
```
leovalais: Nit: using assert_eq with the `Error` variant would be more idiomatic than if-else panicking imo

editoast/src/error.rs
```diff
@@ -290,6 +292,65 @@ impl EditoastError for editoast_models::model::Error {
     }
 }
 
+impl EditoastError for core::Error {
```
leovalais, line 311: The `inventory::submit!` is missing for translation keys, a few examples in this file or in macro expansion of `EditoastError`.

editoast/src/error.rs
```diff
@@ -290,6 +292,65 @@ impl EditoastError for editoast_models::model::Error {
...
+                    serde_json::to_value(raw_error).unwrap(),
+                ),
+            ]
+            .into(),
+            core::Error::UnparsableErrorOutput => Default::default(),
+            core::Error::ConnectionClosedBeforeMessageCompleted => Default::default(),
+            core::Error::ConnectionResetByPeer => Default::default(),
+            core::Error::BrokenPipe => Default::default(),
+            core::Error::MqClientError(_) => Default::default(),
+            core::Error::StandardCoreError(_) => Default::default(),
```
leovalais:
```suggestion
            _ => Default::default(),
```

editoast/src/views/core_error_response.rs
```diff
@@ -0,0 +1,38 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, ToSchema)]
+pub struct CoreErrorResponse {
+    #[serde(with = "StatusCodeRemoteDef", default = "default_status_code")]
+    #[schema(value_type = u16, minimum = 100, maximum = 599)]
+    pub status: StatusCode,
+    #[serde(rename = "type")]
+    pub error_type: String,
+    pub context: HashMap<String, Value>,
+    pub message: String,
+}
```
leovalais: `pub(in crate::views)` when necessary

editoast/src/views/mod.rs
```diff
@@ -1,4 +1,5 @@
 mod authz;
+pub mod core_error_response;
```
```suggestion
mod core_error_response;
use core_error_response::CoreErrorResponse;
```

avoids repetition in import paths

editoast/src/views/core_error_response.rs
```diff
@@ -0,0 +1,38 @@
...
+
+impl From<Error> for CoreErrorResponse {
```
leovalais:
```suggestion
impl From<core::Error> for CoreErrorResponse {
```

leovalais (commented):
thanks for the changes!

editoast/src/core/mod.rs
woshilapin: hint: `Error` does not implement `PartialEq` which I suppose led to this version of the code. Sure, you could add `#[cfg_attr(test, derive(PartialEq))]`, but you could also do the following I believe (not tested).

```rust
assert!(matches!(result, Error::StandardCoreError(expected_error)));
```

leovalais: I'm fine with both, with a slight preference for `assert_eq` still as it will show the actual value on failure unlike `matches!` which will consume it.

hamz2a: I can't implement `PartialEq` because `lapin::Error` and `serde_json::Error` inside `Error::MqClientError` don't implement this trait.

editoast/src/views/core_error_response.rs
hamz2a: Done.

editoast/src/views/mod.rs
hamz2a: Done.

editoast/src/core/mod.rs
leovalais: Oh ok, then I'd say that `matches!` is our best option then

editoast/src/views/core_error_response.rs
hamz2a: `pub(crate) struct CoreErrorResponse` -> `pub(crate) pub enum SimulationResponse` (core::simualtion) -> `pub(crate) enum Response` (core::stdcm) -> `pub(crate) pub enum StdcmResponseOrError` (models::stdcm_log) 

`StdcmResponseOrError` is a field of `StdcmLog` that implements the `Model` and throws this error:
```shell
error[E0446]: crate-private type `models::stdcm_log::StdcmResponseOrError` in public interface
  --> src/models/stdcm_log.rs:29:48
   |
23 | pub(crate) enum StdcmResponseOrError {
   | ------------------------------------ `models::stdcm_log::StdcmResponseOrError` declared as crate-private
...
29 | #[derive(Clone, Debug, Serialize, Deserialize, Model, ToSchema)]
   |                                                ^^^^^ can't leak crate-private type
   |
   = note: this error originates in the derive macro `Queryable` (in Nightly builds, run with -Z macro-backtrace for more info)
  ```

File: editoast/src/error.rs (hunk 8 above)
Comment (hamz2a) [reply editoast/src/error.rs]: Done.

File: editoast/src/core/mod.rs (hunk 6 above)
Comment (hamz2a) [reply editoast/src/core/mod.rs]: The `match` only checks the type of the variant (`Error::StandardCoreError`) and ignores the error's content `expected_error`. Is this what we want here?

Comment (woshilapin) [reply editoast/src/core/mod.rs]: Oh, right, with this syntax, it must bind to a new name `expected_error`. So maybe you can do it like this?

```rust
assert!(matches!(result, Error::StandardCoreError(standard_core_error) if standard_core_error == expected_error));
```

But honestly, at this point, I start to wonder if the version you have is not more readable in the end! I won't mind either version.

leovalais: Yeah I agree the match expression isn't great, it kinda looks like perl... About PartialEq you could use derivative to ignore lapin::Error: https://mcarton.github.io/rust-derivative/latest/cmp.html

hamz2a: Thank you for the suggestion, but strangely, it only works for `struct` and not for `enum`.
```rust
#[derive(Derivative)]
#[derivative(PartialEq)]
struct Foo {
    foo: u8,
    #[derivative(PartialEq = "ignore")]
    bar: u8,
}

#[derive(Derivative)]
#[derivative(PartialEq)]
enum Foo {
    Foo,
    #[derivative(PartialEq = "ignore")] // error: Unknown attribute `ignore` for trait `PartialEq`
    Bar,
}
```

editoast/src/error.rs
hamz2a: Done.

editoast/src/core/mod.rs
leovalais: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=9755b45b30128f52be01c75848a5208c

editoast/src/views/core_error_response.rs
leovalais: this is a `views` definition, why does changing its visibility impact the model? 🤔 looks like a circular dependency

editoast/src/core/mod.rs
leovalais: It looks like `derivative` has become unmaintained :/ 

https://github.com/OpenRailAssociation/osrd/issues/11071

editoast/src/core/mod.rs
hamz2a: Done.

editoast/src/core/mod.rs
hamz2a: Thanks for noticing that. In the meantime, should I keep the current version?

editoast/src/views/core_error_response.rs
hamz2a:
```shell
StdcmLog > StdcmResponseOrError > StdcmResponse > SimulationResponse > CoreErrorResponse
|--------------------------- models -------------------------------| |----- views ------|
```
`>` : include

editoast/src/core/mod.rs
leovalais: I'd suggest to still use derivative in this precise case, as it looks like we can just `sed/derivative/educe` so that creates little additional work.

editoast/src/views/core_error_response.rs
leovalais: well then something is off, remember that `model` should **not** use any definition from `views`, as per our target carte architecture

editoast/src/core/mod.rs
hamz2a: Done.

editoast/src/views/core_error_response.rs
hamz2a: It was a choice made for `StdcmLog`.

leovalais: 🤔 Could you link the github discussion where this choice was discussed?

hamz2a: In this PR [link](https://github.com/OpenRailAssociation/osrd/pull/9873), a dependency was created between `StdcmLog` and `InternalError`. In your [comment](https://github.com/OpenRailAssociation/osrd/pull/11041#discussion_r1983438943) 😅, you mentioned not using `internalError`.

leovalais: Yeah, but we shouldn't make `core` depend on `views` either 😅 The goal of this PR and the error refactoring as a whole is first and foremost to enable the split in crates to continue.

hamz2a: what about moving `CoreErrorResponse` into the `core` module like `SimulationResponse`, which implements `ToSchema` ?

editoast/src/core/mod.rs
```diff
@@ -233,83 +208,80 @@ pub struct Bytes;
...
+            core_error.context.insert("url".to_owned(), url.into());
+            return Error::StandardCoreError(core_error);
+        }
+        Error::UnparsableErrorOutput
+    }
+}
+
+#[derive(Debug, Deserialize, PartialEq)]
 pub struct StandardCoreError {
     #[serde(skip)]
```
leovalais, line 269: 😨

We should deserialize the status code as well (you can place the serde remote definition in `editoast_common`).

editoast/src/views/core_error_response.rs
leovalais: I think you can use `StandardCoreError` directly. See https://github.com/OpenRailAssociation/osrd/pull/11041#discussion_r1991339129

editoast/src/core/mod.rs
```diff
@@ -233,83 +208,80 @@ pub struct Bytes;
...
         Ok(())
     }
 }
 
 #[allow(clippy::enum_variant_names)]
-#[derive(Debug, Error, EditoastError)]
-#[editoast_error(base_id = "coreclient")]
-pub enum CoreError {
+#[derive(Debug, Error, Derivative)]
+#[derivative(PartialEq)]
```
leovalais:
```suggestion
#[derive(Debug, Error, PartialEq)]
```

no need for derivative here since there's nothing to ignore

hamz2a: Done.

editoast/src/views/core_error_response.rs
hamz2a: As discussed, we will leave the `InternalError` in `SimulationResponse` and `PathfindingCoreResult` for now, and handle it in a separate PR.

editoast/src/core/mod.rs
leovalais: missing commit maybe?

editoast/src/core/mod.rs
leovalais: As discussed, the refactoring required to handle "responses as errors" will be done after this PR is merged

leovalais (approved):
LGTM, thanks for the modifications (and for the work put into the crates split!).

As discussed, I'll let you:

1. check with @flomonster how to trim the `SimulationResponse` according to the intended meaning of the `SimulationFailed`
2. check with @younesschrifi how to trim the `PathfindingFailed` kind of objects to let in `mod core` only what's necessary to perform requests, keep response interpretation out of the module and remove the `views` dependency

editoast/src/core/mod.rs
hamz2a: Done.

PR #11044, opened 2025-03-06
front: remove grid margins

### LINKED ISSUE #11020: front: lmr: remove grid margins
Apparently the grid margin is meant to estimate how long the signals must be green before they're passed by the train. But we already have an accurate estimation with the sight distance, something they don't have in contexts where grid margin is used. So we add two redundant margins and we end up too pessimistic. 

We don't want to remove the feature (yet?), but we should set the default value to 0.

See https://github.com/OpenRailAssociation/osrd/pull/9984 for the last PR on this subject, we probably need to do the same kind of change again.

eckter:
They were redundant with the "green signal" requirements of our conflict detection module, we'd miss out on solutions for no good reason.

Fix https://github.com/OpenRailAssociation/osrd/issues/11020


clarani (approved):
LGTM :white_check_mark:

PR #11046, opened 2025-03-06
front: fix new rolling stock categories not reseting

### LINKED ISSUE #11024: front: other categories field isn't reset in rs editor
### What happened?

After the creation of a RS, when I try to create another RS, the "Other categories" field is already filled with the previous value

https://github.com/user-attachments/assets/f963adb1-e8cf-49d4-b2a6-aa4371115d2b

Reset the field when form is done

1. Create a RS with very simple values and fill the "Primary category"
2. video

Recette (SNCF)

Firefox

-

Synar:
Close #11024

### DIFF
--- a/front/src/modules/rollingStock/helpers/utils.ts
+++ b/front/src/modules/rollingStock/helpers/utils.ts
@@ -118,6 +118,7 @@ export const getRollingStockEditorDefaultValues = (
       }
     : {
         ...newRollingStockValues,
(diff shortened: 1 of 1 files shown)


louisgreiner (approved):
Tested and working great ! Thanks :-)

SharglutDev (approved):
Lgtm and tested ✅

PR #11055, opened 2025-03-06
Stop using valkey for tests

flomonster:


Khoyo (approved):
LGTM. Although, I'd really love to have some tests of the caching, I don't think we are testing it right now.

leovalais (approved):
@Khoyo We don't indeed. But for those a mocked redis interface is likely to be enough, so nothing invalidating this change imo.

Khoyo:
Sure, this was a "by the way, that would be real nice to have", not at all a blocker :)

woshilapin (approved):

PR #11059, opened 2025-03-07
editoast: move 'simulation_response' one module up

woshilapin:
Proposition to move some code around to avoid visibility shenanigans.


younesschrifi (approved):
LGTM ! :)

leovalais:
We had a call yesterday (🤔 ?) about the module tree being wrong, we planned to put everything under `views::timetable` and de-correlate the module tree from the the url path hierarchy. Wdyt?

PR #11060, opened 2025-03-07
docker-compose: use an env var to select image release channel

Castavo:
This allows to pull the images corresponding to specific published versions of OSRD. 
Usage: `RELEASE_CHANNEL=stable TAG=vX.X.X docker compose pull`.

I chose the name "release channel" because it's what rust uses to qualitfy nightly, beta and stable [in their docs](https://doc.rust-lang.org/book/appendix-07-nightly-rust.html).

Also, add an env var for osrdyne, in order to use the same uri for core workers as the ones for the other services.

I did not update `docker/docker-compose.pr-tests.yml` because PR versions are only published on edge.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -86,7 +86,7 @@ services:
   core:
     # This is a dummy container to build the core image
     # and document/provide parameters to other compose files
-    image: ghcr.io/openrailassociation/osrd-edge/osrd-core:${TAG-dev}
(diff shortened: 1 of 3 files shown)


Castavo:
The fact that the image url in `osrdyne.yml` doesn't use the `$TAG` env var leads me to think that we can't pin specific core version actually...

It makes this PR a little less useful

younesschrifi (approved):
Nice PR :)

woshilapin (approved):
Nice idea!

PR #11064, opened 2025-03-07
editoast: add 'NoopSpanExporter' to avoid dependencies

woshilapin:
The recent [update of opentelemetry dependencies](https://github.com/OpenRailAssociation/osrd/pull/11007) led me to remove `NoopSpanExporter`, not understand why we would manually implement ourselves something that [already exists in `opentelemetry_sdk`](https://docs.rs/opentelemetry_sdk/latest/opentelemetry_sdk/testing/trace/struct.NoopSpanExporter.html). But then I did some digging and found [that](https://github.com/OpenRailAssociation/osrd/pull/10122#discussion_r1895733870) explaining why we did it. So bringing back the old code, with a comment this time explaining why we do it.

### DIFF
--- a/editoast/Cargo.lock
+++ b/editoast/Cargo.lock
@@ -352,25 +352,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-process"
-version = "2.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


leovalais (approved):
It's indeed better with a comment, thanks!

hamz2a (approved):
LGTM, thank you!

PR #11069, opened 2025-03-10
editoast: adapt conflict endpoint to paced train

younesschrifi:
Adapt conflict endpoint to add paced trains that are involved with paced trains. 


editoast/src/core/conflict_detection.rs
```diff
@@ -78,16 +78,65 @@ pub struct ConflictDetectionResponse {
...
+            if let Some((paced_train_id, index)) = key.get(train_id) {
+                match index {
+                    Some(idx) => paced_train_occurrence_ids.push(PacedTrainOccurrenceId {
+                        paced_train_id: *paced_train_id,
+                        index: *idx,
+                    }),
+                    None => train_schedule_ids.push(*paced_train_id),
+                }
+            } else {
+                train_schedule_ids.push(train_id.parse::<i64>().unwrap());
```
Khoyo: I don't think this can never happen. Editoast shouldn't panic because of a bug in core

editoast/src/core/conflict_detection.rs
```diff
@@ -78,16 +78,65 @@ pub struct ConflictDetectionResponse {
...
+    pub end_time: DateTime<Utc>,
+    /// Type of the conflict
+    #[schema(inline)]
+    pub conflict_type: ConflictType,
+    /// List of requirements causing the conflict
+    pub requirements: Vec<ConflictRequirement>,
+}
+
+impl Conflict {
+    pub fn adapt_conflict(self, key: HashMap<String, (i64, Option<i32>)>) -> ConflictResponse {
```
Khoyo: This needs to be documented (also, `to_conflict_response()` might be a better name)

editoast/src/core/conflict_detection.rs
```diff
@@ -78,16 +78,65 @@ pub struct ConflictDetectionResponse {
...
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, ToSchema)]
+pub struct ConflictResponse {
     /// List of train schedule ids involved in the conflict
     pub train_schedule_ids: Vec<i64>,
-    /// List of paced train occurrences involved in the conflict.
-    /// Each occurrence is identified by a `paced_train_id` and its `index`
-    #[serde(default)]
```
Khoyo, line 90: The doc should stay

core/src/main/java/fr/sncf/osrd/conflicts/Conflicts.kt
```diff
@@ -26,7 +26,7 @@ interface ResourceRequirement {
 // TODO: when dropping v1, remove these structs and directly use Requirements
 class TrainRequirements(
     @Json(name = "train_id")
-    override val trainId: Long, // Not the usual RJS ids, but an actual DB id
+    override val trainId: String, // Not the usual RJS ids, but an actual DB id
```
Khoyo: It's not necessarily that anymore

editoast/src/views/timetable.rs
```diff
@@ -391,7 +391,7 @@ pub struct ElectricalProfileSetIdQueryParam {
     tag = "timetable",
     params(TimetableIdParam, InfraIdQueryParam, ElectricalProfileSetIdQueryParam),
     responses(
-        (status = 200, description = "List of conflict", body = Vec<Conflict>),
+        (status = 200, description = "List of conflict", body = Vec<ConflictResponse>),
     ),
 )]
 async fn conflicts(
```
Khoyo: That function is becoming way too long, could we cut it up a bit?

editoast/src/core/conflict_detection.rs
```diff
@@ -78,16 +78,65 @@ pub struct ConflictDetectionResponse {
...
+            }
+        }
+
+        ConflictResponse {
+            train_schedule_ids,
+            paced_train_occurrence_ids,
+            work_schedule_ids: self
+                .work_schedule_ids
+                .into_iter()
+                .map(|id| id.parse::<i64>().unwrap())
```
Khoyo: This unwrap is wrong too

Khoyo (commented):
Looks good overall, but this new behavior needs to be documented. I'd also love some explanation for the change in the commit message. (Which should be `core, editoast:` not `editoast: core:`)

editoast/src/views/timetable/stdcm/failure_handler.rs
```diff
@@ -90,12 +91,28 @@ impl SimulationFailureHandler {
...
-                    .train_schedule_ids
-                    .retain(|id| id != &virtual_train_id);
-                conflict
+                    .train_ids
+                    .retain(|id| id != &virtual_train_id.to_string());
+                ConflictResponse {
+                    train_schedule_ids: conflict
+                        .train_ids
+                        .iter()
+                        .map(|ts| ts.parse::<i64>().unwrap())
```
Khoyo: Same for this unwrap

editoast/src/views/timetable/stdcm/failure_handler.rs
```diff
@@ -90,12 +91,28 @@ impl SimulationFailureHandler {
...
+                        .collect(),
+                    paced_train_occurrence_ids: vec![],
+                    work_schedule_ids: conflict
+                        .work_schedule_ids
+                        .iter()
+                        .map(|ws| ws.parse::<i64>().unwrap())
```
Khoyo: And this one

core/src/main/java/fr/sncf/osrd/conflicts/IncrementalConflictDetector.kt
```diff
@@ -8,7 +8,7 @@ import java.util.TreeMap
 import kotlin.math.max
 import kotlin.math.min
 
-const val DEFAULT_WORK_SCHEDULE_ID: Long = -1
+const val DEFAULT_WORK_SCHEDULE_ID: String = "-1"
```
eckter: We can have a more meaningful value now (like `default work schedule ID`).

eckter (approved):
Core looks good

editoast/src/views/timetable/stdcm/failure_handler.rs
```diff
@@ -90,12 +91,28 @@ impl SimulationFailureHandler {
         let conflicts: Vec<_> = conflict_detection_response
             .conflicts
             .into_iter()
-            .filter(|conflict| conflict.train_schedule_ids.contains(&virtual_train_id))
+            .filter(|conflict| conflict.train_ids.contains(&virtual_train_id.to_string()))
```
Khoyo: Nit: this to_string is not free, do it out of the loop

editoast/src/views/timetable/stdcm/failure_handler.rs
```diff
@@ -90,12 +91,28 @@ impl SimulationFailureHandler {
...
             .map(|mut conflict| {
                 conflict
-                    .train_schedule_ids
-                    .retain(|id| id != &virtual_train_id);
-                conflict
+                    .train_ids
+                    .retain(|id| id != &virtual_train_id.to_string());
```
Khoyo: Just do a filter after the map below ;)

editoast/src/core/conflict_detection.rs
```diff
@@ -78,16 +78,65 @@ pub struct ConflictDetectionResponse {
...
+            start_time: self.start_time,
+            end_time: self.end_time,
+            conflict_type: self.conflict_type,
+            requirements: self.requirements,
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, ToSchema)]
+pub struct ConflictResponse {
```
leovalais: This looks like a view response with logic that seems superfluous in the core client. Shouldn't it be located in `views`?

editoast/src/core/conflict_detection.rs
leovalais: Good point. `into_conflict_response()` would be more usual though (and clippy may even whine about that `to_*` with a `self` move if I recall correctly)

editoast/src/core/conflict_detection.rs
```diff
@@ -78,16 +78,65 @@ pub struct ConflictDetectionResponse {
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize, ToSchema)]
 pub struct PacedTrainOccurrenceId {
     paced_train_id: i64,
-    index: i64,
+    index: i32,
 }
 
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize, ToSchema)]
```
leovalais: Are all these derives still necessary with the introduction of `ConflictResponse`?

editoast/src/core/conflict_detection.rs
```diff
@@ -78,16 +78,65 @@ pub struct ConflictDetectionResponse {
...
 pub struct PacedTrainOccurrenceId {
     paced_train_id: i64,
-    index: i64,
+    index: i32,
```
leovalais: Why i32 specifically?

editoast/src/core/conflict_detection.rs
```diff
@@ -78,16 +78,65 @@ pub struct ConflictDetectionResponse {
...
+    pub requirements: Vec<ConflictRequirement>,
+}
+
+impl Conflict {
+    pub fn adapt_conflict(self, key: HashMap<String, (i64, Option<i32>)>) -> ConflictResponse {
+        let mut train_schedule_ids = Vec::new();
+        let mut paced_train_occurrence_ids = Vec::new();
+
+        for train_id in &self.train_ids {
+            if let Some((paced_train_id, index)) = key.get(train_id) {
```
leovalais:
```suggestion
            if let Some((train_id, index)) = key.get(train_id) {
```

editoast/src/views/timetable.rs
```diff
@@ -442,23 +446,84 @@ async fn conflicts(
     )
     .await?;
 
+    let mut paced_trains_ts = vec![];
+    for paced_train in paced_trains.clone() {
+        if paced_train.step.num_seconds() == 0 {
```
leovalais: Do we even allow zero-stepped paced trains? :o 
What's the reason behind that?

editoast/src/views/timetable.rs
```diff
@@ -442,23 +446,84 @@ async fn conflicts(
...
+        let mut start_time = paced_train.start_time;
+
+        let all_occurrences: Vec<TrainSchedule> = (0..occurrences)
+            .map(|_| {
+                let mut first_occurrence = paced_train.clone().into_first_occurrence();
+                first_occurrence.start_time = start_time;
+                start_time += step;
+                first_occurrence
+            })
+            .collect();
```
leovalais: Weird style, either go fully functionnal with `map`/`reduce` or fully imperative, it's very unusual to have a `map` with side-effects.

editoast/src/views/timetable.rs
```diff
@@ -468,8 +533,12 @@ async fn conflicts(
 
     // 3. Call core
     let conflict_detection_response = conflict_detection_request.fetch(&core_client).await?;
-
-    Ok(Json(conflict_detection_response.conflicts))
+    let conflicts = conflict_detection_response.conflicts;
+    let conflicts_response = conflicts
+        .into_iter()
+        .map(|conflict| conflict.adapt_conflict(map_string_to_id.clone()))
```
leovalais: Is this clone really necessary?

editoast/src/views/timetable.rs
```diff
@@ -442,23 +446,84 @@ async fn conflicts(
...
             TrainRequirements {
                 start_time: train.start_time,
                 spacing_requirements: final_output.spacing_requirements,
                 routing_requirements: final_output.routing_requirements,
             },
         );
     }
+    let mut occurrences = HashMap::new();
+    for (train, sim) in paced_trains_ts.into_iter().zip(paced_train_simulations) {
+        let (sim, _) = sim;
```
```suggestion
    for (train, (sim, _)) in paced_trains_ts.into_iter().zip(paced_train_simulations) {
```

editoast/src/views/timetable.rs
```diff
@@ -442,23 +446,84 @@ async fn conflicts(
...
+        valkey_client.clone(),
+        core_client.clone(),
+        &paced_trains_ts,
+        &infra,
+        electrical_profile_set_id,
+    )
+    .await?;
+
+    let mut map_string_to_id: HashMap<String, (i64, Option<i32>)> = HashMap::new();
     // 2. Build core request
```
leovalais: These comments usually indicate good function splitting points.

editoast/src/core/conflict_detection.rs
leovalais: I think editoast should actually.

1. From an API perspective, a bug originating from core is no different from an editoast bug. If editoast performs a division by zero, we expect a panic which results in a 5xx (as it should). I don't get why core should have a special treatment (for debuggability reasons at the limit?).
2. From a client crate perspective, we only support the latest core version/API. The client doesn't have to be fault tolerant or check that the correct core version is used. (We could do both, but we currently don't, so panics are then appropriate imo.)

But an `expect` with a nice message would be better.

leovalais (dismissed):
Thanks! I think there's some endpoint logic leaking the `core` module. A few questions as well ⬇️

editoast/src/core/conflict_detection.rs
Khoyo: Maybe an expect is warranted, but only if this is moved out of the future `core` subcrate. Maybe the view wants to crash, maybe it doesn't. Having stuff panic because of an error from another service is general.

Unwrap might be ok when it's plainly obvious that the error cannot happen (eg. checking len() then taking the first element). Otherwise we really shouldn't use it at all, expect should be a minimum.

At least have the offending string in the expect message ;)

leovalais: 100% agree about expect vs unwrap, and I also do agree that this post-processing shouldn't occur in the `core` module/crate. The client shouldn't have any post-processing logic beyond what serde does imo (or at most a very simple mapping to type things more nicely).

editoast/src/core/conflict_detection.rs
flomonster: I suggest `u64`.

editoast/src/views/timetable.rs
flomonster: It should be forbidden. This field should be strictly positive.

editoast/src/core/conflict_detection.rs
flomonster: What do you think about replacing `(i64, Option<i32>)` (which can be a bit confusing) with : 
```rust
enum TrainId {
    TrainSchedule(int),
    PacedTrainOccurrence(PacedTrainOccurrenceId), // Could be also: PacedTrainOccurence{ id: i64, index: u64 }
}
```

editoast/src/core/conflict_detection.rs
```diff
@@ -13,18 +14,17 @@ use super::stdcm::WorkSchedule;
 
 editoast_common::schemas! {
     ConflictDetectionResponse,
-    Conflict,
+    ConflictResponse,
```
emersion: Out of curiosity, what motivated this rename? (It would've been nice to have a separate commit just for the rename, with an explanation in the commit message.)

editoast/src/views/timetable.rs
younesschrifi: Sadly yes :(

editoast/src/views/timetable.rs
```diff
@@ -367,6 +375,77 @@ pub struct ElectricalProfileSetIdQueryParam {
...
+                    TrainId::PacedTrainOccurrence(paced_train_occurrence) => {
+                        paced_train_occurrence_ids.push(paced_train_occurrence.clone())
+                    }
+                }
+            } else {
+                let train_id = train_id
+                    .parse::<i64>()
+                    .unwrap_or_else(|_| panic!("Failed to parse train_id '{}'", train_id));
+                train_schedule_ids.push(train_id);
+            }
```
flomonster: This case is weird. If core sends a train ID that does not match any train ID provided in input, then something is wrong.  In this case, I would throw an error, or at least log a warning and skip it.

See this comment: https://github.com/OpenRailAssociation/osrd/pull/11069#discussion_r1995170230

editoast/src/views/timetable.rs
```diff
@@ -424,34 +507,128 @@ async fn conflicts(
...
+
+    // Build train schedule train requirements
+    for (train, sim) in trains.into_iter().zip(train_simulations) {
         let (sim, _) = sim;
         let final_output = match sim {
             SimulationResponse::Success { final_output, .. } => final_output,
             _ => continue,
         };
+
+        map_string_to_id.insert(train.id.to_string(), TrainId::TrainSchedule(train.id));
```
flomonster: Nit, to avoid calling `to_string` twice.

```suggestion
        let key = train.id.to_string();
        map_string_to_id.insert(key.clone(),  TrainId::TrainSchedule(train.id));
```

flomonster (approved):
Thanks for this PR. LGTM once all comments are fixed.

SharglutDev (approved):
Lgtm for front (not tested)

editoast/src/core/conflict_detection.rs
```diff
@@ -103,6 +97,12 @@ pub struct Conflict {
     pub requirements: Vec<ConflictRequirement>,
 }
 
+#[derive(Debug, Clone)]
+pub enum TrainId {
+    TrainSchedule(i64),
+    PacedTrainOccurrence(PacedTrainOccurrenceId),
+}
+
```
leovalais, line 105: This definition seems misplaced as it's not used anywhere in `core`

editoast/src/views/timetable.rs
```diff
@@ -367,6 +378,80 @@ pub struct ElectricalProfileSetIdQueryParam {
...
+            end_time: conflict.end_time,
+            conflict_type: conflict.conflict_type,
+            requirements: conflict.requirements,
+        })
+    }
+}
+
 /// Retrieve the list of conflict of the timetable (invalid trains are ignored)
 #[utoipa::path(
     get, path = "",
```
leovalais, line 457: A comment about the excessive length of this function has been resolved but not addressed:

https://github.com/OpenRailAssociation/osrd/pull/11069#discussion_r1995200627
https://github.com/OpenRailAssociation/osrd/pull/11069#discussion_r1995291602

editoast/src/views/timetable.rs
```diff
@@ -367,6 +378,80 @@ pub struct ElectricalProfileSetIdQueryParam {
...
+}
+
+impl Conflict {
+    /// Converts the conflict data into a `ViewConflict` format.
+    ///
+    /// This function processes train schedule ids and paced trains generated ids ("paced_train_id#occurrence_id}")
+    ///  and maps them to either a `train_schedule_id` or a `paced_train_occurrence_id` based on the provided key mapping.
+    /// The train_id_map follows this structure:
+    /// - `train_id: (train_schedule_id, None)` if `train_id` corresponds to a train schedule ID.
+    /// - `train_id: (paced_train_id, occurrence_id)` if `train_id` corresponds to a generated paced train ID.
```
leovalais, line 409: * No definition named `ViewConflict`, better link types with `[TypeName]` to ensure doc consistency across renamings.
* "The train_id_map follows this structure": no it doesn't
* "This function processes train schedule ids and paced trains generated ids ("paced_train_id#occurrence_id}")": that's misleading and the id format isn't relevant here (the '#' logic isn't present anywhere in the function)

editoast/src/views/timetable.rs
```diff
@@ -367,6 +378,80 @@ pub struct ElectricalProfileSetIdQueryParam {
...
+    fn into_conflict_response(
+        conflict: CoreConflict,
+        train_id_map: HashMap<String, TrainId>,
```
leovalais, line 412: Do you really need a map? Can't you parse both `TrainId` variants directly?

editoast/src/views/timetable.rs
```diff
@@ -424,34 +513,124 @@ async fn conflicts(
...
+    for paced_train in paced_trains.clone() {
+        let occurrences =
+            (paced_train.duration.num_seconds() / paced_train.step.num_seconds()) as usize;
+        let step = paced_train.step;
+        let mut start_time = paced_train.start_time;
+
+        let mut all_occurrences = Vec::with_capacity(occurrences);
+
+        for _ in 0..occurrences {
+            let mut first_occurrence = paced_train.clone().into_first_occurrence();
```
leovalais, line 526: No need to do that at every iteration

editoast/src/views/timetable.rs
```diff
@@ -424,34 +513,124 @@ async fn conflicts(
...
+            first_occurrence.start_time = start_time;
```
leovalais, line 527:
```suggestion
occurrence.start_time = first_occurrence.start_time + i * step;
```

editoast/src/views/timetable.rs
```diff
@@ -424,34 +513,124 @@ async fn conflicts(
...
+            all_occurrences.push(first_occurrence);
+            start_time += step;
```
leovalais, line 529: I believe a `paced_train.into_occurrence(n: usize)` would help here, maybe even a `paced_train.iter_occurrences()`

editoast/src/views/timetable.rs
```diff
@@ -424,34 +513,124 @@ async fn conflicts(
...
+    // Build paced train requirements
+    for (train, (sim, _)) in paced_trains_ts.into_iter().zip(paced_train_simulations) {
+        let final_output = match sim {
+            SimulationResponse::Success { final_output, .. } => final_output,
+            _ => continue,
+        };
+
+        let entry = occurrences.entry(train.id).or_insert(0);
+        let occurrence_id = *entry;
+        let key = format!("{}#{}", train.id, occurrence_id);
```
leovalais, line 604: Better embed this logic in `TrainId`

editoast/src/views/timetable.rs
```diff
@@ -424,34 +513,124 @@ async fn conflicts(
...
+}
+
+fn build_conflict_core_request(
+    infra: Infra,
+    trains: Vec<TrainSchedule>,
+    train_simulations: Vec<(SimulationResponse, PathfindingResult)>,
+    paced_trains_ts: Vec<TrainSchedule>,
+    paced_train_simulations: Vec<(SimulationResponse, PathfindingResult)>,
+) -> (ConflictDetectionRequest, HashMap<String, TrainId>) {
+    let mut map_string_to_id: HashMap<String, TrainId> = HashMap::new();
```
leovalais, line 570: As said earlier, I think we can get rid of the map since all `TrainId`s are embedded in the request and will provided back in the response.

editoast/src/views/timetable.rs
```diff
@@ -424,34 +513,124 @@ async fn conflicts(
...
+        .map(|conflict| Conflict::into_conflict_response(conflict, map_string_to_id.clone()))
+        .collect();
+    Ok(Json(conflicts_response?))
+}
+
+fn build_conflict_core_request(
+    infra: Infra,
+    trains: Vec<TrainSchedule>,
+    train_simulations: Vec<(SimulationResponse, PathfindingResult)>,
+    paced_trains_ts: Vec<TrainSchedule>,
```
leovalais, line 567: Why not just provide the `PacedTrain` along with its occurrences' simulations? That'd avoid:

1. mixing two different objects (I'd even be partial towards having an `Occurrence` type distinct from `TrainSchedule`)
2. having the logic of reconstructing the occurrence list for each paced train

editoast/src/views/timetable.rs
```diff
@@ -424,34 +513,124 @@ async fn conflicts(
...
+        paced_trains_ts,
+        paced_train_simulations,
+    );
+
+    // 3. Call core
+    let conflict_detection_response = conflict_detection_request.fetch(&core_client).await?;
+    let conflicts = conflict_detection_response.conflicts;
+    let conflicts_response: Result<Vec<Conflict>> = conflicts
+        .into_iter()
+        .map(|conflict| Conflict::into_conflict_response(conflict, map_string_to_id.clone()))
```
leovalais, line 558: `from_core_response` would be more idiomatic if done this way

editoast/src/views/timetable.rs
```diff
@@ -367,6 +378,80 @@ pub struct ElectricalProfileSetIdQueryParam {
...
+                        paced_train_occurrence_ids.push(paced_train_occurrence.clone())
+                    }
+                }
+            } else {
+                let train_id = train_id
+                    .parse::<i64>()
+                    .map_err(|_| TimetableError::ParseError {
+                        train_id: train_id.clone(),
+                    })?;
+                train_schedule_ids.push(train_id);
```
leovalais, line 431: When does Core return a train ID not provided in the request?

Also nit (barely) but didn't we agree on panicking should Core return a faulty ID?

leovalais (commented):
This PR was merged while I was vacationing, I could only take a look now, sorry about the delay... I left a few comments about some things I found weird, and unaddressed but resolved discussions ⬇️

editoast/src/views/timetable.rs
leovalais: Seems unaddressed, unresolving

PR #11070, opened 2025-03-10
core: replace homemade bloom filter with guava's

eckter:
And a few related adjustments / cleanups.

Some pathfinding requests took way too long because we sometimes saturated the bloom filter. As a result, we'd iterate through the whole path at every single step to look for duplicates, which is painfully slow. We'd end up taking more than a minute, with even timeouts in more extreme cases.

1. Remove homemade bloom filter: it was simple enough, but guava's has more convenient init values
2. Said values have been tweaked to account for what I've empirically observed on a timetable import
3. Remove redundant duplicate checks for blocks, and related code (they'd have duplicated tracks anyway)
4. Add time measurement when reproducing requests, just because why not

I didn't run *extensive* benchmarks, but the requests that took more than a minutes are now <30s. Still way too slow in my opinion, but not as extreme as before. 

It also speeds up stdcm in the same way, the difference is just less noticeable there.

### DIFF
--- a/core/kt-osrd-utils/src/main/kotlin/fr/sncf/osrd/utils/AppendOnlySet.kt
+++ b/core/kt-osrd-utils/src/main/kotlin/fr/sncf/osrd/utils/AppendOnlySet.kt
@@ -1,5 +1,7 @@
 package fr.sncf.osrd.utils
 
+import com.google.common.hash.BloomFilter
+
 /**
  * Append-only set implementation. The internal structure is a linked list. Used to store data on
  * diverging paths while minimizing copies. See also `AppendOnlyLinkedList`.
@@ -11,7 +13,7 @@ class AppendOnlySet<T>(
 
     /** Add the given value to the set. O(1). */
     fun add(k: T) {
-        keyFilter.add(k)
+        keyFilter.put(k)
         list.add(k)
     }
 
@@ -25,12 +27,18 @@ class AppendOnlySet<T>(
      * end. Pre-filtered using a bloom filter.
      */
     fun contains(key: T): Boolean {
-        if (!keyFilter.mayContain(key)) return false
+        if (!keyFilter.mightContain(key)) return false
         return list.findLast { it == key } != null
     }
 }
 
 /** Returns a new empty set */
-fun <T> appendOnlySetOf(): AppendOnlySet<T> {
-    return AppendOnlySet(appendOnlyLinkedListOf(), emptyBloomFilter())
+inline fun <reified T> createAppendOnlySet(
+    expectedInsertions: Int,
+    falsePositiveRate: Double,
+): AppendOnlySet<T> {
+    return AppendOnlySet(
+        appendOnlyLinkedListOf(),
+        emptyBloomFilter(expectedInsertions, falsePositiveRate)
+    )
 }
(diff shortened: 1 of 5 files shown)


core/kt-osrd-utils/src/main/kotlin/fr/sncf/osrd/utils/AppendOnlySet.kt
```diff
@@ -25,12 +27,18 @@ class AppendOnlySet<T>(
...
     }
 }
 
 /** Returns a new empty set */
-fun <T> appendOnlySetOf(): AppendOnlySet<T> {
-    return AppendOnlySet(appendOnlyLinkedListOf(), emptyBloomFilter())
+inline fun <reified T> appendOnlySetOf(
+    expectedInsertions: Int = 1500,
+    falsePositiveRate: Double = 0.03,
+): AppendOnlySet<T> {
```
Khoyo: The naming is unfortunate. If `appendOnlySetOf` accept arguments, one would expect they are the elements of the created set, not that it always return an empty set.

Maybe this should just be a secondary constructor of AppendOnlySet instead?

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorer.kt
```diff
@@ -134,7 +137,7 @@ fun initInfraExplorer(
                 appendOnlyLinkedListOf(),
                 appendOnlyLinkedListOf(),
                 appendOnlyMapOf(),
-                appendOnlySetOf(),
+                appendOnlySetOf(maxExpectedTracks, 0.03),
```
Khoyo: I'm not sure why the defaults are repeated here?

core/kt-osrd-utils/src/main/kotlin/fr/sncf/osrd/utils/AppendOnlySet.kt
eckter: Good point, I'll rename the method

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorer.kt
eckter: Oh I intended to remove the default values (they were too "magic" without context imo)

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorer.kt
```diff
@@ -134,7 +137,7 @@ fun initInfraExplorer(
                 appendOnlyLinkedListOf(),
                 appendOnlyLinkedListOf(),
                 appendOnlyMapOf(),
-                appendOnlySetOf(),
+                createAppendOnlySet(maxExpectedTracks, 0.03),
```
bougue-pe: How about defining `falsePositiveRate` close to `maxExpectedTracks` to explicit the value's origin?

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorer.kt
```diff
@@ -359,7 +362,6 @@ private class InfraExplorerImpl(
      * visited.
      */
     private fun hasRepeatedElements(block: StaticIdx<Block>): Boolean {
-        if (blockRoutes.containsKey(block)) return true
```
bougue-pe, line 362: Did you bench perf when keeping this early return?
I'm curious about where the perf gain is (and if this really is better to remove that check).

eckter: Not this specifically, it's more about redundancy than performances. We check for track duplicates in the entirety of the block anyway.

And also to avoid bothering with how to init the bloom filter when creating the map.

bougue-pe (approved):
Thanks !

PR #11073, opened 2025-03-10
front: fix language list

emersion:
i18n.languages contains only the list of fallback langagues, ie. the list of languages used in case a translation doesn't exist in the currently selected language. This isn't what we want here: we want the full list of supported languages.

See this issue for more details:
https://github.com/i18next/i18next/issues/1068

Export our list of supported languages and use that instead.


Akctarus (approved):
lgtm !

kmer2016 (approved):
LGTM

PR #11100, opened 2025-03-11
editoast: add documentation about OpenFGA setup

woshilapin:


leovalais (approved):
Oops™️ Thanks!

editoast/README.md
```diff
@@ -14,13 +14,17 @@ For both tests or run:
...
 
 ## Steps
 
 ```sh
 # apply database migration
 $ cargo install diesel_cli --no-default-features --features postgres
 $ diesel migration run --locked-schema  # avoids bumping modification date (then rebuild)
+# prepare the database for OpenFGA
+docker exec -it osrd-postgres psql -d osrd -c "CREATE SCHEMA openfga;"
+docker exec -it osrd-postgres psql -d osrd -c "GRANT ALL PRIVILEGES ON SCHEMA openfga TO osrd;"
```
flomonster, line 27: This should not be required since it's part of the `init_db.sql`. We have an issue when restoring a backup but it's a bug.
So I don't think we should add these command lines here.

woshilapin: OK, I get that, but we had 3 or 4 different people impacted by this recently. Note that I didn't fully understand the problem you're mentioning, can I help?

flomonster: Yes sure, running:
`./scripts/load-backup.sh BACKUP_FILE` reset privileges on the `openfga` schema. I tried a lot of different things to avoid this behavior without success :shrug:

woshilapin: But can't we generate a new backup that contains these privileges? Or was that already attempted and it doesn't work?

flomonster: I tried and it does not work :disappointed:

woshilapin: From the documentation

> pg_dump only dumps a single database. To back up an entire cluster, or to back up global objects that are common to all databases in a cluster (such as roles and tablespaces), use [pg_dumpall](https://www.postgresql.org/docs/current/app-pg-dumpall.html).

I'm wondering if using [`pg_dumpall`](https://www.postgresql.org/docs/current/app-pg-dumpall.html) would solve the problem?

flomonster: Great idea. Do you want to try this?

bougue-pe: After discussing and testing it, the breaking issue is only happening when postgresql db was inited before the change of `init_db.sql` and no "hand script" was applied. So unless we want to be able to share in the dumps some data from openfga it doesn't look necessary.

I'm about to close this PR in favor of #11167 : let's react there if you think it deserves some more work.

bougue-pe:
superseded by #11167

PR #11101, opened 2025-03-11
editoast: do not ensure latest model when authorization disabled

woshilapin:
I would argue that all of these steps are not necessary when authorization is disabled, what do you think?

### DIFF
--- a/editoast/src/views/mod.rs
+++ b/editoast/src/views/mod.rs
@@ -592,7 +592,9 @@ impl AppState {
             }
         };
         tracing::info!(url = %config.openfga_config.url, "connected to OpenFGA");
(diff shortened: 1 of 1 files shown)


leovalais (approved):
LGTM, it'll also help to spot code paths requesting openfga when authentication is disabled more easily.

Note that the `if` should only apply to the `ensure_authorization_model` call. Otherwise we'd need to make the openfga client optionnal in the state, which brings a whole new set of problems (arguably worse imo 👀).

Note bis: no need to do that for the `TestApp` right now, I already done it in https://github.com/OpenRailAssociation/osrd/pull/10984/files#diff-81c2129f0eebb748f80acf28b466bdc71f6902caa5c610e08fc11f27cbe6e5ae

PR #11107, opened 2025-03-12
core: stdcm: fix offset type inconsistency

eckter:
We'd use an `Offset<Block>` to mean an offset from the start of the simulation on this block, it's now a real `Offset<Block>`.

The function itself is quite weird (probably artifacts from the old java->kt conversion), but I'll rewrite it in a different refactoring.


Erashin (approved):
Yep.

PR #11109, opened 2025-03-12
core: stdcm: reorganize step data to make them dynamic

### LINKED ISSUE #8974: refacto: stdcm: unify stop data

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/8974

Follow-up of https://github.com/OpenRailAssociation/osrd/pull/10989 (which started the wrong way).

May be reviewed commit by commit, but there's been a few changed from earlier commits that have been changed again in later ones. But they're mostly self-contained changes and each commit normally compiles and runs properly (though I've only run unit tests).

---

The main goal of this refactor was to remove the use of step data as a static thing, so that they can be changed dynamically when implementing overtakes. The main challenge was checking for `graph.steps` to test for, e.g., stop duration. A few data redundancy have been removed as well (like stop offsets in `InfraExplorerWithEnvelope`), but they weren't too tangled up.

The main change is the Introduction of a class that keeps track of the steps on the path, `StepTracker`. It includes both stops and non-stop waypoints, while keeping track of what's in the lookahead section. Eventually it may include stops that aren't part of the input.

This solves two main issues: we don't need to check for the input step list anymore, we can instead rely on `StepTracker`'s step list, which would include extra stops. We also don't need to manually keep track of waypoint indexes in edges/nodes (which again wouldn't match with a more dynamic stop list).

There's some more minor refactoring along the way.


core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorer.kt
```diff
@@ -276,10 +278,14 @@ private class InfraExplorerImpl(
...
+    override fun getStepTracker(): StepTracker {
+        return stepTracker
+    }
+
     /**
      * Updates `incrementalPath`, `routes`, `blocks` and returns true if route can be explored.
-     * Otherwise, it returns false and keeps the states as is. `blockRoutes` is updated to keep
-     * track of the route used for each block.
+     * Otherwise, it returns false and the instance is supposed to be dropped. `blockRoutes` is
+     * updated to keep track of the route used for each block.
```
eckter, line 292: Having to always keep the current instance in a consistent state even when it's meant to be discarded was a lot more painful here. We need to update the step tracker to properly handle the following edges.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMEdge.kt
```diff
@@ -55,36 +53,38 @@ data class STDCMEdge(
 
     /** Returns the node at the end of this edge */
     fun getEdgeEnd(graph: STDCMGraph): STDCMNode {
-        var newWaypointIndex = waypointIndex
+        // TODO: maybe integrate this logic to StepTracker?
```
eckter:
```suggestion
```

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/PostProcessingSimulation.kt
```diff
@@ -473,12 +473,6 @@ private fun handlePostProcessingConflict(
             postProcessingLogger.info("        offset $remainingDistance/${edge.length}")
             postProcessingLogger.info("        original time data: ${edge.timeData}")
             postProcessingLogger.info("        updated time data: $updatedTimeData")
-            postProcessingLogger.info(
-                "        original explorer stops: ${edge.infraExplorerWithNewEnvelope.getStops()}"
```
Erashin, line 477: Technically, you could still get those through the infra explorer in one format or another (if you wanna keep a list of train stops, you can use the `generateTrainStops`). Not the most important info to have here in the logs, just saying it in case you missed it.

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * The tricky part is that we need to keep track of which steps are seen on the path as a whole
+ * (including lookahead), and which have actually been reached by the simulations (excluding
+ * lookahead). Unless specified otherwise, fields and methods refer to the whole path.
+ */
+class StepTracker(private val inputSteps: List<STDCMStep>) {
+    // If copying this is too expensive, it may be changed to an AppendOnlyLinkedList.
+    // But the list should be small.
+    private val reachedSteps: MutableList<LocatedStep> = mutableListOf()
+    private val stepsIncludingLookahead: Int
```
Erashin: The variable name doesn't really reflect that this is a number. `stepsSizeIncludingLookahead`, `stepsNbIncludingLookahead` or anything else that'd clarify this.

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
```diff
@@ -0,0 +1,118 @@
...
+        get() = reachedSteps.size
+
+    var stepsExcludingLookahead: Int = 0
```
Erashin: The variable name doesn't really reflect that this is the number of simulated steps/index of the first lookahead step. `stepsSizeExcludingLookahead`, `stepsNbExcludingLookahead` or anything else to clarify this, especially since this variable is public and called in other parts of the code.

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /** True if the last step has been reached (including lookahead). */
+    fun hasReachedDestination(): Boolean {
+        return stepsIncludingLookahead == inputSteps.size
+    }
+
+    /** True if the last step has been reached, with full simulation and no lookahead. */
+    fun hasReachedDestinationExcludingLookahead(): Boolean {
+        return hasReachedDestination() && stepsIncludingLookahead == stepsExcludingLookahead
```
Erashin: It would be clearer if the condition was the same as the above method no?
```suggestion
        return hasReachedDestination() && stepsExcludingLookahead == inputSteps.size
```

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
```diff
@@ -0,0 +1,118 @@
...
+    // Used to compute path offsets
+    private var currentPathOffset: Offset<TravelledPath> = Offset.zero()
+
+    /** Returns all the steps that have been passed on the path, in order. */
+    fun getAllReachedSteps(): List<LocatedStep> {
+        return reachedSteps
+    }
+
+    /** True if the last step has been reached (including lookahead). */
+    fun hasReachedDestination(): Boolean {
```
Erashin: What about `hasSeenDestination()`? Would clarify that we check the final step is at minimum in the LOOKahead part :)
We should probably `assert(stepsIncludingLookahead <= inputSteps.size)`, but this might not be the most appropriate place to put it (we could check it when we add a step to reachedSteps).

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
```diff
@@ -0,0 +1,118 @@
...
+        return stepsIncludingLookahead == inputSteps.size
+    }
+
+    /** True if the last step has been reached, with full simulation and no lookahead. */
+    fun hasReachedDestinationExcludingLookahead(): Boolean {
```
Erashin: What about `hasReachedDestination`?
We should probably as well `assert(stepsExcludingLookahead <= inputSteps.size)`, but same as above, might be more appropriate elsewhere.

core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
```diff
@@ -99,15 +98,17 @@ data class BlockAvailability(
         var maximumDelayToStayAvailable = Double.POSITIVE_INFINITY
         var timeOfNextUnavailability = Double.POSITIVE_INFINITY
 
-        for (step in plannedSteps) {
+        val steps = infraExplorer.getStepTracker().getAllReachedSteps().reversed()
```
Erashin: `reversedSteps`? Anything to make sure people see that we're checking the availability of the steps going backwards here.

Erashin (commented):
This looks good.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMUtils.kt
```diff
@@ -10,28 +10,17 @@ import fr.sncf.osrd.utils.units.Length
 import fr.sncf.osrd.utils.units.Offset
 import java.util.*
 
-/** Returns the offset of the stops on the given block, starting at startOffset */
+/** Returns the offset of the next stop on the given block (if any), starting at startOffset */
 fun getStopOnBlock(
```
bougue-pe:
```suggestion
/** Returns the offset of the first stop (if any) on the current block */
fun getFirstStopOnCurrentBlock(
```

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
```diff
@@ -52,11 +53,25 @@ data class InfraExplorerWithEnvelopeImpl(
...
-        val withStops = EnvelopeStopWrapper(res, stops)
+        val withStops = EnvelopeStopWrapper(res, generateTrainStops())
         envelopeCache = SoftReference(withStops)
         return withStops
     }
 
+    private fun generateTrainStops(): List<TrainStop> {
+        val steps = getStepTracker().getAllReachedSteps()
+        val stopOffsets = steps.filter { it.originalStep.stop }.map { it.pathOffset }
+        val stopDurations = stopTimeData
```
bougue-pe: From the name, I'd expect
```suggestion
        val stopDurations = stopTimeData.map {it.currentDuration}
```
(or zip directly on `stopTimeData.currentDuration` and get rid of this line)

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
```diff
@@ -52,11 +53,25 @@ data class InfraExplorerWithEnvelopeImpl(
...
+        assert(stopDurations.size <= stopOffsets.size)
```
bougue-pe, line 68: This looks suspicious if lists' sizes are not equal.
Here is how I understand it (requires a comment at least, maybe a fn naming change IMO):
Only the last reached stops wouldn't have duration, so wouldn't be listed, and it's OK? (not sure when it would happen and why it would be OK).

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMEdgeBuilder.kt
```diff
@@ -140,15 +148,13 @@ internal constructor(
...
-                startOffset,
-                prevNode.waypointIndex
-            ) != null
-        if (!endAtStop) return null
-        return graph.getFirstStopAfterIndex(prevNode.waypointIndex)!!.duration!!
+        return prevNode.infraExplorer
+            .getStepTracker()
+            .getStepsInLookahead()
+            .filter { it.location.edge == infraExplorer.getCurrentBlock() }
+            .firstOrNull { it.originalStep.stop }
```
bougue-pe, line 158: I don't get what was done before this PR, but from the naming and the comment, shouldn't that be `lastOrNull`?

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPathfinding.kt
```diff
@@ -118,9 +117,7 @@ class STDCMPathfinding(
             ConstraintCombiner(initConstraints(fullInfra, listOf(rollingStock)).toMutableList())
 
         assert(steps.last().stop) { "The last stop is supposed to be an actual stop" }
-        val stops = steps.filter { it.stop }.map { it.locations }
-        assert(stops.isNotEmpty())
-        starts = getStartNodes(stops, listOf(constraints))
+        starts = getStartNodes(steps, listOf(constraints))
```
bougue-pe: Should we filter on `it.stop` here also?

core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
```diff
@@ -148,88 +149,47 @@ data class BlockAvailability(
...
+            (plannedTimingData.arrivalTime - plannedTimingData.arrivalTimeToleranceBefore).seconds -
+                internalMarginForSteps
+        val plannedMaxTimeAtStep =
+            (plannedTimingData.arrivalTime + plannedTimingData.arrivalTimeToleranceAfter).seconds +
+                internalMarginForSteps
+        if (plannedMinTimeAtStep > timeAtStep) {
+            // Train passes through planned timing data before it is available
+            return AvailabilityProperties(
+                max(plannedMinTimeAtStep - timeAtStep, 0.0),
+                stepOffsetOnPath,
```
bougue-pe: Not sure: why removing the conversion?
Did you want to use `step.pathOffset` instead here?

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorer.kt
```diff
@@ -298,28 +304,31 @@ private class InfraExplorerImpl(
             if (!seenFirstBlock) {
                 nBlocksToSkip++
             } else {
+                val startPath = !pathStarted
```
bougue-pe: nit:
```suggestion
                val startsPath = !pathAlreadyStarted
```

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorer.kt
```diff
@@ -298,28 +304,31 @@ private class InfraExplorerImpl(
...
+                val addRoute = block == routeBlocks.first() || startPath
```
bougue-pe:
```suggestion
                val addsRoute = block == routeBlocks.first() || startPath
```

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorer.kt
```diff
@@ -298,28 +304,31 @@ private class InfraExplorerImpl(
...
+                val travelledPathBeginBlockOffset =
```
bougue-pe: If I understood correctly, it would clearer IMO to have explicit type:
```suggestion
                val travelledPathBegin: Offset<Block> =
```

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorer.kt
bougue-pe: The only point I can think of is saving allocations (so reusing). Which may be good in that context...
But I agree that unless someone (@Khoyo ?) has memory/plan in that direction, it's a YAGNI.

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
```diff
@@ -0,0 +1,118 @@
...
+import fr.sncf.osrd.utils.units.Offset
+
+/**
+ * Component of `InfraExplorer` in charge of keeping track of anything related to the input steps.
+ * It keeps tracks of which steps we've passed and where, how many steps have been reached
+ * (specifically when we've reached the destination), and which steps are present in a block range.
+ *
+ * The tricky part is that we need to keep track of which steps are seen on the path as a whole
+ * (including lookahead), and which have actually been reached by the simulations (excluding
+ * lookahead). Unless specified otherwise, fields and methods refer to the whole path.
```
bougue-pe:
```suggestion
 * lookahead).
 * Unless specified otherwise, fields and methods refer to the whole path (including lookahead).
```

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
bougue-pe: I'd also remove the `IncludingLookahead` for consistency and to match the class comment.

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * Go through a block range and register every step in the given range (boundaries included).
+     * New steps are first part of the "lookahead" section, and only move to the "visited" section
+     * upon `moveForward` calls.
+     */
+    fun exploreBlockRange(
+        block: BlockId,
+        rangeStart: Offset<Block>,
```
bougue-pe, line 57: Does that contain the `max(traveledPathBegin, Offset.zero)`? (same for `rangeEnd`)
If so (from the use, I guess), is it possible to be more explicit in naming or in comment? (I imagined exploring part of the block for other reasons, like splitting at every stop, which doesn't seem like a correct use).

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
```diff
@@ -0,0 +1,118 @@
...
+        rangeEnd: Offset<Block>, // No default value as we need the infra to know the block len
+    ): List<LocatedStep> {
+        val res = mutableListOf<LocatedStep>()
+        var mutRangeStart = rangeStart
+        while (true) {
```
bougue-pe: How about iterating over the sorted steps of the range? It would sort only once, remove the `while(true)` red-flag, and the code could be clearer in the end.

`mutRangeStart` broke my brain. Note to self: review that part if that doesn't change.

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
```diff
@@ -0,0 +1,118 @@
...
+            mutRangeStart = location.offset
+        }
+        currentPathOffset += (rangeEnd - mutRangeStart)
+        return res
+    }
+
+    /** Integrate a part of the lookahead into the "actually visited" steps. */
+    fun moveForward(block: BlockId, start: Offset<Block>, end: Offset<Block>) {
+        while (true) {
+            val nextStepLoc = reachedSteps.getOrNull(stepsExcludingLookahead)?.location ?: return
```
bougue-pe: * Not fan of the `while(true)`, I'd probably opt for a `var nextStepLoc` to be tested and updated in the while.

* That said, as `return` and `break` have the same role here, I'd stay to the same for both cases (`break` probably).

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
eckter: I'm not really sure I understand your comment (especially the first question).

The intent is indeed to allow the exploration of a part of a block only. For this method it only happens at the start of the path, but in the one below we do split at every stop.

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
eckter: I replaced the while true with a more usual sublist iteration, but we do have to check everything step-by-step to handle ordering properly.

I've commented `mutRangeStart`, that was indeed a little obscure.

edit: [here](https://github.com/OpenRailAssociation/osrd/pull/11109/commits/09b5967ef81f6ab0b026d5dbe1e0201d2b6a95ed#diff-cf16bc37931f8039dd7a63ad658eaf552824657169fb8595f9d695f88e9a151aR61-R67)

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
eckter: I've turned that loop into a more functional approach entirely, in the end we're only counting matching steps here

core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
eckter: Oh.

That was a mistake. Very nice catch, thanks.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMEdgeBuilder.kt
eckter: We build the edge up to the first encountered stop. I'll clarify the comment. 

[here](https://github.com/OpenRailAssociation/osrd/pull/11109/commits/bd501be5875f4820c599fc7e6917a0067bad058d#diff-cc0a3fa7a96ec34e14a3708937b639c2b23aacc0a2f4d26c6bf85c38480b85adR149-R152)

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPathfinding.kt
eckter: For the start location we consider that the train starts at speed=0, it shouldn't really matter if it's marked as a stop. 

Actually the previous behavior was confusing: in `getStartNodes(stops, ...)`, the `stops` was only meant to init the infra explorers. They're in fact not used anymore and I'll remove it from the function input. The actual start is fetched from `this.steps[0]`.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMUtils.kt
eckter: It is actually "next", we're only considering the lookahead. There may be stops that have already been simulated.

But I've reworded the rest, including function name.

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
eckter: Stop offsets are listed even if they're in the lookahead. 

A "safer" way would be to only iterate on the steps excluding lookahead, but even then it wouldn't necessarily match. The moment when we "move forward" doesn't *precisely* match the moment when when add the stop duration. Sometimes when evaluating the new edge we want its envelope with or without the stop.

eckter (commented):
Very nice reviews, thanks :pray:

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
```diff
@@ -0,0 +1,124 @@
...
+ * The tricky part is that we need to keep track of which steps are seen on the path as a whole
+ * (including lookahead), and which have actually been reached by the simulations (excluding
+ * lookahead).
+ *
+ * Unless specified otherwise, fields and methods refer to the whole path (including lookahead).
+ */
+class StepTracker(private val inputSteps: List<STDCMStep>) {
+    // If copying this is too expensive, it may be changed to an AppendOnlyLinkedList.
+    // But the list should be small.
+    private val reachedSteps: MutableList<LocatedStep> = mutableListOf()
```
bougue-pe: Then for consistency (ditto: getAll`Seen`Steps):
```suggestion
    private val seenSteps: MutableList<LocatedStep> = mutableListOf()
```

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
```diff
@@ -0,0 +1,124 @@
...
+     */
+    fun exploreBlockRange(
+        block: BlockId,
+        rangeStart: Offset<Block>,
+        rangeEnd: Offset<Block>, // No default value as we need the infra to know the block len
+    ): List<LocatedStep> {
+        val res = mutableListOf<LocatedStep>()
+
+        // Start of the range to explore. As we pass steps, we drop the range section before the
+        // visited step (hence mutable). This ensures we don't accep out-of-order steps when they're
```
bougue-pe: nit:
```suggestion
        // visited step (hence mutable). This ensures we don't accept out-of-order steps when they're
```

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
```diff
@@ -0,0 +1,124 @@
...
+        }
+        currentPathOffset += (rangeEnd - mutRangeStart)
+        return res
+    }
+
+    /** Integrate a part of the lookahead into the "actually visited" steps. */
+    fun moveForward(block: BlockId, start: Offset<Block>, end: Offset<Block>) {
+        nStepsExcludingLookahead +=
+            reachedSteps
+                .subList(nStepsExcludingLookahead, reachedSteps.size)
```
bougue-pe: Nice refactor, thanks!
nit:
```suggestion
            reachedSteps
                .drop(nStepsExcludingLookahead)
```

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMEdgeBuilder.kt
```diff
@@ -138,17 +146,18 @@ internal constructor(
         )
     }
 
-    /** Returns the stop duration at the end of the edge being built, or null if there's no stop */
+    /**
+     * Returns the stop duration at the end of the edge being built, or null if there's no stop. The
+     * new edge go up to the first encountered stop on the block that hasn't been simulated yet.
```
bougue-pe: nit:
```suggestion
     * new edge goes up to the first encountered stop on the block that hasn't been simulated yet.
```

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMEdgeBuilder.kt
bougue-pe: Thanks!
nit: How about renaming getEndStopDuration to get`Next`StopDuration (or get`EndOfEdge`StopDuration) then?
(feel free to resolve if none suits you).

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
bougue-pe: Thanks for the explanation 🙏 
I can think of:
* rename fn name to generate`Reached`TrainStops
* comment the fn, something like `match lists of stop offsets and stop durations (sizes may differ, but both are complete from the start of the path)`
Feel free to resolve anyway.

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/PostProcessingSimulation.kt
eckter: Fixed [here](https://github.com/OpenRailAssociation/osrd/pull/11109/commits/09b5967ef81f6ab0b026d5dbe1e0201d2b6a95ed#diff-1e18f89691d0f31e92ec595b5a7919de14273256b83f2e1a18fa94b46ff97225R476-R481)

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
eckter: Fixed in https://github.com/OpenRailAssociation/osrd/pull/11109/commits/09b5967ef81f6ab0b026d5dbe1e0201d2b6a95ed

I adjusted several variable names in one go for this class

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
eckter: Fixed [here](https://github.com/OpenRailAssociation/osrd/pull/11109/commits/09b5967ef81f6ab0b026d5dbe1e0201d2b6a95ed#diff-cf16bc37931f8039dd7a63ad658eaf552824657169fb8595f9d695f88e9a151aR45-R47)

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
eckter: Fixed [here](https://github.com/OpenRailAssociation/osrd/pull/11109/commits/09b5967ef81f6ab0b026d5dbe1e0201d2b6a95ed#diff-cf16bc37931f8039dd7a63ad658eaf552824657169fb8595f9d695f88e9a151aR39-R41)

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
eckter: [Fixed](https://github.com/OpenRailAssociation/osrd/pull/11109/commits/09b5967ef81f6ab0b026d5dbe1e0201d2b6a95ed#diff-cf16bc37931f8039dd7a63ad658eaf552824657169fb8595f9d695f88e9a151aR39-R41)

core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
eckter: [Fixed](https://github.com/OpenRailAssociation/osrd/pull/11109/commits/09b5967ef81f6ab0b026d5dbe1e0201d2b6a95ed#diff-4b129313418f0137d7eb3d01ed0c7c1c4a22942bba84e08c266044e487338adaR101-R103)

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
bougue-pe: Oh, I get it after your refactor and comment thanks for this.

What I had trouble understanding is that `rangeStart` (in the current block) MUST match (represent the same point) the `currentPathOffset` (in the traveled path, encompassing current block).

It's clearer already, I resolve this (and open another suggestion now that I understood it).

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
```diff
@@ -0,0 +1,124 @@
...
+                    location,
+                    step,
+                )
+            res.add(newStep)
+            reachedSteps.add(newStep)
+            currentPathOffset += (location.offset - mutRangeStart)
+            mutRangeStart = location.offset
+        }
+        currentPathOffset += (rangeEnd - mutRangeStart)
+        return res
```
bougue-pe: Not sure how to best enforce that `rangeStart` and `currentPathOffset` represent the same point: a fn comment is the minimal I can think of.
It's probably more of a design question to "unite" those (currently, there is room for misuse of this method).

Still, I can suggest adding:
* `val currentBlockStart: Offset<TraveledPath> = currentPathOffset - rangeStart`
* `val currentPathBlockOffset: Offset<Block> = currentPathOffset - currentBlockStart`

The following matches better my understanding of what's done.

```suggestion
        val currentBlockStart: Offset<TraveledPath> = currentPathOffset - rangeStart

        for (step in inputSteps.drop(nSeenSteps)) {
            val currentPathBlockOffset: Offset<Block> = currentPathOffset - currentBlockStart
            val location =
                step.locations
                    .filter { it.edge == block }
                    .filter { it.offset in currentPathBlockOffset..rangeEnd }
                    .minByOrNull { it.offset } ?: break
            currentPathOffset = currentBlockStart + location.offset
            val newStep =
                LocatedStep(
                    currentPathOffset,
                    location,
                    step,
                )
            res.add(newStep)
            reachedSteps.add(newStep)
        }
        currentPathOffset = currentBlockStart + rangeEnd
        return res
```

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorer.kt
```diff
@@ -302,47 +308,51 @@ private class InfraExplorerImpl(
             if (!seenFirstBlock) {
                 nBlocksToSkip++
             } else {
+                val startPath = !pathAlreadyStarted
```
bougue-pe: nit:
```suggestion
                val startsPath = !pathAlreadyStarted
```

Erashin (approved):
Works for me, modulo PE's other comments.
Be careful before merging, there's a fixup commit not linked to anything ;)

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
eckter: Fixed [here](https://github.com/OpenRailAssociation/osrd/pull/11109/commits/96c488505d211dfea053b75de4a04188d49bf42f#diff-d71ad00e98765b48457612edb0c3dd62b384bde07fe1cb131b3bbce33fcaf180R61-R68) (both rename and comment)

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMEdgeBuilder.kt
eckter: Fixed [here](https://github.com/OpenRailAssociation/osrd/pull/11109/commits/96c488505d211dfea053b75de4a04188d49bf42f#diff-cc0a3fa7a96ec34e14a3708937b639c2b23aacc0a2f4d26c6bf85c38480b85ad)

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
eckter: Fixed [here](https://github.com/OpenRailAssociation/osrd/pull/11109/commits/96c488505d211dfea053b75de4a04188d49bf42f#diff-cf16bc37931f8039dd7a63ad658eaf552824657169fb8595f9d695f88e9a151a)

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
eckter: Fixed

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
eckter: Fixed here and for other `.subList` in the file

core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMEdgeBuilder.kt
eckter: Fixed

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
eckter: Fixed in [this commit](https://github.com/OpenRailAssociation/osrd/pull/11109/commits/80ce0a3ca1110ef58ad7ceff2ca8f267ebb2f3ec)

I've made a few changes, especially for offset conversions, but it's mostly the same

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorer.kt
eckter: Fixed

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/StepTracker.kt
```diff
@@ -0,0 +1,120 @@
...
+        get() = seenSteps.size
+
+    var nStepsExcludingLookahead: Int = 0
+        private set
+
+    // Used to compute path offsets
+    private var currentPathOffset: Offset<TravelledPath> = Offset.zero()
+
+    /** Returns all the steps that have been passed on the path, in order. */
+    fun getAllReachedSteps(): List<LocatedStep> {
```
bougue-pe: Not sure about the "All":
```suggestion
    fun getAllSeenSteps(): List<LocatedStep> {
```

bougue-pe (approved):
Thanks for the work!
Good luck for the merge of fixups, this (and explanations 🙏) really helped review.

bougue-pe (approved):

PR #11115, opened 2025-03-13
scripts: kill openfga when restoring a db backup

eckter:
Otherwise we get an error like "DB is in use, can't load backup".

### DIFF
--- a/scripts/cleanup-db.sh
+++ b/scripts/cleanup-db.sh
@@ -19,6 +19,7 @@ root_path=$(realpath "$(dirname "$0")/..")
 # These variables are necessary to load the infra on the correct instance (the pr-infra or the dev one)
 OSRD_POSTGRES="osrd-postgres"
 OSRD_EDITOAST="osrd-editoast"
+OSRD_OPENFGA="osrd-openfga"
 OSRD_VALKEY="osrd-valkey"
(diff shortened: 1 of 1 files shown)


woshilapin (approved):
Makes sense.

flomonster (approved):
LGTM

scripts/cleanup-db.sh
```diff
@@ -19,6 +19,7 @@ root_path=$(realpath "$(dirname "$0")/..")
 # These variables are necessary to load the infra on the correct instance (the pr-infra or the dev one)
 OSRD_POSTGRES="osrd-postgres"
 OSRD_EDITOAST="osrd-editoast"
+OSRD_OPENFGA="osrd-openfga"
```
Khoyo, line 22: You need to adjust it in the PR_TEST case below

Khoyo (changes requested):

scripts/cleanup-db.sh
woshilapin: Or maybe, you don't do anything and let https://github.com/OpenRailAssociation/osrd/pull/11167 go through 🤷

eckter:
Closing: redundant with https://github.com/OpenRailAssociation/osrd/pull/11167

PR #11120, opened 2025-03-13
core: convert envelope files to kotlin

eckter:
Part of https://github.com/OpenRailAssociation/osrd/pull/10991, just for an easier way to review and comment this


core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.java
```diff
@@ -1,652 +1 @@
-package fr.sncf.osrd.envelope.part;
```
eckter, line 1: Not deleted

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
```diff
@@ -0,0 +1,677 @@
...
+
+    // endregion
+    // region CONSTRUCTORS
+    /** Creates an EnvelopePart */
+    init {
+        runSanityChecks()
+    }
+
+    /** Creates an EnvelopePart */
+    @ExcludeFromGeneratedCodeCoverage
```
eckter, line 66:
```suggestion
```

Doesn't sound relevant anymore

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
```diff
@@ -0,0 +1,677 @@
...
+        speeds: DoubleArray,
+        timeDeltas: DoubleArray
+    ) : this(makeAttrs(attrs), positions, speeds, timeDeltas)
+
+    /** Return the given metadata attribute */
+    @Suppress("UNCHECKED_CAST")
+    fun <T : SelfTypeHolder?> getAttr(attrType: Class<T>): T? {
+        return attrs[attrType] as T?
+    }
+
```
eckter, line 79: Technically we could change it into this, type parameters aren't really mandatory anymore in KT. It makes it more concise and type safe. Buuuut it may mess up java call sites, and it's generally not a trivial change. (note: we'd need to make `attrs` public here for the inline)

(the same applies for other attr methods)

```suggestion
    /** Return the given metadata attribute */
    inline fun <reified T : SelfTypeHolder> getAttr(): T? {
        return attrs[T::class.java] as T?
    }
```

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
```diff
@@ -0,0 +1,677 @@
...
+        attrs: Iterable<SelfTypeHolder>,
+        positions: DoubleArray,
+        speeds: DoubleArray,
+        timeDeltas: DoubleArray
+    ) : this(makeAttrs(attrs), positions, speeds, timeDeltas)
+
+    /** Return the given metadata attribute */
+    @Suppress("UNCHECKED_CAST")
+    fun <T : SelfTypeHolder?> getAttr(attrType: Class<T>): T? {
+        return attrs[attrType] as T?
```
eckter, line 77: Separate comment because this one's more important IMO

```suggestion
    fun <T : SelfTypeHolder?> getAttr(attrType: Class<T>): T {
        return attrs[attrType] as T
```

An optional return implies that there's no error if the value isn't in the map. Which could be a possibility, but that's not how it's implemented.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
```diff
@@ -0,0 +1,677 @@
...
+    val maxSpeed: Double
+        /** Returns the maximum speed of the envelope part */
+        get() {
+            if (!java.lang.Double.isNaN(maxSpeedCache)) return maxSpeedCache
+
+            var maxSpeed = Double.NEGATIVE_INFINITY
+            for (speed in speeds) if (speed > maxSpeed) maxSpeed = speed
+            this.maxSpeedCache = maxSpeed
+            return maxSpeed
+        }
```
eckter, line 148:
```suggestion
    /** Returns the maximum speed of the envelope part */
    fun getMaxSpeed(): Double {
        if (!java.lang.Double.isNaN(maxSpeedCache)) return maxSpeedCache

        var maxSpeed = Double.NEGATIVE_INFINITY
        for (speed in speeds) if (speed > maxSpeed) maxSpeed = speed
        this.maxSpeedCache = maxSpeed
        return maxSpeed
    }
```

Same for other getters that aren't one-liners

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
```diff
@@ -0,0 +1,677 @@
...
+
+    val minSpeed: Double
+        /** Returns the minimum speed of the envelope part */
+        get() {
+            if (!java.lang.Double.isNaN(minSpeedCache)) return minSpeedCache
```
eckter, line 153: Could use an import (doing it in an IDE would do it for the whole file)

```suggestion
            if (!jisNaN(minSpeedCache)) return minSpeedCache
```

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
```diff
@@ -0,0 +1,677 @@
...
+        return positions.size - 1
+    }
+
+    val maxSpeed: Double
+        /** Returns the maximum speed of the envelope part */
+        get() {
+            if (!java.lang.Double.isNaN(maxSpeedCache)) return maxSpeedCache
+
+            var maxSpeed = Double.NEGATIVE_INFINITY
+            for (speed in speeds) if (speed > maxSpeed) maxSpeed = speed
```
eckter, line 145:
```suggestion
            val maxSpeed = speeds.maxOrNull() ?: Double.NEGATIVE_INFINITY
```

Maybe I'm being too nitpicky here? It's just nice to have those tools available. Same kind of change can be made in different places, but I'll just comment it once (in case you'd prefer to ignore this)

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
```diff
@@ -0,0 +1,677 @@
...
+
+    val beginPos: Double
+        get() = getBeginPos(0)
+
+    fun getEndPos(stepIndex: Int): Double {
+        return positions[stepIndex + 1]
+    }
+
+    val endPos: Double
+        get() = getEndPos(stepCount() - 1)
```
eckter, line 173: There's a sneaky default parameter here that we lost because of the getter syntax

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
```diff
@@ -0,0 +1,677 @@
...
+
+    val beginSpeed: Double
+        get() = getBeginSpeed(0)
+
+    fun getEndSpeed(stepIndex: Int): Double {
+        return speeds[stepIndex + 1]
+    }
+
+    val endSpeed: Double
+        get() = getEndSpeed(stepCount() - 1)
```
eckter, line 190: Same (default parameters)

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
```diff
@@ -0,0 +1,677 @@
...
+     */
+    fun interpolateTotalTime(position: Double): Double {
+        return (interpolateTotalTimeUS(position).toDouble()) / 1000000
+    }
+
+    /**
+     * Given a speed return a position. The envelopePart must be bijective in order for this method
+     * to work
+     */
+    fun interpolatePosition(speed: Double): Double? {
```
eckter, line 348: It's weird that we return a nullable, considering we assert that the inputs would give a valid solution

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
```diff
@@ -0,0 +1,677 @@
...
+        var newBeginPos = beginPosition
+        var newEndPos = endPosition
+        var beginIndex = 0
+        if (newBeginPos <= beginPos) newBeginPos = Double.NEGATIVE_INFINITY
+        if (newBeginPos != Double.NEGATIVE_INFINITY) beginIndex = findRight(newBeginPos)
+        var endIndex = stepCount() - 1
+        if (newEndPos >= endPos) newEndPos = Double.POSITIVE_INFINITY
+        if (newEndPos != Double.POSITIVE_INFINITY) endIndex = findLeft(newEndPos)
+        return slice(beginIndex, newBeginPos, endIndex, newEndPos)
+    }
```
eckter, line 441: Could really use default values to avoid that duplication, but the parameter ordering isn't the most convenient for an easy change

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
```diff
@@ -0,0 +1,677 @@
...
+    ): EnvelopePart? {
+        return slice(
+            beginStepIndex,
+            beginPosition,
+            Double.NaN,
+            endStepIndex,
+            endPosition,
+            Double.NaN
+        )
+    }
```
eckter, line 466: Same, default parameters would be much nicer here

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/Envelope.kt
```diff
@@ -0,0 +1,384 @@
...
+    fun get(i: Int): EnvelopePart {
+        return parts[i]
+    }
+
+    override fun getBeginPos(): Double {
+        return parts[0].beginPos
+    }
+
+    override fun getEndPos(): Double {
+        return parts[parts.size - 1].endPos
```
eckter, line 90:
```suggestion
        return parts.last().endPos
```

Very nitpick.

Okay I should probably zoom out and stop with such extreme details.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
eckter: Wait wait wait

I forgot that the `[]` operator returns null when it's not in the map. My bad.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
Erashin: Did it for minSpeed as well.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/part/EnvelopePart.kt
Erashin: nah this fucks everything up in Java classes.

Erashin:
Yeah ok ngl I only took a couple details into account. The rest kinda fuck the java part up, and ngl, this ain't my code, this is just a kotlin reformat. We discussed it with @eckter we agree it's fine as is.
Taken into account in the origin pr.

core/envelope-sim/src/main/java/fr/sncf/osrd/envelope/Envelope.kt
```diff
@@ -0,0 +1,384 @@
...
+import fr.sncf.osrd.reporting.exceptions.ErrorType
+import fr.sncf.osrd.reporting.exceptions.OSRDError
+import java.util.*
+import java.util.stream.Stream
+
+@SuppressFBWarnings("URF_UNREAD_PUBLIC_OR_PROTECTED_FIELD")
+class Envelope private constructor(parts: Array<EnvelopePart>) :
+    Iterable<EnvelopePart>, SearchableEnvelope, EnvelopeInterpolate {
+    private val parts: Array<EnvelopePart>
+    @JvmField val continuous: Boolean
```
Khoyo, line 16: Is the `@JvmField` (and those below) on purpose?

Erashin: Yep, the java files accessing this property wouldn't work if not.

BTW, for context, this pr will be closed, its code is in 2 clean commits in another PR, and the fixes have been put there.

Khoyo: > Yep, the java files accessing this property wouldn't work if not.

The fact that those were *changed* from using a getter to direct access is exactly why I ask this :p Without this annotation, no change would have been necessary.

Erashin: Ah didn't get the nuance. This was done by the java to kotlin intellij converter.

PR #11126, opened 2025-03-13
front: fix display invalid reason in train schedule item

### LINKED ISSUE #11125: front: invalid reasons are not properly displayed in the train schedule item
### What happened?

Image

Invalid reasons in TrainScheduleItem should be displayed in the same way as in PacedTrainItem

1. Go to scenario page 
2. Create a train schedule invalid

Recette (SNCF)

Chrome

dev

Uriel-Sautron:
close #11125


theocrsb (approved):
LGTM and tested

kmer2016 (approved):
LGTM!

PR #11131, opened 2025-03-14
Enable permissions on infra

### LINKED ISSUE #10580: permission: conditionner l'affichage en fonction des privilèges

### LINKED ISSUE #10582: permission: Implémentation de la modale de partage des privilèges sur l'infra
voir le commentaire plus bas

SarahBellaha:
Closes #10580 
Closes #10582 

## Description

This PR adds advanced permission management for infrastructures, allowing detailed control of user actions based on their permissions.

[Mock-up](https://www.sketch.com/s/2fb064cd-c483-46ac-bd30-72705fd63e75/v/dPrOeZ/p/A0AF0F86-C222-422D-A4C0-36D538E779C4/canvas#Inspect)

### Added Features:

- **Infrastructure Permissions Management:**
  Allows authorized users to manage and view permissions of other users for each infrastructure.

- **Action Restrictions in Infrastructure Editor:**
  Prevents unauthorized users from performing actions such as saving or deleting infrastructure elements.

- **Controlled Access to Infrastructure Management Actions:**
  Limits actions such as renaming, duplicating, locking/unlocking, deleting, or downloading infrastructures based on the user's permissions.

### New Component:

- **`InfraSelectorGrantsManager`:** A component designed for visually managing user permissions directly in the infrastructure list. 

### New Custom Hooks:

- **`useCheckPrivileges`:** Checks if the current user has privileges to perform specific actions.

- **`useProtectedAction`:** Manages logic for protected actions, displaying error messages when users attempt to save modifications or delete infrastructure elements without required permissions (writer or owner).

- **`useResourcesGrants`:** Provides the user grants on a list of resources


front/src/modules/infra/components/InfraSelector/InfraSelectorModalBodyStandard.tsx
```diff
@@ -49,6 +54,20 @@ export default function InfraSelectorModalBodyStandard({
...
+    mockedEditoastApi.endpoints.getGrantsByResourceType.useQuery('infra');
+
+  const payload = useMemo(() => ({ infra: infrasList.map((infra) => infra.id) }), [infrasList]);
+
+  const userResourcesGrants = useResourcesGrants(payload);
+
+  const userGrantByInfra = useMemo(
+    () => keyBy(userResourcesGrants?.infra, 'id'),
+    [userResourcesGrants]
+  );
```
clarani: If I understand correctly, the grants of the current user are re-fetched each time the infras are filtered :/

I think it would be better to handle this only when the infras are fetched when the component is mounted.

front/src/modules/infra/components/InfraSelector/InfraSelectorModalBodyStandard.tsx
```diff
@@ -49,6 +54,20 @@ export default function InfraSelectorModalBodyStandard({
...
+  const { data: subjectsList } = mockedEditoastApi.endpoints.getSubjects.useQuery();
+
+  const { data: resourceGrants } =
+    mockedEditoastApi.endpoints.getGrantsByResourceType.useQuery('infra');
+
+  const payload = useMemo(() => ({ infra: infrasList.map((infra) => infra.id) }), [infrasList]);
+
+  const userResourcesGrants = useResourcesGrants(payload);
+
+  const userGrantByInfra = useMemo(
```
clarani: The type here is Map<number (infraId), { id: number, grant: Grant }>. It could be simpler, only Map<number (infraId), Grant>

front/src/modules/infra/components/InfraSelector/InfraSelectorModalBodyStandard.tsx
```diff
@@ -49,6 +54,20 @@ export default function InfraSelectorModalBodyStandard({
   const { closeModal } = useContext(ModalContext);
   const navigate = useNavigate();
 
+  const { data: subjectsList } = mockedEditoastApi.endpoints.getSubjects.useQuery();
```
clarani: Hum I think the subject are only used to get the users. Can you create a `users` const to extract directly the users from `subjectsList` ?

front/src/modules/infra/components/InfraSelector/InfraSelectorGrantsManager.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  useEffect(() => {
+    const getUserGrantsByInfraId = async () => {
+      const subjectsIdList = subjectList.map((subject) => subject.id);
+      try {
+        await postUsersGrantsByResourceId({
+          subjects_id: subjectsIdList,
+          resource_type: 'infra',
+          resource_id: infraId,
+        });
```
clarani: If I understand correctly, this endpoint is used to fetch all the grants of the subjects list (users ids) for a specific infra. Would it be possible to have the same endpoint which will return the grant of the users on the infras at once ?

clarani (commented):
Thanks for the PR :pray:

I was just thinking, maybe it would be easier to read if all the data for the grants was fetched and pre-processed at the same level (in InfraSelectorModal), do you think it would be possible ?

Also, concerning the infra editor, why not simply disable the forms for users who can't write the infras, as if the infra was locked ?

front/src/styles/scss/applications/operationalStudies/_scenario.scss
SharglutDev, line 1: There are still some `ambiant` syntax in this file

front/src/styles/scss/applications/operationalStudies/_scenario.scss
```diff
@@ -729,6 +729,10 @@
         }
 
         .filter-panel {
+          padding: 7px 20px 23px 7px;
+          background-color: var(--ambientB5);
+          border-top: solid 1px var(--white100);
+          border-bottom: solid 1px rgba(0, 0, 0, 0.25);
```
SharglutDev: Why do we have this ?

front/public/locales/en/infraManagement.json
```diff
@@ -25,6 +27,12 @@
   },
   "goToEditionMode": "Edit",
   "goToStandardMode": "Back to the selection",
+  "grants": {
+    "edit": "write",
+    "full": "full access",
+    "none": "no access",
+    "read": "read"
```
SharglutDev, line 34: In the [mockup](https://www.sketch.com/s/2fb064cd-c483-46ac-bd30-72705fd63e75/v/dPrOeZ/p/A0AF0F86-C222-422D-A4C0-36D538E779C4/canvas#Inspect) you linked, seems like the privileges are `none`, `read`, `edit` and `full`.

front/src/modules/infra/styles/_infraSelector.scss
SharglutDev, line 1: The new card style from this mockup (https://www.sketch.com/s/2fb064cd-c483-46ac-bd30-72705fd63e75/v/dPrOeZ/p/F5DF6E8F-B190-4F9A-99D4-962872914E64/canvas) will be done later ? It is referents in comments of the mpockup you linked

front/src/modules/infra/components/InfraSelector/InfraSelectorGrantsManager.tsx
```diff
@@ -0,0 +1,133 @@
...
+type InfraSelectorGrantsManagerProps = {
+  infraId: number;
+  userGrant: string;
+  resourceGrants?: {
+    [grant: string]: string[];
+  };
+  subjectList?: GetSubjectsResponse;
+};
+
+export default function InfraSelectorGrantsManager({
```
SharglutDev: nit: I know the old infra selector components don't matches this syntax but would you mind to change these react componenets for `const` instead of `function` ?

front/src/modules/infra/components/InfraSelector/InfraSelectorGrantsManager.tsx
```diff
@@ -0,0 +1,133 @@
...
+                  getOptionValue={(option) => option.value}
+                  onChange={(option) => updateUserInfraGrant(id, option?.value)}
+                  {...generateGrantSelectProps({
+                    resourceGrants,
+                    subjectGrant: grant,
+                    connectedUserGrant: userGrant,
+                    t,
+                  })}
+                />
+              </span>
```
SharglutDev, line 92: We don't add the info button yet ?

Capture d’écran 2025-03-24 à 15 36 30

front/src/modules/infra/styles/_infraSelector.scss
```diff
@@ -127,6 +127,65 @@
       line-height: 20px;
       font-weight: 500;
     }
+
+    .infra-selector-grants-manager {
+      margin: 0 16px 8px;
+      padding: 5px 0 7px;
+      border-bottom: 1px solid var(--grey20);
+      button:focus {
+        outline: none;
```
SharglutDev: Has it been specified that we don't want the default browser behavior for this ?

front/src/modules/infra/components/InfraSelector/InfraSelectorGrantsManager.tsx
```diff
@@ -0,0 +1,98 @@
...
+          </span>
+        </button>
+      </div>
+      {displayGrantSection && (
+        <div
+          className="infra-selector-subject-list"
+          role="button"
+          tabIndex={0}
+          onClick={(e) => e.stopPropagation()}
+        >
```
SharglutDev, line 73: With this nested button, we have a warning in console : 

Capture d’écran 2025-03-24 à 16 20 25

front/src/modules/infra/styles/_infraSelector.scss
```diff
@@ -127,6 +127,65 @@
...
+        color: var(--primary60);
+      }
+      .grey-label {
+        color: var(--grey30);
+      }
+    }
+    .infra-selector-subject {
+      &-list {
+        display: flex;
+        flex-direction: column;
```
SharglutDev: Not needed

```suggestion
```

front/src/modules/infra/styles/_infraSelector.scss
```diff
@@ -127,6 +127,65 @@
...
+        border-radius: 6px;
+        margin: 15px 8px 8px;
+        background-color: var(--ambientC10);
+      }
+      &-card {
+        display: flex;
+        justify-content: space-between;
+        height: 47px;
+        margin: 0 8px;
```
SharglutDev: You can simplify with 
```suggestion
        margin-inline: 8px;
```

front/src/modules/infra/styles/_infraSelector.scss
```diff
@@ -127,6 +127,65 @@
...
+        padding-left: 7px;
```
SharglutDev: There seems to be a 4px padding on the right
```suggestion
        padding-inline: 7px 4px;
```
Mockup : 
Capture d’écran 2025-03-24 à 16 32 32

front/src/modules/infra/styles/_infraSelector.scss
```diff
@@ -127,6 +127,65 @@
...
+        &:not(:last-child) {
+          border-bottom: 1px solid var(--grey20);
+        }
+      }
+      &-name {
+        display: flex;
+        align-items: center;
+        color: black;
```
SharglutDev: Same color but more scalable if we ever change the `black100` value
```suggestion
        color: var(--black100);
```

front/src/modules/infra/styles/_infraSelector.scss
```diff
@@ -127,6 +127,65 @@
...
+      }
+      &-grant {
+        .feed-back {
+          padding-top: 7px;
+          padding-bottom: 8px;
+        }
```
SharglutDev: Since a few weeks, we added a prop `narrow` to some ui-core components (input, select, combobox, pickers).
This prop cancel all default component paddings. By using it on the `Select` component here, you can remove these overriding styles and just put an `align-items: center;` on the `.infra-selector-subject-card`

front/src/modules/infra/styles/_infraSelector.scss
```diff
@@ -127,6 +127,65 @@
...
+        * > .osrd-ui-custom-select {
+          background-color: white;
+          margin: 0;
+          width: 136px;
+          height: 32px;
```
SharglutDev: Oh wow I wonder if we should allow the `Select` component to accept some custom props for this kind of use case or if it's better to override the style like that

front/src/modules/infra/styles/_infraSelector.scss
```diff
@@ -127,6 +127,65 @@
...
+        align-items: center;
+        color: black;
+      }
+      &-grant {
+        .feed-back {
+          padding-top: 7px;
+          padding-bottom: 8px;
+        }
+        * > .osrd-ui-custom-select {
+          background-color: white;
```
SharglutDev: Same as the black comment
```suggestion
          background-color: var(--white100);
```

front/src/modules/infra/utils/generateGrantSelectProps.ts
```diff
@@ -0,0 +1,69 @@
...
+    resourceGrants,
+    userGrant: connectedUserGrant,
+    requiredGrant: subjectGrant,
+  });
+
+  if (!isUserGrantGreaterThanSubject) {
+    return {
+      value: currentSubjectGrantObject,
+      options: [currentSubjectGrantObject],
+      disabled: true,
```
SharglutDev: The styles for this use case correspond to a `readOnly` not a `disabled` (but there is still a bug for the double arrow color, we should open an issue in osrd-ui)

SharglutDev (commented):
Thank you for your PR ! Reviewed styles and trads for now

front/src/modules/infra/components/InfraSelector/InfraSelectorModalBodyStandard.tsx
```diff
@@ -38,6 +43,7 @@ export default function InfraSelectorModalBodyStandard({
   filter = '',
   setFilter,
   infrasList,
+  infraIdsList = [],
```
SharglutDev: How the default value works ? It's used when we don't provide the prop to the component. But is it the same when we pass the prop as `undefined` to the component ?

Maybe in `InfraSelectorModal`, we could do 
```
  const infraIdsList = useMemo(
    () => infrasList?.results.map((infra) => infra.id) ?? [],
    [infrasList]
  );
```
This way this component always receives a `number[]`

front/src/applications/editor/Editor.tsx
```diff
@@ -119,6 +122,21 @@ const Editor = () => {
...
+    useProtectedAction({
+      resourceType: 'infra' as ResourceType,
+      resourceId: infraID,
+      requiredPrivileges: ['can_write'],
+    }),
+    [infraID]
+  );
+
+  const userCanEditInfra = useCheckPrivileges({
+    resourceType: 'infra' as ResourceType,
```
SharglutDev: Not needed
```suggestion
      resourceType: 'infra',
      resourceId: infraID,
      requiredPrivileges: ['can_write'],
    }),
    [infraID]
  );

  const userCanEditInfra = useCheckPrivileges({
    resourceType: 'infra',
```

front/src/applications/editor/Editor.tsx
```diff
@@ -157,6 +175,7 @@ const Editor = () => {
         viewport,
         mapStyle,
       },
+      protectedAction,
```
SharglutDev, line 179: Should it be in the depths as well ?

front/public/locales/fr/translation.json
```diff
@@ -157,6 +157,7 @@
         }
       },
       "infra-locked": "Attention, vous ne pourrez pas sauvegarder de modification.",
+      "infra-readonly": "Attention, vous ête en lecture seule sur cette infra",
```
SharglutDev:
```suggestion
      "infra-readonly": "Attention, vous êtes en lecture seule sur cette infra",
```

front/src/applications/editor/Editor.tsx
```diff
@@ -119,6 +122,21 @@ const Editor = () => {
...
   };
 
+  const protectedAction = useCallback(
+    useProtectedAction({
+      resourceType: 'infra' as ResourceType,
+      resourceId: infraID,
+      requiredPrivileges: ['can_write'],
+    }),
+    [infraID]
+  );
```
SharglutDev: Is it "react friendly" to put a custom hook in a `useCallback` ? Should to callback be inside the hook itself ?

front/src/common/authorization/hooks/useResourcesGrants.ts
```diff
@@ -0,0 +1,97 @@
...
+  name: string;
+}[];
+
+/**
+ * provides the user grants on a list of resources
+ * @param payload the resource type and the resource ids
+ * @returns the user grants on the resources
+ */
+const useResourcesGrants = (payload: Partial<Record<ResourceType, number[]>>) => {
+  const hasResources = Object.entries(payload).some(([_, ids]) => ids && ids.length > 0);
```
SharglutDev: `ids` is always defined here
```suggestion
  const hasResources = Object.entries(payload).some(([_, ids]) => ids.length > 0);
  
```

SharglutDev: The variable name is a bit misleading. This guarantees that one of the resource type has at least one grants but `.some()` will stop as soon as he finds one `ResourceType` matching the condition. If the first and second `ResourceType` have grants we won't know for the second one.

If I understood correctly, for now we handle only the grants for `infra`. We can maybe leave a TODO above this line to remember to find a better way to handle multiple resource grants when we will deal with `timetable` grants.

front/src/common/authorization/hooks/useResourcesGrants.ts
```diff
@@ -0,0 +1,97 @@
+import { useEffect, useMemo, useState } from 'react';
+
+import { mockedEditoastApi, type Grant, type ResourceType } from 'common/api/mock/mockEditoastApi';
+
+export type SubjectsGrantWithDetails = {
+  id: number;
+  type: string;
+  grant: Grant;
+  name: string;
+}[];
```
SharglutDev: Not sure to understand what `id`, `type` and `name` refer to. Maybe we could add some doc.

front/src/common/authorization/hooks/useResourcesGrants.ts
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * provides the user grants on a list of resources
+ * @param payload the resource type and the resource ids
+ * @returns the user grants on the resources
+ */
+const useResourcesGrants = (payload: Partial<Record<ResourceType, number[]>>) => {
+  const hasResources = Object.entries(payload).some(([_, ids]) => ids && ids.length > 0);
+  const [fetchUserGrantByResourceId, { data: userResourcesGrants }] =
+    mockedEditoastApi.endpoints.postUserResourcesGrants.useMutation();
```
SharglutDev, line 21: It's pretty annoying for this endpoint and `postUsersGrantsByResourceId` to be `mutation` when they in fact seem to be `get` endpoints (if I'm not mistaking).
If their signatures don't change when the back is delivered, we should put these 2 endpoints in `openapi-editoast-config.ts` to transform them into `query` endpoints so we can use them in a more intuitive way.

SharglutDev, line 21: Just to be sure : this endpoint returns the grants for the logged user (me) and `postUsersGrantsByResourceId` return the grants for all users including me ?

front/src/common/authorization/utils/checkPrivileges.ts
```diff
@@ -0,0 +1,32 @@
...
+  requiredGrant,
+  requiredPrivileges = [],
+}: CheckPrivilegesParams) => {
+  const userPrivileges = resourceGrants[userGrant] || [];
+  const privilegesToCheck = requiredGrant
+    ? resourceGrants[requiredGrant] || []
+    : requiredPrivileges;
+
+  return privilegesToCheck.every((privilege) => userPrivileges.includes(privilege));
+};
```
SharglutDev, line 30: We use grants everywhere but here we switch to privilege. Maybe we should stick to one naming.

front/src/modules/infra/utils/generateGrantSelectProps.ts
```diff
@@ -0,0 +1,69 @@
...
+  subjectGrant: string;
+  connectedUserGrant: string;
+  t: TFunction;
+}) => {
+  const currentSubjectGrantObject = {
+    label: t(`grants.${GrantsLabel[subjectGrant as GrantsLabelKeys]}`),
+    value: subjectGrant,
+  };
+
+  const isUserGrantGreaterThanSubject = checkPrivileges({
```
SharglutDev, line 28: Jus to be sure : does a `reader` can give `read` grant to another user with no grant ?

front/src/reducers/infra/generateGrantSelectProps.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+      options: [
+        { label: 'grants.none', value: 'NONE' },
+        { label: 'grants.read', value: 'READER' },
+        { label: 'grants.edit', value: 'WRITER' },
+      ],
+      disabled: false,
+    });
+  });
+
+  it('should handle empty resourceGrants gracefully', () => {
```
SharglutDev: How can a resource not have grants ?

front/src/modules/infra/components/InfraSelector/InfraSelectorEditionItem.tsx
```diff
@@ -37,7 +44,7 @@ export default function InfraSelectorEditionItem({
               type="button"
               aria-label={t('infraManagement:actions.delete')}
               title={t('infraManagement:actions.delete')}
-              onClick={() => setRunningDelete(true)}
+              onClick={() => protectWithOwnerGrant(() => setRunningDelete(true))}
```
SharglutDev, line 47: I guess it will be fixed when back is connected, but for now if I create an infra, I can't delete it myself

front/src/styles/scss/applications/operationalStudies/_scenario.scss
SarahBellaha: Looks like a rebase error ^^'

front/public/locales/en/infraManagement.json
SarahBellaha: Yes, i know, but it's more consistent with the french text to translate it that way. @thibautsailly saw it and agreed, i should have specified it in the linked issue.

front/src/modules/infra/styles/_infraSelector.scss
SarahBellaha: It will be done later :)

front/src/modules/infra/components/InfraSelector/InfraSelectorGrantsManager.tsx
SarahBellaha: We don't have the information needed yet for that icon.
It'll be used to display information like in the screenshot ⬇️ 

Capture d’écran 2025-03-26 à 10 52 38

front/src/modules/infra/styles/_infraSelector.scss
SarahBellaha: Well not really, but it was kinda implied by the fact that there is no outline around "details" and "collapse" on the mockup.

front/src/common/authorization/hooks/useResourcesGrants.ts
SarahBellaha: This variable guarantees that we have at least one **resource id** requested for a given `ResourceType`, not grants.
I agree that we'll have to find a better way to handle multiple `ResourceType` in the payload, i added a TODO 👍

front/src/common/authorization/hooks/useResourcesGrants.ts
SarahBellaha: It refers to the id, type (user or group), name and grant of a subject. This type name wasn't very well chosen, i changed it :)

front/src/common/authorization/hooks/useResourcesGrants.ts
SarahBellaha: Yes, we can do that for the `postUserResourcesGrants` when the back is delivered.
The other endpoint will no longer be a mutation :)

SarahBellaha: Right 👍

front/src/common/authorization/utils/checkPrivileges.ts
SarahBellaha: Grants and Privileges are two different things. The grant is what gives you a set of privileges over a resource.
Ex: If i have the grant "**WRITER**", i have the privileges `['can_read', 'can_share_read', 'can_write', 'can_share_write']`

front/src/modules/infra/utils/generateGrantSelectProps.ts
SarahBellaha: Yes he can :)
On the other hand, he will not be able to share this grant with a user having a higher grant than his own, as it would imply a downgrading of the user's grant.

front/src/modules/infra/components/InfraSelector/InfraSelectorEditionItem.tsx
SarahBellaha: Yes it's normal as, for now, we just display mocked data that isn't updated. When the back is connected, a user who creates or duplicates an infra will automatically be granted OWNER on this resource and will therefore be able to delete it.

front/src/modules/infra/styles/_infraSelector.scss
SharglutDev: Hmm down side is it's not outlined anymore if we tab on it

front/src/modules/infra/styles/_infraSelector.scss
```diff
@@ -127,6 +127,60 @@
...
+        height: 47px;
+        margin-inline: 8px;
+        padding-inline: 7px 4px;
+        &:not(:last-child) {
+          border-bottom: 1px solid var(--grey20);
+        }
+      }
+      &-name {
+        display: flex;
+        align-items: center;
```
SharglutDev: We can remove this now that we have `align-items: center;` in the card
```suggestion
```

front/src/common/authorization/hooks/useResourcesGrants.ts
SharglutDev: Hope that `type` becomes a tuple in the back PR :)

front/src/common/authorization/utils/checkPrivileges.ts
SharglutDev: Oh ok, got it, thanks for the explanation !

SharglutDev (approved):
Except the cursor pointer on the user list that should probably be removed (can be on next PR), lgtm and tested, good job :)
> [!NOTE]
> To be merged only on April 11th

PR #11169, opened 2025-03-14
Add editoast impersonate authn

flomonster:
Part of #11182

### DIFF
--- a/editoast/editoast_authz/src/authorizer.rs
+++ b/editoast/editoast_authz/src/authorizer.rs
@@ -2,6 +2,7 @@ use std::collections::HashSet;
 use tracing::debug;
 use tracing::Level;
 
+use crate::subject::UserIdentity;
 use crate::subject::UserInfo;
 use crate::Error;
 use crate::Regulator;
@@ -17,20 +18,23 @@ pub struct Authorizer<S: StorageDriver> {
 }
 
 impl<S: StorageDriver> Authorizer<S> {
+    /// Initialize an authorizer for the given user.
+    /// If the user doesn't exist, an error is returned.
     #[tracing::instrument(skip_all, fields(%user), ret(level = Level::DEBUG), err)]
     pub async fn try_initialize(
-        user: UserInfo,
+        user: UserIdentity,
         regulator: Regulator<S>,
     ) -> Result<Self, Error<S::Error>> {
-        let user_id = regulator
+        let user_info = regulator
             .driver
-            .ensure_user(&user)
+            .get_user_info_by_identity(&user)
             .await
-            .map_err(Error::Storage)?;
-        debug!(%user, %user_id, "user authenticated");
+            .map_err(Error::Storage)?
+            .ok_or(Error::UnknownUser { identity: user })?;
+        debug!(%user_info, "user authenticated");
         let authorizer = Self {
-            user,
-            user_id,
+            user: user_info.info,
+            user_id: user_info.id,
             regulator,
         };
         Ok(authorizer)
@@ -65,6 +69,7 @@ mod tests {
     use super::*;
     use crate::subject::GroupInfo;
     use crate::subject::GroupName;
+    use crate::subject::User;
     use crate::subject::UserIdentity;
     use crate::Role;
     use futures::stream;
@@ -84,29 +89,29 @@ mod tests {
 
     #[tokio::test]
     async fn check_user_roles() {
+        let user_identity = || "toto".to_owned();
         let user = || UserInfo {
-            identity: "toto".to_owned(),
+            identity: user_identity(),
             name: "Sir Toto, the One and Only".to_owned(),
         };
         let regulator = Regulator::new(crate::openfga!(), MockAuthDriver::default());
         let regulator = move || regulator.clone();
 
-        // setup user (using the authorizer)
+        // setup user
         let user_id = {
-            let authorizer = Authorizer::try_initialize(user(), regulator())
-                .await
-                .unwrap();
-            let users = authorizer.regulator.driver.users.lock().unwrap();
+            let regulator = regulator();
+            let User { id, .. } = regulator.driver.ensure_user(&user()).await.unwrap();
+            let users = regulator.driver.users.lock().unwrap();
             assert_eq!(
                 users.iter().next(),
-                Some((&"toto".to_owned(), &0)),
+                Some((&"toto".to_owned(), &id)),
                 "new user should have been created"
             );
-            authorizer.user_id()
+            id
         };
 
         let id = {
-            let authorizer = Authorizer::try_initialize(user(), regulator())
+            let authorizer = Authorizer::try_initialize(user_identity(), regulator())
                 .await
                 .unwrap();
             authorizer.user_id()
@@ -124,19 +129,19 @@ mod tests {
                 .expect("roles should be granted");
         }
 
-        assert!(Authorizer::try_initialize(user(), regulator())
+        assert!(Authorizer::try_initialize(user_identity(), regulator())
             .await
             .unwrap()
             .check_roles(HashSet::from([Role::OperationalStudies]))
             .await
             .expect("should check roles successfully"));
-        assert!(Authorizer::try_initialize(user(), regulator())
+        assert!(Authorizer::try_initialize(user_identity(), regulator())
             .await
             .unwrap()
             .check_roles(HashSet::from([Role::Stdcm, Role::Admin]))
             .await
             .expect("should check roles successfully"));
-        assert!(!Authorizer::try_initialize(user(), regulator())
+        assert!(!Authorizer::try_initialize(user_identity(), regulator())
             .await
             .unwrap()
             .check_roles(HashSet::from([Role::Admin]))
@@ -151,13 +156,13 @@ mod tests {
                 .expect("roles should be stripped");
         }
 
-        assert!(!Authorizer::try_initialize(user(), regulator())
+        assert!(!Authorizer::try_initialize(user_identity(), regulator())
             .await
             .unwrap()
             .check_roles(HashSet::from([Role::OperationalStudies]))
             .await
             .expect("should check roles successfully"));
-        assert!(Authorizer::try_initialize(user(), regulator())
+        assert!(Authorizer::try_initialize(user_identity(), regulator())
             .await
             .unwrap()
             .check_roles(HashSet::from([Role::Stdcm]))
@@ -165,7 +170,7 @@ mod tests {
             .expect("should check roles successfully"));
 
         // no roles
-        assert!(Authorizer::try_initialize(user(), regulator())
+        assert!(Authorizer::try_initialize(user_identity(), regulator())
             .await
             .unwrap()
             .check_roles(HashSet::from([]))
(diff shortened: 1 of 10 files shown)


editoast/editoast_authz/src/regulator.rs
```diff
@@ -30,17 +31,17 @@ pub struct Regulator<S: StorageDriver> {
...
-    ) -> impl Future<Output = Result<Option<i64>, Self::Error>> + Send;
+    ) -> impl Future<Output = Result<Option<UserInfoWithID>, Self::Error>> + Send;
 
     fn get_group_id(
         &self,
         group_name: &GroupName,
     ) -> impl Future<Output = Result<Option<i64>, Self::Error>> + Send;
 
-    fn get_user_info(
+    fn get_user_info_from_id(
```
hamz2a:
```suggestion
    fn get_user_info_by_id(
```

editoast/editoast_authz/src/authorizer.rs
```diff
@@ -17,20 +18,44 @@ pub struct Authorizer<S: StorageDriver> {
...
+    pub async fn initialize(
+        user: UserIdentity,
+        regulator: Regulator<S>,
+    ) -> Result<Self, Error<S::Error>> {
+        let user_info = regulator
+            .driver
+            .get_user_info(&user)
+            .await
+            .map_err(Error::Storage)?
+            .ok_or(Error::UnknownUser(user))?;
```
hamz2a: It already exists a [variant](https://github.com/OpenRailAssociation/osrd/blob/dev/editoast/editoast_authz/src/regulator.rs#L154) for unknown users.

editoast/editoast_authz/src/lib.rs
```diff
@@ -47,6 +47,8 @@ pub async fn ensure_latest_authorization_model(
 pub enum Error<StorageError: std::error::Error> {
     #[error("unknown subject {0}")]
     UnknownSubject(i64),
+    #[error("unknown user {0}")]
+    UnknownUser(String),
```
hamz2a:
```suggestion
```

hamz2a (approved):
Thanks for the PR, I left some non-blocking comments.

editoast/editoast_authz/src/lib.rs
flomonster: In this case, I don't have the user's ID (`i64`) only its identity. That's why I think we need a new variant.

hamz2a: Got it!

editoast/editoast_authz/src/authorizer.rs
```diff
@@ -17,20 +18,44 @@ pub struct Authorizer<S: StorageDriver> {
...
+            user_id: user_info.id,
+            regulator,
+        };
+        Ok(authorizer)
+    }
+
+    /// Initialize an authorizer for the given user.
+    /// If the user doesn't exist, an error is returned.
+    #[tracing::instrument(skip_all, fields(%user), ret(level = Level::DEBUG), err)]
+    pub async fn initialize(
```
woshilapin: Not using `try_` prefix when there is another function `try_initialize` looks weird since both functions returns `Result`. Maybe we should improve the naming, one with `try_initialize / try_impersonate`? Or `try_initialize_and_create / try_initialize`?

editoast/editoast_authz/src/authorizer.rs
```diff
@@ -17,20 +18,44 @@ pub struct Authorizer<S: StorageDriver> {
 }
 
 impl<S: StorageDriver> Authorizer<S> {
+    /// Initialize an authorizer for the given user.
+    /// If the user doesn't exist, it will be created.
     #[tracing::instrument(skip_all, fields(%user), ret(level = Level::DEBUG), err)]
     pub async fn try_initialize(
```
woshilapin: The only semantic difference between `try_initialize` and `initialize` is that one is creating the user in DB if it doesn't exist and the other fail. But the difference in code is that one is using `ensure_user` and the other `get_user_info` and that doesn't help to understand the semantic difference. It feels that `try_initialize` should instead explicitly do the two operations: first use `get_user_info`, and only if no user exists, trigger the creation. It might need some refactoring (moving the logic of `ensure_user` where it's `None` in a separate function). What do you think?

editoast/editoast_authz/src/authorizer.rs
flomonster: I chose `try_initialize_and_create`

editoast/editoast_authz/src/authorizer.rs
flomonster: I'm not sure I see the problem. If I understand correctly, you consider that the logic of creating the user if it doesn't exist is too low in the API (`ensure_user`) when it should be handled at a higher level?

woshilapin: That is one way to put it. At least, doing this change would make clearer the purpose of `try_initialize` and `try_initialize_and_create` (the difference between both would make it easy to understand the difference).

I guess another problem is that `ensure_user` and `get_user_info` have almost similar behavior and very different name.

And lastly, `ensure_user` is a weird name for what it does (get user info and creating it in DB if not here yet).

Maybe the problem is the mixing of all these independent weirdness, or maybe just fixing one of this weirdness would help make sense of all of it better 🤷 

Note: If you feel like you disagree with my assessment, feel free to resolve, I believe having renamed the functions into `try_initialize` and `try_initialize_and_create` is already a nice enough step forward.

SharglutDev (approved):

leovalais:
Quick review on the phone, will look into it more closely later. 

- There is no need to have two different constructors for the authorizer, expect the user to exist and catch the error and ensure it if it doesn't. 

- Instead of modifying the trait StorageDriver, what about making the new methods you need new functions with a default implementation, overriden in the Pg implementation to perform only a single query?

flomonster:
> Quick review on the phone, will look into it more closely later.
> 
>     * There is no need to have two different constructors for the authorizer, expect the user to exist and catch the error and ensure it if it doesn't.
> 
>     * Instead of modifying the trait StorageDriver, what about making the new methods you need new functions with a default implementation, overriden in the Pg implementation to perform only a single query?

Fixed, I kept the StorageDriver symmetric between users and groups. The only change is the `ensure_user` which returns more information. But it seems required.

editoast/editoast_authz/src/authorizer.rs
flomonster: In response to @leovalais's comment, I've minimized StorageDriver modifications. If we want to improve this trait I suggest doing it in another PR.

editoast/editoast_authz/src/authorizer.rs
```diff
@@ -17,20 +18,44 @@ pub struct Authorizer<S: StorageDriver> {
...
+            Err(Error::UnknownUser(_)) => {
+                regulator
+                    .driver
+                    .ensure_user(&user)
+                    .await
+                    .map_err(Error::Storage)?;
+                Self::try_initialize(user.identity, regulator).await
+            }
+            Err(e) => Err(e),
+        }
```
leovalais: I think this logic could be inlined in the authentication middleware since it's the only place it's used really. (For tests, just using `regulator.ensure_user` does the same.)

editoast/editoast_authz/src/authorizer.rs
```diff
@@ -321,6 +352,24 @@ mod tests {
...
+                .get(user_identity)
+                .map(|&id| UserInfoWithID {
+                    info: UserInfo {
+                        identity: user_identity.clone(),
+                        name: "Mocked User".to_owned(),
+                    },
+                    id,
+                }))
+        }
+
```
leovalais: Perf doesn't matter here, let's keep the blanket implementation for coverage

editoast/editoast_authz/src/lib.rs
leovalais:
```suggestion
    #[error("unknown user {identity}")]
    UnknownUser { identity: String },
```

nit, but more explicit

editoast/editoast_authz/src/subject.rs
```diff
@@ -8,6 +8,12 @@ pub struct UserInfo {
     pub name: UserName,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct UserInfoWithID {
```
leovalais:
```suggestion
pub struct User {
```

editoast/src/models/auth_driver.rs
```diff
@@ -93,15 +116,15 @@ impl StorageDriver for PgAuthDriver {
...
             async move {
-                let user_id = self.get_user_id(&user.identity).await?;
-                match user_id {
-                    Some(user_id) => {
-                        tracing::debug!(user_id, "user already exists in db");
-                        Ok(user_id)
+                let user_info = self.get_user_info_by_identity(&user.identity).await?;
+                match user_info {
+                    Some(user_info) => {
+                        tracing::debug!(user_info.id, "user already exists in db");
```
leovalais:
```suggestion
                        tracing::debug!(?user_info, "user already exists in db");
```

editoast/src/views/mod.rs
```diff
@@ -357,6 +380,12 @@ pub enum AuthorizationError {
     #[error("Forbidden — user has insufficient privileges")]
     #[editoast_error(status = 403)]
     Forbidden,
+    #[error("Forbidden — user must be an admin to impersonate")]
+    #[editoast_error(status = 403)]
+    ForbiddenImpersonate,
```
```suggestion
    ForbiddenImpersonation,
```

but I think the `Forbidden` variant semantics already covers this case

editoast/src/views/mod.rs
```diff
@@ -357,6 +380,12 @@ pub enum AuthorizationError {
...
+    #[error("Not Found — impersonated user '{0}' not found")]
+    #[editoast_error(status = 403)]
+    ImpersonatedUserNotFound(String),
```
```suggestion
    #[error("Not Found — impersonated user '{identity}' not found")]
    #[editoast_error(status = 403)]
    ImpersonatedUserNotFound { identity: String },
```

nit++

leovalais (commented):
Thanks for changing what we discussed yesterday, a proper review this time... ⬇️

editoast/src/views/mod.rs
flomonster: I prefer to have a different error so that we can easily identify that the problem comes from the impersonalization feature. Many things can cause a simple “forbidden” error.

editoast/editoast_authz/src/authorizer.rs
flomonster: I don't think I understood your comment. What is the authentication middleware?

leovalais: The middleware in `views/mod.rs` where the headers are parsed

editoast/src/views/mod.rs
leovalais: Ok fair

leovalais (approved):
LGTM and tested provided https://github.com/OpenRailAssociation/osrd/pull/11169#discussion_r2012316004

A few questions about edge cases

1. Can an Admin impersonate another Admin? Not sure why it would be problematic, but still asking
2. Currently a request with `x-identity == x-impersonate` fails, but I'd argue that everyone should be allowed to "impersonate themselves"

Nit: could `x-impersonated-user` be renamed to `x-impersonate`, it's annoying to type when crafting curl requests 😅

flomonster:
> LGTM and tested provided [#11169 (comment)](https://github.com/OpenRailAssociation/osrd/pull/11169#discussion_r2012316004)
> 
> A few questions about edge cases
> 
>     1. Can an Admin impersonate another Admin? Not sure why it would be problematic, but still asking
> 
>     2. Currently a request with `x-identity == x-impersonate` fails, but I'd argue that everyone should be allowed to "impersonate themselves"
> 
> 
> Nit: could `x-impersonated-user` be renamed to `x-impersonate`, it's annoying to type when crafting curl requests 😅

1. I don't think it'sproblematic
2. In my opinion, only admin can impersonate (and have access to the interface to impersonate). If we say that a user can impersonate himself why not allow a user to impersonate a user with fewer permissions, which is complicated to implement.

We can change the header name, I'm on it.

leovalais (approved):

PR #11172, opened 2025-03-17
front: map: factorize sources

### LINKED ISSUE #11106: Map: refactor data sources and layers
We use https://visgl.github.io/react-map-gl/

The convenient way to describe data source and layer is to do `<Source …props><Layer …props /></Source>`
However, if we have multiple layers that use the same source, with declare two sources. This results in downloading twice the data.

We should factorize it and separate source and layers

Tristramg:
This pull requests separates the map Sources from the Layers. This allows having only one data source if it is used in multiple layers.

There should be no visible change on the map (try adding layers such as OSM Tracks, plateforms, activate terrain…).

In the network console we should notice that there no longer are call multiple times to the same data source

Fixes #11106

### DIFF
--- a/front/src/common/Map/Layers/Background.tsx
+++ b/front/src/common/Map/Layers/Background.tsx
@@ -1,4 +1,4 @@
-import { Source, type LayerProps } from 'react-map-gl/maplibre';
+import { type LayerProps } from 'react-map-gl/maplibre';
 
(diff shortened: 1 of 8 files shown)


Synar:
Code lgtm ^^

Synar (approved):

emersion (approved):
LGTM!

PR #11174, opened 2025-03-17
editoast, front, core, gateway, osrdyne, railjson_generator: add justfiles

Tristramg:


emersion:
Note, we don't use Conventional Commits ("chores:" prefix). We use the code area as the prefix.

editoast/justfile
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env just --justfile
+
+# Display available recipes
+help:
+    @just --list --unsorted
+
+# Run the application; does not start other services
+run:
+    cargo run
```
bougue-pe: Adding `EDITOAST_CORE_SINGLE_WORKER` as core is running on all infra (hard to provide the envvar for a one-infra dedicated core worker), and previous scripts:
```suggestion
    ./assets/sprites/generate-atlas.sh
    ./assets/fonts/generate-glyphs.sh
    diesel migration run --locked-schema
    @echo "Starting editoast in single worker mode. Configure the other services accordingly"
    @echo "If you use osrd-compose add the `sw` flag"
    EDITOAST_CORE_SINGLE_WORKER=true cargo run -- runserver
```

editoast/justfile
```diff
@@ -0,0 +1,29 @@
...
+
+# Install required dependencies
+install:
+    cargo build
+
+# Run the tests
+test:
+    cargo test -- --test-threads=4
```
woshilapin:
```suggestion
    cargo test --workspace -- --test-threads=4
```

woshilapin: nitpick: Why the need for `--test-threads`? I would just remove it. If you really need it, then what about this instead? (or through an environment variable)
```suggestion
test args:
    cargo test -- {{args}}
```

editoast/justfile
```diff
@@ -0,0 +1,29 @@
...
+
+# Format the source code
+format:
+    cargo fmt
+
+# Verify code lints without applying any fix
+lint:
+    cargo clippy
```
woshilapin:
```suggestion
    cargo clippy --workspace --all-targets --all-features
```

editoast/justfile
woshilapin, line 1: All the remarks on `editoast` probably also apply to `gateway` and `osrdyne` (and if it's mono-crate, it probably doesn't harm to support for future workspace's members).

python/railjson_generator/justfile
```diff
@@ -0,0 +1,33 @@
...
+    @just --list --unsorted
+
+# Generate small infra
+run:
+    mkdir -p small_infra_out
+    poetry run scripts/small_infra.py small_infra_out
+
+# Install required dependencies
+install:
+    poetry install
```
woshilapin: Do we want to default to `--all-extras`?

```suggestion
    poetry install --all-extras
```

woshilapin (approved):

python/railjson_generator/justfile
woshilapin, line 1: Any reason why `osrd_schemas` doesn't have a `justfile`?

Same for `osm_to_railjson`.

editoast/justfile
Tristramg: that’s what the readme says, should we change the readme aswell? Is it a decision that should be made in this pull request?

core/justfile
```diff
@@ -0,0 +1,32 @@
...
+gradle := if os() == "windows" { "gradlew.bat" } else { "./gradlew" }
+javaflags := "-ea -Xmx12g"
+
+# Display available recipes
+help:
+    @just --list --unsorted
+
+# Run the application; does not start other services
+run:
+    ALL_INFRA=true java {{javaflags}} -jar build/libs/osrd-all.jar worker --editoast-url http://localhost:8090/
```
eckter: ~~With the scalability and core containers being spawned automatically, it's sadly not quite as simple as that.~~

~~When we just want to run a single worker process (the most common use case when running locally), we need to add the env variable `ALL_INFRA=TRUE`.~~ But we also need to run the rest of the stack in "single worker" mode: `sw` flag in `osrd-compose`, or the env variables set in [docker-compose.single-worker.yml](https://github.com/OpenRailAssociation/osrd/blob/dev/docker/)

---

**EDIT:** I managed to misread and missed the flag I was commenting about, it's already there.

But it can still be a little misleading, that only works when the rest of the stack is configured accordingly. Maybe we could print something before running the command, to give a reminder?

core/justfile
```diff
@@ -0,0 +1,32 @@
...
+
+# Install required dependencies
+install:
+    {{ gradle }} shadowJar
+
+# Run the tests
+test:
+    {{ gradle }} test
```
eckter:
```suggestion
# Run the tests
test:
    {{ gradle }} check
```

Apparently the tests aren't properly configured, `gradle test` doesn't actually run them. But `gradle check` does. But it does run more than just tests.

I'll open an issue for the invalid test config: https://github.com/OpenRailAssociation/osrd/issues/11183

editoast/justfile
bougue-pe:
```suggestion
    cargo clippy --workspace --all-features --all-targets
```

editoast/justfile
```diff
@@ -0,0 +1,29 @@
...
+
+# Fix all the lints that can be automatically applied
+fix-lint:
+    cargo clippy --fix
```
bougue-pe: (all-targets is probably activated by default, but better to match with the lint)
```suggestion
    cargo clippy --workspace --all-features --all-targets --fix
```

python/railjson_generator/justfile
Tristramg: There are no extra dependencies, so I suppose it’s the same issue as with cargo: should we consider that its defaults are _in general_ bad, and we should change them, even if they don’t apply?

editoast/justfile
```diff
@@ -0,0 +1,29 @@
...
+help:
+    @just --list --unsorted
+
+# Run the application; does not start other services
+run:
+    cargo run
+
+# Install required dependencies
+install:
+    cargo build
```
bougue-pe: As discussed, we may install `diesel`, `spreet` and `build_pbf_glyphs` (the rest is too OS-specific for now)

editoast/justfile
Tristramg: I also added --all-features, as the CI runs that, I hope it’s ok

core/justfile
Tristramg: I added a comment in https://github.com/OpenRailAssociation/osrd/pull/11174/commits/3691d77d4980ac87c925113c4e096892e2b3461b

python/railjson_generator/justfile
Tristramg: laziness :grimacing: ?

There is no clear entrypoint, it looks more like a collection of functions, I’m not really sure what to do with it

editoast/justfile
```diff
@@ -0,0 +1,29 @@
...
+
+# Run the tests
+test:
+    cargo test --workspace -- --test-threads=4
+
+# Format the source code
+format:
+    cargo fmt
```
bougue-pe: Mainly adding taplo (ditto for all cargos):
```suggestion
    cargo fmt --all
    taplo fmt
```

core/justfile
bougue-pe, line 1: We may document a bit in the main readme.md the use of `just`

editoast/justfile
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env just --justfile
+
+# Display available recipes
+help:
+    @just --list --unsorted
+
```
bougue-pe, line 6: It may more be worth a `build` target (or in `test` with diff?), but regenerating openapi may be nice too:
`cargo run openapi > openapi.yaml`

front/justfile
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env just --justfile
+
+# Display available recipes
+help:
+    @just --list --unsorted
+
```
bougue-pe, line 6: Same than for editoast, maybe more a `build` target (or in `test` with diff?):
`npm generate-types`

bougue-pe (commented):
Thanks for the work on this!

python/railjson_generator/justfile
Tristramg: As discussed, I added --all-extras

python/railjson_generator/justfile
Tristramg: Added

editoast/justfile
Tristramg: The number of threads is now a by default 4, but can be overriden (and just allows to pass any arbitatry arguments after)

core/justfile
```diff
@@ -9,6 +9,8 @@ help:
 
 # Run the application; does not start other services
 run:
+    @echo "Starting core in single worker mode. Configure the other services accordingly"
+    @echo "If you use osrd-compose add the sw flag"
```
Khoyo:
```suggestion
    @echo "Starting core in single worker mode. If you use osrd-compose add the sw flag"
```

Khoyo (approved):
Bit of a nit, LGTM either way

editoast/justfile
Khoyo: We should probably call `scripts/sync-openapi.sh` instead

editoast/justfile
Khoyo: I don't think we should do system wide installs in the justfile...

editoast/osm_to_railjson/justfile
```diff
@@ -0,0 +1,29 @@
...
+install:
+    cargo build
+
+# Run the tests
+test:
+    cargo test
+
+# Format the source code
+format:
+    cargo fmt
```
bougue-pe:
```suggestion
    cargo fmt --all
    taplo fmt
```

gateway/justfile
```diff
@@ -14,15 +14,15 @@ install:
...
+    cargo test --workspace
 
 # Format the source code
 format:
     cargo fmt
 
 # Verify code lints without applying any fix
 lint:
-    cargo clippy
+    cargo clippy --workspace --all-features
```
bougue-pe:
```suggestion
    cargo clippy --workspace --all-features --all-targets
```

And this should also be spread to `fix-lint` below (whether all-targets is in or not)

osrdyne/justfile
```diff
@@ -14,15 +14,15 @@ install:
...
+    cargo test --workspace 
 
 # Format the source code
 format:
     cargo fmt
 
 # Verify code lints without applying any fix
 lint:
-    cargo clippy
+    cargo clippy --workspace --all-features
```
bougue-pe:
```suggestion
    cargo clippy --workspace --all-features --all-targets
```

And this should also be spread to `fix-lint` below (whether all-targets is in or not)

gateway/justfile
```diff
@@ -14,15 +14,15 @@ install:
 
 # Run the tests
 test:
-    cargo test
+    cargo test --workspace
 
 # Format the source code
 format:
     cargo fmt
```
bougue-pe:
```suggestion
    cargo fmt --all
    taplo fmt
```

osrdyne/justfile
```diff
@@ -14,15 +14,15 @@ install:
 
 # Run the tests
 test:
-    cargo test
+    cargo test --workspace 
 
 # Format the source code
 format:
     cargo fmt
```
bougue-pe:
```suggestion
    cargo fmt --all
    taplo fmt
```

editoast/osm_to_railjson/justfile
```diff
@@ -0,0 +1,29 @@
...
+
+# Verify code lints without applying any fix
+lint:
+    cargo clippy --all-features
```
bougue-pe:
```suggestion
    cargo clippy --all-features --all-targets
```

And this should also be spread to `fix-lint` below (whether workspace and all-targets are in or not)

editoast/justfile
bougue-pe: Indeed, makes sense to just put what's system-agnostic.
`cargo build` may even be removed (no strong opinion on that).
Then just echo a list of what should be available? (Nothing is fine too IMO)

editoast/justfile
bougue-pe: Yes, it's probably better indeed.
Still not sure where to put that, though (before `cargo run` is also an option).

editoast/osm_to_railjson/justfile
Tristramg: I don’t think that we want a --workspace here. That triggers a check on all editoast, not just osm_to_railjson

editoast/justfile
Tristramg: I opted for a just a log: c9ee55bb

core/justfile
Tristramg: df29d9f9

editoast/justfile
Tristramg: (woops, previous comment was on the wrong discussion)

I considered it’s a linting question. Could be discussed, but it feels like all those little chores that could be done automatically and that we forget to do.

See 5f6d54f6

front/justfile
Tristramg: Considering that it’s a change in editoast that triggers a change in openapi.yaml which in turn needs to run generate-types, shouldn’t it be an editoast job?

As suggested in 5f6d54f6, the lint-fix would generate an up-to-date openapi.yaml and generate the types thanks to sync-openapi.sh

editoast/justfile
Tristramg: I hope I forgot nothing in 589d85d9

Tristramg (commented):
Thank you for all the feedbacks.

I hope I didn’t miss anything, but otherwise I consider it as ready

editoast/justfile
```diff
@@ -0,0 +1,39 @@
...
+run:
+    ./assets/sprites/generate-atlas.sh
+    ./assets/fonts/generate-glyphs.sh
+    diesel migration run --locked-schema
+    @echo 'Starting editoast in single worker mode. If you use osrd-compose add the `sw` flag'
+    EDITOAST_CORE_SINGLE_WORKER=true cargo run -- runserver
+
+# Install required dependencies
+install:
+    @echo 'This recipe won’t install binaries. Please use `cargo install`, `cargo binstall` or you preferd tool to install'
```
woshilapin:
```suggestion
    @echo 'This recipe won’t install binaries. Please use `cargo install`, `cargo binstall` or you preferred tool to install'
```

editoast/osm_to_railjson/justfile
```diff
@@ -0,0 +1,30 @@
...
+    cargo test
+
+# Format the source code
+format:
+    cargo fmt --all
+    taplo fmt
+
+# Verify code lints without applying any fix
+lint:
+    cargo clippy --all-features --all-targets
```
woshilapin, line 26: nitpick: for coherence with other Rust projects.

```suggestion
    cargo clippy --workspace --all-features --all-targets
```

editoast/osm_to_railjson/justfile
```diff
@@ -0,0 +1,30 @@
...
+
+# Fix all the lints that can be automatically applied
+fix-lint:
+    cargo clippy --all-features --all-targets --fix
```
woshilapin, line 30: nitpick: for coherence with other Rust projects.

```suggestion
    cargo clippy --workspace --all-features --all-targets --fix
```

editoast/osm_to_railjson/justfile
```diff
@@ -0,0 +1,30 @@
...
+run *args:
+    cargo run {{args}}
+
+# Install required dependencies
+install:
+    cargo build
+
+# Run the tests
+test:
+    cargo test
```
woshilapin, line 17: nitpick: for coherence with other Rust projects.

```suggestion
    cargo test --workspace
```

python/osrd_schemas/justfile
```diff
@@ -0,0 +1,31 @@
...
+help:
+    @just --list --unsorted
+
+# Generate small infra
+run:
+    @echo "This component is a library"
+
+# Install required dependencies
+install:
+    poetry install --all-extras
```
woshilapin: I believe `poetry sync` is more reliable than `install` (will remove unused dependencies too).

editoast/justfile
woshilapin: I've added some comments about missing `--workspace` for `osm_to_railjson`. I'm resolving the current thread.

woshilapin (commented):
A few more comments, nothing big.

editoast/osm_to_railjson/justfile
Tristramg: We don’t want --workspace for osm_to_railjson: it would check editoast as a whole, and not just this module. So I think here it would be better not

python/osrd_schemas/justfile
Tristramg: changed in 870540bf

editoast/justfile
```diff
@@ -0,0 +1,39 @@
...
+run:
+    ./assets/sprites/generate-atlas.sh
+    ./assets/fonts/generate-glyphs.sh
+    diesel migration run --locked-schema
+    @echo 'Starting editoast in single worker mode. If you use osrd-compose add the `sw` flag'
+    EDITOAST_CORE_SINGLE_WORKER=true cargo run -- runserver
+
+# Install required dependencies
+install:
+    @echo 'This recipe won’t install binaries. Please use `cargo install`, `cargo binstall` or you preferred tool to install'
```
bougue-pe:
```suggestion
    @echo 'This recipe won’t install binaries. Please use `cargo install`, `cargo binstall` or your preferred tool to install'
```

python/osrd_schemas/justfile
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env just --justfile
```
bougue-pe, line 1: There is a room for a `justfile` in `./tests` too.
Integration tests are in python, so fmt, lint and test are legitimate targets there, probably inspired by this "python" justfile?

bougue-pe (approved):

python/osrd_schemas/justfile
Tristramg: After discussing with @woshilapin on this issue, we decided to let it be for now: it uses an old version of poetry. We wait until it goes to uv+ruff

Tristramg:
I squashed everything, I’m waiting for the last reviews

woshilapin:
I discover there is an [(unstable) formatting feature](https://just.systems/man/en/formatting-and-dumping-justfiles.html?highlight=format#formatting-and-dumping-justfiles), I just push a commit to format the justfiles.

flomonster (approved):
LGTM (not tested)

eckter (approved):

clarani (approved):

PR #11187, opened 2025-03-18
gateway: add path blocking

ghost:
In order to expose a public version of OSRD, we will need to block various paths directly in the gateway.

This PR aims to add this feature to the gateway.

### DIFF
--- a/gateway/actix_proxy/src/lib.rs
+++ b/gateway/actix_proxy/src/lib.rs
@@ -28,6 +28,7 @@ use opentelemetry::{
     KeyValue,
 };
 use percent_encoding::{utf8_percent_encode, AsciiSet};
+use regex::RegexSet;
 
 use awc::error::{ConnectError, SendRequestError as AwcSendRequestError};
 use futures_util::StreamExt;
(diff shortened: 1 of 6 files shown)


gateway/actix_proxy/src/lib.rs
```diff
@@ -79,6 +80,7 @@ pub struct Proxy {
     upstream_path_prefix: String,
     timeout: Option<Duration>,
     tracing_name: Option<String>,
+    blocked_paths: Vec<Regex>,
```
Khoyo: Use a [`RegexSet`](https://docs.rs/regex/latest/regex/struct.RegexSet.html)

gateway/actix_proxy/src/lib.rs
```diff
@@ -173,6 +183,10 @@ impl Proxy {
             .build()
             .expect("failed to build uri")
     }
+
+    pub fn check_if_path_blocked(&self, path: &str) -> bool {
```
Khoyo:
```suggestion
    pub fn is_path_blocked(&self, path: &str) -> bool {
```

gateway/actix_proxy/src/lib.rs
```diff
@@ -135,6 +138,12 @@ impl Proxy {
             upstream_path_prefix.push('/');
         }
 
+        let parsed_blocked_path = blocked_paths
+            .unwrap_or_default()
+            .into_iter()
+            .map(|raw| Regex::new(&raw).expect("Regex should be parsable"))
```
Khoyo: nit: expect messages should start with a lowercase

Khoyo (approved):

PR #11188, opened 2025-03-18
core: refactor type Path to BlockPath

### LINKED ISSUE #10595: core: refactor path type for consistency
Main issue : https://github.com/OpenRailAssociation/osrd/issues/10239

louisgreiner:
Close https://github.com/OpenRailAssociation/osrd/issues/10595

I tried not to change the current behavior, even if sometimes it looks that the wrong type is used, to ease the review of this PR. The hunt of the wrong typing can be done in another PR. For instance, we could take a look on the calls to these types of functions:
- `toTravelledPath`
- `fromTravelledPath`

For the review:
- I mainly renamed `Path` into `BlockPath`
- I, only a few times (but it might be the most important part for review), fixed the typing from `Path` into `TravelledPath`. Theses changes are important, because the behaviour of the both types are different:
```patch
- Path
+ TravelledPath
```

I also suspect a wrong typing for [these variables](https://github.com/OpenRailAssociation/osrd/blob/f7a7db46bed212a61be6b641e2efd09f58c1fe2c/core/src/main/java/fr/sncf/osrd/conflicts/IncrementalPath.kt#L140-L141), but I did not changed them.

Finally, I also, in the second commit, renamed `getBlockPath` into `getBlockZonePaths` to avoid confusion.

### DIFF
--- a/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingSimulator.kt
+++ b/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingSimulator.kt
@@ -2,7 +2,6 @@ package fr.sncf.osrd.envelope_sim.etcs
 
 import fr.sncf.osrd.envelope.Envelope
 import fr.sncf.osrd.envelope_sim.EnvelopeSimContext
-import fr.sncf.osrd.sim_infra.api.Path
 import fr.sncf.osrd.sim_infra.api.TravelledPath
 import fr.sncf.osrd.utils.units.Offset
 
@@ -28,7 +27,7 @@ interface ETCSBrakingSimulator {
 }
 
 data class LimitOfAuthority(
-    val offset: Offset<Path>,
+    val offset: Offset<TravelledPath>,
     val speed: Double,
 ) {
     init {
--- a/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/pipelines/MaxSpeedEnvelope.kt
+++ b/core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/pipelines/MaxSpeedEnvelope.kt
@@ -18,7 +18,6 @@ import fr.sncf.osrd.envelope_sim.etcs.EndOfAuthority
 import fr.sncf.osrd.envelope_sim.etcs.LimitOfAuthority
 import fr.sncf.osrd.envelope_sim.overlays.EnvelopeDeceleration
 import fr.sncf.osrd.reporting.exceptions.OSRDError
-import fr.sncf.osrd.sim_infra.api.Path
 import fr.sncf.osrd.sim_infra.api.TravelledPath
 import fr.sncf.osrd.utils.units.Offset
 import fr.sncf.osrd.utils.units.meters
@@ -111,7 +110,7 @@ object MaxSpeedEnvelope {
         val cursor = EnvelopeCursor.backward(envelope)
         val limitsOfAuthority = mutableListOf<LimitOfAuthority>()
         while (cursor.findPartTransition(MaxSpeedEnvelope::increase)) {
-            val offset = Offset<Path>(cursor.position.meters)
+            val offset = Offset<TravelledPath>(cursor.position.meters)
             if (etcsRanges.contains(offset.distance)) {
                 limitsOfAuthority.add(
                     LimitOfAuthority(
--- a/core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/api/LoadedSignalingInfra.kt
+++ b/core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/api/LoadedSignalingInfra.kt
@@ -86,7 +86,7 @@ interface LoadedSignalInfra {
 interface BlockInfra {
     val blocks: StaticIdxSpace<Block>
 
-    fun getBlockPath(block: BlockId): StaticIdxList<ZonePath>
+    fun getBlockZonePaths(block: BlockId): StaticIdxList<ZonePath>
 
     fun getBlocksInZone(zone: ZoneId): StaticIdxList<Block>
 
--- a/core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/api/PathProperties.kt
+++ b/core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/api/PathProperties.kt
@@ -16,7 +16,9 @@ data class IdxWithOffset<T, U>(
     val offset: Offset<U>,
 )
 
-typealias IdxWithPathOffset<T> = IdxWithOffset<T, Path>
+typealias IdxWithBlockPathOffset<T> = IdxWithOffset<T, BlockPath>
+
+typealias IdxWithTravelledPathOffset<T> = IdxWithOffset<T, TravelledPath>
 
 data class TrackLocation(
     @get:JvmName("getTrackId") val trackId: TrackSectionId,
@@ -24,11 +26,10 @@ data class TrackLocation(
 )
 
 /**
- * A marker type for Length and Offset. "Path" *mostly* refers to "BlockPath", where start is the
- * beginning of the first block (NOT the real start of the train).
+ * A marker type for Length and Offset. In BlockPath, start refers to the beginning of the first
+ * block (NOT the real start of the train).
  */
-// TODO: rename this to BlockPath and make sure it's used appropriately
-sealed interface Path
+sealed interface BlockPath
 
 /**
  * A marker type for Length and Offset. In TravelledPath, start refers to the real start of the head
@@ -40,7 +41,7 @@ sealed interface TravelledPath
 interface PathProperties {
     fun getSlopes(): DistanceRangeMap<Double>
 
-    fun getOperationalPointParts(): List<IdxWithPathOffset<OperationalPointPart>>
+    fun getOperationalPointParts(): List<IdxWithTravelledPathOffset<OperationalPointPart>>
 
     fun getGradients(): DistanceRangeMap<Double>
 
@@ -65,10 +66,10 @@ interface PathProperties {
     @JvmName("getLength") fun getLength(): Distance
 
     @JvmName("getTrackLocationAtOffset")
-    fun getTrackLocationAtOffset(pathOffset: Offset<Path>): TrackLocation
+    fun getTrackLocationAtOffset(pathOffset: Offset<TravelledPath>): TrackLocation
 
     @JvmName("getTrackLocationOffset")
-    fun getTrackLocationOffset(location: TrackLocation): Offset<Path>?
+    fun getTrackLocationOffset(location: TrackLocation): Offset<TravelledPath>?
 
     fun <T> getRangeMapFromUndirected(
         getData: (chunkId: TrackChunkId) -> DistanceRangeMap<T>
@@ -87,8 +88,8 @@ interface PathProperties {
 fun buildPathPropertiesFrom(
     infra: RawSignalingInfra,
     chunks: DirStaticIdxList<TrackChunk>,
-    pathBeginOffset: Offset<Path>,
-    pathEndOffset: Offset<Path>,
+    pathBeginOffset: Offset<BlockPath>,
+    pathEndOffset: Offset<BlockPath>,
     routes: List<RouteId>? = null,
 ): PathProperties {
     val chunkPath = buildChunkPath(infra, chunks, pathBeginOffset, pathEndOffset)
@@ -126,7 +127,10 @@ fun makeTrackLocation(track: TrackSectionId, offset: Offset<TrackSection>): Trac
  * it's generic.
  */
 @JvmName("getTrackLocationOffsetOrThrow")
-fun getTrackLocationOffsetOrThrow(path: PathProperties, location: TrackLocation): Offset<Path> {
+fun getTrackLocationOffsetOrThrow(
+    path: PathProperties,
+    location: TrackLocation
+): Offset<TravelledPath> {
     return path.getTrackLocationOffset(location)
         ?: throw RuntimeException("Can't find location on path")
 }
--- a/core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/PathPropertiesImpl.kt
+++ b/core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/PathPropertiesImpl.kt
@@ -28,13 +28,13 @@ data class ChunkPath(
      * Offset of the head of the train when it starts its path, compared to the start of the first
      * element in `chunks`.
      */
-    val beginOffset: Offset<Path>,
+    val beginOffset: Offset<BlockPath>,
 
     /**
      * Offset of the head of the train when it ends its path, compared to the start of the first
      * element in `chunks`.
      */
-    val endOffset: Offset<Path>
+    val endOffset: Offset<BlockPath>
 ) {
     val length: Distance = endOffset.distance - beginOffset.distance
 }
@@ -48,7 +48,8 @@ data class PathPropertiesImpl(
         return getRangeMap { dirChunkId -> infra.getTrackChunkSlope(dirChunkId) }
     }
 
-    override fun getOperationalPointParts(): List<IdxWithPathOffset<OperationalPointPart>> {
+    override fun getOperationalPointParts():
+        List<IdxWithTravelledPathOffset<OperationalPointPart>> {
         return getElementsOnPath { dirChunkId ->
             infra.getTrackChunkOperationalPointParts(dirChunkId.value).map { opId ->
                 IdxWithOffset(opId, infra.getOperationalPointPartChunkOffset(opId))
@@ -145,8 +146,8 @@ data class PathPropertiesImpl(
         return chunkPath.endOffset - chunkPath.beginOffset
     }
 
-    override fun getTrackLocationAtOffset(pathOffset: Offset<Path>): TrackLocation {
-        val offset: Offset<Path> = pathOffset + chunkPath.beginOffset.distance
+    override fun getTrackLocationAtOffset(pathOffset: Offset<TravelledPath>): TrackLocation {
+        val offset: Offset<TravelledPath> = pathOffset + chunkPath.beginOffset.distance
         var lengthPrevChunks = 0.meters
         for (chunk in chunkPath.chunks) {
             val chunkLength = infra.getTrackChunkLength(chunk.value)
@@ -167,7 +168,7 @@ data class PathPropertiesImpl(
         throw RuntimeException("The given path offset is larger than the path length")
     }
 
-    override fun getTrackLocationOffset(location: TrackLocation): Offset<Path>? {
+    override fun getTrackLocationOffset(location: TrackLocation): Offset<TravelledPath>? {
         val offset =
             getOffsetOfTrackLocationOnChunks(infra, location, chunkPath.chunks) ?: return null
         if (offset < chunkPath.beginOffset || offset > chunkPath.endOffset) return null
@@ -263,13 +264,13 @@ data class PathPropertiesImpl(
 
     /**
      * Use the given function to get punctual data from a chunk, and concatenates all the values on
-     * the path
+     * the path. // TODO change type: types should be DirStaticIdxList<TrackChunk>
      */
     private fun <T> getElementsOnPath(
         getData: (chunk: DirTrackChunkId) -> Iterable<IdxWithOffset<T, TrackChunk>>
-    ): List<IdxWithPathOffset<T>> {
-        val res = ArrayList<IdxWithPathOffset<T>>()
-        var chunkOffset = Offset<Path>(0.meters)
+    ): List<IdxWithTravelledPathOffset<T>> {
+        val res = ArrayList<IdxWithBlockPathOffset<T>>()
+        var chunkOffset = Offset<BlockPath>(0.meters)
         for (chunk in chunkPath.chunks) {
             for ((element, offset) in getData.invoke(chunk)) {
                 val projectedOffset = projectPosition(chunk, offset)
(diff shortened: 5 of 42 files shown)


core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingSimulator.kt
```diff
@@ -28,7 +28,7 @@ interface ETCSBrakingSimulator {
 }
 
 data class LimitOfAuthority(
-    val offset: Offset<Path>,
+    val offset: Offset<BlockPath>,
```
bougue-pe: From the inconsistency with EndOfAuthority, it's suspicious.
As it's built from (and relates to) the envelope, it looks like it should be a `TraveledPath` (to propagate to some changes of this PR, of course... Trying to not comment those places).

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/api/LoadedSignalingInfra.kt
```diff
@@ -86,7 +86,7 @@ interface LoadedSignalInfra {
 interface BlockInfra {
     val blocks: StaticIdxSpace<Block>
 
-    fun getBlockPath(block: BlockId): StaticIdxList<ZonePath>
+    fun getBlockZonePaths(block: BlockId): StaticIdxList<ZonePath>
```
bougue-pe, line 89: After a little discussion, we have a (double) issue to tackle here :
* `blockPath` (not the type but the named concept) already exists, and is a path _inside_ a block, made of...
* `ZonePath`s which represent path _inside_ a zone, made of entry, exit and a sequence of `Node` (actually node-position, but it's redundant with next node and exit).

This contradicts current PR enforcing the idea that a `Object`Path is a _sequence_ of `Object`s.

---

My favorite action on this matter would be to rename:
* `ZonePath` -> `NodePath` (or `SingleZoneNodePath`?)
* `blockPath` -> `nodePathList` (or `singleBlockNodePathList`?)

```suggestion
    fun getNodePathList(block: BlockId): StaticIdxList<NodePath>
```

This is a tedious commit but OK IMO:
* `grep zonepath` and rename in core (not only the type but variables and methods), maybe a bit more?
* any existing `blockPath` in `dev` branch should probably disappear (and new ones may appear in other commits, representing a sequence of blocks).

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/api/PathProperties.kt
```diff
@@ -40,7 +39,7 @@ sealed interface TravelledPath
 interface PathProperties {
     fun getSlopes(): DistanceRangeMap<Double>
 
-    fun getOperationalPointParts(): List<IdxWithPathOffset<OperationalPointPart>>
+    fun getOperationalPointParts(): List<IdxWithBlockPathOffset<OperationalPointPart>>
```
bougue-pe: It looks like we should work with Offset<`TravelledPath`> here.

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/PathPropertiesImpl.kt
```diff
@@ -28,13 +28,13 @@ data class ChunkPath(
...
      */
-    val beginOffset: Offset<Path>,
+    val beginOffset: Offset<BlockPath>,
 
     /**
      * Offset of the head of the train when it ends its path, compared to the start of the first
      * element in `chunks`.
      */
-    val endOffset: Offset<Path>
+    val endOffset: Offset<BlockPath>
```
bougue-pe, line 37: Those are Offset<`DirStaticIdxList<TrackChunk>`>...
We may typealias `DirStaticIdxList<TrackChunk>` to `ChunkSequence` (as I guess `ChunkPath` is not available at that point, and convenient to keep).
To be propagated again.

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/PathPropertiesImpl.kt
```diff
@@ -299,8 +299,8 @@ data class PathPropertiesImpl(
      * Keeps only the elements that are not outside the path, and shift the offsets to start at 0
      */
     private fun <T> filterAndShiftElementsOnPath(
-        res: List<IdxWithPathOffset<T>>
-    ): List<IdxWithPathOffset<T>> {
+        res: List<IdxWithBlockPathOffset<T>>
+    ): List<IdxWithBlockPathOffset<T>> {
```
bougue-pe: To be confirmed once propagated, but it looks like those are Offset<`ChunkSequence`> in and Offset<`TraveledPath`> out.

EDIT: relates to discussion above (will probably be resolved at the same time)

bougue-pe (commented):
🙏 Thanks for diving on it 💣 

Please some other @OpenRailAssociation/osrd-core maintainer read my comments 😅 

I didn't review all, but there is some thinking already, and it can be incremental (one commit only changing a type and propagating it)... The work may even be separated in one or mutliple other PRs indeed.

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingSimulator.kt
eckter: Agree on `TravelledPath` there

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/pipelines/MaxSpeedEnvelope.kt
```diff
@@ -111,7 +111,7 @@ object MaxSpeedEnvelope {
         val cursor = EnvelopeCursor.backward(envelope)
         val limitsOfAuthority = mutableListOf<LimitOfAuthority>()
         while (cursor.findPartTransition(MaxSpeedEnvelope::increase)) {
-            val offset = Offset<Path>(cursor.position.meters)
+            val offset = Offset<BlockPath>(cursor.position.meters)
```
eckter: `TravelledPath` here as well

Generally speaking, anything directly referencing envelopes is `TravelledPath`

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/api/LoadedSignalingInfra.kt
eckter: I agree on renaming it, but I'm not sure `NodePath` is better. If anything, the boundaries are detectors. It's still a path though (as in related to a sequence of tracks and other stuff). And it's strongly paired with a single zone. 

I don't have any better suggestion for now, but maybe we could discuss it IRL

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/api/PathProperties.kt
eckter: Agree! But I got confused and was about to say the opposite. This specific class is tricky to deal with.

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
```diff
@@ -92,7 +92,7 @@ private fun makeEndOfPathStop(
 fun getSignalOffsets(
     infra: FullInfra,
     zonePaths: List<StaticIdx<ZonePath>>,
-    pathStartOffset: Offset<Path>,
+    pathStartOffset: Offset<TravelledPath>,
```
eckter: Technically in this context this offset refers to the start of the zone paths. But it's built referencing the start of the first route.

Either way, it's not a travelled path offset.

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/path_properties/PathPropResponse.kt
```diff
@@ -32,7 +32,7 @@ data class OperationalPointResponse(
     val id: String,
     val part: OperationalPointPartResponse,
     val extensions: OperationalPointExtensions?,
-    val position: Offset<Path>,
+    val position: Offset<BlockPath>,
```
eckter: Related to the comment above as it's just propagated, but this is a travelled path.

Generally speaking, anything exposed to the public API is very likely to be travelled paths. Other services aren't really equipped to deal with the distinction between the two types.

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelope.kt
```diff
@@ -71,7 +71,7 @@ interface InfraExplorerWithEnvelope : InfraExplorer {
     fun getFullSpacingRequirements(): List<ResultTrain.SpacingRequirement>
 
     /** Returns the length of the simulated section of the path */
-    fun getSimulatedLength(): Length<Path>
+    fun getSimulatedLength(): Length<BlockPath>
```
eckter: :melting_face: 

So that's a `TravelledPath`, but it should be fixed in the other bugfix PR (we can expect a conflict there)

core/src/main/kotlin/fr/sncf/osrd/utils/OffsetConversions.kt
```diff
@@ -24,8 +26,8 @@ fun trainPathBlockOffset(
     blockInfra: BlockInfra,
     blockPath: StaticIdxList<Block>,
     chunkPath: ChunkPath
-): Offset<Path> {
-    val zonePaths = blockPath.flatMap { blockInfra.getBlockPath(it) }
+): Offset<TravelledPath> { //
```
eckter:
```suggestion
): Offset<TravelledPath> {
```

core/src/main/kotlin/fr/sncf/osrd/utils/OffsetConversions.kt
```diff
@@ -37,8 +39,8 @@ fun trainPathZonePathOffset(
     infra: RawInfra,
     zonePaths: List<ZonePathId>,
     chunkPath: ChunkPath
-): Offset<Path> {
-    var prevChunksLength = Offset<Path>(0.meters)
+): Offset<TravelledPath> {
```
eckter: These functions *could* have their distinct return types (zone path / chunk path), but not in this PR

eckter (commented):
Very nice job!

I still commented every possible mismatch, but we may fix them in a different PR

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/api/LoadedSignalingInfra.kt
Khoyo: Soooo... ZonePath, contrarily to the two other paths, is an infra object, so it's not really the same kind of Path. I doesn't have any kind of offset for example, since it's not linked to a train, it's just a potential path that could be taken through a zone. It's more like a Route or a Block in that sense, and Routes and Blocks aren't considered paths.

If we find a new name, I think we should focus on dropping Path from it... ;)

> blockPath -> nodePathList (or singleBlockNodePathList?)

A block path is a list of blocks (and offsets), which is very much *not* just a list of zonePath. You can have two identical paths through zones, but that don't use the same signaling systems and thus not the same blocks.

A *Route* is a nodePathList, but I don't think that's a better name...

I just realized you meant "blockPath" the path through a block, not blockPath the train path in terms of blocks. Is it really a thing?

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/etcs/ETCSBrakingSimulator.kt
louisgreiner: Done in commit #1

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/api/PathProperties.kt
louisgreiner: Done in commit #2

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/PathPropertiesImpl.kt
louisgreiner: Should we spread the "sequence" naming to all these generic classes ? `DirStaticIdxList` -> `Sequence`

from `fr/sncf/osrd/utils/indexing/DirStaticIdxCollectionInterfaces.kt` for instance

core/envelope-sim/src/main/kotlin/fr/sncf/osrd/envelope_sim/pipelines/MaxSpeedEnvelope.kt
louisgreiner: Done in comment #3

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
louisgreiner: Should I revert it to `Offset<BlockPath>` then?

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelope.kt
louisgreiner: Are you sure? Or maybe these comments are misleading : https://github.com/OpenRailAssociation/osrd/blob/881fe17efc53b6e7e43b401df0949bfbf48be80d/core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/interfaces/BlockAvailabilityInterface.kt#L31-L47

eckter: It's currently bugged. Ignore this one and it will be fixed in the other PR (https://github.com/OpenRailAssociation/osrd/pull/11210).

The type in `getAvailability`'s input is correct (`BlockPath`). But `getSimulatedLength()` actually returns a `TravelledPath`. There's a missing conversion.

core/src/main/java/fr/sncf/osrd/standalone_sim/SafetySpeed.kt
louisgreiner: Done in commit 5

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/path_properties/PathPropResponse.kt
louisgreiner: Done

core/src/main/kotlin/fr/sncf/osrd/utils/OffsetConversions.kt
louisgreiner: Done

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/PathPropertiesImpl.kt
bougue-pe: Changing those to `TravelledPath` is as wrong as `BlockPath`...
So here are options I have:
1. Really use (or typealias) `DirStaticIdxList<TrackChunk>` (may be quite long depending on propagation)
2. Add a TODO comment, and keep `TravelledPath` or `BlockPath` (big preference for the later as it says "more" that it's wrong) -> And if changing back to `BlockPath` is an issue, we may have a missing conversion to apply.

EDIT: `filterAndShiftElementsOnPath` signature is way more expressive with a `BlockPath` here for solution 2 (true for other places impacted).

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorer.kt
```diff
@@ -196,7 +196,8 @@ private class InfraExplorerImpl(
         val blockPathProperties = path.withRoutes(listOf(route))
         pathPropertiesCache[getCurrentBlock()] = blockPathProperties
```
bougue-pe, line 198: To clarify a bit below:
```kt
        val blockLength = blockInfra.getBlockLength(getCurrentBlock())
        val endOffset: Offset<Block> = if (length == null) blockLength else offset.plus(length)
        if (offset.distance == 0.meters && endOffset == blockLength) {
            return blockPathProperties
        }
        // In that case, start of the block is start of the travelled path
        return PathPropertiesView(blockPathProperties, offset.cast(), endOffset.cast())
```

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingResultConverter.kt
```diff
@@ -124,7 +124,7 @@ private fun makePendingOPWaypoint(
 private fun makePendingUserDefinedWaypoint(
     infra: RawSignalingInfra,
     path: PathProperties,
-    pathOffset: Offset<Path>
+    pathOffset: Offset<BlockPath>
```
bougue-pe: This must match `makePendingOPWaypoint`:
```suggestion
    pathOffset: Offset<TravelledPath>
```
And it propagates correctly to `PathProperties.getTrackLocationAtOffset` if I'm correct.

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingResultConverter.kt
```diff
@@ -78,7 +78,7 @@ private fun makeUserDefinedWaypoints(
         offsets.add(waypoint.offset)
     }
     val res = ArrayList<PathWaypointResult>()
-    var lengthPrevBlocks: Offset<Path> = Offset(0.meters)
+    var lengthPrevBlocks: Offset<BlockPath> = Offset(0.meters)
```
bougue-pe: Having an `Offset<TravelledPath>` makes more sense here (and so matches the new signature of `makePendingUserDefinedWaypoint`), so below would become:
```kt
    val startFirstRange = rawPath.ranges[0].start
    var startBlockOffset = Offset<TravelledPath>(-startFirstRange.distance)
    for (blockRange in rawPath.ranges) {
        for (waypoint in userDefinedWaypointsPerBlock.getOrDefault(blockRange.edge, ArrayList())) {
            if (blockRange.start <= waypoint && waypoint <= blockRange.end) {
                val pathOffset = startBlockOffset + waypoint.distance
                res.add(makePendingUserDefinedWaypoint(infra, path, pathOffset))
            }
        }
        startBlockOffset += blockInfra.getBlockLength(blockRange.edge).distance
    }
```

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/standalone_sim/SimulationRequest.kt
```diff
@@ -137,8 +137,8 @@ class MarginValueAdapter {
 }
 
 class SimulationPowerRestrictionItem(
-    val from: Offset<Path>,
-    val to: Offset<Path>,
+    val from: Offset<BlockPath>,
+    val to: Offset<BlockPath>,
```
bougue-pe:
```suggestion
    val from: Offset<TravelledPath>,
    val to: Offset<TravelledPath>,
```

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/api/LoadedSignalingInfra.kt
bougue-pe: OK to deal with the rename(s) in another PR and resolve this.
Not sure how to track it: link to this conversation in the parent issue?

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/PathPropertiesImpl.kt
eckter: Definitely not `TravelledPath` here

Introducing new types for the different object types would be best, but maybe out of scope for this PR. `BlockPath` isn't so bad in the meantime.

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/api/LoadedSignalingInfra.kt
louisgreiner: Tracked here : https://github.com/OpenRailAssociation/osrd/issues/11259

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/PathPropertiesImpl.kt
louisgreiner: Done in comment 6, and tracked by `"TODO change type"` in the code. See also [new issue](https://github.com/OpenRailAssociation/osrd/issues/11259#issue-2948941652)

core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/impl/PathPropertiesImpl.kt
louisgreiner: Same as above, Done in comment 6, and tracked by "TODO change type" in the code. See also https://github.com/OpenRailAssociation/osrd/issues/11259#issue-2948941652

core/src/main/kotlin/fr/sncf/osrd/api/api_v2/standalone_sim/SimulationRequest.kt
louisgreiner: Done in [fix mistyping from makePendingUserDefinedWaypoint](https://github.com/OpenRailAssociation/osrd/pull/11188/commits/cdd68864a1d6d1100ddc87d31698b52f3c29caac)

core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingResultConverter.kt
louisgreiner: Done in [fix mistyping from makePendingUserDefinedWaypoint](https://github.com/OpenRailAssociation/osrd/pull/11188/commits/cdd68864a1d6d1100ddc87d31698b52f3c29caac)

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorer.kt
louisgreiner: Done in [fix mistyping from makePendingUserDefinedWaypoint](https://github.com/OpenRailAssociation/osrd/pull/11188/commits/cdd68864a1d6d1100ddc87d31698b52f3c29caac)

bougue-pe (approved):
🎉 You reached the end of this part of the refactor!

Thanks again for the work on this, it will be much more understandable to work with those offsets 🙏

eckter (approved):

PR #11192, opened 2025-03-19
editoast: move hash function to PathfindingInput impl

hamz2a:
Refactor the code to be more idiomatic Rust by moving the hashing logic into the PathfindingInput struct, improving organization and readability.


woshilapin (approved):
Seems reasonable. Thank you!

flomonster:
For consistency, can we do the same for simulation and path projection?

hamz2a:
> For consistency, can we do the same for simulation and path projection?

Yes, absolutely!

PR #11196, opened 2025-03-19
Remove InternalError dependency for Model's Update traits

leovalais:
* Updates `Update`, `Save`, `Patch` and `UpdateBatch`
* Propagates the change where needed, either by coercing to InternalError
  or by adding new error variants
* Rework `last_modification` for projects, studies and scenario to:
    1. actualize the date in a transaction
    2. update the project if a study is modified, and the study if a
       scenario is modified
* Fixes API limits of `Connection::transaction` by not exposing
  `diesel::result::Error`
* Adds the `exists_or_fail` function
* Add the parsing of foreign key violations in model's error
* Rework project image suppression to be:
    1. atomic by nesting everything into a transaction
    2. simplifying the behavior of `update_and_prune` and
       `delete_and_prune`
    3. fixing the database allowing a document to be deleted even if it
       was referenced by a project
    4. removing the unicity constraint on the image FK
    5. adding some missing tests for these functions
* Simplifies the workflow of a rolling stock test
* Makes a towed rolling stock test **less** (but *still*) flaky
* Removes the `check_study_project`, `check_study_project_scenario` kind
  of functions:
    1. they fit ill in the new workflow
    2. the check will be removed once we fix our URL scheme
    3. replaced by plain `if`s

### DIFF
--- a/editoast/src/models/projects.rs
+++ b/editoast/src/models/projects.rs
@@ -1,17 +1,20 @@
 use chrono::NaiveDateTime;
 use chrono::Utc;
+use diesel_async::scoped_futures::ScopedBoxFuture;
+use diesel_async::scoped_futures::ScopedFutureExt;
 use editoast_derive::Model;
 use serde::Deserialize;
 use serde::Serialize;
 use utoipa::ToSchema;
 
-use crate::error::Result;
+use crate::error::InternalError;
 use crate::models::prelude::*;
 use crate::models::Document;
 use crate::models::Study;
 use crate::models::Tags;
 use crate::views::projects::ProjectError;
 use crate::SelectionSettings;
+use editoast_models::model;
 use editoast_models::DbConnection;
 
 editoast_common::schemas! {
@@ -21,6 +24,7 @@ editoast_common::schemas! {
 #[derive(Clone, Debug, Serialize, Deserialize, Model, ToSchema, PartialEq)]
 #[model(table = editoast_models::tables::project)]
 #[model(gen(ops = crud, list))]
+#[model(error = Error)]
 pub struct Project {
     pub id: i64,
     pub name: String,
@@ -36,15 +40,58 @@ pub struct Project {
     pub image: Option<i64>,
 }
 
+#[derive(Debug, thiserror::Error, derive_more::From)]
+pub enum Error {
+    #[error(transparent)]
+    #[from(forward)]
+    Database(model::Error),
+    #[error(transparent)]
+    Legacy(#[from] InternalError),
+}
+
+#[tracing::instrument(skip(conn), ret, err)]
+async fn try_delete_document(conn: &DbConnection, doc_id: i64) -> Result<(), Error> {
+    let res = conn
+        .transaction(|mut conn| {
+            async move {
+                match Document::delete_static(&mut conn, doc_id).await {
+                    Ok(false) => unreachable!(
+                        "cannot happen as the Document has to be there because of the FK on `image`"
+                    ),
+                    Ok(true) => Ok(()),
+                    // We want the delete to occur in a transaction in order to rollback it if the deletion fails.
+                    // The deletion can fail if the document is still used by another project (FK violation). This
+                    // is acceptable, it's what this function does.
+                    // However, if a FK violation occurs, the transaction must rolloback otherwise each subsequent
+                    // query will fail. If the violation occurs, `e` is an `Err`, therefore we return it in order
+                    // to let `transaction` rollback. We then match on the error below in order to accept the
+                    // FK violation, which is not an error in our workflow.
+                    Err(e) => Err(e),
+                }
+            }
+            .scope_boxed()
+        })
+        .await;
+    match res {
+        Ok(_) => Ok(()),
+        Err(model::Error::ForeignKeyViolation { constraint })
+            if constraint == "project_image_id_fkey" =>
+        {
+            Ok(())
+        }
+        Err(e) => Err(self::Error::from(e)),
+    }
+}
+
 impl Project {
     /// This function takes a filled project and update to now the last_modification field
-    pub async fn update_last_modified(&mut self, conn: &mut DbConnection) -> Result<()> {
+    pub async fn update_last_modified(&mut self, conn: &mut DbConnection) -> Result<(), Error> {
         self.last_modification = Utc::now().naive_utc();
         self.save(conn).await?;
         Ok(())
     }
 
-    pub async fn studies_count(&self, conn: &mut DbConnection) -> Result<u64> {
+    pub async fn studies_count(&self, conn: &mut DbConnection) -> Result<u64, InternalError> {
         let project_id = self.id;
         let studies_count = Study::count(
             conn,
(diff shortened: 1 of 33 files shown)


editoast/src/models/projects.rs
```diff
@@ -36,15 +40,63 @@ pub struct Project {
...
+    #[from(forward)]
+    Database(model::Error),
+    #[error(transparent)]
+    Legacy(#[from] InternalError),
+}
+
+#[tracing::instrument(skip(conn), ret, err)]
+async fn try_delete_document(conn: &DbConnection, doc_id: i64) -> Result<(), Error> {
+    #[derive(Debug, thiserror::Error, derive_more::From)]
+    enum Failible {
```
woshilapin:
```suggestion
    enum Faillible {
```

editoast/src/models/projects.rs
```diff
@@ -36,15 +40,63 @@ pub struct Project {
...
+        #[from(forward)]
+        Db(model::Error),
+    }
+    let res = conn
+        .transaction(|mut conn| {
+            async move {
+                match Document::delete_static(&mut conn, doc_id).await {
+                    Ok(false) => unreachable!(
+                    "cannot happen as the Document has to be there because of the FK on `image`"
+                ),
```
woshilapin: Formatting (indentation) seems off here.

editoast/src/models/projects.rs
```diff
@@ -36,15 +40,63 @@ pub struct Project {
...
+                    Ok(true) => Ok(()),
+                    Err(model::Error::ForeignKeyViolation { constraint })
+                        if constraint == "project_image_id_fkey" =>
+                    {
+                        // We can't just return Ok(()) here because we want the transaction to rollback
+                        // which it does not do if we return Ok(()).
+                        // We want the transaction to rollback because if an error occurs in a transaction
+                        // (which is the case here, FK violation), any subsequent query will fail automatically.
+                        Err(Failible::FailedButOk)
```
woshilapin: I'm not sure that I understand why you do need the intermediate representation `Faillible`. What about returning `Ok(())` directly here? And 2 lines below, you should be able to return `Err(e)` directly. Then line 85, you should be able to only do `return res.map_err(Self::Error::from)`. Or did I miss something?

editoast/src/models/projects.rs
```diff
@@ -36,15 +40,63 @@ pub struct Project {
...
     /// This function takes a filled project and update to now the last_modification field
-    pub async fn update_last_modified(&mut self, conn: &mut DbConnection) -> Result<()> {
+    pub async fn update_last_modified(&mut self, conn: &mut DbConnection) -> Result<(), Error> {
         self.last_modification = Utc::now().naive_utc();
         self.save(conn).await?;
         Ok(())
     }
 
-    pub async fn studies_count(&self, conn: &mut DbConnection) -> Result<u64> {
+    pub async fn studies_count(&self, conn: &mut DbConnection) -> Result<u64, InternalError> {
```
woshilapin, line 94: Should it be `Error` instead of `InternalError`?

editoast/src/models/projects.rs
leovalais: If we return `Ok(())` the `transaction` function won't detect the error and won't rollback (or in this case `ROLLBACK TO SAVEPOINT` since we're nested in another transaction). The `transaction` function only rollbacks if we return an `Err` variant. We want this transaction to rollback because if an error occurs in a transaction Postgres won't run subsequent queries. 

So we create a save point (through a nested `transaction` call) that we rollback if we find a foreign key violation (image used by another project).

editoast/src/models/projects.rs
leovalais: `models::prelude::Count::count` still converts to `InternalError`, `List` and `Count` are the next step, then `Retrieve` and `Exist` to finish

But eventually, it should be `Error` indeed

editoast/src/models/projects.rs
leovalais: 8d9bf9e02

editoast/src/models/projects.rs
leovalais: 9d363a1df

editoast/src/models/projects.rs
```diff
@@ -36,15 +40,63 @@ pub struct Project {
...
+                }
+            }
+            .scope_boxed()
+        })
+        .await;
+    match res {
+        Ok(()) | Err(Faillible::FailedButOk) => Ok(()),
+        Err(Faillible::Db(e)) => Err(self::Error::from(e)),
+    }
+}
```
woshilapin: Why not handling the specific error at the very end and avoiding the entire `Faillible`.

```rust
async fn try_delete_document(conn: &DbConnection, doc_id: i64) -> Result<(), Error> {
    let res = conn
        .transaction(|mut conn| {
            async move {
                match Document::delete_static(&mut conn, doc_id).await {
                    Ok(false) => unreachable!(
                        "cannot happen as the Document has to be there because of the FK on `image`"
                    ),
                    Ok(true) => Ok(()),
                    Err(e) => Err(model::Error::from(e)),
                }
            }
            .scope_boxed()
        })
        .await;
    match res {
        Ok(()) => Ok(()),
        Err(model::Error::ForeignKeyViolation { constraint })
            if constraint == "project_image_id_fkey" =>
        {
            // In this case, the transaction has rollbacked because the document
            // is used elsewhere
            Ok(())
        }
        Err(e) => Err(self::Error::from(e)),
    }
}
```

editoast/src/models/projects.rs
```diff
@@ -36,15 +40,63 @@ pub struct Project {
...
+                        Err(Faillible::FailedButOk)
+                    }
+                    Err(e) => Err(Faillible::Db(e)),
+                }
+            }
+            .scope_boxed()
+        })
+        .await;
+    match res {
+        Ok(()) | Err(Faillible::FailedButOk) => Ok(()),
```
woshilapin: Should we really return an `Ok` when `FailedButOk`? If I understand correctly, we're trying to remove a document, and it cannot be removed because it is still linked to another project. So indeed, it's "fine" to not remove it, but the client of that operation might be surprised to still see this document around. Maybe we need a special error case to return in this case: `StillUsed`, or `DependOnExistingProject`? It's assuming that I correctly understood the situation.

Because I guess that we're using that function when trying to remove a project, and if we have a special error case for that, then the `remove_project` function will be able to handle correctly this case by saying "I successfully removed the project, but not the document, since it's used somewhere else... and that's fine!".

editoast/src/models/projects.rs
```diff
@@ -54,44 +106,91 @@ impl Project {
...
-        Ok(project)
+        new_doc_id: Option<i64>,
+    ) -> Result<(), Error> {
+        conn.transaction(|mut conn| {
+            async move {
+                let old_doc_id = self.image;
+                self.image = new_doc_id;
+                self.save(&mut conn).await?;
+                if let Some(doc_id) = old_doc_id {
+                    if new_doc_id.is_none() || new_doc_id.is_some_and(|new| new != doc_id) {
```
woshilapin: I do not understand why we need to check for `new_doc_id.is_none()`? We want to delete the document if the old one is `Some` (check done a line above) and it is different from the new one only (which is the second condition of this OR test), right?

editoast/src/models/projects.rs
```diff
@@ -54,44 +106,91 @@ impl Project {
...
 
-    pub async fn delete_and_prune_document(
-        conn: &mut DbConnection,
+    /// Deletes a project and prunes the image if it is not used by another project
+    #[tracing::instrument(skip(conn), ret, err)]
+    pub async fn delete_and_prune_document(self, conn: &mut DbConnection) -> Result<(), Error> {
+        conn.transaction(|mut conn| {
+            async move {
+                if !self.delete(&mut conn).await? {
+                    tracing::error!(
```
woshilapin: nitpick: not really a problem, is it?
```suggestion
                    tracing::warn!(
```

editoast/src/models/projects.rs
```diff
@@ -173,4 +272,129 @@ pub mod tests {
...
+        );
+
+        // project2 -> image1, project3 -> image2, project4 -> nothing
+
+        let p3_id = project3.id;
+        project3
+            .delete_and_prune_document(&mut db_pool.get_ok())
+            .await
+            .expect("should work");
+        dbg!(image2.id);
```
woshilapin: Can be removed.

editoast/src/views/projects.rs
```diff
@@ -353,13 +375,37 @@ async fn patch(
     if !authorized {
         return Err(AuthorizationError::Forbidden.into());
     }
-    let conn = &mut db_pool.get().await?;
+
+    let mut conn = db_pool.get().await?;
     if let Some(Some(image)) = form.image {
-        check_image_content(conn, image).await?;
+        check_image_content(&mut conn, image).await?;
+        form.image = None;
```
woshilapin: I don't understand this line. Wouldn't that always remove the image from the project? The local variable `image` in this `if` is only used in `check_image_content` and then forget after that... meaning that `form.image` is the only reference to it. And this line remove this only reference. Or I missed something, but maybe the code is not clear enough.

editoast/src/views/projects.rs
```diff
@@ -353,13 +375,37 @@ async fn patch(
...
+                let mut project = project_changeset
+                    .update_or_fail(&mut conn, project.id, || ProjectError::NotFound {
+                        project_id: project.id,
+                    })
+                    .await?;
+                if let Some(new_doc_id) = image {
+                    project
+                        .update_and_prune_document(&mut conn, new_doc_id)
+                        .await?;
+                }
```
woshilapin: With the code above always making `None` of `form.image` and then the `image` variable initialized from it, I would expect this `if` block to never be executed.

woshilapin (commented):
Finally got the time to read this PR. Your oral explanations helped a lot reviewing this.

editoast/src/models/projects.rs
leovalais: agreed 44e570246

editoast/src/models/projects.rs
leovalais: 🐒 652f93a46

editoast/src/models/projects.rs
leovalais: If we remove the image of the project and no other project uses it, it becomes dangling. I changed the condition, maybe it's more straightforward like this: 050f845b9

woshilapin: Not sure I like it better (unwrap 😨). To me, you want to return `false`, only when both `old_doc_id` and `new_doc_id` are `Some` and contains the same value. Something like this should do it.

```rust
if new_doc_id.zip(old_doc_id).map(|(new, old)| new != old).unwrap_or(true) {
```

leovalais: Unless I'm wrong this is essentially equivalent to `old ≠ new`, what about 759319dea to get rid of the unwrap?

editoast/src/models/projects.rs
leovalais: Good catch, much better indeed! 7ee267275

editoast/src/models/projects.rs
leovalais: In this case it's not an error per se imo since the project deletion succeeded, more like a note about a side-effect. What do we do in the endpoint when this case happens?

editoast/src/views/projects.rs
leovalais: You're right, I think I fixed the problem and added a test af92c9eb1

editoast/src/models/projects.rs
woshilapin: Much better way of expressing what I thought I understood. So indeed, fine like it is.

editoast/src/models/projects.rs
woshilapin: Yes, thanks... except for the weird `e => e` that also match with the pattern `Ok(true)`. Although this is fine (because you indeed want to forward an `Ok` and the lines below handles this correctly), I feel it is a bit misleading and we probably want to make the `Ok(true)` an independent pattern match, what do you think?

editoast/src/models/projects.rs
woshilapin: Resolving, I believe this is solved in https://github.com/OpenRailAssociation/osrd/pull/11196#discussion_r2016997967.

woshilapin (approved):
Nice move forward.

front/public/locales/fr/errors.json
```diff
@@ -64,7 +64,8 @@
       "ObjectNotFound": "{{obj_type}} '{{obj_id}}' introuvable dans le cache de l'infrastructure"
     },
     "infra": {
-      "NotFound": "",
+      "Database": "Erreur interne",
+      "NotFound": "Infrastructure {{infra_id}} n'existe pas",
```
```suggestion
      "NotFound": "Infrastructure '{{infra_id}}' n'existe pas",
```

SharglutDev (approved):
Lgtm ✅ You can resolve the comment when fixed :)

PR #11197, opened 2025-03-19
editoast: authz api with openfga

sim51:
This PR implements the API of ticket #10579 

To unmock the front, this task is needed: 

- [x] API `subjects`

Tasks for the futur : 

- [ ] While revoking an owner, check at least one owner will remains 
   - [ ]  for now at least, can't self-remove which prevents orphan resources
- [ ] Check the size of the FGA queries
- [x] Check that adding a grant that already exists doesn't fail
- [x] Check that revoking a grant that doesn't exists doesn't fail
- [ ] Check transactional feature (for batch or when doing deletes & writes)
- [ ] Handle subject (ie. User & Group) instead of just User
- [ ] Adding multi-check request on FGA client

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -140,6 +140,7 @@ services:
       TELEMETRY_ENDPOINT: "http://jaeger:4317"
       OSRD_MQ_URL: "amqp://osrd:password@osrd-rabbitmq:5672/%2f"
       EDITOAST_OPENFGA_URL: "http://openfga:8091"
+      EDITOAST_ENABLE_AUTHORIZATION: ${EDITOAST_ENABLE_AUTHORIZATION:-false}
     command:
       - /bin/sh
       - -c
--- a/editoast/editoast_authz/src/authorizer.rs
+++ b/editoast/editoast_authz/src/authorizer.rs
@@ -57,6 +57,126 @@ impl<S: StorageDriver> Authorizer<S> {
     pub async fn check_roles(&self, roles: HashSet<Role>) -> Result<bool, Error<S::Error>> {
         self.regulator.check_roles(self.user_id, roles).await
     }
+
+    pub async fn check_infra_grant_reader(&self, infra_id: i64) -> Result<bool, Error<S::Error>> {
+        self.regulator
+            .check_infra_grant_reader(self.user_id, infra_id)
+            .await
+    }
+
+    pub async fn check_infra_grant_writer(&self, infra_id: i64) -> Result<bool, Error<S::Error>> {
+        self.regulator
+            .check_infra_grant_writer(self.user_id, infra_id)
+            .await
+    }
+
+    pub async fn check_infra_grant_owner(&self, infra_id: i64) -> Result<bool, Error<S::Error>> {
+        self.regulator
+            .check_infra_grant_owner(self.user_id, infra_id)
+            .await
+    }
+
+    pub async fn check_infra_privilege_can_read(
+        &self,
+        infra_id: i64,
+    ) -> Result<bool, Error<S::Error>> {
+        self.regulator
+            .check_infra_privilege_can_read(self.user_id, infra_id)
+            .await
+    }
+
+    pub async fn check_infra_privilege_can_share_read(
+        &self,
+        infra_id: i64,
+    ) -> Result<bool, Error<S::Error>> {
+        self.regulator
+            .check_infra_privilege_can_share_read(self.user_id, infra_id)
+            .await
+    }
+
+    pub async fn check_infra_privilege_can_write(
+        &self,
+        infra_id: i64,
+    ) -> Result<bool, Error<S::Error>> {
+        self.regulator
+            .check_infra_privilege_can_share_write(self.user_id, infra_id)
+            .await
+    }
+
+    pub async fn check_infra_privilege_can_share_ownership(
+        &self,
+        infra_id: i64,
+    ) -> Result<bool, Error<S::Error>> {
+        self.regulator
+            .check_infra_privilege_can_share_ownership(self.user_id, infra_id)
+            .await
+    }
+
+    pub async fn grant_infra_reader(
+        &self,
+        user_id: i64,
+        infra_id: i64,
+    ) -> Result<(), Error<S::Error>> {
+        self.regulator
+            .grant_infra_reader(self.user_id, user_id, infra_id)
+            .await?;
+        Ok(())
+    }
+
+    pub async fn grant_infra_writer(
+        &self,
+        user_id: i64,
+        infra_id: i64,
+    ) -> Result<(), Error<S::Error>> {
+        self.regulator
+            .grant_infra_writer(self.user_id, user_id, infra_id)
+            .await?;
+        Ok(())
+    }
+
+    pub async fn grant_infra_owner(
+        &self,
+        user_id: i64,
+        infra_id: i64,
+    ) -> Result<(), Error<S::Error>> {
+        self.regulator
+            .grant_infra_owner(self.user_id, user_id, infra_id)
+            .await?;
+        Ok(())
+    }
+
+    pub async fn revoke_infra_reader(
+        &self,
+        user_id: i64,
+        infra_id: i64,
+    ) -> Result<(), Error<S::Error>> {
+        self.regulator
+            .revoke_infra_reader(self.user_id, user_id, infra_id)
+            .await?;
+        Ok(())
+    }
+
+    pub async fn revoke_infra_writer(
+        &self,
+        user_id: i64,
+        infra_id: i64,
+    ) -> Result<(), Error<S::Error>> {
+        self.regulator
+            .revoke_infra_writer(self.user_id, user_id, infra_id)
+            .await?;
+        Ok(())
+    }
+
+    pub async fn revoke_infra_owner(
+        &self,
+        user_id: i64,
+        infra_id: i64,
+    ) -> Result<(), Error<S::Error>> {
+        self.regulator
+            .revoke_infra_owner(self.user_id, user_id, infra_id)
+            .await?;
+        Ok(())
+    }
 }
 
 impl<S: StorageDriver> std::fmt::Debug for Authorizer<S> {
@@ -89,6 +209,61 @@ mod tests {
         counter: Arc<RwLock<i64>>,
         users: Arc<Mutex<HashMap<UserIdentity, i64>>>,
         groups: Arc<Mutex<HashMap<GroupName, i64>>>,
+    }
+
+    #[tokio::test]
+    async fn check_user_grants() {
+        let user_identity = || "toto".to_owned();
+        let user = || UserInfo {
+            identity: user_identity(),
+            name: "Sir Toto, the One and Only".to_owned(),
+        };
+        let regulator = Regulator::new(crate::openfga!(), MockAuthDriver::default());
+        let regulator = move || regulator.clone();
+
+        let user_id = regulator()
+            .driver
+            .ensure_user(&user())
+            .await
+            .expect("toto should be created")
+            .id;
+
+        let authorizer = Authorizer::try_initialize(user_identity(), regulator())
+            .await
+            .unwrap();
+
+        authorizer
+            .regulator
+            .grant_infra_reader_unchecked(authorizer.user_id(), 1)
+            .await
+            .expect("Update grants should be successful");
+        let is_reader = authorizer
+            .check_infra_grant_reader(1)
+            .await
+            .expect("Check grants should be successful");
+        assert_eq!(is_reader, true,);
+
+        authorizer
+            .regulator
+            .grant_infra_writer_unchecked(user_id, 1)
+            .await
+            .expect("Update grants should be successful");
+        let is_writer = authorizer
+            .check_infra_grant_writer(1)
+            .await
+            .expect("Check grants should be successful");
+        assert_eq!(is_writer, true,);
+
+        authorizer
+            .regulator
+            .grant_infra_writer_unchecked(user_id, 1)
+            .await
+            .expect("Update grants should be successful");
+        let is_owner = authorizer
+            .check_infra_grant_writer(1)
+            .await
+            .expect("Check grants should be successful");
+        assert_eq!(is_owner, true,);
     }
 
     #[tokio::test]
@@ -429,6 +604,11 @@ mod tests {
                     .into_iter()
                     .map(|(name, id)| Ok((id, GroupInfo { name }))),
             ))
+        }
+
+        async fn infra_exists(&self, _infra_id: i64) -> Result<bool, Self::Error> {
+            // Mock implementation, always return true
+            Ok(true)
         }
     }
 }
--- a/editoast/editoast_authz/src/lib.rs
+++ b/editoast/editoast_authz/src/lib.rs
@@ -49,12 +49,27 @@ pub async fn ensure_latest_authorization_model(
 pub enum Error<StorageError: std::error::Error> {
     #[error("unknown subject {0}")]
     UnknownSubject(i64),
+    #[error("unknown resource {0}")]
+    UnknownResource(i64),
     #[error("unknown user {identity}")]
     UnknownUser { identity: String },
     #[error(transparent)]
     OpenFga(#[from] fga::client::RequestFailure),
     #[error(transparent)]
+    OpenFgaParsing(#[from] fga::model::ParsingError),
+    #[error(transparent)]
     Storage(StorageError),
+    #[error("Unauthorized")]
+    Unauthorized,
+}
+
+impl<StorageError: std::error::Error> From<fga::client::QueryError> for Error<StorageError> {
+    fn from(err: fga::client::QueryError) -> Self {
+        match err {
+            fga::client::QueryError::Parsing(parsing_error) => Self::OpenFgaParsing(parsing_error),
+            fga::client::QueryError::Request(request_failure) => Self::OpenFga(request_failure),
+        }
+    }
 }
 
 #[cfg(test)]
--- a/editoast/editoast_authz/src/regulator.rs
+++ b/editoast/editoast_authz/src/regulator.rs
@@ -1,6 +1,7 @@
 use std::collections::HashSet;
 use std::future::Future;
 
+use fga::client::UserList;
 use fga::fga;
 use fga::model::Relation;
 use futures::stream;
@@ -91,6 +92,9 @@ pub trait StorageDriver: Clone {
             Self::Error,
         >,
     > + Send;
+
+    fn infra_exists(&self, infra_id: i64)
+    -> impl Future<Output = Result<bool, Self::Error>> + Send;
 }
 
 impl<S: StorageDriver> Regulator<S> {
@@ -161,6 +165,7 @@ impl<S: StorageDriver> Regulator<S> {
             .openfga
             .list_users(Group::member().query_users(&group))
             .await?;
+
         debug_assert!(
             members.public_access.is_none(),
             "we don't write public accesses for groups"
@@ -224,7 +229,8 @@ impl<S: StorageDriver> Regulator<S> {
     #[tracing::instrument(skip(self), ret(level = Level::DEBUG), err)]
     pub async fn user_roles(&self, user_id: i64) -> Result<HashSet<Role>, Error<S::Error>> {
         // no need to check for user inexistence, an empty set will be returned in this case
-        let roles = Role::list_roles(&self.openfga, User::role(), &fga!(User:user_id)).await?;
+        let roles =
+            Role::list_roles(&self.openfga, model::User::role(), &fga!(User:user_id)).await?;
         Ok(roles.into_iter().collect())
     }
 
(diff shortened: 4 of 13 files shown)


sim51:
Following a discussion with @flomonster , we found that we have an issue with the API design here.
Those API will be used for the infra management. 

What we need is : 
- list the infra and get the permission of the current user for each infra
- for an infra, retrieving the list of user that has access to it with their grant
- update / remove a grant of a user on an infra
- add a user to the infra with a specific grant.

What we've got :
- Route `/infra` endpoint returns the list of infra that the user has access. 
- Calling `/authz/me/grants` with the infra IDS give the user permissions
- POST `/authz/{resource_type}/{resource_id}` can update/remove and add a user to the infra. 
- To add a new user, you need to search it before, so we need a search user API.  

What should be refacto : 
- [x] removing route `/subjects`
- [x] adding route GET `/anthz/{resource_type}/{resource_id}` which should return the users with their grants on the resource 
- [x] adding a search user API => See #11226

sim51:
Remainings
- [x] The `admin` role should bypass checks
- [ ] Grant is unique, a user cannot have READER & WRITER

docker-compose.yml
leovalais, line 1: No need to change this file imo

editoast/editoast_authz/src/grant.rs
leovalais, line 1: Not sure if that's intentional while we only deal with infras, but the grant list is specific to each resource. I'd suggest not to abstract that right away.

Likewise, the relationship between the grants and privileges can vary for each resource. We may want to find a generic way to do declaratively deal with this issue later.

editoast/editoast_authz/src/privilege.rs
leovalais, line 1: Same comment as for `Grant`

editoast/editoast_authz/src/authorizer.rs
```diff
@@ -49,6 +57,136 @@ impl<S: StorageDriver> Authorizer<S> {
...
+            .await
+    }
+
+    pub async fn update_grants(
+        &self,
+        authorizations: Vec<SubjectResourceGrant>,
+    ) -> Result<(), Error<S::Error>> {
+        let mut grants_to_execute: Vec<SubjectResourceGrant> = Vec::new();
+
+        for auth in authorizations.clone() {
```
woshilapin: I'm not sure you need to `clone` the entire `Vec`. You can probably just read it.

```suggestion
        for auth in &authorizations {
```

editoast/editoast_authz/src/privilege.rs
```diff
@@ -0,0 +1,37 @@
...
+    Serialize,
+    Deserialize,
+    EnumString,
+    AsRefStr,
+    EnumIter,
+    Display,
+    ToSchema,
+)]
+pub enum Privilege {
+    #[allow(non_camel_case_types)]
```
woshilapin: Any reason for deviating from the standard enum case in Rust?

editoast/editoast_authz/src/authorizer.rs
```diff
@@ -49,6 +57,136 @@ impl<S: StorageDriver> Authorizer<S> {
...
+                            auth.resource_type,
+                            auth.resource_id,
+                            crate::Privilege::can_share_ownership,
+                        )
+                        .await?;
+                    if !can_share_ownership {
+                        return Err(Error::Unauthorized());
+                    }
+                }
+            }
```
woshilapin: This whole piece of code can be more succinct I believe.

```rust
for auth in &authorizations {
  let privilege = match auth.grant {
    Grant::READER => crate::Privilege::can_share_read,
    Grant::WRITER => crate::Privilege::can_share_write,
    Grant::OWNER => crate::Privilege::can_share_ownership,
  };
  let can_share =  = self
    .regulator
    .check_privilege(
      self.user_id,
      auth.resource_type,
      auth.resource_id,
      privilege,
    )
    .await?;
    if !can_share {
      return Err(Error::Unauthorized());
    }
}

```

front/openapi.yaml
woshilapin, line 1: What is this file? Should probably be removed.

woshilapin (commented):
This PR is huge, and the description doesn't explain what it does in details. Can't this be separated in multiple PR?

I've started reviewing this and saw a few minor comments so I'm pushing them, but I didn't finish the review since I'm missing the purpose of all of it and the details of why it's done like this (maybe this is traced in a ticket?)

leovalais:
@woshilapin: @sim51 presented me this PR yesterday, I'm still in the middle of reviewing it. I'll resume that once the impersonation is un-blocked.

editoast/editoast_authz/src/resource.rs
leovalais, line 1: Same comment as for `Grant` and `Privilege`. We want a `Resource` enum for the views still, but it should be located in the route module and only have the "API value" meaning.

editoast/editoast_authz/src/regulator.rs
```diff
@@ -26,6 +35,47 @@ pub struct Regulator<S: StorageDriver> {
...
+    pub resource_id: i64,
+    pub grant: Grant,
+}
+
+#[derive(Clone, Deserialize, ToSchema)]
+pub struct ResourceGrant {
+    pub resource_type: Resource,
+    pub resource_id: i64,
+    pub grant: Grant,
+}
```
leovalais: I don't think we need such genericity just yet. We'll only support infras and rolling stocks this year, and there's plenty experimentation to do to determine the best way to be generic over resources and relations in a type-safe way.

My suggestion would be to restrict the scope of this PR as much as possible:

1. Move the structures required for the routes in the `views` module and drop the pub visibility.
2. Let's keep the user and group management functions in the `Authorizer` as these are quite special because of their relation with the database. They will probably stay there even once we find a way to be generic over resources. But they shouldn't use these definitions.
3. I'd strongly suggest to keep things as simple as possible for infra grants and privileges. Just having a bunch of free functions `can_read_infra`, `can_write_infra`, `share_infra_read`, `get_infra_user_can_read`, etc. will likely be enough for quite a while. (We can of course find a bit nicer API like making these associated `model::Infra` functions for example ^^)

The last point is especially important imo as we'll need to experiment quite a bit with APIs, and early genericity will only make that more difficult with Rust.

editoast/editoast_authz/src/lib.rs
```diff
@@ -48,9 +61,11 @@ pub enum Error<StorageError: std::error::Error> {
     #[error("unknown subject {0}")]
     UnknownSubject(i64),
     #[error(transparent)]
-    OpenFga(#[from] fga::client::RequestFailure),
+    OpenFga(#[from] fga::client::QueryError),
```
leovalais: I'd suggest to `impl From<QueryError>` instead and include both of its variant in this error and let `?` do the work:

```rust
pub enum QueryError {
    #[error(transparent)]
    Parsing(#[from] ParsingError),
    #[error(transparent)]
    Request(#[from] RequestFailure),
}
```

editoast/editoast_authz/src/lib.rs
```diff
@@ -48,9 +61,11 @@ pub enum Error<StorageError: std::error::Error> {
...
     #[error(transparent)]
     Storage(StorageError),
+    #[error("Unauthorized")]
+    Unauthorized(),
```
leovalais: I'd refrain from interpreting an authorization missing as a failure *in the context of `editoast_authz`*. This crate provides the required mechanism to determine whether an operation is authorized, so a missing authorization is a valid result in this context. Interpreting this as an error is a job for the `views` imo.

If we had instead a type like:

```rust
enum Authorization<ProtectedData> {
	Authorized(ProtectedData),
	Unauthorized
}

fn retrieve_protected_data(...) -> Result<Authroized<T>
, Error>
```

it makes the authorization obvious in the signature (instead of hidden as one of many variants in a common error type) and forces the user to explicitly match on the authorization and deal with the unauthorized case.

editoast/editoast_authz/src/regulator.rs
```diff
@@ -291,4 +390,402 @@ impl<S: StorageDriver> Regulator<S> {
...
+        Ok(result)
+    }
+
+    /// Return the list of subject that have access to the resource
+    pub async fn get_resource_users(
+        &self,
+        resource_type: Resource,
+        resource_id: i64,
+        page: u64,
+        page_size: u64,
```
leovalais: I don't think that any pagintation should occur in `editoast_authz` since the crate is not responsible for the storage of resources. Since that's the job of `models`, any query partitioning should occur there. And "pagination", which is an API concept, should stay in views.

We can of course design a nice API to link everything smoothly, we don't have to carry u64s around.

editoast/editoast_authz/src/privilege.rs
sim51: This enum (and others like Resource or Grant) reflects the openfga model.

front/openapi.yaml
sim51: my bad, sorry

editoast/editoast_authz/src/privilege.rs
woshilapin: OK for the naming, but I'm not sure we should deviate from the usual case of enums in Rust. One reason that actually mislead me when reviewing, I saw `Privilege::can_share_read` given as a parameter of a function, and I thought that this parameter was itself a function (because of the case), instead of an enum's variant.

sim51: Thanks to @flomonster, I know ho to use the `rename_all` of serde and the `serialize_all` for strum. 
So enums follow the rust convention, and on a to_string or json, they follow the openfga convention

leovalais (approved):
Thanks for this first implementation! In this first review I only focused on architecture, the details will come later.

It seems to me that this implementation is too generic in the `editoast_authz` area. We don't know where we're going in order to deal with many resources/grant/privileges conveniently and safely. Being overly generic now will only make experimenting towards that goal more difficult. More details about that in the comments ⬇️

My take would be to keep grant and privilege management as simple and as direct as possible, especially in `editoast_authz`. We can move the generic grant / privilege / etc. definitions in `views` where we're sure they fit, and maybe move them in `editoast_authz` later if the future architecture is compatible.

leovalais (changes requested):
Approved by mistake, sorry about that! 😓

sim51:
@leovalais I changed the code of this PR with the branch I shared with you.
This new code includes : 
- resource named functions in regulator (no generic anymore)
- refacto of authz API with struct only present in the API code

editoast/editoast_authz/src/regulator.rs
```diff
@@ -126,6 +127,30 @@ impl<S: StorageDriver> Regulator<S> {
...
+            .filter_map(|Group(group)| match group.parse() {
+                Ok(id) => Some(id),
+                Err(_) => {
+                    tracing::error!(group, "unparsable group id — skipping it");
+                    None
+                }
+            })
+            .collect())
+    }
+
```
leovalais: Linking [this PR](https://github.com/OpenRailAssociation/osrd/pull/11505) which also implemented this function. The implementations must not be duplicated.

editoast/editoast_authz/src/authorizer.rs
```diff
@@ -91,6 +278,61 @@ mod tests {
...
+
+        let authorizer = Authorizer::try_initialize(user_identity(), regulator())
+            .await
+            .unwrap();
+
+        authorizer
+            .regulator
+            .grant_infra_reader(authorizer.user_id(), 1)
+            .await
+            .expect("Update grants should be successfull");
```
leovalais:
```suggestion
            .expect("Update grants should be successful");
```

editoast/editoast_authz/src/lib.rs
leovalais: Up on this, that'd avoid a lot of `map_err`s

editoast/editoast_authz/src/lib.rs
leovalais: I still think the distinction is important, but this can be done afterwards

editoast/editoast_authz/src/subject.rs
```diff
@@ -42,3 +46,23 @@ impl AsRef<UserInfo> for User {
...
+    EnumString,
+    AsRefStr,
+    EnumIter,
+    Display,
+    ToSchema,
+)]
+pub enum Subject {
+    User,
+    Group,
+}
```
leovalais: Likely misplaced

editoast/editoast_authz/src/authorizer.rs
```diff
@@ -57,6 +57,193 @@ impl<S: StorageDriver> Authorizer<S> {
...
+    pub async fn grant_infra_reader(
+        &self,
+        user_id: i64,
+        infra_id: i64,
+    ) -> Result<(), Error<S::Error>> {
+        // Check that the connected user has the right to add the grants
+        let can_share = self
+            .regulator
+            .check_infra_privilege_can_share_read(self.user_id, infra_id)
+            .await?;
```
leovalais: I think this check should occur in the regulator, like for the Admin role

editoast/editoast_authz/src/regulator.rs
```diff
@@ -307,4 +356,435 @@ impl<S: StorageDriver> Regulator<S> {
...
+            if let Some(user_info) = user_info {
+                users.push(UserSubject {
+                    id: user_id,
+                    info: user_info,
+                });
+            }
+        }
+        Ok(users)
+    }
```
leovalais, line 725: Can this function be split? `model::User::parse_userlist` which only parses the list, and maybe `StorageDriver::get_user_info_batch(impl Iterator<Item = i64>) -> Result<impl futures::stream::TryStream<Ok = (i64, UserInfo), Error = Self::Error>, Self::Error>` with a default implementation? (We'll optimize the amount of queries later.)

editoast/editoast_authz/src/regulator.rs
```diff
@@ -307,4 +356,435 @@ impl<S: StorageDriver> Regulator<S> {
...
+        let has_grant = self.check_infra_grant_reader(user_id, infra_id).await?;
+        if !has_grant {
+            let user = fga!(User:user_id);
+            let infra = fga!(Infra:infra_id);
+            let mut writes = self.openfga.prepare_writes();
+            writes.push(&Infra::reader().tuple(&user, &infra));
+            writes
+                .execute()
+                .await
+                .map_err(|e| Error::OpenFga(e.into()))?;
```
leovalais:
```suggestion
	self.openfga
		.prepare_writes()
		.write(&Infra::reader().tuple(&user, &infra))
        .execute()
        .await?;
```

editoast/src/views/authz.rs
```diff
@@ -1,41 +1,143 @@
...
+    InfraGrant,
+    SubjectResourceGrant,
+    SubjectGrant,
+    InfraPrivilege,
+    BodyUpdateGrants,
+    WhoamiResponse,
+    ResourceGrant,
+    SubjectResource,
+    Subject,
+    SubjectItem,
```
leovalais: Better inline schemas that are only used once

editoast/src/views/authz.rs
```diff
@@ -1,41 +1,143 @@
...
 }
 
-#[derive(serde::Deserialize, utoipa::IntoParams)]
-struct UserIdPathParam {
-    /// A user ID (not to be mistaken for its identity, cf. editoast user model documentation)
-    user_id: i64,
+#[derive(Debug, Display, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "snake_case")]
+#[strum(serialize_all = "snake_case")]
+pub enum Resource {
```
leovalais: doesn't need to be pub

editoast/src/views/authz.rs
```diff
@@ -1,41 +1,143 @@
...
+    Infra,
+}
+
+#[derive(Debug, Display, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "snake_case")]
+#[strum(serialize_all = "snake_case")]
+#[allow(clippy::enum_variant_names)]
```
leovalais:
```suggestion

```

editoast/src/views/authz.rs
```diff
@@ -1,41 +1,143 @@
...
+    CanWrite,
+    CanShareWrite,
+    CanDelete,
+    CanShareOwnership,
+}
+
+#[derive(Debug, Display, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "snake_case")]
+#[strum(serialize_all = "snake_case")]
+pub enum InfraGrant {
```
leovalais: `pub(in crate::views)`

editoast/src/views/authz.rs
```diff
@@ -1,41 +1,143 @@
...
+    subject_id: i64,
+    grant: InfraGrant,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, ToSchema, PartialEq)]
+pub struct SubjectItem {
+    pub id: i64,
+    pub name: String,
+    pub r#type: Subject,
+}
```
leovalais: pub

editoast/src/views/authz.rs
```diff
@@ -46,22 +148,16 @@ enum AuthzError {
     Authorizer(#[from] AuthorizerError),
     #[error("Authorization error")]
     Authz(#[from] AuthorizationError),
+    #[error("Bad grants operation : {message}")]
+    BadGrantsOperation { message: String },
```
leovalais:
```suggestion
    SimultaneousGrantsAndRevokes,
```

let's be precise + the `#[error(MESSAGE)]` defines the message that's shown, the content is for context data usually

editoast/src/views/authz.rs
```diff
@@ -46,22 +148,16 @@ enum AuthzError {
...
-    NoSuchUser { user_id: i64 },
-}
-
-#[derive(Debug, serde::Serialize, utoipa::ToSchema)]
-#[cfg_attr(test, derive(serde::Deserialize, PartialEq))]
-struct WhoamiResponse {
-    id: i64,
-    name: String,
-    roles: Vec<Role>,
+    ResourceTypeNotFound { resource_type: String },
```
leovalais:
```suggestion
    UnknownResourceType { resource_type: String },
```

nit++: `-NotFound` strikes me as a an error for data that may or may not be there, whereas here we know for sure the resource type in question doesn't exist

editoast/src/views/authz.rs
```diff
@@ -83,161 +179,579 @@ async fn whoami(Extension(auth): AuthenticationExt) -> Result<Json<WhoamiRespons
...
+
+    // Check that infra exists
+    let infra_id = resource_id_param.resource_id;
+    let conn = &mut db_pool.get().await?;
+    Infra::retrieve_or_fail(conn, infra_id, || InfraApiError::NotFound { infra_id }).await?;
+
+    // Validate pagination params
+    let (page, page_size) = pagination_params
+        .validate(1000)?
+        .warn_page_size(100)
```
leovalais:
```suggestion
        .validate(100)?
```

the warn is only there for API retrocompatibility

editoast/src/views/authz.rs
```diff
@@ -83,161 +179,579 @@ async fn whoami(Extension(auth): AuthenticationExt) -> Result<Json<WhoamiRespons
...
-    Extension(auth): AuthenticationExt,
-    State(AppState { regulator, .. }): State<AppState>,
-) -> Result<Json<Roles>> {
-    if !auth
-        .check_roles([Role::OperationalStudies].into())
-        .await
-        .map_err(AuthorizationError::from)?
-    {
-        return Err(AuthorizationError::Forbidden.into());
+async fn users_grants_for_resource_id(
```
leovalais, line 262: This function doesn't check if the infra IDs returned by OpenFGA still exist.

Here's the workflow I'd suggest:

1. Fetch from OpenFGA an `HashMap<i64, InfraGrant>` for each grant
2. Use `views::pagination::PaginatedList` to handle pagination and querying the infras from the DB
3. Match the result against the ID map and build the response
4. I don't remember the API we designed but it might be a good idea to include the `PaginationStats` in the response.

editoast/src/views/authz.rs
```diff
@@ -83,161 +179,579 @@ async fn whoami(Extension(auth): AuthenticationExt) -> Result<Json<WhoamiRespons
...
+    State(AppState {
+        regulator, db_pool, ..
+    }): State<AppState>,
+    Path(resource_type_param): Path<ResourceTypeParam>,
+    Path(resource_id_param): Path<ResourceIdParam>,
```
leovalais: we usually destructure path params when possible

editoast/src/views/authz.rs
```diff
@@ -83,161 +179,579 @@ async fn whoami(Extension(auth): AuthenticationExt) -> Result<Json<WhoamiRespons
...
 
-    check_user_exists(user_id, &regulator).await?;
-
-    regulator
-        .grant_user_roles(user_id, HashSet::from_iter(roles))
-        .await
-        .map_err(AuthzError::from)?;
-    Ok(axum::http::StatusCode::NO_CONTENT)
+    let mut infra_privileges: HashMap<InfraGrant, Vec<InfraPrivilege>> = HashMap::new();
+    // Order is important here, we use it in the user_grants function
```
leovalais: ?

editoast/src/views/authz.rs
```diff
@@ -83,161 +179,579 @@ async fn whoami(Extension(auth): AuthenticationExt) -> Result<Json<WhoamiRespons
...
-    State(AppState { regulator, .. }): State<AppState>,
-    Json(RoleListBody { roles }): Json<RoleListBody>,
-) -> Result<impl axum::response::IntoResponse> {
-    if !auth
-        .check_roles([Role::Admin].into())
-        .await
-        .map_err(AuthorizationError::from)?
-    {
-        return Err(AuthorizationError::Forbidden.into());
+    Json(body): Json<BodyUpdateGrants>,
```
leovalais:
```suggestion
    Json(BodyUpdateGrants { grant, revoke }): Json<BodyUpdateGrants>,
```

editoast/src/views/authz.rs
```diff
@@ -1,41 +1,143 @@
...
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, ToSchema, PartialEq)]
+struct SubjectGrant {
+    subject: SubjectItem,
+    grant: InfraGrant,
+}
+
+#[derive(Clone, Debug, Deserialize, ToSchema, PartialEq)]
+pub struct BodyUpdateGrants {
```
leovalais: pub + the fact that either `grant` xor `revoke` is expected should be documented in the openapi

editoast/src/views/authz.rs
```diff
@@ -83,161 +179,579 @@ async fn whoami(Extension(auth): AuthenticationExt) -> Result<Json<WhoamiRespons
...
-        .revoke_user_roles(user_id, HashSet::from_iter(roles))
-        .await
-        .map_err(AuthzError::from)?;
-    Ok(axum::http::StatusCode::NO_CONTENT)
+    Ok(StatusCode::NO_CONTENT)
 }
 
 #[cfg(test)]
-mod tests {
+pub mod test {
```
leovalais:
```suggestion
mod tests {
```

no reason to make it pub and it's `tests` by convention

editoast/src/views/authz.rs
```diff
@@ -83,161 +179,579 @@ async fn whoami(Extension(auth): AuthenticationExt) -> Result<Json<WhoamiRespons
...
     use super::*;
+    use crate::models::fixtures::create_small_infra;
+    use crate::views::test_app::TestApp;
+    use crate::views::test_app::TestRequestExt;
+    use editoast_authz::subject::UserInfo;
+    use pretty_assertions::assert_eq;
+    use rstest::rstest;
+    use serde_json::json;
+
+    #[rstest]
```
leovalais, line 524:
```suggestion
    #[tokio::test(flavor = "multi_thread")] // maybe with worker_threads = 1 as well
```

we're trying to get rid of `rstest` where pssible

editoast/src/views/authz.rs
```diff
@@ -83,161 +179,579 @@ async fn whoami(Extension(auth): AuthenticationExt) -> Result<Json<WhoamiRespons
...
+            .by_user(&user);
+        let users: Vec<SubjectGrant> = app
+            .fetch(request_all)
+            .assert_status(StatusCode::OK)
+            .json_into();
+        assert_eq!(users.len(), 1);
+    }
+
+    #[rstest]
+    async fn grants_test() {
```
leovalais, line 607: The case of granting/revoking the same grant twice should be tested since there's pre-processing involved on our end

editoast/src/views/test_app.rs
```diff
@@ -412,6 +427,29 @@ impl<'a> UserBuilder<'a> {
...
+                            regulator
+                                .grant_infra_owner(user.id, infra_id)
+                                .await
+                                .expect("Infra owner should be granted successfully");
+                        }
+                        InfraGrant::Writer => {
+                            regulator
+                                .grant_infra_writer(user.id, infra_id)
+                                .await
+                                .expect("Infra owner should be granted successfully");
```
```suggestion
                                .expect("Infra writer should be granted successfully");
```

editoast/src/views/authz.rs
leovalais: Also can the definition of these schemas be moved above the endpoint that use them instead of having them all at the beginning of the file?

leovalais (commented):
Thanks a lot for the changes! There's still a few things to change imo but I think we'll be able to merge this soon.

There's a few comments I made for later when we'll revisit this code to factorize or/and make it generic. I'll try to work on that later in the PI. But these are ofc not blocking this PR.

editoast/editoast_authz/src/regulator.rs
sim51: I'm agree, but the PR is not yet merged.
It will be easy to do the rebase, code is at 99% the same

editoast/editoast_authz/src/lib.rs
sim51: This error is not used (at least for now) when we retrieve some data, but when we do some grant &revoke operations, like in `grant_infra_writer` . These functions return nothing, but throw an error if the user is not authorized to do the action, which make sense for me. 

I don't know what are the good practices in rust for managing exception, but in the java world working with exception like that is normal.

If I implement ` grant_infra_writer`  in your way,  It becomes something like that 
`pub async fn grant_infra_owner( &self,  user_id: i64,  infra_id: i64,) -> Result<Authorization<()>, Error<S::Error>> `, and then I need to check the returned result to know if the action was well performed or not.  
This is weird for me, because we already have a `Result` which can be in Success or in Error mode.

editoast/src/views/authz.rs
sim51: `InfraGrant` is used in `test_app`  for the `with_infra_grant` function

editoast/src/views/authz.rs
sim51: I removed struct that are already inline (ie BodyUpdateGrants & WhoamiResponse). 

But  I have one issue : how to inline sub types ?
Example : `BodyUpdateGrants` is inline, but it's definition uses `SubjectResource` & `SubjectResourceGrant` which are not includes in the inline of `BodyUpdateGrants`.  So how can I inline them ?

editoast/src/views/authz.rs
sim51: This function returns users (with their grant) related to the given resource, so your notes can't be related to it. 
For users, we already call the DB to retrieve the user's info, otherwise it is omitted.

In the regulator, all functions like `check_infra_grant_owner` check that the given user exists, but do nothing on `infraId`. We need to discuss about it .  
The only way I see to be able to add this check in the regulator, is to add a `infra_exists` function in the `StorageDriver`.

About the API :
-  "/me" => nothing to do
- "/me/grants" => I added a check to verify that the infra given in the body exists before to query openfga
- "/grants" => to do, it's where we call functions like "grant_infra_owner". IMHO, the regulator should be in charge of checking that the infra exists before
- "/{resource_type}/{resource_id}" =>  we already check the existence of users
 - "/grants/{resource_type}"=>  nothing to do

editoast/src/views/authz.rs
sim51: The following test already exists `adding_a_grant_that_already_exists`. 
Is it what you means by "grant twice" ?

editoast/editoast_authz/src/authorizer.rs
sim51: For me the authorizer is related to  the connected user,  that's why all checks related to the connected user are in it. On the other side,  the regulator is focus on grants jobs.

So if you ask the regulator if the user `42` has access to the resource `FRANCE`, and `42` is an `Admin`, the answer must be YES. 
And if you ask the regulator to add the `READ` grant on the resource `FRANCE`, it will do it  (this can be useful for a CLI for example). 

If you want to check that the caller (ie. the connected user) can add the `READ` grant on the resource `FRANCE`, use the `authorizer`.

Does it make sense ?

editoast/src/views/authz.rs
sim51: This is needed because every value starts with `can`

editoast/src/views/authz.rs
sim51: Don't know why, but when changing to tokio, my tests never ends, even when I removed the `multi_thread`

editoast/editoast_authz/src/authorizer.rs
sim51: Following our discussion : 
- put all business logic in the regulator, so this check must be put in the regulator
- app need to assign grant to user (ex: when a user clone a infra), so we need a function for that. We name it `grant_infra_reader_unchecked`

editoast/src/views/authz.rs
sim51: Yes we need to check `infraId` in the regulator and so adding a `infra_exists` on the `StorageDriver`

editoast/src/views/authz.rs
sim51: `#[schema(inline)]`

sim51:
@leovalais I did a new commit with : 
- check of infra exist in regulator
- adding all business logic in regulator
- remove a max of struct in the schema 
- the error management to avoid the map_error we did together

What is missing : 
- the split of the parse_userlist function 
- the `Unauthorized` code way (ie. Error vs Return type)

IMHO, both are not blocking for now

flomonster (approved):
Approve docker-compose change as devops.

> [!IMPORTANT]
> I did not review the editoast part. I let @leovalais handle this one :)

editoast/editoast_authz/src/lib.rs
leovalais: Following our discussion, this still requires some debate. So let's just fix the form here:

```suggestion
    #[error("Unauthorized")]
    Unauthorized,
```

likewise for returning it: `Err(Error::Unauthorized)` without parentheses

editoast/editoast_authz/src/regulator.rs
leovalais: sure, that was a ping to @woshilapin in case his PR is merged second

editoast/editoast_authz/src/regulator.rs
```diff
@@ -177,7 +206,7 @@ impl<S: StorageDriver> Regulator<S> {
         Ok(())
     }
 
-    /// Removes some users from a group
+    /// Remove some users from a group
```
leovalais:
```suggestion
    /// Removes some users from a group
```

🤔 

> What does this function do? It removes X from Y

+ inconsistency with the other docs

editoast/src/views/authz.rs
```diff
@@ -1,41 +1,78 @@
...
+    Hash,
+    Serialize,
+    Deserialize,
+    EnumString,
+    AsRefStr,
+    EnumIter,
+    Display,
+    ToSchema,
+)]
+pub enum Subject {
```
leovalais:
```suggestion
enum Subject {
```

editoast/src/views/authz.rs
```diff
@@ -46,18 +83,20 @@ enum AuthzError {
     Authorizer(#[from] AuthorizerError),
     #[error("Authorization error")]
     Authz(#[from] AuthorizationError),
+    #[error("{message}")]
+    SimultaneousGrantsAndRevokes { message: String },
```
leovalais:
```suggestion
    #[error("Grant and Revoke cannot be defined at the same time")]
    SimultaneousGrantsAndRevokes,
```

editoast/src/views/authz.rs
```diff
@@ -83,163 +122,624 @@ async fn whoami(Extension(auth): AuthenticationExt) -> Result<Json<WhoamiRespons
...
     }
 
-    check_user_exists(user_id, &regulator).await?;
+    if let Some(grants) = grant {
+        for grant in grants {
+            if grant.resource_type != Resource::Infra {
+                return Err(ResourceError::UnknownResourceType {
+                    resource_type: grant.resource_type.to_string(),
+                })?;
+            }
```
leovalais:
```suggestion
```

can never happen

editoast/src/views/authz.rs
```diff
@@ -83,163 +122,624 @@ async fn whoami(Extension(auth): AuthenticationExt) -> Result<Json<WhoamiRespons
...
+        }
+    }
+    if let Some(revoke) = revoke {
+        for grant in revoke {
+            if grant.resource_type != Resource::Infra {
+                return Err(ResourceError::UnknownResourceType {
+                    resource_type: grant.resource_type.to_string(),
+                })?;
+            }
```
leovalais:
```suggestion
```

likewise

editoast/src/views/authz.rs
```diff
@@ -83,163 +122,624 @@ async fn whoami(Extension(auth): AuthenticationExt) -> Result<Json<WhoamiRespons
...
-            .clone(),
-    }))
+    let mut result = HashMap::new();
+    let conn = &mut db_pool.get().await?;
+
+    if let Some(infra_ids) = body.get(&Resource::Infra) {
+        for infra_id in infra_ids {
+            // check that the infra exists before to check the grants
+            let infra = Infra::retrieve(conn, *infra_id).await?;
+            if infra.is_some() {
```
leovalais: Not sure what's more convenient API-wise between a 404 or an incomplete response. Maybe @kmer2016 or @SarahBellaha have a preference?

leovalais:
```suggestion
            if Infra::exists(conn, *infra_id).await? {
```

editoast/src/views/authz.rs
```diff
@@ -83,163 +122,624 @@ async fn whoami(Extension(auth): AuthenticationExt) -> Result<Json<WhoamiRespons
...
+        return Err(ResourceError::UnknownResourceType {
+            resource_type: resource_type.to_string(),
+        })?;
     }
 
-    check_user_exists(user_id, &regulator).await?;
+    // Check that infra exists
+    let infra_id = resource_id;
+    let conn = &mut db_pool.get().await?;
+    Infra::retrieve_or_fail(conn, infra_id, || InfraApiError::NotFound { infra_id }).await?;
```
leovalais:
```suggestion
    Infra::exists_or_fail(conn, infra_id, || InfraApiError::NotFound { infra_id }).await?;
```

Not sure if it's worth checking that anymore since we trust the result of the regulator. We can match on the `Err` of the regulator call to return this specific error.

editoast/src/views/authz.rs
leovalais, line 1: The struct definitions here `derive` a lot of things, which I'm sure they're not all used. Can we keep only what's necessary? (these hurt compilation times when they pile up)

editoast/src/views/authz.rs
```diff
@@ -83,163 +122,624 @@ async fn whoami(Extension(auth): AuthenticationExt) -> Result<Json<WhoamiRespons
...
-        .revoke_user_roles(user_id, HashSet::from_iter(roles))
-        .await
-        .map_err(AuthzError::from)?;
-    Ok(axum::http::StatusCode::NO_CONTENT)
+    Ok(StatusCode::NO_CONTENT)
 }
 
 #[cfg(test)]
-mod tests {
+mod test {
```
leovalais:
```suggestion
mod tests {
```

leovalais (commented):
Another round of review, I'll test the endpoints manually tomorrow and we should be good after that :)

editoast/src/views/authz.rs
leovalais: Off topic comment, but now infra checks are done in the regulator using a StorageDriver function

editoast/src/views/authz.rs
sim51: I t can be a 400, but not a 404 (the resource/url exists).

I'm agree that here I did a design choice that can be discussed : I prefer to have a response on this endpoint, even if I give in the list one infraID that doesn't exist, but I will have the data for the others).

In fact for the 404 infraID, there is no value, which means no authorization, which can be considered as correct.

leovalais: > It can be a 400, but not a 404 (the resource/url exists).

If the user asks about his rights about an infra with a wrong ID, it means it tries to read a resource that doesn't exits thus justifying the 404 code doesn't it?

> I'm agree that here I did a design choice that can be discussed

In editoast we conventionally fail if we detect the slightest error in the request. I don't really mind it this way though. But then:

1. To decide, asking API consumers about what they prefer can help tilt the balance. If they do not answer I'm fine with leaving things as they are (@kmer2016 @SarahBellaha 👀)
2. Since this is a bit of a paradigm shift for editoast, the documentation should emphasize this behavior.

editoast/src/views/authz.rs
sim51: By removing those lines (which can be deleted for now because we only have one ), the `resource_type` attribute is not needed anymore in the code.  But we need to keep it for the API contract to prepare the futur.
So to avoid that,  you will see in the code  a ` match grant.resource_type` with only one arm.

editoast/src/views/authz.rs
sim51: I added a comment about this in the description of the openapi

editoast/editoast_authz/src/authorizer.rs
leovalais: outdated

editoast/editoast_authz/src/authorizer.rs
leovalais: outdated

leovalais (approved):
LGTM. I think we reached a good foundation to gradually improve from now on. Thanks a lot for this work and for the back and forth. I'll test this and make a list of what to do next.

> [!NOTE]
> The commits need to be squashed

leovalais:
# Next steps

Kinda ordered

## To consider the feature complete

- [ ] Check infra authorization in endpoints
- [ ] Manage `user:id#manager` tuples
- [ ] While revoking an owner, check at least one owner will remains
- [ ] Handle subject (ie. User & Group) instead of just User
- [ ] Fix bug: `/authz/me/grants` should return return the actual grants, not compute them from privileges (an Admin is always owner of everything for instance)
- [ ] Second guessing [this choice](https://github.com/OpenRailAssociation/osrd/pull/11197#discussion_r2063535262): we can't tell apart "the infra doesn't exist" and "the infra exist but the user doesn't have any grant on it" by just looking at the response

## Enhancements / refactoring

- [ ] Check if `/authz/grants/{resource_type}` keys (grants) can be typed in the openapi to help the frontend
- [ ] A LOT OF LOGGING IS MISSING
- [ ] https://github.com/OpenRailAssociation/osrd/pull/11197#discussion_r2056146813
- [ ] Settle on `Error::Unauthorized` https://github.com/OpenRailAssociation/osrd/pull/11197#discussion_r2015099284
- [ ] Adding multi-check request on FGA client
- [ ] Tie FGA checks with DB exploration, the 2 strategies need to be implemented, **with pagination**
- [ ] Remove `InternalError` from `AuthDriverError` (important, blocks the split in crates)
- [ ] Transactional OpenFGA updates
- [ ] Check the size of the FGA queries
- [ ] Compute `/authz/grants/{resource_type}` response using the `/expand` API of OpenFGA
- [ ] Experiment with *type-safe* genericity
	- [ ] Factorize all Authorizer and Regulator boilerplate
	- [ ] Allow dealing with different resources
	- [ ] To try: represent relations that are used to store data differently from inferring-only relations in `fga` => allows us to represent grants and privileges differently upstream + safer requests

leovalais:
Found an error related bug: if you're trying to make someone owner of an infra you're only writer on, it fails with a 500 instead of a 403

kmer2016 (approved):
LGTM

editoast/editoast_authz/src/regulator.rs
sim51: mark at resolved to merge the PR

sim51:
I fixed the issue for the 403

PR #11198, opened 2025-03-19
core: stdcm: fix conflict introduced by planned steps

eckter:
The associated unit test used to fail and has been fixed. 

I thought I managed to reproduce a bug in a unit test. Turns out it's a different cause and doesn't actually fix any of the real requests I'm trying to fix. Still, it's a fix.

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
@@ -236,7 +236,11 @@ class STDCMPostProcessing(private val graph: STDCMGraph) {
 
     /**
      * Identify the max possible time we can add to the previous stop, assuming we can remove some
-     * time from next stops
+     * time from next stops.
+     *
+     * For example, with a train stopping at A, having a requested timing at B and then stopping
+     * again at C: We're trying to find by how much we can "move" stop duration from C to A, to
+     * delay B passage time without affecting what's outside the [A, C] range.
      */
     private fun findMaxPossibleTimeToAdd(
         lastStopIndexBeforeNode: Int,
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/fr/sncf/osrd/stdcm/graph/STDCMPostProcessing.kt
```diff
@@ -246,7 +250,13 @@ class STDCMPostProcessing(private val graph: STDCMGraph) {
         lastTimeData: TimeData,
     ): Double {
         if (lastStopIndexBeforeNode == 0) return lastTimeData.maxFirstDepartureDelaying
-        var maxTimeDiff = Double.POSITIVE_INFINITY
+
+        // Init the result with the delay we can add without conflict between the last stop and the
+        // planned node
+        var maxTimeDiff =
+            if (node.stopDuration == null) node.timeData.maxDepartureDelayingWithoutConflict
```
Erashin, line 257: I think there might be a pb with `maxDepartureDelayingWithoutConflict`'s description in `TimeData`. Though I fully agree with this fix, the way the value is described, it makes me think it should work for delaying departure as well as later stops :/

eckter: It does work for departures and stops, which is precisely the issue here. 

When a node isn't a stop, there's no issue, we just pick `maxDepartureDelayingWithoutConflict`.

But when it is a stop, the time data is considered *after* that stop. That value wouldn't tell us anything about before the stop, just by how much we can lengthen it. So we need to look it up with `stopTimeData.last().maxDepartureDelayBeforeStop`, which keeps track of the max delay before the stop (for this kind of scenario).

I agree it's confusing though.

Erashin: Ah fuck just noticed the diff, true my bad, it's the delay after, whereas here we're looking for the delay before said stop. Yeaaahhhh this is fun haha.

Erashin (approved):
Unfortunate that we need such a generic test, but works for me. GJ.

PR #11203, opened 2025-03-20
front: support no infra edit override

ghost:
In order to publish a public version of OSRD, we have to prevent infra edition that would otherwise spwan multiples cores.

We've already introduced a blocking mechanism for those routes on the gateway, here is a flag in overrides in order to prevent the front from displaying the edit feature, thus avoid exposing the final user to crashes when the feature is purposefully disabled.


clarani (approved):
LGTM :white_check_mark:

PR #11205, opened 2025-03-20
editoast: railjson: loosen the railjson version check for persisting

Castavo:
This allows for railjsons with compatible versions to easily be loaded.

It's either this, or we have to write migrations for all railjson versions update, even if we don't need to update the content of the railjsons. (Currently, you can't download and reupload some railjsons on deployed envs)

I made two choices which are very much debatable:
* Checking only the major version
* Using a very naive way to check the semver. I thought about using a regex, but then we should rather use a crate (like [semver](https://docs.rs/semver/latest/semver/index.html)), but I thought another dependency could be too much.

### DIFF
--- a/editoast/editoast_schemas/src/infra/railjson.rs
+++ b/editoast/editoast_schemas/src/infra/railjson.rs
@@ -52,3 +52,10 @@ pub struct RailJson {
     /// `Detector` is a device that identifies the presence of a train in a TVD section (Track Vacancy Detection section), indicating when a track area is occupied.
     pub detectors: Vec<Detector>,
 }
+
+pub fn major_version(version: &str) -> &str {
+    match version.split_once('.') {
+        Some((major, _)) => major,
+        None => version,
+    }
+}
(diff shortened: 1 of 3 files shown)


editoast/editoast_schemas/src/infra/railjson.rs
```diff
@@ -17,6 +17,10 @@ use super::TrackSection;
 
 pub const RAILJSON_VERSION: &str = "3.4.13";
 
+pub fn min_version_prefix() -> String {
```
woshilapin: Is it "minimum" version accepted or the "minor" from semantic versioning? I would either expand to `minimum_version` or `major_version` (or `railjson_major_version`).

editoast/editoast_schemas/src/infra/railjson.rs
```diff
@@ -17,6 +17,10 @@ use super::TrackSection;
...
+    RAILJSON_VERSION.split('.').next().unwrap().to_owned() + "."
+}
```
woshilapin: I'm not sure that we need to reconstruct a `String` with a dot `.`. How about this function is used like this instead

```rust
fn major_version(version: &str) -> &str {
  version.split_once('.').unwrap()
}

// And then use it like
if major_version(version) != major_version(RAILJSON_VERSION)
```

woshilapin (commented):
I'm fine with the semantic of the change, not yet just sure about the implementation. I've made some propositions.

Castavo:
@woshilapin thanks for the suggestions, I was quite unhappy with the implem too

woshilapin (approved):
Thank you.

PR #11208, opened 2025-03-20
front: fix itinerary warning and timestops display

### LINKED ISSUE #8832: front: train edition: if the path is invalid but debug mode is available, the rest of the tabs think it's valid
### What happened?

When a train has an invalid path and the debug mode is available, the tab isn't shown in red and the schedule table is present.

In the video, at first, an example of the desired behavior, then the actual behavior

https://github.com/user-attachments/assets/a15e6e7b-da3d-4420-9c41-2b18e5960954

I expect the form to behave the same way whether the PF is fully broken or whether the debug mode is available:
* pathfinding tab surrounded by a red border
* schedule table not available

1. Create a train schedule on a path where for example the electrification is incompatible
2. Witness

Dev (SNCF)

Firefox

6edea6d

RomainValls:
closes #8832 

fix itinerary warning and timestops display when pathfinding error.

### DIFF
--- a/front/src/modules/timesStops/TimesStopsInput.tsx
+++ b/front/src/modules/timesStops/TimesStopsInput.tsx
@@ -7,6 +7,7 @@ import type { Operation } from 'react-datasheet-grid/dist/types';
 import { useTranslation } from 'react-i18next';
 
 import { useScenarioContext } from 'applications/operationalStudies/hooks/useScenarioContext';
+import type { PathfindingState } from 'modules/pathfinding/types';
 import { isVia, matchPathStepAndOp } from 'modules/pathfinding/utils';
 import type { SuggestedOP } from 'modules/trainschedule/components/ManageTrainSchedule/types';
 import {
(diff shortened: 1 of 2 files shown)


front/src/modules/timesStops/TimesStopsInput.tsx
```diff
@@ -134,6 +137,10 @@ const TimesStopsInput = ({
   );
 
   useEffect(() => {
+    if (pathfindingState.error) {
+      setRows([]);
+      return;
+    }
```
theocrsb, line 143: We don't want to display the itinerary in the 'schedules and stops' section when there is an incompatibility?

RomainValls: I'm not sure about this. On @Castavo 's comment on the issue :  the PF is broken but the tab isn't red and the schedule table is present.

Castavo: Yes, we discussed it with @maelysLeratRosso, allowing to set schedules and stops on a path a train can't actually use isn't desirable. 
There should be no difference in behavior whether the pathfinding failed because of the constraints or because the path is impossible.

theocrsb (approved):
LGTM and tested

SharglutDev (approved):
Lgtm and tested, good job !

Issue #10239, opened 2025-01-06
core: `Path` typing isn't consistent


eckter:
[`Path`](https://github.com/OpenRailAssociation/osrd/blob/58d7306a3ffb3dda50f3307030e0c09953b9a6ef/core/kt-osrd-sim-infra/src/main/kotlin/fr/sncf/osrd/sim_infra/api/PathProperties.kt#L26) is an interface that can be used to type  `Length`, `Offset`, and such.

But we have significant issues with what it actually means. When we handle several types of paths, it generally means "0 = start of the first block included in the path". But there are cases where 0 means the departure point (as in, the travelled path).

That's a significant issue. We should unify this, make the semantics consistent, and document it. We should probably remove the ambiguous `Path` at least temporarily, and introduce a `BlockPath` type instead.

---

Might be addressed at some point (no sub-issue created, TDB if we must track it):
- [ ] type offsets in envelope framework, as noted https://github.com/OpenRailAssociation/osrd/pull/11210#discussion_r2007244553
- [ ] Rename some types for consistency about use of "sequence"/"path"/"..." https://github.com/OpenRailAssociation/osrd/issues/11259

PR #11210, opened 2025-03-20
core: stdcm: fix path / travelled path confusion on block availability

### LINKED ISSUE #11121: lmr: technical problem error on MAC-LYD-VCE
### What happened?

I tried a simulation with these inputs : 

Image

I got the following error : 

Image

I expected to have a success

As above

Recette (SNCF)

Google Chrome

2c2c893

eckter:
Well that was a massive bug considering how sneaky it is. It's in `BlockAvailability` (a class that's mocked in most tests), and `Offset<Path> == Offset<TravelledPath>` in most test cases.

`InfraExplorerWithEnvelope.getSimulatedLength()` returned a `Length<Path>`, when it should really be `TravelledPath`. 

I'd add a unit test with stops in `BlockAvailabilityTests`, but that would massively conflict with the other pending PRs. 

Fix some (most?) cases of https://github.com/OpenRailAssociation/osrd/issues/11121.

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
@@ -10,6 +10,7 @@ import fr.sncf.osrd.envelope.EnvelopeInterpolate
 import fr.sncf.osrd.envelope_sim.PhysicsRollingStock
 import fr.sncf.osrd.railjson.schema.schedule.RJSTrainStop.RJSReceptionSignal.SHORT_SLIP_STOP
 import fr.sncf.osrd.sim_infra.api.Path
+import fr.sncf.osrd.sim_infra.api.TravelledPath
 import fr.sncf.osrd.standalone_sim.EnvelopeStopWrapper
 import fr.sncf.osrd.standalone_sim.result.ResultTrain.SpacingRequirement
 import fr.sncf.osrd.stdcm.graph.StopTimeData
@@ -179,10 +180,14 @@ data class InfraExplorerWithEnvelopeImpl(
         return this
     }
 
-    override fun getSimulatedLength(): Length<Path> {
+    override fun getSimulatedLength(): Length<TravelledPath> {
         if (envelopes.isEmpty()) return Length(0.meters)
         val lastEnvelope = envelopes[envelopes.size - 1]
         return Length(Distance.fromMeters(lastEnvelope.startOffset + lastEnvelope.envelope.endPos))
+    }
+
+    override fun getSimulationEndPathOffset(): Offset<Path> {
+        return getIncrementalPath().fromTravelledPath(getSimulatedLength())
     }
 
     override fun clone(): InfraExplorerWithEnvelope {
--- a/core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/BlockAvailabilityTests.kt
+++ b/core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/BlockAvailabilityTests.kt
@@ -106,10 +106,12 @@ class BlockAvailabilityTests {
         nBlocksInPath: Int,
         nBlocksSimulated: Int,
         originalSteps: List<STDCMStep> = listOf(),
-        rollingStock: RollingStock = REALISTIC_FAST_TRAIN
+        rollingStock: RollingStock = REALISTIC_FAST_TRAIN,
+        startOffset: Offset<Block> = Offset(0.meters)
     ): InfraExplorerWithEnvelope {
         assert(nBlocksInPath >= nBlocksSimulated)
         assert(nBlocksInPath <= 5)
+        assert(startOffset <= blockLengths[0])
 
         val steps = originalSteps.toMutableList()
         val lastStep =
@@ -125,7 +127,7 @@ class BlockAvailabilityTests {
         var infraExplorer =
             initInfraExplorerWithEnvelope(
                     infra,
-                    PathfindingEdgeLocationId(blocks[0], Offset(0.meters)),
+                    PathfindingEdgeLocationId(blocks[0], startOffset),
                     rollingStock,
                     steps,
                 )
@@ -133,12 +135,14 @@ class BlockAvailabilityTests {
         while (infraExplorer.getLookahead().size + 1 < nBlocksInPath) infraExplorer =
             infraExplorer.cloneAndExtendLookahead().find { filterExplorer(it) }!!
         for (i in 0 ..< nBlocksSimulated) {
+            var envelopeLength = blockLengths[i].distance.meters
+            if (i == 0) envelopeLength -= startOffset.distance.meters
             infraExplorer =
                 infraExplorer.addEnvelope(
                     Envelope.make(
                         EnvelopePart.generateTimes(
                             listOf(EnvelopeProfile.CONSTANT_SPEED),
-                            doubleArrayOf(0.0, blockLengths[i].distance.meters),
+                            doubleArrayOf(0.0, envelopeLength),
                             doubleArrayOf(30.0, 30.0)
                         )
                     )
@@ -157,7 +161,7 @@ class BlockAvailabilityTests {
             availability.getAvailability(
                 explorer,
                 Offset(0.meters),
-                explorer.getSimulatedLength(),
+                explorer.getSimulationEndPathOffset(),
                 0.0
             )
         }
@@ -172,7 +176,7 @@ class BlockAvailabilityTests {
             availability.getAvailability(
                 explorer,
                 Offset(0.meters),
-                explorer.getSimulatedLength(),
+                explorer.getSimulationEndPathOffset(),
                 0.0
             )
         }
@@ -189,7 +193,7 @@ class BlockAvailabilityTests {
             availability.getAvailability(
                 explorer,
                 Offset(0.meters),
-                explorer.getSimulatedLength(),
+                explorer.getSimulationEndPathOffset(),
                 0.0
             ) as BlockAvailabilityInterface.Unavailable
         assertEquals(duration, res.duration)
@@ -206,7 +210,7 @@ class BlockAvailabilityTests {
             availability.getAvailability(
                 explorer,
                 Offset(0.meters),
-                explorer.getSimulatedLength(),
+                explorer.getSimulationEndPathOffset(),
                 0.0
             ) as BlockAvailabilityInterface.Unavailable
         assertEquals(duration, res.duration)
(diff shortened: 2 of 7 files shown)


core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/BlockAvailabilityTests.kt
```diff
@@ -268,12 +272,35 @@ class BlockAvailabilityTests {
...
+                )
+            )
+        val res =
+            availability.getAvailability(
+                explorer,
+                Offset(0.meters),
+                explorer.getSimulationEndPathOffset(),
+                0.0
+            ) as BlockAvailabilityInterface.Available
+        assertEquals(t - explorer.getFullEnvelope().totalTime, res.maximumDelay)
```
Erashin: nit: You could also add an assert on the values of getSimulationLength and getSimulationEndPathOffset, just to make sure no one ever switches em again.

Erashin (approved):
Well that must have been fun to debug. Good job, this is huge!
For the test, I'm fine with merging this as is, putting said test in an issue and doing it once the other PRs are merged.

core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/BlockAvailabilityTests.kt
eckter: I've added a test on `InfraExplorerWithEnvelope` directly

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
```diff
@@ -192,12 +193,16 @@ data class InfraExplorerWithEnvelopeImpl(
         return this
     }
 
-    override fun getSimulatedLength(): Length<Path> {
+    override fun getSimulatedLength(): Length<TravelledPath> {
         if (envelopes.isEmpty()) return Length(0.meters)
         val lastEnvelope = envelopes[envelopes.size - 1]
         return Length(Distance.fromMeters(lastEnvelope.startOffset + lastEnvelope.envelope.endPos))
```
bougue-pe, line 186: For another PR? 
Can we type the `LocatedEnvelope.startOffset`?

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
```diff
@@ -192,12 +193,16 @@ data class InfraExplorerWithEnvelopeImpl(
...
     }
 
+    override fun getSimulationEndPathOffset(): Offset<Path> {
+        return getIncrementalPath().fromTravelledPath(getSimulatedLength())
```
bougue-pe, line 190: For another PR? 
Providing an `PathOffsetBuilder` inside `IncrementalPath` would probably be clearer and better mutualization.

core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/BlockAvailabilityTests.kt
```diff
@@ -268,12 +272,35 @@ class BlockAvailabilityTests {
...
         assertEquals(minStartTime, res.timeOfNextConflict)
     }
 
+    /**
+     * In this case Offset<Path> != Offset<EnvelopePath>, checks that offsets types aren't mixed up
+     */
+    @Test
+    fun testNonZeroStartOffset() {
+        val explorer = makeExplorer(5, 5, startOffset = Offset(4_000.meters))
+        val t = 1200.0
```
bougue-pe:
```suggestion
        val requirementStartTime = 1200.0
```

core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/BlockAvailabilityTests.kt
```diff
@@ -268,12 +272,35 @@ class BlockAvailabilityTests {
...
                 Offset(0.meters),
-                explorer.getSimulatedLength(),
+                explorer.getSimulationEndPathOffset(),
                 0.0
             ) as BlockAvailabilityInterface.Available
         assertEquals(minStartTime, res.timeOfNextConflict)
     }
 
+    /**
+     * In this case Offset<Path> != Offset<EnvelopePath>, checks that offsets types aren't mixed up
```
bougue-pe: As there is ongoing work around naming and `EnvelopePath` doesn't exist, I think we shouldn't introduce something new.
It looks like current type for envelope sequence is `EnvelopeConcat`.
`EnvelopeSimPath` exists, so it may be used, even though it's not a sequence of designated object.
We may come up with `EnvelopePartPath` to have a sequence of designated object?

Not sure about the meaning here, but maybe just `Offset<Envelope>` ?

bougue-pe (commented):
Thanks for this bugfix!

Apart from the naming, I guess my first 2 comments can (should?) be tracked a sibling issues of https://github.com/OpenRailAssociation/osrd/issues/10595.
If agreeing on the usefulness of suggestions, I can create issues.

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
eckter: The envelope framework is still mostly based on untyped `double` and they're not *that* trivial to integer-based `Distances`. The plan was to not migrate it, but we've been dealing with this "legacy" module for a while now...

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
eckter: I'm not sure I understand the suggestion, you're saying we should have something like `InfraExplorer.fromTravelPath`? Or build and expose a `PathOffsetBuilder` in `InfraExplorer`?

core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/BlockAvailabilityTests.kt
eckter: In the current implementation, in every case `Offset<Envelope> == Offset<TravelledPath>`.

But it's true that it's not *necessarily* the case, and we could introduce that type.

There's a slight ambiguity though: for an envelope ranging from position 50 to 100, where does `Offset<Envelope>(50.meters)` point to? Is it the start or the end?

core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/BlockAvailabilityTests.kt
eckter: Fixed in https://github.com/OpenRailAssociation/osrd/pull/11210/commits/d997b34fb92f0156018bf09a22cd9aa8a87436fa

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
bougue-pe: I was thinking to attach a `PathOffsetBuilder` to `IncrementalPath` (and remove its `from`/`toTravelledPath` fn), but it may make more sense to attach it to `InfraExplorer`?

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
bougue-pe: OK, I only noted it in the "meta" issue on the matter of offset typing https://github.com/OpenRailAssociation/osrd/issues/10239
Resolving this.

core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/BlockAvailabilityTests.kt
bougue-pe: In my understanding that "`Object`Path" is a sequence of instances of `Object`: `Offset<Envelope>(50.meters)` points to the end, because it's not a path.

Anyway, I think this comment should move a bit, as I still don't understand if it refers to EnvelopeConcat or Envelope:
```suggestion
     * In this case Offset<Path> != Offset<EnvelopeConcat>, checks that offsets types aren't mixed up
```
?

eckter: Oh I missunderstood the initial comment (specifically I didn't read my own code carefully enough).

I intended to write `Offset<TravelledPath>` there, and I agree we can stick to well established types even in comments. Even if envelope offset typing would be nice.

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
eckter: I have to admit I like `IncrementalPath.toTravelledPath` methods. They're the most commonly used methods to convert paths, the `PathOffsetBuilder` is the exception and it's *mostly* local to `ScheduleMetadataExtractor`. We could instantiate a builder inside `IncrementalPath` for code factorization, but incremental paths are instanciated and cloned *a lot*, so the new objects may be expensive.

We could add `InfraExplorer.toTravelledPath` as an utility method (as it's very often chained after `getIncrementalPath` calls), but I'd see it as just wrapping `IncrementalPath.toTravelledPath`.

core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/BlockAvailabilityTests.kt
eckter: I've changed it to `TravelledPath`, is it OK for you? I feel like I'm dodging the rest of the discussion, but I don't know if it's still relevant.

bougue-pe: I'm perfectly OK to leave the rest of the discussion to https://github.com/OpenRailAssociation/osrd/issues/10239 (and related work).
Thanks!

core/src/main/kotlin/fr/sncf/osrd/stdcm/infra_exploration/InfraExplorerWithEnvelopeImpl.kt
bougue-pe: OK, perf matters here 🙂 
We could bring `PathOffsetBuilder`'s code a bit closer to what's in `IncrementalPath` to keep the best of both (tiny bit more typing like `IncrementalPath`, tiny bit less indirection like `PathOffsetBuilder`).
It's a detail, and I resolve this.

bougue-pe (approved):

bougue-pe (approved):
🙏 If you have a unit test for the second commit's bug, it's nice to embed (we can consider that e2e tests are enough, I guess).

PR #11215, opened 2025-03-20
Fix gateway mock provider

flomonster:
This PR cleans the `x-remote-user-identity` and `x-remote-user-name` when using the mock provider.

Before: `mock.identity/Example User`
After: `mock/mocked`

---------

I also fixed the username that was not being passed to editoast. As the mock provider was registered as both a session provider and an identity provider, it always matched the identity (which doesn't provide the username). 
To solve this problem, I removed the mock provider's identity implementation. By doing so, the `require_login` field became unused, so I deleted it.

### DIFF
--- a/gateway/README.md
+++ b/gateway/README.md
@@ -83,11 +83,6 @@ type = "Mocked"
 provider_id = "mock"
 # The username reported to the front-end
 username = "Example User"
-# The unique user id sent to the back-end
-user_id = "example-user-id"
-# Whether users actually need to log in to call targets where require_auth = true.
-# It's used for testing.
-require_login = true
 ```
 
 # Enable open telemetry
--- a/gateway/actix_auth/src/providers/mock.rs
+++ b/gateway/actix_auth/src/providers/mock.rs
@@ -1,28 +1,21 @@
 use actix_web::HttpRequest;
-use serde::{Deserialize, Serialize};
+use serde::Deserialize;
+use serde::Serialize;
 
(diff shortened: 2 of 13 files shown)


gateway/README.md
```diff
@@ -83,11 +83,8 @@ type = "Mocked"
 provider_id = "mock"
 # The username reported to the front-end
 username = "Example User"
-# The unique user id sent to the back-end
+# The unique user id sent to the back-end (default: "mocked")
 user_id = "example-user-id"
```
Khoyo: Maybe we should change the example config to use the default?

gateway/actix_auth/src/providers/mock.rs
```diff
@@ -1,28 +1,21 @@
...
-        }
+    pub fn new(username: String, user_id: Option<String>) -> Self {
+        Self { username, user_id }
     }
 
     fn get_user_id(&self) -> &str {
-        self.user_id
-            .as_deref()
-            .unwrap_or_else(|| self.username.as_ref())
+        self.user_id.as_deref().unwrap_or("mocked")
```
Khoyo: I think `MockProvider::user_id` should'nt be optional, and that the config parser should handle setting the default

Khoyo (approved):
Otherwise lgtm

sim51:
Tested it. 
Now I have a `mock/mocked (Example User)` in my database., so it works!

sim51 (approved):

PR #11221, opened 2025-03-21
Correct Loader Font, Clear Alternative Simulation Info, and Cancel Alternative Results

### LINKED ISSUE #10572: Alternative Solutions Search and Display for Failed STDCM Simulations
When an STDCM simulation fails, two alternative queries should be triggered in parallel with slight adjustments to the time tolerance parameters (one earlier and one later). During this process, the loader should be updated to inform the user that an alternative solutions search is in progress.

achrafmohye:
Closes #10572

### DIFF
--- a/front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
+++ b/front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
@@ -35,9 +35,10 @@ type StcdmResultsProps = {
   onStartNewQueryWithData: () => void;
   buttonsVisible: boolean;
   showStatusBanner: boolean;
+  hasSeenInfoMessage: boolean;
 };
 
-const StcdmResults = ({
+const StdcmResults = ({
   isCalculationFailed,
   isDebugMode,
   onSelectSimulation,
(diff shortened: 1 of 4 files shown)


front/src/applications/stdcm/components/StdcmResults/StdcmResults.tsx
```diff
@@ -35,6 +35,7 @@ type StcdmResultsProps = {
   onStartNewQueryWithData: () => void;
   buttonsVisible: boolean;
   showStatusBanner: boolean;
+  hasSeenInfoMessage: boolean;
 };
 
 const StcdmResults = ({
```
Caracol3: NIT : It's probably something that's been inherited for a while, but we could rename StcdmResults to StdcmResults to keep things consistent.

achrafmohye: Nice catch, Done !

Caracol3 (approved):
lgtm (tested)

kmer2016 (approved):
LGTM !

PR #11224, opened 2025-03-21
front: add loader for macro editor

emersion:
Instead of showing a blank screen when the iframe is loading and we're fetching all train schedules, show a loader.

### DIFF
--- a/front/src/applications/operationalStudies/components/NGE/NGE.tsx
+++ b/front/src/applications/operationalStudies/components/NGE/NGE.tsx
@@ -20,6 +20,7 @@ interface NGEElement extends HTMLElement {
 type NGEProps = {
   dto?: NetzgrafikDto;
   onOperation?: (op: NGEEvent, netzgrafikDto: NetzgrafikDto) => void;
+  onLoad?: () => void;
 };
 
 const frameSrc = `
@@ -43,7 +44,7 @@ const frameSrc = `
  * Abstracts away low-level NGE details. Doesn't contain any OSRD-specific
  * logic.
  */
-const NGE = ({ dto, onOperation }: NGEProps) => {
+const NGE = ({ dto, onOperation, onLoad }: NGEProps) => {
   const frameRef = useRef<HTMLIFrameElement>(null);
 
   const [ngeRootElement, setNgeRootElement] = useState<NGEElement | null>(null);
@@ -55,6 +56,8 @@ const NGE = ({ dto, onOperation }: NGEProps) => {
       const ngeRoot = frame.contentDocument!.createElement('sbb-root') as NGEElement;
       frame.contentDocument!.body.appendChild(ngeRoot);
       setNgeRootElement(ngeRoot);
+
+      if (onLoad) onLoad();
     };
 
     frame.addEventListener('load', handleFrameLoad);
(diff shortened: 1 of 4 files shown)


louisgreiner (commented):
EDIT: nevermind, I thought it was another feature fix

It may be because my computer is slow and wants to go in weekend, but I still can see the default `netzgrafikDto` after the loading screen

https://github.com/user-attachments/assets/65269ded-1c13-4f74-b9b1-be771bfb6c50

louisgreiner (approved):
Working great.

Sadly, the blinking issue is still there. Do you see any other solution than [this](https://github.com/osrd-project/netzgrafik-editor-frontend/pull/13)?

emersion:
Oh yeah, this PR doesn't aim to fix the blinking default DTO. So that's expected.

front/src/applications/operationalStudies/components/NGE/NGE.tsx
```diff
@@ -20,6 +20,7 @@ interface NGEElement extends HTMLElement {
 type NGEProps = {
   dto?: NetzgrafikDto;
   onOperation?: (op: NGEEvent, netzgrafikDto: NetzgrafikDto) => void;
+  onLoad?: () => void;
```
SharglutDev, line 23: Seems like the function we pass is always defined ?

front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
```diff
@@ -60,6 +61,7 @@ const ScenarioContent = ({
   } = useScenarioData(scenario, infra);
   const macroEditorState = useRef<MacroEditorState>();
   const [ngeDto, setNgeDto] = useState<NetzgrafikDto>();
+  const [ngeLoaded, setNGELoaded] = useState(false);
```
SharglutDev: nit : feels like we often use `isLoading` to handle the loading state of an element rather than the "end" state

SharglutDev (approved):
Lgtm and tested, really nice to finally have this loader. Left just two nits

front/src/applications/operationalStudies/components/NGE/NGE.tsx
emersion: Yeah, but like `onOperation` (and in general component event handlers) I didn't want to force users to define it.

front/src/applications/operationalStudies/components/Scenario/ScenarioContent.tsx
emersion: Hm, what do you mean?

SharglutDev: In most of the codebase (mainly because of rtk query `isLoading` but not only that), we use `setIsloading(true)` when the loading starts and `setIsLoading(false)` when it ends. Here you used the opposite pattern : "has something finished loaded or not".
But that's really not important :)

emersion: Switched to `isLoading`!

emersion:
Also fixed a bug where the editor would get stuck in loading state when clicking twice on the macro tab.

achrafmohye (approved):
LGTM :)

PR #11226, opened 2025-03-21
editoast: adding endpoint for searching users

sim51:


flomonster (approved):
LGTM

editoast/src/views/search.rs
```diff
@@ -615,6 +615,34 @@ pub(super) struct SearchResultItemProject {
     tags: Vec<String>,
 }
 
+#[derive(Search, Serialize, ToSchema)]
+#[search(
+    table = "search_user",
+    migration(src_table = "authn_user"),
+    joins = "INNER JOIN authn_user ON authn_user.id = search_user.id",
```
leovalais, line 622: No need to join over the real table since all the necessary data to build the response item is already in cache.

leovalais (commented):
LGTM about the code, but by design a user is only allowed to see the union of members and managers from all its groups. This feature breaks that. Was this considered @sim51 @flomonster ?

PR #11231, opened 2025-03-24
front, ci : add a custom report for e2e tests

### LINKED ISSUE #11193: Add a detailed GitHub report for E2E tests
### Description and goal

Generate a summary of your Playwright tests within the GitHub Actions workflow, including detailed results and traces of failed tests
- Generate a personalized playwright report (.json) using playwright build it test-results
- Display the personalized report in github summary
### Acceptance criteria

- [x] Test reporter that display detailed results
- [ ] Link playwright trace to failed tests

Maymanaf:
Closes #11193  
💀 Don't mind the third commit — it intentionally causes some tests to fail to better visualize the report.

You can view the report from the following screenshot:  
image


scripts/generate-github-summary.sh
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env bash
+set -e
+
+REPORT_FILE="$1"
+OUTPUT_FILE="$2"
```
Khoyo, line 5: There should be at least some a check that we have 2 arguments, or print a usage message to stdout

(not !!mandatory!! if you have `set -u`)

scripts/generate-github-summary.sh
```diff
@@ -0,0 +1,117 @@
...
+
+ICON_TESTS="📝"
+ICON_SUCCESS="✅"
+ICON_FAIL="❌"
+ICON_SKIP="⏭️"
+ICON_FLAKY="🍂"
+ICON_TIME="⏱️"
+
+TOTAL=$(jq '.results.summary.tests // 0' "$REPORT_FILE")
```
Khoyo, line 14: Not an environment variable or a constant, should be lowercase (ditto for all other variables)

See https://google.github.io/styleguide/shellguide.html#s7.2-variable-names

scripts/generate-github-summary.sh
```diff
@@ -0,0 +1,117 @@
...
+  "| \(.name) | \(
+    if .status == "passed" then "✅ passed"
+    elif .status == "failed" then "❌ failed"
+    elif .status == "skipped" then "⏭️ skipped"
+    elif .status == "timedOut" then "⌛ timed out"
+    else .status end
+  ) | \((.duration // 0) / 1000) | \(
+    if .flaky then "**Yes**" else "No" end
+  ) |"
+' "$REPORT_FILE")
```
Khoyo, line 109: This should be a function, eg `generate_summary()`. Then you could just echo stuff to `stdout` instead of doing clunky string concatenation

.github/actions/generate-test-summary/action.yml
```diff
@@ -0,0 +1,51 @@
...
+          exit 1
+        fi
+
+    - name: Generate test summary
+      shell: bash
+      id: generate
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
+      run: |
+        chmod +x ./scripts/generate-github-summary.sh
```
Khoyo, line 50: The script should be executable already. If it isn't, it isn't the ci job to fix it.

Additionally, it should live next to the action (in `.github/actions/generate-test-summary/`), unless it's intended for developer use.

scripts/generate-github-summary.sh
```diff
@@ -0,0 +1,117 @@
...
+
+echo "$SUMMARY" > "$OUTPUT_FILE"
+{
+  echo "summary-markdown<<EOF"
+  echo "$SUMMARY"
+  echo "EOF"
+} >> "$GITHUB_OUTPUT"
+echo "failed-count=$FAILED" >> "$GITHUB_OUTPUT"
```
Khoyo, line 117:
```suggestion
{
  echo "summary-markdown<<EOF"
  generate_summary | tee "$OUTPUT_FILE"
  echo "EOF"
} >> "$GITHUB_OUTPUT"
echo "failed-count=$FAILED" >> "$GITHUB_OUTPUT"
```

scripts/generate-github-summary.sh
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env bash
+set -e
```
Khoyo, line 2: You probably want `set -euo pipefail` unless you have a good reason not to

scripts/generate-github-summary.sh
```diff
@@ -0,0 +1,117 @@
...
+ICON_SKIP="⏭️"
+ICON_FLAKY="🍂"
+ICON_TIME="⏱️"
+
+TOTAL=$(jq '.results.summary.tests // 0' "$REPORT_FILE")
+PASSED=$(jq '.results.summary.passed // 0' "$REPORT_FILE")
+FAILED=$(jq '.results.summary.failed // 0' "$REPORT_FILE")
+SKIPPED=$(jq '.results.summary.skipped // 0' "$REPORT_FILE")
+START=$(jq '.results.summary.start // 0' "$REPORT_FILE")
+STOP=$(jq '.results.summary.stop // 0' "$REPORT_FILE")
```
Khoyo, line 19: Why would those field not be present, and why are we not treating that as an error? Are we not in control of the report file?

scripts/generate-github-summary.sh
```diff
@@ -0,0 +1,117 @@
...
+FLAKY_COUNT=$(jq '[.results.tests[] | select(.flaky == true)] | length' "$REPORT_FILE")
```
Khoyo, line 20: Why isn't this done on the JS side?

scripts/generate-github-summary.sh
```diff
@@ -0,0 +1,117 @@
...
+DURATION_SEC=$(( (STOP - START) / 1000 ))
+DURATION=$(printf '%02d:%02d:%02d' $((DURATION_SEC/3600)) $((DURATION_SEC%3600/60)) $((DURATION_SEC%60)))
+
+REPO="${GITHUB_REPOSITORY}"
+RUN_ID="${GITHUB_RUN_ID}"
+TOKEN="${GITHUB_TOKEN}"
+
+ARTIFACTS=$(curl -s -H "Authorization: Bearer $TOKEN" \
+  -H "Accept: application/vnd.github+json" \
+  "https://api.github.com/repos/${REPO}/actions/runs/${RUN_ID}/artifacts")
```
Khoyo, line 31: I believe we can assume gh is installed:
```suggestion
ARTIFACTS=$(gh api repos/${REPO}/actions/runs/${RUN_ID}/artifacts)
```

scripts/generate-github-summary.sh
```diff
@@ -0,0 +1,117 @@
...
+
+TRACE_ID=$(echo "$ARTIFACTS" | jq -r '.artifacts[] | select(.name == "playwright-traces") | .id' | head -n1)
+VIDEO_ID=$(echo "$ARTIFACTS" | jq -r '.artifacts[] | select(.name == "playwright-videos") | .id' | head -n1)
+
+TRACE_URL="https://github.com/${REPO}/actions/runs/${RUN_ID}/artifacts/${TRACE_ID}"
+VIDEO_URL="https://github.com/${REPO}/actions/runs/${RUN_ID}/artifacts/${VIDEO_ID}"
```
Khoyo, line 37: Why not just use the `url` field instead of taking the `id` to reconstruct the url?

scripts/generate-github-summary.sh
```diff
@@ -0,0 +1,117 @@
...
+START=$(jq '.results.summary.start // 0' "$REPORT_FILE")
+STOP=$(jq '.results.summary.stop // 0' "$REPORT_FILE")
+FLAKY_COUNT=$(jq '[.results.tests[] | select(.flaky == true)] | length' "$REPORT_FILE")
+
+DURATION_SEC=$(( (STOP - START) / 1000 ))
+DURATION=$(printf '%02d:%02d:%02d' $((DURATION_SEC/3600)) $((DURATION_SEC%3600/60)) $((DURATION_SEC%60)))
+
+REPO="${GITHUB_REPOSITORY}"
+RUN_ID="${GITHUB_RUN_ID}"
+TOKEN="${GITHUB_TOKEN}"
```
Khoyo, line 27: You can just use the `GITHUB_*` variables directly, I'm not sure what the point is to rename them

Synar:
Thanks for the pr! Why did you split the report in both a ts part and bash/jq part? It seems a bit complex to me for what it does, could we do all the formatting in ts?

.github/actions/generate-test-summary/action.yml
```diff
@@ -0,0 +1,51 @@
+name: "Generate Test Summary"
```
Khoyo, line 1: Is there a reason this job specifically needs a custom action?

Khoyo (commented):
Did not review the typescript yet.

I'm quite surprised that a shell script is necessary at all... Why cant the custom report just generate the report we want, in an actual high level language like typescript ?

front/tests/reporter/generate-personalized-report.ts
```diff
@@ -0,0 +1,333 @@
...
+interface ReporterConfigOptions {
+  outputFile?: string;
+  outputDir?: string;
+  minimal?: boolean;
+  screenshot?: boolean;
+  annotations?: boolean;
+  testType?: string;
+}
+
+class GeneratePersonalizedReport implements Reporter {
```
ghost, line 30: I feel like this class is a mix of a state that is sometimes changed by functions, sometimes not, and feel truly messy with confusing function names and structure.

Why does it not output the markdown directly ? That would prevent the usage on `jq` and even the need for an external script.

front/tests/reporter/generate-personalized-report.ts
```diff
@@ -0,0 +1,333 @@
...
+      return null;
+    }
+  }
+
+  static updateStart(startTime: Date): number {
+    return Math.floor(new Date(startTime).getTime() / 1000);
+  }
+
+  static calculateStopTime(startTime: Date, duration: number): number {
+    return Math.floor(new Date(startTime).getTime() + duration) / 1000;
```
ghost, line 241: The operation order is not correct resulting in non-integer seconds.

front/tests/reporter/generate-personalized-report.ts
```diff
@@ -0,0 +1,333 @@
...
+    } catch (error) {
+      console.error(
+        'Error parsing browser metadata:',
+        error instanceof Error ? error.message : error
+      );
+      return null;
+    }
+  }
+
+  static updateStart(startTime: Date): number {
```
ghost, line 236: Why is this method named "update" and do not touch the state ?

front/tests/reporter/generate-personalized-report.ts
```diff
@@ -0,0 +1,333 @@
...
+    }
+  }
+
+  private setFilename(filename: string): void {
+    this.reporterConfigOptions.outputFile = filename.endsWith('.json')
+      ? filename
+      : `${filename}.json`;
+  }
+
+  private updatePersonalizedTestResultsFromTestResult(
```
ghost, line 118: Why is this method apart of `updateSummaryFromTestResult` ? What is the motivation for having 2 methods for this that are called once ?

front/tests/reporter/generate-personalized-report.ts
```diff
@@ -0,0 +1,333 @@
...
+    if (testCase.results.length === 0) return;
+
+    const latestResult = testCase.results.at(-1);
+    if (latestResult) {
+      this.updatePersonalizedTestResultsFromTestResult(testCase, latestResult);
+      this.updateSummaryFromTestResult(latestResult);
+    }
+  }
+
+  private setFilename(filename: string): void {
```
ghost, line 112: This method seems weird. It is called from `onBegin` with a member of the class with the default attached ? What was the intent here ?

ghost (approved):

ghost (changes requested):
Miss click with the last review.

Maymanaf:
> Did not review the typescript yet.
> 
> I'm quite surprised that a shell script is necessary at all... Why cant the custom report just generate the report we want, in an actual high level language like typescript ?

https://github.com/OpenRailAssociation/osrd/pull/11231#issuecomment-2980260398

ghost:
> > Thanks for the pr! Why did you split the report in both a ts part and bash/jq part? It seems a bit complex to me for what it does, could we do all the formatting in ts?
> 
> The ts code uses the Playwright functions to generate the report. To display it in the GitHub summary, it needs to be in Markdown format, which is a bit complex to handle directly in ts [Adding a job summary](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#adding-a-job-summary)

I feel that's a bad reason, because I do not adhere with the premise that doing it entirely in TS is harder/less maintainable than doing it with bash/jq, writing the JSON or writing the markdown in the onEnd hook seems almost the same difficulty. Furthermore, adding `bash` and `jq` to the stack is not simpler, bash and jq are way more error-prone than just the typescript implementation.

Maymanaf:
> > > Thanks for the pr! Why did you split the report in both a ts part and bash/jq part? It seems a bit complex to me for what it does, could we do all the formatting in ts?
> > 
> > 
> > The ts code uses the Playwright functions to generate the report. To display it in the GitHub summary, it needs to be in Markdown format, which is a bit complex to handle directly in ts [Adding a job summary](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#adding-a-job-summary)
> 
> I feel that's a bad reason, because I do not adhere with the premise that doing it entirely in TS is harder/less maintainable than doing it with bash/jq, writing the JSON or writing the markdown in the onEnd hook seems almost the same difficulty. Furthermore, adding `bash` and `jq` to the stack is not simpler, bash and jq are way more error-prone than just the typescript implementation.

Yes that's why I deleted my comment , but I guess it was too late 😄

Maymanaf:
Thanks for the reviews ! 
As everyone suggested, I'll remove `generate-github-summary.sh` and create a new ts file with a `generate_summary() `function instead

Maymanaf:
I'm closing this PR to open a new one (I'll link both for reference), since a significant part of the work will change. It's also the only way I can properly test in CI without spamming you with each push

PR #11233, opened 2025-03-24
front: use enum for StdcmResultsOperationalPoint.stopType

emersion:
Improves type safety: stopType can only take a few fixed strings instead of any string.


Akctarus (approved):
Tested and LGTM !

kmer2016 (approved):
LGTM !

PR #11234, opened 2025-03-24
front: map: switch to pmtiles for OSM and terrain

Tristramg:
The vpn now let us access to Hetzner’s object storage.

The map (including terrain) fetches on https://osm.nbg1.your-objectstorage.com/.

There should no longer any request to osm.osrd.fr

Please test if the map loads when the VPN is on

### DIFF
--- a/front/src/common/Map/const.ts
+++ b/front/src/common/Map/const.ts
@@ -4,8 +4,9 @@ import config from 'config/config';
 export const MAP_URL = `${config.proxy_editoast}/layers`;
 export const SPRITES_URL = `${config.proxy_editoast}/sprites`;
 export const FONTS_URL = `${config.proxy_editoast}/fonts`;
-export const OSM_URL = 'https://osm.osrd.fr/data/v3.json';
-export const TERRAIN_URL = 'https://osm.osrd.fr/data/terrain.json';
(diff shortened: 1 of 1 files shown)


emersion (approved):
LGTM (not tested)

emersion:
Note, it would be nicer to use the "tuiles.enliberté.fr" domain name instead of hardcoding a Hetzner-specific URL. Could be done with an HTTP redirect perhaps, or a JSON style.

flomonster (approved):

Tristramg:
A redirection works ! Thank you for the tip.
However, despite being 301 (permanently moved), firefox when panning on the map regularly pokes at tuiles.enliberté.fr. This introduces a small latency that is a bit annoying.

CNAME is currently not possible (this should change in the future).
There are provided styles such as https://tuiles.enliberté.fr/styles/bright.json that works in general (the bucket is hidden inside), but for OSRD there are custom styles

(plus tuiles.enliberté.fr is currently still blocked on the VPN — the domain is to recent)

emersion:
Right, that's a bit tricky indeed. We could maybe fetch the style, extract the `sources` field, and add our own style on top… but that's a bit cumbersome. There is no way to extend/combine styles at the moment (https://github.com/mapbox/mapbox-gl-js/issues/4225).

PR #11235, opened 2025-03-24
front: fix timetable train and paced train count displayed

### LINKED ISSUE #11177: The total of trains is not displayed correctly when select a type
### What happened?

Image

It should display 1/3 servies selected and 21 trains 
==> display the total of trains    

1. Go a scenario Page
2. Select a service or a train
3. Observe the label of the train count 

Recette (SNCF), Local

Firefox

7afc69b

Akctarus:
fix #11177


theocrsb:
Maybe we should add the German translation?

theocrsb (changes requested):
The English translation has not been added, but the French translation has the expected behavior

Akctarus:
> The English translation has not been added, but the French translation has the expected behavior

good catch, I had forgotten to push the last commit

Akctarus:
> The English translation has not been added, but the French translation has the expected behavior

I don't think so, we're waiting for an alternative to Transifex to take care of the german translation.

theocrsb (approved):

kmer2016 (approved):
LGTM !

theocrsb:
There's just this one case where the 's' is missing

image

Akctarus:
> There's just this one case where the 's' is missing
> 
> image

fixed

theocrsb (approved):

Issue #9372, opened 2024-10-17
general issue: path data APIs aren't robust to coordinate errors


eckter:
Most of OSRD exclusively deals with "offsets" to locate elements on their respective tracks or paths. When we want to display them, the front-end is in charge of projecting them on linear geographic elements.

The problem is, we often have mismatches between the referenced track lengths and their linestrings. Imported infrastructures have data quality issues, and generated infrastructures just don't have coordinates that match their distances.

In some cases, we can normalize offsets. For example, an object placed at offset=50m on a track of length=100m, we can place it at the middle of the linestring, even if that linestring is 500m long.

But that doesn't work out well when we need to project data over *a whole train path*. Any data error on any given track will throw off the normalization.

And the issue is that when the front-end projects objects on the path, it mostly relies on the `path_properties` endpoint. That endpoint returns the geographic data for the whole path. Worse, it doesn't even include the expected path length, making it impossible to normalize anything. Most projections end up being wrong.

Causes several bugs:

* https://github.com/OpenRailAssociation/osrd/issues/9228
* https://github.com/OpenRailAssociation/osrd/issues/9110

---

Possible fixes:

We can't think of any *satisfying* fix (but we're open to suggestions). For now we have a few general ideas:

1. Normalize the distances by ranges, included alongside the path geography (e.g. first 200m of geo path is worth 500m of expected path length)
2. All objects returned by the back-end include their geo data, no more front-end projection
3. All offsets include both a "path length" offset and a "geo length" offset

maelysLeratRosso:
Could it also trigger this [bug ](https://github.com/OpenRailAssociation/osrd/issues/10221)?

-> EDIT: Probably not

bougue-pe:
Workshop with @flomonster @clarani @eckter @SharglutDev @maelysLeratRosso @sim51 @Tristramg 

## TLDR

> [!IMPORTANT]  
>
> **This solution is on hold after a possibly better alternative has been expressed (see next comment)**

> Solution on hold (for current bugs and future designs):
> * When backend returns a topological offset/length (ranges) to be projected on a geographical linestring (so far: path):
>    Backend must also provide a corresponding ratio of the considered geographical linestring
> * All inputs to the backend must use only topological length (already OK)
>
> * ⚠ offsets on a tracksection are not impacted by this, as the ratio is pretty straightforward: tracksection contains its length and geographical length (from linestring).

The goals for the workshop are :

[...]

PR #11243, opened 2025-03-25
editoast: add geographical ratio to topological offset in path_properties

### LINKED ISSUE #11191: core: add `geom_projection` to the /path_properties response
part of https://github.com/OpenRailAssociation/osrd/issues/9228:

1. core provides `geom_projection` in response of `/path_properties`

Here is the detailed description of the `geom_projection` content: https://github.com/OpenRailAssociation/osrd/issues/9372#issuecomment-2765815821

hamz2a:
closes https://github.com/OpenRailAssociation/osrd/issues/11191

### DIFF
--- a/editoast/src/views/path/properties.rs
+++ b/editoast/src/views/path/properties.rs
@@ -18,7 +18,6 @@ use serde::Deserialize;
 use serde::Serialize;
 use serde_qs::axum::QsQuery;
 use std::collections::hash_map::DefaultHasher;
-use std::collections::HashMap;
 use std::hash::Hash;
 use std::hash::Hasher;
 use tracing::info;
@@ -76,10 +75,10 @@ impl PathPropertiesInput {
         cumulative_sums
     }
 
-    pub fn find_track_section_offset(&self) -> HashMap<String, u64> {
+    fn find_track_section_offsets(&self) -> Vec<(String, u64)> {
         let track_length_cumulative_sums = self.get_track_length_cumulative_sums();
-        let mut res = HashMap::new();
-        for path_item_position in self.path_item_positions.iter() {
+        let mut offsets = Vec::new();
+        for path_item_position in &self.path_item_positions {
             let Some((track_range, inferior_sum)) = self
                 .track_section_ranges
                 .iter()
@@ -98,12 +97,24 @@ impl PathPropertiesInput {
                 Direction::StopToStart => track_range.end - offset_on_track_range,
             };
 
-            res.insert(
+            offsets.push((
                 (*track_range.track_section).clone(),
                 offset_on_track_section,
-            );
+            ));
         }
-        res
+        offsets
+    }
+
+    fn get_ratio(&self, track_section: &TrackSection) -> Vec<(String, f64)> {
+        self.find_track_section_offsets()
+            .iter()
+            .filter(|(tr, _)| tr == track_section.id.as_ref())
+            .map(|(_, position)| *position as f64)
+            .map(|position| {
+                let ratio = position / (track_section.length * 1000_f64);
+                (track_section.id.to_string(), ratio)
+            })
+            .collect::<Vec<_>>()
     }
 }
 
@@ -241,39 +252,28 @@ async fn post(
     .await?;
 
     // Get track section offsets.
-    let track_section_offsets = path_properties_input.find_track_section_offset();
+    let track_section_offsets = path_properties_input.find_track_section_offsets();
 
     // Get the IDs of the track sections.
-    let object_ids = path_properties_input
-        .find_track_section_offset()
+    let track_section_ids = track_section_offsets
         .iter()
         .map(|ts| ts.0.clone()) // Extract track section IDs.
         .collect();
 
     // Fetch the track section objects using the IDs.
-    let objects = infra
-        .get_objects(conn, ObjectType::TrackSection, &object_ids)
-        .await?;
+    let track_sections = infra
+        .get_objects(conn, ObjectType::TrackSection, &track_section_ids)
+        .await?
+        .into_iter()
+        .map(|track_section| track_section.railjson)
+        .map(serde_json::from_value::<TrackSection>)
+        .collect::<Result<Vec<_>, serde_json::Error>>()?;
 
-    // Prepare a vector to store position ratios.
-    let mut path_item_position_ratio = Vec::new();
-
-    for object in objects {
-        let track_section =
-            serde_json::from_str::<TrackSection>(object.railjson.to_string().as_str())?;
-
-        // Get the position of the track section from the offsets.
-        let position = track_section_offsets[track_section.id.as_ref()] as f64;
-
-        // Calculate the geographic length of the track section.
-        let geo_length = track_section.geo_bbox().diagonal_length();
-
-        // Calculate the position ratio within the track section.
-        let geo_position = position * geo_length / (track_section.length * 1000_f64);
-        let ratio = geo_position / geo_length;
-
-        path_item_position_ratio.push((track_section.id.to_string(), ratio));
-    }
+    let path_item_position_ratio = track_sections
+        .iter()
+        .map(|ts| path_properties_input.get_ratio(ts))
+        .flatten()
+        .collect::<Vec<(String, f64)>>();
 
     // 1) Try to retrieve all the informations from Valkey
     let mut path_properties = retrieve_path_properties(
@@ -374,11 +374,14 @@ fn path_properties_input_hash(
 #[cfg(test)]
 mod tests {
     use axum::http::StatusCode;
+    use editoast_schemas::infra::{Direction, TrackSection};
     use rstest::rstest;
     use serde_json::json;
 
     use super::PathProperties;
+    use crate::core::pathfinding::TrackRange;
     use crate::models::fixtures::create_small_infra;
+    use crate::views::path::properties::PathPropertiesInput;
     use crate::views::test_app::TestAppBuilder;
 
     #[rstest]
@@ -398,5 +401,56 @@ mod tests {
         assert!(response.electrifications.is_some());
         assert!(response.geometry.is_some());
         assert!(response.operational_points.is_some());
+    }
+
+    #[rstest]
+    fn test_get_track_length_cumulative_sums() {
+        let track_range_1 = TrackRange::new("TR1", 0, 1300, Direction::StartToStop);
+        let track_range_2 = TrackRange::new("TR2", 0, 1000, Direction::StartToStop);
+        let path_properties_input = PathPropertiesInput {
+            track_section_ranges: vec![track_range_1, track_range_2],
+            path_item_positions: vec![0, 2300],
+        };
+        assert_eq!(
+            path_properties_input.get_track_length_cumulative_sums(),
+            vec![0, 1300]
+        );
+    }
+
+    #[rstest]
+    fn test_find_track_section_offset() {
+        let track_range_1 = TrackRange::new("TR1", 0, 400, Direction::StartToStop);
+        let track_range_2 = TrackRange::new("TR2", 0, 600, Direction::StartToStop);
+        let path_properties_input = PathPropertiesInput {
+            track_section_ranges: vec![track_range_1, track_range_2],
+            path_item_positions: vec![0, 500, 1000],
+        };
+        assert_eq!(
+            path_properties_input.find_track_section_offsets(),
+            vec![
+                ("TR1".to_string(), 0),
+                ("TR2".to_string(), 100),
+                ("TR2".to_string(), 600)
+            ]
+        );
+    }
+
+    #[rstest]
+    fn test_get_ratio() {
+        let track_range_1 = TrackRange::new("TR1", 0, 4_000, Direction::StartToStop);
+        let track_range_2 = TrackRange::new("TR2", 0, 5_000, Direction::StartToStop);
+        let path_properties_input = PathPropertiesInput {
+            track_section_ranges: vec![track_range_1, track_range_2],
+            path_item_positions: vec![0, 5_000, 9_000],
+        };
+        let track_section = TrackSection {
+            id: "TR2".into(),
+            length: 5.0,
+            ..Default::default()
+        };
+        assert_eq!(
+            path_properties_input.get_ratio(&track_section),
+            vec![("TR2".to_string(), 0.2), ("TR2".to_string(), 1.0),]
+        );
     }
 }
--- a/editoast/src/views/path/properties.rs
+++ b/editoast/src/views/path/properties.rs
@@ -65,13 +65,12 @@ pub struct PathPropertiesInput {
 
 impl PathPropertiesInput {
     fn get_track_length_cumulative_sums(&self) -> Vec<u64> {
-        let mut cumulative_sums = Vec::new();
+        let mut cumulative_sums = Vec::with_capacity(self.track_section_ranges.len());
         let mut cumulative_sum = 0;
 
-        for track_section in self.track_section_ranges.iter() {
-            let range_length = track_section.end - track_section.begin;
-            cumulative_sum += range_length;
+        for track_range in &self.track_section_ranges {
             cumulative_sums.push(cumulative_sum);
+            cumulative_sum += track_range.length();
         }
 
         cumulative_sums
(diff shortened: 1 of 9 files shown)


front/src/modules/pathfinding/hooks/usePathfinding.ts
```diff
@@ -96,9 +96,10 @@ const usePathfinding = ({
   ) => {
     const pathPropertiesParams: PostInfraByInfraIdPathPropertiesApiArg = {
       infraId,
-      props: ['electrifications', 'geometry', 'operational_points'],
+      props: ['electrifications', 'geometry', 'operational_points', 'path_item_position_ratio'],
```
SharglutDev, line 99: You ask for the `path_item_position_ratio` but you don't take it l.105 to do something with it ?

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
```diff
@@ -60,6 +60,7 @@ const useGetProjectedTrainOperationalPoints = ({
           props: ['operational_points'],
           pathPropertiesInput: {
             track_section_ranges: pathfindingResult.track_section_ranges,
+            path_item_positions: pathfindingResult.path_item_positions,
```
SharglutDev, line 63: Here you pass it in`pathPropertiesInput` but you don't put it in `props` ?

front/src/modules/pathfinding/hooks/usePathfinding.ts
hamz2a: This PR only covers the backend. A separate front PR will be created to integrate these changes. I made a few front adjustments to match the updated API, just enough to let the front build.

front/src/modules/simulationResult/components/SpaceTimeChart/useGetProjectedTrainOperationalPoints.ts
hamz2a: Sorry, but I didn't understand your question?

SharglutDev: You answered my question in the previous comment, thank you :)

editoast/src/views/path/properties.rs
```diff
@@ -52,6 +58,57 @@ editoast_common::schemas! {
...
+        for track_section in self.track_section_ranges.iter() {
+            let range_length = track_section.end - track_section.begin;
+            cumulative_sum += range_length;
+            cumulative_sums.push(cumulative_sum);
+        }
+
+        cumulative_sums
+    }
+
+    pub fn find_track_section_offset(&self) -> HashMap<String, u64> {
```
woshilapin: We should probably add some unit tests for this algorithm.

editoast/src/views/path/properties.rs
```diff
@@ -52,6 +58,57 @@ editoast_common::schemas! {
...
+        let track_length_cumulative_sums = self.get_track_length_cumulative_sums();
+        let mut res = HashMap::new();
+        for path_item_position in self.path_item_positions.iter() {
```
woshilapin: nitpick: this can be written like this (not strictly the same but with equivalent behavior, iteration on references of the `Item`)

```suggestion
        for path_item_position in &self.path_item_positions {
```

editoast/src/views/path/properties.rs
```diff
@@ -52,6 +58,57 @@ editoast_common::schemas! {
...
+    /// The first value is always `0` (beginning of the path) and the last one is always equal to the `length` of the path in mm
+    pub path_item_positions: Vec<u64>,
+}
+
+impl PathPropertiesInput {
+    fn get_track_length_cumulative_sums(&self) -> Vec<u64> {
+        let mut cumulative_sums = Vec::new();
+        let mut cumulative_sum = 0;
+
+        for track_section in self.track_section_ranges.iter() {
```
woshilapin: nitpick: this can be written like this (not strictly the same but with equivalent behavior, iteration on references of the `Item`)

```suggestion
        for track_section in &self.track_section_ranges {
```

editoast/src/views/path/properties.rs
```diff
@@ -52,6 +58,57 @@ editoast_common::schemas! {
...
+            let range_length = track_section.end - track_section.begin;
+            cumulative_sum += range_length;
```
woshilapin:
```suggestion
            cumulative_sum += track_section.length();
```

editoast/src/views/path/properties.rs
```diff
@@ -52,6 +58,57 @@ editoast_common::schemas! {
...
     /// List of track sections
     pub track_section_ranges: Vec<TrackRange>,
+    /// The path offset in mm of each path item given as input of the pathfinding
+    /// The first value is always `0` (beginning of the path) and the last one is always equal to the `length` of the path in mm
+    pub path_item_positions: Vec<u64>,
+}
+
+impl PathPropertiesInput {
+    fn get_track_length_cumulative_sums(&self) -> Vec<u64> {
+        let mut cumulative_sums = Vec::new();
```
woshilapin:
```suggestion
        let mut cumulative_sums = Vec::with_capacity(self.track_section_ranges.len());
```

editoast/src/views/path/properties.rs
```diff
@@ -52,6 +58,57 @@ editoast_common::schemas! {
...
+                .position(|&cumulative_sum| *path_item_position <= cumulative_sum);
+
+            if let Some(track_section_index) = track_section_index {
+                let track_range = self.track_section_ranges[track_section_index].clone();
+
+                let inferior_sum = if track_section_index > 0 {
+                    track_length_cumulative_sums[track_section_index - 1]
+                } else {
+                    0
+                };
```
woshilapin: Looking at this piece of code, it feels that `track_length_cumulative_sums` should contain the start of a track section, and not the end. For example, with 3 tracks of a length of respectively 12, 15, and 3, the current version of the code would produce `[12, 27, 30]` but I propose that it should produce `[0, 12, 27]`. This way, each cell of the vector would directly contain the `inferior_sum`. With this change of semantic, the 5 lines here would just become a no-operation.

editoast/src/views/path/properties.rs
```diff
@@ -52,6 +58,57 @@ editoast_common::schemas! {
...
+        cumulative_sums
+    }
+
+    pub fn find_track_section_offset(&self) -> HashMap<String, u64> {
+        let track_length_cumulative_sums = self.get_track_length_cumulative_sums();
+        let mut res = HashMap::new();
+        for path_item_position in self.path_item_positions.iter() {
+            let track_section_index = track_length_cumulative_sums
+                .iter()
+                .position(|&cumulative_sum| *path_item_position <= cumulative_sum);
```
woshilapin: How about using `.zip` and `.find` instead, to avoid dealing with indexes.

```suggestion
            let track_section_index = self.track_section_ranges
                .iter()
                .zip(track_length_cumulative_sums)
                .find(|(track_range, cumulative_sum)| *path_item_position <= cumulative_sum);
```

If you apply the suggestion below about having the `track_length_cumulative_sums` being the start of each track section (instead of the end), then you might even make the selection here of the right track with the following

```suggestion
            let track_section_index = self.track_section_ranges
                .iter()
                .zip(track_length_cumulative_sums)
                .find(|(track_range, cumulative_sum)| *path_item_position <= cumulative_sum + track_range.length());
```

editoast/src/views/path/properties.rs
woshilapin, line 1: I actually summarized most of my proposition for this file into https://github.com/OpenRailAssociation/osrd/pull/11267.

woshilapin (commented):
Not finished yet but here is a first list of comments.

editoast/src/views/path/properties.rs
```diff
@@ -175,15 +237,53 @@ async fn post(
...
+        infra_id,
+    })
+    .await?;
+
+    // Get track section offsets.
+    let track_section_offsets = path_properties_input.find_track_section_offset();
+
+    // Get the IDs of the track sections.
+    let object_ids = path_properties_input
+        .find_track_section_offset()
```
woshilapin: Let's reuse the variable just above.

```suggestion
    let object_ids = track_section_offsets
```

editoast/src/views/path/properties.rs
```diff
@@ -175,15 +237,53 @@ async fn post(
...
+    let infra = Infra::retrieve_or_fail(conn, infra_id, || PathfindingError::InfraNotFound {
+        infra_id,
+    })
+    .await?;
+
+    // Get track section offsets.
+    let track_section_offsets = path_properties_input.find_track_section_offset();
+
+    // Get the IDs of the track sections.
+    let object_ids = path_properties_input
```
woshilapin: nitpick: Maybe `object_ids` is a pretty vague naming. Aren't they `track_section_ids`?

editoast/src/views/path/properties.rs
```diff
@@ -207,13 +307,14 @@ async fn post(
             geometry: Some(computed_path_properties.geometry),
             operational_points: Some(computed_path_properties.operational_points),
             zones: Some(computed_path_properties.zones),
+            path_item_position_ratio: Some(path_item_position_ratio),
```
woshilapin, line 306: It's interesting because we pay the price of constructing the entire object `PathProperties` with all the properties... and only after, we filter some of them out. This might need a refactor (in another PR), what do you think? (and not necessarily you who need to do it)

editoast/src/views/path/properties.rs
```diff
@@ -175,15 +237,53 @@ async fn post(
...
+
+    // Get the IDs of the track sections.
+    let object_ids = path_properties_input
+        .find_track_section_offset()
+        .iter()
+        .map(|ts| ts.0.clone()) // Extract track section IDs.
+        .collect();
+
+    // Fetch the track section objects using the IDs.
+    let objects = infra
```
woshilapin: nitpick: Same here, let's name that `track_sections`, what do you think?

editoast/src/views/path/properties.rs
```diff
@@ -175,15 +237,53 @@ async fn post(
...
+
+        // Calculate the geographic length of the track section.
+        let geo_length = track_section.geo_bbox().diagonal_length();
+
+        // Calculate the position ratio within the track section.
+        let geo_position = position * geo_length / (track_section.length * 1000_f64);
+        let ratio = geo_position / geo_length;
+
+        path_item_position_ratio.push((track_section.id.to_string(), ratio));
+    }
```
woshilapin: suggestion: this all piece can be written in a more chainable way if you'd like

```rust
    let path_item_position_ratio = objects
        .into_iter()
        .map(|object| object.railjson)
        .map(serde_json::from_value::<TrackSection>)
        .filter_map_ok(|track_section| {
            let position = *track_section_offsets.get(track_section.id.as_ref())? as f64;
            let geo_length = track_section.geo_bbox().diagonal_length();
            let geo_position = position * geo_length / (track_section.length * 1000_f64);
            let ratio = geo_position / geo_length;
            Some((track_section.id.to_string(), ratio))
        })
        .collect::<Result<Vec<_>, serde_json::Error>>()?;
```

editoast/src/views/path/properties.rs
```diff
@@ -175,15 +237,53 @@ async fn post(
...
+
+    // Prepare a vector to store position ratios.
+    let mut path_item_position_ratio = Vec::new();
+
+    for object in objects {
+        let track_section =
+            serde_json::from_str::<TrackSection>(object.railjson.to_string().as_str())?;
+
+        // Get the position of the track section from the offsets.
+        let position = track_section_offsets[track_section.id.as_ref()] as f64;
```
woshilapin: nitpick: Prefer `.get()` instead of the panickable `[]`.

editoast/src/views/path/properties.rs
```diff
@@ -175,15 +237,53 @@ async fn post(
...
+    let objects = infra
+        .get_objects(conn, ObjectType::TrackSection, &object_ids)
+        .await?;
+
+    // Prepare a vector to store position ratios.
+    let mut path_item_position_ratio = Vec::new();
+
+    for object in objects {
+        let track_section =
+            serde_json::from_str::<TrackSection>(object.railjson.to_string().as_str())?;
```
woshilapin: hint: You can use `serde_json::from_value` instead

```suggestion
            serde_json::from_value::<TrackSection>(object.railjson)?;
```

editoast/src/views/path/properties.rs
```diff
@@ -175,15 +237,53 @@ async fn post(
...
+
+        // Get the position of the track section from the offsets.
+        let position = track_section_offsets[track_section.id.as_ref()] as f64;
+
+        // Calculate the geographic length of the track section.
+        let geo_length = track_section.geo_bbox().diagonal_length();
+
+        // Calculate the position ratio within the track section.
+        let geo_position = position * geo_length / (track_section.length * 1000_f64);
+        let ratio = geo_position / geo_length;
```
woshilapin: If we multiply by `geo_length` then divide by it, aren't these 2 lines equivalent to

```suggestion
        let ratio = position / (track_section.length * 1000_f64);
```

editoast/src/views/path/properties.rs
hamz2a: Done.

editoast/src/views/path/properties.rs
hamz2a: Done.

editoast/src/views/path/properties.rs
hamz2a: Done.

editoast/src/views/path/properties.rs
hamz2a: Done.

editoast/src/views/path/properties.rs
hamz2a: Done.

editoast/src/views/path/properties.rs
hamz2a: Thank you for spotting this. I will do it in another PR, just after this one.

editoast/src/views/path/properties.rs
hamz2a: Done.

editoast/src/views/path/properties.rs
hamz2a: Thanks for the suggestion! I added another function `get_ratio` in `PathPropertiesInput` to simplify the code even more.

flomonster (changes requested):
I think there's a misunderstanding of the ticket. The idea was to enrich the back-end endpoints that return an offset in mm along the path with a ratio of the geometry.

In my opinion, this means adding `path_item_position_ratio` to the pathfinding endpoint. 

On the path properties side, it may be necessary to enrich certain properties of the geometric ratio.

- slopes: Not required (not display on a map)
- curves: Not required (not display on a map)
- electrifications: Not required (not display on a map)
- geometry: Not required (does not make sense)
- operational_points: Might make sense ? I don't know if it's actually used. We should check if only selected waypoint are display on the map or all operational points along it.
- zones: Not required (not display on a map)

@bougue-pe Can you confirm my statement?

editoast/src/views/path/properties.rs
```diff
@@ -175,15 +244,42 @@ async fn post(
...
+        .into_iter()
+        .map(|track_section| track_section.railjson)
+        .map(serde_json::from_value::<TrackSection>)
+        .collect::<Result<Vec<_>, serde_json::Error>>()?;
+
+    let path_item_position_ratio = track_sections
+        .iter()
+        .map(|ts| path_properties_input.get_ratio(ts))
+        .flatten()
+        .collect::<Vec<(String, f64)>>();
```
woshilapin, line 276: `track_sections` is not used afterwards, so I would collapse the 2 chain of iterators into one to avoid doing 2 `.collect()`, what do you think?

editoast/src/views/path/properties.rs
woshilapin: How about testing some edge cases too (like what if some of the arrays are empty, or if some offset is bigger than the total length, etc.°

bougue-pe:
To track down things a bit: There is new a more global proposition expressed in https://github.com/OpenRailAssociation/osrd/issues/9372#issuecomment-2765815821
Once (and if) confirmed that it's the way we want to take, the work initiated in this PR will probably change a lot.

flomonster:
@hamz2a Since we change our strategy to fix the bug. Is it okay for you to close this PR? 🙄 
If you'd like to work on the new issue, please go ahead and assign it to yourself and create a new pull request.

hamz2a:
Closing the PR as we’ve changed the strategy to fix the bug.

PR #11251, opened 2025-03-25
editoast: manage authorization on infras

sim51:
Must be merge after #11197

This PR adds the authorization management on the infrastructure. 
It also enable per default the authorization system on editoast. So you have to add the role "Admin"  (or `OperationalStudies`) on your user : 

* In your browser go to http://localhost:4000/api/authz/me and remember your user ID
* In the editoast folder, type the command : ` cargo run roles add 373 Admin` where `373` is your user's ID

### DIFF
--- a/editoast/editoast_authz/src/regulator.rs
+++ b/editoast/editoast_authz/src/regulator.rs
@@ -425,13 +425,81 @@ impl<S: StorageDriver> Regulator<S> {
 
         // Bypass if user is an admin
         if self.is_admin(user).await? {
-            return Ok(Authorization::Bypassed(()));
+            return Ok(Authorization::Bypassed);
         }
 
         // Calling openfga
         let check = self
             .openfga
             .check(model::Infra::can_read().check(user, infra))
+            .await?;
+        Ok(Authorization::from_privilege_check(check))
+    }
+
+    #[tracing::instrument(skip(self), ret(level = Level::DEBUG), err)]
+    pub async fn authorize_infra_write(
+        &self,
+        user: &User,
+        infra: &Infra,
+    ) -> Result<Authorization<()>, Error<S::Error>> {
+        // Check if the infra exists
+        if !self
+            .driver
+            .infra_exists(infra.0)
+            .await
+            .map_err(Error::Storage)?
+        {
+            return Err(Error::UnknownResource(infra.0));
+        }
+
+        // Check if user exists
+        if !self.user_exists(user.0).await? {
+            return Err(Error::UnknownSubject(user.0));
+        }
+
+        // Bypass if user is an admin
+        if self.is_admin(user).await? {
+            return Ok(Authorization::Bypassed);
+        }
+
+        // Calling openfga
+        let check = self
+            .openfga
+            .check(model::Infra::can_write().check(user, infra))
+            .await?;
+        Ok(Authorization::from_privilege_check(check))
+    }
+
+    #[tracing::instrument(skip(self), ret(level = Level::DEBUG), err)]
+    pub async fn authorize_infra_delete(
+        &self,
+        user: &User,
+        infra: &Infra,
+    ) -> Result<Authorization<()>, Error<S::Error>> {
+        // Check if the infra exists
+        if !self
+            .driver
+            .infra_exists(infra.0)
+            .await
+            .map_err(Error::Storage)?
+        {
+            return Err(Error::UnknownResource(infra.0));
+        }
+
+        // Check if user exists
+        if !self.user_exists(user.0).await? {
+            return Err(Error::UnknownSubject(user.0));
+        }
+
+        // Bypass if user is an admin
+        if self.is_admin(user).await? {
+            return Ok(Authorization::Bypassed);
+        }
+
+        // Calling openfga
+        let check = self
+            .openfga
+            .check(model::Infra::can_delete().check(user, infra))
             .await?;
         Ok(Authorization::from_privilege_check(check))
     }
@@ -459,7 +527,7 @@ impl<S: StorageDriver> Regulator<S> {
 
         // Bypass if user is an admin
         if self.check_roles(user, [Role::Admin].into()).await? {
-            return Ok(Authorization::Bypassed(()));
+            return Ok(Authorization::Bypassed);
         }
 
         // Calling openfga
@@ -493,7 +561,7 @@ impl<S: StorageDriver> Regulator<S> {
 
         // Bypass if user is an admin
         if self.check_roles(user, [Role::Admin].into()).await? {
-            return Ok(Authorization::Bypassed(()));
+            return Ok(Authorization::Bypassed);
         }
 
         // Calling openfga
@@ -527,7 +595,7 @@ impl<S: StorageDriver> Regulator<S> {
 
         // Bypass if user is an admin
         if self.check_roles(user, [Role::Admin].into()).await? {
-            return Ok(Authorization::Bypassed(()));
+            return Ok(Authorization::Bypassed);
         }
 
         // Calling openfga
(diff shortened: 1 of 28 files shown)


editoast/src/views/stdcm_logs.rs
```diff
@@ -474,6 +483,7 @@ mod tests {
         let user = app.user("bob", "Bob").with_roles([Role::Admin]).create();
         let _ = execute_stdcm_request(&app, Some(&user)).await;
         let request = app.get("/stdcm_logs").by_user(&user);
+        println!("Request: {:?}", request);
```
```suggestion
```

editoast/src/views/stdcm_logs.rs
```diff
@@ -320,8 +320,16 @@ mod tests {
...
     async fn execute_stdcm_request(app: &TestApp, user: Option<&subject::User>) -> String {
-        let small_infra = create_small_infra(&mut app.db_pool().get_ok()).await;
         let timetable = create_timetable(&mut app.db_pool().get_ok()).await;
+        // create an infra and add the provided user as infra owner
+        let small_infra = create_small_infra(&mut app.db_pool().get_ok()).await;
+        if let Some(user) = user {
+            let _ = app
+                .regulator()
+                .grant_infra_owner_unchecked(user.id, small_infra.id)
+                .await;
```
leovalais: use expect to avoid silent failures

editoast/src/views/mod.rs
```diff
@@ -200,6 +201,78 @@ impl Authentication {
...
+                        tracing::debug!(
+                            "User is not authorized to delete infra {infra_id}: {reason}"
+                        );
+                        Ok(false)
+                    }
+                    _ => Ok(true),
+                }
+            }
+        }
+    }
```
leovalais: These functions can be easily factorize and could delegate the task of converting the unauthorization into an error to the `Authorization` object.

We could have something in the lines of:

```rust
async fn authorize<E: Into<AuthorizationError>>(self, f: impl AsyncFnOnce(Authorizer<PgAuthDriver>) -> Result<Authorization<()>, E>) -> Result<(), AuthorizationError> {
	match self {
		Authenticated(authorizer) => f(authorizer).await?.allowed().map_err(|Unauthorized| AuthorizationError::Forbidden)
		...
	}
}
```

so we can just write in each endpoint:

```rust
auth.authorize(async |authorizer| authorizer.authorize_infra_read(...).await).await?;
```

That brings us a one-liner and hides error management.

wdyt?

editoast/src/views/infra/mod.rs
```diff
@@ -222,21 +221,30 @@ async fn list(
...
+                conn,
+                settings.filter(move || Infra::ID.eq_any(infra_ids.clone())),
+            )
+            .await?
+        }
+        Authorization::Bypassed => Infra::list_paginated(conn, settings).await?,
+        Authorization::Denied { reason: _ } => {
+            return Err(AuthorizationError::Unauthorized.into());
+        }
     };
```
leovalais: nit: the match could more precisely just return the settings, the list_paginated doesn't really need to change

leovalais (commented):
LGTM, just details. I'll test that asap ^^'

editoast/src/views/mod.rs
sim51: IMHO, all the `check` functions (check_roles, check_infra_can_read, ...) should return a Forbidden or Unauthenticated error, so we can have  a one line call in views. 

But making an `authorize` function that takes in argument a function with the authorizer can help to factorize the code.  We need to be sure that the arms `Authentication::SkipAuthorization` returns true, and `Unauthenticated` returns false (we don't have the authorizer when skipping auth). 
So I would rename this function `checkAuthorize` (for me a check returns true/false or Ok/Error).

On the other side, devs should know the API of the authorizer and not only the one from Auth, but I think it's a acceptable tradeoff, to avoid code duplication in checks function.

editoast/src/views/infra/mod.rs
sim51: The match also has the Deny arm, that's why I did that, even if at this step we are sure that the user is connected and with the good permissions.
Otherwise I need a default arms where I give access to  the full access to the infra list.

editoast/src/views/mod.rs
sim51: I did the job for the infra_check, but we also need to do it check role. 
I did not do it, otherwise this PR will modify all views and so will be huge.

sim51:
FYI, I also added in the readme, how to add the 'admin' role to the default user.

sim51:
FYI my last commit fix an issue on the `allowed_ten_try` function, you can check the commit log to see why.

sim51:
FYI, fix an issue about admin that can not revoke a grant

editoast/src/views/infra/mod.rs
leovalais: Default is implemented for `SelectionSettings`. I meant something in the lines of:

```suggestion
	let settings = match auth.list_authorized_infra().await? {
        Authorization::Granted(infra_ids) => {
            settings.filter(move || Infra::ID.eq_any(infra_ids.clone()))
        }
        Authorization::Bypassed => Default::default(),
        Authorization::Denied { reason } => {
            unreachable!("user is authenticated at this point: {reason}")
        }
    };
    let (infras, stats) = Infra::list_paginated(conn, settings).await?;
```

There is an `unreachable!` macro for this use case. It panics if it we reach it. It helps highlight the bug instead of silently ignoring it.

editoast/src/views/mod.rs
leovalais: Thanks for making the change!

* Agreed about the `check` prefix since we only accept `Authorization<()>`.  (I have in mind that we could later make that function generic over some `T`, in which case I'm not so sure about the prefix, but that's for later.  Better being correct now.)  Nitpick: I believe `check_authorized` would be more correct.
* I think everyone programming with permissions should be comfortable using the `Authorizer` at least.  In my mind, the `Authentication` is just a way to force us to deal with the different authentication modes of editoast.  It shouldn't deal with permissions (or be at most a *very thin* wrapper only where it makes sense).
* The long-term plan we have for roles is to have a middleware in an `axum` router perform the role-checking.  This way we can configure it for entire path sub-trees and remove all explicit role checks in endpoints.  So I'm not sure if we even need to bother making the change.

leovalais (commented):
Still LGTM, thanks for the changes!

editoast/src/views/infra/mod.rs
sim51: I was not aware of this macro, thanks. 
Will change that

README.md
```diff
@@ -43,6 +43,9 @@ To compile and run the application with an example infrastructure:
 # build and run the entire stack
 docker compose up -d --build
 
+# Add the admin role to the default user
+docker exec osrd-editoast editoast roles add mock/mocked Admin
```
flomonster: 1. This won't work if the user did not open the application beforehand. (No request has been sent and so the user does not exist in DB).
2. It's kind of painful to run another command.

An idea could be to adapt the main docker compose to create the user and add it's role before running editoast. Something like:

```yaml
    command:
      - /bin/sh
      - -c
      - "diesel migration run --locked-schema && editoast user add 'mock/mocked' 'Example User' && editoast roles add 'mock/mocked' Admin && exec editoast runserver"
```

We might want to format this command a little bit 😅

sim51: I implemented your solution in my last commit and it works, thanks!

flomonster (approved):
Ops: LGTM

editoast/editoast_authz/src/regulator.rs
```diff
@@ -635,6 +703,30 @@ impl<S: StorageDriver> Regulator<S> {
         Ok(users)
     }
 
+    /// Get IDS of infras a subject can read
+    pub async fn list_authorized_infra(
+        &self,
+        user: &User,
+    ) -> Result<Authorization<Vec<i64>>, Error<S::Error>> {
```
leovalais:
```suggestion
    ) -> Result<Authorization<Vec<Infra>>, Error<S::Error>> {
```

you can use `authz::model` types in return position as well (avoids the iteration in the end)

leovalais (approved):

editoast/editoast_authz/src/regulator.rs
sim51: done

PR #11258, opened 2025-03-26
editoast: rename paced train fields

younesschrifi:
This PR helps to unify front and back naming for the paced train model. 
We've chosen `time_window` instead of `duration` and `interval` instead of `step`.

### DIFF
--- a/editoast/editoast_schemas/src/paced_train.rs
+++ b/editoast/editoast_schemas/src/paced_train.rs
@@ -12,10 +12,10 @@ editoast_common::schemas! {
 pub struct Paced {
     /// Duration of the paced train, an ISO 8601 format is expected
     #[schema(value_type = chrono::Duration, example = "PT2H")]
-    pub duration: PositiveDuration,
+    pub time_window: PositiveDuration,
     /// Time between two occurrences, an ISO 8601 format is expected
     #[schema(value_type = chrono::Duration, example = "PT15M")]
-    pub step: PositiveDuration,
+    pub interval: PositiveDuration,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, ToSchema)]
--- a/editoast/migrations/2025-03-26-090235_rename_paced_train_fields/up.sql
+++ b/editoast/migrations/2025-03-26-090235_rename_paced_train_fields/up.sql
@@ -1,0 +1,2 @@
+ALTER TABLE paced_train RENAME COLUMN step TO interval;
+ALTER TABLE paced_train RENAME COLUMN duration TO time_window;
--- a/editoast/editoast_models/src/tables.rs
+++ b/editoast/editoast_models/src/tables.rs
@@ -452,8 +452,8 @@ diesel::table! {
         speed_limit_tag -> Nullable<Varchar>,
         power_restrictions -> Jsonb,
         options -> Jsonb,
-        duration -> Interval,
-        step -> Interval,
+        time_window -> Interval,
+        interval -> Interval,
     }
 }
 
--- a/editoast/migrations/2025-03-26-090235_rename_paced_train_fields/down.sql
+++ b/editoast/migrations/2025-03-26-090235_rename_paced_train_fields/down.sql
@@ -1,0 +1,2 @@
+ALTER TABLE paced_train RENAME COLUMN interval TO step;
+ALTER TABLE paced_train RENAME COLUMN time_window TO duration;
--- a/editoast/openapi.yaml
+++ b/editoast/openapi.yaml
@@ -1706,17 +1706,17 @@ paths:
                   paced:
                     type: object
                     required:
-                    - duration
-                    - step
+                    - time_window
+                    - interval
                     properties:
-                      duration:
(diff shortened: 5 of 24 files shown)


editoast/editoast_schemas/src/paced_train.rs
```diff
@@ -12,10 +12,10 @@ editoast_common::schemas! {
 pub struct Paced {
     /// Duration of the paced train, an ISO 8601 format is expected
     #[schema(value_type = chrono::Duration, example = "PT2H")]
-    pub duration: PositiveDuration,
+    pub time_window: PositiveDuration,
     /// Time between two occurrences, an ISO 8601 format is expected
     #[schema(value_type = chrono::Duration, example = "PT15M")]
-    pub step: PositiveDuration,
+    pub interval: PositiveDuration,
```
leovalais, line 18: do we whave a CHECK constraint in DB to ensure this always holds?

leovalais (approved):
LGTM if we have this constraint

editoast/migrations/2025-03-26-090235_rename_paced_train_fields/up.sql
woshilapin, line 1: Do we have experience with such migrations of renaming existing fields? Because the migration will be run first, and then the existing deployed `editoast` instance will still try to access the old names for a while (until the new version is deployed). So that means some kind of downtime for everything that touches paced train I suppose?

But maybe this is acceptable?

flomonster: We already perform this kind of migration. In my opinion, it's perfectly acceptable.

Maymanaf (approved):

editoast/migrations/2025-03-26-090235_rename_paced_train_fields/up.sql
Khoyo: Since the migration will only happen when an editoast pod is already starting up, the downtime should be minimal. This is not *zero* downtime but I don't think we try to maintain that.

editoast/editoast_schemas/src/paced_train.rs
flomonster: We don't have constraint on these fields:

```
osrd=# \d paced_train;
                                          Table "public.paced_train"
         Column          |           Type           | Collation | Nullable |             Default
-------------------------+--------------------------+-----------+----------+----------------------------------
 id                      | bigint                   |           | not null | generated by default as identity
 train_name              | character varying(128)   |           | not null |
 labels                  | text[]                   |           | not null |
 rolling_stock_name      | character varying(128)   |           | not null |
 timetable_id            | bigint                   |           | not null |
 start_time              | timestamp with time zone |           | not null |
 schedule                | jsonb                    |           | not null |
 margins                 | jsonb                    |           | not null |
 initial_speed           | double precision         |           | not null |
 comfort                 | smallint                 |           | not null |
 path                    | jsonb                    |           | not null |
 constraint_distribution | smallint                 |           | not null |
 speed_limit_tag         | character varying(128)   |           |          |
 power_restrictions      | jsonb                    |           | not null |
 options                 | jsonb                    |           | not null |
 duration                | interval                 |           | not null |
 step                    | interval                 |           | not null |
Indexes:
    "paced_train_pkey" PRIMARY KEY, btree (id)
    "pga_idx_fk_paced_train_timetable_id" btree (timetable_id)
Foreign-key constraints:
    "paced_train_timetable_id_fkey" FOREIGN KEY (timetable_id) REFERENCES timetable(id) ON DELETE CASCADE
```

younesschrifi: Do I have to add a constraint ?

flomonster: No, we don't have a constraint for other Positive duration fields.

leovalais: > we don't have a constraint for other Positive duration fields

and that's a bad thing imo

But we can do that in another PR, agreed on that

SharglutDev (approved):

front/src/modules/trainschedule/components/ManageTrainSchedule/AddTrainScheduleButton.tsx
```diff
@@ -79,8 +79,8 @@ const AddTrainScheduleButton = ({
           const pacedTrainPayload: PacedTrain = {
             ...basePacedTrainPayload,
             paced: {
-              duration: validTimetableItemConfig.timeRangeDuration,
-              step: validTimetableItemConfig.cadence,
+              time_window: validTimetableItemConfig.timeRangeDuration,
+              interval: validTimetableItemConfig.cadence,
```
emersion: Note, we should also write a follow-up PR to adjust our types to match.

SharglutDev: Created the issue here : https://github.com/OpenRailAssociation/osrd/issues/11297

shenriotpro (approved):
OK for tests/

PR #11266, opened 2025-03-26
Fix possible race condition of OpenFGA model upload

leovalais:


flomonster (approved):

editoast/editoast_authz/src/lib.rs
```diff
@@ -31,12 +31,14 @@ pub async fn ensure_latest_authorization_model(
             client.update_authorization_model(&model).await?;
         }
         delta => {
-            let model = fga::compile_model(AUTHORIZATION_MODEL);
-            tracing::warn!(
+            // This can happen if multiple pods are spawned simultaneously resulting
+            // in a race condition where multiple models are uploaded at the same time.
+            tracing::error!(
```
woshilapin, line 36: We're tracing an error but don't panic, is it fine?

leovalais: It's intentional, we detected an error case and we can't do anything about it currently. Logging is still useful to notice the issue. Panicking would send editoast in a startup crash loop.

PR #11269, opened 2025-03-26
editoast: add /authz/me route and improve authn testing utils

leovalais:
requires changes from #11169

### DIFF
--- a/editoast/src/views/authz.rs
+++ b/editoast/src/views/authz.rs
@@ -15,12 +15,15 @@ use super::AuthorizerError;
 use super::Regulator;
 
 crate::routes! {
-    "/authz/roles" => {
-        "/me" => list_current_roles,
-        "/{user_id}" => {
-            list_user_roles,
-            grant_roles,
-            strip_roles,
+    "/authz" => {
+        "/me" => whoami,
+        "/roles" => {
+            "/me" => list_current_roles,
+            "/{user_id}" => {
+                list_user_roles,
+                grant_roles,
+                strip_roles,
+            },
         },
     },
 }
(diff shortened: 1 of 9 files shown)


editoast/src/views/authz.rs
```diff
@@ -53,6 +56,53 @@ enum NoSuchUserError {
...
+)]
+async fn whoami(
+    State(AppState { config, .. }): State<AppState>,
+    Extension(auth): AuthenticationExt,
+) -> Result<Json<WhoamiResponse>> {
+    // This is especially necessary as this endpoint is always queried by the frontend
+    // when loading. Making it return 401 results in a blank page.
+    // We return `Admin` as when no authorization is enabled, we want everyone to have
+    // access to full feature set of OSRD.
+    if !config.enable_authorization {
```
flomonster: We could handle this in the `Authorizer`. The `SkipAuthorization` could contain username, identity information. 

Here a suggestion of new workflow:
1. Try retrieve user information using regulator
2. If we must skip authorization then use `SkipAuthorization{user_identity, id, name}` variant
    - If no information retrieve then provide default values (like you're doing in this endpoint)
3. Otherwise create an `Authenticate` variant with all information
    - If no information retrieve then fail with a 403 (current behavior)

leovalais: Not sure what you mean. `authentication == SkipAuthorization` only happens in two cases:

- the header was put there by Core to bypass the authorization, and Core will never call `/authz/me`. This header is also blocked by the gateway, so you can't put it there using curl unless querying editoast directly.
- the authorization is disabled using the server config

In the latter, I'm not sure what the benefit is, especially now that we have impersonation?

editoast/src/views/authz.rs
```diff
@@ -53,6 +56,53 @@ enum NoSuchUserError {
...
+        return Ok(Json(WhoamiResponse {
+            // TODO: don't return -1 and a hardcoded name, return a different schema instead, requires fontend changes
```
woshilapin:
```suggestion
            // TODO: don't return -1 and a hardcoded name, return a different schema instead, requires frontend changes
```

woshilapin (commented):
Only reviewed the last commit (as I understood the first one was part of another merged PR). Didn't see anything problematic, but might be missing the big picture 🤷

woshilapin (approved):

editoast/src/views/authz.rs
leovalais: I did that, not sure if that's what you meant, but I think the operations order is a bit more clear like this anyway: d17fefd84

flomonster: It's almost what I meant. Thanks :pray:

editoast/src/views/authz.rs
```diff
@@ -53,6 +57,50 @@ enum NoSuchUserError {
...
+    let authorizer = auth.authorizer()?;
+    let user_roles = authorizer
+        .user_roles()
+        .await
+        .map_err(AuthzError::Authorizer)?;
+    Ok(Json(WhoamiResponse {
+        id: authorizer.user_id(),
+        name: authorizer.user_name().to_owned(),
+        roles: user_roles.into_iter().collect(),
+    }))
```
flomonster: We can now simplify this with something like this:

```suggestion
      Ok(Json(WhoamiResponse {
        id: auth.user_id()?.unwrap_or(-1), // This already exists
        name: auth.user_name()?.unwrap_or("OSRD user".into()) // This function must be created like `user_id`.
        roles: auth.user_roles()?, // Same here the return type should be something like `Result<HashSet<Role>, AuthorizationError>`
    }))
```

The idea behind this suggestion is to limit the usage of the `authorizer` as much as possible. What do you think?

kmer2016 (approved):

editoast/src/views/authz.rs
leovalais: 2cc3bd402

But I'm not a huge fan of inventing data (-1, "OSRD user") just to match a schema. I think that whoami should return a special response (think schema) when we're skipping the authorization. Otherwise the clients will build on top of these magic values and things can get brittle very fast.

Fine by me for now, we have better things to do, but I think that implementation should change eventually.

flomonster (approved):
LGTM

PR #11270, opened 2025-03-27
front: mailto for remarks must not open an blank tab

### LINKED ISSUE #11206: stdcm: Mailto for remarks opens a blank tab
### What happened?

https://github.com/user-attachments/assets/0f159c44-03ef-424c-9568-48bd414a26bc

When clicking on the button to send a feedback, it opens a new blank tab.

It appears to happen only on several browsers (firefox for all, chrome for some).

Clicking on the button should not open a new blank tab.

1. Go to LMR and launch a simulaiton
2. Try to send a feedback

Recette (SNCF)

Firefox

82cd094

theocrsb:
closes [#11206](https://github.com/OpenRailAssociation/osrd/issues/11206)

### DIFF
--- a/front/src/applications/stdcm/components/StdcmResults/StdcmFeedback.tsx
+++ b/front/src/applications/stdcm/components/StdcmResults/StdcmFeedback.tsx
@@ -60,7 +60,7 @@ ${separator}
   const mailtoLink = `mailto:${stdcmFeedbackMail}?subject=${subject}&body=${body}`;
 
   const handleEmailClick = () => {
-    window.open(mailtoLink, '_blank');
+    window.open(mailtoLink, '_self');
   };
 
   return (


RomainValls (approved):
LGTM

SharglutDev:
Hmm I don't think we should remove this new tab. I don't have a mail app on my computer (my mail is opened in the browser) so now if I try to send a mail, I lose the stdcm page, it's pretty annoying. What do you think @Arthur-Lefebvre ?

Arthur-Lefebvre:
After some readings, it's better to have a target="_blank" for all users.

The default behavior, which is requested, is to open in a new tab if you don't have an e-mail application set, or your in-browser mail app in a new tab if you don't have it. I think we want to avoid the LMR tab to be lost.

However, the behaviour seems to be highly related to your browser and your settings. I found multiple links about it, the most interstings are : 
- https://stackoverflow.com/questions/17517600/mailto-links-do-nothing-in-chrome-but-work-in-firefox
- https://mtsknn.fi/blog/tricky-mailto-links/

In conclusion, like in the second link, I think we have to stick with the target='_blank' as closing a blank tab is less anoying than the loss of your simulation. Unless there is another solution ?

PR #11271, opened 2025-03-27
core: stdcm: fix 20s margin bug

### LINKED ISSUE #11186: core: stdcm: properly handle 20s margin before stop end
In France, we consider that when a train restarts after stopping, it must see a green signal at least 20 secondes before departing.

This is something we have in the conflict detection (see [here](https://github.com/OpenRailAssociation/osrd/blob/dev/core/src/main/java/fr/sncf/osrd/standalone_sim/ScheduleMetadataExtractor.kt#L42-L43)). But we haven't properly thought about it for STDCM. It causes *issues*. There's some 500 (such as https://github.com/OpenRailAssociation/osrd/issues/11121).

But it also causes false negatives. When stopping, we check that there's no conflict during the stop duration, and *only then* we check if we should lengthen it to avoid conflicts. But the API can't let us tell appart conflicts caused during the stop by being there and conflicts caused by this margin, so we just drop the solution if there's any conflict. Even when they could have been easily avoided by delaying the departure. 

Fixing this is not easy. There are several options, some more "clean" than others:

1. Refine the incremental spacing resource generator API to tell as input "the train stop for at least this long but the departure isn't planned yet" (may not be enough)
2. Refine the *output* instead of input, to tell apart the resources that come from those 20s (? - unsure)
3. Tweak the inputs/outputs to reach a similar result to 1 or 2, but without changing the spacing resource generator (may lead to confusing and messy pipelines)

eckter:
Fix https://github.com/OpenRailAssociation/osrd/issues/11186

Fix this issue by:

1. Tweaking inputs for spacing resource requirements: infra explorers that end at stop don't generate requirements for anything after the stop itself
2. Tweaking `BlockAvailability` to account for this margin when identifying conflicts after a stop

I'm not 100% convinced this is the right approach, but it works. I'll leave it there, maybe not merge during the feature freeze, and see if we can come up with something better.

Also add two unit tests: one for STDCM that reproduces this precise case (and fails without the last commit), and another one for spacing ressource generation to properly identify its behavior around partial stops.

Most of the diff comes from new tests, the PR isn't as large as it seems.

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/conflicts/IncrementalRequirementEnvelopeAdapter.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/conflicts/IncrementalRequirementEnvelopeAdapter.kt
@@ -12,6 +12,12 @@ class IncrementalRequirementEnvelopeAdapter(
     private val rollingStock: PhysicsRollingStock,
     private val envelopeWithStops: EnvelopeInterpolate?,
     override var simulationComplete: Boolean,
+
+    // If set to true, we consider that the train doesn't leave its current stop (yet).
+    // Used to evaluate conflicts up to a stop, including the stop itself,
+    // but without considering the reservation on blocks/routes after the stop
+    // (especially to avoid being bothered by the reservation margin before restarting after a stop
+    // on closed signal).
     private val infiniteLastStop: Boolean = false,
 ) : IncrementalRequirementCallbacks {
     override fun maxSpeedInRange(
@@ -46,6 +52,9 @@ class IncrementalRequirementEnvelopeAdapter(
             pastStop = endPos
         }
         if (pastStop == endPos && infiniteLastStop) {
+            // The requested stop is at the end of the simulated path,
+            // and the requested behavior is to make that stop "infinite"
+            // (pushing next reservation's start-time to "never").
             return Double.POSITIVE_INFINITY
         }
         return envelopeWithStops.interpolateDepartureFrom(pastStop)
@@ -96,6 +105,7 @@ class IncrementalRequirementEnvelopeAdapter(
             rollingStock,
             envelopeWithStops, // This is effectively read-only, we don't need a deep copy here
             simulationComplete,
+            infiniteLastStop,
         )
     }
 }
--- a/core/src/main/kotlin/fr/sncf/osrd/conflicts/IncrementalRequirementEnvelopeAdapter.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/conflicts/IncrementalRequirementEnvelopeAdapter.kt
@@ -12,6 +12,7 @@ class IncrementalRequirementEnvelopeAdapter(
     private val rollingStock: PhysicsRollingStock,
     private val envelopeWithStops: EnvelopeInterpolate?,
     override var simulationComplete: Boolean,
+    private val infiniteLastStop: Boolean = false,
 ) : IncrementalRequirementCallbacks {
     override fun maxSpeedInRange(
         pathBeginOff: Offset<TravelledPath>,
@@ -43,6 +44,9 @@ class IncrementalRequirementEnvelopeAdapter(
         var pastStop = (stopOffset.distance).meters
         if (pastStop > endPos) {
             pastStop = endPos
+        }
+        if (pastStop == endPos && infiniteLastStop) {
+            return Double.POSITIVE_INFINITY
         }
         return envelopeWithStops.interpolateDepartureFrom(pastStop)
     }
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
@@ -4,6 +4,7 @@ import fr.sncf.osrd.conflicts.*
 import fr.sncf.osrd.envelope_utils.DoubleBinarySearch
 import fr.sncf.osrd.sim_infra.api.BlockPath
 import fr.sncf.osrd.sim_infra.api.TravelledPath
+import fr.sncf.osrd.standalone_sim.CLOSED_SIGNAL_RESERVATION_MARGIN
 import fr.sncf.osrd.standalone_sim.result.ResultTrain.SpacingRequirement
 import fr.sncf.osrd.stdcm.infra_exploration.InfraExplorerWithEnvelope
 import fr.sncf.osrd.stdcm.infra_exploration.LocatedStep
@@ -213,6 +214,24 @@ data class BlockAvailability(
             if (needFullRequirements) infraExplorer.getFullSpacingRequirements()
             else infraExplorer.getSpacingRequirements()
 
+        // Ensures that we account for the 20s margin where a train must see a green signal before
+        // its departure: when the current segment starts at a stop, we want to consider conflicts
+        // up to 20s before the "start time".
+        val startsAtStop =
+            infraExplorer
+                .getStepTracker()
+                .getSeenSteps()
+                .drop(1) // Skip departure step
+                .filter { it.originalStep.stop }
+                .map {
+                    infraExplorer.getIncrementalPath().fromTravelledPath(it.travelledPathOffset)
+                }
+                .contains(startOffset)
+        var minRelevantTime = startTime
+        if (startsAtStop) {
+            minRelevantTime -= CLOSED_SIGNAL_RESERVATION_MARGIN
+        }
+
         // Modify the spacing requirements to adjust for the start time,
         // and filter out the ones that are outside the relevant time range
         val shiftedSpacingRequirements =
@@ -220,7 +239,7 @@ data class BlockAvailability(
                 .map {
                     SpacingRequirement(
                         it.zone,
-                        max(pathStartTime + it.beginTime, startTime),
+                        max(pathStartTime + it.beginTime, minRelevantTime),
                         min(pathStartTime + it.endTime, endTime),
                         it.isComplete
                     )
--- a/core/src/test/kotlin/fr/sncf/osrd/stdcm/FullSTDCMTests.kt
+++ b/core/src/test/kotlin/fr/sncf/osrd/stdcm/FullSTDCMTests.kt
@@ -231,10 +231,10 @@ class FullSTDCMTests {
     }
 
     /**
-     * Very long stop with an occupancy that starts after the stop end. The requested arrival time
-     * may make us stay longer at the stop location, to the point of causing a conflict. Not finding
-     * a solution is valid, we're looking for crashes (specifically postprocessing assertions).
-     * Reproduces a bug.
+     * Very long stop with an occupancy that starts after the stop end at the stop location. The
+     * requested arrival time may make us stay longer at the stop location, to the point of causing
+     * a conflict. Not finding a solution is valid, we're looking for crashes (specifically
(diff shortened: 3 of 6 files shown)


core/src/test/kotlin/fr/sncf/osrd/stdcm/FullSTDCMTests.kt
```diff
@@ -266,6 +266,40 @@ class FullSTDCMTests {
...
+                .setInfra(infra)
+                .setStartTime(0.0)
+                .setStartLocations(start.blockLocations)
+                .addStep(STDCMStep(stop.blockLocations, 5_000.0, true))
+                .setEndLocations(end.blockLocations)
+                .setBlockAvailability(makeBlockAvailability(requirements))
+                .setMaxRunTime(Double.POSITIVE_INFINITY)
+                .setMaxDepartureDelay(0.0)
+                .run()!!
+        assertTrue(res.stopResults.first().duration > 5_000.0)
```
Erashin: We could test that it's `>= 5_000.0 + CLOSED_SIGNAL_RESERVATION_MARGIN`.

core/src/main/kotlin/fr/sncf/osrd/conflicts/IncrementalRequirementEnvelopeAdapter.kt
```diff
@@ -12,6 +12,7 @@ class IncrementalRequirementEnvelopeAdapter(
     private val rollingStock: PhysicsRollingStock,
     private val envelopeWithStops: EnvelopeInterpolate?,
     override var simulationComplete: Boolean,
+    private val infiniteLastStop: Boolean = false,
```
Erashin, line 21: We probably need a comment here. The name is ok-ish (tbh, i can't find anything better), but we need some context as to what this does exactly from the get-go.

core/src/main/kotlin/fr/sncf/osrd/conflicts/IncrementalRequirementEnvelopeAdapter.kt
```diff
@@ -44,6 +45,9 @@ class IncrementalRequirementEnvelopeAdapter(
         if (pastStop > endPos) {
             pastStop = endPos
         }
+        if (pastStop == endPos && infiniteLastStop) {
```
Erashin, line 54: Probably needs a comment as well.

core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
```diff
@@ -213,14 +214,32 @@ data class BlockAvailability(
...
+            infraExplorer
+                .getStepTracker()
+                .getSeenSteps()
+                .drop(1) // Skip departure step
+                .filter { it.originalStep.stop }
+                .map {
+                    infraExplorer.getIncrementalPath().fromTravelledPath(it.travelledPathOffset)
+                }
+                .contains(startOffset)
+        var minRelevantTime = startTime
```
Erashin, line 230: What do you think about `startTimeWithMargin`?

core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
```diff
@@ -213,14 +214,32 @@ data class BlockAvailability(
             if (needFullRequirements) infraExplorer.getFullSpacingRequirements()
             else infraExplorer.getSpacingRequirements()
 
+        // Ensures that we account for the 20s margin where a train must see a green signal before
+        // its departure: when the current segment starts at a stop, we want to consider conflicts
+        // up to 20s before the "start time".
+        val startAtStop =
```
Erashin: `startsAtStop`

Erashin (commented):
Hmm. Honestly think this solution is fine. Can't think of anything else off the top of my head.

core/src/test/kotlin/fr/sncf/osrd/stdcm/FullSTDCMTests.kt
bougue-pe: Not sure I understand correctly:
Should we check that stop departure _time_ is `~= 7000 + MARGIN`?
The check on duration seems a bit pointless: we check that we find a valid solution, but that's all, no?

eckter: The intent for this assert was specifically to test that the stop was lengthened compared to what was initially requested. Otherwise the test wouldn't have actually tested anything.

This assert ensures that the test itself is functional, it doesn't have much to do with the intended behavior itself.

core/src/main/kotlin/fr/sncf/osrd/conflicts/IncrementalRequirementEnvelopeAdapter.kt
eckter: Fixed in https://github.com/OpenRailAssociation/osrd/pull/11271/commits/a9bd554fe497c5a3287ddb47988e47b46d42c7dd

core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
eckter: The intent for this variable is to bound the results to the relevant range, the margin thing explains how it's computed but I'd prefer it to describe what it actually does

core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
eckter: Fixed in https://github.com/OpenRailAssociation/osrd/pull/11271/commits/a9bd554fe497c5a3287ddb47988e47b46d42c7dd

core/src/main/kotlin/fr/sncf/osrd/conflicts/IncrementalRequirementEnvelopeAdapter.kt
eckter: Fixed in https://github.com/OpenRailAssociation/osrd/pull/11271/commits/a9bd554fe497c5a3287ddb47988e47b46d42c7dd

core/src/test/kotlin/fr/sncf/osrd/stdcm/FullSTDCMTests.kt
eckter: Commented in https://github.com/OpenRailAssociation/osrd/pull/11271/commits/b906adf8ac2c3c3cb6649e639775d5b1d9175272

Erashin: Typo: `can be`

core/src/main/kotlin/fr/sncf/osrd/stdcm/preprocessing/implementation/BlockAvailability.kt
Erashin: Ok, fine for me then

core/src/main/kotlin/fr/sncf/osrd/conflicts/IncrementalRequirementEnvelopeAdapter.kt
```diff
@@ -12,6 +12,11 @@ class IncrementalRequirementEnvelopeAdapter(
     private val rollingStock: PhysicsRollingStock,
     private val envelopeWithStops: EnvelopeInterpolate?,
     override var simulationComplete: Boolean,
+
+    // If set to true, we consider that the train doesn't leave its current stop (yet).
+    // Used to evaluate conflicts up to a stop, including the stop itself,
+    // but without considering the reservation margin for what's coming after the stop.
```
bougue-pe:
```suggestion
    // but without considering the reservation on blocks/routes after the stop (especially avoid being bothered by reservation margin before restarting from a stop on closed signal).
```

core/src/main/kotlin/fr/sncf/osrd/conflicts/IncrementalRequirementEnvelopeAdapter.kt
```diff
@@ -44,6 +49,11 @@ class IncrementalRequirementEnvelopeAdapter(
         if (pastStop > endPos) {
             pastStop = endPos
         }
+        if (pastStop == endPos && infiniteLastStop) {
+            // The requested stop is at the end of the path,
+            // and the last stop is flagged as "infinite".
```
bougue-pe: Not sure:
```suggestion
            // The requested stop is at the end of the simulated path,
            // and the requested behavior is to make that stop "infinite" (pushing next reservation's start-time to "never").
```

core/src/test/kotlin/fr/sncf/osrd/stdcm/FullSTDCMTests.kt
```diff
@@ -266,6 +266,41 @@ class FullSTDCMTests {
...
+                .setStartTime(0.0)
+                .setStartLocations(start.blockLocations)
+                .addStep(STDCMStep(stop.blockLocations, 5_000.0, true))
+                .setEndLocations(end.blockLocations)
+                .setBlockAvailability(makeBlockAvailability(requirements))
+                .setMaxRunTime(Double.POSITIVE_INFINITY)
+                .setMaxDepartureDelay(0.0)
+                .run()!!
+        assertTrue(res.stopResults.first().duration >= 5_000.0) {
+            "if a solution can be found without lengthening the stop, the test itself is broken"
```
bougue-pe: From the "comment" I expected:
```suggestion
        assertTrue(res.stopResults.first().duration > 5_000.0) {
            "if a solution can be found without lengthening the stop, the test itself is broken"
```

bougue-pe (approved):
LGTM after your side-explanations (tried to suggest places to report it).
Not so sharp on STDCM, I let you see if @Erashin must approve too 🙂

Erashin (approved):

core/src/main/kotlin/fr/sncf/osrd/conflicts/IncrementalRequirementEnvelopeAdapter.kt
eckter: Applied in https://github.com/OpenRailAssociation/osrd/pull/11271/commits/8f2263f3fccd1fde472d3eed77eb90c0c351a928

core/src/main/kotlin/fr/sncf/osrd/conflicts/IncrementalRequirementEnvelopeAdapter.kt
eckter: Applied in https://github.com/OpenRailAssociation/osrd/pull/11271/commits/8f2263f3fccd1fde472d3eed77eb90c0c351a928

core/src/test/kotlin/fr/sncf/osrd/stdcm/FullSTDCMTests.kt
eckter: Good catch, I accidentally included a local change I should have reverted. Fixed in https://github.com/OpenRailAssociation/osrd/pull/11271/commits/8f2263f3fccd1fde472d3eed77eb90c0c351a928

bougue-pe (approved):

Issue #11285, opened 2025-03-28
core: stdcm: use `StepTracker` in `STDCMHeuristic`


eckter:
Should have been part of https://github.com/OpenRailAssociation/osrd/issues/8974, requirement for overtakes. 

The heuristic relies on the number of steps on the path, which will become dynamic.

PR #11286, opened 2025-03-28
core: stdcm: fix off-by-one in heuristic step count

eckter:
The new `StepTracker.nStepsExcludingLookahead`, when compared to the old `edge.nPassedSteps`, has one difference: it (generally) includes the departure location. This resulted in the heuristic returning `0` (i.e. being disabled) between the last intermediate step and the destination.

This caused timeouts.

We should ideally integrate `StepTracker` in the heuristic and we'll need to do that before overtakes. It should have been part of https://github.com/OpenRailAssociation/osrd/issues/8974 but I kind of forgot about it. I'll open a different issue: https://github.com/OpenRailAssociation/osrd/issues/11285

### DIFF
--- a/core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/STDCMHeuristicTests.kt
+++ b/core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/STDCMHeuristicTests.kt
@@ -253,6 +253,10 @@ class STDCMHeuristicTests {
                 0.0,
                 null,
             )
-        return heuristic.invoke(defaultEdge, nodeOffsetOnEdge?.let { Offset(it) }, nbPassedSteps)
+        return heuristic.invoke(
+            defaultEdge,
+            nodeOffsetOnEdge?.let { Offset(it) },
+            nbPassedSteps + 1 // TODO: remove this +1 when we'll use StepTrackers directly
+        )
     }
 }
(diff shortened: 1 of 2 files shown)


core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/STDCMHeuristicTests.kt
```diff
@@ -253,6 +253,10 @@ class STDCMHeuristicTests {
                 0.0,
                 null,
             )
-        return heuristic.invoke(defaultEdge, nodeOffsetOnEdge?.let { Offset(it) }, nbPassedSteps)
+        return heuristic.invoke(
+            defaultEdge,
+            nodeOffsetOnEdge?.let { Offset(it) },
+            nbPassedSteps + 1 // TODO: remove this +1 when we'll use StepTrackers directly
```
bougue-pe, line 259: Will a test crash when using StepTrackers? (hoping that it will)

Also is it possible to add a test that would have crashed before this PR? Not sure, but can we assert != 0 before returning the value (excluding arrival)?

bougue-pe (approved):
Thanks for this!
(OK to merge even without more tests especially with the emergency of the bug)

Erashin (approved):

core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/STDCMHeuristicTests.kt
eckter: There was a failing test.

I can work on writing a new test that would fail with the bug (in a new PR)

eckter: Actually the current fix isn't working properly, so I'll work on better tests in this PR

eckter:
Closing: it's not working properly, it's easier to do https://github.com/OpenRailAssociation/osrd/issues/11285 directly than to debug these incompatible systems

PR #11289, opened 2025-03-28
core: stdcm: integrate step tracker into heuristic computation

### LINKED ISSUE #11285: core: stdcm: use `StepTracker` in `STDCMHeuristic`

### LINKED ISSUE #11290: core: stdcm: requests take way longer than they should, including timeouts
### What happened?

We have an off-by-one issue with the stdcm heuristic, causing it to be effectively "turned off" during the last section of the path. This results in lots of timeouts and very slow computations.

_No response_

Run any kind of request on a large infra. Also the logged progress tracker is broken. 

Recette (SNCF), Local, Dev (SNCF)

Firefox

6188f9b5267087ea5a55db7b7ffe0de9c37f90b3

eckter:
Follow-up of https://github.com/OpenRailAssociation/osrd/pull/11286

Fix https://github.com/OpenRailAssociation/osrd/issues/11285

Fix https://github.com/OpenRailAssociation/osrd/issues/11290

Can be reviewed by commit:

1. Small refactoring to have a proper class instead of a typealias on a function type. No functional change but lots of code moving, can interfere when reading the diff all commits at once
2. Use `StepTracker` instead of raw number of passed steps as function input
3. Adapt unit tests

Because the new tests use an `InfraExplorer` directly, the previous bug would have been catched there (it was an "integration" bug between infra explorer / step tracker / heuristic).

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/stdcm/STDCMHeuristic.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/stdcm/STDCMHeuristic.kt
@@ -20,12 +20,86 @@ import kotlin.math.min
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 
-/**
- * This typealias defines a function that can be used as a heuristic for an A* pathfinding. It takes
- * an edge, and offset on this edge, and a number of passed steps as input, and returns an
- * estimation of the remaining time needed to get to the end.
- */
-typealias STDCMAStarHeuristic = (STDCMEdge, Offset<Block>?, Int) -> Double
+data class STDCMAStarHeuristic(
+    val blockInfra: BlockInfra,
+    val steps: List<STDCMStep>,
+    val remainingTimeEstimations: List<MutableMap<BlockId, Double>>,
+    val mrspBuilder: CachedBlockMRSPBuilder,
+    val bestTravelTime: Double,
+) {
+    /**
+     * This function defines a function that can be used as a heuristic for an A* pathfinding. It
+     * takes an edge, and offset on this edge, and a number of passed steps as input, and returns an
+     * estimation of the remaining time needed to get to the end.
+     */
+    fun invoke(edge: STDCMEdge, offset: Offset<Block>?, nPassedSteps: Int): Double {
+        if (nPassedSteps >= steps.size - 1) return 0.0
+        val lookahead = edge.infraExplorer.getLookahead()
+        val currentBlock = edge.block
+        val allBlocks = mutableListOf(currentBlock)
+        allBlocks.addAll(lookahead)
+
+        // We don't consider the time of the last lookahead block to avoid issues
+        // if it contains the destination, and we don't need it (the destination is
+        // already locked-in).
+
+        // Account for the steps that will be passed in the lookahead
+        val expectedIndex =
+            getExpectedStepIndex(
+                allBlocks.dropLast(1),
+                offset ?: edge.blockOffsetFromEdge(edge.length),
+                nPassedSteps
+            )
+        if (expectedIndex >= remainingTimeEstimations.size) return 0.0
+
+        val timeAfterStartOfLastBlock =
+            remainingTimeEstimations[expectedIndex][allBlocks.last()]
+                ?: return Double.POSITIVE_INFINITY
+
+        // Compute the time it takes from the current point until the start
+        // of the last block of the lookahead, then from that point the destination.
+        var timeUntilStartOfLastBlock = 0.0
+        for (j in 0 until allBlocks.size - 1) {
+            timeUntilStartOfLastBlock += mrspBuilder.getBlockTime(allBlocks[j], null)
+        }
+        val timeSinceFirstBlock = mrspBuilder.getBlockTime(edge.block, offset)
+        timeUntilStartOfLastBlock -= timeSinceFirstBlock
+
+        val remainingTime = timeUntilStartOfLastBlock + timeAfterStartOfLastBlock
+
+        return remainingTime
+    }
+
+    /** Returns the numbers of passed waypoints at the end of the block list */
+    private fun getExpectedStepIndex(
+        blocks: List<BlockId>,
+        firstOffset: Offset<Block>,
+        currentIndex: Int
+    ): Int {
+        var stepIndex = currentIndex
+        var blockIndex = 0
+        var currentOffset = firstOffset
+        while (true) {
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/fr/sncf/osrd/stdcm/STDCMHeuristic.kt
```diff
@@ -20,12 +20,84 @@ import kotlin.math.min
...
+            )
+
+        val timeAfterStartOfLastBlock =
+            remainingTimeEstimations[expectedIndex][allBlocks.last()]
+                ?: return Double.POSITIVE_INFINITY
+
+        // Compute the time it takes from the current point until the start
+        // of the last block of the lookahead, then from that point the destination.
+        var timeUntilStartOfLastBlock = 0.0
+        for (j in 0 until allBlocks.size - 1) timeUntilStartOfLastBlock +=
```
Erashin: Any way to change the format here? Either EoL or brackets?

core/src/main/kotlin/fr/sncf/osrd/stdcm/STDCMHeuristic.kt
```diff
@@ -71,35 +69,23 @@ data class STDCMAStarHeuristic(
...
-                // Move on to the next block
-                blockIndex++
-                currentOffset = Offset(0.meters)
-            }
+        val stepTrackerCopy = stepTracker.clone()
+        for (block in blocks) {
+            stepTrackerCopy.moveForward(block, Offset.zero(), blockInfra.getBlockLength(block))
         }
+        val nPlannedSteps =
+            stepTrackerCopy.getSeenSteps().take(stepTrackerCopy.nStepsExcludingLookahead).count {
```
Erashin: Might be worth to add a `getReachedSteps()` in `StepTracker`.
-> `stepTrackerCopy.getReachedSteps().count { it.isPlanned }`

core/src/main/kotlin/fr/sncf/osrd/stdcm/STDCMHeuristic.kt
```diff
@@ -48,6 +48,7 @@ data class STDCMAStarHeuristic(
                 allBlocks.subList(0, allBlocks.size - 1),
                 stepTracker,
             )
+        if (expectedIndex >= remainingTimeEstimations.size) return 0.0
```
Erashin, line 51: Should probably be in one of the previous commits.

core/src/main/kotlin/fr/sncf/osrd/stdcm/STDCMHeuristic.kt
eckter: Fixed in https://github.com/OpenRailAssociation/osrd/pull/11289/commits/b525d5556e33cb0c556478e735010b499a285bd8

core/src/main/kotlin/fr/sncf/osrd/stdcm/STDCMHeuristic.kt
eckter: Fixed in https://github.com/OpenRailAssociation/osrd/pull/11289/commits/1edeb4e15c8bbd4a9256b3fc5dfe80ddc5000191

core/src/main/kotlin/fr/sncf/osrd/stdcm/STDCMHeuristic.kt
eckter: Fixed, but I couldn't do that with a fixup, I ammended the commits

Erashin (approved):

core/src/main/kotlin/fr/sncf/osrd/stdcm/STDCMHeuristic.kt
```diff
@@ -20,14 +20,71 @@ import kotlin.math.min
...
+        allBlocks.addAll(lookahead)
+
+        // We don't consider the time of the last lookahead block to avoid issues
+        // if it contains the destination, and we don't need it (the destination is
+        // already locked-in).
+
+        // Account for the steps that will be passed in the lookahead
+        val expectedIndex =
+            getExpectedStepIndex(
+                allBlocks.subList(0, allBlocks.size - 1),
```
bougue-pe: nit:
```suggestion
                allBlocks.removeLast()
```

core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/STDCMHeuristicTests.kt
```diff
@@ -227,11 +219,25 @@ class STDCMHeuristicTests {
...
+                appendOnlyLinkedListOf(),
+                SpacingRequirementAutomaton(
+                    infra,
+                    infra.fullInfra().loadedSignalInfra,
+                    infra,
+                    infra.fullInfra().signalingSimulator,
+                    IncrementalRequirementEnvelopeAdapter(REALISTIC_FAST_TRAIN, null, false),
+                    infraExplorer.getIncrementalPath(),
+                ),
+                REALISTIC_FAST_TRAIN,
```
bougue-pe: Not sure that it matters, but I'm curious about the change from `STANDARD_TRAIN` to `REALISTIC_FAST_TRAIN`.

bougue-pe (approved):

core/src/main/kotlin/fr/sncf/osrd/stdcm/STDCMHeuristic.kt
eckter: Fixed in https://github.com/OpenRailAssociation/osrd/commit/cefdaaf8289288725dfc6c096d5808ec96089ebf

Though I needed `.dropLast()` as I still need to return the value here and not just modify the list

core/src/test/kotlin/fr/sncf/osrd/stdcm/preprocessing/STDCMHeuristicTests.kt
eckter: I had just picked the first one suggested by my IDE. But I've changed it to `STANDARD_TRAIN` for consistency in https://github.com/OpenRailAssociation/osrd/commit/99d5f7c6b2a10e45314909df1da68beaf75d9108

PR #11292, opened 2025-03-28
 schemas, railjson-generator, tests: replace 'poetry' with 'uv' 🚀 

woshilapin:
Moving all the Python projects of `osrd` to `uv` which is faster, and handle the Python version and the virtual environment too.

It means:
- removing `poetry.lock` in favor of `uv.lock`
- updating all of the README.md
- adding some `.gitignore` for `*.egg-info` (to be honest, I'm not entirely sure I understand why and when these are produced)
- updating the `pyproject.toml` for local dependencies (there is not a correct way of defining these in the standard, so we define them as normal dependencies, and we rely on `uv` specific source specification to provide these dependencies)
- update the CI scripts (using `astral-sh/setup-uv` action)


eckter (approved):
Very nice!

There's still some mentions of poetry in a few places (readmes and justfiles). You can `git grep poetry` to find them.

Other than that, LGTM

woshilapin:
@eckter Thanks for catching some of those. Some of them happened in between the time I opened this branch, but some, I missed completely (like `dependabot` and `taplo`'s configuration).

flomonster (approved):
LGTM

PR #11301, opened 2025-03-31
front: fix panning over a train path.

### LINKED ISSUE #10725: When panning on the space time, the mouse grabs a train and starts moving it
click and start dragging the chart. if the mouse touches a train it start moving the train instead of panning.

https://github.com/user-attachments/assets/9d982c66-b348-4379-a9e5-cb7750a6eb27

### LINKED ISSUE #11275: GET: Train moves instead of graph when clicking in blank space
### What happened?

https://github.com/user-attachments/assets/0dee0656-f6f8-4b8e-b298-4fc5dd7c883d

If you click briefly on a train, then if you try to move in th grap by clicking in blanck space, it moves the train instead.

It's like the first click selected the train for moving.

The train has to move only if clicked. In this situation, the graph has to move.

1. Open a scenario with a train
2. Click briefly on a train
3. Click on a white space of the GET, maintain and drag your mouse

Recette (SNCF)

Firefox

872260c

anisometropie:
closes #10725
closes #11275

### DIFF
--- a/front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
+++ b/front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
@@ -102,6 +102,8 @@ const ManchetteWithSpaceTimeChartWrapper = ({
 
   const [tmpSelectedTrain, setTmpSelectedTrain] = useState(selectedTrainScheduleId);
 
+  const [previousPanning, setPreviousPanning] = useState(false);
+
   useEffect(() => {
     setTmpSelectedTrain(selectedTrainScheduleId);
   }, [selectedTrainScheduleId]);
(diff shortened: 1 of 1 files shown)


theocrsb (approved):
LGTM and tested

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -301,7 +303,12 @@ const ManchetteWithSpaceTimeChartWrapper = ({
     }
 
     // if not dragging, we check if we should start dragging
+    // Only a mouse hover that starts already over a path should register
+    // if we start panning, and then the mouse hover over the path,
```
SharglutDev:
```suggestion
    // if we start panning, and then the mouse hovers over the path,
```

front/src/modules/simulationResult/components/ManchetteWithSpaceTimeChart/ManchetteWithSpaceTimeChart.tsx
```diff
@@ -102,6 +102,8 @@ const ManchetteWithSpaceTimeChartWrapper = ({
 
   const [tmpSelectedTrain, setTmpSelectedTrain] = useState(selectedTrainScheduleId);
 
+  const [previousPanning, setPreviousPanning] = useState(false);
```
SharglutDev, line 105: Could we use a `useRef` instead to avoid rerender ?
```suggestion
  const previousPanningRef = useRef(false);
```

SharglutDev (approved):
Lgtm and tested, thanks ! You can resolve the comment when fixed

PR #11302, opened 2025-03-31
core: tools: reproduce requests using railjson infra file

eckter:
Small quality of life improvement: this lets us reproduce request payloads without running any other service, infras can be fetched from railjson files using `--railjson infra.json` parameter

### DIFF
--- a/core/src/main/java/fr/sncf/osrd/cli/ReproduceRequest.kt
+++ b/core/src/main/java/fr/sncf/osrd/cli/ReproduceRequest.kt
@@ -4,13 +4,18 @@ import com.beust.jcommander.Parameter
 import com.beust.jcommander.Parameters
 import com.squareup.moshi.JsonAdapter
 import fr.sncf.osrd.api.ElectricalProfileSetManager
+import fr.sncf.osrd.api.FullInfra
 import fr.sncf.osrd.api.InfraManager
+import fr.sncf.osrd.api.InfraProvider
+import fr.sncf.osrd.api.makeSignalingSimulator
 import fr.sncf.osrd.api.pathfinding.PathfindingBlocksEndpoint
 import fr.sncf.osrd.api.pathfinding.pathfindingRequestAdapter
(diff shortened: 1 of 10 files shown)


core/src/main/java/fr/sncf/osrd/cli/ReproduceRequest.kt
```diff
@@ -93,3 +106,22 @@ class ReproduceRequest : CliCommand {
...
+/**
+ * Implement the InfraProvider interface using a railjson file. Used to reproduce requests without
+ * needing to run any other part of the stack.
+ */
+data class FileInfraProvider(val path: String) : InfraProvider {
+    override fun getInfra(
+        infraId: String?,
+        expectedVersion: String?,
+        diagnosticRecorder: DiagnosticRecorder?
+    ): FullInfra? {
```
Erashin: Nothing can return null here, and we've got a method for that :)
```suggestion
    ): FullInfra {
        val rjs = RJSParser.parseRailJSONFromFile(path)
        return FullInfra.fromRJSInfra(rjs, makeSignalingSimulator())
```

Erashin (approved):
Quite an elegant solution, I like it!

core/src/main/java/fr/sncf/osrd/cli/ReproduceRequest.kt
eckter: Fixed

Erashin: I think you didn't see the second part of the fix with the return ;)

eckter: Second part of the fix? I'm not sure I get it

Oh wait I just realized there's a `FullInfra.fromRJSInfra()`, I don't know if that's what you had in mind but I will use this. (done in last commit)

Erashin (approved):

bougue-pe (approved):
❤️ Thanks!

PR #11317, opened 2025-03-31
front, ci: auto sort translation keys

Synar:
Sort all existing translation keys, add a translation key sorting script for ease of use, and use it to enforce translation keys sorting in the ci.

This enforcement was planned [here]([https://github.com/osrd-project/osrd-confidential/issues/623)


Khoyo (approved):
LGTM

The script should probably be run by the justfile (eg. lint), which pr does this depends on whether #11174 gets merged first

front/docker/Dockerfile
```diff
@@ -1,4 +1,7 @@
 FROM front_build AS tests
 
+# Add jq
+RUN apk add --no-cache jq
```
Khoyo: the running env for the front tests is debian bookworm based

RomainValls (approved):
LGTM 👍(except for the insensitive, it seems we're going with the sensitive sorting?)

front/docker/Dockerfile
Synar: Fixed ^^

front/docker/Dockerfile
```diff
@@ -1,4 +1,9 @@
 FROM front_build AS tests
 
+# Add jq
+RUN apt-get update -yqq && \
+    apt-get install -yqq --no-install-recommends jq && \
+    rm -rf /var/lib/apt/lists/*
+
```
SharglutDev: Should we explain why we need to install `jq` ?

SharglutDev (approved):
Lgtm

> [!NOTE]
> Should be merged only on April 11th

flomonster (changes requested):
We will use weblate soon. I'm woried that this PR won't be compatible with it.
I don't think that weblate will sort keys (check [doc](https://docs.weblate.org/en/latest/formats/i18next.html)). It will be painful to reorder keys before merging a PR opened by the bot.

@emersion @woshilapin @sim51 what are your thoughts about it (since you know a little bit about weblate).

Synar:
@flomonster have you checked https://docs.weblate.org/en/weblate-5.9.1/admin/addons.html#customize-json-output?
I am not super familiar with weblate, but, assuming we can manage weblate add-ons, it seems to me that it should work

(Especially since the section on i18n you linked has a direct link to the customize-json-output section, which has a sort-keys flag.

For the case sensitive vs insensitive question, I suppose we can just do the same thing weblate does then. (Or if casing really really matters we could always open a weblate pr to add a flag like was done for other outputs I suppose, but that's probably not needed ^^))

flomonster (approved):
Sorry I did not dig enough the doc 😅

flomonster:
We might still adapt the way keys are sorted to match with weblate. But it's an issue for later 😉

front/docker/Dockerfile
Synar: I think so ^^. But I realized I could just run the script outside of the container even in the CI (and jq is already available in the CI env by default) so we actually don't need this at all.

woshilapin:
> emersion woshilapin sim51 what are your thoughts about it (since you know a little bit about weblate).

I believe this is small enough and useful enough improvement to not wait for a Weblate that might come later in the future. In the meantime, it's probably useful for those who work on translation. And we'll deal with the adaptation of that once Weblate will be there.

woshilapin (approved):

Synar:
I'm starting to think I should wait for the night to try and merge this, with how many conflicts keep popping up ^^''

PR #11321, opened 2025-04-01
front: fix scroll down when button is clicked

### LINKED ISSUE #11299: front: stdcm window scrolls down when clicking "as soon as possible"
### What happened?

When clicking "dès que possible" (= "as soon as possible"), the window scrolls down, whereas "horaire précis" does not have the same behavior

https://github.com/user-attachments/assets/3bc393b0-ad34-4f73-8423-a770aa9b225f

don't scroll down (keep same view) when changing the input of the simulation

1. follow video

Production (SNCF)

Firefox

82cd094

theocrsb:
closes [#11299](https://github.com/OpenRailAssociation/osrd/issues/11299)


Uriel-Sautron (approved):
LGTM, tested

kmer2016 (approved):

PR #11323, opened 2025-04-01
front: prevent creation of trains with names that makes train unusable

### LINKED ISSUE #10874: Impossible to create a train with a long name (more than 128 characters)
### What happened?

Impossible to create a train with a long name (more than 128 characters)

Image

I don't think it should be possible to create a train with a name that prevents it from being used.

Go to scenario page
Try to add train with a train whose name is too long
See error

Local

Firefox

development

achrafmohye:
Closes #10874

### DIFF
--- a/front/src/modules/rollingStock/components/RollingStockEditor/RollingStockEditorFormHelpers.tsx
+++ b/front/src/modules/rollingStock/components/RollingStockEditor/RollingStockEditorFormHelpers.tsx
@@ -23,6 +23,7 @@ import type {
   RollingStockParametersValues,
   SchemaProperty,
 } from 'modules/rollingStock/types';
+import { SMALL_INPUT_MAX_LENGTH } from 'utils/strings';
 
 type RollingStockMetadataFormProps = {
   rollingStockValues: RollingStockParametersValues;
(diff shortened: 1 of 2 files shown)


theocrsb:
I can still create a train with more than 120 characters :/

achrafmohye:
@theocrsb it appears that the bug can be reproduced in two ways one in the rolling stock editor, the second one is by setting a long **Number, name** in the scenario page without actually having an extra long rolling stock name, now iit fixes both

front/src/utils/strings.ts
```diff
@@ -88,6 +88,7 @@ export const TEXT_AREA_MAX_LENGTH = 4096;
 export const SMALL_TEXT_AREA_MAX_LENGTH = 1024;
 export const TEXT_INPUT_MAX_LENGTH = 255;
 export const SMALL_INPUT_MAX_LENGTH = 128;
+export const TRAIN_NAME_MAX_LENGTH = 120;
```
emersion: Asked a question here about the value 120: https://github.com/OpenRailAssociation/osrd/issues/10874#issuecomment-2769765754

theocrsb (approved):
LGTM and tested

front/src/utils/strings.ts
achrafmohye: @emersion used the `SMALL_INPUT_MAX_LENGTH ` like discussed what do you think ?

emersion (approved):
LGTM!

front/src/utils/strings.ts
achrafmohye: can you resolve this comment, or is the comment is always valid ?

PR #11328, opened 2025-04-02
front: avoid blinking with example DTO when launching NGE

emersion:
NGE will use an example DTO (with Swiss train stations) if it's initialized without a specific DTO. Later, we set the DTO to the final state. There is a small time window where NGE displays the example data before the real data. This results in blinking.

Set NGE's initial DTO as an empty state to avoid this blinking.

Requires this NGE patch:
https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/pull/419

### DIFF
--- a/front/src/applications/operationalStudies/components/NGE/NGE.tsx
+++ b/front/src/applications/operationalStudies/components/NGE/NGE.tsx
@@ -10,6 +10,7 @@ import ngeVendor from '@osrd-project/netzgrafik-frontend/dist/netzgrafik-fronten
 
 import i18n from 'i18n';
 
+import { EMPTY_DTO } from './consts';
 import type { NetzgrafikDto, NGEEvent } from './types';
 
 interface NGEElement extends HTMLElement {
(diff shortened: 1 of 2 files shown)


front/package.json
```diff
@@ -8,7 +8,7 @@
     "@nivo/core": "^0.88.0",
     "@nivo/line": "^0.88.0",
     "@openapi-contrib/openapi-schema-to-json-schema": "^5.1.0",
-    "@osrd-project/netzgrafik-frontend": "0.0.0-snapshot.90c0b8c5111ab498e0963758f6c306fe4f38d729",
+    "@osrd-project/netzgrafik-frontend": "0.0.0-snapshot.91b559fbb68fff395f543b8343a923b1e0d13eaa",
```
louisgreiner: version with fixed fork:

```suggestion
    "@osrd-project/netzgrafik-frontend": "0.0.0-snapshot.b1f76019cf13070d3d1764b1b643e98c711c1c07",
```

louisgreiner (approved):
Tested with suggested version, nice :-)

front/package.json
emersion: The version bump has been performed in deda5858d1ead1c4f4bcb08db841d83eac5775a0.

kmer2016 (approved):
LGTM!

PR #11331, opened 2025-04-02
front: fix Duration serialization in redux persist

### LINKED ISSUE #11277: Diffrent errors when enabling/disabling paced trains
### What happened?

It's hard to reproduce, but try to follow the video and enable/disable paced trains
https://github.com/user-attachments/assets/1e3658e3-9d7a-4210-bc23-8e8aaeb205b3

Can add unique train or paced train without any issue 

1. Go to the Scenario page

2. Activate paced train

3. Add a paced train

4. Deactivate paced train

5. Add a unique train

6. Activate paced train (duplicate of #2, kept for consistency)

7. Refresh the page

8. Add a unique train / paced train

Recette (SNCF)

Firefox, Google Chrome

872260c

kmer2016:
closes #11277
The bug occurred because of two conflicting persistence configurations:
- One at root level in persistConfig
- One at slice level with persistReducer

This caused Duration objects to not be properly serialized when stored in localStorage, leading to runtime errors when trying to call Duration methds on non-Duration objects.

Solution:
- Remove slice level persistence
- Keep only root level persistence with transforms

### DIFF
--- a/front/src/reducers/index.ts
+++ b/front/src/reducers/index.ts
@@ -1,9 +1,8 @@
-import type { Action, Reducer, ReducersMapObject, AnyAction } from 'redux';
-import type { PersistConfig } from 'redux-persist';
-import { createTransform, persistCombineReducers, persistReducer } from 'redux-persist';
+import type { Action, ReducersMapObject } from 'redux';
+import { createTransform, persistCombineReducers } from 'redux-persist';
 import storage from 'redux-persist/lib/storage'; // defaults to localStorage
 import createCompressor from 'redux-persist-transform-compress';
-import { createFilter } from 'redux-persist-transform-filter';
+import { createFilter, createBlacklistFilter } from 'redux-persist-transform-filter';
 
 import { osrdEditoastApi } from 'common/api/osrdEditoastApi';
 import { osrdGatewayApi } from 'common/api/osrdGatewayApi';
@@ -53,11 +52,18 @@ const userWhiteList = ['account', 'userPreferences', 'impersonatedUser'];
 
 const mainWhiteList = ['lastInterfaceVersion'];
 
-const saveMapFilter = createFilter('map', mapWhiteList);
+const operationalStudiesConfBlackList = ['usingSpeedLimits'];
 
-const saveUserFilter = createFilter('user', userWhiteList);
+const saveMapFilter = createFilter(mapSlice.name, mapWhiteList);
 
-const saveMainFilter = createFilter('main', mainWhiteList);
+const saveUserFilter = createFilter(userSlice.name, userWhiteList);
+
+const saveMainFilter = createFilter(mainSlice.name, mainWhiteList);
+
+const operationalStudiesFilter = createBlacklistFilter(
+  operationalStudiesConfSlice.name,
+  operationalStudiesConfBlackList
+);
 
 // Deserialize date strings coming from local storage
 const operationalStudiesDateTransform = createTransform(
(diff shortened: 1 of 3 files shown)


Uriel-Sautron (approved):
Thank you for this fix, tested.

Yohh (approved):

front/src/reducers/index.ts
```diff
@@ -77,7 +76,10 @@ const operationalStudiesDateTransform = createTransform(
     cadence: Duration.parse(cadence),
     timeRangeDuration: Duration.parse(timeRangeDuration),
   }),
-  { whitelist: ['operationalStudiesConf'] }
+  {
+    whitelist: ['operationalStudiesConf'],
+    blacklist: ['usingSpeedLimits'],
```
SharglutDev: Feels weird to have this in `blacklist` in a transform concerning dates

emersion (commented):
It seems like `pathStepsTransform` is no longer called.

To reproduce: click "Add" in a scenario, go to the "times and stops" tab, this error pops up:

```
11:24:25.806 Uncaught (in promise) TypeError: stopFor.total is not a function
    formatSuggestedViasToRowVias utils.ts:81
    formatSuggestedViasToRowVias utils.ts:74
    fetchAndFormatRows TimesStopsInput.tsx:153
    TimesStopsInput TimesStopsInput.tsx:163
    React 15
    handleTabClick Tabs.tsx:35
    onClick Tabs.tsx:51
    React 23
    <anonymous> index.tsx:28
utils.ts:81:45
```

SharglutDev (commented):
Wow forgot to send the comment sorry

front/src/reducers/index.ts
kmer2016: You're right! I've moved the blacklist to its dedicated filter and kept the transform focused on date/duration conversions only.

SharglutDev:
Looks good to me. But I'll leave a better redux specialist than me take a turn on this PR too :)

front/src/reducers/index.ts
```diff
@@ -77,7 +76,10 @@ const operationalStudiesDateTransform = createTransform(
     interval: Duration.parse(interval),
     timeWindow: Duration.parse(timeWindow),
   }),
-  { whitelist: ['operationalStudiesConf'] }
+  {
+    whitelist: ['operationalStudiesConf'],
+    blacklist: ['usingSpeedLimits'],
```
emersion: I'm not sure it ever makes sense to use `whitelist` together with `blacklist`? If `whitelist` is supplied, the filter will never run on anything else?

front/src/reducers/index.ts
```diff
@@ -77,7 +76,10 @@ const operationalStudiesDateTransform = createTransform(
     interval: Duration.parse(interval),
     timeWindow: Duration.parse(timeWindow),
   }),
-  { whitelist: ['operationalStudiesConf'] }
+  {
+    whitelist: ['operationalStudiesConf'],
```
emersion: Could we use a `.name` instead of a hardcoded string here?

kmer2016: Yes, already done.

emersion: Ah right, one commit uses strings but the following one replaces these with constants.

emersion (approved):
LGTM, thank you for untangling this!

anisometropie (approved):
looking good

Yohh (approved):

PR #11332, opened 2025-04-02
railjson_generator: import infra scripts as modules

woshilapin:
This will ease the conversion to `uv` which doesn't install the current package `railjson_generator` in the virtual environment (will be useful for https://github.com/OpenRailAssociation/osrd/pull/11292).

I've removed the side effect of `infra-scripts/*.py` (i.e. writing to files) and instead, make these scripts provide the same interface: exporting an `Infra` object named `infra` (same for `external_inputs`). `railjson_generator` takes care of deciding where and when to write.


python/railjson_generator/railjson_generator/__main__.py
```diff
@@ -1,17 +1,39 @@
...
 def run_script(gen_script: Path, output_dir: Path):
     assert gen_script.is_file(), f"{gen_script} is not a valid file name"
     script_name = gen_script.stem
     script_output = output_dir / script_name
     script_output.mkdir(parents=True, exist_ok=True)
     print("running generation script", script_name)
-    subprocess.run([sys.executable, gen_script, script_output], check=True)
+    module = import_module_from_path(script_name, gen_script)
+    module.infra.save(script_output / "infra.json")
+    if hasattr(module, "external_inputs"):
```
shenriotpro: i'm not sure why you're doing this here, but there may be a simpler way, for instance using module.scenario_data

shenriotpro:
Thanks, can we turn these scripts into functions and add `if __name__=='__main__'`s for backwards compatibility?

python/railjson_generator/railjson_generator/__main__.py
woshilapin: I'm not entirely sure of what you mean. Do you mean something like [this](https://github.com/OpenRailAssociation/osrd/compare/706433803a0c662940200c4e0b71aaac9c34079d..32564931252274dedb6f0f35d6cf5c5eaefe03c5)?

shenriotpro: I mean let's avoid hasattr if we can :)

woshilapin:
> Thanks, can we turn these scripts into functions and add `if __name__=='__main__'`s for backwards compatibility?

@shenriotpro I've done it [this way](https://github.com/OpenRailAssociation/osrd/compare/32564931252274dedb6f0f35d6cf5c5eaefe03c5..4c7b0feb4b7f03d57af54a55894666640daa3b53), does it work for you?

python/railjson_generator/railjson_generator/__main__.py
woshilapin: Then I succeeded, I believe 😄

shenriotpro:
> > Thanks, can we turn these scripts into functions and add `if __name__=='__main__'`s for backwards compatibility?
> 
> @shenriotpro I've done it [this way](https://github.com/OpenRailAssociation/osrd/compare/32564931252274dedb6f0f35d6cf5c5eaefe03c5..4c7b0feb4b7f03d57af54a55894666640daa3b53), does it work for you?

Thanks, the `__main__` part is great, I think it'd be nice to have the rest in a function as well, but that's a nit

tests/infra-scripts/circle_infra.py
```diff
@@ -1,27 +1,33 @@
...
-switch2 = builder.add_point_switch(track_d.begin(), track_b.end(), track_c.end(), label="switch2")
+    # Build infra
+    return ScenarioData(infra=builder.build())
 
-# Build infra
-infra = builder.build()
 
-# Save railjson
-infra.save(OUTPUT_DIR / "infra.json")
+scenario_data = _build_scenario_data()
```
shenriotpro, line 33: nit: i'd remove this and call the (public) function from the other module
the idea is that one can import this without running all the code

shenriotpro (approved):
Thanks

tests/infra-scripts/circle_infra.py
woshilapin: Yes, but apart from doing that, this file does nothing. And that would mean calling it twice (one from `railjson_generator`, and another one from the `__main__` block below). I see the lazy attitude being a good thing... but in this case, since the file does only one thing and it's exactly this, I don't understand the goal of not calling the function here.

flomonster (approved):
Looks good.

> [!NOTE]
> We should think about dropping the railjson_generator library.
> It's not used anymore by anyone.

tests/infra-scripts/circle_infra.py
woshilapin: Since it's "nit", I'm resolving, but feel free to reopen if it's more than an itch for you.

PR #11336, opened 2025-04-02
front: add tailwind prettier plugin

ghost:
Add tailwind auto prettier sorting classes
https://tailwindcss.com/blog/automatic-class-sorting-with-prettier


SharglutDev:
Hmm since we don't have tailwind installed on osrd, I don't see why we would need this plugin here ?

ghost:
> Hmm since we don't have tailwind installed on osrd, I don't see why we would need this plugin here ?

This project won't be using ts anytime soon... so no need

Issue #11249, opened 2025-03-25
editoast: rename TrainScheduleBase to TrainSchedule


Wadjetz:
### Description and goal

- [x] Rename schema TrainScheduleBase to TrainSchedule
- [ ] Rename PacedTrain.train_schedule_base to PacedTrain.train_schedule ????
- [x] Rename TrainScheduleResult to TrainScheduleResponse

Fallow up from https://github.com/OpenRailAssociation/osrd/pull/10945#pullrequestreview-2714084430

SharglutDev:
And `TrainScheduleResult` to `TrainScheduleResponse` :)

PR #11339, opened 2025-04-03
editoast: rename TrainScheduleResult to TrainScheduleResponse

Wadjetz:
part of https://github.com/OpenRailAssociation/osrd/issues/11249

Rename TrainScheduleResult to TrainScheduleResponse

This is only a part of the ticket to facilitate the review. I will submit another PR for renaming TrainScheduleBase.

### DIFF
--- a/front/src/common/api/osrdEditoastApi.ts
+++ b/front/src/common/api/osrdEditoastApi.ts
@@ -5,7 +5,7 @@ import {
   type GetSpritesSignalingSystemsApiResponse,
   generatedEditoastApi,
   type Property,
-  type TrainScheduleResult,
+  type TrainScheduleResponse,
   type PacedTrainResponse,
 } from './generatedEditoastApi';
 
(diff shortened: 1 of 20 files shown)


front/src/common/api/osrdEditoastApi.ts
```diff
@@ -16,14 +16,14 @@ const osrdEditoastApi = generatedEditoastApi
...
-        TrainScheduleResult[],
+        TrainScheduleResponse[],
         { timetableId: number }
       >({
         queryFn: async ({ timetableId }, { dispatch }) => {
           const pageSize = 25;
           let page = 1;
           let reachEnd = false;
-          const result: TrainScheduleResult[] = [];
+          const responses: TrainScheduleResponse[] = [];
```
SharglutDev: nit : we kept `result` for the next custom endpoint with `PacedTrainResponse`.
For harmonization, we should either keep `result` or use `response` (without `s`) for both

SharglutDev (approved):
Lgtm, thanks

woshilapin (approved):
Thank you.

leovalais (approved):
Thanks! :)

Maymanaf (approved):

PR #11346, opened 2025-04-03
front: fix scenario train count not taking into account paced trains

### LINKED ISSUE #11345: Train count on scenario mini cards doesn't take into account paced trains
### What happened?

Image

The number shoud take into account paced trains

Add train schedule and paced trains in a scenario and check the scenario mini card in the study

Local

Firefox

.

SharglutDev:
fix #11345


theocrsb (approved):
LGTM and tested

Caracol3 (approved):
LGTM (tested)

kmer2016 (approved):

PR #11349, opened 2025-04-04
editoast: rename TrainScheduleBase to TrainSchedule

### LINKED ISSUE #11249: editoast: rename TrainScheduleBase to TrainSchedule

Wadjetz:
fix https://github.com/OpenRailAssociation/osrd/issues/11249


editoast/src/models/paced_train.rs
```diff
@@ -10,7 +10,7 @@ use editoast_schemas::train_schedule::Margins;
 use editoast_schemas::train_schedule::PathItem;
 use editoast_schemas::train_schedule::PowerRestrictionItem;
 use editoast_schemas::train_schedule::ScheduleItem;
-use editoast_schemas::train_schedule::TrainScheduleBase;
+use editoast_schemas::train_schedule::TrainSchedule as TrainScheduleSchema;
```
woshilapin: nitpick/suggestion: Maybe we could do the same as with `model`, re-export all objects in the root level of `editoast_schemas`, so then you can use them as `editoast_schemas::TrainSchedule` 🤷 In particular, in files where both `models` and `schema` are used, they could be referred throughout the file as `editoast_schemas::TrainSchedule` and `models::TrainSchedule` (when there is a name that is used for 2 different things, I tend to prefer to explicitly prefix it inline everywhere, instead of `use`-ing it for one, and prefixing for the other).

Note that you could still `use editoast_schemas as schemas;` to shorten the prefix to only what we really need to disambiguate.

I'd be happy to hear any opinion about that suggestion?

woshilapin (approved):

Maymanaf (approved):

flomonster (approved):

Math-R (approved):
Lgtm :)

editoast/editoast_schemas/src/train_schedule/train_schedule_base.rs
leovalais, line 1: Shouldn't this module be renamed as well? Or better put the `TrainSchedule` in `trains_schedule.rs` directly?

editoast/src/models/paced_train.rs
leovalais: (maybe we can even drop the `editoast_` prefix for crates? 👀 It brings very little benefits imo)

leovalais (approved):
Thanks!

editoast/src/models/paced_train.rs
woshilapin: I agree... probably in another PR though 😄

woshilapin (approved):

editoast/src/models/paced_train.rs
flomonster: We did that to match what other repos are doing like https://github.com/serde-rs/serde.

leovalais: Yeah I recall talking about that, but in the end since unlike `serde` or `diesel` we're not publishing our crates, it makes less sense to follow this practice imo (and it's cumbersome to write ~above all else~)

PR #11353, opened 2025-04-07
front: synchronize scenario page across windows

emersion:
Use a BroadcastChannel to synchronize timetable item inserts, updates and deletes across browser windows.

Ref https://github.com/osrd-project/osrd-confidential/issues/970

### DIFF
--- a/front/src/applications/operationalStudies/hooks/useScenarioData.ts
+++ b/front/src/applications/operationalStudies/hooks/useScenarioData.ts
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState, useRef } from 'react';
 
 import { keyBy, sortBy } from 'lodash';
 import { useSelector } from 'react-redux';
@@ -22,6 +22,7 @@ import type {
 } from 'reducers/osrdconf/types';
 import { getTrainIdUsedForProjection } from 'reducers/simulationResults/selectors';
 import { getShowPacedTrains } from 'reducers/user/userSelectors';
+import { useAppDispatch } from 'store';
 import {
   formatEditoastTrainIdToPacedTrainId,
   formatEditoastTrainIdToTrainScheduleId,
@@ -34,7 +35,13 @@ import { mapBy } from 'utils/types';
 import useAutoUpdateProjection from './useAutoUpdateProjection';
 import usePathProjection from './usePathProjection';
 
+type ScenarioBroadcastMessage =
+  | { type: 'upsertTimetableItems'; timetableItems: TimetableItemWithTimetableId[] }
+  | { type: 'removeTimetableItems'; timetableItemIds: TimetableItemId[] }
+  | { type: 'updateTrainDepartureTime'; timetableItemId: TimetableItemId; newDeparture: Date };
+
 const useScenarioData = (scenario: ScenarioResponse, infra: InfraWithState) => {
+  const dispatch = useAppDispatch();
   const electricalProfileSetId = useSelector(getOperationalStudiesElectricalProfileSetId);
   const trainIdUsedForProjection = useSelector(getTrainIdUsedForProjection);
   const showPacedTrains = useSelector(getShowPacedTrains);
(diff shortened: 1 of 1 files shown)


theocrsb (approved):
LGTM and tested

SharglutDev (commented):
Thank you for the PR ! I have a weird behavior : when working with 2 windows, if I delete a train from a scenario with multiple trains from one window, it's properly removed on the 2nd window but if I leave the scenario and go back into it, the scenario is empty.

https://github.com/user-attachments/assets/c4e524ce-95b4-4ccb-a61d-f2efdc54c4e9



Same behavior when adding a train, it's properly added but when leaveing/entering the scenario, it's not there anymore except if I refresh (sorry for the video, had a weird bug when I couldn't click on any button at one point, seems related to the recording tool)

https://github.com/user-attachments/assets/e51b4af2-7510-40c9-aa8a-5c7b9d8537ff

emersion:
> I have a weird behavior : when working with 2 windows, if I delete a train from a scenario with multiple trains from one window, it's properly removed on the 2nd window but if I leave the scenario and go back into it, the scenario is empty.

Good catch! Fixed by adding a call to `osrdEditoastApi.util.invalidateTags()`.

SharglutDev:
Just to be sure, synchronization should only work for the timetable ? We don't want to synchronize also when changing the scenario details ? (name, description)

SharglutDev:
Another weird behavior : the space time chart (and its loader) is blinking and there are some warning for simmulation summary and project path in console

https://github.com/user-attachments/assets/287816b7-d629-4704-bc4c-bfbebbe70652

emersion:
> Just to be sure, synchronization should only work for the timetable ? We don't want to synchronize also when changing the scenario details ? (name, description)

This PR is only the first part of https://github.com/osrd-project/osrd-confidential/issues/970. The second part can be done in a second PR.

After some investigation, part 2 (rtk-query integration) appears to be significantly more complicated. I'm not sure it's worth the effort. I'll talk to a PO and maybe open an issue on the react-toolkit repo.

> the space time chart (and its loader) is blinking

Oops, good catch. It seems like with my previous changes we entered an infinite loop, with a ping-pong game of messages broadcast back and forth between the two windows. Should be fixed now!

> there are some warning for simmulation summary and project path in console

Seems to be related to the lazy loading. I haven't found a way to get rid of these yet. I _think_ they should go away with my lazy loading redesign, will check.

emersion:
> part 2 (rtk-query integration) appears to be significantly more complicated

To expand on this a bit: we'd probably need to define a second `BroadcastChannel` inside `osrdEditoastApi`, set `onQueryStarted` in `baseQuery`, wait for mutations to complete, and broadcast a message to invalidate mutation tags in that case. Alternatively, could perhaps be implemented as a middleware (ref https://github.com/reduxjs/redux-toolkit/blob/3592b084c1921964137598b3de8deadf39038685/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts#L48).

front/src/applications/operationalStudies/hooks/useScenarioData.ts
```diff
@@ -322,6 +335,78 @@ const useScenarioData = (scenario: ScenarioResponse, infra: InfraWithState) => {
     [timetableItems, rollingStocks]
   );
 
+  const upsertTimetableItemsWithBroadcast = useCallback(
+    (timetableItemsToUpsert: TimetableItemWithTimetableId[]) => {
```
SharglutDev, line 339: nit : to be consistent with the other two functions (or the other way, be more preciser for the other 2 functions)
```suggestion
    (ids: TimetableItemWithTimetableId[]) => {
```

SharglutDev (approved):
Lgtm and tested, very nice job !

front/src/applications/operationalStudies/hooks/useScenarioData.ts
emersion: The other two functions take IDs as inputs, while this one takes a full timetable item (ie. an object instead of a string).

SharglutDev: woops, got baited by the id at the end of the type (TGIF)

PR #11355, opened 2025-04-07
editoast: refacto conflict endpoint

younesschrifi:
Some comments have not been addressed in this [PR](https://github.com/OpenRailAssociation/osrd/pull/11069): .
This PR is used to correct any requests that may have been made concerning the endpoint of conflicts that have not been addressed or any improvments about the endpoint.

### DIFF
--- a/editoast/src/models/paced_train.rs
+++ b/editoast/src/models/paced_train.rs
@@ -71,6 +71,19 @@ impl PacedTrain {
             options: self.options,
         }
     }
+
+    pub fn iter_occurrences(&self) -> impl Iterator<Item = TrainSchedule> {
+        let base_occurrence = self.clone().into_first_occurrence();
+
+        (0..self.num_occurrences()).map(move |occurrence_idx| TrainSchedule {
+            start_time: base_occurrence.start_time + self.interval * occurrence_idx as i32,
+            ..base_occurrence.clone()
+        })
+    }
+
+    pub fn num_occurrences(&self) -> usize {
+        (self.time_window.num_seconds() / self.interval.num_seconds()) as usize
+    }
 }
 
 impl From<paced_train::PacedTrain> for PacedTrainChangeset {
--- a/editoast/src/views/timetable.rs
+++ b/editoast/src/views/timetable.rs
@@ -1,6 +1,9 @@
 pub mod stdcm;
 
 use std::collections::HashMap;
+use std::fmt::Display;
+use std::str::FromStr;
+use std::sync::Arc;
 
 use axum::Extension;
 use axum::extract::Json;
@@ -14,9 +17,11 @@ use chrono::Utc;
 use derivative::Derivative;
 use editoast_authz::Role;
 use editoast_derive::EditoastError;
+use editoast_models::DbConnection;
 use editoast_models::DbConnectionPoolV2;
 use editoast_schemas::paced_train::PacedTrain;
 use editoast_schemas::train_schedule::TrainSchedule;
+use itertools::Either;
 use itertools::Itertools;
 use serde::Deserialize;
 use serde::Serialize;
@@ -30,14 +35,12 @@ use super::pagination::PaginationQueryParams;
 use super::pagination::PaginationStats;
 use super::path::pathfinding::PathfindingResult;
 use crate::AppState;
-use crate::RetrieveBatch;
+use crate::ValkeyClient;
 use crate::core::AsCoreRequest;
 use crate::core::conflict_detection::Conflict as CoreConflict;
 use crate::core::conflict_detection::ConflictDetectionRequest;
 use crate::core::conflict_detection::ConflictRequirement;
 use crate::core::conflict_detection::ConflictType;
-use crate::core::conflict_detection::PacedTrainOccurrenceId;
-use crate::core::conflict_detection::TrainId;
 use crate::core::conflict_detection::TrainRequirements;
 use crate::core::simulation::SimulationResponse;
 use crate::error::Result;
@@ -50,6 +53,7 @@ use crate::models::timetable::TimetableWithTrains;
 use crate::models::train_schedule::TrainScheduleChangeset;
 use crate::views::AuthenticationExt;
 use crate::views::AuthorizationError;
+use crate::views::CoreClient;
 use crate::views::train_schedule::TrainScheduleForm;
 use crate::views::train_schedule::TrainScheduleResponse;
 use crate::views::train_schedule::train_simulation_batch;
@@ -384,7 +388,7 @@ pub struct Conflict {
     /// List of paced train occurrences involved in the conflict.
     /// Each occurrence is identified by a `paced_train_id` and its `index`
     #[schema(inline)]
-    pub paced_train_occurrence_ids: Vec<PacedTrainOccurrenceId>,
+    paced_train_occurrence_ids: Vec<PacedTrainOccurrenceId>,
     /// List of work schedule ids involved in the conflict
     pub work_schedule_ids: Vec<i64>,
     /// Datetime of the start of the conflict
@@ -399,37 +403,24 @@ pub struct Conflict {
 }
 
 impl Conflict {
-    /// Converts the conflict data into a `ViewConflict` format.
-    ///
-    /// This function processes train schedule ids and paced trains generated ids ("paced_train_id#occurrence_id}")
+    /// This function processes train ids from Core Response
     ///  and maps them to either a `train_schedule_id` or a `paced_train_occurrence_id` based on the provided key mapping.
-    /// The train_id_map follows this structure:
-    /// - `train_id: (train_schedule_id, None)` if `train_id` corresponds to a train schedule ID.
-    /// - `train_id: (paced_train_id, occurrence_id)` if `train_id` corresponds to a generated paced train ID.
-    fn into_conflict_response(
-        conflict: CoreConflict,
-        train_id_map: HashMap<String, TrainId>,
-    ) -> Result<Self> {
-        let mut train_schedule_ids = Vec::new();
-        let mut paced_train_occurrence_ids = Vec::new();
+    fn from_core_response(conflict: CoreConflict) -> Result<Self> {
+        let (train_schedule_ids, paced_train_occurrence_ids): (Vec<_>, Vec<_>) = conflict
+            .train_ids
+            .iter()
+            .partition_map(|train_id| match train_id.parse() {
+                Ok(TrainId::TrainSchedule(id)) => Either::Left(id),
+                Ok(TrainId::PacedTrainOccurrence {
+                    paced_train_id,
+                    index,
+                }) => Either::Right(PacedTrainOccurrenceId {
+                    paced_train_id,
+                    index,
+                }),
+                Err(_) => unreachable!("Unreachable case encountered while partitioning train IDs"),
+            });
 
-        for train_id in &conflict.train_ids {
-            if let Some(train_id) = train_id_map.get(train_id) {
-                match train_id {
-                    TrainId::TrainSchedule(train_id) => train_schedule_ids.push(*train_id),
-                    TrainId::PacedTrainOccurrence(paced_train_occurrence) => {
-                        paced_train_occurrence_ids.push(paced_train_occurrence.clone())
-                    }
-                }
-            } else {
-                let train_id = train_id
-                    .parse::<i64>()
-                    .map_err(|_| TimetableError::ParseError {
-                        train_id: train_id.clone(),
-                    })?;
-                train_schedule_ids.push(train_id);
-            }
-        }
         let work_schedule_ids = conflict
             .work_schedule_ids
             .into_iter()
@@ -448,6 +439,55 @@ impl Conflict {
             conflict_type: conflict.conflict_type,
             requirements: conflict.requirements,
         })
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, ToSchema)]
+struct PacedTrainOccurrenceId {
+    paced_train_id: i64,
+    index: u64,
+}
+
+#[derive(Debug, Clone)]
+enum TrainId {
+    TrainSchedule(i64),
+    PacedTrainOccurrence { paced_train_id: i64, index: u64 },
+}
+
+impl Display for TrainId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::TrainSchedule(id) => write!(f, "{id}"),
+            Self::PacedTrainOccurrence {
+                paced_train_id,
+                index,
+            } => write!(f, "{paced_train_id}#{index}"),
+        }
+    }
+}
+
+impl FromStr for TrainId {
+    type Err = &'static str;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s.contains('#') {
+            let parts: Vec<&str> = s.split('#').collect();
+            if parts.len() != 2 {
+                return Err("Invalid PacedTrainOccurrenceId format");
+            }
+
+            let paced_train_id = parts[0].parse::<i64>().map_err(|_| "Invalid train id")?;
+            let index = parts[1]
+                .parse::<u64>()
+                .map_err(|_| "Invalid occurence id")?;
+            Ok(TrainId::PacedTrainOccurrence {
+                paced_train_id,
+                index,
+            })
+        } else {
+            let id = s.parse::<i64>().map_err(|_| "Invalid train id")?;
+            Ok(TrainId::TrainSchedule(id))
+        }
     }
 }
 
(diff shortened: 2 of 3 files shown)


editoast/src/views/timetable.rs
```diff
@@ -400,35 +404,19 @@ pub struct Conflict {
...
-                    TrainId::TrainSchedule(train_id) => train_schedule_ids.push(*train_id),
-                    TrainId::PacedTrainOccurrence(paced_train_occurrence) => {
-                        paced_train_occurrence_ids.push(paced_train_occurrence.clone())
-                    }
+            match TrainId::from_str(train_id) {
+                Ok(TrainId::TrainSchedule(id)) => train_schedule_ids.push(id),
+                Ok(TrainId::PacedTrainOccurrence(paced)) => paced_train_occurrence_ids.push(paced),
+                Err(_) => {
+                    unreachable!();
                 }
```
leovalais: note: [`partition_map`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.partition_map) covers this use case in a functional way ;)

editoast/src/views/timetable.rs
```diff
@@ -400,35 +404,19 @@ pub struct Conflict {
...
-                match train_id {
-                    TrainId::TrainSchedule(train_id) => train_schedule_ids.push(*train_id),
-                    TrainId::PacedTrainOccurrence(paced_train_occurrence) => {
-                        paced_train_occurrence_ids.push(paced_train_occurrence.clone())
-                    }
+            match TrainId::from_str(train_id) {
+                Ok(TrainId::TrainSchedule(id)) => train_schedule_ids.push(id),
+                Ok(TrainId::PacedTrainOccurrence(paced)) => paced_train_occurrence_ids.push(paced),
+                Err(_) => {
+                    unreachable!();
```
```suggestion
                    unreachable!("core cannot return an ID we didn't send, so all of them are parsable and valid: we either sent an invalid ID or core is faulty");
```

editoast/src/views/timetable.rs
```diff
@@ -400,35 +404,19 @@ pub struct Conflict {
...
         let mut paced_train_occurrence_ids = Vec::new();
 
         for train_id in &conflict.train_ids {
-            if let Some(train_id) = train_id_map.get(train_id) {
-                match train_id {
-                    TrainId::TrainSchedule(train_id) => train_schedule_ids.push(*train_id),
-                    TrainId::PacedTrainOccurrence(paced_train_occurrence) => {
-                        paced_train_occurrence_ids.push(paced_train_occurrence.clone())
-                    }
+            match TrainId::from_str(train_id) {
```
```suggestion
            match train_id.parse() {
```

editoast/src/views/timetable.rs
```diff
@@ -452,6 +440,63 @@ impl Conflict {
...
+impl Display for PacedTrainOccurrenceId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}#{}", self.paced_train_id, self.index)
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum TrainId {
+    TrainSchedule(i64),
+    PacedTrainOccurrence(PacedTrainOccurrenceId),
```
```suggestion
    PacedTrainOccurrence {
    	paced_train_id: i64,
    	index: u64,
    },
```

is the other definition necessary?

editoast/src/views/timetable.rs
```diff
@@ -452,6 +440,63 @@ impl Conflict {
...
+}
+
+impl FromStr for PacedTrainOccurrenceId {
+    type Err = String;
```
```suggestion
    type Err = &'static str;
```

(not 100% sure though)

editoast/src/views/timetable.rs
```diff
@@ -557,20 +560,77 @@ async fn conflicts(
...
+    let timetable_trains = TimetableWithTrains::retrieve_or_fail(conn, timetable_id, || {
+        TimetableError::NotFound { timetable_id }
+    })
+    .await?;
+
+    let (trains, _): (Vec<_>, _) =
+        TrainSchedule::retrieve_batch(conn, timetable_trains.train_ids).await?;
+
+    let (paced_trains, _): (Vec<_>, _) =
+        models::PacedTrain::retrieve_batch(conn, timetable_trains.paced_train_ids).await?;
```
leovalais: nit:

```suggestion
    let (trains, paced_trains): (Vec<_>, Vec<_>) =
    	tokio::join!(
    		TrainSchedule::retrieve_batch_unchecked(conn, timetable_trains.train_ids),
    		models::PacedTrain::retrieve_batch_unchecked(conn, timetable_trains.paced_train_ids)
    	).await?;
```

editoast/src/views/timetable.rs
```diff
@@ -557,20 +560,77 @@ async fn conflicts(
...
+        &mut db_pool.get().await?,
+        valkey_client.clone(),
+        core_client.clone(),
+        &paced_train_to_ts,
+        infra,
+        electrical_profile_set_id,
+    )
+    .await?;
+
+    Ok((train_simulations, paced_train_simulations))
```
leovalais: nit: probably can be `join!`ed as well

editoast/src/views/timetable.rs
```diff
@@ -581,58 +641,54 @@ fn build_conflict_core_request(
...
         let key = train.id.to_string();
-        map_string_to_id.insert(key.clone(), TrainId::TrainSchedule(train.id));
         trains_requirements.insert(
             key,
             TrainRequirements {
                 start_time: train.start_time,
-                spacing_requirements: final_output.spacing_requirements,
-                routing_requirements: final_output.routing_requirements,
+                spacing_requirements: final_output.spacing_requirements.clone(),
+                routing_requirements: final_output.routing_requirements.clone(),
```
leovalais: clone?

editoast/src/views/timetable.rs
```diff
@@ -581,58 +641,54 @@ fn build_conflict_core_request(
...
-        let entry = occurrences.entry(train.id).or_insert(0);
-        let occurrence_id = *entry;
-        let key = format!("{}#{}", train.id, occurrence_id);
-        map_string_to_id.insert(
-            key.clone(),
-            TrainId::PacedTrainOccurrence(PacedTrainOccurrenceId {
+    let mut current_index = 0;
+    for train in paced_trains.iter() {
+        let occurrences = (train.duration.num_seconds() / train.step.num_seconds()) as usize;
+        for (index, (sim, _)) in paced_train_simulations[current_index..occurrences]
```
leovalais: Since the simulations of a paced trains are contiguous, how about consuming from an iterator at each loop in order to:

1. catch the case "a simulation is missing" to panic with a nicer error message
2. avoid cloning the values

```rust
let mut it = paced_train_simulations.into_iter();
for train in &paced_trains { // no need to .iter() in for loops
        let occurrences = train.num_occurrences(); // a function would be nice
        for (index, (sim, _)) in it.take(occurrences).enumerate() {
```

editoast/src/views/timetable.rs
```diff
@@ -581,58 +641,54 @@ fn build_conflict_core_request(
...
-        *entry += 1;
+                index: index as u64,
+            }
+            .to_string();
+            trains_requirements.insert(
+                key,
+                TrainRequirements {
+                    start_time: train.start_time,
+                    spacing_requirements: final_output.spacing_requirements.clone(),
+                    routing_requirements: final_output.routing_requirements.clone(),
```
leovalais: I'm talking about these clones in ⬆️

editoast/src/views/timetable.rs
```diff
@@ -557,20 +560,77 @@ async fn conflicts(
...
+
+    Ok((train_simulations, paced_train_simulations))
+}
+
 fn build_conflict_core_request(
     infra: Infra,
     trains: Vec<TrainSchedule>,
     train_simulations: Vec<(SimulationResponse, PathfindingResult)>,
-    paced_trains_ts: Vec<TrainSchedule>,
+    paced_trains: Vec<models::PacedTrain>,
```
```suggestion
    paced_trains: &[models::PacedTrain],
```

we can avoid taking ownership in functions where this is not necessary

editoast/src/views/timetable.rs
```diff
@@ -557,20 +560,77 @@ async fn conflicts(
...
+
+    Ok((trains, paced_trains))
+}
+
+async fn retrieve_simulations(
+    db_pool: Arc<DbConnectionPoolV2>,
+    valkey_client: Arc<ValkeyClient>,
+    core_client: Arc<CoreClient>,
+    trains: &[TrainSchedule],
+    paced_trains: Vec<models::PacedTrain>,
```
leovalais: likewise about ownership

editoast/src/models/paced_train.rs
```diff
@@ -71,6 +71,20 @@ impl PacedTrain {
             options: self.options,
         }
     }
+
+    pub fn iter_occurrences(self) -> Vec<TrainSchedule> {
```
leovalais: could we have

```suggestion
    pub fn iter_occurrences(&self) -> impl Iterator<Item = TrainSchedule> {
```

leovalais (commented):
Thanks a lot for the follow up! A first round of review while this is still a draft:

editoast/src/views/timetable.rs
leovalais: You're right, I thought `take` would use a `&mut self` on the iterator but it takes ownership instead 😮

You indeed need `by_ref` as advised in the function doc in order to reborrow `it`, sorry about the confusion!

tests/tests/test_timetable.py
```diff
@@ -62,7 +62,7 @@ def test_conflicts(
     )
 
     stopping_paced_train_payload = stopping_train_schedule_payload[0]
-    stopping_paced_train_payload["start_time"] = "2024-05-22T08:05:00.000Z"
+    stopping_paced_train_payload["start_time"] = "2024-05-22T08:00:00.000Z"
```
shenriotpro: Why is this changing?

editoast/src/models/paced_train.rs
```diff
@@ -71,6 +71,29 @@ impl PacedTrain {
...
+            if i < occurrences {
+                let mut next = current.clone();
+                next.start_time = start_time + step * i as i32;
+                i += 1;
+                Some(next)
+            } else {
+                None
+            }
+        })
+    }
```
woshilapin: I believe this entire function can be much shorter and simpler.

```rust
    pub fn iter_occurrences(&self) -> impl Iterator<Item = TrainSchedule> + use<'_> {
        let base_occurrence = self.clone().into_first_occurrence();

        (0..self.num_occurrences()).map(move |occurrence_idx| TrainSchedule {
            start_time: base_occurrence.start_time + self.step * occurrence_idx as i32,
            ..base_occurrence.clone()
        })
    }
```

editoast/src/views/timetable.rs
```diff
@@ -558,31 +548,89 @@ async fn conflicts(
     let conflicts = conflict_detection_response.conflicts;
     let conflicts_response: Result<Vec<Conflict>> = conflicts
         .into_iter()
-        .map(|conflict| Conflict::into_conflict_response(conflict, map_string_to_id.clone()))
+        .map(Conflict::from_core_response)
         .collect();
     Ok(Json(conflicts_response?))
 }
 
+async fn retrieve_trains_and_paced_trains(
```
woshilapin, line 562: `retrieve_trains_and_paced_trains` and `retrieve_simulations` are somewhat similar, but one takes a `DbConnectionPoolV2` and the other one takes a `DbConnection`. Maybe it's a good time to make them both take a `DbConnection`?

editoast/src/views/timetable.rs
```diff
@@ -593,47 +641,49 @@ fn build_conflict_core_request(
...
-            key.clone(),
-            TrainId::PacedTrainOccurrence(PacedTrainOccurrenceId {
-                paced_train_id: train.id,
-                index: occurrence_id,
-            }),
-        );
+    let mut it = paced_train_simulations.into_iter();
+    for train in paced_trains {
+        let occurrences = &train.num_occurrences();
+        let items: Vec<_> = it.by_ref().take(*occurrences).collect();
```
woshilapin, line 652: I don't think you need to `.collect()` here since you only need the count of it if I understand correctly.

```suggestion
        let items: Vec<_> = it.by_ref().take(*occurrences).count();
```

woshilapin (commented):
It's a bit hard to understand the purpose of this Pull Request. I see some structures move from file to other file that make sense, some additional helper functions that also makes sense, but there is much more modifications in this Pull Request. But I'm not sure that I can make sense of all of it. Maybe a description would help?

editoast/src/views/timetable.rs
younesschrifi: thanks, done!

editoast/src/views/timetable.rs
younesschrifi: thanks, done :)

editoast/src/models/paced_train.rs
younesschrifi: thanks, done 👍🏽

editoast/src/views/timetable.rs
younesschrifi: I forgot that an iterator can only be consumed once... I need `items` and I use it below

editoast/src/models/paced_train.rs
```diff
@@ -71,6 +71,19 @@ impl PacedTrain {
             options: self.options,
         }
     }
+
+    pub fn iter_occurrences(&self) -> impl Iterator<Item = TrainSchedule> + use<'_> {
```
woshilapin: nitpick: I actually find weird in terms of API that `into_first_occurrence`, which yields a `TrainSchedule`, takes ownership of `self`, but `iter_occurences`, which also yields instances of `TrainSchedule`, takes only a reference of `self`. Maybe this function should also take ownership and let the caller decide if a `.clone()` is needed. What do you think?

note: might need to be renamed `into_iter_occurrences` if we choose to do this.

woshilapin (approved):

editoast/src/models/paced_train.rs
```diff
@@ -71,6 +71,19 @@ impl PacedTrain {
             options: self.options,
         }
     }
+
+    pub fn into_iter_occurrences(self) -> impl Iterator<Item = TrainSchedule> {
```
leovalais:
```suggestion
    pub fn iter_occurrences(&self) -> impl Iterator<Item = TrainSchedule> {
```

you're still cloning anyway (I think the train_simulation_batch stack can be changed to take `&[&TrainSchedule]` and we'd avoid even more cloning, but let's keep this pR simple)

editoast/src/views/timetable.rs
```diff
@@ -451,6 +442,49 @@ impl Conflict {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}#{}", self.paced_train_id, self.index)
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum TrainId {
+    TrainSchedule(i64),
+    PacedTrainOccurrence { paced_train_id: i64, index: u64 },
+}
```
```suggestion
#[derive(Debug, Clone)]
enum TrainId {
    TrainSchedule(i64),
    PacedTrainOccurrence { paced_train_id: i64, index: u64 },
}

impl Display for PacedTrainOccurrenceId {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
    	match self {
    		Self::TrainSchedule(id) => write!(f, "{id}"),
    		Self::PacedTrainOccurrence { paced_train_id, index } => write!(f, "{paced_train_id}#{index}"),
    	}
    }
}
```

editoast/src/views/timetable.rs
```diff
@@ -558,31 +548,87 @@ async fn conflicts(
...
+    paced_trains: &[models::PacedTrain],
+    infra: &Infra,
+    electrical_profile_set_id: Option<i64>,
+) -> Result<(
+    Vec<(SimulationResponse, PathfindingResult)>,
+    Vec<(SimulationResponse, PathfindingResult)>,
+)> {
+    let paced_train_to_ts = paced_trains
+        .iter()
+        .flat_map(|pt| pt.clone().into_iter_occurrences())
```
```suggestion
        .flat_map(|pt| pt.iter_occurrences())
```

editoast/src/views/timetable.rs
```diff
@@ -593,47 +639,48 @@ fn build_conflict_core_request(
...
-            },
-        );
+        for (index, (sim, _)) in items.into_iter().enumerate() {
+            let final_output = match sim {
+                SimulationResponse::Success { final_output, .. } => final_output,
+                _ => continue,
+            };
 
-        *entry += 1;
+            let key = PacedTrainOccurrenceId {
```
```suggestion
            let key = TrainId::PacedTrainOccurrence {
```

leovalais (commented):
A few last quick changes, otherwise LGTM thanks!

editoast/src/models/paced_train.rs
leovalais: Ideally, the signature should look like this imo:

```rust
pub fn iter_occurrences(&self) -> impl Iterator<Item = &'self TrainSchedule>
```

The problem is that every `train_simulation_batch` functions and friends take a `&[TrainSchedule]` and not a `&[&TrainSchedule]` (as they should since they do not consume the schedule), making the iterator item useless.

So since we want a `Item = TrainSchedule`, we still need to clone the base schedule once per occurrence. My point being, since we need to clone anyway, something we can't hide from the caller, let's clone once more for the `into_first_occurrence` and not consume `self`, bringing us closer to a more sound API.

Wdyt? @woshilapin @younesschrifi (unresolving)

editoast/src/models/paced_train.rs
leovalais: I didn't see the comment about this, I'll close this one to keep the discussion there : https://github.com/OpenRailAssociation/osrd/pull/11355#discussion_r2053743491

editoast/src/models/paced_train.rs
leovalais: moved to https://github.com/OpenRailAssociation/osrd/pull/11355#discussion_r2053743491

editoast/src/views/timetable.rs
```diff
@@ -558,31 +552,87 @@ async fn conflicts(
...
     // Build train schedule train requirements
-    for (train, sim) in trains.into_iter().zip(train_simulations) {
+    for (train, sim) in trains.iter().zip(train_simulations) {
         let (sim, _) = sim;
         let final_output = match sim {
             SimulationResponse::Success { final_output, .. } => final_output,
             _ => continue,
         };
-
         let key = train.id.to_string();
```
```suggestion
        let key = TrainId::TrainSchedule(train.id).to_string();
```

editoast/src/views/timetable.rs
```diff
@@ -451,6 +442,53 @@ impl Conflict {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, ToSchema)]
+pub struct PacedTrainOccurrenceId {
+    pub paced_train_id: i64,
+    pub index: u64,
+}
```
leovalais: doesn't need to be pub

leovalais (approved):
Tested

editoast/src/models/paced_train.rs
leovalais: @younesschrifi let's give @woshilapin the chance to agree or disagree before resolving and merge the pr 😄

PR #11356, opened 2025-04-07
front: move TrainSpaceTimeData into modules/simulationResults

clarani:


SharglutDev (approved):
Should `usePathProjection` be moved as well ?

clarani:
`usePathProjection` is only used in OperationalStudies, so I think it's fine to keep it in this folder :)

PR #11357, opened 2025-04-07
front: occupancy blocks: add blinking color

### LINKED ISSUE #9936: Blinking signals don't appear stripped
### What happened?

Blinking signals should appear stripped in the STD (unlike non-blinking signals)

Image

_No response_

- In the infrastructure editor, check the "Is the signal yellow blinking" checkbox of a BAL signal.
- Create a train that goes through it.

Local

Firefox

1d3b9f7474b39faa1cdf30e0677f2d7c7edf6e04

Akctarus:
closes #9936


kmer2016 (commented):
The code is ok, but the PR is neither testable nor mergeable in its current state, because we have to wait for the osrd-ui release.

Uriel-Sautron (approved):
LGTM

kmer2016 (approved):

PR #11383, opened 2025-04-08
Fix import railjson version check

### LINKED ISSUE #10893: infrastructure file cannot be imported
### What happened?

I tried to export the railjson file of an infrastructure and re-import it
I got the following error

Image

I expected to be able to reimport my file

export and re-import an infratructure

Recette (SNCF)

Google Chrome

776f9b3

flomonster:
close #10893

We already check if the major version matches the application one.

### DIFF
--- a/editoast/openapi.yaml
+++ b/editoast/openapi.yaml
@@ -5157,7 +5157,6 @@ components:
       - $ref: '#/components/schemas/EditoastLayersErrorViewNotFound'
       - $ref: '#/components/schemas/EditoastLinesErrorsLineNotFound'
       - $ref: '#/components/schemas/EditoastListErrorsErrorsWrongErrorTypeProvided'
-      - $ref: '#/components/schemas/EditoastListErrorsRailjsonWrongRailjsonVersionProvided'
       - $ref: '#/components/schemas/EditoastMacroNodeErrorNotFound'
       - $ref: '#/components/schemas/EditoastModelError'
       - $ref: '#/components/schemas/EditoastNoSuchUserErrorNoSuchUser'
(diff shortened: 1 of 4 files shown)


emersion:
Do we need to perform migrations on older RailJSON files?

flomonster:
> Do we need to perform migrations on older RailJSON files?

No, it's not required. It could be performed in another PR.

emersion:
I mean, what if someone submits a RailJSON file with e.g. version 3.4.11? The JSON migration in `2024-05-02-082903_remove_schematic/up.sql` won't be applied.

flomonster:
> I mean, what if someone submits a RailJSON file with e.g. version 3.4.11? The JSON migration in `2024-05-02-082903_remove_schematic/up.sql` won't be applied.

It's not an issue since we are supposed (and I think it's the case) to be compatible. When the railjson is submitted, the railjson version in DB will be bumped to the last one.

> [!NOTE]
> We don't want to apply migration when a railjson is submitted (cause it requires a lot of work).

SharglutDev (approved):
Lgtm (not tested)

Khoyo:
In the past, minor version bumps for Railjson were those where the new version is incompatible but handled by db migrations (eg. the split of TVM signaling, where core will refuse to load any infra with `TVM` signals). I'm guessing this guideline will have to change, and all incompatible railjson changes will have to be a major version bump?

I'm not sure if there is any of those in the current rjs major version, if so we should probably do a major version bump to prevent incompatible infras being loaded.

woshilapin (approved):
OK, wondered why just removing was ok... and then found the previous #11205 that is some code down in the `persist` function. Seems good.

PR #11384, opened 2025-04-08
front: fix train selection after edition

### LINKED ISSUE #11338: When editing a train/service, the selected item changes and is the first train/occurrence of the scenario
### What happened?

I select a train or occurrence in the scenario (not the first one). Then I edit it. Then the edition is done but the selected item becomes the first train or occurrence in the scenario (chronologically speaking). It is confusing because it brings me to another train, with another path and so one.

https://github.com/user-attachments/assets/6ca8a4ef-ef2c-473d-b73c-b13a4b0e9893

I expected to have still the same selected item not to be changed when I edit a train (which is not necessarily the selected item !)

As in the video 

Recette (SNCF)

Google Chrome

b42f78b

clarani:
closes #11338

### DIFF
--- a/front/src/modules/trainschedule/helpers/pacedTrain.ts
+++ b/front/src/modules/trainschedule/helpers/pacedTrain.ts
@@ -1,0 +1,9 @@
+/* eslint-disable import/prefer-default-export */
+import type { PacedTrainWithDetails } from '../components/Timetable/types';
+
+export const getOccurrencesNb = ({ duration, step }: PacedTrainWithDetails['paced']) => {
+  if (step.ms === 0) {
+    throw new Error('Step cannot be 0');
+  }
+  return Math.ceil(duration.ms / step.ms);
+};
--- a/front/src/applications/operationalStudies/hooks/useAutoUpdateProjection.ts
+++ b/front/src/applications/operationalStudies/hooks/useAutoUpdateProjection.ts
@@ -12,6 +12,7 @@ import {
 } from 'reducers/simulationResults/selectors';
 import { useAppDispatch } from 'store';
 import {
+  extractPacedTrainIdFromOccurrenceId,
   formatEditoastTrainIdToOccurrenceId,
   formatPacedTrainIdToEditoastTrainId,
   isPacedTrain,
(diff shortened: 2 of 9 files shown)


front/src/modules/trainschedule/helpers/pacedTrain.ts
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable import/prefer-default-export */
+import type { PacedTrain } from 'common/api/osrdEditoastApi';
+import { Duration } from 'utils/duration';
+
+export const getOccurrencesNb = (pacedTrain: PacedTrain) => {
```
SharglutDev: We could use this new function in `useOccurrences` l.24 and `paced-train-sections` l.84 ?

front/src/modules/trainschedule/components/ManageTrainSchedule/hooks/useUpdateTrainSchedule.ts
```diff
@@ -101,10 +109,13 @@ const useUpdateTrainSchedule = (
           timetable_id: timetableId!,
           id: trainIdToEdit,
         };
+        const occurrenceCount = getOccurrencesNb(pacedTrain);
+        if (occurrenceCount > 1) {
```
SharglutDev: Why not `> 0` ?

front/src/modules/trainschedule/components/ManageTrainSchedule/hooks/useUpdateTrainSchedule.ts
```diff
@@ -101,10 +109,13 @@ const useUpdateTrainSchedule = (
...
+          trainIdToSelect = formatPacedTrainIdToOccurrenceId(trainIdToEdit, 0);
```
SharglutDev: Doesn't change much while we don't have exceptions but if I selected occurrence with index 2, after editing I'll always have the first occurrence selected (even if the interval or time window didn't change).

front/src/modules/trainschedule/helpers/pacedTrain.ts
clarani: Hum in `useOccurrences`, we manipulate `PacedTrainWithDetails` and not `PacedTrain`.
Same for `paced-train-sections`

front/src/modules/trainschedule/components/ManageTrainSchedule/hooks/useUpdateTrainSchedule.ts
clarani: Good catch

front/src/modules/trainschedule/components/ManageTrainSchedule/hooks/useUpdateTrainSchedule.ts
clarani: Yes but for the moment, we don't allow the user to edit an occurrence

front/src/modules/trainschedule/helpers/pacedTrain.ts
SharglutDev: Oh right, maybe we could change `getOccurrencesNb` to pass `paced` (`PacedTrain["paced"]` / `PacedTrainWithDetails["paced"]`) in it and not the whole paced train ?

front/src/modules/trainschedule/components/ManageTrainSchedule/hooks/useUpdateTrainSchedule.ts
SharglutDev: Yes but we allow the user to select one, so it could start with occurrence 2 selected and end up with occurrence 1 selected. But sounds painful to handle now, we can keep it for exceptions (we could add this case in the ACs)

theocrsb (approved):
lgtm and tested

front/src/modules/trainschedule/components/ManageTrainSchedule/hooks/useUpdateTrainSchedule.ts
SharglutDev, line 1: nit: we should at some point rename this file since it also handles paced trains

front/src/modules/trainschedule/components/ManageTrainSchedule/hooks/useUpdateTrainSchedule.ts
```diff
@@ -101,10 +111,18 @@ const useUpdateTrainSchedule = (
           timetable_id: timetableId!,
           id: trainIdToEdit,
         };
+
+        // if the selected train is an occurence of the edited paced train, keep it selected
```
SharglutDev:
```suggestion
        // if the selected item is an occurrence of the edited paced train, keep it selected
```

SharglutDev (approved):
Lgtm and tested, good job ! You can resolve the last comments

front/src/modules/trainschedule/components/ManageTrainSchedule/hooks/useUpdateTrainSchedule.ts
clarani: Done :)

front/src/modules/trainschedule/components/ManageTrainSchedule/hooks/useUpdateTrainSchedule.ts
clarani: Done !

PR #11386, opened 2025-04-08
LMR: Canceling a simulation for alternative solutions doesn't display the results of the simulation

### LINKED ISSUE #11341: LMR: Canceling a simulation for alternative solutions doesn't display the results of the simulation
### What happened?

When a simulaiton is canceled during the search for the alternative results, the result of the first simulaiton is not displayed.

https://github.com/user-attachments/assets/ea592edf-22a9-41c2-bd00-d25b4983071e

According to the AC of https://github.com/osrd-project/osrd-confidential/issues/442, the result of the first simulation should be displayed.

1. Open LMR
2. Launch a simulation with alternative solutions
3. Cancel it when the message indicating the research for the alternative solutions appears 

Recette (SNCF)

Firefox

b42f78b

achrafmohye:
Closes #11341

### DIFF
--- a/front/src/applications/stdcm/hooks/useStdcm.ts
+++ b/front/src/applications/stdcm/hooks/useStdcm.ts
@@ -56,7 +56,7 @@ const useStdcm = ({
   const osrdconf = useSelector(getStdcmConf);
   const timetableId = useSelector(getStdcmTimetableID);
   const infraId = useSelector(getStdcmInfraID);
-  const requestPromise = useRef<ReturnType<typeof postTimetableByIdStdcm>>();
+  const requestPromise = useRef<ReturnType<typeof postTimetableByIdStdcm>[]>();
   const isCancelledRef = useRef(false);
 
   const currentSimulationInputs = useStdcmForm();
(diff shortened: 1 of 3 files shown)


front/src/applications/stdcm/hooks/useStdcm.ts
```diff
@@ -155,6 +155,16 @@ const useStdcm = ({
     payload: PostTimetableByIdStdcmApiArg
   ) => {
     try {
+      const currentSimulation = await createSimulation(
+        currentSimulationInputs,
+        payload,
+        response,
+        undefined
+      );
```
kmer2016: No need to pass undefined as the param is optional
```suggestion
    try {
      const currentSimulation = await createSimulation(currentSimulationInputs, payload, response);
```

front/src/applications/stdcm/hooks/useStdcm.ts
```diff
@@ -178,15 +188,14 @@ const useStdcm = ({
...
       const downstreamInputs = adjustInputByDirection(currentSimulationInputs, 'downstream');
 
-      const [currentSimulation, downstreamSimulation, upstreamSimulation] = await Promise.all([
-        createSimulation(currentSimulationInputs, payload, response, undefined),
+      const [downstreamSimulation, upstreamSimulation] = await Promise.all([
         createSimulation(downstreamInputs, payloadDownstream, resDown, 'downstream'),
         createSimulation(upstreamInputs, payloadUpstream, resUp, 'upstream'),
       ]);
 
       if (isCancelledRef.current) return;
```
kmer2016: Do we need to check `isCancelledRef` here too ?

SharglutDev:
> [!WARNING]
> Something seems to be wrong with your commit's name

kmer2016 (approved):
LGTM ! Good job

SarahBellaha (commented):
https://github.com/user-attachments/assets/9be83d63-cc05-4406-8b6f-8503962ca511


The behavior is still a bit strange. When we click on the cancel button after the 1st simulation has been computed, it doesn't really abort the 2 aditionnal requests. Which leads to a delay between the moment you cancel and the moment you have the 1st simulation result displayed under.
We should be able to abort those requests to avoid that.


⬇️  I also think that this message should not be displayed when we cancel additional requests.

image

SarahBellaha (approved):
LGTM, thx !

PR #11387, opened 2025-04-08
editoast: replace `InternalError` by `StandardCoreError` in core module

hamz2a:
Part of https://github.com/OpenRailAssociation/osrd/issues/10868

### DIFF
--- a/editoast/src/core/mod.rs
+++ b/editoast/src/core/mod.rs
@@ -23,6 +23,7 @@ use serde_json::Value;
 use thiserror::Error;
 use tracing::error;
 use tracing::trace;
+use utoipa::ToSchema;
 
 #[cfg(test)]
 use crate::core::mocking::MockingError;
@@ -264,7 +265,7 @@ impl Error {
     }
 }
 
-#[derive(Debug, Deserialize, PartialEq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, ToSchema)]
 pub struct StandardCoreError {
     #[serde(skip)]
     pub status: StatusCode,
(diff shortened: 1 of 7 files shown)


editoast/src/core/mod.rs
```diff
@@ -273,10 +274,26 @@ pub struct StandardCoreError {
...
+            Error::StandardCoreError(standard_core_error) => standard_core_error,
+            _ => Self {
+                status: StatusCode::INTERNAL_SERVER_ERROR,
+                error_type: "editoast:coreclient:StandardCoreError".to_string(),
+                context: HashMap::default(),
+                message: "Core returned an error in an unknown format".to_string(),
+                cause: CoreErrorCause::Internal,
+            },
+        }
+    }
```
leovalais, line 293: This fallback case is wrong. What about `MqClientError` for instance?

leovalais (commented):
I'm lost in the implementation plan of this refactoring: I don't understand how this PR helps with the error refactoring since it's basically just a `s/InternalError/StandardCoreError`? Shouldn't `PathfindingCoreResult` and `SimulationResponse` be simplified first to eliminate non-success variants? And therefore transferring the responsibility to match on the error to the views?

hamz2a:
> I'm lost in the implementation plan of this refactoring: I don't understand how this PR helps with the error refactoring since it's basically just a `s/InternalError/StandardCoreError`? Shouldn't `PathfindingCoreResult` and `SimulationResponse` be simplified first to eliminate non-success variants? And therefore transferring the responsibility to match on the error to the views?

If I understand your point correctly, you're suggesting that we should completely remove the `SimulationFailed` variant from `SimulationResponse` and the `InternalError` from `PathfindingCoreResult`, right?

leovalais:
Yes, I think that's what we discussed a while ago in your first PR? The same would be done for the pathfinding I believe

leovalais:
https://github.com/OpenRailAssociation/osrd/pull/11041#pullrequestreview-2718628162

hamz2a:
> Yes, I think that's what we discussed a while ago in your first PR? The same would be done for the pathfinding I believe

I thought we could solve this by replacing `InternalError` with `StandardCoreError`, but in the end, the simplest solution is to just remove the variant that corresponds to the error.

leovalais:
> the simplest solution is to just remove the variant that corresponds to the error

It sure is, but we need to put it back in the views to keep our API unchanged

editoast/src/core/mod.rs
```diff
@@ -264,7 +265,7 @@ impl Error {
     }
 }
 
-#[derive(Debug, Deserialize, PartialEq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, ToSchema)]
```
woshilapin, line 268: As discussed earlier, we might not want to expose (e.g. derive `ToSchema`) for errors coming from `core`.

hamz2a:
This PR is no longer up to date.
https://github.com/OpenRailAssociation/osrd/pull/11649

PR #11388, opened 2025-04-08
core: move files to have a clean kotlin api package

Erashin:
Moved files around to cleanly separate the api files (endpoints, requests, responses, parsers and post processing) from the rest.
Added bonus: corrected a couple of warnings.


core/src/main/kotlin/fr/sncf/osrd/api/RollingStockParser.kt
```diff
@@ -1,5 +1,6 @@
-package fr.sncf.osrd.api.standalone_sim
+package fr.sncf.osrd.api
 
+import fr.sncf.osrd.api.standalone_sim.PhysicsConsistModel
```
bougue-pe, line 3: Not sure: Should it move too?

eckter (approved):

bougue-pe (approved):
Thanks !
(feel free to resolve comment)

core/src/main/kotlin/fr/sncf/osrd/api/RollingStockParser.kt
Erashin: Hmm. Honestly there's a couple of objects still used in multiple endpoints that could be cleaned up. Might be worth to create a whole new package all-together for theses objects (maybe call it `model` or something)., For now I'll just leave this as is.

PR #11390, opened 2025-04-08
front: redesign lazy loading for simulation summaries

### LINKED ISSUE #11410: Trains are not loaded when you first leave and return to a scenario.
### What happened?

When you open a scenario with paced trains and unique trains, only paced trains load.

If you quit and re-open it, both unique and paced trains load.

After a few minutes, this behavior is reproduced.

https://github.com/user-attachments/assets/1787139f-eab0-43fb-af00-6609b46b0e94

It also behaves the same way if you reload the scenario page.

Both unique and paced trains should load.

1. Create a scenario with both paced and unique trains
2. Close it
3. Wait a few minutes
4. Try to epen it and see only paced trains load
5. Close it
6. Re-open it and see all trains

Recette (SNCF)

Firefox

de411c7

### LINKED ISSUE #11572: After page reload only the paced train is displayed
### What happened?

https://github.com/user-attachments/assets/93c4fe78-a5c8-4a03-b582-95489f8750e8

All the timetable items are displayed 

Check video
Scenario : https://rec-osrd.reseau.sncf.fr/operational-studies/projects/6/studies/664/scenarios/1483

Recette (SNCF)

Google Chrome

20afde5

emersion:
closes #11410 
closes #11572 

TODO:
- [x] Use the new hook in operational studies
- [x] Use the new hook in STDCM
- [x] Docs

### Simon, what's your beef with `useLazyLoadTimetableItems()`?

`useLazyLoadTimetableItems()` has served us well, but was in my honest opinion a bit difficult to work with because:

- Lack of encapsulation: state setters are passed around (`setTimetableItemIdsToProject()`, `setTimetableItemSummariesById()`, etc), leaking internals to other components and hooks.
- Unclear data flow: `useEffect()` and `useState()` are over-used, making it difficult to understand where the inputs are coming from. When a `useEffect()` is called, it's tricky to figure out the difference with the previous state. For instance, when `timetableItemIdsToFetch` is updated, what are the new items that we need to fetch, these already fetched, these currently being fetched, etc?
- Confusion about variable capture: having the whole logic inside a hook makes it complicated to figure out when variables are captured in async arrow functions. An async arrow function captures variables when it starts executing, but the variables aren't updated when the hook is called again. Multiple async functions triggered by the hook may run in parallel.
- Remove the `outOfSync` workaround: this got added because of some mismatch between the timetable item path and the simulation result path. I'm not 100% sure all edge cases are fixed, but the new architecture should make it easier to properly fix the issue without workarounds. The workaround caused issues with the NGE paced train integration (some trains would disappear sometimes).

### What is it that we need?

The replacement needs to tick these boxes:

- Simulation results are incrementally returned: we'll load simulations in batches, and we want to display a batch in the UI as soon as it's received from the server.
- Simulation & projection can be done standalone: STDCM debug mode needs to filter simulated trains before projecting them.
- Play well with `upsertTimetableItem()`/`removeTimetableItem()`/`updateTrainDepartureTime()`: we sometimes need to request a single train to be simulated again, or dropped from the results.
- Don't abuse rtk-query hooks: they only allow one subscribed request, so we shouldn't call these in a loop.

### Why split the logic into two files?

The goal is separation of concerns: `TrainSimulationLazyLoader` is only responsible for performing HTTP requests sequentially, while `useLazySimulateTrains()` performs React integration (by exposing a hook and holding all results in a state) and post-processing steps (to mix the simulation results with timetable items data).

### Why the rename?

I used the word "simulate" instead of "load". The hook doesn't just load data from the database: it requests simulations to be computed by the backend. I prefer to make this explicit and clearly indicate that this is different from `GET /timetable/<id>/train_schedules`.

### Why introduce a class, instead of a type and functions?

It felt more natural to define class methods rather than standalone functions mutating an object. Classes can also hide their internal fields (by defining defaults or initializing them in constructors), and are stricter than type aliases (any random object which happens to have the right fields doesn't satisfy the type checks).

In general, I think it makes sense to use type aliases to define the shape of _data structures_, and use classes for associated functions which need to keep state.

### Got another question?

Please ask me to add a section about it!


front/src/applications/operationalStudies/hooks/useLazySimulateTrains.ts
```diff
@@ -0,0 +1,92 @@
...
+        if (onProgress) onProgress(summaries);
+      },
+    });
+
+    loader.simulateTimetableItems([...timetableItemsByIdRef.current.keys()]);
+
+    loaderRef.current = loader;
+    return () => {
+      loader.cancel();
+      loaderRef.current = undefined;
```
clarani, line 75: Why do you need to set the current ref to undefined here ? Is it for  `allTrainsSimulated` ?

clarani (commented):
Super clean :) It's very nice, thanks !!

front/src/applications/operationalStudies/hooks/useLazySimulateTrains.ts
emersion: Mostly a matter of avoiding a ref left behind with a loader object that we shouldn't use anymore.

front/src/applications/stdcm/hooks/useProjectedTrainsForStdcm.ts
```diff
@@ -142,7 +136,7 @@ const useProjectedTrainsForStdcm = (stdcmResponse?: StdcmSuccessResponse) => {
 
   return {
     spaceTimeData,
-    projectionLoaderData: { allTrainsProjected, totalTrains: trainIds.length },
+    projectionLoaderData: { allTrainsProjected, totalTrains: timetable?.length ?? 0 },
```
SharglutDev, line 139: Not related to your PR but if we project only the trains "framing" the stdcm result, would we want to show only the number of trains in this window for the totalTrains

front/src/applications/operationalStudies/hooks/useScenarioData.ts
```diff
@@ -89,20 +85,21 @@ const useScenarioData = (scenario: ScenarioResponse, infra: InfraWithState) => {
...
+    removeSimulatedTimetableItems,
+  } = useLazySimulateTrains({
+    infraId: scenario.infra_id,
+    electricalProfileSetId,
+    rollingStocks,
+    onProgress: (summaries) => {
+      // TODO Paced trains : remove this filter when paced trains are handled in projection
+      const trainScheduleIds = [...summaries.keys()].filter(isTrainSchedule);
+      setTimetableItemIdsToProject((prev) => new Set([...prev, ...trainScheduleIds]));
+    },
```
SharglutDev, line 108: In the future, will it be necessary to refacto lazy projection as well ?

front/src/applications/operationalStudies/hooks/useLazySimulateTrains.ts
```diff
@@ -0,0 +1,107 @@
...
+import { useAppDispatch } from 'store';
+
+import formatTimetableItemSummaries from '../helpers/formatTimetableItemSummaries';
+import TrainSimulationLazyLoader from '../helpers/TrainSimulationLazyLoader';
+
+type UseLazySimulateTrainsOptions = {
+  infraId: number;
+  electricalProfileSetId: number | undefined;
+  rollingStocks: LightRollingStockWithLiveries[] | null;
+  onProgress?: (results: Map<TimetableItemId, TimetableItemWithDetails>) => void;
```
SharglutDev, line 18: For now, on both calls of this hook, `onProgress` is defined. Do you plan another use for it where it could be undefined ?

front/src/applications/operationalStudies/hooks/useLazySimulateTrains.ts
```diff
@@ -0,0 +1,107 @@
...
+      electricalProfileSetId,
+      onProgress: (rawSummaries: Map<TimetableItemId, SimulationSummaryResult>) => {
+        const summaries = formatTimetableItemSummaries(
+          [...rawSummaries.keys()],
+          rawSummaries,
+          timetableItemsByIdRef.current,
+          rollingStocks
+        );
+        setSimulatedTrainsById((prev) => new Map([...prev.entries(), ...summaries.entries()]));
+        if (onProgress) onProgress(summaries);
```
SharglutDev, line 66: This is a bit weird `onProgress` calling `onProgress`. Feels like a recursive function. Would you consider renaming one of them ?

SharglutDev (commented):
Great job for this refacto ! It's very clear.

front/src/applications/stdcm/hooks/useProjectedTrainsForStdcm.ts
emersion: Good point. Also, we might want to show a loader when we're simulating all trains.

front/src/applications/operationalStudies/hooks/useScenarioData.ts
emersion: Yup, that's next on my TODO list :P

front/src/applications/operationalStudies/hooks/useLazySimulateTrains.ts
emersion: It could be undefined when there is no need to launch a projection (we might not need to in the future, when all space time charts are hidden in the modular scenario view). I made it optional since results are provided via a return value too. If you prefer, we can make it required for now, and make it optional if desirable later.

front/src/applications/operationalStudies/hooks/useLazySimulateTrains.ts
emersion: I think it makes sense to have them both have the same name, since they have the exact same purpose.

front/src/applications/operationalStudies/hooks/useLazySimulateTrains.ts
SharglutDev: No don't bother, it's totally fine by me :)

front/src/applications/operationalStudies/hooks/useLazySimulateTrains.ts
```diff
@@ -0,0 +1,107 @@
...
+
+import type {
+  LightRollingStockWithLiveries,
+  SimulationSummaryResult,
+} from 'common/api/osrdEditoastApi';
+import type { TimetableItemWithDetails } from 'modules/trainschedule/components/Timetable/types';
+import type { TimetableItemId, TimetableItemWithTimetableId } from 'reducers/osrdconf/types';
+import { useAppDispatch } from 'store';
+
+import formatTimetableItemSummaries from '../helpers/formatTimetableItemSummaries';
```
louisgreiner: I may be doing it wrong, but the imports are broken when I try to test it locally

```suggestion
import formatTimetableItemSummaries from 'modules/simulationResult/helpers/formatTimetableItemSummaries';
```

emersion: Not sure what would cause this, relative imports are supposed to work fine and CI doesn't show any error.

louisgreiner: let's ignore it then

SharglutDev (approved):
Lgtm and tested, very nice !

front/src/applications/operationalStudies/hooks/useLazySimulateTrains.ts
```diff
@@ -0,0 +1,107 @@
...
+          rollingStocks
+        );
+        setSimulatedTrainsById((prev) => new Map([...prev.entries(), ...summaries.entries()]));
+        if (onProgress) onProgress(summaries);
+      },
+    });
+
+    loader.simulateTimetableItems([...timetableItemsByIdRef.current.keys()]);
+
+    loaderRef.current = loader;
```
anisometropie, line 72: why do you create a new one each time in this useEffect ? any reason not to keep the same instance and just feed it more trains to project once the need arises ?

front/src/applications/operationalStudies/helpers/TrainSimulationLazyLoader.ts
```diff
@@ -0,0 +1,146 @@
...
+   */
+  cancel() {
+    this.pending = [];
+    this.cancelled = true;
+  }
+
+  async processPending() {
+    while (this.pending.length > 0) {
+      const batch = this.pending.slice(0, BATCH_SIZE);
+      this.pending = this.pending.slice(BATCH_SIZE);
```
anisometropie, line 72: creating new arrays every cycle is wasteful in JS, if we have thousands of trains I think it might make a difference. 
Here you can work in place and start taking trains from the end of the array to avoid shifting.

what do you think ?

```suggestion
  const batch = this.pending.splice(-BATCH_SIZE);
  await this.processBatch(batch);
```

front/src/applications/operationalStudies/helpers/TrainSimulationLazyLoader.ts
```diff
@@ -0,0 +1,146 @@
...
+              },
+            },
+            { subscribe: false }
+          )
+        )
+        .unwrap();
+    }
+
+    const rawTrainScheduleSummaries = await trainSchedulePromise;
+    const rawPacedTrainSummaries = await pacedTrainPromise;
```
anisometropie, line 128: Feels a bit weird to define and `await` empty promises.
why don’t you just set `rawPacedTrainSummaries = {}` in case we have nothing to process (and avoid using an unecessary promise in that case altogether) and `rawPacedTrainSummaries = await pacedTrainPromise` otherwise ?

front/src/applications/operationalStudies/helpers/TrainSimulationLazyLoader.ts
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  /**
+   * Queue train IDs for simulation.
+   */
+  simulateTimetableItems(ids: TimetableItemId[]) {
+    if (this.cancelled) {
+      throw new Error('simulateTimetableItems() called after cancel()');
+    }
+    this.pending = [...this.pending, ...ids];
```
anisometropie: here we don’t really care about being immutable either, I would do it in place too.

front/src/applications/operationalStudies/hooks/useLazySimulateTrains.ts
anisometropie: That puzzled me too, but indeed we have no shadowing happening here. it’s just an object key holding an anonymous function 

Screenshot_20250428_193252

anisometropie: on the other hand 

Screenshot_20250428_193456

anisometropie (changes requested):
Solid rework of this part of the code. that’s really a step in the right direction to solve our hooks spaghetti hell.

good job !

I have a few remarks, mostly performance related.

front/src/applications/operationalStudies/hooks/useLazySimulateTrains.ts
emersion: The parameters of the loader (infra ID, etc) change. We want to cancel any in-flight requests. Re-creating the whole loader is simpler and more robust.

front/src/applications/operationalStudies/helpers/TrainSimulationLazyLoader.ts
emersion: Unless there is a proof that this actually matters in a benchmark, I'd say this is premature optimization.

front/src/applications/operationalStudies/helpers/TrainSimulationLazyLoader.ts
emersion: We want to run both requests in parallel, not one after the other.

front/src/applications/operationalStudies/helpers/TrainSimulationLazyLoader.ts
emersion: I've changed it to `.push(...ids)` since that doesn't impact readability.

front/src/applications/operationalStudies/helpers/TrainSimulationLazyLoader.ts
anisometropie: hm, right there is no nice solution to keep it running in parallel.

front/src/applications/operationalStudies/helpers/TrainSimulationLazyLoader.ts
anisometropie: This is known to use more memory, which over times will demand more garbage collection and in the end slow down the application. our application is already slow as it is, what do we lose from optimizing here ?

emersion: I find `splice()` to be confusing and best avoided, because it does a lot of different things and figuring out what the call actually does requires pulling up the docs and carefully reading each clause. I value readability over performance, and in general I think this has very little to no impact in practice because it's just an array of a few thousand integers in the worst case.

The app is slow, but making this change will not fix that slowness. The slowness will not worsen because of this change. Fixing the slowness requires making benchmarks and interpreting CPU/memory usage graphs.

clarani (approved):
LGTM :white_check_mark:

anisometropie (approved):

PR #11399, opened 2025-04-09
front: display feedbackcard below both results and failure sections

### LINKED ISSUE #11389: The mail adress is not available when no simulation found on stdcm
### What happened?

On STDCM, when no simulation is found, the feedback email address below the simulation is not accessible.

- The user should be able to access the feedback email address when a simulation fails.

1. Launch a simulation with all mandatory inputs filled and set the max speed to 30 km/h.
2. If the itinerary is long enough, no simulation should be returned.
3. The feedback email address is not available.

Recette (SNCF), Production (SNCF)

Firefox, Google Chrome

b42f78b

Caracol3:
close #11389 

The form data used in the mail is no longer the output data but the input data, because the user should be able to send an email with prefilled values even if the simulation did not succeed.

### DIFF
--- a/front/src/applications/stdcm/components/StdcmResults/StdcmFeedback.tsx
+++ b/front/src/applications/stdcm/components/StdcmResults/StdcmFeedback.tsx
@@ -1,11 +1,8 @@
 import { Comment } from '@osrd-project/ui-icons';
-import dayjs from 'dayjs';
 import { useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
 
-import type { StdcmResultsOutput } from 'applications/stdcm/types';
 import { getSelectedSimulation } from 'reducers/osrdconf/stdcmConf/selectors';
-import { dateTimeFormatting } from 'utils/date';
 import useDeploymentSettings from 'utils/hooks/useDeploymentSettings';
 
 const StdcmFeedback = () => {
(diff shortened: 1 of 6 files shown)


front/src/styles/scss/applications/stdcm/_results.scss
```diff
@@ -355,9 +357,8 @@
       display: flex;
       flex-direction: column;
       flex-shrink: 0;
-      width: 836px;
+      width: 804px;
```
Uriel-Sautron: You can remove this width, you move it in `.results-with-feedback`

Caracol3: done

kmer2016 (approved):
LGTM

front/src/applications/stdcm/components/StdcmResults/StdcmFeedback.tsx
```diff
@@ -16,19 +13,18 @@ const StdcmFeedback = ({ stdcmFeedbackMail }: StdcmFeedbackProps) => {
...
-  );
+  const { inputs } = selectedSimulation;
+  const { departureTime } = inputs;
+  const data = inputs.consist;
+
+  const trainName = data?.tractionEngine?.name ?? '-';
+  const consistCode = data?.speedLimitByTag ?? '-';
+  const consistLength = data?.totalLength ? `${data.totalLength} m` : '-';
+  const consistMass = data?.totalMass ? `${data.totalMass} t` : '-';
+  const maxSpeed = data?.maxSpeed ? `${Math.round(data.maxSpeed * 3.6)} km/h` : '-';
```
Yohh: nit: what about consistData?

Yohh: you should find a way to avoid those optional `data`, then maybe destructure the keys line 18

Yohh: you should find a way to avoid those optional `data`, then maybe destructure the keys line 18

Caracol3: done

Yohh (approved):

PR #11400, opened 2025-04-09
front: fix rounding times for timetable items cards

### LINKED ISSUE #11337: Time rounding up is different beteween paced trains and train schedules
### What happened?

For a same start time for a paced train and train schedule, it is rounded to the minute under and to the closest minute respectively.
The two in the screenshot have a start time at 14:40:41.

Image

I expect the start time to be rounded the same way: to the closest minute (seen with @maelysLeratRosso)

1. Upload [this timetable](https://github.com/user-attachments/files/19586997/timetable.2.json)
2. Witness the different rounding

Recette (SNCF)

Firefox

b42f78b

louisgreiner:
Close https://github.com/OpenRailAssociation/osrd/issues/11337

Not sure if I should refactor more the code nearby, I did the minimal changes, I'm open for refactoring if needed

### DIFF
--- a/front/src/modules/trainschedule/components/Timetable/TrainScheduleItem.tsx
+++ b/front/src/modules/trainschedule/components/Timetable/TrainScheduleItem.tsx
@@ -29,6 +29,7 @@ import {
 
 import TimetableItemActions from './TimetableItemActions';
 import type { TrainScheduleWithDetails } from './types';
+import { formatFullDate, formatTrainDuration, roundAndFormatToNearestMinute } from './utils';
 
 type TrainScheduleItemProps = {
   isInSelection: boolean;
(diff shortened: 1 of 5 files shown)


front/src/modules/trainschedule/components/Timetable/TrainScheduleItem.tsx
```diff
@@ -44,7 +44,7 @@ type TrainScheduleItemProps = {
 };
 
 const formatFullDate = (d: Date) => dayjs(d).format('D/MM/YYYY HH:mm:ss');
-const formatDateHours = (d: Date) =>
+export const formatDateHours = (d: Date) =>
```
SharglutDev: We should move this function either in `Timetable/utils.ts` or in `utils/date.ts`

SharglutDev (approved):
Lgtm and tested, just one suggestion

Akctarus (approved):
LGTM !

Maymanaf (approved):

PR #11404, opened 2025-04-10
core, editoast: rename v2 endpoints

Erashin:
Remove v2 from tsv2 endpoint names.

This rename does not impact ~the DB table object names nor~ (I'm a moron, those were the migrations haha) the error names, which still end with _v2. ~The subject is on the table though, talk this through between yourselves editoast maintainers (to be done in another PR though if the go-ahead is given).~ -> TBD in another PR.


woshilapin (approved):
Thank you, seems ok to go the extra-mile for the errors and tables to me.

flomonster (approved):
LGTM. I don't see a `_v2` table, I'm not sure I see the thing you're talking about. Anyway, we should also adapt the errors :+1:

bougue-pe (approved):

eckter (approved):
:heart:

PR #11405, opened 2025-04-10
core: simulation: use blocks from pathfinding

### LINKED ISSUE #10297: core: use the signaling system output from pathfinding everywhere in simu
remove recoverBlockPath (only deprecated now) and replace it

bougue-pe:
Previously, the block path was recomputed, possibly leading to inconsistent results.

fix https://github.com/OpenRailAssociation/osrd/issues/10297

🔍 please review by commit (the 3rd one was just used for test and is going to be squashed with the 4th)

Note: `ScheduleMetadataExtractor.kt` and `ScheduleMetadataExtractorV2.kt` are going to be merged in a separate PR.

### DIFF
--- a/core/src/main/java/fr/sncf/osrd/standalone_sim/ScheduleMetadataExtractor.kt
+++ b/core/src/main/java/fr/sncf/osrd/standalone_sim/ScheduleMetadataExtractor.kt
@@ -11,16 +11,12 @@ import fr.sncf.osrd.signaling.SignalingTrainState
 import fr.sncf.osrd.signaling.ZoneStatus
 import fr.sncf.osrd.sim_infra.api.*
 import fr.sncf.osrd.sim_infra.utils.routesOnBlock
-import fr.sncf.osrd.standalone_sim.result.ResultPosition
-import fr.sncf.osrd.standalone_sim.result.ResultSpeed
 import fr.sncf.osrd.standalone_sim.result.ResultTrain.RoutingRequirement
 import fr.sncf.osrd.standalone_sim.result.ResultTrain.RoutingZoneRequirement
 import fr.sncf.osrd.train.RollingStock
-import fr.sncf.osrd.utils.CurveSimplification
 import fr.sncf.osrd.utils.indexing.StaticIdxList
 import fr.sncf.osrd.utils.units.*
 import kotlin.collections.set
-import kotlin.math.abs
 
 // Reserve clear track with a margin for the reaction time of the driver
 const val CLOSED_SIGNAL_RESERVATION_MARGIN = 20.0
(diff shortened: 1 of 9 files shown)


core/src/main/kotlin/fr/sncf/osrd/api/stdcm/STDCMEndpoint.kt
```diff
@@ -188,6 +185,7 @@ class STDCMEndpoint(private val infraManager: InfraManager) : Take {
                 path.chunkPath,
                 infra,
                 path.routePath.toIdxList(),
+                path.blocks.ranges.map { it.edge }.toIdxList(), // TODO: filter duplicates?
```
bougue-pe: Any opinion is welcome on the need to filter duplicates.

eckter: Not necessary, duplicates are removed when building the result

eckter (approved):
Nice job :+1:  and nice commit history

core/src/main/java/fr/sncf/osrd/standalone_sim/ScheduleMetadataExtractor.kt
```diff
@@ -40,31 +33,36 @@ class PathOffsetBuilder(val startOffset: Distance) {
...
-//   with multiple signaling systems)
-fun deprecatedRecoverBlockPath(
-    simulator: SignalingSimulator,
+data class DetailedBlock(
+    val block: BlockId,
+    // the index of the route in the path
+    val routeIndex: Int,
+)
+
+fun buildDetailedBlockPath(
```
Khoyo: I'm not sure I like that name anymore since it becomes only the blocks and the route indexes. It should also probably be called directly by the routing requirements function, or even removed by directly building the route to block lookup table there?

bougue-pe: Moved in that direction in https://github.com/OpenRailAssociation/osrd/pull/11405/commits/ea7849a06ca56b8ca2614f5a259542f687eaa4ab (hope that matches your idea, but it's indeed better IMO).

core/src/main/kotlin/fr/sncf/osrd/api/stdcm/STDCMEndpoint.kt
bougue-pe: Comment removed in https://github.com/OpenRailAssociation/osrd/pull/11405/commits/41733876d497982e627307bc5cb033a0fd3e5b36

bougue-pe:
Rebased, only last 2 commits are new since reviews.
One final approval would be nice, but it's OK to merge as-is IMO (@Khoyo took a look at the main commit): probably gonna autosquash-merge it by tomorrow anyway.

Erashin (approved):
Works for me gj. You said you were gonna squash and merge, just in case you forgot, you have a `[wip]` commit that needs to disappear :)

PR #11407, opened 2025-04-10
Merge of the /infra/{infra_id}/speed_limit_tags and /speed_limit_tags endpoints

### LINKED ISSUE #9783: editoast, front: merge speed-limit tag endpoints
### Description and goal

Currently in front, when displaying a list of speed-limit tags, 2 calls are always done and merged:
1. `/infra/{infra_id}/speed_limit_tags` (all tags of the given infra)
2. `/speed_limit_tags` (all tags hard-configured in fallback hierarchy)

This issue is about doing the merging process in editoast instead of front:
- remove `/speed_limit_tags` endpoint from editoast
- improve `/infra/{infra_id}/speed_limit_tags` to have it send a merged list of all tags (config + given infra) in editoast
- only one call without merge everywhere in front

### Acceptance criteria

- [x] On all possible picks of speed-limit tags (see below), the choice is identical (all values from [config](https://github.com/OpenRailAssociation/osrd/blob/dev/assets/static_resources/speed_limit_tags.yml) + possible "exotic" extra-values from the infra itself, like after #7975)
- [x] When displaying a choice/list for speed-limit tags, only one api is called by the front:
  - [x] operational study: tag choice for train
  - [x] infra editor: tag choice for speed-limit
  - [x] infra editor: tag choice for map
  - [x] map explorer: tag choice for map
  - [x] any other case ? None found
  - [x] STDCM: probably not applicable as there is a shorter hardcoded list

SarrahElm:
Close #9783 

Merge of the /infra/{infra_id}/speed_limit_tags and /speed_limit_tags endpoints in editoast.
Deletion of the merge of the two speed limit tag lists in the front.


editoast/src/views/mod.rs
```diff
@@ -603,7 +600,6 @@ pub struct AppState {
     pub valkey: Arc<ValkeyClient>,
     pub infra_caches: Arc<DashMap<i64, InfraCache>>,
     pub map_layers: Arc<MapLayers>,
-    pub speed_limit_tag_ids: Arc<SpeedLimitTagIds>,
```
flomonster: We should keep this state to retrieve the list of speed limit tags.

editoast/src/models/infra/speed_limit_tags.rs
```diff
@@ -23,10 +24,16 @@ impl Infra {
...
             .load::<SpeedLimitTags>(conn.write().await.deref_mut())
             .await?;
-        Ok(speed_limits_tags)
+        let speed_limit_tags_hardcoded = SpeedLimitTagIds::load().0;
+        let mut all_speed_limit_tags: Vec<SpeedLimitTags> = speed_limit_tags_hardcoded
+            .into_iter()
+            .map(|el| SpeedLimitTags { tag: el })
+            .collect();
+        all_speed_limit_tags.append(&mut speed_limits_tags);
+        Ok(all_speed_limit_tags)
```
flomonster: This method should only return speed limit tags that are used in the infra.
The merge of the list should be performed by the endpoint.

flomonster: This might have duplicated values. We should use [HashSet](https://doc.rust-lang.org/std/collections/struct.HashSet.html) to ensure unicity.

flomonster (changes requested):
Thanks for your PR.
Ask questions if you need it :)

editoast/src/views/infra/mod.rs
```diff
@@ -569,9 +571,15 @@ async fn get_speed_limit_tags(
...
-    ))
+    let set_speed_limits_tags: HashSet<String> =
+        speed_limits_tags.into_iter().map(|el| (el.tag)).collect();
+    let speed_limit_tags_ids = SpeedLimitTagIds::load().0;
+    let set_speed_limit_tags_ids: HashSet<String> = speed_limit_tags_ids.into_iter().collect();
+    let all_speed_limit_tags: Vec<String> = set_speed_limit_tags_ids
+        .union(&set_speed_limits_tags)
+        .cloned()
+        .collect();
+    Ok(Json(all_speed_limit_tags))
```
flomonster: 1. We can change the endpoint signature to return a `Result<Json<HashSet<String>>>`. So we don't have to transform the `HashSet` into a `Vec`.
2. Instead of building two HashSet and calling `HashSet::union`, we can just build an iterator over both list and then collect as HashSet.
3. We don't need to call `SpeedLimitTagIds::load()`. We can retrieve this list from the state `State(builtin_tags): State<SpeedLimitTagIds>`

```suggestion
    let infra_tags = infra.get_speed_limit_tags(conn).await?;
    let union_tags = infra_tags
        .into_iter()
        .map(|el| (el.tag))
        .chain(builtin_tags.0)
        .collect();
    Ok(Json(union_tags))
```

flomonster (changes requested):

editoast/src/views/infra/mod.rs
leovalais: About 3., we can't directly do that. Either extract the tags from the state directly `State(AppState { speed_limit_tags_ids, .. }): State<AppState>` or implement `FromRef<AppState> for Arc<SpeedLimitTagIds>` to have the shorthand and then you can indeed `State(builtin_tags): State<Arc<SpeedLimitTagIds>>` (note the `Arc`). Whichever you prefer is fine by me :)

leovalais: We already impl FromRef for the DB pool in views/mod.rs, you can do something similar

flomonster (approved):

front/src/common/SpeedLimitByTagSelector/useStoreDataForSpeedLimitByTagSelector.ts
```diff
@@ -54,9 +49,7 @@ export const useStoreDataForSpeedLimitByTagSelector = ({
     }
   }, []);
 
-  const speedLimitsByTags = isStdcm
-    ? COMPOSITION_CODES
-    : compact(uniq(concat(speedLimitTags, speedLimitsTagsByInfraId)));
+  const speedLimitsByTags = isStdcm ? COMPOSITION_CODES : compact(uniq(speedLimitsTagsByInfraId));
```
SharglutDev: We don't need to `compact` anymore, we can only have `string` returned by the endpoint.
```suggestion
  const speedLimitsByTags = isStdcm ? COMPOSITION_CODES : uniq(speedLimitsTagsByInfraId);
```

Or the other files you changed, we can also remove `compact` but to do so you'll have to do the same logic as here when you call the endpoint : `data: speedLimitsTagsByInfraId = []`

SharglutDev (commented):
Thank you for this PR !

SharglutDev:
Lgtm and tested, good job !

flomonster:
@SharglutDev I think you forgot to approve the PR :sweat_smile:

SharglutDev (approved):
Woops 🫣

editoast/src/views/infra/mod.rs
```diff
@@ -1058,10 +1066,13 @@ pub mod tests {
...
 
-        let speed_limit_tags: Vec<String> =
+        let mut speed_limit_tags: Vec<String> =
             app.fetch(req).assert_status(StatusCode::OK).json_into();
 
-        assert_eq!(speed_limit_tags, vec!["test_tag"]);
+        let mut test_tags = builtin_tags.0.clone();
+        test_tags.push("test_tag".to_string());
+
+        assert_eq!(speed_limit_tags.sort(), test_tags.sort());
```
bougue-pe, line 1075: (late-remark) porting here this part of test would be nice: https://github.com/OpenRailAssociation/osrd/pull/11407/files#diff-eaf4e9096f2499f33e97891fed95b08b42ce30f1d2fc9090e3297e02022a8c62L57-L58

bougue-pe (commented):
✅ Manually checked on a local  `dev` version from today that it works on the cases described in issue (changed the issue to "Done").

PR #11412, opened 2025-04-10
front: fix trainschedules multideletion

### LINKED ISSUE #11408: Multiple deletion does not work in mission mode
### What happened?

In operational studies, mission mode, if I delete several unique trains at once (no mission) then no delete request is sent. So it appears as if my trains are deleted, but if I refresh, they come back

I expect my trains to be deleted

go in mission mode, in a scenario with 2 unique trains. Try to delete these trains with the boxes.

Recette (SNCF)

Google Chrome

de411c7

SharglutDev:
fix #11408

### DIFF
--- a/front/src/modules/trainschedule/components/Timetable/TimetableToolbar.tsx
+++ b/front/src/modules/trainschedule/components/Timetable/TimetableToolbar.tsx
@@ -152,13 +152,13 @@ const TimetableToolbar = ({
       try {
         let deletingTrainSchedulesPromise;
         let deletingPacedTrainsPromise;
-        if (editoastSelectedPacedTrainIds.length > 0) {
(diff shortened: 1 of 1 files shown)


front/src/modules/trainschedule/components/Timetable/TimetableToolbar.tsx
clarani, line 1: Hum there are some issues in the naming :thinking:

clarani (changes requested):

front/src/modules/trainschedule/components/Timetable/TimetableToolbar.tsx
SharglutDev: Fixed :)

clarani (approved):
LGTM

theocrsb (approved):
LGTM and tested

PR #11417, opened 2025-04-10
front: fix console warning at first load of a non empty timetable

### LINKED ISSUE #11257: Input warning in console
### What happened?

> [!NOTE]
> The related bug happens regardless of the paced train switch mode

When entering in a scenario with trains (doesn't happen if scenario is empty) for the first time, there is a warning in console : 

Image

No warning

1. Go to a scenario with trains
2. Open the console and notice the warning

Local

Firefox

.

SharglutDev:
When loading the timetable of a scenario for the first time, while the timetable items are loading, we first displayed the checkbox corresponding to an empty timetable and then we switched to the other one.
As the initial checkbox had checked property non defined, React displayed the "uncontrolled" warning because he considered that the same checkbox was switching from an undefined state to a defined one.
Fix it by merging the two checkboxes.

fix #11257

### DIFF
--- a/front/src/modules/trainschedule/components/Timetable/TimetableToolbar.tsx
+++ b/front/src/modules/trainschedule/components/Timetable/TimetableToolbar.tsx
@@ -275,27 +275,22 @@ const TimetableToolbar = ({
             'with-details': isInSelection,
           })}
         >
-          {timetableItems.length === 0 ? (
-            <div className="train-count">
-              <Checkbox small readOnly label={t('timetable.noTrain')} />
-            </div>
-          ) : (
-            <div className="train-count">
-              <Checkbox
-                label={computedItemLabel()}
-                small
-                checked={
-                  selectedTimetableItemIds.length === timetableItemsWithDetails.length &&
-                  selectedTimetableItemIds.length > 0
-                }
-                isIndeterminate={
-                  selectedTimetableItemIds.length !== timetableItemsWithDetails.length &&
-                  selectedTimetableItemIds.length > 0
-                }
-                onChange={() => toggleAllTrainsSelecton()}
-              />
-            </div>
-          )}
+          <div className="train-count">
+            <Checkbox
+              label={timetableItems.length > 0 ? computedItemLabel() : t('timetable.noTrain')}
+              small
+              readOnly={timetableItems.length === 0}
+              checked={
+                selectedTimetableItemIds.length === timetableItemsWithDetails.length &&
+                selectedTimetableItemIds.length > 0
+              }
+              isIndeterminate={
+                selectedTimetableItemIds.length !== timetableItemsWithDetails.length &&
+                selectedTimetableItemIds.length > 0
+              }
+              onChange={() => toggleAllTrainsSelecton()}
+            />
+          </div>
 
           {timetableItems.length > 0 && (
             <div>


front/src/modules/trainschedule/components/Timetable/TimetableToolbar.tsx
```diff
@@ -152,13 +152,13 @@ const TimetableToolbar = ({
...
+          deletingTrainSchedulesPromise = deleteTrainSchedules({
             body: { ids: editoastSelectedTrainScheduleIds },
           }).unwrap();
         }
         if (editoastSelectedPacedTrainIds.length > 0) {
-          deletingTrainSchedulesPromise = deletePacedTrains({
+          deletingPacedTrainsPromise = deletePacedTrains({
             body: { ids: editoastSelectedPacedTrainIds },
           }).unwrap();
         }
```
theocrsb: Does this change correspond to this PR?

theocrsb (approved):
For the rest, PR works

front/src/modules/trainschedule/components/Timetable/TimetableToolbar.tsx
SharglutDev: Oh right, not sure what happened 🤔 Fixed

front/src/modules/trainschedule/components/Timetable/TimetableToolbar.tsx
clarani, line 1: Hum why not use the condition only to change the label ? :thinking:
Maybe we could have only 1 checkbox

```
<div className="train-count">
    <Checkbox
      label={timetableItems.length > 0 ? computedItemLabel(): t('timetable.noTrain')}
      small
      checked={
        selectedTimetableItemIds.length === timetableItemsWithDetails.length &&
        selectedTimetableItemIds.length > 0
      }
      isIndeterminate={
        selectedTimetableItemIds.length !== timetableItemsWithDetails.length &&
        selectedTimetableItemIds.length > 0
      }
      onChange={() => toggleAllTrainsSelecton()}
    />
</div>
```

SharglutDev: Done (added the `readOnly` prop as well) and changed the commit description

Caracol3:
Shouldn't we use `disabled` instead of `readonly` for the checkbox? According to the MDN documentation, the readonly attribute is not appropriate for checkboxes as it has no effect on them and it will still be clickable:

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly.

SharglutDev:
> Shouldn't we use `disabled` instead of `readonly` for the checkbox? According to the MDN documentation, the readonly attribute is not appropriate for checkboxes as it has no effect on them and it will still be clickable:
> 
> https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly.

This was discussed with @thibautsailly when implementing this new interface. He used the readOnly style of the checkbox for the mockup so @theocrsb implemented it like that.

You're right for the convention but if you check the `Checkbox` component on osrd-ui, you'll see that we actually don't use the `readOnly` attribute on it when passing this props :)

Caracol3:
LGTM

Caracol3 (approved):
LGTM

thibautsailly:
The MDN document says "The difference between [disabled](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/disabled) and readonly is that read-only controls can still function and are still focusable, whereas disabled controls can not receive focus and are not submitted with the form and generally do not function as controls until they are enabled."
"Still function" doesn't mean being interactive and clickable, it means that the value stored for the field will be submitted with the rest of the form. Therefore, I don't understand why a checkbox can't be read-only.

Example.
A first group of radio buttons asks what you want to drink: water, soda or tea with the meal you're ordering.
A second group of checkbox lets you choose if you want some ice, a straw, and/or a lid for your cup.

If you're ordering tea, you will get a lid to prevent burns: selecting tea will select the lid and will not let you remove it because the vendor doesn't want to be responsible for a skin burn. In this case, the readonly checked checkbox has a valid function of letting you know you'll get a lid on your tea cup.

[See IBM's design system reference](https://carbondesignsystem.com/patterns/read-only-states-pattern/)

clarani (approved):
LGTM :white_check_mark:

PR #11424, opened 2025-04-11
core: stdcm: restore full stdcm tests

eckter:
Reintroduce a test file that was removed with the tsv1 drop. 

There's also some cleanup, some methods weren't used anymore


Erashin (approved):
Gj!

PR #11426, opened 2025-04-11
editoast: remove unused variant from `core::stdcm::Response`

hamz2a:


leovalais:
Can't core return this variant?

flomonster (approved):
LGTM

flomonster:
> Can't core return this variant?

Using grep I don't see occurrences of this error 🤷

PR #11427, opened 2025-04-11
front stdcm warningbox improvements

### LINKED ISSUE #11278: Improve messages displayed in WarningBox
### What happened?

# Description

#### Improvements must be done regarding the warning box in STDCM.
Follow-up to ticket #9707 

# AC

- [x] Warnings should be displayed as different sections of the same message bloc, in this priority order:
1. Origin and destination timing preferences incompatibility
2. Invalid information
3. Missing informations

- [x] Each section should order their warning in the reading order of the form: consist, origin, intermediate OPs, destination.

- [x] Invalid fields must be included in the WarningBox.

- [x] If a field was previously filled and is then cleared, a warning should appear on focus out (onBlur).

# Tests E2E
- [x] Add invalid fields to the tests for this user scenario: https://github.com/osrd-project/osrd-tests/issues/106

_No response_

1. Go to STDCM
2. Fill in the origin and destination with the same OP
3. Fill in all consist fields except one
4. Lower the value of one of the consist fields so that it becomes invalid
5. Click the button to launch the simulation

Recette (SNCF), Dev (SNCF)

Firefox, Google Chrome

872260c

Caracol3:
close #11278

### DIFF
--- a/front/public/locales/en/stdcm.json
+++ b/front/public/locales/en/stdcm.json
@@ -12,6 +12,7 @@
         "negative": "Must be positive.",
         "range": "The max speed must be between {{low}} and {{high}}km/h"
       },
+      "missingValue": "Missing value",
       "totalLength": {
         "negative": "Must be positive.",
         "range": "The total length must be between {{low}} and {{high}}m"
@@ -175,8 +176,13 @@
   "simulationSheet": "Simulation sheet",
   "spaceTimeGraphic": "Space-Time graph",
   "stdcmErrors": {
-    "bothPointAreScheduled": "The calculation cannot take into account both an origin and a destination schedule. You have to choose one or the other:",
     "infraNotLoaded": "Infrastructure is loading",
+    "invalidFields": {
+      "maxSpeed": "the maximum speed of the consist",
+      "totalLength": "the total length of the consist",
+      "totalMass": "the total mass of the consist"
+    },
+    "invalidInformations": "Invalid information for calculation:",
     "missingFields": {
       "destination": "the destination",
       "maxSpeed": "the maximum speed of the consist",
@@ -185,16 +191,22 @@
       "timeOrigin": "the departure time at the origin",
       "totalLength": "the total length of the consist",
       "totalMass": "the total mass of the consist",
-      "tractionEngine": "the traction engine"
+      "tractionEngine": "the traction engine",
+      "vias": "At least one intermediate OP"
     },
     "missingInformations": "Missing informations for calculation:",
-    "missingLocation": "Missing location",
     "noPaths": "Incompatibility with other train paths.",
     "noResults": "No path found",
-    "noScheduledPoint": "The calculation requires either the origin or destination schedule.",
     "pathfindingFailed": "No path have been found for these waypoints.",
     "requestFailed": "Calculation error for last minute train path",
-    "zeroLengthPath": "The path cannot have a null length."
+    "routeErrors": {
+      "bothPointAreScheduled": "The calculation cannot take into account both an origin and a destination schedule. You have to choose one or the other:",
+      "global": "Please check the errors related to the following route:",
+      "noScheduledPoint": "The calculation requires either the origin or destination schedule.",
+      "viaStopDurationMissing": "Please specify the stop duration for the intermediate OP.",
+      "viaStopDurationTooShort": "The driver switch stop time is usually at least 3 minutes.",
+      "zeroLengthPath": "Origin and destination cannot be the same."
+    }
   },
   "trainPath": {
     "addVia": "Add intermediate OP",
@@ -220,7 +232,6 @@
     "time": "Time",
     "tolerance": "Tolerance",
     "type": "Type",
-    "vias": "Intermediate OP",
-    "warningMinStopTime": "The driver switch stop time is usually at least 3 minutes."
+    "vias": "Intermediate OP"
   }
 }
--- a/front/public/locales/fr/stdcm.json
+++ b/front/public/locales/fr/stdcm.json
@@ -12,6 +12,7 @@
         "negative": "Doit être positif.",
         "range": "La vitesse max. doit être comprise entre {{low}} et {{high}}km/h"
       },
+      "missingValue": "Valeur manquante",
       "totalLength": {
         "negative": "Doit être positif.",
         "range": "La longueur totale doit être comprise entre {{low}} et {{high}}m"
@@ -175,8 +176,13 @@
   "simulationSheet": "Fiche de simulation",
   "spaceTimeGraphic": "Graphique Espace-Temps",
   "stdcmErrors": {
-    "bothPointAreScheduled": "Le calcul ne peut pas prendre en compte à la fois un horaire d'origine et un horaire de destination. Il faut choisir l'un ou l'autre :",
     "infraNotLoaded": "Infrastructure en cours de chargement",
+    "invalidFields": {
+      "maxSpeed": "la vitesse maximale du convoi",
+      "totalLength": "la longueur totale du convoi",
+      "totalMass": "le poids total du convoi"
+    },
+    "invalidInformations": "Informations invalides pour le calcul :",
     "missingFields": {
       "destination": "la destination",
       "maxSpeed": "la vitesse maximale du convoi",
@@ -185,16 +191,22 @@
       "timeOrigin": "l'heure de départ à l'origine",
       "totalLength": "la longueur totale du convoi",
       "totalMass": "le poids total du convoi",
-      "tractionEngine": "l'engin de traction"
+      "tractionEngine": "l'engin de traction",
+      "vias": "au moins un PR intermédiaire"
     },
     "missingInformations": "Informations manquantes pour le calcul :",
-    "missingLocation": "Localisation manquante",
     "noPaths": "Incompatibilité avec d'autres sillons.",
     "noResults": "Aucun sillon trouvé",
-    "noScheduledPoint": "Le calcul a besoin de l'horaire d'origine ou de celui de la destination.",
     "pathfindingFailed": "Aucun chemin n'a été trouvé pour ces points de jalonnement.",
     "requestFailed": "Erreur de calcul Sillon de dernière minute",
-    "zeroLengthPath": "Le chemin ne peut pas avoir de longueur nulle."
+    "routeErrors": {
+      "bothPointAreScheduled": "Le calcul ne peut pas prendre en compte à la fois un horaire d'origine et un horaire de destination. Il faut choisir l'un ou l'autre :",
+      "global": "Veuillez vérifier les erreurs suivantes liées au trajet :",
+      "noScheduledPoint": "Le calcul a besoin de l'horaire d'origine ou de celui de la destination.",
+      "viaStopDurationMissing": "Le temps d'arrêt pour le PR intermédiaire.",
+      "viaStopDurationTooShort": "Le temps d'arrêt d'une relève conducteur est d'au moins 3 minutes.",
+      "zeroLengthPath": "Origine et destination ne peuvent pas être identiques."
+    }
   },
   "trainPath": {
     "addVia": "Ajouter un PR intermédiaire",
@@ -220,7 +232,6 @@
     "time": "Heure",
     "tolerance": "Tolérance",
     "type": "Type",
-    "vias": "PR intermédiaire",
-    "warningMinStopTime": "Le temps d'arrêt d'une relève conducteur est d'au moins 3 minutes."
+    "vias": "PR intermédiaire"
   }
 }
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
@@ -1,9 +1,10 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 
 import { Input, ComboBox, useDefaultComboBox } from '@osrd-project/ui-core';
 import { useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
 
+import { consistErrorFields } from 'applications/stdcm/consts';
 import useConsistFieldStatus from 'applications/stdcm/hooks/useConsistFieldStatus';
 import useStdcmTowedRollingStock from 'applications/stdcm/hooks/useStdcmTowedRollingStock';
 import type { ConsistErrors } from 'applications/stdcm/types';
@@ -48,9 +49,16 @@ const ConsistCardTitle = ({
 export type StdcmConsistProps = {
   isDebugMode: boolean;
   disabled?: boolean;
+  consistErrors: ConsistErrors;
+  setConsistErrors: React.Dispatch<React.SetStateAction<ConsistErrors>>;
 };
 
-const StdcmConsist = ({ isDebugMode, disabled = false }: StdcmConsistProps) => {
+const StdcmConsist = ({
+  isDebugMode,
+  disabled = false,
+  consistErrors,
+  setConsistErrors,
+}: StdcmConsistProps) => {
   const { t } = useTranslation('stdcm');
 
   const speedLimitByTag = useSelector(getStdcmSpeedLimitByTag);
@@ -68,12 +76,6 @@ const StdcmConsist = ({ isDebugMode, disabled = false }: StdcmConsistProps) => {
     mass: true,
     length: true,
     speed: true,
-  });
-
-  const [consistErrors, setConsistErrors] = useState<ConsistErrors>({
-    totalMass: { display: false },
-    totalLength: { display: false },
-    maxSpeed: { display: false },
   });
 
   const {
@@ -102,6 +104,13 @@ const StdcmConsist = ({ isDebugMode, disabled = false }: StdcmConsistProps) => {
   const lengthFieldStatus = createFieldStatus('totalLength');
   const speedFieldStatus = createFieldStatus('maxSpeed');
 
+  const getMissingFieldMessage = (value?: number): string | null => {
+    if (!value) {
+      return t('consist.errors.missingValue');
+    }
+    return null;
+  };
+
   const { filteredRollingStockList: rollingStocks } = useFilterRollingStock({ isStdcm: true });
 
   const { filteredTowedRollingStockList: towedRollingStocks } = useFilterTowedRollingStock({
@@ -113,9 +122,7 @@ const StdcmConsist = ({ isDebugMode, disabled = false }: StdcmConsistProps) => {
     return secondPart ? `${rs.name} - ${secondPart}` : rs.name;
   };
 
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-call
   const rollingStockComboBoxDefaultProps = useDefaultComboBox(rollingStocks, getLabel);
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-call
   const towedRollingStockComboBoxDefaultProps = useDefaultComboBox(
     towedRollingStocks,
     (trs: TowedRollingStock) => trs.name
@@ -141,38 +148,66 @@ const StdcmConsist = ({ isDebugMode, disabled = false }: StdcmConsistProps) => {
     setStatusMessagesVisible((prevState) => ({ ...prevState, [key]: false }));
   };
 
-  useEffect(() => {
-    const totalMassError = validateTotalMass({
-      tractionEngineMass: rollingStock?.mass,
-      towedMass: towedRollingStock?.mass,
-      totalMass,
-    });
-
-    const totalLengthError = validateTotalLength({
-      tractionEngineLength: rollingStock?.length,
-      towedLength: towedRollingStock?.length,
-      totalLength,
-    });
-
-    const maxSpeedError = validateMaxSpeed(maxSpeed, rollingStock?.max_speed);
-
-    // Hide the tooltip if the user just changed the value for an invalid one (should be displayed only on focus out)
-    // but we also want to hide it as soon as the user corrects it
-    setConsistErrors((prevErrors) => ({
-      totalMass: {
-        message: totalMassError,
-        display: prevErrors.totalMass.display && !!totalMassError,
-      },
-      totalLength: {
-        message: totalLengthError,
-        display: prevErrors.totalLength.display && !!totalLengthError,
-      },
-      maxSpeed: {
-        message: maxSpeedError,
-        display: prevErrors.maxSpeed.display && !!maxSpeedError,
+  const handleBlurError = (field: keyof ConsistErrors, error?: string) => {
+    setConsistErrors((prev) => ({
+      ...prev,
+      [field]: {
+        message: error,
+        display: !!error,
+        type: error === t('consist.errors.missingValue') ? 'missing' : 'invalid',
       },
     }));
-  }, [rollingStock, towedRollingStock, totalMass, totalLength, maxSpeed]);
+  };
+
+  const totalMassError = useMemo(
+    () =>
+      getMissingFieldMessage(totalMass) ??
+      validateTotalMass({
+        tractionEngineMass: rollingStock?.mass,
+        towedMass: towedRollingStock?.mass,
+        totalMass,
+      }),
+    [totalMass]
+  );
+
+  const totalLengthError = useMemo(
+    () =>
+      getMissingFieldMessage(totalLength) ??
+      validateTotalLength({
+        tractionEngineLength: rollingStock?.length,
+        towedLength: towedRollingStock?.length,
+        totalLength,
+      }),
+    [totalLength]
+  );
+
+  const maxSpeedError = useMemo(
+    () => getMissingFieldMessage(maxSpeed) ?? validateMaxSpeed(maxSpeed, rollingStock?.max_speed),
+    [maxSpeed]
+  );
+
+  useEffect(() => {
+    const errors = {
+      totalMass: totalMassError,
+      totalLength: totalLengthError,
+      maxSpeed: maxSpeedError,
+    };
+
+    consistErrorFields.forEach((field) => {
+      setConsistErrors((prev) => ({
+        ...prev,
+        [field]: {
+          ...prev[field],
+          display:
+            prev[field].display &&
+            (prev[field].type === 'missing'
+              ? // Hide tooltip if the error was a missing one and is fixed regardless if the fied is still invalid
+                !!(errors[field] === t('consist.errors.missingValue'))
+              : !!errors[field]),
+        },
+      }));
+    });
+  }, [totalMassError, totalLengthError, maxSpeedError]);
 
   return (
     <StdcmCard
@@ -218,13 +253,10 @@ const StdcmConsist = ({ isDebugMode, disabled = false }: StdcmConsistProps) => {
           type="number"
           min={0}
           value={totalMass ?? ''}
-          onChange={onTotalMassChange}
-          onBlur={() => {
-            setConsistErrors({
-              ...consistErrors,
-              totalMass: { ...consistErrors.totalMass, display: !!consistErrors.totalMass.message },
-            });
+          onChange={(e) => {
+            onTotalMassChange(e);
           }}
+          onBlur={() => handleBlurError('totalMass', totalMassError)}
           disabled={disabled}
           statusWithMessage={massFieldStatus}
           onCloseStatusMessage={() => handleCloseStatusMessage('mass')}
@@ -237,16 +269,10 @@ const StdcmConsist = ({ isDebugMode, disabled = false }: StdcmConsistProps) => {
           type="number"
           min={0}
           value={totalLength ?? ''}
-          onChange={onTotalLengthChange}
-          onBlur={() => {
-            setConsistErrors({
-              ...consistErrors,
-              totalLength: {
-                ...consistErrors.totalLength,
-                display: !!consistErrors.totalLength.message,
-              },
-            });
+          onChange={(e) => {
+            onTotalLengthChange(e);
           }}
+          onBlur={() => handleBlurError('totalLength', totalLengthError)}
           disabled={disabled}
           statusWithMessage={lengthFieldStatus}
           onCloseStatusMessage={() => handleCloseStatusMessage('length')}
@@ -269,13 +295,10 @@ const StdcmConsist = ({ isDebugMode, disabled = false }: StdcmConsistProps) => {
           type="number"
           min={0}
           value={maxSpeed ?? ''}
-          onChange={onMaxSpeedChange}
-          onBlur={() => {
-            setConsistErrors({
-              ...consistErrors,
-              maxSpeed: { ...consistErrors.maxSpeed, display: !!consistErrors.maxSpeed.message },
-            });
+          onChange={(e) => {
+            onMaxSpeedChange(e);
           }}
+          onBlur={() => handleBlurError('maxSpeed', maxSpeedError)}
           disabled={disabled}
           statusWithMessage={speedFieldStatus}
           onCloseStatusMessage={() => handleCloseStatusMessage('speed')}
--- a/front/src/applications/stdcm/components/StdcmForm/StdcmDestination.tsx
+++ b/front/src/applications/stdcm/components/StdcmForm/StdcmDestination.tsx
@@ -21,6 +21,7 @@ const StdcmDestination = ({ disabled = false, onItineraryChange }: StdcmItinerar
       title={<img src={DestinationIcon} alt="destination" className="stdcm-destination-icon" />}
       disabled={disabled}
       className="extremity"
+      testId="stdcm-card-destination"
     >
       <StdcmOperationalPoint
         location={destination.location}
--- a/front/src/applications/stdcm/components/StdcmWarningBox.tsx
+++ b/front/src/applications/stdcm/components/StdcmWarningBox.tsx
@@ -7,10 +7,7 @@ import { useTranslation } from 'react-i18next';
 
 import { StdcmConfigErrorTypes, type StdcmConfigErrors } from '../types';
 
-const SHORT_TEXT_ERRORS = [
-  StdcmConfigErrorTypes.INFRA_NOT_LOADED,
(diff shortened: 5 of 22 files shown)


front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
```diff
@@ -28,86 +30,142 @@ const checkStdcmConfigErrors = ({
...
-    origin.arrivalType === ArrivalTimeTypes.RESPECT_DESTINATION_SCHEDULE;
+      if (
+        (stopType === 'driverSwitch' || stopType === 'serviceStop') &&
+        (stopFor === undefined || stopFor === null)
+      ) {
+        routeErrors.push(StdcmConfigErrorTypes.VIA_STOP_DURATION_MISSING);
+      }
 
-  const isDestinationASAP = destination.arrivalType === ArrivalTimeTypes.ASAP;
+      if (stopType === 'driverSwitch' && stopFor !== undefined && stopFor !== null && stopFor < 3) {
```
SharglutDev: Same at l.124 (for duration issue)
```suggestion
      if (stopType === 'driverSwitch' && !isNil(stopFor) && stopFor.total('minute') < 3) {
```

front/src/applications/stdcm/components/StdcmWarningBox.tsx
SharglutDev, line 1: When a consist field is empty, it triggers both invalid / missing line in the warning box

Capture d’écran 2025-04-18 à 11 16 21

SharglutDev, line 1: If I have both errors "same origin / destination" and "precise time", if I clear the origin or destination, both errors are removed from the warning box when it should only be the one regarding "same origin / destination"

front/src/applications/stdcm/components/StdcmForm/StdcmDestination.tsx
SharglutDev, line 1: To check if the error was there before but if i set a path with origin / 1 via / destination and I clear the destination CI, I have an error in console in `useStaticPathFinding.ts`

front/public/locales/en/stdcm.json
```diff
@@ -127,6 +128,12 @@
   "stdcmErrors": {
     "bothPointAreScheduled": "The calculation cannot take into account both an origin and a destination schedule. You have to choose one or the other:",
     "infraNotLoaded": "Infrastructure is loading",
+    "invalidInformations": "Invalid information for calculation:",
+    "invalidFields": {
+      "totalMass": "the total mass of the consist is invalid",
+      "totalLength": "the total length of the consist is invalid",
+      "maxSpeed": "the maximum speed of the consist is invalid"
```
SharglutDev: These values are in a section `Invalid information for calculation`, is it necessary to repeat that they are invalids ? (same in french)
```suggestion
      "totalMass": "the total mass",
      "totalLength": "the total length",
      "maxSpeed": "the maximum speed"
```

front/public/locales/fr/stdcm.json
```diff
@@ -135,16 +142,19 @@
...
     },
     "missingInformations": "Informations manquantes pour le calcul :",
-    "missingLocation": "Localisation manquante",
     "noPaths": "Incompatibilité avec d'autres sillons.",
     "noResults": "Aucun sillon trouvé",
     "noScheduledPoint": "Le calcul a besoin de l'horaire d'origine ou de celui de la destination.",
     "pathfindingFailed": "Aucun chemin n'a été trouvé pour ces points de jalonnement.",
     "requestFailed": "Erreur de calcul Sillon de dernière minute",
-    "zeroLengthPath": "Le chemin ne peut pas avoir de longueur nulle."
+    "routeErrors": "Veuillez vérifier les erreurs liées au trajet suivantes :",
```
SharglutDev: nit : less confusing like this maybe ?
```suggestion
    "routeErrors": "Veuillez vérifier les erreurs suivantes liées au trajet :",
```

front/src/applications/stdcm/hooks/useStdcmErrors.ts
SharglutDev, line 1: Looks like this hook is not used ?

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
```diff
@@ -48,9 +48,20 @@ const ConsistCardTitle = ({
...
-const StdcmConsist = ({ isDebugMode, disabled = false }: StdcmConsistProps) => {
+const StdcmConsist = ({
+  isDebugMode,
+  disabled = false,
+  consistErrors = {
+    totalMass: { display: false },
+    totalLength: { display: false },
+    maxSpeed: { display: false },
+  },
+  setConsistErrors = () => {},
```
SharglutDev: Their types in props should be changed, they are always defined
```suggestion
  consistErrors,
  setConsistErrors,
```

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
```diff
@@ -102,6 +107,13 @@ const StdcmConsist = ({ isDebugMode, disabled = false }: StdcmConsistProps) => {
   const lengthFieldStatus = createFieldStatus('totalLength');
   const speedFieldStatus = createFieldStatus('maxSpeed');
 
+  const getMissingFieldMessage = (value?: number): string | undefined => {
+    if (value === undefined || value === null) {
```
SharglutDev: Value can't be null. 0 is not accepted in the inputs (always replaced by undefined), so we should be good with this
```suggestion
    if (!value) {
```

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
```diff
@@ -102,6 +107,13 @@ const StdcmConsist = ({ isDebugMode, disabled = false }: StdcmConsistProps) => {
...
+      return t('consist.errors.missingValue');
+    }
+    return undefined;
```
SharglutDev: Just a personal opinion but I find it better to return `null` in this case than `undefined`. It's that we will be "missing" something but we want to return something that isn't a string.
```suggestion
    return null;
```

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
SharglutDev, line 1: If I enter an invalid value in one of the input and then clear the input, the tooltip error (missing value) pops instantly, not when focus out the input (same for the reversed, first an empty value and then the invalid one)

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
```diff
@@ -142,37 +154,52 @@ const StdcmConsist = ({ isDebugMode, disabled = false }: StdcmConsistProps) => {
...
+      });
 
-    const maxSpeedError = validateMaxSpeed(maxSpeed, rollingStock?.max_speed);
+    const maxSpeedError =
+      getMissingFieldMessage(maxSpeed) ?? validateMaxSpeed(maxSpeed, rollingStock?.max_speed);
 
     // Hide the tooltip if the user just changed the value for an invalid one (should be displayed only on focus out)
     // but we also want to hide it as soon as the user corrects it
-    setConsistErrors((prevErrors) => ({
+    setConsistErrors({
```
SharglutDev: I think we can keep using `prevErrors` here and remove all consistErrors from the depths (you'll have to pass the true type of `setConsistErrors` in props to make it work)
```suggestion
    setConsistErrors((prevErrors) => ({...
```

front/src/applications/stdcm/components/StdcmWarningBox.tsx
SharglutDev, line 1: Sometimes when the warning box is up, changing the path now longer shows the pathfinding status

front/src/applications/stdcm/utils/getInvalidFields.ts
```diff
@@ -0,0 +1,30 @@
+import type { ConsistErrors, InvalidFields } from '../types';
+
+const getInvalidFields = (consistErrors: ConsistErrors): InvalidFields[] => {
+  const invalidFields: InvalidFields[] = [];
+
+  if (consistErrors.totalMass.display && consistErrors.totalMass.message) {
+    invalidFields.push({
+      fieldName: 'totalMass',
+      message: consistErrors.totalMass.message,
```
SharglutDev: Maybe I missed something but do we ever use the `message` property ?

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
```diff
@@ -28,86 +30,142 @@ const checkStdcmConfigErrors = ({
...
 
-  if (pathfindingStatus && pathfindingStatus === 'failure') {
-    return { errorType: StdcmConfigErrorTypes.PATHFINDING_FAILED };
-  }
-
-  const isOriginRespectDestinationSchedule =
-    origin.arrivalType === ArrivalTimeTypes.RESPECT_DESTINATION_SCHEDULE;
+      if (
+        (stopType === 'driverSwitch' || stopType === 'serviceStop') &&
+        (stopFor === undefined || stopFor === null)
```
SharglutDev:
```suggestion
        isNil(stopFor)
```

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
```diff
@@ -28,86 +30,142 @@ const checkStdcmConfigErrors = ({
...
+      ) {
+        routeErrors.push(StdcmConfigErrorTypes.VIA_STOP_DURATION_MISSING);
```
SharglutDev, line 117: Capture d’écran 2025-04-18 à 15 49 32

This is a little repetitive

front/src/applications/stdcm/components/StdcmWarningBox.tsx
```diff
@@ -69,6 +69,45 @@ const StdcmWarningBox = ({
...
+      {!hasInvalidFields && !hasMissingFields && !hasRouteErrors && (
+        <p
+          className={cx('mb-0', {
+            'text-center': SHORT_TEXT_ERRORS.includes(errorType),
+            'text-justify': !SHORT_TEXT_ERRORS.includes(errorType),
+          })}
+        >
+          {t(`stdcmErrors.${errorType}`)}
+        </p>
+      )}
```
SharglutDev, line 81: What case is it if we have neither of these 3 error types ?

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
```diff
@@ -28,86 +30,142 @@ const checkStdcmConfigErrors = ({
...
+  }
+
+  const activeErrorTypes = [finalMissingFields, finalInvalidFields, finalRouteErrors].filter(
+    Boolean
+  ).length;
+
+  let errorType: StdcmConfigErrorTypes;
+
+  if (activeErrorTypes > 1) {
+    errorType = StdcmConfigErrorTypes.MULTIPLE_ERRORS;
```
SharglutDev, line 161: We don't have a translation for this error type ? Is it used somewhere ?

front/src/applications/stdcm/types.ts
```diff
@@ -197,14 +200,23 @@ export type MissingFields =
...
 
+export type InvalidFields = {
+  fieldName: 'totalMass' | 'totalLength' | 'maxSpeed';
+  message: string;
+};
+
 export type StdcmConfigErrors = {
   errorType: StdcmConfigErrorTypes;
   errorDetails?: {
+    routeErrors?: StdcmConfigErrorTypes[] | undefined;
```
SharglutDev: Same l.219
```suggestion
    routeErrors?: StdcmConfigErrorTypes[];
```

front/tests/pages/stdcm/consist-section.ts
```diff
@@ -119,6 +119,36 @@ class ConsistSection {
...
+    await handleAndVerifyInput(this.lengthField, value);
+  }
+
+  async setInvalidMaxSpeed(value: string): Promise<void> {
+    await handleAndVerifyInput(this.maxSpeedField, value);
+  }
+
+  async clearLength(): Promise<void> {
+    await this.lengthField.fill('');
+    await expect(this.lengthField).toHaveValue('');
```
SharglutDev: Can't we do that ?
```suggestion
    await handleAndVerifyInput(this.lengthField, '');
```

front/tests/pages/stdcm/destination-section.ts
```diff
@@ -132,6 +132,13 @@ class DestinationSection extends STDCMPage {
     await expect(this.dynamicDestinationCh).toHaveValue(chValue);
     await expect(this.destinationArrival).toHaveValue(arrivalType);
   }
+
+  async clearDestination(): Promise<void> {
+    const clearButton = this.destinationCard.locator('.clear-icon');
+    await clearButton.click({ force: true });
```
SharglutDev: Why do we need the `force: true` ?

front/tests/assets/constants/missing-fields.ts
```diff
@@ -22,3 +22,12 @@ export type MissingFields = (typeof ALL_MISSING_FIELDS_KEY)[number];
 
 export const getFieldsLabel = (fields: readonly MissingFields[], translations: StdcmTranslations) =>
   fields.map((field) => translations.stdcmErrors.missingFields[field]);
+
+export const ALL_INVALID_FIELDS_KEY = ['totalMass', 'totalLength', 'maxSpeed'] as const;
+
+export type InvalidFields = (typeof ALL_INVALID_FIELDS_KEY)[number];
+
+export const getInvalidFieldsLabel = (
```
SharglutDev: Looks like it's not used

SharglutDev (commented):
Thanks for the PR ! We can see the end of the tunnel 💪

front/src/applications/stdcm/hooks/useStdcmErrors.ts
Caracol3: Good catch, i forget to erase it.

front/public/locales/en/stdcm.json
Caracol3: done

front/public/locales/fr/stdcm.json
Caracol3: Indeed.
Changed

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
Caracol3: done

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
Caracol3: done

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
Caracol3: I agree, changed.

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
Caracol3: done

front/src/applications/stdcm/components/StdcmForm/StdcmDestination.tsx
Caracol3: I tried this use case on dev and indeed we have this problem too.

front/src/applications/stdcm/components/StdcmWarningBox.tsx
Caracol3: Now we only have the `infraNotLoaded`warning, but we can have other ones in the future.

front/src/applications/stdcm/types.ts
Caracol3: done

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
Caracol3: done

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
Caracol3: Changed

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
Caracol3: Better this way !  Changed

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
Caracol3: `MULTIPLE_ERRORS` is used to identify cases where multiple types of errors are present (missing fields, invalid fields, route errors).
There is maybe a better way to do so but since the UI already renders each category with its specific heading, we don’t display `MULTIPLE_ERRORS` directly, so no translation is needed for me.

front/src/applications/stdcm/utils/getInvalidFields.ts
Caracol3: Indeed `message` is not used after all.
I removed it

front/tests/assets/constants/missing-fields.ts
Caracol3: Deleted

front/tests/pages/stdcm/consist-section.ts
Caracol3: done

front/tests/pages/stdcm/destination-section.ts
Caracol3: It was hard to trigger the click of the clear-icon on the combobox so i tried this at the beginning.
Now the logic is a bit different so it's maybe useless. 
I will try to remove it to see how it works without.

front/public/locales/fr/stdcm.json
RomainValls, line 1: Spotted a translation error that doesn't come from your PR, but maybe you could rectify it : Please wait is translated in "Veuillez patientez", should be "patienter". Thanks 😄

front/src/applications/stdcm/components/StdcmWarningBox.tsx
RomainValls, line 1: Nit: When we add values to fields that had no values, the error in the warning box are removed. 
If we delete the values again, the warning box doesn't appear, unless we click on "Get simulation". 

It is ok for me to keep this behavior, but since errors are removed dynamically, maybe they can re appear dynamically too.

RomainValls (approved):
LGTM (there's a nit comment that can be skipped, and a minor translation error I spotted, GJ !)

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
```diff
@@ -28,86 +31,139 @@ const checkStdcmConfigErrors = ({
...
-  if (shouldCheckMandatoryFields) {
-    if (!rollingStockID) missingFields.push('tractionEngine');
-    if (isNil(totalMass)) missingFields.push('totalMass');
-    if (isNil(totalLength)) missingFields.push('totalLength');
-    if (isNil(maxSpeed)) missingFields.push('maxSpeed');
-    if (!origin.location) {
-      missingFields.push('origin');
+  missingFields.push(
+    ...filterMissingFields({
+      missingFields: prevFormErros?.errorDetails?.missingFields,
```
RomainValls: prevFormErros should be prevFormErrors :)

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
Caracol3: Good catch !
Done

front/src/applications/stdcm/components/StdcmWarningBox.tsx
Caracol3: done

Caracol3: The issue is now fixed when both errors are present and you change the arrival type from "precise time" to "as soon as possible": the "same origin/destination" error correctly remains in the warning box.
However, in the reverse case (e.g., when you remove the destination), both errors disappear — and that's expected, because displaying the "precise time" error requires both origin and destination to be defined.

Caracol3: Fixed

Caracol3: Discussed with @thibautsailly.
This is the behavior we want:
we want the error to be removed as soon as the field is corrected, but we do not want to reintroduce the error unless the user clicks the launch button again.

We assume that if a field becomes empty again after the error was shown, it was a deliberate action by the user, and they are aware that they need to fix it.
We don't want to be intrusive after an error has been fixed. If needed, the error will only reappear when the user explicitly tries to launch a simulation.

front/src/applications/stdcm/utils/checkStdcmConfigErrors.ts
Caracol3: done

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
SharglutDev: Now when we correct a missing/invalid field, the popup doesn't disappear when the error disappears (e.g. : clear a field, focus out > the popup show, I enter something in the input > the popup doesn't hide until I focused out)

front/src/applications/stdcm/components/StdcmWarningBox.tsx
SharglutDev: Looks like we don't have any warning when the pathfinding fails.

image

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
SharglutDev: Not sure if I wasn't supposed to test again after your commit but I still have the issue : 

https://github.com/user-attachments/assets/41332446-824c-47bf-81c4-d69be20e21ed

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
```diff
@@ -96,12 +98,19 @@ const StdcmConsist = ({ isDebugMode, disabled = false }: StdcmConsistProps) => {
       statusMessagesVisible,
       rollingStock,
       towedRollingStock
-    );
+    ) as StatusWithMessage | undefined;
```
SharglutDev: Seems unnecessary
```suggestion
    );
```

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
SharglutDev, line 1: Seems like we can remove both `eslint-disable`

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
```diff
@@ -141,38 +150,63 @@ const StdcmConsist = ({ isDebugMode, disabled = false }: StdcmConsistProps) => {
     setStatusMessagesVisible((prevState) => ({ ...prevState, [key]: false }));
   };
 
-  useEffect(() => {
-    const totalMassError = validateTotalMass({
+  const totalMassError =
```
SharglutDev: We should put these 3 `const` in `useMemo` (we can probably omit `rollingStock` and `towedRollingStock` from the depths since it will recompute mass/length/vmax each time anyway)

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
SharglutDev, line 1: We still have a wrong behavior if the value is invalid. The tooltip and warning box disappears even though the value is still invalid (coming from the new `clearMissingError` in `onChange` inputs) :


https://github.com/user-attachments/assets/7da69ce6-ec75-4f17-a55b-56df156a10de


Moreover, I think there is a bit of repetitive code with these 3 `useEffects` and handling the hide of the popup in the `useEffect` and in the `onChange` as well.

Proposal : 
1. Remove `clearMissingErrorIfNotEmpty`
2. Merge the 3 `useEffect` (like it was before but a bit refactored) and handle the clear inside : 
```
  useEffect(() => {
    const fields: (keyof ConsistErrors)[] = ['totalMass', 'totalLength', 'maxSpeed'];
    const errors = {
      totalMass: totalMassError,
      totalLength: totalLengthError,
      maxSpeed: maxSpeedError,
    };

    fields.forEach((field) => {
      setConsistErrors((prev) => ({
        ...prev,
        [field]: {
          ...prev[field],
          display:
            prev[field].display &&
            (prev[field].type === 'missing'
              ? // Hide tooltip if the error was a missing one and is fixed regardless if the fied is still invalid
                !!(errors[field] === t('consist.errors.missingValue'))
              : !!errors[field]),
        },
      }));
    });
  }, [totalMassError, totalLengthError, maxSpeedError]);
```

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
```diff
@@ -218,12 +252,19 @@ const StdcmConsist = ({ isDebugMode, disabled = false }: StdcmConsistProps) => {
...
-              totalMass: { ...consistErrors.totalMass, display: !!consistErrors.totalMass.message },
-            });
+            setConsistErrors((prev) => ({
+              ...prev,
+              totalMass: {
+                message: totalMassError,
+                display: !!totalMassError,
+                type: totalMassError === t('consist.errors.missingValue') ? 'missing' : 'invalid',
+              },
+            }));
```
SharglutDev: The 3 `onBlur` are pretty similar, we could create a function for that : 
```
const handleBlurError = (field: keyof ConsistErrors, error?: string) => {
  setConsistErrors((prev) => ({
    ...prev,
    [field]: {
      message: error,
      display: !!error,
      type: error === t('consist.errors.missingValue') ? 'missing' : 'invalid',
    },
  }));
};

```
```suggestion
          onBlur={() => handleBlurError('totalMass', totalMassError)}
```

front/src/applications/stdcm/types.ts
```diff
@@ -75,19 +75,16 @@ export type StdcmResultsOperationalPoint = {
   stopRequested: boolean;
 };
 
+export type FieldError = {
+  message?: string;
+  display: boolean;
+  type?: 'missing' | 'invalid';
```
SharglutDev: This field is always defined I think ?
```suggestion
  type: 'missing' | 'invalid';
```

front/src/applications/stdcm/utils/getInvalidFields.ts
```diff
@@ -0,0 +1,17 @@
+import type { ConsistErrors, InvalidFields } from '../types';
+
+const getInvalidFields = (consistErrors: ConsistErrors): InvalidFields[] => {
+  const InvalidKeys: (keyof ConsistErrors)[] = ['totalMass', 'totalLength', 'maxSpeed'];
```
SharglutDev: If you implemented the proposal to clean `useEffect`, we have this line repeated at several places. We could extract it in a `const.ts` (the `const` should start with a lowercase)

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
Caracol3: It's now fixed

Caracol3: done

front/src/applications/stdcm/types.ts
Caracol3: done

Caracol3: done

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
Caracol3: Thanks, it's clearer now !

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
Caracol3: done

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
Caracol3: done

Caracol3: Indeed, when fixing the `missingValue` bug i totally put aside the behavior of the `invalidField`.
Fixed now

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
Caracol3: done

front/src/applications/stdcm/utils/getInvalidFields.ts
Caracol3: done

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
```diff
@@ -141,38 +148,67 @@ const StdcmConsist = ({ isDebugMode, disabled = false }: StdcmConsistProps) => {
...
+
+    return {
+      totalMassError: massError,
+      totalLengthError: lengthError,
+      maxSpeedError: speedError,
+    };
+  }, [totalMass, totalLength, maxSpeed]);
+
+  useEffect(() => {
+    const fields = consistErrorFields;
```
SharglutDev: I think we can use `consistErrorFields` directly on the `forEach`
```suggestion
```

front/src/applications/stdcm/utils/getInvalidFields.ts
```diff
@@ -0,0 +1,18 @@
+import { consistErrorFields } from '../consts';
+import type { ConsistErrors, InvalidFields } from '../types';
+
+const getInvalidFields = (consistErrors: ConsistErrors): InvalidFields[] => {
+  const InvalidKeys = consistErrorFields;
```
SharglutDev: Angular leftover with the uppercase for the `const` ? 😅 

We could also remove this line and use `consistErrorFields` on the `reduce`

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
SharglutDev: Not sure how React work for that but maybe it would be better to keep the separation you made for these 3 `const` and put each one of them in `useMemo` (sorry I wasn't very precise).
If we don't I'm afraid each time one of these fields is updated, all of the field errors will be recomputed as well resulting to maybe launching 3 times the other `useEffect` for `consistErrors` when it should be launched only once ?

SharglutDev (commented):
Tested and everything works as expected, good job ! We are close to the end :)
Just a couple of small comments and fixing css linting + e2e

front/src/applications/stdcm/utils/getInvalidFields.ts
Caracol3: done

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
Caracol3: You're right.
Done

front/src/applications/stdcm/components/StdcmForm/StdcmConsist.tsx
Caracol3: done

SharglutDev:
Lgtm and tested, nice work ! I'll leave @Maymanaf put his approve and let you clean your commits so I can test one more time before approving :)

front/tests/pages/stdcm/destination-section.ts
```diff
@@ -93,7 +93,10 @@ class DestinationSection extends STDCMPage {
     // Verify default values
     await expect(this.dynamicDestinationCh).toHaveValue(chValue);
     await expect(this.destinationArrival).toHaveValue(arrivalType.default);
-    await expect(this.warningBox).toContainText(translations.stdcmErrors.noScheduledPoint);
+    await this.page.getByTestId('launch-simulation-button').click();
```
Maymanaf:
```suggestion
    await this.launchSimulationButton.click();
```

front/tests/pages/stdcm/destination-section.ts
```diff
@@ -129,6 +132,13 @@ class DestinationSection extends STDCMPage {
     await expect(this.dynamicDestinationCh).toHaveValue(chValue);
     await expect(this.destinationArrival).toHaveValue(arrivalType);
   }
+
+  async clearDestination(): Promise<void> {
+    const clearButton = this.destinationCard.locator('.clear-icon');
```
Maymanaf: Could you move the locator declaration to the constructor

front/tests/pages/stdcm/consist-section.ts
```diff
@@ -117,6 +117,35 @@ class ConsistSection {
...
+    await handleAndVerifyInput(this.tonnageField, value);
+  }
+
+  async setValidLength(value: string): Promise<void> {
+    await handleAndVerifyInput(this.lengthField, value);
+  }
+
+  async setValidMaxSpeed(value: string): Promise<void> {
+    await handleAndVerifyInput(this.maxSpeedField, value);
+  }
```
Maymanaf: You can directly use `setLength`, `setTonnage`, and `setMaxSpeed`
The step comments in the test make it clear.

front/tests/pages/stdcm/destination-section.ts
Caracol3: done

front/tests/pages/stdcm/destination-section.ts
Caracol3: done

front/tests/pages/stdcm/consist-section.ts
Caracol3: You're right, done

Maymanaf (approved):
LGTM for e2e tests ✅

SharglutDev (approved):
Lgtm and tested, great job ! If you could just squash your commits since they lost a bit of their independence after fixing up the old ones :) The one for tests can remain.

PR #11438, opened 2025-04-15
ci: remove codecov

Khoyo:
It's been broken for months now, apparently nobody cares for it, and building with instrumentation costs ci time.


.github/workflows/build.yml
```diff
@@ -825,20 +773,10 @@ jobs:
         run: |
           docker run --name=front-test --net=host -v $PWD/output:/output \
             ${{ fromJSON(needs.build.outputs.stable_tags).front-tests }} \
-            sh -c "npm run test-coverage && cp -R coverage /output"
+            sh -c "CI=true npm run test"
```
emersion: Seems like `CI=true` is new, what does it do?

.github/workflows/build.yml
```diff
@@ -825,20 +773,10 @@ jobs:
         run: |
           docker run --name=front-test --net=host -v $PWD/output:/output \
```
emersion: We should be able to get rid of the `/output` bind mount.

.github/workflows/build.yml
```diff
@@ -825,20 +773,10 @@ jobs:
...
             ${{ fromJSON(needs.build.outputs.stable_tags).front-tests }} \
-            sh -c "npm run test-coverage && cp -R coverage /output"
```
emersion, line 828: We could get rid of the `test-coverage` script in `package.json`. Potentially drop the `@vitest/coverage-v8` dependency?

.github/workflows/build.yml
Khoyo: I'm not fully certain, but it at least prevents the test runner from watching for file changes after running tests...

gateway/Dockerfile
```diff
@@ -39,7 +39,7 @@ RUN rustup component add llvm-tools && \
     cargo install grcov
 COPY --from=planner /gateway/recipe.json recipe.json
 
-ENV RUSTFLAGS="-Cinstrument-coverage"
+ENV RUSTFLAGS=""
```
flomonster: Since it's empty, should we remove this flag?

flomonster (approved):
LGTM

.github/workflows/build.yml
emersion: Yeah, it's an equivalent to `vitest run`. I'd prefer to be more explicit and either use `npx vitest run`, or `npm run test run` (run run!).

front/package.json
```diff
@@ -113,7 +113,6 @@
     "@typescript-eslint/eslint-plugin": "^8.26.1",
     "@typescript-eslint/parser": "^8.29.1",
     "@vitejs/plugin-react-swc": "^3.8.1",
-    "@vitest/coverage-v8": "^3.1.1",
```
emersion, line 116: I think `package-lock.json` needs to be synced (can be done with `npm install`).

.github/workflows/build.yml
emersion: Gentle ping, has this comment been forgotten, or do you disagree?

emersion (approved):
Looks good apart from one remaining comment: https://github.com/OpenRailAssociation/osrd/pull/11438/files#r2050533907

Thanks a lot for doing this!

flomonster (approved):
LGTM, thanks for this change :pray:

.github/workflows/build.yml
Khoyo: It was forgotten ^^'

PR #11439, opened 2025-04-15
Prepare Model Retrieve error refactoring

leovalais:
- **editoast: add Retrieve::retrieve_real for real error forwarding**
- **editoast: allow deperecate retrieve**
- **editoast: remove InternalError dependency for Model Exist**
- **editoast: remove InternalError dependency for retrieve_batch_unchecked**
- **editoast: remove InteralError dependency for RetrieveBatch**

### DIFF
--- a/editoast/src/models/timetable.rs
+++ b/editoast/src/models/timetable.rs
@@ -10,6 +10,7 @@ use editoast_derive::Model;
 use futures_util::stream::TryStreamExt;
 use std::ops::DerefMut;
 
+use crate::error::EditoastError;
 use crate::error::Result;
 use crate::models::Retrieve;
 use crate::models::prelude::*;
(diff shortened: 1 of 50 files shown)


editoast/src/models/timetable.rs
```diff
@@ -117,8 +118,8 @@ pub struct TimetableWithTrains {
     pub paced_train_ids: Vec<i64>,
 }
 
-impl Retrieve<i64> for TimetableWithTrains {
-    async fn retrieve(conn: &mut DbConnection, timetable_id: i64) -> Result<Option<Self>> {
+impl TimetableWithTrains {
+    pub async fn retrieve(conn: &mut DbConnection, timetable_id: i64) -> Result<Option<Self>> {
```
woshilapin: Might be hard to track these when we will remove all the `.retrieve()` calls. Does it makes sense to add a `#[deprecated]` on it (and on the `retrieve_or_fail` below?

woshilapin (approved):
Thanks for the work to make this change incremental.

editoast/src/models/timetable.rs
leovalais: Removed InternalError for these functions instead 7ce289c22

leovalais: Good catch though

PR #11486, opened 2025-04-17
editoast: update `pathfinding` to handle `OsrdError` from `core` service

hamz2a:
The `core` service returns `OsrdError`, so this update adds the correct `OsrdError` type and updates pathfinding to handle it properly.
It’s part of a larger refactor of core error handling. If this approach works well, I will apply the same changes to `core::simulation` (which requires more refactoring and a new structure for the core response, which is why it’s not included in this PR).

### DIFF
--- a/editoast/src/core/mod.rs
+++ b/editoast/src/core/mod.rs
@@ -305,6 +305,21 @@ impl Display for StandardCoreError {
     }
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
+pub struct RawError {
+    #[serde(rename = "type")]
+    pub error_type: String,
+    pub message: String,
+    pub context: HashMap<String, serde_json::Value>,
+    pub cause: ErrorCause,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
+pub enum ErrorCause {
+    Internal,
+    User,
+}
+
 #[cfg(test)]
 mod tests {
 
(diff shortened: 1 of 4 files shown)


woshilapin (approved):
If I understand correctly, this is a way to remove `InternalError` from `core`'s module (or at least a part of that task). But since it's not entirely removed anyway, it's still converted into an `InternalError` anyway, but much later, not in `core` anymore. If I understood correctly, it seems fine, thank you.

editoast/src/core/mod.rs
```diff
@@ -305,6 +305,21 @@ impl Display for StandardCoreError {
     }
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
+pub struct OsrdError {
```
leovalais: Let's not call it OsrdError since it's not the right format (cause field). What about `core::RawError` or something like that?

editoast/src/core/mod.rs
```diff
@@ -305,6 +305,21 @@ impl Display for StandardCoreError {
...
+    #[serde(rename = "type")]
+    pub error_type: String,
+    pub message: String,
+    pub context: HashMap<String, Value>,
```
leovalais:
```suggestion
    pub context: HashMap<String, serde_json::Value>,
```

leovalais (approved):
Besides the naming LGTM

editoast/src/core/mod.rs
hamz2a: Done.

editoast/src/core/mod.rs
hamz2a: Done.

PR #11459, opened 2025-04-15
Refacto paced trains and train schedules modules

younesschrifi:
After discussions with @leovalais concerning the`timetable`, `paced trains` and `train schedule` modules, it was agreed that we needed to improve the editoast tree structure of these modules, as many structures and functions were used in timetable, paced trains and train schedule. 

We decided to put everything under: timetable.

##  Tree structure

here's the tree structure:    
       - views
          - mod.rs
          - timetable.rs
          - timetable
              - paced_train.rs
              - train_schedule.rs
              - simulation.rs   

- enum `SimulationSummaryResult` moves from `mod.rs` to `simulation.rs`
- functions `train_simulation_batch`, `consist_train_simulation_batch` and `build_simulation_request` move from  `train_schedule.rs` to `simulation.rs`
- struct `InfraIdQueryParam` moves from `views/mod.rs` to `views/infra.rs`
- struct `ListId` moves from `views/mod.rs` to `views/paced_train.rs` and `views/train_schedule.rs`

PR #11202, opened 2025-03-20
editoast: refactor simulation and projection hash functions

hamz2a:
https://github.com/OpenRailAssociation/osrd/pull/11192#issuecomment-2737156150

PR #11501, opened 2025-04-17
editoast: refactor `core::simulation`

hamz2a:
> [!WARNING]
> must be merged after https://github.com/OpenRailAssociation/osrd/pull/11486.

The commits improve clarity and structure. The views were updated to use the correct response type instead of `core::SimulationResponse`. `SimulationResponse` and `SimulationRequest` were renamed to `Response` and `Request` for consistency. The `InternalError` type was removed from `Response` to simplify handling. Lastly, `SimulationSummaryResult` was moved to `views::core::simulation` for better organization.

> [!TIP]
> Please review the PR one commit at a time for easier understanding and clearer context.

### DIFF
--- a/editoast/src/core/simulation.rs
+++ b/editoast/src/core/simulation.rs
@@ -26,10 +26,10 @@ use serde::Deserialize;
 use serde::Serialize;
 use utoipa::ToSchema;
 
+use super::RawError;
 use super::pathfinding::TrackRange;
 use crate::core::AsCoreRequest;
 use crate::core::Json;
-use crate::error::InternalError;
 
 editoast_common::schemas! {
     CompleteReportTrain,
@@ -495,7 +495,7 @@ pub enum Response {
         electrical_profiles: ElectricalProfiles,
     },
     SimulationFailed {
-        core_error: InternalError,
+        core_error: RawError,
     },
 }
 
--- a/editoast/src/core/simulation.rs
+++ b/editoast/src/core/simulation.rs
@@ -30,7 +30,6 @@ use super::pathfinding::TrackRange;
 use crate::core::AsCoreRequest;
 use crate::core::Json;
 use crate::error::InternalError;
-use crate::views::path::pathfinding::PathfindingFailure;
 
 editoast_common::schemas! {
     CompleteReportTrain,
@@ -40,7 +39,6 @@ editoast_common::schemas! {
     RoutingZoneRequirement,
     ZoneUpdate,
     ReportTrain,
-    SimulationResponse,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, Derivative, ToSchema)]
@@ -480,7 +478,7 @@ pub struct SimulationRequest {
     pub electrical_profile_set_id: Option<i64>,
 }
 
-#[derive(Serialize, Deserialize, PartialEq, Clone, Debug, ToSchema)]
+#[derive(Serialize, Deserialize, PartialEq, Clone, Debug)]
 #[serde(tag = "status", rename_all = "snake_case")]
 // We accepted the difference of memory size taken by variants
 // Since there is only on success and others are error cases
@@ -491,16 +489,10 @@ pub enum SimulationResponse {
         base: ReportTrain,
         /// Simulation that takes into account the regularity margins
         provisional: ReportTrain,
-        #[schema(inline)]
         /// User-selected simulation: can be base or provisional
         final_output: CompleteReportTrain,
-        #[schema(inline)]
         mrsp: SpeedLimitProperties,
-        #[schema(inline)]
         electrical_profiles: ElectricalProfiles,
-    },
-    PathfindingFailed {
-        pathfinding_failed: PathfindingFailure,
     },
     SimulationFailed {
         core_error: InternalError,
@@ -513,21 +505,6 @@ impl AsCoreRequest<Json<SimulationResponse>> for SimulationRequest {
 
     fn infra_id(&self) -> Option<i64> {
         Some(self.infra)
-    }
-}
-
-impl SimulationResponse {
-    pub fn simulation_run_time(&self) -> Option<u64> {
-        if let SimulationResponse::Success { provisional, .. } = self {
-            Some(
-                *provisional
-                    .times
-                    .last()
-                    .expect("core error: empty simulation result"),
-            )
-        } else {
-            None
-        }
     }
 }
 
--- a/editoast/src/core/simulation.rs
+++ b/editoast/src/core/simulation.rs
@@ -461,7 +461,7 @@ pub struct SimulationPowerRestrictionRange {
 
 #[derive(Debug, Serialize, Derivative)]
 #[derivative(Hash)]
-pub struct SimulationRequest {
+pub struct Request {
     pub infra: i64,
     pub expected_version: i64,
     pub path: SimulationPath,
@@ -499,7 +499,7 @@ pub enum Response {
     },
 }
 
-impl AsCoreRequest<Json<Response>> for SimulationRequest {
+impl AsCoreRequest<Json<Response>> for Request {
     const METHOD: reqwest::Method = reqwest::Method::POST;
     const URL_PATH: &'static str = "/standalone_simulation";
 
--- a/editoast/src/core/simulation.rs
+++ b/editoast/src/core/simulation.rs
@@ -483,7 +483,7 @@ pub struct SimulationRequest {
 // We accepted the difference of memory size taken by variants
 // Since there is only on success and others are error cases
 #[allow(clippy::large_enum_variant)]
-pub enum SimulationResponse {
+pub enum Response {
     Success {
         /// Simulation without any regularity margins
         base: ReportTrain,
@@ -499,7 +499,7 @@ pub enum SimulationResponse {
     },
 }
 
-impl AsCoreRequest<Json<SimulationResponse>> for SimulationRequest {
+impl AsCoreRequest<Json<Response>> for SimulationRequest {
     const METHOD: reqwest::Method = reqwest::Method::POST;
     const URL_PATH: &'static str = "/standalone_simulation";
 
--- a/editoast/src/views/stdcm_logs.rs
+++ b/editoast/src/views/stdcm_logs.rs
@@ -175,7 +175,6 @@ mod tests {
     use crate::core::simulation::CompleteReportTrain;
     use crate::core::simulation::ElectricalProfiles;
     use crate::core::simulation::ReportTrain;
-    use crate::core::simulation::SimulationResponse;
     use crate::core::simulation::SpeedLimitProperties;
     use crate::models::fixtures::create_fast_rolling_stock;
     use crate::models::fixtures::create_small_infra;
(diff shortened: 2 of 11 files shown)


editoast/src/views/core/simulation.rs
woshilapin, line 1: In https://github.com/OpenRailAssociation/osrd/pull/11459, all simulation related functionalities were moved to `views::timetable::simulation` if I'm not wrong, there might be conflicts with your work.

editoast/src/views/core/simulation.rs
```diff
@@ -0,0 +1,92 @@
...
+    fn default() -> Self {
+        Self::Success {
+            base: Default::default(),
+            provisional: Default::default(),
+            final_output: Default::default(),
+            mrsp: Default::default(),
+            electrical_profiles: Default::default(),
+        }
+    }
+}
```
woshilapin: Because I see this, I'm wondering what does a `Default` implementation semantically means? To me, it seems weird to implement `Default` for such an object. Can we remove it (I'm really asking as a question, not a "we should do that")? If not, where does it fail and why? I'm totally ok for doing that in another PR since it's not entirely related.

woshilapin (approved):
Thank you so much for the precise description of the PR and the separation in commit, it was a breeze to review. And thank you for following up on these refacto.

editoast/src/views/core/simulation.rs
hamz2a: We use default [here](https://github.com/OpenRailAssociation/osrd/blob/hai/editoast-refactor-core-simulation/editoast/src/views/train_schedule.rs#L417).

editoast/src/views/train_schedule.rs
```diff
@@ -334,7 +334,7 @@ pub async fn train_simulation_batch(
     train_schedules: &[models::TrainSchedule],
     infra: &Infra,
     electrical_profile_set_id: Option<i64>,
-) -> Result<Vec<(SimulationResponse, PathfindingResult)>> {
+) -> Result<Vec<(Response, PathfindingResult)>> {
```
leovalais: I'm not sure it's right to convert into a response this early. I'd say it's up to the endpoints to decide when they want to turn calculations into api responses

editoast/src/views/train_schedule.rs
leovalais, line 1: could a more qualified path be used like `core::simulation::Request` in order to provide context? "Request" can refer to many things in views.

editoast/src/core/simulation.rs
```diff
@@ -513,7 +513,7 @@ pub enum Response {
         electrical_profiles: ElectricalProfiles,
     },
     SimulationFailed {
-        core_error: InternalError,
+        core_error: RawError,
```
leovalais, line 498: 🥳 Are we ready to move `core` in its own crate now???

leovalais (changes requested):
Thanks for this (hopefully last) refactoring work in `core`! A few comments:

* Changes from commit 2 seem to leak into commit 1, which likely doesn't build (not tested)
* I'd wait for #11459 to be merged as your last commit is part of this PR, which performs a more complete reorganization
* Likewise with that PR, your changes in `core/simulation.rs` should be moved in the future `timetable/simulation.rs`
* Great to have a dedicated response type for the views! But can you use qualified paths for these? `Response` can mean many things in the views (core's response, our view response, axum's response?)
* I think *core*'s simulation response is converted into an API response too early. When we request a `train_simulation_batch`, we expect simulation results, not an API response. Should these two types diverge, we'd loose information for any post-processing.
* You commit messages are too long, the end leak into the description part 😁

editoast/src/views/core/simulation.rs
leovalais: Yuck! It was there before so it's fine for this PR but I'd rather have a vec initialized with None rather than having API responses implement Default...

editoast/src/views/train_schedule.rs
hamz2a: Done.

editoast/src/core/simulation.rs
leovalais: no we're not https://github.com/OpenRailAssociation/osrd/pull/11202#issuecomment-2831016396

editoast/src/views/train_schedule.rs
hamz2a: Thanks for the feedback!
In this PR, I only renamed the type, so I think it would be better to handle that change in a separate PR.

hamz2a:
> * Changes from commit 2 seem to leak into commit 1, which likely doesn't build (not tested)

I'm not sure I fully understood the first comment — I tested each commit individually and confirmed that they all build.

> * You commit messages are too long, the end leak into the description part 😁

I've also rewritten the commit messages to keep them short.

editoast/src/core/simulation.rs
hamz2a: Can [this change](https://github.com/OpenRailAssociation/osrd/pull/11592) help solve the issue ?

editoast/src/views/core/simulation.rs
woshilapin: Should be solved by https://github.com/OpenRailAssociation/osrd/pull/11614.

editoast/src/core/simulation.rs
leovalais: Not sure if it solves it, but it's a necessary step, thanks!

editoast/src/views/train_schedule.rs
leovalais: I disagree. This PR is not just renaming things. It also introduces a semantic shift for some types. And just like a name change should be propagated everywhere, we should ensure that the new semantic fits well everywhere it's involved, or else the change is not complete imo.

editoast/src/views/core/simulation.rs
woshilapin: The other PR is in the merge queue. I believe that you will only need to remove this `Default` implementation.

editoast/src/views/timetable/simulation.rs
```diff
@@ -39,13 +39,14 @@ use utoipa::ToSchema;
...
-    SimulationSummaryResult,
+    SummaryResult,
 }
 
 #[derive(Debug, Serialize, ToSchema)]
 #[cfg_attr(test, derive(PartialEq, serde::Deserialize))]
 #[serde(tag = "status", rename_all = "snake_case")]
-pub enum SimulationSummaryResult {
+#[schema(as = SimulationSummaryResult)]
+pub enum SummaryResult {
```
leovalais: What about just `Summary`? `Result` is ambiguous in the Rust context + it's obviously the result of some calculation.

editoast/src/views/core/simulation.rs
leovalais: Unresolving

Imo there should be only two `simulation.rs` files, one in `core`, one in `timetable`. In the `views`, to make everything as clear as possible, I suggest using `core::simulation::Response` and `simulation::Response` to represent the API response. Wdyt?

leovalais:
> I'm not sure I fully understood the first comment — I tested each commit individually and confirmed that they all build.

Oh ok I've been probably bamboozled by the unqualified paths

editoast/src/views/timetable/simulation.rs
hamz2a: `SummaryResponse` ?

hamz2a:
> Imo there should be only two simulation.rs files, one in core, one in timetable. In the views, to make everything as clear as possible, I suggest using core::simulation::Response and simulation::Response to represent the API response. Wdyt?

I’m working on the rebase and cleaning up the commits right now. I agree, I’ll do that just after.

editoast/src/views/core/simulation.rs
hamz2a: I’m working on the rebase and cleaning up the commits right now. I agree, I’ll do that just after.

editoast/src/views/timetable/simulation.rs
leovalais: Why the suffix?

hamz2a: To be more clear, do you not like it?

editoast/src/views/train_schedule.rs
hamz2a: we are using `views::timetable::simulation::Response` instead of `core::simulation::Response` in the [`train_simulation_batch` function](https://github.com/OpenRailAssociation/osrd/blob/hai/editoast-refactor-core-simulation/editoast/src/views/timetable/simulation.rs#L203), because in the `consist_train_simulation_batch` function, when a pathfinding failure occurs, we return the [`simulation::Response::PathfindingFailed` variant](https://github.com/OpenRailAssociation/osrd/blob/hai/editoast-refactor-core-simulation/editoast/src/views/timetable/simulation.rs#L308):

```rust
PathfindingResult::Failure(pathfinding_failed) => {
    simulation_results[index] = Some(simulation::Response::PathfindingFailed {
        pathfinding_failed: pathfinding_failed.clone(),
    });
    continue;
}
```

However, the `PathfindingFailed` variant **does not exist** in [`core::simulation::Response`](https://github.com/OpenRailAssociation/osrd/blob/hai/editoast-refactor-core-simulation/editoast/src/core/simulation.rs#L486). Therefore, to handle this properly, we need to use `views::timetable::simulation::Response`, which includes that variant.

editoast/src/views/timetable/simulation.rs
leovalais: I thought it was obvious that it's a response, especially since this term is very heavily used. But let's keep what you prefer

editoast/src/views/stdcm_logs.rs
```diff
@@ -280,8 +280,8 @@ mod tests {
         }
     }
 
-    fn simulation_response() -> SimulationResponse {
-        SimulationResponse::Success {
+    fn simulation_response() -> simulation::Response {
+        simulation::Response::Success {
```
```suggestion
    fn simulation_response() -> core::simulation::Response {
        core::simulation::Response::Success {
```

editoast/src/views/train_schedule.rs
leovalais: First, thanks for the detailed explanation with the links! This is problematic because it means that these functions confuse core responses with API responses, and that it wasn't just a typing issue. I've been thinking that the whole `train_simulation_batch` and friends should be rewritten to have a more structured API instead of having a bunch of free functions floating around. They encapsulate a lot of behavior and their signature are overly complex. Now with this issue with error typing and semantics (the same goes for the PathfindingResult also returned) I think that rewriting becomes necessary. I'll try to experiment that this PI, I already have a few ideas... 😶‍🌫️ 

As for this comment, resolving it properly involves a lot of work, so let's merge this PR and deal with that later. You can continue working on `crate core` as the rewriting I'm talking about is a `views`-only task.

leovalais (approved):
LGTM. I see an issue with the views simulation API which I'll try to tackle at least partially later, but I think that for now the priority is to have `crate core` complete. Thanks for this PR!

Nit about commit names: https://english.stackexchange.com/questions/100296/which-preposition-to-use-with-rename

editoast/src/views/stdcm_logs.rs
hamz2a: Done.

editoast/src/views/train_schedule.rs
hamz2a: Okay, sounds good! I’ll keep working on core. Looking forward to seeing where that goes.

PR #11506, opened 2025-04-18
Editoast bump rust edition to 2024

### LINKED ISSUE #10903: Move to rust edition 2024
### Description and goal

Rust edition 2024 is out 🎉 .
We should migrate our code base to the new edition 🚀  (https://doc.rust-lang.org/edition-guide/editions/transitioning-an-existing-project-to-a-new-edition.html)

### Acceptance criteria

- [ ] editoast
- [x] osrdyne
- [x] gateway
- [x] bump flake nix

flomonster:
close #10903 

Review commits separately:
1. Bump dependencies
1. Migrate to edition 2024 and fix build
    - Require explicit typing for `AsyncCommands` due to [never type fallback](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/never-type-fallback.html).
    - `env::set_var` is now an unsafe function
    - Fix fga! macro lifetime issue
1. Run cargo fmt

> [!IMPORTANT]
> This migration requires rustc version 1.85 or higher


editoast/fga/src/lib.rs
```diff
@@ -446,10 +446,7 @@ macro_rules! fga {
 
     // fga!(Group:"my_friends"#member) => userset syntax "group:my_friends#member"
     ($ty:ident : $id:literal # $relation:ident) => {
-        {
-            use $crate::model::Relation as _;
-            $ty::$relation().userset(&fga!($ty:$id))
-        }
+        $ty::$relation().userset(&fga!($ty:$id))
```
woshilapin: This change seems weird. `.userset()` is a function of `Relation`'s trait, so adding the `use` inside the macro seems a security for the macro to always work (in file that already import that traits but also in traits that don't import that trait).

editoast/src/valkey_utils.rs
```diff
@@ -11,16 +14,13 @@ use deadpool_redis::redis::RedisError;
...
-use futures::future;
+use deadpool_redis::redis::aio::ConnectionLike;
+use deadpool_redis::redis::cmd;
 use futures::FutureExt;
-use serde::de::DeserializeOwned;
+use futures::future;
 use serde::Serialize;
-use tracing::{debug, span, trace, Level};
+use serde::de::DeserializeOwned;
+use tracing::{Level, debug, span, trace};
```
woshilapin: nitpick: Since it's part of the diff, can you flatten that `use`, please?

editoast/src/views/layers.rs
```diff
@@ -301,7 +301,9 @@ mod tests {
             ("http://localhost:8090/test", "http://localhost:8090/test/"),
             ("http://localhost:8090/test/", "http://localhost:8090/test/"),
         ] {
-            std::env::set_var("ROOT_URL", root_url);
+            unsafe {
```
woshilapin, line 308: I'd say it would be nice to have a best practice of adding a comment when we have a `unsafe` in code (the same than with `#[allow(...)]`). And actually, it's not fine in this case because we usually run tests in parallel threads, and this is exactly the case for non-being safe. Sure, we could say that this is the only test that actually do that (but who will detect if another test is added?). Another possible solution would be to rely on something like `cargo-nextest` which parallelize tests with processes (not threads).

It seems that we already had the problem anyway so I'm fine with a comment justifying that it's the only test like this.

woshilapin (approved):
I'm really happy to see this new Rust formatter, not giving up on macros 😄

editoast/src/valkey_utils.rs
flomonster: Yep, do we have a command line to flatten everything?

editoast/src/views/layers.rs
flomonster: I will add a comment. About the parallelization issue, we already had this issue before this PR.
I think the fix should be to stop retrieving the ROOT_URL directly from endpoints. The env should be loaded at startup and stored in a config state, which is used later by the app. This way, it makes it easier to test it.

Are you okay if I only add a comment with my suggestion and add a good first issue to fix it later?

editoast/fga/src/lib.rs
flomonster: Having a block section here makes the borrow checker fail. I will try another way to fix it, keeping the `use`. 
The main issue here is using recursive use of the macro.

leovalais: It's not serious enough to block the 2024 edition, but keeping the macro hygienic is important. If the recursion is the problem, you can totally inline the output of the branch `$ty:type : $id:ident`

editoast/src/valkey_utils.rs
leovalais: Not that I'm aware of, but you can have a `rustfmt.toml` somewhere with `imports_granularity = "item"` and `cargo +nightly fmt` if I'm not mistaken

https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=granula#imports_granularity

leovalais:
Nit, not tested: shouldn't 2nd and 3rd commit be merged?

Ping me if you have troubles with the macro, there are patterns that are here for convenience which can be removed if there's an issue with them

editoast/src/valkey_utils.rs
woshilapin: Note, we did not enforce that in the CI, because it would also impose this on all developers and that functionality depends on `nightly` so it felt too much to impose on everyone.

editoast/src/views/layers.rs
woshilapin: Yes, as stated in my last sentence.

> It seems that we already had the problem anyway so I'm fine with a comment justifying that it's the only test like this.

editoast/fga/src/lib.rs
woshilapin: Or inline the trait with something like this, I believe this should work too (not tested)

```rust
$crate::model::Relation::userset($ty::$relation(), &fga!($ty:$id))
```

It's actually probably better since it does not import a trait for the rest of the file, the use of the trait would be completely scoped here.

flomonster: Fixed

Issue #11508, opened 2025-04-18
`editoast-stateful` scalability


Khoyo:
moved from https://github.com/osrd-project/osrd-confidential/issues/906

## Description

Allow scaling the `editoast-stateful` pod in order to improve the service availability. It currently isn't since it isn't possible to distribute the InfraCache. 

> [!IMPORTANT]
> The goal is to remove this SPOF. If the pod gets moved by carpenter we might incur significant downtime.

## Acceptance criteria

- `editoast-stateful` can scale
- the cache is always up to date, on all pods.
- if a pod dies, it can get back up quickly.

### Editoast

- Nice to have (good first issue): #11457
- Migrate `infra.version` from string to integer. #11510 
    - Improve core to avoid loading infra when the requested infra version is lower than the current one.
- Create a wrapper in `valkey_utils` to handle `zadd` and `zrangebyscore` #12385 
- Adapt the edition endpoint to store cache operation in valkey using `zadd` operation. #12432 
    - score -> the newest `infra_version`  
    - key -> `infra_patches.{OSRD_VERSION}.{infra_id}`
    - value -> List of `CacheOperation` **IMPORTANT** we don't store the json patch but the CacheOperation itself.
- Adapt `InfraCache` to store the `infra_version` of the current infra #12432 
- Adapt `InfraCache::get_or_load()` #12595 
    - Should check if the infra_version matches the current cached one
    - If not we call `ZRANGEBYSCORE` from `cached_version+1` to `infra_version`
    - If we retrieve the expected number of patches, we can apply them to the current InfraCache
    - Otherwise trash and reload from psql

> [!IMPORTANT]
> We should call `InfraCache::get_or_load` only once per endpoint.

### Ops

- Move editoast-stateful in `StatefulSet`
- Change the number of replicas to `2`
- Nice to have: Filter forwarded request precisely in gateway: #11304

PR #11511, opened 2025-04-18
core, editoast: migrate infra versions from string to integers

### LINKED ISSUE #11510: Convert `infra.version` from strings to integers
The infra version strings are in practice always integers. We should recognize that in their datatypes.

It also allows core to bypass downloading a requested older version if a new one is already in cache.

Part of [#11508](https://github.com/OpenRailAssociation/osrd/issues/11508)

Khoyo:
Closes #11510.


woshilapin (commented):
Was this decision discussed? From the PR and the ticket, you seem to have taken this decision alone. Anyway, seems ok, but I'm not entirely sure that I agree at this point: what if we, at some point, need some sort of more structured versioning, with some notion of bug fix and breaking changes? A string is more flexible (although less typed, sure).

Khoyo:
Sorry if the linking was bad, this is part of https://github.com/OpenRailAssociation/osrd/issues/11508 implementation plan (go back to the confidential ticket for the various meeting notes)

As far as infra version are concerned, we either need real versioning (with notion of parents and conflicts), or we need an ever increasing number. Bug fixes and breaking changes just don't make sense for infras, in the same way they don't make sense for a png.

woshilapin (approved):

flomonster (approved):
LGTM

Erashin (approved):

SharglutDev (approved):
Lgtm (not tested)

leovalais (approved):

core/src/main/java/fr/sncf/osrd/api/InfraManager.java
```diff
@@ -172,7 +174,8 @@ public FullInfra load(String infraId, String expectedVersion, DiagnosticRecorder
                 // try downloading the infra again if:
                 //  - the existing cache entry hasn't reached a stable state
                 //  - we don't have the right version
-                var obsoleteVersion = expectedVersion != null && !expectedVersion.equals(cacheEntry.version);
+                var obsoleteVersion = expectedVersion != null
+                        && (cacheEntry.version == null || expectedVersion.compareTo(cacheEntry.version) > 0);
```
bougue-pe: nit: can't you make it more readable with:
```suggestion
                        && (cacheEntry.version == null || expectedVersion > cacheEntry.version);
```

bougue-pe (approved):
LGTM, reviewed only core part.

bougue-pe (approved):

PR #11512, opened 2025-04-20
front: fix psl panels speed rounding in infra editor

### LINKED ISSUE #6547: Wrong announcement value when creating an LPV
### What happened?

When creating un LPV with 30km/h value the proposed announcement panel shows 35km/h

The proposed announcement panel should show 30 km/h

1. Go to the speed editor
2. Create an LPV at 30 km/h
3. Add the announcement panel

- **Browser**: Chrome

- **Env**: Recette

1482e8c

Synar:
Close #6547 

This is the simplest fix so that entering a speed that is already a multiple of 5 will give us the same value at the end. 

However if ceiling rather than rounding is an intended behavior we want to keep, we could either : 
- substract a small numerical value (-0.001 should dominate all float errors) before ceiling
- ensure we pass the actual value in km/h to the sign component without converting to ms first

### DIFF
--- a/front/src/applications/editor/tools/rangeEdition/speedSection/EditPSLSection.tsx
+++ b/front/src/applications/editor/tools/rangeEdition/speedSection/EditPSLSection.tsx
@@ -27,15 +27,14 @@ const getNewAnnouncementSign = (
   speedLimit: number
 ) => {
   const firstRange = trackRanges[0];
-  const speedInKmH = msToKmh(speedLimit);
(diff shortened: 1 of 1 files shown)


Synar:
@bgiuliana Can you confirm the intended behavior? ^^

Akctarus (approved):
tested, lgtm !

Synar:
Discussed this with Giuliana and in fact we want to round to the nearest km, change coming soon ^^

emersion (approved):
LGTM!

PR #11513, opened 2025-04-20
front: fix disabled speed tag select in infra editor

### LINKED ISSUE #7265: Composition code selection is not possible in the infrastructure editor if the "speed limit" switch is not active
### What happened?

I try to choose a composition code in the infrastructure editor (in the configuration of the elements displayed on the map), when the "speed limit" switch is not active, but it is not possible. It occurs only in the infratructure editor : in the map viewer for example, I can choose a composition code regardless of the switches (for the Permanent speed limits for example, which is the expected behaviour). 

I expected to be able to choose a composition code, regardless of the active switches, like in the map viewer.

1. Open the infrastructure editor on any infra
2. Click on "Choose the types of elements to display" button on the right
3. Make sure the "speed limit" switch is not active
4. Try to change the composition code -> it is impossible (to compare, one can display the speed limits and observe that now, one can change the composition code)

- **Browser**: Chrome
- **OS**: W11
- **Env**: Staging internal

1482e8c

Synar:
Close #7265

Alternatively, we could disable the select if both speed limits and permanent speed limits are not selected (and make it clearer through styling). Still, the issue implies we should remove this disable all together.

### DIFF
--- a/front/src/applications/editor/components/LayersModal.tsx
+++ b/front/src/applications/editor/components/LayersModal.tsx
@@ -194,7 +194,7 @@ const LayersModal = ({ initialLayers, selection, frozenLayers, onChange }: Layer
             id="speedLimitTag"
             className="form-control"
             value={layersSettings.speedlimittag || DEFAULT_SPEED_LIMIT_TAG}
-            disabled={!isArray(allSpeedLimitTags) || !selectedLayers.has('speed_sections')}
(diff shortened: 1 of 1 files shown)


SharglutDev (approved):
Lgtm and tested ✅

RomainValls (approved):
LGTM, tested 👍

PR #11539, opened 2025-04-22
core: reproduce-request: exclude infra loading from time measurement

eckter:
Load the infra *before* running the request itself

### DIFF
--- a/core/src/main/java/fr/sncf/osrd/cli/ReproduceRequest.kt
+++ b/core/src/main/java/fr/sncf/osrd/cli/ReproduceRequest.kt
@@ -70,8 +70,12 @@ class ReproduceRequest : CliCommand {
         try {
             val httpClient = OkHttpClient.Builder().readTimeout(120, TimeUnit.SECONDS).build()
             val infraManager =
-                if (railjson != null) FileInfraProvider(railjson!!)
-                else InfraManager(editoastUrl, editoastAuthorization, httpClient)
+                if (railjson != null) {
+                    val rjs = parseRailJSONFromFile(railjson)
+                    val signalingSimulator = makeSignalingSimulator()
+                    val infra = FullInfra.fromRJSInfra(rjs, signalingSimulator)
+                    FileInfraProvider(infra)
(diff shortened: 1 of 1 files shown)


core/src/main/java/fr/sncf/osrd/cli/ReproduceRequest.kt
```diff
@@ -70,8 +70,12 @@ class ReproduceRequest : CliCommand {
...
             val httpClient = OkHttpClient.Builder().readTimeout(120, TimeUnit.SECONDS).build()
             val infraManager =
-                if (railjson != null) FileInfraProvider(railjson!!)
-                else InfraManager(editoastUrl, editoastAuthorization, httpClient)
+                if (railjson != null) {
+                    val rjs = parseRailJSONFromFile(railjson)
+                    val signalingSimulator = makeSignalingSimulator()
+                    val infra = FullInfra.fromRJSInfra(rjs, signalingSimulator)
+                    FileInfraProvider(infra)
+                } else InfraManager(editoastUrl, editoastAuthorization, httpClient)
```
bougue-pe, line 78: For the editoast provider, the loading is still part of the timing, no?

The only idea I have on this is to pre-load request to extract infra info, then pre-load infra before actual timed request. Not sure it's worthy.

bougue-pe (approved):
It's already worthy as-is 👍

Erashin (approved):

core/src/main/java/fr/sncf/osrd/cli/ReproduceRequest.kt
eckter: > For the editoast provider, the loading is still part of the timing, no?

True

> The only idea I have on this is to pre-load request to extract infra info, then pre-load infra before actual timed request. Not sure it's worthy.

It's tricky: the infra ID is part of the request itself, which we can't really parse at this point.

bougue-pe: Thanks for the answer, I resolve this as it's not so important currently.

PR #11541, opened 2025-04-22
front: add translation for unknown step in simulation tables

Synar:
'Unknown' was reported raw in all languages in the report sheet for steps without a name, I believe it should be translated instead


front/public/locales/en/stdcm-simulation-report-sheet.json
```diff
@@ -32,6 +32,7 @@
   "stopType": "motif",
   "towedMaterial": "towed material",
   "track": "track",
+  "unknown": "Aucun",
```
emersion: This should probably be "Unknown" still :P

emersion (approved):
LGTM apart from the one comment

front/public/locales/en/stdcm-simulation-report-sheet.json
Synar: My bad haha

SarahBellaha (approved):
Lgtm !

PR #11544, opened 2025-04-22
core: pathfinding: add some telemetry trace spans

eckter:
I haven't *really* tested, it's difficult to test datadog related code locally (I don't know if it's even possible). But it shouldn't work differently from the spans in stdcm, it's pretty straightforward.

### DIFF
--- a/core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
+++ b/core/src/main/kotlin/fr/sncf/osrd/api/pathfinding/PathfindingBlocksEndpoint.kt
@@ -26,6 +26,8 @@ import fr.sncf.osrd.utils.indexing.*
 import fr.sncf.osrd.utils.units.Length
 import fr.sncf.osrd.utils.units.Offset
 import fr.sncf.osrd.utils.units.meters
+import io.opentelemetry.api.trace.SpanKind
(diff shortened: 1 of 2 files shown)


Erashin (approved):

woshilapin:
You can test locally with `jaeger` (should be started with the docker compose stack).

From what I recall, we used to need `@WithSpan` for OpenTelemetry and `@Trace` for Datadog. But now, we're able to send OpenTelemetry to an endpoint (the Datadog agent) who understand OpenTelemetry.

eckter:
> You can test locally with `jaeger` (should be started with the docker compose stack).
> 
> From what I recall, we used to need `@WithSpan` for OpenTelemetry and `@Trace` for Datadog. But now, we're able to send OpenTelemetry to an endpoint (the Datadog agent) who understand OpenTelemetry.

Oh, I tried to check the local jaeger and I did find the "global" span, but it was missing all the subdivisions. Same for the stdcm traces actually, despite working properly on datadog.  I kind of assumed these spans weren't processed properly by jaeger.

I'm probably using jaeger wrong. I'll try to make it work, but maybe I'll ask for help at some point, if you'd be available for a quick teams call

eckter:
After looking a little into it, it *seems* like editoast is sending the trace, but not core. 

I'll look further into this later on, for now I still feel fairly confident in adding these few spans.

PR #11545, opened 2025-04-22
Add `ViewError` derive macro

leovalais:
closes https://github.com/osrd-project/osrd-confidential/issues/1490
design doc: https://osrd.fr/en/docs/reference/design-docs/editoast-errors/

> [!NOTE]
> This PR does not cover the whole design doc, but I feel like this is a good stopping point for now. We can enhance the macro later as we need it.

# What's in here

* `views::error` defines `trait ViewError` that all errors implement thanks to the macro. A type implementing `ViewError` can be automatically converted into a `struct EditoastError` that has the error format we want and is an `axum` response as well. (Not an `utoipa` schema though!)
* `views::error` also tests most of the features of the macro by comparing the error being produced.
* `trait ViewError` can also produce an `utoipa::openapi::Responses` built from the metadata the macro generates as an `OpenApiResponse` (custom type)

---

* `derive(ViewError)` generates the implementation of `trait ViewError` according to the macro parameters provided
* `derive(ViewError)` also generates an implementation for `axum::IntoResponses` so that we can use the error type in the `Err` position of the `Result` in handlers return type (see documents.rs)
* `derive(ViewError)` also generates the implementation of `utoipa::IntoResponses` so that the error type can be used in the `responses()` section of `utoipa::path` macro so that the OpenAPI now **also tracks each endpoint's error cases**

---

* `editoast_derive::view_error` is architectured like so:
	* `view_error.rs` defines the macro entry point and the snapshot tests. Snapshot tests are not meant to be exhaustive (tests in `views::error` exist for that reason) but rather to have a few representatives of what the macro can generate. This helps with documentation, development and review.
	* `args.rs` defines the shape of the macro, its arguments and the attributes it reads using `darling`
	* `Args::parse` is where the macro logic lies. It turns a set of arguments into a `Codegen` type implementing `ToTokens` that has all the necessary information conveniently organized in order to generate the code.
	* `codegen.rs` low-level representation of the parameters of the expected generated code

---

* In `views::documents` we get rid of the former `derive(EditoastError)` type and split it into several `derive(ViewError)` types so that no endpoint wrongfully declare an error response that cannot happen.
* Note that `derive(ViewError)` is compatible with `derive(EditoastError)` if needed, but it should be avoided. `derive(ViewError)` is meant for **small, specialized** errors, not for a huge enum packing a bunch of them regardless of whether they can be returned or not.
* `views::documents::DatabaseError` is there for now since it's the first usage, but it'll likely go up as soon as it's shared by other endpoints.

---

> [!WARNING]
> Frontend tooling has **not** been updated so error translation will be inconsistent while we migrate. This was accepted at the time we wrote the document to avoid having to maintain hybrid tooling. "Bryan is in the kitchen" it is.

# Review

**All commits will be squashed.** I thought keeping the development log would be useful to watch the macro get incrementally more complex instead of dropping a single, albeit atomic, commit.

> [!TIP]
> I'm available to peer review anytime. 

The last commit is a proposition, not 100% sure about this.

# Next steps

* Support error types with generics. It enables a pattern for error composition we might want to adopt. PR incoming "soon".
* Incrementally convert endpoints to this new error paradigm.
* Provide a convenient way to unit-test responses with an `EditoastError` body, ideally by trying to convert it to its original `impl ViewError` type (with limitations).

# Comparison with the design doc

Thanks chatgpt 😇 (except for `REVIEW` annotations)

## Missing from the current implementation

These are whole planned capabilities with no current equivalent:

- Selective context fields and renaming:
  `context(reason, recovery_id = "recovery")` is unsupported. Context remains all eligible fields or none.
- Context provider functions:
  `context_with = context_provider` is unsupported.
- Optional `source_location` metadata pointing to the defining file and line.
- The proposed generic, JSON-preserving wrapper for Core errors. (REVIEW: we'll do that in views, it's not related to derive(ViewError) lol)
- OpenAPI post-processing that verifies errors sharing a label have compatible statuses and schemas. Existing OpenAPI title validation is unrelated to error-label consistency. (REVIEW: we may do that when we update the fontend openapi-sync)

## Implemented

These planned features are implemented fully or nearly fully:

- A distinct view-layer `ViewError` trait based on `std::error::Error`.
- Delayed conversion into the standard HTTP payload: typed errors survive until response serialization.
- `derive(ViewError)` integration with `thiserror`.
- Generated `ViewError`, `utoipa::IntoResponses`, and `axum::IntoResponse` implementations.
- Support for enums, named structs, tuple structs, and newtypes.
- Empty context by default and complete-field context through `#[view_error(context)]`.
- Context generation immediately before response serialization.
- Source exclusion for:
  - field-level `#[source]`;
  - field-level `#[from]`;
  - fields named `source`;
  - `#[error(transparent)]`.

  This is almost complete: `backtrace` fields are not recognized specially, and source fields cannot be explicitly opted back into context. (REVEIW: I think that's fine)
- Default status 500 and status selection through `StatusCode` associated constants such as `NOT_FOUND`.
  
  The planned numeric form `status = 401` and `user`/`internal` shorthands remain unsupported, but equivalent status configuration exists. (REVIEW: that's intentional)
- Configurable type-level labels with `name = "…"`.
- Configurable variant-level labels with `name = "…"`.
- Reusable view-error structs and field-level `#[view_error]` substitution.
- Forwarding of the embedded error’s label, sub-label, status, context, and OpenAPI responses.
- Endpoint-specific OpenAPI responses, demonstrated by `DocumentError` in [documents.rs](osrd/editoast/src/views/documents.rs:52).
- OpenAPI descriptions for labels, statuses, context schemas, and thiserror message templates.

There are still several deviations within otherwise implemented features:

- The document specifies labels shaped like `editoast:ErrorType::Variant`; defaults now produce normalized labels such as `editoast::database::internal`.
- `#[view_error]` does not itself substitute the runtime message. The wrapper’s `Display` output is used, so complete substitution requires `#[error(transparent)]` or another message that delegates to the embedded error. (REVIEW: that's intentional — more flexibility and a thiserror annotation is required anyway)

## Not in the plan but implemented

- Default label normalization:
  Rust identifiers are converted to `snake_case`, then a trailing `_error` or `_errors` is removed. For example, `DatabaseError::Internal` becomes `database::internal`. Explicit `name` values are preserved verbatim.
- Explicit labels may contain precomposed namespaces, such as `name = "document::not_found"`.
- Unit structs are supported.
- `#[view_error]` may identify any field; it does not need to be a thiserror source or carry `#[from]`.
- A forwarded error may coexist with additional fields. Those fields are ignored during substitution. (REVIEW: for when we need context to match on the error, but we don't want to include it in the response)
- Compatibility awareness for `derive_more::From`:
  its conversion annotations do not become thiserror sources. (REVEIW: *very* important)
- Dedicated diagnostics reject unsupported `derive_more` opt-out forms `#[from(skip)]` and `#[from(ignore)]`.
- Field-level `#[view_error(...)]` arguments receive a dedicated error rather than being silently accepted.
- Labels are instance methods, enabling forwarding through enum variants instead of being fixed associated constants. (REVIEW: oopsie in the design doc)
- Forwarded OpenAPI declarations are composed by extending the wrapper response list with the embedded `ViewError` response lists.


woshilapin:
Should I take a look at this PR?

leovalais:
It didn't move since our call where you answered my questions. I'll get back to it after the permission work is "done". I'll ping you when it's ready to review since you approved the general direction :)

leovalais:
`jj absorb` a few revisions

editoast/src/views/error.rs
```diff
@@ -0,0 +1,689 @@
...
+use utoipa::openapi::ResponsesBuilder;
+use utoipa::openapi::Schema;
+use utoipa::openapi::SchemaFormat;
+use utoipa::openapi::Type;
+use utoipa::openapi::schema::SchemaType;
+
+/// Describes a view error and provides the necessary information to build the API response
+///
+/// Not meant to be implemented directly. Use the `derive(ViewError)` proc-macro instead.
+pub(crate) trait ViewError: std::error::Error + utoipa::IntoResponses + Sized {
```
Khoyo: why the `pub(crate)` instead of a pub?

editoast/src/views/error.rs
```diff
@@ -0,0 +1,689 @@
...
+
+#[derive(Debug, serde::Serialize)]
+#[cfg_attr(test, derive(PartialEq, Eq))]
+pub(crate) struct EditoastError {
+    pub(crate) r#type: String,
+    #[serde(with = "crate::error::StatusCodeRemoteDef")]
+    pub(crate) status: http::StatusCode,
+    pub(crate) message: String,
+    pub(crate) context: HashMap<String, serde_json::Value>,
+}
```
Khoyo: ditto

editoast/src/views/error.rs
```diff
@@ -0,0 +1,689 @@
...
+    pub(crate) sub_label: Option<&'static str>,
+    pub(crate) status: http::StatusCode,
+    pub(crate) message_template: Option<&'static str>,
+    pub(crate) context: Vec<ContextEntry>,
+}
+
+pub(crate) struct ContextEntry {
+    pub(crate) key: &'static str,
+    pub(crate) schema: RefOr<Schema>,
+}
```
Khoyo: ditto

editoast/src/views/documents.rs
```diff
@@ -7,26 +7,44 @@ use axum::http::StatusCode;
...
-    NotFound { document_key: i64 },
+#[derive(Debug, thiserror::Error, ViewError)]
+pub(in crate::views) enum DatabaseError {
+    #[error("database error: {0}")]
+    #[view_error(status = INTERNAL_SERVER_ERROR)]
+    Internal(#[from] editoast_models::Error),
+    #[error("database unavailable: {0}")]
+    #[view_error(status = SERVICE_UNAVAILABLE)]
+    Unavailable(#[from] database::DatabasePoolError),
+}
```
Khoyo, line 28: Maybe the `DatabaseError` should live somewhere more common, eg. `views::error`. I'd expect almost all endpoints to need it.

editoast/editoast_derive/src/lib.rs
```diff
@@ -283,6 +284,14 @@ pub fn route(attr: TokenStream, input: TokenStream) -> TokenStream {
         .into()
 }
 
+#[proc_macro_derive(ViewError, attributes(view_error))]
```
Khoyo, line 287: We need some documentation here.

editoast/src/views/documents.rs
leovalais: views::documents::DatabaseError is there for now since it's the first usage, but it'll likely go up as soon as it's shared by other endpoints.

Khoyo: Fair enough

editoast/src/views/error.rs
leovalais: It was to prevent non-views modules from defining ViewErrors. But [f8863ff](https://github.com/OpenRailAssociation/osrd/pull/11545/commits/f8863ff3f89b3a2ceb9964d3d3fe35bd3db4ff58) is what I meant to do...

leovalais:
Idea: Or<T1,...,Tn> where T: ViewError instead of OrDatabaseError<T> (not for this PR)

front/src/common/api/generatedEditoastApi.ts
aschantraine, line 1: This file is not supposed to change.

front/src/common/api/osrdRailwayManagerApi.ts
aschantraine, line 1: This file is not supposed to change.

editoast/openapi.yaml
```diff
@@ -431,6 +431,60 @@ paths:
...
+                      type: string
+                    status:
+                      type: integer
+                      format: int32
+                      enum:
+                      - '500'
+                    type:
+                      type: string
+                      enum:
+                      - editoast::database::internal
```
aschantraine, line 460: We use to have `:` instead of `::`. Is there a reason for this change ?
If not, I think we should keep `:`.

front/src/common/api/generatedEditoastApi.ts
leovalais: Oh, I guess my autoformatter has the wrong version 🤷 I'll revert that when I squash all commits in the end, let's keep the comment unresolved.

front/src/common/api/osrdRailwayManagerApi.ts
leovalais: https://github.com/OpenRailAssociation/osrd/pull/11545#discussion_r3917602474

editoast/openapi.yaml
leovalais: Not a technical one. This is a scoping character and the goal of `ViewError` is to have a code-first representation of the errors. I preferred to use `::` as it's the scoping operator in Rust. It feels less arbitrary and most importantly less error-prone. If we have to write such scopes manually (like I did with `DocumentKeyError`), it's much more obvious to me to use `::` in the middle of writing Rust code. I was bit by that when writing old `EditoastError`s manually a while ago and only noticed when I saw the CI being red 😢 

That being said I don't really care much, I can change it back to `:` if you still prefer.

Khoyo:
This is so much better than what we have right now.

Khoyo (approved):

leovalais:
@woshilapin I have the ship it squirrel letz go

editoast/src/views/error.rs
```diff
@@ -0,0 +1,691 @@
...
+use utoipa::openapi::schema::SchemaType;
+
+/// Describes a view error and provides the necessary information to build the API response
+///
+/// Not meant to be implemented directly. Use the `derive(ViewError)` proc-macro instead.
+pub(in crate::views) trait ViewError:
+    std::error::Error + utoipa::IntoResponses + Sized
+{
+    fn responses() -> Vec<OpenApiResponse>;
+    fn label(&self) -> &'static str;
```
woshilapin, line 25: I’m not entirely sure of the usecase for having both `label` and `sub_label` separated. Couldn’t we have `label` 
that directly returns `MyError::Variant`?

editoast/editoast_derive/src/snapshots/editoast_derive__view_error__tests__enums__mixed__forwarded_view_error.snap
woshilapin, line 1: It seems that you have 6 snapshots, but all are duplicated, so there is only 3 different snapshots (just by looking at the first 40 lines). Is it something intended from `insta` that I don’t understand, or it’s a mistake?

editoast/editoast_derive/src/view_error/codegen.rs
woshilapin, line 1: There is a bunch of code in `args.rs` and `codegen.rs`, and I believe there is a case for more tests probably? I’ve seen the few tests in `view_error.rs` which are mostly smoke tests, but don’t exercise a lot of the use cases of the macro (use of `#[thiserror::from]` vs `#[thiserror::source]`, or use of `#[derive_more::from]`, or giving a status like `NOT_A_VALID_HTTP_STATUS`, and probably a bunch of other situations). Maybe some of them can be exercised at code level, instead of going full-feature with `insta`? For example, you already have an `insta` test for a variant that would contain a field named `source`, so testing for `FieldArgs::is_source` for all the cases might enough instead of going full-`insta`.

woshilapin (approved):
All the comments are fine to address in future PR. The current PR can be merged as-is. The comments can be resolved and put in issues as follow-up work.

editoast/src/views/error.rs
leovalais: We could, but that would require more work on the macro-side (generating the concatenated string). Just a trade-off wdyt?

woshilapin: I’d say having both function make the API more complex that it need to be, for no real benefit. 'label' is just way to identify an error, the fact that it is identified by two parts or only one is irrelevant. But again, it’s neither a hard no, neither should it be blocking for merging, we can address that in a subsequent PR.

editoast/editoast_derive/src/snapshots/editoast_derive__view_error__tests__enums__mixed__forwarded_view_error.snap
leovalais: It was a mistake, I don't know what I did...

editoast/openapi.yaml
leovalais: Changed to `:`.

leovalais:
Merging this to allow @neigeantre and @QuentinHayK to work on errors. Follow-up tasks:

- [ ] Documentation
- [ ] Merge `ViewError::label` and `ViewError::sub_label`
- [ ] Unit-test `args.rs` where possible

editoast/editoast_derive/src/lib.rs
leovalais: https://github.com/OpenRailAssociation/osrd/pull/11545#issuecomment-5779946810

editoast/src/views/error.rs
leovalais: https://github.com/OpenRailAssociation/osrd/pull/11545#issuecomment-5779946810

editoast/editoast_derive/src/view_error/codegen.rs
leovalais: There may be a bunch to tests to add indeed. https://github.com/OpenRailAssociation/osrd/pull/11545#issuecomment-5779946810

> or giving a status like NOT_A_VALID_HTTP_STATUS

That won't work because the macro doesn't parse this text, only forwards it to `http::`.

PR #11546, opened 2025-04-22
editoast: return explicit type in `paced_train` endpoints response

hamz2a:
https://github.com/OpenRailAssociation/osrd/pull/11546#issuecomment-2821566699

### DIFF
--- a/editoast/src/views/paced_train.rs
+++ b/editoast/src/views/paced_train.rs
@@ -6,7 +6,7 @@ use axum::extract::Json;
 use axum::extract::Path;
 use axum::extract::Query;
 use axum::extract::State;
-use axum::response::IntoResponse;
+use axum::http::StatusCode;
 use editoast_authz::Role;
 use editoast_derive::EditoastError;
 use editoast_models::DbConnectionPoolV2;
@@ -118,7 +118,7 @@ async fn get_by_id(
     State(db_pool): State<DbConnectionPoolV2>,
     Extension(auth): AuthenticationExt,
     Path(PacedTrainIdParam { id: paced_train_id }): Path<PacedTrainIdParam>,
-) -> Result<impl IntoResponse> {
+) -> Result<Json<PacedTrainResponse>> {
     let authorized = auth
         .check_roles([Role::OperationalStudies, Role::Stdcm].into())
         .await
@@ -154,7 +154,7 @@ async fn update_paced_train(
     Extension(auth): AuthenticationExt,
     Path(PacedTrainIdParam { id: paced_train_id }): Path<PacedTrainIdParam>,
     Json(paced_train_base): Json<PacedTrain>,
-) -> Result<impl IntoResponse> {
+) -> Result<StatusCode> {
     let authorized = auth
         .check_roles([Role::OperationalStudies].into())
         .await
@@ -171,7 +171,7 @@ async fn update_paced_train(
         })
         .await?;
 
-    Ok(axum::http::StatusCode::NO_CONTENT)
+    Ok(StatusCode::NO_CONTENT)
 }
 
 /// Delete a paced train
@@ -189,7 +189,7 @@ async fn delete(
     Json(ListId {
         ids: paced_train_ids,
     }): Json<ListId>,
-) -> Result<impl IntoResponse> {
+) -> Result<StatusCode> {
     let authorized = auth
         .check_roles([Role::OperationalStudies].into())
         .await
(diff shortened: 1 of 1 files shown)


leovalais:
A bit of context about the change motivation would be nice here... Axum seems to favor the `impl IntoResponse` style https://docs.rs/axum/latest/axum/response/index.html, even though not always. I'm fine with the `Json<T>` extractor as it helps "go to definition" of the response more easily, but for the status codes I'd stick with the `impl IntoResponse`, *even more so* as this PR doesn't cover all editoast endpoint definitions where status codes are returned.

hamz2a:
Sorry for not adding more context in the PR description. I need to modify the endpoints for `paced_trains`, and I couldn’t find the return types in the handler signatures, which is the main info I was looking for to make the changes. So I thought it would be helpful to have them directly in the signature, rather than having to dig through the function itself.

leovalais:
I'd suggest to stick to `axum`'s doc style as much as possible. Furthermore the changes of this PR are not consistent with all other endpoint signatures, which do stick to `axum`'s style. Should everyone agree on getting rid of `impl IntoResponse`, such change must apply to the whole API, not just paced trains.

hamz2a:
If I understand correctly, it's ok to change the endpoint that returns the JSON, not the others ?

leovalais:
Unless everyone agrees to do otherwise, I'd say yes.

woshilapin:
For what I can understand of the problem, `impl IntoResponse` is easy to write, harder to read and understand. And I usually prefer better readability since code is more often read than written (and also, figuring out the type when we write the code is usually easy).
Also, another reason why `impl IntoResponse` might not be such a great idea is for compilation errors. Consider this piece of code.

```rust
fn iter_maybe(with: bool) -> impl Iterator {
    if with {
        return Some(0).into_iter();
    }
    vec![0].into_iter()
}
```

All exit points of the function returns an `Iterator` and yet, it doesn't compile. And the error message may not be so obvious about what is going on (in the case of `axum`, it might be even worse).

Finally, the fact that `axum` is using `impl IntoResponse` in the documentation might just be a documenting reason, not a "best practice" reason: indeed, it documents well the important part of what a handler should return.

hamz2a:
@leovalais what do you think ?

leovalais:
> Also, another reason why impl IntoResponse might not be such a great idea is for compilation errors. Consider this piece of code.

Agreed, but considering that most of our endpoints returning a non-json response usually have a single `Ok` return expression, I'm not sure how much of a problem this is.

> For what I can understand of the problem, impl IntoResponse is easy to write, harder to read and understand. And I usually prefer better readability since code is more often read than written (and also, figuring out the type when we write the code is usually easy).

Agreed on the principle, but that doesn't match what we (and most of the community Rust code I've read for that matter) do with iterator for instance. When returning an iterator, we **always** prefer to write `impl Iterator`, even if the iterator type is easy to write: `std::iter::repeat(12usize)`. Why should we treat API responses differently?

Another issue with that is: where do we draw the line between explicit typing and `impl IntoResponse`? Here's the return expression of `GET /documents/:id`:

```rust
	Ok((
        StatusCode::OK,
        [
            (CONTENT_TYPE, doc.content_type),
            (CACHE_CONTROL, "public, max-age=3600".to_string()),
        ],
        doc.data,
    ))
```

That seems to fit the  `impl IntoResponse` use-case perfectly imo. So when do we accept to use it and when do we not? That introduces a degree of freedom that leads to inconsistencies in signatures, which *do degrades readability* (not the end of the world admittedly, but that's the topic of the discussion here 😄).

> Finally, the fact that axum is using impl IntoResponse in the documentation might just be a documenting reason, not a "best practice" reason: indeed, it documents well the important part of what a handler should return.

I disagree. First I'd be very surprised if they'd adopt a pattern in their documentation they didn't encourage. And by looking at their [`examples`](https://github.com/tokio-rs/axum/tree/main/examples) directory it seem they have the same convention we have: explicit `Json<T>`, `impl IntoResponse` otherwise, explicit typing for errors (because of the implicit `.into()` of the ? likely).

---

That being said, I'm not vehemently against changing our convention (even though it's indeed not my preference, for the reasons above), but in this case the change should cover our **whole API**.

woshilapin:
We had another round of discussion with @leovalais. I'm clearly not really fan of NOT using `impl IntoResponse` in the case of a complex return type. And also, it feels that the edge case you encounter @hamz2a (somewhat duplicating an endpoint and being impacted by not having the type explicit) might be pretty rare (usually, we will create a new endpoint, or modifying one, and rarely care about the exact type, apart from it being `impl IntoResponse`).

How do you feel about keeping `impl IntoResponse` everywhere, and paying the price on occasion of having a slightly worst experience of finding the exact type (I say slightly because, it's usually not that bad, going at the end of the function to figure out what is the type of the last `return`)?

